PR #60, opened 2017-08-17
Update Agent tests to new repos.

simitt:


tests/agent/nodejs/app.js
```diff
@@ -1,11 +1,8 @@
 'use strict'
 
-var opbeat = require('opbeat').start({
-  appName: 'nodejs-testapp',
-  secretToken: '123',
-  apiHost: 'localhost',
-  apiPort: '8080',
-  flushInterval: 1,
```
simitt: I only removed it because the current agent docu didn't have this property.. added it again.

PR #63, opened 2017-08-17
Fix agents code which is used for manual testing

ruflin:
Some of the code change and now the agent code has to be updated. Fix node.js and python agent.

### DIFF
--- a/tests/agent/nodejs/Makefile
+++ b/tests/agent/nodejs/Makefile
@@ -6,4 +6,4 @@ stop:
 	docker-compose down -v
 
 setup:
-	npm install elastic-apm --save
+	npm install elastic/apm-agent-nodejs
--- a/tests/agent/python/app.py
+++ b/tests/agent/python/app.py
@@ -7,16 +7,16 @@ from elasticapm.contrib.flask import ElasticAPM
 
 app = Flask(__name__)
 app.debug = True
-app.config['APM'] = {
+
+app.config['ELASTICAPM'] = {
     'SERVERS': [os.environ.get('apm-server', 'http://localhost:8080')],
     'DEBUG': True,
 }
 
 apm = ElasticAPM(
     app,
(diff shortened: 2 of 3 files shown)


tests/agent/python/app.py
```diff
@@ -14,8 +14,7 @@
 
 apm = ElasticAPM(
     app,
-    organization_id='123',
-    app_id='123',
+    app_name='test-app',
     secret_token='123',
```
watson: Similar to the Node agent, this really shouldn't be used any more right?

ruflin: I'm also making it empty. As it is convenient to have it in if you want to test the apm-server with a token.

tests/agent/nodejs/Makefile
```diff
@@ -6,4 +6,4 @@ stop:
 	docker-compose down -v
 
 setup:
-	npm install elastic-apm --save
+	npm install elastic/apm-agent-nodejs --save
```
watson: - Since there's no `package.json` in the `tests/agent/nodejs` folder, we should remove the `--save` option
- The node agent have now been deployed to npm as `elastic-apm`, so we should consider if we want this test to always take the latest stable version from npm (in which case it should be changed to `npm install elastic-apm`), or if we always want the latest HEAD from master in which case we can leave it as is

ruflin: For now I would keep the latest from master as this is for manual testing and normally what I want to do is testing the most recent changes of the node agent to see how they work with the apm-server.

watson: We should still remove the `--save` option though

ruflin: It should actually be removed. Github shows up here the "cached" diff.

ruflin:
jenkins, retest it

PR #66, opened 2017-08-17
SSL system tests

jalvz:
Also some refactor:

    -Show returned status code as assertion error message.
    -Create test classes so to reuse configuration for different scenarios.


tests/system/apmserver.py
```diff
@@ -14,17 +18,99 @@ def setUpClass(cls):
         cls.beat_path = "../../"
         super(BaseTest, cls).setUpClass()
 
+    def get_payload(self):
```
simitt: this is the `transaction` payload, I suggest to either rename the method to `get_transaction_payload` or pass in a string indicating which payload to return.

jalvz: done f201cb53c8a552d3567730996c707b4b4fbc1306 👍

PR #69, opened 2017-08-18
git-ignore temporary files produced by approval's testing.

jalvz:


jalvz:
Yeah, I don't have a preference, but i'm doing this to be consistent with how we do for SSL test certs which is a similar (albeit not same) situation, in which you'll find files on disk only if tests fails.

https://github.com/elastic/apm-server/commit/4d56ee9ceb148a8ebd063c5d72fe6e2ea446e06a

simitt:
The difference is that once the `approvals` tool is run, the `received` files are removed when they were approved by the user, so in a standard case there should be no leftover.

PR #73, opened 2017-08-18
Update Dashboards to Latest Version

simianhacker:
This PR updates the dashboards to the latest version. In this update the response time charts for TSVB use the new duration formatter instead of relying on a calculation. This also fixes a few broken links and changes the default index pattern to `apm-server-*`.


ruflin:
Thanks for the update. Will do some more manual tests as soon as https://github.com/elastic/apm-server/pull/63 is in.

PR #77, opened 2017-08-21
Update timestamp in JSON Spec for edge-case

roncohen:
Also brings back logging in case we missed other edge cases in the regular expression.

### DIFF
--- a/docs/spec/errors/error.json
+++ b/docs/spec/errors/error.json
@@ -98,7 +98,7 @@
         "timestamp": {
             "type": "string",
             "description": "Recorded time of the error, UTC based and formatted as YYYY-MM-DDTHH:mm:ss.sssZ",
-            "pattern": "^(\\d{4})-(\\d{2})-(\\d{2})T(\\d{2}):(\\d{2}):(\\d{2})(\\.\\d{1,6})?Z"
+            "pattern": "^(\\d{4})-(\\d{2})-(\\d{2})T(\\d{2}):(\\d{2}):(\\d{2})(\\.\\d{1,6})?Z$"
         }
     },
     "required": ["timestamp"],
--- a/docs/spec/transactions/transaction.json
+++ b/docs/spec/transactions/transaction.json
@@ -29,7 +29,7 @@
         "timestamp": {
             "type": "string",
             "description": "Recorded time of the transaction, UTC based and formatted as YYYY-MM-DDTHH:mm:ss.sssZ",
-            "pattern": "^(\\d{4})-(\\d{2})-(\\d{2})T(\\d{2}):(\\d{2}):(\\d{2})(\\.\\d{1,6})?Z"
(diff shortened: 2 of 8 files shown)


docs/spec/errors/error.json
```diff
@@ -98,7 +98,7 @@
         "timestamp": {
             "type": "string",
             "description": "Recorded time of the error, UTC based and formatted as YYYY-MM-DDTHH:mm:ss.sssZ",
-            "pattern": "^(\\d{4})-(\\d{2})-(\\d{2})T(\\d{2}):(\\d{2}):(\\d{2})(\\.\\d{1,6})?Z"
+            "pattern": "^(\\d{4})-(\\d{2})-(\\d{2})T(\\d{2}):(\\d{2}):(\\d{2})(\\.\\d{1,6})?Z$"
```
simitt: good find

ruflin: @simitt Is there something like a concept of "variable" in json schema so we could reuse the same pattern for all timestamps instead of copy / pasting it?

jalvz:
well spotted. if regexes are so hard to get right, why don't we validate with the actual parser that we actually use later on? then we are 100% sure it is correct.

for instance, what happens now if timestamp has month `13` or day `99`?

jalvz:
@roncohen @simitt 
Timestamp regex is still broken, day `99` passes but golang won't parse and returns default date:
https://github.com/elastic/apm-server/pull/87/files#diff-b5754f55ab03f8813d8f04b7cad34e7bR4

https://github.com/elastic/apm-server/pull/87/files#diff-5aa69e8e3f44bbabd43dc32a7cab0f6eR15

To start with, I think we should merge https://github.com/elastic/apm-server/pull/86 so we can properly test.

watson:
I wouldn't say that the regex is broken. Though a malicious client can sneak cleverly crafted dates past the validator, I think the spirit of the validation is to make sure that agents don't use the wrong format. A regex can never validate that the date is valid, only that the format is as expected.

But if we want to check that the date actually is valid before storing it in the database, it should be a separate step after the JSON Schema validation. But personally I'm ok with not reporting this back to the client, as this will most likely be someone sending bad data intentionally. So in that case we can do that check after ending the HTTP request.

jalvz:
What if an agent swaps day and month? They both are 2 digits, so format would be wrong but regex would accept it.
From a server perspective I don't care what is the cause of bad input. Bad input is bad input, and must be rejected.
If timestamp is not parseable for any reason, the stored date in ES is `0001-01-01 00:00:00 +0000 UTC`, which should not happen IMO.

roncohen:
As @watson said, we can't catch all edge cases with the regex.

@jalvz validating the timestamp by trying to actually parse it sounds like the way to go. Wanna give it a go? If that proves to be complex, I'm OK with dropping incoming transactions/errors with unparsable timestamps and writing out a good log message.

PR #78, opened 2017-08-21
Refactor server related code

jalvz:
This PR is motivated by a confusion that I faced adding basic TLS support.

+ rename variables to make more clear what is a beater abstraction and config.
+ merge beater and serve packages.
+ remove beater config and 'server' namespace in ypm-server.yml

Hopefully this makes it more clear where configuration comes from, etc.


beater/beater.go
```diff
@@ -3,37 +3,37 @@ package beater
...
+	server *server.ApmServer
+	config BeaterConfig
 	client beat.Client
 }
 
 // Creates beater
-func New(_ *beat.Beat, cfg *common.Config) (beat.Beater, error) {
-	config := config.DefaultConfig
-	if err := cfg.Unpack(&config); err != nil {
+func New(_ *beat.Beat, ucfg *common.Config) (beat.Beater, error) {
```
ruflin: `cfg` is kind of an unwritten convention for the `common.Config`. I think we don't use ucfg to not conflict with the package name.

beater/beater.go
```diff
@@ -3,37 +3,37 @@ package beater
...
 
-	"github.com/elastic/apm-server/config"
 	"github.com/elastic/apm-server/server"
 	"github.com/elastic/beats/libbeat/beat"
 	"github.com/elastic/beats/libbeat/common"
 	"github.com/elastic/beats/libbeat/logp"
 )
 
-type ApmServer struct {
+type ApmBeater struct {
```
ruflin: We could also call this just `Apm` or only `Beater`. I'm slightly in favor of `Apm`.

beater/config.go
```diff
@@ -1,15 +1,15 @@
...
 
-package config
+package beater
 
 import "github.com/elastic/beats/libbeat/common"
 
-type Config struct {
-	Host   string         `config:"host"`
-	Server *common.Config `config:"server"`
+type BeaterConfig struct {
```
ruflin: Now it's `beater.BeaterConfig`. +1 on moving the package but the name should stay just `Config` because then it becomes `beater.Config`.

beater/config.go
```diff
@@ -1,15 +1,15 @@
...
+	Host            string         `config:"host"`
+	ApmServerConfig *common.Config `config:"server"`
 }
 
-var DefaultConfig = Config{
+var DefaultBeaterConfig = BeaterConfig{
```
ruflin: Same here, I don't think Beater is needed in the name.

server/server.go
```diff
@@ -15,34 +15,22 @@ import (
...
-	"github.com/elastic/beats/libbeat/common"
 	"github.com/elastic/beats/libbeat/logp"
 )
 
 var debugHandler = flag.Bool("debugHandler", false, "Enable for having catchall endpoint enabled")
 
-type Server struct {
-	config Config
-	http   *http.Server
+type ApmServer struct {
```
ruflin: If we call this just `Apm` we will have `server.Apm` which would explain it better. Agree that `server.Server` is not nice.

server/server.go
```diff
@@ -15,34 +15,22 @@ import (
...
+	config     Config
+	httpServer *http.Server
```
ruflin: If we use `Apm` for the struct, we could also use here just `server`

server/server.go
```diff
@@ -67,35 +55,36 @@ func (s *Server) create(successCallback func([]beat.Event), host string) *http.S
...
-func (s *Server) Stop() error {
+func (apm *ApmServer) Stop() error {
 	c := context.TODO()
-	err := s.http.Shutdown(c)
+	err := apm.httpServer.Shutdown(c)
 	return err
 }
 
-func (s *Server) createHandler(p processor.Processor, successCallback func([]beat.Event)) func(w http.ResponseWriter, r *http.Request) {
+func createHandler(p processor.Processor, config Config, successCallback func([]beat.Event)) func(w http.ResponseWriter, r *http.Request) {
```
ruflin: I actually added this to server in the past exactly to be able to use the `config` and not have to pass it as param.

beater/beater.go
```diff
@@ -45,11 +45,14 @@ func (bt *ApmServer) Run(b *beat.Beat) error {
 		go bt.client.PublishAll(events)
 	}
 
-	bt.server, err = server.New(bt.config.Server)
-	if err != nil {
+	ucfg := bt.config.ApmServerConfig
+	apmServerConfig := server.DefaultConfig
+	if err := ucfg.Unpack(&apmServerConfig); err != nil {
```
ruflin: the common Config is used so that the Unpacking can happen in the package itself and the knowledge about the config must not be shared outside. Above you access `server.DefaultConfig` which could also be private in the package. This also prevents potential circular imports and coupling between the beater and the server package.

beater/beater.go
jalvz: I don't think there is a risk (also `http` is a package name and code was working with var named name `http`)
`cfg` / `config` / `conf` is too generic,  this way I know what type I have. `cfg` and `config` was a big part of my confusion before.

beater/config.go
jalvz: I was very confused by the several layers of configs we have. I know this is not as short as usual Golang idioms, but then I clearly know what kind of configuration is this.
This all PR in general comes out of experience trying to understand the code :)

server/server.go
jalvz: hmm, what are the benefits of that? `createHandler` still needs 3 things either case.
this way:
+ no accidental mutation can happen.
+ more explicit signature. the fact is that not the whole server is needed, just the config.
+ decouples server from handler, this now can be tested in isolation without having to create a server (which would be nice, btw).

beater/config.go
jalvz: it is not, because it is not used outside the package. ill make it private 👍

beater/beater.go
jalvz: Hmm, I don't think so. Based on what is that?

I think there is no more coupling between beater and server than before: 
+ Beater already knew about server (which is ok): https://github.com/elastic/apm-server/blob/master/beater/apm-server.go#L48
+ Server doesn't know about Beater.

Besides, knowledge about the config is *not shared outside. The thing is just unpacked, not inspected. We have the same pattern in other parts of the code base.

Just this snippet actually allows me to remove 2 dependencies:

https://github.com/elastic/apm-server/pull/78/files#diff-91bbeda7eb98a7adc57b9e47e2cf5c2bL18

https://github.com/elastic/apm-server/pull/78/files#diff-110fdf0113b0dcfcdca7c9352c1b4c4fL6

So circular imports are actually less likely, mostly because the beater config now is where it belongs.

`DefaultConfig` is turned public, yes, just like `DefaultBeaterConfig` which was already public before, I don't see how it is different. In turn, now I can make both `DefaultBeaterConfig` and `BeaterConfig` private, so encapsulation game is stronger `2 - 1` :D

Does this make sense or is there anything here that you don't think is true?

ruflin:
Your PR made me rethink our current configs because there seems to be something wrong. Looking at our YAML file I think it becomes obvious:

```
apm-server:
  host: "localhost:8080"

  server.max_unzipped_size:
  server.max_header_bytes:
  server.read_timeout: 2s
  server.write_timeout: 2s
  server.secret_token:
```

Two things here:

* Why is there a config inside a server?
* Why is host outside the server but when looking at the code, it's only used inside the server.

To solve the above I think there are two solutions. Either put `host` under `server` or remove the `server` namespace all together as the `beater` config itself is the server. 

As I don't like repetition I'm more leaning in the direction of removing the server and bring all on the same level as `host`. On the implementation this has the affect, that host is not treated separately anymore but part of one config option. This config object could either be directly in the `beater` which would probably mean to get rid of the server package and move `server` into the `beater` package. Alternatively we could use an `inline` config based on `common.Config` and then the unpacking only happens in the server.

Haven't played around with the code yep but merging the server and beater package would probably make the code the cleanest for now. I extracted the `server` package in the past because I felt it doesn't belong in the `beater` as all I personally think should be in the beater package are the implementations of the interface methods. But also the `beats` are not fully aligned there.

This does not answer all your comments above, but I thought perhaps we should first align on a high level before we go into the small details.

jalvz:
I think it was a good decision to separate `beater` and `server` packages separated, and is probably ok to keep them that way. Moving `host` config to `server` namespace makes a lot of sense. I can deff see us adding non-server configs related.

jalvz:
@ruflin thanks for the insights 👍 
removed server namespace and merged server and beater packages. please have a look when you have time

_meta/beat.yml
```diff
@@ -4,13 +4,13 @@
 
 apm-server:
   # Defines the host and port the server is listening on
-  host: "localhost:8080"
 
-  #server.max_unzipped_size:
-  #server.max_header_bytes:
-  #server.read_timeout: 2s
-  #server.write_timeout: 2s
+  #host: "localhost:8080"
```
ruflin: @simitt @roncohen Could we please have your thoughts on this change config change? We will keep it like this for probably quite a while.

beater/beater.go
ruflin: Agree there should not be a conflict. I'm just very used to the internal `beats` conventions but `ucfg` also works for me.

beater/beater.go
```diff
@@ -45,19 +43,17 @@ func (bt *ApmServer) Run(b *beat.Beat) error {
 		go bt.client.PublishAll(events)
 	}
 
-	bt.server, err = server.New(bt.config.Server)
-	if err != nil {
-		return err
-	}
-	bt.server.Start(callback, bt.config.Host)
-	defer bt.server.Stop()
+	server := apmServer(bt.config, callback)
```
ruflin: I definitively prefer the old way. Using `New` to create an object is a common way to do it.

beater/beater.go
```diff
@@ -45,19 +43,17 @@ func (bt *ApmServer) Run(b *beat.Beat) error {
...
+	start(server, bt.config.SSL)
```
ruflin: Why did you change this? You pass a reference to start and stop so you could directly use the object itself. It seems to me `start` and `stop` methods on the server itself are very natural. 

Also in case you use the object, you don't have to pass config.SSL as the config is already passed the constructor above.

In general I'm in favor of using objects where we can as it leads to less params passed to methods and works very well in the idea as I directly see which methods are available on an object.

But as this is also more a difference on the way the two of us code, I would like to get the input from @simitt and @roncohen on their preference as otherwise we go back and forth on this one.

tests/system/test_base.py
```diff
@@ -1,9 +0,0 @@
-from apmserver import ServerBaseTest
-import requests
-
-
-class Test(ServerBaseTest):
```
ruflin: Why did you remove this one?

jalvz: found it is duplicated

beater/server_test.go
```diff
@@ -1,4 +1,4 @@
-package server
+package beater
```
jalvz: I say it was probably ok to keep them separated, but thought you preferred to merge them. I don't have a strong preference.

beater/beater.go
jalvz: > It seems to me start and stop methods on the server itself are very natural.

I agree, but there isn't such server anymore, other than the builtin. I'm simply removing as much unneeded code as I can.

There is only 1 config, there is already 1 struct (`beater`) holding that config and passing it around, it doesn't make sense to have yet another struct (`server`) holding the same config to pass it around again.

Before it could make sense, because there was 2 configs, and there was 1 struct for each (among other stuff). It was also a source of confusion as I was reading things like `server.config`, `config.server`, `config.config` etc. 
Now the only `server` is golang builtin server, and the only config is the `beater`.

jalvz:
> I'm aware that the config changes cannot be completely decoupled from the refactoring as they also require some changes.

Yeah, it would be quite difficult to decouple that. What are you asking me is to do this again, in 2 different PRs. Honestly, I rather close this and let it go. I have more pushing things to do, this drained me a lot of time already and can't spend more time on it. At least for now.

ruflin:
@jalvz Lets do it then in 1 PR. Giving up on this one is not an option as this is a critical change we need :-) If you don't have the time for this one I can also take it over.

_meta/beat.yml
simitt: I am +1 on removing the `server` namespace, as it seems redundant within `apm-server`.

jalvz: Thanks Nicolas, everyone needs to be aware of this, indeed. 2 things:
+ removed `server` inner namespace (eg: `apm-server.secret_token` instead of `apm-server.server.secret_token`
+ commented out `host`, as it is defaulted already in the codebase.

jalvz:
@ruflin what else is needed other than changelog?

beater/server.go
```diff
@@ -68,34 +54,35 @@ func (s *Server) create(successCallback func([]beat.Event), host string) *http.S
...
 	c := context.TODO()
-	err := s.http.Shutdown(c)
+	err := server.Shutdown(c)
 	return err
 }
 
-func (s *Server) createHandler(p processor.Processor, successCallback func([]beat.Event)) func(w http.ResponseWriter, r *http.Request) {
+type handler func(w http.ResponseWriter, r *http.Request)
+
+func createHandler(p processor.Processor, config Config, publish successCallback) handler {
```
simitt: I completely agree with not having `createHandler` as a method of the `server`, but for `new`, `start` and `stop` I prefer the previous handling. This way only starting and stopping would be methods of the server, which seems totally legimit to me.

beater/beater.go
simitt: I see the advantage of how you refactored the config passing. For creating, starting and stopping I would be in favor of having this as methods of the server though. What do you think about:
`type apmServer struct { server *http.Server }`

and changing:
`func apmserver(config Config, publish successCallback) *http.Server` -> `func create(config Config, publish successCallback) apmServer`
`func start(server *http.Server, ssl *SSLConfig) `-> `func (*apmServer) start(ssl *SSLConfig) `
`func stop(server *http.Server) error`-> `func (*apmServer) stop() error`

simitt:
+1 on having this in 1 PR. I don't see a lot of gain but a lot of work in splitting this up.

beater/server.go
jalvz: but why should i create a type just to call a method on it which is exposed otherwise? just a placeholder for only 1 thing? 
it doesn't make any sense to me. what is the actual value that the type provides? 
If is just to type `a.b(c)` instead of `c(a, b)` ?

I don't mind either way, but then eventually we would end we things like `apmServer.server.something` which are annoyingly redundant, and then you have to distinguish between builtin server and placeholder server for builtin server?

simitt: I can also live with both ways (and none of both feels 100% right). Since I can't come up with a better suggestion right now, and the overall changes feel like a huge improvement, the current state is fine with me.

_meta/beat.yml
roncohen: Looks good to me 👍

beater/server.go
```diff
@@ -55,7 +41,7 @@ func (s *Server) create(successCallback func([]beat.Event), host string) *http.S
 
 	// TODO: Remove or find nicer way before going GA
 	if *debugHandler {
-		logp.Warn("Debug handler enabled")
+		logp.Warn("Debug handler Enabled")
```
roncohen: why the capital `E` here?

beater/server.go
ruflin: On the same page with Silvia that the overall change is more important. At some point we probably need to agree on a style on this otherwise we keep changing it forth and back, but this time is not now.

ruflin:
I merged it as I think it's more important that we have the config change then 100% agree on the coding style.

beater/server.go
ruflin: I think it should not matter, as this part will be removed anyways in a follow up PR.

beater/server.go
```diff
@@ -1,48 +1,34 @@
...
 	"crypto/subtle"
 	"encoding/json"
-	"flag"
 	"fmt"
 	"io"
-	"io/ioutil"
 	"net/http"
 	"strings"
 
+	"flag"
```
ruflin: @jalvz Could you clean up the imports here in the follow up PR https://github.com/elastic/apm-server/pull/85

beater/server.go
jalvz: lol, i think it was the IDE magic refactor button 🙈

beater/server.go
jalvz: Sure thing, but just to understand: what is the *real value of creating an ad-hoc type, so to write `a.b()` instead of `b(a)`, otherwise not needed?

It is an object wrapping another object that doesn't do anything else, or am i missing something?  It is because it "looks" good and "feels" more natural?

simitt: It just abstracts which methods belong to one struct. I am not a fan of having one struct with a huge amount of methods, since this makes it really hard to test, but in this case it would be two methods that obviously need this struct to make any sense. 
TBH I am generally a bit conflicted on this topic.. maybe I can find some good reading/examples and then come back to this discussion a bit later.

PR #81, opened 2017-08-22
fix env causing mkdir: invalid option

OmgImAlexis:
Before this change I was getting this error on OSX.

```
Step 5/10 : RUN mkdir -p $apm-server_PATH
 ---> Running in 503d035a8229
mkdir: invalid option -- 's'
Try 'mkdir --help' for more information.
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,12 +8,12 @@ RUN set -x && \
     apt-get clean
 
 # Setup work environment
-ENV apm-server_PATH /go/src/github.com/elastic/apm-server
+ENV APM_SERVER_PATH /go/src/github.com/elastic/apm-server
 
-RUN mkdir -p $apm-server_PATH
-WORKDIR $apm-server_PATH
+RUN mkdir -p $APM_SERVER_PATH
+WORKDIR $APM_SERVER_PATH
 
-COPY . $apm-server_PATH
+COPY . $APM_SERVER_PATH
 
 RUN make
 


ruflin:
@OmgImAlexis Thanks for the fix, LGTM. Could you sign the CLA? https://www.elastic.co/contributor-agreement

For this docker file keep in mind that this is not intended for production.

OmgImAlexis:
I can't sign the CLA as the page never loads. I've tried multiple browsers and with my VPN so I have a feeling it's a problem with the CLA server.

watson:
@OmgImAlexis I just tried here and it seems to be working. Is it the https://www.elastic.co/contributor-agreement page that doesn't load or the one after that?

OmgImAlexis:
The iframe from echosign. I also tried without any extensions installed incase it was my adblock but it still won't load.

watson:
@OmgImAlexis That's weird. I can confirm that it works for me. Could you try and load it on another device just to see if the issue is isolated to your computer?

OmgImAlexis:
This is as far as I get. I even tried on my phone on data. Does echosign use anything like Cloudflare? The sydney connection goes down quite often so that may be the issue.

screen shot 2017-08-22 at 7 48 08 pm

ruflin:
@karmi Perhaps you can comment on this? Did you see a similar issue in the past?

ruflin:
@OmgImAlexis Some of our co-workers in Sydney unfortunately confirmed what you reported :-( Will investigate what I can do. Really sorry about that.

ruflin:
@OmgImAlexis Could you try to change the setting in your browser to accept all cookies?

OmgImAlexis:
I've tried 4 different browsers in normal and private/incognito mode, tried 3 different laptops and my phone. Also tried a NYC VPN as well as my home and 4g data connection. Every single time the echosign form just hangs.

ruflin:
Hm, I'm running out of ideas. I tried it again myself and it takes a few seconds for the form to load, but finally it loads. Which checkbox did you use? Individual contributor?

I now copied the URL of the iframe to open it directly: https://secure.na1.echosign.com/public/esignWidget?wid=957HMA2E3B4XU9&hosted=false&token=&firstName=&lastName=&nameEditable=true Does this help?

OmgImAlexis:
That works perfectly yet the iframe still doesn't work at all.

What do I put for title?

ruflin:
@OmgImAlexis What about `Software Engineer` :-)

ruflin:
@OmgImAlexis Merged. Thanks for going through all this hoops. Still not 100% sure what went wrong TBH :-(

OmgImAlexis:
At least you know a way around the issue incase it happens to someone else.

PR #86, opened 2017-08-22
Include timestamp in approval tests

jalvz:
with the beat.Event introduction, we no longer put @timestamp in the generated documents, but we should still test timestamps


ruflin:
Good catch. I'm glad I'm not the `@timestamp` user. He must get tons of notifications from Elastic :-)

PR #88, opened 2017-08-22
Fix swallowed error in tests package

alrs:
json_schema_test.go was missing an assert on a defined error.


ruflin:
@alrs Thanks for the contribution.

PR #89, opened 2017-08-22
Unexport payload in error and transaction packages

### LINKED ISSUE #75: Shouldn't payload be a pointer
At this [line](https://github.com/elastic/apm-server/blob/master/processor/transaction/processor.go#L29),shouldn't the payload be a pointer.

Similarly at this [line](https://github.com/elastic/apm-server/blob/master/processor/validator.go#L26) ,I think the data should be a pointer as we are passing [p.payload](https://github.com/elastic/apm-server/blob/master/processor/transaction/processor.go#L38).

crackerplace:
Unexports payload in error and transaction packages.

@ruflin 

Fixes #75

### DIFF
--- a/processor/error/processor.go
+++ b/processor/error/processor.go
@@ -25,12 +25,12 @@ func NewProcessor() pr.Processor {
 }
 
 type processor struct {
-	payload pr.Payload
+	payload *payload
 	schema  *jsonschema.Schema
 }
 
 func (p *processor) Transform() []beat.Event {
-	return p.payload.Transform()
+	return p.payload.transform()
 }
 
 func (p *processor) Validate(reader io.Reader) error {
--- a/processor/error/processor.go
+++ b/processor/error/processor.go
@@ -20,12 +20,12 @@ const (
 func NewProcessor() pr.Processor {
 	schema := pr.CreateSchema(errorSchema, processorName)
 	return &processor{
-		&Payload{},
+		&payload{},
 		schema}
 }
 
 type processor struct {
-	payload *Payload
+	payload pr.Payload
 	schema  *jsonschema.Schema
 }
 
--- a/processor/error/event.go
+++ b/processor/error/event.go
@@ -47,7 +47,7 @@ func (e *Event) DocType() string {
 	return "error"
 }
 
-func (e *Event) Mappings(pa *Payload) (string, []m.DocMapping) {
+func (e *Event) Mappings(pa *payload) (string, []m.DocMapping) {
 	return e.Timestamp,
 		[]m.DocMapping{
 			{Key: "processor", Apply: func() common.MapStr {
--- a/processor/error/payload.go
+++ b/processor/error/payload.go
@@ -12,7 +12,7 @@ type payload struct {
 	Events []Event   `json:"errors"`
 }
 
-func (pa *payload) Transform() []beat.Event {
+func (pa *payload) transform() []beat.Event {
 	var events []beat.Event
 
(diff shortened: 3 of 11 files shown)


crackerplace:
@ruflin Due to this refactoring I was able to fix one more minor inconsistency.For error processor we were using [payload](https://github.com/elastic/apm-server/blob/master/processor/error/processor.go#L28) itself and not pr.Payload.Whereas for transaction processor it was pr.Payload.

ruflin:
@crackerplace Could you sign the CLA? https://www.elastic.co/contributor-agreement Please comment afterwards again so it will recheck the CLA.

ruflin:
jenkins, test it

crackerplace:
@ruflin Yes,I have signed the cla.

processor/error/processor.go
```diff
@@ -20,12 +20,12 @@ const (
...
 	schema := pr.CreateSchema(errorSchema, processorName)
 	return &processor{
-		&Payload{},
+		&payload{},
 		schema}
 }
 
 type processor struct {
-	payload *Payload
+	payload pr.Payload
```
ruflin: Why did you change this? I think it's good to depend on the local payload. We should change it to the same as we have in the `transaction` processor.

ruflin: Just saw your comment on this. +1 on fixing the inconsistency, but i would go with the private payload in both cases.

crackerplace: Yes,Changing this to the local payload looks better than the interface.Will change it.

crackerplace:
@ruflin what is the jenkins test command.Could not find in docs.

ruflin:
The jenkins command triggers a build on our jenkins slave: https://beats-ci.elastic.co/job/elastic+apm-server+pull-request/36/ It is not automatically done for community contributors so I have to trigger it "manually". 

Ping me when you pushed the change.

crackerplace:
@simitt I just wanted to take a step back and point out something,As per the code in apm-server master,though we are embedding the [Payload](https://github.com/elastic/apm-server/blob/master/processor/processor.go#L18) interface in Processor interface,our error or transaction [processor](https://github.com/elastic/apm-server/blob/master/processor/transaction/processor.go#L28) structs are not truly confirming to this Processor interface.Because we are not embedding [payload](https://github.com/elastic/apm-server/blob/master/processor/transaction/processor.go#L23) struct in processor struct(as name is used) and again implementing transform [method](https://github.com/elastic/apm-server/blob/master/processor/transaction/processor.go#L33-L35) which actually is not needed if we embed.

Now coming to your changes to make transform private,you have also removed the payload interface and that looks better as now the former inconsistency is also removed.This looks good to me and I have no issues including this in this PR.

thanks

simitt:
By making `transform()` a private method, the `Payload` interface was no longer fulfilled, so yes I had to adapt this in this step. I did not see a point in further defining it with a private method thus I completely removed it.
We could also merge your PR and I come up with the follow up PR but I think it makes more sense to have the changes in one commit.

processor/error/processor.go
ruflin: This still seems to use the global one?

ruflin:
@crackerplace I left you a last minor comment. I'm also ok with merging as is and we take over from here. What ever your preference is.

crackerplace:
@ruflin Yes,pushed the latest changes.

crackerplace:
@simitt thanks

ruflin:
jenkins, test it

PR #92, opened 2017-08-23
Update beats to bd6981d

ruflin:
Important Changes / Additions:
* Add support for cloud id. See https://github.com/elastic/beats/pull/4964 for more details. This could simplify the onboard for users that use apm-server with the Cloud in the future.
* Be able to overwrite index pattern in the index template. See https://github.com/elastic/beats/pull/4769 for more details. This allows to clean up our system tests.

### DIFF
--- a/apm-server.reference.yml
+++ b/apm-server.reference.yml
@@ -126,6 +126,19 @@ apm-server:
 #processors:
 #- add_docker_metadata: ~
 
+#============================= Elastic Cloud ==================================
+
+# These settings simplify using apm-server with the Elastic Cloud (https://cloud.elastic.co/).
+
+# The cloud.id setting overwrites the `output.elasticsearch.hosts` and
+# `setup.kibana.host` options.
+# You can find the `cloud.id` in the Elastic Cloud web UI.
+#cloud.id:
+
+# The cloud.auth setting overwrites the `output.elasticsearch.username` and
+# `output.elasticsearch.password` settings. The format is `<user>:<pass>`.
+#cloud.auth:
+
 #================================ Outputs ======================================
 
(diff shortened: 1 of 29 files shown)


apm-server.reference.yml
```diff
@@ -617,10 +631,14 @@ output.elasticsearch:
 # Set to false to disable template loading.
 #setup.template.enabled: true
 
-# Template name. By default the template name is apm-server.
-# The version of the beat will always be appended to the given name
-# so the final name is apm-server-%{[beat.version]}.
-#setup.template.name: "apm-server"
+# Template name. By default the template name is "apm-server-%{[beat.version]}"
```
simitt: should we keep havig. `beat.version` here or change to `apmserver.version`?

apm-server.reference.yml
```diff
@@ -617,10 +631,14 @@ output.elasticsearch:
...
+# The template name and pattern has to be set in case the elasticsearch index pattern is modified.
+#setup.template.name: "apm-server-%{[beat.version]}"
+
+# Template patttern. By default the template patter is "-%{[beat.version]}-*" to apply to the default index settings.
```
ruflin: That is a typo I need to fix on the beats side.

apm-server.reference.yml
ruflin: The fields used here must be part of the event. We have `beat.version` but we don't have `apm-server.version` in our events.

ruflin:
@simitt For updating _beats: Yes

Our use case for overwriting the index pattern name is for testing. On the beats side especially for logging people create their own templates. But in general it's only expert users that do this and I would not expect it to happen on the apm-server side at first.

PR #94, opened 2017-08-23
Make apm-server to fail fast

jalvz:
Current behaviour is e.g. that apm-server can happily start with a port already used. 
In such case a socket cannot be opened and the apm-server stays up without complaint, albeit not being able to accept requests.

This makes the apm-server to finish with an error when the underlying http server connection fails.

### DIFF
--- a/beater/beater.go
+++ b/beater/beater.go
@@ -3,14 +3,16 @@ package beater
 import (
 	"fmt"
 
+	"net/http"
+
 	"github.com/elastic/beats/libbeat/beat"
 	"github.com/elastic/beats/libbeat/common"
 	"github.com/elastic/beats/libbeat/logp"
 )
 
 type beater struct {
-	done   chan struct{}
 	config Config
+	server *http.Server
 	client beat.Client
 }
 
@@ -22,7 +24,6 @@ func New(_ *beat.Beat, ucfg *common.Config) (beat.Beater, error) {
 	}
 
 	bt := &beater{
-		done:   make(chan struct{}),
 		config: beaterConfig,
 	}
 	return bt, nil
@@ -43,17 +44,18 @@ func (bt *beater) Run(b *beat.Beat) error {
 		go bt.client.PublishAll(events)
 	}
 
-	server := newServer(bt.config, callback)
-	start(server, bt.config.SSL)
-	defer stop(server)
+	bt.server = newServer(bt.config, callback)
+	err = run(bt.server, bt.config.SSL)
+	logp.Err(err.Error())
 
-	logp.Info("apm-server is running! Hit CTRL-C to stop it.")
-	// Blocks until service is shut down
-	<-bt.done
-
-	return nil
+	if err == http.ErrServerClosed {
+		return nil
+	}
+	return err
 }
 
+// Graceful shutdown
 func (bt *beater) Stop() {
-	close(bt.done)
+	logp.Info("stopping apm-server...")
+	stop(bt.server)
 }
--- a/beater/server.go
+++ b/beater/server.go
@@ -62,18 +62,21 @@ func newServer(config Config, publish successCallback) *http.Server {
 	}
 }
 
-func start(server *http.Server, ssl *SSLConfig) {
+func run(server *http.Server, ssl *SSLConfig) error {
+	logp.Info("starting apm-server! Hit CTRL-C to stop it.")
 	if ssl.isEnabled() {
-		go server.ListenAndServeTLS(ssl.Cert, ssl.PrivateKey)
+		return server.ListenAndServeTLS(ssl.Cert, ssl.PrivateKey)
 	} else {
-		go server.ListenAndServe()
+		return server.ListenAndServe()
 	}
 }
 
-func stop(server *http.Server) error {
+func stop(server *http.Server) {
 	c := context.TODO()
 	err := server.Shutdown(c)
-	return err
+	if err != nil {
+		logp.Err(err.Error())
+	}
 }
 
 type handler func(w http.ResponseWriter, r *http.Request)
--- a/beater/server_test.go
+++ b/beater/server_test.go
@@ -49,7 +49,7 @@ func setupHTTP(t *testing.T) (*http.Server, *http.Request) {
 	cfg := defaultConfig
 	cfg.Host = host
 	apm := newServer(cfg, func(_ []beat.Event) {})
-	start(apm, cfg.SSL)
+	go run(apm, cfg.SSL)
 	waitForServer(false, host)
 	data, _ := tests.LoadValidData("transaction")
 	req, err := http.NewRequest("POST", transaction.Endpoint, bytes.NewReader(data))
@@ -75,18 +75,13 @@ func setupHTTPS(t *testing.T, useCert bool, domain string) (*http.Server, string
 	}
 
 	t.Log("Starting server on ", host)
-	start(apm, cfg.SSL)
+	go run(apm, cfg.SSL)
 
 	waitForServer(true, host)
 
 	data, _ := tests.LoadValidData("transaction")
 
 	return apm, host, data
-}
-
-func tearDown(t *testing.T, apm *http.Server) {
-	err := stop(apm)
-	assert.Nil(t, err, err)
 }
 
 func randomAddr() string {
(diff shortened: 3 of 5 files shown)


beater/beater.go
```diff
@@ -44,16 +43,19 @@ func (bt *beater) Run(b *beat.Beat) error {
 	}
 
 	server := newServer(bt.config, callback)
-	start(server, bt.config.SSL)
+	bt.errCh = start(server, bt.config.SSL)
 	defer stop(server)
 
 	logp.Info("apm-server is running! Hit CTRL-C to stop it.")
```
ruflin: It's a bit strange that `running` is printed before it is actually running.

beater/server.go
```diff
@@ -62,12 +62,18 @@ func newServer(config Config, publish successCallback) *http.Server {
...
 }
 
-func start(server *http.Server, ssl *SSLConfig) {
-	if ssl.isEnabled() {
-		go server.ListenAndServeTLS(ssl.Cert, ssl.PrivateKey)
-	} else {
-		go server.ListenAndServe()
-	}
+func start(server *http.Server, ssl *SSLConfig) chan(error) {
+	c := make(chan(error), 1)
```
ruflin: Instead of having the go routines in here, we could call `start` inside a go routine. Like this we would not need an error channel but could directly handle it in `beater.go`. Something like:

```
go func() {
	err := start(..., ...)
	if err != nil {
		logp.Err(err)
		close(bt.done)
	}

}

<-bt.done
```

In this case I would probably rename the function from `start` to `run`.

ruflin:
For the check to see if the port is available, I was thinking to do something like:

```
conn, err := net.Dial("tcp", host)
conn.close()
if err == nil {
    // Connection available -> do not start
    return fmt.Errorf("Port already taken")
}
```

But I'm also asking myself if both checks are potentially overkill and what you proposed is sufficient. The main reason I would want 2 checks is because the second check is async, mean it could tell the user the server is "normally" running but it will stop right afterwards. As we talk here about a ms delay, it's probably not an issue.

ruflin:
@jalvz Could you also update the Changelog? I labeled this PR bug as I think that apm-server is not stopping when the http server is not started is actually a bug. Good catch.

beater/beater.go
```diff
@@ -9,7 +9,7 @@ import (
 )
 
 type beater struct {
-	done   chan struct{}
+	errCh  chan error
```
jalvz: > I'm hesitant to call this errCh as the show down is not only an erro
Not true, it is always an error: https://github.com/golang/go/blob/d6cada7fa4ef92190c2b59c9048413f9280e4a74/src/net/http/server.go#L2680

I found `errCh` as variable name in golang source code itself, surely is idiomatic enough.

> there are potentially 2 different channels needed
I'm pretty sure that 2 channels are not required to shutdown a server.

beater/beater.go
jalvz: `start` happens in L46, and log in L49. 
Before the refactor merged yesterday that was indeed the case, but not anymore.

jalvz:
> It would be nice if the first case would be synchronous, means we check if the port is taken or not without trying to start the server.

That is precisely what `ListenAndServe` does. It would certainly feel overkill to do it twice.

> Could you also update the Changelog? I labeled this PR bug.

Sure thing. I found this some time ago, I thought it was intended that way. Glad to hear otherwise.


This PR still is not what want, btw. I'll update in a sec.

beater/server.go
jalvz: I prefer this way instead. This is easy to test. `beater.go` is a libbit dependency and doesn't lend itself to much testing.

beater/server_test.go
```diff
@@ -20,6 +20,7 @@ import (
 	"github.com/elastic/apm-server/processor/transaction"
 	"github.com/elastic/apm-server/tests"
 	"github.com/elastic/beats/libbeat/beat"
+	"github.com/pkg/errors"
```
jalvz: package is used in a test

beater/beater.go
ruflin: See my follow up comment. If we change the implementation we don't event need an error channel. WDYT?

beater/beater.go
ruflin: Hm,  I see. It's still the case that the log is potentially printed before the http server is running, but don't know how we could do it otherwise as it happens in a go routine and then is blocking. 

Perhaps we should rephrase the log to "started" as I would say running means the http listener is able to receive calls?

beater/server.go
ruflin: The advantage of the above is that we don't need to pass around an error channel and the it is directly visible in the `beater.go` on what is happening. 

For testing you could also extract that part in a method so you can test it separately. 

In general the part I don't like about the previous code is that we start go routines in two different places for kind of the same thing.

beater/beater.go
```diff
@@ -3,14 +3,17 @@ package beater
 import (
 	"fmt"
 
+	"net/http"
```
ruflin: new line above.

beater/beater.go
ruflin: By "not only an error" I was not referring to the message coming from the Listener, but the way the `done` channel is used. If the channel contains an error I'm definitively ok with the name, but I think we don't even need to send an error over the channel if the code is changed.

beater/beater.go
```diff
@@ -43,17 +45,21 @@ func (bt *beater) Run(b *beat.Beat) error {
...
 
 	logp.Info("apm-server is running! Hit CTRL-C to stop it.")
-	// Blocks until service is shut down
-	<-bt.done
 
-	return nil
+	err = <-bt.errCh
+	logp.Err(err.Error())
+
+	if err == http.ErrServerClosed {
```
ruflin: Nice, didn't spot that one in the first review.

beater/beater.go
```diff
@@ -43,17 +45,21 @@ func (bt *beater) Run(b *beat.Beat) error {
...
+	bt.server = newServer(bt.config, callback)
+	bt.errCh = start(bt.server, bt.config.SSL)
 
 	logp.Info("apm-server is running! Hit CTRL-C to stop it.")
-	// Blocks until service is shut down
-	<-bt.done
 
-	return nil
+	err = <-bt.errCh
+	logp.Err(err.Error())
```
ruflin: This should probably also happen after line 56, as otherwise always an err is logged?

roncohen:
Great work guys. I'm wondering if we really need to call `ListenAndServe` in its own go-routine? If we could call them them synchronously, this code would become way simpler.

beater/beater.go
```diff
@@ -43,17 +45,21 @@ func (bt *beater) Run(b *beat.Beat) error {
...
+	if err == http.ErrServerClosed {
+		return nil
+	}
+	return err
 }
 
+// Graceful shutdown
 func (bt *beater) Stop() {
-	close(bt.done)
+	stop(bt.server)
```
ruflin: I would not recommend to use `stop` here but use `defer` instead. This has proven to be better on edge cases like when Stop is called before the server is actually started. Also `Stop` might potentially be removed in the future.

beater/server.go
```diff
@@ -62,18 +62,26 @@ func newServer(config Config, publish successCallback) *http.Server {
...
+
+	return c
 }
 
-func stop(server *http.Server) error {
+func stop(server *http.Server) {
 	c := context.TODO()
 	err := server.Shutdown(c)
-	return err
+	if err != nil {
```
ruflin: I would leave it to the caller to decide if an error should be logged or not. This also makes it easier for testing to see if an error was returned or not.

roncohen:
Big +1 for making the Listen calls synchronous as long as we can.

jalvz:
@ruflin @roncohen is anything else required to merge?

beater/beater.go
```diff
@@ -43,17 +44,18 @@ func (bt *beater) Run(b *beat.Beat) error {
...
-	server := newServer(bt.config, callback)
-	start(server, bt.config.SSL)
-	defer stop(server)
-
-	logp.Info("apm-server is running! Hit CTRL-C to stop it.")
-	// Blocks until service is shut down
-	<-bt.done
+	bt.server = newServer(bt.config, callback)
+	err = run(bt.server, bt.config.SSL)
+	logp.Err(err.Error())
```
ruflin: Comment seems to have disappeared during rebase. As we return the error, we don't need to log it at all. It will be printed out by the beat afterwards (if it is not nil)

jalvz: I checked the open comments and I think its fine. Error is always not nil, but the returned value to the beat might be nil, so there is a case for this log. thanks

PR #97, opened 2017-08-24
Add debug logging for events received

ruflin:
The following things are now logged:

* Size of request
* Agent name + version request coming from
* Number of events in one request
* App name

### DIFF
--- a/beater/server.go
+++ b/beater/server.go
@@ -80,7 +80,7 @@ type handler func(w http.ResponseWriter, r *http.Request)
 
 func createHandler(p processor.Processor, config Config, publish successCallback) handler {
 	return func(w http.ResponseWriter, r *http.Request) {
-		logp.Debug("handler", "Request - URI: %s ; Method: %s", r.RequestURI, r.Method)
+		logp.Debug("handler", "Request: URI=%s, method=%s, content-length=%d", r.RequestURI, r.Method, r.ContentLength)
 
 		if !checkSecretToken(r, config.SecretToken) {
 			sendError(w, r, 401, "Invalid token", true)
--- a/processor/error/payload.go
+++ b/processor/error/payload.go
@@ -4,6 +4,7 @@ import (
 	pr "github.com/elastic/apm-server/processor"
 	m "github.com/elastic/apm-server/processor/model"
 	"github.com/elastic/beats/libbeat/beat"
+	"github.com/elastic/beats/libbeat/logp"
 )
 
 type payload struct {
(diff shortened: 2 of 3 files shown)


beater/server.go
```diff
@@ -80,7 +80,7 @@ type handler func(w http.ResponseWriter, r *http.Request)
 
 func createHandler(p processor.Processor, config Config, publish successCallback) handler {
 	return func(w http.ResponseWriter, r *http.Request) {
-		logp.Debug("handler", "Request - URI: %s ; Method: %s", r.RequestURI, r.Method)
+		logp.Debug("handler", "Request - URI: %s ; Method: %s; Content Length: %d", r.RequestURI, r.Method, r.ContentLength)
```
roncohen: I think we should use the same format as `metrics.go` in beats (`field=value`):

    2017/08/24 13:56:45.531578 metrics.go:39: INFO Non-zero metrics in the last 30s: beat.memstats.memory_alloc=-24968 beat.memstats.memory_total=21488 libbeat.pipeline.events.retry=1

ruflin: Changed. This reminds me that we should start introducing metrics in our code that we can later use for monitoring. Will create a github issue.

processor/error/payload.go
```diff
@@ -15,6 +16,8 @@ type payload struct {
 func (pa *payload) transform() []beat.Event {
 	var events []beat.Event
 
+	logp.Debug("error", "Transform error events: events=%d, app=%s, agent=%s:%s", len(pa.Events), pa.App.Name, pa.App.Agent.Name, pa.App.Agent.Version)
```
roncohen: any particular reason you went with `name:version`? We used to do `name/version`

roncohen:
LGTM 👍 

Is there a mode where we can output just these and not get the full debug logs out? Probably using the debug log filter on the command line?

That would be useful for anyone who wants to know how much data is coming in from a specific language etc.

ruflin:
@roncohen Exactly. If you would only want to see the one from the server you would run `./apm-server -e -d "handler", or for all 3: `./apm-server -e -d "handler,error,transaction"`.

For your use case this does look too complex in my opinion, so we should definitively discuss in a follow up if we should have some namespaces across packages and even document these. Like this we could tell users on what `filters` they should use for which use case. So far we did not do a very good job at this in beats as we treated it mainly for internal use.

processor/error/payload.go
ruflin: Not really. I think I was used to it because of docker. Can change it if you like :-)

PR #98, opened 2017-08-24
Seperate Processors and Payloads

### LINKED ISSUE #96: Discrepancy between agent.name and language.name in context
Steps to reproduce: Expand any document in the Discover table to review the context info.

screenshot 2017-08-24 10 57 34

roncohen:
Fixes #96.

Tests were updated so they reuse processors, like we do in the actual server code.

### DIFF
--- a/beater/server.go
+++ b/beater/server.go
@@ -101,13 +101,25 @@ func createHandler(p processor.Processor, config Config, publish successCallback
 
 		// Limit size of request to prevent for example zip bombs
 		limitedReader := io.LimitReader(reader, config.MaxUnzippedSize)
-		err = p.Validate(limitedReader)
+
+		buf, err := ioutil.ReadAll(limitedReader)
 		if err != nil {
-			sendError(w, r, 400, fmt.Sprintf("Data Validation error: %s", err), true)
+			// If we run out of memory, for example
+			sendError(w, r, 500, fmt.Sprintf("Data read error: %s", err), true)
+		}
+
+		err = p.Validate(buf)
+		if err != nil {
+			sendError(w, r, 400, fmt.Sprintf("Data validation error: %s", err), true)
 			return
 		}
 
-		list := p.Transform()
+		list, err := p.Transform(buf)
+
+		if err != nil {
+			sendError(w, r, 500, fmt.Sprintf("Data transformation error: %s", err), true)
+			return
+		}
 
 		w.WriteHeader(202)
 		publish(list)
--- a/processor/transaction/processor.go
+++ b/processor/transaction/processor.go
@@ -1,7 +1,7 @@
 package transaction
 
 import (
-	"io"
+	"encoding/json"
 
 	pr "github.com/elastic/apm-server/processor"
 	"github.com/elastic/beats/libbeat/beat"
@@ -20,22 +20,26 @@ const (
 
 func NewProcessor() pr.Processor {
 	return &processor{
-		payload: &payload{},
-		schema:  pr.CreateSchema(transactionSchema, processorName),
+		schema: pr.CreateSchema(transactionSchema, processorName),
 	}
 }
 
 type processor struct {
-	payload *payload
-	schema  *jsonschema.Schema
+	schema *jsonschema.Schema
 }
 
-func (p *processor) Transform() []beat.Event {
-	return p.payload.transform()
+func (p *processor) Validate(buf []byte) error {
+	return pr.Validate(buf, p.schema)
 }
 
-func (p *processor) Validate(reader io.Reader) error {
-	return pr.Validate(reader, p.schema, p.payload)
+func (p *processor) Transform(buf []byte) ([]beat.Event, error) {
+	var pa payload
+	err := json.Unmarshal(buf, &pa)
+	if err != nil {
+		return nil, err
+	}
+
+	return pa.transform(), nil
 }
 
 func (p *processor) Name() string {
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -14,7 +14,7 @@ https://github.com/elastic/apm-server/compare/x...master[View commits]
 ==== Bugfixes
 - changed `context.system.title` to `context.system.process_title`, removed `transaction.context`, `trace.context` (already available on top level). {pull}10[10]
 - changed type of `context.request.body` from `object` to `text`. {pull}27[27]
-
+- incoming transactions and errors could reuse data from previous POSTs due to payload state being kept in processors {pull}98[98].
 
 ==== Added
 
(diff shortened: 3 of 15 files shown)


ruflin:
This does not only seem to fix the bug we had in our code but I think it also makes the code more readable. 👍 

@roncohen You need to rebase. Seems you used an old version of master :-)

beater/server.go
```diff
@@ -101,13 +101,13 @@ func createHandler(p processor.Processor, config Config, publish successCallback
...
 		limitedReader := io.LimitReader(reader, config.MaxUnzippedSize)
-		err = p.Validate(limitedReader)
+		payload, err := p.Validate(limitedReader)
 		if err != nil {
 			sendError(w, r, 400, fmt.Sprintf("Data Validation error: %s", err), true)
 			return
 		}
 
-		list := p.Transform()
+		list := payload.Transform()
```
simitt: We made the `payload.transform()` method private. I think `p.Transform(payload)` should be the way to go.

processor/transaction/processor.go
```diff
@@ -20,22 +19,17 @@ const (
...
 
-func (p *processor) Transform() []beat.Event {
-	return p.payload.transform()
-}
-
-func (p *processor) Validate(reader io.Reader) error {
-	return pr.Validate(reader, p.schema, p.payload)
+func (p *processor) Validate(reader io.Reader) (pr.Payload, error) {
+	var payload Payload
+	return &payload, pr.Validate(reader, p.schema, &payload)
```
simitt: what do you think about returning the validated `payload` from the `pr.Validate` method?

beater/server.go
simitt: I would not expose the `payload` here, but rather handle it as an interna of `processor`. The publicly known and agreed interface is `Processor` which offers a `Validate(..)` and a `Transform(..)` method.

processor/transaction/processor.go
roncohen: `pr.Validate` is used both by errors and processors. It's generic: https://github.com/elastic/apm-server/blob/master/processor/validator.go#L26

But the Payloads are different. One is from the error package and one is from the transaction package. Suggestions are welcome.

beater/server.go
roncohen: I considered that, but it means the interface will have a method: `Transform(Payload)` which takes a `Payload` interface. If we remove the `Transform` method from the `Payload` interface, there's nothing left. So it take an empty interface, which could be anything. 

Also, the `Transform` method would look like this. Notice the need to type assert `pr.Payload` to an actual struct:

```
func (p *processor) Transform(payload pr.Payload) []beat.Event {
	var events []beat.Event
	pa := payload.(Payload)
	for _, tx := range pa.Events {

		events = append(events, pr.CreateDoc(tx.Mappings(pa)))

		for _, tr := range tx.Traces {
			events = append(events, pr.CreateDoc(tr.Mappings(pa, tx)))
		}
	}

	return events
}
```

processor/transaction/processor.go
roncohen: or did you mean to both take it as the third argument by reference (like in the suggested above) _and_ return it?

beater/server.go
simitt: I think we currently mix `Validation` and `Transformation` together, which is more appearant now than it was before. 
How about clearly separating those two steps, e.g.
```
var buf bytes.Buffer
dataReader := io.TeeReader(reader, &buf)
err := processor.Validate(dataReader)
//do some error check
processor.Transform(buf.Bytes)
```

I have not done a test implementation, but I think it should also be easier to keep the internal logic private then.

processor/transaction/processor.go
crackerplace: @simitt If I were to add few points here,I like what you said and returning payload sounds better.

As we are registering the processors only once using NewProcessor,they are singleton.So,ideally our processor methods which act on input request should be stateless and and not mutate data(payload) but instead work like functions.

simitt: @roncohen as this all plays together, please see my [comment above](https://github.com/elastic/apm-server/pull/98/files#r135033960)

roncohen:
I took another round on this now:
* Payload is private and the interface is gone again.
* I took JSON unmarshalling out of `Validate` and put it in `Transform` now as suggested by @simitt.
* Both `Validate` and `Transform` simply take `[]byte` now. `Validate` used to take an `io.Reader` but I also changed that to `[]byte` so it consistent now. The same `[]byte` can be used for both calls after each other. The alternative was for both of them to take an io.Reader and then use TeeReader to split the stream, but in the end that would mean all the data will be in memory anyway and we'd have the added complexity of the TeeReader.

If you're reviewing this, you should probably look at the holistic view and not the individual commits, as the latest one reverts some of the changes done in the first.

beater/server.go
```diff
@@ -101,13 +101,20 @@ func createHandler(p processor.Processor, config Config, publish successCallback
 
 		// Limit size of request to prevent for example zip bombs
 		limitedReader := io.LimitReader(reader, config.MaxUnzippedSize)
-		err = p.Validate(limitedReader)
+
+		buf, err := ioutil.ReadAll(limitedReader)
```
ruflin: I remember there was a reason I passed the reader instead of the buffer to Validate, but I can't see why anymore.

crackerplace:
@roncohen Just incase this helps,don't we want to handle error [here](https://github.com/roncohen/apm-server/blob/fix-96/processor/transaction/processor.go#L42) and [here](https://github.com/roncohen/apm-server/blob/fix-96/processor/transaction/processor.go#L42) so as to avoid calling publish.

processor/transaction/processor.go
```diff
@@ -39,7 +39,8 @@ func (p *processor) Transform(buf []byte) []beat.Event {
 	err := json.Unmarshal(buf, &pa)
 	if err != nil {
 		logp.Err("Error while unmarshaling data: %v", err)
-		return nil
+		// Effectively drop the data
+		return []beat.Event{}
```
ruflin: @roncohen I would probably even prefer if Transform could also return an error, so logging and treating of the error would happen in the server. Like this we only do it in one place. Also it would make testing `Transform` simpler I would think.

roncohen:
Following @crackerplace's comment I've changed it to return an empty list of `Event`s instead of `nil` if JSON unmarshalling in `Transform` fails - which shouldn't happen if our JSON Spec is always good.

ruflin:
@roncohen Left an additional comment on the last commit. I would prefer that we even return an error as I think it makes the code cleaner.

ruflin:
@roncohen Run `make fmt`, it seems pep8 does not like your python formatting.

roncohen:
@ruflin OK, fixed.

roncohen:
Rebased and update commit msg.

ruflin:
@roncohen Great step forward in cleaning up the code (and fixing the bug ;-) )
@crackerplace Thanks for your reviews here.

Issue #76, opened 2017-08-20
Processor methods look redundant in packages


crackerplace:
This might be similar to this [issue](https://github.com/elastic/apm-server/issues/26).

Looks like the processor.go in [error](https://github.com/elastic/apm-server/blob/master/processor/error/processor.go) and [transaction](https://github.com/elastic/apm-server/blob/master/processor/transaction/processor.go) package are no different and looks like we are having redundancy.

Either we can use go embedding or refactor in a better way.Will update with some ideas.

ruflin:
Agree that this could be potentially simplified through one of the two methods you mentioned. If you have suggestions, feel free to also directly open a PR. Sometimes discussing code is easier :-)

crackerplace:
Can I find some more info on the role of the processors ? Might be that helps.

A direct way can be as below

```
type Processor interface {
	Validate(io.Reader) error
	Payload
	Name() string
}

type processor struct {} // this struct can implement the default methods.

func (p *processor) Transform() []beat.Event {
	return p.payload.Transform()
}
```
Now for transaction processor we can embed the based processor so that it can access all methods of the base.

```
type TxProcessor struct {
  Processor
}
```

ruflin:
The role of processors is to take data, validate it, transform it and convert one input event to one or multiple output events.

Composing the Processors from a base processor is very/same as we do with MetricSets in metricbeat: https://github.com/elastic/beats/blob/master/metricbeat/module/apache/status/status.go#L44 So  it's definitively a direction I like.

I would also like to get @simitt and @jalvz involved on this as they were thinking about how to better structure the code quite a bit.

crackerplace:
Sure.Will implement once we have some consensus.

crackerplace:
Will start on this.

PR #102, opened 2017-08-25
[WIP/Idea] Remove Validate from Processor

ruflin:
Currently each processor requires to have a validate function. It kind of bugged me that these Validate functions are exactly the same. This change moves the Validate out of the processor by adding the schema to the processor registry. Like this also schema creating and handling is outside the processor.

I initially expected it to be a small change but it has lots of dependencies so it got more complex.

Pro:

* Schema handling outside of processor
* Schema is not persisted in processor itself
* Smaller interface

Cons:

* ProcessorData is not nice -> better name?
* Having schema inside processor would allow to use different "schemas" which are not json

I was initially playing with the idea to make `Schema()` an interface method on the processor like we have for `Name()` but as the schema is only use in one place, I prefer to pass it directly to the registry.

The current code is working, tests are not. I want to use this PR to discuss if this idea is something that we want to move forward with or drop.

### DIFF
--- a/beater/server.go
+++ b/beater/server.go
@@ -78,7 +78,7 @@ func stop(server *http.Server) error {
 
 type handler func(w http.ResponseWriter, r *http.Request)
 
-func createHandler(p processor.Processor, config Config, publish successCallback) handler {
+func createHandler(p processor.ProcessorData, config Config, publish successCallback) handler {
 	return func(w http.ResponseWriter, r *http.Request) {
 		logp.Debug("handler", "Request: URI=%s, method=%s, content-length=%d", r.RequestURI, r.Method, r.ContentLength)
 
@@ -108,13 +108,13 @@ func createHandler(p processor.Processor, config Config, publish successCallback
 			sendError(w, r, 500, fmt.Sprintf("Data read error: %s", err), true)
 		}
 
-		err = p.Validate(buf)
+		err = processor.Validate(buf, p.Schema)
 		if err != nil {
 			sendError(w, r, 400, fmt.Sprintf("Data validation error: %s", err), true)
 			return
 		}
 
-		list, err := p.Transform(buf)
+		list, err := p.Processor.Transform(buf)
 
 		if err != nil {
 			sendError(w, r, 500, fmt.Sprintf("Data transformation error: %s", err), true)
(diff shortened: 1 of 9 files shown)


roncohen:
I like not repeating ourselves with the schema validation, but i also like that validation as a concept is part of processing. It makes it clear that processors can have additional validation steps that are not part of the JSON Schema and makes handling of each data type (errors, transactions) a compartmentalized entity, which I think is neat.

This might be overkill, but how about using composition/delegation. Something like this:
```golang
type Validator interface {
  Validate(buf []byte) error
}

type Transformer interface {
  Transform(buf []byte) []Event
}

type Processor struct {
  Validator Validator
  Transformer Transformer
}

type SchemaValidator struct {
}

func (s *SchemaValidator) Validate(buf []byte) {
  // Validate as usual
}

func NewSchemaValidator(schemaFile string) {
   // load schema and return validator
}

var errorProcessor = Processor{
  Validator: NewSchemaValidator("file")
  Transformer: NewErrorTransformer()
}

var transactionProcessor = Processor{
  Validator: NewSchemaValidator("file")
  Transformer: NewTransactionTransformer()
}
```

ruflin:
I like the idea of composing our processors. This gives us the flexibility of reusing code but still flexibility in case the predefined code does not work to use our own. That is also what @crackerplace is pointing at in https://github.com/elastic/apm-server/issues/76

crackerplace:
What @roncohen specified looks good.

```type Processor struct {
	Name      string
	Path      string
	Validator Validator
	Transformer Transformer
}

p := new Processor(Validator:JsonValidator{Schema:TransactionSchema},
                  Transformer:ExtractEvents{})
```

One point which I was thinking on is about how would it be if we pass a context to processors for each request as we might want to pass additional attributes excluding the request bytes.

ruflin:
@crackerplace I kind of like the idea but not sure yet what we would use it for as I hope the processors stay simple. So I would only introduce this when we have a use case for it.

I'm closing this PR as I think it already trigger very good discussions on where we could go with the code and the code I have in this PR is not the one to go forward with. But lets continue the disucssion here.

@crackerplace Do you plan to work on this?

crackerplace:
@ruflin Sure.I will work on this.

PR #103, opened 2017-08-25
Improve logging for production

ruflin:
In production I expect a log message on the error level that I have to do something. For our 400 response no changes are needed on the server as the problem is potentially on the agent side.

Now only if 500 happens an error is logged, all other are logged on debug level if enabled. Otherwise the production log could be flooded with error messages because of invalid requests or tokens which can be pretty normal if the server has a public API.

### DIFF
--- a/beater/server.go
+++ b/beater/server.go
@@ -75,7 +75,7 @@ func createHandler(p processor.Processor, config Config, publish successCallback
 		logp.Debug("handler", "Request: URI=%s, method=%s, content-length=%d", r.RequestURI, r.Method, r.ContentLength)
 
 		if !checkSecretToken(r, config.SecretToken) {
-			sendError(w, r, 401, "Invalid token", true)
+			sendError(w, r, 401, "Invalid token", false)
 			return
 		}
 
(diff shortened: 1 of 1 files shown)


roncohen:
I see where you're coming from on this, but I actually prefer to have it logged out in both places. As a user, you might not be monitoring your application logs very closely. 

Also the actual amount of logs that this PR removes should be minimal because we expect 400s to only happen when there's a mismatch between what agents send and what the server expects.

ruflin:
I hope the log entries that I removed are not create by our own agents. They should already create valid data and request and for the purpose of testing or seeing what is wrong, I added the debug logging for all the error request which I was missing previously. I'm mainly thinking about everything out in the internet that pings any public API. This applies mainly in case people expose the API endpoint publicly, but as the Application and the server are not necessarly running at the same place, this could be a common use case.

ruflin:
@roncohen BTW: I could see us logging this on `Info` level instead of `Debug`.

jalvz:
I also prefer to keep 4XX logs until proven they are a problem.

ruflin:
The logs are still there, the question is on which log level. As both of you don't seem the like the `debug` level, what about `info`?

jalvz:
`info` is fine by me

ruflin:
Changed to `Info`

roncohen:
@ruflin what do you think about using `code=400` here? 

```
2017/08/28 08:16:21.908104 server.go:75: DBG Request: URI=/v1/errors, method=POST, content-length=141
2017/08/28 08:16:21.908232 server.go:124: INFO Data validation error: Problem validating JSON document against schema: I[#] S[#] doesn't validate with "error#"
  I[#/app] S[#/properties/app/required] missing properties: "name", code: 400
```

ruflin:
@roncohen Changed. I'm used to use `:` in beats but also there we are very inconsistent. We only used the `=` so far for the metrics logging case. Both options work for me.

Issue #99, opened 2017-08-24
Add metrics to apm-server


ruflin:
Libbeat offers to possiblity to register metrics (see example here: https://github.com/elastic/beats/blob/master/filebeat/prospector/log/harvester.go#L40). These metrics can then be logged every x seconds or also be used for monitoring. We should add the following metrics. Please comment with additional metrics we should add:

* Request counter
* Counter for the different request errors
  * Invalid content
  * Size limit
  * ...
* Incoming transaction events counter
* Incoming error events counter
* Outgoing transaction events counter
* Outgoing error events counter
* Outgoing trace events counter

PR #104, opened 2017-08-25
Introduce metrics to apm-server

ruflin:
This is a first take at adding metrics to the APM server. This is a basic set of metrics to get stats for requests, response and processor internals. We should adjust, add and modify these metrics over time when we see which ones are valuable in production, which ones are not and which ones are missing. This only sets the foundation.

These metrics can be later used to be shipped for monitoring. To access the metrics right now start apm-server as following:

```
./apm-server -e -E http.enabled=true
```

Then you can access the metrics under http://localhost:5066/stats?pretty

Closes https://github.com/elastic/apm-server/issues/99

### DIFF
--- a/beater/server.go
+++ b/beater/server.go
@@ -16,6 +16,14 @@ import (
 	"github.com/elastic/apm-server/processor"
 	"github.com/elastic/beats/libbeat/beat"
 	"github.com/elastic/beats/libbeat/logp"
+	"github.com/elastic/beats/libbeat/monitoring"
+)
+
+var (
+	serverMetrics  = monitoring.Default.NewRegistry("apm-server.server")
+	requestCounter = monitoring.NewInt(serverMetrics, "requests.counter")
+	responseValid  = monitoring.NewInt(serverMetrics, "response.valid")
+	responseErrors = monitoring.NewInt(serverMetrics, "response.errors")
 )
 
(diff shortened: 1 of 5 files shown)


roncohen:
This is a great start! Does it get logged out by default?

It would be great to have number of successfully processed events per processor and have the number of errors per processor as an explicit metric here. I see what we can calculate it as `validations - transformations` but having it explicitly would be really useful.

ruflin:
The metrics are logged every 30s. This can be adjusted with `logging.metrics.period: 30s`.

Will add the error counters, but it will make the Validation code a bit less clean.

beater/server.go
```diff
@@ -36,7 +43,9 @@ func newServer(config Config, publish successCallback) *http.Server {
 	}
 
 	mux.HandleFunc("/healthcheck", func(w http.ResponseWriter, r *http.Request) {
+		requestCounter.Inc()
```
jalvz: i'd remove  `healthcheck`counters, don't think they provide any value.

ruflin: I would like to keep this in to make sure `# of request = actual number of requests`. Otherwise if for whatever reason `healthcheck` is hammered and the server is slowing down but the request counter stays low it might imply the it's not a request problem.

PR #105, opened 2017-08-25
Cleanup logging for PR #94

ruflin:
Address review comments from https://github.com/elastic/apm-server/pull/94

* In any case on shutdown the apm-server was logging an error even if it was shut down nicely. The logging message was removed because if it is an error, it will be logged by the beat anyways.
* Cleanup imports
* Logging should happen as late as possible. This makes also testing easier. Stop logging was moved out.

### DIFF
--- a/beater/beater.go
+++ b/beater/beater.go
@@ -2,7 +2,6 @@ package beater
 
 import (
 	"fmt"
-
 	"net/http"
 
 	"github.com/elastic/beats/libbeat/beat"
@@ -46,8 +45,6 @@ func (bt *beater) Run(b *beat.Beat) error {
 
 	bt.server = newServer(bt.config, callback)
 	err = run(bt.server, bt.config.SSL)
-	logp.Err(err.Error())
-
 	if err == http.ErrServerClosed {
 		return nil
 	}
@@ -57,5 +54,9 @@ func (bt *beater) Run(b *beat.Beat) error {
 // Graceful shutdown
 func (bt *beater) Stop() {
 	logp.Info("stopping apm-server...")
-	stop(bt.server, bt.config.ShutdownTimeout)
+
+	err := stop(bt.server, bt.config.ShutdownTimeout)
+	if err != nil {
+		logp.Err("Error stopping apm-server: %s", err.Error())
+	}
 }
--- a/beater/server.go
+++ b/beater/server.go
@@ -54,18 +54,20 @@ func run(server *http.Server, ssl *SSLConfig) error {
 	}
 }
 
-func stop(server *http.Server, timeout time.Duration) {
+func stop(server *http.Server, timeout time.Duration) (err error) {
 	ctx, cancel := context.WithTimeout(context.Background(), timeout)
 	defer cancel()
 
-	err := server.Shutdown(ctx)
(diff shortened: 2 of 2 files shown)


beater/server.go
```diff
@@ -54,18 +54,20 @@ func run(server *http.Server, ssl *SSLConfig) error {
...
 
-	err := server.Shutdown(ctx)
+	err = server.Shutdown(ctx)
 	if err != nil {
-		logp.Err(err.Error())
-		err = server.Close()
-		if err != nil {
-			logp.Err(err.Error())
-		}
+		defer func() {
```
jalvz: This is a lot problems just to log an error in the caller instead of the callee, which makes no difference at all.

beater/beater.go
```diff
@@ -46,8 +45,6 @@ func (bt *beater) Run(b *beat.Beat) error {
 
 	bt.server = newServer(bt.config, callback)
 	err = run(bt.server, bt.config.SSL)
-	logp.Err(err.Error())
```
jalvz: at very least log the error when the beat doesn't, inside the `if err == http.ErrServerClosed` block

beater/server.go
ruflin: It makes a difference for testing. You can't test if something was logged, but you can test if something returned an error.

beater/beater.go
ruflin: I think I miss the reason you want to log that. Isn't that the normal shutdown of the server so everything went as expected? Should this be an `Info` log that the server properly shut down?

beater/server.go
jalvz: I see no test taking advantage of that difference here. 
If you can come up with a test that exercises all the code paths (graceful shutdown ok, graceful shutdown timeout, and server close error), then the patch is welcome.

ruflin: We don't have these tests yet. But in general I prefer to have logging in as few places as possible and let errors bubble up instead of logging them out.

beater/beater.go
jalvz: how do you know is expected? how do you know that accidentally someone did not send a SIGTERM to the process?
that log line is simply saying *why the server stopped. 
and again, `srv.Serve` *always returns a non nil error, you are not respecting the api.

ruflin: For me a sigterm is a proper stopping as some on purpose wanted to stop the server. If the server then shuts down properly, no Error should be logged. 

I don't think that is not respecting the API. For example `EOF` when closing a file is also an error. But often EOF is an expected error. So here we make the decision that the `ErrServerClosed` is expected.

ruflin:
Closing this for now after conversation with @jalvz as the logging is going to stay mostly the same.

@jalvz Could you have a look again at comment https://github.com/elastic/apm-server/pull/105/files#r135464828 ?

beater/beater.go
jalvz: What I mean is that a signal can be delivered unintentionally, because eg. you give a wrong `pid` to `kill` or because you `Ctrl-C` in the wrong tab, and you might be stopping the server without noticing. The log simply says why the server stops. This is bikeshedding anyways... `Info` is good enough for me.

PR #106, opened 2017-08-25
Increment version to 0.2.0

ruflin:


jalvz:
why?

ruflin:
Mainly for our internal testing. We recently had quite a few breaking changes in so if we update the version we can see in our test environment which events come from the new version.

PR #107, opened 2017-08-25
Force HTTP server to close if it doesn't shutdown gracefully.

jalvz:
Exit code will still be 0.

### DIFF
--- a/_meta/beat.yml
+++ b/_meta/beat.yml
@@ -10,6 +10,7 @@ apm-server:
   #max_header_bytes:
   #read_timeout: 2s
   #write_timeout: 2s
+  #shutdown_timeout: 5s
 
   # Authorization token to be checked. If a token is set here the agents must
   # send their token in the following format: Authorization: Bearer <secret-token>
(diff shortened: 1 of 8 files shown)


_meta/beat.yml
```diff
@@ -10,6 +10,7 @@ apm-server:
   #max_header_bytes:
   #read_timeout: 2s
   #write_timeout: 2s
+  #shutdown_timeout: 5s
```
ruflin: Reminds me of: https://github.com/elastic/beats/blob/master/filebeat/_meta/common.reference.p2.yml#L254 Nice that we have the same naming.

ruflin:
@jalvz One thing we should discuss at one stage is if `5s` is a good default. We struggled in beats finding a good default which lead us in the end to us to not enable the timeout by default. So people that need a timeout can set it themself to the right value.

jalvz:
Just wait to see if many people complaints about seeing a context timeout. I wouldn't even make this configurable...

PR #108, opened 2017-08-28
Use one beat client per request.

jalvz:


beater/server.go
```diff
@@ -123,9 +123,14 @@ func createHandler(p processor.Processor, config Config, publish successCallback
 			return
 		}
 
+		err = publish(list)
+		if err != nil {
+			sendError(w, r, 500, fmt.Sprintf(err.Error()), true)
```
ruflin: I suggest to change this to:
```
sendError(w, r, 500, fmt.Sprintf("error publishing events: %s", err.Error()),
```
This will make it easier to find the error in case it is cryptic.

ruflin:
I quickly checked the branch locally and I think the reason it breaks the system tests it that the system tests load x events into es and check if x events exists. But now with this change x+1 events exist. I see two options for a fix:

* Change it to x+1 in the tests
* Specific a more specific query to exclude the onboarding event

Option 1 is probably much simpler.

jalvz:
This conflicts with PR109. Should be closed once 109 is in master...

ruflin:
I suggest to get this change in and then rebase https://github.com/elastic/apm-server/pull/109 on top of it?

ruflin:
Closing as this was covered in https://github.com/elastic/apm-server/pull/118

PR #109, opened 2017-08-28
Respond with 503 immediately if internal queue is full.

roncohen:
This changes the APM Server to immediately return 503 when the queue is full. I tried this as an experiment to get the desired semantics while requiring no changes to libbeat.

Start the server:
```
$ ./apm-server -e -E queue.mem.events=40
```

Send data:
```
$ for i in {0..60}; do curl localhost:8200/v1/errors -w "$i - %{http_code}\n" -d @tests/data/valid/error/minimal_payload_log.json  -H "Content-Type: application/json" ; done
0 - 202
1 - 202
2 - 202
3 - 202
4 - 202
5 - 202
6 - 202
7 - 202
8 - 202
9 - 202
10 - 202
11 - 202
12 - 202
13 - 202
...
58 - 202
59 - 202
{"error":"Error adding data to internal queue: Queue is full"}60 - 503
```

Details:
* You'll see it accepts 60 instead 40 events. Long explanation:
I'm reusing `Client` for between requests. The publish callback is protected by a mutex so there's no concurrent usage of the `Client`. I had reuse `Client` because:
  * The code relies on the `DropIfFull` publishing mode which will return immediately if the queue is full: https://github.com/elastic/beats/blob/master/libbeat/publisher/queue/memqueue/produce.go#L121. Each new `Client` has a `Broker` and each `Broker` creates a golang channel with capacity of 20 events: https://github.com/elastic/beats/blob/master/libbeat/publisher/queue/memqueue/broker.go#L85. This channel acts a small buffer that feeds the main queue. If we reuse the `Client`, they share this channel and when it fills up we'll start throwing 503s, which is the desired behaviour. If we don't reuse `Client`, every incoming request will have its own small buffer with capacity for 20 events. 
* There's a danger of partially publishing a set of output documents. 
Some context: Each event coming into our intake-API can generate multiple outgoing documents. Let's call the output documents resulting from a single incoming event for a "batch".  In general, partially dropping a batch is bad because that could lead to inconsistencies in the UI. A small number of short-lived inconsistencies cannot be avoided because of different inherent Elasticsearch/libbeat internals, but if we have a scenario where Elasticsearch has an ingestion rate which is fluctuating around the ingestion rate of the apm-server and the queue has filled up, we can have a problem where batches get half dropped, resulting in significant, permanent inconsistencies in the UI. A better semantic here would be to drop the whole batch if the queue does not have capacity for the all the documents in the batch.
* I don't think any of the above are blockers for moving forward with this.


urso:
> The code relies on the DropIfFull publishing mode which will return immediately if the queue is full: https://github.com/elastic/beats/blob/master/libbeat/publisher/queue/memqueue/produce.go#L121. Each new Client has a Broker and each Broker creates a golang channel with capacity of 20 events:

Each client has a connection to the broker. The broker is a shared resource. The internal golang channel with capacity of 20 events is some very internal implementation details which might be removed in the future.

urso:
> There's a danger of partially publishing a set of output documents.
Some context: Each event coming into our intake-API can generate multiple outgoing documents. Let's call the output documents resulting from a single incoming event for a "batch". In general, partially dropping a batch is bad because that could lead to inconsistencies in the UI. A small number of short-lived inconsistencies cannot be avoided because of different inherent Elasticsearch/libbeat internals, but if we have a scenario where Elasticsearch has an ingestion rate which is fluctuating around the ingestion rate of the apm-server and the queue has filled up, we can have a problem where batches get half dropped, resulting in significant, permanent inconsistencies in the UI. A better semantic here would be to drop the whole batch if the queue does not have capacity for the all the documents in the batch.

libbeat has no idea about the input batch. Even PublishAll forwards one event after another. Single events from multiple clients are potentially merged into one batch. Plus bulk requests to Elasticsearch are limited in the number of documents to be indexed. That is, the problem potentially exists on many levels. For UI consistency reasons I guess we want to publish either all or nothing. 

Some brainstorming following here:

One workaround I can think of is introducing another publisher layer with N beat.Client workers + a shared work queue/channel. The workers will use GuaranteedSend for retry and the channel for the workers drops if the queue is not served. e.g.

```
chDocs := make(chan []beat.Event, 50)
for i := 0; i < N; i++ {
  go func() { // worker (create N times)
    client, err := beat.Pipeline.ConnectWith(beat.ClientConfig{ PublishMode: beat.GuaranteedMode })
    for docs := range chDocs {
      client.PublishAll(docs) // blocks once broker is full -> no worker available to serve chDocs
    }
  }()
}

...

callback := func(events []beat.Event) error {
  select {
  case chDocs <- events:
    return nil
  default:
    fmt.Errorf("channel full")
  }
}

```

You don't want to use an unbuffered channel, as the select goes with the default case if no worker is already waiting for a client to publish requests (chance to drop events, besides queue is not full).

Due to creating batches of bulk_max_size, potentially multiple clients' events getting intermixed (if multiple workers are active) and potential out-of-order publishing if load-balancing is enabled, there is still a chance some partial events being published yet. But thanks to infinite retry the missing events will follow (given the beat with in memory queue is not stopped early).

Using the ACK handler + storing some intermediate context, one might consider to publish an 'OK' event after all docs have been indexed. Presence of an 'OK' event signals the UI all events are present, plus might contain some more meta-data/metrics (e.g. duration until all events are indexed).

beater/beater.go
```diff
@@ -29,19 +30,75 @@ func New(_ *beat.Beat, ucfg *common.Config) (beat.Beater, error) {
 	return bt, nil
 }
 
+type Eventer struct {
+	eventsLeft int
```
urso: test running with multiple agents send (sending events) fast and binary compiled with `-race`. Might be you want the type to be `atomic.Int32`.

ruflin:
@roncohen Thanks for pushing this PR. +1 on having this first in the apm-server (if possible) so we can iterate on it quickly.

As Steffen mentioned, missing events could happen an several levels. It could also be that a bulk request to ES only accepts a few of the documents for whatever reason. +1 on ignoring this as I think the UI should be able to cope with this. Right now as far as I can understand this can only happen to `transactions` with `traces`. A solution here could be that the transaction contains a `counter` on how many traces there should be, and if there are less `traces` in ES and marked in the `transaction`, the UI knows it's incomplete. 

My understanding of the returning a 503 directly when the queue is full has 2 reasons:

* Make sure the agents know as soon as possible that the event failed so they can move one
* Potentially tell a load balancer to send events somewhere else

For the first one: Assuming there is no load balancer, do we plan to allow the agents to send events to multiple `apm-server` instances, mean defining multiple hosts? I'm leaning in the direction of no and get users to use a load balancer or DNS load balancing, this keeps things simpler on the agent side.

ruflin:
To make sure we don't forget it, can you add a CHANGELOG entry?

ruflin:
jenkins, test it

roncohen:
> All clients/producers publish message on the very same channel.

You are right of course. My bad. I did a version that created a client per incoming http request which didn't work. The behaviour I was seeing there came from closing the client after `Eventer.Published()` was called for all events, but then when I call `client.Close()` in the end messages were cancelled. I tried changing it to only call `client.Close()` when messages have been ACKed instead, but i get `ACK handlers with DropIfFull mode not supported`.

urso:
The DropIfFull mode does not support an ACK handler due to the complexity in bookeeping this would require if an event based ACK handler is installed.

Oh, I see why events are dropped on close: https://github.com/elastic/beats/blob/master/libbeat/publisher/pipeline/pipeline.go#L324

The DropOnCancel is more of a hint. But the memory queue indeed tries to remove not-published events from the queue for all modes.

urso:
See elastic/beats#5044. The PR enables support for cancelling only on GuaranteedSend.

roncohen:
Updated now to use a `Client` per incoming request and used an atomic counter

roncohen:
relies on https://github.com/elastic/apm-server/pull/113

beater/beater.go
```diff
@@ -29,23 +30,69 @@ func New(_ *beat.Beat, ucfg *common.Config) (beat.Beater, error) {
 	return bt, nil
 }
 
-func (bt *beater) Run(b *beat.Beat) error {
+type Eventer struct {
+	eventsToBeQueued atomic.Int32
+	eventsToBeAcked  atomic.Int32
```
simitt: Is this `eventsToBeAcked` used for anything?

roncohen:
Fixed by https://github.com/elastic/apm-server/pull/118

PR #110, opened 2017-08-29
Push onboarding doc on startup.

jalvz:
It checks that HTTP Server is running. Onboarding doc is used to tell the user that connections between APM Server and Kibana are wired properly.

### DIFF
--- a/beater/beater.go
+++ b/beater/beater.go
@@ -47,7 +47,7 @@ func (bt *beater) Run(b *beat.Beat) error {
 
 	var checkServer = func() bool {
 		secure := bt.config.SSL.isEnabled()
-		return isServerUp(secure, bt.config.Host, 30, time.Second*10)
+		return isServerUp(secure, bt.config.Host, 10, time.Second*6)
 	}
 	go onboarding.NotifyUp(checkServer, callback)
 
--- a/beater/beater.go
+++ b/beater/beater.go
@@ -2,9 +2,10 @@ package beater
 
 import (
 	"fmt"
-
 	"net/http"
+	"time"
 
+	"github.com/elastic/apm-server/onboarding"
 	"github.com/elastic/beats/libbeat/beat"
 	"github.com/elastic/beats/libbeat/common"
 	"github.com/elastic/beats/libbeat/logp"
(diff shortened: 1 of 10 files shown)


processor/onboarding/onboarding.go
```diff
@@ -0,0 +1,33 @@
+package onboarding
```
ruflin: I would suggest not to put this under `processors` namespace but only `onboarding`. Reason is that from my point of view it doesn't have the typical properties of a processor like and endpoint handler or any special fields.

beater/server_test.go
```diff
@@ -90,34 +89,9 @@ func randomAddr() string {
...
-		}
-		return res.StatusCode
-	}
-
-	for i := 0; i <= 1000; i++ {
-		time.Sleep(time.Second / 50)
-		if check() == 200 {
-			return
-		}
+	if isServerUp(secure, host, 1000, time.Second/50) {
```
ruflin: Nice that we can reuse this here.

beater/beater.go
```diff
@@ -44,11 +45,17 @@ func (bt *beater) Run(b *beat.Beat) error {
...
+		return isServerUp(secure, bt.config.Host, 30, time.Second*10)
+	}
+	go onboarding.NotifyUp(checkServer, callback)
+
 	bt.server = newServer(bt.config, callback)
 	err = run(bt.server, bt.config.SSL)
-	logp.Err(err.Error())
 
 	if err == http.ErrServerClosed {
+		logp.Info(err.Error())
```
ruflin: thanks, appreciate it.

processor/onboarding/_meta/fields.yml
```diff
@@ -0,0 +1,5 @@
+- key: apm-onboarding
```
ruflin: I assume you added this because a processor requires it. See my comment below, I think we should not make it a processor so this would not be required. But I see the point that potentially we want to have a special field like `up` or something like that that indicates it's available.

jalvz:
Current status: integration tests at the moment make this PR very hard to ship without very dirty hacks.
They are so slow that we need to wait on a given count of documents to actually assert a document count... which means than a wrong count error actually surfaces a timeout. 
To make things worse, that count is distorted by the onboarding doc.
For instance a test that inserts 4 transactions might observe at some point a log indicating 4 insertions or 5. If it is 4, it could be the onboarding doc + 3 of the transactions, or the 4 transactions. 

We need to rework the integration tests before being able to move forward with this.

ruflin:
Closing in favor of https://github.com/elastic/apm-server/pull/117

PR #111, opened 2017-08-29
Add log message for which address server is started on

ruflin:


beater/server.go
```diff
@@ -46,6 +46,8 @@ func newServer(config Config, publish successCallback) *http.Server {
 		responseValid.Inc()
 	})
 
+	logp.Info("Starting server on address: %s", config.Host)
```
beniwohli: how about "Starting APM server on ..." to make it clear which server?

roncohen: `APM Server` or `apm-server`, please :)

ruflin: @beniwohli I'm actually thinking we should leave out apm part. As otherwise it will get confusing with the overall start of the APM Server.

What about `Start listening on address: ...`

beniwohli: Sounds fine :)

roncohen: I suggest simply: `Listening on %s`

ruflin:
Changed to the second proposal from @roncohen Good logging is hard :-)

PR #112, opened 2017-08-29
Roll new version for changelog, add missing directives and fix links

jalvz:


jalvz:
the backslash is such a subtle thing that i think is nice to have it a commit for it, in case someone later on thinks is a typo then there is something in the git history about it

ruflin:
@jalvz 3 commits SGTM. But I think in this case commit 2 and 3 probably need a bit longer message to make it clear it's about the changelog?

jalvz:
i worry that generated docs will be wrong with this backslash...

ruflin:
We can fix it when we get there :-)

PR #113, opened 2017-08-29
Update beats to 083834b

roncohen:
Need to utilize https://github.com/elastic/beats/pull/5044


_beats/dev-tools/cmd/index_template/index_template.go
```diff
@@ -40,7 +40,7 @@ func main() {
 		*version = "2.0.0"
 	}
 
-	tmpl, err := template.New(beatVersion, *version, *index, template.TemplateSettings{})
+	tmpl, err := template.New(beatVersion, *index, *version, template.TemplateConfig{})
```
ruflin: Good timing, wanted to do it for this change too.

PR #117, opened 2017-08-30
Push onboarding doc on startup

ruflin:
For the UI an onboarding doc is needed to see when the server is actually started. A doc is pushed with the field `up: true` and the `@timestamp` when it was pushed. This happens on each startup.

The code checks that the HTTP Server is running and as soon as this is the case, sends the doc.

This PR takes over the previous work from #110

### DIFF
--- a/beater/beater.go
+++ b/beater/beater.go
@@ -2,7 +2,6 @@ package beater
 
 import (
 	"fmt"
-
 	"net/http"
 
 	"github.com/elastic/beats/libbeat/beat"
@@ -37,11 +36,13 @@ func (bt *beater) Run(b *beat.Beat) error {
 	}
 	defer pub.Stop()
 
-	bt.server = newServer(bt.config, pub.Send)
-	err = run(bt.server, bt.config.SSL)
-	logp.Err(err.Error())
+	go notifyListening(bt.config, pub.Send)
 
+	bt.server = newServer(bt.config, pub.Send)
+
+	err = run(bt.server, bt.config.SSL)
 	if err == http.ErrServerClosed {
+		logp.Info("Listener stopped: %s", err.Error())
 		return nil
 	}
 	return err
--- a/beater/onboarding.go
+++ b/beater/onboarding.go
@@ -1,0 +1,28 @@
+package beater
+
+import (
+	"time"
+
+	"github.com/elastic/beats/libbeat/beat"
+	"github.com/elastic/beats/libbeat/common"
+	"github.com/elastic/beats/libbeat/logp"
+)
+
+func notifyListening(config Config, reporter reporter) {
+
+	var isServerUp = func() bool {
+		secure := config.SSL.isEnabled()
+		return isServerUp(secure, config.Host, 10, time.Second)
+	}
+
+	if isServerUp() {
+		logp.Info("Publishing onboarding document")
+
+		event := beat.Event{
+			Timestamp: time.Now(),
+			Fields:    common.MapStr{"listening": config.Host},
+		}
+		events := []beat.Event{event}
+		reporter(events)
+	}
+}
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -14,6 +14,7 @@ https://github.com/elastic/apm-server/compare/f9a2086ceed0b918e1a0b3d8ddc140fc21
 
 
 ==== Added
+- Send document to output on start of server {pull}117[117]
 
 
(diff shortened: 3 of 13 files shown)


ruflin:
This PR will conflict with https://github.com/elastic/apm-server/pull/118 or https://github.com/elastic/apm-server/pull/109 As these are more complex I prefer that one of the two gets merged and then I rebase this one on top of it.

beater/onboarding.go
```diff
@@ -0,0 +1,28 @@
...
+	"github.com/elastic/beats/libbeat/beat"
+	"github.com/elastic/beats/libbeat/common"
+	"github.com/elastic/beats/libbeat/logp"
+)
+
+func notifyListening(config Config, reporter reporter) {
+
+	var isServerUp = func() bool {
+		secure := config.SSL.isEnabled()
+		return isServerUp(secure, config.Host, 10, time.Second*1)
```
simitt: what is the `*1` for?

beater/beater.go
```diff
@@ -37,11 +36,13 @@ func (bt *beater) Run(b *beat.Beat) error {
 	}
 	defer pub.Stop()
 
+	go notifyListening(bt.config, pub.Send)
```
simitt: Could we trigger this after `err = run(bt.server, bt.config.SSL)` to ensure the server is up instead of polling for 10 seconds. 
The problem I see with the current handling is that the server has to be started within n seconds otherwise we never send a document to ES.

beater/beater.go
```diff
@@ -37,11 +36,13 @@ func (bt *beater) Run(b *beat.Beat) error {
...
+
 	bt.server = newServer(bt.config, pub.Send)
-	err = run(bt.server, bt.config.SSL)
-	logp.Err(err.Error())
 
+	err = run(bt.server, bt.config.SSL)
 	if err == http.ErrServerClosed {
+		logp.Info("Listener stopped: %s", err.Error())
```
simitt: We do send a document when the server is started, but we only log an information when it is stopped. Hence, having a document that the server was started in ES is not a relieable indicator for the server actually running. 

I think the original idea was to push a document and regulary update this document with the information that the server is running. When the last update of the document is too long ago this would indicate the server having stoped.

beater/onboarding.go
ruflin: leftover

beater/beater.go
ruflin: As `run` is blocking, we can't start it afterwards. I think it's pretty save that the server is starting in 10s or otherwise returns an error (I hope ).

beater/beater.go
ruflin: Such additional features will be covered by monitoring: https://github.com/elastic/beats/issues/3422

ruflin:
`*1` removed and CHANGELOG added.

beater/onboarding.go
```diff
@@ -0,0 +1,28 @@
...
+		secure := config.SSL.isEnabled()
+		return isServerUp(secure, config.Host, 10, time.Second)
+	}
+
+	if isServerUp() {
+		logp.Info("Publishing onboarding document")
+
+		event := beat.Event{
+			Timestamp: time.Now(),
+			Fields:    common.MapStr{"listening": config.Host},
```
jalvz: hey, why was this field added?

ruflin: @jalvz See PR message. Sorry, it was not there before, only in the commit message. Forgot to update the PR message too.

jalvz: @formgeist is relevant to show in the on-boarding page in which port is the APM Server running?

formgeist: @jalvz I don't think so, we'll just check that it's there and move on. Won't be printing the port to the user in the UI. Unless you feel that might be a good thing to do?

jalvz: Ok, if not used for on-boarding is better to remove it, this is more maintenance work potentially duplicated in beats monitoring. Are you ok with removing it @ruflin ?

ruflin: The reason I put this in as I think it will help us and the users to debug potential problems with the apm-server. So in case a user does not have access to apm-server and see it running, but his agent can't send data to it, we can ask him to check this doc about the host config that is set. I expect this to be one of the most common getting started errors to happen that the wrong host / port is set. I would keep it in as it does not add any complexity as far as I can see but brings potential value. I don't think we need it in the UI for now as the doc is always accessible through discovery.

PR #118, opened 2017-08-30
Attempt at generating a 503 if queue is full

urso:
This solution uses an intermediate go-routine accepting batches of events from the http handlers and forwarding these as single events to the beats publisher pipeline. The events are forwarded with infinite retrying enabled. Sending events to the go-routine's channel fails immediately if the channel is full. In this case a 503 response will be returned to the agent.

### DIFF
--- a/beater/beater.go
+++ b/beater/beater.go
@@ -13,7 +13,6 @@ import (
 type beater struct {
 	config Config
 	server *http.Server
-	client beat.Client
 }
 
 // Creates beater
@@ -30,21 +29,15 @@ func New(_ *beat.Beat, ucfg *common.Config) (beat.Beater, error) {
 }
 
 func (bt *beater) Run(b *beat.Beat) error {
-
 	var err error
 
-	bt.client, err = b.Publisher.Connect()
+	pub, err := newPublisher(b.Publisher, bt.config.ConcurrentRequests)
 	if err != nil {
 		return err
 	}
-	defer bt.client.Close()
+	defer pub.Stop()
 
-	callback := func(events []beat.Event) {
-		// Publishing does not wait for publishing to be acked
-		go bt.client.PublishAll(events)
-	}
-
-	bt.server = newServer(bt.config, callback)
+	bt.server = newServer(bt.config, pub.Send)
 	err = run(bt.server, bt.config.SSL)
 	logp.Err(err.Error())
 
--- a/beater/pub.go
+++ b/beater/pub.go
@@ -1,0 +1,86 @@
+package beater
+
+import (
+	"errors"
+	"sync"
+
+	"github.com/elastic/beats/libbeat/beat"
+)
+
+// publisher forwards batches of events to libbeat. It uses GuaranteedSend
+// to enable infinite retry of events being processed.
+// If the publisher's input channel is full, an error is returned immediately.
+// Number of concurrent requests waiting for processing do depend on the configured
+// queue size. As the publisher is not waiting for the outputs ACK, the total
+// number requests(events) active in the system can exceed the queue size. Only
+// the number of concurrent HTTP requests trying to publish at the same time is limited.
+type publisher struct {
+	events chan []beat.Event
+	client beat.Client
+	wg     sync.WaitGroup
+}
+
+var (
+	errFull              = errors.New("Queue is full")
+	errInvalidBufferSize = errors.New("Request buffer must be > 0")
+)
+
+// newPublisher creates a new publisher instance. A new go-routine is started
+// for forwarding events to libbeat. Stop must be called to close the
+// beat.Client and free resources.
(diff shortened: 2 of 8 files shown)


beater/beater.go
```diff
@@ -30,21 +29,15 @@ func New(_ *beat.Beat, ucfg *common.Config) (beat.Beater, error) {
...
 	if err != nil {
 		return err
 	}
-	defer bt.client.Close()
-
-	callback := func(events []beat.Event) {
-		// Publishing does not wait for publishing to be acked
-		go bt.client.PublishAll(events)
-	}
+	defer pub.Stop()
```
urso: - pub.Stop() blocks until the internal queue is drained. 
- the HTTP server must be shutdown before calling stop. Sending to `pub` after `Stop` generates a panic.

roncohen: would be useful to have these as comments

roncohen:
Looks good @urso. Nice work.
Also like the refactor you did in `processRequest` 👍

beater/beater.go
ruflin: @urso can you add a doc comment here?

beater/pub.go
```diff
@@ -0,0 +1,86 @@
...
+import (
+	"errors"
+	"sync"
+
+	"github.com/elastic/beats/libbeat/beat"
+)
+
+// publisher forwards batches of events to libbeat. It uses GuaranteedSend
+// to enable infinite retry of events being processed.
+// If the publisher it's input is channel is full, an error is returned
```
ruflin: s/it's/its/ (I think)

beater/pub.go
```diff
@@ -0,0 +1,86 @@
...
+// number requests(events) active in the system can exceed the queue size. Only
+// the number of concurrent HTTP requests trying to publish at the same time is limited.
+type publisher struct {
+	events chan []beat.Event
+	client beat.Client
+	wg     sync.WaitGroup
+}
+
+var (
+	errFull              = errors.New("Queue is full")
```
ruflin: want to start the errors lower case?

beater/server.go
```diff
@@ -26,14 +27,19 @@ var (
 	responseErrors = monitoring.NewInt(serverMetrics, "response.errors")
 )
 
-type successCallback func([]beat.Event)
+type reporter func([]beat.Event) error
 
-func newServer(config Config, publish successCallback) *http.Server {
+var (
+	errInvalidToken    = errors.New("Invalid token")
```
ruflin: lower case?

beater/server.go
```diff
@@ -82,79 +88,77 @@ func stop(server *http.Server, timeout time.Duration) {
...
-	if log {
-		logp.Err(error)
-	} else {
-		logp.Info("%s, code=%d", error, code)
+	// Limit size of request to prevent for example zip bombs
+	limitedReader := io.LimitReader(reader, maxSize)
+	buf, err := ioutil.ReadAll(limitedReader)
+	if err != nil {
+		// If we run out of memory, for example
+		return reportError(500, fmt.Errorf("Data read error: %s", err))
```
ruflin: lower case? :-D I know that was here before, so don't mind ;-)

urso:
For testing I did start apm-server with: `./apm-server -e -E queue.mem.events=40`

and did send data with `for i in {0..80}; do curl localhost:8200/v1/errors -w "$i - %{http_code}\n" -d @tests/data/valid/error/minimal_payload_log.json  -H "Content-Type: application/json" ; done`:

```
0 - 200
1 - 200
2 - 200
3 - 200
4 - 200
5 - 200
6 - 200
...
79 - 200
{"error":"Error adding data to internal queue: Queue is full"}80 - 503
```

roncohen:
Great work @urso. I'm ready to merge unless @ruflin insists on lower case error messages ;)

roncohen:
There's a couple of test failures left

ruflin:
@ruflin does not insist ;-)

roncohen:
@urso i fixed the tests here: https://github.com/roncohen/apm-server/commit/dbfe10959c15c2249ac2b503924a83460163335a

not super happy about that code, so feedback is appreciated

urso:
@ron thanks for the commit. I adapted server.go to return 202 and added the additional request body check from you commit. For fixing the tests I decided to 'lower the resistence' by introducing a `sendStatus` function which could be tested in isolation.

As I tried to understand the tests I started refactoring them a little. With this change in place the request handling and response writing can be tested more in isolation. Will cleanup tests and add some more tests in followup PR.

> lower case error messages 

Well, I didn't want to make the call in this PR. As these error messages are used in the response as is, I did decide to keep them uppercase.

ruflin:
@urso Could you also squash your commits and rebase?

roncohen:
Good work on the refactored SSL tests. Happy to merge when squashed!

jalvz:
It would be good to have some test for this...

PR #120, opened 2017-09-01
Set default number of shards to 1

ruflin:
This change sets the number of shards to 1 in the standard config file. That means if someone does not change the configuration, each index will have one shard. This should improve the getting started experience. The default in elasticsearch is 5. More experienced users can change the value to their needs when going into production.

The same is done here for metricbeat: https://github.com/elastic/beats/blob/master/metricbeat/_meta/setup.yml#L14

To not have the default changed also in the reference file which represent the internal defaults, the `beat.reference.yml` file is introduced which is used as base for the `apm-server.reference.yml`. Configs which are not critical for the getting started were removed from the `beat.yml` file.


ruflin:
@beniwohli Jep, on my list for the dev guide.

PR #123, opened 2017-09-01
Moving server to a struct

ruflin:
This is some refactoring for the server object. I reduces the number of params that have to be passed to each function by more relying on the config object. Also this allows auto-complete in the IDE so see which methods are available on `server`.

* Move handler logic to its own file
* Make error handling local to server run method

No functional changes were made in this PR.

### DIFF
--- a/beater/handler.go
+++ b/beater/handler.go
@@ -1,0 +1,172 @@
+package beater
+
+import (
+	"compress/gzip"
+	"compress/zlib"
+	"crypto/subtle"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"net/http"
+	"strings"
+
+	"github.com/elastic/apm-server/processor"
+	"github.com/elastic/beats/libbeat/logp"
+)
+
+var (
+	errInvalidToken    = errors.New("Invalid token")
+	errPOSTRequestOnly = errors.New("Only post requests are supported")
+)
+
+type handler func(w http.ResponseWriter, r *http.Request)
+
+func createHandler(p processor.Processor, config Config, report reporter) handler {
+	return func(w http.ResponseWriter, r *http.Request) {
+		logp.Debug("handler", "Request: URI=%s, method=%s, content-length=%d", r.RequestURI, r.Method, r.ContentLength)
+		requestCounter.Inc()
+
+		code, err := processRequest(r, p, config, report)
+		sendStatus(w, r, code, err)
+	}
+}
+
+func processRequest(r *http.Request, p processor.Processor, config Config, report reporter) (code int, err error) {
+	if !checkSecretToken(r, config.SecretToken) {
+		return reportInfo(401, errInvalidToken)
+	}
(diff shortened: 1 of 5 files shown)


crackerplace:
+1 for moving the handle logic to separate file as I was thinking of doing that.

beater/server.go
```diff
@@ -29,12 +19,12 @@ var (
...
 type reporter func([]beat.Event) error
 
-var (
-	errInvalidToken    = errors.New("Invalid token")
-	errPOSTRequestOnly = errors.New("Only post requests are supported")
-)
+type server struct {
+	http.Server
+	config Config
+}
```
simitt: TBH this feels a bit like back & forth, because it is reverting some of the changes in PR https://github.com/elastic/apm-server/pull/78. I don't see the additional value of reverting this.
I generally still like the idea of grouping together methods, but as @jalvz pointed out earlier it does not make too much sense in this special case, since the `http.Server` offers the methods and it would basically just be a wrapper. Storing the config in the server stores quite some values as duplicates and adds more state to the server object. 
Could you quickly motivate this change a bit more so I can understand the value that it adds?

beater/handler.go
```diff
@@ -0,0 +1,172 @@
...
+)
+
+var (
+	errInvalidToken    = errors.New("Invalid token")
+	errPOSTRequestOnly = errors.New("Only post requests are supported")
+)
+
+type handler func(w http.ResponseWriter, r *http.Request)
+
+func createHandler(p processor.Processor, config Config, report reporter) handler {
```
simitt: Generally I like the idea of moving the handler part to its own file. When you move the logic, you should definitly also move the tests. This could be the tricky part since it is quite strictly coupled. 
So from my side +1 on seperating the logic, but not without the tests. I'd see this as a basis to see where we still have a lot of coupling, mixing and holding states instead of passing in values, which makes decoupling and testing harder.

beater/server.go
ruflin: The motivation is to have all the server logic "hidden" inside server. Config is only passed once on setup and from the on it's up to the server to decide when and where the config is needed.

Second part is the error handling. There was knowledge outside the server about how to handle specific errors which I don't think beater should know about. 

In general I think it's important that we can change the server without having to change anything in the beater `Run` method. 

The server object is composed of the http.Server. In my opinion this adds value as it simplifies the interface to interact with the server. 


For the duplicated config values: As the server instance only exists one I don't think this should be an issue.

beater/handler.go
ruflin: Will do if we agree on moving forward with this.

ruflin:
After a conversation with @simitt we decided to have a Github meta issue related to refactoring tasks that we want to do so we can discuss there which tasks we actually see relevant and which ones not: https://github.com/elastic/apm-server/issues/128

For now closing this PR as it touches code base created by @jalvz and it requires his input on the changes. Also the changes on the handler extractions should be separated.

PR #130, opened 2017-09-04
Update beats in 6.x to 1ee5c98

ruflin:
This makes the 6.x branch depend on the 6.x beats branch. Currently 6.x is almost identical to master in beats except with some delay.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -18,5 +18,5 @@ var RootCmd *cmd.BeatsRootCmd
 
 func init() {
 	var runFlags = pflag.NewFlagSet(Name, pflag.ExitOnError)
-	RootCmd = cmd.GenRootCmdWithRunFlags(Name, "7.0.0-alpha1", beater.New, runFlags)
+	RootCmd = cmd.GenRootCmdWithRunFlags(Name, "", beater.New, runFlags)
 }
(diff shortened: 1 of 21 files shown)


cmd/root.go
```diff
@@ -18,5 +18,5 @@ var RootCmd *cmd.BeatsRootCmd
 
 func init() {
 	var runFlags = pflag.NewFlagSet(Name, pflag.ExitOnError)
-	RootCmd = cmd.GenRootCmdWithRunFlags(Name, "7.0.0-alpha1", beater.New, runFlags)
+	RootCmd = cmd.GenRootCmdWithRunFlags(Name, "", beater.New, runFlags)
```
simitt: This uses a fallback to use the `default_version` of beats, which currently is `6.1.0`, so this is aligned with the other changes. However, why is the default version in master not set to `7.0.0-alpha` when everything else is?

ruflin: Master should be set to `7.0.0-alpha1` automatically that is why I removed https://github.com/elastic/apm-server/pull/132 here. Where in master is it not set to 7.0.0-alpha1?

simitt: I have not had updated to this https://github.com/elastic/apm-server/pull/129/files#diff-942011c37b00c68bed7206f659059385.

PR #131, opened 2017-09-04
Update beats to 1ee5c9 in 6.0 branch

ruflin:
A small trick is applied to still get beats code from the 6.x branch but have the versions from the 6.0 branch. The two required files are fetched seperately. The changes were made in the code instead of modifying the two files manually to be able to just run `make update-beats` also in the future without overwriting it again.

### DIFF
--- a/_beats/libbeat/docs/version.asciidoc
+++ b/_beats/libbeat/docs/version.asciidoc
@@ -1,4 +1,4 @@
-:stack-version: 7.0.0-alpha1
+:stack-version: 6.1.0
 :doc-branch: master
 :go-version: 1.8.3
 :release-state: unreleased
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@ TEST_ENVIRONMENT=true
 ES_BEATS?=./_beats
 PREFIX?=.
 NOTICE_FILE=NOTICE
-BEATS_VERSION?=master
+BEATS_VERSION?=6.x
 
 # Path to the libbeat Makefile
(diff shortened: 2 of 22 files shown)


_beats/libbeat/docs/version.asciidoc
```diff
@@ -1,4 +1,4 @@
-:stack-version: 7.0.0-alpha1
+:stack-version: 6.1.0
```
simitt: I assume this is on purpose? So in the docs we already link to the `6.1.0` version although we are in the `6.0` version right?

ruflin: We could do that in the docs build if we want, but for now there is only a docs build for master. The docs build is independent of the unified release build.

simitt: This means that this version here does not cause anything, because docs are only built in master?
Just wanted to ensure that it is on purpose to point to `6.1.0` here instead of `6.0.x`.

ruflin: Currently for apm-server we only build the master branch: https://github.com/elastic/docs/blob/master/conf.yaml#L728 Assuming we would add the 6.0 branch for building, the doc build would still work but all links would point to 6.1. 

The reason it points to 6.1 is because we also use the 6.x beats branch here. From a docs perspective it's not "correct" but as I don't think we plan to do a doc build based on 6.0 I would not fix this issue.

simitt: ok, thx for additional explanation

PR #134, opened 2017-09-04
Add beta log message to 6.x branch.

ruflin:


beater/beater.go
```diff
@@ -16,6 +17,7 @@ type beater struct {
 
 // Creates beater
 func New(_ *beat.Beat, ucfg *common.Config) (beat.Beater, error) {
+	cfgwarn.Beta("You are using an beta release of apm-server.")
```
simitt: `a beta release`, again does it make sense to use `APM Server instead?

ruflin: will change it accordingly.

roncohen:
LGTM 👍

Issue #128, opened 2017-09-04
Refactoring / Cleanup apm-server


ruflin:
This meta issue should help to discuss and track refactorings which we want to do in apm-server to improve the code base. Each person should update and edit the tasks on this list.

* [ ] Improve handler testing
* [ ] Improve server and config handling
* [ ] ...

simitt:
I suggest to close this meta issue, as I think we are at a pretty good stage right now and don't need a specific meta issue for future refatoring. 
What do you think @roncohen , @graphaelli , @jalvz ?

PR #135, opened 2017-09-04
Chain handlers

crackerplace:
@ruflin 

Refactored the handler logic to use handler chaining.I did not move the handlers to separate file assuming that's something being discussed in parallel.

Not sure if I should include this change in #128.

### DIFF
--- a/beater/server.go
+++ b/beater/server.go
@@ -89,6 +89,7 @@ func stop(server *http.Server, timeout time.Duration) {
 
 func authHandler(secretToken string, h http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		requestCounter.Inc()
 		if !isAuthorized(r, secretToken) {
 			sendStatus(w, r, 401, errInvalidToken)
 			return
@@ -100,7 +101,6 @@ func authHandler(secretToken string, h http.Handler) http.Handler {
 func appHandler(p processor.Processor, config Config, report reporter) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		logp.Debug("handler", "Request: URI=%s, method=%s, content-length=%d", r.RequestURI, r.Method, r.ContentLength)
-		requestCounter.Inc()
 		code, err := processRequest(r, p, config.SecretToken, config.MaxUnzippedSize, report)
 		sendStatus(w, r, code, err)
 	})
--- a/beater/server.go
+++ b/beater/server.go
@@ -97,8 +97,8 @@ func logHandler(h http.Handler) http.Handler {
 func authHandler(secretToken string, h http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		if !isAuthorized(r, secretToken) {
-			logp.Err(errInvalidToken.Error())
-			sendStatus(w, r, 401, errInvalidToken)
+			code, err := reportInfo(401, errInvalidToken)
+			sendStatus(w, r, code, err)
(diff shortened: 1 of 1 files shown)


ruflin:
@crackerplace For the part of moving the handler to an other file: I think we are all on the same page that it should move to it's own file but during doing this, also the tests should be cleaned up which could be a bigger task then expected.

Thanks for the PR. Will have a look tomoroow.

crackerplace:
@ruflin Having the handlers in the same file should also be fine as we don't have many handlers to deal with and can move once we get there so as to avoid more complexity from the points that you have mentioned.

beater/server.go
```diff
@@ -157,29 +159,18 @@ func reportInfo(code int, err error) (int, error) {
...
 }
 
-// checkSecretToken checks the Authorization header. It must be in the form of:
-//
+// isAuthorized checks the Authorization header. It must be in the form of:
 //   Authorization: Bearer <secret-token>
-//
 // Bearer must be part of it.
-func checkSecretToken(req *http.Request, secretToken string) bool {
-	// No token configured
```
ruflin: Can you bring this comment back?

beater/server.go
```diff
@@ -157,29 +159,18 @@ func reportInfo(code int, err error) (int, error) {
...
+func isAuthorized(req *http.Request, secretToken string) bool {
```
ruflin: I like the renaming as it's more generic.

beater/server.go
```diff
@@ -38,12 +38,9 @@ func newMuxer(config Config, report reporter) *http.ServeMux {
...
-		handler := createHandler(p, config, report)
-
+		handler := appHandler(p, config, report)
 		logp.Info("Path %s added to request handler", path)
-
-		mux.HandleFunc(path, handler)
+		mux.Handle(path, authHandler(config.SecretToken, handler))
 	}
 
 	mux.HandleFunc("/healthcheck", func(w http.ResponseWriter, r *http.Request) {
```
ruflin: This change just made me realise that we don't check the token for the `/healthcheck` handler I think?

beater/server.go
```diff
@@ -90,16 +87,23 @@ func stop(server *http.Server, timeout time.Duration) {
...
+			sendStatus(w, r, 401, errInvalidToken)
+			return
+		}
+		h.ServeHTTP(w, r)
+	})
+}
 
+func appHandler(p processor.Processor, config Config, report reporter) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		logp.Debug("handler", "Request: URI=%s, method=%s, content-length=%d", r.RequestURI, r.Method, r.ContentLength)
```
simitt: With this change a request is is not logged any more when it is not authorized.

roncohen:
Nice work @crackerplace. How about adding a handler for counting and one for logging?

beater/server.go
crackerplace: This looks fine for me.

beater/server.go
crackerplace: yes,I was looking for a better way to log.Could not use reportInfo in sendStatus as I wanted to.

crackerplace:
@roncohen Sure.Added a log handler.Logging the metric counter from the same handler as well.Let me know if you think we need a separate counting handler.

crackerplace:
@ruflin Do we need responseValid ?

I somehow think the request rate and error rate serve the purpose.The other point being tracking all the return points in case we want to account for a successful response and increment responseValid at that point might become tricky.

beater/server.go
crackerplace: done

PR #136, opened 2017-09-06
Clean up intake API JSON schema indentation

watson:
We should consider a linter 😃

### DIFF
--- a/processor/transaction/schema.go
+++ b/processor/transaction/schema.go
@@ -32,11 +32,11 @@ var transactionSchema = `{
             "required": ["name", "version"]
         },
         "argv": {
-            "type": ["array","null"],
+            "type": ["array", "null"],
             "minItems": 0
         },
         "framework": {
-            "type": ["object","null"],
+            "type": ["object", "null"],
             "properties": {
                 "name": {
                     "type": "string",
@@ -55,7 +55,7 @@ var transactionSchema = `{
             "maxLength": 1024
         },
         "language": {
-            "type": ["object","null"],
+            "type": ["object", "null"],
             "properties": {
                 "name": {
                     "type": "string",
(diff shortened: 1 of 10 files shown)


processor/transaction/schema.go
```diff
@@ -135,74 +135,74 @@ var transactionSchema = `{
...
-      },
-      "additionalProperties": false
-    },
-    "response": {
-        "type": ["object", "null"],
-        "properties": {
-            "finished": {
-                "type": ["boolean", "null"]
+                "$schema": "http://json-schema.org/draft-04/schema#",
+    "$id": "doc/spec/context.json",
```
watson: I can `make fmt` to update this file, but it seems to mess up a few indentation

ruflin: I don't think `make fmt` cleans up json files. It's the `.editorconfig` which should take care of the indentation: https://github.com/elastic/apm-server/blob/master/.editorconfig#L10

But that does not do any linting.

ruflin:
@watson Do you know a simple linter we could use for json files? Currently IntellJ is my linter for local files :-)

processor/transaction/schema.go
watson: This isn't a JSON file, but a go file which was generated automatically by running `make fmt`.

watson: I ran it because the tests were failing and complaining about changes to the schema. I looked in the README and it suggested running `make fmt`. After I ran that the tests passed.

watson:
@ruflin I only know [eslint](https://www.npmjs.com/package/eslint-plugin-json), which is a JavaScript linter. But that should be possible to have it lint json files as well using [eslint-plugin-json](https://www.npmjs.com/package/eslint-plugin-json)

processor/transaction/schema.go
simitt: The schema in this file is created automatically and does not care about intentation. If you think this should be fixed we can create an Issue to improve this.

simitt:
Thanks @watson for taking the time to lint these files.

PR #144, opened 2017-09-08
Change timestamp fields type to time.Time

### LINKED ISSUE #137: Untestable transformation error
An error in `Transform` makes server to return a 500 error:
https://github.com/elastic/apm-server/blob/master/beater/server.go#L137

This has 2 issues:

+ Semantic: definition of 500 is "The server failed to fulfil a valid request", but in our specific case a potential error returned by `Transform` would happen with an _invalid_ request.

+ Not testable: https://github.com/elastic/apm-server/blob/master/beater/server.go#L139 is not tested, and we don't know under which conditions might be executed, so we can not even test it end to end (without mocks).

### LINKED ISSUE #141: APM Server accepts invalid timestamps
If an agent sends a timestamp with eg. month and day swapped: 
+ APM Server will accept it because it conforms the regex
+ APM Server will drop the event because it can't parse it
+ 202 will be returned

jalvz:
Also make server to return 400 on transformation errors.

This relies on default Golang Unmarshaling, so we don't need to parse timestamps ourselves. It makes the error returned in transformation useful and testable.

Closes #141

Closes #137


processor/error/payload_test.go
```diff
@@ -80,7 +78,7 @@ func TestPayloadTransform(t *testing.T) {
 		outputEvents := test.Payload.transform()
 		for j, outputEvent := range outputEvents {
 			assert.Equal(t, test.Output[j], outputEvent.Fields, fmt.Sprintf("Failed at idx %v; %s", idx, test.Msg))
-			assert.Equal(t, expectedTime, outputEvent.Timestamp)
+			assert.WithinDuration(t, time.Now(), outputEvent.Timestamp, time.Second)
```
simitt: I suggest to rather store `Time.now()` as a variable somewhere and check the exact time. This looks like it could become a flaky test very easily. 

Do you know the `timecop` tools? I know it from ruby, but saw that there is also a [go-timecop](https://github.com/bluele/go-timecop) implementation. This can be quite useful when testing something time related (probably not necessary here but worth to mention.)

processor/transaction/payload_test.go
```diff
@@ -159,7 +156,7 @@ func TestPayloadTransform(t *testing.T) {
 		outputEvents := test.Payload.transform()
 		for j, outputEvent := range outputEvents {
 			assert.Equal(t, test.Output[j], outputEvent.Fields, fmt.Sprintf("Failed at idx %v; %s", idx, test.Msg))
-			assert.Equal(t, expectedTime, outputEvent.Timestamp)
+			assert.WithinDuration(t, time.Now(), outputEvent.Timestamp, time.Second)
```
simitt: same as for testing the `error payload`

processor/error/payload_test.go
jalvz: done!

processor/transaction/trace.go
```diff
@@ -40,7 +42,7 @@ func (t *Trace) Transform(transactionId string) common.MapStr {
 	return tr
 }
 
-func (t *Trace) Mappings(pa *payload, tx Event) (string, []m.DocMapping) {
+func (t *Trace) Mappings(pa *payload, tx Event) (time.Time, []m.DocMapping) {
```
simitt: shouldn't we still return a `string` that we parse out of the `time.Time` object with the proper `time.RFC3339Nano` format that we agreed on?

jalvz: When you have a `time.Time` in a struct that you unmarshal into, Golang automatically converts `string` to `RFC3339Nano` already. No need to do it ourselves.

https://play.golang.org/p/ShKw8ZKqAC

roncohen:
good work @jalvz!

roncohen:
@jalvz could you packport to `6.0` ?

PR #145, opened 2017-09-08
Remove timestamp validation regex

jalvz:
Depends on #144


docs/spec/errors/error.json
```diff
@@ -91,8 +91,7 @@
         },
         "timestamp": {
             "type": "string",
-            "description": "Recorded time of the error, UTC based and formatted as YYYY-MM-DDTHH:mm:ss.sssZ",
-            "pattern": "^(\\d{4})-(\\d{2})-(\\d{2})T(\\d{2}):(\\d{2}):(\\d{2})(\\.\\d{1,6})?Z$"
+            "description": "Recorded time of the error, UTC based and formatted as YYYY-MM-DDTHH:mm:ss.sssZ"
```
simitt: Why are you removing this pattern? When we use a json schema we should also validate input against it, to have as much validation as possible in one place and it also serves as documentation. 
I'd rather go the other direction and use a more thorough regex here then the one we started with. We had lots of discussions on how strict we should build the json schema and if we also want semantic validation here or not. However, in my opininon the goal should be to put everything intake related that could cause errors or where we want to handle errors, already into the schema because I don't see the value of having a schema otherwise.

jalvz: I thought of removing this because:
+There is no way to create a (reasonable) regex that validates a date (think of leap years etc)
+This means that we need to validate the timestamp when we parse it no matter what.
+If we are validating the timestamp when we parse it, then there is no point in doing it twice, 1 with regex and 1 with unmarshal.

But if you think is better to keep it duplicated i'm ok with that as well (that is why i did this in a different PR). I agree with general thinking on json schema, is just that a regex is not good enough for this.

jalvz:
Actually let me close this, i think there is a better way. Let me know if #144 is alright

PR #146, opened 2017-09-08
Update to leatest beats master

simitt:


ruflin:
There is something of with the CI in this PR. I tries to download and test 6.0.0-beta1 but I would expect it to use 7.0.0-alpha1. Will investigate why.

ruflin:
The issue is that as soon as a release is out the snapshot builds are removed from S3. For master this normally only happens every 12 months as it points to the next major. But now with 6.0 coming out this just happened. I opened https://github.com/elastic/beats/pull/5142 to point to the next major version. As soon as it is merged, we need to update beats again.

PR #147, opened 2017-09-08
Load dashboards update

makwarth:
- Replaced manual curl command with ./apm-server setup

cc @ruflin @roncohen @simianhacker


ruflin:
That is definitively the way loading should work in the future. The `setup` commands loads all the dashboards. There is currently one issue with it in apm-server: We use script fields in the index pattern which are not supported by `fields.yml` yet. So if you use setup, it will load the dashboards and generate the index pattern from the apm-server side. This will break some of the dashboards as 2 script fields are missing in the index pattern.

makwarth:
@ruflin Right, but they will be supported in 6.0?

ruflin:
Here is the change that needs to happen first on the beats side: https://github.com/elastic/beats/issues/4899

simitt:
As everything needed is now implemented in `beats` and `apm-server` we could reopen this PR, but the remote branch is already deleted. @makwarth do you still have the local branch? Otherwise I can just create a new one..

makwarth:
Great!

>ruflin deleted the makwarth-patch-1 branch 14 days ago

^ Should I restore the branch @simitt?

simitt:
@makwarth could you please rebase and solve the conflicts?

PR #148, opened 2017-09-11
Override `make index-template` to rename ES template.

simitt:
Generate ES template called `apm` instead of `apm-server`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,6 +18,10 @@ update-beats:
 	BEATS_VERSION=$(BEATS_VERSION) sh _beats/update.sh
 	$(MAKE) update
 
+.PHONY: index-template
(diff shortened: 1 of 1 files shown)


ruflin:
Be aware that the `make index-template` command is mainly used for development purpose. The template itself is generated and loaded dynamically based on the `fields.yml` file and the params passed to it there. I assume the end goal is to use `apm-*` index, templates and index pattern? I assume it will involve quite a few changes which need to happen at the same time to not break things.

I'm assuming in case `index-template` command already exists, you will get an "overload" error in the Makefile.

simitt:
I got an overload info but it still worked

simitt:
I will close this in favor of #152

PR #149, opened 2017-09-11
SSL docs

jalvz:


jalvz:
This is checked out from Ron's branch. @roncohen do you intend to create a PR or merge your branch?

roncohen:
Thanks @jalvz! We should get the other PR merged as soon as possible.

Is this SSL or TLS btw.? I found this:

```Current browsers do not support SSL 2.0. Transport Layer Security (TLS) is the successor to SSL. TLS 1.0 was defined in RFC 2246 in January 1999. The differences between TLS 1.0 and SSL 3.0 were significant enough that they did not interoperate```

roncohen:
Ended up making a new PR with the same commit to get this in today: https://github.com/elastic/apm-server/pull/173

PR #150, opened 2017-09-11
Use format instead of a regex in json schema to validate timestamps 

jalvz:
Depends on #144


jalvz:
@simitt this validates date more accurately than a regex, see http://json-schema.org/latest/json-schema-validation.html#rfc.section.8.3.1

roncohen:
this effectively changes the API right? this has bigger implications and we had a very long discussion on why we restricted it to this specific format earlier. For example, this allows timezones again after we specifically agreed to disallow it.

jalvz:
right, didn't consider timezone. ill fix and write a test 👍

PR #152, opened 2017-09-11
Set index prefix to `apm` instead of `apm-server`

simitt:
Change index prefix from `apm-server` to `apm`.

This is based on https://github.com/elastic/beats/pull/5161, and therefore in progress, because the beats framework needs to be updated before this has any affect.


roncohen:
great work!
This probably has an impact on dashboards, so we should be careful about merging before dashboards gets an update.

simitt:
Can only be merged after https://github.com/elastic/beats/pull/5161

simitt:
The PR https://github.com/elastic/beats/pull/5161 in the beats framwork was merged. After updating to the latest beats master and applying the changes of this PR the default index name is changed to `apm-*` instaed of `apm-server-*`.

PR #154, opened 2017-09-11
Don't allow timezone offsets for timestamps

jalvz:
This complements `date-format` used for proper date validation with a `Z$` regex to make sure that timezone offsets are not allowed (as per spec)

### DIFF
--- a/docs/spec/errors/error.json
+++ b/docs/spec/errors/error.json
@@ -92,6 +92,7 @@
         "timestamp": {
             "type": "string",
             "format": "date-time",
+            "pattern": "Z$",
             "description": "Recorded time of the error, UTC based and formatted as YYYY-MM-DDTHH:mm:ss.sssZ"
         }
     },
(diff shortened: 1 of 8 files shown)


jalvz:
Ah, forgot about this. @roncohen wasn't very happy with the implementation. 

Logic is duplicated (same thing is done in a298a0d4039d7df6c854ab37fda5c540542b18a9), but I think is better to have *independent* duplicated logic, than half-right logic implicitly relying on the other part, that is why i did this.

If you guys think otherwise you can revert 52b48dc41d5deafb38ea29c8c87bae9cab3e0e8b and this.

simitt:
Don't see where this duplicates more logic than before. All the validation happens in the `json schema`. Since we cannot use this to also transform into our internal data structure there is a second step anyways that does the transformation. In this step theoretically an error could occur, but should never happen. 

But I was not aware of ongoing discussions about this, so sorry if I merged something too soon. @roncohen please let us know if this is ok with you, otherwise we can just revert the merge commit.

roncohen:
No big deal. This is fine 👍

PR #156, opened 2017-09-11
Ensure libbeat infrastructure has time to run before returning 503

roncohen:
In benchmarks (we've not reproduced this outside, yet), we sometimes get a full-queue even though we've using a "null" output. This happens because the go scheduler is not running the go routines that take events out of the queue as much time as we'd like it to. By introducing a one second timeout, this is an attempt at giving the go runtime some time to run those go routines some time to run. 

I'm not a big fan of the UX of taking an extra second to return a 503 to the user in the case where the queue is really full, but it might be the best solution for now.

### DIFF
--- a/beater/pub.go
+++ b/beater/pub.go
@@ -3,6 +3,7 @@ package beater
 import (
 	"errors"
 	"sync"
+	"time"
 
 	"github.com/elastic/beats/libbeat/beat"
 )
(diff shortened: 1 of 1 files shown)


roncohen:
btw. this is closer to urso's original experiment that also used a timeout.

roncohen:
we could probably set `ConcurrentRequests` quite low with this as incoming requests will now allow the libbeat code to spend up to a second to accept the data in the local queue

PR #161, opened 2017-09-12
Fix the dockerized python app for manual testing

jalvz:


ruflin:
Thanks for taking a stab at this one. The problem happened in this PR that opbeans was accidentially renamed to apm-server: https://github.com/elastic/apm-server/pull/60/files#diff-e9420078c93ff1a0cc39dbedaa4e6946L8 This made the current code very confusing.

My goal with the docker image was to fetch and load opbeans and then run it. I did an alternative change here: https://github.com/elastic/apm-server/compare/master...ruflin:fix-opbeans-setup?expand=1 The change is working in that I get a container running with opbeans. The part that is not working yet is on how do I get this container to send data to my local apm-server instance. 

Unfortunately all the above is not documented in the README of the python agent testing directory. It only mentions the simple app which I normally run directly locally and you packaged above into the Docker container. So there are actually 2 test environments.

simitt:
Is this going to be closed in favor of https://github.com/elastic/apm-server/pull/180?

beniwohli:
Yes, good point!

simitt:
@jalvz I think all the fixes should have been merged with the PR #180, thus I am closing this one.

PR #169, opened 2017-09-14
Fix link from [APM] Applications to [APM] Transactions.

simitt:
Context.app.name needs to be passed as string.

This is the part that actually changed: `query:(language:lucene,query:'context.app.name:\\\"{{value}}\\\"'`.

If you go to the [APM] Transactions dashboard from the [APM] Applications dashboard, the `context.app.name` is passed as a filter parameter. This needs to be passed with enclosing `""` as otherwise no transactions for the app can be found if the app name includes a space. 

This is an edge case as it only exists when spaces are used in the app_name.
It also concerns the 6.0 branch - @ruflin is it still possible to backport this after @simianhacker has reviewed it?


roncohen:
Yeah, let's get this backported.

PR #170, opened 2017-09-14
Avoid cmd output from Makefile.

simitt:
Only Info, Warning and Errors should be printed, instead of every
command.


simitt:
@ruflin would you mind reviewing this?

PR #171, opened 2017-09-14
Fix wrong content-type in response.

### LINKED ISSUE #167: Use correct content type for JSON responses
As @watson points out in [#166](https://github.com/elastic/apm-server/issues/166#issuecomment-329136461), the server currently always returns a `Content-Type` header `text/plain`, even if the response is JSON.

simitt:
WriteHeader(code) needs to be called after headers are set.
closes #167.


roncohen:
Good work!

PR #174, opened 2017-09-14
Moving APM Server specific reference docs into separate directory

roncohen:


roncohen:
Parking this for now as it needs some beats involvement and it's not really critical.

PR #180, opened 2017-09-15
Fix broken opbeans setup

beniwohli:
This also adds services for both flask and django test apps to docker-compose

### DIFF
--- a/tests/agent/python/Makefile
+++ b/tests/agent/python/Makefile
@@ -1,9 +1,7 @@
-setup:
-	pip install -Ur requirements.txt
-
 start:
 	docker-compose build
-	docker-compose up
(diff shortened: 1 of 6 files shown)


tests/agent/python/Makefile
```diff
@@ -1,9 +1,7 @@
-setup:
-	pip install -Ur requirements.txt
-
 start:
 	docker-compose build
-	docker-compose up
+	IP=$$(ifconfig | grep -E "([0-9]{1,3}\.){3}[0-9]{1,3}" | grep -v 127.0.0.1 | awk '{ print $$2 }' | cut -f2 -d: | head -n1) docker-compose up
```
ruflin: Nice trick

PR #181, opened 2017-09-15
Minor logging fixes

jalvz:
+ Logs `202` responses with debug level
+ Fixes monitoring counting (503 were counted as valid _and_ invalid)
+ Removes unused param and named return args from method
+ Removes duplicated messages from errors (eg: something like "data validation error" is already in the original error, there is no need to wrap it)
+ Simplifies code a bit

### DIFF
--- a/beater/server.go
+++ b/beater/server.go
@@ -30,8 +30,8 @@ var (
 type reporter func([]beat.Event) error
 
 var (
-	errInvalidToken    = errors.New("Invalid token")
-	errPOSTRequestOnly = errors.New("Only post requests are supported")
+	errInvalidToken    = errors.New("invalid token")
+	errPOSTRequestOnly = errors.New("only POST requests are supported")
 )
 
 func newMuxer(config Config, report reporter) *http.ServeMux {
@@ -97,8 +97,7 @@ func logHandler(h http.Handler) http.Handler {
 func authHandler(secretToken string, h http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		if !isAuthorized(r, secretToken) {
-			code, err := reportInfo(401, errInvalidToken)
-			sendStatus(w, r, code, err)
+			sendStatus(w, r, 401, errInvalidToken)
 			return
 		}
 		h.ServeHTTP(w, r)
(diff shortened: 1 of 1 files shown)


beater/server.go
```diff
@@ -225,9 +207,13 @@ func sendStatus(w http.ResponseWriter, r *http.Request, code int, err error) {
 	w.WriteHeader(code)
 
 	if err == nil {
+		responseValid.Inc()
+		logp.Debug("request", "request successful, code=%d", code)
 		return
 	}
 
+	logp.Err("%s, code=%d", err.Error(), code)
```
jalvz: Since at this point I know `err` is not `nil`, semantically makes sense to log it with `logp.Err`, whatever it is.
This reverts a change @ruflin did a while ago, which checked if `code` is `4XX`, and in that case logs it as Info.

Argument IIRC was to not flood log with too much stuff. I don't think that is a concern, but I was ok with that at the time. 
Now I suggest to change it back mostly to keep the code simpler and afford an `if` branch that doesn't give any benefit. Semantically, a `4XX` is still an error, even if the culprit is not the server.

beater/server.go
```diff
@@ -135,34 +128,23 @@ func processRequest(
 	buf, err := ioutil.ReadAll(limitedReader)
 	if err != nil {
 		// If we run out of memory, for example
-		return reportError(500, fmt.Errorf("Data read error: %s", err))
+		return 500, err
```
simitt: You remove the additional error information here. In case this info is not already reflected in the original error (I haven't checked that) you could do something like 
`return 500, errors.New(fmt.Sprintf("xyz %s", err.Error()))`

jalvz: You mean the "Data read error" part? that was intended, for what i could see, errors are more or less self-descriptive (eg "unexpected EOF").
Do you think is needed to wrap it?

simitt: no as I said, I did not look into the original error - if you checked, it is fine with me.

ruflin: The more descriptive errors were added on the one hand for the users but also to quickly nail down where the error exactly happened. `unexpected EOF` could happen in a few places.

beater/server.go
```diff
@@ -135,34 +128,23 @@ func processRequest(
...
 	}
-	return 202, nil
-}
-
-func reportError(code int, err error) (int, error) {
-	logp.Err(err.Error())
-	return code, err
-}
 
-func reportInfo(code int, err error) (int, error) {
```
ruflin: This reverts the cleanup which was done by @urso in https://github.com/elastic/apm-server/pull/118 which I quite liked.

beater/server.go
ruflin: From an operator perspective I think an error means there is action needed to "fix" the apm-server. But return a 400 error can happen pretty often especially when exposed to the internet without requiring an action from the operator. I would still keep the old logging.

beater/server.go
jalvz: done 499ba3e31c72a174cfa12e4f69b2177e0a8e3668

beater/server.go
jalvz: I don't think that is true, biggest cleanup parts are `processRequest` `sendStatus`, and `sendJSON` which don't change here. This replaces 2 functions just performing a side effect with a single line here: https://github.com/elastic/apm-server/pull/181/files#diff-f913ad38aa36e2c00a5c7f20e5ec3d20R216

Is it wrong?

simitt: I see this as a further improvement of @urso 's nice cleanup. He cleaned up how we manage errors and extraced two functions to call. Now those two functions are called automatically when returning the response. It is a revert in the sense that those functions are removed again, but the code is not put back to where it was before, but rather enhanced the idea from urso (at least this is how I understand it).

beater/server.go
jalvz: By experience, i can tell you it does *not happen often. 

And if it happens, it usually means that there is a bug in a client, so it needs fixing as well.

Note that agent owners don't have the chance to collect errors from the agent, they need either to get it reported from the end user or from the apm server operator, so is nice if we can highlight it on this side, don't you think?

ruflin: TBH I didn't run apm-server in production yet so that is probably the only way to tell us :-) Both works for me. We are on the same page that if the client sends data which leads to a 404 we should log it as error. Let's go with error and see how it works out in production 👍

PR #183, opened 2017-09-18
Update linking to json files in docs.

simitt:


docs/elasticsearch-docs.asciidoc
```diff
@@ -1,21 +1,38 @@
 [[elasticsearch-docs]]
 == Elasticsearch
 
-The documents produced by APM Server have a specific structure.
-There are examples of such documents as well as documentation 
-
+The APM Server produces documents with a specific structure.
+Based on the defined structure a template mapping can be used for Index creation in Elasticsearch.
```
dedemorton: There is no reason to capitalize index here.

docs/elasticsearch-docs.asciidoc
```diff
@@ -1,21 +1,38 @@
...
+See examples below for different document types.
```
dedemorton: I would say "See the examples in this section for different document types."

docs/elasticsearch-docs.asciidoc
```diff
@@ -1,21 +1,38 @@
...
+[source,json]
+----
+include::./data/elasticsearch/error.json[]
+----
 
 
+===  Indexed Fields
 
-include::./fields.asciidoc[See indexed fields in Elasticsearch]
+See which fields are indexed in Elasticsearch:
```
dedemorton: Instead of adding a section, it would be better to add a cross reference that points to the section about exported fields. When you look at the output, you get a topic that contains hardly any content.

image

I don't think there's any real value in this section. Maybe at the top of the file (before you provide examples), you should have a sentence that says:  "To see which fields are indexed in Elasticsearch, see Exported Fields" (make Exported Fields an active link).

docs/elasticsearch-docs.asciidoc
```diff
@@ -1,21 +1,41 @@
 [[elasticsearch-docs]]
 == Elasticsearch
 
-The documents produced by APM Server have a specific structure.
-There are examples of such documents as well as documentation 
-
+The APM Server produces documents with a specific structure.
+The defined structure is used to create an index in Elasticsearch.
+To see which fields are indexed in Elasticsearch, see {ref}/exported-fields.html[Exported Fields].
```
simitt: @dedemorton it seems I am doing something wrong here with the linking, but I can't see what the issue is, could you have a look?

dedemorton: sure...looking now...

dedemorton: You still need to include fields.asciidoc. I just meant that you should remove the "Indexed field" section and intro. I didn't mean that you should remove the include directive. :-)

I think it would be better to add the include to the index.asciidoc file, though, rather than the bottom of elasticsearch-docs.asciidoc.

So add the following directive to line 30 in index.asciidoc:

include::./fields.asciidoc[]

Then you should be able to use: To see which fields are indexed in Elasticsearch, see <<exported-fields>>.

PR #184, opened 2017-09-18
Use asciidoc attributes for links to documentation

watson:


docs/index.asciidoc
```diff
@@ -1,4 +1,6 @@
 = APM Server Docs (Experimental)
+:branch: current
```
dedemorton: For the getting started, the index in the conf.yaml is set to docs/guide/apm-getting-started.asciidoc, so you need to add the branch attribute and include directive to that file, not docs/index.asciidoc.

So add the following as lines 1 and 2 of apm-getting-started.asciidoc (top of the file is better):

```
:branch: current
include::{asciidoc-dir}/../../shared/attributes.asciidoc[]
```

For docs/index.asciidoc, just move the attribute and directive to the top of the file.

watson: What's the reason for having it both in `docs/guide/apm-getting-started.asciidoc` and in `docs/index.asciidoc`? Wouldn't it be sufficient to have it just on the index that's pointed to from conf.yaml?

dedemorton: Because the getting started guide is a separate book that applies to both the server and agents. We put it in the apm-server repo because we didn't have a better place, but we didn't want to make it part of the server docs because it applies to both.

watson: fixed 😃

dedemorton:
LGTM!

docs/guide/apm-getting-started.asciidoc
```diff
@@ -1,4 +1,7 @@
-[[apm-getting-started]]
+:branch: current
+include::{asciidoc-dir}/../../shared/attributes.asciidoc[]
```
ruflin: I wonder why we need this in both files? I would assume in the `index.asciidoc` header should be enough?

ruflin: @watson Ignore this, just saw the conversation with dede that `guide` is a different book.

PR #187, opened 2017-09-19
Refactor APM Server docs structure and linking.

simitt:
This is an attempt to improve the current APM Server doc experience. 

Based on additional attributes in https://github.com/elastic/docs/pull/237

### DIFF
--- a/docs/error-api.asciidoc
+++ b/docs/error-api.asciidoc
@@ -4,7 +4,8 @@
 The APM Server exposes an API Endpoint to send error records. 
 Unless you are implementing an agent, you don't need to know about the specifics of this API.
 
-To send error records you need to send a HTTP POST request to APM Server `errors` endpoint. Information pertaining to the error record must be sent as a JSON object to the endpoint.
+To send error records you need to send a HTTP POST request to APM Server `errors` endpoint. 
+Information pertaining to the error record must be sent as a JSON object to the endpoint.
 
 --
 
@@ -18,6 +19,7 @@ Find more information about:
 [[error-schema-definition]]
 [float]
 === Schema Definition
+
 The APM Server uses a JSON Schema for validating the transaction requests. 
 Find details on how the schema is defined:
 
--- a/docs/error-api.asciidoc
+++ b/docs/error-api.asciidoc
@@ -1,11 +1,25 @@
+[[error-api]]
 == Error API
+
+The APM Server exposes an API Endpoint to send error records. 
+Unless you are implementing an agent, you don't need to know about the specifics of this API.
 
 To send error records you need to send a HTTP POST request to APM Server `errors` endpoint. Information pertaining to the error record must be sent as a JSON object to the endpoint.
 
 --
-See <<error-api-examples,Examples>> for sending a request to the error endpoint.
 
-See the schema definitions for how the error payload has to look like:
+Find more information about: 
+
+* <<error-schema-definition>>
+* <<error-api-examples>>
+
+--
+
+[[error-schema-definition]]
+[float]
+=== Schema Definition
+The APM Server uses a JSON Schema for validating the transaction requests. 
+Find details on how the schema is defined:
 
 * <<error-payload-schema>>
 * <<error-error-schema>>
@@ -16,10 +30,10 @@ See the schema definitions for how the error payload has to look like:
 * <<error-request-schema>>
 * <<error-user-schema>>
 
---
 
 [[error-payload-schema]]
-=== Payload
+[float]
+==== Payload
 
 [source,json]
 ----
@@ -27,7 +41,8 @@ include::./spec/errors/payload.json[]
 ----
 
 [[error-error-schema]]
-=== Error 
+[float]
+==== Error 
 
 [source,json]
 ----
@@ -35,7 +50,8 @@ include::./spec/errors/error.json[]
 ----
 
 [[error-app-schema]]
-=== App
+[float]
+==== App
 
 [source,json]
 ----
@@ -43,7 +59,8 @@ include::./spec/app.json[]
 ----
 
 [[error-system-schema]]
-=== System
+[float]
+==== System
 
 [source,json]
 ----
@@ -51,7 +68,8 @@ include::./spec/system.json[]
 ----
 
 [[error-context-schema]]
-=== Context 
+[float]
+==== Context 
 
 [source,json]
 ----
@@ -59,7 +77,8 @@ include::./spec/context.json[]
 ----
 
 [[error-stacktraceframe-schema]]
-=== Stacktrace Frame 
+[float]
+==== Stacktrace Frame 
 
 [source,json]
 ----
@@ -67,7 +86,8 @@ include::./spec/stacktrace_frame.json[]
 ----
 
 [[error-request-schema]]
-=== Request 
+[float]
+==== Request 
 
 [source,json]
 ----
@@ -75,7 +95,8 @@ include::./spec/request.json[]
 ----
 
 [[error-user-schema]]
-=== User
+[float]
+==== User
 
 [source,json]
 ----
@@ -83,9 +104,10 @@ include::./spec/user.json[]
 ----
 
 [[error-api-examples]]
+[float]
 === Examples
 
-See the example on how to send an error request to the APM Server.
+Send an example request to the APM Server:
 
 ["source","sh",subs="attributes"]
 ------------------------------------------------------------
@@ -94,6 +116,14 @@ curl http://localhost:8200/v1/errors \
   --data @docs/data/intake-api/generated/error/payload.json
 ------------------------------------------------------------
 
+See examples on how an error request to the APM Server can look like:
+
+* <<payload-with-error>>
+* <<payload-with-minimal-exception>>
+* <<payload-with-minimal-log>>
+
+[[payload-with-error]]
+[float]
 ==== Payload with an Error
 
 [source,json]
@@ -101,6 +131,8 @@ curl http://localhost:8200/v1/errors \
 include::./data/intake-api/generated/error/payload.json[]
 ----
 
+[[payload-with-minimal-exception]]
+[float]
 ==== Payload with an Error with minimal Exception Information
 
 [source,json]
@@ -108,6 +140,8 @@ include::./data/intake-api/generated/error/payload.json[]
 include::./data/intake-api/generated/error/minimal_payload_exception.json[]
 ----
 
+[[payload-with-minimal-log]]
+[float]
 ==== Payload with an Error with minimal Log Information
 
 [source,json]
--- a/docs/error-api.asciidoc
+++ b/docs/error-api.asciidoc
@@ -7,14 +7,10 @@ Unless you are implementing an agent, you don't need to know about the specifics
 To send error records you need to send a HTTP POST request to APM Server `errors` endpoint. 
 Information pertaining to the error record must be sent as a JSON object to the endpoint.
 
---
-
 Find more information about: 
 
 * <<error-schema-definition>>
 * <<error-api-examples>>
-
---
 
 [[error-schema-definition]]
 [float]
--- a/docs/guide/apm-getting-started.asciidoc
+++ b/docs/guide/apm-getting-started.asciidoc
@@ -4,6 +4,7 @@ include::{asciidoc-dir}/../../shared/attributes.asciidoc[]
 [[apm-gettting-started]]
 = Getting Started with Elastic APM
 
+[[overview]]
 == Overview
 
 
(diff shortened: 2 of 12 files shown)


docs/error-api.asciidoc
```diff
@@ -1,11 +1,25 @@
+[[error-api]]
```
watson: Note to my self that this will break the Node agent docs because they are linking to the old anchor

watson:
@simitt I noticed that you put the `[float]` after the id, eg:

```
[[error-api-examples]]
[float]
=== Examples
```

In the Node and Python docs we have always put it above, eg:

```
[float]
[[error-api-examples]]
=== Examples
```

I imagine it works in both cases, but it would be nice if we kept it the same across all APM docs. Would you mind moving it above?

docs/error-api.asciidoc
```diff
@@ -1,11 +1,25 @@
...
+
 To send error records you need to send a HTTP POST request to APM Server `errors` endpoint. Information pertaining to the error record must be sent as a JSON object to the endpoint.
 
 --
-See <<error-api-examples,Examples>> for sending a request to the error endpoint.
 
-See the schema definitions for how the error payload has to look like:
+Find more information about: 
+
+* <<error-schema-definition>>
```
watson: If there isn't a text after the link it will be rendered in cursive (`<<foo>>` vs `<<foo,Foo>>`).

So I personally always include the text even if it's the same as the headline that the link is pointing to. This way we'll not get mixed text-styling in case another bullet is added later where the text is included.

But maybe I'm doing it wrong 😅

docs/error-api.asciidoc
```diff
@@ -1,11 +1,25 @@
...
+* <<error-api-examples>>
+
+--
+
+[[error-schema-definition]]
+[float]
+=== Schema Definition
+The APM Server uses a JSON Schema for validating the transaction requests.
```
watson: I normally add a blank line after headlines. It makes it easier to read in my opinion.

docs/error-api.asciidoc
```diff
@@ -94,20 +116,32 @@ curl http://localhost:8200/v1/errors \
   --data @docs/data/intake-api/generated/error/payload.json
 ------------------------------------------------------------
 
+See examples on how an error request to the APM Server can look like:
+
+* <<payload-with-error>>
```
watson: same as above

docs/guide/apm-getting-started.asciidoc
```diff
@@ -25,9 +25,9 @@ It also makes it much easier to keep an eye on how many times specific errors ha
 === APM Components
 Elastic APM consists of four components:
 
-* Elasticsearch
-* APM agents
-* APM Server
+* {ref}/index.html[Elasticsearch]
+* {apm-agents-ref}/index.html[APM agents]
```
watson: I can't see where this is defined? If we need this I guess it should be added to https://github.com/elastic/docs/blob/master/shared/attributes.asciidoc

docs/overview.asciidoc
```diff
@@ -0,0 +1,39 @@
+[[overview]]
+== Overview 
+
+--
```
watson: What does the `--` do?

docs/overview.asciidoc
```diff
@@ -0,0 +1,39 @@
...
+APM Server is a central component in the Elastic APM system.
+It receives data from APM agents and transforms them into Elasticsearch documents.
+
+To get an overview about the whole Elastic APM system, please also have a look at the documentation for
+
+* {apm-node-ref}/index.html[APM Node.js Agent]
+* {apm-py-ref}/index.html[APM Python Agent]
+* {ref}/index.html[Elasticsearch]
+
+See how to {apm-get-started}/index.html[Get Started] with the Elastic APM system.
```
watson: I can't see where `apm-get-started` is defined?

docs/overview.asciidoc
```diff
@@ -0,0 +1,39 @@
...
+--
+Welcome to APM Server Docs.
+
+NOTE: The documentation pages are still work in progress.
+For more details also check APM Server https://github.com/elastic/apm-server[Github repository].
+
+APM Server is a central component in the Elastic APM system.
+It receives data from APM agents and transforms them into Elasticsearch documents.
+
+To get an overview about the whole Elastic APM system, please also have a look at the documentation for
```
watson: I think we should add line breaks after commas as well as periods

docs/security.asciidoc
```diff
@@ -1,19 +1,22 @@
...
 
-==== SSL setup
-
-To enable SSL you need a private key and a certificate issued by a certification authority (CA).
-Then you can specify the path to those files in the configuration properties apm-server.ssl.key and apm-server.ssl.certificate respectively.
-This will make the APM Server to serve HTTPS requests instead of HTTP, so you also need to enable SSL in the agent.
-Please check the agent documentation for how to do it.
+To enable SSL/TLS you need a private key and a certificate issued by a certification authority (CA).
+Then you can specify the path to those files in the configuration properties `apm-server.ssl.key` and `apm-server.ssl.certificate` respectively.
+This will make the APM Server to serve HTTPS requests instead of HTTP. Hence, you also need to enable SSL in the agent.
```
watson: Add a line break after the the period

docs/security.asciidoc
```diff
@@ -1,19 +1,22 @@
...
-Please check the agent documentation for how to do it.
+To enable SSL/TLS you need a private key and a certificate issued by a certification authority (CA).
+Then you can specify the path to those files in the configuration properties `apm-server.ssl.key` and `apm-server.ssl.certificate` respectively.
+This will make the APM Server to serve HTTPS requests instead of HTTP. Hence, you also need to enable SSL in the agent. 
+For agent specific details, please check the {apm-agent-ref}/index.html[agent documentation] for how to do it.
 
+[float]
 ==== Secret token
 
+You can configure a secret token which is sent with every request from the APM agents to the server. This string is used to ensure that only your agents can send data to your APM servers. Both the agents and the APM servers have to be configured with the same secret token.
```
watson: Add line breaks after each period

docs/transaction-api.asciidoc
```diff
@@ -1,11 +1,24 @@
+[[transaction-api]]
 == Transaction API
+The APM Server exposes an API Endpoint to send transaction records.
```
watson: Add an empty line after the headline

docs/transaction-api.asciidoc
```diff
@@ -1,11 +1,24 @@
+[[transaction-api]]
 == Transaction API
+The APM Server exposes an API Endpoint to send transaction records. 
+Unless you are implementing an agent, you don't need to know about the specifics of this API.
```
watson: Add a line break after the comma

docs/transaction-api.asciidoc
```diff
@@ -1,11 +1,24 @@
...
 
 To send transaction records you need to send a HTTP POST request to APM Server `transactions` endpoint. Information pertaining to the record must be sent as a JSON object to the endpoint.
 
 --
-See <<transaction-api-examples,Examples>> for sending a request to the transaction endpoint.
 
-See the schema definitions for how the transaction payload has to look like:
+Find more information about: 
+
+* <<transaction-schema-definition>>
```
watson: same as above

docs/transaction-api.asciidoc
```diff
@@ -103,20 +124,32 @@ curl http://localhost:8200/v1/transactions \
   --data @docs/data/intake-api/generated/transaction/payload.json
 ------------------------------------------------------------
 
+See examples on how a transaction request to the APM Server can look like:
+
+* <<payload-with-transactions>>
```
watson: same as above

docs/transaction-api.asciidoc
```diff
@@ -1,11 +1,24 @@
...
+
+* <<transaction-schema-definition>>
+* <<transaction-api-examples>>
+
+--
+
+[[transaction-schema-definition]]
+[float]
+=== Schema Definition
+The APM Server uses a JSON Schema for validating the transaction requests.
```
watson: Add an empty line after the headline

docs/guide/apm-getting-started.asciidoc
simitt: yes, I haven't pushed those changes yet

simitt: https://github.com/elastic/docs/pull/237

simitt:
@watson I can change the `float` handling, but I sticked to the [docs description](https://github.com/elastic/docs/blob/master/README.asciidoc#chunking-selected-sections).

docs/error-api.asciidoc
simitt: I don't see it being rendered in cursive.

If you have a look at the [current doc](https://www.elastic.co/guide/en/apm/server/current/_transaction_api.html) and the [plaing asciidoc](https://raw.githubusercontent.com/elastic/apm-server/master/docs/error-api.asciidoc) it is also linked without a name and not rendered cursive. Thus I think this should be fine.

watson:
> I can change the `float` handling, but I sticked to the [docs description](https://github.com/elastic/docs/blob/master/README.asciidoc#chunking-selected-sections).

Good point. Then it's @beniwohli and me that have done it wrong. We'll update ours instead I guess

docs/overview.asciidoc
simitt: It's a [part intro](https://github.com/elastic/docs/blob/master/README.asciidoc#part-intro)

docs/error-api.asciidoc
watson: Weird - It must be yet another one of those differences with how it renders locally vs online. Stick to you what you got then 👍

docs/overview.asciidoc
simitt: will change some line breaks.

simitt:
@watson thx for the review, I updated some parts.

docs/overview.asciidoc
watson: Shouldn't it then have a `[partintro]` line just above the `--`?

watson: Reading the asciidoc docs, it seems that this is only needed for parts which are level 0 (`=`). This is probably also why it's not complaining about the missing `[partintro]`

docs/error-api.asciidoc
dedemorton: I actually prefer not specifying link text and letting the build resolve the text. This is especially important if you are referring to a full topic heading. It gets weird for users if you change the topic heading, but don't remember to change the link text. It's a maintenance issue. If you do use link text, it should be generic enough that users won't be confused if the actual topic heading changes.

watson: @dedemorton I agree 👍 For some reason when I rendered the Node docs locally, these kind of links looked different than links with a link text. But apparently that's just because our local build script doesn't match the one we use in production.

dedemorton: @watson It's possible that Clint does some extra magic, but doesn't your script call https://github.com/elastic/docs/blob/master/build_docs.pl? AFAIK, the doc build simply calls build_docs.pl with --all specified (and uses the conf.yaml to set parameters).

docs/security.asciidoc
```diff
@@ -1,19 +1,32 @@
...
-This will make the APM Server to serve HTTPS requests instead of HTTP, so you also need to enable SSL in the agent.
-Please check the agent documentation for how to do it.
-
+[[ssl-setup]]
+[float]
+==== SSL/TLS setup
+
+To enable SSL/TLS you need a private key and a certificate issued by a certification authority (CA).
+Then you can specify the path to those files in the configuration properties 
+`apm-server.ssl.key` and
```
ruflin: Is that on purpose split over 3 lines?

docs/error-api.asciidoc
ruflin: @watson can you open a PR on the node agent side which is merged at a similar time like this one so we don't break the doc build?

docs/security.asciidoc
simitt: yes to make it more readable

docs/error-api.asciidoc
watson: @dedemorton Yes, I'm using the build_docs.pl script. But for some reason this is what I'm seeing:

In the Node.js docs, if I write:

```asciidoc
* <<express,Get started with Express>>
* <<hapi>>
* <<koa,Get started with Koa>>
```

It's generated like so locally:

screen shot 2017-09-20 at 10 13 48

docs/error-api.asciidoc
watson: Good idea: https://github.com/elastic/apm-agent-nodejs/pull/43

watson: @ruflin could you verify that I changed the link correctly? I.e. that the link is going to be like I predicted

PR #190, opened 2017-09-19
Clean up asciidoc style in getting started guide

watson:
This PR does the following:

- Adds ID's to all headlines
- Removes `$` from all bash scripts
- Adds line breaks after all commas and periods (maybe a little too aggressive, but at least consistent)
- Break up long bash commands into multiple lines
- Add blank line after all headlines
- Remove double blank lines

### DIFF
--- a/docs/guide/apm-getting-started.asciidoc
+++ b/docs/guide/apm-getting-started.asciidoc
@@ -1,30 +1,34 @@
 :branch: current
 include::{asciidoc-dir}/../../shared/attributes.asciidoc[]
 
-[[apm-gettting-started]]
+[[gettting-started]]
 = Getting Started with APM
 
 [[overview]]
 == Overview
 
-
 APM is an application performance monitoring system built on the Elastic Stack.
 It uses Elasticsearch as its data store and allows you to monitor performance of thousands of applications in real time.
 
-With APM, you can automatically collect detailed performance information from inside your applications and it requires only minor changes to your application.
+With APM,
+you can automatically collect detailed performance information from inside your applications and it requires only minor changes to your application.
 APM will automatically instrument your application and measure the response time for incoming requests.
 It also automatically measures what your application was doing while it was preparing the response.
-For example, external database queries are measured to make it easy to detect slow queries.
-It also measures calls to caches, external HTTP requests etc.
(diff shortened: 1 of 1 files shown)


watson:
@simitt rebased

I simplified some of the ID's as well as it will make for better/smaller links I think. Let me know if you don't like it 😃

docs/guide/apm-getting-started.asciidoc
```diff
@@ -50,46 +60,58 @@ APM Server builds Elasticsearch documents from the data received from agents.
...
+To visualize the data after it's sent to Elasticsearch,
+you can use the pre-built,
+open source Kibana dashboards that come with APM Server.
 
-We designed Elastic APM to run on anything from a regular laptop to thousands of machines, and it's easy to get started.
+We designed Elastic APM to run on anything from a regular laptop to thousands of machines,
+and it's easy to get started.
 
-[[install-and-run-elastic-apm]]
+[[elastic-apm]]
```
simitt: It was my understanding that those IDs are used in the actually created URL. From this viewpoint I like having `install-and-run` in the ID. Same for the IDs below where you removed the `install` part.

watson: Yes they are (at least for level 1 and 2). I was just thinking that shorter URL's were better. But I agree that I might have been carried away with this one in particular. Maybe it should have been `install-elastic-apm` or something. I just feel the full headline is too bloated if used as a filename. But maybe it's just me? The ones below I still think are nicer in the the shorter version though.

simitt: As one can probably see the whole doc as an entrypoint to APM, I am ok with having very generic IDs here. I personally would have left the prefix `install-` but I think this way it is also fine.

simitt:
The PR is fine with me, please rebase again.

watson:
@simitt Rebased and went over all the ID's again.

This will now build the following page structure:

- Overview, `overview.html`
  - APM Components, `components.html`
- Install and run Elastic APM, `install-and-run.html`

I think those a pretty nice filenames. Everything on the `install-and-run.html` page are `[float]`, so they will not get their own pages. How do you like this?

PR #191, opened 2017-09-20
Fix link to security docs page

watson:


watson:
This was breaking the global docs build, so I took some liberties and merged my self 😅

PR #192, opened 2017-09-20
Add tests for beater config.

simitt:
* Add tests for configuration handling.
* Fix bug for check wheterh SSL is enabled.

As a first attempt to add and cleanup tests in the beater module I quickly added some tests to the config. If I remember the logic, whether or not SSL should be enabled, correctly then I found a bug in the current checks.

### DIFF
--- a/beater/config_test.go
+++ b/beater/config_test.go
@@ -1,0 +1,120 @@
+package beater
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/elastic/go-ucfg/yaml"
+)
+
+func TestConfig(t *testing.T) {
+	truthy := true
+	cases := []struct {
+		config         []byte
+		expectedConfig Config
+	}{
+		{
+			config: []byte(`{
+        "host": "localhost:3000",
+        "max_unzipped_size": 64,
+        "max_header_bytes": 8,
+        "read_timeout": 3s,
+        "write_timeout": 4s,
+        "shutdown_timeout": 9s,
+        "secret_token": "1234random",
+        "ssl": {
+					"enabled": true,
+					"key": "1234key",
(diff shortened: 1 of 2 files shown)


beater/config.go
```diff
@@ -23,7 +23,11 @@ type SSLConfig struct {
 }
 
 func (c *SSLConfig) isEnabled() bool {
-	return c != nil && (c.Enabled == nil || *c.Enabled)
+	return c != nil && ((c.Enabled == nil && c.hasKeyOrCert()) || (c.Enabled != nil && *c.Enabled))
```
jalvz: The bug is that `&SSLConfig{Enabled: nil}` was returning `true`?
I don't know what is the expected behaviour, this is a copy&paste from other beats implementations so I think @ruflin knows better wether this is correct or not. 
If so, other beats might require fixing as well.

simitt: If I remember correctly then the behavior should be that if `enabled: nil` it only returns `true` when there is a `cert` and\or? a `key`.
The tests were written in a way where this was not appearant unfortunately. 

@ruflin how should it be?

ruflin: The general idea is that we check if the namespace is there or not. If the namespace is set in the config, we assume it's enabled except if enabled is set to false. We don't check if specific fields are set inside. If the right fields are set should be part of the Validation phase.

In beats in most cases when `Enabled` is checked, it does not know about the inner details of the config object as it's common.Config and still needs to be unpacked. We didn't use here the beats feature directly because it would make the code in this case more complicated as we would have to define ssl as common config and then unpack it.

In summary, I would keep the behaviour as is mainly as we never had issues with it in beats.

If someone sets `apm-server.ssl:` in the config, it means enabled.

simitt: sounds reasonable - I'll change it back.

PR #196, opened 2017-09-21
Remove Kibana dashboard from Readme.

simitt:
* Put Kibana description completely in asciidoc file.
* Add missing float.


simitt:
@watson It is a bit rephrased but I assume that's on purpose since it is more recent, see [Kibana in asciidoc](https://github.com/elastic/apm-server/pull/196/files#diff-6e15198c8dbbd4bd2eb495aebea6f896L138). Only thing I had to add was the picture.

PR #202, opened 2017-09-21
Move beats update script to script folder

ruflin:
The beats update script is moved to the script folder where it belongs. Under _beats should only be files which are coming from beats itself. This change allows to completely remove and create the _beats folder each time which also make sure old files are removed.

Additional changes:

* Rename LICENSE to LICENSE.txt: This was forgotten in a previous PR
* Cleanup beats Golang dependencies each time to make sure obsolete files are removed


NOTICE.txt
```diff
@@ -190,16 +190,6 @@ Apache License 2.0
...
-Dependency: github.com/elastic/beats
-Version: master
-Revision: 57aaa90929a49ba9b11cdc800c49930df3e24161
-License type (autodetected): Apache License 2.0
-./vendor/github.com/elastic/beats/LICENSE.txt:
---------------------------------------------------------------------
-Apache License 2.0
-
-
---------------------------------------------------------------------
```
simitt: why is this removed? shouldn't it be re-added after updating beats?

script/update_beats.sh
```diff
@@ -48,5 +50,9 @@ cp $GIT_CLONE/.go-version .go-version
 # Get system test dependencies
 cp -r $GIT_CLONE/testing testing
 
+# Generate .gitignore file
+echo "libbeat/_meta/fields.generated.yml" > .gitignore
+echo "libbeat/fields.yml" >> .gitignore
```
simitt: Would it maybe be better to remove those files again instead of adding them to `.gitignore`?

NOTICE.txt
ruflin: Not 100% sure what happened here before but it seems it was a duplicate, as it's also a few lines up (187).

script/update_beats.sh
ruflin: Removing does not work as these files are generated during `make update` and I want to prevent someone else checking them in.

ruflin:
@simitt Can you check the diff again? It seems here it shows the rename of LICENSE to LICENSE.txt.

PR #216, opened 2017-10-02
More end to end tests

jalvz:
This PR puts some code for automatic-ish tests of agents + apm-server.
A few considerations:

+ "smoke" test is a misnomer, but all of "integration", "system" and "end-to-end" were already taken, yet none of those use actual agents...

+ this is not meant to be 100% automated: because of the very asnyc nature of the workflow  (and differences between agents) there are quite a few arbitrary `time.sleep`, so this is not robust enough eg. for a CI environment; but hopefully robust enough to give people an idea of how components are playing together.

+ i am unsure about directory structure: on one hand i'd like to reuse the dependencies of the flask app, on the other hand is wrong to have `smoke.py` in the same directory as the flask app... also the tests must be executed from current dir... i even think all of this should go to a different repo.

+ there are a few more rough edges, but in general i'm looking for some feedback on how to improve it.


tests/agent/python/README.md
```diff
@@ -5,11 +5,10 @@ This is to setup an environment for the manual testing of the [python agent](htt
 
 # Setup
 
-There are two possible setups. If you already have Python installed, you can run the following to commands to get started, preferably with an activated virtualenv::
+There are two possible setups. If you already have Python installed, you can run the following to commands to get started, with an activated virtualenv and requirements installed::
```
simitt: minor typo `run the following to commands` should be `two commands` I assume..

tests/agent/nodejs/app.js
```diff
@@ -1,28 +1,42 @@
...
-  appName: 'nodejs-testapp',
+  appName: 'test-app', flushInterval: 1
+})
 
-  // Use if APM Server requires a token
-  secretToken: '',
 
-  // Set custom APM Server URL (default: http://localhost:8200)
-  serverUrl: '',
+var app = require("express")();
```
simitt: How about moving it to `tests/agent/nodejs/express/app.js` as we also have it namespaced for `flask` in `python`?

tests/agent/python/flaskapp/smoke.py
```diff
@@ -0,0 +1,257 @@
...
+from datetime import datetime, timedelta
+import subprocess
+import logging
+import json
+import copy
+import uuid
+
+p_uid = uuid.uuid4().hex
+
+num_reqs = 0    # use with care
```
simitt: This comment somehow scares me, but I don't know why. Could you explain briefly why this should be used with care?

tests/agent/python/flaskapp/smoke.py
```diff
@@ -0,0 +1,257 @@
...
+
+    except AssertionError:
+        num_reqs == 0
+        ioloop.IOLoop.instance().stop()
+        logger.error("Bad response, aborting: {} - {} ({})".format(r.code, r.error, r.request_time))
+
+
+def run_process(lang, nworkers=4):
+    if lang == 'node':
+        cmd = ['node', '../../nodejs/app.js']
```
simitt: We are at `tests/agent/python/flaskapp/smoke.py`. If this should be used as smoke test for both agents, could you please move it to a neutral directory?

tests/agent/python/flaskapp/smoke.py
```diff
@@ -0,0 +1,258 @@
...
+    def __enter__(self):
+        self.p = run_process(self.l)
+        time.sleep(3)
+        logger.info("Process ready {}".format(self.p.pid))
+
+    def __exit__(self, *args):
+        logger.info("Flushing queues")
+        time.sleep(2)
+        self.p.terminate()
+        time.sleep(10)
```
simitt: What exactly is it that needs to happen in the time you sleep at `enter` and `exit` ? Could we use something like `wait_for_xyz` to reduce the waiting time, or adapt it based on how many requests are made?

tests/agent/python/flaskapp/smoke.py
```diff
@@ -0,0 +1,258 @@
...
+        d1 = d1[k]
+    return d1
+
+
+def es_query(field, val):
+    return {"query": {"term": {field: val}}}
+
+
+def reset():
+    f = '../../../../_meta/kibana/default/index-pattern/apmserver.json'
```
simitt: This is the kibana index pattern you are using, I think you should consider using `apm-server/apm-server.template-es.json` which is created when the `make update` is run.

jalvz: hmmm not for me, seems like. i cant find that file after running `make update`

tests/agent/python/flaskapp/smoke.py
jalvz: right, ill explain better in the code 👍 
short story, because it is a global variable

tests/agent/python/flaskapp/smoke.py
jalvz: processes (eg gunicorn) need to start and be ready to accept requests, node agent needs time to flush the queue (that is the `2`), python agent also needs time to flush the queue upon SIGTERM (thats the `10`), then apm-server / libbeat needs time to pipe events to ES...

i don't like the `wait-for-log` approach for this, because then any kind of error looks like a timeout error and you don't know what is going on underneath, which is not great if you are executing this manually and watching the output. 
OTHO, if tests fail consistently under a reasonable waiting time, you then know something is up.

I don't worry much about execution time, because i don't intend to run it very often, at least for now...

tests/agent/python/flaskapp/smoke.py
simitt: you're right, it is not part of the `make update`, you have to run `make index-template` to get it.

tests/agent/python/flaskapp/smoke.py
simitt: Thank you for the explanation. I was worried about the time to sleep being set according to the number of requests you used to make. But it sounds like the time to wait is a fixed time based on agent handling. With this I am fine, since as you said the tests are not intended to be included in the fully automated test runs.

tests/agent/nodejs/app.js
simitt: @jalvz what do you think about moving this file to a subfolder as for python? 
Apart from that LGTM.

jalvz: yes, sounds good!

PR #218, opened 2017-10-03
Update beats

simitt:


ruflin:
@simitt as govendor does not check the dependencies for things under `dev-tools` the kibana package must be added manually as it is not used anywhere else

PR #219, opened 2017-10-04
Move kibana index pattern info to fields.yml.

simitt:
With those changes we can auto-load the kibana dashboards by calling `make import-dashboards`. It enables users to simply run `./apm-server setup` for loading the dashboards.

Remove index-pattern from dashboards, as this should be created by the
kibana index pattern generator.

This requires https://github.com/elastic/beats/pull/5318 to be merged in. 

Currently the dashboards include a defined `index-pattern`. Everything in this index pattern that is not produced by the creation script in the beats framework yet needs to be supported by the script (see all the PRs I created in beats). Additionally the information added for the dashboards needs to be moved to the `fields.yml`. 
This is what is necessary to make the dashboard scripts load automatically with the standard `index-pattern` and not the hardcoded one. 

@makwarth cc


docs/fields.asciidoc
```diff
@@ -328,6 +330,20 @@ Agent version.
...
 
+[float]
+=== `view errors`
+
+type: keyword
+
+format: url
+
+[float]
+=== `error id icon`
```
ruflin: I wonder if we should have a special note on these fields that these are script field. This is a change that would require to our docs script afterwards.

processor/error/_meta/fields.yml
```diff
@@ -2,24 +2,44 @@
   title: APM Error
   description: Error-specific data for APM
   fields:
+    - name: view errors
```
ruflin: not too happy with the name of the field here. I know it's what is used now. I'm thinking if we should also have here some conventions like not using spaces in the future? This is something that would have to be done when creating the dashboard.

simitt: I did not intend to change any outcome of the dashboards, so no namings, linkings etc. i just moved the current parts to the right places to enable automatic loading and automatic adapting of the index pattern (which was pretty outdated already). 

Hence, if you are unhappy with the naming, please bring this up with @simianhacker and the rest of the APM team.

docs/fields.asciidoc
simitt: We could add something to the docs script as soon as there is a `script` set.

simitt:
I tested that `make import-dashboards` work. 
Running `./apm-server setup dashboards` does not work, because of the directory structure in dev env (what @ruflin explained to me recently). 
`./apm-server setup -E setup.dashboards.directory=${PWD}/_meta/kibana` works just fine (and is basically what is used by the make cmd). 

Dashboards are not working with 5.x, but afaik this should not be related to my changes.

processor/error/_meta/fields.yml
ruflin: Definitively. Wasn't meant to deal with in this PR.

ruflin:
@simitt Great to have this in now. As a follow up we should also change our docs to use the `setup` command.

processor/error/_meta/fields.yml
simitt: I just realized that my former comment might have come off a bit harsh, this was not my intention.

PR #221, opened 2017-10-04
Fix dashboard loading for Kibana 5x

simitt:
For automated loading of dashboards into kibana 5.6 from what I can see we need to split up the dashboards and visualizations as done below, since the importer cannot handle the default format. 

There is still one issue when linking to `traces` from `transactions` and same for `error details`.
Maybe you have some input on that.

This depends on beats PR https://github.com/elastic/beats/pull/5329.
Please do not merge before this PR is in and we've update apm-server to make use of it.


ruflin:
I tested this with production data on Elastic Stack 5.6 and all looks good. The graph "Request per Minute" is empty but I assume that is related to https://github.com/elastic/apm-server/pull/226

On 6.0.0-rc1 I tested with a very small data set. I didn't see any errors showing up. 

For me this is ready to be merged.

PR #225, opened 2017-10-05
Updated copy

makwarth:
- Experimental -> Alpha
- Updated blog post link


roncohen:
Great! We also need to do it for the `page_header.html` like here: https://github.com/elastic/apm-agent-nodejs/pull/46/files#diff-b40e969cab8a3a76759352587744c4ca

PR #227, opened 2017-10-10
Frontend processors

jalvz:
* Create a processor `type` attribute that can be used to map each processor type to different chains of handlers 
* Add frontend type endpoints. The processing pipeline is the same as for backend endpoints for now, but they might require different handlers
* Handle `/healthcheck` like any other processor


beater/mux.go
```diff
@@ -0,0 +1,220 @@
...
+	"encoding/json"
+	"errors"
+	"net/http"
+
+	"crypto/subtle"
+
+	"github.com/elastic/beats/libbeat/monitoring"
+)
+
+type chainable func(processor.Processor, Config, reporter) http.Handler
```
roncohen: let's call it something like `processorChain` / `processorHandler` / `processorHandlerChain` for more descriptiveness

beater/mux.go
```diff
@@ -0,0 +1,220 @@
...
+	responseValid  = monitoring.NewInt(serverMetrics, "response.valid")
+	responseErrors = monitoring.NewInt(serverMetrics, "response.errors")
+
+	errInvalidToken    = errors.New("invalid token")
+	errPOSTRequestOnly = errors.New("only POST requests are supported")
+
+	handlerMap = map[int]chainable{
+		processor.Backend:  backendHandler,
+		processor.Frontend: frontendHandler,
+		processor.Nop:      nopHandler,
```
roncohen: We should call it something related to healthcheck as long as we don't have other things relying on the no-op stuff. In other words, lets make it genereric when there's a use for that. For now, i suggest renaming to `processor.HealthCheck` and `healthCheckHandler`

beater/mux.go
```diff
@@ -0,0 +1,220 @@
...
+			return
+		}
+		h.ServeHTTP(w, r)
+	})
+}
+
+// isAuthorized checks the Authorization header. It must be in the form of:
+//   Authorization: Bearer <secret-token>
+// Bearer must be part of it.
+func isAuthorized(req *http.Request, secretToken string) bool {
```
roncohen: It's great to have the `mux.go` and `mux_test.go` files, but I don't believe this all fits in here. mux is short for "HTTP request multiplexer". In other words, it has to do with http routing. We're also doing header parsing, authentication etc. here now. I suggest moving a big part of it back to `server.go`.

beater/mux.go
```diff
@@ -0,0 +1,220 @@
...
+	}
+	header := req.Header.Get("Authorization")
+	parts := strings.Split(header, " ")
+	if len(parts) != 2 || parts[0] != "Bearer" {
+		return false
+	}
+	return subtle.ConstantTimeCompare([]byte(parts[1]), []byte(secretToken)) == 1
+}
+
+func appHandler(p processor.Processor, config Config, report reporter) http.Handler {
```
roncohen: maybe `processHandler` is a more appropriate name here. It does the actual processing.

roncohen: or `processRequestHandler`

processor/processor.go
```diff
@@ -10,10 +10,17 @@ import (
 
 type NewProcessor func() Processor
 
+const (
+	Backend  = iota
+	Frontend = iota
+	Nop      = iota
```
roncohen: `HealthCheck`

processor/transaction/processor.go
```diff
@@ -22,18 +23,28 @@ var (
...
+func NewFrontendProcessor() pr.Processor {
+	return &processor{
+		schema: pr.CreateSchema(transactionSchema, processorName),
+		typ:    pr.Frontend,
 	}
 }
 
 type processor struct {
 	schema *jsonschema.Schema
+	typ    int
```
roncohen: `type` ?

script/output_data/output_data.go
```diff
@@ -39,7 +39,7 @@ func generate() error {
 
 		f, err := os.Open(path)
 		if err != nil {
-			return err
+			continue
```
roncohen: What's going on here?

beater/mux.go
jalvz: Actually the thing I wanted to isolate is the server creation / start / stop (so what remains in `server.go` and put aside all the details. It was getting to big of a file. Maybe `mux` is just not a good name.  what about `handling` / `handlers` or `processing`... ?

processor/transaction/processor.go
jalvz: `type` is a reserved word, can't be a variable name

script/output_data/output_data.go
jalvz: this looks for files like `some_path/tests/valid/something.json` to feed some tests, which doesn't make sense for the health checker as doesn't accept any payload

jalvz:
renamings done here e6654b0807b76dfbc48cd74962f25993c4f704e6

processor/transaction/processor.go
roncohen: ah right. How about `processorType` ? we already have `processorName` for example

script/output_data/output_data.go
roncohen: Could we check the processor type here and skip if it's the healthcheck? That way we'll still get the error for the other processors if something is wrong.

processor/transaction/processor.go
jalvz: is that really necessary? i'm usually told to use short and concise (ie idiomatic) names, and seems you are pushing the opposite way. `typ` is a var name extensively used in Go's source code itself. 
I don't mind either way, but we should align on this to avoid having the same discussion over and over.

That variable is only used in a getter:
```
func (p *processor) Type() int {
	return p.typ
}
```
Isn't it clear enough? i'm calling this as `processor.Type()`

simitt: +1 on calling it `typ`. 
It is only used internally, and as @jalvz pointed out `typ` is heavuly used in the `go` source code itself.

roncohen:
LGTM 👍  
Looks like there a test missing that checks that FrontendEndpoint return 2xx when it's enabled. When we add functionality that makes them different from BackendEndpoints, we should be sure to add those tests.

beater/handlers.go
```diff
@@ -76,6 +78,16 @@ func logHandler(h http.Handler) http.Handler {
 	})
 }
 
+func frontendSwitchHandler(feSwitch bool, h http.Handler) http.Handler {
```
simitt: How about naming this more general like `checkEnabledHandler`? Everything it does is check and react on a boolean in a very generic way.
If you rename it this could be reused for also disabling backend in the future if one only wants to run frontend..

beater/config.go
```diff
@@ -14,6 +14,7 @@ type Config struct {
 	SecretToken        string        `config:"secret_token"`
 	SSL                *SSLConfig    `config:"ssl"`
 	ConcurrentRequests int           `config:"concurrent_requests" validate:"min=1"`
+	EnableFrontend     bool          `config:"enable_frontend"`
```
simitt: could you add the setting to the config_tests?

jalvz:
this would be the `enabled` test: https://github.com/elastic/apm-server/pull/227/commits/677103f08e704af477ebf8632d4a3c89e565ce18#diff-24302485aab47c0697eb5cb545fbe394R79

i test `!= 403` instead of `= 200` for isolation, in that test i don't care what happens down the road, only that the switch lets the request through.

beater/config.go
jalvz: agh sorry, i was too fast :(
didn't know about config_tests, will do in separate PR

PR #229, opened 2017-10-11
Central processor registration

jalvz:
Register processors centrally, so that processor "configuration" / metadata like endpoint and handlers can be all in one place

This also removes the workarounds with blank identifier imports, and instantiates one processor per request to avoid harmful state

Fixes https://github.com/elastic/apm-server/issues/230


beater/server.go
```diff
@@ -34,11 +36,18 @@ var (
...
+	"/v1/transactions": transaction.NewProcessor,
+	"/v1/errors":       err.NewProcessor,
+}
+
 func newMuxer(config Config, report reporter) *http.ServeMux {
 	mux := http.NewServeMux()
 
-	for path, p := range processor.Registry.Processors() {
-		handler := appHandler(p, config, report)
+	for path, constructor := range Routes {
```
simitt: would you mind naming this only `c` or `construc` or similar, to avoid having the same name for the type and the variable?

beater/server.go
```diff
@@ -108,14 +117,16 @@ func authHandler(secretToken string, h http.Handler) http.Handler {
...
-		code, err := processRequest(r, p, config.MaxUnzippedSize, report)
+		code, err := processRequest(r, constructor, config.MaxUnzippedSize, report)
 		sendStatus(w, r, code, err)
 	})
 }
 
-func processRequest(r *http.Request, p processor.Processor, maxSize int64, report reporter) (int, error) {
+func processRequest(r *http.Request, new constructor, maxSize int64, report reporter) (int, error) {
+
+	var p processor.Processor = new()
```
simitt: I like this improvement, to have one dedicated processor per request. 
Currently we do call the [`CreateSchema`](https://github.com/elastic/apm-server/blob/master/processor/validator.go#L11) method though when creating a new processor. With your changes this would be done for every request. I don't think is the way to go. Could you update that the schema is created once per endpoint and then reused per request?

jalvz: good point! we totally should do that.
besides, this is not meant to be merged (at least just yet), this conflicts with PR227 and i'd like to see what everyone thinks of https://github.com/elastic/apm-server/issues/230

simitt: I totally support this improvement, as I think it makes things clearer and less error prone.

beater/server.go
```diff
@@ -34,11 +36,18 @@ var (
 	errPOSTRequestOnly = errors.New("only POST requests are supported")
 )
 
+type constructor func() processor.Processor
```
roncohen: how about `newProcessor` or `processorFactory` ?

jalvz:
Rebased and implemented review comments, ready to `squash & merge`.

This undoes parts of https://github.com/elastic/apm-server/pull/227, now there isn't need for `Type` and multiple constructors.

Found no issues with circular imports, on the contrary this removes more package dependencies than adds.

PR #233, opened 2017-10-13
Small refactor.

roncohen:
This refactors the frontend handler so the config check happens once instead of for each request. 

- Removes `frontendSwitchHandler`
- Introduces generic `errorHandler`

### DIFF
--- a/beater/handlers.go
+++ b/beater/handlers.go
@@ -58,9 +58,12 @@ func backendHandler(p processor.Processor, config Config, report reporter) http.
 }
 
 func frontendHandler(p processor.Processor, config Config, report reporter) http.Handler {
-	return logHandler(
-		frontendSwitchHandler(config.EnableFrontend,
-			processRequestHandler(p, config, report)))
+	if config.EnableFrontend {
+		return logHandler(processRequestHandler(p, config, report))
+	} else {
+		return logHandler(errorHandler(403, errForbidden))
+	}
+
 }
 
 func healthCheckHandler(_ processor.Processor, _ Config, _ reporter) http.Handler {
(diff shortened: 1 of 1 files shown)


beater/handlers.go
```diff
@@ -58,9 +58,12 @@ func backendHandler(p processor.Processor, config Config, report reporter) http.
 }
 
 func frontendHandler(p processor.Processor, config Config, report reporter) http.Handler {
-	return logHandler(
-		frontendSwitchHandler(config.EnableFrontend,
-			processRequestHandler(p, config, report)))
+	if config.EnableFrontend {
```
jalvz: im not convinced about this. 

i wanted to not have any logic at all on the top level handlers (backend, frontend) and just chain downstream handlers, so they can be better reused. chaining would be their only responsibility.
this ways you are doing too many things imo: composing _and_ branching.

i'd liked more eg. @simitt idea of renaming the feSwitchHandler so it could be used in the backend if needed. this way you would have to duplicate the logic, and is not like this affords  code or something.

beater/handlers.go
```diff
@@ -70,21 +73,17 @@ func healthCheckHandler(_ processor.Processor, _ Config, _ reporter) http.Handle
 		}))
 }
 
-func logHandler(h http.Handler) http.Handler {
+func errorHandler(code int, err error) http.Handler {
```
jalvz: this looks like this is the only way to send errors, but that is not true. how would you reuse this eg. for errors returned in `processRequestHandler`?

simitt:
I am confused, how does the check for the `frontend_handler` happen once and not per request with this change? From my understanding it still happens per request.

roncohen:
@simitt when `frontendHandler` and `backendHandler` are called, they return a handler chain. A bunch of http.Handler that are wired to call each other when you put a request in at the top.

In the current version when a request is passed to `frontendSwitchHandler`, it will look at the config and decide whether to return an error or to pass the request on the next handler `processRequestHandler`.

With this change, we make the check in `frontendHandler` and return a simple chain that always gives an error if frontend endpoints are disabled.

beater/handlers.go
roncohen: `errorHandler` is just a convenience handler when there are handler chains that should always end up in an error (like the frontend handler when frontend is disabled).

beater/handlers.go
roncohen: To me it's cleaner that the logic related to handling frontend endpoints is only in one place. The rest of the chain is generic.

roncohen:
I don't feel strongly about this so if you don't think this is cleaner I'm fine to close it.

simitt:
Ofc, thanks for the explanation.

I'd be fine with this change then. The overhead of checking per request could sum up if the server gets overloaded, but since it is still just a simple check I don't see an issue with the current solution either.

PR #235, opened 2017-10-13
Use libbeat LoadCertificate function on server side

sterchelen:
To add the encrypted private key loading feature, I've :
- update the `vendor/github.com/elastic/beats/libbeat/outputs` package which put in public mode the `LoadCertificate` function.
-  use the `LoadCertificate` function and inject the certificate in the `http.Server`.
- Modify tests in consequence.
- Modify reference yaml files.
- Add more log on certificate / private key loading.


sterchelen:
@ruflin @jalvz Here is the change to use the libbeat `LoadCertificate` function.

There is no test for a private key with a pass_phrase. To allow that we should think to reuse the following function with little modification to decrypt the pem block: https://github.com/elastic/beats/blob/d15cab2b3ec265a7f8bf3514c0e4fbc01c498a40/libbeat/outputs/transport/transptest/testing.go#L180 

Do I make the modification on beats repository or I copy the function inside our repository and apply the modification ?

ruflin:
As the functionality is inherited from beats now, it should be up to beats to make sure it works as expected. Or is the usage / implementation different here?

sterchelen:
@ruflin ,
The implementation is exactly the same as beats ! I will made a pull request on beats to implement this test :).

sterchelen:
PR open on beats repository --> https://github.com/elastic/beats/pull/5431
It modify the function that generate a certificate for testing purpose by adding a password parameter.

When the PR will be merged, I'll can add more tests on the `beater/server_test.go` file with a focus on private key with password handling.

sterchelen:
Gentlemen,

I've use the beats modification that I've made to create testing certificate. I've also added a new unit test to test the server in https mode with a encrypted certificate key.

jalvz:
Great @sterchelen ! Ill have a look soon. In the meanwhile, can you rebase master and fix conflicts?

sterchelen:
I've already merged master on it :)

_meta/beat.reference.yml
```diff
@@ -20,7 +20,8 @@ apm-server:
   #ssl.enabled: false
   #ssl.certificate : "path/to/cert"
   #ssl.key : "path/to/private_key"
-
+  #ssl.pass_phrase: "foobar"
```
jalvz: correct attribute here is `ssl.key_passphrase`

beater/server_test.go
```diff
@@ -13,7 +13,10 @@ import (
 	"testing"
 	"time"
 
-	"github.com/kabukky/httpscerts"
```
jalvz: this dependency is not used anymore, should be removed then (also remember to run `make notice` after removing it to remove its entry from the NOTICE file as well)

beater/server_test.go
```diff
@@ -160,16 +163,29 @@ func TestServerNoContentType(t *testing.T) {
 	assert.Equal(t, http.StatusBadRequest, rr.Code, rr.Body.String())
 }
 
+func TestServerSecureOkWithPasswordKey(t *testing.T) {
```
jalvz: This is good!
I miss 1 more test that "breaks" if the passphrase generated and the one passed in the server config do not match.
Note: in that scenario the server won't be able to start, so you need to assert that a timeout happens instead, the one in `waitForServer`. That is hardcoded to 10 seconds, so you will also need to modify that function to accept a `timeout` param and pass a lower value for that test because 10 secs every time it runs would be very slow.
Does this makes sense to you?

beater/server.go
```diff
@@ -35,11 +37,22 @@ func run(server *http.Server, config Config) error {
 
 	ssl := config.SSL
 	if ssl.isEnabled() {
-		return server.ListenAndServeTLS(ssl.Cert, ssl.PrivateKey)
+		logp.Info("Loading server certificate: %s", config.SSL.Certificate.Certificate)
+		logp.Info("Loading private key: %s", config.SSL.Certificate.Key)
+		cert, err := outputs.LoadCertificate(&config.SSL.Certificate)
+		if err != nil {
+			return err
```
jalvz: This code path should also be tested, I think

jalvz:
Great work @sterchelen ! I think this is very very close :)

beater/server_test.go
sterchelen: This dependency is already removed. You are looking in the wrong side of the diff :)

beater/server.go
sterchelen: @jalvz While I'm writing a test for this path, I saw that we are completely hiding error from the `setupServer` function (`server_test` file). Particularly on the `go run` statement.

If we check the error while running the server, we could remove all the `checkErrMsg`tests.

For instance, the `TestServerSecureBadDomain` test, would only have to check if an error has occured with the following msg : `http: TLS handshake error from 127.0.0.1:60368: remote error: tls: bad certificate`.
Again, for the `TestServerBadProtocol` test, we would only have to check for an error like `http: TLS handshake error from 127.0.0.1:60386: tls: oversized record received with length 21536` which tells us that we are trying to reach the https server with the http protocol.

Do you agree ?

jalvz: @sterchelen that sounds good to me, but how do check the error while running the server?

beater/server_test.go
jalvz: Sorry, I meant the actual code from the dependency, not the import.
https://github.com/elastic/apm-server/blob/master/vendor/vendor.json#L1163-L1168 and https://github.com/elastic/apm-server/tree/master/vendor/github.com/kabukky/httpscerts

sterchelen:
@jalvz  sorry for the delay ... So much professional work :)

_meta/beat.reference.yml
```diff
@@ -20,7 +20,7 @@ apm-server:
   #ssl.enabled: false
   #ssl.certificate : "path/to/cert"
   #ssl.key : "path/to/private_key"
-
+  #ssl.key_passphrase: "foobar"
```
graphaelli: let's make this an empty string to match other elastic pkgs

apm-server.reference.yml
```diff
@@ -20,7 +20,7 @@ apm-server:
   #ssl.enabled: false
   #ssl.certificate : "path/to/cert"
   #ssl.key : "path/to/private_key"
-
+  #ssl.key_passphrase: "foobar"
```
graphaelli: let's make this an empty string to match other elastic pkgs

beater/client.go
```diff
@@ -29,7 +29,11 @@ func isServerUp(secure bool, host string, numRetries int, retryInterval time.Dur
 
 	for i := 0; i <= numRetries; i++ {
 		if check() {
-			logp.Info("HTTP Server ready")
+			secureInfo := ""
+			if secure {
+				secureInfo = " (secure mode activated)"
+			}
```
graphaelli: building `secureInfo` can be done outside the for loop.  Also good with removing it altogether.

beater/server.go
```diff
@@ -2,10 +2,13 @@ package beater
 
 import (
 	"context"
+	"crypto/tls"
 	"net/http"
 	"time"
 
 	"github.com/elastic/apm-server/version"
+	"github.com/elastic/beats/libbeat/outputs"
+
```
graphaelli: extra newline here

beater/server_test.go
```diff
@@ -274,7 +301,7 @@ func randomAddr() string {
 	return l.Addr().String()
 }
 
-func waitForServer(secure bool, host string) {
+func waitForServer(secure bool, host string, timeout int) {
```
graphaelli: would prefer a context to timeout here but we can add that later

beater/server_test.go
```diff
@@ -290,13 +317,24 @@ func waitForServer(secure bool, host string) {
 		return res.StatusCode
 	}
 
-	for i := 0; i <= 1000; i++ {
-		time.Sleep(time.Second / 50)
+	for i := 0; i < timeout; i++ {
 		if check() == http.StatusOK {
 			return
 		}
+		time.Sleep(time.Second * 1)
```
graphaelli: How long does this normally take?  I expect we can bring this way down.

beater/server_test.go
```diff
@@ -290,13 +317,24 @@ func waitForServer(secure bool, host string) {
...
 	}
-	panic("server run timeout (10 seconds)")
+	// for i := 0; i <= timeout*100; i++ {
+	// 	time.Sleep(time.Second / 50)
+	// 	if check() == http.StatusOK {
+	// 		return
+	// 	}
+	// }
```
graphaelli: remove commented out code

beater/server_test.go
```diff
@@ -241,13 +269,12 @@ var testData = func() []byte {
...
-	key := path.Join(tmpCertPath, t.Name()+".key")
-	t.Log("generating certificate in ", cert)
-	httpscerts.Generate(cert, key, domain)
+func withSSL(t *testing.T, domain string, passwordKey string) *SSLConfig {
+	name := path.Join(tmpCertPath, t.Name())
+	t.Log("generating certificate in ", name)
+	transptest.GenCertForTestingPurpose(t, domain, name, passwordKey)
 
-	return &SSLConfig{Cert: cert, PrivateKey: key}
+	return &SSLConfig{Enabled: newTrue(), Certificate: outputs.CertificateConfig{Certificate: name + ".pem", Key: name + ".key", Passphrase: passwordKey}}
```
graphaelli: since this is the only usage of `newTrue()`, a local `truthy` var like elsewhere would be preferred.

beater/server_test.go
graphaelli: Doesn't look like this has been addressed yet.

beater/client.go
sterchelen: Agree with building it outside the for loop. Not agree with removing it. IMO, logged that the server is in secure mode is relatively important to know ?

beater/server_test.go
sterchelen: Do you mean the name of the variable is not intelligible ?

beater/server_test.go
sterchelen: Resolved !

sterchelen:
@graphaelli Does the PR seems good to you ?

sterchelen:
@graphaelli  @ruflin  Any news ?

sterchelen:
Hi !
I will update the PR by the end of the week !
Sounds correct to you ?

jalvz:
Sounds fantastic! 👍

jalvz:
Took this over to https://github.com/elastic/apm-server/pull/1010
Thanks @sterchelen !

PR #240, opened 2017-10-18
Automatic merge from master to 6.0 branch

roncohen:


_beats/dev-tools/packer/version.yml
```diff
@@ -1 +1 @@
-version: "6.0.0-rc2"
+version: "6.1.0"
```
ruflin: I think that will break packaging. I would expect that `update-beats` would overwrite this file and set it to the correct version.

ruflin: @roncohen Just saw that you changed the base branch after update beats. What did you rebase this on? 6.0 or 6.x? If 6.x that would explain the above issue.

roncohen: This was the command: `./_beats/dev-tools/merge_pr master 6.0`

ruflin: Ok, that sounds good and you see afterwards you also updated beats. Need to check what is off here.

PR #244, opened 2017-10-19
support OPTIONS method needed for preflight browser requests

jalvz:


beater/handlers.go
```diff
@@ -133,6 +134,18 @@ func isAuthorized(req *http.Request, secretToken string) bool {
...
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.Method == "OPTIONS" {
+			w.Header().Set("Access-Control-Allow-Origin", "*")
+			w.Header().Set("Access-Control-Allow-Methods", "POST")
+			sendStatus(w, r, 200, nil)
+		} else {
+			h.ServeHTTP(w, r)
+		}
+	})
+}
```
simitt: The `OPTION` request is also a request, just with a different http method.
Therefore I suggest to move the differentiation to the `processRequest` or the `processRequestHandler` method.

jalvz: yes, but i only want it for frontend, so i need a different handler that i can chain only for frontend endpoints

beater/handlers.go
```diff
@@ -133,6 +134,18 @@ func isAuthorized(req *http.Request, secretToken string) bool {
 	return subtle.ConstantTimeCompare([]byte(parts[1]), []byte(secretToken)) == 1
 }
 
+func optionsHandler(h http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.Method == "OPTIONS" {
+			w.Header().Set("Access-Control-Allow-Origin", "*")
+			w.Header().Set("Access-Control-Allow-Methods", "POST")
```
roncohen: @jahtalab Do we need the `Access-Control-Allow-Headers` as well? What should it include?

roncohen: @jahtalab Did you investigate it we can get rid of the preflight OPTIONS all together?

jalvz: hmm i probably need to allow header `Content-Type`...

jahtalab: @roncohen ,
We can't get rid of preflight request, since `Content-Type:application/json` is not considered [cors safelisted request header](https://fetch.spec.whatwg.org/#cors-safelisted-request-header). We'd have to switch to one of `application/x-www-form-urlencoded`, `multipart/form-data`, or `text/plain` to get rid of OPTIONS request.

To recap:
Response to OPTIONS should include these 
`Access-Control-Allow-Origin "*";`
`Access-Control-Allow-Headers "Content-Type";`
`Access-Control-Allow-Methods "POST";`

Note: We have to update `Access-Control-Allow-Headers` if we choose to use custom headers on the agent side but for now the above should be enough.

And the response to the original POST request should include:
`Access-Control-Allow-Origin "*";`

In production mode `Access-Control-Allow-Origin` should be configured by the user to something other than "*"

Docs:
https://fetch.spec.whatwg.org/
https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS

roncohen: @jahtalab is it worth switching to `multipart/form-data` or `application/x-www-form-urlencoded` to get rid of the OPTIONS request? It could be a lot of requests we'd end up saving

jalvz: OPTIONS can be cached by the browser if i send the max age header, if that is a problem.
not sure what is reasonable... 1 day?

roncohen: Maybe there are special cases when browsers will cache the OPTIONS request, but I never managed to get them to do it. When i tried it out they seem to ignore the max-age header.

jalvz:
@roncohen @simitt  what you reviewed was the rather trivial allow `*`, which doesn't do much other than allowing all browser requests.

This now allows users to configure a valid origin, and implements the proper CORS compliant checks.

One open question is if / how to allow multiple origins. Options are:

1. Don't
2. Wildcard pattern matching
3. Regex pattern matching
4. Comma separated list of origins

1 is too limited, and 3 is too much.
4 is better, but if you have many subdomains is a pain. 
2 is good to allow subdomains, but we have to implement wildcard pattern matching ourselves (i think it is ok).

@jahtalab voted 2. 
 2 + 4 could also be interesting (doesn't add that much complexity)

jalvz:
hmmm why CI didn't trigger?

simitt:
@jalvz I'd also vote for either option 2 or 4, or if not too much of a hassle combine them. I think it is good enough to do this in a follow up PR and merge this one in.

The question I asked myself is if we really want to set `*` as default option, since I assume quite some customers won't change the default settings.

jalvz:
re. `*` as default: 

I thought about that too, but if eg. you try `apm-server` in a dev env you probably don't want to have to deal with it and just want it to "work" (which won't be the case anyways because you need to enable frontend, but still)

Maybe something for a bootstrap check in prod, if we ever get around that?

jalvz:
I found https://github.com/ryanuber/go-glob for wildcard pattern matching, i looked into it and code is solid and well tested, it would be quite straightforward to plug it.
Important: if we do that, we also need a `Vary: Origin` header to not break the preflight cache.

beater/config.go
```diff
@@ -15,6 +15,7 @@ type Config struct {
 	SSL                *SSLConfig    `config:"ssl"`
 	ConcurrentRequests int           `config:"concurrent_requests" validate:"min=1"`
 	EnableFrontend     bool          `config:"enable_frontend"`
+	Origin             string        `config:"request_origin"`
```
roncohen: YAML has a way to specify lists. We could probably utilize that to make it simple to add multiple here?

I also think we should name the configuration option something like `AllowOrigin`.

beater/handlers.go
```diff
@@ -133,6 +137,60 @@ func isAuthorized(req *http.Request, secretToken string) bool {
...
+
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+
+		// origin header is always set by the browser
+		origin := r.Header.Get("Origin")
+		validOrigin := allowedOrigin == "*" || origin == allowedOrigin
+
+		if r.Method == "OPTIONS" {
+
+			requestHeaders := r.Header.Get("Access-Control-Request-Headers")
```
roncohen: For the OPTIONS request, it should be enough to simply return a static list of allowed headers and methods. We don't need to actually worry about what headers or methods the browser is asking for.

As long as the OPTIONS response includes what it needs (e.g. `Request-Method` is `POST` and the answer is `Allow-Methods: POST, OPTIONS`) it will allow the POST.

As far as i can tell, this means the logic for splitting `requested` goes away.

beater/config.go
roncohen: Sorry, i realize that is (4) from above. Seems like it would be simple to add, so I'm +1 on that.

roncohen:
Getting 2+4 in sounds great. 

I took a look at go-glob and it looks OK, but I also took a look at https://golang.org/pkg/path/filepath/#Match and this also seems to work quite well:

```golang
ok, err := filepath.Match("*.mydomain.com", "api.mydomain.com");
```

it gives users a bit more options without going full regex and it's in the stdlib, which is think is a big win.

urso:
Some annoyances with `path.Match` is the differences between `\\` on windows and linux (on linux it's an escape character). `path.Match` treats the path separator `/` somewhat special (splits path into segments) while `go-glob` does not treat `/` in anyways special. e.g. `*.mydomain.com/*` will not match `api.mydomain.com/a/b` when `path.Match` is used. But `go-glob` will kind match, as the trailing `/*` behaves more like `/**`.

Question: When configuring `*.mydomain.com`, do we want `mydomain.com` to also match? In both glob engines the `.` matches any, but exactly one character. User will have to use `*mydomain.com` in this case, but the will also match `notmydomain.com` ;)

I'm looking into integrating our own glob support with more proper support for `**` into beats. Here we could make the split-set configurable. e.g. empty split-set == string match. Or use `[./]` as split-set for URLs, so domain names can be matched more properly.

roncohen:
Thanks @urso!

Paths names (everything after the host/domain in this context) is not allowed here, so the special delimiter/slash handling should not matter. Only domain names are pertinent.

I'm OK with `*.mydomain.com` not matching `mydomain.com` (it does match `www.mydomain.com` of course). Either way, we'll need to document it clearly.

jalvz:
4 is done here: 86590b7d288d64a5c24ca59c0131f632f5341fb6

I can do 2 in a separated PR, but I wouldn't go with `filepath.Match`, it's not meant for this. 

Just wildcard matching a list of values is as powerful as we need; anything more than that would be documentation overhead at very least.

roncohen:
@jalvz nice simplification!

One more: If the origin does not match in the OPTIONS, we can simply return a 403 (same as for POST).

roncohen:
OK let's push (2) to a separate PR 👍

jalvz:
> If the origin does not match in the OPTIONS, we can simply return a 403

That is not correct, it wouldn't even be cached. OPTIONS is meant to tell the browser if it can send or not a request; saying no is a perfectly valid answer and is expected to be 200 OK.

The way to tell the browser to not proceed is by not echoing the origin in the (required) ACAO header.

beater/handlers.go
```diff
@@ -133,6 +137,55 @@ func isAuthorized(req *http.Request, secretToken string) bool {
 	return subtle.ConstantTimeCompare([]byte(parts[1]), []byte(secretToken)) == 1
 }
 
+func corsHandler(allowedOrigins []string, h http.Handler) http.Handler {
+
+	var isAllowed = func(origin string) bool {
+		for _, allowed := range allowedOrigins {
+			if origin == allowed {
```
roncohen: I'd do:

```
if allowed == "*" || allowed == origin {
    return true
}
```

that way, if someone adds a `*` at the end of the list, it will work as expected.

roncohen: and then 148 would just be: `return false`

beater/handlers.go
```diff
@@ -133,6 +137,55 @@ func isAuthorized(req *http.Request, secretToken string) bool {
...
+			}
+
+			// tell browsers to cache response requestHeaders for up to 1 hour (browsers might ignore this)
+			w.Header().Set("Access-Control-Max-Age", "3600")
+			// origin must be part of the cache key so that we can handle multiple allowed origins
+			w.Header().Set("Vary", "Origin")
+
+			// required if Access-Control-Request-Method and Access-Control-Request-Headers are in the requestHeaders
+			w.Header().Set("Access-Control-Allow-Methods", supportedMethods)
+			w.Header().Set("Access-Control-Allow-Headers", supportedHeaders)
```
roncohen: I think we need these on the POST as well. @jahtalab can you confirm?

tests/system/test_requests.py
```diff
@@ -129,3 +129,46 @@ def test_ok(self):
...
         r = requests.post(self.transactions_url, json=transactions)
         assert r.status_code == 202, r.status_code
+
+
+class CorsTest(CorsBaseTest):
+
+    def test_ok(self):
+        transactions = self.get_transaction_payload()
+        r = requests.post(self.transactions_url, json=transactions, headers={'Origin': 'http://www.elastic.co'})
+        assert r.status_code == 202, r.status_code
```
roncohen: Could use an assert of the `Access-Control-Allow-Origin == 'http://www.elastic.co'`

roncohen:
OK @jalvz, let's stick to just removing the ACAO header for the OPTIONS request.

beater/server_test.go
```diff
@@ -74,12 +74,28 @@ func TestServerFrontendSwitch(t *testing.T) {
...
+	req, _ := http.NewRequest("POST", FrontendTransactionsURL, bytes.NewReader(testData))
+
+	apm.Handler = newMuxer(Config{EnableFrontend: true,
+		AllowOrigins: []string{"http://notmydomain.com", "http://neitherthisone.com"}},
+		nil)
+
+	rec := httptest.NewRecorder()
+	apm.Handler.ServeHTTP(rec, req)
+
+	assert.Equal(t, 403, rec.Code, rec.Body.String())
```
roncohen: This is testing that if we respond with 403 if there's no `Origin` set, AFAICT. Seems like that's already tested by the Python code in `test_no_origin`.

beater/handlers.go
jahtalab: @roncohen , We only need Access-Control-Allow-Origin on the POST request, As I summarized in another comment

beater/handlers.go
jalvz: `http.example1.com, http.example2.com, *` is not valid semantics; it doesn't make sense. ACAO is either `*` (any) or an origin as described in spec:
`* | <origin>` 
and we extend it server side to `* | <origin ...>`

current implementation reflects that

roncohen: @jalvz this is not about the spec, this is about configuration. If you have a list of AllowedOrigin domains and then decide to add `*` because something isn't working as expected, or you want to temporarily allow all domains, it will not do what you expect.

beater/server_test.go
```diff
@@ -74,12 +74,28 @@ func TestServerFrontendSwitch(t *testing.T) {
 	apm.Handler.ServeHTTP(rec, req)
 	assert.Equal(t, 403, rec.Code, rec.Body.String())
 
-	apm.Handler = newMuxer(Config{EnableFrontend: true}, nil)
+	apm.Handler = newMuxer(Config{EnableFrontend: true, AllowOrigins: []string{"*"}}, nil)
```
simitt: As you want to test the `FrontendSwitchHandler` I suggest to also set `AllowOrigins` for the test that returns a 403. Otherwise it could be that it is forbidden because of the origin.

beater/server_test.go
simitt: The python tests test against a running apm-server instance. Those tests should be rather high level as they usually take longer to run. 
I think testing that all headers are set properly and testing through all the different combinations of cors handling should be test in unit tests.

beater/handlers.go
jalvz: >  it will not do what you expect
it actually does
i know is configuration, what i meant is that our config *extends* the spec from `* or origin` to `* or list of origins`; as something like `origin1, *, origin2` doesn't make any sense to me. 
what users expect will depend on how we document it.

i changed it now.

does this PR need any further changes?

roncohen:
Nice @jalvz! Getting close.

I think we should get into the habit of writing the docs alongside the actual code changes, so we don't forget. I think for now it would be enough to add an entry in the `beat.reference.yml` file explaining this new setting.

_meta/beat.reference.yml
```diff
@@ -20,3 +20,10 @@ apm-server:
   #ssl.enabled: false
   #ssl.certificate : "path/to/cert"
   #ssl.key : "path/to/private_key"
+
+  # Comma separated list of permitted origins for frontend. User-agents will send
+  # a origin header that will be validated against this list.
+  # An origin is made of a protocol scheme, host and port, without the url path.
+  # Allowed origins in this setting can have * to match anything (eg.: http://*.example.com)
```
roncohen: Nice!

Would `http://*.example.com` actually work? Haven't seen it before and can't find that in the spec. Seems to allow only either the exact domain or the wildcard `*`, but not the mix. Maybe I missed something?

jalvz: not yet, that's work for another PR...

> 2. Wildcard pattern matching
> I can do 2 in a separated PR, but I wouldn't go with filepath.Match...
> OK let's push (2) to a separate PR

roncohen: OK. Let's add this line when it works. If you remove it for now, I'll merge 💪

jalvz: done....

roncohen:
Great work @jalvz !

PR #245, opened 2017-10-23
test frontend switch handler

jalvz:


tests/system/test_requests.py
```diff
@@ -55,6 +54,11 @@ def test_validation_2_fail(self):
         assert r.status_code == 400, r.status_code
         assert "Problem validating JSON document against schema" in r.content, r.content
 
+    def test_frontend_disabled(self):
```
roncohen: I think this should make it clear that the default is "off". Could be achieved by just renaming test name to `test_frontend_default_disabled` or similar.

roncohen:
Looks good @jalvz! Only one comment.

PR #246, opened 2017-10-24
Make context.app.language.version optional.

### LINKED ISSUE #247: Make `context.app.language.version` optional
The `context.app.language.version` field is currently required if `context.app.language.name` is set. 

After an internal discussion we have decided that it should be optional. 

That will allow @watson to implement the `name` field for the Node.js agent (https://github.com/elastic/apm-agent-nodejs/issues/6).

simitt:
As just discussed detecting a version for a language is not always possible, so we should make the `language.version` field optional to support having only a `language.name` given.
closes #247 
@beniwohli , @watson fyi.


docs/spec/app.json
```diff
@@ -48,7 +48,7 @@
                     "maxLength": 1024
                 }
             },
-            "required": ["name", "version"]
```
jalvz: version type shouldn't be also ["string", "null"] ?

tests/data/invalid/app/no_lang_version.json
```diff
@@ -1,10 +0,0 @@
-{
-    "name": "app1",
-    "agent": {
-      "name": "python",
-      "version": "1.0"
```
jalvz: maybe also remove the version from the valid json, to make sure it works without it, but if you tested it is fine

docs/spec/app.json
simitt: will add it

Issue #214, opened 2017-09-29
Request Per Minute [APM] visualization uses wrong field 


jjfalling:
Request Per Minute [APM] should use transaction.result as the filter field for the status code (or APM is sending the wrong field ;) )

watson:
Yes, I think you're right. The dashboards are probably using `context.response.status_code` at the moment. This field might not always be present (whereas `transaction.result` should be).

The `transaction.result` field is a string and there's no rule that it should conform to look like an HTTP status code.

Also, even if the two fields are both present, there's no guarantee that they are the same. I think we even discussed if `transaction.result` - in case of an incoming http request - should be converted from the string `200` to `2xx` already in the agent instead of when viewed in Kibana?

simitt:
@beniwohli and @watson what do you think about making `transaction.result` a mandatory field? This way we could always rely on this field for visualizations.

beniwohli:
I'm ok with that!

watson:
Sounds like a good idea 👍

simitt:
Thank you @jjfalling for bringing this up, we changed the dashboards to use `transaction.result` now.

PR #250, opened 2017-10-25
Make transaction.result a mandatory field.

### LINKED ISSUE #237: Make `transaction.result` mandatory.
Should be mandatory, as follow up on Issue #214 .

simitt:
closes #237 

@beniwohli , @watson please let us know when we you have added this as a mandatory field so we can merge this PR.

### DIFF
--- a/docs/data/intake-api/generated/transaction/minimal_app.json
+++ b/docs/data/intake-api/generated/transaction/minimal_app.json
@@ -15,6 +15,7 @@
             "name": "GET /api/types",
             "type": "request",
             "duration": 32.592981,
+            "result": "200",
             "timestamp": "2017-05-30T18:53:27.154Z"
         }
     ]
--- a/docs/data/intake-api/generated/transaction/minimal_payload.json
+++ b/docs/data/intake-api/generated/transaction/minimal_payload.json
@@ -11,6 +11,7 @@
             "id": "945254c5-67a5-417e-8a4e-aa29efcbfb79",
             "name": "GET /api/types",
             "type": "request",
+            "result": "success",
             "duration": 32.592981,
             "timestamp": "2017-05-09T15:04:05.999999Z"
         }
--- a/docs/data/intake-api/generated/transaction/minimal_trace.json
+++ b/docs/data/intake-api/generated/transaction/minimal_trace.json
@@ -11,6 +11,7 @@
             "id": "945254c5-67a5-417e-8a4e-aa29efcbfb79",
             "name": "GET /api/types",
             "type": "request",
+            "result": "success",
             "duration": 32.592981,
             "timestamp": "2017-05-30T18:53:27.154Z",
             "traces": [
--- a/docs/data/intake-api/generated/transaction/null_values.json
+++ b/docs/data/intake-api/generated/transaction/null_values.json
@@ -38,6 +38,7 @@
             "type": "request",
             "duration": 13.980558,
             "timestamp": "2017-05-30T18:53:42Z",
+            "result": "200",
             "context": {
                 "request": null,
                 "response": null,
@@ -52,6 +53,7 @@
             "type": "request",
             "duration": 13.980558,
             "timestamp": "2017-05-30T18:53:42.281999Z",
+            "result": "200",
             "context": {
                 "request": {
                     "socket": null,
@@ -91,6 +93,7 @@
             "type": "request",
             "duration": 13.980558,
             "timestamp": "2017-05-30T18:53:42.281Z",
+            "result": "200",
             "context": {
                 "request": {
                     "socket": {
--- a/docs/data/intake-api/generated/transaction/payload.json
+++ b/docs/data/intake-api/generated/transaction/payload.json
@@ -36,6 +36,7 @@
             "name": "GET /api/types",
             "type": "request",
             "duration": 32.592981,
+            "result": "success",
             "timestamp": "2017-05-30T18:53:27.154Z",
             "result": "200",
             "context": {
@@ -183,6 +184,7 @@
             "name": "GET /api/types",
             "type": "request",
             "duration": 13.980558,
+            "result": "failure",
             "timestamp": "2017-05-30T18:53:42.281Z",
             "traces": []
         },
@@ -191,6 +193,7 @@
             "name": "GET /api/types",
             "type": "request",
             "duration": 13.980558,
+            "result": "200",
             "timestamp": "2017-05-30T18:53:42Z"
         },
         {
@@ -198,6 +201,7 @@
             "name": "GET /api/types",
             "type": "request",
             "duration": 13.980558,
+            "result": "200",
             "timestamp": "2017-05-30T18:53:42.281999Z",
             "traces": [
                 {
--- a/docs/data/intake-api/generated/transaction/system_null.json
+++ b/docs/data/intake-api/generated/transaction/system_null.json
@@ -13,6 +13,7 @@
             "name": "GET /api/types",
             "type": "request",
             "duration": 32.592981,
+            "result": "200",
             "timestamp": "2017-05-09T15:04:05.999999Z",
             "context": null,
             "traces": null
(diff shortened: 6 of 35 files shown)


beniwohli:
In the Python agent, this field is already always set

watson:
@beniwohli I was about to say the same for the Node agent, but then remembered that people can make custom transactions where this might not be the case. Isn't that possible in the Python agent as well?

watson:
The Node.js agent currently defaults the `result` to `200`, so if the user makes a custom transaction but doesn't set the `result`, then there shouldn't be an issue. But the user could accidentally set the result to `null`. What should we do in that case? Should the APM Server just reject the request or should the agent refuse the process the transaction?

simitt:
How about setting a default value in such a case, like `unset`?

watson:
@simitt I can do that. Is `unset` a good value? @beniwohli?

watson:
I just had a look at the new tests in this PR and I think we should add a test for when `result` is set to `null`, just to make sure we don't allow that

watson:
In general, most fields that we require are more likely to be `null` than to be missing since the agents will be hardwired to set them, but might set them to `null` by accident.

beniwohli:
@watson I don't think we specifically need to have a default value. If it's not set, the APM Server will reject the transaction with a nice error message, telling the user what to do. As you pointed out, we have several required fields that theoretically can be set to `null` by the user, so I don't think we need to have special handling for `transaction.result`

watson:
@beniwohli the APM server will reject the entire request that in most cases contain multiple transactions - not only the single transaction, right?

beniwohli:
Hmm, that's true I believe. I can live with a default setting of `unset`.

simitt:
As @beniwohli pointed out this would be the same for every mandatory field that sends a value `null`.

watson:
I just had a chat with @roncohen and we came to the conclusion that we should default to a value, e.g. `unset` or something similar in most cases where we have a required field and we don't allow `null`. Otherwise the user might end up loosing a lot of valid data that could have been stored simly because it was submitted in the same batch as a transaction with an invalid `null` value

watson:
Are an empty string, `false`, or `0` allowed result values?

beniwohli:
I'm having difficulties with this decision to be honest. A value is either required or it isn't. Introducing some magic `this_is_supposed_to_be_required_but_¯\_(ツ)_/¯` value will bite us. Just ask MySQL...

I'd prefer if the server could validate the payload on a per-transaction-basis and only discard those transactions that fail. If that's not an option, I'd rather get rid of the required-ness then introducing a magic default value.

watson:
@beniwohli While I understand your concern, I think there's two levels to this and it might help to see them separately:

1. What is required by the APM Server API
1. What does the agents help with to give the best possible experience for our users

I don't feel that just because we have the field required in the API that the agents can't help out a little to ensure that these requirements are met. Especially in situations where it's really easy for users to make a mistake.

Alternatively we could just make it so that the UI defaults to `unset` if there's no result instead of the agents defaulting to it. That way we don't need to make the `result` property required at all. Would that be a possibility @sqren?

simitt:
Although I suggested a default value, I also do not really see the advantage of having a `required` field, but then setting dummy values on the agent. 

Not sure if we should verify requests on a per-transaction-base. In this case we would need to return a proper response, with an error message for every invalid transaction in the request. This seems quite some overhead, where I don't see a lot of gain right now. 

I think we have a couple of different options here:
*mandatory keys in json schema*:
- Agents do not allow user to override those values with null values.
   From my perspective this would be a clean solution, but not sure how easy this is on the agent side.
- Agents provide dummy value in case value is overwritten by user.
   Would actually not make too much sense, as it makes the user believe to unset the value, but he actually doesnt.
- Whole request fails in case key is not provided.
   Could lead to really bad user experience, as also valid data are not stored.
- Only transactions that to not fulfill requirements fail.
   Could also lead to bad user experience, as expected data are missing.

*no mandatory keys in json schema*:
- Validate that agents send all expected keys by having tests for this in place. Full freedom for customization with the downside of potentially broken UI in case expected values are not provided.

sqren:
> Alternatively we could just make it so that the UI defaults to unset if there's no result instead of the agents defaulting to it. That way we don't need to make the result property required at all. Would that be a possibility @sqren?

The only place we use `transaction.result` is for the rpm chart where value ranges are grouped (200-299, 300-399 etc.)
screen shot 2017-10-25 at 17 22 00

So if a value is `null`, or `unset` it just won't affect the data. We talked about showing an "Other" graph, where I suppose `unset` values would belong.

So the answer is: yes, it's possible, but right now we are not going to show values outside of the defined ranges.

sqren:
If we end up making this change, can we also change 'transaction.result' to be an integer instead of a string? Currently we rely on [lexicographic sorting](https://www.elastic.co/guide/en/elasticsearch/guide/current/_ranges.html#_ranges_on_strings)

This is both slow, and will make some queries harder/impossible to perform. 

(sorry for highjacking with something tangential)

simitt:
@sqren thank you for the input.
We specifically have a `string` here, e.g. for allowing non-web transactions to send values like `sucess`, etc.

jalvz:
for what is worth I agree with @beniwohli. 
Validate each transaction is not trivial, but doable - and the right thing to do.

beniwohli:
I had a look through the spec with regards to required fields. In my mind, there are two categories here that we need to look at separately:

 * user-settable fields
 * fields set by the agent

User-settable fields are relatively few: `app.name`, `transaction.name`, `transaction.type`, `trace.name`, `trace.type` (please correct me if I overlooked any)

Required fields set by the agent are quite a bit more.

`app.name` is a bit special, but for the rest, I think it's OK to define defaults that are used if the user doesn't provide a value, and fail loudly if they provide a null value (ideally as a log message at set-time, not as a validation error at parse time on the server).

As for required fields that are not user-defined, any null values there are IMO a bug in the agent.

beniwohli:
@sqren regarding type of `transaction.result`, going back to integer is not really an option as @simitt explained. However, I think we should turn the HTTP status code into `HTTP Yxx` already on the agent, then you don't need to do anything fancy with they keyword.

sqren:
> However, I think we should turn the HTTP status code into HTTP Yxx already on the agent, then you don't need to do anything fancy with they keyword.

Not sure about this. What if we want to do some other aggregations in the future, eg. errors in general (400-599)? We won't be able to do that. @simianhacker also mentioned that we won't be able to make histograms, since they only for on numeric fields.

There are probably workarounds/hacks to these, and again: I understand your arguments for a string-based result field. Just want to make sure we make the decision with open eyes, and understand the implications it has for querying.

beniwohli:
@sqren the full HTTP status is also available in `context.response.status`.

What kind of histograms do we make where the `transaction.status` plays a role?

simitt:
We do have following fields:
- `transaction.result`, string, should always have a value even when instrumenting non-web applications, for web applications it is probably aligned to be the `http` status code
- `context.response.status_code`, number, reflects the `http` status code for instrumenting web applications

As long as we have those two fields I don't understand the discussion about the data type.

sqren:
> What kind of histograms do we make where the transaction.status plays a role?

Today none. In the future we might.

> @sqren the full HTTP status is also available in context.response.status.

That's a good point, and I thought about that. it's paradoxical that this conversation started because the dashboards were using `context.response.status_code` but [were changed](https://github.com/elastic/apm-server/issues/214) to use `transaction.result` since the former is not always available.

If we can use `context.response.status_code` instead, we'll just do that. Then again, I'm not sure why the dashboards were not able to do the same?

simitt:
As a summary, we just discussed to have this field mandatory and that the agents ensure a value is sent. In case the user overwrites with a null value, some default value is sent. 
@beniwohli , @watson can you please ping when the agents have this implemented so we can merge it?

watson:
Currently the Node.js agent converts any value given as the transaction result into a string. So if the value is `null`, it's simply converted to the string `"null"` or if the value is the number `0` it's converted to the string `"0"`. I actually kind of like that approach as it allows the user to see what the result was actually set to. Wouldn't that be ok?

PR #251, opened 2017-10-25
Add test for updating beats framework.

simitt:
Run testsuite when beats framework is updated to ensure compatibility with the beats framework.

This gets triggered after a PR is merged into `beats` `master` or `release` branch, see https://github.com/elastic/infra/pull/3006, and allows to get rid of the dependency in `beats` (https://github.com/elastic/beats/pull/5099).


jalvz:
🎉

PR #253, opened 2017-10-25
Add tests for json schema allowing null values

simitt:
@beniwohli , @watson I added one big testfile for `transaction` and one for `error` to check that all optional keys can actually be set to `null`. Found a few issues and fixed them.
related to https://github.com/elastic/apm-server/pull/250#issuecomment-339286364

The number of changed files when making some minor change to the schema is huge, since additionally to manually adapt the according test files, we also
- generate schema `go` files from it
- generate documentation out of test files
- generate approved files
The generation of those files is convenient, but leads to huge PRs for minor changes.


beniwohli:
@simitt for clarification, is the resulting document going to be the same, no matter if the field is missing or set to `null`? Or will the JSON that is sent to ES reflect the fact that the field was there, but set to `null`?

simitt:
@beniwohli I am currently looking into that.
The created json document reflects _some_ of the values set to nil. Afaik the beats framework does cut off all `null` values underneith. If this is true the same doc would be sent to ES. But then we also might change some internal behavior.

watson:
Is it a good idea to allow `null` in so many places?

simitt:
As far as I remember it was explicitly asked to allow either `null` values or not providing a key at all for all non-mandatory fields. 
From the `apm-server` side it would allow for a bit easier testing and handling if we'd say a non-mandatory field still needs to have a value in case the key is sent. But it doesn't make a huge difference.

watson:
Not sure why, but I just feel it's a bit "sloppy" to allow `null`. It makes the document larger and we end up storing a lot of empty fields in the database. If possible I'd much rather restrict us. But there might be a good reason why we allow `null`?

simitt:
@beniwohli if I remember correctly you requested to also allow null values for optional fields. 
In case a serializer does not remove null values I can see the advantage.

simitt:
The `null` values are treated as following:
- For the models where a specific `type struct` is defined, `empty` and `null` values get removed by the processor.
- For other models `null` values get removed by the beats logic. In case all values within an object are null this leads to `empty` objects, that get strored in ES.

We could either 
- forbid `null` values at all
- explicitly remove them in the processor logic for all models
- leave it as is (which I would not recommend)

From a processing side node allowing null values would be the most efficient thing.

simitt:
After some more discussions, we will allow sending null values for optional fields. They are not indexed in ES and removed by either the processor or beats anyways.

jalvz:
> After some more discussions, we will allow sending null values for optional fields

what was the reason, finally?

processor/error/schema.go
```diff
@@ -380,7 +380,7 @@ var errorSchema = `{
                     "description": "The record severity.",
                     "type": ["string", "null"],
                     "default": "error",
-                    "enum": ["debug", "info", "warning", "error", "fatal"],
+                    "enum": ["debug", "info", "warning", "error", "fatal", null],
```
jalvz: is this one necessary, if type is already `string, null` ?

simitt: yes unfortunately it is.

simitt:
> what was the reason, finally?

We allways allowed sending null values, just haven't had specs for it. 
When adding the specs the question came up if we really want to allow it. Since the standard serializers do not remove null values in common libraries for python and node.js we decided to still allow it. Also it doesn't make sense to forbid sending null values when the key itself is already optional.

PR #254, opened 2017-10-25
Add isAuthorized unit tests

kalafut:
Hi @simitt and @jalvz .  Per our discussion, I was playing around with a fail delay in isAuthorized(). I ultimately decided that such a delay could actually cause more problems than it would prevent in the case of a highly parallel attack (tying up goroutines).  But I'd already written up the unit tests for that function, so feel free to use them if you want.

Thanks,
Jim


kalafut:
CLA is now signed.

simitt:
jenkins, test it

jalvz:
@kalafut looks good, thanks for contributing!

PR #257, opened 2017-10-26
Rate limiter per IP for frontend agents

jalvz:
This is cherry-picked from https://github.com/elastic/apm-server/pull/207, which is somewhat old so after merging it doesn't look much like it anymore.

The main goal of this is to add per IP rate limiting. 

Approach in #207 with `context.Wait` is smart, but it becomes hairy when you need one `ctx` per IP, a call to `Allow` is much simpler. 

Main functional difference is that in #207, is that if the bucket is refilled with a token within `readTimeout` seconds, then the request goes trough; in here it gets rejected right away.
I don't consider this a substantial difference.

For now, i also got rid of the burst to avoid over-configuration.

### DIFF
--- a/_meta/beat.reference.yml
+++ b/_meta/beat.reference.yml
@@ -22,6 +22,8 @@ apm-server:
   #ssl.key : "path/to/private_key"
 
   #frontend.enabled: false
+
+  # Rate Limit per second and IP address
   #frontend.rate_limit: 10
 
   # Comma separated list of permitted origins for frontend. User-agents will send
--- a/_meta/beat.reference.yml
+++ b/_meta/beat.reference.yml
@@ -21,8 +21,11 @@ apm-server:
   #ssl.certificate : "path/to/cert"
   #ssl.key : "path/to/private_key"
 
+  #frontend.enabled: false
+  #frontend.rate_limit: 10
+
   # Comma separated list of permitted origins for frontend. User-agents will send
   # a origin header that will be validated against this list.
   # An origin is made of a protocol scheme, host and port, without the url path.
   # If an item in the list is a single *, everything will be allowed
-  #allow_origins : *
+  #frontend.allow_origins : *
--- a/beater/config.go
+++ b/beater/config.go
@@ -5,17 +5,23 @@ import (
 )
 
 type Config struct {
-	Host               string        `config:"host"`
-	MaxUnzippedSize    int64         `config:"max_unzipped_size"`
-	MaxHeaderBytes     int           `config:"max_header_bytes"`
-	ReadTimeout        time.Duration `config:"read_timeout"`
-	WriteTimeout       time.Duration `config:"write_timeout"`
-	ShutdownTimeout    time.Duration `config:"shutdown_timeout"`
-	SecretToken        string        `config:"secret_token"`
-	SSL                *SSLConfig    `config:"ssl"`
-	ConcurrentRequests int           `config:"concurrent_requests" validate:"min=1"`
-	EnableFrontend     bool          `config:"enable_frontend"`
-	AllowOrigins       []string      `config:"allow_origins"`
+	Host               string          `config:"host"`
+	MaxUnzippedSize    int64           `config:"max_unzipped_size"`
+	MaxHeaderBytes     int             `config:"max_header_bytes"`
+	ReadTimeout        time.Duration   `config:"read_timeout"`
+	WriteTimeout       time.Duration   `config:"write_timeout"`
+	ShutdownTimeout    time.Duration   `config:"shutdown_timeout"`
+	SecretToken        string          `config:"secret_token"`
+	SSL                *SSLConfig      `config:"ssl"`
+	ConcurrentRequests int             `config:"concurrent_requests" validate:"min=1"`
+	EnableFrontend     bool            `config:"enable_frontend"`
+	Frontend           *FrontendConfig `config:"frontend"`
+}
+
+type FrontendConfig struct {
+	Enabled      *bool    `config:"enabled"`
+	RateLimit    int      `config:"rate_limit"`
+	AllowOrigins []string `config:"allow_origins"`
 }
 
 type SSLConfig struct {
@@ -28,6 +34,10 @@ func (c *SSLConfig) isEnabled() bool {
 	return c != nil && (c.Enabled == nil || *c.Enabled)
 }
 
+func (c *FrontendConfig) isEnabled() bool {
+	return c != nil && (c.Enabled == nil || *c.Enabled)
+}
+
 var defaultConfig = Config{
 	Host:               "localhost:8200",
 	MaxUnzippedSize:    10 * 1024 * 1024, // 10mb
@@ -37,5 +47,5 @@ var defaultConfig = Config{
 	WriteTimeout:       2 * time.Second,
 	ShutdownTimeout:    5 * time.Second,
 	SecretToken:        "",
-	AllowOrigins:       []string{"*"},
+	Frontend:           &FrontendConfig{Enabled: new(bool), RateLimit: 10, AllowOrigins: []string{"*"}},
 }
--- a/beater/config.go
+++ b/beater/config.go
@@ -14,7 +14,6 @@ type Config struct {
 	SecretToken        string          `config:"secret_token"`
 	SSL                *SSLConfig      `config:"ssl"`
 	ConcurrentRequests int             `config:"concurrent_requests" validate:"min=1"`
-	EnableFrontend     bool            `config:"enable_frontend"`
 	Frontend           *FrontendConfig `config:"frontend"`
 }
 
--- a/beater/handlers.go
+++ b/beater/handlers.go
@@ -126,7 +126,9 @@ func ipRateLimitHandler(rateLimit int, h http.Handler) http.Handler {
 	cache, _ := lru.New(rateLimitCacheSize)
 
 	var deny = func(ip string) bool {
-		cache.ContainsOrAdd(ip, rate.NewLimiter(rate.Limit(rateLimit), rateLimit*rateLimitBurstMultiplier))
+		if !cache.Contains(ip) {
+			cache.Add(ip, rate.NewLimiter(rate.Limit(rateLimit), rateLimit*rateLimitBurstMultiplier))
+		}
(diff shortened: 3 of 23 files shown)


beater/handlers.go
```diff
@@ -102,15 +107,64 @@ func frontendSwitchHandler(feSwitch bool, h http.Handler) http.Handler {
 		if feSwitch {
 			h.ServeHTTP(w, r)
 		} else {
-			sendStatus(w, r, 403, errForbidden)
+			sendStatus(w, r, http.StatusForbidden, errForbidden)
```
simitt: nice!

beater/handlers.go
```diff
@@ -102,15 +107,64 @@ func frontendSwitchHandler(feSwitch bool, h http.Handler) http.Handler {
...
+		}
+	})
+}
+
+func rateLimitHandler(rateLimit int, h http.Handler) http.Handler {
```
simitt: Can we call it something like `rateLimitPerIpHandler` as we'll probably need a more generic one later on for backend?

beater/handlers.go
```diff
@@ -102,15 +107,64 @@ func frontendSwitchHandler(feSwitch bool, h http.Handler) http.Handler {
...
+		return strings.TrimSpace(client)
+	}
+
+	var real = func() string {
+		return r.Header.Get("X-Real-IP")
+	}
+
+	for _, s := range []string{remoteAddr(), forwarded(), real()} {
+		if s != "" {
+			ip = s
```
simitt: this way you are overwriting the `ip` in every iteration. How about 
```
if s != ""{ 
  return s
}
```

roncohen:
nice work @jalvz!

some high level thoughts:
* I absolutely agree on keeping it simple with `Allow`. It's also more resource friendly as connections can be dropped instantly
* the config name should indicate this is for frontend endpoints, it's not in effect for other endpoints
* do we have any indication of whether 10k is the right LRU size? Some idea of what 10k rate limiters cost memory and CPU wise? they all have a Mutex and a timer, AFAICT.

jalvz:
thanks for feedback, good points 

LRU size: i don't know tbh. maybe 1k or even less is good enough. each limiter has mutex, but not a timer. it compares last event timestamp with current time to know how many tokens to fill the bucket with, it doesn't look a heavy structure.

should we consider also a global rate limiter for both backend and frontend? if so, i think would be safe to decrease the cache size so i'm 👍  on it

burst: what i don't want is a configuration for this, it is testing overhead, documentation overhead, etc. i'm ok with hardcoding something. how much is reasonable? 2X?

beater/handlers.go
```diff
@@ -102,15 +107,64 @@ func frontendSwitchHandler(feSwitch bool, h http.Handler) http.Handler {
...
+
+func rateLimitHandler(rateLimit int, h http.Handler) http.Handler {
+
+	cache, _ := lru.New(10000)
+
+	var deny = func(ip string) bool {
+		if rateLimit < 0 {
+			return false
+		}
+		if !cache.Contains(ip) {
```
roncohen: Looks like there's a `ContainsOrAdd`.

beater/handlers.go
roncohen: I have a big dislike for named return values because it often makes simple things complicated. Here's an alternative:

```golang
for _, s := range []string{real(), forwarded(), remoteAddr()} {
    if s != "" {
        return s
    }
}
return ""
```

roncohen:
> should we consider also a global rate limiter for both backend and frontend? if so, i think would be safe to decrease the cache size so i'm 👍 on it

How would it work? I think we should focus on the immediate problem, which is frontend rate limiting - and i think what we have is a good start. But we need to update the config name to indicate this is for frontend logging.

_meta/beat.reference.yml
```diff
@@ -20,3 +20,6 @@ apm-server:
   #ssl.enabled: false
   #ssl.certificate : "path/to/cert"
   #ssl.key : "path/to/private_key"
+
+  # Rate limiter to limit number of events per second. 0 means disabled.
+  #ratelimit.size: 0
```
roncohen: Setting a default rate limit is difficult, but I think we should something large here so it will really only matter if something in the agent is broken or if there's a malicious usage. For example 10 per second seems like a high enough number that you'll never run into the rate limit under normal usage.

roncohen:
I couldn't build it. It needs `github.com/hashicorp/golang-lru/simplelru` in the vendor tree, but `simplelru` never made it there.

jalvz:
> But we need to update the config name to indicate this is for frontend

Beats doesn't solve this with naming. We have `enable_frontend` already,  which means that we should have a nested `frontend` config with `enabled` and `rate_limit` attributes, like eg. SSL.

The problem then is that beats `isEnabled` logic (`c != nil && (c.Enabled == nil || *c.Enabled)`) won't fit here because we want frontend disabled by default, which means that `frontend.enabled` must exist _and_ be `false` (which is weird?). 

We could have a different logic, but then it wouldn't be consistent with the rest (ssl, outputs, etc).

roncohen:
@jalvz good point regarding the frontend namespace. Is there no way to have a `frontend.enabled` and make it default to false?

jalvz:
>> should we consider also a global rate limiter for both backend and frontend? if so, i think would be safe to decrease the cache size so i'm 👍 on it

> How would it work? 

it would work the same but for all requests and without per IP. If we get 9 reqs/sec from 1 gazillion different ip we need to protect the apm-server as well.

jalvz:
jenkins retest

jalvz:
jenkins retest this please

jalvz:
jenkins, retests this please

jalvz:
jenkins retest please

jalvz:
jenkins retest this

jalvz:
@roncohen @simitt this is rebased and green now, what else i need to do here?

changelog is missing but i figured out that better create a separated pr with a fully updated changelog; ill try to go through all PR's since last entry and see what needs to be added.

beater/handlers.go
```diff
@@ -106,15 +116,62 @@ func frontendSwitchHandler(feSwitch bool, h http.Handler) http.Handler {
...
+		}
+	})
+}
+
+func ipRateLimitHandler(rateLimit int, h http.Handler) http.Handler {
+
+	cache, _ := lru.New(rateLimitCacheSize)
+
+	var deny = func(ip string) bool {
+		cache.ContainsOrAdd(ip, rate.NewLimiter(rate.Limit(rateLimit), rateLimit*rateLimitBurstMultiplier))
```
roncohen: I see now that having a separate `Contains` and then `Add` allows us to initialize way fewer rate limiters. This now effectively means we'll create a rate limiter for every request that goes into this handler.  I looked into what the cost of creating those are. I was mostly worried about the `sync.Mutex` that gets created every time, but it looks like it shouldn't be that expensive.

I'm OK with changing back to `Contains` and then `Add`, but also OK to stay with this.

beater/config.go
```diff
@@ -5,17 +5,23 @@ import (
...
+	Host               string          `config:"host"`
+	MaxUnzippedSize    int64           `config:"max_unzipped_size"`
+	MaxHeaderBytes     int             `config:"max_header_bytes"`
+	ReadTimeout        time.Duration   `config:"read_timeout"`
+	WriteTimeout       time.Duration   `config:"write_timeout"`
+	ShutdownTimeout    time.Duration   `config:"shutdown_timeout"`
+	SecretToken        string          `config:"secret_token"`
+	SSL                *SSLConfig      `config:"ssl"`
+	ConcurrentRequests int             `config:"concurrent_requests" validate:"min=1"`
+	EnableFrontend     bool            `config:"enable_frontend"`
```
roncohen: This should be removed.

_meta/beat.reference.yml
```diff
@@ -21,8 +21,11 @@ apm-server:
   #ssl.certificate : "path/to/cert"
   #ssl.key : "path/to/private_key"
 
+  #frontend.enabled: false
+  #frontend.rate_limit: 10
```
roncohen: You used to have a little note here explaining that it's "per second" and that zero means disabled. Would be great to get that back.

roncohen:
OK to have the changelog for all the missing ones in another PR 👍

roncohen:
We also need to document the whole `frontend.enabled` setting. Explain that it opens up another set of endpoints that are not protected by the secret token etc.
WDYT @jalvz ?

_meta/beat.reference.yml
jalvz: per seconds: yes, absolutely
zero means disabled: i wanted to use that as configuration for enable/disable frontend, but it was rejected. it seams silly now to say you can use `frontend.enabled = false` _and/or_  `frontend.rate_limit = 0` to disable frontend.

beater/config.go
jalvz: 🤦‍♂️

beater/handlers.go
jalvz: that limiter probably goes to the heap. I'll change it back.

jalvz:
We need to document all-things-frontend. But Il also do it in another frontend documentation PR, maybe when we have a working agent, more tests and we know a bit better what we are doing etc.

jalvz:
done 217c7358963ae91617252f0b606b1a7c887b1a0f

_meta/beat.reference.yml
roncohen: what do you suggest is a good way to disable rate limiting?

jalvz: what it was considered is `rate_limit = 0` to disabled frontend. 
i don't think there was a note or option to have frontend enabled and rate limit disabled. you mean that? why would someone _want_ that? 
still, if you want it, set it to a really high number like `10000000`, but i think is silly to document that.

roncohen: yeah, that's what I mean. I think we'll have some people telling they want to disable the rate limit completely. We'll need to tell them there's no way to disable, but they can set it very high - which also seems a bit silly.
I'm OK to merge now and we can see if this ever becomes a problem.

roncohen:
Mind squashing?

jalvz:
its easier to squash & merge with big green button than rebase, force push again and consume some jenkins resources... i don't mind who clicks the button.

PR #258, opened 2017-10-26
Install govendor for testing.

simitt:


script/jenkins/update-beats.sh
```diff
@@ -24,4 +24,5 @@ cleanup() {
 }
 trap cleanup EXIT
 
+go get -u github.com/kardianos/govendor
```
simitt: `govendor` was missing on jenkins.
Unfortunately I haven't found a proper way yet to let the tests run in a local jenkins env before creating a PR.

PR #259, opened 2017-10-27
Trance :partyparrot: -> trace (thanks @formgeist)

roncohen:


docs/fields.asciidoc
```diff
@@ -446,7 +446,7 @@ type: keyword
 Equal to message, but with placeholders replaced.
 
 [[exported-fields-apm-trace]]
-== APM Trance fields
+== APM Trace fields
```
simitt: You need to change this in https://github.com/elastic/apm-server/blob/cdf862d41993b6f6229a4bb34a2b1ff0607c07b0/processor/transaction/_meta/fields.yml#L58,
as the documentation gets generated out of it running `make update`.

roncohen: 🤦‍♂️

roncohen:
thanks @simitt. Need more ☕️

PR #281, opened 2017-11-02
Backport to 6.0: support OPTIONS method needed for preflight browser requests (#244)

roncohen:
This allows users to configure a valid origin, so that the apm-server won't accept requests from different origins. It handles preflight OPTIONS as per CORS spec.

It makes it possible to specify a list of allowed origins as well as `*` for any origin.


roncohen:
jenkins, test it

PR #282, opened 2017-11-02
update changelog

jalvz:


jalvz:
@simitt @roncohen let me know if there is something missing/wrong.

PR #283, opened 2017-11-02
update beats framework

jalvz:


jalvz:
hackday project: create a GH user with a cron job in some machine that does this every week or something

PR #286, opened 2017-11-02
`make update-beats` prints the recommended commit msg

roncohen:
pasted_image_at_2017_11_02_05_07_pm_720


roncohen:
It now says `Update beats framework to ...` instead

PR #288, opened 2017-11-03
disable urllib warnings so that it doesn't break tests with unexpected exit code 2

jalvz:
eg: https://beats-ci.elastic.co/job/elastic+apm-server+pull-request+multijob-windows/349/beat=apm-server,label=windows/testReport/junit/test_requests/SecureTest/test_https_ok/

### DIFF
--- a/tests/system/test_requests.py
+++ b/tests/system/test_requests.py
@@ -10,7 +10,6 @@ import time
 from datetime import datetime
 from collections import defaultdict
 import threading
-import urllib3
 
 try:
     from StringIO import StringIO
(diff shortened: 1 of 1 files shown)


tests/system/test_requests.py
```diff
@@ -117,6 +117,9 @@ def test_deflate_error(self):
 class SecureTest(SecureServerBaseTest):
 
     def test_https_ok(self):
+        import urllib3
+
+        urllib3.disable_warnings()
```
simitt: good finding!
Why is this a volatile error though?

jalvz: it started happening  on recent builds, not sure. maybe is caused by the beats merge...

PR #292, opened 2017-11-09
Revert "Make transaction.result a mandatory field. (#250)"

simitt:
This reverts commit 11af2e8f9bc936fbd26d5bf444cdd77c6c8ab48d.

As the mandatory field does not make sense for frontend agents, which share the same schema as backend agents, this field is going to be optional again.


processor/transaction/package_tests/TestProcessTransactionFull.approved.json
```diff
@@ -312,7 +312,6 @@
                 },
                 "id": "85925e55-b43f-4340-a8e0-df1906ecbf7a",
                 "name": "GET /api/types",
-                "result": "failure",
```
roncohen: This is the "full" payload, so optional values should be present.

PR #293, opened 2017-11-09
Split app spec in 2 files

jalvz:
Background: the upcoming `sourcemap` data model will need some app attributes, but not all.
This splits the `app` spec in 2, `app_core` that will be used by `sourcemap`, and `app` that references `app_core` and includes all the other attributes.

This doesn't change current functionality / data models.
_____

To "merge" 2 sets of attributes at the same level, this uses the `allOf` keyword at the top of the spec file, which is legal jsonSchema. 
However, we never did this before and the package tests didn't know how to handle the new semantics.
I had to rework the `TestPayloadAttributesInSchema` to use the original schema files instead of the generated strings. There is less code (eg. `flattenSchemaNames` and `TestGetSchemaProperties` got removed), but is not necessarily simpler. 
There is probably a better way, but after trying all week,  I settled on this.

On the other hand I had to remove the `TestJsonSchemaKeywordLimitation` tests, I couldn't make them work. It would be feasible if _all_ the strings were limited to 1024, but having to cross-check the keywords in the `fields.yml` is too much and I haven't been able to use that code.

I am unsure of the value of these tests, tbh. They are useful, but they basically require a jsonschema compliant parser which is probably more complex than the thing under test. 
I think we can keep the ones that they are working, but if in the future they give more problems I vote to remove them.

There are of course ways around (we could eg. duplicate the app attributes instead of embedding one spec into another) but doesn't change the fact that they are hard to maintain.

 We should also get rid of the generated string-ified schemas when we get a chance.

### DIFF
--- a/tests/json_schema.go
+++ b/tests/json_schema.go
@@ -3,9 +3,13 @@ package tests
 import (
 	"encoding/json"
 	"fmt"
-	"io"
-	"strings"
+
 	"testing"
+
+	"io/ioutil"
+
+	"path/filepath"
+	"runtime"
 
 	"github.com/fatih/set"
 	"github.com/stretchr/testify/assert"
@@ -24,18 +28,25 @@ type Mapping struct {
 	to   string
 }
 
-func TestPayloadAttributesInSchema(t *testing.T, name string, undocumentedAttrs *set.Set, schema string) {
-	var payload interface{}
+// not exhaustive list, just the words currently used in our spec
+var jsonSchemaReservedWords = []string{"$schema", "$id", "type", "description", "anyOf", "required", "title", "items",
+	"minItems", "allOf", "properties", "maxLength", "pattern", "enum", "default", "format", "additionalProperties",
+	"regexProperties", "dependencies"}
+
+func TestPayloadAttributesInSchema(t *testing.T, name string, undocumented *set.Set, specDir string) {
+
+	schema, specDir := loadSchema(specDir, "")
+	schemaNames := set.New()
+	flattenJsonKeys(schemaNames, schema, specDir, "")
+
+	var payload map[string]interface{}
 	UnmarshalValidData(name, &payload)
 	jsonNames := set.New()
-	flattenJsonKeys(payload, "", jsonNames)
-	jsonNamesDoc := set.Difference(jsonNames, undocumentedAttrs).(*set.Set)
+	flattenJsonKeys(jsonNames, payload, "", "")
+	jsonNames = set.Difference(jsonNames, undocumented).(*set.Set)
 
-	schemaStruct, _ := schemaStruct(strings.NewReader(schema))
-	schemaNames := set.New()
-	flattenSchemaNames(schemaStruct, "", addAllPropNames, schemaNames)
+	missing := set.Difference(jsonNames, schemaNames).(*set.Set)
 
-	missing := set.Difference(jsonNamesDoc, schemaNames).(*set.Set)
 	if missing.Size() > 0 {
 		msg := fmt.Sprintf("Json Transaction Payload fields missing in Schema %v", missing)
 		assert.Fail(t, msg)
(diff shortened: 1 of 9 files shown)


roncohen:
I would just copying `name` and `version` to the sourcemap model as a start. We might add an `environment`, but other than that, we have no intention of changing those.

I think it's a reasonable solution compared to the added complexity of adding features to our own homemade JSON Schema parser.  

`app_core` in the PR means "everything related to the app that we need in the sourcemap model". In the future, we might think other things are application core values, but there's no need for them in the sourcemap model, and then we'd have to do another special case.

jalvz:
I disagree with that suggestion and reasoning:
 
* You can compose and reuse json schemas however you want, if something doesn't fit in `app_core` because is not needed for sourcemaps, put it somewhere else and compose properly. 

* I'm not adding "features" to our json schema parser, i only changed it to use the json file _instead_ of the stringified one. It is not simple, but not more complex than before neither.

* These tests are hard to understand regardless `app_core`. The problem is that complexity and the stringified schemas, not `app_core`. I don't want to do the wrong thing just so to avoid touching the complex code. 

* Code duplication is actually the worst thing. I really, really, really don't want to duplicate code / data models / specs.

simitt:
I agree with splitting the schema files into smaller parts and combining them where needed.

Removing the `keyword` tests is probably fine, as the schema doesn't change that much any more.

tests/json_schema.go
```diff
@@ -48,98 +59,48 @@ func TestPayloadAttributesInSchema(t *testing.T, name string, undocumentedAttrs
...
+	json.Unmarshal(schemaData, &schema)
+	return schema, filepath.Dir(specFile)
 }
 
-func flattenJsonKeys(data interface{}, prefix string, flattened *set.Set) {
+// walks a json collecting nested keys separated by dots, ie. {a:{b:c:{d:1}} ==> a.b.c.d
+// accumulates the result in the first argument, which is mutated recursively
+// it has some rudimentary knowledge of jsonSchema.
+// jsonSchema keys are traversed, but not accumulated in the result
+func flattenJsonKeys(flattened *set.Set, data interface{}, currentSpecDir string, keyPrefix string) {
```
roncohen: Is it correctly understood that this now is intended to be used both for payloads and for JSON Schema files?

tests/json_schema.go
```diff
@@ -48,98 +59,48 @@ func TestPayloadAttributesInSchema(t *testing.T, name string, undocumentedAttrs
...
+func loadSchema(spec string, currentDir string) (map[string]interface{}, string) {
+	if currentDir == "" {
+		_, current, _, _ := runtime.Caller(0)
+		currentDir = filepath.Join(filepath.Dir(current), "../docs/spec/")
 	}
+	specFile := filepath.Join(currentDir, spec)
+	schemaData, _ := ioutil.ReadFile(specFile)
+	var schema map[string]interface{}
+	json.Unmarshal(schemaData, &schema)
+	return schema, filepath.Dir(specFile)
```
roncohen: Isn't `filepath.Dir(specFile)` always equal to `currentDir` here?

tests/json_schema_test.go
```diff
@@ -165,103 +165,3 @@ func testDataAgainstSchema(t *testing.T, testData []schemaTestData, schemaPath s
 		assert.True(t, filesToTest.Has(f.Name()), fmt.Sprintf("Did you miss to add the file %v to `json_schema_tests`?", filepath.Join(path, f.Name())))
 	}
 }
-
-func TestGetSchemaProperties(t *testing.T) {
```
roncohen: This got removed because `flattenJsonKeys` is now doing the work that `flattenSchemaNames` used to do, but the JSON Spec part of `flattenJsonKeys` isn't tested as far as I can see.

roncohen:
Is it correctly understood that you had to change `TestPayloadAttributesInSchema` to use the original schema files instead of the generated strings because flattenSchemaNames doesn't support `allOf` ?

tests/json_schema.go
jalvz: no, not all specs are in the same directory

tests/json_schema.go
jalvz: yes

roncohen:
> I'm not adding "features" to our json schema parser, i only changed it to use the json file instead of the stringified one. It is not simple, but not more complex than before neither.

You're right. What I meant was this adds a "feature" to `flattenJsonKeys` so that it now knows about `$ref`. That functionality is in two places now: https://github.com/elastic/apm-server/blob/master/script/inline_schemas/inline_schemas.go#L43. 

The schemas we use in production are the ones that are inlined, but this changes our test to no longer use the actual schema we use in production.

tests/json_schema_test.go
jalvz: It is not, but i know it works because the tests using it pass.

jalvz:
> Is it correctly understood that you had to change TestPayloadAttributesInSchema to use the original schema files instead of the generated strings because flattenSchemaNames doesn't support allOf ?


`flattenSchemaNames` was problematic, but it is not as easy as supporting `allOf`. I've been working with those tests for the whole week, I tried many things and I got many errors.

jalvz:
Are you worried that this doesn't work, or what is the issue?

roncohen:
> Are you worried that this doesn't work, or what is the issue?

Just trying to minimize more complexity and pitfalls. 

* We have the `$ref` logic implemented in two separate places now. The tests have their own implementation which is different from what we use in production.
* `flattenJsonKeys` now does the job of the old `flattenJsonKeys` + `flattenSchemaNames`. As far as i can see it only works right now because we don't have `title`, `properties` or similar in our payload keys.

jalvz:
I don't think is fair to say this adds complexity because `flattenJsonKeys` is more complex . There is a lot of stuff removed as well, one thing pays for the other. If you have a simpler working implementation, i'm happy to see it!

And to put things in perspective: these tests check that all the documented fields from our manually hardcoded *samples* are in the spec and vice versa. Those samples are not used for anything production related, `$ref` inlining logic is not production code and is pretty much an impl. detail.

Duplicating the spec and all that comes out of it (data types, etc), which is the only alternative proposed so far, is much worse.

I can suggest another alternative if you dislike this that much: remove it all and add an entry in a code review checklist saying: "if you change the schema, make sure to reflect it in the sample files".

simitt:
I took another look at this. The changes you want to make are rather small from a functional perspective. It feels very inbalanced that adapting the specs to this change takes such a huge amount of changes. On the other hand we did run into troubles before, when not having the tests in place.
Since the only way of having those tests in a form we have them right now is to parse the schema I suggest we discuss having other tests in place.

@jalvz what do you think about using `approval` tests for it. 
E.g. for the keyword limitation we create a test case that outputs one json file recording all the keys that have the limit set in the schema, and recording all the keys that are keywords in the `meta.yml`. This would indicate a semi-automated process when something changes. The approved file would fail, and show the changes, the reviewer thinks about it and approves (or not).

jalvz:
@roncohen came up another way, which i just merged: https://github.com/elastic/apm-server/pull/298, so this is obsolete now

PR #296, opened 2017-11-10
Disable (vendored) urllib3 warnings

jalvz:
urllib3 might or might not be vendored inside requests, this disables urllib warnings in any case to keep CI happy.

### DIFF
--- a/tests/system/test_requests.py
+++ b/tests/system/test_requests.py
@@ -117,9 +117,19 @@ class Test(ServerBaseTest):
 class SecureTest(SecureServerBaseTest):
 
     def test_https_ok(self):
-        import urllib3
 
-        urllib3.disable_warnings()
+        try:
+            import urllib3
+            urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
+        except ImportError:
+            pass
+
+        try:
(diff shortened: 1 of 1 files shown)


jalvz:
isn't this installing the latest version? https://github.com/jalvz/apm-server/blob/3ad33b3129c0be3b0e4057efc53948c381a2af79/tests/system/requirements.txt#L1

beniwohli:
It installs the latest version at the time you run it the first time. If you first installed requests when the version was lower than 2.16, it won't upgrade it. You need to say that you want `requests>=2.16.0`

jalvz:
Sure, but in CI we always install it fresh, right? 
https://github.com/jalvz/apm-server/blob/c26ae295e8c00ed96d877d3c7639445584dde796/_beats/libbeat/scripts/Makefile#L206

Moreover, `pip install` gets passed a `-U` argument

beniwohli:
In that case I don't understand why you'd need to handle the request-with-vendored-urllib3 case :/

jalvz:
yeah. i had issues with installing `concurrent.futures` also in windows, so it could be that those requirements are not installed properly in windows and that `requests` comes from somewhere else. i didn't look much into it tbh... the import trick seems ok enough

PR #301, opened 2017-11-14
Adapt note for Kibana dashboards.

simitt:
Remove note about Kibana incompatibility.
Add information for loading dashboards when Kibana is not running on localhost.


simitt:
@roncohen, @jalvz I'd really like to get the docs change out, could one of you please review.

Issue #28, opened 2017-08-09
Creating docs and ES index template out of fields.yml


simitt:
The `fields.yml` file is used as a base to create documentation, kibana index-patterns and indexing templates for ES.

Not all options that are allowed by [Elasticsearch](https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-params.html) are implemented for index template creation right now. Also there are some limitations per mapping param that might not be that obvious. Eg 
- `enabled:false` can only be used for `mapping types` and `object` fields, see [official doc](https://www.elastic.co/guide/en/elasticsearch/reference/current/enabled.html).
- difference between `enabled` and `index`. 
- ..

It would be nice if we could document all known fields, with additional information if the field is indexed, searchable, etc. For this we need to implement additional logic in beats, see https://github.com/elastic/beats/pull/4853

Right now we have set `dynamic: false` only for `context`. This means that no additional attributes are indexed dynamically. In case we don't want to automatically index all other fields we either have to implement proper functionality for the index template creation or also set `dynamic: false` for other objects.

simitt:
@roncohen not sure if you saw the state here, is it ok with you to keep the docs to only include what is indexed?

roncohen:
@simitt Any idea how much work it would take to make sure all fields and child fields are written to documentation?

alvarolobato:
The index patterns are being moved to kibana, this changes probably doesn't make too much sense anymore. Will re-evaluate later if needed and create an up to date issue.

PR #302, opened 2017-11-14
Add support for sourcemap uploads

jalvz:


jalvz:
@roncohen this is ready for review!

roncohen:
Nice @jalvz!

A couple comments:
* We discussed this briefly earlier, but how would you feel about letting the sourcemap library handle validation of the uploaded sourcemap instead of doing it in JSON Schema?  We can still use JSON Schema to validate the other fields (`app_name`, `app_version`), but just let the sourcemap itself be validated as a string and then make the sourcemap validator validate by passing the sourcemap to the library. There might be other things than just the JSON structure that it needs to validate and we can't validate those with JSON Schema anyway. Just an idea.

* Did you check how the sourcemap is stored in ES? We need to make sure it's not being indexed.

* `TestProcessSourcemapNullValues.approved.json` and `TestProcessSourcemapMinimalPayload.approved.json` look the same to me. Do you know what the difference is? Maybe we can skip some of those tests?

* Error messages given by the JSON validation for sourcemaps will refer to field names that are unknown to the user (`app.name` vs `app-name`, `bundle-filepath` vs. `bundle_filepath` etc.). We have to align the map we create [here](https://github.com/elastic/apm-server/pull/302/files#diff-1dd27bd625e71803ce168baf149b4b32R30) so it has the same names as what the users puts in, so that the error messages refer to the same fields.

jalvz:
re. handling validation with parsing library: 
not sure what you have in mind, like this https://github.com/elastic/apm-server/pull/302/commits/c319f55f513304d3644744c9cd79a2c735ac2256?

re. approved files: they look the same because tests ensure that passing attributes with null values is equivalent to not passing attributes.

jalvz:
4th points above implemented here d4ab32d5b985ef853b4f427e3fe943280d84e021

beater/handlers.go
```diff
@@ -36,6 +37,7 @@ const (
 	BackendErrorsURL        = "/v1/errors"
 	FrontendErrorsURL       = "/v1/client-side/errors"
 	HealthCheckURL          = "/healthcheck"
+	SourcemapsURL           = "/sourcemaps"
```
roncohen: This is not versioned. It might be fine, but could you explain the rationale briefly?

_meta/beat.yml
```diff
@@ -7,7 +7,7 @@ apm-server:
   # Defines the host and port the server is listening on
   host: "localhost:8200"
 
-  frontend.enabled: false
+  frontend.enabled: true
```
roncohen: This would enable frontend endpoints by default. I suggest we keep it off by default for now and reevaluate once we have sourcemap processing of errors in.

processor/sourcemap/_meta/fields.yml
```diff
@@ -6,14 +6,20 @@
...
+          fields:
+            - name: name
+              type: keyword
+              description: >
+                The name of the app this sourcemap belongs to.
+            - name: version
+              type: keyword
+              description: >
+                App version.
+
```
roncohen: Tiny thing: Could we put the `sourcemap` field in here so it shows up in the docs but it will still not get indexed?

roncohen:
Good work @jalvz! Getting close 👍

processor/sourcemap/_meta/fields.yml
jalvz: I don't know. How do you do that? I can't find anywhere else where we do that.

simitt: No that's not possible, see https://github.com/elastic/apm-server/issues/28

roncohen: right 👍

roncohen:
Cool @jalvz. Could you squash?

jalvz:
done. i wouldnt like to squash more than this, but if you are not happy with it feel free to squash and merge yourself.

script/output_data/output_data.go
```diff
@@ -28,7 +28,7 @@ func generate() error {
 
 	for path, mapping := range beater.Routes {
 
-		if path == beater.HealthCheckURL {
+		if path == beater.HealthCheckURL || path == beater.SourcemapsURL {
```
roncohen: Is there still a reason for skipping sourcemaps here?

simitt: What do you think about adding sourcemaps to a `docs/generated/sourcemaps` folder, and not to the `intake-api` ? So we would have it documented but clearly distinguished from standard intake API.
I suggest to merge this and add the doc script in a seperate PR.

jalvz: no, updated

roncohen:
awesome! Great work @jalvz !

PR #309, opened 2017-11-17
6.x.backports

simitt:
I didn't use the `merge_pr` script for backporting, because it failed since the two branches differ. 

What I did instead is following:
- `git cherry-pick c617729c9b60f3dd5e2c1490d3abab10cc069c31^..036a6b2a7531bcaeaf9086cb519bb13708149a93`
(Note: using `^` includes the from commit).
- resolving conflicts on commits that updated the beats framework by using `git checkout --theirs filename`.
- run  `git diff --name-status elastic/master` to verify that all commits that should be backported are there. 
- run `make update-beats` to update to the latest `6.x` beats. 

As we haven't backported for a long time this seemed to be the cleanest solution.


jalvz:
great work! 🎉

roncohen:
Nice work!

PR #313, opened 2017-11-17
Note about custom index patterns breaking the UI

makwarth:


watson:
We have a convention of breaking lines in AsciiDoc after each punctuation mark. This makes the diff produced by future changes as minimal as possible. The actual rendering of the text will not be affected by this 😃

PR #320, opened 2017-11-20
Update beats framework to dc74607

roncohen:


_beats/testing/environments/latest.yml
```diff
@@ -3,7 +3,7 @@
 version: '2.1'
 services:
   elasticsearch:
-    image: docker.elastic.co/elasticsearch/elasticsearch:6.0.0-rc1
+    image: docker.elastic.co/elasticsearch/elasticsearch:6.0.0-rc2
```
simitt: Shouldn't this be `6.0.0` now?

_beats/libbeat/docs/version.asciidoc
```diff
@@ -1,6 +1,6 @@
 :stack-version: 6.1.0
```
simitt: I think this should be `6.0.1`?

roncohen: We have our own, that also needs to get updated: https://github.com/elastic/apm-server/blob/6.0/docs/copied-from-beats/version.asciidoc

_beats/testing/environments/latest.yml
roncohen: Good find: https://github.com/elastic/beats/pull/5648

PR #326, opened 2017-11-20
Import script now needs 5x instead of 5.x.

simitt:


simitt:
@roncohen could you have a look please. The `make import-dashboards` cmd which used from the `apm-server setup` cmd did not work anymore with the current folder structure for Kibana 5.x

PR #329, opened 2017-11-21
Change beats version and docs version to 6.x, update beats.

simitt:


vendor/github.com/elastic/beats/libbeat/common/op/cancel.go
```diff
@@ -1,28 +0,0 @@
-package op
```
jalvz: How come is this removed? 🤔 
I can see this here: https://github.com/elastic/beats/blob/6.x/libbeat/common/op/cancel.go

simitt: That's a very good question, the commit didn't change in beats for `common/op`. I'll investigate..

PR #336, opened 2017-11-21
Use Go 1.9.2 everywhere

roncohen:


simitt:
@roncohen could you figure out why the travis test fails? It also failed on my local env but not on Jenkins, but I haven't yet figure out how it is connected to this changes.

ruflin:
The reason it fails is because Golang 1.9.2 seems to have a bit of a different python setup and it requires pip to be updated. Add the following line to the docker image: https://github.com/elastic/beats/blob/master/filebeat/Dockerfile#L10 

```
RUN pip install --upgrade setuptools
```

Dockerfile
```diff
@@ -7,6 +7,8 @@ RUN set -x && \
          netcat python-pip virtualenv && \
     apt-get clean
 
+RUN pip install --upgrade setuptools
+
```
simitt: AFAIK the `RUN` cmd gets cached by docker. Whereas `COPY, ADD` invalidate the cache, so every `RUN` cmd after those will get reevaluated every time the docker image is built, the `RUN` cmds before won't.

Having this in mind I can see that two different images can have two different versions of the setuptools installed leading to the error we saw on travis. But having this fix in place here could still lead to the same issue later, as it won't get reevaluated when rebuilding the image.
I therefore suggest to move this `RUN` cmd down after the first `COPY`cmd.

PR #338, opened 2017-11-21
Set beats dependency to 6.1 for 6.1 branch

ruflin:


ruflin:
@simitt @roncohen @jalvz Please have a look at https://github.com/elastic/beats/pull/5681 It should not affect apm I think but in case it does, please run again the update after this branch is merged.

ruflin:
I added the most recent change in an additional commit to it. Please squash if you merge.

simitt:
@ruflin I don't think the ` vendor/github.com/elastic/beats/libbeat/common/op/*` files should be removed? I had the same issue yesterday when updating beats and it was solved by doing a completely new checkout of `apm-server` and removing local `beats`.

Issue #291, opened 2017-11-06
Model change: transaction.id -> transaction.transaction_id


roncohen:
@gingerwizard pointed out that it would be nice to have the field `transaction_id` always be called `transaction_id` across our output model. Today we have `transaction.id` and `trace.transaction_id`. Let's change it so it's `transaction.transaction_id` instead.

This make it easier to filter for a specific transaction for example in a dashboard and show the transaction, all the traces etc. with a single filter

ruflin:
Why not going with `transaction.id` for the `trace` and the `transaction` object? So the trace object would look more like:

```
{
    "context": {
        ...
    },
    "processor": {
        ...
    },
    "trace": {
        "id": 0,
        ...
        "type": "db.postgresql.query"
    },
    "transaction": {
        "id": "945254c5-67a5-417e-8a4e-aa29efcbfb79",
    }
}
```

Ignoring the change in implementation this would require, I think it would be what the user expects.

roncohen:
Nice idea. Let's do that.

roncohen:
Fixed in #345

ruflin:
@roncohen Looking at #345 it seems it's now `trace.transaction.id`. The idea was to have `transaction.id` to make it the same for transaction and trace documents.

roncohen:
Created new issue to track this: https://github.com/elastic/apm-server/issues/357

PR #345, opened 2017-11-23
move trace.transaction_id to trace.transaction.id

jalvz:
implements #291 

note: this will break kibana UI badly

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -5,7 +5,23 @@
 
 
 === APM Server version HEAD
-https://github.com/elastic/apm-server/compare/f9a2086ceed0b918e1a0b3d8ddc140fc21af0e74\...master[View commits]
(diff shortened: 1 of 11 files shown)


roncohen:
@sqren @vanjacosic @makwarth cc

roncohen:
Nice work!

makwarth:
@jalvz @roncohen which version is this for?

makwarth:
Leading up to GA in 6.2, I think we need to add breaking changes like this to a topic on APM weekly or something like that. It's too easy to miss mentions.

makwarth:
Also, this includes a new task for the APM Server team to update the dashboards.

roncohen:
Good point regarding dashboards. I've added an issue now: https://github.com/elastic/apm-server/issues/346. 

We can try to bring it up breaking changes in the weekly meeting. @jalvz will you put it on the agenda?

makwarth:
There's also the [meta issue](https://github.com/elastic/apm-dev/issues/10). Add a Breaking changes checklist?

PR #358, opened 2017-11-29
Add system.ip field to context

### LINKED ISSUE #349: Add agent IP address to context
It would be really useful to collect source host / IP address from agents on every request. Apart from having the info, this feature would allow `add_kubernetes_metadata` to match originating pod and enrich events with metadata from Kubernetes.

I've been playing with that idea and did this:

https://github.com/elastic/apm-server/compare/master...exekias:system-ip?expand=1

along with these settings:
```
processors:
- add_kubernetes_metadata:
    indexers:
      - ip_port:
    matchers:
      - fields:
          lookup_fields: ["context.system.ip"]
```

so far it works.

exekias:
This PR adds `context.system.ip` field. This expose IP of the app serving received traces and will enable matching it by source in the future

closes #349

### DIFF
--- a/beater/handlers.go
+++ b/beater/handlers.go
@@ -7,7 +7,7 @@ import (
 	"strings"
 
 	"github.com/elastic/apm-server/processor"
-	"github.com/elastic/beats/libbeat/common"
+	"github.com/elastic/apm-server/utility"
 	"github.com/elastic/beats/libbeat/logp"
 
 	"compress/gzip"
@@ -20,8 +20,6 @@ import (
 
 	"github.com/hashicorp/golang-lru"
 	"golang.org/x/time/rate"
-
-	"net"
 
 	"github.com/ryanuber/go-glob"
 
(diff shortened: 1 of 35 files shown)


exekias:
Did a 2nd attempt, let me know what you think, as I needed to update processors to inject HTTP request, so they can extract info from it.

First try (5405257) was simpler but broke tests as it didn't follow the convention on how processors are tested.

simitt:
@exekias all the other fields regarding [system](https://github.com/elastic/apm-server/blob/master/docs/spec/system.json) are collected directly by the agents and sent to the server. I expected the `system.ip` to also be collected by agents. Could you please elaborate a bit on why you would handle this different?

beater/handlers.go
```diff
@@ -146,42 +145,14 @@ func ipRateLimitHandler(rateLimit int, h http.Handler) http.Handler {
 	}
 
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		if deny(extractIP(r)) {
+		if deny(utility.ExtractIP(r)) {
```
simitt: nice improvement!

exekias:
closing in favor of #393

PR #366, opened 2017-11-29
add app.environment

graphaelli:


watson:
How do we make sure the agents start implementing these new API features? Should we create issues when we merge these kind of PR's on the agent repos so that we don't forget?

PR #368, opened 2017-11-30
Update beats and fix vendor issue.

simitt:
Fix vendoring artefacts from spf13 with updating beats framework.

It looks like the spf13 packages were in the vendor folder but not properly added to the `vendor/vendor.json` file. This PR fixes this while also updating to the newest version when updating the beats framework.

(1) manually remove `vendor/github.com/spf13`
(2) run `make update-beats` to get dependencies added properly.


simitt:
Now also the forth and back with adding and removing `libbeat/common/op/*` files should be fixed.
(1) manually run `govendor remove github.com/elastic/beats/libbeat/common/op`.
(2) rerun `make update-beats`.

PR #378, opened 2017-12-05
Updated change log

roncohen:
Added missing items and removed mention of #250 as it never made it into a release.


roncohen:
jenkin, retest this

PR #379, opened 2017-12-05
Reduce type conversions in the processing pipeline.

jalvz:
This is an attempt to simplify data processing by removing (some) duplicated type conversions.

Right now, we convert `reader` to `[]byte` in the handler, and we pass that byte array to the `Validate` and `Transform` interfaces. 
In `Validate`, `[]byte` is converted back to `reader`, and that `reader` to a `map[string]interface{}` (this happens inside the validation library).
During `Transform`ation we also convert `[]byte` to a `struct`.
The sourcemaps processor is more convoluted, it has (I think) 6 (un)marshal steps.

This PR converts `reader` directly to a `map[string]interface{}`, which is passed as-is to the `Validate` and `Transform` interfaces. 
In `Validate` no more conversions happen.
In `Transform` `map[string]interface{}` is converted to a `struct`. 
This also removes 2 type conversions in the sourcemap processors.

This is possible now thanks to a recent addition in the json schema validation library (https://github.com/santhosh-tekuri/jsonschema/commit/b7a5b2f2d7c3acfbee30bd8c3779f4f163832b21).

Ideas more than welcome as usual (particularly for further simplification of the sourcemap processor  ).

### DIFF
--- a/processor/sourcemap/package_tests/processor_test.go
+++ b/processor/sourcemap/package_tests/processor_test.go
@@ -12,15 +12,15 @@ import (
 // ensure all valid documents pass through the whole validation and transformation process
 func TestSourcemapProcessorOK(t *testing.T) {
 	requestInfo := []tests.RequestInfo{
-		{Name: "TestProcessSourcemapFull", Path: "tests/data/valid/sourcemap/payload.json"},
-		{Name: "TestProcessSourcemapMinimalPayload", Path: "tests/data/valid/sourcemap/minimal_payload.json"},
+		{Name: "TestProcessSourcemapFull", Path: "data/valid/sourcemap/payload.json"},
+		{Name: "TestProcessSourcemapMinimalPayload", Path: "data/valid/sourcemap/minimal_payload.json"},
 	}
 	tests.TestProcessRequests(t, sourcemap.NewProcessor(), requestInfo, map[string]string{"@timestamp": "***IGNORED***"})
 }
 
 // ensure invalid documents fail the json schema validation already
 func TestSourcemapProcessorValidationFailed(t *testing.T) {
-	data, err := tests.LoadInvalidDataAsInterface("sourcemap")
+	data, err := tests.LoadInvalidData("sourcemap")
 	assert.Nil(t, err)
 	p := sourcemap.NewProcessor()
 	err = p.Validate(data)
--- a/processor/sourcemap/package_tests/processor_test.go
+++ b/processor/sourcemap/package_tests/processor_test.go
@@ -19,8 +19,8 @@ func TestSourcemapProcessorOK(t *testing.T) {
 }
 
 // ensure invalid documents fail the json schema validation already
-func TestTransactionProcessorValidationFailed(t *testing.T) {
-	data, err := tests.LoadInvalidData("sourcemap")
+func TestSourcemapProcessorValidationFailed(t *testing.T) {
+	data, err := tests.LoadInvalidDataAsInterface("sourcemap")
 	assert.Nil(t, err)
 	p := sourcemap.NewProcessor()
 	err = p.Validate(data)
--- a/processor/sourcemap/processor.go
+++ b/processor/sourcemap/processor.go
@@ -7,6 +7,8 @@ import (
 
 	parser "github.com/go-sourcemap/sourcemap"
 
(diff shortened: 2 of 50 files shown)


processor/validator.go
```diff
@@ -20,10 +19,9 @@ func CreateSchema(schemaData string, url string) *jsonschema.Schema {
 	return schema
 }
 
-func Validate(buf []byte, schema *jsonschema.Schema) error {
-	reader := bytes.NewReader(buf)
-	if err := schema.Validate(reader); err != nil {
-		return fmt.Errorf("Problem validating JSON document against schema: %v", err)
+func Validate(raw interface{}, schema *jsonschema.Schema) error {
+	if err := schema.ValidateInterface(raw); err != nil {
+		return fmt.Errorf("Problem validating JSON document against schema: %raw", err)
```
graphaelli: `%v` instead of `%raw`?

processor/sourcemap/processor.go
```diff
@@ -33,36 +35,31 @@ func NewProcessor() pr.Processor {
 	return &processor{schema}
 }
 
-func (p *processor) Validate(buf []byte) error {
+func (p *processor) Validate(raw interface{}) error {
 	validationCount.Inc()
 
-	var pa payload
-	err := json.Unmarshal(buf, &pa)
+	bytes, err := json.Marshal(raw.(map[string]interface{})["sourcemap"])
```
graphaelli: this could panic with something like `sourcemap.NewProcessor().Validate(map[string]string{})`

processor/sourcemap/package_tests/processor_test.go
```diff
@@ -20,7 +20,7 @@ func TestSourcemapProcessorOK(t *testing.T) {
 
 // ensure invalid documents fail the json schema validation already
 func TestTransactionProcessorValidationFailed(t *testing.T) {
```
graphaelli: not introduced here, but should these tests be named `TestSourcemapProcessorValidation`... ?

processor/sourcemap/processor.go
jalvz: Right. Is this 16d630f695a6b1aeb35f8acfbef3f622b444cfcc better?

processor/sourcemap/processor.go
```diff
@@ -33,36 +35,31 @@ func NewProcessor() pr.Processor {
 	return &processor{schema}
 }
 
-func (p *processor) Validate(buf []byte) error {
+func (p *processor) Validate(raw map[string]interface{}) error {
 	validationCount.Inc()
 
-	var pa payload
-	err := json.Unmarshal(buf, &pa)
+	bytes, err := json.Marshal(raw["sourcemap"])
```
roncohen: in `DecodeSourcemapFormData` we decode the sourcemap to a map and here we decode it back. Would it be possible to just read the file to a `[]byte`, put the `[]byte` in the returned payload use it here in parser.Parse here?
Just an idea.

roncohen:
Didn't know about `mapstructure` :+1:

processor/sourcemap/processor.go
jalvz: Not easily. It would make testing quite more painful.

processor/sourcemap/processor.go
graphaelli: looks great

jalvz: Cool, thanks!

Fixed the other issues you pointed out as well. Did you find anything else?

tests/json_schema_test.go
```diff
@@ -179,14 +145,14 @@ func testDataAgainstSchema(t *testing.T, testData []schemaTestData, schemaPath s
...
 
-func testDataAgainstProcessor(t *testing.T, p processor.Processor, testData []schemaTestData, filePath string) {
-	for _, d := range testData {
-		data, err := ioutil.ReadFile(filepath.Join("data/invalid", filePath, d.File))
-		assert.Nil(t, err)
-		err = p.Validate(data)
-		assert.Contains(t, err.Error(), d.Error)
-	}
-}
+//func testDataAgainstProcessor(t *testing.T, p processor.Processor, testData []SchemaTestData) {
```
graphaelli: looks like this moved, any reason to keep this?

PR #381, opened 2017-12-06
Sourcemapping

simitt:
This implements the sourcemap mapping when uploading `transactions` or `requests` to the frontend endpoint. 

There are some points I would like to discuss:
- If `frontend` is enabled, but there are errors with retrieving `ElasticsearchClients` or instantiating a cache, should the whole `apm-server` startup panic or should we only log an error? Right now I [logged an error](https://github.com/simitt/apm-server/blob/6235741c802a1f5f216717e9d2ff1ba9d6c55e99/beater/beater.go#L45).
- If the connection to ES is lost, or some json/byte[] error or similar occurs, should this be logged to the `stacktrace.sourcemap.Error` field? Right now I only store mapping errors to the `stacktrace.sourcemap.Error`, which are: no (unique) sourcemap found for this specified key, no mapping found for lineno+colno.
- please take a look at [utility tests](https://github.com/simitt/apm-server/blob/41508038f2da994f096123818d6bdaa8cff1b3e2/utility/sourcemap_accessor_test.go). I added only non-ES-dependent tests, everything involving ES was added as system tests. It would be feasible to add an `interface` defining the ES client methods and then creating a fake implementation so we could have isolated unit tests. However this would lead to the impression that anything implementing the proper interface could be past in while we really want to have an ES client.

CHANGELOG is missing so far, I will add it once we decided on all the details.


simitt:
@jalvz could you please have a look on a high level perspective. 
I still miss some system integration tests for the actual sourcemap fetching from ES.

simitt:
jenkins, retest this please

utility/sourcemap_accessor.go
```diff
@@ -0,0 +1,180 @@
...
+}
+
+func NewSourcemapAccessor(esClients []elasticsearch.Client, cacheSize int) (*SourcemapAccessor, error) {
+	if esClients == nil || len(esClients) == 0 {
+		err := SmapError{Msg: "Sourcemap ESClient cannot be initialized.", Kind: InitError}
+		logp.Err(err.Error())
+		return nil, err
+	}
+
+	smapCache, err := lru.New(cacheSize)
```
roncohen: is there a way to get the number of used or available slots in the lru cache? i think this would be a useful number to have for debugging potential memory issues. We could use the beats metrics to keep track of it.

simitt:
I added a commit to invalidate the cache for the sourcemap key if a sourcemap is uploaded. Generally this key should most probably not be in the cache, except when the same sourcemap key gets uploaded multiple times. 
I gave it a second thought, and this could indeed happen e.g. when adding the sourcemap upload to a deploy script and the deploy fails after the sourcemap was uploaded, it might be re-run. 

I am working on another commit now to fetch the latest sourcemap for the given key.

roncohen:
>  fetch the latest sourcemap for the given key
sounds good 👍 

I'm still digging into this. Should have something later today

simitt:
I updated the handling when multiple sourcemaps with the same key are uploaded and adapted the PRs description.

simitt:
Since this branch had conflicts with master (after merging a beats update) I rebased although the PR was open.

roncohen:
@simitt a couple of preliminary ideas from the first pass: 

* What do you think about saving the `SmapAccessor` on the `Server` and passing the `Server` to processors? This way we get type safety, and it's not unlikely that we'll need other stuff from `Server` in the processors in the future? I also considered adding an initialize step to `ProcessorFactory` so we could move sourcemap related initialization there and wouldn't have to pass `settings map[string]interface{}` into all processors, but i like passing `Server` better.
* At the moment, it's possible to specify other outputs than Elasticsearch, for example Kafka. Sourcemapping should only be available if an Elasticsearch output is selected. Lots of big organizations use Kafka that feeds into Logstash or Elasticsearch. Down the line, we can consider the option for setting a separate Elasticsearch instance to use for sourcemapping. That way, you can still use your Kafka output with source mapping. 
* We need a temporal expiration of source maps from the cache. It's great that we'll automatically expire the cache if a new one is uploaded, but you could be running multiple APM Servers behind a load balancer. Only one of them will get the source map removed from their local cache on upload
* > If the connection to ES is lost, or some json/byte[] error or similar occurs, should this be logged to the stacktrace.sourcemap.Error field? 

  Great question. We have the queue in beats that allows for some Elasticsearch downtime. If Elasticsearch is just down temporarily, we would ideally somehow pause sourcemapping until Elasticsearch is back so we keep the nice property that we can work with a bit of Elasticsearch downtime.

simitt:
@roncohen thank you for feedback, you raised some very interesting questions.

- ad passing `server` to `processor`. 
   I don't really see the point in passing the complete server to processor. However, I like the idea of passing a defined struct rather than `map[string]interface`.

- ad Kafka/ES: I am not sure I fully understand what you are saying here. Do you mean we should consider adding a completely seperate config option for ES hosts for sourcemapping so we also allow it if otherwise Kafka is used? If so I suggest to add this as an ehancement *after* we added smapping, as this PR is already pretty big. 

- ad temporal expiration of sourcemaps:
  This is indeed an issue! But also expiring the cache after a certain time would not solve this, as it still could happen that the wrong sourcemap is used. We would need something like a distributed caching system. 

- ad lost connection of ES:
  The way I implemented it the connection reconnects again after a temporary ES connection loss. However, all the requests that were processed in the mean time lack of the sourcemapping. 
The difference to the queue buffer we use for event publishing is that for sourcemapping loosing the connection would have an impact right in the middle of transforming requests to events.

beater/beater.go
```diff
@@ -38,7 +40,22 @@ func (bt *beater) Run(b *beat.Beat) error {
...
-	bt.server = newServer(bt.config, pub.Send)
+	settings := map[string]interface{}{}
+	if bt.config.Frontend.isEnabled() {
+		cfg := b.Config.Output.Config()
+		esClients, err := elasticsearch.NewElasticsearchClients(cfg)
+		if err != nil {
+			logp.Err(err.Error())
+		}
+		cacheSize := bt.config.Frontend.SourcemapCacheSize
+		smapAccessor, err := utility.NewSourcemapAccessor(esClients, cacheSize)
```
graphaelli: If this fails we have bigger problems.

utility/sourcemap_accessor.go
```diff
@@ -0,0 +1,192 @@
...
+		"sort": []map[string]interface{}{
+			{
+				"@timestamp": map[string]interface{}{
+					"order": "desc",
+				},
+			},
+		},
+	}
+
+	index := "apm-*"
```
graphaelli: needs to be configurable?

utility/sourcemap_accessor.go
```diff
@@ -0,0 +1,192 @@
...
+			logp.Err(fmt.Sprintf("Sourcemap fetching from ES Error %s", err.Error()))
+			return nil, err
+		}
+		s.AddToCache(smapId, smapConsumer)
+	}
+	return smapConsumer, err
+}
+
+func (s *SourcemapAccessor) AddToCache(smapId SmapID, smap *sourcemap.Consumer) {
+	s.smapCache.Add(smapId.key(), smap)
```
graphaelli: To Ron's [comment](https://github.com/elastic/apm-server/pull/381/files#r157163591) - we could track evictions using the return value of `Add` here.

beater/beater.go
```diff
@@ -38,7 +40,22 @@ func (bt *beater) Run(b *beat.Beat) error {
 
 	go notifyListening(bt.config, pub.Send)
 
-	bt.server = newServer(bt.config, pub.Send)
+	settings := map[string]interface{}{}
+	if bt.config.Frontend.isEnabled() {
+		cfg := b.Config.Output.Config()
+		esClients, err := elasticsearch.NewElasticsearchClients(cfg)
+		if err != nil {
+			logp.Err(err.Error())
```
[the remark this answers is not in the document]
simitt: I was thinking about it, as long as you are using ES for backend + Sourcemapping this shouldn't fail if you have configured it right. So yes for those cases we could panic.
The reason I only logged a message here is that the beats framework also allows for a configuration other than ES, eg. Kafka. If you are not using ES, then with this handling Sourcemapping simply doesn't work, but the server itself should still work. (Although I didn't test to configure something else than ES.)

beater/beater.go
simitt: The question here is do we want to let APM fail when there is an issue with Sourcemapping? In my opinion Sourcemapping is an add-on, that is why I decided to log errors here and not panic.

utility/sourcemap_accessor.go
simitt: true, I will change to pass in the index with the config.

utility/sourcemap_accessor.go
```diff
@@ -0,0 +1,192 @@
...
+		logp.Err(err.Error())
+		return nil, err
+	}
+	return &SourcemapAccessor{esClients: esClients, smapCache: smapCache}, nil
+}
+
+func (s *SourcemapAccessor) Fetch(smapId SmapID) (*sourcemap.Consumer, error) {
+	smapConsumer, err := s.fetchFromCache(smapId)
+	if err != nil {
+		logp.Err(fmt.Sprintf("Sourcemap fetching from Cache Error %s", err.Error()))
```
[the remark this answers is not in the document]
simitt: So you would avoid fetching the Sourcemap from ES in case something went wrong with the caching? Not sure about that tbh.

simitt:
@graphaelli regarding your [comment](https://github.com/elastic/apm-server/pull/381#pullrequestreview-83951603) I don't really see how we could do this. Either we cache the checksums, then we would have the same issues with multiple servers. Or we fetch it from ES every time, in which case we should rather get rid of caching in first place.

simitt:
@roncohen regarding your [comment](https://github.com/elastic/apm-server/pull/381#issuecomment-352021935) to allow configuring ES for Sourcemapping when otherwise e.g. Kafka is used. 
With the current setup in beats it is only allowed to define 1 output source. This means as soon as Kafka is configured ES cannot be configured as output source for sourcemapping. When we want to allow this we need to adapt the output handling in beats afaict. I think this needs some more discussion beforehand.

utility/sourcemap_accessor.go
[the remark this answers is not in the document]
graphaelli: No, definitely my mistake

beater/beater.go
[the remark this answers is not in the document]
graphaelli: Do we plan to allow storing sourcemapping somewhere other than ES?  I understand if we allow providing sourcemaps through alternate means but if we're doing the mapping in apm server by fetching the sourcemap from ES, I think the ES client config has to be correct (but ES doesn't necessarily have to be available at startup time)

beater/beater.go
graphaelli: Agreed, here I think it's fine to log and continue.  I expect this failure, basically allocating the cache, to be very rare.

graphaelli:
I didn't see this before writing this [https://github.com/elastic/apm-server/pull/381#discussion_r157494351](comment).  I expected that the output configuration and sourcemap storage are separate.  Agreed on having a discussion on this before proceeding.

graphaelli:
@simitt re: [comment](https://github.com/elastic/apm-server/pull/381#issuecomment-352357880)  - I intended for that checksum/warning to happen at sourcemap ingest time.  I'm not sure a race between fetching the currently stored sourcemap's checksum and updating but can look into it.

beater/beater.go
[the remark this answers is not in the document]
simitt: Agree on that. 
However, right now the same pipeline is used for uploading `sourecemaps` as for uploading `transactions` or `errors`. If something else than ES is used, uploading sourcemaps would probably just fail right now. But it would also mean that no `output.hosts.elasticsearch` would be configured, which would lead to an error at this point. When panicing we would basically always fail to startup when someone configures a different output. 
When logging an error and continuing, as it is done here, we would just disable sourcemapping. I think this shouls be the way to go until we support sourcemapping with ES although another output is configured for `transactions` or `errors`.

simitt:
@graphaelli I think we still would have the same issue, as the sourcemap gets uploaded on one of n servers. 
Writing a checksum to the server would still need to somehow notify all other servers of this.

graphaelli:
@simitt agreed that the cache invalidation piece isn't resolved by the checksum on the server.  I think what we're after is a user notification that a different sourcemap is [possibly] being used for the same service name/version/path.  This shouldn't happen with good deployment practices so I don't think we need to solve it, but it's probably a good idea to notify users when it does happen.  

I'm proposing we store a checksum in the document with the sourcemap so that subsequent sourcemap ingestions can check if they're overwriting the same name/version/path sourcemap with different content.

simitt:
Closing this to bring in the latest changes and reopening afterwards.

simitt:
@roncohen and @graphaelli I added changes and rebased the `sourcemapping` branch according to our latest discussions. Please review.

simitt:
@roncohen, @graphaelli I've rebased to master again - could you please review this, as it has a lot of changes in it.

roncohen:
Good work @simitt!

At the moment, it will try to apply source maps also for errors coming from backend services. We should only apply sourcemaps to errors coming from browsers.

A couple of ideas to solve this:
- Let the agent indicate whether sourcemapping should be applied
- Only apply sourcemapping to errors coming in through the `client-side` endpoint

simitt:
good point @roncohen , I updated the PR with proper handling.

roncohen:
cool @simitt 👍 

One more high-level comment: 
We have sourcemapping concerns spread out over a relatively large area of the code base at the moment. For example `newMuxer` which was intended to just connect processors to routes, is largely made up of code relating to sourcemaps. Sourcemapping code could perhaps be refactored out of some of the files/models that are common for all processors and into the files/models relating to the processors that need it. 

It might also make sense to see if sourcemapping related code can be separated out of `model/stacktrace_frame.go`, but that's less of a problem.

I'm OK with going forward with this structure for now, but it's something we should discuss for future work.

I'll take a close look tomorrow.

simitt:
@roncohen thank you for the input, I changed the handling how the `sourcemap_accessor` is created out of the configuration https://github.com/elastic/apm-server/pull/381/commits/72358818bfa8a9c8a6ddf8c8ccaa61ea0870e136.

apm-server.reference.yml
```diff
@@ -21,17 +21,28 @@ apm-server:
...
+    # An origin is made of a protocol scheme, host and port, without the url path.
+    # Allowed origins in this setting can have * to match anything (eg.: http://*.example.com)
+    # If an item in the list is a single *, everything will be allowed
+    #allow_origins : *
+
+    #sourcemapping:
+      #cache:
+        #expiration: 300 #seconds
+        #cleanup_interval: 600 #seconds
+      #index: "apm-"
```
roncohen: It took me a while to understand that this is the prefix we use when searching for sourcemaps and that an asterisk gets added automatically. It's unrelated to where we store them when it gets uploaded. I think it would be good to have a line or two describing this some more. By the way, it looks like the default value is `apm`, not `apm-`.

beater/handlers.go
```diff
@@ -10,9 +10,9 @@ import (
 	"net/http"
 	"strings"
 
-	"github.com/hashicorp/golang-lru"
+	lru "github.com/hashicorp/golang-lru"
```
roncohen: any particular reason to add this?

utility/sourcemap_accessor.go
```diff
@@ -0,0 +1,202 @@
...
+
+func (s *SourcemapAccessor) RemoveFromCache(smapId SmapID) {
+	s.smapCache.Delete(smapId.key())
+	logp.Debug("sourcemap", "Removed smapId %v. Cache now has %v entries.", smapId.key(), s.smapCache.ItemCount())
+}
+
+func (s *SourcemapAccessor) fetchFromCache(smapId SmapID) (*sourcemap.Consumer, error) {
+	if s.smapCache == nil {
+		msg := "SourceMap Cache requested but not initialized"
+		return nil, SmapError{Msg: msg, Kind: InitError}
```
roncohen: `s.smapCache` can only be nil, if someone didn't use `NewSourcemapAccessor` to create their `SourcemapAccessor`. 

`AddToCache`  and `RemoveFromCache` already assumes `s.smapCache` is not nil.

I think we should also do that there. It will also save us some lines in `Fetch`

utility/sourcemap_accessor.go
```diff
@@ -0,0 +1,202 @@
...
+	}
+	if cached, found := s.smapCache.Get(smapId.key()); found {
+		return cached.(*sourcemap.Consumer), nil
+	}
+	return nil, nil
+}
+
+func (s *SourcemapAccessor) fetchFromES(smapId SmapID) (*sourcemap.Consumer, error) {
+	if s.esClients == nil || len(s.esClients) == 0 {
```
roncohen: Similar to the above, this can't happen if we assume people use `NewSourcemapAccessor`, which i think we can.

apm-server.reference.yml
simitt: There is a description right above `frontend`, but I'll move it to be directly above the `index` field and try to make it more descriptive.

beater/handlers.go
simitt: I removed all imports after a merge conflict and by using `goimport` to automatically resolve imports afterwards, this was added automatically. Afaik if a package starts with `golang-` or `go-` this prefix is removed by default when using it.

_meta/beat.reference.yml
```diff
@@ -21,14 +21,25 @@ apm-server:
...
+    # An origin is made of a protocol scheme, host and port, without the url path.
+    # Allowed origins in this setting can have * to match anything (eg.: http://*.example.com)
+    # If an item in the list is a single *, everything will be allowed
+    #allow_origins : *
+
+    #sourcemapping:
+      #cache:
+        #expiration: 300 #seconds
+        #cleanup_interval: 600 #seconds
+      #Sourcemaps are stored in the same index as transaction and error documents.
```
roncohen: Could you please add a space after the `#` for explanations/comments?

_meta/beat.reference.yml
```diff
@@ -21,14 +21,25 @@ apm-server:
...
-  # Rate Limit per second and IP address
-  #frontend.rate_limit: 10
-
-  # Comma separated list of permitted origins for frontend. User-agents will send
-  # a origin header that will be validated against this list.
-  # An origin is made of a protocol scheme, host and port, without the url path.
-  # Allowed origins in this setting can have * to match anything (eg.: http://*.example.com)
-  # If an item in the list is a single *, everything will be allowed
-  #frontend.allow_origins : *
+  #frontend:
```
roncohen: What do you think about adding a short message here that tells people that frontend support is experimental? It will be in the docs, but it would be good to document here as well.

tests/system/apmserver.py
```diff
@@ -91,63 +124,179 @@ def config(self):
...
+    def check_smap(self, doc, updated, err=None):
+        if "stacktrace" not in doc:
+            return
+        for frame in doc["stacktrace"]:
+            smap = frame["sourcemap"]
+            if err is None:
+                assert 'error' not in smap
+            else:
+                assert err in smap["error"]
+            assert smap["updated"] == updated
```
roncohen: It would be good to have an assertion that the `abs_path`, `lineno` etc. is set to the values we expect them during the sourcemap process. Maybe there is a test for that, that I overlooked?

Not a blocker, in any case.

simitt: Very good point, I think I am lacking tests in the actual sourcemap mapping, will add them.

utility/sourcemap_accessor.go
```diff
@@ -0,0 +1,190 @@
...
+
+func (s *SourcemapAccessor) fetchFromCache(smapId SmapID) *sourcemap.Consumer {
+	if cached, found := s.smapCache.Get(smapId.key()); found {
+		return cached.(*sourcemap.Consumer)
+	}
+	return nil
+}
+
+func (s *SourcemapAccessor) fetchFromES(smapId SmapID) (*sourcemap.Consumer, error) {
+	body := map[string]interface{}{
```
graphaelli: not a huge deal but this query could just pull the sourcemap `"_source": "sourcemap.sourcemap"`

tests/system/apmserver.py
simitt: Actually I did test the actual mapping here https://github.com/simitt/apm-server/blob/269ee82205b3072c02a307fe3cac4643960537c8/model/stacktrace_frame_test.go#L222
by using a test sourcemap https://github.com/simitt/apm-server/blob/269ee82205b3072c02a307fe3cac4643960537c8/model/stacktrace_frame_test.go#L32

simitt:
@roncohen and @graphaelli I have rebased to master and incorporated your PR comments. Thanks for the thorough review. Are you fine with merging it now?

graphaelli:
@simitt awesome effort here!  I vote merge

roncohen:
@simitt great work here. Good to merge!

roncohen:
I think you forgot a space here: https://github.com/elastic/apm-server/pull/381/files#diff-386819a19a12310e5e0ea62392099482R42

PR #383, opened 2017-12-07
Stricter Json Schema validation: set additionalProperties to high level json entities.

jalvz:
This will allow us eg. to detect if someone is using an agent sending `traces` to a server expecting `spans` (since `spans` are optional and `traces` are ignored; right now validation succeeds if someone uses an invalid agent, and the only symptom would be missing traces in Kibana).

It also serves to detect inconsistencies in our tests etc.

This has the drawback that any addition will be a breaking change (that is why I did it only for `payloads`, `transactions`, `spans` and `errors`).

### DIFF
--- a/docs/data/intake-api/generated/transaction/transaction_empty_values.json
+++ b/docs/data/intake-api/generated/transaction/transaction_empty_values.json
@@ -14,7 +14,6 @@
             "name": "GET /api/types",
             "type": "request",
             "duration": 32.592981,
-            "start": 24.01,
             "timestamp": "2017-05-09T15:04:05.999999Z",
             "spans": []
         }
--- a/docs/spec/errors/error.json
+++ b/docs/spec/errors/error.json
@@ -96,6 +96,7 @@
             "description": "Recorded time of the error, UTC based and formatted as YYYY-MM-DDTHH:mm:ss.sssZ"
         }
     },
+    "additionalProperties": false,
     "required": ["timestamp"],
     "anyOf": [
         {
--- a/docs/spec/errors/payload.json
+++ b/docs/spec/errors/payload.json
@@ -19,5 +19,6 @@
             "$ref": "../system.json"
         }
     },
+    "additionalProperties": false,
     "required": ["service", "errors"]
 }
--- a/docs/spec/transactions/payload.json
+++ b/docs/spec/transactions/payload.json
@@ -19,5 +19,6 @@
             "minItems": 1
         }
     },
+    "additionalProperties": false,
     "required": ["service", "transactions"]
 }
--- a/docs/spec/transactions/span.json
+++ b/docs/spec/transactions/span.json
@@ -66,6 +66,7 @@
             "maxLength": 1024
         }
     },
+    "additionalProperties": false,
     "dependencies": {
         "parent": {
             "required": ["id"]
--- a/docs/spec/transactions/transaction.json
+++ b/docs/spec/transactions/transaction.json
@@ -45,5 +45,6 @@
             "maxLength": 1024
         }
     },
+    "additionalProperties": false,
     "required": ["id", "name", "duration", "type", "timestamp"]
 }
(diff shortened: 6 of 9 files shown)


watson:
Having it fail for any additional field will make it much harder to upgrade in the future, so my gut feeling is 👎. I agree this is nice for testing, but I don't like our users installing it. Could we maybe have a switch where we can enable this stricter mode for testing purposes?

jalvz:
Clarification: 

Testing is the less of the things. From 6.2 going forward we'll have a rather complex matrix of incompatibilities between agents and servers, eg. node 0.5 won't get along with server 6.2.

The consequence of that will be traces not showing up, which is a symptom quite far away from the root cause - incompatible versions. 

Support will go nuts, there might be many reasons for why traces don't show up, this just one of them. They'll need a cheatsheet to know what agents versions works with what stack, etc.

With this PR, validation errors will point the issue unambiguously. There aren't many attributes at the top levels (`payload`, `transaction`, etc.) so is not unfair to say that add something there is breaking. This PR doesn't affect `context`, `request`, `service`, and such.

==

An alternative is to have the server know which versions don't work together for every agent, validate them in addition to json schema, and return explicit error messages.

That is probably cleaner, but we would also have to maintain it and keep it up to date, etc.

Any option works for me. Would be great to have a decision soon.

jahtalab:
I think this is great for testing agents and makes catching bugs early easier and also to make sure we don't send data that apm-server doesn't know about.
Furthermore, I can't think of any use-case in which having additional properties apart from what apm-server already recognizes would make sense, if there's a need for an additional property the apm-server should add it explicitly.

I'm not sure if it makes upgrading harder @watson ,  would you care to elaborate?

watson:
@jahtalab In case we later add a field to the payload API, if the server allows for and just ignores unknown fields, the order in which you upgrade the agents and the server doesn't matter. Whereas if the server returns an error like it would with this PR, the logs of the service running the agent will be flooded with errors until the server have also been upgraded.

jahtalab:
@watson , I think generally we should settle that if apm-server makes breaking changes, apm-server should be upgraded before the agents (or should it be the other way around) and also communicate this to our users (AFAIR, there was a consensus that apm-server should be upgraded first when we were talking about agents not being part of unified build).

Based on that, this issue becomes whether removing a field from the first level of the payload (eg `trace`) should be considered a breaking change.

mikker:
I like this idea. I do see your point, @watson. But I’m sure that we can get around generating too much trouble with 1) picking an upgrade order and sticking to it and 2) help our users with deprecation warnings and the like.

mikker:
I like errors. Fail early if stuff isn’t right and things get sooo much easier to debug. At least in a anything-goes-language like Ruby.

watson:
@jalvz wrote:

> From 6.2 going forward we'll have a rather complex matrix of incompatibilities between agents and servers, eg. node 0.5 won't get along with server 6.2.
> 
> The consequence of that will be traces not showing up, which is a symptom quite far away from the root cause - incompatible versions.
> 
> Support will go nuts, there might be many reasons for why traces don't show up, this just one of them. They'll need a cheatsheet to know what agents versions works with what stack, etc.

I could see this being an issue if we where post GA, but to my knowledge we agreed to break things at this stage to get it out of the way and not introduce mechanisms for managing backwards compatibility and the like. In my mind this also means not caring too much about supporting old combinations of agents and server. This is a privilege we can take now that we're still pre-GA 😃 

I don't not think support will go nuts as they are currently not supporting this product as far as I know?

@mikker While I do agree that it's nice to fail early, I think this PR doesn't really improve on that (bear with me): There's a million ways to send up wrongly spelled properties that will not get caught by this change, so we need our integration and unit tests to validate the entire deep JSON structure no matter what. So this extra check doesn't really give any added protection as far as I can see. But please correct me if I'm wrong.

So I think the main issue I have with this is whether or not the added benefit of the check is actually helping or just gives a false sense of protection. I'm not convinced of the use-case for making a special rule for the root of the JSON structure, when we previously have seen it as a feature that we can add new fields to the other parts of the JSON without failures.

simitt:
Closing this for now. Instead we are looking into:
- having automated tests running a (strict) validations against the json schema in the agents.
- looking into the json validation library to see if we could adapt in a proper way to log warnings.

PR #385, opened 2017-12-07
rename in_app to library_frame, reversing meaning

### LINKED ISSUE #384: Rename in_app to lib(rary)_frame
Since "app" is being renamed to "service", the `in_app` property on stack trace frames doesn't make sense any more. Calling it `in_service` can be really confusing, so its been decided to revert the meaning of it and call it either `library_frame` or `lib_frame`. So when ever `in_app` was `true` previously this new property should be `false` and vice versa.

graphaelli:
closes #384

### DIFF
--- a/model/stacktrace_frame.go
+++ b/model/stacktrace_frame.go
@@ -6,17 +6,17 @@ import (
 )
 
 type StacktraceFrame struct {
-	AbsPath     *string `mapstructure:"abs_path"`
-	Filename    string
-	Lineno      int
-	Colno       *int
-	ContextLine *string `mapstructure:"context_line"`
-	Module      *string
-	Function    *string
-	InLibrary   *bool `mapstructure:"library_frame"`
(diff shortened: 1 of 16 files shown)


model/stacktrace_frame.go
```diff
@@ -13,7 +13,7 @@ type StacktraceFrame struct {
 	ContextLine *string `mapstructure:"context_line"`
 	Module      *string
 	Function    *string
-	InApp       *bool `mapstructure:"in_app"`
+	InLibrary   *bool `mapstructure:"library_frame"`
```
simitt: I'd prefer to also call it `LibraryFrame` here as it is more consistent.

graphaelli: I was trying to avoid repeating `Frame`, as in `StackFrame.LibraryFrame` but looking at it again I don't think it's a big deal and will change it.

PR #386, opened 2017-12-08
Update beats framework.

simitt:


_beats/libbeat/_meta/kibana/default/index-pattern/libbeat.json
```diff
@@ -1,16 +1,16 @@
 {
+  "version": "6.0.1",
```
simitt: seems to not be updated in beats yet, but shouldn't cause issues

simitt:
jenkins retest this please

PR #388, opened 2017-12-08
remove stacktrace transform function

graphaelli:


processor/transaction/span.go
```diff
@@ -9,16 +9,14 @@ import (
...
-	Context          common.MapStr
-	Parent           *int
-
-	TransformStacktrace m.TransformStacktrace
+	Id         *int
+	Name       string
+	Type       string
+	Start      float64
+	Duration   float64
+	Stacktrace m.Stacktrace `mapstructure:"stacktrace"`
```
roncohen: I could be wrong, but i think we only needed the mapstructure tag to be set because the fieldname was different from what our json representation was. Now that it's the same (`stacktrace` vs. `StacktraceFrames`) i assume we don't need this explicitly anymore?

simitt: Since this is not a bug but just an improvement, I suggest we fix this with one of the next PRs touching the `span.go` file.

PR #391, opened 2017-12-11
Log useful data at info level for all requests

### LINKED ISSUE #390: More useful access log
Currently, the apm-server only prints this for successful requests

    request successful, code=202

and an error message plus the status code for unsuccessful ones.

To properly debug issues, it would be useful if the apm-server logged something more akin to an access log of an http server, including IP address of sender, user agent of sender, requested path, and HTTP method.

I also think `INFO` would be a better level than `DEBUG` for these log messages.

jalvz:
fixes #390

### DIFF
--- a/beater/handlers.go
+++ b/beater/handlers.go
@@ -344,9 +344,9 @@ func sendStatus(w http.ResponseWriter, r *http.Request, code int, err error) {
 	w.Header().Set("Content-Type", content_type)
 	w.WriteHeader(code)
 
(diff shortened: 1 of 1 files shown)


beniwohli:
@jalvz are those tabs separating the values? I was thinking of something like this:

    method=GET path="/some/path" code=200 useragent="..."

of course, this would need escaping of any `"` in the user agent, but the benefit is that you could parse it with something like the [logstash kv plugin](https://www.elastic.co/guide/en/logstash/current/plugins-filters-kv.html)

jalvz:
Yes. Tabs should be easier to parse than key value pairs? 
With or without plugins (eg. awk)

PR #392, opened 2017-12-11
mark spans with only_library_frames

### LINKED ISSUE #372: Add `only_library_frames` bool to span
Add a boolean `has_in_app_frame` to every `span` to determine whether it has in_app frames. 
This can be done while [iterating over frames](https://github.com/elastic/apm-server/blob/master/model/stacktrace.go#L18). 

This field should be indexed, so it needs to be added to the [fields](https://github.com/elastic/apm-server/blob/master/processor/transaction/_meta/fields.yml).

graphaelli:
closes #372


model/stacktrace_frame.go
```diff
@@ -43,3 +43,10 @@ func (s *StacktraceFrame) Transform() common.MapStr {
 
 	return m
 }
+
+func (s *StacktraceFrame) Library() bool {
+	if s == nil || s.LibraryFrame == nil {
+		return false
```
simitt: I am not sure about the handling of nil pointers here, with this handling it signals an `in_app` library in the stacktrace. It depends if we want to count `nil` as `library` or not. @elastic/apm thoughts?

processor/error/event.go
```diff
@@ -94,7 +94,7 @@ func (e *Event) addException() {
 		e.enhancer.Add(ex, "code", e.Exception.Code.(string))
 	}
 
-	st := e.Exception.Stacktrace.Transform()
+	st, _ := e.Exception.Stacktrace.Transform()
```
simitt: The information should also be handled for `errors`.

processor/transaction/package_tests/TestProcessTransactionNullValues.approved.json
```diff
@@ -229,6 +229,7 @@
                     "us": 3781
                 },
                 "name": "SELECT FROM product_types",
+                "only_library_frames": false,
```
simitt: I would expect the `only_library_frames` information to be part of the `stacktrace` which however would be a breaking change in the ES model and would require to re-introduce something like `frames`, e.g.
```
"stacktrace": {
  "only_library_frames": false,
  "frames": [ { .. }, ..  ]
}
```

@roncohen , @jalvz what do you say?

jalvz: I agree, but I'm fine either way.

CHANGELOG.asciidoc
```diff
@@ -17,6 +17,7 @@ https://github.com/elastic/apm-server/compare/71df0d96445df35afe27f38bcf734a0828
 - Updated systemd doc url {pull}354[354]
 
 ==== Added
+- span.only_library_frames {pull}392[392]
```
jalvz: I think this has to be understandable for any apm-server users, so would be great to be a bit more descriptive

model/stacktrace_frame.go
jalvz: I understand that if `LibraryFrame` is `nil` is because we don't know; and if we don't know is better to consider it as `in_app` frame to not hide (maybe relevant) data to the user.

The `if s == nil` part I assume is good standard Go practice; but I don't think we can have `nil` frames with our current spec.

simitt: sounds reasonable

graphaelli: exactly my thinking @jalvz

processor/error/event.go
graphaelli: here we're ignoring the boolean for only_library_frames since we don't use that on error traces - should we be?

processor/transaction/package_tests/TestProcessTransactionNullValues.approved.json
graphaelli: if we're going to add more stacktrace-specific metadata later it's better to bite the bullet now and make this change, which I agree with - @simitt @jalvz ?

CHANGELOG.asciidoc
graphaelli: certainly, on it

processor/error/event.go
simitt: @roncohen, @sqren does it make sense to also provide the `only_library_frames` information for errors?

processor/transaction/package_tests/TestProcessTransactionNullValues.approved.json
graphaelli: pushed up 925753f63148cff6f8ca9f8952df731b64ac9ad7 per @simitt's spec in https://github.com/elastic/apm-server/pull/392#discussion_r156291157, fine to remove that commit depending on @roncohen's decision

PR #393, opened 2017-12-11
Enrich processing with additional data generated within the APM Server.

jalvz:
This is similar to #358. It is much easier now thanks to dd3f9b9bb48020ead3fb67fb626655936b13a6a9

We can augment the payload with any arbitrary data, skip the validation, and pass it directly to the transform methods. 

Like this is not needed to pass `request` objects to the processors.


jalvz:
@exekias If you want, you can rebase 358 and do something similar to this instead. I think your PR is good; and except for the complication of having to pass the requests around, the rest seems right.

If you can't/don't have time, I or someone else in the team can take it.

Thanks any case!

exekias:
Awesome :100:, I see this one is pretty advanced so I'm ok with you taking over! I've been traveling and have plenty of stuff on my pipe, so closing mine

jalvz:
@simitt would be great if you can have a look. thanks for the idea about injecting the augmented data, not sure if i got it right.

jalvz:
So this went out of hand.
To have the `beater` handler unaware of the decoding details for data augmentation, the decoding has to go to the `processor` package. 

This resulted in an circular dependency cycle hell so I had to move a lot of things around. There is now a `decoder` package,  and a `loader` package inside the `tests` to load all the test files.

Also, the generated docs for the intake api only have full payloads now.

processor/decoder.go
```diff
@@ -0,0 +1,36 @@
+package processor
```
simitt: What do you think about also moving this to the `decoder` package?

processor/error/event.go
```diff
@@ -90,6 +90,15 @@ func (e *Event) Transform(config *pr.Config, service m.Service) common.MapStr {
...
 }
 
+// This updates the event in place
+func (e *Event) contextTransform(pa *payload) common.MapStr {
+	if e.Context == nil {
+		e.Context = make(map[string]interface{})
+	}
+	utility.InsertInMap(e.Context, "user", pa.User)
+	return e.Context
+}
```
simitt: I suggest to create a `model.Context` which still can be a `common.MapStr`, but to be consistent with the other attributes, also has its own `transform` method, which then can add `user`, `system`, ..

processor/error/package_tests/TestProcessErrorUserAgentAndIP.approved.json
```diff
@@ -0,0 +1,31 @@
...
+                "service": {
+                    "agent": {
+                        "name": "python",
+                        "version": "1.0"
+                    },
+                    "name": "1234_service-12a3"
+                },
+                "user": {
+                    "ip": "127.0.0.1",
+                    "user-agent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_12_5)",
```
simitt: I still find it confusing to have the `user-agent` at several places now, and do not quite understand why it is placed within `user`.
I'd rather go for a neutral place, like e.g. `context.user-agent`, depending on the actual agent it would be clear if it concerns an end user or not, but I know that there are lots of discussions and different opinions about that.

processor/error/payload.go
```diff
@@ -17,6 +17,7 @@ type payload struct {
 	System  *m.System
 	Process *m.Process
 	Events  []Event `mapstructure:"errors"`
+	User    map[string]interface{}
```
simitt: Why don't you use a `common.MapStr`here as we do for most of the other things?

processor/transaction/event.go
```diff
@@ -27,43 +27,52 @@ type Dropped struct {
...
+				"total": *t.SpanCount.Dropped.Total,
 			},
 		}
 		enh.Add(tx, "span_count", s)
 	}
 	return tx
 }
 
+// This updates the event in place
+func (t *Event) contextTransform(pa *payload) common.MapStr {
```
simitt: same not as for `errors` I'd move this to `model.Context`

utility/common.go
```diff
@@ -13,3 +13,27 @@ func CleanUrlPath(p string) string {
...
 	return url.String()
 }
+
+// InsertInMap modifies `data` *in place*, inserting `values` at the given `key` or at the top level if the `key`is "".
+// If `key` doesn't exist, it gets created.
+// If `key` exists and its value is not a map, InsertInMap does nothing.
+func InsertInMap(data map[string]interface{}, key string, values map[string]interface{}) {
+	if data == nil || values == nil {
+		return
+	}
```
simitt: Why not create the map here in case it is nil?

utility/common.go
```diff
@@ -13,3 +13,27 @@ func CleanUrlPath(p string) string {
 	url.Path = path.Clean(url.Path)
 	return url.String()
 }
+
+// InsertInMap modifies `data` *in place*, inserting `values` at the given `key` or at the top level if the `key`is "".
+// If `key` doesn't exist, it gets created.
+// If `key` exists and its value is not a map, InsertInMap does nothing.
+func InsertInMap(data map[string]interface{}, key string, values map[string]interface{}) {
```
simitt: I think this should be possible already when using `common.MapStr`'s `Put` method.

utility/common.go
jalvz: Because i am not `return`ing data, if i create the map there i wouldn't be able to "see it" from the outside as it would a be (new) local reference.
I could `return` it but i'm very much against mutating _and_ returning data, is very hacky and counter intuitive. 
I could also make a full copy of the map and return that, but that is expensive as it is used at the payload level.

utility/common.go
jalvz: I tried to replicate the behaviour copying the test for this with common.MapStr and the test fails pretty much every case as it doesn't expand the map in the way I need to.

This implementation is a mental leftover from a previous attempt, so Im changing this to be much simpler.

processor/decoder.go
jalvz: good idea, done here: 85ec52dcc2637ac4b674f325373dfdb6f7f80f2f

we can leave the map[string]interface{} to common.MapStr to another iteration

decoder/decoder_test.go
```diff
@@ -94,3 +96,38 @@ func TestDecodeSizeLimitGzip(t *testing.T) {
...
+	data, err := DecodeSourcemapFormData(req)
+	assert.NoError(t, err)
+
+	assert.Len(t, data, 4)
+	assert.Equal(t, "js/bundle_no_mapping.js.map", data["bundle_filepath"])
+	assert.Equal(t, "My service", data["service_name"])
+	assert.Equal(t, "0.1", data["service_version"])
+	assert.NotNil(t, data["sourcemap"].(string))
+	assert.Equal(t, len(fileBytes), len(data["sourcemap"].(string)))
+}
```
simitt: I really like this change to havedecoding functions in one place! Could you please also add some basic tests for the other decode data?

jalvz: Done! c0649a7e0f2ed7d7407ce3cd3d683de86920da71

utility/common.go
```diff
@@ -13,3 +13,23 @@ func CleanUrlPath(p string) string {
 	url.Path = path.Clean(url.Path)
 	return url.String()
 }
+
+// InsertInMap modifies `data` *in place*, inserting `values` at the given `key` or at the top level if the `key`is "".
```
simitt: This description is outdated.

Could you also add a note that `key` has to be a top-level key

processor/error/package_tests/json_schema_test.go
```diff
@@ -27,6 +27,7 @@ func TestPayloadAttributesInSchema(t *testing.T) {
 		"errors.context.request.cookies.c1",
 		"errors.context.request.cookies.c2",
 		"errors.context.tags.organization_uuid",
+		"errors.context.system.ip",
```
simitt: This shouldn't be necessary - if so it would be a bug because it actually meant that it is not added to the `fields.yml` (where it is).

jalvz: good catch, thanks! 
fixed

simitt:
jenkins, retest this please

simitt:
Great PR @jalvz , looking forward to having this in!

roncohen:
Nice work @jalvz !

PR #394, opened 2017-12-11
added some more descriptions to the JSON schema

beniwohli:
not super detailed, but better than nothing :)

### DIFF
--- a/docs/spec/context.json
+++ b/docs/spec/context.json
@@ -18,9 +18,11 @@
             "type": ["object", "null"],
             "properties": {
                 "finished": {
+                    "description": "A boolean indicating the the response was finished",
                     "type": ["boolean", "null"]
                 },
                 "headers": {
+                    "description": "A mapping of HTTP headers of the response object",
                     "type": ["object", "null"],
                     "properties": {
                         "content-type": {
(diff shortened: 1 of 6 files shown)


docs/spec/context.json
```diff
@@ -32,6 +34,7 @@
                     "type": ["boolean", "null"]
                 },
                 "status_code": {
+                    "description": "The HTTTP status code of the response.",
```
roncohen: Hyper text test transfer protocol 😎

docs/spec/context.json
```diff
@@ -41,6 +44,7 @@
         },
         "tags": {
             "type": ["object", "null"],
+            "description": "A mapping of tags with values.",
```
roncohen: flat mapping? to further distinguish itself from custom

roncohen:
nice @beniwohli. We should also make sure these end up on fields.yml

docs/spec/context.json
beniwohli: good catch 👍

beniwohli:
@roncohen I suppose there's no automated way to get these into `fields.yml`? 😁

roncohen:
No, but we don't have to do it all at once :)

simitt:
https://github.com/elastic/apm-server/issues/397

PR #395, opened 2017-12-12
Move the healthcheck handler to the root and include about info.

### LINKED ISSUE #375: Print build date/time and commit hash on startup?
For easy debuggability, especially of snapshot builds, it would be nice if the apm-server would print out its build date/time and commit hash, similar to how it prints the version:

    2017/12/04 14:44:27.223895 beat.go:203: INFO Setup Beat: apm-server; Version: 7.0.0-alpha1

A more involved version would be an endpoint where this information could be queried, but this has certain security implications: making it easy to discover which version of the apm-server is running could make it easy for attackers to find outdated and possibly insecure versions of the apm-server in the wild.

jalvz:
About info contains the version, git ref and build date(*), and is returned only if a request is authenticated.

About is also printed on startup.
The git ref and the updated timestamp are generated with `make update`.

Closes #375

(*) @beniwohli asked for the build date. This is not exactly the build date, but the timestamp of when you run `make update`. Im not sure how I would get the build date...

### DIFF
--- a/beater/handlers.go
+++ b/beater/handlers.go
@@ -11,7 +11,6 @@ import (
 	"compress/gzip"
 	"compress/zlib"
 	"encoding/json"
-	"errors"
 	"net/http"
 
 	"crypto/subtle"
@@ -23,6 +22,7 @@ import (
 
 	"github.com/ryanuber/go-glob"
 
+	"github.com/elastic/apm-server/beater/about"
 	err "github.com/elastic/apm-server/processor/error"
 	"github.com/elastic/apm-server/processor/healthcheck"
 	"github.com/elastic/apm-server/processor/sourcemap"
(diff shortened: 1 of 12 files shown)


beater/handlers.go
```diff
@@ -107,10 +107,16 @@ func sourcemapHandler(pf ProcessorFactory, config Config, report reporter) http.
...
+func healthCheckHandler(_ ProcessorFactory, config Config, _ reporter) http.Handler {
 	return logHandler(
 		http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-			sendStatus(w, r, http.StatusOK, nil)
+			if r.URL.Path != "/" {
+				sendStatus(w, r, http.StatusNotFound, nil)
+			} else if isAuthorized(r, config.SecretToken) {
+				sendStatus(w, r, http.StatusOK, about.About())
+			} else {
+				sendStatus(w, r, http.StatusOK, nil)
```
jalvz: we could also forbid completely non authenticated health-check requests...

beniwohli:
@jalvz I looked into this a while ago, and most solutions recommend using ldflags, e.g. https://www.atatus.com/blog/golang-auto-build-versioning/

PR #405, opened 2017-12-16
Fix attribute to resolve link to agents ref

dedemorton:
The asciidoc reference was incorrect, so it wasn't resolving. As a result, there is missing text in the published doc.

Note that the doc build does not flag missing asciidoc attributes as errors.  There's an issue open in the docs repo, but there's no resolution on it yet: https://github.com/elastic/docs/issues/220

### DIFF
--- a/docs/security.asciidoc
+++ b/docs/security.asciidoc
@@ -2,7 +2,7 @@
 [float]
 === Security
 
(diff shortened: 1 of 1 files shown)


watson:
@dedemorton thanks for spotting and fixing it 😃

beniwohli:
This probably also needs to go into 6.1 and 6.x? cc @elastic/apm-server

PR #417, opened 2017-12-19
Update beats for 6.1.1

graphaelli:


_beats/libbeat/_meta/kibana/default/index-pattern/libbeat.json
```diff
@@ -12,5 +12,5 @@
       "version": 1
     }
   ],
-  "version": "6.1.0"
+  "version": "6.1.1"
```
ruflin: Not related to this PR but I think this file should not be checked in in apm-server. Need to double check.

PR #430, opened 2017-12-21
Support transaction.marks.

### LINKED ISSUE #423: Support `transaction.marks` for experimental frontend support.
Custom `transaction.marks` should be supported. They should support the format 

```json
{
  "transactions": [
    {
      ...
      "marks": {
        "navigationTiming": {
          "appBeforeBootstrap": 608.9300000000001,
          "appAfterBootstrap": 623.0600000000001,
          "timeToComplete": 5707.695000000001,
          "navigationStart": -21,
          ....
        }
      }
    }
  ]
}
```
So basically `marks` is an object consisting of `objects` containing simple `key(string):value(number)` pairs. 
The data need to be indexed to ensure they are searchable and groupable.

simitt:
Support user defined groups of marks in the form of  key:value.
Ensure marks are indexed.
closes #423

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -23,6 +23,7 @@ https://github.com/elastic/apm-server/compare/71df0d96445df35afe27f38bcf734a0828
 - Add context.request.url.full {pull}436[436]
 - report more detail on max data size error {pull}442[442]
 - Add transaction.id to errors {pull}437[437]
+- Support for `transaction.marks` {pull}430[430].
 
 ==== Deprecated
 
--- a/docs/spec/transactions/mark.json
+++ b/docs/spec/transactions/mark.json
@@ -1,0 +1,11 @@
+{
+    "$schema": "http://json-schema.org/draft-04/schema#",
+    "$id": "docs/spec/transactions/mark.json",
+    "type": ["object", "null"],
+    "description": "A mark captures the timing of a significant event during the lifetime of a transaction. Every mark is a simple key value pair and can be set by the user or the agent.",
+    "regexProperties": true,
+    "patternProperties": {
+        "^[^.*\"]*$": { "type": "number" }
+    },
+    "additionalProperties": false
+}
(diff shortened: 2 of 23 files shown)


docs/spec/transactions/mark.json
```diff
@@ -0,0 +1,11 @@
+{
+    "$schema": "http://json-schema.org/draft-04/schema#",
+    "$id": "docs/spec/transactions/mark.json",
+    "type": ["object", "null"],
+    "description": "A flat mapping of a user-defined mark.",
```
simitt: @jahtalab maybe you can come up with a better description here.

docs/spec/transactions/transaction.json
```diff
@@ -43,6 +43,18 @@
             "type": "string",
             "description": "Keyword of specific relevance in the service's domain (eg: 'request', 'backgroundjob', etc)",
             "maxLength": 1024
+        },
+        "marks": {
+            "type": ["object", "null"],
+            "description": "User defined objects for grouping marks.",
```
simitt: @jahtalab same here - maybe some more descriptive information comes to your mind?

docs/spec/transactions/mark.json
jahtalab: @simitt 
I think I should be something like: 
1. "A mark captures the timing of a significant event. They're organized into groups and can be set by the user or by the agent."
2. "A mark captures the timing of a significant event during the lifetime of a transaction. They can be set by the user or by the agent."

The second one doesn't specify the grouping since I think it is a technical detail.
Also, maybe we should use the same definition in both places.

CHANGELOG.asciidoc
```diff
@@ -19,6 +19,7 @@ https://github.com/elastic/apm-server/compare/71df0d96445df35afe27f38bcf734a0828
 ==== Added
 - Support for uploading sourcemaps {pull}302[302].
 - app.environment {pull}366[366]
+- Support for uploading `transaction.marks` {pull}430[430].
```
roncohen: I suggest: "Added support for `transaction.marks`"

simitt: This is already in the section `Added` so I don't see value in repeating this, but will remove the `uploading`.

simitt:
@roncohen I rebased on master to remove conflicts.

PR #432, opened 2017-12-22
Upload Sourcemap as binary.

### LINKED ISSUE #420: Sourcemaps: consider using `binary` field type
When displaying sourcemap documents in Kibana/Discover, Kibana more or less throws up its hands and gives up (sourcemaps can be multi-megabyte large).

AFAIK, Discover doesn't try to display [`binary`](https://www.elastic.co/guide/en/elasticsearch/reference/current/binary.html) fields, which would solve this. It is also semantically the better choice IMO, as sourcemaps should be pretty much opaque for Elasticsearch.

We should also consider using gzip. The space savings should be pretty huge*, and the time savings in transferring the data from ES to apm-server should easily make up for the time used to decompress, especially on one of the lower gzip levels.

As an example, the sourcemap from https://github.com/elastic/opbeans/ is 

 * 4.3Mbyte uncompressed
 * 1.1Mbyte at gzip level 1
 * 996kbyte at gzip level 4
 * 924kbyte at gzip level 9

--
\* maybe not, due to built-in compression of Elasticsearch.

simitt:
closes #420


beniwohli:
@simitt can you check if Kibana doesn't try to load/render the binary field in Discover? Otherwise we need to enable source filtering for the field somehow (I guess in `fields.yml`, but no idea if libbeat has handling for that yet)

processor/sourcemap/decoder.go
```diff
@@ -19,14 +20,14 @@ func DecodeSourcemapFormData(req *http.Request) (map[string]interface{}, error)
 	}
 	defer file.Close()
 
-	var parsedSourcemap map[string]interface{}
-	err = json.NewDecoder(file).Decode(&parsedSourcemap)
+	var sourcemap bytes.Buffer
+	_, err = io.Copy(&sourcemap, file)
```
roncohen: alternatively, I think you can do:
`buf, err := ioutil.ReadAll(file)` and use `buf` below.

roncohen:
I took a look and it looks like `[]byte` gets converted to a list of numbers when it goes into Elasticsearch:

image

I suspect this is an artifact of how the Elasticsearch Go client we're using handles `[]byte`.

The tests don't catch this as they don't test the part. Integration tests could be useful here to test the full thing, including the Elasticsearch Go Client.


Also `data/valid/sourcemap/payload.json` still has the sourcemap itself as a JSON structure and not as a the `[]byte` they actually end up as: https://github.com/elastic/apm-server/pull/432/files#diff-1dd27bd625e71803ce168baf149b4b32R30

How about simply storing it to a regular unindexed `string` instead of `[]byte`? I gave it a shot it here: https://github.com/roncohen/apm-server/commit/d8e805be2d289657f886459814177e43846f7b58

simitt:
@roncohen storing it as string is also fine with me. Could you please open a PR for your implementation. 
Regarding integration tests: there is the open PR https://github.com/elastic/apm-server/pull/381 implementing the sourcemap mapping for incoming requests, which adds system tests for uploading and then using sourcemaps. I'll need to rebase this once the change in the upload is added.

beniwohli:
do you use `binary` as the field type in the elasticsearch index template?

simitt:
@beniwohli the type `binary` is currently not supported in the Kibana Index Pattern generator in beats.

processor/sourcemap/payload.go
```diff
@@ -35,7 +36,7 @@ func mappings(pa *payload) (time.Time, []m.DocMapping) {
 				return common.MapStr{
 					"bundle_filepath": pa.BundleFilepath,
 					"service":         common.MapStr{"name": pa.ServiceName, "version": pa.ServiceVersion},
-					"sourcemap":       pa.Sourcemap,
+					"sourcemap":       base64.StdEncoding.EncodeToString(pa.Sourcemap),
```
simitt: @roncohen what do you think of storing a base64 encoded string instead?
This way we still have the `[]byte` for validating the sourcemap when uploading it and we simply can Decoded the base64 string when applying the sourcemapping for requests.

beniwohli: that's basically what the `binary` field type does AFAIK. Wouldn't it be better in the long term to add support for this field to beats instead of building our own thing on top of a field that isn't meant for binary data storage?

I haven't looked at how ES implements the `binary` field, but I could imagine that it stores it in a less wasteful format than base64 (33% overhead in the best case!), and only uses base64 for transmission.

graphaelli: I wasn't sure what the kibana index pattern generator was so I attempted to add binary field support in elastic/beats#5974.

simitt: nice @graphaelli I am working on the same thing atm ;) 
I am also working on adding support for source_filtering as I have figured out now how to exclude information from being shown in discover.

beniwohli: Awesome stuff @simitt & @graphaelli!!

simitt: @beniwohli the binary type accepts a base64 encoded string, that's why I changed it to this here, sorry for not being clear enough that this also includes adding binary support for creating the Kibana index pattern and ES template out of fields.yml. 
see https://www.elastic.co/guide/en/elasticsearch/reference/current/binary.html#binary

beniwohli: ah, I assumed that the go elasticsearch library would handle that transparently

simitt: Adding source filteringin beats:  https://github.com/elastic/beats/pull/5975

simitt:
To finalize and sum this up, I uploaded a sourcemap via this branch and making use of https://github.com/elastic/beats/pull/5975.:

- `Sourcemap.sourcemap` is sent as base64 encoded string. 
- The ES index template has defined `binary` as datatype for storing `sourcemap.sourcemap`, and `doc_values` set to `false`.
- Kibana Index Pattern defines `sourceFilters` section that excludes the defined pattern from showing up in the discovery section.
- As soon as elastic/beats#5974 is merged, the explicit definition of setting `aggregation, index, etc` to `false` in the fields.yml is not necessary any more. 

@roncohen , @graphaelli , @beniwohli I think with merging this PR and the two PRs in beats we should be good for uploading sourcemaps in binary format.

simitt:
Closed in favor of #446 .

PR #435, opened 2017-12-22
consider exception or log message for grouping key

### LINKED ISSUE #402: Grouping key: Use `log.message` or `exception.message` if there's nothing else
There are edge cases in which agents will send up a `log.message` or `exception.message` but no stack trace or `param_message`. In these cases, APM Server should take `message` into account when calculating the grouping key.

graphaelli:
if nothing else is available

closes #402

### DIFF
--- a/processor/error/event.go
+++ b/processor/error/event.go
@@ -4,6 +4,7 @@ import (
 	"crypto/md5"
 	"encoding/hex"
 	"fmt"
+	"hash"
 	"io"
 	"strconv"
 	"time"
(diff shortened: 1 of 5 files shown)


processor/error/event.go
```diff
@@ -123,40 +124,68 @@ func (e *Event) addGroupingKey() {
...
-		}
-		return s != nil
+func newGroupingKey() *groupingKey {
+	return &groupingKey{
+		hash:  md5.New(),
+		empty: true,
+	}
+}
+
+func (k *groupingKey) add(s *string) bool {
```
graphaelli: these methods will panic if k is `nil` but since it's private and used in exactly one place I skipped guards.

PR #437, opened 2017-12-22
add transaction.id to errors

### LINKED ISSUE #409: Connecting errors with transactions
It would be valuable if errors had a reference to the transaction that was active when the error occurred.

I suggest that we simply add an optional `transactionId` to the list of [error properties](https://github.com/elastic/apm-server/blob/master/docs/spec/errors/error.json) in the intake API. In the Elasticsearch model I imagine this should be stored as `transaction.id`.

If possible it would be nice to get this out in the data model before 6.2. We can always improve the usage of this in the UI later, but having it in the model before 6.2 would be beneficial.

Crazy bonus idea: This might even allow us to remove duplicate information from errors that are already stored on the transaction. But that would potentially have serious downsides once we introduce sampling, as the active transaction might not have been sampled.

cc @elastic/apm

graphaelli:
to connect errors to transactions.  closes #409


roncohen:
@graphaelli nice work! 
But we should put it in the same place as in with transaction documents. That is, on the root. In other words, we need to move it one level up. That way, the field is the same across all our documents.

graphaelli:
@roncohen that makes sense now, pushed up a commit for that.  One side-effect is that `transaction.id` is now a common field - the remaining transaction properties (`name`, `duration`, etc) stay valid only on transaction docs.

ruflin:
As the `transaction.id` is the field for correlation, it's fine that `name` etc. are only available in the transaction event.

@graphaelli This will need a rebase.

graphaelli:
@ruflin I was going to see if merging #443 eliminates the conflict here, do you have happen to know?

ruflin:
I think you will have to rebase anyways even with #443 as the union is not in this PR yet and Github will not retrigger (I think).

graphaelli:
taking your word for it, rebased & force pushed

tests/data/valid/error/payload.json
```diff
@@ -245,5 +245,8 @@
                 "message": "Cannot read property 'baz' of undefined"
             }
         }
-    ]
+    ],
+    "transaction": {
```
roncohen: I believe this should be inside the elements in `errors`. Agents can buffer up errors from different transactions

graphaelli: ugh, my mistake, that makes much more sense.

tests/json_schema.go
```diff
@@ -60,6 +60,7 @@ func TestJsonSchemaKeywordLimitation(t *testing.T, fieldPaths []string, schema s
 	mapping := []Mapping{
 		{"errors.context", "context"},
 		{"transactions.context", "context"},
+		{"errors.transaction.id", "transaction.id"},
```
graphaelli: @roncohen wasn't that comfortable with this, any way to do this better?

roncohen: Making sure i got this correctly: The test checks that everything that get stored as a `keyword` has the length restriction and vice-versa. We use this mapping to map the json schema names to the final elasticssearch document names. If that's correct, i don't have a better way to do it that what you did here.

graphaelli: That's how I understand that test as well, thanks for verifying.

PR #439, opened 2017-12-23
Increase default 'MaxUnzippedSize' from 10mb to 50mb

roncohen:


graphaelli:
does this need a note in the changelog?

simitt:
I don't think we need to add a CHANGELOG entry for this in the current stadium.

roncohen:
Normally, I would say this warrants a changelog entry as we're changing the defaults for a configuration option. I'll add it.

PR #441, opened 2017-12-24
Add required field `transaction.sampled`

roncohen:
This adds a required field `transaction.sampled` to the intake API and the Elasticsearch output document for transactions.

This doesn't change the requirements around what data is included when sampled is `false`/`true`. That is, agents can still send `spans` and set `sampled: false`. We might want to add this enforcement later, but I wanted to get this in now as it's a breaking change.

**Why required?**
I considered defaulting it to `true` when `spans` is missing in the payload in order to maintain backwards compatibility with current agents, but decided against it because now is the time to make breaking changes and because i believe having it default to `true` when `spans` is missing in the payload will add complexity in the long run. Happy to discuss though.

### DIFF
--- a/docs/data/elasticsearch/transaction.json
+++ b/docs/data/elasticsearch/transaction.json
@@ -115,6 +115,7 @@
         },
         "name": "GET /api/types",
         "result": "success",
+        "sampled": true,
         "type": "request"
     }
 }
--- a/docs/data/intake-api/generated/transaction/minimal_payload.json
+++ b/docs/data/intake-api/generated/transaction/minimal_payload.json
@@ -2,8 +2,8 @@
     "service": {
         "name": "service1",
         "agent": {
-          "name": "python",
-          "version": "1.0"
+            "name": "python",
+            "version": "1.0"
         }
     },
     "transactions": [
@@ -12,8 +12,7 @@
             "name": "GET /api/types",
             "type": "request",
             "duration": 32.592981,
-            "timestamp": "2017-05-09T15:04:05.999999Z",
-            "sampled": false
+            "timestamp": "2017-05-09T15:04:05.999999Z"
         }
     ]
 }
--- a/docs/data/intake-api/generated/transaction/minimal_payload.json
+++ b/docs/data/intake-api/generated/transaction/minimal_payload.json
@@ -12,7 +12,8 @@
             "name": "GET /api/types",
             "type": "request",
             "duration": 32.592981,
-            "timestamp": "2017-05-09T15:04:05.999999Z"
+            "timestamp": "2017-05-09T15:04:05.999999Z",
+            "sampled": false
         }
     ]
 }
--- a/docs/data/intake-api/generated/transaction/minimal_service.json
+++ b/docs/data/intake-api/generated/transaction/minimal_service.json
@@ -15,7 +15,8 @@
             "name": "GET /api/types",
             "type": "request",
             "duration": 32.592981,
-            "timestamp": "2017-05-30T18:53:27.154Z"
+            "timestamp": "2017-05-30T18:53:27.154Z",
+            "sampled": false
         }
     ]
 }
--- a/docs/data/intake-api/generated/transaction/minimal_service.json
+++ b/docs/data/intake-api/generated/transaction/minimal_service.json
@@ -2,11 +2,11 @@
     "service": {
         "name": "service1",
         "agent": {
-          "name": "python",
-          "version": "1.0"
+            "name": "python",
+            "version": "1.0"
         },
         "language": {
-          "name": "javascript"
+            "name": "javascript"
         }
     },
     "transactions": [
@@ -15,8 +15,7 @@
             "name": "GET /api/types",
             "type": "request",
             "duration": 32.592981,
-            "timestamp": "2017-05-30T18:53:27.154Z",
-            "sampled": false
+            "timestamp": "2017-05-30T18:53:27.154Z"
         }
     ]
 }
--- a/docs/data/intake-api/generated/transaction/minimal_span.json
+++ b/docs/data/intake-api/generated/transaction/minimal_span.json
@@ -2,8 +2,8 @@
     "service": {
         "name": "service1",
         "agent": {
-          "name": "python",
-          "version": "1.0"
+            "name": "python",
+            "version": "1.0"
         }
     },
     "transactions": [
@@ -13,7 +13,6 @@
             "type": "request",
             "duration": 32.592981,
             "timestamp": "2017-05-30T18:53:27.154Z",
-            "sampled": true,
             "spans": [
                 {
                     "name": "GET /api/types",
--- a/docs/data/intake-api/generated/transaction/minimal_span.json
+++ b/docs/data/intake-api/generated/transaction/minimal_span.json
@@ -13,6 +13,7 @@
             "type": "request",
             "duration": 32.592981,
             "timestamp": "2017-05-30T18:53:27.154Z",
+            "sampled": true,
             "spans": [
                 {
                     "name": "GET /api/types",
--- a/docs/data/intake-api/generated/transaction/null_values.json
+++ b/docs/data/intake-api/generated/transaction/null_values.json
@@ -32,7 +32,7 @@
             "result": "200",
             "context": null,
             "spans": null,
-            "sampled": false
+            "sampled": null
         },
         {
             "id": "85925e55-b43f-4340-a8e0-df1906ecbf78",
@@ -40,7 +40,7 @@
             "type": "request",
             "duration": 13.980558,
             "timestamp": "2017-05-30T18:53:42Z",
-            "sampled": false,
+            "sampled": null,
             "context": {
                 "request": null,
                 "response": null,
@@ -55,7 +55,7 @@
             "type": "request",
             "duration": 13.980558,
             "timestamp": "2017-05-30T18:53:42.281999Z",
-            "sampled": true,
+            "sampled": null,
             "context": {
                 "request": {
                     "socket": null,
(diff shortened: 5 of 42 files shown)


roncohen:
feedback appreciated @elastic/apm-agent-devs @elastic/apm-ui

watson:
@roncohen you mention considering defaulting it to `true` - you meant `false` right?

But your consideration made me think that we have other fields with default values where we are currently allowed to omit the property if we just want the default value. This reasoning should apply to all such fields right?

But in general, if we want the property to be present with its default value in the database, I think it would be more user-friendly if it was the job of the APM Server to populate it in case it wasn't present in the intake JSON. Wouldn't that be better?

mikker:
I like it 👍

roncohen:
>  we are currently allowed to omit the property if we just want the default value.

which fields a those? 🤔

graphaelli:
`library_frame` for one will default to false if omitted

beniwohli:
FYI, my sampling PR (elastic/apm-agent-python/pull/116) already contains this field. I hope to merge it soon.

As for default value vs. required, I more or less agree with @watson. Fields that have an obvious default value shouldn't be required by the API. The more required fields we have, the harder it is to interact with the APM server through things like curl, which makes it harder to play around with, which is important for authors of new agents.

simitt:
What exactly does the `transaction.sampled` field indicate? I am not sure I fully understand the value of this. 
In case it only indicates whether or not the `transaction` includes `spans` then I would not add this to the public API but rather just set the flag accordingly in the transforming process.

beniwohli:
@simitt the absence of `spans` doesn't necessarily indicate that the transaction wasn't sampled. It could just mean that the transaction didn't have any spans. Having a boolean makes it explicit, which is good IMO.

We could say that `spans: null` means "not sampled", while `spans: []` means "sampled, but didn't have spans", but that seems like somewhat magic to me, and is easy to get wrong in the agent.

simitt:
Thanks @beniwohli for explanation. 
In this case I suggest to not set a default value (in case we make the field optional) as this could only be a rough estimation. 
However, I have no strong feelings about making it required or not.

docs/fields.asciidoc
```diff
@@ -638,6 +638,14 @@ type: keyword
 The result of the transaction. HTTP status code for HTTP-related transactions.
 
 
+[float]
+=== `transaction.sampled`
+
+type: boolean
+
+Unsampled transactions have no spans.
```
mikker: Am I the only one who's brain explodes making sense of double negatives? _Sampled transactions have spans_.

roncohen:
>  Fields that have an obvious default value shouldn't be required by the API.

To be clear, you're arguing that `sampled` should default to `true` no matter what the rest of the transaction contains?

beniwohli:
@roncohen yes. My concern with adding more required fields is that the API becomes less explorable with tools like curl.

Specifically in this case, I think it's ok to assume that a transaction has been sampled unless the agent specifically tells us otherwise. If an agent doesn't support sampling, by definition all transactions are sampled. If the agent supports sampling, it'll know what value to set `sampled` to.

jahtalab:
I also agree with @beniwohli about not making it required.
Regarding the naming, people (including myself) seem to be confused about `sampled`. Personally I think we should reverse the meaning and use "partial" or "optimized" or "reduced" for the name (something like that but I'm not sure). This is also consistent with assuming if the field is not there then the agent doesn't support sampling.

mikker:
The logic also confused me. Agree with @jahtalab .

beniwohli:
I like `sampled`. It says what it is on the tin. `partial` or `optimized` don't describe _why_ data is missing, `sampled: False` does IMO.

roncohen:
Made 'sampled' optional and it now defaults to true. Also rebased to fix conflicts.

sqren:
@roncohen When are you planning to merge this?
I'm working on https://github.com/elastic/x-pack-kibana/issues/3739 but can't verify my ES queries  until the documents have the `sampled` field.

roncohen:
Thanks for your input @mikker, @jahtalab and @beniwohli!
I'd like to hear from @watson @Qard @graphaelli @simitt (and whoever has an opinion) on "sampled" vs. "partial", "optimized", "reduced".

watson:
I think `sampled` is still the best option

simitt:
To me the confusing part is that `sampled` indicates whether the `context` and the `spans` have been sampled, rather than the `transaction` itself, as I'd expect a non-sampled `transaction` not to show up at all. 

But overall I'd also go with `sampled`.

watson:
I think the details of the missing `context` and `spans` are purely an implementation detail. For Node.js at least, there's a lot more that we're not doing when a transaction is not sampled. As I see it, the main reason why we're still sending up transaction objects for non-sampled transactions, is because we can't afford to change the API in a breaking way right now (this might change in the future). So I still think it's important that we convey the reason behind the missing data - i.e. that the transaction isn't sampled.

docs/data/elasticsearch/transaction.json
```diff
@@ -115,6 +115,7 @@
         },
         "name": "GET /api/types",
         "result": "success",
+        "sampled": true,
```
simitt: please remove from the `minimal` example as it is not a required field.

docs/data/intake-api/generated/transaction/minimal_payload.json
```diff
@@ -12,7 +12,8 @@
             "name": "GET /api/types",
             "type": "request",
             "duration": 32.592981,
-            "timestamp": "2017-05-09T15:04:05.999999Z"
+            "timestamp": "2017-05-09T15:04:05.999999Z",
+            "sampled": false
```
simitt: please remove from the minimal example as it is not a required field.

docs/data/intake-api/generated/transaction/minimal_service.json
```diff
@@ -15,7 +15,8 @@
             "name": "GET /api/types",
             "type": "request",
             "duration": 32.592981,
-            "timestamp": "2017-05-30T18:53:27.154Z"
+            "timestamp": "2017-05-30T18:53:27.154Z",
+            "sampled": false
```
simitt: please remove from the minimal example as it is not a required field.

docs/data/intake-api/generated/transaction/minimal_span.json
```diff
@@ -13,6 +13,7 @@
             "type": "request",
             "duration": 32.592981,
             "timestamp": "2017-05-30T18:53:27.154Z",
+            "sampled": true,
```
simitt: please remove from the minimal example as it is not a required field.

docs/data/intake-api/generated/transaction/null_values.json
```diff
@@ -31,14 +31,16 @@
             "timestamp": "2017-05-30T18:53:27.154Z",
             "result": "200",
             "context": null,
-            "spans": null
+            "spans": null,
+            "sampled": false
```
simitt: please rset to `null` to test that you actually allow null values (this test will fail, as there is a null option missing in the spec).

docs/data/intake-api/generated/transaction/system_null.json
```diff
@@ -15,7 +15,8 @@
             "duration": 32.592981,
             "timestamp": "2017-05-09T15:04:05.999999Z",
             "context": null,
-            "spans": null
+            "spans": null,
+            "sampled": false
```
simitt: please remove as this tests that `system` can be null

docs/data/intake-api/generated/transaction/transaction_empty_values.json
```diff
@@ -16,7 +16,8 @@
             "duration": 32.592981,
             "start": 24.01,
             "timestamp": "2017-05-09T15:04:05.999999Z",
-            "spans": []
+            "spans": [],
+            "sampled": true
```
simitt: please remove as it is not a required field.

docs/fields.asciidoc
```diff
@@ -655,6 +655,14 @@ type: object
 A user-defined mapping of groups of marks. 
 
 
+[float]
+=== `transaction.sampled`
+
+type: boolean
+
+Sampled transactions have spans. Defaults to true
```
simitt: How about: `If the transaction is sampled, then all available information has been recorded, otherwise information about spans and context are discarded.

docs/spec/transactions/transaction.json
```diff
@@ -55,6 +55,10 @@
                 }
             },
             "additionalProperties": false
+        },
+        "sampled": {
+            "type": "boolean",
```
simitt: `"type": ["boolean", "null"]` otherwise you cannot send the key with an empty value.

processor/transaction/_meta/fields.yml
```diff
@@ -45,6 +45,10 @@
           description: >
             A user-defined mapping of groups of marks. 
 
+        - name: sampled
+          type: boolean
+          description: >
+            Sampled transactions have spans. Defaults to true
```
simitt: see comment about description in json spec.

tests/data/invalid/transaction/invalid_id.json
```diff
@@ -4,5 +4,6 @@
     "type": "request",
     "duration": 1.0,
     "timestamp": "2017-05-30T18:53:27.154Z",
-    "spans": []
+    "spans": [],
+    "sampled": true
```
simitt: please remove as it is not a required field, also from the other test json files.

simitt:
@roncohen you probably didn't run `approval`, as those still include default values for minimal examples.

docs/data/elasticsearch/transaction.json
roncohen: I don't think this is a `minimal` example?

roncohen:
@simitt "approved" files deal with the output right? those should have the `sampled` value as it defaults to `true` if it's not supplied in the intake api.

simitt:
@roncohen yeah right, weird that some specs are still failing for approved files. Is there an issue with files that you had added and are removed again?

roncohen:
> Is there an issue with files that you had added and are removed again?

Not sure what you mean. Could you elaborate?

PR #442, opened 2017-12-26
report more detail on max data size error

### LINKED ISSUE #438: When max unzipped limit is hit, APM Server responds with unhelpful 'unexpected EOF' error
It's possible to send gzip compressed payloads to APM Server. When the APM Server receives a compressed payload it will stop decompression after `MaxUnzippedSize` bytes. The purpose of the limit is to guard against zip bombs. If the payload is not fully read after `MaxUnzippedSize` bytes, it will result in an unhelpful EOF error when the JSON parsing of the payload is attempted.

The error is `Problem validating JSON document against schema: unexpected EOF`

We should emit a more helpful error message if the limit is hit.

graphaelli:
before:
```
{"error":"Decoding error: Data read error: unexpected EOF"}
```

after:
```
{"error":"while decoding: data read error: http: request body too large"}
```

closes #438

### DIFF
--- a/beater/handlers.go
+++ b/beater/handlers.go
@@ -5,14 +5,13 @@ import (
 	"compress/zlib"
 	"crypto/subtle"
 	"encoding/json"
-	"errors"
 	"fmt"
-	"io"
 	"net"
 	"net/http"
 	"strings"
 
 	"github.com/hashicorp/golang-lru"
+	"github.com/pkg/errors"
 	"github.com/ryanuber/go-glob"
 	"golang.org/x/time/rate"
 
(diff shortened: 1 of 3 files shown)


beater/handlers.go
```diff
@@ -323,10 +316,9 @@ func decodeLimitJSONData(maxSize int64) decoder {
 			}
 		}
 		v := make(map[string]interface{})
-		err := json.NewDecoder(io.LimitReader(reader, maxSize)).Decode(&v)
-		if err != nil {
+		if err := json.NewDecoder(http.MaxBytesReader(nil, reader, maxSize)).Decode(&v); err != nil {
```
roncohen: Didn't know about `MaxBytesReader`. Exactly what we need. Excellent!

beater/handlers.go
```diff
@@ -273,7 +267,7 @@ func processRequest(r *http.Request, pf ProcessorFactory, maxSize int64, report
 
 	data, err := decode(r)
 	if err != nil {
-		return http.StatusBadRequest, errors.New(fmt.Sprintf("Decoding error: %s", err.Error()))
+		return http.StatusBadRequest, errors.Wrap(err, "while decoding")
```
roncohen: 👍

PR #443, opened 2017-12-26
use merge=union for changelog

graphaelli:
from the docs:
  Run 3-way file level merge for text files, but take lines from both
  versions, instead of leaving conflict markers. This tends to leave the
  added lines in the resulting file in random order and the user should
  verify the result. Do not use this if you do not understand the
  implications.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+CHANGELOG.asciidoc merge=union


graphaelli:
Before, got this merge conflict:

```
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@@ -20,7 -19,7 +20,11 @@@ https://github.com/elastic/apm-server/c
  ==== Added
  - app.environment {pull}366[366]
++<<<<<<< HEAD
 +- Add context.request.url.full {pull}436[436]
++=======
+ - Add error.transaction.id {pull}437[437]
++>>>>>>> add error.transaction.id
```

After, no conflict, changelog ended up like:
```
==== Added
- app.environment {pull}366[366]
- Add context.request.url.full {pull}436[436]
- Add error.transaction.id {pull}437[437]
```

ruflin:
We use the same on the beats side: https://github.com/elastic/beats/blob/master/.gitattributes The only problem we encounter from time to time on cherry picks for backports that some lines are "sneaked" in. Meaning an other line is added to the commit which should not be there. But that is normally catched with the review.

PR #444, opened 2017-12-27
rename config param concurrent_requests to queue_size

graphaelli:
to better reflect purpose and sync with beats terminology.

Also bumps queue size default up from 20 to 100.


roncohen:
thanks @graphaelli. I agree that concurrent_requests is not the best name.

**Here's some context:**
Beats has a separate internal queue that we'd like to use as the primary means of queuing. Events coming in get added to our own queue (which has size of `concurrent_requests`) and then they get added to the queue in beats from there, here: https://github.com/elastic/apm-server/blob/master/beater/pub.go#L84-L86. The call to `PublishAll` will block if the queue in beats is full.

If that queue is full, we'll fill up our own queue and eventually reject the incoming data here: https://github.com/elastic/apm-server/blob/master/beater/pub.go#L75-L79

Side note: Ideally, instead of waiting a second, we'd reject immediately when the queue is full. We tried removing that timeout but we saw that even when there's room in the internal beats queue, the go scheduler may not choose to run the go-routine that would unblock this [PublishAll](https://github.com/elastic/apm-server/blob/master/beater/pub.go#L85) call and the `Send` call would return with an error.

It's worth noting that our "own queue" deals with batches of events (`[]Event`) while the the beats queue has type `Event`. Processors generate batches of events. Some of these events relates to each other (Span relates to Transaction for example). If Elasticsearch is barely managing to index incoming events, we could end in a situation where the beats queue is borderline full all the time. Now, if we handled each event separately instead of in batches, it could easily happen than batches of events will get partially published to Elasticsearch. So we added this queue of batches in order to minimize the possibility that batches would get partially published in heavy load environments. 

The reason we initially named it "concurrent_requests" was that when the beats queue is not full, we should be able to have `concurrent_requests` number of concurrent requests while the PublishAll call finishes. 

Ideally, we would find a way to get rid of our own queue all together. An alternative is to just let the incoming requests call `PublishAll` as part of their request/response cycle. We'd need serialize calls  `PublishAll` while ensuring that incoming requests are not just piling up if it's blocked. Not saying we need to do that right now.

graphaelli:
closing this in favor of investigating how to rework our queue

PR #445, opened 2017-12-28
Move process related fields to their own namespace

roncohen:
This affects the intake API as well as the output documents.

This puts us on the path to align better with other elastic projects in the future.

### DIFF
--- a/_meta/fields.common.yml
+++ b/_meta/fields.common.yml
@@ -141,6 +141,21 @@
               type: keyword
               description: >
                 The platform of the system the event was recorded on.
+        - name: process
+          type: group
+          description: >
+            Information pertaining to the running process where the data was collected
+          fields:
+
+            - name: pid
+              type: long
+              description: >
+                Numeric process ID of the service process.
+
+            - name: title
+              type: keyword
+              description: >
+                Service process title.
 
         - name: service
           type: group
@@ -164,16 +179,6 @@
               type: keyword
               description: >
                 Version of the service emitting this event.
-
-            - name: pid
-              type: long
-              description: >
-                Numeric process ID of the service process.
-
-            - name: process_title
-              type: keyword
-              description: >
-                Service process title.
 
             - name: environment
               type: keyword
--- a/docs/spec/errors/payload.json
+++ b/docs/spec/errors/payload.json
@@ -1,12 +1,15 @@
 {
     "$schema": "http://json-schema.org/draft-04/schema#",
-    "$id": "docs/spec/errors/wrapper.json",
-    "title": "Errors Wrapper",
-    "description": "List of errors wrapped in an object containing some other attributes normalized away form the errors themselves",
+    "$id": "docs/spec/errors/payload.json",
+    "title": "Errors payload",
+    "description": "List of errors wrapped in an object containing some other attributes normalized away from the errors themselves",
     "type": "object",
     "properties": {
         "service": {
             "$ref": "../service.json"
+        },
+        "process": {
+            "$ref": "../process.json"
         },
         "errors": {
             "type": "array",
--- a/docs/spec/process.json
+++ b/docs/spec/process.json
@@ -1,0 +1,22 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "$id": "doc/spec/process.json",
+  "title": "Process",
+  "type": ["object", "null"],
+  "properties": {
+      "pid": {
+          "description": "Process ID of the service",
+          "type": ["number", "null"]
+      },
+      "title": {
+          "type": ["string", "null"],
+          "maxLength": 1024
+      },
+      "argv": {
+        "description": "Command line arguments used to start this process",
+        "type": ["array", "null"],
+        "minItems": 0
+    }
+  },
+  "required": ["pid"]
+}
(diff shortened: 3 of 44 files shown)


_meta/fields.common.yml
```diff
@@ -141,6 +141,21 @@
               type: keyword
               description: >
                 The platform of the system the event was recorded on.
+        - name: process
+          type: group
+          description: >
+            Information pertainin to the running process where the data was collected
```
graphaelli: typo in `pertainin`

docs/spec/errors/payload.json
```diff
@@ -1,13 +1,16 @@
 {
     "$schema": "http://json-schema.org/draft-04/schema#",
     "$id": "docs/spec/errors/wrapper.json",
```
graphaelli: while you're at it, `payload.json`

processor/error/schema.go
```diff
@@ -7,8 +7,8 @@ func Schema() string {
 var errorSchema = `{
     "$schema": "http://json-schema.org/draft-04/schema#",
     "$id": "docs/spec/errors/wrapper.json",
```
graphaelli: again `payload.json`

roncohen:
thanks @graphaelli ! Fixed nits and rebased to resolve conflicts.

roncohen:
Breaking change: @elastic/apm-ui  @elastic/apm-agent-devs

beniwohli:
@roncohen when is this supposed to land? If 6.2, can you add it to the 6.2 checklist?

roncohen:
@beniwohli done!

docs/spec/process.json
```diff
@@ -0,0 +1,22 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "$id": "doc/spec/process.json",
+  "title": "Process",
+  "type": ["object", "null"],
+  "properties": {
+      "pid": {
+          "description": "Process ID of the service",
+          "type": ["number", "null"]
```
simitt: why is `null` allowed when it is a required field?

model/process.go
```diff
@@ -0,0 +1,27 @@
+package model
+
+import (
+	"github.com/elastic/apm-server/utility"
+	"github.com/elastic/beats/libbeat/common"
+)
+
+type Process struct {
+	Pid   *int
```
simitt: When `pid` is required you could use `int` instead of `*int` here.

processor/transaction/event.go
```diff
@@ -34,14 +34,20 @@ func (t *Event) Transform() common.MapStr {
...
+		{Key: "context", Apply: func() common.MapStr { return t.Context }},
+		{Key: "context.service", Apply: pa.Service.Transform},
+		{Key: "context.system", Apply: pa.System.Transform},
+	}
+
+	if pa.Process != nil {
+		mapping = append(mapping, m.DocMapping{
+			Key: "context.process", Apply: pa.Process.Transform,
+		})
+	}
```
simitt: Why do you handle this differently than for `errors` here https://github.com/elastic/apm-server/pull/445/files#diff-a163837d2241a4bad19d7e4baa4c617bR62?

I suggest to rather add a nil check when processing the mappings, as e.g. `system` is also not a required field.

processor/error/package_tests/TestProcessErrorNullValues.approved.json
```diff
@@ -3,6 +3,7 @@
         {
             "@timestamp": "2017-05-09T15:04:05.999999Z",
             "context": {
+                "process": {},
```
simitt: I think `process` should be able to be `null`, so this should be checked here. 
If `process` is given, it should however have a `pid`, at least this is required in the spec above. If you remove the `null` type in the spec, sending an empty hash for `process` should not be possible any longer.

processor/transaction/package_tests/TestProcessTransactionNullValues.approved.json
```diff
@@ -3,6 +3,7 @@
         {
             "@timestamp": "2017-05-30T18:53:27.154Z",
             "context": {
+                "process": {},
```
simitt: same comment as above - empty `process` does not reflect spec, where `pid` is required.

tests/data/valid/transaction/null_values.json
```diff
@@ -22,6 +22,11 @@
         "architecture": null,
         "platform": null
     },
+    "process": {
+        "pid": null,
```
simitt: again - this doesn't make sense to me - either `process.pid` is required, or it is allowed to be null.

processor/transaction/event.go
roncohen: oversight on my part, good catch

roncohen:
great feedback. Thanks @simitt !

simitt:
@roncohen some tests are failing, apart from that LGTM

beniwohli:
Implemented for Python in https://github.com/elastic/apm-agent-python/pull/125

makwarth:
@roncohen could you elaborate on the UI implications?

roncohen:
@makwarth yes. The data is moved around in the error document and the transactions document:

Transaction: https://github.com/elastic/apm-server/pull/445/files#diff-f63f8f194850ee66420e85c23183a538

Error: https://github.com/elastic/apm-server/pull/445/files#diff-ce810586885ee0e452b0c394d2d61ee0

PR #446, opened 2018-01-02
String source map

roncohen:
Based on @simitt's PR https://github.com/elastic/apm-server/pull/432 but updated to store as string.


simitt:
@roncohen could run `make update` please so the `NOTICE` changes accordingly.

roncohen:
@simitt done

simitt:
@roncohen what do you think about this https://github.com/elastic/apm-server/pull/432/files#r159254075?

simitt:
After several discussions and changes in beats I think we should upload the information as a base64 encoded string, stored as binary in ES, as it is done by PR #432 and close this one.

PR #448, opened 2018-01-04
Support dropped span_count on transactions.

### LINKED ISSUE #280: More efficient way to send/store repeated spans?
@gingerwizard brought up an interesting problem. In a purposely-made inefficient view, he queries a list of 15000 items one by one instead of a single query, to demonstrate a problem that can be detected and analyzed with APM.

The problem is, due to the huge amount of traces, the payload surpasses our payload size limit, and the transaction is lost.

In the Opbeat agents, we did some optimization by only storing the stacktrace once, but that's not done in the new Elastic APM agents.

A very simplistic approach could be to check if the just-ending trace is equivalent (same trace name, same stack trace, maybe even same context) with the preceding trace. If yes, increment a counter field, and add the duration to the preceding trace, don't create a new trace.

Even this simplistic approach has some drawbacks though. We lose some information (the duration of each individual repetition), and the UI would need to expose the count somehow. Also, things like "Top N queries" would get more complicated to implement.

Any ideas?

simitt:
closes #280.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -25,6 +25,8 @@ https://github.com/elastic/apm-server/compare/71df0d96445df35afe27f38bcf734a0828
 - Add transaction.id to errors {pull}437[437]
 - Support for `transaction.marks` {pull}430[430]
 - Support for sourcemap mapping on incoming frontend requests {pull}381[381]
+- Support for `transaction.marks` {pull}430[430].
+- Support for `transaction.span_count.dropped.total` {pull}448[448].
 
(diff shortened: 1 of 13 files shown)


roncohen:
Good work @simitt. I think it would be useful to add the field to `fields.yml` to get it indexed and in the documentation. WDYT?

simitt:
I was thinking about indexing it, but decided against it as the UI doesn't need it. But giving it a second thought it is probably worth to be indexed.

simitt:
jenkins, retest this please.

Issue #431, opened 2017-12-22
Streamline logging


roncohen:
Now that we have [structured logging](https://github.com/elastic/beats/pull/5901#issuecomment-352594717), we should go through APM Server and streamline logging.

Generally, we should have things like `service.name` and `agent.name` as fields whenever there's a rejected payload, etc.

PR #458, opened 2018-01-05
Streamline logging

roncohen:
Initial work on streamlining logging and moving towards structured logging.

beats introduced structured logging. The fact that we output JSON on the command line was part of this and it's already merged in beats.

This PR only deals with what we can control in APM Server. I'm looking into if we can use a key-value style instead of the JSON when started with `-e`.

With this, we will always log something like this for each request:
```
2018-01-05T16:19:39.317+0100	INFO	[request]	beater/handlers.go:137	handled request	{"request_id": "180acb9f-2128-4955-97e6-2485b6330d8c", "response_code": 200, "method": "GET", "URL": "/healthcheck", "content_length": 0, "remote_address": "127.0.0.1", "user_agent": "Go-http-client/1.1"}
```

Errors get another line:

```
2018-01-05T16:19:52.433+0100	ERROR	[request]	beater/handlers.go:365	error handling request	{"request_id": "b54dde2e-590e-4084-9ffe-cded450de649", "error": "Problem validating JSON document against schema: I[#] S[#] doesn't validate with \"error#\"\n  I[#] S[#/required] missing properties: \"service\", \"errors\""}
2018-01-05T16:19:52.434+0100	INFO	[request]	beater/handlers.go:137	handled request	{"request_id": "b54dde2e-590e-4084-9ffe-cded450de649", "response_code": 400, "method": "POST", "URL": "/v1/errors", "content_length": 132, "remote_address": "127.0.0.1", "user_agent": "curl/7.54.0"}
```

closes https://github.com/elastic/apm-server/issues/431

### DIFF
--- a/beater/handlers.go
+++ b/beater/handlers.go
@@ -3,6 +3,7 @@ package beater
 import (
 	"compress/gzip"
 	"compress/zlib"
+	"context"
 	"crypto/subtle"
 	"encoding/json"
 	"fmt"
@@ -13,6 +14,7 @@ import (
 	lru "github.com/hashicorp/golang-lru"
 	"github.com/pkg/errors"
 	glob "github.com/ryanuber/go-glob"
+	uuid "github.com/satori/go.uuid"
 	"golang.org/x/time/rate"
 
 	"github.com/elastic/apm-server/processor"
@@ -20,6 +22,7 @@ import (
 	"github.com/elastic/apm-server/processor/healthcheck"
 	"github.com/elastic/apm-server/processor/sourcemap"
 	"github.com/elastic/apm-server/processor/transaction"
+	"github.com/elastic/apm-server/utility"
 	"github.com/elastic/beats/libbeat/logp"
 	"github.com/elastic/beats/libbeat/monitoring"
 )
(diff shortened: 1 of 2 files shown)


beater/handlers.go
```diff
@@ -108,12 +111,40 @@ func healthCheckHandler(_ ProcessorFactory, _ Config, _ reporter) http.Handler {
...
+
 		requestCounter.Inc()
-		h.ServeHTTP(w, r)
+
+		reqLogger := logger.With("request_id", reqID)
+
+		lr := r.WithContext(
+			context.WithValue(r.Context(), reqLoggerContextKey, reqLogger),
+		)
+		lr.Context().Value(reqLoggerContextKey)
```
simitt: Is this line a leftover?

beater/handlers.go
```diff
@@ -108,12 +111,40 @@ func healthCheckHandler(_ ProcessorFactory, _ Config, _ reporter) http.Handler {
...
+
+		lw := utility.NewRecordingResponseWriter(w)
+
+		h.ServeHTTP(lw, lr)
+
+		reqLogger.Infow("handled request", "response_code", lw.Code,
+			"method", r.Method, "URL", r.URL, "content_length", r.ContentLength,
+			"remote_address", extractIP(r), "user_agent", r.Header.Get("User-Agent"))
+
+		if lw.Code > 399 {
```
simitt: Should this be the other way around? `error if Code > 399`

roncohen:
doh :facepalm:. Thanks!

roncohen:
jenkins, restest this please

beater/handlers.go
roncohen: fixed, thanks! 🤦‍♂️

beater/handlers.go
```diff
@@ -331,15 +361,17 @@ func sendStatus(w http.ResponseWriter, r *http.Request, code int, err error) {
...
 	w.WriteHeader(code)
 
-	logp.Info("%s	%s %s	%d	%s", extractIP(r), r.Method, r.URL, code, r.Header.Get("User-Agent"))
 	if err == nil {
-		responseValid.Inc()
 		return
 	}
 
-	logp.Err("%s, code=%d", err.Error(), code)
+	logger, ok := r.Context().Value(reqLoggerContextKey).(*logp.Logger)
```
jalvz: @roncohen 
in what case `ok` would be false?
what is the benefit of putting and getting the logger in the context instead of doing another `logger := logp.NewLogger("request")` here? doesn't seem an expensive operation

roncohen: `ok` would only be false if `logHandler` somehow wasn't applied. I didn't want to run the risk of breaking a customer app in production, so i did this fallback here. We can do a `logp.NewLogger("request").Err(..)` here if `ok` is false. 

I put the logger in the context because the logger has context that we don't otherwise pass around for example `request_id`.

jalvz: got it, thanks

PR #461, opened 2018-01-08
Smapping changes

simitt:
I gave the structure and testing of the current sourcemap implementation another thought and refactored the sourcemap to move it out of `utility` into its own `sourcemap` package. 
This helped splitting the functionality up and adding more unit tests for better coverage.


beater/config.go
```diff
@@ -66,23 +65,23 @@ func (s *Sourcemapping) isSetup() bool {
 	return s != nil && (s.esConfig != nil)
 }
 
-func (c *FrontendConfig) SmapAccessor() utility.SmapAccessor {
+func (c *FrontendConfig) SmapMapper() sourcemap.Mapper {
 	smap := c.Sourcemapping
 	if c.isEnabled() && smap.isSetup() {
-		if smap.accessor == nil {
-			smapConfig := utility.SmapConfig{
+		if smap.mapper == nil {
```
graphaelli: While we're refactoring, how about flipping the conditional, `return nil` early?

beater/config.go
```diff
@@ -66,23 +65,23 @@ func (s *Sourcemapping) isSetup() bool {
...
 				CacheCleanupInterval: smap.Cache.CleanupInterval,
 				ElasticsearchConfig:  smap.esConfig,
-				Index:                smap.Index,
+				Index:                smap.Index + "*",
 			}
-			smapAccessor, err := utility.NewSourcemapAccessor(smapConfig)
+			smapMapper, err := sourcemap.NewSmapMapper(smapConfig)
 			if err != nil {
-				logp.Err(fmt.Sprintf("Error creating Sourcemap Accessor: %v", err.Error()))
+				logp.Err(err.Error())
```
graphaelli: Didn't consider this before but perhaps this should return an error instead?  That way the handler could do the logging instead of the library

docs/data/elasticsearch/error.json
```diff
@@ -193,7 +192,7 @@
                     },
                     "filename": "file/name.py",
                     "function": "foo",
-                    "library_frame": true,
+                    "library_frame": false,
```
graphaelli: any specific reason to flip all of these?  Concerned that since it's the default we're losing coverage on the `true` code path.

sourcemap/accessor.go
```diff
@@ -0,0 +1,76 @@
...
+		return consumer, nil
+	}
+	if found {
+		return nil, errSmapNotAvailable(id)
+	}
+	consumer, err := s.es.fetch(id)
+	if err != nil {
+		return nil, err
+	}
+	s.cache.add(id, consumer)
```
graphaelli: This can be negative cache, to work with the comment about avoiding ES fetch above?  Is there a concern about timing between indexing a new sourcemap upload and this being called?  Is a shorter TTL on the negative cache possible?

model/stacktrace_frame.go
```diff
@@ -59,49 +59,43 @@ func (s *StacktraceFrame) Transform(service Service, smapAccessor utility.SmapAc
...
-		Path:           cleanedPath,
-	})
+	sourcemapId := s.buildSourcemapId(service)
+	mapping, err := mapper.Apply(sourcemapId, s.Lineno, *s.Colno)
 	if err != nil {
 		logp.Err(fmt.Sprintf("Sourcemap fetching Error %s", err.Error()))
-		e, isSmapError := err.(utility.SmapError)
-		if !isSmapError || e.Kind == utility.MapError {
+		e, issourcemapError := err.(sourcemap.Error)
+		if !issourcemapError || e.Kind == sourcemap.MapError || e.Kind == sourcemap.KeyError {
```
graphaelli: Is there a test to cover this?  if `!issourcemapError` `e` will be `nil` right?

docs/data/elasticsearch/error.json
simitt: There is still one `library_frame: true` left.
The reason for flipping is to get more variety, which we need for the next feature. 
Usually I'd add it to the next PR, but since I had all of this already in I left this.

sourcemap/accessor.go
simitt: It is on purpose to add nil values to the cache if they cannot be found. 

Reason for this is:
We do not offer a seperate option to enable/disable sourcemapping, neither do we offer to enable/disable it per service. Right now, as soon as `frontend.enabled=true` sourcemapping is enabled. So for every incoming `trancsaction` and `error` the processor checks whether there is a sourcemap available and tries to apply it. For customers/services that want to have frontend support but haven't uploaded sourcemaps, this would mean that we would fetch ES for every stacktraceframe of every record. 

It would be possible to set a different TTL for negative cache values, however I'd rather avoid having another config option here. Hardcoding and not providing a config option would also be weird, when the general TTL is configurable.

Regarding timing issue:
Depending on the TTL it could ofc be that a sourcemap is uploaded but the cache still holds the nil value. TTL default is 5 mins, so I wouldn't be too concerned about this right now. 
As this feature is experimental my main concern with caching was more directed to not influence the general APM server and ES performance too much with it.

model/stacktrace_frame.go
simitt: [This test](https://github.com/simitt/apm-server/blob/d5e687adf0935ddca571de56336a5de27174111f/model/stacktrace_frame_test.go#L100) covers it, see the [mocked return values](https://github.com/simitt/apm-server/blob/d5e687adf0935ddca571de56336a5de27174111f/model/stacktrace_frame_test.go#L218)

simitt: As I am not using `e`, but `err` to retrieve the message I cannot see a problem here.

graphaelli: Ugh, sorry, I was thinking the conditional wouldn't short-circuit, leaving `e` as `nil`, disregard.

sourcemap/accessor.go
graphaelli: Sounds good.  An option to disable sourcemapping while keeping frontend enabled seems like a great idea to start.

simitt: https://github.com/elastic/apm-server/issues/472

sourcemap/elasticsearch.go
```diff
@@ -0,0 +1,115 @@
...
+	for _, client := range e.clients {
+		_, result, err = client.Connection.SearchURIWithBody(e.index, "", nil, body)
+		if err == nil {
+			return result, nil
+		}
+	}
+	if err != nil {
+		return nil, Error{Msg: err.Error(), Kind: AccessError}
+	}
+	return result, nil
```
jalvz: @simitt this line never will be reached, right? Also, if the program reaches L50, `err` will always be `not nil` because setup ensures that there is at least 1 client.

sourcemap/elasticsearch.go
```diff
@@ -0,0 +1,115 @@
...
+}
+
+type smapElasticsearch struct {
+	clients []es.Client
+	index   string
+}
+
+func NewElasticsearch(config *common.Config, index string) (*smapElasticsearch, error) {
+	esClients, err := es.NewElasticsearchClients(config)
+	if err != nil || esClients == nil || len(esClients) == 0 {
```
jalvz: looking at es.NewElasticsearchClients, part of this condition is redundant, `esClients == nil || len(esClients) == 0` can not happen here.
shall i create a pr for this?

sourcemap/elasticsearch.go
simitt: hm, yes I think you're right.

processor/sourcemap/payload.go
```diff
@@ -20,14 +21,14 @@ type payload struct {
...
+func (pa *payload) transform(config *pr.Config) []beat.Event {
 	var events = []beat.Event{pr.CreateDoc(mappings(pa))}
 	sourcemapCounter.Add(1)
 
-	if smapAccessor == nil {
+	if config == nil || config.SmapMapper == nil {
 		logp.Err("Sourcemap Accessor is nil, cache cannot be invalidated.")
 	} else {
-		smapAccessor.RemoveFromCache(utility.SmapID{
+		config.SmapMapper.NewSourcemapAdded(smap.Id{
```
jalvz: This is tricky, if the sourcemap can't be saved in ES for whatever reason, it will only be saved in the cache.
Once it expires, sourcemaps will suddenly stop working.

Quite a corner case, but 🤷‍♂️

jalvz: aahh, scratch that. that is removing it, not adding it 🤦‍♂️ 
all good then

PR #464, opened 2018-01-09
Avoid overwriting default config of beat.

simitt:
* Add config tests
* Change handling to not overwrite default config values.

The `defaultConfig` is only used once in production code. However, adding the config tests revealed that the beater currently changes the `defaultConfig` when [unpacking the ucfg](https://github.com/simitt/apm-server/blob/master/beater/beater.go#L19).
This change avoids having the `defaultConfig` overwritten, which could lead to issues in the future when used after initializing the beat.

### DIFF
--- a/beater/config.go
+++ b/beater/config.go
@@ -33,8 +33,8 @@ type Sourcemapping struct {
 	Cache *Cache `config:"cache"`
 	Index string `config:"index"`
 
-	elasticsearch *common.Config
-	accessor      utility.SmapAccessor
+	esConfig *common.Config
+	accessor utility.SmapAccessor
 }
 
 type Cache struct {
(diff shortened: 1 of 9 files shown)


beater/config.go
```diff
@@ -81,25 +87,27 @@ func (c *FrontendConfig) SmapAccessor() utility.SmapAccessor {
...
-	SecretToken:        "",
-	Frontend: &FrontendConfig{
-		Enabled:      new(bool),
-		RateLimit:    10,
-		AllowOrigins: []string{"*"},
-		Sourcemapping: &Sourcemapping{
-			Cache: &Cache{
-				Expiration:      300,
-				CleanupInterval: 600,
+func defaultConfig() Config {
```
graphaelli: While we're at it, it seems strange to copy `Config`s around rather than passing a pointer to _the_ config around.  What do you think about this returning a pointer?

simitt: good point, changing to pass around *the* config everywhere.

graphaelli: well played!

PR #473, opened 2018-01-10
Update beats framework to 5aafc20

graphaelli:


graphaelli:
mostly interested in pulling elastic/beats#6010 and elastic/beats#6032

Issue #466, opened 2018-01-10
Race when starting APM Server on a used port


roncohen:
When starting APM Server on a port that is already in use, it will occasionally `panic: send on closed channel` instead of the intended message 

```
2018-01-10T13:14:42.170+0100	ERROR	instance/beat.go:666	Exiting: listen tcp 127.0.0.1:8200: bind: address already in use
Exiting: listen tcp 127.0.0.1:8200: bind: address already in use
```

Full output:

```$ ./apm-server -e
2018-01-10T13:14:40.812+0100	INFO	instance/beat.go:467	Home path: [/Users/ron/Projects/elastic/goapps/src/github.com/elastic/apm-server] Config path: [/Users/ron/Projects/elastic/goapps/src/github.com/elastic/apm-server] Data path: [/Users/ron/Projects/elastic/goapps/src/github.com/elastic/apm-server/data] Logs path: [/Users/ron/Projects/elastic/goapps/src/github.com/elastic/apm-server/logs]
2018-01-10T13:14:40.812+0100	INFO	instance/beat.go:474	Beat UUID: 999256f1-bc70-4224-86d0-1bf6891a9d6c
2018-01-10T13:14:40.812+0100	INFO	instance/beat.go:212	Setup Beat: apm-server; Version: 7.0.0-alpha1
2018-01-10T13:14:40.812+0100	INFO	elasticsearch/client.go:123	Elasticsearch url: http://localhost:9200
2018-01-10T13:14:40.813+0100	INFO	pipeline/module.go:76	Beat name: VeryFastPicket.local
2018-01-10T13:14:40.813+0100	INFO	instance/beat.go:300	apm-server start running.
2018-01-10T13:14:40.813+0100	INFO	[monitoring]	log/log.go:87	Starting metrics logging every 30s
2018-01-10T13:14:40.813+0100	INFO	beater/handlers.go:79	Path /v1/transactions added to request handler
2018-01-10T13:14:40.813+0100	INFO	beater/handlers.go:79	Path /v1/client-side/transactions added to request handler
2018-01-10T13:14:40.813+0100	INFO	beater/handlers.go:79	Path /v1/errors added to request handler
2018-01-10T13:14:40.813+0100	INFO	beater/handlers.go:79	Path /v1/client-side/errors added to request handler
2018-01-10T13:14:40.813+0100	INFO	beater/handlers.go:79	Path /healthcheck added to request handler
[...]

PR #474, opened 2018-01-11
create listener separate from starting server

graphaelli:
Inspired by #466, this doesn't perfectly resolve that, but makes it even less likely.  This does resolve issues running `server_test`s while port 8200 is already in use - that currently hangs until `waitForServer` times out.  It also eliminates suspected issues from opening a random port, closing it, and immediately reopening it via `randomAddr()`.

### DIFF
--- a/beater/beater.go
+++ b/beater/beater.go
@@ -2,6 +2,7 @@ package beater
 
 import (
 	"fmt"
+	"net"
 	"net/http"
 
 	"github.com/elastic/beats/libbeat/beat"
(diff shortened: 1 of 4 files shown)


graphaelli:
Todo: ensure listeners close on server shutdown

roncohen:
could we get rid of `waitForServer` entirely?

graphaelli:
Since we're spinning off the server into it's own goroutine keep it is probably preventing flakiness in these tests.

graphaelli:
Confirmed that listeners are closed on `Shutdown`, even if the server is started with `Serve` instead of `ListenAndServe`

PR #481, opened 2018-01-11
Detect library_frames for frontend stacktraces.

### LINKED ISSUE #426: Detect if a stacktraceframe is library_frame or not.
Set boolean `library_frame` for every stacktrace frame if not set.

Allow for configurable regexp indicating a frame is a `library_frame`. 

@jahtalab will look into whether default values make sense if no config is given.

simitt:
closes #426 

@jahtalab fyi

### DIFF
--- a/model/stacktrace_frame.go
+++ b/model/stacktrace_frame.go
@@ -2,6 +2,7 @@ package model
 
 import (
 	"fmt"
+	"regexp"
 
 	pr "github.com/elastic/apm-server/processor"
 	"github.com/elastic/apm-server/sourcemap"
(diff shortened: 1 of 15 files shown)


model/stacktrace_frame.go
```diff
@@ -43,6 +44,9 @@ func (s *StacktraceFrame) Transform(config *pr.Config, service Service) common.M
 	enhancer.Add(m, "module", s.Module)
 	enhancer.Add(m, "function", s.Function)
 	enhancer.Add(m, "vars", s.Vars)
+	if config != nil && config.LibraryPattern != nil && s.LibraryFrame == nil {
+		s.LibraryFrame = s.isLibraryFrame(config.LibraryPattern)
```
roncohen: `isLibraryFrame` can never return `nil`, so it should return a `bool` instead of `*bool` and then we should store it in a variable here and set `s.LibraryFrame` to the address of that var.

PR #482, opened 2018-01-11
Exclude certain stacktrace frames from grouping key.

simitt:
* Only exclude if frontend record.
* Hardcode exclude patterns for now.

@jahtalab fyi


processor/error/event.go
```diff
@@ -204,3 +214,14 @@ func (e *Event) calcGroupingKey() string {
 func (e *Event) add(key string, val interface{}) {
 	e.enhancer.Add(e.data, key, val)
 }
+
+func use_for_grouping(fr *m.StacktraceFrame) bool {
+	if fr.Filename == "" || strings.HasPrefix(fr.Filename, "/webpack") {
+		return false
+	}
+	file := filepath.Base(fr.Filename)
+	if strings.HasPrefix(fr.Filename, ("/" + file)) {
```
graphaelli: What is this supposed to exclude?

processor/error/event.go
```diff
@@ -204,3 +214,14 @@ func (e *Event) calcGroupingKey() string {
 func (e *Event) add(key string, val interface{}) {
 	e.enhancer.Add(e.data, key, val)
 }
+
+func use_for_grouping(fr *m.StacktraceFrame) bool {
```
graphaelli: should be camel case (`useForGrouping`)

processor/error/event_test.go
```diff
@@ -225,21 +225,48 @@ func TestExplicitGroupingKey(t *testing.T) {
...
 
+func TestFramesUsableForGroupingKey(t *testing.T) {
+	frontendConfig := pr.Config{Frontend: true}
+	st1 := m.Stacktrace{
+		m.StacktraceFrame{Filename: "/a/b/c", Lineno: 123},
+		m.StacktraceFrame{Filename: "webpack", Lineno: 77},
+		m.StacktraceFrame{Filename: "~/tmp", Lineno: 45},
+		m.StacktraceFrame{Filename: "tmp", Lineno: 89},
+		m.StacktraceFrame{Filename: "", Lineno: 13},
+		m.StacktraceFrame{Filename: "/webpack", Lineno: 7},
```
graphaelli: how about adding another element to this path, so that it wouldn't match https://github.com/elastic/apm-server/pull/482/files#diff-a163837d2241a4bad19d7e4baa4c617bR223 in case the first filter was broken

jahtalab:
@simitt , Do you mind creating an issue to make hardcoded patterns configurable later (Just want to make sure we don't forget this going forward)

simitt:
@jahtalab @roncohen as this was only supposed to be a temporary solution until we have a config option for it, I didn't want to invest more time to this.
However after @roncohen 's comment I played around a bit with regex and immediatly made it configurable, reflecting the same pattern as default regex that we had decided on for the standard exclude algorithm.

(This PR is based on PR #485 which contains a bugfix for stacktrace frames. )

beater/handlers.go
```diff
@@ -96,8 +96,9 @@ func frontendHandler(pf ProcessorFactory, config *Config, report reporter) http.
 		logp.Err(err.Error())
 	}
 	prConfig := processor.Config{
-		SmapMapper:     smapper,
-		LibraryPattern: regexp.MustCompile(config.Frontend.LibraryPattern),
+		SmapMapper:          smapper,
+		LibraryPattern:      regexp.MustCompile(config.Frontend.LibraryPattern),
+		ExcludeFromGrouping: regexp.MustCompile(config.Frontend.ExcludeFromGrouping),
```
graphaelli: I didn't notice that we were doing this already with `LibraryPattern` but are we ok with a panic if a bad regex is used?  That seems too unfriendly for GA but maybe ok for alpha frontend support?

graphaelli: Maybe something like:
```
diff --git a/beater/beater.go b/beater/beater.go
index bce4eac..feb0124 100644
--- a/beater/beater.go
+++ b/beater/beater.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"net"
 	"net/http"
+	"regexp"
 
@@ -26,6 +27,16 @@ func New(b *beat.Beat, ucfg *common.Config) (beat.Beater, error) {
 		beaterConfig.setElasticsearch(b.Config.Output.Config())
 	}
 
+	if beaterConfig.Frontend.isEnabled() {
+		if _, err := regexp.Compile(beaterConfig.Frontend.LibraryPattern); err != nil {
+			return nil, err
+		}
+
+		if _, err := regexp.Compile(beaterConfig.Frontend.ExcludeFromGrouping); err != nil {
+			return nil, err
+		}
+	}
+
 	bt := &beater{
 		config: beaterConfig,
 	}
```

But with errors wrapped with helpful messaging.

roncohen:
Great stuff @silvia. I’m on my phone at the moment, but it looks like we should move the exclude_from_grouping config out of frontend namespace in the config as ExcludeFromGrouping is not only for fronted anymore.

simitt:
Thanks @roncohen. I intentially left it under the frontend namespace. This way we could also provide a backend specific config option, so that we support both at the same time. Internally we then only have to pass along the backend specific pattern to the backend handlers and the frontend specific pattern to the frontend handlers.

beater/beater.go
```diff
@@ -21,9 +23,18 @@ func New(b *beat.Beat, ucfg *common.Config) (beat.Beater, error) {
 	if err := ucfg.Unpack(beaterConfig); err != nil {
 		return nil, fmt.Errorf("Error reading config file: %v", err)
 	}
-
-	if b.Config != nil && b.Config.Output.Name() == "elasticsearch" {
-		beaterConfig.setElasticsearch(b.Config.Output.Config())
+	if beaterConfig.Frontend.isEnabled() {
+		_, err := regexp.Compile(beaterConfig.Frontend.LibraryPattern)
```
graphaelli: I would prefer the one line assignment and nil check here, to limit `err`'s lifetime

roncohen:
> This way we could also provide a backend specific config option, so that we support both at the same time.

But at the moment we have a configuration option under `frontend` that also affects `backend`, as far as i can tell. We need to either move the configuration option or make it only affect `frontend` data.

_meta/beat.reference.yml
```diff
@@ -49,6 +49,14 @@ apm-server:
     # If the regexp matches, the stacktrace frame is considered to be a library frame.
     #library_pattern: "node_modules|bower_components|~"
 
+    # Regexp to be matched against a stacktrace frame's `file_name`.
+    # If the regexp matches, the stacktrace frame is not used for calculating error groups.
+    # The default pattern excludes stacktrace frames that have
+    # - an empty filename,
+    # - a filename starting with '/webpack'
+    # - a filename in the format of an absolute path only including the filename, e.g. /myfile.
+    #exclude_from_grouping: "^\\s*$|^/webpack|^[/][^/]*$"
```
roncohen: The `/webpack` part of this pattern looks fine. I understand the rest is something we have from the original Opbeat code (`filename.startswith('/' + frame['filename'].split('/')[-1])`). Do we know why it's like that? If not, I suggest we remove it for now (remove the `|^[/][^/]*` part) and add it back in if we come across the use-case.

simitt:
@roncohen please have a look at the [config options](https://github.com/simitt/apm-server/blob/b94f44cc465a08da167d49ee90a380415c1c2d4f/beater/handlers.go#L87) we pass to the handler. 
The backend handler does not pass in any information that comes from the frontend section, so we only use the `exclude_from_grouping` key in the frontend. It is the same handling that I also used for passing along `sourcemap` and `library_pattern` information.

_meta/beat.reference.yml
simitt: great, I will change that.

roncohen:
@simitt Missed that 👍

_meta/beat.reference.yml
simitt: @roncohen do you mean to only remove `|^[/][^/]*` or also the part describing the empty filename `^\\s*$|`?

roncohen: I don't know where the empty file pattern part came from. I don't recall seeing that anywhere?

simitt: I was also part of the original Opbeat logic, I can also remove it, just asking because you explicitly mentioned to remove `|^[/][^/]* `.

PR #485, opened 2018-01-12
Use Stacktrace Frame pointers.

simitt:
Allow using transformed stack trace frame values for calculation.

Right now we change the values of stacktrace frames in the `Transform` method, but as the `stacktrace` doesn't hold pointer references to the stacktrace frames the changed values are not available for the `error` objects holding `stacktraces`. This leads to the usage of the wrong frame information, in case it was changed, as e.g. for sourcemapping.

### DIFF
--- a/processor/error/event_test.go
+++ b/processor/error/event_test.go
@@ -5,11 +5,16 @@ import (
 	"encoding/hex"
 	"fmt"
 	"io"
+	"io/ioutil"
+	"os"
+	"path/filepath"
 	"testing"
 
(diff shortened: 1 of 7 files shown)


processor/error/event_test.go
```diff
@@ -396,3 +399,43 @@ func md5With(args ...string) []byte {
 	}
 	return md5.Sum(nil)
 }
+
+func TestSourcemapping(t *testing.T) {
```
graphaelli: great test

PR #486, opened 2018-01-12
[6.x] Use Stacktrace Frame pointers. (#485)

simitt:
 - Use Stacktrace Frame pointers.  (#485)


simitt:
@roncohen , @graphaelli I merge this PR, as the original commit has been approved and this is a bugfix that should get backported to 6.x.

PR #489, opened 2018-01-13
Fix default example for frontend.allow_origins

simitt:
The config define `frontend.allow_origins` to be an `[]string`.


simitt:
closed in favor of a bigger refactoring #492

PR #493, opened 2018-01-15
Add type to description for `context.tags`, `transaction.marks`.

simitt:


docs/spec/transactions/mark.json
```diff
@@ -2,7 +2,7 @@
     "$schema": "http://json-schema.org/draft-04/schema#",
     "$id": "docs/spec/transactions/mark.json",
     "type": ["object", "null"],
-    "description": "A mark captures the timing of a significant event during the lifetime of a transaction. Every mark is a simple key value pair and can be set by the user or the agent.",
+    "description": "A mark captures the timing of a significant event during the lifetime of a transaction. Every mark is a simple key value pair, where the value has to be a number, and can be set by the user or the agent.",
```
beniwohli: might be interesting to say what the physical value of the number is. Milliseconds, microseconds?

simitt: mh, but that's up to the agent that records the values, as they are user defined input we cannot specify that really.

beniwohli: they are on the same timeline as the transaction itself, right? So I'd assume that they are represented in the same way as transaction timings

simitt: added

Issue #413, opened 2017-12-19
Remove references to untested configurations


graphaelli:
Because APM inherits so much infrastructure from beats, there are a number of unsupported/undocumented configuration options exposed in [apm-server.reference.yml](https://github.com/elastic/apm-server/blob/75c0e67ead2e3d8958b675ed4502eceb62b8c93c/apm-server.reference.yml).  They should be removed for now and re-added as development supports them.

Any documentation should be removed accordingly.

jalvz:
close this?

PR #496, opened 2018-01-15
Remove untested options from config inherited from beats.

simitt:
This is meant to be a temporary fix for #413 to remove config options inherited from libbeat that might not make sense for apm or are completely untested. 

@roncohen and @graphaelli please review if you think the config options that I suggest to use make sense. I haven't tested all of them for this PR, but I think those are the ones that we should have in.


_beats/libbeat/scripts/Makefile
```diff
@@ -269,12 +269,9 @@ update: python-env collect
 	@echo "Updating generated files for ${BEAT_NAME}"
 
 	@# Update config files.
-	@cat _meta/beat.yml ${ES_BEATS}/libbeat/_meta/config.yml | sed -e "s/beatname/${BEAT_NAME}/g;s/beat-index-prefix/${BEAT_INDEX_PREFIX}/g" > ${BEAT_NAME}.yml
-	@if [ -e _meta/beat.reference.yml ]; then \
-        cat _meta/beat.reference.yml ${ES_BEATS}/libbeat/_meta/config.reference.yml | sed -e "s/beatname/${BEAT_NAME}/g;s/beat-index-prefix/${BEAT_INDEX_PREFIX}/g" > ${BEAT_NAME}.reference.yml ; \
-    else \
-        cat _meta/beat.yml ${ES_BEATS}/libbeat/_meta/config.reference.yml | sed -e "s/beatname/${BEAT_NAME}/g;s/beat-index-prefix/${BEAT_INDEX_PREFIX}/g" > ${BEAT_NAME}.reference.yml ; \
-    fi;
+	@cp _meta/beat.yml ${BEAT_NAME}.yml
```
graphaelli: maybe it's simpler to remove these few lines altogether and leave the files in `_meta` as is?

simitt: I think we then would need to ensure that the server by default finds the file from `_meta` at startup if no config option is provided. I'd rather not mess with this, as copying the files over is pretty straightforward.

simitt:
jenkins, retest this please.

PR #502, opened 2018-01-15
Change logic for setting function name when applying smap and add test for minified source mapping file. 

simitt:
Add testfiles for sourcemapping. @jahtalab please verify the output `processor/error/package_tests/TestProcessErrorFrontendMinifiedSmap.approved.json`, which is created by applying the source map to the error `tests/data/valid/error/frontend_app.e2e-bundle.min.json`

Fix `function` with the following algorithm:
- set previous function name after applying sourcemap
- set <anonymous>  for the function of the first stack trace frame
- set <unknown> if the sourcemapping for the previous stack trace frame did not return a function name. 
- if single frames in between are not valid for source mapping (no mapping found etc) preserve the function name from the last successfully applied frame for the next one.


CHANGELOG.asciidoc
```diff
@@ -20,6 +20,7 @@ https://github.com/elastic/apm-server/compare/71df0d96445df35afe27f38bcf734a0828
 - Updated systemd doc url {pull}354[354]
 - Updated readme doc urls {pull}356[356]
 - Use updated stack trace frame values for calculating error `grouping_keys` {pull}485[485] 
+- Change logic for setting function name after applying sourcemap {pull}502[502]
```
roncohen: Please add the PR number to the line describing the source map feature, like so:

`- Support for sourcemap mapping on incoming frontend requests {pull}381[381] {pull}502[502]`

instead of the line added here.

This is how i think about the changelog in general: The changelog should explain the difference between the versions. If a bug was not released in a version, we should not have a bug-fix for it in the changelog. I'll bring it up on the next APM Server meeting to discuss.

simitt: makes sense, will change!

model/stacktrace.go
```diff
@@ -7,12 +7,18 @@ import (
...
+func (st Stacktrace) Transform(config *pr.Config, service Service) []common.MapStr {
+	var fr *StacktraceFrame
 	var frames []common.MapStr
+	noFrames := len(st)
+	frames = make([]common.MapStr, noFrames)
 
-	for _, fr := range *st {
-		frame := fr.Transform(config, service)
-		frames = append(frames, frame)
+	fct := "<anonymous>"
```
roncohen: would be great to have the algorithm described here (or elsewhere) in comments, like you did in the PR description.

simitt: Sure I can add a description. It would have been easier to follow if we applied it after the original sourcemapping was applied, but this would mean to walk through the whole stacktrace a second time.

model/stacktrace.go
```diff
@@ -7,12 +7,50 @@ import (
...
+//   * should be moved down one stack trace frame,
+//   * the function name of the first frame is set to <anonymous>
+//   * if one frame is not found in the source map, this frame is left out and
+//   the function name from the previous frame is set for the one but next frame
+// - colno
+// - lineno
+// - abs_path is set to the cleaned abs_path
+// - sourcmeap.updated is set to true
+func (st Stacktrace) sourcemapAndTransform(config *pr.Config, service Service) []common.MapStr {
+	noFrames := len(st)
```
graphaelli: nitpick but this sounds like a boolean, frameCount? numFrames?

model/stacktrace.go
```diff
@@ -7,12 +7,50 @@ import (
...
+func (st Stacktrace) transform(config *pr.Config) []common.MapStr {
+	var frames []common.MapStr
+	for _, fr := range st {
+		frame := fr.Transform(config)
 		frames = append(frames, frame)
 	}
 	return frames
 }
+
+// source map algorithm:
```
graphaelli: love these comments

model/stacktrace_frame.go
```diff
@@ -24,23 +24,19 @@ type StacktraceFrame struct {
...
+	Sourcemap Sourcemap
+}
+
+type Sourcemap struct {
+	Updated *bool
+	Error   *string
 }
 
-func (s *StacktraceFrame) Transform(config *pr.Config, service Service) common.MapStr {
+func (s *StacktraceFrame) Transform(config *pr.Config) common.MapStr {
```
graphaelli: thinking of this as an api, what should happen if Transform is called more than once?

simitt: Really good question. As the decision was made to overwrite the original values and not to create a new stacktrace frame out of it, running `Transform` multiple times would lead to a different output.

model/stacktrace.go
```diff
@@ -7,12 +7,52 @@ import (
...
+// apply source mapping frame by frame
+// if no source map could be found, set updated to false and set sourcemap error
+// otherwise use source map library for mapping and update
+// - filename: only if it was found
+// - function:
+//   * should be moved down one stack trace frame,
+//   * the function name of the first frame is set to <anonymous>
+//   * if one frame is not found in the source map, this frame is left out and
+//   the function name from the previous frame is set for the one but next frame
+//   * if a mapping could be applied but nor function name is found, the
```
graphaelli: `s/nor/no/`

model/stacktrace.go
```diff
@@ -7,12 +7,52 @@ import (
...
+// source map algorithm:
+// apply source mapping frame by frame
+// if no source map could be found, set updated to false and set sourcemap error
+// otherwise use source map library for mapping and update
+// - filename: only if it was found
+// - function:
+//   * should be moved down one stack trace frame,
+//   * the function name of the first frame is set to <anonymous>
+//   * if one frame is not found in the source map, this frame is left out and
+//   the function name from the previous frame is set for the one but next frame
```
roncohen: > previous frame is set for the one but next frame

not sure what was meant here

model/stacktrace.go
```diff
@@ -7,12 +7,52 @@ import (
...
 }
+
+// source map algorithm:
+// apply source mapping frame by frame
+// if no source map could be found, set updated to false and set sourcemap error
+// otherwise use source map library for mapping and update
+// - filename: only if it was found
+// - function:
+//   * should be moved down one stack trace frame,
+//   * the function name of the first frame is set to <anonymous>
```
roncohen: it's only set to `<anonymous>` if there's a problem, AFAICT.

model/stacktrace.go
graphaelli: @roncohen I've pushed a commit to hopefully clear this up.  Essentially `fct` stays the same if the mapping can not be applied and is used as the function name for the next frame.

graphaelli:
jenkins, retest this please

model/stacktrace.go
simitt: Not exactly, `<anonymous>` is set for the first frame. As the logic is to always use the sourcemapped function name of the previous frame, the first frame doesn't have a predecessor so it gets an <anonymous>.
For all the others, if no function name can be found for a frame,  `<unknown>` is set as function name for the next frame.

model/stacktrace_frame.go
```diff
@@ -80,32 +81,36 @@ func (s *StacktraceFrame) isLibraryFrame(pattern *regexp.Regexp) bool {
 		(s.AbsPath != nil && pattern.MatchString(*s.AbsPath))
 }
 
-func (s *StacktraceFrame) applySourcemap(service Service, mapper sourcemap.Mapper) {
+func (s *StacktraceFrame) applySourcemap(mapper sourcemap.Mapper, service Service, prevFunction string) string {
```
simitt: @roncohen fyi - this method gets a `prevFunction` string passed in.
If source mapping cannot be applied, the function returns the unchanged `prevFunction` string.
If source mapping is successful, the `prevFunction` is either changed to the new function name or if this is empty to `<unknown>`.

PR #510, opened 2018-01-16
Change logic for setting function name when applying smap and add test for minified source mapping file. (#502)

graphaelli:

* Change logic for setting function name when applying smap and add test for minified source mapping file.  (#502)


graphaelli:
jenkins, retest this please

PR #512, opened 2018-01-17
[DNM] Switch vendoring tool

simitt:
This is meant as a POC for using [dep](https://github.com/golang/dep) instead of [govendor](govendor not a go package or not in gopath) as we had issues with using govendor after updating to go 1.9


jalvz:
why this has been closed?

simitt:
I created the PR accidentially, it was only meant to play around a bit with `dep` and figure out if it fulfills all our requirements.

PR #520, opened 2018-01-18
Change error.culprit after successful source mapping.

### LINKED ISSUE #455: Update culprit for sourcemapping
Update `error.culprit` with `filename` and `function` of stacktrace frame causing error after applying sourcemapping.
Take the information from the latest stack trace frame.

simitt:
closes #455

### DIFF
--- a/model/stacktrace_frame.go
+++ b/model/stacktrace_frame.go
@@ -42,14 +42,13 @@ func (s *StacktraceFrame) Transform(config *pr.Config) common.MapStr {
 	enhancer.Add(m, "module", s.Module)
 	enhancer.Add(m, "function", s.Function)
 	enhancer.Add(m, "vars", s.Vars)
-	if config != nil && config.LibraryPattern != nil && s.LibraryFrame == nil {
-		libraryFrame := s.isLibraryFrame(config.LibraryPattern)
-		s.LibraryFrame = &libraryFrame
+	if config != nil && config.LibraryPattern != nil {
+		s.setLibraryFrame(config.LibraryPattern)
 	}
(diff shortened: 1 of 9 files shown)


simitt:
ping @elastic/apm-server for review

model/stacktrace_frame.go
```diff
@@ -42,14 +42,13 @@ func (s *StacktraceFrame) Transform(config *pr.Config) common.MapStr {
 	enhancer.Add(m, "module", s.Module)
 	enhancer.Add(m, "function", s.Function)
 	enhancer.Add(m, "vars", s.Vars)
-	if config != nil && config.LibraryPattern != nil && s.LibraryFrame == nil {
-		libraryFrame := s.isLibraryFrame(config.LibraryPattern)
-		s.LibraryFrame = &libraryFrame
+	if config != nil && config.LibraryPattern != nil {
+		s.setLibraryFrame(config.LibraryPattern)
```
jalvz: i'm lost with the latest developments, but why is necessary to mutate the argument here?
`Transform` shouldn't just produce a `common.MapStr`?

simitt: We are using the `Stacktrace frame` information in the `error` doc for calculating the `grouping key`, and there we don't use the `mapStr` that came out of `transform`, but the actual values of the struct. 
As we should use the values after source mapping (in case it was applied) those values need to be changed.

@graphaelli already raised the question whether the `transform` action is idempotent, which is not the case atm. I created an issue now to follow up on this https://github.com/elastic/apm-server/issues/566

PR #522, opened 2018-01-19
Minor fixes to changelog

roncohen:
Removed duplicate: 
* Support for `transaction.marks` {pull}430[430]

And added whitespace to "Increase default 'ConcurrentRequests .." to fix rendering.


CHANGELOG.asciidoc
```diff
@@ -34,11 +34,10 @@ https://github.com/elastic/apm-server/compare/71df0d96445df35afe27f38bcf734a0828
 - Support for `transaction.marks` {pull}430[430]
 - Support for uploading sourcemaps {pull}302[302].
 - Support for sourcemap mapping on incoming frontend requests {pull}381[381], {pull}502[502]
-- Support for `transaction.marks` {pull}430[430].
```
graphaelli: shouldn't this be the one that stays since it's newer?

roncohen: maybe :) what's the difference?

roncohen: updated now

PR #449, opened 2018-01-04
update readme doc urls

graphaelli:
This moves control of `_beats/dev-tools/packer/readme.md.j2` into the apm repo rather than trying to templatize it under beats.  README ends up as:

```
# Welcome to apm-server 7.0.0-alpha1

Sends events to Elasticsearch or Logstash

## Getting Started

To get started with apm-server, you need to set up Elasticsearch on your localhost first. After that, start apm-server with:

     ./apm-server -c apm-server.yml -e

This will start apm-server and send the data to your Elasticsearch instance. To load the dashboards for apm-server into Kibana, run:

    ./apm-server setup -e

For further steps visit the [Getting started](https://www.elastic.co/guide/en/apm/get-started/master/) guide.

## Documentation

Visit [Elastic.co Docs](https://www.elastic.co/guide/en/apm/server/master/) for the full apm-server documentation.
```
 
closes #356

PR #523, opened 2018-01-19
Fix update script for _beats. Update beats framework.

simitt:
(1) The current [update_beats script](https://github.com/elastic/apm-server/blob/master/script/update_beats.sh) doesn't seem to work properly.
Steps to reproduce:
```
rm -rf _beats/*
make update-beats
```
The diff then shows that expected files under `_beats` are not re-added.
[Changing the `include`](https://github.com/simitt/apm-server/blob/58b4cea3713f53f132e326c8e8ad29017913961b/script/update_beats.sh) statements in the script leads to expected behavior. 

(2) `make update-beats` updates to the latest beats framework. 

Point (1) needs to be backported to `6.x` to allow properly updating to recent beats changes.


simitt:
jenkins, retest this please.

simitt:
jenkins, retest this please

_beats/dev-tools/packer/readme.md.j2
```diff
@@ -1,19 +0,0 @@
-# Welcome to {{.beat_name}} {{.version}}
```
graphaelli: missed this on first review.  Removing this will undo the readme fixes from #449

_beats/libbeat/scripts/Makefile
```diff
@@ -132,7 +132,8 @@ clean:: ## @build  Cleans up all files generated by the build steps
 	@rm -f docker-compose.yml.lock
 	@rm -f ${BEAT_NAME} ${BEAT_NAME}.test ${BEAT_NAME}.exe ${BEAT_NAME}.test.exe
 	@rm -f _meta/fields.generated.yml fields.yml
```
graphaelli: what do you think about restoring the `rm -rf _beats`?

graphaelli: sorry, meant this for the other makefile

graphaelli: cancel that, that will make it more complicated to keep files we're overriding, like the readme

script/update_beats.sh
```diff
@@ -22,19 +22,27 @@ git clone https://github.com/elastic/beats.git ${GIT_CLONE}
...
-    --include="script/***" \
     --include="dev-tools/***" \
-    --include="libbeat/scripts/***" \
-    --include="libbeat/_meta/***" \
+    --include="script/***" \
+    --include="testing/***" \
+    --include="libbeat/" \
     --include=libbeat/Makefile \
+    --include="libbeat/_meta/***" \
+    --exclude="libbeat/_meta/fields.generated.yml" \
```
ruflin: @simitt Haven't tested it but was wondering if this could cause issues as I think the generated fields.yml from libbeat is used to generate the full fields.yml for apm-server.

simitt: Thanks for pointing that out @ruflin I'll look into it and test it again. 
This file was never added to apm-server though - not in the original version before PR #449 , and not before this PR. There was an issue with the current `rsync` cmd and after fixing it, this file needed to be specifically excluded to reach the same state as before the change.

ruflin: @simitt Great, thanks for checking.

PR #529, opened 2018-01-19
[6.2] Fix update script for _beats. Update beats framework.

graphaelli:
Backports the fix from #523 + make update-beats + post fix from #496


_beats/libbeat/docs/version.asciidoc
```diff
@@ -5,3 +5,4 @@
 :python: 2.7.9
 :docker: 1.12
 :docker-compose: 1.11
+:branch: 6.x
```
graphaelli: this surprises me a bit, was expecting 6.2, but that's what is on the beats 6.2 branch at this commit - https://github.com/elastic/beats/blob/a270b60dec662a9ed1c0302e949c65d51af7f772/libbeat/docs/version.asciidoc

PR #541, opened 2018-01-24
make autopep8 happy

graphaelli:
clear up new `make check` issue - https://beats-ci.elastic.co/job/elastic+apm-server+master+beats-update/274/console

### DIFF
--- a/script/payload-compare.py
+++ b/script/payload-compare.py
@@ -1,8 +1,8 @@
-import random
-import json
-import zlib
 from uuid import uuid4
 from collections import defaultdict
+import json
+import random
+import zlib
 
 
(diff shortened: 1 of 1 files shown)


graphaelli:
Backports incoming to clear up build failures against 6.x branches.

@simitt any idea what caused this to suddenly trigger?

simitt:
I'd assume it being triggered by using the latests released `autopep8` version `1.3.4`, released on 24/01/2018.

graphaelli:
aha, now I see that the version isn't pinned.  I'll propose an update to pin those deps, in the beats repo, to avoid nasty surprises.

PR #546, opened 2018-01-25
x-pack security docs

### LINKED ISSUE #545: document x-pack security configuration
Document apm-server required cluster and index permissions.  Something similar to the [beats docs](https://www.elastic.co/guide/en/beats/auditbeat/6.0/beats-user-access.html).

graphaelli:
* add X-Pack configuration documentation
* expand on TLS configuration

closes #545

### DIFF
--- a/docs/copied-from-beats/security/basic-auth.asciidoc
+++ b/docs/copied-from-beats/security/basic-auth.asciidoc
@@ -1,0 +1,95 @@
+[role="xpack"]
+[[beats-basic-auth]]
+=== Configuring Authentication Credentials for {beatname_uc}
+
+When sending data to a secured cluster through the `elasticsearch`
+output, {beatname_uc} must either provide basic authentication credentials
+or present a client certificate.
+
+To configure authentication credentials for {beatname_uc}:
+
+. Create a role that has the `manage_index_templates` and
+`monitor` cluster privileges, and `read`, `write`, and `create_index`
+privileges for the indices that {beatname_uc} creates. You can create roles from the
+**Management / Roles** UI in {kib} or through the `role` API.
+For example, the following request creates a ++{beat_default_index_prefix}_writer++ role:
++
+["source","sh",subs="attributes,callouts"]
+---------------------------------------------------------------
+POST _xpack/security/role/{beat_default_index_prefix}_writer
+{
+  "cluster": ["manage_index_templates", "monitor"],
+  "indices": [
+    {
+      "names": [ "{beat_default_index_prefix}-*" ], <1>
+      "privileges": ["write","create_index"]
+    }
+  ]
+}
+---------------------------------------------------------------
+<1> If you use a custom {beatname_uc} index pattern, specify that pattern
+instead of the default ++{beat_default_index_prefix}-*++ pattern.
+
+. Assign the writer role to the user that {beatname_uc} will use to connect to
+{es}:
+
+.. To authenticate as a native user, create a user for the {beatname_uc} to use
+internally and assign it the writer role. You can create users from the
+**Management / Users** UI in {kib} or through the `user` API. For example, the
+following request creates a ++{beat_default_index_prefix}_internal++ user that has the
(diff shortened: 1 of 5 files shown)


roncohen:
Looks like this accidentally removes the part that describes how to get APM Server to serve SSL

simitt:
@roncohen afaics it exchanges our specific explanation about SSL with the one from beats, https://github.com/elastic/apm-server/pull/546/files#diff-6ff90a0c0297bd657a48b2249be6e1e8R23

graphaelli:
you guys are too fast :)  I was planning on testing the ssl changes but if you know the right parameters right off, please feel free to push updates.

graphaelli:
Excellent, thanks.  This could also use elastic/beats#6184 - ok to give that a little time to get in before redoing this?

graphaelli:
copied changes from elastic/beats#6184 over and re-added server tls configuration section.

docs/copied-from-beats/security/basic-auth.asciidoc
```diff
@@ -0,0 +1,95 @@
...
+
+. Create a role that has the `manage_index_templates` and
+`monitor` cluster privileges, and `read`, `write`, and `create_index`
+privileges for the indices that {beatname_uc} creates. You can create roles from the
+**Management / Roles** UI in {kib} or through the `role` API.
+For example, the following request creates a ++{beat_default_index_prefix}_writer++ role:
++
+["source","sh",subs="attributes,callouts"]
+---------------------------------------------------------------
+POST _xpack/security/role/{beat_default_index_prefix}_writer
```
jalvz: Since this is APM Server documentation, might worth be explicit about that this call is made to ES. There is nothing wrong neither with writing a full example that users can copy & paste like
`curl -u elastic:changeme -H "Content-Type: application/json" -X POST http://localhost:9200/_xpack/security/role/...`

docs/copied-from-beats/security/basic-auth.asciidoc
```diff
@@ -0,0 +1,95 @@
...
+
+.. To use PKI authentication, configure the `certificate` and
+`key` settings:
++
+["source","js",subs="attributes,callouts"]
+--------------------------------------------------
+output.elasticsearch:
+    hosts: ["localhost:9200"]
+    index: "{beat_default_index_prefix}"
+    ssl.certificate: "/etc/pki/client/cert.pem" <1>
```
jalvz: to avoid confusion for people not familiar with ssl, this example could also have the `username` and `password` attributes as well to make clear that ssl attributes do not replace username/password (which looks a bit like that if you compare this snippet with the previous one)

docs/security.asciidoc
```diff
@@ -7,6 +7,16 @@ We recommend using firewall rules to ensure only authorized systems can connect.
...
 
+[[secret-token]]
+[float]
+==== Secret token
+
+You can configure a secret token which is sent with every request from the APM agents to the server.
+This string is used to ensure that only your agents can send data to your APM servers.
+Both the agents and the APM servers have to be configured with the same secret token.
+
+NOTE: The usage of a secret token only provides any security when used in combination with having SSL/TLS configured.
```
jalvz: I would also mention that secret token is only for backend endpoints and sourcemap uploads, as frontend requests can not be authenticated.

docs/copied-from-beats/security/basic-auth.asciidoc
graphaelli: great point, I'll think about how to refactor the docs so we can put that in.  Maybe just a header before we include this page.

docs/security.asciidoc
graphaelli: good call, I'll follow up with that separately.

graphaelli:
Right, apm-server will not run with the `apm_reader` role.  We'll need to figure out if we need to clarify that `apm_writer` is for the server and `apm_reader` is for the kibana user consuming the data - I'm not sure if that's confusing right now but we'll see soon enough.

PR #551, opened 2018-01-26
Add troubleshooting section to docs.

simitt:


docs/copied-from-beats/getting-help.asciidoc
```diff
@@ -0,0 +1,26 @@
...
+//// descriptions here generic enough to work for all Beats that include
+//// this file. When using cross references, make sure that the cross
+//// references resolve correctly for any files that include this one.
+//// Use the appropriate variables defined in the index.asciidoc file to
+//// resolve Beat names: beatname_uc and beatname_lc.
+//// Use the following include to pull this content into a doc file:
+//// include::../../libbeat/docs/getting-help.asciidoc[]
+//////////////////////////////////////////////////////////////////////////
+
+Start by searching the https://discuss.elastic.co/c/apm[discussion forum] for your issue.
```
roncohen: Ah, i missed this. `copied-from-beats` should contain only verbatim copies - and it does so far. We need to either:
* make a separate copy and keep it outside `copied-from-beats`
* or make this parameterizable so we can change it in our docs and contribute the parameterized version back to beats so we can copy it to `copied-from-beats`.

Issue #425, opened 2017-12-21
Making `transaction.name` optional?


simitt:
It recently came up that it might not always be possible to collect the `transaction.name`. 

Right now this is a required field on the server. This leads to agents setting default values if they cannot detect a name. 

Following proposals came up:
1. Keep current logic that agents set default names.
1. Keep current logic and additionally set a boolean flag indicating if the `transaction.name` was retrieved or a default value was set.
1. Making `transaction.name` *optional* on the server side and not sending the key up to the server if it couldn't be set.

@elastic/apm-agent-devs what do you think?

watson:
For reference, this is what we currently do in the Node.js agent: A transaction is given the name `unnamed` if no other name have been provided provided, or `GET unknown route (unnamed)` in case an HTTP request is associated with the transaction but no other name have been provided.

watson:
Speaking purely from an agent perspective (i.e. not considering the implications for ES or Kibana), I think option 2 or 3 are the best. There's a slight advantage in nr 2 as we in the Node.js agent can come up with a slightly better default name in case an HTTP request is associated with the transaction. But as I see it this is pretty unlikely with our current setup. But honestly I'm personally fine with all 3 options - unless someone can point out an issue with one or more of the solutions that I didn't think of.

roncohen:
It doesn't seem like we'll have time to do the UI work for this, so let's postpone it to 6.3 or later for now.

mikker:
I like the way it is. I get that it’s weird to have to make up a “no name” name as default but I think that’s the best option of the three suggested.

If I have to pick between 2 and 3? Then 3.

simitt:
After having an offline discussion for this we will implement version (3) to make the `transaction.name` optional.

PR #554, opened 2018-01-26
Make transaction.name optional.

simitt:
Remove attribute from minimal testfiles.
implements #425 

@elastic/apm-agent-devs fyi


watson:
@simitt which Elastic version will this be included in?

simitt:
@watson `6.3`.

PR #555, opened 2018-01-26
Update manual testing guide and docker-compose.

simitt:


TESTING.md
```diff
@@ -28,15 +28,17 @@ Following workflow is intended:
...
-* Open `localhost:5601` in your browser and log into Kibana with `elastic` and `changeme`
-* Create the `apm-server-*` index pattern
-* In Kibana go to the Discovery tab and you should see data
-
-For manual testing with specific agents, check instructions at `tests/agent/[LANG]/README.md`
-
+For manual testing with the elastic stack there is a test environment based on docker. 
+To get this running execute the following commands.
+
+* Run `make start-env` to start docker containers for the elastic stack
```
jalvz: This command is outdated, right?

jalvz: ah wait, I see you changed the docker-compose bellow... let me check it out then 👍

tests/docker-compose.yml
```diff
@@ -1,33 +1,58 @@
...
-    build: .
-    ports:
-      - "8200:8200"
-    command: ./apm-server -E output.elasticsearch.hosts=[http://elastic:changeme@elasticsearch:9200] -E apm-server.host=0.0.0.0:8200 -e
-
   elasticsearch:
-    extends:
-      file: ../_beats/testing/environments/snapshot.yml
-      service: elasticsearch
+    image: docker.elastic.co/elasticsearch/elasticsearch-platinum:${STACK_VERSION:-6.1.2}
```
jalvz: minor thing: I'd remove the default and just point out in the docs to set the version explicitly like `STACK_VERSION=version make start-dev`, that way we don't keep an old default when new versions roll out and we don't have to remember to update it.

PR #557, opened 2018-01-26
Add `process.ppid` as optional field.

### LINKED ISSUE #547: Add "parent process id" field
@watson brought up that node recently added [`process.ppid`](https://nodejs.org/api/process.html#process_process_ppid), which returns the parent process ID. Python exposes this information as well ([`os.getppid()`](https://docs.python.org/3/library/os.html#os.getppid)), and so does Ruby ([`Process.ppid`](http://ruby-doc.org/core-2.2.0/Process.html#method-c-ppid)).

This field could be especially interesting in scenarios where there is a single master process, and several worker processes. This is a fairly common way to deploy web apps.

simitt:
closes #547 

@elastic/apm-agent-devs fyi

### DIFF
--- a/_meta/fields.common.yml
+++ b/_meta/fields.common.yml
@@ -146,11 +146,15 @@
           description: >
             Information pertaining to the running process where the data was collected
           fields:
-
             - name: pid
               type: long
               description: >
                 Numeric process ID of the service process.
+
+            - name: ppid
+              type: long
+              description: >
+                Numeric ID of the service's parent process.
 
             - name: title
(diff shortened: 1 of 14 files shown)


beniwohli:
that was quick :D

watson:
What happens if the agent sends up `ppid` to a 6.2 server? Will it just ignore it, or will the request fail? I'm just trying to asses if this is a breaking feature in the agents, or if we can just add it at any time

simitt:
It should be ignored.

PR #563, opened 2018-01-29
[6.x] Add `process.ppid` as optional field. (#557)

simitt:
 - Add `process.ppid` as optional field.  (#557)


watson:
Awesome!

Btw, I think it would be a good idea if we ping @elastic/apm-agent-devs on PR's that change the intake API so that they the respective agents remember to implement it 😃

watson:
Oh sorry. Nvm - I was too quick there. I can see that you did so in the original PR

PR #571, opened 2018-01-29
[6.x] Update testify and dependencies

simitt:


simitt:
close this until we can update to latest beats

PR #572, opened 2018-01-29
[6.2] Update testify and dependencies.

simitt:


simitt:
close until we can update to beats

PR #578, opened 2018-01-30
Prevent automatically inheriting configs from beats.

simitt:
Config files are not copied over from beats any longer.
The control over the config files now is within apm-server,
to enable different default settings and more fine granular control
over enabled or suggested config options.

As I think we should keep updating libbeat's `Makefile` when running `update-beats` I didn't prevent this from being synced, but rather set the libbeat config files to be empty files and not being synced. This way the cmd in the Makefile doesn't need to be changed. 
However, with this change it is our responsibility to keep up-to-date with added config options in beats and manually sync them.

fixes https://github.com/elastic/apm-server/issues/625


jalvz:
As with this change we are down to a more human-friendly number of config options, we could get rid of the `reference.yml` and maintain only 1 config file.

With the reference around it becomes quite arbitrary to decide which options "deserve" to be included in`beat.yml` commented out and which not (eg you added now the `ssl.*` ones which I believe they were added and removed back and forth).

graphaelli:
I don't feel strongly about it, but if i had to chose I'd prefer the single file with all config options and remove the reference file.

simitt:
I updated the PR to have all options in the `apm-server.yml`. 
When running `make update` the `reference` file is still created, as this is done within the libbeat Makefile cmd, but it is not committed anymore. 
We need to do some changes on the beats side to completely get rid of the `apm-server.reference.yml` file.

apm-server.yml
```diff
@@ -1,54 +1,300 @@
-################### APM Server Configuration Example #########################
+######################## APM Server Configuration #############################
```
ruflin: In the Beats case we call it example because we already enable some modules by default and also change some of the default configs inside for a better getting started experience like overwriting number of shards / replicas. For APM perhaps the reference file could even be removed?

apm-server.yml
```diff
@@ -61,39 +307,28 @@ setup.kibana:
   # IPv6 addresses should always be defined as: https://[2001:db8::1]:5601
   #host: "localhost:5601"
 
-#============================= Elastic Cloud ==================================
```
ruflin: I would think this is something that is also usable for APM?

simitt:
Thank you for the feedback @ruflin. I did not yet check the packaging, and have to filter for any other places where the `reference` yml is used. 
The general idea is to manually sync up with config options inherited from beats, to only list those that we actively test and support. 
Some use cases are different in apm, e.g. when using the (experimental) `source map` feature, you need to upload source maps to an elasticsearch instance, so it is not compatible with using a different output atm.

jalvz:
can this be merged or do you want to do something else here?

simitt:
please leave it open, I'll come back to it and have another look.

jalvz:
We are inheriting a lot of docs from beats. If we own & control the configuration, does make sense to keep using beats docs?

Said in other way, don't we risk to get docs & code out of sync? (eg. if we restrict some config option on our side but we get docs from beats, that won't be reflected).

We can of course be vigilant about that, but we could be putting ourselves in another kind of mess in which we lose keeping track of what docs are / should / aren't / shouldn't be inherited...

simitt:
That's a good point @jalvz . However, we don't automatically inherit docs from beats, but copy them manually. I'd see it the responsibility of the committer & reviewer that update the docs to keep the docs in sync with the config files. But ofc that's an error-prone process.

What do you suggest instead? I think as long as we haven't tested the configs we shouldn't describe them, neither in the config nor in the docs.

jalvz:
So, recap: we keep copying docs from beats, being careful of not copying too much.

simitt:
After checking what needs to be changed to get rid of the `reference.yml` I suggest following steps:
* keep `reference.yml` in a first step and just ensure it isn't copied over from beats, so we can move on, e.g. with separating indices etc.
* update handling in beats packaging to also work without `reference.yml` (already in progress)
* merge from beats and update in apm-server to get rid of `reference.yml`.

If not splitting this up I think it gets a mess, so I will revert this PR again to only get rid of the copying part for now.

_meta/beat.reference.yml
```diff
@@ -72,8 +71,351 @@ apm-server:
 
   # golang expvar support - https://golang.org/pkg/expvar/
   #expvar:
-  # Set to true to Expose expvar
-  #enabled: false
+    # Set to true to Expose expvar
+    #enabled: false
+
+    # Url to expose expvar
+    #url: "/debug/vars"
```
simitt: shifted so it is properly positioned within the `expvar` namespace.

simitt:
@jalvz could you please review again with a close look on what is changed in the `apm-server.yml` and the `apm-server.reference.yml` as I have slightly updated descriptions etc.

Sorry for all the noise.

_meta/beat.reference.yml
```diff
@@ -72,8 +71,351 @@ apm-server:
...
+
+  # Dictionary of HTTP parameters to pass within the url with index operations.
+  #parameters:
+    #param1: value1
+    #param2: value2
+
+  # Number of workers per Elasticsearch host.
+  #worker: 1
+
+  # Optional index name. The default is "apm" plus date
```
jalvz: the default is apm-VERSION-DATE, right?

jalvz:
Why the diff for beat.reference.yml is different from apm-server.reference.yml? isn't copied over?

simitt:
Until now `apm-server.reference.yml` was created by merging `_meta/beat.reference.yml` and `_beats/libbeat/_meta/config.reference.yml`. This PR removes the config from `beats` and only uses what is defined in `_meta/beat.reference.yml` so the diff cannot be the same.

jalvz:
right, of course 🤦‍♂️

PR #582, opened 2018-01-30
Enhancement separate smap index

simitt:
This PR is based on PR #578.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -45,6 +45,7 @@ https://github.com/elastic/apm-server/compare/71df0d96445df35afe27f38bcf734a0828
 - Change `error.culprit` after successfully applying sourcemapping {pull}520[520]
 - Make `transaction.name` optional {pull}554[554]
 - Remove config files from beats. Manually add relevant config options {pull}578[578]
+- Use seperate index for uploaded `source maps` {pull}582[582].
 
 ==== Deprecated
 
--- a/beater/config.go
+++ b/beater/config.go
@@ -1,6 +1,7 @@
 package beater
 
 import (
+	"regexp"
 	"time"
 
 	"github.com/elastic/apm-server/sourcemap"
@@ -33,14 +34,17 @@ type FrontendConfig struct {
 	LibraryPattern      string         `config:"library_pattern"`
 	ExcludeFromGrouping string         `config:"exclude_from_grouping"`
 	SourceMapping       *SourceMapping `config:"source_mapping"`
+
+	beatVersion string
 }
 
 type SourceMapping struct {
-	Cache *Cache `config:"cache"`
-	Index string `config:"index_pattern"`
+	Cache        *Cache `config:"cache"`
+	IndexPattern string `config:"index_pattern"`
 
 	esConfig *common.Config
 	mapper   sourcemap.Mapper
+	index    string
 }
 
 type Cache struct {
@@ -83,10 +87,12 @@ func (c *FrontendConfig) SmapMapper() (sourcemap.Mapper, error) {
 	if smap.mapper != nil {
 		return c.SourceMapping.mapper, nil
 	}
+	re := regexp.MustCompile("%.*{.*beat.version.?}")
+	c.SourceMapping.index = re.ReplaceAllLiteralString(smap.IndexPattern, c.beatVersion)
 	smapConfig := sourcemap.Config{
 		CacheExpiration:     smap.Cache.Expiration,
 		ElasticsearchConfig: smap.esConfig,
-		Index:               smap.Index,
+		Index:               c.SourceMapping.index,
 	}
 	smapMapper, err := sourcemap.NewSmapMapper(smapConfig)
 	if err != nil {
@@ -96,7 +102,7 @@ func (c *FrontendConfig) SmapMapper() (sourcemap.Mapper, error) {
 	return c.SourceMapping.mapper, nil
 }
 
-func defaultConfig() *Config {
+func defaultConfig(beatVersion string) *Config {
 	return &Config{
 		Host:               "localhost:8200",
 		MaxUnzippedSize:    50 * 1024 * 1024, // 50mb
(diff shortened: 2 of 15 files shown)


CHANGELOG.asciidoc
```diff
@@ -43,6 +44,7 @@ https://github.com/elastic/apm-server/compare/71df0d96445df35afe27f38bcf734a0828
 - Add `process.ppid` as optional field {pull}564[564]
 - Change `error.culprit` after successfully applying sourcemapping {pull}520[520]
 - Make `transaction.name` optional {pull}554[554]
+- Change generationg of config files and supported options {pull}578[578]
```
graphaelli: typo

graphaelli:
Is this really a breaking change?  Defaults in 6.2 would put sourcemaps in `apm-*`, this puts sourcemaps in `apm-sourcemap*`.  If a user used this and then downgraded back to 6.2, apm-server would continue to look for sourcemaps in `apm-*` and would find them, even if they were in `apm-sourcemap*`.  Maybe moot since this feature is beta.

sourcemap/elasticsearch.go
```diff
@@ -82,9 +82,9 @@ func query(id Id) map[string]interface{} {
 			"bool": map[string]interface{}{
 				"must": []map[string]interface{}{
 					{"term": map[string]interface{}{"processor.name": "sourcemap"}},
-					{"term": map[string]interface{}{"sourcemap.bundle_filepath": id.Path}},
-					{"term": map[string]interface{}{"sourcemap.service.name": id.ServiceName}},
-					{"term": map[string]interface{}{"sourcemap.service.version": id.ServiceVersion}},
+					{"match": map[string]interface{}{"sourcemap.bundle_filepath": id.Path}},
```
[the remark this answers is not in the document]
simitt: A`term` query only works on keywords. When supporting custom index names for sourcemap no ES template is (automatically) loaded for this index, indexing a`text` and `keyword` entry. With `match` we can cover this.

simitt:
@graphaelli you are right, it is not a breaking change (I wrote this when I started out with using a `sourcemap` index, but then figured, prefixing it with `apm` might be a better idea) - will change.

tests/system/apmserver.py
```diff
@@ -303,11 +303,25 @@ def config(self):
...
+                    "smap_index": "sourcemap"})
+        return cfg
+
+
 class ExpvarBaseTest(ServerBaseTest):
     config_overrides = {}
 
     def config(self):
-        cfg = super(ServerBaseTest, self).config()
+        cfg = super(ExpvarBaseTest, self).config()
```
jalvz: good eyes :)

sourcemap/elasticsearch.go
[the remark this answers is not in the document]
roncohen: does that mean the sourcemap itself is now going to get indexed?

simitt:
Checking indexing again and reopen PR afterwards.

simitt:
@elastic/apm-server  after our discussions today, I have changed the sourcemap index to default to `apm-{beat.version}-sourcemap`, and the query index pattern to `apm-*-sourcemap`.

jalvz:
is possible to have this independent from 578 (for easier review)?

simitt:
@jalvz after PR #578 was merged, this PR is cleaner now - please review.

beater/config.go
```diff
@@ -83,10 +87,12 @@ func (c *FrontendConfig) SmapMapper() (sourcemap.Mapper, error) {
 	if smap.mapper != nil {
 		return c.SourceMapping.mapper, nil
 	}
+	re := regexp.MustCompile("%.*{.*beat.version.?}")
```
jalvz: I think this method should be documented stating that it modifies its object, and that if a second time is called, nothing happens, ie. if i got it right this behaves like this:
```
c.beatVersion = 1
c.SmapMapper()
print  c.SourceMapping.index  // apm-1

c.beatVersion = 2
c.SmapMapper()
print  c.SourceMapping.index  // still apm-1
```

jalvz:
I think this needs a test somewhere that POSTes a sourcemap and eg a transaction, and then asserts that the sourcemap and the transactions are indeed in different indices.

beater/config.go
simitt: I don't think that the `beat.version` can change (from a logical perspective) without restarting the server.

simitt:
Could you take a look at this test https://github.com/elastic/apm-server/blob/master/tests/system/test_integration.py#L287.

First a `sourcemap` is uploaded and verified that it is there, by using the [sourcemap index](https://github.com/simitt/apm-server/blob/25136bb16f0a2640ca85e4b47b43164f8093f70c/tests/system/apmserver.py#L253) which is [specifically set](https://github.com/simitt/apm-server/blob/25136bb16f0a2640ca85e4b47b43164f8093f70c/tests/system/apmserver.py#L312) and [set in the config](https://github.com/simitt/apm-server/blob/25136bb16f0a2640ca85e4b47b43164f8093f70c/tests/system/config/apm-server.yml.j2#L65).
Then a `error` is [uploaded and verified](https://github.com/simitt/apm-server/blob/25136bb16f0a2640ca85e4b47b43164f8093f70c/tests/system/apmserver.py#L171)

As the `index` and the `smap_index` are defined to be different, this tests that the docs are in different indices. Please let me know if I am missing something. else here

beater/config.go
jalvz: But the point is the same. I don't say it's a bug, just that I would appreciate documentation when any kind of side effects are going on; or rename it to make it evident (eg `SetSourceMapperIfNil`)

jalvz:
that works, thanks

beater/config.go
simitt: that's true, will change.

simitt: @jalvz I did refactor a bit, please have a look.

As this PR doesn't introduce the problem, I suggest to create an Issue and address once we work on config reloading.

Good finding.

jalvz: LGTM. an issue is not necessary for me. It wasn't a comment on config reloading, but on code explicitness. thanks.

PR #583, opened 2018-01-31
remove deprecated usages of logp

jalvz:


beater/beater.go
```diff
@@ -43,6 +43,7 @@ func New(b *beat.Beat, ucfg *common.Config) (beat.Beater, error) {
 
 func (bt *beater) Run(b *beat.Beat) error {
 	var err error
+	logger := logp.NewLogger("beater")
```
roncohen: i suggest adding `logger` as a member on the `beater` and creating it in `New`

model/stacktrace_frame.go
```diff
@@ -97,9 +96,9 @@ func (s *StacktraceFrame) applySourcemap(mapper sourcemap.Mapper, service Servic
 	sourcemapId := s.buildSourcemapId(service)
 	mapping, err := mapper.Apply(sourcemapId, s.Lineno, *s.Colno)
 	if err != nil {
-		logp.Err(fmt.Sprintf("failed to apply sourcemap %s", err.Error()))
-		e, issourcemapError := err.(sourcemap.Error)
-		if !issourcemapError || e.Kind == sourcemap.MapError || e.Kind == sourcemap.KeyError {
+		logp.NewLogger("stacktrace").Errorf("failed to apply sourcemap %s", err.Error())
```
roncohen: what do you think about adding a `var logger = logp.NewLogger("stacktrace")` as a global near the top of the file here instead? similar to how we do in python

roncohen: need to check if it's safe for multi-threaded usage

processor/error/payload.go
```diff
@@ -22,7 +22,7 @@ type payload struct {
 func (pa *payload) transform(config *pr.Config) []beat.Event {
 	var events []beat.Event
 
-	logp.Debug("error", "Transform error events: events=%d, service=%s, agent=%s:%s", len(pa.Events), pa.Service.Name, pa.Service.Agent.Name, pa.Service.Agent.Version)
+	logp.NewLogger("error").Debugf("Transform error events: events=%d, service=%s, agent=%s:%s", len(pa.Events), pa.Service.Name, pa.Service.Agent.Name, pa.Service.Agent.Version)
```
roncohen: the level is `debug`, so maybe `transform` is a better selector here?

sourcemap/cache.go
```diff
@@ -28,12 +28,12 @@ func newCache(expiration time.Duration) (*cache, error) {
 
 func (c *cache) add(id Id, consumer *sourcemap.Consumer) {
 	c.goca.Set(id.Key(), consumer, gocache.DefaultExpiration)
-	logp.Debug("sourcemap", "Added id %v. Cache now has %v entries.", id.Key(), c.goca.ItemCount())
+	logp.NewLogger("sourcemap").Debugf("Added id %v. Cache now has %v entries.", id.Key(), c.goca.ItemCount())
```
roncohen: here also it might make sense to have a global logger for this file

roncohen:
Nice to get selectors on everything 👍

sourcemap/cache.go
jalvz: Beats documentation warns against using `logp.NewLogger` globally.
 
Can you have a look at the question i wrote you in your logp PR? I don't know what to do with that case (https://github.com/elastic/apm-server/blob/master/beater/handlers.go#L388-L393).

roncohen: ah, didn't see that 👍

roncohen: let's make "sourcemap" a constant instead?

jalvz: added here e305b844f3f52fcefa646e21c4abbb9cb365a8ce

ruflin:
I added @andrewkroh as reviewer as I think it would be good to get his eyes on this change. It's one of the first changes using the new logger API which was not done by Andrew AFAIK.

beater/client.go
```diff
@@ -29,7 +29,7 @@ func isServerUp(secure bool, host string, numRetries int, retryInterval time.Dur
 
 	for i := 0; i <= numRetries; i++ {
 		if check() {
-			logp.Info("HTTP Server ready")
+			logp.NewLogger("HTTP Client").Info("HTTP Server ready")
```
andrewkroh: While there is absolutely no requirement to use single word keys, conventionally we have named the loggers using lowercase and no spaces like `http_client` or `httpclient` or `http.client`.

sourcemap/cache.go
```diff
@@ -10,7 +10,10 @@ import (
 	"github.com/elastic/beats/libbeat/logp"
 )
 
-const MIN_CLEANUP_INTERVAL_SECONDS float64 = 60
+const (
+	MIN_CLEANUP_INTERVAL_SECONDS float64 = 60
+	LOGGER_SELECTOR                      = "sourcemap"
```
andrewkroh: Constants are usually written in [MixedCase](https://golang.org/doc/effective_go.html#mixed-caps).

jalvz:
thanks @andrewkroh !

PR #587, opened 2018-01-31
Validate configuration on startup

jalvz:


beater/beater.go
```diff
@@ -21,18 +22,15 @@ type beater struct {
...
-		if _, err := regexp.Compile(beaterConfig.Frontend.LibraryPattern); err != nil {
-			return nil, errors.New(fmt.Sprintf("Invalid regex for `library_pattern`: %v", err.Error()))
-		}
-		if _, err := regexp.Compile(beaterConfig.Frontend.ExcludeFromGrouping); err != nil {
-			return nil, errors.New(fmt.Sprintf("Invalid regex for `exclude_from_grouping`: %v", err.Error()))
-		}
-		if b.Config != nil && b.Config.Output.Name() == "elasticsearch" {
-			beaterConfig.setElasticsearch(b.Config.Output.Config())
-		}
+		beaterConfig.setElasticsearch(b.Config.Output.Config())
```
simitt: why did you remove th check for `b.Config != nil && b.Config.Output.Name() == "elasticsearch"` here?

beater/beater.go
```diff
@@ -72,3 +70,27 @@ func (bt *beater) Stop() {
...
+		}
+	}
+	if beatConfig != nil && beatConfig.Output.Name() == "elasticsearch" {
+		clients, err := es.NewElasticsearchClients(beatConfig.Output.Config())
+		if err != nil {
+			return fmt.Errorf("elasticsearch not available: %v", err)
+		}
+		for _, client := range clients {
+			if err := client.Connect(); err != nil {
+				return fmt.Errorf("elasticsearch not available: %v", err)
```
simitt: This would mean if multiple ES connections are configured, but one is not available the server doesn't start. What benefit do you see from this?

jalvz: Yeah, maybe 1 up is enough as you suggest. I'm fine with that.

I did this small refactor also as a initial work on bootstrap checks, there are many more things we can add here (dependencies / incompatibilities between options, ssl, etc) but i don't have the full picture of how everything plays together.

simitt:
Which advantage do you see in validating that ES is up when starting the server? 
Atm if logging is enabled a warning is printed that no ES host is reachable. There is an internal in-memory queue where some data can be buffered. 
How would you handle a connection loss after a server was started? 

Generally I like the current handling where the server could overcome short connection losses, but maybe I am overlooking something?

jalvz:
In case it can't connect to ES on startup, the difference between a running apm-server and a killed one is whether you passed `-E setup.dashboards.enable=true`. This doesn't make any sense to me.

Buffering messages is nice, but i think is a different case.

simitt:
I totally agree that it is weird the server doesn't start when there is no ES connection and you set `-E setup.dashboards.enable=true`. If this is the only reason, I'd rather address this with the dashboard loading setting though tbh.

jalvz:
You mean to make beats buffer dashboard data until ES becomes available?

simitt:
no, I meant looking into what exactly is going on when `-E setup.dashboards.enable=true` is set, as this should be loaded via Kibana and not ES any more afaik.

jalvz:
i see. closing this for now, then

PR #590, opened 2018-01-31
document code contribution workflow

graphaelli:


CONTRIBUTING.md
```diff
@@ -42,3 +44,13 @@ Once your changes are ready to submit for review:
...
     We might not be able to review your code as fast as we would like to, but we'll do our best to dedicate it the attention it deserves.
     Your effort is much appreciated!
+
+### Workflow
+
+All feature development and most bug fixes hit the master branch first.  Pull requests should be reviewed by someone with commit access.  Once approved, the author of the pull request, or reviewer if the author does not have commit access, should "Squash and merge".
+
+### Backports
+
+After the pull request has been reviewed, a committer will tag it with the target version(s).  Once a version is released, new features are frozen and will not be backported.  Bug fixes may be backported on a case by case basis.  The committer of the original pull request, typically the author, is responsible for backporting the changes to the target versions.  Each backport is performed through its own pull request, tagged with the target version and "backport", and merged with "Create a merge commit".  Straightforward backports may be merged without review.
```
roncohen: I think we should tag it with target versions up front, so the discussion around which versions it should get backported to is part of the review.

protip for asciidoc: each sentence on its own line so giving per sentence comments is easier.

roncohen: > Once a version is released, new features are frozen and will not be backported.  

this is technically correct, but might seem hard to understand if the reader is not familiar with our support scheme. If possible, it would be nice to try and elaborate. Maybe describe the support scheme (latest minor from the previous major).

roncohen: why do we create merge commits for backports?

graphaelli: @roncohen Ah great, will move each sentence to its own line.

Regarding timing of tagging, I was thinking only of the case where a contributor proposed the change, so they know they don't have to deal with with versions to target / we'll take care of it, but I think your point is actually more important - what do you think of "during review"?

graphaelli: I've been meaning to ask about the merge commit policy - I think it was so we could track what got backported where but that doesn't actually work with the way we branch, unless I'm misunderstanding something.

PR #596, opened 2018-02-01
Fix broken link and remove beta label

roncohen:


simitt:
@roncohen I think you should also remove the Beta headers from the docs.

roncohen:
@simitt https://github.com/elastic/apm-server/commit/7b6948229f73537e03e1f74cff15dadbb032a665#diff-3db1a33866d6bcfa775cb0afac1c5bc6

simitt:
Ah sry, I think I checked out your branch before you pushed this so didn't see they are gone.

PR #610, opened 2018-02-01
Log a warning for sourcemap uploads with duplicated caching keys

### LINKED ISSUE #538: Log error if sourcemaps with the same key are uploaded
Uploaded sourcemaps are uniquely identified by their `service.name`, `service.version` and `bundle_filepath`. In case multiple sourcemaps with the same key information are uploaded, an error should be logged. 

see [former discussions](https://github.com/elastic/apm-server/pull/381#pullrequestreview-83951603).

jalvz:
fixes #538

### DIFF
--- a/sourcemap/mapper.go
+++ b/sourcemap/mapper.go
@@ -65,8 +65,7 @@ func (m *SmapMapper) Apply(id Id, lineno, colno int) (*Mapping, error) {
 
 func (m *SmapMapper) NewSourcemapAdded(id Id) {
 	_, err := m.Accessor.Fetch(id)
-	sourcemapErr, ok := err.(Error)
-	if err == nil || (ok && sourcemapErr.Kind == MapError) {
+	if err == nil {
 		logp.NewLogger("sourcemap").Warnf("Overriding sourcemap for service %s version %s and file %s",
 			id.ServiceName, id.ServiceVersion, id.Path)
 	}
--- a/sourcemap/mapper.go
+++ b/sourcemap/mapper.go
@@ -64,10 +64,10 @@ func (m *SmapMapper) Apply(id Id, lineno, colno int) (*Mapping, error) {
 }
 
 func (m *SmapMapper) NewSourcemapAdded(id Id) {
+	m.Accessor.Remove(id)
 	_, err := m.Accessor.Fetch(id)
 	if err == nil {
 		logp.NewLogger("sourcemap").Warnf("Overriding sourcemap for service %s version %s and file %s",
 			id.ServiceName, id.ServiceVersion, id.Path)
 	}
-	m.Accessor.Remove(id)
 }
--- a/sourcemap/mapper.go
+++ b/sourcemap/mapper.go
@@ -64,10 +64,11 @@ func (m *SmapMapper) Apply(id Id, lineno, colno int) (*Mapping, error) {
 }
 
 func (m *SmapMapper) NewSourcemapAdded(id Id) {
-	m.Accessor.Remove(id)
 	_, err := m.Accessor.Fetch(id)
-	if err == nil {
+	sourcemapErr, ok := err.(Error)
+	if err == nil || (ok && sourcemapErr.Kind == MapError) {
 		logp.NewLogger("sourcemap").Warnf("Overriding sourcemap for service %s version %s and file %s",
 			id.ServiceName, id.ServiceVersion, id.Path)
 	}
+	m.Accessor.Remove(id)
 }
--- a/sourcemap/mapper.go
+++ b/sourcemap/mapper.go
@@ -5,6 +5,7 @@ import (
 	"time"
 
 	"github.com/elastic/beats/libbeat/common"
+	"github.com/elastic/beats/libbeat/logp"
 )
 
 type Mapper interface {
@@ -63,5 +64,10 @@ func (m *SmapMapper) Apply(id Id, lineno, colno int) (*Mapping, error) {
 }
 
 func (m *SmapMapper) NewSourcemapAdded(id Id) {
+	_, err := m.Accessor.Fetch(id)
+	if err == nil {
+		logp.NewLogger("sourcemap").Warnf("Overriding sourcemap for service %s version %s and file %s",
+			id.ServiceName, id.ServiceVersion, id.Path)
+	}
 	m.Accessor.Remove(id)
 }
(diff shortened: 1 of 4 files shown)


sourcemap/mapper.go
```diff
@@ -63,5 +64,10 @@ func (m *SmapMapper) Apply(id Id, lineno, colno int) (*Mapping, error) {
 }
 
 func (m *SmapMapper) NewSourcemapAdded(id Id) {
+	_, err := m.Accessor.Fetch(id)
+	if err == nil {
+		logp.NewLogger("sourcemap").Warnf("Overriding sourcemap for service %s version %s and file %s",
+			id.ServiceName, id.ServiceVersion, id.Path)
+	}
```
simitt: I think this should be done after `m.Accessor.Remove(id)` as we use negative caching. The cache should first be emptied for this entry, otherwise it could happen that a `nil` value is read from cache (no error) although no entry is there.

jalvz: ah, missed that `Fetch` looks up in the cache first. Good catch, fixed now

jalvz:
jenkins retest this please

sourcemap/mapper.go
jalvz: hmm, i don't think that works... If I `remove` first and `fetch` later, the `fetch` will put a `<id, nil>` entry in the cache, undoing the previous `remove` and leaving the cache inconsistent. So `remove` necessarily needs to happen later.
I changed it again now, but it is a bit fiddly. does this makes sense to you?

sourcemap/mapper.go
```diff
@@ -63,5 +64,11 @@ func (m *SmapMapper) Apply(id Id, lineno, colno int) (*Mapping, error) {
 }
 
 func (m *SmapMapper) NewSourcemapAdded(id Id) {
+	_, err := m.Accessor.Fetch(id)
+	sourcemapErr, ok := err.(Error)
+	if err == nil || (ok && sourcemapErr.Kind == MapError) {
```
simitt: I have looked at the `Fetch` method again, and here a check for `err == nil` is sufficient. 
The only two possibilities that no error is returned is when the sourcemap is found in the cache or found in ES.

However, if a nil sourcemap is cached or nothing is found in ES but no error was thrown, then a `MapError` is returned from `Fetch`, but in this case there should not be logged a warning here. 

To summarize: Your first attempt was perfectly fine, sorry for my misleading comment.

tests/system/test_integration.py
```diff
@@ -143,33 +145,37 @@ def count_library_frames(self, doc, lf):
...
+            file_name='bundle.js.map', bundle_filepath=path)
         assert r.status_code == 202, r.status_code
         self.wait_for_sourcemaps()
 
         self.load_docs_with_template(self.get_error_payload_path(),
                                      'http://localhost:8200/v1/errors',
                                      'error',
                                      1)
-        self.assert_no_logged_warnings()
+        self.assert_no_logged_warnings(self.ignore_override_warnings)
```
simitt: I don't think that this is right, there should not be a warning issued, as the sourcemap is uploaded only once, so please don't ignore it here.

tests/system/test_integration.py
```diff
@@ -143,33 +145,37 @@ def count_library_frames(self, doc, lf):
...
+            file_name='bundle.js.map', bundle_filepath=path)
         assert r.status_code == 202, r.status_code
         self.wait_for_sourcemaps()
 
         self.load_docs_with_template(self.get_error_payload_path(),
                                      self.errors_url,
                                      'error',
                                      1)
-        self.assert_no_logged_warnings()
+        self.assert_no_logged_warnings(self.ignore_override_warnings)
```
simitt: same - I don't think the warning should be issued.

tests/system/test_integration.py
```diff
@@ -185,7 +191,7 @@ def test_backend_transaction(self):
                                      'http://localhost:8200/v1/transactions',
                                      'transaction',
                                      2)
-        self.assert_no_logged_warnings()
+        self.assert_no_logged_warnings(self.ignore_override_warnings)
```
simitt: same as above

tests/system/test_integration.py
```diff
@@ -201,7 +207,7 @@ def test_frontend_transaction(self):
                                      self.transactions_url,
                                      'transaction',
                                      2)
-        self.assert_no_logged_warnings()
+        self.assert_no_logged_warnings(self.ignore_override_warnings)
```
simitt: same as above

tests/system/test_integration.py
```diff
@@ -210,12 +216,13 @@ def test_no_sourcemap(self):
...
                                      1)
-        self.check_frontend_error_sourcemap(False, expected_err="No Sourcemap available for")
+        self.check_frontend_error_sourcemap(
+            False, expected_err="No Sourcemap available for")
 
     @unittest.skipUnless(INTEGRATION_TESTS, "integration test")
     def test_no_matching_sourcemap(self):
         r = self.upload_sourcemap('bundle_no_mapping.js.map')
-        self.assert_no_logged_warnings()
+        self.assert_no_logged_warnings(self.ignore_override_warnings)
```
simitt: same as above

tests/system/test_integration.py
```diff
@@ -267,7 +278,7 @@ def test_sourcemap_mapping_cache_usage(self):
                                      self.errors_url,
                                      'error',
                                      1)
-        self.assert_no_logged_warnings()
+        self.assert_no_logged_warnings(self.ignore_override_warnings)
```
simitt: same as above

tests/system/test_integration.py
```diff
@@ -280,15 +291,19 @@ def test_sourcemap_mapping_cache_usage(self):
...
```
simitt: this should also not lead to a warning

tests/system/test_integration.py
```diff
@@ -297,7 +312,7 @@ def test_sourcemap_cache_expiration(self):
...
```
simitt: same as above

sourcemap/mapper.go
jalvz: Damn, its hard to think about this. `Fetch` could also return `KeyError`, but it doesn't matter because in that case shouldn't be logged neither.

jalvz:
> What do you think of also issueing a warning when multiple sourcemaps are found when fetching it?
done here: 03313824452601ea3f968ada8962495f0c38dd22
also fixed the previous commit mess

jalvz:
@simitt something else here?

PR #611, opened 2018-02-01
Attempt at fixing docs failure

roncohen:


lcawl:
I think this is worth a shot, so LGTM

PR #509, opened 2018-01-16
optionally expose expvar

graphaelli:
Primarily for use with the [metricbeat golang module](https://www.elastic.co/guide/en/beats/metricbeat/6.1/metricbeat-module-golang.html).

PR #615, opened 2018-02-01
[6.x] optionally expose expvar (#509)

graphaelli:
 - optionally expose expvar 

* add changelog and config file entries (#509)

### DIFF
--- a/_meta/beat.reference.yml
+++ b/_meta/beat.reference.yml
@@ -70,6 +70,14 @@ apm-server:
       # a matching index pattern needs to be specified here.
       #index_pattern: "apm-*"
 
+  # golang expvar support - https://golang.org/pkg/expvar/
+  #expvar:
+  # Set to true to Expose expvar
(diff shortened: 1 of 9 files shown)


_meta/beat.reference.yml
```diff
@@ -70,6 +70,14 @@ apm-server:
       # a matching index pattern needs to be specified here.
       #index_pattern: "apm-*"
 
+  # golang expvar support - https://golang.org/pkg/expvar/
+  #expvar:
+  # Set to true to Expose expvar
+  #enabled: false
```
ruflin: Should `enabled` and `url` be in the `expvar` namespace` if yes, it needs 2 additional spaces.

simitt: You are right, but it will be fixed with https://github.com/elastic/apm-server/pull/640/files#diff-386819a19a12310e5e0ea62392099482R73

PR #618, opened 2018-02-05
document and git-ignore the new keystore provided by beats

### LINKED ISSUE #559: Test and document keystore functionality from beats for APM Server
Beats merged a keystore (https://github.com/elastic/beats/pull/5687) for 6.2. We should make ourselves familiar with it and see if it "just works" for APM Server. If that is the case, we should document it. See https://github.com/elastic/beats/pull/6173.

This doesn't strictly need to happen before 6.2 release, but would be good to investigate at least.

jalvz:
fixes #559

### DIFF
--- a/docs/copied-from-beats/command-reference.asciidoc
+++ b/docs/copied-from-beats/command-reference.asciidoc
@@ -19,7 +19,15 @@
 :keystore-command-short-desc: Manages the <<keystore,secrets keystore>>
 :modules-command-short-desc: Manages configured modules
 :run-command-short-desc: Runs {beatname_uc}. This command is used by default if you start {beatname_uc} without specifying a command
+
+ifeval::["{has_ml_jobs}"=="yes"]
 :setup-command-short-desc: Sets up the initial environment, including the index template, Kibana dashboards (when available), and machine learning jobs (when available)
+endif::[]
+
+ifeval::["{has_ml_jobs}"!="yes"]
+:setup-command-short-desc: Sets up the initial environment, including the index template, Kibana dashboards (when available)
+endif::[]
+
 :test-command-short-desc: Tests the configuration
 :version-command-short-desc: Shows information about the current version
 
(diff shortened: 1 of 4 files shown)


jalvz:
jenkins retest this please

jalvz:
jenkins retest it please

docs/copied-from-beats/command-reference.asciidoc
```diff
@@ -16,23 +16,22 @@
 
 :export-command-short-desc: Exports the configuration or index template to stdout
 :help-command-short-desc: Shows help for any command
+:keystore-command-short-desc: Manages the <<keystore,secrets keystore>>
 :modules-command-short-desc: Manages configured modules
 :run-command-short-desc: Runs {beatname_uc}. This command is used by default if you start {beatname_uc} without specifying a command
-
-ifeval::["{has_ml_jobs}"=="yes"]
```
simitt: Please don't remove this, there is a PR https://github.com/elastic/beats/pull/6184/files open in beats that introduces this distinction, as it doesn't make sense to describe ML here for APM.

jalvz: I copied over a fresh version of the command-reference that has an entry for the keystore.
Once that PR gets merged and copied over, we will have the right thing, no?

I think is better to wait for that PR to get merged first, than manually editing copied files.

jalvz:
Yes, I tested and it works.

docs/copied-from-beats/command-reference.asciidoc
simitt: As soon as this gets merged, it is in the official docs. 

Thus, I'd not suggest to merge in some documentation that is not supposed to show up for APM Server, but rather to copy over the part for the keystore as is and leave the other part of the file untouched for now.

simitt:
As the https://github.com/elastic/beats/pull/6184/files is merged, could you please update this PR so we can merge it.

jalvz:
I did already, the change up there in L387 seems to be consistent with the updated docs

PR #619, opened 2018-02-05
Add line breaks after punctuation in CONTRIBUTING.md

watson:
Content have not been changed, only line breaks. This makes future diffs more useful and follows our documentation guidelines used in our regular docs.


watson:
Done: #626

PR #623, opened 2018-02-05
docs: fix typo in 6.1 doc page header

watson:
I hope this is the correct approach for fixing something directly on a version branch

### DIFF
--- a/docs/guide/page_header.html
+++ b/docs/guide/page_header.html
@@ -1,1 +1,1 @@
-You are looking at documentation for an beta release.
+You are looking at documentation for a beta release.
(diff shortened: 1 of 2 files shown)


jalvz:
I believe it has to be fixed in master and then backported? (in case is also wrong in master, ofc)

simitt:
@jalvz generally yes, but the docs headers are removed in master, so I'd say this is the proper approach to fix it.

jalvz:
ok, cool. i just `grep`ed "an beta release" in master and showed 42 occurrences, so something is wrong somewhere, but i have no idea of how docs are generated.

watson:
@jalvz that's probably from your local `build` directory - which is not checked into git. So it's just the result of building the docs for version 6.1 I think

jalvz:
Yes, you are right, sorry for the noise. I was actually confused because `make docs` doesn't remove the `build` folder, so i had the wrong stuff locally.

PR #624, opened 2018-02-06
target for clean and docs together

jalvz:
make docs without removing '/build' first behaves in a unexpected way

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -80,3 +80,6 @@ check-full: check
 notice: python-env
 	@echo "Generating NOTICE"
 	@$(PYTHON_ENV)/bin/python ${ES_BEATS}/dev-tools/generate_notice.py . -e '_beats' -s "./vendor/github.com/elastic/beats" -b "Apm Server"
+
+.PHONY: force-update-docs
+force-update-docs: clean docs


watson:
@jalvz What is the unexpected behaviour you're referring to? Deleting the `build` folder is an expensive operation as it requires a 4GB download to rebuild, so it's not something you normally like to do for every re-build. My apologies if I've misunderstood the intent of the PR 😃

jalvz:
i was confused by `make docs` not using a fresh copy of  the docs, so i was building the wrong thing. 
`make docs` is still available. `update-docs` is a convenience shortcut that at least for me is simpler to remember than `make clean docs`

watson:
Ah, got'ya 😃 

I think there's an ambiguity between `make docs` and `make update-docs` - i.e. it's not as if `make docs` doesn't update the docs locally. And I feel that `clean` in the realm of makefiles is fairly well understood as cleaning out any pre-compiled or pre-downloaded assets before "making". Just my 10 cents, I have no strong feelings

simitt:
I also think that the name `update-docs` is a bit ambiguous. How about naming it something like `init-docs`? Or adding some hint of `force` to the command, `force-update-docs`?

Note:
I actually tried out 
* adding a header
* running `make docs`
* removing header
* running `make docs`

The docs were created properly and didn't show the header anymore, but the created html for the headers was still in the `build` folder. So looking at the output shows expected docs, but grepping in the folder does not give expected output.

jalvz:
I changed the name now, but if this is only useful for me I can close this and create an alias for myself...

watson:
My makefile foo isn't the best, but `make force-update-docs` is just an alias for the already existing `make clean docs` right? Or do the two commands to something different?

PR #628, opened 2018-02-06
Fix broken doc links.

### LINKED ISSUE #627: Missing sentences in the docs
When looking at https://www.elastic.co/guide/en/apm/server/master/configuration-template.html, I noticed some incomplete sentences:

> setup.template.settings
A dictionary of settings to place into the settings.index dictionary of the Elasticsearch template. For more details about the available Elasticsearch mapping options, please

> setup.template.settings._source
A dictionary of settings for the _source field. For the available settings,

@watson found out that https://github.com/elastic/beats/blob/master/libbeat/docs/shared-beats-attributes.asciidoc is not yet copied from beats, which contains certain variables like the Elasticsearch URL.

That's why the asciidoc rendering fails whenever the `{elasticsearch}` template variable is referenced. 

However, the build does not fail with an error but the corresponding line does not get rendered which is quite scary.

So we should at first fix this by copying over the file from beats. But we should also see if we get the build to fail or at least to warn if there are missing variables. Furthermore we should look at ways to reduce copy/paste from beats docs by using git submodules for example. But I guess that one is for another issue.

simitt:
fixes #627 

Fixed: 
screen shot 2018-02-06 at 17 12 57

released atm:
screen shot 2018-02-06 at 17 13 28

### DIFF
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -12,6 +12,10 @@ include::{asciidoc-dir}/../../shared/attributes.asciidoc[]
 :discuss_forum: apm
 :github_repo_name: apm-server
 
+:libbeat: http://www.elastic.co/guide/en/beats/libbeat/{doc-branch}
+:logstashdoc: https://www.elastic.co/guide/en/logstash/{doc-branch}
+:elasticsearch: https://www.elastic.co/guide/en/elasticsearch/reference/{doc-branch}
+
 ifdef::env-github[]
 NOTE: For the best reading experience,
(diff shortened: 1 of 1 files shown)


watson:
@simitt would it be possible instead to import the missing file from beats? With this approach don't we risk that the variables in beats and apm-server gets out of sync. Every time they update them in beats we'll need to have a process of discovering this and update ours right?

simitt:
@watson I only added the variables to where some others were defined already to stay consistent.
Additionally we have the rule not to differ in `copied_from_beats` folder from beats docs, but some variables need to be set differently.

However I fully agree that this is risky to get out of sync.

PR #641, opened 2018-02-07
Use type integer instead of number where applicable

felixbarny:
In most cases where type: number has been used in the json schema files,
floating point numbers are not neccessary.

This change allows automated tools like the jsonschema2pojo converter to
generate Integer instead of Double fields.

In these instances, I didn't change `number` to `integer`:

```
transactions/Mark.patternProperties
transactions/Span.duration
transactions/Span.start
transactions/Transaction.duration
```

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -43,6 +43,7 @@ https://github.com/elastic/apm-server/compare/71df0d96445df35afe27f38bcf734a0828
 - Updated readme doc urls {pull}356[356]
 - Use updated stack trace frame values for calculating error `grouping_keys` {pull}485[485]
 - Fix panic when a signal is delivered before the server is instantiated {pull}580[580]
+- Use type integer instead of number in JSON schemas where applicable {pull}641[641]
 
 ==== Added
 - Include build time and revision in version information {pull}396[396]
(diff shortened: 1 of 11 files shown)


simitt:
@felixbarny could you please add a Changelog entry.

@roncohen should we consider changing the scheme to only allow `integer` over `number` for several attributes a breaking change?

roncohen:
It seems unlikely that this will break anything as the intention with `Number` in those places was always actually `Integer`. 

@elastic/apm-agent-devs do you see anything here that would break?

If not, I suggest we merge this so it gets scheduled for 6.3

jahtalab:
LGTM

watson:
LGTM

beniwohli:
Same!

simitt:
Great! @felixbarny could you please rebase your PR to master, then we can merge it as soon as CI is green.

simitt:
Can you please create a backport PR for `6.x` by using https://github.com/sqren/backport

felixbarny:
I'm a bit confused, I thought we are not backporting this?

simitt:
According to https://github.com/elastic/apm-server/pull/641#issuecomment-366938492 we want to have this in `6.3` - so we need to backport it to `6.x`. 
We are not backporting it to `6.2` as a bugfix.

felixbarny:
right!

Issue #566, opened 2018-01-29
Preserve original stack trace frame and keep transform action idempotent.


simitt:
Right now the `transform` action is not idempotent anymore, as the original values are changed within 
- source mapping
- setting library frame

Investigate if idempotency should be ensured by keeping the original values around in the structs.
cc @graphaelli (you originally raised this question in a source map PR)

simitt:
After discussing this offline, we decided to preserve the original stack trace frame values within a separate namespace. This should also lead to idempotency of transform action.

PR #647, opened 2018-02-07
Keep original values when transforming.

simitt:
Persist original values when
* applying source mapping
* setting library_frame indicator
by storing original values within `original` namespace.

Ensure transform operation is idempotent,
by using original values if applied multiple times.

implements #566


simitt:
jenkins, retest this please.

PR #648, opened 2018-02-07
Add note to docs for dashboard loading.

### LINKED ISSUE #519: Loading dashboards with non-default index doesn't work for 5.x
Overwriting dashboards index by setting the config option `setup.dashboards.index` to use a non-default index pattern doesn't work.
Reason for this is that before 6.x the dashboards are not loaded via the Kibana API.

Add note to documentation about this.

simitt:
Describe that `setup.dashboards.index`
only works for >= 6.0
closes #519 

As elastic/beats#6197 was merged, I updated it also in apm-server.


simitt:
jenkins, retest this please

PR #651, opened 2018-02-08
Set array item types to string

felixbarny:
relates to #641


CHANGELOG.asciidoc
```diff
@@ -44,6 +44,7 @@ https://github.com/elastic/apm-server/compare/71df0d96445df35afe27f38bcf734a0828
 - Use updated stack trace frame values for calculating error `grouping_keys` {pull}485[485]
 - Fix panic when a signal is delivered before the server is instantiated {pull}580[580]
 - Use type integer instead of number in JSON schemas where applicable {pull}641[641]
+- Set array item types to string {pull}651[651]
```
beniwohli: can you add "in JSON schemas" similar to your former entry? That makes the changelog easier to read for people who don't have all the context

felixbarny: done

CHANGELOG.asciidoc
```diff
@@ -44,6 +44,7 @@ https://github.com/elastic/apm-server/compare/71df0d96445df35afe27f38bcf734a0828
 - Use updated stack trace frame values for calculating error `grouping_keys` {pull}485[485]
 - Fix panic when a signal is delivered before the server is instantiated {pull}580[580]
 - Use type integer instead of number in JSON schemas where applicable {pull}641[641]
+- Set array item types to string in JSON schemas {pull}651[651]
```
roncohen: This should not be under 6.2 :)

simitt: Same for `Use type integer instead of number in JSON schemas where applicable {pull}641[641]`

felixbarny: woops 😬

PR #654, opened 2018-02-08
Improve sourcemap docs

jahtalab:
cc @simitt

### DIFF
--- a/docs/sourcemaps.asciidoc
+++ b/docs/sourcemaps.asciidoc
@@ -77,10 +77,10 @@ Send an example source map to the APM Server:
 ["source","sh",subs="attributes"]
 ---------------------------------------------------------------------------
 curl -X POST http://127.0.0.1:8200/v1/client-side/sourcemaps \
-  -F service_name="test-service" 
-  -F service_version="1.0" 
-  -F bundle_filepath="/static/js/main.js.map" 
-  -F sourcemap=@docs/data/intake-api/generated/sourcemap/bundle.js.map 
+  -F service_name="test-service" \
+  -F service_version="1.0" \
+  -F bundle_filepath="http://localhost/static/js/bundle.js" \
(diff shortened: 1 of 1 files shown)


docs/sourcemaps.asciidoc
```diff
@@ -77,10 +77,10 @@ Send an example source map to the APM Server:
...
 ---------------------------------------------------------------------------
 curl -X POST http://127.0.0.1:8200/v1/client-side/sourcemaps \
-  -F service_name="test-service" 
-  -F service_version="1.0" 
-  -F bundle_filepath="/static/js/main.js.map" 
-  -F sourcemap=@docs/data/intake-api/generated/sourcemap/bundle.js.map 
+  -F service_name="test-service" \
+  -F service_version="1.0" \
+  -F bundle_filepath="http://localhost/static/js/bundle.js" \
+  -F sourcemap=@bundle.js.map
```
simitt: Why are you changing the location of the file? This is the path assuming you are in the root folder of the `apm-server`.

jahtalab: @simitt , IMO, an example should be as simple as possible without much distraction, Since this example is supposed be read by devs who want to use that in their application, I don't think having a long path that is specific to the apm-server is good.

simitt:
Looks like we cannot decide on which structure is better, @roncohen which version do you prefer?

docs/sourcemaps.asciidoc
simitt: I revisited my opinion on this, as I realized that most users probably have downloaded the apm-server package and not checked out the source code.

jahtalab:
I'm closing this in favor of https://github.com/elastic/apm-server/pull/771 to clean up the branch on elastic/apm-server repo and also request the changes to be merged into master.

PR #659, opened 2018-02-09
use libbeat version

graphaelli:
before:

```
$ ./apm-server version
apm-server version 7.0.0-alpha1 (amd64), libbeat 7.0.0-alpha1 [unknown built unknown]
$ ./apm-server -e
...
INFO	instance/beat.go:213	Setup Beat: apm-server; Version: 7.0.0-alpha1
```

after:
```
$ ./apm-server version
apm-server version 7.0.0-alpha1 (amd64), libbeat 7.0.0-alpha1 [1432fce151b137543631079b21a40b5b006e178c built 2018-02-09 14:53:23 +0000 UTC]

$ ./apm-server -e
...
INFO	[server]	beater/server.go:30	Starting apm-server [1432fce151b137543631079b21a40b5b006e178c built 2018-02-09 14:53:23 +0000 UTC]. Hit CTRL-C to stop it.
```

I do wish it had `git describe --dirty` info but that's for another time.
I'll follow up with a change to beats to make `generate.py` do the right thing for new beats.
I'll also separately follow up on this not working in releases.


graphaelli:
@jalvz I think it originally came out of your pr - #395 - we should revisit that when time permits, It would be great to expose this for display in the ui / for agents to interrogate / etc

jalvz:
yeah, i meant i didnt know about the `version` object in libbeat (or i didnt remember at least)
totally agree with exposing it

PR #660, opened 2018-02-12
docs: fix missing app -> service renames

watson:
Fixes elastic/apm-agent-nodejs#237

### DIFF
--- a/docs/guide/index.asciidoc
+++ b/docs/guide/index.asciidoc
@@ -185,7 +185,7 @@ Currently,
 Elastic APM has agents for Node.js and for Python.
 
 Setting up a new service to be monitored requires installing the agent in the service,
-coming up with a good app name for the service,
+coming up with a good name for the service,
 and then configuring the agents so they know the address of your APM Server and the service name.
 
 [[choose-service-name]]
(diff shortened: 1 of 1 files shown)


docs/guide/index.asciidoc
```diff
@@ -220,8 +220,8 @@ Then configure the elastic-apm-node module inside your service by adding the fol
 ----------------------------------
 // Add this to the VERY top of the first file loaded in your app
```
simitt: you could also rename `app` to `service` here, but not mandatory..

beniwohli:
@watson would it maybe make sense to not include a configuration snippet and only link to the agent docs? The more places we have this snippet, the more we have to remember to update it next time.

watson:
@beniwohli Absolutely. I never really questioned having it on this page as it was there from the beginning. But I'll revisit this later 👍

docs/guide/index.asciidoc
watson: From a Node.js developers perspective, I think it's more clear to use "app" in this location. But I'm happy to reevaluate if you feel it's confusing

PR #661, opened 2018-02-12
Resolve attrs in docs example.

simitt:


beniwohli:
LGTM, good find! I guess we should apply the same to https://github.com/elastic/beats/blob/3cb07cb8daade784a8028e10a5ae714243777a4b/libbeat/docs/loggingconfig.asciidoc?

PR #663, opened 2018-02-12
[6.2] docs: fix missing app -> service renames (#660)

watson:
Backports the following commits to 6.2:
 - docs: fix missing app -> service renames  (#660)


watson:
The failing tests should not have anything to do with this PR, so it should be good to merge

PR #667, opened 2018-02-13
bump elastic stack tests to 6.2.2

graphaelli:
testing here before proposing the change upstream


graphaelli:
looks good, opened elastic/beats#6370

PR #670, opened 2018-02-14
Update beats framework to 06622f6

graphaelli:
includes:
* update go to 1.9.4
* fix version subcommand in packages


graphaelli:
This should pass CI after the next snapshot build succeeds

graphaelli:
snapshots are still failing, #672 in temporarily, rebased this on that and retained known good es change

PR #672, opened 2018-02-15
temporarily test against es 6.2.2

graphaelli:
since 7.0.0-alpha1 builds are currently having an issue


graphaelli:
can be reverted later or beat sync will do it for us.

PR #675, opened 2018-02-16
First attempt at ad-hoc map to struct convertion to improve performance

jalvz:
Some test results with `hey-apm` on my mid '14 mac (8gb):

+ Heavy profile, 1 minute:

*This branch*: Real Memory Size tops at **370Mb**, and CPU at **160%**
```
    http://localhost:8200/v1/errors 0
      [202] 315 responses (100.00%)
      total 315 responses (5.21 rps)

    http://localhost:8200/v1/errors 1
      [202] 103 responses (100.00%)
      total 103 responses (1.70 rps)

    http://localhost:8200/v1/transactions 2
      [202] 5067 responses (99.74%)
      [503] 13 responses (0.26%)
      total 5080 responses (83.98 rps)

    http://localhost:8200/healthcheck 3
      [200] 60 responses (100.00%)
      total 60 responses (0.99 rps)
```

*Master*: Real Memory Size tops at **570Mb**, and CPU at **350%**
```
    http://localhost:8200/v1/errors 0
      [202] 181 responses (88.73%)
      [503] 23 responses (11.27%)
      total 204 responses (3.36 rps)

    http://localhost:8200/v1/errors 1
      [202] 62 responses (87.32%)
      [503] 9 responses (12.68%)
      total 71 responses (1.17 rps)

    http://localhost:8200/v1/transactions 2
      [202] 2709 responses (91.12%)
      [503] 264 responses (8.88%)
      total 2973 responses (48.97 rps)

      2 errors:
      [2] Post http://localhost:8200/v1/transactions: EOF
```

+ Maxspans profile, 5 minutes:

*This branch*: Real Memory Size tops at **2,3Gb**
```
  http://localhost:8200/v1/transactions 0
      [202] 473 responses (80.58%)
      [503] 114 responses (19.42%)
      total 587 responses (1.95 rps)

    125 errors:
      [108] Post http://localhost:8200/v1/transactions: EOF
      [17]  Post http://localhost:8200/v1/transactions: net/http: request canceled (Client.Timeout exceeded while awaiting headers)

    http://localhost:8200/healthcheck 1
      [200] 299 responses (100.00%)
      total 299 responses (0.99 rps)
```

*Master*: Real Memory Size tops at **3Gb**
```
    http://localhost:8200/v1/transactions 0
      [400] 2 responses (100.00%)
      total 2 responses (0.01 rps)

    410 errors:
      [340] Post http://localhost:8200/v1/transactions: net/http: request canceled (Client.Timeout exceeded while awaiting headers)
      [54]  Post http://localhost:8200/v1/transactions: EOF
      [12]  write: broken pipe
      [4] read: connection reset by peer

    http://localhost:8200/healthcheck 1
      [200] 192 responses (100.00%)
      total 192 responses (0.64 rps)

    36 errors:
      [36]  Get http://localhost:8200/healthcheck: net/http: request canceled (Client.Timeout exceeded while awaiting headers)
```

This PR is not correct yet, it needs some more nil validations here and there, but is good enough to get an idea of how will it look like and perform.

Any questions and comments more than welcome.


axw:
(Yesterday I went down a similar route of hand-writing for encoding, in apm-agent-go. Encoding is quite a lot easier, because you know up-front what it should look like, but it's still messy to hand-write the encoding. I abandoned that for now because it's a premature optimisation for the agent.)

IIUC, mapstructure is required only because validation and transformation are split into two stages. Is that correct? Is there a reason why those cannot be combined? If not, perhaps we can consider using [goswagger](https://goswagger.io/). That does mean we would need to rewrite the schema (IMO Swagger/OpenAPI is probably better anyway), but what you get is:

 - model code generation
 - validation code generation, operating on the decoded structs
 - support for [easyjson](https://github.com/mailru/easyjson). easyjson is a code generator that produces static JSON marshal/unmarshal methods, which don't rely on reflection, and are tailored to the types to which they belong

Swagger/OpenAPI specs also define API paths, content-types, status codes; I think the APM API spec would benefit from those too.

jalvz:
What happens is that the validation library (any validation library, I think) works on a `map[string]interface{}`, and the transformation code works on regular structs; so no matter what we need 2 decoding steps:

1. `[]byte` => map  *and* `[]byte` => struct
2. `[]byte` => map => struct
3. `[]byte` => struct = map

I think Silvia found that 2 is better than 1; 2 and 3 are probably quite similar, at the end is 2 passes of copying data & reflection, etc.
Make the transformation code on maps would imply imo that the entire codebase is as messy as the decoding functions in this PR, as you have type assert all the time to access nested attributes, etc.

The swagger idea is quite an undertaking because it would impact the agents, break backwards compatibility, etc.

There is also a technical issue which is that our structs don't match exactly the spec, because we augment the data sent from the agents with a couple of attributes of our own (IP and user-agent for now), and I would guess that with swagger you can not generate a model from the spec, extend that model, and then validate it against the spec (maybe I am wrong).

Still, Im open to the idea... easyjson is very appealing. 

What to the rest think?

axw:
> The swagger idea is quite an undertaking because it would impact the agents, break backwards compatibility, etc.

I'm not a Swagger expert, but I'm not sure about that: Swagger is based on JSON Schema. It doesn't support all of what JSON Schema does, but the APM spec isn't that complicated, so I would be surprised if it doesn't map cleanly.

I'm going to take a quick look to see if it's possible, as I can then use the same code (the marshalling side) from apm-agent-go.

> There is also a technical issue which is that our structs don't match exactly the spec, because we augment the data sent from the agents with a couple of attributes of our own (IP and user-agent for now), and I would guess that with swagger you can not generate a model from the spec, extend that model, and then validate it against the spec (maybe I am wrong).

Possibly a problem, but I don't think so. You could have your own types that embed the model types, extending with the enrichment fields. If you're changing things that are in the model and want your changes to be validated, that should be OK too:  the Validate methods that goswagger generates are separate from unmarshalling, so you can unmarshal, enrich, and then validate.

jalvz:
Ok. I didn't know that, I will bring it up today. Thanks!

axw:
@jalvz here's my first pass on a Swagger spec: https://gist.github.com/axw/db86c1ba241da6c65f5516f54568f5f9

I built the models with goswagger (`swagger generate model`), and ran easyjson to generate static marshal/unmarshal methods. I then ran a little microbenchmark comparison, with some dummy data: a transactions payload with service, system, process and transactions all populated; 100 transactions, each identical, with most fields set.

The first benchmark uses encoding/json.Unmarshal to unmarshal the data into a `map[string]interface{}`. The second benchmark calls the easyjson-generated UnmarshalJSON method on the model struct, *and then calls its Validate method*.

```
BenchmarkUnmarshalMapStdlib-4                500           2481138 ns/op          814791 B/op      16313 allocs/op
BenchmarkUnmarshalSwaggerModel-4            1000           1461362 ns/op          385345 B/op       9651 allocs/op
```

So unmarshalling via easyjson *and* calling the validation logic generated by goswagger uses less memory and CPU than just unmarshalling into `map[string]interface{}` (with no validation).

I suspect this is still not optimal, but I'm not going to spend any more time on this right now. I'm also not sure I'll use this for apm-agent-go now. The generated types are a little ugly, and I'm not yet sure how much the model types will be used directly. If nothing else, I'll probably use easyjson; validation won't be performed agent-side anyway.

simitt:
Thanks for all the work on this @axw, definitly interesting to look into swagger.

PR #687, opened 2018-02-21
Enhance utility methods and get rid of normalization processor in beats.

simitt:
This requires following change in beats: https://github.com/elastic/beats/pull/6419

With this PR we get rid of using the beats normalizer, but do all the necessary normalization steps already in the `utility.MapStrEnhancer`. 

implements https://github.com/elastic/apm-server/issues/711


model/system.go
```diff
@@ -22,5 +22,8 @@ func (s *System) Transform() common.MapStr {
 	utility.Add(system, "platform", s.Platform)
 	utility.Add(system, "ip", s.IP)
 
+	if len(system) == 0 {
```
jalvz: i don't think you need this, `utility.Add` is already handling empty `MapStr`s

processor/transaction/span_context.go
```diff
@@ -0,0 +1,30 @@
...
+func (c *SpanContext) Transform(m common.MapStr) common.MapStr {
+	if m == nil {
+		m = common.MapStr{}
+	} else {
+		for k, v := range m {
+			utility.Add(m, k, v)
+		}
+	}
+	utility.Add(m, "service", c.service)
+	if len(m) == 0 {
```
jalvz: same as before, this should be needed

model/context.go
```diff
@@ -0,0 +1,40 @@
...
+	} else {
+		for k, v := range m {
+			utility.Add(m, k, v)
+		}
+	}
+	utility.Add(m, "service", c.service)
+	utility.Add(m, "process", c.process)
+	utility.Add(m, "system", c.system)
+	utility.MergeAdd(m, "user", c.user)
+	if len(m) == 0 {
```
jalvz: this shouldnt be needed

model/context.go
```diff
@@ -0,0 +1,40 @@
...
+		system:  system.Transform(),
+		user:    user,
+	}
+}
+
+func (c *Context) Transform(m common.MapStr) common.MapStr {
+	if m == nil {
+		m = common.MapStr{}
+	} else {
+		for k, v := range m {
```
jalvz: this reads weird at first sight, can you add a short comment saying that you do this to remove nulls? same for spanContext

utility/map_str_enhancer.go
```diff
@@ -4,44 +4,99 @@ import (
...
 		}
 	case []common.MapStr:
 		if valArr := val.([]common.MapStr); len(valArr) > 0 {
 			m[key] = valArr
+		} else {
+			delete(m, key)
 		}
+	case float64:
+		m[key] = common.Float(val.(float64))
 	default:
```
jalvz: beats normalizer also calls `UTC()` on time objects

processor/transaction/span_context.go
simitt: It is necessary as it checks whether ot not `m` has elements, if this check is removed an empty hash would be returned instead of `nil` and stored in ES.

utility/map_str_enhancer.go
simitt: This is done when decoding time objects, https://github.com/simitt/apm-server/blob/ad8d5d26d243e955cfcb252852a864ee7dac95cb/utility/rfc3399decoder.go#L15. I think this should be enough.

processor/transaction/span_context.go
jalvz: why? aren't you already doing `if len(valMap) > 0 `  in `utility.Add`?

utility/map_str_enhancer.go
```diff
@@ -4,44 +4,99 @@ import (
...
+			if len(valMap) > 0 {
+				m[key] = valMap
+			} else {
+				delete(m, key)
+			}
+		} else {
+			delete(m, key)
+		}
+	case map[string]interface{}:
+		valMap := val.(map[string]interface{})
```
jalvz: this wouldn't handle something like `{"a: {"b": null}}`, right?

utility/map_str_enhancer.go
```diff
@@ -4,44 +4,99 @@ import (
...
 	default:
 		if val != nil {
 			m[key] = val
+		} else {
+			delete(m, key)
+		}
+	}
+}
+
+// MergeAdd modifies `data` *in place*, inserting `values` at the given `key`.
```
jalvz: wrong arg names in comment

utility/map_str_enhancer.go
```diff
@@ -51,3 +106,14 @@ func MillisAsMicros(ms float64) common.MapStr {
 	m["us"] = int(ms * 1000)
 	return m
 }
+
+func AddStrWithDefault(m common.MapStr, key string, val *string, defaultVal string) {
```
jalvz: this is not used anywhere?

utility/rfc3399decoder.go
```diff
@@ -8,7 +8,11 @@ import (
 func RFC3339DecoderHook(f reflect.Type, t reflect.Type, data interface{}) (interface{}, error) {
 	// stolen from https://github.com/mitchellh/mapstructure/issues/41
 	if t == reflect.TypeOf(time.Time{}) && f == reflect.TypeOf("") {
-		return time.Parse(time.RFC3339, data.(string))
+		parsed, err := time.Parse(time.RFC3339, data.(string))
+		if err != nil {
+			return nil, err
+		}
+		return time.Time(parsed.UTC()), nil
```
jalvz: ah, i see now you added the UTC call here. forget the previous comment then

utility/map_str_enhancer.go
simitt: This is not a new function, I only moved it, but you're right, so I'll remove it.

jalvz:
I think would be good to copy over the changes of beats #6419 here so this compiles and pass the tests - they will be wiped after rebasing anyways.
That way we could also do some performance tests on this as is, I think is going to be a very noticeable jump.

utility/map_str_enhancer.go
jalvz: as a reminder, if you don't replace the maps with `MapStr` in the decoding in this PR (which is ok, this already big), then I think we have to do it as soon as this is merged - is not great to have duplicated functions like this.

processor/transaction/span_context.go
simitt: `m` is the base map to which everything is added. If nothing is added `nil` should be returned.

utility/map_str_enhancer.go
simitt: It should, as I iterate through all the elements (see below). 
I added a spec to show this case though.

processor/transaction/span_context.go
jalvz: ok, i don't get it. i commented out all the `if len(m) == 0` in this PR, and all the approval tests pass. Can you add an approval test that exercise this code path? I don't understand how affects the final result.

simitt:
I did have it copied locally and did some performance improvements - the beats PR is approved, so we should have it in soon.

CHANGELOG.asciidoc
```diff
@@ -17,6 +17,7 @@ https://github.com/elastic/apm-server/compare/4daa36bd5c144cf9182afc62dc8042af66
 ==== Added
 
 - Enriched data with IP and UserAgent {pull}393[393].
+- Remove usage of beats normalizer for beat.Events {pull}687[687]
```
jalvz: this is an implementation detail, i don't think is something users need to be aware of.

utility/map_str_enhancer.go
jalvz: ah ofc 🤦‍♂️ sorry

processor/transaction/span_context.go
simitt: you're right that this is checked when the `context` is added to the events, removed it.

jalvz:
thanks for https://github.com/elastic/apm-server/pull/687/commits/bf20be3c42054ee656ba6d7dc26466ad7738f09b
i could reproduce ~15% improvement, which was expected.

simitt:
I'd like to dig a little bit more into how the data parts that do not go through a Go struct are stored in ES before merging this.

simitt:
@elastic/apm-server this is ready for final review.

roncohen:
@simitt I'm investigating something at the moment, please hold on merging

roncohen:
with #709 in I'm OK to merge this

simitt:
@roncohen I've rebased to also include the ES integration test now, as you can see one empty hash is removed, otherwise it stays the same. (see https://github.com/elastic/apm-server/pull/687/files#diff-f63f8f194850ee66420e85c23183a538R122)

simitt:
jenkins, retest this please

PR #688, opened 2018-02-21
Use embedded structs over nested structs.

simitt:


CHANGELOG.asciidoc
```diff
@@ -17,6 +17,7 @@ https://github.com/elastic/apm-server/compare/4daa36bd5c144cf9182afc62dc8042af66
 ==== Added
 
 - Enriched data with IP and UserAgent {pull}393[393].
+- Use embedded structs instead of nested structs {pull}688[688].
```
graphaelli: a little worried that putting a changelog entry in for this implies the go API is stable - is that something we want?

simitt: you're right

simitt:
I checked if it had a performance impact to switch, while it doesn't I like the style better. (So no big reason).

roncohen:
I don't understand this. Why did we choose to embed some of these? To me, embedding should be used when you want the methods of the embedded structs to be available on the embedder.

graphaelli:
I agree with that as motivation for switching to embedding.  I approved this because I would have defaulted to writing to this way in the first place - I think it's less surprising to define the structs this way.

roncohen:
I still don't understand :) 
Did we make the change because we wanted to expose the methods on the embedder?

Also, I think we missed this?

https://github.com/elastic/apm-server/blob/60c75826aa09a8c8ea6b758a1ca3956fbd8b94ea/processor/transaction/event.go#L21

simitt:
@roncohen you're right that the `SpanCount` is missing.

Regarding motivation:
Embedded structs are used to expose methods on the embedder, but as the [docs:](https://golang.org/doc/effective_go.html#embedding) state:

> Embedding can also be a simple convenience. 

I didn't have any strong opinion about whether or not embedded or nested structs should be used, but I changed to use embedded structs while checking for performance improvements, and then decided to create a PR anyways, as I preferred this notation.

If you prefer it the other way it is also fine with me to revert this commit, as I didn't really find any best practice docs for this.

roncohen:
As far as I can tell, the convenience they mention in the docs still relate to making the methods available on the embedder - which is not what we're trying to do here, as far as i can see:

——

Embedding can also be a simple convenience. This example shows an embedded field alongside a regular, named field.

    type Job struct {
        Command string
        *log.Logger
    }

The Job type now has the Log, Logf and other methods of *log.Logger. We could have given the Logger a field name, of course, but it's not necessary to do so. And now, once initialized, we can log to the Job:

    job.Log("starting now...")

———

To me it seems like embedding should be done when you decide it makes sense to have the methods of the embedded on the embedder. If you agree that it's not what motivated us here, then I suggest we revert this.

PR #690, opened 2018-02-21
[6.x] Use type integer instead of number where applicable

felixbarny:
 - Use type integer instead of number where applicable (#641)


simitt:
jenkins, test this please

PR #700, opened 2018-02-26
Add security note

jalvz:


utility/ip.go
```diff
@@ -6,6 +6,9 @@ import (
 	"strings"
 )
 
+// Obtains the IP of a request, looking up X-Forwarded-For and X-Real-IP headers.
+// X-Forwarded-For has a list of IPs, of which the first is the one of the original client.
+// This value however might not be necessarily trusted, as it can be forged by a malicious user.
```
simitt: Would it make sense to also add a note to the description for `user.ip` and `system.ip` in the [fields.yml](https://github.com/elastic/apm-server/blob/master/_meta/fields.common.yml)?

jalvz: good idea

PR #704, opened 2018-02-28
graceful shutdown of apm server

### LINKED ISSUE #671: Fix bug preventing server from being stopped.
When a termination cmd is sent to the server, but the memory queue still buffers some amount of docs to be writtern to the output, the server doesn't stop, nor can it drain the queue. 
With just a quick investigation [this](https://github.com/elastic/apm-server/blob/master/beater/pub.go#L80) seems to be causing the issue.

jalvz:
Fixes #671

This uses the `apm-server.shutdown_timeout` config option to wait for the events buffer to drain, instead of as a timeout for the HTTP server. The HTTP server is closed immediately, instead.

### DIFF
--- a/beater/beater.go
+++ b/beater/beater.go
@@ -18,6 +18,7 @@ type beater struct {
 	mutex   sync.Mutex // guards server and stopped
 	server  *http.Server
 	stopped bool
+	logger  *logp.Logger
 }
 
 // Creates beater
@@ -41,23 +42,23 @@ func New(b *beat.Beat, ucfg *common.Config) (beat.Beater, error) {
 	bt := &beater{
 		config:  beaterConfig,
 		stopped: false,
+		logger:  logp.NewLogger("beater"),
 	}
 	return bt, nil
 }
 
 func (bt *beater) Run(b *beat.Beat) error {
 	var err error
-	logger := logp.NewLogger("beater")
 
-	pub, err := newPublisher(b.Publisher, bt.config.ConcurrentRequests)
+	pub, err := newPublisher(b.Publisher, bt.config.ConcurrentRequests, bt.config.ShutdownTimeout)
 	if err != nil {
 		return err
 	}
-	defer pub.Stop(bt.config.ShutdownTimeout)
+	defer pub.Stop()
 
 	lis, err := net.Listen("tcp", bt.config.Host)
 	if err != nil {
-		logger.Errorf("failed to listen: %s", err)
+		bt.logger.Errorf("failed to listen: %s", err)
 		return err
 	}
 	go notifyListening(bt.config, pub.Send)
@@ -73,7 +74,7 @@ func (bt *beater) Run(b *beat.Beat) error {
 
 	err = run(bt.server, lis, bt.config)
 	if err == http.ErrServerClosed {
-		logger.Infof("Listener stopped: %s", err.Error())
+		bt.logger.Infof("Listener stopped: %s", err.Error())
 		return nil
 	}
 	return err
@@ -81,7 +82,8 @@ func (bt *beater) Run(b *beat.Beat) error {
 
 // Graceful shutdown
 func (bt *beater) Stop() {
-	logp.NewLogger("beater").Info("stopping apm-server...")
+	bt.logger.Infof("stopping apm-server... waiting maximum of %v seconds for queues to drain",
+		bt.config.ShutdownTimeout.Seconds())
 	bt.mutex.Lock()
 	if bt.server != nil {
 		stop(bt.server)
--- a/beater/beater.go
+++ b/beater/beater.go
@@ -47,16 +47,17 @@ func New(b *beat.Beat, ucfg *common.Config) (beat.Beater, error) {
 
 func (bt *beater) Run(b *beat.Beat) error {
 	var err error
+	logger := logp.NewLogger("beater")
 
 	pub, err := newPublisher(b.Publisher, bt.config.ConcurrentRequests)
 	if err != nil {
 		return err
 	}
-	defer pub.Stop()
+	defer pub.Stop(bt.config.ShutdownTimeout)
 
 	lis, err := net.Listen("tcp", bt.config.Host)
 	if err != nil {
-		logp.Err("failed to listen: %s", err)
+		logger.Errorf("failed to listen: %s", err)
 		return err
 	}
 	go notifyListening(bt.config, pub.Send)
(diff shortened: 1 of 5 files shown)


beater/pub.go
```diff
@@ -88,13 +82,10 @@ func (p *publisher) drain() {
...
 func (p *publisher) Send(batch []beat.Event) error {
-	p.m.RLock()
-	defer p.m.RUnlock()
-	if p.stopped {
-		return errChanneClosed
-	}
 
 	select {
+	case <-p.done:
+		return errChanneClosed
```
roncohen: I think this could panic. Try this:
```
package main

import (
    "fmt"
)

func main() {
    c := make(chan int, 10)
    d := make(chan int, 10)
    close(d)
    close(c)

    for i := 0; i < 10; i++ {
        fmt.Println(i)
        select {
        case <- c:
        case d <- 0:
        }
    }
    fmt.Println("OK")
}
```

beater/pub.go
```diff
@@ -50,34 +47,31 @@ func newPublisher(pipeline beat.Pipeline, N int) (*publisher, error) {
...
 
-// drain signals the publisher to stop accepting event batches, blocking
-// execution until all the queued events have been published.
-func (p *publisher) drain() {
-	p.m.Lock()
-	p.stopped = true
-	p.m.Unlock()
-	for {
+// drain waits until the events queue is empty or apm-server.shutdown_timeout seconds have elapsed
+func (p *publisher) drain(d time.Duration) {
```
simitt: I think we don't need to check for the queue length at all, as we use the `beat.Client`. If we add the `bt.config.timeout` to the `beat.ClientConfig` in the `newPublisher` then the proper draining of the queue before shutting down should be handled by the beats framework, like:

```
func newPublisher(pipeline beat.Pipeline, N int, timeout time.Duration) (*publisher, error) {
        ...

	client, err := pipeline.ConnectWith(beat.ClientConfig{
		PublishMode: beat.GuaranteedSend,

		// TODO: We want to wait for events in pipeline on shutdown?
		//       If set >0 `Close` will block for the duration or until pipeline is empty
		WaitClose: timeout,
	})
	        ...
}
```

beater/pub.go
jalvz: I don't think that is equivalent. Or at least I am not able to make the apm server panic.
Of course it still might be this is wrong even if i can't reproduce the panic.

beater/pub.go
jalvz: Interesting, didn't notice that. But putting the timeout there doesn't seem to give the expected result...

simitt: I used this recently for shutting down the server and it always behaved as expected - what was the issue you run into with it?

jalvz: yeah, the timeout is effective, but within this PR it panics. 
this is the simplest approach i came up with that seeems to work; is there any other way to close the server properly that doesn't add more timeouts, locks and code in general?

(i actually don't even know what is the point is the shutdown timeout. for what reasons might be the apm server stopped? does it matter if we have drain the queue or not? do we expect users to stop their agents before stopping the server?)

simitt: I didn't try it in the context of this PR, but in addition what we currently have in master (and removing manually draining the queue).

roncohen: There are two different queues in play here. APM Servers "own" queue `p.events` and the libbeat's queue. `WaitClose: timeout` affects only the libbeat queue.

One of the reasons for draining the queue is when you want to do a rolling upgrade to your APM Servers. The process is like this:

1. You take the first APM Server out of the load balancer so it doesn't get any more requests. 
1. You signal the APM Server to shut down when it's finished delivering all the data it has queued up. If it takes longer than some timeout, we want it to just drop the data and exit.
1. You restart the APM Server, put it back in the load balancer
1. Do it to the next APM Server

beater/pub.go
roncohen: OK. Did you find anything in the code that indicates this is not equivalent? It's important that we understand whether it's not panicking because we were just unlucky or there could be some subtlety that makes our code different from that example.

beater/beater.go
```diff
@@ -47,16 +47,17 @@ func New(b *beat.Beat, ucfg *common.Config) (beat.Beater, error) {
 
 func (bt *beater) Run(b *beat.Beat) error {
 	var err error
+	logger := logp.NewLogger("beater")
```
roncohen: There are two identical calls to `logp.NewLogger("beater")`. I suggest we put this logger on the beater instead. WDYT?

beater/pub.go
jalvz: I don't think is a subtlety, more like the complexity of the apm server is not translatable to that example.
But seeing that Silvia's suggestion works and it uses an attribute that is meant for that already, i'd do it that way. It has a lock, but 1 lock more / 1 lock less doesn't matter much at this point.

roncohen:
cool @jalvz. 

This only deals with the beats queue. It seems to me that we can have a situation where there are elements in `p.events` that have not yet been put into the beats queue. So the beats queue could be empty but there are events in `p.events`. In these situations, the server could close prematurely. Ideally, we'd wait to make sure both queues are empty or the timeout has been reached, before we shut down. We can move forward with this now though and look into waiting for the other queue to also be empty in a separate PR.

A small thing I saw while trying this out:

```
^C
2018-03-12T07:19:27.672+0100	INFO	[beater]	beater/beater.go:84	stopping apm-server...
2018-03-12T07:19:27.672+0100	INFO	[beater]	beater/beater.go:76	Listener stopped: http: Server closed
[...]
2018-03-12T07:19:32.676+0100	INFO	[monitoring]	log/log.go:133	Uptime: 2m1.785253392s
2018-03-12T07:19:32.676+0100	INFO	[monitoring]	log/log.go:110	Stopping metrics logging.
2018-03-12T07:19:32.676+0100	INFO	instance/beat.go:315	apm-server stopped.
```

for a while, it hangs there after you sent the CTRL+C. It would be very nice if it logged out the max wait time:`stopping apm-server. Waiting maximum 5 seconds for queues to drain".

beater/pub.go
```diff
@@ -42,7 +41,7 @@ func newPublisher(pipeline beat.Pipeline, N int) (*publisher, error) {
 
 		// TODO: We want to wait for events in pipeline on shutdown?
 		//       If set >0 `Close` will block for the duration or until pipeline is empty
```
simitt: Could you please remove the TODO

jalvz:
jenkins retest this please

PR #706, opened 2018-02-28
Push transactions and errors to different indices

### LINKED ISSUE #279: Split up indices for Errors and Transactions
We should discuss if we want to generate seperate indices for `errors` and for `transactions`. The documents are semantically different and splitting up indices would allow easier handling in some cases.

jalvz:
fixes #279

### DIFF
--- a/_meta/beat.yml
+++ b/_meta/beat.yml
@@ -79,6 +79,14 @@ output.elasticsearch:
       when.contains:
         processor.event: "sourcemap"
 
+    - index: "apm-%{[beat.version]}-%{+yyyy.MM.dd}-error"
+      when.contains:
+        processor.name: "error"
+
+    - index: "apm-%{[beat.version]}-%{+yyyy.MM.dd}-transaction"
+      when.contains:
+        processor.name: "transaction"
+
   # Optional protocol and basic auth credentials.
   #protocol: "https"
   #username: "elastic"
--- a/_meta/beat.yml
+++ b/_meta/beat.yml
@@ -79,11 +79,11 @@ output.elasticsearch:
       when.contains:
         processor.event: "sourcemap"
 
-    - index: "apm-%{[beat.version]}-%{+yyyy.MM.dd}-error"
+    - index: "apm-%{[beat.version]}-error-%{+yyyy.MM.dd}"
       when.contains:
         processor.name: "error"
 
-    - index: "apm-%{[beat.version]}-%{+yyyy.MM.dd}-transaction"
+    - index: "apm-%{[beat.version]}-transaction-%{+yyyy.MM.dd}"
       when.contains:
         processor.name: "transaction"
 
--- a/tests/system/apmserver.py
+++ b/tests/system/apmserver.py
@@ -306,8 +306,8 @@ class SplitIndicesTest(ElasticTest):
     def setUpClass(cls):
         super(SplitIndicesTest, cls).setUpClass()
 
-        cls.index_name_transaction = "test-apm-12-12-2017-transaction"
-        cls.index_name_error = "test-apm-12-12-2017-error"
+        cls.index_name_transaction = "test-apm-transaction-12-12-2017"
+        cls.index_name_error = "test-apm-error-12-12-2017"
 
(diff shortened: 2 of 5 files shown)


jalvz:
jenkins retest this please

_meta/beat.yml
```diff
@@ -79,6 +79,14 @@ output.elasticsearch:
       when.contains:
         processor.event: "sourcemap"
 
+    - index: "apm-%{[beat.version]}-%{+yyyy.MM.dd}-error"
```
roncohen: What do you think about doing 
`apm-%{[beat.version]}-error-%{+yyyy.MM.dd}`
or 
`apm-error-%{[beat.version]}-%{+yyyy.MM.dd}`
instead?

Having the date at the end seems more natural for me. It would mean that when sorting alphabetically, you'll get them grouped by version and data type first, and then date.

tests/system/config/apm-server.yml.j2
```diff
@@ -67,6 +67,17 @@ output.elasticsearch:
       when.contains:
         processor.event: "sourcemap"
   {% endif %}
+
+  {% if index_name_transaction %}
+  indices:
+    - index: {{ index_name_transaction }}
+      when.contains:
+        processor.name: "transaction"
+    - index: {{ index_name_error }}
```
simitt: Although this works with the test handling, it would be cleaner to also check `{% if index_name_error %}` is set

tests/system/test_integration.py
```diff
@@ -2,27 +2,12 @@
...
 from apmserver import ElasticTest, ExpvarBaseTest
-from apmserver import ClientSideBaseTest, SmapIndexBaseTest, SmapCacheBaseTest
+from apmserver import ClientSideBaseTest, SmapIndexBaseTest, SmapCacheBaseTest, SplitIndicesTest
 from beat.beat import INTEGRATION_TESTS
 
 
 class Test(ElasticTest):
 
-    @unittest.skipUnless(INTEGRATION_TESTS, "integration test")
-    def test_onboarding_doc(self):
```
simitt: I liked having the explicit `test_onboarding_doc`, and not test it implicit within the `test_load_docs_with_template_and_add_transaction`.

tests/system/apmserver.py
```diff
@@ -212,7 +215,7 @@ def check_for_no_smap(self, doc):
             assert "sourcemap" not in frame
 
 
-class ClientSideBaseTest(ServerBaseTest):
+class ClientSideBaseTest(ElasticTest):
```
simitt: With this change, also the [ClientSideTest](https://github.com/jalvz/apm-server/blob/905aef392efc1a6cd07f86ac8bcb5a6fe3cd4a86/tests/system/test_requests.py#L136) and the [RateLimitTest](https://github.com/jalvz/apm-server/blob/905aef392efc1a6cd07f86ac8bcb5a6fe3cd4a86/tests/system/test_requests.py#L214) will be derived from `ElasticTest` which includes the index creation and template loading to ES and slows down tests.

jalvz: yeah, but multiple inheritance doesn't play well here with the way in which we set configuration attributes because if they conflict you overwrite the configuration from one base class with the other one (i actually ran into that). so i think is better to have always 1 base class for simplicity, even if it is slower (it takes ages anyway).

pushed e8a93f982cb21468234142d6bb959aa5bec99cf7 (and 0d14abc258a444419998566e679d92fb6f113c3d)

simitt: ok

simitt:
Please check with @elastic/apm-ui that this does not break anything in the curated UI.

sqren:
Thanks for the heads up @simitt. I use `apm*` as index pattern, so AFAICT this still plays well with the new ones (`apm-%{[beat.version]}-transaction-%{+yyyy.MM.dd}` and `apm-%{[beat.version]}-error-%{+yyyy.MM.dd}`).

jalvz:
I did check manually that this works with dashboards indeed 👍

PR #707, opened 2018-03-08
Update beats framework to cefe024

graphaelli:
* fixes package build by moving to debian 7 base image


graphaelli:
`make package` succeeds locally still

roncohen:
thanks @graphaelli !

PR #709, opened 2018-03-09
Add integration test for ES documents.

simitt:
Verify ES documents for transactions, spans and errors
against gold documents. Approved files need to be adapted
whenever test files are changed.

@roncohen after adding this I can check that PR #687 does not change ES docs.

### DIFF
--- a/tests/system/test_integration.py
+++ b/tests/system/test_integration.py
@@ -4,6 +4,7 @@ import unittest
 from apmserver import ElasticTest, ExpvarBaseTest
 from apmserver import ClientSideBaseTest, SmapIndexBaseTest, SmapCacheBaseTest
 from beat.beat import INTEGRATION_TESTS
+import json
 
 
 class Test(ElasticTest):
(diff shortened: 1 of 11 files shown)


tests/system/test_integration.py
```diff
@@ -60,8 +71,36 @@ def test_load_docs_with_template_and_add_error(self):
...
+            rec = rec_entry['_source']
+            rec_id = rec[doc_type]['id']
+            for appr_entry in approved:
+                appr = appr_entry['_source']
+                if rec_id == appr[doc_type]['id']:
+                    checked = True
+                    self.assert_docs(rec[doc_type], appr[doc_type])
+                    self.assert_docs(rec['context'], appr['context'])
+                    self.assert_docs(rec['@timestamp'], appr['@timestamp'])
+                    self.assert_docs(rec['processor'], appr['processor'])
```
roncohen: could we just compare the docs in full instead of the individual keys `context`, `@timestamp` and `processor` ? for example, what about `marks` ?

roncohen:
great work @simitt !

tests/system/test_integration.py
simitt: `marks` are part of `context` 
comparing the docs in full does not work as e.g. `_id` is transient.

PR #712, opened 2018-03-09
Remove enum for error.log.level

### LINKED ISSUE #665: Consider removing `enum` limitation on log level in JSON Schema
At the moment, the log level is limited to a fairly concise list of levels: https://github.com/elastic/apm-server/blob/4feeae5a028b8f6626f839b7e44edba033541c8e/docs/spec/errors/error.json#L66

This mostly works for Python, where the available log levels are [well defined](https://docs.python.org/3/library/logging.html#logging-levels) and seldom changed. But other logging frameworks allow easy customization of log levels, e.g. [Log4j](https://logging.apache.org/log4j/2.x/manual/customloglevels.html).

I don't think that the X-Pack UI depends on the log levels being restricted. Please correct me if I'm wrong @sqren @formgeist.

simitt:
Adds more flexibility to support custom log levels,
while ES mapping is still keyword.

closes #665

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -20,6 +20,7 @@ https://github.com/elastic/apm-server/compare/4daa36bd5c144cf9182afc62dc8042af66
 
 - Enriched data with IP and UserAgent {pull}393[393], {pull}701[701].
 - Push errors and transactions to different ES indices {pull}706[706].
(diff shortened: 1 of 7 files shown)


simitt:
I don't think this has any side effects, as we simply allow more values. As long as the agents send the so far specified enums, those enums are still shown, without any changes in ES.

PR #718, opened 2018-03-12
[6.1] Update beats framework to d8326d1

graphaelli:
* fix packaging issues


vendor/vendor.json
```diff
@@ -6,269 +6,275 @@
...
 			"path": "github.com/elastic/beats/libbeat/autodiscover",
-			"revision": "09f84bfbb817e086b8d7ac12bf567c6c47046e87",
-			"revisionTime": "2018-01-31T10:29:17Z",
+			"revision": "d8326d125db9dff73d223e51e2cc36114ac75b2a",
+			"revisionTime": "2018-03-11T15:50:53Z",
+			"version": "6.1",
+			"versionExact": "6.1"
+		},
+		{
+			"path": "github.com/elastic/beats/libbeat/autodiscover/builder",
```
simitt: This entry looks a bit odd, not having a checksumSHA1

graphaelli: hmm, true.  A bunch of the libbeat packages are missing that key

PR #730, opened 2018-03-16
Use manual decoding for performance improvements.

simitt:
Use hardcoded decodig for `transactions` and `errors` to improve overall performance. 

This implementation is based on @jalvz POC for manual decoding performance increase.

@elastic/apm-server I haven't add more tests yet, but the existing ones which are pretty exhaustive are all passing.
Early feedback on how I structured the decoding is highly appreciated. 

Performance Results for using the same settings for all 4 branches (`apm-server -N`):

(update: changed numbers to final measurements)

master:  90 rps
change-decoding: 230 rps
enhance-decoding: 350 rps
jalvz-original manual decoding: 361 rps 

Fixes https://github.com/elastic/apm-server/issues/733


model/process.go
```diff
@@ -1,17 +1,35 @@
...
 	Argv  []string
 }
 
+func (p *Process) Decode(input interface{}) error {
+	raw, ok := input.(map[string]interface{})
+	if raw == nil {
+		return nil
+	}
+	if !ok {
+		return errors.New("Invalid type for process")
```
jalvz: i think this line can not be ever executed, if type assert fails `raw` will be `nil` and function will return 2 lines above.

simitt: found and fixed that, it should be `if input == nil..`

model/service.go
```diff
@@ -32,6 +34,29 @@ type Agent struct {
 	Version string
 }
 
+func (s *Service) Decode(input interface{}) error {
+	raw, ok := input.(map[string]interface{})
+	if raw == nil {
+		return nil
+	}
+	if !ok {
+		return errors.New("Invalid type for service")
```
jalvz: same here, in case you didn't fix already

model/stacktrace.go
```diff
@@ -1,46 +1,58 @@
...
-// - lineno
-// - abs_path is set to the cleaned abs_path
-// - sourcmeap.updated is set to true
-func (st Stacktrace) sourcemapAndTransform(config *pr.Config, service Service) []common.MapStr {
-	frameCount := len(st)
+func (st *Stacktrace) Transform(config *pr.Config, service Service) []common.MapStr {
+	if st == nil {
+		return nil
+	}
+	// source map algorithm:
```
jalvz: why this comment is here? i think it fits better as `applySourcemap`

processor/error/event.go
```diff
@@ -180,17 +242,19 @@ func (e *Event) calcGroupingKey() string {
...
-		if st == nil || len(st) == 0 {
+		if st.Frames == nil || len(st.Frames) == 0 {
 			st = e.Log.Stacktrace
 		}
 	}
 
-	for _, fr := range st {
-		if fr.ExcludeFromGrouping {
-			continue
+	if st.Frames != nil {
```
jalvz: i think you don't need this

model/stacktrace.go
simitt: because the comment describes the sourcemapping for the whole stacktrace and not just for a single frame.

processor/error/payload.go
```diff
@@ -18,9 +19,45 @@ type payload struct {
...
 	Process *m.Process
-	User    common.MapStr
+	User    *m.User
 
-	Events []Event `mapstructure:"errors"`
+	Events []Event
+}
+
+func (pa *payload) decode(raw map[string]interface{}) error {
+	if err := pa.Service.Decode(raw["service"]); err != nil {
```
jalvz: I think all this `if` boilerplate is much simplified if you use functions instead of methods; and propagate error through calls.

eg,  for system:

```
func DecodeSystem(input interface{}, err error) (*System, error) {
	if err != nil {
		return nil, err
	}
	raw, _ := input.(map[string]interface{})
	df := utility.DataFetcher{}
	return &System{
		Hostname: df.StringPtr(raw, "hostname"),
		Platform: df.StringPtr(raw, "platform"),
		Architecture: df.StringPtr(raw, "architecture"),
		IP: df.StringPtr(raw, "ip"),
	}, df.Err
}
```

Then in `decode` you can do just like this:

```
var err error
pa.Service, err = m.DecodeService(raw["system"], err)
pa.System, err = m.DecodeSystem(raw["system"], err)
pa.Process, err = m.DecodeProcess(raw["process"], err)
...
return err
```

utility/common.go
```diff
@@ -33,3 +33,14 @@ func InsertInMap(data map[string]interface{}, key string, values map[string]inte
 	}
 
 }
+
+func GetDeep(raw map[string]interface{}, keys ...string) map[string]interface{} {
```
jalvz: I think it is better GetDeep as a function than as a method, but right now they are duplicated (leftover i guess) so this one should be removed

utility/data_fetcher.go
```diff
@@ -0,0 +1,167 @@
...
+
+type DataFetcher struct {
+	Err error
+}
+
+func (d *DataFetcher) Float64(base map[string]interface{}, key string, keys ...string) float64 {
+	val := d.getDeep(base, keys...)[key]
+	if valFloat, ok := val.(float64); ok {
+		return valFloat
+	} else if val == nil {
```
jalvz: I think this case is not needed, neither `nilErr` error (attribute required-ness already happens in json schema validation)

utility/data_fetcher.go
```diff
@@ -0,0 +1,167 @@
...
+		d.Err = nilErr
+	} else {
+		d.Err = typeErr
+	}
+	return 0.0
+}
+
+func (d *DataFetcher) IntPtr(base map[string]interface{}, key string, keys ...string) *int {
```
jalvz: I think this should be enough:

```
func (d *DataFetcher) IntPtr(base map[string]interface{}, key string, keys ...string) *int {
	val := d.getDeep(base, keys...)[key]
	if val == nil {
		return nil
	} else if valFloat, ok := val.(float64); ok || val == nil {
		valInt := int(valFloat)
		return &valInt
	}
	 d.Err = typeErr
	return nil
}
```

utility/data_fetcher.go
```diff
@@ -0,0 +1,167 @@
...
+	d.Err = typeErr
+	return nil
+}
+
+func (d *DataFetcher) Interface(base map[string]interface{}, key string, keys ...string) interface{} {
+	return d.getDeep(base, keys...)[key]
+}
+
+func (d *DataFetcher) Int(base map[string]interface{}, key string, keys ...string) int {
+	if val := d.IntPtr(base, key, keys...); val != nil {
```
jalvz: this should be enough:
```return *d.IntPtr(base, key, keys...)```

File: utility/data_fetcher.go [hunk 10]
```diff
@@ -0,0 +1,167 @@
...
+		return nil
+	} else if valStr, ok := val.(string); ok {
+		return &valStr
+	}
+	d.Err = typeErr
+	return nil
+}
+
+func (d *DataFetcher) String(base map[string]interface{}, key string, keys ...string) string {
+	if val := d.StringPtr(base, key, keys...); val != nil {
```
jalvz: this should be enough:
```
return *d.StringPtr(base, key, keys...)
```

utility/data_fetcher.go
```diff
@@ -0,0 +1,167 @@
...
+		d.Err = nilErr
+	}
+	return ""
+}
+
+func (d *DataFetcher) StringArr(base map[string]interface{}, key string, keys ...string) []string {
+	val := d.getDeep(base, keys...)[key]
+	if val == nil {
+		return nil
+	} else if valArr, ok := val.([]string); ok {
```
jalvz: i think this case is not needed

jalvz:
remember to update the notice file

utility/data_fetcher.go
simitt: You're right that this should be caught in the validation already. However, this method doesn't know anything about a validation, and I'd rather keep it this way, to still have a defined handling in case the validation changes or a bug gets introduced.

utility/data_fetcher.go
simitt: `d.IntPtr` could be `nil` here, so I cannot just return `*d.IntPtr`.

utility/data_fetcher.go
simitt: same as above

utility/data_fetcher.go
simitt: I'd prefer leaving it in as it would be really unexpected behavior when querying for a StringArray would not return anything when the value for the specified key is a `[]string`

utility/data_fetcher.go
jalvz: The goal of validation is not that you can make assumptions about the data later on? 

We are making assumptions already, explicitly or implicitly (eg: we assume we can save this data in ES conforming its mapping because we have validated it). So this double-checking is actually redundant.

utility/data_fetcher.go
jalvz: As reply for this and `Int` / `String` implementations above:
In apm-server data is meaningful - we know what is a transaction, error, span etc. 
Why have so much generic code as if we don't know what data is?

This case is never executed in production, yet you took the time to write it and test, me to read it, the compiler to compile it, etc...

utility/data_fetcher.go
jalvz: That is fair if you want to make sure it doesn't panic, but what difference makes the 
`val := d.getDeep(base, keys...)[key]` line?
couldn't just return 0 if val is nil?

simitt: oh you were referring to this line, that I already changed, just haven't pushed yet.

utility/data_fetcher.go
simitt: Not checking for a nil or a type error could lead to storing incomplete data in ES as we would go on with the processing, although some value that should be given might be missing.

I'd rather keep this check in than taking the risk of having unexpected data in ES.
The validation implementation is completely separated from this implementation, and as long as we do not get the gostructs as a result out of the validation, I'm afraid we have to do the same validations twice, or risking that the logic can drift apart. 

Regarding assumptions about being able to store the data in ES - we do have integration tests, that ensure the data can actually be stored, so I don't really get your point here.

simitt:
@elastic/apm-server this is ready for review now

processor/error/payload.go
```diff
@@ -18,9 +19,40 @@ type payload struct {
...
+		pa.Service = *service
+	}
+	pa.System, err = m.DecodeSystem(raw["system"], err)
+	pa.Process, err = m.DecodeProcess(raw["process"], err)
+	pa.User, err = m.DecodeUser(raw["user"], err)
+	if err != nil {
+		return nil, err
+	}
+
+	df := utility.DataFetcher{Err: err}
```
jalvz: if `err != nil`, you are returning in L41, so here Err: is going to be always nil, so you don'y need to pass it.

processor/error/event.go
```diff
@@ -16,35 +17,96 @@ import (
...
+		e.Exception = &Exception{
+			Message:    *exMsg,
+			Code:       df.Interface(ex, "code"),
+			Module:     df.StringPtr(ex, "module"),
+			Attributes: df.Interface(ex, "attributes"),
+			Type:       df.StringPtr(ex, "type"),
+			Handled:    df.BoolPtr(ex, "handled"),
+			Stacktrace: m.Stacktrace{},
+		}
+		str, err = m.DecodeStacktrace(ex["stacktrace"], err)
```
jalvz: nitpick: when i read `str = ...`  I thought why `DecodeStacktrace` would return a string... i'd rename this variable :)
(and same for the Log part bellow and Span)

simitt:
Thanks for the thorough review @jalvz , I updated everything.

processor/error/package_tests/TestProcessErrorAugmentedUserAgentAndIP.approved.json
```diff
@@ -12,7 +12,7 @@
                 },
                 "user": {
                     "ip": "127.0.0.1",
-                    "user-agent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_12_5)",
+                    "user_agent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_12_5)",
```
roncohen: Seems like this would be a breaking change. I don't think anyone is using it at the moment, so we can probably move forward.

simitt: It's not a breaking change as it hasn't been released yet.

simitt:
This is inspired by [`json.NewDecoder().Decode(v interface{}) error`](https://golang.org/pkg/encoding/json/#Decoder.Decode).

The idea was to have the full control and responsibility inside the decoding function.

processor/error/package_tests/TestProcessErrorAugmentedUserAgentAndIP.approved.json
simitt: I'd rather see it as a bugfix, as we convert all other values to use '_'.

roncohen:
I see. I think it's different because `Decoder.Decode` must take a `interface{}` because it's generic. It needs to work on arbitrary JSON. It needs to be able to decode an `int` as well as a `map[string]interface{}`. Our functions are not generic. WDYT?

roncohen:
Lets move forward as it is. We can do a refactor subsequently if necessary.

simitt:
I don't have a very strong opinion on this, I don't see a big advantage for one or the other. What would be the benefit of changing it?

simitt:
I'm not fine with just merging it in and then do a refactor again. If you think it is worth the effort and improves the code, I can change it. It is just not clear to me which benefit this would bring.

roncohen:
Generally, functions signatures are for signaling what types the function expects. This helps tooling, other consumers of those functions, safety etc

If we define functions to take `interface{}` and then do type checking inside we're essentially doing work that should be left to the type system. In this particular case, you can argue that we must do the type check _somewhere_ because the data happens to come from a JSON that was decoded to a generic type. So if we need to type check somewhere, one could argue that its better in the individual functions to make the call site clearer. If that's the case, e.g. you feel the code will be much harder to read if we do the type checking on the side of the caller, I'm OK to go forward as it.

jalvz:
i think an argument for doing it in the function because the assertion+check happens only once (inside the function) instead of many times (each caller).

roncohen:
Hm, yeah OK that makes sense for these @jalvz. Let's leave it 👍

roncohen:
There are just two callers, but you could argue that due to the nature of these functions, the scenarios they are used in means the data always come from an `interface{}`. That's what I tried to say here: 

> you can argue that we must do the type check somewhere because the data happens to come from a JSON that was decoded to a generic type. So if we need to type check somewhere, one could argue that its better in the individual functions to make the call site clearer.

simitt:
@roncohen I pushed the changes for the `user-agent` and the renaming.

PR #745, opened 2018-03-21
Use config.Config over processor.Config.

simitt:
* Remove the dependency of the `model` package to the `processor` package. 
* Pass config by value, as it is not going to be changed inside.


processor/error/payload.go
```diff
@@ -55,13 +54,18 @@ func decodeError(raw map[string]interface{}) (*payload, error) {
...
-	context := m.NewContext(&pa.Service, pa.Process, pa.System, pa.User)
-
+func (pa *payload) transform(c pr.Config) []beat.Event {
 	logp.NewLogger("transform").Debugf("Transform error events: events=%d, service=%s, agent=%s:%s", len(pa.Events), pa.Service.Name, pa.Service.Agent.Name, pa.Service.Agent.Version)
-
 	errorCounter.Add(int64(len(pa.Events)))
+
+	context := m.NewContext(&pa.Service, pa.Process, pa.System, pa.User)
+
+	config := m.Config{
```
jalvz: why this if config is passed by value already?

simitt: because it is creating a `model.Config` that can be passed to the transformations, to get rid of the `processor` dependency in  `model`

jalvz: oh, i didn't notice that there is `m.Config` and `pr.Config` which are the same. Having duplicated models to avoid dependencies is not a good pattern.
generally if there is a 2-way dependency A<=>B, they best way to solve it is to extract the common part to a different package C, so it becomes A=>C, and B=>C.

if you don't have time for this i can try to find better way to do it, but duplicated models is not good imo. Passing by value instead of reference is 👍

simitt: The reason behind this is that there might be some more config options for the processor, so I kept the original `processor.Config`. If you don't think this is worth it, we can also just remove this one completely and pass in the `model.Config` already from the `beater`.

jalvz: i think it would be better - if we have more configs options for the processor we can figure it out at the time it happens...

simitt: I still think that it would be fair to create a `processor.Config` from `beater`, as beater doesn't need to know anything about the insights of processing. Processor then extracts the necessary information for processing it and passes it along to the model package in a new model. And right now it happens to be the same information.

However, I created a separate `config` package now, as I don't think calling `model.Config` from `beater` is the right thing to do.

jalvz: i don't have any strong preferences on where things should be, or what functions should know what, so this is fine by me.
i'm not opposing to `processor.Config` neither, just don't want to have duplicated code for no good reason.

config/config.go
```diff
@@ -0,0 +1,13 @@
+package config
+
+import (
+	"regexp"
+
+	"github.com/elastic/apm-server/sourcemap"
+)
+
+type Config struct {
```
jalvz: we have so many configs that i would prefer to have more explicit names for each one, like eg. `FrontendConfig` for this one, but up to you

simitt: I would not change it to `FrontendConfig` with this PR as this was only meant to get rid of package dependencies.
AFAIR the `LibraryPattern` and `ExcludeFromGrouping` could theoretically also be used for backend endpoints as discussed when we introduced these settings, so I would revisit this in a separate PR.

Issue #715, opened 2018-03-09
Return after request was validated


simitt:
Handle transformation async to avoid blocking open connections. 

Goal is to run the `transform` method async as the method cannot return an error.

This requires to return `503` already after the validation.

graphaelli:
> This requires to return 503 already after the validation.

Meaning we need a way to communicate back pressure without performing the transformation?  We'll still return a `400` if the validation itself fails right?  If so, agreed.

simitt:
@roncohen what do you think about building this in for 6.3?

roncohen:
@simitt i think that would be great if possible. Do you already have an idea how we can do this while maintaining the same semantics as now with the regards to full queue, e.g return 503 to new requests that attempt to put things into a full queue? From my understanding of the queue, it's going to be difficult.

simitt:
What do you think about putting it into the queue before the actual transformation happens? 
This way it would be put into the queue at an earlier stage, but should basically keep the current semantics.

roncohen:
sounds like a great idea!

simitt:
closed by https://github.com/elastic/apm-server/pull/746

PR #746, opened 2018-03-21
Early return to caller.

simitt:
* Return payload from decoding
* Add payload to channel and run transformation on payload.
* Free up resources as early as possible

implements #715

With these changes we do return earlier to the caller, which helps freeing up resources in the agents. 
This also means that less processing happens before putting the data into the channel and after fetching data from the channel processing is heavier. This can lead to more `503` as the blocking part for moving data out of the channel gets slower.

### DIFF
--- a/beater/handlers.go
+++ b/beater/handlers.go
@@ -43,7 +43,7 @@ const (
 	supportedMethods = "POST, OPTIONS"
 )
 
-type ProcessorFactory func(conf.Config) processor.Processor
+type ProcessorFactory func() processor.Processor
 
 type ProcessorHandler func(ProcessorFactory, *Config, reporter) http.Handler
 
@@ -302,7 +302,7 @@ func processRequestHandler(pf ProcessorFactory, config conf.Config, report repor
 }
 
 func processRequest(r *http.Request, pf ProcessorFactory, config conf.Config, report reporter, decode decoder.Decoder) (int, error) {
-	processor := pf(config)
+	processor := pf()
 
 	if r.Method != "POST" {
 		return http.StatusMethodNotAllowed, errPOSTRequestOnly
@@ -317,12 +317,12 @@ func processRequest(r *http.Request, pf ProcessorFactory, config conf.Config, re
 		return http.StatusBadRequest, err
 	}
 
-	list, err := processor.Transform(data)
+	payload, err := processor.Decode(config, data)
 	if err != nil {
 		return http.StatusBadRequest, err
 	}
 
-	if err = report(list); err != nil {
+	if err = report(payload); err != nil {
 		return http.StatusServiceUnavailable, err
 	}
 
--- a/beater/handlers.go
+++ b/beater/handlers.go
@@ -317,12 +317,13 @@ func processRequest(r *http.Request, pf ProcessorFactory, config conf.Config, re
 		return http.StatusBadRequest, err
 	}
 
-	payload, err := processor.Decode(config, data)
+	payload, err := processor.Decode(data)
 	if err != nil {
 		return http.StatusBadRequest, err
 	}
 
-	if err = report(payload); err != nil {
+	event := event{payload: payload, config: config}
+	if err = report(event); err != nil {
 		return http.StatusServiceUnavailable, err
 	}
 
--- a/beater/handlers.go
+++ b/beater/handlers.go
@@ -322,8 +322,7 @@ func processRequest(r *http.Request, pf ProcessorFactory, config conf.Config, re
 		return http.StatusBadRequest, err
 	}
 
-	decodedRequest := decodedRequest{payload: payload, config: config}
-	if err = report(decodedRequest); err != nil {
+	if err = report(pendingReq{payload: payload, config: config}); err != nil {
 		return http.StatusServiceUnavailable, err
 	}
 
--- a/beater/handlers.go
+++ b/beater/handlers.go
@@ -322,8 +322,8 @@ func processRequest(r *http.Request, pf ProcessorFactory, config conf.Config, re
 		return http.StatusBadRequest, err
 	}
 
-	event := event{payload: payload, config: config}
-	if err = report(event); err != nil {
+	decodedRequest := decodedRequest{payload: payload, config: config}
+	if err = report(decodedRequest); err != nil {
 		return http.StatusServiceUnavailable, err
 	}
 
--- a/beater/onboarding.go
+++ b/beater/onboarding.go
@@ -8,7 +8,7 @@ import (
 	"github.com/elastic/beats/libbeat/logp"
 )
 
-func notifyListening(config *Config, reporter reporter) {
+func notifyListening(config *Config, pubFct func(beat.Event)) {
 
 	var isServerUp = func() bool {
 		secure := config.SSL.isEnabled()
@@ -22,7 +22,6 @@ func notifyListening(config *Config, reporter reporter) {
 			Timestamp: time.Now(),
 			Fields:    common.MapStr{"listening": config.Host},
 		}
-		events := []beat.Event{event}
-		reporter(events)
+		pubFct(event)
 	}
 }
--- a/beater/pub.go
+++ b/beater/pub.go
@@ -10,24 +10,23 @@ import (
 	"github.com/elastic/apm-server/config"
 	pr "github.com/elastic/apm-server/processor"
 	"github.com/elastic/beats/libbeat/beat"
-	sysinfo "github.com/elastic/go-sysinfo"
 )
 
-// publisher forwards batches of events to libbeat. It uses GuaranteedSend
-// to enable infinite retry of events being processed.
+// publisher forwards batches of decodedRequests to libbeat. It uses GuaranteedSend
+// to enable infinite retry of decodedRequests being processed.
 // If the publisher's input channel is full, an error is returned immediately.
 // Number of concurrent requests waiting for processing do depend on the configured
 // queue size. As the publisher is not waiting for the outputs ACK, the total
-// number requests(events) active in the system can exceed the queue size. Only
+// number requests(decodedRequests) active in the system can exceed the queue size. Only
 // the number of concurrent HTTP requests trying to publish at the same time is limited.
 type publisher struct {
-	events  chan event
-	client  beat.Client
-	m       sync.RWMutex
-	stopped bool
+	decodedRequests chan decodedRequest
+	client          beat.Client
+	m               sync.RWMutex
+	stopped         bool
 }
 
-type event struct {
+type decodedRequest struct {
 	payload pr.Payload
 	config  config.Config
 }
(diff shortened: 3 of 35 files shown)


beater/handlers.go
```diff
@@ -316,14 +316,12 @@ func processRequest(r *http.Request, pf ProcessorFactory, prConfig *processor.Co
...
-	list, err := processor.Transform(data)
+	payload, err := processor.Decode(data)
 	if err != nil {
 		return http.StatusBadRequest, err
 	}
 
-	if err = report(list); err != nil {
-		return http.StatusServiceUnavailable, err
-	}
+	go report(payload)
```
simitt: I removed the usage of the channel and instead use a go routine for further processing, as the check to not overfloat the server with requests is now already done at an earlier stage with @roncohen 's changes.

However, as the `transform` and the `publishAll` is done within this go routine, after releasing the request from the channel, and those actions might be more heavy than the `validation` and `decoding` it could lead to more go routines processing than the number of `concurrent_requests` is set to. 
WDYT?

simitt:
Closing this for now, as it needs more investigation.

simitt:
After some more investigation I updated and reopened the PR. Basic tests show the same throughput as for current master, which is fair as the goal of the PR is to return to the caller asap and not to increase throughput.

beater/pub.go
```diff
@@ -16,16 +18,16 @@ import (
 // number requests(events) active in the system can exceed the queue size. Only
 // the number of concurrent HTTP requests trying to publish at the same time is limited.
 type publisher struct {
-	events  chan []beat.Event
-	client  beat.Client
-	m       sync.RWMutex
-	stopped bool
+	payloads chan pr.Payload
```
roncohen: Having the channel be a `pr.Payload` instead of `*pr.Payload` means we'll be copying all the data. Have you thought about making this channel of pointers instead? It might not make any difference.

beater/pub.go
```diff
@@ -66,30 +68,30 @@ func (p *publisher) Stop() {
...
 	case <-time.After(time.Second * 1): // this forces the go scheduler to try something else for a while
 		return errFull
 	}
 }
 
 func (p *publisher) run() {
-	for batch := range p.events {
-		p.client.PublishAll(batch)
+	for payload := range p.payloads {
+		p.client.PublishAll(payload.Transform())
```
roncohen: This effectively makes transformation single-threaded. In a multi-core machine, we should be able to utilize multiple cores for transformation. 

A way to solve this could be to start multiple go routines (for example, as many as there are cores. We can get the core-count by using `runtime.NumCPU()`). 

Something to be aware of: As far as I remember, beats `Client`s are not safe to use from multiple go routines. Perhaps each go routine could just have its own `Client`.

roncohen:
Good work @simitt - and great that throughput is unchanged. 

Do you have a sense for if the memory usage pattern is changed with this PR? E.g. do we use more memory now that transformation is decoupled from the request/response cycle? Or do we have adequate queueing/limits in place?

processor/sourcemap/payload.go
```diff
@@ -16,23 +16,25 @@ var (
...
 )
 
-type payload struct {
+type Payload struct {
 	ServiceName    string
 	ServiceVersion string
 	Sourcemap      string
 	BundleFilepath string
+
+	config config.Config
```
jalvz: might be a bit fiddly to put configuration and user data in the same struct...

processor/transaction/payload.go
```diff
@@ -17,19 +17,22 @@ var (
...
 )
 
-type payload struct {
+type Payload struct {
 	Service m.Service
 	System  *m.System
 	Process *m.Process
 	User    *m.User
-	Events  []Event
+	Events  []*Event
```
jalvz: not clear why this, slice items are already references

beater/pub.go
simitt: This shouldn't make a big difference, as the Payload mostly consists of Pointers and Slices.

simitt:
Will investigate according to PR comments and reopen.

processor/transaction/payload.go
simitt: Because I am removing the references to the `Event` immediately after processing it. If garbage collection kicks in the Event can already be cleaned up.

https://github.com/elastic/apm-server/blob/83472af331ff13c45f62cd06769dca92a52fdec0/processor/transaction/payload.go#L92-L94

processor/sourcemap/payload.go
simitt: You are right, going to change that.

beater/pub.go
simitt: Looking into the client code at beats, `PublishAll` uses a `c.mutex.Lock()`, so calling this from multiple go routines is not an issue:
https://github.com/elastic/beats/blob/master/libbeat/publisher/pipeline/client.go#L34

simitt:
Thanks @roncohen and @jalvz for the valuable feedback, I adapted the code accordingly.

I compared the behavior of this branch with the current master (local test runs) with following outcome:

(1) heavy load,  default settings: 
~ same throughput, ~ same RAM usage 

(first run`return_early`, second run `master`)
screen shot 2018-04-06 at 12 09 48


(2) heavy load,  `concurrent_requests=512`, `queue.mem.events=512`:
~ same throughput, `return_early` uses less RAM

(first run`return_early`, second run `master`)
screen shot 2018-04-06 at 12 30 28


(3) nodesjs test script,  default settings: 
throughput: `return_early` ~1384 docs per second, `master` ~1187 docs per second
~same RAM usage 

(first run`return_early`, second run `master`)
screen shot 2018-04-06 at 12 44 54

beater/pub.go
```diff
@@ -52,13 +62,27 @@ func newPublisher(pipeline beat.Pipeline, N int, shutdownTimeout time.Duration)
...
+		go p.run()
 	}
 
-	go p.run()
 	return p, nil
 }
 
+func numCPUToUse() int {
+	maxCPU := sysinfo.Go().MaxProcs
+	numCPU := runtime.NumCPU()
```
jalvz: why just `runtime.NumCPU()` is not enough?

beater/pub.go
```diff
@@ -1,11 +1,16 @@
...
+	"runtime"
 	"sync"
 	"time"
 
+	"github.com/pkg/errors"
+
+	"github.com/elastic/apm-server/config"
+	pr "github.com/elastic/apm-server/processor"
 	"github.com/elastic/beats/libbeat/beat"
+	sysinfo "github.com/elastic/go-sysinfo"
```
jalvz: does this works in windows?

beater/pub.go
```diff
@@ -16,21 +21,26 @@ import (
...
 // the number of concurrent HTTP requests trying to publish at the same time is limited.
 type publisher struct {
-	events  chan []beat.Event
+	events  chan event
 	client  beat.Client
 	m       sync.RWMutex
 	stopped bool
 }
 
+type event struct {
```
jalvz: since we have already error events, transactions events, and beat events, would be nice to unclog a bit our names and call this differently...

beater/pub.go
```diff
@@ -16,21 +21,26 @@ import (
...
+	payload pr.Payload
+	config  config.Config
```
jalvz: i think this is a better place to put the config 👍

beater/pub.go
simitt: Because in the `apm-server.yml` you can configure a `max_procs` with description: Sets the maximum number of CPUs that can be executing simultaneously.

If this is set then we should also respect if for the amount of consuming go routines.

jalvz: right. in that case, shouldn't `runtime.GOMAXPROCS(0)` be enough?

simitt: I don't want to set the variable here, as this is already done when starting the apm-server in the beats code, but only want to get the number of configured CPUs.
`sysinfo.Go().MaxProcs` returns exactly what has been set when starting APM-Server. In case this hasn't been set or is an invalid number the default `runtime.NumCPU` is used.

jalvz: precisely. `runtime.GOMAXPROCS(0)` doesn't set anything, just gets you that number.
If user don't specify anything, GOMAXPROCS defaults to num of cpus, so i think you don't need any logic and any test.

simitt: I checked [`sysinfo.Go().MaxProcs`](https://github.com/elastic/go-sysinfo/blob/master/system.go#L37) internas, and as it also calls `runtime.GOMAXPROCS(0)` I am fine with changing it.

I kept the tests, as I still see value in ensuring we create the expected amount of go routines for the transformation and publishing.

beater/onboarding.go
```diff
@@ -8,7 +8,7 @@ import (
 	"github.com/elastic/beats/libbeat/logp"
 )
 
-func notifyListening(config *Config, reporter reporter) {
+func notifyListening(config *Config, pubFct func(e beat.Event)) {
```
graphaelli: only nitpicking since you need to push again to resolve merge conflicts, can remove the `e` named param here

beater/pub.go
```diff
@@ -52,44 +61,47 @@ func newPublisher(pipeline beat.Pipeline, N int, shutdownTimeout time.Duration)
...
 	case <-time.After(time.Second * 1): // this forces the go scheduler to try something else for a while
 		return errFull
 	}
 }
 
 func (p *publisher) run() {
-	for batch := range p.events {
-		p.client.PublishAll(batch)
+	for decodedRequest := range p.decodedRequests {
+		p.client.PublishAll(decodedRequest.payload.Transform(decodedRequest.config))
```
graphaelli: this works but reads strangely (to me).  What do you think of flipping it around to `func (dr *decodedRequest) Publish(client beat.Client)` ?  I'm not sure that's any better.

At least, can we rename the `decodedRequest` iterator to something distinct from the `decodedRequest` struct type?

simitt: The `client.PublishAll([] beat.Event)` is defined in beats, so flipping it here would only defer calling this method, which I think is even more confusing. 
I tried to come up with a better naming though. WDYT?

simitt:
jenkins, retest this please.

beater/pub.go
graphaelli: Yeah, and after sleeping on it, it's even more gross to have the decodedRequest know anything about publishing anyway.  I like rename.

beater/pub.go
```diff
@@ -52,44 +61,47 @@ func newPublisher(pipeline beat.Pipeline, N int, shutdownTimeout time.Duration)
...
 	case <-time.After(time.Second * 1): // this forces the go scheduler to try something else for a while
 		return errFull
 	}
 }
 
 func (p *publisher) run() {
-	for batch := range p.events {
-		p.client.PublishAll(batch)
+	for req := range p.pendingRequests {
+		p.client.PublishAll(req.payload.Transform(req.config))
```
roncohen: could we put the config as property on the publisher so we didn't have to pass it in every `pendingReq` ? It doesn't change know

simitt: https://github.com/elastic/apm-server/pull/746#discussion_r179128363

Actually I also think it is cleaner to separate the config from the payload data.

simitt: Oh sorry, you said on the publisher, I misread. We cannot put it on the publisher, as it is different per endpoint.

watson:
Woot! 🎉Can't wait to see this reduce resource usage in the agents 😃

PR #751, opened 2018-03-21
limit number of new connections to accept simultaneously

### LINKED ISSUE #736: Limit open concurrent tcp connections
The APM Server will happily accept and create a go routine for each new incoming connection. 
https://github.com/elastic/apm-server/pull/731 will ensure that we don't start processing more than `concurrent_requests`, but it should be possible to limit incoming requests in some way. 

Since go routines are quite cheap and users may legitimately want to accept many connections, the default restriction should be to a high number, say 1000.

graphaelli:
Introduce a limit to the number of connections to accept simultaneously.

closes #736

### DIFF
--- a/_meta/beat.reference.yml
+++ b/_meta/beat.reference.yml
@@ -24,6 +24,9 @@ apm-server:
   # Maximum number of requests permitted to be sent to the server concurrently.
   #concurrent_requests: 5
 
+  # Maximum number of new connections to accept simultaneously
+  # max_connections: 0
+
   # Authorization token to be checked. If a token is set here the agents must
   # send their token in the following format: Authorization: Bearer <secret-token>.
   # It is recommended to use an authorization token in combination with SSL enabled.
--- a/_meta/beat.reference.yml
+++ b/_meta/beat.reference.yml
@@ -24,7 +24,7 @@ apm-server:
   # Maximum number of requests permitted to be sent to the server concurrently.
   #concurrent_requests: 5
 
-  # Maximum number of new connections to accept simultaneously
+  # Maximum number of new connections to accept simultaneously (0 means unlimited)
   # max_connections: 0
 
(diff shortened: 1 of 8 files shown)


graphaelli:
note https://github.com/golang/go/issues/24458, I don't think it's a showstopper.

_meta/beat.reference.yml
```diff
@@ -24,6 +24,9 @@ apm-server:
   # Maximum number of requests permitted to be sent to the server concurrently.
   #concurrent_requests: 5
 
+  # Maximum number of new connections to accept simultaneously
```
simitt: Could you add a note that `0` means unlimited (even though it is kind of expected).

graphaelli:
I have not done any performance testing with this, good idea.  Do you think merge (once tests are straightened out) should depend on that or just curious?  I could imagine contention on that semaphore being an issue, but we use the same approach already the concurrent_requests gate.

_meta/beat.reference.yml
graphaelli: done

jalvz:
i think it would be nice to have some proof (= find 1 use case) that this actually helps; but admittedly i should have said that in the issue before you started the implementation, so personally i'm fine with this (my 2c)

simitt:
I did do some basic testing with it and saw that the amount of handled requests can drop dramatically whith the limit enabled. This might uncover some more general issue, but I prefer to investigate a bit more before merging it in.

jalvz:
ok, i just spend some time testing it with different profiles, loads and settings (800Kb / 6Mb payload, 10qps / 100qps, 10/100/1k/10k/unlimited connections) and i couldn't find any noticeable difference in any case (in function of the number of connections, that is)

graphaelli:
@simitt @jalvz thanks for running perf tests, I've incorporated your feedback, can you approve for merge (or request changes)?

PR #762, opened 2018-03-22
[6x] Update beats framework to 011adba

simitt:


_beats/libbeat/processors/add_host_metadata/_meta/fields.yml
```diff
@@ -20,15 +20,14 @@
...
           description: >
             OS platform (e.g. centos, ubuntu, windows).
         - name: os.version
-          type: object
+          type: keyword
           description: >
             OS version.
         - name: os.family
-          type: object
+          type: keyword
```
simitt: this is not a breaking change, as it hasn't been released yet.

PR #763, opened 2018-03-22
[6.x] use libbeat version (#659)

graphaelli:
 - use libbeat version  (#659)


graphaelli:
requires a patch in beats first, closing until that goes in

PR #768, opened 2018-03-22
listen on unix domain socket

### LINKED ISSUE #689: Bind to a Unix Domain Socket or Windows Named Pipe instead of a TCP port
I've been thinking that it would be a good feature if we allowed the APM Server to bind to a Unix Domain Socket or Windows Named Pipe instead of a TCP port. This removes the overhead of the TCP handshake and allows the agent to do proper IPC when the APM Server is installed on the edge machine. This doesn't change the fact that the agents will be speaking HTTP to the server, but the underlying socket will be different.

I originally started researching this because I hoped it would allow for greater throughput, but it seems we currently have a bottleneck elsewhere, e.g:

Throughput when binding to TCP socket:

```
$ autocannon -m POST -H 'Content-Type: application/json' -b '{"service":{"name":"foo","agent":{"name":"foo","version":"1"}},"errors":[{"timestamp":"2017-01-01T00:00:00Z","log":{"message":"foo"}}]}' http://localhost:8200/v1/errors
Running 10s test @ http://localhost:8200/v1/errors
10 connections

Stat         Avg    Stdev   Max
Latency (ms) 2.31   2.26    70.98
Req/Sec      3555.2 324.04  3928
Bytes/Sec    434 kB 38.1 kB 479 kB

36k requests in 10s, 4.34 MB read
```

Throughput when binding to Unix Domain Socket:

```
$ autocannon -S /tmp/apm-server.sock -m POST -H 'Content-Type: application/json' -b '{"service":{"name":"foo","agent":{"name":"foo","version":"1"}},"errors":[{"timestamp":"2017-01-01T00:00:00Z","log":{"message":"foo"}}]}' http://localhost/v1/errors
Running 10s test @ http://localhost/v1/errors (/tmp/apm-server.sock)
10 connections

Stat         Avg    Stdev Max
Latency (ms) 2.45   2.47  67.17
Req/Sec      3397.4 408.5 4044
Bytes/Sec    416 kB 51 kB 493 kB

34k requests in 10s, 4.14 MB read
```

Binding to a Unix Domain Socket or a Windows Named Pipe like this also means that the system admin don't have to consider the risk of the open TCP port.

Updating the APM Server to bind to a Unix Domain Socket is fairly easy. All I did was change the following in `beater/server.go`:

```diff
 func newServer(config *Config, report reporter) *http.Server {
        mux := newMuxer(config, report)

-       return &http.Server{
-               Addr:           config.Host,
+       server := &http.Server{
                Handler:        mux,
                ReadTimeout:    config.ReadTimeout,
                WriteTimeout:   config.WriteTimeout,
                MaxHeaderBytes: config.MaxHeaderSize,
        }
+
+       unixListener, err := net.Listen("unix", "/tmp/apm-server.sock")
+       if err != nil {
+               panic(err)
+       }
+       server.Serve(unixListener)
+
+       return server
 }
```

This doesn't correctly close the socket on close, and will fail if the `.sock` file is present on boot, but it was enough for my little test.

What do you think?

graphaelli:
with `apm-server.host=unix:/path`

closes #689

### DIFF
--- a/beater/beater.go
+++ b/beater/beater.go
@@ -68,6 +68,14 @@ func (bt *beater) Run(b *beat.Beat) error {
 		bt.logger.Errorf("failed to listen: %s", err)
 		return err
 	}
+	// in case host is :0 or similar
+	if network == "tcp" {
+		addr := lis.Addr().(*net.TCPAddr).String()
+		if bt.config.Host != addr {
+			bt.logger.Infof("host resolved from %s to %s", bt.config.Host, addr)
+			bt.config.Host = addr
+		}
+	}
 	go notifyListening(bt.config, pub.Send)
 
 	bt.mutex.Lock()
--- a/beater/beater.go
+++ b/beater/beater.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"net"
 	"net/http"
+	"net/url"
 	"regexp"
 	"sync"
 
@@ -56,7 +57,13 @@ func (bt *beater) Run(b *beat.Beat) error {
 	}
 	defer pub.Stop()
 
-	lis, err := net.Listen("tcp", bt.config.Host)
+	path := bt.config.Host
+	network := "tcp"
+	if parsed, err := url.Parse(path); err == nil && parsed.Scheme == "unix" {
+		network = parsed.Scheme
+		path = parsed.Path
+	}
+	lis, err := net.Listen(network, path)
 	if err != nil {
 		bt.logger.Errorf("failed to listen: %s", err)
 		return err
--- a/beater/beater.go
+++ b/beater/beater.go
@@ -48,25 +48,24 @@ func New(b *beat.Beat, ucfg *common.Config) (beat.Beater, error) {
 	return bt, nil
 }
 
-func (bt *beater) Run(b *beat.Beat) error {
-	var err error
-
-	pub, err := newPublisher(b.Publisher, bt.config.ConcurrentRequests, bt.config.ShutdownTimeout)
-	if err != nil {
-		return err
+// parseListener extracts the network and path for a configured host address
+// all paths are tcp unix:/path/to.sock
+func parseListener(host string) (string, string) {
+	if parsed, err := url.Parse(host); err == nil && parsed.Scheme == "unix" {
+		return parsed.Scheme, parsed.Path
 	}
-	defer pub.Stop()
+	return "tcp", host
+}
 
-	path := bt.config.Host
-	network := "tcp"
-	if parsed, err := url.Parse(path); err == nil && parsed.Scheme == "unix" {
-		network = parsed.Scheme
-		path = parsed.Path
-	}
+// listen starts the listener for bt.config.Host
+// bt.config.Host may be mutated by this function in case the resolved listening address does not match the
+// configured bt.config.Host value.
+// This should only be called once, from Run.
+func (bt *beater) listen() (net.Listener, error) {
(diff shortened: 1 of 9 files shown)


graphaelli:
still thinking about a test for this, might have to be in a separate integration test

beater/beater_test.go
```diff
@@ -204,43 +204,124 @@ func (d *DummyOutputClient) Close() error {
...
+	}
+}
+
+func setupBeater(t *testing.T, ucfg *common.Config) (*beater, func()) {
+	// create a pipeline
+	pip, err := DummyPipeline()
+	assert.NoError(t, err)
+
+	// create a beat
+	apm, err := instance.NewBeat("test-apm-server", "", "")
```
graphaelli: spoke with beats team about this since it felt gross - suggested it was better to duplicate the logic than pull `instance` in

jalvz:
`ip/ExtractIP` will be foobar'ed with this, so you probably need handle the case of  not existing IPs, maybe here: https://github.com/jalvz/apm-server/blob/640e667fcef6677287b9e5830cad88ec2a44f80b/utility/common.go#L31

graphaelli:
@jalvz we only extractIP from client requests right?  I expect that wouldn't change, but I can add a test to be sure.

jalvz:
what do you mean with client requests? we do it for all the requests that come in, either backend or frontend.

graphaelli:
I mean as opposed to the listening address.  I'm not sure where something that's not an IP would be exposed to `extractIP`?

jalvz:
if you make a request to an endpoint; that function gets called, right?

graphaelli:
resolved in chat, update needed to omit remote IP if empty.  @jalvz reminds me we need docs for this.

beater/beater.go
```diff
@@ -56,11 +57,25 @@ func (bt *beater) Run(b *beat.Beat) error {
 	}
 	defer pub.Stop()
 
-	lis, err := net.Listen("tcp", bt.config.Host)
+	path := bt.config.Host
```
roncohen: Tiny nitpick: Could we extract this into its own function?

watson:
Is the `unix:` prefix a common thing to use? I've always heard that it should always end in `.sock` to be valid. But it's all anecdotal, so just curious.

watson:
@graphaelli Great work!

Did you ever look into supporting named pipes on Windows as well? (the equivalent of domain sockets in the Windows world).

I know I'm too late to the party here, but we might want to consider if the prefix `unix:` is good or bad in the context of allowing future support for named pipes. A more general purpose prefix could be `ipc:` - or we might consider not having a prefix at all (FYI: this is the regex I normally use to detect if the user have provided a named pipe on Windows: `/^\\\\[?.]\\pipe/`).

graphaelli:
I did not consider windows named pipes support.  We'd need separate logic for creating that kind of listener anyway so that should be a separate issue (and I'd love to hear a need before we spend time on it) - a quick look indicates http://github.com/Microsoft/go-winio (MIT licensed) would make this pretty straightforward, we'd just have to decide on how users would specify that type of listener. 

That might all be moot as (news to me) it seems unix socket support [is coming to windows](https://blogs.msdn.microsoft.com/commandline/2017/12/19/af_unix-comes-to-windows/), so I expect `unix:c:\tmp\apm-server.sock` will just work there eventually.  Of course we'll have to fix up the tests to use that instead of skipping them on windows.

Again, would love to hear user demand for getting these done but they don't seem too challenging given the libraries already available.

PR #774, opened 2018-03-26
Split indices by `processor.event` instead of `name`.

### LINKED ISSUE #734: Move `spans` to their own index.
Seperate spans from transaction documents by storing in their own index.

simitt:
This creates a seperate index for `spans`
rather than storing them in the same index as transactions.

closes #734

### DIFF
--- a/tests/system/test_integration.py
+++ b/tests/system/test_integration.py
@@ -222,8 +222,16 @@ class SplitIndicesIntegrationTest(SplitIndicesTest):
         self.load_docs_with_template(self.get_transaction_payload_path(name="payload.json"),
                                      'http://localhost:8200/v1/transactions',
                                      'transaction',
-                                     9,
+                                     4,
                                      query_index="test-apm-transaction-12-12-2017")
+
+    @unittest.skipUnless(INTEGRATION_TESTS, "integration test")
+    def test_span_index(self):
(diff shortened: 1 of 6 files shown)


tests/system/test_integration.py
```diff
@@ -222,9 +222,17 @@ def test_transaction_index(self):
...
                                      'transaction',
-                                     9,
+                                     4,
                                      query_index="test-apm-transaction-12-12-2017")
 
+    @unittest.skipUnless(INTEGRATION_TESTS, "integration test")
+    def test_span_index(self):
+        self.load_docs_with_template(self.get_transaction_payload_path(name="payload.json"),
+                                     'http://localhost:8200/v1/transactions',
+                                     'transaction',
```
graphaelli: what do you think of adding another param to `load_docs_with_template` for `processor.event` and having the query check more specifically against `name` and `event`, instead of just `name`, as splitting will be on `event` once this goes in?

simitt: I did [update the test](https://github.com/simitt/apm-server/blob/855fa9077c6a6aeb5fd4cb4e50ca98e61425be13/tests/system/test_integration.py#L204) slightly different - WDYT?

graphaelli: lgtm, thanks

PR #780, opened 2018-03-27
[6.x] limit number of new connections to accept simultaneously (#751)

graphaelli:
 - limit number of new connections to accept simultaneously  (#751)
 - disable conn limit test (#781)


graphaelli:
resolving flakey test first

PR #783, opened 2018-03-27
handle large integers in json

### LINKED ISSUE #775: Span.id is limited to 999999
When trying to send a span with a Span id >= 1 000 000, the APM server responds with a 400 and this error message:
 ```
Problem validating JSON document against schema: I[#] S[#] doesn't validate with "transaction#"
  I[#/transactions/0/spans/0/id] S[#/properties/transactions/items/properties/spans/items/properties/id/type] expected integer or null, but got number
```

The JSON Schema spec does not limit the range of the `integer` data type: http://json-schema.org/latest/json-schema-validation.html#rfc.section.6.1.1, so there might be something wrong with the go JSON schema validator .

### PR DESCRIPTION
**Pull request by graphaelli**
no performance impact on decoding detected with:
```
func decodeFunc(useNumber bool) func(input []byte, dst interface{}) error {
        return func(input []byte, dst interface{}) error {
                d := json.NewDecoder(bytes.NewReader(input))
                if useNumber {
                        d.UseNumber()
                }
                return d.Decode(dst)
        }
}

func BenchmarkDecode(b *testing.B) {
	benchmarks := []struct {
		name  string
		input []byte
	}{
		{"empty", []byte(`{}`)},
		{"strings", []byte(`{"a": "b"}`)},
		{"int", []byte(`{"a": 1000000000}`)},
		{"float", []byte(`{"a": 1.0}`)},
	}

	for _, bm := range benchmarks {
		for d := 0; d <= 1; d++ {
			useNumber := d == 1
			decode := decodeFunc(useNumber)
			suffix := "Default"
			if useNumber {
				suffix = "UseNumber"
			}
			b.Run(bm.name+suffix, func(b *testing.B) {
				for i := 0; i < b.N; i++ {
					decode(bm.input, &data)
				}
			})
		}
	}
}

BenchmarkDecode/emptyDefault-8         	 3000000	       455 ns/op
BenchmarkDecode/emptyUseNumber-8       	 3000000	       453 ns/op
BenchmarkDecode/stringsDefault-8       	 1000000	      1042 ns/op
BenchmarkDecode/stringsUseNumber-8     	 1000000	      1046 ns/op
BenchmarkDecode/intDefault-8           	 1000000	      1139 ns/op
BenchmarkDecode/intUseNumber-8         	 1000000	      1100 ns/op
BenchmarkDecode/floatDefault-8         	 1000000	      1084 ns/op
BenchmarkDecode/floatUseNumber-8       	 1000000	      1060 ns/op
PASS
```

closes #775

### DIFF
--- a/decoder/decoder.go
+++ b/decoder/decoder.go
@@ -24,7 +24,7 @@ func DecodeLimitJSONData(maxSize int64) Decoder {
 		if err != nil {
 			return nil, err
 		}
-		return decodeJSONData(reader)
+		return DecodeJSONData(reader)
 	}
 }
 
@@ -61,7 +61,7 @@ func readRequestJSONData(maxSize int64) Reader {
 	}
 }
 
-func decodeJSONData(reader io.ReadCloser) (map[string]interface{}, error) {
+func DecodeJSONData(reader io.ReadCloser) (map[string]interface{}, error) {
 	v := make(map[string]interface{})
 	d := json.NewDecoder(reader)
 	if err := d.Decode(&v); err != nil {
--- a/decoder/decoder.go
+++ b/decoder/decoder.go
@@ -64,6 +64,7 @@ func readRequestJSONData(maxSize int64) Reader {
 func DecodeJSONData(reader io.ReadCloser) (map[string]interface{}, error) {
 	v := make(map[string]interface{})
 	d := json.NewDecoder(reader)
+	d.UseNumber()
 	if err := d.Decode(&v); err != nil {
 		// If we run out of memory, for example
 		return nil, errors.Wrap(err, "data read error")
--- a/decoder/decoder.go
+++ b/decoder/decoder.go
@@ -5,9 +5,9 @@ import (
 	"compress/zlib"
 	"encoding/json"
 	"fmt"
-	"net/http"
-
+	"io"
 	"io/ioutil"
+	"net/http"
 	"strings"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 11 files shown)

### CODE REVIEW

File: utility/map_str_enhancer.go [hunk 1]
```diff
@@ -15,7 +16,7 @@ func Add(m common.MapStr, key string, val interface{}) {
 		return
 	}
 
-	switch val.(type) {
+	switch value := val.(type) {
 	case *bool:
 		if newVal := val.(*bool); newVal != nil {
```
graphaelli: holding off on refactoring in this pr to use `value` and avoid double type assertion, will do that next

decoder/decoder.go
```diff
@@ -44,13 +57,19 @@ func DecodeLimitJSONData(maxSize int64) Decoder {
...
-		}
-		return v, nil
+		return http.MaxBytesReader(nil, reader, maxSize), nil
+	}
+}
+
+func DecodeJSONData(reader io.ReadCloser) (map[string]interface{}, error) {
+	v := make(map[string]interface{})
+	d := json.NewDecoder(reader)
+	d.UseNumber()
```
simitt: nice!

utility/data_fetcher.go
```diff
@@ -38,6 +46,13 @@ func (d *ManualDecoder) IntPtr(base map[string]interface{}, key string, keys ...
 		if valFloat == float32(valInt) {
 			return &valInt
 		}
+	} else if valNumber, ok := val.(json.Number); ok {
```
simitt: As this should be the default now, to you mind moving it up to the first `else if`?

graphaelli: good idea, done

PR #785, opened 2018-03-28
Clarify semantics of request.url.raw

felixbarny:


docs/spec/request.json
```diff
@@ -58,7 +58,7 @@
             "properties": {
                 "raw": {
                     "type": ["string", "null"],
-                    "description": "The raw, unparsed URL of the request, e.g https://example.com:443/search?q=elasticsearch#top.",
+                    "description": "The raw, unparsed URL of the HTTP request line, e.g https://example.com:443/search?q=elasticsearch. This URL may be absolute or relative. For more details, see https://www.w3.org/Protocols/rfc2616/rfc2616-sec5.html#sec5.1.2.",
```
simitt: Afair this was specifically called `request.url` and not `http.url`, so not sure if the description is proper. As this was heavly driven by the experience and need of the @elastic/apm-agent-devs , could you please have a look at this.

watson: The reason why we renamed this from `http.url` to `request.url` was not because it wasn't http related, but because we needed to also add information about the response. So having one object called `http` and another called `response` was confusing. So we renamed them to `request` and `response`.

But it might have been better if the property clearly stated that it was http related.

axw: FWIW, I'd love it if we had `{"http":{"request":..., "response":...}}`

watson: That would be much better. Though that would deviate from our "property name" => "tab headline" rule in the APM UI. But it might be worth it. We'll have to discuss this with that in mind though /cc @sqren

sqren: If that structure makes more sense, the UI should be able to accommodate that.

beniwohli: Just remember that this is a backwards incompatible change, and any user (not just the APM UI) would need to change their code to accommodate this. I'm not sure if it's worth the effort.

Also, I'm really not a huge fan of deeply nested data structures, this would add another layer to an already quite deeply nested `request` object

watson: Moved that part of the discussion to a separate issue: https://github.com/elastic/apm-server/issues/794

simitt:
Can you please run `make update`?

felixbarny:
Sure. But why are there generated files checked into git?

axw:
> Sure. But why are there generated files checked into git?

Silvia or someone else can answer specifically for the APM server case, but it's pretty common to check in generated Go source code so that you can use [go get](https://golang.org/cmd/go/#hdr-Download_and_install_packages_and_dependencies) to build from a fresh git clone, without any additional steps.

PR #787, opened 2018-03-28
[6.x] handle large integers in json (#783)

graphaelli:
 - handle large integers in json  (#783)

### DIFF
--- a/processor/error/event.go
+++ b/processor/error/event.go
@@ -3,6 +3,7 @@ package error
 import (
 	"crypto/md5"
 	"encoding/hex"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"hash"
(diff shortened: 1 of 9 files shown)


graphaelli:
jenkins, retest this please

processor/error/event.go
```diff
@@ -165,13 +166,15 @@ func (e *Event) addException(config config.Config, service m.Service) {
 	utility.Add(ex, "type", e.Exception.Type)
 	utility.Add(ex, "handled", e.Exception.Handled)
 
-	switch e.Exception.Code.(type) {
+	switch code := e.Exception.Code.(type) {
 	case int:
```
axw: I don't think this case will ever be hit, will it? Without UseNumber, it'll be string or float64; with UseNumber, it'll be string or json.Number.

simitt: I think you're right, this probably can be removed.

PR #789, opened 2018-03-29
Update beats framework to 1a25bea

graphaelli:


graphaelli:
merge #790 first and then redo `make update-beats`

PR #790, opened 2018-03-30
use transitive dep versions in NOTICE

### LINKED ISSUE #460: NOTICE has incorrect revision information for deps of deps
While investigating #450, discovered that `NOTICE.txt` has revisions information from `vendor/vendor.json`for dependencies of dependencies.  Unfortunately those revisions are those of the `origin`, not the package being vendored.

eg, https://github.com/elastic/apm-server/blob/7bfb7ccd5065cc7abc123212d81513ececc69c33/NOTICE.txt#L34-L35

https://github.com/elastic/beats/commit/2463fd833024b2782962b18060ce642af2bdf1f4 is actually the commit in beats, which is the origin for that dependency.

graphaelli:
instead of beats version at time it pulled in transitive version.  No other transitive deps provide this information at this time.

`_beats/dev-tools/generate_notice.py` is already updated in beats

closes #460

### DIFF
--- a/_beats/dev-tools/generate_notice.py
+++ b/_beats/dev-tools/generate_notice.py
@@ -44,7 +44,7 @@ def read_versions(vendor):
     return libs
 
 
-def gather_dependencies(vendor_dirs):
+def gather_dependencies(vendor_dirs, overrides=None):
     dependencies = {}   # lib_path -> [array of lib]
     for vendor in vendor_dirs:
         libs = read_versions(vendor)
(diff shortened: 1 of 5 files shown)


_beats/dev-tools/generate_notice.py
```diff
@@ -273,7 +282,12 @@ def detect_license_summary(content):
...
                 dirs.remove(exclude)
 
+    overrides = {}  # revision overrides only for now
+    if args.beats_origin:
+        govendor = json.load(args.beats_origin)
+        overrides = {package['path']: package for package in govendor["package"]}
+
     print("Get the licenses available from {}".format(vendor_dirs))
-    create_notice(notice, args.beat, args.copyright, vendor_dirs, args.csvfile)
+    create_notice(notice, args.beat, args.copyright, vendor_dirs, args.csvfile, overrides=overrides)
```
simitt: I think this could lead to inconsistencies if the same third party lib is also used as a direct lib, as the versions could differ (haven't tried it out though). As I don't think that this is the case for us atm I'm fine with merging this and discuss if we need further changes in a follow up.

simitt:
After merging this and trying to upgrade beats I figured out that not all transitive dependencies are updated properly in the notice file. There seems to be an issue with subdirectories.
E.g. for `golang.org/x/net` we still show the beats version, as only subdirectories are listed in the `vendor/vendor.json`, but the LICENSE file is stored under`golang.org/x/net`.

PR #792, opened 2018-04-03
overrides timestamps coming from frontend agents

### LINKED ISSUE #723: Force a timestamp on RUM data
Our experience has shown that we cannot rely on timestamps in data ( _durations/elapsed time_ is fine) coming from browsers. It turns out that people routinely have computers that are completely our of sync with time or they have purposely backdated their clocks.

For this reason, we need the APM Server to set the current time on all the data that comes in through the client side endpoints. 

Alternatively, we could make timestamp optional, make sure the JS agent doesn't set a timestamp, and let the APM Server set a timestamp on all the entries that don't have a timestamp set.

jalvz:
fixes #723


roncohen:
nice @jalvz. Seems like timestamp should be optional if we're not going to use it. WDYT? We could make the logic that if the timestamp is missing in the payload, APM Server will add it. And then we change the frontend agent to make sure it's not included in the payload. cc @jahtalab

jalvz:
> We could make the logic that if the timestamp is missing in the payload, APM Server will add it
sure. that is basically re-do this with an schema change, so closing this.

PR #798, opened 2018-04-04
[6.x] Clarify semantics of request.url.raw (#785)

simitt:
 - Clarify semantics of request.url.raw  (#785)


simitt:
cc @felixbarny

PR #797, opened 2018-04-03
override revision for missed subpackages

graphaelli:
Follow up to #790, where some transitive deps were not overridden.  This change generates overrides for each part of an import path to cover those cases.

For example this was missed:

`NOTICE.txt` reported:
```
Dependency: github.com/pierrec/xxHash
Revision: 768e58ede27f7a80b38c983a83537f8669a0cb04
```

`vendor/vendor.json`:
```
{
	"checksumSHA1": "IT4sX58d+e8osXHV5U6YCSdB/uE=",
	"origin": "github.com/elastic/beats/vendor/github.com/pierrec/xxHash/xxHash32",
	"path": "github.com/pierrec/xxHash/xxHash32",
	"revision": "768e58ede27f7a80b38c983a83537f8669a0cb04",
	"revisionTime": "2018-03-20T11:14:13Z"
}
```

`_beats/vendor/vendor.json`
```
{
	"checksumSHA1": "IT4sX58d+e8osXHV5U6YCSdB/uE=",
	"path": "github.com/pierrec/xxHash/xxHash32",
	"revision": "5a004441f897722c627870a981d02b29924215fa",
	"revisionTime": "2016-01-12T16:53:51Z"
}
```

With this change, `NOTICE.txt` reports:
```
Dependency: github.com/pierrec/xxHash
Revision: 5a004441f897722c627870a981d02b29924215fa
```
as expected.

This also resolves the unknown version and license warnings.

PR #803, opened 2018-04-04
[6.x] use transitive dep versions in NOTICE (#790) | override revision for missed subpackages (#797)

graphaelli:
 - use transitive dep versions in NOTICE  (#790)
 - override revision for missed subpackages  (#797)


graphaelli:
the `NOTICE.txt` updates were regenerated with `make notice`, not backported.
`_beats/vendor/vendor.json` was copied from elastic/beats/6.x@a8d5d73f2389e7e0e7d1e80c779cd73c3329ac97

PR #811, opened 2018-04-06
Log additional request info in error case.

### LINKED ISSUE #793: Access log: using warning level for status code >= 400
Currently, the level for "access log" style log messages is hard coded to `info`: https://github.com/elastic/apm-server/blob/edb22ae93c62309902cfa3d7454881c90a1e9259/beater/handlers.go#L179-L181

Most people probably want to run the APM Server with a lower level, e.g. warning, and hence won't see any access logs at all. This is probably fine in normal operation, but you'd probably still want to see logs for error responses. Could we set the level dynamically depending on the response status code, e.g. if response code >= 400 set it to `warning`?

simitt:
Currently we only log a verbose request info for every incoming request on log level `info`. 

If the log level is set higher than info, this verbose information would be helpful for investigating request errors.

This PR ensures that the verbose information is also logged on log.level >= info if an error occurs handling a request. The additional log information is not added to the already created error log entry, as this would lead to log duplication in case log.level <= info is used. 

closes #793

### DIFF
--- a/beater/handlers.go
+++ b/beater/handlers.go
@@ -176,13 +176,21 @@ func logHandler(h http.Handler) http.Handler {
 
 		h.ServeHTTP(lw, lr)
 
-		reqLogger.Infow("handled request", "response_code", lw.Code,
-			"method", r.Method, "URL", r.URL, "content_length", r.ContentLength,
-			"remote_address", utility.ExtractIP(r), "user-agent", r.Header.Get("User-Agent"))
+		logKey := "handled request"
+		logInfo := []interface{}{
+			"response_code", lw.Code,
+			"method", r.Method,
+			"URL", r.URL,
+			"content_length", r.ContentLength,
+			"remote_address", utility.ExtractIP(r),
+			"user-agent", r.Header.Get("User-Agent"),
+		}
 
 		if lw.Code > 399 {
+			reqLogger.Errorw(logKey, logInfo...)
 			responseErrors.Inc()
 		} else {
+			reqLogger.Infow(logKey, logInfo...)
 			responseValid.Inc()
 		}
 	})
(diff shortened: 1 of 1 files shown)


beater/handlers.go
```diff
@@ -342,11 +350,10 @@ func sendStatus(w http.ResponseWriter, r *http.Request, code int, err error) {
...
 
 	logger, ok := r.Context().Value(reqLoggerContextKey).(*logp.Logger)
-	if ok {
-		logger.Errorw("error handling request", "error", err.Error())
-	} else {
-		logp.NewLogger("request").Errorf("error handling request: %v", err)
+	if !ok {
+		logger = logp.NewLogger("request")
 	}
+	logger.Errorw("error handling request", "error", err.Error())
```
beniwohli: Shouldn't this be inside the if-clause?

simitt: This if clause only checks if there is a valid logger given in the context, otherwise a new logger is created. In both cases the error that occured should be logged. 
This code is only reached if there was an error, otherwise it returns here https://github.com/elastic/apm-server/pull/811/files/3db639d45486fc0d2700741512af430c312e81b7#diff-0ee6b4dd463eb95280cc3653bd4701b6R348.

beniwohli: ah right. I'll retract my comment then!

roncohen:
cool @simitt. This would log two errors every time there is a problem. Is there a way to refactor so we only log once, but get the actual error message in the log?

simitt:
sure there is, but either 
* we only log the `request` when no error happened and add the same information to the error log 
* or we will have a duplicated entry in case an error happens and the log level is set to <= `info`. 

I went ahead with the first option.

simitt:
jenkins, retest this please

beater/handlers.go
```diff
@@ -176,13 +176,17 @@ func logHandler(h http.Handler) http.Handler {
...
 		h.ServeHTTP(lw, lr)
 
-		reqLogger.Infow("handled request", "response_code", lw.Code,
-			"method", r.Method, "URL", r.URL, "content_length", r.ContentLength,
-			"remote_address", utility.ExtractIP(r), "user-agent", r.Header.Get("User-Agent"))
-
 		if lw.Code > 399 {
 			responseErrors.Inc()
 		} else {
+			reqLogger.Infow("handled request", []interface{}{
```
roncohen: cool! What do you think about moving the parameters here (`response_code`, `method` etc.) up to when we're creating the `reqLogger`: 

```
reqLogger := logger.With(
    "request_id", reqID
    "response_code", lw.Code,
    "method", r.Method,
    "URL", r.URL,
    "content_length", r.ContentLength,
    "remote_address", utility.ExtractIP(r),
    "user-agent", r.Header.Get("User-Agent"),
)
```

this will allow you to remove it from `sendStatus`

Issue #800, opened 2018-04-04
Document options for dealing with 503: Queue is full responses


roncohen:
We need to document in which scenarios the `503: Queue is full` can occur and document which options users have to troubleshoot when it happens.

bastbnl:
@roncohen I ran into a 503 error too but it had a different cause. I'm running apm-server behind nginx and I had to configure `client_max_body_size` on the location to make sure elasticsearch (which is behind nginx as well) accepted the body since nginx will return 503 when the request body is over a certain size:

```
2018/06/16 17:06:57 [error] 9134#9134: *58437 client intended to send too large body: 1337630 bytes, client: x, server: , request: "POST /_bulk HTTP/1.1", host: "x:9200"
```

It's really not caused by `apm-server`, but maybe this would save somebody time when using this fairly common deployment pattern.

Happy to create a PR with additional details on the documentation.

PR #817, opened 2018-04-09
Documents how to handle 503s

roncohen:
bleh

Fixes https://github.com/elastic/apm-server/issues/800

### DIFF
--- a/docs/troubleshooting.asciidoc
+++ b/docs/troubleshooting.asciidoc
@@ -44,14 +44,15 @@ The internal queue helps to:
 * handle large spikes of data arriving at the APM Server at the same time
 * send documents to Elasticsearch in bulk, instead of individually
 
-
 When the internal queue is full,
 APM Server returns an HTTP 503 status with the message "Queue is full".
 
-A full queue generally means that the agents have collected more data than APM server is able to deliver to Elasticsearch. This might happen when APM Server is not configured properly for the size of your Elasticsearch cluster,
+A full queue generally means that the agents have collected more data than APM server is able to deliver to Elasticsearch.
+This might happen when APM Server is not configured properly for the size of your Elasticsearch cluster.
 or because your Elasticsearch cluster is underpowered or not configured properly for the given workload.
 
-The queue can also fill up if Elasticsearch is unavailable for a prolonged period, or a sudden spike of data arrives at the APM Server.
+The queue can also fill up if Elasticsearch is unavailable for a prolonged period,
+or a sudden spike of data arrives at the APM Server.
 
 To solve the "503: Queue is full" problem,
 you have a few options:
@@ -81,17 +82,16 @@ you can try to:
 [[tune-output-config]]
 ==== Tune APM Server output parameters for your Elasticsearch cluster
 
-If your Elasticsearch cluster is sized properly, but not ingesting the amount of data you expect, you can tweak APM Server options to make better use of the cluster:
+If your Elasticsearch cluster is sized properly,
+but not ingesting the amount of data you expect,
+you can tweak APM Server options to make better use of the cluster:
 
 * Increase `output.elasticsearch.workers`
 * Make sure that `output.elasticsearch.bulk_max_size` is high, for example 5120. The default of 50 is very conservative.
 +
-If you increase `bulk_max_size`, make sure to also increase `queue.mem.size`.
+If you increase `bulk_max_size`,
+make sure to also increase `queue.mem.size`.
 A good rule of thumb is that `queue.mem.size` should equal `workers` multiplied by `bulk_max_size`.
-
-See {ref}/tune-for-indexing-speed.html[Tune for indexing speed],
-particularly the sections about how to increase the refresh interval, disable swapping,
-use faster hardware, and set the indexing buffer size.
 
 [[increase-cluster-ingest]]
 ==== Tune Elasticsearch for higher ingestion
@@ -113,7 +113,8 @@ Reducing the transaction sample rate does not affect the collection of metrics s
 [[increase-queue-size]]
 ==== Increase internal queue size
 
-A larger internal queue allows Elasticsearch to be unavailable for longer periods, and it alleviates problems that might result from sudden spikes of data. 
+A larger internal queue allows Elasticsearch to be unavailable for longer periods,
+and it alleviates problems that might result from sudden spikes of data. 
 You can increase the queue size by setting `apm-server.queue.mem.size`. 
 Be aware that increasing `apm-server.queue.mem.size` can significantly affect APM Server memory usage. 
 
@@ -122,7 +123,8 @@ Be aware that increasing `apm-server.queue.mem.size` can significantly affect AP
 
 Large payloads coming from agents may result in "503: Request timed out waiting to be processed" messages.
 You can reduce the payload size by decreasing the `max_queue_size` in the agents. 
-This will result in agents sending smaller payloads to the APM Server, but the requests will be more frequent.
+This will result in agents sending smaller payloads to the APM Server,
+but the requests will be more frequent.
 See the documentation for the {apm-py-ref}/configuration.html#config-max-queue-size[Python] and {apm-node-ref}/agent-api.html#max-queue-size[Node.js] agents for more information.
 
 [[add-apm-server-nodes]]
--- a/docs/troubleshooting.asciidoc
+++ b/docs/troubleshooting.asciidoc
@@ -88,7 +88,8 @@ but not ingesting the amount of data you expect,
 you can tweak APM Server options to make better use of the cluster:
 
 * Increase `output.elasticsearch.workers`
-* Make sure that `output.elasticsearch.bulk_max_size` is high, for example 5120. The default of 50 is very conservative.
+* Make sure that `output.elasticsearch.bulk_max_size` is high, for example 5120.
+  The default of 50 is very conservative.
 +
 If you increase `bulk_max_size`,
 make sure to also increase `queue.mem.size`.
@@ -99,8 +100,8 @@ A good rule of thumb is that `queue.mem.size` should equal `workers` multiplied 
 
 Increasing the Elasticsearch ingestion rate is a large topic.
 See {ref}/tune-for-indexing-speed.html[Tune for indexing speed],
-particularly the sections about how to increase the refresh interval, disable swapping,
-use faster hardware, and set the indexing buffer size.
+particularly the sections about how to increase the refresh interval,
+disable swapping, use faster hardware, and set the indexing buffer size.
 
 [[reduce-data]]
 ==== Reduce the amount of data collected
--- a/docs/troubleshooting.asciidoc
+++ b/docs/troubleshooting.asciidoc
@@ -29,56 +29,69 @@ include::copied-from-beats/getting-help.asciidoc[]
 
 include::copied-from-beats/debugging.asciidoc[]
 
+[[common-problems]]
 == Common problems
+[[queue-is-full]]
 === 503: Queue is full
 APM Server has an internal queue which stores documents until they can be delivered to Elasticsearch.
 It is designed this way to alleviate problems that might occur if Elasticsearch is intermittently unavailable or if a large spike of data arrives at the APM Server at the same time.
 
-When the internal queue is full, APM Server will return a HTTP 503 status with the message "Queue is full".
+When the internal queue is full,
+APM Server will return a HTTP 503 status with the message "Queue is full".
 
-A full queue generally means APM Server that the amount of data being collected by agents and sent to APM Server is greater than the amount of data that the Elasticsearch cluster can ingest.
-If Elasticsearch is unavailable for a prolonged period or if a sudden spike of data arrives at the APM Server, the queue can also fill up.
+A full queue generally means that the amount of data being collected by agents and sent to APM Server is greater than the amount of data that the Elasticsearch cluster can ingest.
+If Elasticsearch is unavailable for a prolonged period or if a sudden spike of data arrives at the APM Server,
+the queue can also fill up.
 
-To solve the "503: Queue is full" problem, you have a few options:
+To solve the "503: Queue is full" problem,
+you have a few options:
 
 * <<reduce-data,Reduce the amount of data collected>>
 * <<increase-ingestion-rate,Increase the ingestion rate of Elasticsearch>>
-* <<increase-queue-size,Increase the size of the APM Server queue.>>
+* <<increase-queue-size,Increase the size of the APM Server queue>>
 
-
+[[request-timed-out]]
 === 503: Request timed out waiting to be processed
 
 There is a limit to the number of requests that the APM Server will process at the same time.
-The APM Server will return a HTTP 503 status with the message "Request timed out waiting to be processed" when the limit is reached and the request from an agent is blocked from delivering the request.
+The APM Server will return a HTTP 503 status with the message "Request timed out waiting to be processed" when the limit is reached and the request from an agent is blocked.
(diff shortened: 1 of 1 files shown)


docs/troubleshooting.asciidoc
```diff
@@ -28,3 +28,59 @@ include::copied-from-beats/getting-help.asciidoc[]
...
 include::copied-from-beats/debugging.asciidoc[]
+
+== Common problems
+=== 503: Queue is full
+APM Server has an internal queue which stores documents until they can be delivered to Elasticsearch.
+It is designed this way to alleviate problems that might occur if Elasticsearch is intermittently unavailable or if a large spike of data arrives at the APM Server at the same time.
+
+When the internal queue is full, APM Server will return a HTTP 503 status with the message "Queue is full".
+
+A full queue generally means APM Server that the amount of data being collected by agents and sent to APM Server is greater than the amount of data that the Elasticsearch cluster can ingest.
```
alvarolobato: I think `APM Server` is a typo here `means APM Server that`

docs/troubleshooting.asciidoc
```diff
@@ -28,3 +28,59 @@ include::copied-from-beats/getting-help.asciidoc[]
...
+
+* <<reduce-data,Reduce the amount of data collected>>
+* <<increase-ingestion-rate,Increase the ingestion rate of Elasticsearch>>
+* <<increase-queue-size,Increase the size of the APM Server queue.>>
+
+
+=== 503: Request timed out waiting to be processed
+
+There is a limit to the number of requests that the APM Server will process at the same time.
+The APM Server will return a HTTP 503 status with the message "Request timed out waiting to be processed" when the limit is reached and the request from an agent is blocked from delivering the request.
```
alvarolobato: `and the request from an agent is blocked from delivering the request.` 
This is the case where you are not enqueuing on time right? I think the last part of the sentence is not clear or maybe just `from delivering the request` nest to be removed.

docs/troubleshooting.asciidoc
```diff
@@ -28,3 +28,59 @@ include::copied-from-beats/getting-help.asciidoc[]
 == Debug
 
 include::copied-from-beats/debugging.asciidoc[]
+
+== Common problems
+=== 503: Queue is full
+APM Server has an internal queue which stores documents until they can be delivered to Elasticsearch.
+It is designed this way to alleviate problems that might occur if Elasticsearch is intermittently unavailable or if a large spike of data arrives at the APM Server at the same time.
+
+When the internal queue is full, APM Server will return a HTTP 503 status with the message "Queue is full".
```
watson: Add a line break after punctuations, i.e:

```diff
-When the internal queue is full, APM Server will return a HTTP 503 status with the message "Queue is full".
+When the internal queue is full,
+APM Server will return a HTTP 503 status with the message "Queue is full".
```

The same goes for the rest of the document, but I'll comment more on it

docs/troubleshooting.asciidoc
```diff
@@ -28,3 +28,59 @@ include::copied-from-beats/getting-help.asciidoc[]
...
+
+A full queue generally means APM Server that the amount of data being collected by agents and sent to APM Server is greater than the amount of data that the Elasticsearch cluster can ingest.
+If Elasticsearch is unavailable for a prolonged period or if a sudden spike of data arrives at the APM Server, the queue can also fill up.
+
+To solve the "503: Queue is full" problem, you have a few options:
+
+* <<reduce-data,Reduce the amount of data collected>>
+* <<increase-ingestion-rate,Increase the ingestion rate of Elasticsearch>>
+* <<increase-queue-size,Increase the size of the APM Server queue.>>
```
watson: Remove trailing period

docs/troubleshooting.asciidoc
```diff
@@ -28,3 +28,59 @@ include::copied-from-beats/getting-help.asciidoc[]
...
+There is a limit to the number of requests that the APM Server will process at the same time.
+The APM Server will return a HTTP 503 status with the message "Request timed out waiting to be processed" when the limit is reached and the request from an agent is blocked from delivering the request.
+This limit is determined by the `apm-server.concurrent_requests` configuration parameter.
+
+To alleviate this problem, you can try to:
+
+* <<reduce-data,Reduce the amount of data collected>>
+* <<reduce-payload-size,Decrease the payload size>>
+
+=== Solutions
```
watson: I would add a `[[solutions]]` id above this headline, move all id's like `[[reduce-data]]` above their respective headline, and add an empty line above and below each headlines, i.e:

```diff
+[[solutions]]
 === Solutions 
+
+[[reduce-data]]
 ==== Reduce the amount of data collected
-[[reduce-data]]
+
```

The same goes for the other headlines in this doc

docs/troubleshooting.asciidoc
```diff
@@ -28,3 +28,59 @@ include::copied-from-beats/getting-help.asciidoc[]
...
+[[reduce-data]]
+The most obvious way to reduce the number of documents to be indexed, is to reduce the _transaction sample rate_.
+Transaction sample rate is controlled in the configuration of agents (for example for {apm-py-ref}/configuration.html#config-transaction-sample-rate[Python] and {apm-node-ref}/agent-api.html#transaction-sample-rate[Node.js]).
+
+Reducing transaction sample rate does not affect the collection of metrics such as Requests Per Second.
+
+==== Increase the Elasticsearch ingestion rate
+[[increase-ingestion-rate]]
+Increasing Elasticsearch ingestion rate is a large topic.
+Please see {ref}/tune-for-indexing-speed.html, particularly `Increase the refresh interval`, `Disable swapping`,
```
watson: This URL should probably not be shown as is, but instead be linked to as a piece of text, e.g:

```
{ref}/tune-for-indexing-speed.html[Tune for indexing speed]
```

roncohen:
Updated according to comments.

@dedemorton could use your input here

docs/troubleshooting.asciidoc
```diff
@@ -28,3 +28,72 @@ include::copied-from-beats/getting-help.asciidoc[]
 == Debug
 
 include::copied-from-beats/debugging.asciidoc[]
+
+[[common-problems]]
+== Common problems
+[[queue-is-full]]
+=== 503: Queue is full
+APM Server has an internal queue which stores documents until they can be delivered to Elasticsearch.
+It is designed this way to alleviate problems that might occur if Elasticsearch is intermittently unavailable or if a large spike of data arrives at the APM Server at the same time.
```
simitt: I assume with `the internal queue` you are summarizing all queues in place until data are written to ES. If so I suggest to also add that the queue is used to send data in batches to ES for optimized write.

docs/troubleshooting.asciidoc
```diff
@@ -28,3 +28,72 @@ include::copied-from-beats/getting-help.asciidoc[]
...
+
+When the internal queue is full,
+APM Server will return a HTTP 503 status with the message "Queue is full".
+
+A full queue generally means that the amount of data being collected by agents and sent to APM Server is greater than the amount of data that the Elasticsearch cluster can ingest.
```
simitt: How about `than the amount of data that can be processed by the APM Server and ingested by the ES cluster` ?

Otherwise people might only start to fine tune the ES cluster, but it could also be that they need to set up additional APM Server

docs/troubleshooting.asciidoc
```diff
@@ -28,3 +28,72 @@ include::copied-from-beats/getting-help.asciidoc[]
...
+If Elasticsearch is unavailable for a prolonged period or if a sudden spike of data arrives at the APM Server,
+the queue can also fill up.
+
+To solve the "503: Queue is full" problem,
+you have a few options:
+
+* <<reduce-data,Reduce the amount of data collected>>
+* <<increase-ingestion-rate,Increase the ingestion rate of Elasticsearch>>
+* <<increase-queue-size,Increase the size of the APM Server queue>>
```
simitt: How about 
- increase the number of running APM Servers?

docs/troubleshooting.asciidoc
```diff
@@ -28,3 +28,72 @@ include::copied-from-beats/getting-help.asciidoc[]
...
+
+There is a limit to the number of requests that the APM Server will process at the same time.
+The APM Server will return a HTTP 503 status with the message "Request timed out waiting to be processed" when the limit is reached and the request from an agent is blocked.
+This limit is determined by the `apm-server.concurrent_requests` configuration parameter.
+
+To alleviate this problem,
+you can try to:
+
+* <<reduce-data,Reduce the amount of data collected>>
+* <<reduce-payload-size,Decrease the payload size>>
```
simitt: * increase the number of `concurrent_requests`
* increase the number of APM Servers

docs/troubleshooting.asciidoc
roncohen: Until recently, increasing APM Server processing power would never fix "Queue is full". Maybe that changed with the "early return" PR?

simitt: If you have multiple APM Server instances running behind a load balancer adding additional machines basically increases the number of concurrent_requests to be processed and also the internal memory queue, as you have more instances running that handle incoming requests.

roncohen: the general point i was trying to make in these two is:

Queue is full: Elasticsearch is the bottleneck
Request timed out waiting to be processed: APM Server is the bottleneck

I'll try to make this clearer

dedemorton:
@roncohen Looked through this and most of my comments are minor editorial nitpicks . Might be better for me to edit the content after you've merged it (rather than clutter up the review with a scattershot of minor comments). Sound OK?

roncohen:
I updated this after @simitt suggested offline that it would benefit from details on ES ingest vs. misconfigured APM Server output. 

@dedemorton if i merge it goes up on the site. Maybe you could just push your commits to this PR? I think i can give you access.

docs/troubleshooting.asciidoc
```diff
@@ -79,56 +81,55 @@ you can try to:
...
-  * If you increase `bulk_max_size`, make sure to also increase `queue.mem.size`.
-  A good rule of thumb for `queue.mem.size` should equal `workers` multiplied by `bulk_max_size`.
++
+If you increase `bulk_max_size`, make sure to also increase `queue.mem.size`.
+A good rule of thumb is that `queue.mem.size` should equal `workers` multiplied by `bulk_max_size`.
 
-Please see {ref}/tune-for-indexing-speed.html[Tune for indexing speed],
-particularly `Increase the refresh interval`, `Disable swapping`,
-`Use faster hardware` and `Indexing buffer size`.
+See {ref}/tune-for-indexing-speed.html[Tune for indexing speed],
```
dedemorton: Note that I've removed the hard-coded section titles because this sort of info gets stale over time. If we use a general description, it will be applicable even when the section titles change.

docs/troubleshooting.asciidoc
```diff
@@ -28,3 +29,107 @@ include::copied-from-beats/getting-help.asciidoc[]
...
+[[queue-is-full]]
+=== 503: Queue is full
+APM Server has an internal queue that buffers documents until they can be delivered to Elasticsearch.
+The internal queue helps to: 
+
+* alleviate problems that might occur if Elasticsearch is intermittently unavailable
+* handle large spikes of data arriving at the APM Server at the same time
+* send documents to Elasticsearch in bulk, instead of individually
+
+
```
watson: Nitpick: remove double line break

docs/troubleshooting.asciidoc
```diff
@@ -28,3 +29,107 @@ include::copied-from-beats/getting-help.asciidoc[]
...
 
 include::copied-from-beats/debugging.asciidoc[]
+
+[[common-problems]]
+== Common problems
+
+This section describes common problems you might encounter with APM Server.
+
+[[queue-is-full]]
+=== 503: Queue is full
```
watson: Nitpick: Add blank line after headlines

docs/troubleshooting.asciidoc
```diff
@@ -28,3 +29,110 @@ include::copied-from-beats/getting-help.asciidoc[]
...
+
+[[tune-output-config]]
+==== Tune APM Server output parameters for your Elasticsearch cluster
+
+If your Elasticsearch cluster is sized properly,
+but not ingesting the amount of data you expect,
+you can tweak APM Server options to make better use of the cluster:
+
+* Increase `output.elasticsearch.workers`
+* Make sure that `output.elasticsearch.bulk_max_size` is high, for example 5120. The default of 50 is very conservative.
```
watson:
```diff
-* Make sure that `output.elasticsearch.bulk_max_size` is high, for example 5120. The default of 50 is very conservative.
+* Make sure that `output.elasticsearch.bulk_max_size` is high,
+  for example 5120.
+  The default of 50 is very conservative
```

docs/troubleshooting.asciidoc
```diff
@@ -28,3 +29,110 @@ include::copied-from-beats/getting-help.asciidoc[]
...
+
+[[queue-is-full]]
+=== 503: Queue is full
+
+APM Server has an internal queue that buffers documents until they can be delivered to Elasticsearch.
+The internal queue helps to: 
+
+* alleviate problems that might occur if Elasticsearch is intermittently unavailable
+* handle large spikes of data arriving at the APM Server at the same time
+* send documents to Elasticsearch in bulk, instead of individually
```
watson:
```diff
-* send documents to Elasticsearch in bulk, instead of individually
+* send documents to Elasticsearch in bulk,
+  instead of individually
```

docs/troubleshooting.asciidoc
```diff
@@ -28,3 +29,110 @@ include::copied-from-beats/getting-help.asciidoc[]
...
+If you increase `bulk_max_size`,
+make sure to also increase `queue.mem.size`.
+A good rule of thumb is that `queue.mem.size` should equal `workers` multiplied by `bulk_max_size`.
+
+[[increase-cluster-ingest]]
+==== Tune Elasticsearch for higher ingestion
+
+Increasing the Elasticsearch ingestion rate is a large topic.
+See {ref}/tune-for-indexing-speed.html[Tune for indexing speed],
+particularly the sections about how to increase the refresh interval, disable swapping,
```
watson:
```diff
-particularly the sections about how to increase the refresh interval, disable swapping,
+particularly the sections about how to increase the refresh interval,
+disable swapping,
```

docs/troubleshooting.asciidoc
```diff
@@ -28,3 +29,110 @@ include::copied-from-beats/getting-help.asciidoc[]
...
+use faster hardware, and set the indexing buffer size.
```
watson:
```diff
-use faster hardware, and set the indexing buffer size.
+use faster hardware,
+and set the indexing buffer size.
```

roncohen:
thanks Thomas. I've made some additional changes but not all the ones you suggested.
The guideline that we need a newline for every punctuation is not a hard rule in my opinion.
There are plenty of situations where splitting up into tiny lines will decrease readability in the source and not significantly help comment-ability.

watson:
@roncohen Fair enough. I was going all in as the human linter here 😄

But just a little background on my thinking: AFAIK the line break rule were added not for readability but to improve future diff'ing. Just as with code formatting rules I feel it's important to follow it to the letter as it removes any ambiguity on when to apply the rules. You just always apply them without having to consider each case.

As a reviewer this also makes it easier to know when to request a change or not - E.g. "Did you intend to leave out the line break in this case?",  "Should I ask?",  "Will he think I'm being too nit-picky?"

This also helps when the next person comes along to make changes to the document, as it's easier to see what the rules are if they they are always followed. Else you might get the broken window syndrome where you don't care to follow the rules because you can see that others don't always follow them either.

PR #819, opened 2018-04-09
makes timestamp optional

### LINKED ISSUE #723: Force a timestamp on RUM data

jalvz:
fixes #723

### DIFF
--- a/utility/data_fetcher.go
+++ b/utility/data_fetcher.go
@@ -145,6 +145,9 @@ func (d *ManualDecoder) MapStr(base map[string]interface{}, key string, keys ...
 
 func (d *ManualDecoder) TimeRFC3339(base map[string]interface{}, key string, keys ...string) time.Time {
 	val := getDeep(base, keys...)[key]
+	if val == nil {
+		return time.Time{}
+	}
 	if valStr, ok := val.(string); ok {
 		if valTime, err := time.Parse(time.RFC3339, valStr); err == nil {
 			return valTime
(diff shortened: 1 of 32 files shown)


processor/transaction/payload.go
```diff
@@ -65,6 +67,10 @@ func (pa *payload) transform(config config.Config) []beat.Event {
 
 	var events []beat.Event
 	for _, event := range pa.Events {
+		var zeroTime = time.Time{}
+		if event.Timestamp == zeroTime {
```
roncohen: Looks like this `IsZero` method could be useful here: https://golang.org/pkg/time/#Time.IsZero

utility/data_fetcher.go
```diff
@@ -145,6 +145,9 @@ func (d *ManualDecoder) MapStr(base map[string]interface{}, key string, keys ...
 
 func (d *ManualDecoder) TimeRFC3339(base map[string]interface{}, key string, keys ...string) time.Time {
 	val := getDeep(base, keys...)[key]
+	if val == nil {
+		return time.Time{}
```
simitt: How about introducing a `TimeRFC3339WithDefault` method that returns a `time.Now()` instead of a zero time in case it is empty? The decoding happens before the transformation, so the timestamp would be more accurate, and no further handling in the `transform` function would be needed.

jalvz:
jenkins retest it please

roncohen:
this is great. Good work @jalvz @simitt

PR #825, opened 2018-04-11
[6.x] Early return to caller. (#746)

simitt:
 - Early return to caller.  (#746)


simitt:
Please *do not merge* for now!

PR #826, opened 2018-04-11
Move common problems up

roncohen:
Common problems should be the first part of the trouble shooting section, so people will see it before they start asking questions in discuss etc.

image


roncohen:
thanks @dedemorton !

PR #834, opened 2018-04-13
Update beats framework to 41a118f

graphaelli:
* addresses import issue introduced in elastic/beats#6825


graphaelli:
splitting this up into two pull requests for easier backporting.

PR #836, opened 2018-04-13
wait for publishers to start

graphaelli:
to remove flakiness from TestNumberGoRoutines.

### DIFF
--- a/beater/pub_test.go
+++ b/beater/pub_test.go
@@ -11,9 +11,11 @@ import (
 )
 
 func TestNumberGoRoutines(t *testing.T) {
+	runtime.GC() // force goroutine accounting?
 	before := runtime.NumGoroutine()
 	_, err := newPublisher(pip{}, 1, time.Duration(0))
 	assert.NoError(t, err)
+	runtime.GC()
 	assert.Equal(t, before+runtime.GOMAXPROCS(0), runtime.NumGoroutine())
 }
 
--- a/beater/pub.go
+++ b/beater/pub.go
@@ -37,6 +37,10 @@ var (
 	errChannelClosed     = errors.New("Can't send batch, publisher is being stopped")
 )
 
+type doner interface {
+	Done()
+}
+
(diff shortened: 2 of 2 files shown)


graphaelli:
still trying to think of another good reason.

graphaelli:
One possible justification for this is that we want the consumers of that channel up before the writers to not block sends that should have succeeded - they should only fail when the queue is legitimately full, not when our writers started up too slowly.  Weak: yes.

beater/pub_test.go
```diff
@@ -11,9 +11,11 @@ import (
 )
 
 func TestNumberGoRoutines(t *testing.T) {
+	runtime.GC() // force goroutine accounting?
 	before := runtime.NumGoroutine()
 	_, err := newPublisher(pip{}, 1, time.Duration(0))
 	assert.NoError(t, err)
+	runtime.GC()
```
simitt: What do you think about something like `wait_until_true_with_a_timeout`, eg. 
```
func TestNumberGoRoutines(t *testing.T) {
	expected := runtime.NumGoroutine() + runtime.GOMAXPROCS(0)
	_, err := newPublisher(pip{}, 1, time.Duration(0))
	assert.NoError(t, err)
	for i := 0; i < 10; i++ {
		if expected == runtime.NumGoroutine() {
			break
		} else {
			time.Sleep(100 * time.Millisecond)
		}
	}
	assert.Equal(t, expected, runtime.NumGoroutine())
}
```

It is also not very nice, but I'd prefer it over adding `sync.Wait` to production code for making a test work.

graphaelli: That's actually where I started but was considered this will not catch the case where `NumGoroutine` continues to grow after the condition is already met.

simitt:
As just discussed offline, I create a PR https://github.com/elastic/apm-server/pull/845 to get rid of the test, as it doesn't add a lot of value, but noise.

graphaelli:
after discussion IRL we're going to remove this test altogether

PR #837, opened 2018-04-13
Allow other tests to be run from any directory

graphaelli:
follow on from #835

### DIFF
--- a/tests/system/test_integration.py
+++ b/tests/system/test_integration.py
@@ -40,14 +40,16 @@ class Test(ElasticTest):
         rs = self.es.search(index=self.index_name, body={
             "query": {"term": {"processor.event": "transaction"}}})
         assert rs['hits']['total'] == 4, "found {} documents".format(rs['count'])
-        approved = json.load(open('transaction.approved.json'))
+        with open(self._beat_path_join(os.path.dirname(__file__), 'transaction.approved.json')) as f:
+            approved = json.load(f)
         self.check_docs(approved, rs['hits']['hits'], 'transaction')
 
         # compare existing ES documents for spans with new ones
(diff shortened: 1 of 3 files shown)


tests/system/test_requests.py
```diff
@@ -1,20 +1,21 @@
...
-from datetime import datetime
-from collections import defaultdict
-import threading
+
 
 try:
     from StringIO import StringIO
 except ImportError:
-    import io
+    from io import StringIO
```
graphaelli: shouldn't this work for both python2.7 in python3?

beniwohli: yup, as long as you only feed unicode strings into `io.StringIO`, it should behave the same as `StringIO.StringIO` on 2.7. The latter also accepts bytestrings

tests/system/test_integration.py
```diff
@@ -40,14 +40,16 @@ def test_load_docs_with_template_and_add_transaction(self):
         rs = self.es.search(index=self.index_name, body={
             "query": {"term": {"processor.event": "transaction"}}})
         assert rs['hits']['total'] == 4, "found {} documents".format(rs['count'])
-        approved = json.load(open('transaction.approved.json'))
+        with open(self._beat_path_join(os.path.dirname(__file__), 'transaction.approved.json')) as f:
```
beniwohli: not introduced with this PR, but still: I would recommend to use [`codecs.open(..., encoding='utf-8')`](https://docs.python.org/3/library/codecs.html#codecs.open) with content that may very well contain non-ASCII characters

graphaelli: good call, this actually bit us recently elsewhere - I'll make an issue to address both.

PR #831, opened 2018-04-12
Is beats synced?

graphaelli:

PR #838, opened 2018-04-13
[6.x] Is beats updated? (#831)

graphaelli:
 - Is beats updated?  (#831)


graphaelli:
travis failure is due to flakey test per #836

PR #845, opened 2018-04-16
Remove flaky test for goroutine consumers.

simitt:
As this test doesn't add a lot of value,
but noise, it is removed again.


graphaelli:
jenkins, retest this please

PR #856, opened 2018-04-18
Monitor status codes for requests.

simitt:


simitt:
jenkins, retest this please

beater/handlers.go
```diff
@@ -348,7 +359,24 @@ func sendStatus(w http.ResponseWriter, r *http.Request, code int, err error) {
 	if !ok {
 		logger = logp.NewLogger("request")
 	}
-	logger.Errorw("error handling request", []interface{}{"response_code", code, "error", err.Error()}...)
+	logger.Errorw("error handling request", "response_code", code, "error", err.Error())
+
+	if code == http.StatusServiceUnavailable {
```
jalvz: you can also use switch if you prefer:
```switch err {
		case errFull:
			errMap[statusQueueFull].Inc()
		case errConcurrencyLimitReached:
			errMap[statusConcurrencyLimit].Inc()
		case errChannelClosed:
			errMap[statusChannelClosed].Inc()
		default:
			logger.Warnw("error counter missing", "response_code", code, "error", err.Error())
		}```

File: beater/handlers.go [hunk 2]
```diff
@@ -55,8 +55,22 @@ type routeMapping struct {
 var (
 	serverMetrics  = monitoring.Default.NewRegistry("apm-server.server")
 	requestCounter = monitoring.NewInt(serverMetrics, "requests.counter")
```
roncohen: should we call this "count" instead of "counter" to align them?

beater/handlers.go
```diff
@@ -357,6 +364,32 @@ func sendStatus(w http.ResponseWriter, r *http.Request, code int, err error) {
 	}
 }
 
+func logErrMetrics(c context.Context, code int, err error) {
+	logger, ok := c.Value(reqLoggerContextKey).(*logp.Logger)
+	if !ok {
+		logger = logp.NewLogger("request")
+	}
+	logger.Errorw("error handling request", "response_code", code, "error", err.Error())
```
roncohen: I would move this out as it doesn't relate to metrics

beater/handlers.go
```diff
@@ -54,9 +54,24 @@ type routeMapping struct {
...
+		http.StatusBadRequest:       monitoring.NewInt(serverMetrics, "response.errors.badrequest"),
+		http.StatusForbidden:        monitoring.NewInt(serverMetrics, "response.errors.forbidden"),
+		http.StatusUnauthorized:     monitoring.NewInt(serverMetrics, "response.errors.unauthorized"),
+		http.StatusTooManyRequests:  monitoring.NewInt(serverMetrics, "response.errors.ratelimit"),
+		http.StatusMethodNotAllowed: monitoring.NewInt(serverMetrics, "response.errors.method"),
+		statusQueueFull:             monitoring.NewInt(serverMetrics, "response.errors.queue"),
+		statusConcurrencyLimit:      monitoring.NewInt(serverMetrics, "response.errors.concurrency"),
+		statusChannelClosed:         monitoring.NewInt(serverMetrics, "response.errors.closed"),
+	}
+	errCountMissing = "metrics.apm-server.server.response.errors count missing for"
```
roncohen: I think this assumes `response_code` will come right after in the error message, but the `response_code` will be just another key/value in the log output, coming out in random order.

Something like this:
```
metrics.apm-server.server.response.errors count missing for      {"key1": "value1", "response_code": "999"}
```
so maybe this should be something like:

`internal counter missing for given response code`

or similar

simitt:
Closing this, as everything would be overwritten by https://github.com/elastic/apm-server/pull/862

PR #859, opened 2018-04-18
Prevent automatically inheriting configs from beats. (#578)

roncohen:
Config files are not copied over from beats any longer.
The control over the config files now is within apm-server,
to enable different default settings and more fine granular control
over enabled or suggested config options.
```
Conflicts:
	_beats/libbeat/_meta/config.reference.yml
	_beats/libbeat/_meta/config.yml
	_meta/beat.reference.yml
	_meta/beat.yml
	apm-server.reference.yml
```


roncohen:
on 6.2, after `update-beats`, `check-full` fails. This is an attempt to fix it.

CHANGELOG.asciidoc
```diff
@@ -48,6 +48,7 @@ https://github.com/elastic/apm-server/compare/6.1...6.2[View commits]
 - Increase default 'ConcurrentRequests' from 20 to 40 {pull}492[492]
 - Change config option `frontend.sourcemapping.index` to `frontend.source_mapping.index_pattern` and remove adding a '*' by default.{pull}492[492].
 - Add Config option for excluding stack trace frames from `grouping_key` calculation {pull}482[482]
+- Remove config files from beats. Manually add relevant config options {pull}578[578]
```
simitt: There is a changelog entry already from PR #496 .

> - Remove untested config options from config yml files {pull}496[496]

The manual changes were made before the script to update beats was changed, but basically had the same effect. As you can see from the changes in the apm config yml files in this PR it is basically cosmetic changes, but no config options from beats are removed, as they were already before. 
So maybe a more accurate entry would be only `Remove config files from beats.`

simitt:
@roncohen it would be great to get this in, so we can actually run `make update-beats`. Could you please respond to my comment about the CHANGELOG?

simitt:
@roncohen it would be great to get this in, so we can update the beats framework.

PR #862, opened 2018-04-19
Add monitoring for agents and all response codes and reasons

jalvz:


beater/handlers.go
```diff
@@ -52,17 +52,63 @@ type routeMapping struct {
...
-	serverMetrics  = monitoring.Default.NewRegistry("apm-server.server")
-	requestCounter = monitoring.NewInt(serverMetrics, "requests.counter")
-	responseValid  = monitoring.NewInt(serverMetrics, "response.valid")
-	responseErrors = monitoring.NewInt(serverMetrics, "response.errors")
+	serverMetrics = monitoring.Default.NewRegistry("apm-server.server")
+	counter       = func(s string) *monitoring.Int {
+		return monitoring.NewInt(serverMetrics, s)
+	}
+	requestCounter = counter("request.count")
+	responseCount  = counter("response.count")
```
simitt: `responseCounter` to stay aligned

beater/handlers.go
```diff
@@ -52,17 +52,63 @@ type routeMapping struct {
...
+	responseErrors = counter("response.errors.count")
```
simitt: Now that you split up `response.valid` I'd also expect a `responseSucesses` or removing the `responseErrors`

beater/handlers.go
```diff
@@ -52,17 +52,63 @@ type routeMapping struct {
...
+
 var (
-	serverMetrics  = monitoring.Default.NewRegistry("apm-server.server")
-	requestCounter = monitoring.NewInt(serverMetrics, "requests.counter")
-	responseValid  = monitoring.NewInt(serverMetrics, "response.valid")
-	responseErrors = monitoring.NewInt(serverMetrics, "response.errors")
+	serverMetrics = monitoring.Default.NewRegistry("apm-server.server")
+	counter       = func(s string) *monitoring.Int {
+		return monitoring.NewInt(serverMetrics, s)
+	}
```
simitt: I'd prefer adding a `newServerResponse(err error, code int, name string)` method that abstracts how the counter gets created.

beater/handlers.go
```diff
@@ -52,17 +52,63 @@ type routeMapping struct {
...
+		errors.New("forbidden request"), http.StatusForbidden, counter("response.errors.forbidden"),
+	}
+	unauthorized = serverResponse{
+		errors.New("invalid token"), http.StatusUnauthorized, counter("response.errors.unauthorized"),
+	}
+	requestTooLarge = serverResponse{
+		errors.New("request body too large"), http.StatusRequestEntityTooLarge, counter("response.errors.toolarge"),
+	}
+	decodeCounter = counter("response.errors.decode")
+	cantDecode    = func(err error) serverResponse {
```
simitt: Could you use `cannotDecode` over `cantDecode` please.

beater/handlers_test.go
```diff
@@ -7,69 +7,52 @@ import (
...
+	for i := 1; i <= 5; i++ {
+		for _, res := range []serverResponse{accepted, ok, forbidden, unauthorized, requestTooLarge, rateLimited,
+			methodNotAllowed, tooManyConcurrentRequests, cantValidate(errors.New("")),
+			cantDecode(errors.New("")), cantPublish(errors.New(""))} {
+			sendStatus(w, req, res)
+			assert.Equal(t, int64(i), res.counter.Get())
+		}
+	}
+	assert.Equal(t, int64(55), responseCount.Get())
+	assert.Equal(t, int64(45), responseErrors.Get())
```
simitt: Nice test!

beater/handlers.go
```diff
@@ -284,76 +326,85 @@ func corsHandler(allowedOrigins []string, h http.Handler) http.Handler {
...
 
 	payload, err := processor.Decode(data)
 	if err != nil {
-		return http.StatusBadRequest, err
+		return cantDecode(err)
 	}
 
 	if err = report(pendingReq{payload: payload, config: config}); err != nil {
-		return http.StatusServiceUnavailable, err
+		return cantPublish(err)
```
simitt: Why don't you distinguish between `errChannelClosed` and `errFull` error here? I think a seperate counter for a full queue would totally make sense.

roncohen:
cool! Very nice to have an actual value for each type of response instead of relying of status-code + an error.

I suggest coming up with a prefix or postfix for the name of the variables for responses. At the moment we have `ok` and `accepted` as global variable names. Maybe `okResponse`, `acceptedResponse` etc.

beater/handlers.go
jalvz: with a `new` function how do you do the ones that take an error from the processor (eg cannotDecode), you cant create a counter in  `newServerResponse` because there can only be one monitor instance with the given name, so you have to pass it from the outside. you know what i mean?

beater/handlers.go
jalvz: mostly to keep the code simpler and not have to check here the error message from queueFull /  channelClosed, which can change and this wouldn't work anyways.
Im fine having them together because `errchannelClosed` only happens during shutdown and metrics are sent every 30 seconds, so only 1 doc for the whole process lifetime would have an inaccurate server full count.

but i ca separate them if you prefer

jalvz:
In case it went unnoticed, this starts returning `413s` upon too large requests (before was `400`)
How should i changelog it:  breaking change / added / bugfix ?

beater/handlers.go
simitt: As you are splitting all the others, and the `queue_full` is one of the most common/interesting errors I'd prefer to have them separated.

simitt:
I'd consider that a breaking change. Do you need to change the return code?

jalvz:
well, it is the correct code, and quite a relevant for our use case...

beater/handlers.go
simitt: that's true, would be more difficult

simitt:
Let's go with defining it as `bugfix` then.

jalvz:
thanks for the review, i think i implemented everything i could. can you have another look?

beater/handlers.go
```diff
@@ -52,17 +52,70 @@ type routeMapping struct {
...
+	fullQueueCounter  = counter("response.error.queue")
+	fullQueueResponse = func(err error) serverResponse {
+		return serverResponse{
+			errors.New("queue "), http.StatusServiceUnavailable, fullQueueCounter,
+		}
+	}
+	serverShuttingDownCounter  = counter("response.error.closed")
+	serverShuttingDownResponse = func(err error) serverResponse {
+		return serverResponse{
+			errors.New("timeout waiting to be processed"), http.StatusServiceUnavailable, serverShuttingDownCounter,
```
roncohen: error message should probably be "server is shutting down" here

beater/handlers.go
```diff
@@ -52,17 +52,70 @@ type routeMapping struct {
...
+		nil, http.StatusAccepted, counter("response.valid.acceptedResponse"),
+	}
+	forbiddenResponse = serverResponse{
+		errors.New("forbiddenResponse request"), http.StatusForbidden, counter("response.errors.forbiddenResponse"),
+	}
+	unauthorizedResponse = serverResponse{
+		errors.New("invalid token"), http.StatusUnauthorized, counter("response.errors.unauthorizedResponse"),
+	}
+	requestTooLargeResponse = serverResponse{
+		errors.New("request body too large"), http.StatusRequestEntityTooLarge, counter("response.errors.toolarge"),
```
roncohen: this counter is called `response.errors.toolarge`, but the one above is `response.errors.forbiddenResponse`. We should make it consistent whether we include `Response` postfix. I'm fine with either of the two

beater/handlers.go
```diff
@@ -52,17 +52,70 @@ type routeMapping struct {
...
+	methodNotAllowedResponse = serverResponse{
+		errors.New("only POST requests are supported"), http.StatusMethodNotAllowed, counter("response.error.method"),
+	}
+	tooManyConcurrentRequestsResponse = serverResponse{
+		errors.New("timeout waiting to be processed"), http.StatusServiceUnavailable, counter("response.error.concurrency"),
+	}
+	fullQueueCounter  = counter("response.error.queue")
+	fullQueueResponse = func(err error) serverResponse {
+		return serverResponse{
+			errors.New("queue "), http.StatusServiceUnavailable, fullQueueCounter,
```
roncohen: `queue is full`

PR #865, opened 2018-04-19
Monitor per agent

simitt:


_meta/beat.reference.yml
```diff
@@ -78,6 +78,8 @@ apm-server:
   # or the IP and User Agent of the real user (frontend requests). It defaults to true.
   #capture_personal_data: true
 
+  #agents: ["nodejs", "python"]
```
roncohen: this defaults to "nodejs", "python", "ruby", "apm-js", "java", "go"
Let's add the rest of the agents here so that the config file reflects the defaults

roncohen: also in `_meta/beat.yml`,

config/config.go
```diff
@@ -10,4 +10,41 @@ type Config struct {
 	LibraryPattern      *regexp.Regexp
 	ExcludeFromGrouping *regexp.Regexp
 	SmapMapper          sourcemap.Mapper
+	Agents              strContainer
```
roncohen: we've used a `set.Set` from github.com/fatih/set in other places. Maybe that fits here too?

beater/config.go
```diff
@@ -23,6 +23,7 @@ type Config struct {
 	Expvar              *ExpvarConfig   `config:"expvar"`
 	Frontend            *FrontendConfig `config:"frontend"`
 	AugmentEnabled      bool            `config:"capture_personal_data"`
+	Agents              []string        `config:"agents"`
```
roncohen: This config is only for metrics, so i suggest we put it under a metrics namespace:
```
metrics:
  agents: ....
```

or see if we can come up with another way to indicate that this config is used to determine which agents we'll be collecting metrics from.

roncohen: i think `agent_metrics` would be an OK name.

processor/metric.go
```diff
@@ -0,0 +1,56 @@
...
+		val.Inc()
+	}
+}
+
+func (m *metricMap) val(agent, key string) *monitoring.Int {
+	r, ok := m.metrics[agent]
+	if !ok {
+		m.mu.Lock()
+		defer m.mu.Unlock()
+		r, ok = m.metrics[agent]
```
roncohen: excellent 👍

config/config.go
simitt: We've only used the `set.Set` in tests so far. As this `Contains` is called for every request, I prefer to have a `map` over a `set` for performance reasons.

processor/metric.go
simitt: Sure with the config I can create it upfront.

However,, just for clarification - I don't understand why this should not be sure. 
Line 35 reads from the map, if it doesn't find the key, it uses a `sync.Mutex.Lock` and reads **again** to ensure no other thread is writing while reading. Only if the key is still not find, it is written. The writing always happens within the lock, so there is no scenario where read and write occurs at the same time. Can you elaborate on why this should not be threadsafe?

config/config.go
roncohen: `set.Set` uses a map for lookups internally.
It has a mutex and is thread-safe, so there's a slight performance penalty for that, but seems like a nice way to simplify this code a bit.

processor/metric.go
roncohen: The read on line 35 is not protected by a lock, so while another thread is writing to the map on line 41, a different thread could be reading on line 35

simitt:
Closing this for now. We should have a better overview of which metrics to monitor before implementing.

PR #868, opened 2018-04-20
Monitor req sizes

jalvz:


decoder/decoder.go
```diff
@@ -42,21 +65,33 @@ func readRequestJSONData(maxSize int64) Reader {
...
 			var err error
 			reader, err = gzip.NewReader(reader)
 			if err != nil {
 				return nil, err
 			}
 		}
+		if knownCLen {
+			uncompressedLengthAcc.Add(cLen)
+		}
+		io.Pipe()
```
simitt: Is this a leftover as you don't use the returned `PipeReader` and `PipeWriter`

decoder/decoder.go
```diff
@@ -42,21 +65,33 @@ func readRequestJSONData(maxSize int64) Reader {
...
+				gzipLengthAcc.Add(cLen)
+			}
 			var err error
 			reader, err = gzip.NewReader(reader)
 			if err != nil {
 				return nil, err
 			}
 		}
+		if knownCLen {
+			uncompressedLengthAcc.Add(cLen)
```
simitt: What do you think about moving this into the `switch` `default` section to only cover non-compressed requests and not have an overall counter?

decoder/decoder.go
```diff
@@ -13,18 +13,41 @@ import (
...
 
 type Reader func(req *http.Request) (io.ReadCloser, error)
 type Decoder func(req *http.Request) (map[string]interface{}, error)
 
+var (
+	decoderMetrics        = monitoring.Default.NewRegistry("apm-server.decoder", monitoring.PublishExpvar)
+	deflateLengthAcc      = monitoring.NewInt(decoderMetrics, "deflate.content-length")
+	gzipLengthAcc         = monitoring.NewInt(decoderMetrics, "gzip.content-length")
+	uncompressedLengthAcc = monitoring.NewInt(decoderMetrics, "uncompressed.content-length")
+	readerAcc             = monitoring.NewInt(decoderMetrics, "reader.size")
```
simitt: What does `Acc` stand for here?

decoder/decoder.go
jalvz: yes, is a leftover

decoder/decoder.go
jalvz: there isn't an overall counter, all the switch cases return from the function. but i can put this a a default.

decoder/decoder.go
jalvz: accumulator. i am not counting with `Inc()`, but accumulating up large numbers, i want to make explicit how these monitors are used.

decoder/decoder.go
jalvz: actually you are right. i don't know what i am saying.

PR #892, opened 2018-04-26
[6.2] Update beats framework to 7a45f64

graphaelli:


graphaelli:
will reopen when elastic/beats#6953 is merged

PR #894, opened 2018-04-30
 Update beats framework to 78f520f and update Go to 1.10.1

graphaelli:


graphaelli:
good call, will add to changelog and will track down those others.

PR #908, opened 2018-05-02
Add backport config

sqren:
This adds a config for [`backport`](https://github.com/sqren/backport) which removes the need for every user of the tool to create their own config.


sqren:
I don't have merge access. @jalvz can you do it?

PR #921, opened 2018-05-10
Update beats framework to 15d9539

graphaelli:
pulls in
* seccomp config
* updated kafka support (sarama update)


simitt:
WDYT about also updating the `_meta.beat.reference.yml` file for the [seccomp config](https://github.com/elastic/beats/pull/6832/files#diff-85a376edf7b31f5946ddeb979405932b)

graphaelli:
I floated that idea too but haven't heard back - maybe we do that as a follow on?

simitt:
SGTM - maybe you can change the note here that it is pulling it in but unofficially supported, otherwise it could be confusing if someone comes across this PR.

PR #923, opened 2018-05-11
 decoder: only augment with valid IPs

### LINKED ISSUE #920: Events are invalid if remote address is not an IP
The server optionally augments incoming data by adding the "system.ip" field to the client address: https://github.com/elastic/apm-server/blob/a37185ab9917e24822d693d0a01d5888fed046d8/decoder/decoder.go#L169

This fails if the remote address it not an IP. In that case, we'll get errors like:

> Cannot index event publisher.Event ... {"type":"mapper_parsing_exception","reason":"failed to parse [context.system.ip]","caused_by":{"type":"illegal_argument_exception","reason":"'pipe' is not an IP string literal."}}

This particular error was caused by the (WIP) introduction of an in-memory transport that I'm intending to use for the server to trace itself. The remote address is always "pipe".

I haven't tested it, but I expect this would also manifest when using the Unix socket listener support.

axw:
When augmenting the decoded payload, only add the remote address if it is a valid IP. Otherwise, the events fail indexing.

Also, make some minor performance improvements to utility.ExtractIP by pulling in some of the "RemoteAddr" function from https://github.com/elastic/apm-agent-go/tree/master/internal/apmhttputil, which has been optimized to eliminate allocations in typical code paths. Missing is the support for the Forwarded header, which is a little bit more involved.

Fixes #920

### DIFF
--- a/utility/ip.go
+++ b/utility/ip.go
@@ -10,20 +10,29 @@ import (
 // X-Forwarded-For has a list of IPs, of which the first is the one of the original client.
 // This value however might not be necessarily trusted, as it can be forged by a malicious user.
 func ExtractIP(r *http.Request) string {
-	realIP := r.Header.Get("X-Real-IP")
-	if realIP != "" {
+	if realIP := r.Header.Get("X-Real-Ip"); realIP != "" {
 		return realIP
 	}
-
-	forwardedFor := r.Header.Get("X-Forwarded-For")
-	client := strings.Split(forwardedFor, ",")[0]
-	if client != "" {
-		return strings.TrimSpace(client)
+	if xff := r.Header.Get("X-Forwarded-For"); xff != "" {
+		if sep := strings.IndexRune(xff, ','); sep > 0 {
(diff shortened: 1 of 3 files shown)


utility/ip.go
```diff
@@ -10,20 +10,29 @@ import (
 // X-Forwarded-For has a list of IPs, of which the first is the one of the original client.
 // This value however might not be necessarily trusted, as it can be forged by a malicious user.
 func ExtractIP(r *http.Request) string {
```
simitt: I like your original suggestion to rename to `ExtractRemoteAddr`, especially since this is also called in the [logHandler](https://github.com/elastic/apm-server/blob/master/beater/handlers.go#L227) and in the [IP ratelimiter](https://github.com/elastic/apm-server/blob/master/beater/handlers.go#L266). 
Then you could move `net.ParseIP(s)` to `utility.ExtractIP` which internally uses `utility.ExtractRemoteAddr`.

PR #925, opened 2018-05-11
[6.3] Update beats framework to 929dc48

simitt:


simitt:
jenkins, retest this please

_beats/testing/environments/args.yml
```diff
@@ -5,6 +5,6 @@ services:
   args:
     build:
       args:
-        DOWNLOAD_URL: https://snapshots.elastic.co/downloads
-        ELASTIC_VERSION: 6.3.0-SNAPSHOT
+        DOWNLOAD_URL: https://staging.elastic.co/6.3.0-46941e7c/downloads
```
graphaelli: I don't understand why this needs to happen

simitt: not sure either, but as this is pulled in from beats and only used for testing I'd leave it as is.

PR #927, opened 2018-05-11
utility: add RemoteAddr, change ExtractIP sig

axw:
Copy in the RemoteAddr function from apm-agent-go,
with a small change to how/where the Forwarded
header is parsed. Change ExtractIP to call RemoteAddr
and return net.ParseIP of the result.

### DIFF
--- a/beater/handlers.go
+++ b/beater/handlers.go
@@ -224,7 +224,7 @@ func logHandler(h http.Handler) http.Handler {
 			"method", r.Method,
 			"URL", r.URL,
 			"content_length", r.ContentLength,
-			"remote_address", utility.ExtractIP(r),
+			"remote_address", utility.RemoteAddr(r),
 			"user-agent", r.Header.Get("User-Agent"))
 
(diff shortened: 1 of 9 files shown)


beater/handlers.go
```diff
@@ -263,7 +263,7 @@ func ipRateLimitHandler(rateLimit int, h http.Handler) http.Handler {
 	}
 
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		if deny(utility.ExtractIP(r)) {
+		if deny(utility.RemoteAddr(r)) {
```
simitt: Could you also change https://github.com/elastic/apm-server/pull/927/files#diff-0ee6b4dd463eb95280cc3653bd4701b6L227 to `utility.RemoteAddr` ?

axw: Oops, I thought I did - must have reverted it by accident. Pushed a new commit. If you're merging, can you please squash? I haven't done it, as I don't want to mess up the review history.

PR #938, opened 2018-05-11
[6.x] Add backport config (#908)

sqren:
 - Add backport config  (#908)


graphaelli:
why do we need this on the 6.x branch?

sqren:
If you are on any other branch than master, backport currently won't work (since there is no backport config). This is the setup that works for Kibana. Thought it made sense here too but maybe people always work from master?

graphaelli:
Makes sense, just checking if there was some other reason I was missing.  LGTM

sqren:
@graphaelli can you merge?

PR #944, opened 2018-05-14
[6.3] decoder: only augment with valid IPs (#923)

simitt:
Backports the following commits to 6.3:
 - decoder: only augment with valid IPs  (#923)


simitt:
Merging this in after offline discussions with the @elastic/apm-server team.

Issue #863, opened 2018-04-19
Remove apm-server.reference.yml


simitt:
As discussed in earlier PRsthe goal is to get rid of `apm-server.reference.yml` for easier handling, as the file doesn't add any value.

PR #958, opened 2018-05-22
Remove apm-server.reference.yml

simitt:
* Update `beats` with the according changes (see https://github.com/elastic/beats/pull/7092)
* Combine `reference` and `apm-server` yml to one file. 

implements #863


jalvz:
This is fantastic, great job! 🎉 

You need to remove the reference to the reference yml file from the docs in `frontend`, `setting-up-and-running` and `shared-configuring` (just grep it)

simitt:
updated the docs that we copy from beats accordingly in https://github.com/elastic/beats/pull/7164

PR #967, opened 2018-05-24
Move output config options.

simitt:
Ensure the order is as expected for creating docker images.


simitt:
jenkins, retest this please

PR #969, opened 2018-05-28
add telemetry on waiting time for concurrent requests

### LINKED ISSUE #887: [telemetry] track time spent blocked by concurrent requests
Thoughts on a timer on wait time around https://github.com/elastic/apm-server/blob/fcd18547a6dae8e6c7cde88f49456ab7c5915bf6/beater/handlers.go#L154-L160 ?

jalvz:
fixes #887

### DIFF
--- a/beater/handlers_test.go
+++ b/beater/handlers_test.go
@@ -9,6 +9,11 @@ import (
 
 	"fmt"
 
+	"sync"
+	"time"
+
+	"strconv"
+
 	"github.com/stretchr/testify/assert"
 )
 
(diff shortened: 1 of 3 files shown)


beater/handlers_test.go
```diff
@@ -31,6 +36,28 @@ func TestIncCounter(t *testing.T) {
...
+	config := Config{ConcurrentRequests: 2, MaxRequestQueueTime: time.Second}
+	var wg sync.WaitGroup
+	h := concurrencyLimitHandler(&config, noopHandler{&wg})
+	for i := 1; i <= 3; i++ {
+		wg.Add(1)
+		go h.ServeHTTP(nil, nil)
+	}
+	wg.Wait()
+	// 3rd request must wait at least for one to complete (100ms ± sync margin that needs to be quite high for jenkins)
+	assert.InDelta(t, 100, concurrentWait.Get(), 70, strconv.FormatInt(concurrentWait.Get(), 10))
```
simitt: Why did you set delta to `70`?

jalvz: each request takes 100 ms, the third one will be queued for some amount of time until some request completes. 
how much exactly can't be known, depends on the scheduler, etc.

70 is an arbitrary & large enough number for jenkins (in local 10 is enough); meaning eg. that the first request can run for 70ms _before_ the third request comes, in which case it will have to wait 30 ms.

what matters in the end is that the counter is updated

simitt: I just thought it is a bit hard to read with having `70` set as arbitrary number, but I get your point.

jalvz: i agree, that is why i wrote the comment... any ideas to make it more clear?
the wait can be more or less than 100, depending on how the functions are scheduled. would eg. `concurrentWait.Get() > 20` be better?

jalvz: so main thread sends the requests in a `go` block. there can happen many things. 
how i see it (i might be wrong):

- A) one request is picked up by the scheduler and runs right away for 70ms before Go even starts to run another request. when that request finally fires and stumbles on the semaphore, it has to wait for the first one to finish, which only will be 100 - 70 = 30ms (less than 100 wait).

- B) scheduler picks one request, time starts to count (https://github.com/jalvz/apm-server/blob/4e1776a68241954533fbc7d2ba99799cf9cbfaa2/beater/handlers.go#L154) but before it arrives to the semaphore; Go schedules other request and that one enters the semaphore. the first one then will have to wait on the semaphore for the whole 100ms the last one takes plus a few extra milliseconds because it was actually scheduled first, but got preempted too quickly (more than 100ms wait).

jalvz:
jenkins retest this please

PR #972, opened 2018-05-29
Read sourcemap content and fill context lines

### LINKED ISSUE #457: Add actual source to stacktrace frame when applying sourcemapping
Add the `pre_context`, `post_context` and `context_line` information when sourcmap mapping is applied to a stacktrace frame of an incoming request.

jalvz:
fixes #457

### DIFF
--- a/sourcemap/mapper.go
+++ b/sourcemap/mapper.go
@@ -84,7 +84,7 @@ func (m *SmapMapper) NewSourcemapAdded(id Id) {
 }
 
 func subSlice(from, to int, content []string) []string {
-	if from > len(content) {
+	if len(content) == 0 {
 		return content
 	}
 	if from < 0 {
--- a/sourcemap/mapper.go
+++ b/sourcemap/mapper.go
@@ -4,9 +4,13 @@ import (
 	"fmt"
 	"time"
 
+	"strings"
+
 	"github.com/elastic/beats/libbeat/common"
 	"github.com/elastic/beats/libbeat/logp"
 )
+
+const sourcemapContentSnippetSize = 5
 
 type Mapper interface {
 	Apply(Id, int, int) (*Mapping, error)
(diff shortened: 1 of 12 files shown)


sourcemap/mapper.go
```diff
@@ -4,10 +4,14 @@ import (
 	"fmt"
 	"time"
 
+	"strings"
+
 	"github.com/elastic/beats/libbeat/common"
 	"github.com/elastic/beats/libbeat/logp"
 )
 
+const sourcemapContentSnippetSize = 5
```
simitt: This could have a huge impact on the used storage so I suggest to make this configurable.

sourcemap/mapper.go
```diff
@@ -71,3 +82,16 @@ func (m *SmapMapper) NewSourcemapAdded(id Id) {
 	}
 	m.Accessor.Remove(id)
 }
+
+func subSlice(from, to int, content []string) []string {
+	if from > len(content) {
+		return content
```
simitt: If I read correct, then in this case we would store the whole `content`. I suggest to return an empty array and to log an error.

jalvz: this is for when source content is not available, will rewrite to `if len(content) == 0` to make it clear; sorry for that.
not sure about logging an error, sourcemap content is optional (as per spec); i think some tools include it by default and others don't, users can choose.
you still think should be logged?

sourcemap/mapper.go
jalvz: you think so? this is in errors - not spans... and only for frontend with sources in uploaded sourcemaps
i didn't want to bloat more the configuration space, but i can make it configurable if you prefer

sourcemap/mapper.go
simitt: Oh ok. I agree that there is no need for logging when you check `if len(content) == 0 { return content }`, as this reflects something different than checking that `from` is not an overflow of the content array.

sourcemap/mapper.go
simitt: Sourcemapping is generally enabled for frontend endpoints, so it will also be enabled for `transactions`, not only for `errors`. 
If we don't add the config option, it will be hard to give advice to customers in case they want to use sourcemapping but reduce storage.

roncohen:
Awesome to have this in! We should probably add a changelog entry?

PR #974, opened 2018-05-29
clarify root cause for 403 errors, fixes #973

### LINKED ISSUE #973: Improve client-side 403 'forbidden request' error log message
`/v1/client-side` endpoint returns 403 "forbidden request" in two possible situations:

1. If the frontend endpoint is disabled.
2. If the origin header violates `allow_origins` configuration option.

It would make finding the root cause easier, if we clarify this in the error log message for that specific request. 
Possibly logging the origin header as well if it turns out to be the second situation.

see also: https://discuss.elastic.co/t/transactions-throwing-403-forbidden/133646

jalvz:


beater/handlers.go
```diff
@@ -341,7 +344,7 @@ func corsHandler(allowedOrigins []string, h http.Handler) http.Handler {
 			h.ServeHTTP(w, r)
 
 		} else {
-			sendStatus(w, r, forbiddenResponse)
+			sendStatus(w, r, forbiddenResponse(errors.New(origin+" is not allowed")))
```
jahtalab: I would change this to something like `"origin: '" + origin + "' is not allowed"` in case the origin header is empty.

PR #976, opened 2018-05-29
[6.3] clarify root cause for 403 errors, fixes #973 (#974)

### LINKED ISSUE #973: Improve client-side 403 'forbidden request' error log message

jalvz:
Backports the following commits to 6.3:
 - clarify root cause for 403 errors, fixes elastic/apm-server#973  (#974)

### DIFF
--- a/beater/handlers.go
+++ b/beater/handlers.go
@@ -74,8 +74,11 @@ var (
 	acceptedResponse = serverResponse{
 		nil, http.StatusAccepted, counter("response.valid.accepted"),
 	}
-	forbiddenResponse = serverResponse{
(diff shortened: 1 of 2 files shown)


jalvz:
jenkins retest this please

PR #981, opened 2018-05-30
override wait_until function with a try/except block

jalvz:
to not abort on 503 errors

fixes #979

PR #982, opened 2018-05-30
[6.x] override wait_until function with a try/except block to not abort on 503 errors (#981)

jalvz:
 - override wait_until function with a try/except block to not abort on 503 errors  (#981)


jalvz:
mostly to avoid possible merge conflicts in the future

Issue #954, opened 2018-05-15
Document index lifecycle management tips


graphaelli:
eg:

```
# list all apm indices
$ curator_cli --host elasticsearch show_indices --filter_list '[{"filtertype":"pattern","kind":"prefix","value":"apm-"}]'

apm-6.3.0-error-2018.05.10
apm-6.3.0-error-2018.05.11
apm-6.3.0-error-2018.05.12
apm-6.3.0-error-2018.05.13
apm-6.3.0-error-2018.05.14
apm-6.3.0-error-2018.05.15
apm-6.3.0-sourcemap
apm-6.3.0-span-2018.05.10
apm-6.3.0-span-2018.05.11
apm-6.3.0-span-2018.05.12
apm-6.3.0-span-2018.05.13
apm-6.3.0-span-2018.05.14
apm-6.3.0-span-2018.05.15
apm-6.3.0-transaction-2018.05.10
apm-6.3.0-transaction-2018.05.11
apm-6.3.0-transaction-2018.05.12
apm-6.3.0-transaction-2018.05.13
apm-6.3.0-transaction-2018.05.14
apm-6.3.0-transaction-2018.05.15

# delete > 5 days old by name
$ curator_cli --host elasticsearch delete_indices --filter_list '[{"filtertype":"pattern","kind":"prefix","value":"apm-"}, {"filtertype":"age","source":"name","timestring":"%Y.%m.%d","unit":"days","unit_count":5,"direction":"older"}]'

INFO      Deleting selected indices: [u'apm-6.3.0-span-2018.05.10', u'apm-6.3.0-error-2018.05.10', u'apm-6.3.0-transaction-2018.05.10']
INFO      ---deleting index apm-6.3.0-span-2018.05.10
INFO      ---deleting index apm-6.3.0-error-2018.05.10
INFO      ---deleting index apm-6.3.0-transaction-2018.05.10
INFO      "delete_indices" action completed.
```

PR #983, opened 2018-05-30
document curator tips useful for APM data

jalvz:
fixes elastic/apm-server#954

### DIFF
--- a/docs/common-problems.asciidoc
+++ b/docs/common-problems.asciidoc
@@ -32,6 +32,7 @@ you have a few options:
 * <<tune-output-config,Tune the APM Server output configuration to your cluster>>
 * <<increase-cluster-ingest,Tune Elasticsearch for higher ingestion>>
 * <<increase-queue-size,Increase the size of the APM Server queue>>
+* <<delete-old-data,Delete old data>>
 
 [float]
 [[request-timed-out]]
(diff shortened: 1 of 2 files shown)


docs/common-problems.asciidoc
```diff
@@ -96,6 +97,48 @@ and it alleviates problems that might result from sudden spikes of data.
 You can increase the queue size by overriding `queue.mem.events`.
 Be aware that increasing `queue.mem.events` can significantly affect APM Server memory usage.
 
+[float]
+[[delete-old-data]]
+==== Delete old data
+
+A late symptom of a full queue is that Elasticsearch ran out of disk space and started rejecting insertions.
```
dedemorton: We are talking about the internal queue, right? I'd make this statement a bit more direct and say something like:

A full internal queue might cause Elasticsearch to run out of disk space and start rejecting insertions.

docs/common-problems.asciidoc
```diff
@@ -96,6 +97,48 @@ and it alleviates problems that might result from sudden spikes of data.
...
+In addition to other storage optimization measures you will likely need to remove old indices to make room for new data.
```
dedemorton: IMO, it's clear from the context that other optimizations are necessary. I would prune this a bit (to make it easier to scan) and simply say something like: "If this happens, you will likely need to remove old indices to make room for new data."

docs/common-problems.asciidoc
```diff
@@ -96,6 +97,48 @@ and it alleviates problems that might result from sudden spikes of data.
...
+You can achieve this with a tool like Curator,
```
dedemorton: Again, I'd pare away some of the words: To do this, you can use a tool like Curator, and set up a cron job to run it periodically.

docs/common-problems.asciidoc
jalvz: Thanks for taking the time to review the rest as well.
So this is the other way around: Elasticsearch runs out of disk space, and because of it it starts rejecting insertions, and because of it the queue fills up (because nothing gets pulled out). 
Users then get this "queue is full" message in their app logs. 
In this case the root cause is that the disk that elasticsearch is writing to is full.
How would you phrase it better?

dedemorton: Ah, that makes sense! The original sentence was  a little ambiguous. :-) How about something like:

The internal queue might be full because Elasticsearch ran out of disk space and started rejecting insertions. If this happens, you will likely need to remove old indices to make room for new data.

PR #984, opened 2018-05-31
[6.3] Update beats framework to d07c2f2

simitt:
updates to latest beats version and removes non-existant libbeat packages:
-  "github.com/elastic/beats/libbeat/asset"
- "github.com/elastic/beats/libbeat/common/seccomp"
- "github.com/elastic/beats/libbeat/common/transport/tlscommon"


ph:
@simitt LGTM, I've checked they should not be in 6.3 I believe you are right it should have been a vendor error at some point.

PR #996, opened 2018-06-07
remove duplicate json schema tests

graphaelli:
removes `Test*ProcessorValidationFailed` duplicate of a test in the respective `*/package_tests/json_schema_test.go`

### DIFF
--- a/tests/json_schema.go
+++ b/tests/json_schema.go
@@ -158,6 +158,8 @@ func TestDataAgainstProcessor(t *testing.T, p processor.Processor, testData []Sc
 		data, err := loader.LoadData(d.File)
 		assert.Nil(t, err)
 		err = p.Validate(data)
-		assert.Contains(t, err.Error(), d.Error)
+		if assert.Error(t, err) {
+			assert.Contains(t, err.Error(), d.Error)
+		}
 	}
 }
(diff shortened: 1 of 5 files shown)


tests/json_schema.go
```diff
@@ -158,6 +158,8 @@ func TestDataAgainstProcessor(t *testing.T, p processor.Processor, testData []Sc
 		data, err := loader.LoadData(d.File)
 		assert.Nil(t, err)
 		err = p.Validate(data)
-		assert.Contains(t, err.Error(), d.Error)
+		if assert.Error(t, err) {
```
jalvz: how this is removing duplicates?

jalvz:
I don't see how this removes duplicated tests. The jsons in `valid` are positive tests (must validate), and the ones in `invalid` are negative tests (must not validate).
Am I missing something?

tests/json_schema.go
simitt: I generally agree that the tests are duplicates. However, the removed tests did check that `p.Validate(data)` returns an error. Removing those tests and changing this check to `if assert.Error(t, err).. ` we lose the check that the validation fails.

graphaelli: That is confusing.  It doesn't actually lost the check that that validation fails, it just prevents the `assert.Contains` from panicking when err is actually nil.

graphaelli: ```An error is expected but got nil.``` is produced in that case

simitt: ah sry didn't get that.

Issue #947, opened 2018-05-15
Dropped transactions or errors can lead to user confusion


jalvz:
A typical case is:

A user makes a bad deploy and as a result their app starts sending too many errors to the apm-server.
apm-server will saturate and return 503's for errors and also transactions.
User then will see a drop in transactions when the errors spike, and think that the errors are causing the transaction drop.
What happens actually is that the apm-server can't keep up and can't prioritize transactions over errors.

alvarolobato:
The problem here is that you could lose many transactions if the errors spike, because errors would fill the queue and transactions would start to be discarded. This can lead you to draw non accurate conclusions from the data.

Having two queues could improve this situation but could also cause the two time of data to be out of sync in terms of time if one of the queues fall behind of the other. This can lead to more confusion to the user.

One possible solution could be trying to reduce the number of dropped transactions under heavy load, maybe by enabling overflow to disk in beats.

A possible way to improve the situation could be adding a indication in UI with the number of dropped transactions and errors.

We will timebox some time to investigate on a way to improve this situation and propose a way to improve it.

jalvz:
> Having two queues could improve this situation but could also cause the two time of data to be out of sync in terms of time if one of the queues fall behind of the other

@alvarolobato what do you mean with that?

alvarolobato:
@jalvz this was the summary of a discussion during a team meeting.
Basically if you have two queues and you start getting very different load you may start seeing up to date transactions but not getting the errors or the errors from an hour ago (or the opposite), because one of the queues is almost empty and the other one is not being able to keep up with the load.

This causes a bad user experience and may be miss-leading precisely in moments where you need to draw quick conclusions ie. a lot of errors coming from the monitored application.

alvarolobato:
I don't know either my comment was a summary of the pros and cons discussed during the meeting.

alvarolobato:
Putting back to the backlog and decide after all the changes to the V2 protocol and the refactor.

simitt:
Moving this to backlog as I don't think we are blocked here atm, but rather postponed to work on this.

graphaelli:
agreed

PR #997, opened 2018-06-08
Multiple queues in the publisher

jalvz:
This attempts to explore https://github.com/elastic/apm-server/issues/947

Made some measurements in GCP, variance is quite high so take the results with a grain of salt. For the tables bellow I selected results falling in the middle.
This approach prioritizes transactions over errors, overall performance respect to master seems better with `concurrent_requests = 5` (default) but is not so clear with `concurrent_requests = 10`.

There are many more variables untested (other relevant settings like`mem.queue.events`, other values for `concurrent_requests`, other workload patterns, etc.). Given the variability of the results and the amount of knobs we have, exhaustive testing is not really doable.
This code is also more complex (and adds another knob to the mix) so I am reluctant to go with this. 
Happy to see alternative implementations.

Besides, I found that the best runs with `concurrent_requests=10` are much better than the best runs with the default, and with a similar memory usage. Did that has been observed/tested before?

______

- duration of each test = 1min
- transactions payload = ~900kb, 3 workers
- errors payload = ~200kb, 10 workers
- first table: `concurrent_requests=5`, second table: `concurrent_requests=10`
- for each table: first column is `master`, and second is this branch.
- for each table: first row is transactions only, and second row is transactions + errors.

- things to notice: latency and ratio of 503's between rows and columns - as expected ratio of 503's between errors and transactions is higher in this branch.

```
http://localhost:8200/v1/transactions
pushed 3.4Mb / sec , accepted 2.1Mb / sec        pushed 3.5Mb / sec , accepted 2.7Mb / sec
  [202] 139 responses (60.96%)                     [202] 182 responses (76.79%)
  [503] 89 responses (39.04%)                      [503] 55 responses (23.21%)
  total 228 responses (3.80 rps)                   total 237 responses (3.95 rps)
  431.66 ms / request                              329.67 ms / request
396.5Mb (max RSS)                                377.8Mb (max RSS)


http://localhost:8200/v1/transactions
pushed 3.3Mb / sec , accepted 1.9Mb / sec        pushed 3.6Mb / sec , accepted 2.7Mb / sec
  [202] 126 responses (56.00%)                     [202] 183 responses (75.31%)
  [503] 99 responses (44.00%)                      [503] 60 responses (24.69%)
  total 225 responses (3.75 rps)                   total 243 responses (4.05 rps)
  476.19 ms / request                              327.87 ms / request
http://localhost:8200/v1/errors
pushed 1.0Mb / sec , accepted 496.7kb / sec      pushed 1.0Mb / sec , accepted 485.6kb / sec
  [202]	134 responses (47.02%)                     [202] 131 responses (46.45%)
  [503]	151 responses (52.98%)                     [503] 151 responses (53.55%)
  total	285 responses (5.70 rps)                   total 282 responses (5.64 rps)
  373.14 ms / request                              381.68 ms / request
445.0Mb (max RSS)                                463.8Mb (max RSS)                        
```

```
http://localhost:8200/v1/transactions
pushed 3.4Mb / sec , accepted 2.1Mb / sec        pushed 3.4Mb / sec , accepted 2.3Mb / sec
  [202] 140 responses (61.40%)                     [202] 156 responses (67.24%) 
  [503] 88 responses (38.60%)                      [503] 76 responses (32.76%) 
  total 228 responses (3.80 rps)                   total 232 responses (3.87 rps) 
  428.57 ms / request                              384.62 ms / request
421.3Mb (max RSS)                                426.6Mb (max RSS)


http://localhost:8200/v1/transactions
pushed 3.3Mb / sec , accepted 1.8Mb / sec        pushed 3.4Mb / sec , accepted 1.9Mb / sec 
  [202] 119 responses (53.85%)                     [202] 128 responses (56.14%)
  [503] 102 responses (46.15%)                     [503] 100 responses (43.86%)
  total 221 responses (3.68 rps)                   total 228 responses (3.80 rps)
  504.24 ms / request                              468.75 ms / request
http://localhost:8200/v1/errors
pushed 1.9Mb / sec , accepted 630.1kb / sec      pushed 1.8Mb / sec , accepted 215.0kb / sec
  [202] 170 responses (32.88%)                     [202] 58 responses (11.72%)
  [503] 347 responses (67.12%)                     [503] 437 responses (88.28%) 
  total 517 responses (10.34 rps)                  total 495 responses (9.90 rps)
  294.14 ms / request                              862.09 ms / request
456.4Mb (max RSS)                                461.0Mb (max RSS)
```

### DIFF
--- a/beater/handlers.go
+++ b/beater/handlers.go
@@ -387,7 +387,7 @@ func processRequest(r *http.Request, pf ProcessorFactory, config conf.Config, re
 		return cannotDecodeResponse(err)
 	}
 
-	if err = report(pendingReq{payload: payload, config: config}); err != nil {
+	if err = report(pendingReq{payload: payload, config: config, processorName: processor.Name()}); err != nil {
 		if strings.Contains(err.Error(), "publisher is being stopped") {
 			return serverShuttingDownResponse(err)
 		}
--- a/beater/pub.go
+++ b/beater/pub.go
@@ -7,9 +7,15 @@ import (
 
 	"github.com/pkg/errors"
 
+	"math"
+
 	"github.com/elastic/apm-server/config"
 	pr "github.com/elastic/apm-server/processor"
+	perr "github.com/elastic/apm-server/processor/error"
+	"github.com/elastic/apm-server/processor/sourcemap"
+	"github.com/elastic/apm-server/processor/transaction"
 	"github.com/elastic/beats/libbeat/beat"
+	"fmt"
 )
 
 // Publisher forwards batches of events to libbeat. It uses GuaranteedSend
@@ -20,15 +26,18 @@ import (
 // number requests(events) active in the system can exceed the queue size. Only
 // the number of concurrent HTTP requests trying to publish at the same time is limited.
 type publisher struct {
-	pendingRequests chan pendingReq
-	client          beat.Client
-	m               sync.RWMutex
-	stopped         bool
+	errorsC       chan pendingReq
+	transactionsC chan pendingReq
+	sourcemapsC   chan pendingReq
+	client        beat.Client
+	m             sync.RWMutex
+	stopped       bool
 }
 
 type pendingReq struct {
-	payload pr.Payload
-	config  config.Config
+	processorName string
+	payload       pr.Payload
+	config        config.Config
 }
 
 var (
@@ -55,13 +64,15 @@ func newPublisher(pipeline beat.Pipeline, N int, shutdownTimeout time.Duration) 
 	if err != nil {
 		return nil, err
 	}
-
+	transactionsCSize := N - 1
+	errorsCSize := int(math.Max(math.Trunc(float64(transactionsCSize) * 0.2), 1))
+	fmt.Println("transSize  ", transactionsCSize)
+	fmt.Println("errSize ", errorsCSize)
 	p := &publisher{
-		client: client,
-
-		// Set channel size to N - 1. One request will be actively processed by the
-		// worker, while the other concurrent requests will be buffered in the queue.
-		pendingRequests: make(chan pendingReq, N-1),
+		client:        client,
+		transactionsC: make(chan pendingReq, transactionsCSize),
+		errorsC:       make(chan pendingReq, errorsCSize),
+		sourcemapsC:   make(chan pendingReq, 1),
 	}
 
 	for i := 0; i < runtime.GOMAXPROCS(0); i++ {
@@ -78,7 +89,9 @@ func (p *publisher) Stop() {
 	p.m.Lock()
 	p.stopped = true
 	p.m.Unlock()
-	close(p.pendingRequests)
+	close(p.transactionsC)
+	close(p.errorsC)
+	close(p.sourcemapsC)
 	p.client.Close()
 }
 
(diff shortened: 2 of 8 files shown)


jalvz:
Ah, sorry, I didn't know you had that in mind. Then this comment https://github.com/elastic/apm-server/issues/947#issuecomment-395413333 makes more sense (i think).

If the goal is to prioritize transactions over errors, it _looks_ that this does the job - as per tables above:
`master`: `44.00%` `503`'s in transactions / `52.98%` in errors
this branch: `24.69%` `503`'s in transactions / `53.55%` in errors

`master`: `46.15%` `503`'s in transactions / `67.12%` in errors
this branch: `43.86%` `503`'s in transactions / `88.28%` in errors.

So this branch accepts >2X more transactions than errors (for these tests), which was the intention.
My point is that it is more complicated, hard to test and hard to prove how/if it helps users after all.

I don't mind trying out different clients, but that seems even more complicated and I don't understand why should be better - at the end is the same elasticsearch cluster with the same ingestion capacity, same network, etc.

simitt:
Maybe I misunderstand the purpose of the PR, so I elaborate a bit on my thoughts:
I don't think that prioritising transactions over errors or the other way around should be the goal here. Instead we want to avoid reducing the ingestion rate for transactions if error rate suddenly increases (and the other way around). When sharing one internal libbeat queue, that has a max size it is likely that in the event of a sudden increase of documents for one doc type the ingestion rate for the other doc types decreases, as the queue size is limited. When having a separate client per type, where every client has its own queue a sudden increase of documents e.g. errors would fill up the error queue, so errors would start getting dropped. But transactions could be processed normally.

Now I agree that if we generally overload the ES ingestion, transactions and errors would still not be isolated from each other, but that's why we have a max queue size to prevent this.

jalvz:
Some more data:
Top graph shows a constant rate of transactions (15/sec) and how we drop them when the apm-server is overloaded with errors. Bottom graph shows the corresponding error count.

It more or less matches my expectations. @simitt @graphaelli does it make sense to you? 
Any other scenarios you think should be tested?

transactions-annotated
errors

graphaelli:
After clarifying with @jalvz, the results make sense to me.  The top row is the number of transactions ingested, the bottom is the number of errors.   Each column is the result of applying a massive error spike on top of a constant 15/s transaction load against difference branches/configs.

I like the idea of allowing users to specify a preference for dropping transactions or errors (or metrics ..) in the face of memory exhaustion, depending on their needs.  I'm struggling with allocating fixed sized queues for this purpose, and exposing such specific tunables to accomplish that.

Why should apm-server drop errors when there's plenty of memory to handle them otherwise?  Does this problem go away completely with a disk-backed queue?  Does moving to 1 transaction per request, like errors are in practice, change this into strictly a rate limiting problem?

jalvz:
I am very biased by my experience, but I don't see a lot of value in ingesting 200 errors per minute instead of 100, when all are the same. 

Since how little difference is between 0.8 transactions - 0.2 queues; and 0.5 each (3rd and 4th blocks) I would spare customers another config option and just hardcode a ratio.

jalvz:
I tried to use several clients and It doesn't work. I believe it is because `beat.Pipeline` is a stateful object and I can't get more than 1 client out of it. I only get one beat.Pipeline from the `beat.Beat`.

My attempt is here: https://github.com/elastic/apm-server/compare/master...jalvz:multiple-clients?expand=1

Is there something obviously wrong there? Any other ideas?

simitt:
Your implementation generally LGTM - that's basically what I meant to separate the two clients. When `beat.Pipeline` only allows for 1 client and you don't see an improvement on isolating transactions and errors from each other with this approach, I agree that it doesn't make too much sense.

jalvz:
Closing this to explore https://github.com/elastic/apm-server/issues/1057 first.

PR #1000, opened 2018-06-12
Metrics

graphaelli:
Add experimental `/v1/metrics` endpoint to enable agents to submit application level metrics along with transaction and error data.  It is enabled by default and can be disabled by setting `apm-server.metrics.enabled=false`.


axw:
Looks good, apart from the discrepancy between spec and how values are being stored.

processor/metric/payload.go
```diff
@@ -0,0 +1,153 @@
...
+func (pa *Payload) Transform(conf config.Config) []beat.Event {
+	if pa == nil {
+		return nil
+	}
+
+	var events []beat.Event
+	for _, metric := range pa.Metrics {
+		samples := common.MapStr{}
+		for _, sample := range metric.Samples {
+			samples[sample.Name()] = sample.Value()
```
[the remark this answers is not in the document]
graphaelli: oh my, fixing.  Looking at the spec closely now, we have `value/count` in-take and `value/value` indexed - thoughts?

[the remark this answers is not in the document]
axw: > Since this information is for downstream systems - UI/rollups - using value for both would make it possible to change between counter and gauge types for the same metric without need to reindex anything. Of course changing back and forth is error prone, but changing once, after first trying to use the data, seems like a common case.

That sounds like a good argument. Also, the beats metrics proposal uses value in both places.

> So if we use value/value here, should it be value/value in intake also?

I have no strong opinion on this. Probably best to keep it the same in both places.

docs/spec/metrics/sample.json
```diff
@@ -0,0 +1,28 @@
...
+        "object"
+    ],
+    "description": "A single metric sample.",
+    "oneOf": [
+        {
+            "additionalProperties": false,
+            "properties": {
+                "type": {"enum": ["counter"]},
+                "unit": {"type": "string"},
+                "count": {"type": "number"}
```
jalvz: fwiw i also vote for calling this one `value` too

processor/metric/package_tests/TestProcessMetricMinimal.approved.json
```diff
@@ -0,0 +1,30 @@
...
+            "context": {
+                "service": {
+                    "agent": {
+                        "name": "elastic-go",
+                        "version": "0.1.0"
+                    },
+                    "name": "apm-server"
+                }
+            },
+            "metric": {
```
jalvz: Our convention is that "minimal" payloads are only have the required attrs, in this case would be just the gauge or the counter metric, but not the 2.
Look eg. for errors with have a minimal example with an exception and another with a log

processor/metric/payload.go
```diff
@@ -0,0 +1,217 @@
...
+		"type":  "gauge",
+		"value": g.value,
+	}
+	if g.unit != nil {
+		v["unit"] = g.unit
+	}
+	return v
+}
+
+func CounterFrom(name string, sample map[string]interface{}) *Counter {
```
jalvz: To avoid some repetition you could have a function that returns `(string, float64, *string, error)` and then create a and then the `*From` functions just create `Counter` or `Gauge` out of that.

processor/metric/payload.go
```diff
@@ -0,0 +1,217 @@
...
+	sl := make([]Sample, 0)
+
+	samples := raw["samples"].(map[string]interface{})
+	for name, s := range samples {
+		var sample Sample
+		sampleMap := s.(map[string]interface{})
+		sampleType, ok := sampleMap["type"]
+		if !ok {
+			// json schema should prevent this
+			logp.NewLogger("transform").Debugf("unexpected sample: %s: %s", name, s)
```
jalvz: should be a warning?

processor/metric/payload.go
```diff
@@ -0,0 +1,217 @@
...
+
+func DecodePayload(raw map[string]interface{}) (*Payload, error) {
+	if raw == nil {
+		return nil, nil
+	}
+	pa := Payload{}
+
+	var err error
+	service, err := m.DecodeService(raw["service"], err)
+	if service != nil {
```
jalvz: I don't think you need this `if`, is redundant with the one in L171

processor/metric/payload.go
```diff
@@ -0,0 +1,217 @@
...
+		case "counter":
+			if c := CounterFrom(name, sampleMap); c != nil {
+				sample = c
+			}
+		case "gauge":
+			if g := GaugeFrom(name, sampleMap); g != nil {
+				sample = g
+			}
+		}
+		if sample != nil {
```
jalvz: iirc we don't have this kind of behaviour for other types, right? we either accept it all or reject all if 1 item in the list is wrong (i think)

I personally prefer the way here, but we should be consistent imo

graphaelli:
jenkins, retest this please

graphaelli:
jenkins, retest this please

processor/metric/package_tests/TestProcessMetricMinimal.approved.json
graphaelli: fair enough, I'll trim it further.

processor/metric/payload.go
graphaelli: @simitt @roncohen thoughts on this?  We require one of the known types via json schema so the entry to this would either be a mistake there or programmatic usage of this api.  I think individual metrics should be ignored rather than failing the whole set of metrics but I agree with @jalvz that consistency with the other endpoints is important.

docs/spec/metrics/sample.json
graphaelli: @roncohen thoughts?  I don't feel strongly either way and this would be a quick update.

processor/metric/payload.go
```diff
@@ -0,0 +1,217 @@
...
+		ev := beat.Event{
+			Fields: common.MapStr{
+				"processor": processorEntry,
+				"context":   context,
+				"metric":    samples,
+			},
+			Timestamp: metric.Timestamp,
+		}
+		events = append(events, ev)
+	}
```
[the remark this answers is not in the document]
graphaelli: done

beater/config.go
```diff
@@ -111,6 +120,7 @@ func replaceVersion(pattern, version string) string {
 }
 
 func defaultConfig(beatVersion string) *Config {
+	metricsEnabled := true
```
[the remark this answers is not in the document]
graphaelli: Why?  It's token protected just like the other endpoints.  I'll send you a link to the chat where this was  discussed.

docs/spec/metrics/metric.json
```diff
@@ -0,0 +1,39 @@
...
+            },
+            "additionalProperties": false
+        },
+        "tags": {
+            "type": ["object", "null"],
+            "description": "A flat mapping of user-defined tags with string values.",
+            "regexProperties": true,
+            "patternProperties": {
+                "^[^*\"]*$": {
+                    "type": "string",
```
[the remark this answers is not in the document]
graphaelli: I'm not clear on what you're proposing here.  Does this mean we should support a payload like this?:
```
tags: {
  "foo": null
}
```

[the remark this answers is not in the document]
simitt: Yes thats what we usually allow. The null values are removed during the server side processing, we are just less strict on the intake side. If this is not allowed, the whole request would fail.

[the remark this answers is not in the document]
graphaelli: Ok, I can live with that, will update.

processor/metric/package_tests/TestProcessMetricMinimal.approved.json
graphaelli: done

graphaelli:
closing while making significant changes, will reopen for review when done.

docs/spec/metrics/metric.json
```diff
@@ -0,0 +1,38 @@
...
+            },
+            "additionalProperties": false
+        },
+        "timestamp": {
+            "type": "string",
+            "format": "date-time",
+            "pattern": "Z$",
+            "description": "Recorded time of the metric, UTC based and formatted as YYYY-MM-DDTHH:mm:ss.sssZ"
+        }
+    }
```
simitt: The comment above regarding adding `null` as a possible type for all properties should also apply to all the other properties.

processor/metric/payload.go
```diff
@@ -0,0 +1,300 @@
...
+	errMissingKey = errors.New("key not found")
+	errNan        = errors.New("not a json number")
+)
+
+type Sample interface {
+	Name() string
+	Value() common.MapStr
+}
+
+func floatFrom(key string, m map[string]interface{}) (float64, error) {
```
simitt: Why not using `utility/data_fetcher` here? https://github.com/elastic/apm-server/blob/master/utility/data_fetcher.go#L19

processor/metric/payload.go
```diff
@@ -0,0 +1,300 @@
...
+		return 0, errMissingKey
+	}
+	n, ok := v.(json.Number)
+	if !ok {
+		return 0, errNan
+	}
+	return n.Float64()
+}
+
+func stringPtrFrom(key string, m map[string]interface{}) *string {
```
simitt: same - why not using the `utility/data_fetcher` here? https://github.com/elastic/apm-server/blob/master/utility/data_fetcher.go#L69

processor/metric/payload.go
```diff
@@ -0,0 +1,300 @@
...
+	return s.name
+}
+
+func (s *Summary) Value() common.MapStr {
+	v := common.MapStr{
+		"type":  "summary",
+		"count": s.count,
+		"sum":   s.sum,
+	}
+	if s.unit != nil {
```
simitt: Why not using the `utilty.MapStrEnhancer` https://github.com/elastic/apm-server/blob/master/utility/map_str_enhancer.go#L10?

processor/metric/payload.go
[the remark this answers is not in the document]
simitt: For the other endpoints we also sum up the number of events per transformation - not sure if this makes sense here, just want to ensure you haven't overlooked it.

tests/data/valid/metric/null.json
```diff
@@ -0,0 +1,25 @@
...
+            "counter.null.unit": {
+                "type": "counter",
+                "value": 6.520832e+06,
+                "unit": null
+            }
+        },
+        "tags": null,
+        "timestamp": "2017-05-30T18:53:42.281Z"
+    }]
+}
```
simitt: Could you adapt this example to hold null values for all non-required properties? (Except you really don't want to allow null values there).

processor/metric/payload.go
graphaelli: Probably because I forgot it existed, I'll rework this.

docs/spec/metrics/metric.json
```diff
@@ -0,0 +1,39 @@
...
+    "description": "Metric data captured by an APM agent",
+    "properties": {
+        "samples": {
+            "type": "object",
+            "description": "Sampled application metrics collected from the agent",
+            "regexProperties": true,
+            "patternProperties": {
+                "^[^*\"]*$": {
+                    "$ref": "sample.json",
+                    "maxLength": 1024
```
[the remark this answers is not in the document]
graphaelli: Thanks for explanation, that makes sense to me now.  Is there any sense in restricting the length of metric names?

processor/metric/payload.go
graphaelli: finally refactored this while adding summaries

processor/metric/payload.go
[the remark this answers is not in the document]
graphaelli: Thanks, I did see that but I don't see much value in tracking that.  Happy to add it in if someone makes a case.

tests/data/valid/metric/null.json
graphaelli: That won't be possible as metric has a bunch of nesting.  I'll add a few combinations alongside this one.

processor/metric/summary.go
```diff
@@ -34,6 +81,18 @@ func (s *summary) mapstr() common.MapStr {
 	utility.Add(v, "min", s.min)
 	utility.Add(v, "stddev", s.stddev)
 	utility.Add(v, "unit", s.unit)
+	if s.quantiles != nil {
+		quantiles := common.MapStr{}
+		for q, value := range s.quantiles {
+			// index quantiles as percentiles
+			k := fmt.Sprintf("%f", q*100)
+			k = strings.TrimRight(strings.TrimRight(k, "0"), ".") // 99.0000 -> 99
```
axw: alternatively: `strconv.FormatFloat(q*100, 'f', -1, 64)`

processor/metric/generated/schema/payload.go
```diff
@@ -0,0 +1,228 @@
...
+                "^[^*\"]*$": {
+                        "$schema": "http://json-schema.org/draft-04/schema#",
+    "$id": "docs/spec/metrics/sample.json",
+    "type": ["object", "null"],
+    "description": "A single metric sample.",
+    "oneOf": [
+        {
+            "properties": {
+                "type": {"enum": ["counter", "gauge"]},
+                "unit": {"type": ["string", "null"]},
```
felixbarny: A `description` about what the unit affects would be helpful. For example does it influence the way Kibana formats the values? Which are the special values here? For example `byte` vs. `bytes`, `kb` vs. `KiB`.

processor/metric/package_tests/json_schema_test.go
```diff
@@ -0,0 +1,87 @@
...
+								"count":     1,
+								"sum":       1,
+								"quantiles": val{val{0, 6.12}, val{100, 10}, val{0.5, 10}}}}},
+						},
+					},
+				},
+			},
+		},
+	}
+	procSetup.DataValidation(t, payloadData)
```
simitt: You are missing a test for the required/optional attributes like https://github.com/elastic/apm-server/blob/master/processor/transaction/package_tests/json_schema_test.go#L20

utility/map_str_enhancer.go
```diff
@@ -133,3 +137,12 @@ func MillisAsMicros(ms float64) common.MapStr {
 	m["us"] = int(ms * 1000)
 	return m
 }
+
+func Prune(m common.MapStr) common.MapStr {
```
simitt: When using `utility.Add` this should actually not be necessary, as the `Add` method only adds key-value pairs if the value is not null or an empty object.

processor/metric/package_tests/json_schema_test.go
simitt: You are missing tests for optional/required attributes like https://github.com/elastic/apm-server/blob/master/processor/transaction/package_tests/json_schema_test.go#L20.

processor/metric/generated/schema/payload.go
graphaelli: Good idea, lacking in descriptions in general here.

docs/spec/metrics/sample.json
```diff
@@ -0,0 +1,40 @@
...
+                "stddev": {"type": ["number", "null"]},
+                "min": {"type": ["number", "null"]},
+                "max": {"type": ["number", "null"]},
+                "quantiles": {
+                    "type": ["array", "null"],
+                    "items": {
+                        "type": ["array", "null"],
+                        "items": [
+                            {"type": "number", "minimum": 0, "maximum": 1},
+                            {"type": "number"}
```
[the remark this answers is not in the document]
graphaelli: This requires quantiles to be a list of tuples, where each tuple contains:
  first element: number in the range 0-1
  second element: any number

I may have gone a bit overboard with the null allowances, reviewing those now.

utility/map_str_enhancer.go
graphaelli: This is used in a different context than `Add` normally is - here it's against `ManualDecoder.MapStr` result

PR #1022, opened 2018-06-25
Add release notes

roncohen:
Adds the CHANGELOG.asciidoc (that we're already maintaining) to the docs they will get rendered nicely and so we can link to them from download pages.

image

image

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -1,7 +1,7 @@
-
 [[release-notes-head]]
 == APM Server version HEAD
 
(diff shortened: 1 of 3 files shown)


watson:
@roncohen is the idea that this doc is kept up to date manually or is there an automated process?

roncohen:
@watson we're manually keeping the APM Server changelog up to date. This PR just adds that file to the docs so they will get rendered nicely and is more linkable.

alvarolobato:
Thanks Ron, having this in the docs is great.

PR #1024, opened 2018-06-25
Move generated schemas into sub-folders

roncohen:
This moves generated schemas into their own sub-folders in their respective processors.

This is extracted from the work on [intake v2](https://github.com/elastic/apm-server/pull/930). 

Right now, I think it's nice to have auto-generated and human-generated source code separately. 

In the future for intake v2, we'll need to validate entities like errors and transactions separate from the payload, which means we'll need to have the schemas separate for the individual entities.

When we add the individual schemas (errors, transactions) in the future, we'd be putting even more auto-generated files in the same directories as the human-generated code, if we didn't have a sub-folder for it. 

I went ahead and gave the schema constants names (`const PayloadSchema`) instead of using the `Schema()` method, because we'll need to have more schemas in the package in the future.

### DIFF
--- a/processor/error/package_tests/json_schema_test.go
+++ b/processor/error/package_tests/json_schema_test.go
@@ -4,6 +4,7 @@ import (
 	"testing"
 
 	er "github.com/elastic/apm-server/processor/error"
+	"github.com/elastic/apm-server/processor/error/generated/schema"
 	"github.com/elastic/apm-server/tests"
 )
 
@@ -148,5 +149,5 @@ func TestPayloadAttributesInSchema(t *testing.T) {
 		"errors.context.request.cookies.c2",
 		"errors.context.tags.organization_uuid",
 	)
-	tests.TestPayloadAttributesInSchema(t, "error", undocumented, er.Schema())
+	tests.TestPayloadAttributesInSchema(t, "error", undocumented, schema.PayloadSchema)
 }
--- a/script/inline_schemas/inline_schemas.go
+++ b/script/inline_schemas/inline_schemas.go
@@ -3,6 +3,8 @@ package main
 import (
 	"fmt"
 	"io/ioutil"
+	"os"
+	"path"
 	"path/filepath"
 	"regexp"
 	"strings"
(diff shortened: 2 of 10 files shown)


processor/error/package_tests/json_schema_test.go
```diff
@@ -4,6 +4,7 @@ import (
 	"testing"
 
 	er "github.com/elastic/apm-server/processor/error"
+	generatedschemas "github.com/elastic/apm-server/processor/error/generated-schemas"
```
simitt: Pleas use `schema` over `schemas` - as [Rakyll's post](https://rakyll.org/style-packages/) points out package names should not be plural but singular. 

I am not a huge fan of having `generated-schemas` in the path. The package name is `schemas`, the import path ends with `generated-schemas` and it needs to be overwritten with `generatedschemas` - if not absolutely necessary I'd rather avoid that.
How about either 
* getting rid of `generated` and just use `schema` 
* or use `generated/schema` folder structure to have a cleaner import. This would also allow to add further generated files in the future if necessary.

script/inline_schemas/inline_schemas.go
```diff
@@ -12,11 +14,11 @@ const basePath = "./docs/spec/"
...
 	schemaPaths := []struct {
-		path, schemaOut, goVariable, packageName string
+		path, schemaOut, goVariable string
 	}{
-		{"errors/payload.json", "processor/error/schema.go", "errorSchema", "error"},
-		{"transactions/payload.json", "processor/transaction/schema.go", "transactionSchema", "transaction"},
-		{"sourcemaps/payload.json", "processor/sourcemap/schema.go", "sourcemapSchema", "sourcemap"},
+		{"errors/payload.json", "processor/error/generated-schemas/payload.go", "PayloadSchema"},
+		{"transactions/payload.json", "processor/transaction/generated-schemas/payload.go", "PayloadSchema"},
+		{"sourcemaps/payload.json", "processor/sourcemap/generated-schemas/payload.go", "SourcemapSchema"},
```
simitt: The name `SourcemapSchema` is not consistent - please also change to `PayloadSchema`.

roncohen:
thanks @simitt! Updated and rebased.

simitt:
jenkins, retest this please

roncohen:
jenkins, retest this please

PR #1028, opened 2018-06-25
[6.x] Update beats framework to 575c675

graphaelli:


graphaelli:
Wanted to add `make is-beats-updated` to run post-`make update-beats` but for some reason a few packages that are not in `6.x` are added to vendor.json, making that target fail as currently written:

```
github.com/elastic/beats/libbeat/asset
github.com/elastic/beats/libbeat/autodiscover/providers/jolokia
github.com/elastic/beats/libbeat/common/seccomp
github.com/elastic/beats/libbeat/common/transport/tlscommon
github.com/elastic/beats/libbeat/processors/dissect
```

Same thing happened in #1027 on 6.3.

PR #1030, opened 2018-06-26
Restructure package tests

simitt:
Preparation to support different schemas and unify usage of json schema
and fields tests.


processor/error/package_tests/attrs_common.go
```diff
@@ -28,20 +59,29 @@ func TestAttributesPresenceInError(t *testing.T) {
 		"errors.log.stacktrace.lineno",
 		"errors.context.request.method",
 		"errors.context.request.url",
-	)
-	condRequiredKeys := map[string]tests.Condition{
+	))
+}
+
+func condRequiredKeys(c map[string]tests.Condition) map[string]tests.Condition {
```
jalvz: why the `c` argument if you ever only call it with `nil`?

simitt: Because I pulled this out from the distributed_tracing branch. You're right that it doesn't make sense yet, but as soon as we are allowing multiple formats to be sent to the server we will need this. So it didn't make sense to me to add methods where the signature changes with the next PR.

processor/transaction/package_tests/attrs_common.go
```diff
@@ -1,24 +1,55 @@
...
+		tests.Group("transactions.context.request.env."),
+		tests.Group("transactions.context.request.body"),
+		tests.Group("transactions.context.request.cookies"),
+		tests.Group("transactions.context.custom"),
+		tests.Group("transactions.context.tags"),
+		tests.Group("transactions.marks"),
+	))
+}
+
+func jsonSchemaNotInPayloadAttrs(s *tests.Set) *tests.Set {
```
jalvz: same argument for this? (required for DT)

Issue #968, opened 2018-05-25
Take POC for distributed tracing to production ready. 


simitt:
* Add list of mappings from Intake API to ES template.
* Verify that Intake API validations are working properly.
* Add tests.

simitt:
# Adding Benchmarks
Currently the distributed tracing support is added to the V1 endpoint. This is done in a backwards compatible way by having a json schema that supports single service and distributed tracing. The json schema is using `allOf,anyOf,oneOf` and `if,else,then` keywords to achieve this.

Benchmarking `master` against supporting both concepts in one schema, shows worse results when supporting both schemas.

## Errors
Benchmarking the `validation` and `decoding` steps for Errors shows significantly worse results for the `validation` step when supporting both schemas. 

`go test -benchmem -run="none" -benchtime=200ms -bench='.*Processor$' ./processor/error/...`

### Baseline (master)
```
BenchmarkBackendProcessor/TestProcessErrorMinimalServiceValidate-4         	   50000	      6977 ns/op	     225 B/op	       8 allocs/op
BenchmarkBackendProcessor/TestProcessErrorMinimalServiceDecode-4           	  100000	      3238 ns/op	     504 B/op	      10 allocs/op
BenchmarkBackendProcessor/TestProcessErrorMinimalProcessValidate-4         	   30000	      9093 ns/op	     401 B/op	      14 allocs/op
BenchmarkBackendProcessor/TestProcessErrorMinimalProcessDecode-4           	  100000	      3766 ns/op	     568 B/op	      12 allocs/op
BenchmarkBackendProcessor/TestProcessErrorFullValidate-4                   	    5000	     58480 ns/op	   13043 B/op	     161 allocs/op
BenchmarkBackendProcessor/TestProcessErrorFullDecode-4                     	   20000	     14368 ns/op	    3632 B/op	      93 allocs/op
BenchmarkBackendProcessor/TestProcessErrorNullValuesValidate-4             	   10000	     39119 ns/op	   11097 B/op	      93 allocs/op
BenchmarkBackendProcessor/TestProcessErrorNullValuesDecode-4               	   30000	      7862 ns/op	    1184 B/op	      22 allocs/op
BenchmarkBackendProcessor/TestProcessErrorAugmentedIPValidate-4            	   30000	      7920 ns/op	     225 B/op	       8 allocs/op
[...]

Issue #913, opened 2018-05-07
Distributed Tracing POC - Intake API JSON schema changes for distributed tracing


felixbarny:
The first step to get rolling with supporting distributed tracing is to change the JSON schema for the intake API. If that is done, the agents can work independently on supporting this new API, without having to wait until these features are actually implemented in the intake API.

## Changes to the Intake API
We should strive for compatibility with the W3C trace context spec https://github.com/w3c/distributed-tracing. Trace IDs are globally unique as they consist of 128 random bits, just like an UUID. Transaction IDs and Span IDs are unique to their corresponding trace.
* Transaction
  * Add traceId (hex encoded 128 random bits)
  * Change id to hex encoded 64 random bits
  * Add parentId (hex encoded 64 random bits)
    * Can refer to either a Transaction.id or Span.id
    * Remove embedded spans array
* Span
  * Add traceId (hex encoded 128 random bits)
  * Change id to hex encoded 64 random bits
    * A monotonically incrementing counter like in the current model for the Span ID would not work in a distributed environment.
  * Change semantics of parent: can be either a Transaction.id or a Span.id of the span or transaction that caused this span
* Payload
  * Add spans array

## Backwards compatibility
These changes can be applied in a backward compatible way, no need for a v2 for this. That way, we are able to release distributed tracing features before 7.0 and before the details of intake v2 have been sorted out. For example, the Transaction.spans array can be deprecated. UUID validation on the Transaction ID should be removed, otherwise, it’s just adding fields. The type Transaction.id and Span.id can be set to String or 64-bit Integer.

simitt:
First POC available: https://github.com/elastic/apm-server/compare/6.x...simitt:POC-6-x-dt

**ES template changes:**
__transaction__:
*  added `hex_id` in favour of `id` (deprecated)
*  added `trace_id`
* `id` has been a mandatory field in the past, now either `hex_id` or `id` are required.  

__spans__:
*  added `hex_id` in favour of `id` (deprecated)
*  added `parent_id` in favour of `parent` (deprecated).
*  added `trace_id`


__error__:
*  added `hex_id` in favour of `id` (deprecated)
*  added `parent_id` in favour of `transaction.id` (deprecated).
*  added `trace_id`
 
**Changes to the Intake API**:  have been made according to the spec of @felixbarny . 

**Missing from POC**:
Adding validation rules to the json spec for
* required `trace_id` if `id` is in hex encoded format, for error, transaction and span
[...]

PR #1031, opened 2018-06-26
Support distributed tracing in a backwards compatible way

simitt:
based on top of https://github.com/elastic/apm-server/pull/1030

implements https://github.com/elastic/apm-server/issues/968, based on POC https://github.com/elastic/apm-server/issues/913

The json schema decides if an error or transaction is in a `distributed tracing` or a `single service tracing` (old) format based on the `transaction.id` and `error.id`. If the `id` is missing it is always considered a single service tracing format. This means that for distributed tracing an `id` is always required.
The json schema accepts distributed tracing attributes even when a single service tracing format is recognized, but the additional attributes will not be processed or stored. 

Please review especially against 
* Intake API logic, especially which fields are required and optional
* backwards compatibility changes, especially the conversion of the `span.id` form a hex encoded string to a long

cc @felixbarny 

[Updated after internal discussions]:
Short summary regarding **required attributes**:
* Errors : if a `trace_id`, `transaction_id` or `parent_id` is given, then all three of them are required. `id` is optional, and stays in the UUID format used so far. 
* Transactions : `id` and `trace_id` are required; `parent_id` is optional.
* Spans : `id`, `trace_id`, `transaction_id` and `parent_id` are required.


simitt:
Closing this for now as some more discussions are needed, see Issue #968

Issue #1019, opened 2018-06-21
document service name change


graphaelli:
Inspired by https://discuss.elastic.co/t/how-can-i-remove-a-service/136904/2

Users commonly change service names and have to deal with old data.  While often the old data is simply aged out, it would be nice to provide instructions for how to forcibly remove or update service names.

Delete:
```
POST /apm-*/_delete_by_query
{
  "query": {
    "bool": {
      "must": [
        {
          "term": {
            "context.service.name": {
              "value": "old-service-name"
            }
          }
        }
      ]
    }
  }
}
```

Update:

```
POST /apm-*/_update_by_query
{
  "query": {
    "bool": {
      "must": [
        {
          "term": {
            "context.service.name": {
              "value": "old-service-name"
            }
          }
        }
      ]
    }
  },
  "script": {
    "source": "ctx._source['context.service.name'] = 'new-service-name'",
    "lang": "painless"
  }
}
```

Probably best to use the long form of these queries as shown to make it easy for additional requirements to be added.

PR #1036, opened 2018-06-27
Refactor docs and add examples on how to delete and update data.

simitt:
Add description on how to delete and update data by query.
Implements elastic/apm-server#1019

Added an overview to the top and ordered the sections accordingly.

screen shot 2018-06-27 at 16 09 04

screen shot 2018-06-27 at 16 08 40


docs/faq.asciidoc
```diff
@@ -143,22 +168,59 @@ INFO      "delete_indices" action completed.
...
+            "context.service.name": {
+              "value": "old-service-name"
+            }
+          }
+        }
+      ]
+    }
+  }
+}
+------------------------------------------------------------
```
dedemorton: You should add //CONSOLE to the next line after the dashes so that the published topic has Copy as Curl and View In Console links:

```
}
---------------------------------------------------------------
// CONSOLE
```

The doc build will add the links to the bottom of the example. Here's an example from the Beats docs. The links make it really easy for users to run the examples.

image

dedemorton: I didn't check for other examples in the doc, but we should do this whenever possible.

docs/faq.asciidoc
```diff
@@ -143,22 +168,59 @@ INFO      "delete_indices" action completed.
...
-When the APM Server cannot process incoming requests quickly enough,
-you will see "503: Request timed out waiting to be processed" messages.
 
-The best way to avoid this problem is to add more processing power to your APM Server cluster.
-You can either migrate your APM Server processes to more powerful machines or add more machines.
+[float]
+[[update-existing-data]]
+=== How to update existing data?
+In case you want to update specific attributes in existing documents, e.g. update `context.service.name` to a new value, 
+you can do this by sending following request against your Elasticsearch API:
```
dedemorton: same comments as above

docs/faq.asciidoc
```diff
@@ -143,22 +168,59 @@ INFO      "delete_indices" action completed.
...
+      ]
+    }
+  },
+  "script": {
+    "source": "ctx._source['context.service.name'] = 'new-service-name'",
+    "lang": "painless"
+  }
+}
+------------------------------------------------------------
+See https://www.elastic.co/guide/en/elasticsearch/reference/6.0//docs-update-by-query.html for further information on
```
dedemorton: Use an asciidoc attribute, as recommended earlier. Not sure why you would hard code to 6.0 branch, but there's a way to do that if you need to. It's generally better to point to branches at the same release level unless there's a compelling reason to specify a version.

watson: Even if we need to point to a specific branch, I would still recommend that we use the attributes, e.g. `See {ref-60}/docs-update-by-query.html[update by query] for further...`

https://github.com/elastic/docs/blob/3eac44b52d266ec5b7ad8d9d1cc583558cce2e5c/shared/attributes.asciidoc#L2

simitt:
Thank you @dedemorton for reviewing, I added the changes.

docs/faq.asciidoc
```diff
@@ -75,48 +102,51 @@ A good rule of thumb is that `queue.mem.events` should equal `output.elasticsear
...
 
 By default APM indices have the pattern `apm-%{[beat.version]}-{type}-%{+yyyy.MM.dd}`.
 With the curator command line interface you can, for instance, see all your existing indices:
 
-["source","sh",subs="attributes"]
+["source","sh"]
 ------------------------------------------------------------
 curator_cli --host localhost show_indices --filter_list '[\{"filtertype":"pattern","kind":"prefix","value":"apm-"\}]'
+------------------------------------------------------------
+// CONSOLE
```
dedemorton: Unfortunately // CONSOLE only works on REST calls, so you should remove it here. It would be *really cool* if we had similar functionality to open commands in a terminal window, but we don't have that right now.

docs/faq.asciidoc
```diff
@@ -131,34 +161,77 @@ apm-{stack-version}-transaction-{sample_date_2}
 
 And then delete any span indices older than 1 day:
 
-["source","sh",subs="attributes"]
+["source","sh"]
 ------------------------------------------------------------
 curator_cli --host localhost delete_indices --filter_list '[\{"filtertype":"pattern","kind":"prefix","value":"apm-{stack-version}-span-"\}, \{"filtertype":"age","source":"name","timestring":"%Y.%m.%d","unit":"days","unit_count":1,"direction":"older"\}]'
+------------------------------------------------------------
+// CONSOLE
```
dedemorton: Same comment. Does not currently work for non-REST examples.

roncohen:
@simitt great to get delete and update data by query documented! 

I don't think it works so well to simply convert the problems/solutions that we had, to an FAQ. For example, we now have "How to add APM Server nodes?" but the section doesn't explain _how_ to do that at all.

I also don't think "How to reduce the payload size?" will really ever be a question we get asked frequently. Instead, people will ask about the symptom they are seeing: "Why am I getting `503 Request timed out waiting to be processed?"

The FAQ is under "Troubleshooting" but I think that the two added sections (delete and update by query) don't have to do with "Troubleshooting" necessarily. I think of them more like recipes. E.g. how do i set up curator to purge data periodically?

I think a solution could be to have an FAQ at the top level (not inside "Troubleshooting") that looks something like (quick draft):

* Why am I getting "500: Queue is full" ?
* Why am I getting "500: Request timed out waiting to be processed" ?
* Can I adjust the amount of data collected?
* How do i purge old data periodically?
* How can i delete data for a specific service?
* Can i rename a service?

dedemorton:
I really like Ron's suggestions here. TBH, I had a similar reaction, but only had time to do a superficial review. :-/

simitt:
As mentioned in an offline discussion the idea was to align with the structure of [beats FAQ section](https://www.elastic.co/guide/en/beats/metricbeat/current/faq.html). 
APM Server is built on top of beats. We have a lot of shared documentation and usually try to align to make it easier for our users. 

I am fine though with switching to a dedicated top level FAQ section. @roncohen I don't really understand your suggestion to also pull something like `Why am I getting 500 Queue full` into the FAQ. Should this link to the trouble section then?

roncohen:
I was thinking was that the current "common problems" section under "troubleshooting" goes away and we introduce a top level FAQ that contains all that information instead.

simitt:
@roncohen I reworked the PR and have two different proposals now. Please have a look, you can also check out the linked commits and run `make docs` if you want to build it locally.
I personally prefer solution (2). 

(1) [Top level FAQ, move everything over to FAQ](https://github.com/elastic/apm-server/pull/1036/commits/3a54f9662bbd601a832005f6a44cbcc8b09ee387)
screen shot 2018-07-03 at 14 25 10

screen shot 2018-07-03 at 14 25 17

(2) [Top level Recipes, leave common problems in trouble shooting section](https://github.com/elastic/apm-server/pull/1036/commits/6fe11a2227fe23a158d2ff46170884e4c58350de)
screen shot 2018-07-03 at 14 59 01

screen shot 2018-07-03 at 14 59 16

screen shot 2018-07-03 at 14 59 43

dedemorton:
I don't mean to complicate the discussion here, but I worry about a top-level FAQ becoming a catch-all for tasks that really belong in the docs as separate topics. Dropping content into an FAQ is easier than figuring out where it really belongs (and it gives the person contributing docs an "out" for providing content that's incomplete).

There's also an impact on SEO. Keywords within the topic title have a bigger impact on search engine rankings than keywords within the body. Stuff buried in an FAQ might not be found so easily by google.

I actually like the idea of keeping common problems under the troubleshooting section. I'd suggest that we have a separate topic for each problem (for SEO). For each topic/problem, we should  provide the solution directly beneath the description of the problem, if possible.  (I'd like to do something similar for Beats because I feel that the FAQ  format no longer works because the list of issues has gotten a lot longer.)

For solutions that apply to multiple problems, I recommend creating a topic that groups related solutions based on a specific user goal. For example, you could have a topic called "Performance tuning" that groups common things users can do to tune performance. Then you could point to the performance tuning topic (or its sub-sections) from within your troubleshooting topics.

So, in the TOC, you'd have something like:

```
Exported fields
Performance tuning
Troubleshooting
  Common problems
    Problem A
    Problem B
    Etc...
  Get help
  Debug
```
You could put all the performance tuning content in a single topic, or break it out into sub-topics. Does that make sense?

I don't like the idea of having a "Recipes" section for a couple of reasons:

* I would expect it to contain detailed steps for how to build something (rather than information about performing tasks).
* It's not a meaningful container--doesn't convey specific user goals. We already have this problem with our "configuring" section, which is a catch-all for reference and task-based content.

simitt:
Thank you @dedemorton for taking the time to elaborate on this. I try to come up with a structure that takes your arguments into account as a POC and once we agreed on the structure will change the docs.

simitt:
Thanks for the detailled explanation @dedemorton.
I managed to compile the changes and pushed them to a [WIP branch](https://github.com/simitt/apm-server/tree/restructure-documentation/docs).

Here some screenshot of the structure, that does not have a dedicated FAQ section:
screen shot 2018-07-05 at 09 02 35

screen shot 2018-07-05 at 09 02 46

simitt:
Closing for now and reopen when all the agreed changes and restructures are in.

simitt:
I took the suggestions from previous discussions into account and updated the docs.

Cleaning up and refactoring the `exported fields` and `event types` section is not part of this PR, as this can easily be separated from those changes. 

Some excerpts of the newly suggested structure:
screen shot 2018-07-05 at 16 08 06


## Tune Data Ingestion
screen shot 2018-07-05 at 16 08 15

screen shot 2018-07-05 at 16 08 32

## Storage Management
screen shot 2018-07-05 at 16 11 42

screen shot 2018-07-05 at 16 08 58

## Common Problems
screen shot 2018-07-05 at 16 14 08

docs/data_ingestion.asciidoc
```diff
@@ -0,0 +1,103 @@
...
+you can tweak APM Server options to make better use of the cluster:
+
+* Increase `output.elasticsearch.workers` to a number that suits your setup.
+See {ref}/tune-for-indexing-speed.html[tune for indexing speed] for an overview.
+* Ensure `output.elasticsearch.bulk_max_size` is set to a high value, for example 5120.
+  The default of 50 is very conservative.
+* Ensure that `queue.mem.events` is set to a reasonable value compared to your other settings.
+A good rule of thumb is that `queue.mem.events` should equal `output.elasticsearch.worker` multiplied by `output.elasticsearch.bulk_max_size`.
+
+Get detailled information on available <<configuring-output,output configuration options>> and their default values.
```
jalvz: detailed with one `l`

docs/data_ingestion.asciidoc
```diff
@@ -0,0 +1,103 @@
...
+APM Server uses an internal queue to allow buffering incoming requests until they can be delievered to Elasticsearch. 
+A larger internal queue allows Elasticsearch to be unavailable for longer periods,
+and it alleviates problems that might result from sudden spikes of data.
+You can increase the queue size by overriding `queue.mem.events`.
+Be aware that increasing `queue.mem.events` can significantly affect APM Server memory usage.
+
+[[increase-concurrent-requests]]
+[float]
+=== Increase concurrent requests
+To avoid overfloating the APM Server has an upper limit to how many requests are accepted concurrently.
```
jalvz: A native speaker would have a better recommendation, but "overfloating" doesn't sound right to me. Maybe overwhelming, saturating, overflowing. I also think you need a pronoun somewhere:
"to avoid... the APM Server, it has an..."

docs/data_ingestion.asciidoc
```diff
@@ -0,0 +1,103 @@
...
+
+[[reduce-payload-size]]
+[float]
+=== Reduce the payload size
+
+Large payloads coming from agents may result in a server timeout.
+You can reduce the payload size by decreasing the `max_queue_size` in the agents.
+This will result in agents sending smaller payloads to the APM Server,
+but the requests will be more frequent.
+See the documentation for the {apm-py-ref}/configuration.html#config-max-queue-size[Python] and {apm-node-ref}/agent-api.html#max-queue-size[Node.js] agents for more information.
```
jalvz: 💯 linking here to agent docs! 
should we mention keep-alive? (ie. supported by deserved, disabled by default in most agents)

docs/data_ingestion.asciidoc
```diff
@@ -0,0 +1,103 @@
...
+Optionally you can also <<reduce-sample-rate, reduce the sample rate>> or <<reduce-stacktrace, reduce the amount of stacktraces 
+collected>>,
+both leading to collect less data and decreasing the payload size. 
+
+Read more about configuration options directly in the {apm-agents-ref}/index.html[agent documentation].
+
+[[tune-es]]
+== Tune Elasticsearch
+
+Increasing the Elasticsearch ingestion rate is a large topic.
```
jalvz: This way of introducing information to users might be a bit discouraging, i'd remove this sentence or change it to something like "is covered in detail at..."
Same for "Storage management is a large topic" bellow

docs/storage_management.asciidoc
```diff
@@ -0,0 +1,158 @@
...
+--
+
+[[reduce-storage]]
+== Reduce storage
+The amount of storage for APM data depends on several factors. 
+Depending on how many services you are instrumenting and how much traffic the services see, 
+the number of recorded transactions varies.
+You can influence the detail level of the information collected 
+by reducing the sample rate or reducing collected stacktrace information.
+Another factor is how long you want to keep monitoring data around.
```
jalvz: "...is for how long..."? (I actually don't know if this phrasing is correct)

docs/storage_management.asciidoc
```diff
@@ -0,0 +1,158 @@
...
+Depending on your use case, 
+you can either delete data periodically with a tool like {curator-ref-current}[Curator] 
+or by using the {ref}/docs-delete-by-query.html[Delete By Query API].
+
+
+[[delete-data-periodically]]
+[float]
+==== Delete data periodically
+
+It might make sense to delete old APM indices on a periodical basis to make room for new data.
```
jalvz: "periodical" -> "periodic" (I think!)

docs/data_ingestion.asciidoc
jalvz: probably this comment is out of place, thou -> can be added in another pr

docs/data_ingestion.asciidoc
```diff
@@ -0,0 +1,103 @@
...
+* <<tune-es>>
+
+--
+
+[[tune-apm-server]]
+== Tune APM Server
+You can fine tune your setup considering following options:
+
+* <<tune-output-config>>
+* <<increase-queue-size>>
```
jalvz: I know that part of this was there already; but for both `increase-queue-size` and `increase-concurrent-requests` i'd rename titles to "adjust-*" since you can either increase or decrease

docs/data_ingestion.asciidoc
simitt: https://github.com/elastic/apm-server/issues/1089

simitt:
@jalvz thanks for the comments, I pushed a commit following your suggestions.

I agree that the wording is often verbose and repetitive, but also think that this probably can be better addressed by a native speaker going over it separately.

dedemorton:
Created an issue to track editing the content. This would be a good first task for the new writer: https://github.com/elastic/apm-server/issues/1090

PR #1045, opened 2018-06-28
catch errors when starting the beater

jalvz:
... and return immediately instead of timing out

(in support for https://github.com/elastic/apm-server/pull/1010, ie: passing a wrong passphrase will cause the beater to fail, and test should fail right away with a descriptive reason)

### DIFF
--- a/beater/server_test.go
+++ b/beater/server_test.go
@@ -18,6 +18,8 @@ import (
 	"github.com/kabukky/httpscerts"
 	"github.com/stretchr/testify/assert"
 
+	"github.com/stretchr/testify/require"
+
 	"github.com/elastic/apm-server/tests/loader"
 	"github.com/elastic/beats/libbeat/beat"
 	"github.com/elastic/beats/libbeat/common"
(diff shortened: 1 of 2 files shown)


beater/server_test.go
```diff
@@ -45,7 +45,8 @@ func TestMain(m *testing.M) {
 }
 
 func TestServerOk(t *testing.T) {
-	apm, teardown := setupServer(t, nil)
+	apm, teardown, err := setupServer(t, nil)
+	assert.NoError(t, err)
```
simitt: We recently introduced using [testify/require](https://godoc.org/github.com/stretchr/testify/require) for checks that should fail fast and prevent the test from being run any further. I think this might also make sense for some of the `NoError` checks here. 

(see conversation around introducing _require_ https://github.com/elastic/apm-server/pull/1015#discussion_r197250826)

jalvz: done here 3be0983417f7dc4b5427cab78ec8f4f5da4a61f7

PR #1052, opened 2018-06-29
Vendor beats transptest

jalvz:
required for https://github.com/elastic/apm-server/pull/1010


jalvz:
it needs to run update beats for that, which i did in a separate PR because it also happens to incorporate all the packaging stuff (you'll find the transptest code there)

PR #1054, opened 2018-07-02
[6.x] update beats

simitt:
This PR updates to the latest `beats` version `6.x`, including following changes:

- adapting `make check` to fail when the _elastic license header_ is missing in any files, and adapts `make fmt` to add the missing license header in any file. Running those two commands added license headers to all files now. 
- `govendor` `github.com/elastic/beats/libbeat/generator/fields` for changed fields generation and handling. 
- generally updating beats libraries
- updating `github.com/elastic/go-struct` to `v0.0.4`


Makefile
```diff
@@ -39,18 +41,7 @@ before-build:
...
-# Generates imports for all modules and metricsets
-.PHONY: imports
-imports:
-	@mkdir -p include
-	@mkdir -p processor
-
-.PHONY: fields
-fields:
-	@cat _meta/fields.common.yml > _meta/fields.generated.yml
-	@cat processor/*/_meta/fields.yml >> _meta/fields.generated.yml
```
simitt: Changes are aligned with changes made when updated to beats framework in master, see https://github.com/elastic/apm-server/pull/987/files#diff-b67911656ef5d18c4ae36cb6741b7965.

simitt:
`make update` should not remove the license headers, as I added a change https://github.com/elastic/apm-server/pull/1054/files#diff-2cd3c44030e450f619ef38d007f1ba75R59 to run the licensing tool after creating the schema files.

simitt:
@graphaelli good findings! I've updated accordingly.

simitt:
That is very weird. Do you maybe have the `CHECK_HEADERS_DISABLED` env variable set? 
If not could you run `make check` once to ensure you have the licensing tool installed.

jalvz:
fyi im doing the 1.10.3 update on master as well also as part of updating beats

simitt:
@graphaelli you are absolutely right to expect a failure when the `go-licenser` is not installed. I added a panic if the licenser cannot run. This can only panic when the schemas are built, which is at build and not at runtime.

simitt:
update: I removed the `go-license` check from the inline-script again and just run the `add-headers` cmd within the `make update`. This way the logic does not get duplicated  and it seems cleaner to trigger the `go-licenser` from within the Makefile. Also the `go-licenser` would add the headers to all files and not only the ones touched by the inline-script,

CHANGELOG.asciidoc
```diff
@@ -14,6 +14,7 @@ https://github.com/elastic/apm-server/compare/9e0a1e281e56044745f1a4f18dcbf00a7f
 ==== Added
 
 - Listen on default port 8200 if unspecified {pull}[886]886.
+- Update Go to 1.10.3 {pull}894[894].
```
jalvz: wrong PR number?

Issue #540, opened 2018-01-24
Adapt dashboards for frontend data.


simitt:
Currently the OSS dashboards are tailored for data from existing (backend) agents. 
Investigate and adapt dashboards for specific frontend requirements.

simitt:
@jahtalab I assigned this to you as we recently discussed you taking overview the coordination of this.

jahtalab:
Dashboards/ [APM] Services:

![image](https://user-images.githubusercontent.com/6134122/35979737-62e00046-0ce9-11e8-9c01-f484979b4399.png)

The top charts are not relevant for frontend applications: 
General solution would be to show a "Top Services by Transaction per minute" chart
Also in the table "Avg. Resp. Time" is not relevant for frontend applications, should be "Avg. Trans Time". The same with "95th percentile of Resp. Time"

jahtalab:
Dashboard / [APM] Transactions

![image](https://user-images.githubusercontent.com/6134122/35980151-480bc6dc-0cea-11e8-9039-6a7e17b05dce.png)

The same issues as above ^

jahtalab:
@elastic/apm-server , I talked to @roncohen about the dashboards and he decided to adopt a generic solution for dashboards. Meaning we should try not to be specific about backend data. The main difference in the dashboards are in transaction data since the concept of Requests and Response is not relevant for frontend. 

@roncohen, Per our discussion about dashboards, Would you please coordinate how these fixes should be implemented?

simitt:
Is there something that should be done for 6.3? (ping @roncohen @jahtalab )

jahtalab:
To summarise the following changes are needed:

- In [APM] Services dashboard:
   - Chart title "Top Services by Response Time [APM]" should be changed to "Top Services by Transaction Time [APM]"
  - Chart title "Top Services by Request Per Minute [APM]" should be changed to "Top Services by Transaction Per Minute [APM]"
  - Table column header "Avg. Resp. Time" --> "Avg. Trans. Time"
  - Table column header "95th percentile of Resp. Time" --> "95th percentile of Trans. Time"
- In [APM] Transactions dashboard:
  - Chart title "Response Times [APM]" --> "Transaction Times [APM]"
  - The query for "Request Per Minute [APM]" chart should be changed in a way that it includes all transactions (currently it seems it only includes transactions with request field)
  - Chart title "Request Per Minute [APM]" --> "Transaction Per Minute [APM]"
[...]

PR #1055, opened 2018-07-02
Adapt dashboards for RUM agent.

simitt:
When releasing the RUM agent for GA we want the dashboards to be adapted accordingly.

This PR implements necessary changes suggested by @jahtalab in Issue elastic/apm-server#540 .


jahtalab:
image

@simitt , The tooltip seems to be autogenerated? has it always been like that?

simitt:
Just confirmed with the Kibana team that changing the tooltip is currently not supported.

makwarth:
👍

PR #1053, opened 2018-06-29
Update beats

jalvz:
This update includes packaging refactor (https://github.com/elastic/beats/pull/7388)

required before https://github.com/elastic/apm-server/pull/1010

PR #1075, opened 2018-07-04
[6.3] add jenkins release wrapper

graphaelli:
As of #1053, this path is used to trigger package tests on master.  This change wraps the 6.3 version of package tests at the new path so a single jenkins configuration can be used for both versions.

### DIFF
--- a/script/update_beats.sh
+++ b/script/update_beats.sh
@@ -23,6 +23,7 @@ git clone https://github.com/elastic/beats.git ${GIT_CLONE}
 # sync
 rsync -crpv --delete \
     --exclude=dev-tools/packer/readme.md.j2 \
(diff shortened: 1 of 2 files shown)


graphaelli:
The actual test suite affected: https://apm-ci.elastic.co/job/elastic+apm-server+6.3+package/24/console

script/update_beats.sh
```diff
@@ -23,6 +23,7 @@ git clone https://github.com/elastic/beats.git ${GIT_CLONE}
 # sync
 rsync -crpv --delete \
     --exclude=dev-tools/packer/readme.md.j2 \
+    --exclude=dev-tools/jenkins_release.sh \
```
jalvz: great that you remembered this!

PR #1076, opened 2018-07-04
[6.2] add jenkins release wrapper

graphaelli:
As of #1053, this path is used to trigger package tests on master. This change wraps the 6.2 version of package tests at the new path so a single jenkins configuration can be used for both versions.


graphaelli:
The actual test suite affected: https://apm-ci.elastic.co/job/elastic+apm-server+6.2+package/8/console

Issue #1083, opened 2018-07-05
Cleanup fields documentation


simitt:
The current section for [exported fields](https://www.elastic.co/guide/en/apm/server/current/exported-fields.html) has some flaws:
* repetition of _fields_, see [General APM fields fields](https://www.elastic.co/guide/en/apm/server/current/exported-fields-apm.html)
* unexpected structures, e.g. having context fields, but then also putting some of the context fields into other sections, https://www.elastic.co/guide/en/apm/server/current/exported-fields-apm.html
* including Kibana only fields

simitt:
After some discussions we keep the fields as is for the moment.

PR #1088, opened 2018-07-06
Restructure docs for RUM agent and ES related parts.

simitt:
* Remove the auto generated fields listing, and explain how users can
fetch the APM template instead.
* Add a separate RUM section and split up sourcemap docs according to
the existing structure. Use _RUM_ instead of _frontend_.
* Introduce a new part how to explore data in ES and move everything
related.
* This leads to only having a general overview of Event Types left on
top level, for easier high level understanding of APM Events.

Builds on top of https://github.com/elastic/docs/pull/395

implements #1083 

## High level overview:
screen shot 2018-07-06 at 15 28 27

## Changed Event Types
screen shot 2018-07-06 at 15 28 39

screen shot 2018-07-06 at 15 28 46

## RUM w. Source Mapping
screen shot 2018-07-06 at 15 28 58

screen shot 2018-07-06 at 15 29 05

## Exploring data in ES
screen shot 2018-07-06 at 15 29 24
screen shot 2018-07-06 at 15 29 35


docs/exploring-es-data.asciidoc
```diff
@@ -0,0 +1,63 @@
...
+* <<error-indices>>
+* <<sourcemap-indices>>
+* <<healthcheck-indices>>
+
+You can explore the raw data stored in Elasticsearch by querying the indices. 
+
+For getting an overview of existing indices you can run:
+["source","sh"]
+------------------------------------------------------------
+GET _cat/indices
```
graphaelli: Maybe filtering apm would be even more useful `GET _cat/indices/apm*`?

docs/exploring-es-data.asciidoc
```diff
@@ -0,0 +1,63 @@
...
+// CONSOLE
+
+
+If you want to see which _settings_ and _mappings_ are applied to the Elastic APM indices you can fetch the index 
+templates.
+First send a query to see which templates exist:
+
+["source","sh"]
+------------------------------------------------------------
+GET _cat/templates
```
graphaelli: Same as with indices `GET _cat/templates/apm*`?  Just a thought

docs/exploring-es-data.asciidoc
```diff
@@ -0,0 +1,63 @@
+[[exploring-es-data]]
+= Exploring data in Elasticsearch
+
+[partintro]
+--
+By default Elastic APM data are stored in different index types.
```
graphaelli: different index types -> different indices

docs/exploring-es-data.asciidoc
```diff
@@ -0,0 +1,63 @@
+[[exploring-es-data]]
+= Exploring data in Elasticsearch
+
+[partintro]
+--
+By default Elastic APM data are stored in different index types. 
+
+Indices storing monitored data are built in the format of `apm-[version]-[type]-[date]` by default.
```
graphaelli: monitored data sounds odd but I'm not sure what else to call the onboarding docs

docs/exploring-es-data.asciidoc
dedemorton: Should be "data is stored"

docs/exploring-es-data.asciidoc
dedemorton: Not sure you need to say "monitored data" here. Maybe just say something like:
```
The indices are named `apm-[version]-[type]-[date]` by default.
```

Then give an example. The way it's worded right now, sounds like the indices themselves use some kind of special format, but really it's jut the index name that uses the format.

docs/exploring-es-data.asciidoc
```diff
@@ -0,0 +1,62 @@
...
+= Exploring data in Elasticsearch
+
+[partintro]
+--
+By default Elastic APM data are stored in different indices, 
+in the format of `apm-[version]-[type]-[date]`.
+
+This format makes it easy to query for all Elastic APM data at once by using `apm-*`.
+
+It also allows for querying documents that have been collected with a specific APM Server version, e.g. `apm-6.3.0*`.
```
jalvz: i think you should either remove the `e.g. apm-6.3.0*` bit or use a variable pulling the current version

docs/errors.asciidoc
```diff
@@ -3,35 +3,11 @@
 
 An error record represents one error event, captured by Elastic APM agents within one service. 
 It is identified by a unique ID.
+Errors contain information about the `exception` that has been captured
+and some meta information about the setting in which the error occured. 
 
-Some of the key attributes of an error are described in more detail:
+Errors can either relate to monitored <<transactions, transactions>> 
+or be monitored stand alone, 
+depending on how your APM agents are set up.
```
jalvz: reading this, i guess someone will want to know how to set up an agent to relate errors and transactions...

docs/exploring-es-data.asciidoc
```diff
@@ -0,0 +1,62 @@
+[[exploring-es-data]]
+= Exploring data in Elasticsearch
+
+[partintro]
+--
+By default Elastic APM data are stored in different indices, 
+in the format of `apm-[version]-[type]-[date]`.
```
jalvz: I'll add a line here saying that type is one of error, transaction, etc.; or a link to the event types section

jalvz:
Forgot to mention: this could be split in 2 commits (or prs, it doesn't matter):
One for the things that make sense to backport (eg. removing exported fields), and other for the things that don't (RUM), wdyt?

simitt:
I suggest to backport to 6.x and 6.3. In 6.3 we already have experimental RUM support, so I would experimental to those section. 
WDYT @jalvz

simitt:
After rebasing onto master this is the index overview:

screen shot 2018-07-09 at 11 59 45

simitt:
@roncohen I added the `fields` section again as just discussed, please approve if you're fine with.

screen shot 2018-07-09 at 16 56 52

PR #1114, opened 2018-07-10
add source_mapping.elasticsearch configuration option

### LINKED ISSUE #1104: Apply sourcemaps even when output is not Elasticsearch
*updated*: one conservative option is to not allow upload when other output than Elasticsearch is used, since we won't be able to apply them, is better to inform the user with a error code when uploading.

graphaelli:
closes elastic/apm-server#1104

### DIFF
--- a/_meta/beat.yml
+++ b/_meta/beat.yml
@@ -72,7 +72,21 @@ apm-server:
     # to all error and transaction documents sent to the frontend endpoint.
     #source_mapping:
 
-      # Source maps are are fetched from Elasticsearch and then kept in an in-memory cache for a certain time.
+      # Source maps are always fetched from Elasticsearch, by default using the output.elasticsearch configuration.
+      # A different instance must be configured when using any other output.
+      # This setting only affects sourcemap reads - the output determines where sourcemaps are written.
+      #elasticsearch:
+        # Array of hosts to connect to.
+        # Scheme and port can be left out and will be set to the default (http and 9200)
+        # In case you specify and additional path, the scheme is required: http://localhost:9200/path
+        # IPv6 addresses should always be defined as: https://[2001:db8::1]:9200
+        # hosts: ["localhost:9200"]
+
+        # Optional protocol and basic auth credentials.
+        #protocol: "https"
+        #username: "elastic"
+        #password: "changeme"
+
       # The `cache.expiration` determines how long a source map should be cached before fetching it again from Elasticsearch.
       # Note that values configured without a time unit will be interpreted as seconds.
       #cache:
--- a/beater/beater.go
+++ b/beater/beater.go
@@ -48,6 +48,7 @@ type beater struct {
 
 // Creates beater
 func New(b *beat.Beat, ucfg *common.Config) (beat.Beater, error) {
+	logger := logp.NewLogger("beater")
 	beaterConfig := defaultConfig(b.Info.Version)
 	if err := ucfg.Unpack(beaterConfig); err != nil {
 		return nil, errors.Wrap(err, "Error processing configuration")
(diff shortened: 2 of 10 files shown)


beater/server_test.go
```diff
@@ -228,6 +228,77 @@ func TestServerNoContentType(t *testing.T) {
...
+
+		var beatConfig beat.BeatConfig
+		ocfg, err := common.NewConfigFrom(testCase.outputConfig)
+		if !assert.NoError(t, err) {
+			continue
+		}
+		beatConfig.Output.Unpack(ocfg)
+		apm, teardown, err := setupServer(t, ucfg, &beatConfig)
+		if assert.NoError(t, err) {
+			assert.Equal(t, testCase.expected, apm.smapElasticsearchHosts())
```
jalvz: I think I am not seeing the point of this test. You asserting that `smapElasticsearchHosts` returns the expected thing, buy `smapElasticsearchHosts` itself is a test helper function only used here?

beater/beater.go
```diff
@@ -59,15 +60,21 @@ func New(b *beat.Beat, ucfg *common.Config) (beat.Beater, error) {
...
 		}
-		if b.Config != nil && b.Config.Output.Name() == "elasticsearch" {
-			beaterConfig.setElasticsearch(b.Config.Output.Config())
+		if b.Config != nil && beaterConfig.Frontend.SourceMapping.EsConfig == nil {
+			// fall back to elasticsearch output configuration for sourcemap storage if possible
+			if b.Config.Output.Name() == "elasticsearch" {
+				logger.Info("Falling back to elasticsearch output for sourcemap storage")
+				beaterConfig.setSmapElasticsearch(b.Config.Output.Config())
+			} else {
+				logger.Info("Unable to determine sourcemap storage, sourcemaps will not be applied")
```
jalvz: I still think in this case we shouldn't allow users to upload sourcemaps...

graphaelli:
That's because:

> This setting only affects sourcemap reads - the output determines where sourcemaps are written.

@jalvz Do you think this is confusing because you are already familiar with apm-server?

jalvz:
but shouldn't be read from the same place were is written to? 
I am deff missing something.... how do I test manually that this works?

beater/server_test.go
graphaelli: The fallback from `source_mapping.elasticsearch` to `output.elasticsearch` only happens in `beater.New` - `smapElasticsearchHosts` is a test-only function that allows access to the final configuration of `Frontend.SourceMapping`.  These tests create a new beater to test that the fallback is happening (and not happening) correctly.

graphaelli:
One way to test would be to set the apm-server output to logstash and the sourcemap elasticsearch accordingly.  Concretely [1]:

```
output.elasticsearch.enabled: false
output.logstash.enabled: true
output.logstash.hosts: ["logstash:5044"]
apm-server.frontend.source_mapping.elasticsearch.hosts: ["elasticsearch:9200"]
```

Upload a sourcemap - that should flow through logstash into the same elasticsearch instance`apm-server.frontend.source_mapping.elasticsearch` is pointing to

Then post a transaction and see that the source map is fetched and applied.

[1] [apm-integration-testing](https://github.com/elastic/apm-integration-testing) will do the first three with `./scripts/compose.py --apm-server-output=logstash --with-logstash`

beater/beater.go
simitt: I agree that the sourcemap upload should fail, I'd even say not failing is a bug in the current implementation (and has nothing to do with this PR).
But I suggest to add `beaterConfig.Frontend.isEnabled() && smapper != nil` at the `killSwitchHandler` check at https://github.com/elastic/apm-server/blob/master/beater/handlers.go#L235

beater/server_test.go
jalvz: ok, thanks. a comment would help

beater/beater.go
graphaelli: @jalvz opened #1128 where we can fight to the death instead.

graphaelli:
@jalvz mind taking one more look with the updated config and documentation?

_meta/beat.yml
```diff
@@ -72,7 +72,21 @@ apm-server:
     # to all error and transaction documents sent to the frontend endpoint.
     #source_mapping:
 
-      # Source maps are are fetched from Elasticsearch and then kept in an in-memory cache for a certain time.
+      # Source maps are always fetched from Elasticsearch, by default using the output.elasticsearch configuration.
+      # A different instance must be configured when using any other output.
+      # This setting only affects sourcemap reads - the output determines where sourcemaps are written.
```
jalvz: I will phrase this more along the lines:
"if you use a logstash or kafka output, you should set this property to the elasticsearch host that logstash or kafka is writing to"...

docs/configuring-output-after.asciidoc
```diff
@@ -3,7 +3,9 @@
 [float]
 === Sourcemaps
 
-Sourcemaps can be uploaded through various outputs but can currently only be applied when Elasticsearch is the output.
+Sourcemaps can be uploaded through all outputs but must eventually be stored in Elasticsearch.
+When using outputs other than Elasticsearch, `source_mapping.elasticsearch` must be set for source maps to be applied.
+See <<config-sourcemapping-elasticsearch>> for more details.
```
jalvz: can you add a note that then you also need to change source_mapping.index_pattern to whatever matches the used output? (also in the yml file)

_meta/beat.yml
graphaelli: I'm worried that as we add outputs we'll forgot to make updates to this.

docs/configuring-output-after.asciidoc
graphaelli: good call, done in the docs.  The config file has good language around this.

Issue #1060, opened 2018-07-03
Document application metrics


jalvz:

jalvz:
This was about the `metricsets` events. They are already described in the apm server docs, but missing eg from https://www.elastic.co/guide/en/apm/get-started/master/apm-data-model.html.

PR #1121, opened 2018-07-11
make data model compatible with metricbeat

graphaelli:
In order to promote reuse of existing queries and visualizations and
possible future curated UIs


processor/metric/package_tests/TestProcessMetric.approved.json
```diff
@@ -86,15 +81,11 @@
                     "sum": 100,
                     "type": "summary"
```
axw: I think we want to drop `"type": "summary"` too?

graphaelli:
Added another commit to simplify the implementation now that we've scaled back our ambition.  Some unsurprising benchmark improvements as a result:

```
benchmark                                                               old ns/op     new ns/op     delta
BenchmarkProcessor/TestProcessMetricValidate-8                          85297         65487         -23.22%
BenchmarkProcessor/TestProcessMetricDecode-8                            9276          6829          -26.38%
BenchmarkProcessor/TestProcessMetricTransform-8                         14118         10276         -27.21%
BenchmarkProcessor/TestProcessMetricProcessRequest-8                    108431        78837         -27.29%

benchmark                                                               old allocs     new allocs     delta
BenchmarkProcessor/TestProcessMetricValidate-8                          428            284            -33.64%
BenchmarkProcessor/TestProcessMetricDecode-8                            41             31             -24.39%
BenchmarkProcessor/TestProcessMetricTransform-8                         98             59             -39.80%
BenchmarkProcessor/TestProcessMetricProcessRequest-8                    567            374            -34.04%

benchmark                                                               old bytes     new bytes     delta
BenchmarkProcessor/TestProcessMetricValidate-8                          39805         29281         -26.44%
BenchmarkProcessor/TestProcessMetricDecode-8                            1480          984           -33.51%
BenchmarkProcessor/TestProcessMetricTransform-8                         8178          6406          -21.67%
BenchmarkProcessor/TestProcessMetricProcessRequest-8                    49464         36676         -25.85%
```

processor/metric/payload.go
```diff
@@ -37,12 +37,14 @@ var (
...
-	transform(common.MapStr) error
+type sample struct {
+	name  string
+	value float64
+	unit  *string
 }
 
 type metric struct {
-	samples   []sample
+	samples   []*sample
```
simitt: out of curiosity - why the change?

processor/metric/system/_meta/fields.yml
```diff
@@ -0,0 +1,80 @@
+- key: system
```
simitt: How did you come up with this?
I checked it with the `metricbeat` example, but that has `metrics.samples.system` and `metrics.samples.memory` and not `system.memory`.

processor/metric/payload.go
graphaelli: changed sample from an interface to a struct since there is only one sample type now

processor/metric/system/_meta/fields.yml
graphaelli: This is from https://www.elastic.co/guide/en/beats/metricbeat/current/exported-fields-system.html#_memory_fields_5 - does that look right?

simitt: So some metrics have a defined type and format, and all the others are also indexed, but dependent on the first value, right?

graphaelli: exactly

graphaelli:
jenkins, retest this please

graphaelli:
jenkins, retest this please

model/metric/payload.go
```diff
@@ -184,25 +183,14 @@ func (md *metricDecoder) decodeSamples(input interface{}) []sample {
...
-		case "counter":
-			sample = md.decodeCounter(name, sampleMap)
-		case "gauge":
-			sample = md.decodeGauge(name, sampleMap)
-		case "summary":
-			sample = md.decodeSummary(name, sampleMap)
+		samples[i] = &sample{
+			name:  name,
+			value: md.Float64(sampleMap, "value"),
+			unit:  md.StringPtr(sampleMap, "unit"),
```
axw: unit is gone, delete this?

testdata/metric/metricbeat.json
```diff
@@ -0,0 +1,42 @@
...
+        "name": "apm-server",
+        "agent": {
+            "name": "elastic-go",
+            "version": "0.1.0"
+        }
+    },
+    "metrics": [{
+        "samples": {
+            "system.load.norm.1": {
+                "type": "gauge",
```
axw: delete all these "type"s?

graphaelli:
All feedback has been addressed, thanks for sticking with this as the requirements changed a few times since first opening.

I'm planning to follow up on this issue with documentation in #1060 about being careful with metric naming as 1. fields explosion could be a real issue here 2. since fields are on the root of these docs now (not under `metrics` like originally implemented), care needs to be take to avoid conflicts with apm document keys, especially when upgrading.

Issue #1008, opened 2018-06-14
document upgrade instructions


graphaelli:
The installation docs are good but we could use a section for upgrades.   Some thoughts:

* be sure to check for new options in apm-server.yml
* rerun setup to update index templates / patterns / dashboards

PR #1125, opened 2018-07-12
Docs: Add server upgrade instructions.

simitt:
implements elastic/apm-server#1008

related to https://github.com/elastic/stack-docs/pull/81

@dedemorton how can I add APM Server to https://www.elastic.co/products/upgrade_guide?

update:
screen shot 2018-07-17 at 16 36 49

### DIFF
--- a/docs/upgrading.asciidoc
+++ b/docs/upgrading.asciidoc
@@ -1,16 +1,16 @@
 [[upgrading]]
-=== Upgrading APM Server
+== Upgrading APM Server
 
 We do our best to keep APM Server backwards compatible between minor releases.
-However, check out the <<breaking-changes, breaking changes>> section before upgrading.
+However, check out the <<breaking-changes, breaking changes>> section for exceptions.
 
-Before upgrading we recommend to:
+Before upgrading:
 
-* Review changes between your currently used version and the one you want to upgrade to,
-by checking the <<release-notes,release notes>>.
+* Review the <<release-notes,release notes>> and the <<breaking-changes, breaking changes>> 
+for changes between your current version and the one you are upgrading to.
 * Check the {stack-ref}/index.html[Elastic Stack Installation and Upgrade Guide] for guidance on how to upgrade your 
- Elastic Stack used in combination with APM. 
- For a more tailored upgrade guide you can also try out the {upgrade_guide}[Interactive Upgrade Guide].
+ Elastic Stack. 
+ For a more tailored upgrade guide, try out the {upgrade_guide}[Interactive Upgrade Guide].
 
 When ready to actually upgrade your APM Server between minor versions, 
 simply install the new release.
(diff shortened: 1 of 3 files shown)


dedemorton:
@simitt 

@debadair is updating the guide right now (I think she already has it on staging). I'll ping her to let her know that you've asked, and you can follow up with her next week.

debadair:
@simitt I've opened a separate issue for adding APM server to the interactive guide: https://github.com/elastic/stack-docs/issues/83.

It's a bit fragile to update, so I want to tackle that separately from the fixes I currently have staged for release.

docs/upgrading.asciidoc
```diff
@@ -0,0 +1,49 @@
+[[upgrading]]
+=== Upgrading APM Server
```
dedemorton: Have you thought about putting the upgrade content at the top-level in the TOC (to be consistent with other docs in the stack)? It's fine to do it this way as long as you think users will see the content.

docs/upgrading.asciidoc
```diff
@@ -0,0 +1,49 @@
...
+When you have ensured everything is properly configured and updated,
+start the APM Server again.
+
+By default index names and the mapping templates are versioned. 
+When starting the new installation of APM Server new indices with the current version are created, 
+and the correct template for this version is applied automatically. 
+
+[[breaking-changes]]
+[float]
+==== Breaking Changes
```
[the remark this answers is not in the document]
simitt: So far this upgrade guide only considers minor upgrades, as we haven't had a major upgrade.
Within minor upgrades we aim for having no breaking changes, that's why I put the section here and not on a separate page.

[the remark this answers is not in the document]
simitt: I gave it a second thought and now put _Upgrading_ top level and _Breaking changes_ within the upgrade guide but as a separate page. WDYT?

docs/upgrading.asciidoc
```diff
@@ -0,0 +1,50 @@
...
+
+[float]
+==== 6.3
+no breaking changes in APM Server
+
+https://www.elastic.co/guide/en/beats/libbeat/current/breaking-changes-6.3.html[breaking changes in libbeat]
+
+[float]
+==== 6.2
+First stable version.
```
jalvz: what "stable" means? (GA?)

docs/upgrading.asciidoc
```diff
@@ -0,0 +1,50 @@
...
+When everything is properly configured and updated, start the APM server.
+
+When you start the APM server after upgrading, it creates new indices that use the current version,
+and applies the correct template automatically.
+
+[[breaking-changes]]
+=== Breaking Changes
+APM Server is built on top of {beats-ref}/index.html[libbeat].
+As such any breaking change in libbeat is also considered to be a breaking change in APM Server.
+
```
jalvz: This is OK, but separately from this we could discuss/create a policy around what is a breaking change and document it. I think would be useful for corner cases like rephrase/removing logs, etc.

PR #1127, opened 2018-07-12
disable metrics logging by default

jalvz:
Any reason not to do this?

### DIFF
--- a/_meta/beat.yml
+++ b/_meta/beat.yml
@@ -464,7 +464,7 @@ output.elasticsearch:
 # in the last period. For each metric that changed, the delta from the value at
 # the beginning of the period is logged. Also, the total values for
 # all non-zero internal metrics are logged on shutdown. The default is true.
-#logging.metrics.enabled: true
+logging.metrics.enabled: false
 
 # The period after which to log the internal metrics. The default is 30s.
 #logging.metrics.period: 30s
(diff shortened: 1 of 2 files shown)


roncohen:
I've seen that the beats team often make use of the metrics to help debug issues. It's a lot simpler to do if those are on by default.

jalvz:
Because for me it dominates & pollutes the logs to the point where it actually hurts readability. I never used them to debug any issues, not even when we had big performance/scalability problems.
(Another story might be the metrics pushed to ES that can be queried in dashboards, etc. but this is just about the logs)

Also: we are going to release application metrics because monitoring does not fit exactly the required use case, right? 
We don't log the actual application metrics  (just push them to ES); so why monitoring shouldn't be the same?

However, if any of you have insights/experience with effectively troubleshooting apm-server issues with monitoring logs, I would like to learn how you did it!

simitt:
I don't have any strong opinion on that tbh, so I am fine either way.

graphaelli:
I also don't have a strong opinion on doing this for users.  FWIW, I always have this when running locally.

jalvz:
Cool. So is this _approved_ or _rejected_ ?

roncohen:
I don't feel strongly either way 👍

Issue #993, opened 2018-06-06
setup automated benchmark suite


jalvz:
created out of https://github.com/elastic/apm-server/issues/773#issuecomment-394741852

PR #1147, opened 2018-07-17
Briefly explain benchmarking in README

roncohen:
It wasn't immediately clear how to benchmark my changes, so i added this which should make it clearer for other in the future.


axw:
FYI, I created https://github.com/elastic/gobench for storing the Go agent's benchmark data in ES, following Felix's lead. That tool could be used to store the server's benchmark data too. You won't get benchcmp output, but you can compare it visually in Kibana.

Makefile
```diff
@@ -93,3 +93,7 @@ release-manager-snapshot:
 .PHONY: release-manager-release
 release-manager-release:
 	./_beats/dev-tools/run_with_go_ver $(MAKE) release
+
+.PHONY: bench
+bench:
+	@go test -benchmem -run=XXX -benchtime=100ms -bench='.*Processor$$' ./processor/...
```
jalvz: I wouldn't have a `Processor` regex because if we ever create a benchmark somewhere else (or just rename those tests) then `make bench` would silently skip them.

roncohen:
@axw cool! would be nice to have them automatically run on merge to master (or similar) in CI and output to that cluster.

simitt:
@roncohen fyi we have a task to automate the benchmark tests #993 (linked the comment from @axw).

simitt:
@roncohen is there something blocking you to merge this?

PR #1167, opened 2018-07-23
Update beats framework to a314e34

roncohen:


roncohen:
ran into this now: `failed building apm-server-oss type=dmg for platform=darwin/amd64: failed to build internal beat pkg: copy failed: cannot stat source file apm-server.reference.yml: stat apm-server.reference.yml: no such file or directory` cc @andrewkroh

_beats/vendor/vendor.json
```diff
@@ -521,7 +511,7 @@
 		},
 		{
 			"path": "github.com/elastic/go-structform/internal/visitors",
-			"revision": "",
+			"revision": "v0.0.4",
```
simitt: This looks strange, also I could not find `github.com/elastic/go-structform/internal/visitors`, but `github.com/elastic/go-structform/visitors`. Same for `github.com/elastic/go-structform/internal/ubjson` above. 

That looks like a potential issue in beats though, https://github.com/elastic/beats/pull/7676

simitt:
oh, just saw your comment above now.

roncohen:
Added a commit to fix the problem mentioned above `apm-server.reference.yml: no such file or directory`.

PR #1173, opened 2018-07-23
[6.x] Update beats framework to fd9c363

roncohen:
Includes all the commits we needed to to on `master` to get the new packaging working.


jalvz:
https://github.com/roncohen/apm-server/blob/6.x-update/_beats/dev-tools/jenkins_release.sh build the package in the old way,
you need to copy over https://github.com/elastic/apm-server/blob/master/_beats/dev-tools/jenkins_release.sh from master

That is for the packaging test in ci: https://github.com/elastic/infra/pull/5798/files

other than that, lgtm

PR #1177, opened 2018-07-24
Update beats

jalvz:


jalvz:
closing in favour of https://github.com/elastic/apm-server/pull/1181

jalvz:
jenkins retest this please

magefile.go
```diff
@@ -99,6 +99,10 @@ func Update() error {
 	return sh.Run("make", "update")
 }
 
+func Fields() error {
+	return mage.GenerateFieldsYAML(".")
```
andrewkroh: I think this should be `model` rather than `.`. 

(I think the docs that I wrote in `GenerateFieldsYAML` are not 100% accurate w.r.t. how it searches for fields.yml files because I didn't fully understand the subtleties of how global_fields/main.go searches.)

jalvz: that won't skip `_meta/fields.common.yml`?

andrewkroh: `_meta/fields.common.yml` will still be included. The args to `GenerateFieldsYAML` are extra module directories to collect fields.yml.

simitt: We will do a follow up fix then.

PR #1194, opened 2018-07-24
[6.x] log payload when test fails (#1050)

simitt:
 - log payload when test fails  (#1050)


simitt:
jenkins, retest this please.

PR #1197, opened 2018-07-24
Server information

### LINKED ISSUE #957: Add json formatted information to `/` request and deprecate `healthcheck`
Add information about server version etc. in json format to the root endpoint, and deprecate the `healthcheck` endpoint.

graphaelli:
provide basic server information at /
* if no secret token is set
* if secret token is set and request is authorized

```
{
    "build_date": "2018-07-27T18:49:58Z",
    "build_sha": "bc4d9a286a65b4283c2462404add86a26be61dca",
    "version": "7.0.0-alpha1"
}
```

returns 200 OK in any case, to replace /healthcheck eventually

closes #957

### DIFF
--- a/beater/handlers.go
+++ b/beater/handlers.go
@@ -76,6 +76,7 @@ type serverResponse struct {
 	err     error
 	code    int
 	counter *monitoring.Int
+	body    string
 }
 
 var (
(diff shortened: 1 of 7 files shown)


beater/handlers.go
```diff
@@ -243,13 +271,41 @@ func sourcemapHandler(p processor.Processor, beaterConfig *Config, report report
...
+		"version":    version.GetDefaultVersion(),
+	}
+	detailedOkResponse := serverResponse{
+		code:    http.StatusOK,
+		counter: responseOk,
+		body:    serverInfo,
+	}
+
+	var handler http.Handler = http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.URL.Path != "/" {
```
jalvz: is this actually necessary?

docs/server-info.asciidoc
```diff
@@ -0,0 +1,39 @@
...
+The APM Server exposes an API endpoint to query some server information.
+
+[[server-info-endpoint]]
+[float]
+=== Server Information endpoint
+Send an `HTTP GET` request to the server information endpoint:
+
+[source,bash]
+------------------------------------------------------------
+http(s)://{hostname}:{port}/
```
jalvz: for the other types we have a full curl example:

`curl http://localhost:8200/   --header "Content-Type: application/json`

I think having this plus the one bellow is a bit too verbose tbh

beater/handlers.go
graphaelli: Yes, I forgot about how the [default mux](https://golang.org/pkg/net/http/#ServeMux.Handle) pattern matches work until writing https://github.com/elastic/apm-server/pull/1197/files#diff-24302485aab47c0697eb5cb545fbe394R120.

docs/server-info.asciidoc
```diff
@@ -0,0 +1,39 @@
+[[server-info]]
+== Server Information
+
+The APM Server exposes an API endpoint to query some server information.
```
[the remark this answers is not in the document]
graphaelli: ok, done

docs/server-info.asciidoc
graphaelli: I think the parity with the intake api docs is worth it, especially since this is such a short page anyway.

PR #1209, opened 2018-07-25
Add script for jenkins sync test.

simitt:
Gopath and environment needs to be set as the script actually runs `make update` which triggers `mage`..


simitt:
jenkins, retest this please

PR #1222, opened 2018-07-26
utility.Add should not modify source value

roncohen:
`utility.Add` adds a `val` to a MapStr `m`. However, it can sometimes modify the value given as a `val`. 

If you use `utility.Add` concurrently, to add the same value `val` to two different `m`s, you'll get can risk getting the following error:
```
fatal error: concurrent map iteration and map write
```


roncohen:
It turned out to be quite difficult to test because it is not enough to test that the values of the source do not change, because even writing the _same_ values to the "source" map could cause problems. The best I could come up with was to test that the value added is a copy of the source.

PR #1224, opened 2018-07-26
Remove unnecessary type assertions in utility.Add

roncohen:
Depends on https://github.com/elastic/apm-server/pull/1222.


graphaelli:
how/why does this depend on #1222?

roncohen:
was hoping to avoid conflicts by doing it like this, but that was probably a mistake.

PR #1231, opened 2018-07-31
don't build unused DMGs

graphaelli:
Up for discussion - build time is insignificant but the dependencies (xcode) are not.


graphaelli:
jenkins, retest this please

andrewkroh:
Since you forked your own package.yml you could directly implement these customizations in the YAML file now: https://github.com/elastic/apm-server/blob/master/magefile.go#L111

PR #1247, opened 2018-08-02
Reuse Beats ci scripts

jalvz:


script/jenkins/sync.sh
```diff
@@ -4,19 +4,9 @@ set -euox pipefail
 : "${HOME:?Need to set HOME to a non-empty value.}"
 : "${WORKSPACE:?Need to set WORKSPACE to a non-empty value.}"
```
simitt: can't you also remove this?

script/update_beats.sh
```diff
@@ -23,6 +23,9 @@ git clone https://github.com/elastic/beats.git ${GIT_CLONE}
 # sync
 rsync -crpv --delete \
     --exclude=dev-tools/packer/readme.md.j2 \
+    --exclude="dev-tools/jenkins_ci.sh" \
+    --exclude="dev-tools/jenkins_ci.ps1" \
+    --exclude="dev-tools/jenkins_intake.sh" \
```
simitt: sorry I didn't comment on this, why is this excluded?

jalvz: because we are not executing them, and it was distracting me to have unused scripts here.

simitt: Since this PR is going into the direction of aligning more with beats for the tests script, I think we should also try to use those scripts, instead of removing them. WDYT?

jalvz: If so we don't have any flexibility/control of our own scripts, and we are forced to run what they run - eg. they run `check` and we run `check-full`, we need that.

Each beat probably also need specific things, i guess that is why they have `common.bash` and a thin wrapper on top, that sounds right to me.

PR #1256, opened 2018-08-03
Fix packaging and update beats

simitt:
* Include `_beats/dev-tools/packaging/packages.yml` again as needed.
* Update beats framework to b84e083
* Exclude `dmg` build in `customizePackaging` in `magefile.go` (cc @andrewkroh)


script/update_beats.sh
```diff
@@ -27,7 +27,6 @@ rsync -crpv --delete \
     --exclude="dev-tools/jenkins_ci.ps1" \
     --exclude="dev-tools/jenkins_intake.sh" \
     --exclude="dev-tools/packaging/preference-pane/***" \
-    --exclude="dev-tools/packaging/packages.yml" \
```
jalvz: 💯

simitt:
good point! I just checked and did not find any `dmg` specific docs in the Installation section.

PR #1260, opened 2018-08-07
Intake protocol v2

### LINKED ISSUE #1238: Implement Intake V2
Implement intake v2 based on elastic/apm-server#930.

Includes:
* v2 endpoints
* processing of ndjson objects
* allowing spans to be sent separatly
* add `timestamp` and `transaction.id` to spans
* allow for same compression methods as v1
* allow for `transactions, spans, errors, metadata`

roncohen:
TODO:

- [x] Implement the http error response
- [x] Handle unexpected read errors by responding with an InternalServerError early
- [x] When responding early, use "Connection: Close" instead of reading the full request body
- [x] Default timestamp to request time when no timestamps are given
- [x] Handle transaction ids on Spans
- [x] Properly handle special case for invalid metadata
- [ ] ~~Finalize _processor/package_tests_ style tests for v2~~ (deferred https://github.com/elastic/apm-server/issues/1288) 
- [x] "timeout waiting to be processed" error is in v1 format
- [ ] ~~Investigate "full queue" behavior (reporting timeout is 1s atm., which is probably too high)~~ (defered https://github.com/elastic/apm-server/issues/1298)
- [ ] ~~Investigate which limits/configs are useful for intake v2~~ (defered https://github.com/elastic/apm-server/issues/1299)
- [ ] ~~Consider unsetting/increasing `read_timeout` when using v2 to avoid timeouts~~ (also handled by https://github.com/elastic/apm-server/issues/1299)

Closes #1238 

**Config options of note:**
On the server there are some config options that are interesting for v2:
- read timeout (30s default atm.)
if the agent doesn’t complete its request within the `read_timeout`, atm. APM Server will force close the connection with an internal error
- max_unzipped_size (30mb default atm.)
if the unzipped size surpasses `max_unzipped_size` the rest will be rejected
- concurrent_requests (5 default atm.)
for v2, this will be the maximum number of agents that we can receive data from. For v2 we should make it significantly or disable it entirely for v2, as we no longer need to keep it low to conserve memory


simitt:
update: @graphaelli and I looked into the issues, and it seems that the server is not handling the _read_timeout_ properly. The requeing of non-existent requests starts when the read timeout is hit.

beater/v2_handler.go
```diff
@@ -0,0 +1,269 @@
...
+	return nil, cannotValidateResponse(errors.New("did not recognize object type"))
+}
+
+// readBatch will read up to `batchSize` objects from the ndjson stream
+// it returns a slice of eventables, a serverResponse and a bool that indicates if we're at EOF.
+func (v *v2Handler) readBatch(batchSize int, reader *decoder.NDJSONStreamReader, response *streamResponse) ([]transform.Transformable, bool) {
+	var err error
+	var rawModel map[string]interface{}
+
+	eventables := []transform.Transformable{}
```
simitt: If you are using `var eventables []transform.Transformable` instead of creating an empty array, then the formerly experienced CPU load and queuing of non-existing events disappears. 

Still think there are more issues here though, as `eof` seems to not be read properly,

simitt: thx to @graphaelli for pointing out the cpu usage is going on in this method.

roncohen:
great feedback. Thanks @silvia! 
Great find regarding the `read_timeout`. I'll get right on that. 

I think parallelizing decoding makes sense, but I might leave it for a follow up PR. When we do it, we would need to benchmark it make sure there's significant speedups. I'll make an issue it it doesn't go into this PR.

roncohen:
great feedback. Thanks @simitt ! 
Great find regarding the `read_timeout`. I'll get right on that. 

I think parallelizing decoding makes sense, but I might leave it for a follow up PR. When we do it, we would need to benchmark it make sure there's significant speedups. I'll make an issue it it doesn't go into this PR.

simitt:
Created a follow-up task for the v2 json spec changes, https://github.com/elastic/apm-server/issues/1276.

model/span/span.go
```diff
@@ -56,7 +65,7 @@ type Span struct {
 	TransactionId string
 }
 
-func DecodeSpan(input interface{}, err error) (*Span, error) {
+func DecodeSpan(input interface{}, err error) (transform.Transformable, error) {
```
simitt: You need to add setting the `Timestamp` to the `DecodeSpan` method, as for v2 the spans are not sent within `transactions` any more. 
There was an agreement that if the Timestamp is not set it should be set to the time "that the HTTP request from the agent was first received. "

beater/v2_handler.go
```diff
@@ -0,0 +1,269 @@
...
+
+	// augment the metadata object with information from the request, like user-agent or remote address
+	reqMeta, err := v.requestDecoder(r)
+	if err != nil {
+		return nil, cannotDecodeResponse(err)
+	}
+
+	for k, v := range reqMeta {
+		utility.MergeAdd(rawMetadata, k, v.(map[string]interface{}))
+	}
```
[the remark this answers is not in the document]
roncohen: we discussed offline and decided to defer this to subsequent PRs

beater/route_config.go
```diff
@@ -0,0 +1,151 @@
...
+	}
+	metricsRouteType = routeType{
+		metricsHandler,
+		backendMetadataDecoder,
+		func(*Config) transform.Config { return transform.Config{} },
+	}
+	sourcemapRouteType = routeType{
+		sourcemapHandler,
+		backendMetadataDecoder,
+		rumTransformConfig,
```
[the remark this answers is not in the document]
roncohen: it just needs `config.Smapper` to be set. `rumTransformConfig` does that.

roncohen:
I've made a number of improvements and update the description with a todo of the remaining items. Would be great to get another round of feedback.

I created https://github.com/elastic/apm-server/issues/1285 to track that separately.

model/span/span.go
roncohen: Instead of setting the request time during decoding, I'm now doing it during transformation. I added a request time to the transformation context which is used to set the time if it's absent.

roncohen:
After a brief offline discussion, we agreed to defer _processor/package_tests_ style tests (https://github.com/elastic/apm-server/issues/1288) to a separate PR to get this in.

beater/v2_handler.go
```diff
@@ -0,0 +1,297 @@
...
+	metadata, err := metadata.DecodeMetadata(rawMetadata)
+	if err != nil {
+		return nil, err
+	}
+
+	return metadata, nil
+}
+
+func (v *v2Handler) handleRequestBody(r *http.Request, ndReader *decoder.NDJSONStreamReader, report reporter) *StreamResponse {
+	requestTime := requestTime(r)
```
simitt: same, you can move this down to when the `transform.Context` is created.

beater/v2_handler.go
```diff
@@ -0,0 +1,297 @@
...
+	for {
+		transformables, done := v.readBatch(batchSize, ndReader, resp)
+		if transformables != nil {
+			err := report(r.Context(), pendingReq{
+				transformables: transformables,
+				tcontext:       tctx,
+			})
+
+			if err != nil {
+				if strings.Contains(err.Error(), "publisher is being stopped") {
```
simitt: Can you check for `errChannelClosed` instead of checking the error message?

beater/v2_handler.go
```diff
@@ -0,0 +1,297 @@
...
+					resp.Add(ShuttingDownErr, 1)
+					return resp
+				}
```
simitt: Did you miss to cover the [`ctx.Error()`](https://github.com/elastic/apm-server/blob/master/beater/pub.go#L124)?

beater/v2_handler_test.go
```diff
@@ -0,0 +1,223 @@
...
+	assert.NoError(t, err)
+	reqTimestamp, err := time.Parse(time.RFC3339, "2018-01-02T10:00:00Z")
+	assert.NoError(t, err)
+
+	for idx, test := range []struct {
+		body         string
+		contentType  string
+		err          *StreamResponse
+		expectedCode int
+		reported     []transform.Transformable
```
simitt: Could you add a test for invalid headers, ERR_INVALID_JSON, ERR_SHUTTING_DOWN, .. please

decoder/req_decoder.go
```diff
@@ -65,23 +65,23 @@ func (mr monitoringReader) Close() error {
...
-// of json data from the body of the request
-func readRequestJSONData(maxSize int64) Reader {
+// CompressedRequestReader makes a function that uses information from an http request to construct a Limited ReadCloser
+// from the body of the request, handling any decompression necessary
+func CompressedRequestReader(maxSize int64) ReqReader {
 	return func(req *http.Request) (io.ReadCloser, error) {
-		contentType := req.Header.Get("Content-Type")
-		if !strings.Contains(contentType, "application/json") {
-			return nil, fmt.Errorf("invalid content type: %s", req.Header.Get("Content-Type"))
-		}
```
simitt: I don't think you should remove this check for v1

beater/v2_handler.go
```diff
@@ -0,0 +1,297 @@
...
+
+	eventables := []transform.Transformable{}
+	for i := 0; i < batchSize && err == nil; i++ {
+		rawModel, err = reader.Read()
+		if err != nil && err != io.EOF {
+			switch e := err.(type) {
+			case decoder.ReadError:
+				response.AddWithMessage(ServerError, 1, e.Error())
+				// return early, we can't recover from a read error
+				return eventables, true
```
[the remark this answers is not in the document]
roncohen: I was thinking that if there are events that we already read and validated we should `report` them.

beater/v2_handler.go
roncohen: Good find. I just copied it from the original v1 handler, but will update both.

beater/v2_handler.go
roncohen: It turns out that this is actually equivalent to what we have today. 

Unfortunately, we throw away `ctx.Error()` in v1:  https://github.com/elastic/apm-server/blob/master/beater/handlers.go#L156-L161.

I'll check if we can safely change it in both places so it becomes something similar to `errors.Wrap(ctx.Error(), "queue is full")`

decoder/req_decoder.go
roncohen: The check has been moved to here and it's still in effect for v1: https://github.com/elastic/apm-server/pull/1260/files/fb77d985e9ee8623290bc23839542e9b318472c8#diff-5a9d3e83c01f2b2c9aa1bea1afade025R68

roncohen:
thanks a lot @simitt! great feedback again 👍

decoder/req_decoder.go
simitt: Thanks @ron - not sure how I could overlook that.

beater/v2_handler.go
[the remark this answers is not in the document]
simitt: Fair enough to publish those events.

Out of curiosity - why did you decide to read and publish in batches instead of one by one?

beater/v2_handler.go
```diff
@@ -0,0 +1,297 @@
...
+
+	tctx := &transform.Context{
+		RequestTime: requestTime,
+		Config:      v.tconfig,
+		Metadata:    *metadata,
+	}
+
+	for {
+		transformables, done := v.readBatch(batchSize, ndReader, resp)
+		if transformables != nil {
```
simitt: This would still report empty arrays (if no data were read or a read error occurred an empty array can be returned), which doesn't cause errors but is unnecessary, please see https://github.com/elastic/apm-server/pull/1260/files/fb77d985e9ee8623290bc23839542e9b318472c8#r208983595

beater/route_config.go
```diff
@@ -0,0 +1,152 @@
...
+	"github.com/elastic/apm-server/processor/transaction"
+
+	"github.com/elastic/apm-server/decoder"
+	"github.com/elastic/apm-server/transform"
+	"github.com/elastic/beats/libbeat/logp"
+)
+
+var (
+	rootURL                   = "/"
+	BackendTransactionsURL    = "/v1/transactions"
```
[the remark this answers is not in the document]
roncohen: thanks @simitt. You mention that you're not in favour of introducing _backend_. I'm not really _introducing_ the terminology here, just moved the definitions and and then used the same terminology for v2. 

I'm not sure if you're suggesting i just change the v2 terminology or i change all of them, but in any case, i would prefer to to not change terminology for v1 in this PR because i don't think that terminology relates to the changes here. And if we're not changing v1 terminology here, then i don't think we should use a different terminology for v2 because that would make it inconsistent.

beater/v2_handler_test.go
roncohen: thanks! There's a test for invalid content type headers. Did you have other invalid headers in mind?

beater/v2_handler.go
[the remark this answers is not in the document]
roncohen: good question. With v1, entire payload are put into our queue. There's overhead in pushing to go queues and reading from them, so i wanted to make sure we where not paying that overhead for each even, especially because that wasn't an overhead we had in v1.

script/inline_schemas/inline_schemas.go
```diff
@@ -31,12 +31,18 @@ const basePath = "./docs/spec/"
...
-		{"sourcemaps/payload.json", "model/sourcemap/generated/schema/payload.go"},
+		{"errors/payload.json", "model/error/generated/schema/payload.go", "PayloadSchema"},
+		{"transactions/payload.json", "model/transaction/generated/schema/payload.go", "PayloadSchema"},
+		{"metrics/payload.json", "model/metric/generated/schema/payload.go", "PayloadSchema"},
+		{"sourcemaps/payload.json", "model/sourcemap/generated/schema/payload.go", "PayloadSchema"},
+
+		{"metadata.json", "model/metadata/generated/schema/metadata.go", "ModelSchema"},
+		{"errors/error.json", "model/error/generated/schema/error.go", "ModelSchema"},
+		{"transactions/transaction.json", "model/transaction/generated/schema/transaction.go", "ModelSchema"},
+		{"transactions/span.json", "model/span/generated/schema/transaction.go", "ModelSchema"},
```
simitt: this should be `span`

beater/v2_handler_test.go
roncohen: Added tests for ERR_INVALID_JSON, ERR_SHUTTING_DOWN and found an actual bug ;) thanks!

beater/stream_response_test.go
```diff
@@ -0,0 +1,91 @@
...
+		"invalid":0,
+		"dropped":0,
+		"errors":{
+			"ERR_QUEUE_FULL":{
+				"count":23,
+				"message":"queue is full"
+			}
+		}
+	}`
+	expectedJSON = strings.Replace(strings.Replace(expectedJSON, "\n", "", -1), "\t", "", -1)
```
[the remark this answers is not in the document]
roncohen: good idea. I think this goes into _processor/package_tests_ style tests. If you agree, I'll defer it to the subsequent PR as previously discussed.

beater/route_config.go
```diff
@@ -103,6 +111,12 @@ var (
 	}
 )
 
+func v2backendHandler(beaterConfig *Config, h http.Handler) http.Handler {
+	return logHandler(
+		requestTimeHandler(
+			authHandler(beaterConfig.SecretToken, h)))
```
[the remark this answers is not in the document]
roncohen: I see what you mean. However, the current default of 5 makes it impossible to support more than 5 agents in v2 out of the box without changing the default value - which would affect v1 negatively. I think the solution is to create a limited number (`runtime.NumCPU`) of separate goroutines that read from the streams. That would also parallelize decoding and validation. I'll take a look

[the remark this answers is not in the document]
roncohen: I've now changed the base branch to a `v2`. With that in mind, I would like to defer this to a separate PR. v2 is already way safer memory wise than v1, even the "edge case" you mention here. I don't think this particular issue should hold up the many other issues that are waiting for this PR.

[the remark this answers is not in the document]
roncohen: Added it to this one: https://github.com/elastic/apm-server/issues/1285

roncohen:
Let me know you think there's more we need to do here @simitt

beater/v2_handler_test.go
```diff
@@ -0,0 +1,328 @@
+// Licensed to Elasticsearch B.V. under one or more contributor
```
simitt: `errors` and `metrics` are missing completely from these tests. There is an open Issue for handling errors, but none for metrics. Please add an Issue or some tests.

roncohen: missed this one. We have an issue for adding tests for all the different event types here: https://github.com/elastic/apm-server/issues/1288

PR #1271, opened 2018-08-09
clarify that is possible to use a secret token when uploading sourcemaps

### LINKED ISSUE #1268: Document using secret token to upload sourcemaps
We should document that setting a secret token would affect the sourcemap endpoint as well as how to use that token when uploading the sourcemap

jalvz:
fixes #1268

@jahtalab is this good enough?

### DIFF
--- a/docs/sourcemap-api.asciidoc
+++ b/docs/sourcemap-api.asciidoc
@@ -25,6 +25,8 @@ The request must include some fields needed to identify `source map` correctly l
 The source map must follow the
 https://docs.google.com/document/d/1U1RGAehQwRypUTovF1KRlpiOFze0b-_2gc6fAH0KY0k[Source map revision 3 proposal] spec and be attached as a `file upload`.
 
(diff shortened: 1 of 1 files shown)


jahtalab:
Thanks @jalvz ! I would also add a note about the fact that this is the same token used to restrict backend agent's endpoint, so that users that already have the token for backend agents, they know that they need to provide the same token to the sourcemap upload endpoint.

jalvz:
Isn't that obvious? I don't want to be overlay verbose, that sentence links to https://www.elastic.co/guide/en/apm/server/master/securing-apm-server.html#secret-token which fairly clear imo, nowhere mentions that there are 2 different tokens or anything; it says that the secret token "authorizes requests to the APM Server"

jahtalab:
@jalvz , I just think, if someone is getting a permission denied error while uploading the sourcemap, they might not think of the token they have set for backend agents.

But if you think it's fine, then it LGTM.

PR #1272, opened 2018-08-10
Update beats framework to e30da2c

simitt:
* Update to latest beats framework
* Fix update-beats script, as this was broken after https://github.com/elastic/beats/pull/7855 got merged.
* Fix govendor issues with indirect dependency of 
  * `github.com/elastic/go-txfile` as beats origin was missing when vendoring. 
  * `github.com/spf13/pflag`

manual step:
* Reinstall `go-licenser` as used version was outdated.


simitt:
jenkins, retest this please.

simitt:
There is one failing test:
screen shot 2018-08-10 at 13 59 29

Which should have nothing to do with this PR.

simitt:
jenkins, retest this please

NOTICE.txt
```diff
@@ -1721,7 +1721,6 @@ Apache License 2.0
 
 --------------------------------------------------------------------
 Dependency: github.com/spf13/pflag
-Version: v1.0.1
```
graphaelli: strange that this was dropped but otherwise left unchanged

script/update_beats.sh
```diff
@@ -7,6 +7,7 @@ BEATS_VERSION="${BEATS_VERSION:-master}"
 # Find basedir and change to it
 DIRNAME=$(dirname "$0")
 BASEDIR=${DIRNAME}/../_beats
+rm -rf $BASEDIR
```
graphaelli: doesn't this prevent us from owning any files in `_beats/` ourselves?  Maybe that's the goal, but we historically have overridden a few files, including `dev-tools/packer/readme.md.j2` which doesn't actually exist any more (will check what's in our packages these days next).

vendor/vendor.json
```diff
@@ -1839,12 +1874,11 @@
 			"revisionTime": "2017-11-23T00:13:03Z"
 		},
 		{
-			"checksumSHA1": "k5TsfhX2vvfffYbZQ4fMKtL5+cA=",
+			"checksumSHA1": "STxYqRb4gnlSr3mRpT+Igfdz/kM=",
+			"origin": "github.com/elastic/beats/vendor/github.com/spf13/pflag",
 			"path": "github.com/spf13/pflag",
-			"revision": "583c0c0531f06d5278b7d917446061adc344b5cd",
-			"revisionTime": "2018-04-12T12:09:13Z",
-			"version": "v1.0.1",
```
graphaelli: this shouldn't have an origin, we use this directly.  This is very confusing

script/update_beats.sh
graphaelli: following up on that readme, all is well - https://github.com/elastic/apm-server/blob/master/package-README.md.tmpl takes care of our customizations now.  Let's just remove that exclude since it does nothing anyway, looks like it should be fine to not own anything under `_beats`

vendor/vendor.json
simitt: This is also vendored by beats, so I run a `govendor delete github.com/spf13/pflag` and then `make update-beats`, which was adding it as a beats dependency. I think we should have the same versions as beats wherever we use the same dependencies. 
However, I revert this change for now and we can discuss our vendoring approach for such situations.

script/update_beats.sh
simitt: As we don't have any files that we overwrite, I thought it was a good idea to clean out the `_beats` folder regularly, as otherwise we keep files that get removed in beats, which can be confusing, e.g. the [fields.generated.yml](https://github.com/elastic/apm-server/pull/1272/files#diff-66cab9e9cbfa9dc1c2c954f7b0d5f880) that doesn't exist anymore.

PR #1277, opened 2018-08-13
add release notes for 6.4

jalvz:
screen shot 2018-08-13 at 11 11 10

### DIFF
--- a/docs/guide/index.asciidoc
+++ b/docs/guide/index.asciidoc
@@ -56,6 +56,48 @@ To visualize the data after it's sent to Elasticsearch,
 you can use the the dedicated APM UI bundled in X-Pack,
 or the pre-built open source Kibana dashboards that can be loaded directly in the Kibana UI.
 
+[[apm-release-notes]]
+=== Release notes
+
+[float]
+==== APM version 6.4
+
+[float]
+===== Breaking changes
+
+We previously split APM data into separate indices (transaction, span, error, etc.).
+In 6.4 APM Kibana UI starts to leverage those separate indices for queries.
+In case you only update Kibana but run an older version of APM Server you will not be able to see any APM data by default.
+To fix this, use the Kibana APM configuration options to specify the location of the APM index.
+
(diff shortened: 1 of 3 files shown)


docs/guide/index.asciidoc
```diff
@@ -56,6 +56,48 @@ To visualize the data after it's sent to Elasticsearch,
...
+[float]
+==== APM version 6.4
+
+[float]
+===== Breaking changes
+
+We previously split APM data into separate indices (transaction, span, error, etc.).
+In 6.4 APM Kibana UI starts to leverage those separate indices for queries.
+In case you only update Kibana but run an older version of APM Server you will not be able to see any APM data by default.
+To fix this, use the Kibana APM configuration options to specify the location of the APM index.
```
simitt: Could you link to https://github.com/elastic/kibana/blob/2ad265b73f47efe2cfc5d209d8a4f006f3e5ca09/docs/settings/apm-settings.asciidoc?

docs/guide/index.asciidoc
```diff
@@ -56,6 +56,48 @@ To visualize the data after it's sent to Elasticsearch,
...
+
+When upgrading to Kibana 6.4, you might need to refresh your APM index pattern for certain APM UI features to work.
+
+[float]
+===== New features
```
simitt: I think it would make sense to link to the according changelog sections or if not available to the overview for Server, UI and Agents.

docs/guide/index.asciidoc
jalvz: done here eaf82ead415fdc8eb574529899f43899f4f419eb

docs/guide/index.asciidoc
jalvz: agents are not adding their changelogs to the Elastic docs, and I'd prefer not to link to GH because we can't detect broken links. 
(also personally I don't think it looks very nice to follow a link from elastic docs to a GH readme, even thou i don't know if others are doing that)

PR #1302, opened 2018-08-21
Fix filepath to approvals folder.

simitt:


simitt:
Good point, locally they have not passed. I'll check back with Jenkins.

Issue #1276, opened 2018-08-13
Update json spec for v2 Intake related to distributed tracing


simitt:
At least forbid `transaction.spans` to be sent in Intake V2.

Check if other changes are necessary, e.g. regarding `span.id` or `span.transaction_id`.

simitt:
@elastic/apm-agent-devs and @roncohen as Intake v2 is a breaking change anyways, we could also take the opportunity to apply some breaking changes on the schema specs:

* the `span.id` is currently defined as an integer. Are we deferring to change the format to a string until we implement distributed tracing or should we do this from the beginning of v2? In case we change it, we also need to change the ES schema, similar to what we did in the distributed tracing POC. 
* will the `span.id` continue to be optional? The `transaction.id` is required. In case we generally want to use `event.id` for any cross referencing with logs etc we should treat them the same. If there are valid use cases to not have an ID on spans keeping it optional is fine with me. 
* `span.transaction_id` is optional. For a transaction `id` is required. Having `span.transaction_id` optional does only make sense if a span can exist outside a transaction, otherwise I suggest to make it also required.

felixbarny:
> Are we deferring to change the format to a string until we implement distributed tracing or should we do this from the beginning of v2? 

IIRC, there will be no non-DT version of v2, so changing that to a string sounds fine. Regarding the ES schema, it would still have to be backwards compatible I guess?

What is `event.id`? Does it come from ECS?

What is `span.transaction_id` used for again? I think it was for being backwards compatible with the old UI. If we don't need this anymore in version 7 of the stack, we should leave it optional. But maybe I'm missing some of the use-cases for `span.transaction_id`.

simitt:
Thanks @felixbarny !
* I also think we will not release a non-dt v2. I am not sure whether or not to change the format from `long` to `string` now or when we implement dt. The agents might want to test against intake v2 already. Also I am not sure if we agreed on how the format for `span.id` will look exactly. I tend to leave it unchanged for now and change when we introduce dt. 

* `event.id` was my attempt to talk about `transaction.id`, `span.id`, `metric.id`, etc in one term as all of them are events. 

* `span.transaction_id` is not new and still necessary afaik. It is used to query a transaction's spans. Spans are stored as separate documents in ES and we somehow need to relate them to a transaction.

felixbarny:
> The agents might want to test against intake v2 already.

If I'm not mistaken, supporting the non-DT schema with v2 would just complicate things for the agents. 

[...]

PR #1303, opened 2018-08-21
Update Intake API spec for v2

simitt:
implements #1276 

# Changes Intake API Json specs for v2 and distributed tracing:
### JSON spec V2
- remove pattern of `error.id`
- remove `transaction.id`
- add `transaction_id` referring to transaction
- add `parent_id`
- add `trace_id`

### ES schema
- add `error.parent_id`
- add `error.trace_id`

### JSON spec
- remove pattern of `transaction.id`
- remove `spans` (don't forbid them, just don't process them)
- add `parent_id`
- add `trace_id` (required attribute)

### ES schema
- add `transaction.parent_id`
- add `transaction.trace_id`

### JSON spec
- change pattern of `id` from long to string
- remove pattern of `transaction_id`
- remove `parent` referring to long id
- add `parent_id` referring to transaction or span
- add `timestamp`
- add `trace_id` (required attribute)

### ES schema
- add `span.parent_id`
- add `span.trace_id`
- add `span.hex_id`
- deprecate `span.id`
- deprecate `span.parent`

## Open Questions:
* Any changes for required optional fields necessary? (@elastic/apm-agent-devs )

**UPDATE**:
After some discussion with @roncohen we decided to also add distributed tracing support to this PR. 
Thus, `parent_id` and `trace_id` have been added.

* The `span.id` is not used for anything, thus will simply be deprecated. (cc @elastic/apm-ui )


docs/spec/errors/v2_error.json
```diff
@@ -0,0 +1,23 @@
+{
+    "$id": "docs/spec/errors/v1_error.json",
```
graphaelli: v2_error

model/span/_meta/fields.yml
```diff
@@ -12,9 +12,14 @@
...
-          type: long
+          type: long 
+          description: >
+            Deprecated (>= 6.5). A locally unique ID of the span.
+
+        - name: hex_id
+          type: keyword 
           description: >
-            A locally unique ID of the span.
+            Th ID of the span.
```
graphaelli: typo

model/span/event.go
```diff
@@ -52,49 +53,80 @@ func ModelSchema() *jsonschema.Schema {
...
 
-func DecodeSpan(input interface{}, err error) (transform.Transformable, error) {
-	if input == nil || err != nil {
+func V2DecodeEvent(input interface{}, err error) (transform.Transformable, error) {
+	e, raw, err := decodeEvent(input, err)
+	if err != nil {
 		return nil, err
 	}
+	decoder := utility.ManualDecoder{}
+	fmt.Println(raw)
```
roncohen: :)

model/span/event.go
```diff
@@ -43,7 +44,7 @@ var (
 
 	spanDocType = "span"
 
-	processorSpanEntry = common.MapStr{"name": "transaction", "event": spanDocType}
```
roncohen: for transactions, this is called `processorTransEntry`, and for errors it `processorEntry`. Let's align them on `processorEntry`?

roncohen:
Great work @simitt !

> deprecate id

for 7.0, I assume we'll deprecate `hex_id` and just use `id`? If so, deprecate `id` now could send the wrong signal.

simitt:
@roncohen I suggest to go with @felixbarny 's  suggestion to use `id` instead of `span.hex_id` which aligns with ECS, and deprecate `span.id`. I assume for 7.0 we will align with ECS as much as possible anyways.

roncohen:
@simitt gotcha 👍

felixbarny:
It probably makes sense to deprecate `transaction.id` in favor of `id` in this PR as well. Btw, in [fields.asciidoc](https://github.com/elastic/apm-server/blob/master/docs/fields.asciidoc#service-fields) the `transaction.id` is listed under "service fields" as opposed to "APM transaction fields".

simitt:
We discussed the `fields.asciidoc` issues a while ago with the conclusion to leave as is for now and rework the fields doc generally. 

The `transaction.id` and the `error.id` are currently used and referenced at some places in the dashboards and the UI. The `span.id` isn't afaik. I suggest to keep the ES changes as minimal as possible as this will go out with a minor release. We can aim to align the schema in 7.0

felixbarny:
related: https://github.com/elastic/ecs/issues/99#issuecomment-415029185

> The id you linked from the apm use case has become `event.id`: https://github.com/elastic/ecs#-event-fields

simitt:
I don't think that we should introduce ECS compatibility for the id then with this PR. We would end up with: 
```
{
    "span": { "id": 123, "type": "db", ... },
    "event": { "id": "0123456789abcdef" }, 
    ... 
}
```
I find that very confusing, so I stick with the original idea of introducing a `span.hex_id`.

felixbarny:
Yes, let's make that a different concern. We probably need to discuss what the appropriate ECS name should be first.

simitt:
@sqren could you please check and lmk if the current APM UI is using `span.id` somewhere.

sqren:
@simitt `span.id` is not used anywhere. I thought we did, but it is actually overwritten:

https://github.com/elastic/kibana/blob/02a56b39b10db48b89966b64c1014c16644d9421/x-pack/plugins/apm/server/lib/transactions/spans/get_spans.js#L58

CHANGELOG.asciidoc
```diff
@@ -14,6 +14,8 @@ https://github.com/elastic/apm-server/compare/6.4\...master[View commits]
 
 - Provide basic server information at `/` {pull}1197[1197]
 - Add pipeline registration and pipeline usage {pull}1258[1258]
+- Update JSON schema spec for v2 {pull}1303[1303]
+- Make JSON schema spec distributed tracing compliatn {pull}1303[1303]
```
roncohen: compliant

model/span/event.go
roncohen: just wanted to make sure you didn't miss this by mistake. It's not critical, so it's fine if you just decided to defer the discussion.

simitt: I updated now to `processorEntry`.

simitt:
jenkins, retest this please

simitt:
jenkins, retest this please.

simitt:
jenkins, test this please.

PR #1306, opened 2018-08-22
Update beats framework to a2b0f15

graphaelli:
following up on #1305, added origin entries for `go-structform` and many other packages [manually](https://gist.github.com/graphaelli/714f4965ecf9530c45c77b69334030b5)


vendor/vendor.json
```diff
@@ -1980,6 +2001,18 @@
 			"revision": "f85c78b1dd998214c5f2138155b320a4a43fbe36",
 			"revisionTime": "2017-10-30T23:38:06Z"
 		},
+		{
+			"checksumSHA1": "2LpxYGSf068307b7bhAuVjvzLLc=",
+			"path": "golang.org/x/crypto/ed25519",
```
simitt: Shouldn't we also have an origin here?

vendor/vendor.json
```diff
@@ -2002,42 +2035,66 @@
...
+		},
+		{
+			"checksumSHA1": "K4XNY0c60IPBpv5sO6aiCuB8o/0=",
+			"path": "golang.org/x/net/ipv4",
+			"revision": "922f4815f713f213882e8ef45e0d315b164d705c",
+			"revisionTime": "2018-08-02T22:19:15Z"
+		},
+		{
+			"checksumSHA1": "hDszlYGI8m9puPsEIDr59ccTiF8=",
+			"path": "golang.org/x/net/ipv6",
```
simitt: same as above, all newly vendored 3rd party libs should have `origin`.

vendor/vendor.json
graphaelli: you're right.  I'll go through these 1 by 1

graphaelli:
added origins using https://gist.github.com/graphaelli/714f4965ecf9530c45c77b69334030b5.  I'm holding off on adding that to our repo and integrating it into beats-update since that process might change soon, but we should run it one off for 6.4 too.

The next beats update will pull in more than usual as a result, but I didn't want to confuse review here with that update.

graphaelli:
@simitt this is ready for another look

PR #1318, opened 2018-08-27
Fix changelog description.

simitt:


simitt:
Because PR #1029 was targeted against `6.4` only.

PR #1330, opened 2018-08-28
Make docs pass

roncohen:
We need to do a proper round of fixes on the docs for v2. This just gets them passing for now.


docs/transaction-api.asciidoc
```diff
@@ -64,7 +55,7 @@ include::./spec/transactions/transaction.json[]
 
 [source,json]
 ----
-include::./spec/span/span.json[]
+include::./spec/span/v1_span.json[]
```
simitt: needs to be `spec/spans/v1_span.json`

roncohen:
there's an unrelated windows failure, but docs are passing, so I'll merge.

PR #1342, opened 2018-08-30
maintain changelog per minor release

graphaelli:
As a user, I expect https://www.elastic.co/guide/en/apm/server/6.4/release-notes-6.4.0.html and https://www.elastic.co/guide/en/apm/server/master/release-notes-6.4.0.html to be identical. 
 Unfortunately, that is frequently not the case.  Keeping those release notes in sync across branches is a manual and time consuming process and also very important.

This is a proposed implementation for simplifying that process.  As an example, `changelogs/6.4.asciidoc` should be identical in the `6.4` branch and on `master`, making keeping those in sync straightforward.

One drawback is that CHANGELOG.asciidoc does not contain all of the changelog entries in the source repo - I'm not sure if that's a big concern for others, it's not for me.

The process of backporting changes remains the same, though still a bit complicated:

1. Fix bug / add feature on `master` branch, note in `head` changelog (step could be skipped)
2. Backport to `release` branch including note in `head` changelog (does not show up anywhere)
3. At next patch release, move entry from `head` to `minor` release changelog, in new patch release section, on `release` branch
4. backport (forwardport?) that move from `head` to `minor` on `master` branch

Steps 3 and 4 could be staged ahead of release with lower fear of merge conflicts than with a single large changelog.


graphaelli:
I like that, so everyone can understands what will land next instead of hiding it in `head` until the next release.  In that case, how should we handle the case where something is backported but no release actually happens?  Should we remove the `unreleased` section periodically (when the next minor is released) ?

graphaelli:
Per discussion, we'll deal with that situation as it arises - we generally know when another patch release is expected for older versions and don't backport if we're not expecting a release anyway.

PR #1344, opened 2018-08-30
docs(node): update Node.js code snippet

watson:
The upcoming version of the Node.js agent (v1.12.0) will have automatically pick the a serviceName from the local package.json file. It's going to be released now and will be compatible with all versions of the APM Server from v6.2+


watson:
@Qard what do you think of this way of writing it?

PR #1347, opened 2018-08-31
convert stream integration tests to approvals flow.

roncohen:
Simplifies integrations tests and makes use of the nice "approvals" tools for all of them


processor/stream/approved-es-documents/testV2IntakeIntegrationOptionalTimestamps.approved.json
```diff
@@ -0,0 +1,146 @@
+{
+    "events": [
+        {
+            "@timestamp": "2018-08-01T10:00:00Z",
```
simitt: I assume this should test that a given timestamp is not overwritten, but if it is missing the approrpiate timestamp is set. Please use two different timestamps for this (atm the default timestamp is the same as the one used in the testdata file).

Issue #1300, opened 2018-08-20
Add APM specific track to `rally` for ES performance measurements


simitt:
[Rally](https://github.com/elastic/rally) is a benchmarking framework for Elasticsearch. It uses [tracks](https://github.com/elastic/rally-tracks) for benchmarking different use cases.

Create an APM specific track with predefined APM corpora to get more insights on performance behaviour for APM documents. 

This can be helpful for figuring out default settings, e.g. around batch sizes and limiting the number of allowed fields per index.

PR #1349, opened 2018-08-31
Add basic ES rally track for APM.

simitt:
This is a first draft to get an APM rally track defined, see elastic/apm-server#1300 . 

Currently I tested with local APM corpora. I have used several agents from the apm integration testing to create sample data in ES and then fetched those data by the python script (see below).
The data will be uploaded to s3 soon, so everyone can run the rally tests with APM track. 

At the moment this only covers a basic test to ingest `error, transaction, span` data in parallel or separated from each other. This can be used for figuring out proper `bulk` sizes for ingestion and other ES ingestion related default settings. 

In the future we can use this to also define other default settings (field limit, etc) and to performance test queries from @elastic/apm-ui . 

implements #1300 

(cc @danielmitterdorfer )


rally/.gitignore
```diff
@@ -0,0 +1,4 @@
+documents/*.json.offset
```
danielmitterdorfer: When Rally downloads data from a remote server, it will end up in `~/.rally/benchmarks/data/apm-events` instead of the current directory (it is just that Rally will search both the current directory and the data directory I mentioned before for convenience when you develop locally).

rally/challenges/default.json
```diff
@@ -0,0 +1,52 @@
...
+          "operation": {
+            "operation-type": "create-index",
+            "settings": {{index_settings | default({}) | tojson}}
+          }
+        },
+        {
+          "name": "check-cluster-health",
+          "operation": {
+            "operation-type": "cluster-health",
+            "index": "logs-*",
```
danielmitterdorfer: Should this be `apm-*`?

rally/challenges/default.json
```diff
@@ -0,0 +1,52 @@
...
+            "request-params": {
+              "wait_for_status": "{{cluster_health | default('green')}}",
+              "wait_for_no_relocating_shards": "true"
+            }
+          }
+        },
+        {
+          "name": "index-apm-error",
```
danielmitterdorfer: You are bulk-indexing one index after the other so you might be prone to ordering bias (i.e. the results would be different if you change the order of those three operations).

Another problem is that this does not do what you intend to do. The parameter to specify the target indices is called `indices` instead of `index` *and* it needs to be specified on the operation itself (see [reference docs](https://esrally.readthedocs.io/en/stable/track.html#bulk)). I also suggest you just inline the operation. E.g.:

```json
{
  "name": "index-apm-transaction",
  "operation": {
    "operation-type": "bulk",
    "indices": ["apm-6.4.0-transaction"],
    "bulk-size": {{bulk_size | default(1000)}},
    "ingest-percentage": {{ingest_percentage | default(100)}}
  },
  "warmup-time-period": 120,
  "clients": {{bulk_indexing_clients | default(8)}}
}
```

W.r.t. to the `indices` parameter you also need to be specific because Rally tries to match the specified indices against the target indices specified in the `corpora` block. I guess the reason you have specified a `*` is that you want it to match 6.4.0 and 7.0.0 indices. This is *usually* solved a bit differently (by creating a track repository and using branches) but you could also use a track parameter instead and use Rally's template language support to replace the track parameter for you. E.g. you'd specify `--track-param="es_version:6.4.0"` on the command line and instead of specifying `"indices": ["apm-6.4.0-transaction"]` you can also specify `"indices": ["apm-{{es_version}}-transaction"]`.

rally/track.json
```diff
@@ -0,0 +1,55 @@
...
+        {
+          "target-index": "apm-6.4.0-span",
+          "source-file": "documents/span.json",
+          "document-count": 2225454,
+          "uncompressed-bytes": 10897616683
+        }
+      ]
+    }
+  ],
+  "operations": [
```
danielmitterdorfer: You could probably inline operations and challenges which would make it a bit simpler to read. Then you can get rid of the first line (`{% import "rally.helpers" as rally with context %}`) and the `rally.collect` calls. You can see a [simple example track](https://esrally.readthedocs.io/en/stable/adding_tracks.html#example-track) in our docs).

rally/challenges/default.json
```diff
@@ -0,0 +1,52 @@
...
+      "description": "Indexes the whole document corpus using Elasticsearch default settings. We only adjust the number of replicas as we benchmark a single node cluster and Rally will only start the benchmark if the cluster turns green. Document ids are unique so all index operations are append only.",
+      "default": true,
+      "schedule": [
+        {
+          "operation": "delete-index"
+        },
+        {
+          "operation": {
+            "operation-type": "create-index",
+            "settings": {{index_settings | default({}) | tojson}}
```
danielmitterdorfer: I think a short readme file documenting the exposed track parameters would be nice. See e.g. https://github.com/elastic/rally-tracks/tree/master/geonames#parameters

rally/track.json
simitt: I think it will be easier to avoid code duplication in the future, e.g. when adding querying, when we don't inline.

simitt:
Thanks @danielmitterdorfer for the great suggestions! I adapted the PR accordingly.

rally/track.json
```diff
@@ -0,0 +1,55 @@
...
+    },
+    {
+      "name": "apm-span",
+      "body": "index.json",
+      "types": [ "type" ]
+    }
+  ],
+  "corpora": [
+    {
+      "name": "apm-events",
```
danielmitterdorfer: I think a download URL is missing here so other people can use this track as well.

rally/track.json
```diff
@@ -25,19 +25,19 @@
       "target-type": "type",
       "documents": [
         {
-          "target-index": "apm-6.4.0-error",
+          "target-index": "apm-error",
           "source-file": "documents/error.json",
```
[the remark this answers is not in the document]
simitt: I'll change that once I uploaded the corpora.

rally/track.json
simitt: Totally, I haven't uploaded the corpora yet (mentioned that in the PR description.)

danielmitterdorfer: Ok, I somehow missed that (or I've assumed that the data is uploaded meanwhile). In that case I think this is good to merge.

simitt:
Sorry for not being more clear on this: It is still a WIP Pull Request and I'll only merge it when the data is uploaded. Just wanted to get some feedback and give my colleagues the possibility to familiarise themselves with the APM track.

danielmitterdorfer:
> Sorry for not being more clear on this: It is still a WIP Pull Request and I'll only merge it when the data is uploaded. Just wanted to get some feedback and give my colleagues the possibility to familiarise themselves with the APM track.

Ok. Apparently I was overeager. :) Just ping me if you need another review.

simitt:
@elastic/apm-server this PR is ready for review, you can also run the challenge locally, see the README.md

rally/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+If you want to store results to a file, preserve temporary Elasticsearch, use a running Elasticsearch instance, etc. 
+please refer to rally's [command_line_reference](https://esrally.readthedocs.io/en/stable/command_line_reference.html#command-line-flags).
+
+
+## Parameters
+When providing `--track-params` you can override following default parameters for the challenges: 
+
+* ingest_percentage: Percentage defining how much of the document corpus should be indexed. Defaults to 100. 
+* bulk-size: Number of documents per bulk operation. Defaults to 1_000.
```
ruflin: typo? 1_000?

rally/README.md
```diff
@@ -0,0 +1,39 @@
...
+For general information about rally tracks and how to use them, 
+check out the [docs](https://esrally.readthedocs.io/en/stable/index.html).
+and the [rally](https://github.com/elastic/rally) and [rally-tracks](https://github.com/elastic/rally-tracks) github 
+repos.
+
+## Corpora
+Data used in this track are dumped from example Opbeans applications, instrumented with `nodejs`, `python` and 
+`ruby` agents. The dump was created in September 2018. 
+You can also build your own corpora and reference them in the track.
+There is a [python script](https://github.com/elastic/apm-server/blob/d7b9d5027dd6a296792aa5179c0eaff8374d62d8/rally/fetch_data.py).
```
graphaelli: will this link break when this is merged / branch deleted?  probably safe to change the sha to `master`

rally/README.md
```diff
@@ -0,0 +1,39 @@
...
+### Default challenge
+Ingest data of type `error,transactions,spans` in parallel into Elasticsearch.
+You can start the race for the default challenge with `esrally --track-path=<local-path-to-apm-track>`.
+
+### Single Event Type challenge
+If you want to test only a dedicated event type, you can use the second challenge, by running 
+`esrally --track-path=<local-path-to-apm-track> --track-params="event_type:'<event_type>'" --challenge=ingest-event-type`
+
+
+If you want to store results to a file, preserve temporary Elasticsearch, use a running Elasticsearch instance, etc.
```
graphaelli: I"m not sure what `preserve temporary Elasticsearch` - maybe `for additional options` ...

rally/README.md
graphaelli: ha, this is ruby speak for 1000.  python3 also supports it now.

rally/fetch_data.py
```diff
@@ -0,0 +1,53 @@
...
+        with open(f, 'a+') as out:
+            for doc in helpers.scan(es, query=q, index=idx, size=batch_size):
+                out.write(json.dumps(doc["_source"]))
+                out.write("\n")
+
+                ct += 1
+                if ct >= doc_ct:
+                    break
+
+        os.system("wc -l {}".format(f))
```
graphaelli: can we get rid of these system calls?

graphaelli:
Ran everything locally and it went smoothly just following the docs 👍 

Curious - do the corpora need to be so large?  I'm concerned that an automated run of this will be dominated by pulling data down.  This isn't hooked up to a periodic job right now right?

rally/fetch_data.py
simitt: I added them because one has to define the `number of documents` and the `size` when defining corpora.

simitt:
The corpora have to be relatively large so the ingestion part makes sense, but probably could be cut down a bit. However, once documents are downloaded they will be reused, so it only takes long first time downloading. We don't have this set up as an automated benchmark job, and I currently see the purpose of this rally track for having an easy to use setup when ES params need to be defined or UI queries need to be checked for performance.
I suggest to keep as is and overwrite the corpora when necessary.

danielmitterdorfer:
> Curious - do the corpora need to be so large? I'm concerned that an automated run of this will be dominated by pulling data down. This isn't hooked up to a periodic job right now right?

In addition to what Silvia has already said I want to mention that benchmarks are prone to measure mostly warmup effects if the dataset is too small. We need to give the system some time to get into steady state. An alternative approach is to generate data on the fly instead of ingesting date from a file but this approach causes other problems (client-side bottlenecks, reproducibility, unrealistic distribution of data, ...).

graphaelli:
This corpora is pretty repetitive though, are there options for downloading a core set of data and generating the rest from those?

danielmitterdorfer:
> This corpora is pretty repetitive though, are there options for downloading a core set of data and generating the rest from those?

No, this is not possible at the moment. What do you mean by "repetitive"? The reason I'm asking: If the actual data are not realistic that could point out a problem with the benchmark and performance might appear better than it is in practice (Lucene picks up on those patterns and would compress data unrealistically well).

graphaelli:
I meant repetitive as in the docs are basically the same except for timestamps and minor variations in a few fields, all things that could be simulated.

PR #1350, opened 2018-09-03
v2: more tests

roncohen:


beater/v2_handler_test.go
```diff
@@ -19,17 +19,37 @@ package beater
...
+	var resultmap map[string]interface{}
+	err := json.Unmarshal(body, &resultmap)
+	assert.NoError(t, err, "err", string(body))
+
+	resultName := fmt.Sprintf("approved-stream-result/%s", name)
+	verifyErr := tests.ApproveJson(resultmap, resultName, nil)
+	if verifyErr != nil {
+		assert.Fail(t, fmt.Sprintf("Test %s failed with error: %s", name, verifyErr.Error()))
+	}
+}
```
simitt: Would it make sense to move `approveResultBody` to the `test` package? I think I saw this copy&pasted a few times in former PR already.

beater/v2_handler_test.go
```diff
@@ -81,3 +101,80 @@ func TestRequestDecoderError(t *testing.T) {
...
+				assert.Equal(t, 0, w.Body.Len())
+			} else {
+				body := w.Body.Bytes()
+				approveResultBody(t, "TestRequestIntegration"+test.name, body)
+			}
+		})
+	}
+}
+
+func TestReportingProblemsIntegration(t *testing.T) {
```
simitt: I'd appreciate to get rid of the duplicated code in `TestRequestIntegration` and `TestReportingProblemsIntegration`.

roncohen: good tip!

beater/v2_handler_test.go
roncohen: refactored a bit. WDYT?

PR #1352, opened 2018-09-03
v2: add max_line_length

roncohen:


axw:
Here's a simpler version of LineReader, which I think does what you need. As a bonus, it's faster and does no allocation. The returned `[]byte` will be invalidated by a subsequent call to ReadLine (i.e. it's not copied, hence no allocations.)

```go
// LineReader reads length-limited lines from streams using a limited amount of memory.
type LineReader struct {
        reader        io.Reader
        br            *bufio.Reader
        maxLineLength int
        skip          bool
}

func NewLineReader(reader io.Reader, maxLineLength int) *LineReader {
        return &LineReader{
                reader:        reader,
                br:            bufio.NewReaderSize(reader, maxLineLength),
                maxLineLength: maxLineLength,
        }
}

// ReadLine reads the next line from the given reader.
// If it encounters a line that is longer than `maxLineLength` it will
// return the first `maxLineLength` bytes with `ErrLineTooLong`. On the next
// call it will return the next line.
func (lr *LineReader) ReadLine() ([]byte, error) {
        for {
                line, prefix, err := lr.br.ReadLine()
                if err != nil {
                        return nil, err
                }
                if !lr.skip {
                        if prefix {
                                lr.skip = true
                                return line[:lr.maxLineLength], ErrLineTooLong
                        } else {
                                return line, nil
                        }
                } else if !prefix {
                        lr.skip = false
                }
        }
}
```

roncohen:
so much simpler. Thanks @axw!

It looks to me like it swallows any `ReadError`s. Haven't investigate too much yet though

roncohen:
jenkins, retest this please

axw:
@roncohen you're right, bufio.Reader does swallow errors in some cases. See https://github.com/golang/go/blob/b444215116e8d4a63f70f1a1b5a31f60a41a7f75/src/bufio/bufio.go#L390.

So if the reader has some data buffered, the error will be ignored. I think we can read errors to be terminal (i.e. once there's an error there will be no more data), so I don't think this matters in practice.

roncohen:
@axw we added the ReadTimeout test because if an agent doesn't complete the request within `read_timeout`, `Read` will return an error. As far as I can tell, this would be silently eaten and i was thinking that could be a problem.

I took another look to try and see if we can get it to bubble up errors. Let me know what you think

decoder/line_reader.go
```diff
@@ -47,18 +47,27 @@ func NewLineReader(reader io.Reader, maxLineLength int) *LineReader {
...
 	for {
-		line, prefix, err := lr.br.ReadLine()
-		if err != nil {
-			return nil, err
+		line, err := lr.br.ReadSlice('\n')
+		if err == bufio.ErrBufferFull {
+			prefix = true
+		} else if err == io.EOF && lr.skip {
+			lr.skip = false
+			return line, ErrLineTooLong
```
axw: This will return the remainder of a too-long line on io.EOF -- why return it on EOF when we don't in other cases? I *think* this special case should be moved below, so it's like:

```go
if !lr.skip {
   ...
} else if err == io.EOF {
    return nil, io.EOF
} else if !prefix {
    lr.skip = false
}
```

axw:
> @axw we added the ReadTimeout test because if an agent doesn't complete the request within read_timeout, Read will return an error. As far as I can tell, this would be silently eaten and i was thinking that could be a problem.

The point I was trying to make is that http.Server.ReadTimeout is a deadline: once it's passed, all reads will fail. The first error might get swallowed (because there's still some data buffered), but subsequent ones won't. Hence I don't think it's a problem in practice. Your change looks fine apart from the comment I left, though.

watson:
Would it make sense to generalize the name of the config option to not be called something with "line", as that's purely an bi-product of our current content-type. How about `max_event_size`?

roncohen:
@watson yeah, that's better

decoder/line_reader_test.go
```diff
@@ -0,0 +1,204 @@
...
+
+import (
+	"bytes"
+	"io"
+	"testing"
+	"testing/iotest"
+
+	"github.com/stretchr/testify/assert"
+)
+
```
simitt: I suggest to add a test for an empty string

processor/stream/stream_processor.go
```diff
@@ -46,6 +46,34 @@ type StreamReader interface {
...
+				Type:     InvalidInputErrType,
+				Message:  err.Error(),
+				Document: string(s.StreamReader.LatestLine()),
+			}
+		}
+
+		if err == decoder.ErrLineTooLong {
+			return nil, &Error{
+				Type:     InvalidInputErrType,
+				Message:  "event exceeded the permitted size.",
```
simitt: You could also simply use `err.Error` as the message is set in the error, and combine the two error checks as the code is the same.

roncohen: i had that before, but changed it. The Line reader returns an error that says

> Line exceeded permitted length

which is fine for a `LineReader`, but i wanted to give our users a better error message:

> event exceeded the permitted size

Issue #1297, opened 2018-08-17
Make error.exception.message optional


felixbarny:
In Java, the exception message is a nullable property, so I propose making it optional in the JSON schema as well. Are there any dragons hiding in the shadows when doing this?

simitt:
An error document has to either provide an `error.exception` or an `error.log`. For both of them a `message` is required. The `message` is the only required field for both, ensuring that if nothing else is provided, some information is available for calculating how to group errors together. 

In case the Java exception message is null, would the agent have some valuable information that could be set as message?

felixbarny:
The only useful thing would then be the `exception.type`. As a workaround, I'm currently sending `"message": "null"` if the exception message is `null`.

roncohen:
We use the message (either `log.message` or `exception.message`) in the UI, so we'd need to have a way to deal with null messages there too.

roncohen:
@felixbarny just throwing it out there, but maybe what you're doing now (`"null"`) is fine?

felixbarny:
I'm ok with just transmitting `"null"` if that's easier. If we get more reports from other agent devs that they struggle with null values, we could re-evaluate this.

If you are fine with this, feel free to close.

felixbarny:
At the moment, I don't have a use-case for `error.log` so I have not yet implemented them. But logging something without a message does not seem useful to me.

watson:
In the case of languages that can use `type` instead of the `message` (like Java), I think the best thing would be if the UI was updated to fall back to using `type` if `message` wasn't provided in all the places where `message` is expected today.

This would of course require us to make `message` optional. But I wouldn't want to make `type` required either. But I think the JSON schema allows us to make just one of them required without specifying which one, correct?

simitt:
We could do that in the json schema yes. 
Thoughts from the @elastic/apm-ui team regarding @watson 's suggestion?

formgeist:
AFAIK the current errors overview table is showing only error message, which in this case needs to be either or both `error.message` and `error.type` as title? Is the language (agent) consistent, so it's either message or type being used or might that change for each error group/occurrence for a single language (agent)?

watson:
@formgeist In my opinion it's either or. If there's a `message`, show that. If not, show the `type` instead. It should be the same rule for all agents

felixbarny:
In Java, the type would always be present but the message might be absent in some (rather rare) cases.

formgeist:
[...]

PR #1354, opened 2018-09-04
[Intake] require `message` or `type` for error.exception

simitt:
Make either `message` or `type` required for `error.exception`,
without changing the logic for grouping_key.

implements #1297

### DIFF
--- a/model/error/event_test.go
+++ b/model/error/event_test.go
@@ -44,7 +44,8 @@ import (
 )
 
 func baseException() *Exception {
-	return &Exception{Message: "exception message"}
+	msg := "exception message"
+	return &Exception{Message: &msg}
 }
 
 func (e *Exception) withCode(code interface{}) *Exception {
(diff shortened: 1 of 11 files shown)


roncohen:
@elastic/apm-ui heads up, this affects you.

model/error/event_test.go
```diff
@@ -188,17 +189,18 @@ func TestErrorEventDecode(t *testing.T) {
 			},
 		},
 	} {
-		for _, decodeFct := range []func(interface{}, error) (transform.Transformable, error){V1DecodeEvent, V2DecodeEvent} {
+		for idx, decodeFct := range []func(interface{}, error) (transform.Transformable, error){V1DecodeEvent, V2DecodeEvent} {
 			transformable, err := decodeFct(test.input, test.inpErr)
 
 			if test.e != nil {
+				fmt.Println(idx)
```
axw: did you mean to leave that Println in?

simitt: Thanks for catching this, will remove in a follow up.

roncohen:
Sorry, it took so long to review this @simitt.

I just realized we now allow `"exception": {"message": null}` without `type` being specified, e.g.:

```
{ "error": {"id": "cdefab0123456789", "transaction_id": "0000555544446666", "timestamp": "2018-08-09T14:59:05.999Z", "exception": {"message": null}}}
```

Allowing `null` for optional values is in line with how we do optional values, however, we probably shouldn't allow `"message": null` if `type` isn't set. 

I did a quick experiment, and it looks like we can get to what we want by adding something like:
```
"anyOf": [
  {"required": ["message"], "properties": {"message": {"type": "string", "minLength": 1}}},
  {"required": ["type"], "properties": {"type": {"type": "string", "minLength": 1}}}
]
```

Issue #1356, opened 2018-09-05
[v2] Rename "metric" entity to "metricset"


axw:
The metrics API defines a [metric](https://github.com/elastic/apm-server/blob/v2/docs/spec/metrics/metric.json) entity, which is really a set of metrics. I propose that we rename this to "metricset", or something similar. Calling it "metric" might give the impression that it is a singular measurement.

watson:
How about just `metrics` (plural). I try to stick to one-word names if I can in these cases. But maybe I misunderstood the use-case here...

axw:
This is for v2 objects, e.g.

```json
{"metadata":{}}
{"transaction":{}}
{"metrics":{"tags":{}, "samples":{}}}
```

"metrics" as an object name seems a bit odd to me, but better than just "metric". We could flatten the structure so we fold "tags" into each entry of "samples", but that would lead to a lot of repetition.

roncohen:
I like `metricset`

watson:
I'm fine either way. Just wanted to throw in the other option 😃

roncohen:
metricbeat uses 'metricset' as well by the way: https://www.elastic.co/guide/en/beats/metricbeat/current/metricbeat-modules.html

PR #1359, opened 2018-09-06
[v2] rename "metric" entity to "metricset"

roncohen:
Closes https://github.com/elastic/apm-server/issues/1356.


axw:
Thanks @roncohen. Should we also rename metric.json schema to metricset.json? TODO?

Issue #1299, opened 2018-08-20
v2: Investigate which limits/configs are useful for intake v2


roncohen:
We should investigate the behavior and applicability of configurations settings like `MaxUnzippedSize`, `Read/WriteTimeout` and `RateLimit` for v2 and see if there are new limits that make sense to apply instead.

roncohen:
Due to a technical limitation in the Go http.Server,`read_timeout` can only be set server-wide. That means it will be effective for both v1 and v2 endpoints at the same time. Agents have decided to default to a request time of 10s. This is well below our default `read_timeout` of 30s - so we're in the clear regarding `read_timeout`. When we drop the v1 endpoints, we can discuss increasing the default `read_timeout` in the server.

@graphaelli proposed having `max_unzipped_size` per line for v2. I like the idea of a per line limit, but unfortunately we can't repurpose `max_unzipped_size` to mean that for v2, because people might be using v1 and v2 at the same time.

We could introduce a new setting `ndjson_max_event_size` or similar, to protect against clients sending very large events. This could be especially important in the RUM scenario, where anyone can access the API.

`concurrent_requests` was something we introduced primarily to put a bound on the memory usage on the server. With v1, each request body will be read into memory before being processed, so a limit on the number of requests we would do that with, made sense. For v2, we only read up to `batchSize` number of events per request at any one time into memory, before it's handed over to the the libbeat queue. This makes the problem much less serious.  For v2, some agents will keep a http request open all the time. If we blindly applied the same `concurrent_requests` limit with v2, it would limit the number of agents that APM Server could service. 

As mentioned, with v2, we will read up to `batchSize` events from each http request. I suggest we introduce `concurrent_requests_per_ip` to limit the number of concurrent requests from a single IP. We could set it to two, for example. 

`rate_limit` today specifies the number of http requests per second per IP for RUM. That doesn't seem right for v2. What we would probably want is a rate limit on the number of events sent per IP + `concurrent_requests_per_ip` + `ndjson_max_event_size` mentioned above.

to summarize:
*  v1 with current defaults stay the same unchanged
* we introduce these limits for v2: 

**Backend:**
* `ndjson_max_event_size`
* `concurrent_requests_per_ip`

**RUM:**
* `ndjson_max_event_size`
* `concurrent_requests_per_ip`
* `event_rate_limit_per_ip`


`event_rate_limit_per_ip` could perhaps be useful on the backend, with a different limit than for RUM.

simitt:
[...]

PR #1367, opened 2018-09-07
1299 event rate limit per ip

simitt:
[v2] Add rate limit per event.

    * Enabled for RUM endpoints with default value 5000.
    * Throttle read if rate limit is hit while reading the request.
    * Check rate limit in batches of 10.
    * Keep rate limiting for v1 unchanged.

    part of #1299

Benchmark results show roughly 15% less performance with the rate limiter for a big payload (`heavy.ndjson`). Looking into it `WaitN` is the operation slowing down performance. 

```
benchmark                                                   old ns/op     new ns/op     delta
BenchmarkStreamProcessor/errors.ndjson-4                    171791        173410        +0.94%
BenchmarkStreamProcessor/heavy.ndjson-4                     20737919      23456178      +13.11%
BenchmarkStreamProcessor/invalid-event.ndjson-4             49072         62997         +28.38%
BenchmarkStreamProcessor/invalid-json-event.ndjson-4        40350         75697         +87.60%
BenchmarkStreamProcessor/invalid-json-metadata.ndjson-4     5478          6033          +10.13%
BenchmarkStreamProcessor/invalid-metadata-2.ndjson-4        5101          5663          +11.02%
BenchmarkStreamProcessor/invalid-metadata.ndjson-4          9667          10529         +8.92%
BenchmarkStreamProcessor/metrics.ndjson-4                   63356         84340         +33.12%
BenchmarkStreamProcessor/optional-timestamps.ndjson-4       72593         92607         +27.57%
BenchmarkStreamProcessor/spans.ndjson-4                     153029        183702        +20.04%
BenchmarkStreamProcessor/transactions.ndjson-4              202070        212248        +5.04%
BenchmarkStreamProcessor/unrecognized-event.ndjson-4        23905         42183         +76.46%

benchmark                                                   old MB/s     new MB/s     speedup
BenchmarkStreamProcessor/errors.ndjson-4                    25.55        25.32        0.99x
BenchmarkStreamProcessor/heavy.ndjson-4                     19.15        16.93        0.88x
BenchmarkStreamProcessor/invalid-event.ndjson-4             15.61        12.16        0.78x
BenchmarkStreamProcessor/invalid-json-event.ndjson-4        14.55        7.75         0.53x
BenchmarkStreamProcessor/invalid-json-metadata.ndjson-4     85.43        77.57        0.91x
BenchmarkStreamProcessor/invalid-metadata-2.ndjson-4        89.77        80.87        0.90x
BenchmarkStreamProcessor/invalid-metadata.ndjson-4          48.31        44.35        0.92x
BenchmarkStreamProcessor/metrics.ndjson-4                   14.30        10.74        0.75x
BenchmarkStreamProcessor/optional-timestamps.ndjson-4       13.79        10.81        0.78x
BenchmarkStreamProcessor/spans.ndjson-4                     15.95        13.29        0.83x
BenchmarkStreamProcessor/transactions.ndjson-4              20.57        19.58        0.95x
BenchmarkStreamProcessor/unrecognized-event.ndjson-4        16.36        9.27         0.57x

benchmark                                                   old allocs     new allocs     delta
BenchmarkStreamProcessor/errors.ndjson-4                    834            842            +0.96%
BenchmarkStreamProcessor/heavy.ndjson-4                     109574         111281         +1.56%
BenchmarkStreamProcessor/invalid-event.ndjson-4             258            266            +3.10%
BenchmarkStreamProcessor/invalid-json-event.ndjson-4        188            204            +8.51%
BenchmarkStreamProcessor/invalid-json-metadata.ndjson-4     25             25             +0.00%
BenchmarkStreamProcessor/invalid-metadata-2.ndjson-4        20             20             +0.00%
BenchmarkStreamProcessor/invalid-metadata.ndjson-4          47             47             +0.00%
BenchmarkStreamProcessor/metrics.ndjson-4                   267            275            +3.00%
BenchmarkStreamProcessor/optional-timestamps.ndjson-4       358            366            +2.23%
BenchmarkStreamProcessor/spans.ndjson-4                     721            729            +1.11%
BenchmarkStreamProcessor/transactions.ndjson-4              819            827            +0.98%
BenchmarkStreamProcessor/unrecognized-event.ndjson-4        124            132            +6.45%

benchmark                                                   old bytes     new bytes     delta
BenchmarkStreamProcessor/errors.ndjson-4                    45304         45827         +1.15%
BenchmarkStreamProcessor/heavy.ndjson-4                     6720163       6827960       +1.60%
BenchmarkStreamProcessor/invalid-event.ndjson-4             15259         15781         +3.42%
BenchmarkStreamProcessor/invalid-json-event.ndjson-4        12327         13372         +8.48%
BenchmarkStreamProcessor/invalid-json-metadata.ndjson-4     1872          1872          +0.00%
BenchmarkStreamProcessor/invalid-metadata-2.ndjson-4        1712          1712          +0.00%
BenchmarkStreamProcessor/invalid-metadata.ndjson-4          2927          2927          +0.00%
BenchmarkStreamProcessor/metrics.ndjson-4                   17962         18478         +2.87%
BenchmarkStreamProcessor/optional-timestamps.ndjson-4       22466         22983         +2.30%
BenchmarkStreamProcessor/spans.ndjson-4                     39755         40273         +1.30%
BenchmarkStreamProcessor/transactions.ndjson-4              53848         54368         +0.97%
BenchmarkStreamProcessor/unrecognized-event.ndjson-4        9629          10145         +5.36%
```
------------------------
**Outdated Information**:

* Enabled for RUM endpoints with default value 5000.
* Deny request if rate limit is hit when http request is established.
* Throttle read if http request is successfully established, but rate
  limit is hit while reading the request.
* Keep rate limiting for v1 unchanged.

part of elastic/apm-server#1299

For every incoming http request an ip rate limiting check is applied by using [`rate.Allow`](https://godoc.org/golang.org/x/time/rate#Limiter.Allow). This considers the rate limit and the burst size, and returns immediately in case the rate limit is hit for this IP. 
When processing a request, the rate limit is checked before every ndjson line that is read, to ensure the limit per events is not hit. The check is done by using [`rate.Wait`](https://godoc.org/golang.org/x/time/rate#Limiter.Wait) which blocks until the rate limiter allows another hit for this IP or immediately returns an error, if the wait timeout (1s) would be exceeded.  

The rate limit check for incoming requests decreases the allowed events by one, which is not exactly correct. However, when the rate limit is reached, the intention is to use as little resources as possible and return immediately. If the check were removed, requests exceeding the rate limit would be allowed to instantiate an http connection and metadata would be read. 
A consideration to circumvent decreasing the number of allowed events was to use [`rate.Reserve`](https://godoc.org/golang.org/x/time/rate#Limiter.Reserve) followed by [`reservation.Cancel`](https://godoc.org/golang.org/x/time/rate#Reservation.Cancel) to only check if rate limit is hit, but cancel the reservation immediately after. Looking into it, this seems to be a heavy operation and works with best effort but no guarantee that the reservation can be cancelled immediately.


beater/v2_handler.go
```diff
@@ -109,7 +113,8 @@ func (v *v2Handler) Handle(beaterConfig *Config, report publish.Reporter) http.H
 			return
 		}
 
-		res := v.streamProcessor.HandleStream(r.Context(), reqMeta, ndReader, report)
+		ctx := context.WithValue(r.Context(), stream.RateLimitKey, utility.RemoteAddr(r))
```
roncohen: instead of storing the string key (=IP address) here, would it be perhaps be simpler to store the appropriate rate limit object in the context?

rate/limit.go
```diff
@@ -0,0 +1,72 @@
...
+	lru "github.com/hashicorp/golang-lru"
+	r "golang.org/x/time/rate"
+)
+
+const (
+	cacheSize       = 1000
+	burstMultiplier = 2
+)
+
+type LimitHandler interface {
```
roncohen: could we call it a `LimitManager` or similar instead of `LimitHandler`? `handler` gets confusing with all the `http.Handler`s we have.

beater/route_config.go
```diff
@@ -72,16 +73,27 @@ var V1Routes = map[string]v1Route{
...
-	},
-}
+var (
+	v2BackendRoute = v2Route{
+		routeType{
+			v2backendHandler,
+			systemMetadataDecoder,
+			func(*Config) transform.Config { return transform.Config{} },
+		},
+		func(c *Config) rate.LimitHandler { return c.rateLimitHandler },
```
roncohen: Do we need this extra function when we have `v2backendHandler` and `v2rumHandler`? seems like it would be simpler to handle this under the individual handlers (`v2backendHandler` and `v2rumHandler`) and then i think we could get rid of `route.rateLimitHandler(beaterConfig)` [here](https://github.com/elastic/apm-server/pull/1367/files#diff-9829d1c9b6183c7ba00ca10afb6376b9R160)

beater/v2_handler.go
simitt: As the rate limiter is cached within an LRU cache the object could get kicked out of the cache somewhen during the process.

roncohen:
Good work @simitt. I added a few questions around simplifying stuff. I also see that you took out the "up front" `AllowN` call, which is probably fine. As we discussed, the reason to have it in there in the first place was performance, but it might not matter. Would be interesting to have some benchmarks on that. 

If it turns out to make a difference, we could possibly simplify the "up front" `AllowN` code by asking for `batchSize` events up front, then simply count down the number as we read events. Whenever we reach `0`, we ask for `batchSize` more events. If last batch doesn't quite have `batchSize` events, we'll ask for more events from the rate limiter than necessary, but that might not be a big deal.

beater/v2_handler.go
roncohen: could you elaborate on why this is a problem?

roncohen:
Another note is that RUM data is very "bursty" in nature. The user loads a page generating a bunch of events, then after a couple of seconds, the user loads another page, or a route change. We could consider setting the per-second rate limit quite low and then the burst multiplier to something large.

For example, 
* limit per second: 30
* burst multiplier: 10

processor/stream/stream_processor.go
```diff
@@ -229,6 +239,25 @@ func (s *StreamProcessor) HandleStream(ctx context.Context, meta map[string]inte
...
+
+		//Throttle read in case rate limit is hit, by using
+		//blocking wait until rate limiter allows to read next batchsize event.
+		//Consider burst and timeout indicating how long to wait for rate limiter.
+		//Return early if rate limiter returns error.
+		limiter, limiterOk := ctx.Value(RateLimiterKey).(*rate.Limiter)
+		if limiterOk && limiter != nil {
+			ctxT, cancel := context.WithTimeout(ctx, time.Second)
+			defer cancel()
+			err = limiter.WaitN(ctxT, batchSize)
```
roncohen: this decreases the rate limit "bucket" by `batchSize` even if we read fewer items for every batch. 
We could consider doing something like:

```
readsAllowed := 0
for {
    if readAllowed < batchSize {
        missing := batchSize - readAllowed
        limiter.WaitN(ctxT, missing)
        readsAllowed += missing
    }
    n, transformables, done := s.readBatch(batchSize, jsonReader, res)
    readsAllowed = readsAllowed - n
    [...]
}
```

I also think you should move: 
`limiter, limiterOk := ctx.Value(RateLimiterKey).(*rate.Limiter)`
out of the loop. No reason to get it from the context every time. 

finally, `defer` defers the execution of `cancel()` to when the function returns, not until the loop does another iteration. We'll need to call `cancel()` manually here.

beater/v2_handler.go
```diff
@@ -108,8 +116,22 @@ func (v *v2Handler) Handle(beaterConfig *Config, report publish.Reporter) http.H
...
 		}
+		ctx := r.Context()
+
+		// fetch the rate limiter from the cache, if a cache is given
+		var limiter *rate.Limiter
+		if v.cache != nil && beaterConfig.RumConfig.EventRate != nil {
+			key := utility.RemoteAddr(r)
+			if !v.cache.Contains(key) {
+				v.cache.Add(key, rate.NewLimiter(rate.Limit(beaterConfig.RumConfig.EventRate.Limit),
+					beaterConfig.RumConfig.EventRate.Limit*2))
```
roncohen: `2` should be a constant. I also think 2 is [perhaps not right](https://github.com/elastic/apm-server/pull/1367#issuecomment-420235843) for RUM. WDYT?

beater/v2_handler.go
```diff
@@ -108,8 +116,22 @@ func (v *v2Handler) Handle(beaterConfig *Config, report publish.Reporter) http.H
...
+			}
+			entry, _ := v.cache.Get(key)
+			limiter, _ = entry.(*rate.Limiter)
```
roncohen: I don't think you need this cast if you're just storing it in the context in the next line.

beater/v2_handler.go
```diff
@@ -53,6 +59,8 @@ func (v *v2Handler) statusCode(sr *stream.Result) int {
 			code = http.StatusServiceUnavailable
 		case stream.ServerErrType:
 			code = http.StatusInternalServerError
+		case stream.RateLimitErrType:
+			code = http.StatusTooManyRequests
```
roncohen: Queue is full and Rate limit errors now give the same response code, but for different problems. I remember we picked 503 for Queue is full for v1 because we said it requires operator intervention. I'm leaning towards going back to that for v2 also, but no strong feelings.

beater/route_config.go
```diff
@@ -79,37 +81,52 @@ var V1Routes = map[string]v1Route{
...
-		func(*Config) transform.Config { return transform.Config{} },
-	},
-}
+var (
+	v2BackendRoute = v2Route{
+		routeType{
+			v2backendHandler,
+			systemMetadataDecoder,
+			func(*Config) transform.Config { return transform.Config{} },
+			func(c *Config) *lru.Cache { return nil },
```
roncohen: this function is defined 5 times. Could we allow `nil` here to make things simpler? alternatively, we could make a function, call it `noCache` or similar that returns nil and reuse it

beater/v2_handler.go
```diff
@@ -108,8 +116,22 @@ func (v *v2Handler) Handle(beaterConfig *Config, report publish.Reporter) http.H
 			v.sendResponse(logger, w, &sr)
 			return
 		}
+		ctx := r.Context()
+
+		// fetch the rate limiter from the cache, if a cache is given
+		var limiter *rate.Limiter
```
axw: `limiter` isn't used at this scope, so I think you can get rid of this var

beater/v2_handler.go
```diff
@@ -108,8 +116,22 @@ func (v *v2Handler) Handle(beaterConfig *Config, report publish.Reporter) http.H
...
+		if v.cache != nil && beaterConfig.RumConfig.EventRate != nil {
+			key := utility.RemoteAddr(r)
+			if !v.cache.Contains(key) {
```
axw: Looks like there's a [TOCTOU](https://en.wikipedia.org/wiki/Time_of_check_to_time_of_use) bug here. It's technically possible, if unlikely, for the entry to be evicted after the `if` block but before the subsequent `cache.Get`, which would then return nil.

I'd suggest this instead:

```go
limiter, ok := v.cache.Get(key)
if !ok {
    limiter = rate.NewLimiter(...)
    v.cache.Add(key, limiter)
}
ctx = context.WithValue(r.Context(), stream.RateLimiterKey, limiter)
```

processor/stream/stream_processor.go
```diff
@@ -36,6 +39,10 @@ import (
 	"github.com/elastic/apm-server/validation"
 )
 
+const (
+	RateLimiterKey = "rateLimiter"
```
axw: It's not likely to be an issue, but it's best practice to use an unexported struct type as a context key, to avoid clashing with keys from other code. So this would be something like:

```go
type rateLimiterKey struct {}
var RateLimiterKey = rateLimiterKey{}
```

or perhaps in order to avoid an exported var:

```go
type rateLimiterKey struct {}

func ContextWithRateLimiter(ctx context.Context, limiter *rate.Limiter) context.Context {
    return context.WithValue(ctx, rateLimiterKey{}, limiter)
}
```

processor/stream/stream_processor.go
axw: could you instead move the rate limiting into `readBatch`, before reach call to StreamReader.Read? or better yet, wrap StreamReader such that its Reads are rate-limited? e.g.

```diff
diff --git a/processor/stream/stream_processor.go b/processor/stream/stream_processor.go
index 6182d0cc..d284f6a2 100644
--- a/processor/stream/stream_processor.go
+++ b/processor/stream/stream_processor.go
@@ -23,6 +23,7 @@ import (
        "io"
 
        "github.com/santhosh-tekuri/jsonschema"
+       "golang.org/x/time/rate"
 
        "github.com/elastic/apm-server/decoder"
        er "github.com/elastic/apm-server/model/error"
@@ -74,6 +75,19 @@ func (s *srErrorWrapper) Read() (map[string]interface{}, error) {
        return v, err
 }
 
+type rateLimitStreamReader struct {
+       StreamReader
+       ctx context.Context
+       rl  *rate.Limiter
+}
+
+func (r *rateLimitStreamReader) Read() (map[string]interface{}, error) {
+       if err := r.Wait(r.ctx); err != nil {
+               return nil, err
+       }
+       return r.StreamReader.Read()
+}
+
 type StreamProcessor struct {
        Tconfig transform.Config
 }
@@ -222,6 +236,16 @@ func (s *StreamProcessor) HandleStream(ctx context.Context, meta map[string]inte
                return res
        }
 
+       if rl, _ := ctx.Value(rateLimitKey{}); rl != nil {
+               // A rate-limiter has been provided; wrap the stream
+               // reader so that reads are rate-limited.
+               jsonReader = &rateLimitStreamReader{
+                       StreamReader: jsonReader,
+                       ctx:          ctx,
+                       rl:           rl,
+               }
+       }
+
        tctx := &transform.Context{
                RequestTime: utility.RequestTime(ctx),
                Config:      s.Tconfig,
```

axw: By the way, you might also want to take a look at https://github.com/juju/ratelimit as an alternative rate limiter. It's a little more flexible, and I think probably more efficient.

e.g. to rate-limit, but wait for no more than 1 second, with juju/ratelimit you would do this:

```go
var r *ratelimit.Bucket = ...
if !r.WaitMaxDuration(batchSize, time.Second) {
    res.Add(&Error{Type: RateLimitErrType, Message: "rate limit exceeded"})
    return res
}
```

beater/v2_handler.go
axw: Also, looks like there's a attack possible here, due to use of LRU. Say a DDoS is taking place, then the LRU is filled up with many IPs. new IPs coming in will be given a new rate limiter with a full capacity/burst.

I think what you want to do is check if an item was evicted, and reuse that rate limiter, so you'll probably get an empty bucket.

roncohen: Ah. Nice catch

simitt: @axw I agree that the usage of an LRU cache introduces a possible attack surface. However, I brought this up in discussions with @roncohen earlier this week, where he argued for keeping the LRU cache. The alternative would be to use a cache with expiration date, that is either unbounded (could lead to endless memory usage) or limited to a certain number of entries, in which case reaching the number of entries would also be problematic. 

Regarding your proposed solution to avoid the TOCTOU, I agree that your suggestion gets rid of this problem. Both of our solutions could potentially lead to race conditions though. I'll think about this some more.

processor/stream/stream_processor.go
simitt: @axw in my first draft I did call `Wait` on every iteration. I was actually thinking about just wrapping it around the reader. However, we would have to check for every single event, which decreases performance. After discussing with @roncohen he suggested to check for `batchSize` events at once and accept to 'waste' some requests for the last batch that is read. 

@roncohen I don't agree with your suggestion [here](https://github.com/elastic/apm-server/pull/1367/files/2f0de7b8e53bd44cfe9fa09d2b04824109038483#r217454248). 
The only reason for not reading `batchsize` events in a batch would be if there was an early-return-error or if it is the last batch that had less than batchsize entries. In both cases, we would return without an additional iteration. In any other case `readBatch` should have read `batchsize` entries. So the case that we waste `batchsize-n` hits always only applies to the last batch. This would not be solved by your suggestion.

roncohen: ah, true 👍

beater/v2_handler.go
simitt: In my opinion we should not change the `queue is full` from a server side to a client side error without having a good reason for that. I have overlooked that in your [original PR](https://github.com/elastic/apm-server/pull/1324/files#diff-ce6c26e750f0443c52c10a357e08e82aR52). Nevertheless, I suggest that the decision which status code to return for a full queue has to be made completely independent of the error response returned for a rate limit. I suggest to stick with a server side status code for `v2`, as a full queue either reflects an issue with the storage or with the internal memory queue being misconfigured. If it were an issue with the client sending too many requests, then we have to cover this with a rate limiter. 
503 or 507 are probably the two status codes we should discuss for a full queue (but separate of this PR).

beater/route_config.go
simitt: I'll change it to a nil function 👍

beater/v2_handler.go
roncohen: As far as I understand `event_rate.limit:1000` means a single IP can send 1000 events per second (= 2 page loads per second) continuously and burst up to 5000 (= 10 page loads per second). 500 spans is the max that backend agents allow by default. The avg. number of spans is much lower. The 2 page loads per second is then more like a minimum, because most transactions will have much less than 500 spans. That seems excessive to me.

Is it correct that if the burst limit is reached, our rate limiter would reject the request instead of delaying the reading?

processor/stream/stream_processor.go
roncohen: We've used a string wrapped in a context key type here: https://github.com/elastic/apm-server/blob/master/beater/handlers.go#L311 and a `ContextWith` function here: https://github.com/elastic/apm-server/blob/v2/utility/request_time.go#L29

would probably make sense to align the code base on how we do it. I have no strong feelings of which approach as long as it's the same.

beater/v2_handler.go
roncohen: I think there's a couple of different discussions going on here. There's the question about when to call `Wait`/`WaitN` and then there's the question of whether we should get the rate limiter from the LRU when we want to call `Wait`/`WaitN` or we should get the rate limiter from the LRU when the request comes in. 

I [argued](https://github.com/elastic/apm-server/pull/1367#issuecomment-420216773) that if there's a significant performance impact from calling `Wait` for every event, a potential way to solve it would be to consult the rate limiter for each batch instead. If there's no significant performance impact, i think it's much simpler to consult the rate limiter for each event.

I asked whether it would be [simpler](https://github.com/elastic/apm-server/pull/1367#discussion_r216604275) to get the rate limiter from the LRU and store it in the context instead of storing the LRU key in the context. If it's not simpler, I'm fine with getting the rate limiter from the LRU every time.

roncohen: @axw could you elaborate on this?: 
> I think what you want to do is check if an item was evicted, and reuse that rate limiter, so you'll probably get an empty bucket.

I believe @axw suggested offline if we set the LRU size to the same size as the overall `concurrent_requests` limit that we'll implement for v2, that would mitigate the attack.

axw: > I believe @axw suggested offline if we set the LRU size to the same size as the overall concurrent_requests limit that we'll implement for v2, that would mitigate the attack.

That just ensures that the bucket for the current IP won't get evicted mid-way through a request while retaining an upper bound on the number of active rate limiter objects.

>> I think what you want to do is check if an item was evicted, and reuse that rate limiter, so you'll probably get an empty bucket.

Sorry, a bit cryptic. And this is all a bit hypothetical, I haven't actually implemented this before.

Say your LRU has size N, and an attacker has 2*N IPs at their disposal. They cycle through the first N IPs until they get rate-limited, and then cycle through the next N, and then switch back to the first N and so on.

The reason that's an issue is because after N IPs have entered the LRU, then the N+1'th IP will evict one, and as it is implemented, a fresh limiter created with a full bucket, so you can burst immediately. What you could do instead is effectively have a pool of limiters by storing the evicted limiter using lru.NewWithEvict, and then reusing it when lru.Add returns true indicating another entry was evicted.

In pseudo-code:

```
var limiterPool sync.Pool
onEvicted := func(key interface{}, value interface{}) {
    limiterPool.Put(*value.(**rate.Limiter))
}
cache, err := lru.NewWithEvict(N, onEvicted)
if err {...}
...
var limiter *rate.Limiter
value, ok := cache.Get(key)
if ok {
    limiter = *value.(**rate.Limiter)
} else {
    if _, evicted := cache.Add(key, &limiter); evicted {
        limiter = limiterPool.Get().(*rate.Limiter)
    } else {
        limiter = rate.NewLimiter()
    }
}
```

simitt: @roncohen thanks for clarification!

The two of us had an offline discussion, after your PR comments, where we discussed a couple of the comments and these were my take aways:
(1) use `batchsize` for checks if it performs better than checking for every request.
(2) fetch limiter once from cache and not for every event, because you didn't see an advantage for fetching it for every event but found the code to be more complicated. 
(3) don't use another form of cache, but stick with LRU cache. Make the limit configurable and document the downside in the config file. 
(4) get rid of checking whether or not the request itself should be allowed, as an additional limit and check will be applied for that. 

add (1): I did performance checks and `Wait` performs poorly so I switched to check in batches. However, also doing performance checks for `Reserve` I am going to switch back, as this behaves better.

add (2):  I might have misunderstood you there, as I thought you definitely prefer to not fetch the limiter per event but per request. We discussed how a request with many events get more hits in the LRU cache than a request with less events if the limiter is fetched for every event. So this is not only abut code simplicity but about a different semantic.

add (3): The suggestion to align the cache key limit with the amount of concurrent requests helps to ensure a key doesn't get evicted while fetching. It cannot solve the problem that an attack could bypass the whole rate limiting if requests are sent from multiple different ips, where `number_ips > cache_keys`.

ad (4): Introducing an additional `requests_per_ip_per_second` is fine with me, although I tried to avoid this additional config option.

simitt: I just saw @axw 's last comment. @roncohen that is exactly the attack vector that we discussed offline. Since it is an LRU cache you can bypass with requests from too many different IPs. We also discussed this could be especially a problem with RUM agents on high traffic websites. 
I'll look into @axw 's suggestion with the callback.

beater/v2_handler.go
simitt: A request is rejected if the burst limit is reached and the rate limit would still be exceeded after a defined waiting time (1 second). 
This means that if only one request is sent, that exceeds the limit, the first `n=limit` events are processed, and then the processing is throttled. One event after the other is checked, and if the limit is already reached, the event is added to the burst queue. If the burst queue is full and the waiting time until this event could be accepted is larger than a defined timeout (1 second) then the processing fails with an ip rate limiting error. 
If several requests from one IP are sent concurrently then `limit + burst` requests can be processed in a throttled way. If more requests are sent then they are either rejected or some events get processed in a throttled way and then they get rejected. 

Having a large burst size makes sense if we think that there won't be a lot of concurrent requests and some of the requests will be exceptionally large, so we are fine with processing them slowly.

processor/stream/stream_processor.go
simitt: @axw update on `Wait` vs. `Reserve`: Using `Reserve` leads to *always* reserve a token if the amount of checked tokens is < limit.burst, which is always the case when checking per event. See [`ReserveN` always calls `reserveN` with `InfDuration`](https://github.com/golang/time/blob/master/rate/rate.go#L211), and therefore [`reserveN` always updating state](https://github.com/golang/time/blob/master/rate/rate.go#L330).
There is no way of adding a timeout to `Reserve` or `ReserveN`. Using `Wait` would only update, if the `deadline` is met. This means I cannot use `Reserve` for throttling the read. As soon as one request hits the limit, within a second all following requests would fail too, and it cannot relax anymore, so all requests would return with a rate limit hit error.

beater/v2_handler.go
simitt: @axw I looked into your suggestion to reuse evicted limiters and I am not sure this is the way to go. My main issue with it is that this could lead to a lot of slow down for requests from IP addresses that actually do respect the rate limiting, while potentially allowing a larger amount of events from other IPs than intended. 

I also see an issue with re-using an old rate limiter for one request, as the next request for the same IP will still use a new rate limiter

axw: > I also see an issue with re-using an old rate limiter for one request, as the next request for the same IP will still use a new rate limiter

That sounds odd, can I see a diff somewhere?

simitt: I was referring to your pseudo code: 
```
var limiter *rate.Limiter
if _, evicted := cache.Add(key, &limiter); evicted {
        limiter = limiterPool.Get().(*rate.Limiter)
    }
```
The limiter fetched from the `limiterPool` is not added to the pool again. So for the next request another limiter is used. But as cache keys do not expire here, it would be another evicted limiter and not a new one.

simitt:
As discussions show that there are several points to address implementing rate limiting for v2 RUM requests, this summary aims to clarify what is left for discussion.

**Objectives** of rate limiting the number of `requests` and the number of `events` per second:
1. Ensure server cannot be brought in undefined state
2. Protect storage by avoiding unlimited write operations to configured storage. 
3. Spread resources between requests to avoid overwhelming the server with malicious requests, so valid ones cannot be processed.

**Possible Config Option Combinations:**
1. `events_per_ip`: Deny incoming requests for current IP, throttle processing of established requests in case limit is hit. The advantage is that we can achieve the same as with the other combinations but only have to use and maintain one config option. Another advantage is that this is compatible with different agent behaviour (few requests with lots of events vs. many small requests).

2. `events_per_ip`+ `req_per_ip`: Deny incoming requests if `req_per_ip` is hit, throttle processing of events for established requests in case `events_per_ip_per_sec` is hit. The config options seem redundant. If we have a separate `req_per_ip` limiter, why do we need to limit the amount of events over all requests? Which advantage does the combination of these two config options give?

3. `events_per_req`+`req_per_ip`: Deny incoming requests if `req_per_ip` is hit, throttle processing per request in case `events_per_req` is hit for the specific request. This would allow to establish a new event limiter per request, without caching it. We would still need to cache the limiter for the `req_per_ip`. The advantage is that this could easily be combined with an external rate limiter (e.g. nginx, haproxy) while being rather simple to implement. The downside is that it is hard to configure if agents differ a lot in behaviour. 

**Possible Cache Approaches:**

- _cache with unlimited size_ with expiring keys: unbounded memory usage, can bring server in undefined state. 

- _LRU cache_ (similar to current handling): Keys could get evicted without the rate limiter applying, as soon as `requests_from_unique_ips >= 2*cache_size`. This would basically mean unlimited write operations to the configured storage. (Limited ofc by other generic config options that apply, eg. concurrent_req, size_per_event, ..)

- _fixed size cache with expiring keys_: Key expiration could be chosen in a way to ensure possibly queued hits are processed before expiration ( (limit+burst_limit) / limit ), to ensure key is not expired while hits are queued. If cache keys limit is reached, no new requests can be accepted. This could potentially lead to blocking the whole cache with malicious requests and deny all valid requests. However, the regularly configured limit could not be exceeded as the rate limiter would apply. Once cache is full the expiration time would automatically mean a cool down phase, until new requests are accepted, so throttling would be guaranteed. 

- _LRU cache + eviction handling_ [suggested by @axw](https://github.com/elastic/apm-server/pull/1367#discussion_r218026232): The suggestion here was to reuse the evicted limiter, as soon as a new limiter would lead to an eviction. As soon as a cache entry gets evicted and is reused for the other entry, requests from two unrelated IPs are related for this request. One could argue that in a normal use case the cache capacity should not be reached, and if reached, the rate limit should usually not be hit by regular requests. When under attack of malicious requests, this would overall throttle down the request processing, so it would ensure that some upper limits to write operations. It would also slow down request processing for valid requests. 




**Open**:
[ ] Decide which config options to use: I think option (1) would be the most user friendly one, and cannot see any advantage in using option (2). 
[ ] Decide for cache handling: The standard LRU cache handling is not a good fit for this. The cache with unlimited size can bring the server in an undefined state and stop from working, that is also not great. The best solutions seems to be the LRU cache + eviction handling. 
[ ] Add suggestion to use HAProxy, Nginx, .. in front of APM Server(s) for rate limiting in docs. 
[ ] Add metrics information about rate limiter capacity / hits to monitoring tab.


@roncohen , @axw WDYT?

axw:
> [ ] Decide which config options to use: I think option (1) would be the most user friendly one, and cannot see any advantage in using option (2).
> [ ] Decide for cache handling: The standard LRU cache handling is not a good fit for this. The cache with unlimited size can bring the server in an undefined state and stop from working, that is also not great. The best solutions seems to be the LRU cache + eviction handling.
> [ ] Add suggestion to use HAProxy, Nginx, .. in front of APM Server(s) for rate limiting in docs.
> [ ] Add metrics information about rate limiter capacity / hits to monitoring tab.

That all sounds good to me.

roncohen:
thanks for the summary. Generally this sounds good to me as well 👍 

> Add suggestion to use HAProxy, Nginx, .. in front of APM Server(s) for rate limiting in docs.

Your summary didn't mention why this is important. Could you elaborate?

The general Elastic approach is to work well without external dependencies and we should try to reflect this in our product as well as in our documentation. As an example, we're working towards removing the requirement to have a load balancer in front of the APM Server and instead do it in agents.

simitt:
The in-memory cache of the server is limited, so I think we can only reach a best effort rate limiting within the server. Having a load balancer in the agents is great for a regular use case, it doesn't help against any kind of attacks though. For the RUM endpoint there is no possibility to secure the endpoint against requests with a secret token. For these cases having an additional load balancer and rate limiter up front could be very helpful.

simitt:
jenkins, retest this please.

simitt:
jenkins, retest this please.

simitt:
jenkins, retest this please.

_meta/beat.yml
```diff
@@ -49,9 +49,27 @@ apm-server:
...
+      # Defines the maximum amount of events allowed to be sent to the APM Server v2 RUM
+      # endpoint per ip per second. There is an internal burst queue to avoid rejecting short termed spikes. The burst 
+      # queue is set to limit*5.
+      # Defaults to 1000.
+      #limit: 1000
+      # Defines the amount of unique IPs for which a rate limiter is collected.
+      # Ensure a large enough cache size, as the rate limiting cannot properly be applied,
+      # in case the amount of unique IPs exceeds the amount of cache keys configured.
+      # Defaults to 1000.
+      #cache_keys: 1000
```
roncohen: i like this generally, but the warning here could sound like rate limiting just stops working or something when the LRU size is exceeded. I think we can either:
* explain in detail how it works or what happens when the number of IPs exceed the cache size. Something like:
```
# When an IP gets evicted from the LRU, its rate limit is re-used for the new IP
# to ensure that the rate limit is effective even when the concurrent number of unique clients exceed the LRU size.
# Sites with many concurrent clients should consider increasing this limit.
```
* or remove the warning and just say something like:
```
# An LRU cache is used to keep a rate limit per IP for the most recently seen IPs. 
# This setting defines the number of unique IPs that can will be tracked in the cache.
# Sites with many concurrent clients should consider increasing this limit.
```

roncohen: Should we consider naming it `lru_size`? `cache_keys` here is a bit ambiguous. No strong feelings

beater/cache.go
```diff
@@ -0,0 +1,89 @@
...
+	"sync"
+
+	lru "github.com/hashicorp/golang-lru"
+	"github.com/pkg/errors"
+	"golang.org/x/time/rate"
+)
+
+const burstMultiplier = 5
+
+var limiterPool sync.Pool
```
roncohen: this pool should probably be part of `rlCache`. Otherwise, you can create two `rlCache` with different limits who will use the same pool. That means the pool can contain rate limiters with different limits and you might end up calling `getRateLimiter` and getting a rate limit with the wrong limit.

beater/cache.go
```diff
@@ -0,0 +1,89 @@
...
+		return nil, false
+	}
+
+	if limiter, ok := getLimiter(); ok {
+		return limiter
+	}
+
+	rlc.mu.Lock()
+	defer rlc.mu.Unlock()
+	if limiter, ok := getLimiter(); ok {
```
roncohen: I took a look at the LRU implementation. The LRU we use here is basically a wrapper around 	"github.com/hashicorp/golang-lru/simplelru" with locks added. In other words, there's a lock being taken for each call to `Get` and `Add`. In the last part of this method we're adding our own lock on top. I seems like we might as well as use `simplelru` and use our own lock?

I haven't thought this through completely, but instead of the get/lock/get/add/unlock pattern here, could you use  `ContainsOrAdd` by:
1. calling `Get`, if the IP is there the recentness gets updated and the limiter is returned
2. it was not there, call `ContainsOrAdd`, if it's there now, we'll return

beater/cache.go
```diff
@@ -0,0 +1,89 @@
+// Licensed to Elasticsearch B.V. under one or more contributor
```
roncohen: this file provides rate limiting and not caching, so I'd rename it to something related to rate limiting.

beater/config.go
```diff
@@ -68,6 +73,12 @@ type rumConfig struct {
 	beatVersion string
 }
 
+type eventRate struct {
+	Limit       int `config:"limit"`
+	CacheKeys   int `config:"cache_keys"`
+	limiterPool sync.Pool
```
roncohen: seems like a leftover

_meta/beat.yml
```diff
@@ -49,9 +49,27 @@ apm-server:
...
     #rate_limit: 10
 
+    #-- v2 RUM endpoint
+
+    #event_rate:
+      # Defines the maximum amount of events allowed to be sent to the APM Server v2 RUM
+      # endpoint per ip per second. There is an internal burst queue to avoid rejecting short termed spikes. The burst 
+      # queue is set to limit*5.
+      # Defaults to 1000.
+      #limit: 1000
```
roncohen: sorry to keep discussing defaults, but i think 1000 events per second is too high. For reference, the previous discussion is here: https://github.com/elastic/apm-server/pull/1367#discussion_r218004645

following the previous discussing, what about a limit of 500 and a burst of 5? And I think even that is high. It means each browser can make a new request with the maximum number of spans per second - and most websites takes way longer than 1s to load.

_meta/beat.yml
simitt: I prefer to remove the warning, as I would not want to go too much into implementation details in the config yml description.

processor/stream/stream_processor.go
```diff
@@ -227,9 +260,11 @@ func (s *StreamProcessor) HandleStream(ctx context.Context, meta map[string]inte
 		Config:      s.Tconfig,
 		Metadata:    *metadata,
 	}
+	rl := rateLimiterFromContext(ctx)
 
 	for {
-		transformables, done := s.readBatch(batchSize, jsonReader, res)
+
+		transformables, done := s.readBatch(ctx, rl, batchSize, jsonReader, res)
```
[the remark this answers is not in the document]
simitt: You pointed out in a previous comment that we can fetch the `rate_limiter` once and not for every batch. I agree that this is more performant. The `context` is needed for the `WaitN` though. Thus I suggest to stick with the current solution for this.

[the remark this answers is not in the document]
roncohen: to be clear, I'm just suggesting to move

> rl := rateLimiterFromContext(ctx)

inside `readBatch` and removing the `rl` argument to `readBatch`. 

What we discussed was only accessing the LRU once, which would still be the case. I don't have strong feelings about this, but since the rate limiter is already in the context and we need to pass in the context anyway, i don't see the point of taking it out here and passing it in as a separate argument.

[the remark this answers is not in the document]
simitt: I was referring to https://github.com/elastic/apm-server/pull/1367#discussion_r217454248: 
> I also think you should move:
limiter, limiterOk := ctx.Value(RateLimiterKey).(*rate.Limiter)
out of the loop. No reason to get it from the context every time.

This way we save a fetching and typecasting the limiter from the context for every batch.

simitt:
jenkins, retest this please.

processor/stream/stream_processor.go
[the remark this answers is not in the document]
simitt: Me neither, but for clarification, _out of the loop_ in the above example translates to what I've done. If I move it to `readBatch` then it would be _inside the loop_ again.

_meta/beat.yml
```diff
@@ -49,9 +49,24 @@ apm-server:
...
     # Rate limit per second and IP address for requests sent to the RUM endpoint.
     #rate_limit: 10
 
+    #-- v2 RUM endpoint
+
+    #event_rate:
+      # Defines the maximum amount of events allowed to be sent to the APM Server v2 RUM
+      # endpoint per ip per second. Defaults to 300.
+      #limit: 300
+      # An LRU cache is used to keep a rate limit per IP for the most recently seen IPs.
```
roncohen: tiny nitpick: would be nice to have an empty line separating these

processor/stream/stream_processor.go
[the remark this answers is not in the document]
roncohen: I don't think it's the same situation. The example I commented on previously requires that a line is moved up above the `for {` line - and nothing else. We were only using the rate limiter outside readBatch and not passing it in.

I think a good argument to continue passing in the rate limit is that it makes it explicit to `readBatch`. If we pass the rate limit in through the context only it's implicit and difficult for a caller to know that the context can contain a rate limit that `readBatch` will use.

testdata/intake-v2/only-metadata.ndjson
```diff
@@ -0,0 +1 @@
+{"metadata": {"process": {"ppid": 6789, "pid": 1234, "argv": ["node", "server.js"], "title": "node"}, "system": {"platform": "darwin", "hostname": "prod1.example.com", "architecture": "x64"}, "service": {"name": "1234_service-12a3", "language": {"version": "8", "name": "ecmascript"}, "agent": {"version": "3.14.0", "name": "elastic-node"}, "environment": "staging", "framework": {"version": "1.2.3", "name": "Express"}, "version": "5.1.3", "runtime": {"version": "8.0.0", "name": "node"}}}}
```
roncohen: can't find where this file is used

beater/rl_cache.go
```diff
@@ -0,0 +1,88 @@
...
+	cache *simplelru.LRU
+	limit int
+
+	mu             sync.Mutex //guards limiter in cache
+	evictedLimiter *rate.Limiter
+}
+
+func NewRlCache(size, rateLimit int) (*rlCache, error) {
+	if size <= 0 || rateLimit < 0 {
+		return nil, errors.New("Cache initialization went wrong.")
```
roncohen: suggestion for more informative error message:
`cache initialization: size and rateLimit must be greater than zero`

beater/rl_cache.go
```diff
@@ -0,0 +1,88 @@
...
+
+	rlc := rlCache{limit: rateLimit}
+
+	var onEvicted = func(_ interface{}, value interface{}) {
+		rlc.evictedLimiter = (*value.(**rate.Limiter))
+	}
+
+	if c, err := simplelru.NewLRU(size, simplelru.EvictCallback(onEvicted)); err != nil {
+		return nil, err
+	} else {
```
roncohen: don't need the else clause here

beater/route_config.go
```diff
@@ -79,37 +79,60 @@ var V1Routes = map[string]v1Route{
...
+			func(*Config) transform.Config { return transform.Config{} },
+			nilCache,
+		},
+	}
+	v2RumRoute = v2Route{
+		routeType{
+			v2rumHandler,
+			userMetaDataDecoder,
+			rumTransformConfig,
+			func(c *Config) *rlCache {
```
roncohen: sorry, I've changed my mind regarding this. It was generally the idea that the `routeType`s would contain all the stuff that is common for v1 and v2 while `v2Route` allows special config for v2. I think i forgot that when we last discussed this. I think the rlCache should be part of `v2Route` instead because rlCache is only used in v2.

beater/v2_handler_test.go
```diff
@@ -110,25 +111,16 @@ func TestRequestIntegration(t *testing.T) {
...
-			req.Header.Add("Content-Type", "application/x-ndjson")
-
-			w := httptest.NewRecorder()
-
-			c := defaultConfig("7.0.0")
-			report := func(context.Context, publish.PendingReq) error {
-				return test.reportingErr
-			}
-			handler := (&v2BackendRoute).Handler(c, report)
-
```
roncohen: nice to DRY this

processor/stream/stream_processor_test.go
```diff
@@ -156,3 +158,34 @@ func TestIntegration(t *testing.T) {
...
+	for _, test := range []struct {
+		name string
+		lim  *rate.Limiter
+		hit  int
+	}{
+		{name: "NoLimiter"},
+		{name: "LimiterDenyAll", lim: rate.NewLimiter(rate.Limit(0), 2)},
+		{name: "LimiterAllowAll", lim: rate.NewLimiter(rate.Limit(40), 40*5)},
+		{name: "LimiterAllowWithWait", lim: rate.NewLimiter(rate.Limit(10), 10*2), hit: 10},
+		{name: "LimiterForbiddenHit", lim: rate.NewLimiter(rate.Limit(7), 7*2), hit: 10},
```
roncohen: what's the difference between `LimiterForbiddenHit` and `LimiterForbidden`?

tiny naming nit pick: Deny vs. Forbidden. I thought this name was meant to mean that it's a test for _hit_ting the forbidden limit. Maybe we should go with `Deny` here too to keep it clearly separate from "Method Forbidden".

"Hit" here means that some part of the limit has already been used as far as i understand. Maybe simply "limitUsed" or similar would make that clearer?

processor/stream/stream_processor.go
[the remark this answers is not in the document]
simitt: It is obvious that it is not completely the same situation as the code changed. I am not arguing for passing a parameter to a method or not. I am arguing for fetching and typecasting the rate limiter _once_ from the context and not for every batch.

simitt:
@roncohen I've addressed your comments in another round of changes. Are you good with merging the PR?

roncohen:
thanks @simitt. Yes looks good! :shipit: 🚢

Issue #1381, opened 2018-09-12
Update Elastic apm go agent


graphaelli:
May need to update `go-sysinfo` first.

axw:
might need to get rethink how (or if) spans are created with v2, I think each request may now yield many calls into the beats pipeline

graphaelli:
Great point @axw, created #1385 for that. Will continue to use this issue for updated apm-agent-go to the latest release instead.

graphaelli:
closed by #1387

PR #1386, opened 2018-09-14
Update beats framework to 037a4f2

graphaelli:
mostly to pull in an update `go-sysinfo` for #1381.


graphaelli:
jenkins, run package test please

PR #1387, opened 2018-09-14
 update apm-agent-go to v0.5.0

graphaelli:
depends on #1386

PR #1389, opened 2018-09-17
enable sending instrumentation to another apm-server

### LINKED ISSUE #1388: Send instrumentation to another apm-server
Allow an instrumented apm-server to send data to another apm-server, instead of only to itself.

graphaelli:
Adds `apm-server.instrumentation` configuration directives: `hosts` and `secret_token`.  `hosts` can only contain 1 element at the moment but is an array in anticipation of this restriction being lifted following elastic/apm-agent-go#200.

TODO:
- [x] merge elastic/apm-server#1387
- [x] docs/config
- [x] tests

closes elastic/apm-server#1388

### DIFF
--- a/beater/beater.go
+++ b/beater/beater.go
@@ -188,12 +188,6 @@ func initTracer(info beat.Info, config *Config, logger *logp.Logger) (*elasticap
 	} else {
 		os.Setenv("ELASTIC_APM_ACTIVE", "true")
 		logger.Infof("self instrumentation is enabled")
-
-		if config.SelfInstrumentation.Hosts != nil {
-			os.Setenv("ELASTIC_APM_SECRET_TOKEN", config.SelfInstrumentation.SecretToken)
-			os.Setenv("ELASTIC_APM_SERVER_URL", config.SelfInstrumentation.Hosts[0])
-			logger.Infof("self instrumentation directed to %s", config.SelfInstrumentation.Hosts[0])
-		}
 	}
 
 	tracer, err := elasticapm.NewTracer(info.Beat, info.Version)
@@ -201,6 +195,15 @@ func initTracer(info beat.Info, config *Config, logger *logp.Logger) (*elasticap
 		return nil, nil, err
 	}
 	if config.SelfInstrumentation.isEnabled() {
+		if config.SelfInstrumentation.Hosts != nil {
+			t, err := transport.NewHTTPTransport(config.SelfInstrumentation.Hosts[0], config.SelfInstrumentation.SecretToken)
+			if err != nil {
+				return nil, nil, err
+			}
+			tracer.Transport = t
+			logger.Infof("self instrumentation directed to %s", config.SelfInstrumentation.Hosts[0])
+		}
+
 		if config.SelfInstrumentation.Environment != nil {
 			tracer.Service.Environment = *config.SelfInstrumentation.Environment
 		}
--- a/beater/beater.go
+++ b/beater/beater.go
@@ -209,6 +209,7 @@ func initTracer(info beat.Info, config *Config, logger *logp.Logger) (*elasticap
 	if config.SelfInstrumentation.Hosts != nil {
 		t, err := transport.NewHTTPTransport(config.SelfInstrumentation.Hosts[0], config.SelfInstrumentation.SecretToken)
 		if err != nil {
+			tracer.Close()
 			return nil, nil, err
 		}
 		tracer.Transport = t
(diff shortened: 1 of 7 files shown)


beater/beater.go
```diff
@@ -186,6 +188,12 @@ func initTracer(info beat.Info, config *Config, logger *logp.Logger) (*elasticap
 	} else {
 		os.Setenv("ELASTIC_APM_ACTIVE", "true")
 		logger.Infof("self instrumentation is enabled")
+
+		if config.SelfInstrumentation.Hosts != nil {
+			os.Setenv("ELASTIC_APM_SECRET_TOKEN", config.SelfInstrumentation.SecretToken)
+			os.Setenv("ELASTIC_APM_SERVER_URL", config.SelfInstrumentation.Hosts[0])
```
axw: These env vars only take effect in a call to `transport.NewHTTPTransport`. `elasticapm.NewTracer` uses `transport.Default`, which is initialised at package-init time.

You would either have to call `transport.InitDefault` after setting these, or alternatively (my preference) call `transport.NewHTTPTransport` with the arguments directly.

beater/beater.go
```diff
@@ -193,12 +195,25 @@ func initTracer(info beat.Info, config *Config, logger *logp.Logger) (*elasticap
...
+			logger.Infof("self instrumentation directed to %s", config.SelfInstrumentation.Hosts[0])
+		}
+
 		if config.SelfInstrumentation.Environment != nil {
 			tracer.Service.Environment = *config.SelfInstrumentation.Environment
 		}
 		tracer.SetLogger(logp.NewLogger("tracing"))
 	}
 
+	if config.SelfInstrumentation.isEnabled() && config.SelfInstrumentation.Hosts != nil {
```
axw: Should this instead be `if !config.SelfInstrumentation.isEnabled() || config.SelfInstrumentation.Hosts != nil {`? i.e. only set the pipe transport if self-instrumentation is enabled and there are no hosts specified.

beater/beater.go
```diff
@@ -193,12 +195,25 @@ func initTracer(info beat.Info, config *Config, logger *logp.Logger) (*elasticap
 		return nil, nil, err
 	}
 	if config.SelfInstrumentation.isEnabled() {
+		if config.SelfInstrumentation.Hosts != nil {
+			t, err := transport.NewHTTPTransport(config.SelfInstrumentation.Hosts[0], config.SelfInstrumentation.SecretToken)
```
axw: probably worth a comment here stating that Hosts will only ever have one element, and that extending beyond that depends on https://github.com/elastic/apm-agent-go/issues/200

beater/beater.go
```diff
@@ -192,11 +194,27 @@ func initTracer(info beat.Info, config *Config, logger *logp.Logger) (*elasticap
...
 	}
-	if config.SelfInstrumentation.isEnabled() {
-		if config.SelfInstrumentation.Environment != nil {
-			tracer.Service.Environment = *config.SelfInstrumentation.Environment
+	// tracing disabled, setup complete
+	if !config.SelfInstrumentation.isEnabled() {
+		return tracer, nil, nil
+	}
+
+	if config.SelfInstrumentation.Environment != nil {
```
graphaelli: @axw any reason to do this before returning early, when instrumentation is disabled?

axw: nope, not unless/until we make the "enabled" configuration dynamic

beater/beater.go
```diff
@@ -192,11 +194,27 @@ func initTracer(info beat.Info, config *Config, logger *logp.Logger) (*elasticap
...
+		tracer.Service.Environment = *config.SelfInstrumentation.Environment
+	}
+	tracer.SetLogger(logp.NewLogger("tracing"))
+
+	// tracing destined for external host
+	// only first host used until https://github.com/elastic/apm-agent-go/issues/200
+	if config.SelfInstrumentation.Hosts != nil {
+		t, err := transport.NewHTTPTransport(config.SelfInstrumentation.Hosts[0], config.SelfInstrumentation.SecretToken)
+		if err != nil {
+			return nil, nil, err
```
simitt: In https://github.com/elastic/apm-server/pull/1389/files#diff-110fdf0113b0dcfcdca7c9352c1b4c4fR228 `tracer.Close()` is called in an error case. You don't call it here, but return `nil` instead of the tracer, so I think it would never get closed in this case.

graphaelli: Great call, fixed

PR #1398, opened 2018-09-19
[v2] Require dt ids for errors if one is given.

simitt:
Change json spec to require `transaction_id`, `trace_id` and `parent_id` as soon as one of them is given.

Enforcing this conditional requirement on JSON spec level, decreases performance:
```
benchmark                                                   old ns/op     new ns/op     delta
BenchmarkStreamProcessor/errors.ndjson-4                    147158        163343        +11.00%

benchmark                                                   old MB/s     new MB/s     speedup
BenchmarkStreamProcessor/errors.ndjson-4                    29.83        26.88        0.90x

benchmark                                                   old allocs     new allocs     delta
BenchmarkStreamProcessor/errors.ndjson-4                    833            909            +9.12%

benchmark                                                   old bytes     new bytes     delta
BenchmarkStreamProcessor/errors.ndjson-4                    45311         47782         +5.45%
```

The alternative is to move the check to the go code in the `model/error/event.go`. This would mean basically no overhead. However, we decided to use JSON schema and allow agent devs to run their tests against it, so I didn't simply want to move the check out of the schema.


felixbarny:
I'm ok with both

roncohen:
I'm also OK with both. The following should work as well, but it didn't change the benchmark much: 

```diff
                     "maxLength": 1024
                 }
             },
+            "required": ["id"],
             "oneOf": [
-                { "required": ["id", "trace_id", "transaction_id", "parent_id"] },
+                { "required": ["trace_id", "transaction_id", "parent_id"] },
                 { "allOf": [
-                    { "required": ["id"] },
                     { "not": { "required": ["trace_id"] }},
                     { "not": { "required": ["transaction_id"] }},
                     { "not": { "required": ["parent_id"] }}
```

docs/spec/errors/v2_error.json
```diff
@@ -13,22 +13,30 @@
                     "maxLength": 1024
                 },
                 "trace_id": {
-                    "description": "Hex encoded 128 random bits ID of the correlated trace.", 
+                    "description": "Hex encoded 128 random bits ID of the correlated trace. Must be present if transaction_id and parent_id are set.", 
                     "type": ["string", "null"],
                     "maxLength": 1024
                 },
                 "transaction_id": {
```
felixbarny: Do we really need the `transaction_id` here? If possible, I'd like to avoid that. We currently use the `transaction_id` on spans to make it easier for the UI to display a transaction view. I don't think the errors also need a `transaction_id`. The problem with adding the `transaction_id` is that it makes the case really hard where a span is started after the transaction has ended and makes in-process context propagation more awkward.

@axw might also have a opinion about that

felixbarny: I'm currently experimenting with adding a `transactionId` field to the [`TraceContext`](https://github.com/elastic/apm-agent-java/blob/master/apm-agent-core/src/main/java/co/elastic/apm/impl/transaction/TraceContext.java) class. That would ease the handling quite a bit.

simitt:
@roncohen there are a view different approaches on how to build the requirements, but all cases I checked showed roughly the same performance. Thanks for trying it out though.

docs/spec/errors/v2_error.json
simitt: Having the `transaction_id` might be useful if one wants to show all errors for a transaction. Do we plan this @sqren ?
I am ok with making it optional, I am relying on the knowledge of the @elastic/apm-agent-devs regarding which attributes can be provided and which not.

roncohen:
@simitt cool. Is there a reason the "id" separate like in the example I showed? From my perspective it seems simpler to understand

Qard:
Do the `{ "not": { "required": ... } }` bits make it _optional_ or required that it _not be present_?

docs/spec/errors/v2_error.json
sqren: > Having the transaction_id might be useful if one wants to show all errors for a transaction

Yes, we have talked linking from errors -> transactions, and vice versa.  Eg. it would probably be useful to show the error in the timeline somehow. If the error does not have a `transaction_id` I'm not sure how we can easily query for it.

Btw. Will `transaction_id` -> `transaction.id`, and `trace_id` -> `trace.id` to align with https://github.com/elastic/apm-server/pull/1393?

felixbarny: Turned out adding the transaction id to TraceContext works quite well: https://github.com/elastic/apm-agent-java/pull/217

sqren: Neat!

simitt:
After digging into this some more, I updated to use `if/then` statements, for better readability and they also perform a lot better: 
```
benchmark                                                   old ns/op     new ns/op     delta
BenchmarkStreamProcessor/errors.ndjson-4                    147429        155261        +5.31%

benchmark                                                   old MB/s     new MB/s     speedup
BenchmarkStreamProcessor/errors.ndjson-4                    29.78        28.27        0.95x

benchmark                                                   old allocs     new allocs     delta
BenchmarkStreamProcessor/errors.ndjson-4                    833            869            +4.32%

benchmark                                                   old bytes     new bytes     delta
BenchmarkStreamProcessor/errors.ndjson-4                    45307         46520         +2.68%
`
``

Issue #1355, opened 2018-09-05
[v2] adapt application metrics


simitt:
In `v2` error responses are handled differently than in `v1`. 
We have to adapt the collected metrics accordingly and ensure the semantics don't change and/or metrics are shown correctly in the monitoring tab. 

see https://github.com/elastic/apm-server/blob/master/beater/handlers.go#L87

PR #1403, opened 2018-09-21
[v2] Increase metrics counter in v2 responses.

simitt:
implements elastic/apm-server#1355

### DIFF
--- a/beater/v2_handler.go
+++ b/beater/v2_handler.go
@@ -21,53 +21,65 @@ import (
 	"encoding/json"
 	"net/http"
 
+	"github.com/elastic/beats/libbeat/monitoring"
+
 	"github.com/elastic/apm-server/processor/stream"
 	"github.com/elastic/apm-server/publish"
 
 	"github.com/elastic/beats/libbeat/logp"
 
-	"github.com/pkg/errors"
-
 	"github.com/elastic/apm-server/decoder"
 )
 
-var (
-	errInvalidMetadataFormat = errors.New("invalid metadata format")
-)
+const MethodNotAllowedErrType stream.StreamError = 100
 
 type v2Handler struct {
 	requestDecoder  decoder.ReqDecoder
 	streamProcessor *stream.StreamProcessor
 }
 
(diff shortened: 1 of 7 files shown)


beater/v2_handler.go
```diff
@@ -21,53 +21,65 @@ import (
...
 
-	"github.com/pkg/errors"
-
 	"github.com/elastic/apm-server/decoder"
 )
 
-var (
-	errInvalidMetadataFormat = errors.New("invalid metadata format")
-)
+const MethodNotAllowedErrType stream.StreamError = 100
```
graphaelli: any reason this isn't at the bottom of the list in `processor/stream/result.go` ?

beater/v2_handler.go
```diff
@@ -21,53 +21,65 @@ import (
...
 func (v *v2Handler) sendResponse(logger *logp.Logger, w http.ResponseWriter, sr *stream.Result) {
-	statusCode := v.statusCode(sr)
+	statusCode, counter := v.statusCode(sr)
+	responseCounter.Inc()
+	counter.Inc()
 
 	w.WriteHeader(statusCode)
 	if statusCode != http.StatusAccepted {
+		responseErrors.Inc()
 		// this singals to the client that we're closing the connection
```
graphaelli: might as well fix this typo in `singals`

beater/v2_handler.go
simitt: `processor/stream` has no knowledge about the used protocol, it is just reading and processing a stream. The `MethodNotAllowedErrType` reflects that the wrong http method was used.
But I guess the err name is generic enough to move it inside `processor` as it doesn't say anything about the protocol.

simitt:
@graphaelli I think it is fair to only count the error corresponding to the returned error code for `error.response`. 
As I share the uncertainty about generally only reporting this error, I introduced `apm-server.processor.stream.errors.*` and `apm-server.processor.stream.accepted` counters, counting all processed events.

simitt:
@graphaelli retriggering jenkins solved the ci issue. It was some issue with `go vet` and not a flaky test, so tests should be fine.

PR #1408, opened 2018-09-25
Update beats framework to fa956ec

graphaelli:


graphaelli:
jenkins, run package tests please

Issue #1340, opened 2018-08-30
v2: Require either `span.start` or `span.timestamp`


simitt:
A span's `timestamp` is an absolute value when the span has been captured, and `start` the offset relative to the transactions start. 
At the moment `span.start` is required, but `span.timestamp` is not required, as the RUM agent cannot set a `span.timestamp`. 

Except for the RUM agent all other agents do set the `span.timestamp` and thus `span.start` seems redundant there. Thus the suggestion is to require one of both fields to be set. 

Is this ok with all @elastic/apm-agent-devs ?
* [x] go 
* [x] java
* [ ] nodejs
* [x] python
* [x] ruby 
* [x] RUM

And is it ok with @elastic/apm-ui ?
* [ ] @sqren 
* [ ] @formgeist

jahtalab:
A couple of considerations:
* Currently the apm server sets all missing timestamps (transaction and it's spans) to the time it received the request.
* In the RUM agent `span.start` is set instead.
* because of these limitations, in the v2 spec, for the RUM agent, we agreed on sending a transaction and it's spans in the same request (so that all of them get the same timestamp).

axw:
We can't really do this, turns out. I misunderstood, span's timestamp is meant to be exactly the same as the transaction timestamp. It should probably be called transaction_timestamp. Per @watson:

> AFAIK the timestamp is only used in the query when retrieving the spans and transactions from Elasticsearch. When laying out the spans in the waterfall the offset and duration is used.
> 
> By using the same timestamp, the agents can re-use these timestamps across all spans belonging to the same transaction. This keeps resource usage at a minimum.
> 
> It also means that a time-restricted query will always return all spans belonging to a transactions instead of only the ones falling within the time window.
> 
> Additionally, in the v1 protocol the spans didn't have timestamps, and the server would pick the timestamp from the transaction and ad it to the span before storing the span in ES. So this basically mimics what the server is already doing in v1.

axw:
Discussion still ongoing. A couple of counterpoints:
 - storing the transaction timestamp in the span doc the violates ECS definition of @timestamp
 - even if we have the same timestamp for a span and its transaction, we'll still need to consider transactions crossing time boundaries for a distributed trace

We'll take this to the weekly meeting to flesh it out.

felixbarny:
From a chat conversation with @sqren 

@felixbarny
> > AFAIK the timestamp is only used in the query when retrieving the spans and transactions from Elasticsearch.
[...]

PR #1413, opened 2018-10-01
[v2] accept unix epoch timestamps in microseconds

roncohen:
* makes "start" optional if "timestamp" is given
* adds timestamp.us (long) to span, transaction and error documents
* if "timestamp" is not given, it is set to the sum of req-time + start

Fixes https://github.com/elastic/apm-server/issues/1340


_meta/fields.common.yml
```diff
@@ -14,6 +14,16 @@
       type: keyword
       description: Processor event.
 
+    - name: timestamp
+      type: group
+      description:
+      fields:
+        - name: us
+          type: long
+          count: 1
```
roncohen: I wasn't able to figure out what this "count: 1" actually does.

roncohen:
after a discussion with @simitt and @sqren I'm going to remove the new `timestamp.us` from v1 data.

roncohen:
I've made the changes necessary to keep `timestamp.us` out of v1 data in c988dc9.

It adds significant complexity, in particular in the tests. Given the added complexity, i think it's worth re-considering if we shouldn't "just" add `timestamp.us` also to the v1 data.

The UI will probably not be able to take advantage of this added field for v1 data, because the UI needs to take into account old data that doesn't have the field anyway - so it's easier for it to just assume there no `timestamp.us` field for v1 data.

sqren:
If this is not worth it, I'm fine with leaving `timestamp.us` in, and just not use it when handling v1 data.

model/error/event.go
```diff
@@ -381,3 +380,15 @@ func addStacktraceCounter(st m.Stacktrace) {
...
 }
+
+type v2Event struct {
+	*Event
+}
+
+func (e *v2Event) Transform(tctx *transform.Context) []beat.Event {
+	events := e.Event.Transform(tctx)
+	for _, e := range events {
+		utility.Add(e.Fields, "timestamp", utility.TimeAsMicros(e.Timestamp))
```
graphaelli: This isn't sitting well with me, though I see how we got here.  Have you considered adding a member to `Event` to differentiate between v1 and v2 and only add this timestamp object if v2, inside of `Event.Transform()`?

roncohen: much better, thanks!

model/metricset/event_test.go
```diff
@@ -130,6 +131,19 @@ func TestDecode(t *testing.T) {
 		}
 	}
 }
+func TestDecodeV1(t *testing.T) {
```
roncohen: this is parameterizes `testDecode` while in "span" the decode tests are simply copied. I'm fine either way, but we should consider aligning them.

roncohen:
following feedback from @graphaelli things are a bit simpler and I'm OK to move forward with the current functionality.

docs/spec/timestamp_epoch.json
```diff
@@ -0,0 +1,12 @@
+{
+    "$id": "doc/spec/timestamp_epoch.json",
+    "title": "Timestamp Epoch",
+    "description": "Object with 'timestamp' property.",
+    "type": ["object"],
+    "properties": {  
+        "timestamp": {
+            "description": "Recorded time of the span, UTC based and formatted as microseconds since Unix epoch",
```
simitt: This should say `event` rather than `span`.

model/error/event.go
simitt: Have you considered creating two different event attributes for the two timestamps? This would allow setting the proper timestamp in the v1/v2 decoding methods. The transformation method could then only transform what is set, without any additional knowledge.

The reason I am suggesting this is that we could keep the whole v1/v2 separation in the decode method, as we do so far, which is also the reason that so far no separate v2 transform tests have been necessary.

roncohen: I think I see your point. If i understand correctly, you're suggesting to do something like:

```diff
type Event struct {
	Name       string
	Type       string
	Start      *float64
	Duration   float64
	Context    common.MapStr
	Stacktrace m.Stacktrace

-	Timestamp     time.Time
	TransactionId string

	// new in v2
	HexId    string
	ParentId string
	TraceId  string
+	TimestampV2     time.Time

	// deprecated in v2
	Id     *int64
	Parent *int64
+	TimestampV1     time.Time
}
```

```
func V1DecodeEvent(input interface{}, err error) (transform.Transformable, error) {
	...
	e.TimestampV1 = decoder.TimeRFC3339(raw, "timestamp")
	...
}


func V2DecodeEvent(input interface{}, err error) (transform.Transformable, error) {
	...
	e.TimestampV2 = decoder.TimeEpochMicro(raw, "timestamp")
	...
}
```

For spans, it seems to me that the transform method needs to know if it is dealing with a v1 span or a v2 span. Part of the new functionality is to set the timestamp to be reqTime + start if no timestamp is given. For v1 we're just setting it to reqTime. Looking at the `span` transform method, if none of the timestamps are set (timestamp is optional in both v1 and v2) how will we know if we should apply the v2 logic or not?

Besides the problem above, what i like about this current solution is that the decode methods "decode" to a model that is shared by v1 and v2 in the sense there's one field to denote the timestamp, regardless of which format it came in.

Having two timestamp fields for the sake of deciding the final output seems less clean to me, compared to decoding the different input representations with the same semantics into the same field as it is currently. For example, if there was some functionality that required us to use the timestamp in some logic, if we have two timestamps we would need to check them both even though the semantically mean the same thing.

model/metricset/event_test.go
roncohen: So you're OK with the current changes?

model/error/event.go
simitt: Appreciate the explanation, agree that it is cleaner this way for spans.

model/metricset/event_test.go
simitt: yes fine with me

felixbarny:
Note for agent devs: the field which holds the epoch micros is still called `timestamp` in the intake API: https://github.com/elastic/apm-server/blob/v2/docs/spec/timestamp_epoch.json

PR #1415, opened 2018-10-02
[DOCS] Bumped version for 6.4.2

debadair:


debadair:
@graphaelli @dedemorton We need to bump the version for this morning's release.

PR #1424, opened 2018-10-07
[v2] updated changelog

roncohen:
Update change log in preparation for merging v2 into master.

v2 is a whole new API that we're introducing, so including the changes we've made to v2 separately from the introduction of v2 doesn't seem right to me. Users of the next version of APM Server will just see a new protocol added and shouldn't mind the fact that we've been doing it in iterations between the releases.

So I've gone ahead and removed the individual change log entries that we did for v2 and instead linked to the two meta issues: one for v2 and one for DT. This way, people can still see all the individual work that went into it, but as a change log entry it just shows up as the introduction of v2 with distributed tracing support.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -12,18 +12,10 @@ https://github.com/elastic/apm-server/compare/6.4\...master[View commits]
 [float]
 === Added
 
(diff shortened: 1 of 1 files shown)


roncohen:
thanks @simitt !

Issue #1298, opened 2018-08-20
v2: Investigate "full queue" behavior


roncohen:
Reporting timeout is 1s at the moment., which is probably too high. We should investigate the behavior of v2 when the queue is full.

roncohen:
The reason we can't just block for a second every time data comes in, is that eventually the agents "write" call on the socket will start timing out. This sort of timeout can also happen for other reasons e.g. APM Server isn't working properly, or there's a faulty proxy in between. In other words, we can't just say that whenever a write in the agent times out, it means the queue in APM Server is full. 

When APM Server queue is full we could simply write "Write error: timeout" to the agent logs, but it would be a much better user experience if agents are able to write a more meaningful message to their logs.

After discussions with @urso, we can't get rid of the 1s timeout without making significant changes to libbeat. For the first iteration of v2, i suggest we should do the simple thing and keep the 1s timeout in the APM Server, but return a ERR_FULL_QUEUE response to the agent immediately (see https://github.com/elastic/apm-server/issues/1240) if we've been unable to deliver the events to the internal beats queue within the 1s timeout.

WDYT @elastic/apm-agent-devs ?

felixbarny:
Sounds good. But it would be necessary that all HTTP clients used by the agents support receiving a response before the request has been fully written. If it doesn't I'd expect to get an `IOException` `Connection reset by peer`, which can be handled as well, but without having access to the response. Maybe the TCP connection can also not be reused in that case, needs investigation.

What is an easy way to reproduce that situation so I can test that? Reporting lots of transactions to the APM server and not starting Elasticsearch?

simitt:
fyi - @graphaelli already [created a PR](https://github.com/elastic/apm-integration-testing/pull/167) in the apm-integration-testing so that you can use the testing repo for starting the `v2` branch of the server.

felixbarny:
To keep track which agents have already been tested:

* [ ] go 
* [ ] java
* [ ] nodejs
* [ ] python
* [ ] ruby 
* [ ] RUM

watson:
In the Node.js agent the risk of errors and how we handle them are the same for both api-v1 or v2. So there's no difference between the two versions for us. We're all good 👍

urso:
Proposed libbeat changes: elastic/beats#8080

felixbarny:
@watson how did you test the TCP half-close when the APM server queue is full? I tried shutting down ES and sending lots of transactions but then I got read timeouts from the APM server.

[...]

PR #1425, opened 2018-10-07
Merge V2 into master

roncohen:
I've rebased v2 on master. It should apply cleanly. I pushed it to a separate branch in order to avoid force pushing to `v2` for now. 

TODO: 
- [x] go through the commits messages here and make sure they are all prefixed with `v2: ` or `[v2]` 
- [x] https://github.com/elastic/apm-server/pull/1424 updated changelog
- [x] https://github.com/elastic/apm-server/issues/1298#issuecomment-418429722
- [ ] "rebase and merge" and not "squash and merge".


simitt:
Actually I just realised that the changelogs are not properly rebased. Everything we show in [Changelog#Added](https://github.com/elastic/apm-server/blob/v2/CHANGELOG.asciidoc) should be moved to [changelogs/head.asciidoc](https://github.com/elastic/apm-server/blob/master/changelogs/head.asciidoc), and everything else in CHANGELOG should be removed as it is added to separate files in master, see [changelogs/*](https://github.com/elastic/apm-server/tree/master/changelogs).

roncohen:
@simitt thanks, updated!

PR #1430, opened 2018-10-11
Replace UUID lib

### LINKED ISSUE #1427: Replace uuid library
like elastic/beats#8077

graphaelli:
Switch out uuid library due to satori/go.uuid#73.

closes elastic/apm-server#1427

Please be sure to create a merge commit to simplify backporting.


beater/handlers.go
```diff
@@ -103,6 +103,14 @@ var (
 		code:    http.StatusAccepted,
 		counter: counter("response.valid.accepted"),
 	}
+	internalErrorCounter  = counter("response.errors.internal")
```
simitt: In v2 we introduced a ['response.errors.other'](https://github.com/elastic/apm-server/blob/v2/beater/common_handlers.go#L74). It should be the same as `internalErrorCounter`, @graphaelli @roncohen whoever merges first, please ensure to align those two counters.

Issue #1435, opened 2018-10-15
Add additional fields to `span.context.http`. 


simitt:
```
{
  "context": {
    "http": {
      "status_code": 201,
      "method": "PUT"
    }
  }
}
```

`status_code`: optional, long
`method`: optional, keyword

watson:
I might have misunderstood, but I think we agreed to add `context.http.url` as well right?

cc @jahtalab @eyalkoren

simitt:
[span.context.http.url](https://github.com/elastic/apm-server/blob/master/docs/spec/transactions/span.json#L39) is already available.

watson:
Oh, didn't know. Carry on 😄

PR #1438, opened 2018-10-16
Add indexed fields to span.context.http.

simitt:
Add `status_code` and `method` to `span.context.http` and index them. The new fields are added to v1 and v2 as they are an extension of the current schema. 

There wasn't a big discussion about whether or not to index the fields, but they both seem reasonable to index.

The changes do not automatically show up in the UI, so this would require some work from the @elastic/apm-ui  team. 

implements https://github.com/elastic/apm-server/issues/1435


docs/fields.asciidoc
```diff
@@ -93,6 +93,27 @@ A flat mapping of user-defined tags with string values.
...
+
+
+--
+
+*`context.http.status_code`*::
++
+--
+type: long
+
+The status code of the http request.
```
jalvz: nitpick: this is actually the HTTP response.

i am also seeing that we write indifferently HTTP and http, maybe you can use this change to grep and replace to all caps everywhere.

simitt:
@elastic/apm-server any concerns about indexing those fields? Otherwise I go forward with it.

simitt:
After another discussion with @eyalkoren I went ahead with only indexing `context.http.status_code` but not `context.http.method`, as this doesn't seem to be necessary for now.

simitt:
jenkins, retest this please

PR #1439, opened 2018-10-16
add a v1 deprecation warning log

jalvz:


jalvz:
ill do this against master instead

PR #1443, opened 2018-10-16
Add missing changes to master

roncohen:
These two didn't make it when i rebased v2.


simitt:
I think you now have to run the `make fmt` on this PR as the fix is already in master. (maybe?)

PR #1448, opened 2018-10-17
Add a non-test file to package_tests to keep `go test` happier

jalvz:


simitt:
Can you do the same for the `stream/package_tests` ?

PR #1449, opened 2018-10-17
try to see what is going on

jalvz:
WIP/not merge - trying to figure out if we have a flaky test


jalvz:
conclusion: not flaky - we could improve a bit error messaging, but will leave that for another occasion

PR #1455, opened 2018-10-19
v2: Change ctx used for rate limiter

simitt:
Use simple background context for rate limiting instead of request context.


simitt:
Actually I am going to close this PR. I thought it'd be a good idea to not reuse the request's context but create a new one for the rate limiter. However, the current approach returns a copy of the requests context, so it is safe to use it as is.

PR #1458, opened 2018-10-22
Add missing nil check

jalvz:
`getRateLimiter` can return `nil`, but is not checked at the call point https://github.com/elastic/apm-server/compare/master...jalvz:rl-nil-fix?expand=1#diff-ce6c26e750f0443c52c10a357e08e82aR118

`NewRlCache` prevents that scenario now, but we could break it if eg. we change its implementation - this PR makes it more explicit.

I don't know if this has been discussed before, @simitt lmk if I got something wrong.

### DIFF
--- a/beater/rl_cache.go
+++ b/beater/rl_cache.go
@@ -63,10 +63,10 @@ func NewRlCache(size, rateLimit int) (*rlCache, error) {
 	return &rlc, nil
 }
 
-func (rlc *rlCache) getRateLimiter(key string) *rate.Limiter {
+func (rlc *rlCache) getRateLimiter(key string) (*rate.Limiter, bool) {
 	// fetch the rate limiter from the cache, if a cache is given
-	if rlc.cache == nil || rlc.limit == -1 {
-		return nil
+	if rlc == nil || rlc.cache == nil || rlc.limit == -1 {
+		return nil, false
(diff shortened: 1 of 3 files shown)


jalvz:
yeah, that would work too!

i think the value is that when you see a function returning (`*something`, `bool`), it is obvious for the caller that it has to check whether is safe to use `something` or not.

if a function just returns `*something` then is easier to forget a `nil` check and you have to go the function source code to know if it might be `nil` or not.

simitt:
Actually I would have read that differently: `bool` indicates whether something went wrong. There might still be valid use cases where no rate limiter is returned, but nothing went wrong (depending on the implementation ofc).

jalvz:
So would you do something different here?

eg: if we say that `-1` disables rate limit, we could do `if !ok { return true // always allow }` but that is not how it works now if i understand correctly.

simitt:
Not sure I understand the example. Currently you can disable the rate limiter only by setting it to a very large number. That's mainly because this was also how it was implemented for v1, and that is how the rate limiter from the used library works. I don't see the relation to having an additional `bool` value or not.

However, we did discuss this offline and agreed on both options are ok to move forward, so @jalvz will pick whichever solution feels best to him.

PR #1459, opened 2018-10-22
V1 reusable rate limiter

jalvz:


beater/common_handlers.go
```diff
@@ -304,19 +302,15 @@ func killSwitchHandler(killSwitch bool, h http.Handler) http.Handler {
...
 
 const (
-	rateLimitCacheSize       = 1000
-	rateLimitBurstMultiplier = 2
+	rateLimitCacheSize = 1000
 )
 
 func ipRateLimitHandler(rateLimit int, h http.Handler) http.Handler {
-	cache, _ := lru.New(rateLimitCacheSize)
+	rlc, _ := NewRlCache(rateLimitCacheSize, rateLimit, rateLimitBurstMultiplier)
```
simitt: Why did you remove the `rateLimitBurstMultiplier = 2` const and where does `rateLimitBurstMultiplier` now come from?

beater/route_config.go
```diff
@@ -228,7 +228,7 @@ func (v v2Route) Handler(url string, c *Config, report publish.Reporter) http.Ha
 	}
 
 	if url == V2RumURL {
-		if rlc, err := NewRlCache(c.RumConfig.EventRate.LruSize, c.RumConfig.EventRate.Limit); err == nil {
+		if rlc, err := NewRlCache(c.RumConfig.EventRate.LruSize, c.RumConfig.EventRate.Limit, burstMultiplier); err == nil {
```
simitt: I think you should now remove https://github.com/elastic/apm-server/pull/1459/files#diff-48b2aa9b71bf27a37b9122608aa3dfb1R28, and instead create a `v2BurstMultiplier` const here.

beater/common_handlers.go
jalvz: sorry, that was a bad commit - fixed now

PR #1460, opened 2018-10-22
Update beats framework to 2a25eae

graphaelli:


graphaelli:
nice catch, I'm going to redo this to grab the latest commit too

PR #1466, opened 2018-10-23
[APM-CI] APM Server pipeline initial version

kuisathaverat:
image


kuisathaverat:
@simitt @jalvz @graphaelli, I would like you to take a look, and check if everything is correct and make sense. 

These are the stages and my proposal to run them on a PR, master, or both

* Checkout - all
* Intake - master
* linux build - all
* windows build -all
* Test - all
* System and Environment Tests - all
* windows test - all
* Benchmarking - master
* Integration test - master
* Hey APM test - master
* Documentation - master
* Release (package) - manual
* Check kibana Obj. Updated - master

Integration tests are disabled until we make some changes.

simitt:
Here are my thoughts regarding the stages and when to run them:

- Intake - should also be run on PRs. It checks that all files created by `make update` are committed and that everything is in proper `go fmt`. This is equally important for PRs.
- Hey-APM tests - those are tests for a separate repo [hey-apm](https://github.com/elastic/hey-apm), that the server team built and manages. This should not be part of the build pipeline for APM Server but be its own build pipeline. 
- Benchmarking - if I understand correctly then atm benchmarks do not _succeed_  or _fail_ depending on whether benchmarks improved or got worse, but only write the results to some benchmarking file. What is happening with this file, are the results also pushed to some ES instance so we can have a look over time? I think running benchmarks on master is fine for now, for the future it would be nice to reiterate so we see regressions before we merge PRs.
- Integration tests - I am not convinced that we should run Integration Tests for every push to master for APM Server. I'd like to have those tests separated as the root cause for failing tests can also easily be some agent issues. Failing tests in the server pipeline should only indicate an issue with the server though.
- Documentation - if we generally only run it on master we do need some way to trigger the tests for PRs as otherwise ci would not be usable for doc update PRs.
- package - so far those tests were run triggered by a successful master build. I'd like to keep this, as it indicates issues when merging libbeat updates. We also currently have the possibility to manually trigger the tests, which should still be available for PRs where we know it could be broken. 

On a general notice - for everything that is only run on `master` it would be nice to be able to trigger this build stage manually, to see issues before merging PRs.

kuisathaverat:
@simitt 

>Hey-APM tests - those are tests for a separate repo hey-apm, that the server team built and manages. This should not be part of the build pipeline for APM Server but be its own build pipeline.

It will have his own pipeline, if you think that is not useful to run it in this pipeline we removed it, if at some point we want it we can add it again.

>Benchmarking - if I understand correctly then atm benchmarks do not succeed or fail depending on whether benchmarks improved or got worse, but only write the results to some benchmarking file. What is happening with this file, are the results also pushed to some ES instance so we can have a look over time? I think running benchmarks on master is fine for now, for the future it would be nice to reiterate so we see regressions before we merge PRs.

It executes `make bench` and send the benchmarks to The same ES do the agents, to an index named benchmark-server. Right now does not compare results, we can make it, it is a matter of seeing how to check the values in ES.

>Integration tests - I am not convinced that we should run Integration Tests for every push to master for APM Server. I'd like to have those tests separated as the root cause for failing tests can also easily be some agent issues. Failing tests in the server pipeline should only indicate an issue with the server though.

Same as Hey APM. Integration tests will have a single version run (server vs agents) and an Axis version (a few server versions vs a few agent versions vs ELK versions), similar to what we have now.

>On a general notice - for everything that is only run on master it would be nice to be able to trigger this build stage manually, to see issues before merging PRs.

I have created a check parameter per stage to be able to enable-disable stages, I will add a new one to "run as master" to passthrough the `master branch condition` and be able to run steps that normally run only on the master branch, on any PR

Another step that now I have commented, it is one that runs `script/jenkins/update-beats.sh`, Is it interesting? if so, Shall we run it at master, PR, or both?

simitt:
Yes I think Hey-APM and Integration tests should only run in their own pipelines, not in the APM Server. We also have them separated from server tests at the moment.

It would make sense to run `update-beats` in `master`.

We are talking about `PRs` and `master`, however everything we agree on for `master` should be valid also for all `release` branches .

kuisathaverat:
>We are talking about PRs and master, however everything we agree on for master should be valid also for all release branches.

I am going to use this expression, this step will run on master, Digit.Digit, Digit, and v branches, also on tags vDigit.Digit.Digit*, I think that covers all branches except backports and security

```
            allOf {
              anyOf {
                branch 'master'
                branch "\\d+\\.\\d+"
                branch "v\\d?"
                tag "v\\d+\\.\\d+\\.\\d+*"
                environment name: 'Run_As_Master_Branch', value: 'true'
              }
              environment name: 'bench_ci', value: 'true'
            }
```

Another option is to run on all branches and a tag pattern, so we will exclude PRs and some tags 

```
            allOf {
              anyOf {
                not {
                  changeRequest()
                }
                tag "v\\d+\\.\\d+\\.\\d+*"
                environment name: 'Run_As_Master_Branch', value: 'true'
              }
              environment name: 'bench_ci', value: 'true'
            }
```

Other info that I forgot to mention is that now we send coverage data to Codecov
https://codecov.io/gh/elastic/apm-server/pull/1466

kuisathaverat:
@simitt @jalvz  
it is fully functional and has the Github integration enable, if you agree we can merge this PR and start with the next iteration. After that, I will make PRs to every branch that you want to upload the Jenkins fille and make it buildable. If you want I can show you how it works and walk through the UI on a short AON session.

script/jenkins/unit-test.sh
```diff
@@ -0,0 +1,14 @@
...
+
+source ./_beats/dev-tools/common.bash
+
+jenkins_setup
+
+go get -v -u github.com/jstemmer/go-junit-report
+
+export OUT_FILE="build/test-report.out"
+
+make prepare-tests
```
simitt: you are missing a `make update` before `make prepare-tests`.

script/jenkins/test.sh
```diff
@@ -0,0 +1,28 @@
...
+
+jenkins_setup
+
+#cleanup() {
+#  rm -rf $TEMP_PYTHON_ENV
+#  make stop-environment fix-permissions
+#}
+#trap cleanup EXIT
+
+make testsuite
```
simitt: This will also run the `unit-tests` again. How about you remove the `unit-test` step for now and call this step `linux-tests`?

Jenkinsfile
```diff
@@ -0,0 +1,399 @@
...
+              tar(file: "coverage-files.tgz", archive: true, dir: "coverage", pathPrefix: "${BASE_DIR}/build")
+              codecov(repo: 'apm-server', basedir: "${BASE_DIR}")
+            }
+          }
+        }
+        /**
+        Run tests on a windows environment.
+        Finally archive the results.
+        */
+        stage('windows test') {
```
simitt: This could be run in parallel with the linux tests.

Jenkinsfile
```diff
@@ -0,0 +1,399 @@
...
+          echo "Archive packages"
+          /** TODO check if it is better storing in snapshots */
+          //googleStorageUpload bucket: "gs://${JOB_GCS_BUCKET}/${JOB_NAME}/${BUILD_NUMBER}", credentialsId: "${JOB_GCS_CREDENTIALS}", pathPrefix: "${BASE_DIR}/build/distributions/", pattern: '${BASE_DIR}/build/distributions//**/*', sharedPublicly: true, showInline: true
+        }
+      }
+    }
+    /**
+    Checks if kibana objects are updated.
+    */
+    stage('Check kibana Obj. Updated') {
```
simitt: Please run this before the `Release` step.

Jenkinsfile
```diff
@@ -0,0 +1,399 @@
...
+      options { skipDefaultCheckout() }
+      steps {
+        gitCheckout(basedir: "${BASE_DIR}")
+        stash allowEmpty: true, name: 'source', useDefaultExcludes: false
+        script {
+          env.GO_VERSION = readFile("${BASE_DIR}/.go-version")
+        }
+      }
+    }
+    stage('Build'){
```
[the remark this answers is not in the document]
kuisathaverat: this will increase the time of the build in about 1.5 min, I will change it

Jenkinsfile
kuisathaverat: it does

screenshot 2018-12-04 at 13 52 17

Jenkinsfile
kuisathaverat: ok

simitt:
Great work @kuisathaverat, thanks for putting all this together. I will use the new ci once this is merged to see if something is missing or needs to be changed.

PR #1469, opened 2018-10-24
Add changelog for 6.5.0

simitt:


changelogs/6.5.asciidoc
```diff
@@ -0,0 +1,18 @@
+[[release-notes-6.5]]
+== APM Server version 6.5
+
+https://github.com/elastic/apm-server/compare/6.4\...6.5[View commits]
+
+* <<release-notes-6.5.0>>
+
+[[release-notes-6.5.0]]
+=== APM Server version 6.5.0
```
jalvz: did we stop doing sections? (added, bug fixes, etc.)

changelogs/6.5.asciidoc
```diff
@@ -0,0 +1,18 @@
+[[release-notes-6.5]]
+== APM Server version 6.5
+
+https://github.com/elastic/apm-server/compare/6.4\...6.5[View commits]
+
+* <<release-notes-6.5.0>>
```
jalvz: Need to refer this in `docs/release-notes.asciidoc`

changelogs/6.5.asciidoc
simitt: good question, we didn't do it for `6.4.`, but I have no strong feelings, if preferred we can add them.

jalvz: yes, i saw that. i don't know if it was overlooked or dropped on purpose. i think it helps

graphaelli: +1 on adding those sections

PR #1473, opened 2018-10-24
v2: minor refactor of stream handling

roncohen:
* refactors `v2Handler` into smaller pieces and makes it oblivious to NDJSON reading
* rate limiting is now an explicit argument to `HandleStream` instead of packed in the context
* NewNDJSONStreamReader simplified so it just takes a `LineReader`
* LineReader takes a `bufio.Reader` as an argument instead of creating it, allowing for reuse

### DIFF
--- a/beater/route_config.go
+++ b/beater/route_config.go
@@ -225,8 +225,11 @@ func (v v2Route) Handler(url string, c *Config, report publish.Reporter) http.Ha
 	)
 
 	v2Handler := v2Handler{
-		requestDecoder:  reqDecoder,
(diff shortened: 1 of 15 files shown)


roncohen:
@simitt this adds the argument we previously discussed about instead of packing it into the context and then taking it out. Let me know if you still feel we should keep pass it in through the context.

roncohen:
Need to do some more investigation on this.

roncohen:
opened this back up now.

simitt:
looks like the rate limit tests are broken

PR #1482, opened 2018-10-27
Test export subcommands

graphaelli:
depends on #1481, which updates libbeat with elastic/beats#8684.


graphaelli:
Rebased with master after merging the beats update in #1481, to avoid confusion.

PR #1513, opened 2018-11-05
docs: improve requirements section

watson:
Just ran into these issue my self and thought I'd document it

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,8 @@ To get started with APM please see our [Getting Started Guide](https://www.elast
 ### Requirements
 
 * [Golang](https://golang.org/dl/) 1.10.3
+* Python 2 (accessible via `python` not only via `python2`)
+* The Python virtualenv package (e.g. `pip install virtualenv')
 
 ### Install
 


graphaelli:
Having taken our new writer through the development process recently, I agree major improvements are needed here.  That said, I'd prefer we move this into a developing/contributing section as python + virtualenv are only required for making some code changes and running certain tests - with just `go` you can clone the repo and run `make` to produce a working binary.

The `virtualenv` requirement is currently hidden in the [build](https://github.com/elastic/apm-server/blob/master/README.md#build) section.

watson:
Yeah I saw the `virtualenv` was mentioned in the build section, but thought I'd hoist it up to the top to have it in list form. It seemed weird that there were a section called "requirement", but that it didn't include all the requirements. But maybe that's just me 😅 But I could not get it working at all without those two extra requirements

watson:
In fact I also needed to prefix my `make` calls like so:

```
VIRTUALENV_PARAMS="-p python2.7" make
```

But I think maybe that was just my weird setup, so I didn't document it.

simitt:
This PR has been open for months now. I created https://github.com/elastic/apm-server/issues/2118 to track the issue and suggest to close this PR for now.

PR #1516, opened 2018-11-06
Bugfix v2: Return proper error content type.

### LINKED ISSUE #1514: Wrong response Content-Type for /intake/v2/events
The `Content-Type` header in responses to `/intake/v2/events` requests is `text/plain; charset=utf-8` instead of `application/json; charset=utf-8` or `application/json`.

This means that the error output by the agent isn't very readable as the agent will not try to parse it.

Here's a few examples from curl, but it happens also if I send a valid ndjson document that just fails the JSON schema validation:

```
$ curl -i http://localhost:8200/intake/v2/events
HTTP/1.1 400 Bad Request
Date: Mon, 05 Nov 2018 14:32:01 GMT
Content-Length: 72
Content-Type: text/plain; charset=utf-8

{"accepted":0,"errors":[{"message":"only POST requests are supported"}]}% 
```

```
$ curl -i -X POST http://localhost:8200/intake/v2/events
HTTP/1.1 400 Bad Request
Date: Mon, 05 Nov 2018 14:32:01 GMT
Content-Length: 72
Content-Type: text/plain; charset=utf-8

{"accepted":0,"errors":[{"message":"invalid content type: ''"}]}% 
```

```
$ curl -i -X POST -H 'Content-Type: application/x-ndjson' http://localhost:8200/intake/v2/events
HTTP/1.1 400 Bad Request
Date: Mon, 05 Nov 2018 14:35:54 GMT
Content-Length: 66
Content-Type: text/plain; charset=utf-8

{"accepted":0,"errors":[{"message":"EOF while reading metadata"}]}% 
```

simitt:
fixes #1514 

V2 Endpoints now return proper content-type and content in proper format. On Success no body is returned so no content-type is set. On Failure content type is set to `application/json` if the _Accept_ header specifically requests json or it allows _any_ type. Otherwise it is set to `text/plain; charset=utf-8`.

Details to the bugfix: The `WriteHeader` needs to be called after all other headers have been set, but before the actual `Write` method is called. According to the [WriteHeader](https://golang.org/src/net/http/server.go?s=2977:5840#L83) docs there are two points  to consider:
* Changing the header map after a call to WriteHeader (or Write) has no effect unless the modified headers are trailers.
* If WriteHeader is not called explicitly, the first call to Write will trigger an implicit WriteHeader(http.StatusOK).

### DIFF
--- a/beater/v2_handler.go
+++ b/beater/v2_handler.go
@@ -77,37 +77,48 @@ func (v *v2Handler) statusCode(sr *stream.Result) (int, *monitoring.Int) {
 	return highestCode, monitoringCt
 }
 
-func (v *v2Handler) sendResponse(logger *logp.Logger, w http.ResponseWriter, sr *stream.Result) {
+func (v *v2Handler) sendResponse(logger *logp.Logger, w http.ResponseWriter, r *http.Request, sr *stream.Result) {
 	statusCode, counter := v.statusCode(sr)
 	responseCounter.Inc()
 	counter.Inc()
 
-	w.WriteHeader(statusCode)
 	if statusCode != http.StatusAccepted {
 		responseErrors.Inc()
 		// this signals to the client that we're closing the connection
 		// but also signals to http.Server that it should close it:
 		// https://golang.org/src/net/http/server.go#L1254
 		w.Header().Add("Connection", "Close")
-		w.Header().Add("Content-Type", "application/json")
-		buf, err := json.Marshal(sr)
-		if err != nil {
-			logger.Errorw("error sending response", "error", err)
+		setContentType(w, r)
+		w.WriteHeader(statusCode)
+
+		var err error
+		var buf []byte
+		if acceptsJSON(r) {
+			buf, err = json.Marshal(sr)
+			if err != nil {
+				logger.Errorw("error sending response", "error", err)
+			}
+		} else {
+			buf = []byte(sr.String())
 		}
-		_, err = w.Write(buf)
-		if err != nil {
-			logger.Errorw("error sending response", "error", err)
+
+		if buf != nil {
+			_, err = w.Write(buf)
+			if err != nil {
+				logger.Errorw("error sending response", "error", err)
+			}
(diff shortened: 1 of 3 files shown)


beater/v2_handler.go
```diff
@@ -77,37 +77,48 @@ func (v *v2Handler) statusCode(sr *stream.Result) (int, *monitoring.Int) {
...
 		responseErrors.Inc()
 		// this signals to the client that we're closing the connection
 		// but also signals to http.Server that it should close it:
 		// https://golang.org/src/net/http/server.go#L1254
 		w.Header().Add("Connection", "Close")
-		w.Header().Add("Content-Type", "application/json")
-		buf, err := json.Marshal(sr)
-		if err != nil {
-			logger.Errorw("error sending response", "error", err)
+		setContentType(w, r)
```
roncohen: just some ideas, but instead of `setContentType ` could we simplify by changing `sendJSON` and `sendPlain` to take `statusCode int` and then set the content type inside each one. Maybe we could also change `sendJSON` to take an `interface{}` instead and then use that both in v1 and v2. 

Then, instead of having a `sendStatus` in `common_handlers.go` which is only used in v1 (AFAICT), we could perhaps create one that we can use from both v1 and v2:

```
func sendResponse(w http.ResponseWriter, r *http.Request, statusCode int, response interface{}) {
    if acceptsJSON(r) {
         sendJSON(w, r, statusCode, response)
         return
    }
    sendPlain(w, r, statusCode, response)
}
```

simitt: There are a couple of differences between v1 and v2 (in v2 Content-Type is only set on errors, counter for errors is increased in different places, the returned error messages are not the same). So I could align that and change some behaviour in v1, but I wasn't aiming for any change in v1, especially not within a bugfix. 
I think we should refactor that when we remove v1, but if you feel strongly about it I can come up with something.

roncohen: sure @simitt 👍

PR #1520, opened 2018-11-06
beater: slightly simplify response handling

roncohen:
small refactor to simplify response handling


roncohen:
This does remove the “content-type” header for responses that are errors _and_ have empty bodies for v1. We don’t have a test for it and content-type for empty response shouldn't matter. However, if we think that's problematic, I’m also fine to close the issue.

roncohen:
thanks @simitt !

PR #1524, opened 2018-11-06
[6.5] Bugfix v2: Return proper error content type. (#1516)

simitt:
Backports the following commits to 6.5:
 - Bugfix v2: Return proper error content type.  (#1516)


simitt:
jenkins, retest this please

PR #1540, opened 2018-11-12
[WIP] Map ECS fields

graphaelli:
This change introduces mappings to preview the data model to be used in the next major release.


ruflin:
@simitt Everything from an ECS perspective is an event so I would think this is well aligned. Perhaps `event.type` becomes actually `transaction` etc.?

graphaelli:
Per discussion, closing this and recreating directly against 6.x.

graphaelli:
https://github.com/elastic/apm-server/pull/1609

Issue #1515, opened 2018-11-05
Allow framework.name without framework.version?


watson:
In all our agents it's possible for the user to specify which framework name and version should be reported as part of the service in the metadata object.

But if the user only specifies a framework name, but not a version, the server responds with a 400:

```
Problem validating JSON document against schema: I[#] S[#] doesn't validate with "metadata#"
  I[#/service/framework] S[#/properties/service/properties/framework/required] missing properties: "version"
```

I don't think it's important for us that we have both name and version, so I think we should relax the JSON Schema to allow only one of them.

watson:
In the meantime, the agent documentations should make it very clear that if you specify one, you should also specify the other (cc @elastic/apm-agent-devs)

graphaelli:
Does it ever make sense to report the framework version without the name?  What do you think about just making the framework version optional?

watson:
@graphaelli technically not, but I would rather just allow it than refuse to store the event. Say someone configures the agent with:

```js
require('elastic-apm-node').start({
  frameworkName: getFrameworkNameSomehow(),
  frameworkVersion: getFrameworkVersionSomehow()
})
```

And that `getFrameworkNameSomehow` in some cases returned nothing, then it would fail to record anything. I'd rather allow that I think.

axw:
+1 on making framework version optional. However, unless that change is made for 6.5.0, I probably won't make use of it.

watson:
@axw what do you mean? As in you wont expose the config options to the user?

axw:
> @axw what do you mean? As in you wont expose the config options to the user?

I already check for an empty framework version and change it to a placeholder value ("unspecified"). Unless this change goes into 6.5.0 (which will be required by the Go agent's 1.0 release), I'll have to keep doing that or the transactions will be rejected by older server versions.

watson:
Ah ok, the Node agent doesn't use a placeholder value. In general I'd rather send `null` (or not send the property at all) instead of a placeholder value. It takes up less storage and it's easier to align across agents.

axw:
> Ah ok, the Node agent doesn't use a placeholder value. In general I'd rather send null (or not send the property at all) instead of a placeholder value. It takes up less storage and it's easier to align across agents.

In case I'm not being clear: I'd prefer that too, but that's currently not possible for framework.version, and I already support a framework that doesn't have any versioning scheme. I could just not send framework if the version isn't supplied, and put it back in later if we decide we'll go ahead with this change.

[...]

PR #1546, opened 2018-11-14
spec: Allow framework.name/version to be null

simitt:
implements elastic/apm-server#1515


watson:
We should also allow framework.name to be null

simitt:
I did not see that from the issue, can change to both being nullable.

PR #1555, opened 2018-11-14
update release notes for 6.5

simitt:
implements #1515

### DIFF
--- a/docs/guide/apm-release-notes.asciidoc
+++ b/docs/guide/apm-release-notes.asciidoc
@@ -3,6 +3,28 @@
 == Release notes
 
 [float]
+==== APM version 6.5.0
+
+[float]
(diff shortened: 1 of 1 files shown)


docs/guide/apm-release-notes.asciidoc
```diff
@@ -2,6 +2,28 @@
 [[apm-release-notes]]
 == Release notes
 
+[float]
+==== APM version 6.5.0
```
watson: I might be wrong on this but isn't the correct name for the solution "Elastic APM" instead of just "APM"?

graphaelli: Good question.  Leaving it as is for now as that's what we've had since this log started, I'll follow up on it.

PR #1572, opened 2018-11-20
Update release notes for 6.5.1

simitt:
needs to be backported to `6.x` and `6.5`.


graphaelli:
Sorry I'm late on this but did we intentionally omit `No significant changes.` like we have in other patch changelogs?

Issue #1537, opened 2018-11-09
Docs: Update Tuning section to show v1/v2 differences


bmorelli25:
> For everything related to tuning and config options we need some strategy to point out what to tweak and which options to use when agents talk Intake v1 vs. agents talking intake v2, e.g. if somebody uses a ruby agent with intake v1 and a nodejs agent with intake v2, then both - the deprecated and the newly introduced config options - should be set properly.

simitt:
This looks like the right place to collect thoughts on new sections for trouble shooting in Intake v2:
* Explain and increase `apm-server.rum.event_rate` settings: The rate limiter was bound to number of _requests_ per second per ip in v1. Since with v2 the agents make use of long running requests (default 10s) and flush as many events over a single request as possible, the rate limiter has changed to be bound to number of _events_ sent per second per ip. If the rate limit is hit while events on an established request are sent, the request is not immediately terminated. The intake of the events is only throttled to `event_rate.limit`, which means that the events are queued and processed slower. Only when the allowed buffer queue is also full, the request gets terminated with a `429 - rate limit exceeded` http response. If an agent tries to establish a new request, but the rate limit is already hit, a `429` will be sent immediately.

bmorelli25:
closed with #1576

PR #1530, opened 2018-11-07
Docs: WIP V2/6.5 additions and changes

bmorelli25:
⚠️Work in progress, don't merge

WIP for 6.5 Server & Overview documentation. 

Attempting to display V1/V2 information side by side in the docs was difficult and felt confusing and cluttered. Because of this, I decided to replace V1 information with V2, and link to an "upgrade guide" which explains the changes that were made.

Additions/Changes (some still in progress)
* General grammar/spelling fixes
* Event Types
  * Reworded event types: transactions, spans, errors
  * Moved events into APM Overview and renamed -> `docs/guide/apm-data-model.asciidoc`
  * Added sections on dropped spans & missing spans
* Added agent/server compatibility matrix
* elastic/apm-server#1471 - cleaned up distributed tracing and added screenshot
* elastic/apm-server#1484 - added 6.5 upgrade guide
* elastic/apm-server#1485 - documented new ES data model
* #1486 - updated troubleshooting/tuning
* elastic/apm-server#1506 - added java and go agents to APM server docs
* #1532 - update examples for Intake API and ES

New _APM Overview_ structure 
* Overview - **cleaned up agent section**
* Data Model
  * Transactions
  * Spans
  * Errors
* DT
* Install/Run - **removed agent specific installation instructions**
* Agent/Server Compatibility
* Release Notes

Updates needed to agent documentation:
* https://github.com/elastic/apm-agent-nodejs/pull/672
* https://github.com/elastic/apm-agent-java/pull/297

Redirects needed:
https://www.elastic.co/guide/en/apm/server/6.5/event-types.html
https://www.elastic.co/guide/en/apm/server/6.5/transactions.html
https://www.elastic.co/guide/en/apm/server/6.5/errors.html

https://www.elastic.co/guide/en/apm/get-started/6.5/apm-data-model.html
https://www.elastic.co/guide/en/apm/get-started/6.5/transactions.html
https://www.elastic.co/guide/en/apm/get-started/6.5/errors.html

PR #1576, opened 2018-11-20
docs: WIP tuning and config updates

bmorelli25:
This PR adds the additional tuning and configuration updates that weren't covered in #1530.

Data Ingestion
• Added a note about v1 vs v2 (#1537)
• Adding RUM event rate limit (would this make more sense in common problems?) 
• Tuning APM Server using both v1 and v2 (#1537)

Upgrading to 6.5
• Added section on RUM config changes

Common Problems
• Added links where needed

### DIFF
--- a/docs/common-problems.asciidoc
+++ b/docs/common-problems.asciidoc
@@ -51,6 +51,15 @@ For instance, APM Server 6.2.0 changed the Intake API spec and requires a minimu
 
 View the {apm-get-started-ref}/agent-server-compatibility.html[agent/server compatibility matrix] for more information.
 
+[[event-too-large]]
+[float]
+=== HTTP 400: Event too large
+
+NOTE: Version 6.5 of the APM Server introduced a new intake API (v2). This error only applies to users using **v2** of the intake API. You can learn more about this change in the <<intake-api-changes-65, intake API changes>> documentation. 
+
+APM Agents communicate with the APM server by sending events in an HTTP request. Each event is sent as its own line in the HTTP request body. If events are too large, you should consider increasing the <<max_event_size,`max_event_size`>>
+setting in the APM Server, and adjusting relevant settings in the agent.
+
 [[unauthorized]]
 [float]
 === HTTP 401: Invalid token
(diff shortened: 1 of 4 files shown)


docs/common-problems.asciidoc
```diff
@@ -47,7 +47,9 @@ the {apm-agents-ref}/index.html[agent documentation] for details on what is auto
 === HTTP 400: Data decoding error / Data validation error
 
 The most likely cause for this is that you are using incompatible versions of agent and APM Server.
-For instance, APM Server 6.2.0 changed the Intake API spec and requires a minimum version of each agent.
+For instance, APM Server 6.5 redesigned the Intake API and requires a minimum version of each agent.
```
jalvz: Hmm, this is not true because we still keep v1 for backwards compatibility, right? (the other way around is true: latest agent versions require 6.5)
The change we made in 6.2 was more disruptive ;)

docs/common-problems.asciidoc
```diff
@@ -66,7 +68,7 @@ is coming from an origin not whitelisted in `apm-server.rum.allow_origins`. See
 [float]
 === HTTP 413: Request body too large
```
jalvz: should be more visible the fact that this only can happen for v1?

docs/data-ingestion.asciidoc
```diff
@@ -13,12 +13,17 @@ This section explains how to adapt data ingestion according to your needs.
 [[tune-apm-server]]
 == Tune APM Server
 
+Version 6.5 of the APM Server introduced a new <<intake-api,intake API>>. If you have not upgraded your APM Agent, you'll continue to use the deprecated v1 intake API endpoint. If you have upgraded your agent, you'll use the new v2 intake API endpoint. Understanding which endpoint your APM Agents use to communicate with your APM Server is vital to correctly tuning your data ingestion. Read the <<intake-api-changes-65, intake API changes>> documentation to learn more.
```
jalvz: in v2 there is a `max_event_size` added, is that documented already somewhere else, will be done in a future PR, or wont be done? I don't think users will need to tune that, so not opposing to not document it altogether.

If you document this setting, it should be in connection to an apm-server error "event too large"

docs/common-problems.asciidoc
bmorelli25: I initially left out a note here because this `<<configuration-v1-api,`apm-server.max_unzipped_size`>>` links to a section that mentions this is v1 only. But after reading it over again, it doesn't hurt to also mention this in the `common-problems.asciidoc` file - so I've added it in.

docs/data-ingestion.asciidoc
bmorelli25: Yes! It's currently documented in the [config options](https://www.elastic.co/guide/en/apm/server/current/configuration-process.html#configuration-v2-api), and in the Upgrading to 6.5 [server config changes](https://www.elastic.co/guide/en/apm/server/current/upgrading-to-65.html#server-config-changes-65). 

I've linked to these sections with a new "common problem". Is event too large a `413` error?

jalvz: just checked, and event too large is a 400.

bmorelli25: Thanks @jalvz - I've updated the status code

bmorelli25:
@jalvz - are we good to go on this?

Issue #1585, opened 2018-11-27
Random I/O Timeouts


dpankros:
I'm running APM server 6.5.1 on Kubernetes (AWS EKS).  Full version:
```
Server Version: version.Info{Major:"1", Minor:"10+", GitVersion:"v1.10.3-eks", GitCommit:"58c199a59046dbf0a13a387d3491a39213be53df", GitTreeState:"clean", BuildDate:"2018-09-21T21:00:04Z", GoVersion:"go1.9.3", Compiler:"gc", Platform:"linux/amd64"} (AWS EKS; k8s 1.10.3)
```

I originally deployed as a daemonset, but I've since backed off to a single-pod deployment for debugging purposes.  The apm server is behind an ELB that provides TLS termination.  Inside the ELB, everything is http currently using a shared secret.   The one client (rails-client 2.0.1) I am using as a PoC can send messages to the APM server, but about 50% of the time I get an error in the client command line like: 
```
[ElasticAPM] APM Server reponded with an error:
"read tcp 172.31.22.212:8200->172.31.22.40:41602: i/o timeout"
```
Where `172.31.22.212:8200` is the apm-server pod and `172.31.22.40:41602` is a kube node (??!!) and 41602 isn't an open (i.e. LISTENING) port.  It is possible it was a kube-proxy request from the node-ip to the apm-server.  To be honest, I'm not entirely sure what's going on.

On the apm-server for the same request, I see:
```
INFO	[request]	beater/v2_handler.go:111	error handling request	{"request_id": "0e019f58-2b2d-4bf1-b17e-e03a7c2e5d8d", "method": "POST", "URL": "/intake/v2/events", "content_length": -1, "remote_address": "64.132.xxx.xxx", "user-agent": "http.rb/3.0.0", "error": "read tcp 172.31.22.212:8200->172.31.22.40:41602: i/o timeout"}
```
(remote_address slightly obfuscated, sorry)

The strange thing is that it works about half the time, and fails the other half.   I cannot discern a particular pattern, but maybe together we can figure it out.

dpankros:
Currently, almost no traffic.  It's just my development PoC client (rails) running locally.  The ES is  elastic cloud hosted.  The ES server is not overloaded according to the elastic metrics.
APM Server settings are pretty close to default:
```
    cloud:
      id: ${ELASTIC_CLOUD_ID}
      auth: ${ELASTIC_CLOUD_AUTH}

    apm-server:
      host: 0.0.0.0:8200
[...]

PR #1586, opened 2018-11-27
Docs: add timeout issue to common-problems

bmorelli25:
Added timeout issue (#1585) to `common-problems.asciidoc`

### DIFF
--- a/docs/common-problems.asciidoc
+++ b/docs/common-problems.asciidoc
@@ -219,7 +219,8 @@ the APM server and Elasticsearch log a warning like this:
 
 I/O Timeouts can occur when your timeout settings across the stack are not configured correctly,
 especially when using a load balancer.
-The following error may appear in the client command line:
+
+You may see an error like the one below in the agent logs, and/or a similar error on the APM Server side:
 
 ----------------------------------------------------------------------
 [ElasticAPM] APM Server responded with an error:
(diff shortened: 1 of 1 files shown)


docs/common-problems.asciidoc
```diff
@@ -212,3 +213,25 @@ the APM server and Elasticsearch log a warning like this:
...
+----------------------------------------------------------------------
+
+To fix this, ensure timeouts are incrementing from the {apm-agents-ref}[APM Agent],
+through your load balancer, to the <<read_timeout,APM Server>>.
+
+For example:
+
+----------------------------------------------------------------------
+APM Agent --> ELB  --> APM Server 
+   10s        15s         30s
```
roncohen: ELB is specific to AWS ELB. Might be better to use vendor neutral working like LB or Load balancer. Also, 10s and 30s are defaults on the agents and APM Server respectively. Setting the load balancer timeout to 15s should work if the others are set to the default values.

docs/common-problems.asciidoc
```diff
@@ -212,3 +213,25 @@ the APM server and Elasticsearch log a warning like this:
...
+[float]
+=== I/O Timeout
+
+I/O Timeouts can occur when your timeout settings across the stack are not configured correctly,
+especially when using a load balancer.
+The following error may appear in the client command line:
+
+----------------------------------------------------------------------
+[ElasticAPM] APM Server responded with an error:
+"read tcp 123.34.22.313:8200->123.34.22.40:41602: i/o timeout"
```
roncohen: this is from the APM agents perspective. It will show up in the logs. A similar error will show up on the APM Server side.

bmorelli25:
Thanks Ron! I appreciate the nits as I don't have much experience with load balancing

PR #1609, opened 2018-11-30
model: ECS for 6.x

graphaelli:
Make new 6.x indices look like they will in 7.x by aliasing fields where possible and copying data where necessary.

Tasks:
- [x] All non-indexed fields should be documented now
- [x] fields pending decisions on aliasing: `context.http.status_code` `context.response.status_code` and `context.response.finished`
- [x] decision on `user.user_agent` pipeline (seeking confirmation this is no-op since it's not indexed and user configurable)
- [x] Restore host processor metadata fields
- [x] merge in the latest beats framework update (#1647) now that ECS fields have landed there, no major issues expected.  The host metadata fields will be restored at that time.

Reviewers: please pay particular attention to ensure the meanings of the fields in ECS is respected - that is that not only do the data types match, but the values will be usable across applications.  Also please look for an opportunities to fill in in `event.*` that might have been missed.


_meta/ecs-migration.yml
```diff
@@ -0,0 +1,130 @@
...
+  # not in ECS
+- from: context.service.runtime.name
+  to: service.runtime.name
+
+  # not in ECS
+- from: context.service.runtime.version
+  to: service.runtime.version
+
+- from: context.request.url.full
+  to: url.original
```
webmat: @graphaelli Seeing APM's `url.full` got me thinking.

In ECS, the `.original` fields are typically for a piece of information as it is observed. In access logs, the URL are usually a simple path, so `url.original` would then be a URL path, not a full URL.

We've been discussing adding `url.full` in ECS, and today I created the PR here: https://github.com/elastic/ecs/pull/207

I would recommend using `url.full` instead.

graphaelli: Thanks for adding that to the schema, updates made.  That actually cross off one [the unaccounted for fields](https://github.com/graphaelli/apm-server/blob/35ce28f19c335144c3157b7774d504b1b466d31d/tests/system/test_ecs_mappings.py#L83-L87) nicely.

docs/fields.asciidoc
```diff
@@ -558,6 +558,56 @@ type: alias
 --
 
 
+*`meta.cloud.availability_zone`*::
```
webmat: Note that work is already underway to make the aliases render better: https://github.com/elastic/beats/pull/9269 and https://github.com/elastic/beats/issues/9288

graphaelli:
backport of elastic/beats#9269 will fix the bad docs generated for aliased fields.

webmat:
@graphaelli Just clarifying: `host.hostname` is meant to contain exactly what the command `hostname` would return on that server. `host.name` can also default to this value, but is the field a user would override, if they decide to do so.

graphaelli:
@webmat Can you check my understand here - in the APM use case `host.hostname` would be the `hostname` of the application server being instrumented and `host.name` would default to that but possibly overridden by a user.  Also, we'll need to notify users that `host.name` in <6.6 was the apm-server's hostname but now will be the application server's.  The apm-server hostname has been moved under `agent`.

Note that the application server is the browser in the RUM case.

webmat:
@graphaelli We are currently reviewing whether `agent.hostname` will make it into ECS. I can't answer that yet.

cc @MikePaquette

model/metricset/event.go
```diff
@@ -165,6 +165,7 @@ func (me *Metricset) Transform(tctx *transform.Context) []beat.Event {
 	context := common.MapStr{}
 	if me.Tags != nil {
 		context["tags"] = me.Tags
+		fields["labels"] = context["tags"]
```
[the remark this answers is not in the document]
graphaelli: Yeah, this is copying `context.tags` to `labels` for ECS compatibility.  We don't currently provide a way to report ECS-style `tags` via agents.

_meta/ecs-migration.yml
```diff
@@ -0,0 +1,154 @@
...
+  to: host.architecture
+
+- from: context.system.ip
+  to: host.ip
+
+- from: context.system.platform
+  to: host.os.platform
+
+- from: context.request.method
+  to: http.method
```
simitt: Spans do have a `context.http.method`, I think that should also be mapped to `http.method`.

_meta/ecs-migration.yml
```diff
@@ -0,0 +1,154 @@
...
+
+- from: context.request.http_version
+  to: http.version
```
simitt: From non indexed fields we are also missing: 

- `context.response.headers.content-type`, `context.response.headers_sent`, `context.response.status_code`
- `context.request.headers.cookie`, `context.request.headers.user-agent`

One possibility would be to add `http.request` and `http.response` and then 
map to 
- `http.response.headers.content-type` ,`http.response.headers.sent`, `http.response.status_code`
- `http.request.method`, `http.request.headers.cookie`, `http.request.headers.user-agent` 

But there are also other fields in `context.request` and `context.response` that don't necessarily describe http requests. When splitting those fields up we loose some semantic grouping.

_meta/ecs-migration.yml
```diff
@@ -0,0 +1,154 @@
...
+  to: user.name
+
+- from: docker.container.id
+  to: container.id
+
+- from: docker.container.image
+  to: container.image.name
+
+- from: docker.container.name
+  to: container.name
```
simitt: From non indexed fields we are also missing: 

`context.db.instance`, `context.db.statement`, `context.db.type`, `context.db.user`.
I suggest to map these to `database.instance`, `database.statement`, `database.type` and `database.user.name` as I cannot find any ECS definition, but we need to get rid of the `context`.

simitt: Those fields are _unknown_ missing where I don't have a proposal now:

`context.request.body`
`context.request.env`
`context.request.socket`
`context.request.cookies`
`context.custom`

_meta/ecs-migration.yml
```diff
@@ -0,0 +1,154 @@
...
+  to: cloud.project.id
+
+- from: meta.cloud.provider
+  to: cloud.provider
+
+- from: meta.cloud.region
+  to: cloud.region
+
+- from: context.service.agent.name
+  to: agent.name
```
simitt: Not sure if this should be `name` or `type`. Would it actually be the same if two agents are running on two different hosts instrumentation the same app deployed on two hosts?

_meta/fields.common.yml
```diff
@@ -316,3 +316,256 @@
...
+        # context.request.url.port keyword -> long
+        - name: port
+          type: long
+          description: >
+            The port of the request, e.g. 443.
+
+        - name: query
+          type: alias
+          path: context.request.url.search
+          # TODO: multifield original.text
```
simitt: I don't understand this comment, as I don't see any `text` in our current `url` mappings nor in the ECS schema.

_meta/fields.common.yml
```diff
@@ -316,3 +316,256 @@
...
+
+        - name: provider
+          type: alias
+          path: meta.cloud.provider
+
+        - name: region
+          type: alias
+          path: meta.cloud.region
+
+    - name: container
```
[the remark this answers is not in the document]
ruflin: @graphaelli Are you using these fields in the UI or for queries anywhere?

_meta/fields.common.yml
```diff
@@ -316,3 +316,256 @@
...
+          type: alias
+          path: context.system.architecture
+
+        - name: ip
+          type: alias
+          path: context.system.ip
+
+        - name: name
+          type: alias
+          path: context.system.hostname
```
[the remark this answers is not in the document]
graphaelli: nice catch.  added a test for this and also caught `context.service.version` -> `service.version`

_meta/ecs-migration.yml
```diff
@@ -0,0 +1,154 @@
...
+  copy_to: true
+
+- from: context.process.pid
+  to: process.pid
+
+- from: context.process.ppid
+  to: process.ppid
+
+- from: context.process.title
+  to: process.title
```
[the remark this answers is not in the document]
graphaelli: turns out we can document non-indexed fields in `fields.yml`, added

_meta/fields.common.yml
```diff
@@ -316,3 +316,256 @@
...
+      type: group
+      dynamic: false
+      fields:
+        - name: architecture
+          type: alias
+          path: context.system.architecture
+
+        - name: ip
+          type: alias
+          path: context.system.ip
```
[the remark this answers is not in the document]
graphaelli: Per our discussion, not doing this.

_meta/ecs-migration.yml
graphaelli: We can't alias this since it's not indexed and would be 1-many for users that don't split indices on `processor.event`.  I think since it's display only we could have this write to `http.method` in the next major and note that it won't show up for 6.x indices, including 6.6+.   Should we consider having the UI fallback from `http.method` to `context.request.method` for the medium/long term?

simitt: `context.http.method` is not indexed, it is part of the mapping for non-indexed fields in https://github.com/graphaelli/apm-server/pull/1

simitt: So I think we are good with just storing it at an ECS compliant position. Feel free to close this conversation if you agree.

_meta/ecs-migration.yml
simitt: Suggestion in graphaelli#1

_meta/ecs-migration.yml
```diff
@@ -0,0 +1,154 @@
...
+- from: context.request.url.pathname
+  to: url.path
+
+- from: context.request.url.port
+  to: url.port
+  alias: false
+  copy_to: true
+
+- from: context.request.url.raw
+  to: url.original
```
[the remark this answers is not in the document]
simitt: Moved to https://github.com/graphaelli/apm-server/pull/1/files#diff-4406119bc33ef63390e39887cfd4018eR43

_meta/ecs-migration.yml
simitt: Moved to https://github.com/graphaelli/apm-server/pull/1

tests/system/test_ecs_mappings.py
```diff
@@ -0,0 +1,127 @@
...
+
+            # scripted fields
+            "error id icon",
+            "view errors",
+            "view spans",
+        }
+
+        # TBD
+        exception_fields.update({
+            "beat.timezone", "context.http.status_code", "context.response.finished",
```
simitt: I think `beat.timezone` should be inherited from beats, so no need for a specific mapping on our side.

graphaelli: Agreed!  I expect we'll sync with beats and be able to clear these exceptions before this gets merged.

_meta/fields.common.yml
```diff
@@ -316,3 +316,256 @@
...
+
+        - name: name
```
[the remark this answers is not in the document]
graphaelli: moved `host.name` to `host.hostname` and restored the original `host.name` per review from @simitt

vendor/github.com/elastic/beats/libbeat/publisher/pipeline/module.go
```diff
@@ -85,9 +85,6 @@ func Load(
 					"hostname": beatInfo.Hostname,
 					"version":  beatInfo.Version,
 				},
-				"host": common.MapStr{
```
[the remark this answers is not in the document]
graphaelli: https://github.com/elastic/apm-server/pull/1609#discussion_r239607278

_meta/ecs-migration.yml
graphaelli: Agreed.  Started documenting non-indexed fields here: https://github.com/elastic/apm-server/pull/1609/commits/3ce04c0ae7b6f5a3b408a94fb187d253c3b4ae84, we can use that for fields we move to in 7.0 too.

graphaelli:
https://github.com/elastic/beats/pull/9475 opened for handling object fields with `enabled: false`

_meta/ecs-migration.yml
```diff
@@ -0,0 +1,168 @@
...
+  to: agent.version
+
+- from: context.system.architecture
+  to: host.architecture
+
+- from: context.system.ip
+  to: host.ip
+
+- from: context.system.hostname
+  to: host.hostname
```
simitt: Not sure if https://github.com/elastic/ecs/issues/243#issuecomment-445166416 implied that the `host.name` should always be set (@ruflin )?

webmat: Yes, the pattern is that in the following 3 cases, in ECS, `.name` is meant to be a user-overridable field. If it's not overridden, it should contain the value of its sister field. I don't think we've made it obvious enough in the spec yet, however :-)

- `host.name` defaults to `host.hostname`
- `agent.name` defaults to `agent.type`
- `service.name` defaults to `service.type`

_meta/fields.common.yml
webmat: Context: in ECS, almost all fields are indexed as `keyword`. This is what most monitoring use cases need.

There's two future situations we consider:

- If we decide to add full text search support for some fields in ECS
- If a user decides they want full text search on ECS fields that aren't currently defined that way

In both cases, a `.text` multi-field should be added to the field in question. This is not a breaking change in the context of ECS, and this is not a conflict with ECS, if a user does this.

Currently however, ECS does not define any multi-field this way. If `url.original` is a field APM expects to be able to do full text search on (and for any other ECS field where this applies), you're welcome to define this multi-field.

Note that since the overwhelming majority of monitoring use cases use mostly keyword for aggregation on short values, we've flipped the usual convention around. Instead of the ElasticSearch default of doing `[field]` == `text` and `[field].keyword` == `keyword`, in ECS we do `[field]` == `keyword` and `[field].text` == `text`.

webmat: We'd love do know which fields you're defining as `text` in APM. I do think it would be helpful to define a few as such in ECS.

Your input will help us plan this.

_meta/ecs-migration.yml
webmat: Created https://github.com/elastic/ecs/issues/269

_meta/fields.common.yml
graphaelli: This comment is actually left over from before https://github.com/elastic/ecs/pull/137, which changed a number of fields from `text` to `keyword` in ECS.  I'll remove the comment and audit the remaining fields.

_meta/fields.common.yml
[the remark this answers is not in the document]
ruflin: yes, fixed by now.

_meta/fields.common.yml
```diff
@@ -316,3 +316,256 @@
...
+          type: alias
+          path: context.request.url.raw
+
+        - name: path
+          type: alias
+          path: context.request.url.pathname
+
+        # context.request.url.port keyword -> long
+        - name: port
+          type: long
```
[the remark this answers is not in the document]
ruflin: Is this used in the UI for queries? Is it used in the dashboards? We could also mark this as a breaking change and not apply an alias here if it is rarely used.

_meta/ecs-migration.yml
```diff
@@ -0,0 +1,141 @@
...
+  # not in ECS
+- from: context.service.environment
+  to: service.environment
+
+  # not in ECS
+- from: context.service.framework.name
+  to: service.framework.name
+
+  # not in ECS
+- from: context.service.framework.version
```
ruflin: @graphaelli What is the service here we are talking about? Just out of curiosity? Can you make an example?

_meta/ecs-migration.yml
```diff
@@ -0,0 +1,141 @@
...
+  to: service.runtime.name
+
+  # not in ECS
+- from: context.service.runtime.version
+  to: service.runtime.version
+
+- from: context.service.version
+  to: service.version
+
+- from: context.request.url.full
```
ruflin: @webmat I wonder if we should allow to nest `url` also under `request`. Feels pretty natural.

_meta/ecs-migration.yml
graphaelli: It's the application the user is running.  `service.name` is user-defined and displayed on the apm-ui landing page eg https://demo.elastic.co/app/apm.  services are often built on top of frameworks - `framework.name` is django, express, etc

_meta/fields.common.yml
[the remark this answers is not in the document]
graphaelli: It is not used in queries or any bundled dashboards that I know of.  I'm also in favor of marking it as a breaking change and dropping support for the old version of this field.

_meta/ecs-migration.yml
webmat: If there's only one URL in the event, I don't see a reason to do move it around.

We need to control in how many places things can be nested. Each time we do this, we're growing the amounts of expected places that fields can be found in ECS, which is contrary to ECS' goal of making things easier to find across data sources.

processor/transaction/package_tests/TestProcessTransactionFull.approved.json
```diff
@@ -110,6 +110,12 @@
                     "username": "foo"
                 }
             },
+            "labels": {
+                "organization_uuid": "9f0e9d64-c185-4d21-a6f4-4673ed561ec8"
+            },
+            "process": {
+                "args": "node server.js"
```
webmat: `process.args` is meant to be an array of individual terms.

Hmmm actually the spec formats the example as an array, but doesn't state this clearly otherwise. We'll have to clarify that field description...

Anyhow, doing it as an array of terms enables the possibility of doing keyword searches on any of the args, which I think is desirable.

WDYT, @ruflin?

graphaelli: I completely misunderstood that - I thought the docs were saying it's one big keyword and ended up [implementing](https://github.com/elastic/apm-server/pull/1609/files#diff-f7e31f1f55e1a508546a00f6ae75a26fR35) as such.  Thanks for spotting this @webmat.

_meta/ecs-migration.yml
```diff
@@ -0,0 +1,148 @@
...
+  to: user.email
+
+- from: context.user.id
+  to: user.id
+
+- from: context.user.username
+  to: user.name
+
+- from: context.user.ip
+  to: client.ip
```
simitt: Defining the `user.ip` as `client.ip` - would that not also imply that `user` should be nested under `client` instead of being on top level?

_meta/fields.common.yml
```diff
@@ -82,6 +134,60 @@
...
+            indexed: false
+            analyzed: false
+            searchable: false
+            aggregatable: false
+            # intake enforces:
+            # fields:
+            # - name: socket.encrypted
+            #   type: boolean
+            # - name: socket.remote_address
+            #   type: keyword
```
simitt: Is it possible to also define the nested fields here?

_meta/fields.common.yml
```diff
@@ -30,6 +30,67 @@
...
+          dynamic: false
+          fields:
+            - name: method
+              type: keyword
+              index: false
+              aggregatable: false
+              searchable: false
+              doc_values: false
+
+            - name: status_code
```
simitt: This is an indexed field, same datatype as `context.response.status_code` that is aliased to `http.response.status_code` now. I think this should also be mapped to the same field, by copying it over instead of aliasing.

model/ecs.go
```diff
@@ -0,0 +1,47 @@
...
+
+	// context.request.url.port -> url.port (as int)
+	if port, err := fields.GetValue("context.request.url.port"); err == nil {
+		if portStr, ok := port.(string); ok {
+			if portNum, err := strconv.Atoi(portStr); err == nil {
+				utility.MergeAdd(fields, "url", common.MapStr{"port": portNum})
+			}
+		}
+	}
+}
```
simitt: Sorry for my ignorance, do we have a test somewhere that does set the port to a non-number string?

processor/transaction/package_tests/TestProcessTransactionFull.approved.json
ruflin: yes, should be an array on individual keywords.

_meta/ecs-migration.yml
```diff
@@ -0,0 +1,148 @@
...
+- from: context.request.http_version
+  to: http.version
+
+- from: context.tags
+  to: labels
+  alias: false
+  copy_to: true
+
+- from: context.process.argv
+  to: process.args
```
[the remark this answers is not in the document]
graphaelli: Good point, remove altogether since it's a useless (though allowed, somewhat surprisingly) alias.

_meta/ecs-migration.yml
graphaelli: the other user fields, `id, email, name` don't fit under client.  ECS team has provided guidance to proceed as is while they discuss whether to change that recommendation.

_meta/fields.common.yml
graphaelli: I want to so they'll show up in documentation but not as-is.  Did you have other concerns besides docs?

_meta/ecs-migration.yml
webmat: The IP pairs in ECS have really been defined in terms of network monitoring and access log analysis. The APM/RUM usage hasn't been explored as much.

I do think `client.ip` makes cognitive sense here, when you see the field name it's pretty clear what it is. I don't think there's a need to necessarily fill out the other side of the pair (`server.ip`), despite what the spec says.

So that wasn't your actual question, but I wanted to make sure I mentioned this here :-)

Yeah nesting user at `client.user` would make sense, I think. It's one of the expected places for `user`.

_meta/ecs-migration.yml
```diff
@@ -0,0 +1,148 @@
...
+  to: user.id
+
+- from: context.user.username
+  to: user.name
+
+- from: context.user.ip
+  to: client.ip
+
+- from: context.user.user-agent
+  to: user_agent.original.text
```
webmat: `user_agent.original.text` would be a multi-field, which is defined in the context of the "real" field, which is `user_agent.original` in this case.

I guess we're hitting a corner case here. The field (in terms of setting an alias) is moving to `user_agent.original`. I'm not sure it's possible to create an alias to a multi-field. We need to look into this.

But the usage side, if APM intends to use this as "what field does the query target", then yeah it's now `user_agent.original.text`.

cc @ruflin

_meta/ecs-migration.yml
```diff
@@ -0,0 +1,148 @@
...
+
+- from: listening
+  to: observer.listening
```
webmat: What's this field meant to contain? There isn't a `observer.listening` in ECS, at this time.

_meta/fields.common.yml
graphaelli: I convinced myself that this field is metadata on a span and might actually be confusing if it's in `http.response.status_code`, as it's the response code from calling a remote service vs the response code we return from the traced service.

If you would still like to proceed, are you saying that in:
6.6 - put this value in `context.response.status_code` and `context.http.status_code`
7.0 - write it to `http.response.status_code`

model/ecs.go
webmat: Is there a chance this could happen?

_meta/fields.common.yml
webmat: I can see the ambiguity here. This is to track the response of an external call performed by the monitored app, correct?

I think I'd be inclined to still put it in the ECS field. After all, these events will mostly be looked at in the context of this one event stream. Might as well benefit from the streamlined field names.

But if you see a problem with this, we can hold off on the migration of this field.

_meta/ecs-migration.yml
graphaelli: you can alias to a multi-field but not back from a multi-field (marked as a non-reversible alias)

_meta/ecs-migration.yml
graphaelli: This is an awkward field that I'm expecting will never be in ECS.  It's used as a heartbeat so that kibana can verify that an apm-server has checked in, even if it has not processed any traces.

graphaelli:
> It should be indexed in an ECS index.
> Perhaps if APM doesn't make this useable this requirement can be relaxed? @ruflin WDYT?

We've decided it's not worth indexing this field - it's often huge (think java, where it's not even captured by default) and uninteresting for queries - other fields are better suited for search/aggs, like `process.name`, and then args are handy metadata per document.

_meta/ecs-migration.yml
webmat: Ok. Out of scope for now, but perhaps APM could simply send a completely different (and mostly empty) event that's categorized as a heartbeat? E.g. `event.kind:heartbeat`

graphaelli: Agreed, I'll keep an eye on the `event.kind` progress.

_meta/fields.common.yml
graphaelli: > This is to track the response of an external call performed by the monitored app, correct?
correct

I don't have philosophical objections to this plan - my concern is mostly a practical issue of this being our only 1:many field.  I agree that in 7.0.0 we can write this to `http.response.status_code`, after all. an APM event stream in the distributed tracing case is composed of multiple transactions that will each bring a possibly different `http.response.status_code` value.  It's also not a big effort to write this to both places for spans, I just want to be sure we want that in 6.6 because I'm not sure what the benefit of that is, since we can't make that work for pre-6.6 indices.

simitt: ok, I am good with leaving as is for 6.x and reconsider for 7.0

_meta/fields.common.yml
simitt: No concerns, I was hoping we could add them.

_meta/ecs-migration.yml
simitt: I am good either way, just wanted to make sure it was not overlooked.

simitt:
@graphaelli I don't quite understand what the problem with copying `context.tags` to `labels` is based on customer defined keys. I though we would copy all fields nested under `context.tags. 

Apart from that, this PR LGTM.

graphaelli:
The main concern is that it may surprisingly take a significant amount of space.  I'd like to merge this and follow up after we discuss a bit further on a separate issue.

roncohen:
Exciting! Great work!

Issue #1617, opened 2018-12-04
Add span.subtype and span.action


roncohen:
Following https://github.com/elastic/apm/issues/8 we need to add these two properties to APM Server. They should be indexed as keywords and they should be optional.

roncohen:
I can give it a shot

PR #1618, opened 2018-12-04
add optional fields span.subtype and span.action

roncohen:
Implementing https://github.com/elastic/apm-server/issues/1617 (global issue: https://github.com/elastic/apm/issues/8) on the server

### DIFF
--- a/beater/server_test.go
+++ b/beater/server_test.go
@@ -368,7 +368,7 @@ func TestServerSourcemapElasticsearch(t *testing.T) {
 			expected: []string{"localhost:5200"},
 			config: m{
 				"frontend": m{
-					"enabled":                            "true",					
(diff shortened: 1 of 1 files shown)


jalvz:
agents sending subtype and action to a 6.4 install will see their requests rejected, right?

this could use a changelog entry, if it didn't get one

simitt:
The Intake API allows additional fields, so `6.4` would simply ignore the additional fields but process the request. Otherwise we could never add fields, as every update on the agent side would be a breaking change

Good point regarding the changelog.

roncohen:
thanks! will add change log entry

roncohen:
https://github.com/elastic/apm-server/pull/1622

PR #1625, opened 2018-12-05
Update beats and versions for 6.5.2

simitt:
Update beats framework to 9ac0adf and update docs version and changelog.


simitt:
@jalvz why are you suggesting package tests, nothing changed related to packaging?

simitt:
I don't follow, there is no update for mage. Could you please point to what you mean.

PR #1628, opened 2018-12-06
Docs: clarify apm setup instructions

bmorelli25:
Per customer comments, updating the install and run page to be more clear about needing to install Kibana

### DIFF
--- a/docs/guide/install-and-run.asciidoc
+++ b/docs/guide/install-and-run.asciidoc
@@ -4,14 +4,14 @@
 To get started using Elastic APM,
 you need to have:
 
-* an Elasticsearch cluster and Kibana (version 5.6 or above)
-* APM Server
-* APM agents installed in your services
+* {apm-agents-ref}/index.html[APM agents] installed in your services
+* {apm-server-ref-v}/index.html[APM Server]
+* An {ref}/index.html[Elasticsearch] cluster (version 5.6+)
+* {kibana-ref}[Kibana] (version 5.6+)
 
 TIP: For best results, ensure you're using the same version of Elasticsearch, Kibana, and APM Server. 
 
-For information about setting up an Elasticsearch cluster,
-see the Elasticsearch {ref}/getting-started.html[Getting Started].
+First, you'll need to install and start an Elasticsearch cluster and Kibana. Information on how to do this can be found in the {stack-gs}/get-started-elastic-stack.html[Elastic Stack getting started guide]. Specifically, the {stack-gs}/get-started-elastic-stack.html#install-elasticsearch[install Elasticsearch], and {stack-gs}/get-started-elastic-stack.html#install-kibana[install Kibana] sections.
 
 The following sections show how to get started quickly with Elastic APM on a local machine.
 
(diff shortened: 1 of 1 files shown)


docs/guide/install-and-run.asciidoc
```diff
@@ -4,14 +4,14 @@
 To get started using Elastic APM,
 you need to have:
 
-* an Elasticsearch cluster and Kibana (version 5.6 or above)
-* APM Server
-* APM agents installed in your services
+* {apm-agents-ref}/index.html[APM agents] installed in your services
+* {apm-server-ref-v}/index.html[APM Server]
+* An {ref}/index.html[Elasticsearch] cluster (version 5.6+)
```
watson: v6.2 is the minimum supported version of Elasticsearch for use the Elastic APM

docs/guide/install-and-run.asciidoc
```diff
@@ -4,14 +4,14 @@
...
+* {kibana-ref}[Kibana] (version 5.6+)
```
watson: v6.2 is the minimum supported version of Kibana for use the Elastic APM

docs/guide/install-and-run.asciidoc
bmorelli25: well shoot, that's good to know

docs/guide/install-and-run.asciidoc
jalvz: I actually wouldn't mention versions at all; if anything, just encourage users to have apm server + elasticsearch + kibana all in the same version.
Elasticsearch is not so important, but there is a stronger dependency between kibana and apm server.

docs/guide/install-and-run.asciidoc
```diff
@@ -4,14 +4,14 @@
...
+* {apm-agents-ref}/index.html[APM agents] installed in your services
+* {apm-server-ref-v}/index.html[APM Server]
+* An {ref}/index.html[Elasticsearch] cluster (version 6.2+)
+* {kibana-ref}[Kibana] (version 6.2+)
 
 TIP: For best results, ensure you're using the same version of Elasticsearch, Kibana, and APM Server. 
 
-For information about setting up an Elasticsearch cluster,
-see the Elasticsearch {ref}/getting-started.html[Getting Started].
+First, you'll need to install and start an Elasticsearch cluster and Kibana. Information on how to do this can be found in the {stack-gs}/get-started-elastic-stack.html[Elastic Stack getting started guide]. Specifically, the {stack-gs}/get-started-elastic-stack.html#install-elasticsearch[install Elasticsearch], and {stack-gs}/get-started-elastic-stack.html#install-kibana[install Kibana] sections.
```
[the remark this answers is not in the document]
bmorelli25: Perfect. I like this because we can mirror the overview if we leave out cluster

docs/guide/install-and-run.asciidoc
bmorelli25: Typically I would say we need to mention minimum versions, but because 6.1 is beta, I think if we emphasize the "using the same version" of the stack, you're right, it probably doesn't need to be mentioned. I'll change the TIP to IMPORTANT.

PR #1636, opened 2018-12-11
Docs: Add 6.5.3 release notes

bmorelli25:
Needs to be backported to `6.x` and `6.5`

But... None of the "view commit" links work for me. Does anyone know how we might fix them?


graphaelli:
That will be fixed once the release is out and the `v6.5.3` tag is pushed.

graphaelli:
Usually the stack version goes morning of release and release notes follow post release, to ensure release notes are not up for a version that hasn't been published yet.

Issue #1649, opened 2018-12-13
Support wildcards in bundle_filepath when looking up source maps


jahtalab:
This is useful for websites that have many subdomains and it would be impractical to upload the source map file for each one of them.

[discuss reference](https://discuss.elastic.co/t/apm-rum-agent-source-maps-for-multiple-subdomains/160426)

steve21168:
This feature would be greatly appreciated.  Any organization who provides subdomains on a per client basis will run into this issue.

jalvz:
@steve21168 this will be included in the 6.6 release, which will be out very soon.
You don't need to set any wildcards, but after upgrading you need to upload the sourcemap again in the same way as before.
I hope that helps!

wangking873:
It is recommended to match only resource paths and source maps

No, service.name

No, service.version

Only bundle_ filepath



https://cdn.xxx.com/ [hash]/ xxx.js -> sourceMap

steve21168:
@jalvz This was previously working but is now no longer matching my existing sourcemap on `7.9.2`.  Was there any known changes that break this?

PR #1661, opened 2018-12-17
index the url path only of the sourcemaps

jalvz:
fixes https://github.com/elastic/apm-server/issues/1649

this is the best i came up with, as wildcard queries are slow in elasticsearch (particularly if the wildcard is at the beginning, like in here) and caching would be quite complicated too

- needs changelog


jalvz:
in that case the services would be different, so they wouldn't match, right? is there a realistic case for same service, version and url path, but different hostname?

if so, i also could also add a boost to the full url query term and sort by relevance. either case for me it feels like overdoing it, but i dont know.

simitt:
The implementation LGTM, thanks for taking care of this @jalvz ! 
From a use-case point of view, I am not sure if this can cause troubles, @jahtalab could you give this change a thought from a users perspective?

jahtalab:
Just had a zoom call with @jalvz , decided to consider searching for an exact match first and if that fails to search for only the path part!

jalvz:
thanks @jahtalab! 
The straightforward way would be 6401bb8ace79d6e0d07740411c4f57025cb6bd7b, but we could investigate other alternatives indeed.

jalvz:
jenkins retest this please

jalvz:
jenkins retest this, please

jalvz:
jenkins retest this, please

PR #1665, opened 2018-12-18
automate conflicting libbeat field removals

graphaelli:
Path of least resistance to get this done.  Also attempted to move all of these out of libbeat and into each beat's fields.common.yml but the generated documentation ended up all wrong.


graphaelli:
@jalvz I expect you're doing one more beats update on 6.x - you will want this in place before doing that, assuming it's acceptable.

jalvz:
Thanks @graphaelli !
in case this is needed elsewhere, it has been amended here https://github.com/elastic/apm-server/commit/8d8b859c72caa0605b13f139a0c27163b98d5b33

graphaelli:
great catch, thanks @jalvz

PR #1702, opened 2018-12-21
Test against latest elasticsearch 7.0.0 snapshot

graphaelli:
While retaining compatibility with 6.x.

### DIFF
--- a/_beats/testing/environments/snapshot.yml
+++ b/_beats/testing/environments/snapshot.yml
@@ -3,7 +3,7 @@
 version: '2.1'
 services:
   elasticsearch:
-    image: docker.elastic.co/elasticsearch/elasticsearch:7.0.0-alpha1-SNAPSHOT
+    image: docker.elastic.co/elasticsearch/elasticsearch:7.0.0-SNAPSHOT
     healthcheck:
       test: ["CMD", "curl", "-f", "http://localhost:9200"]
       retries: 300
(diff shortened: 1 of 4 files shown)


sourcemap/elasticsearch.go
```diff
@@ -65,7 +65,8 @@ func (e *smapElasticsearch) runESQuery(body map[string]interface{}) (*es.SearchR
 	e.mu.Lock()
 	defer e.mu.Unlock()
 	for _, client := range e.clients {
-		_, result, err = client.Connection.SearchURIWithBody(e.index, "", nil, body)
+		_, result, err = client.Connection.SearchURIWithBody(e.index, "",
+			map[string]string{"rest_total_hits_as_int": "true"}, body)
```
graphaelli: This parameter should be removed after elastic/beats#9735 makes it back through a beats update

_beats/testing/environments/snapshot.yml
```diff
@@ -3,7 +3,7 @@
 version: '2.1'
 services:
   elasticsearch:
-    image: docker.elastic.co/elasticsearch/elasticsearch:7.0.0-alpha1-SNAPSHOT
+    image: docker.elastic.co/elasticsearch/elasticsearch:7.0.0-SNAPSHOT
```
simitt: Is this going to be fixed also in beats, as it would be overwritten otherwise with the next update?

graphaelli: Yes, that's already updated in beats master so we're just waiting on elastic/beats#9735 before doing a beats update.

PR #1710, opened 2018-12-21
Update beats framework to 1b83f5e

graphaelli:


graphaelli:
sync test failure will be fixed by elastic/kibana#27699

Issue #1371, opened 2018-09-10
Investigate indexing `span.context.db.*`


simitt:
Currently `span.context.db.*` properties are stored but not indexed in ES. This makes it impossible to query/aggregate on them, which might be helpful for multiple use cases (aggregate on them to figure out occurence of queries per type, problematic instances, etc.).

Suggested solution:

attribute | type | description | index type | limit (server) | limit (agent) | note
-- | -- | -- | -- | -- | -- | --
instance | string | db instance name | keyword | 1024 | 1024 | -
user | string | username for accessing db | keyword | 1024 | 1024 |-  
type | string | database type | keyword | 1024 |  1024| -
statement | string | concrete query | text | none | ? | should be sanitized for grouping

alvarolobato:
We discussed this in the server meeting. Originally these fields were removed from the indexes because we where concerned for the disk space usage. We've decided to document how to add these fields to the index in `fields.yml` instead of indexing it by default.

graphaelli:
After some investigation and discussion with the beats team, the fields.yml approach seems too manual and error prone to be the right long-term solution.  Most notably, upgrades become very manual, with merging new fields with user-modified fields.yml.  

Adding discuss label for brainstorming alternative solutions.

roncohen:
there's `setup.template.append_fields` which allows users to add arbitrary fields to the template: https://www.elastic.co/guide/en/beats/filebeat/6.x/configuration-template.html

seems slightly better than asking people to edit their `fields.yml`

simitt:
This is still marked as experimental though (although it landed quite a while ago, see https://github.com/elastic/beats/pull/6024/files). Did you talk to the beats team if there are plans to make this feature stable?

roncohen:
Related: https://github.com/elastic/beats/issues/8607

ruflin:
+1 on making append_fields GA. The main challenge I see around it is that users will add fields after they did run the Beat for the first and will have to overwrite the template to make use of it. We should solve this by docs and think about other options how to raise awareness.

simitt:
The docs already mention when `append_fields` can be used, how to overwrite the template and when template changes are applied. With https://github.com/elastic/beats/pull/16576 the `experimental` flag from `append_fields` will be removed.

tobiasstadler:
What is the status? Are their plans to index the fields by default?

simitt:
[...]

PR #1712, opened 2018-12-22
Allow numbers and boolean values for tags.

### LINKED ISSUE #828: Allow setting numbers as tags
Having the ability to set custom number values which are indexed in Elasticsearch allows to create aggregated graphs based on transactions.

For example, you could track the number of search results for a search transaction, along with the search query to get a list of the top searches without results.

simitt:
Store numbers as scaled_floats, scaling_factor 1000000
fixes #828.

depends on https://github.com/elastic/beats/pull/9772


felixbarny:
What is the maximum number that can be stored with this scaled float format? Are strings, numbers and booleans stored in the same place in ES? What if one span has the tag `foo: bar` and another `foo: 42` (different data types for the same tag)?

simitt:
Yes, strings, numbers and booleans are stored at the same place in ES. If the same attribute has different datatypes this will throw an exception.

graphaelli:
jenkins, retest this please

simitt:
@felixbarny I'd like to merge this soonish - do you have any concerns regarding the handling?

felixbarny:
Do users have to possibility to set another mapping type in case they want to set large numbers?

For most use cases, including recording bytes up to 9PB and Unix timestamps in up to millisecond precision the current range is sufficient.

felixbarny:
If the user can't configure the mapping type I think it would make sense to validate the range in the APM Server so that the error appears in the agent logs as well rather than only in the APM Server and Elasticsearch logs.

But if possible I think users should be able to configure the mapping type, although the default you choose sounds good to me.

simitt:
There is a way to add fields to the ES template, see discussions around this https://github.com/elastic/apm-server/issues/1371, which makes use of the experimental feature  [append_fields](https://www.elastic.co/guide/en/beats/filebeat/6.x/configuration-template.html) (experimental). 
This can **not** be used for tags, as it raises an exception when used for fields that already have a template. 

However, one can make use of the `setup.template.path` and `setup.template.overwrite: true` config options to point to a custom `fields.yml`, and add other types [here](https://github.com/elastic/apm-server/pull/1712/files#diff-2d1b3f471cc1866ddb1f08e68b28cb0cR36). The custom fields.yml then needs to be manually updated every time the server is updated, to ensure new mappings are added, so all features  work as expected. 

So to summarize, there is some hackish way to do add custom mapping types.

simitt:
jenkins, retest this please.

simitt:
> If the user can't configure the mapping type I think it would make sense to validate the range in the APM Server so that the error appears in the agent logs as well rather than only in the APM Server and Elasticsearch logs.

That is not possible with the setup we have at the moment, as the server would need to either handle the whole request syncronously so errors can be returned back to the agents, or the server would need to fetch all indexed `tags` with their types and check them for every incoming document, which would require some kind of shared cache to work between multiple servers. 

The same handling applies if e.g. a mapping explosion or any other issue with storing the data to ES happens.

PR #1748, opened 2019-01-09
Change response format of the root handler and prettify JSON response

### LINKED ISSUE #1679: Change `/` response format.
Copied description from https://github.com/elastic/apm-server/issues/1509#issuecomment-443001335:

> Change / response format. The ok doesn't match the new v2 response format so now it seems strange. For reference, es 7 looks like:

> HTTP/1.1 200 OK
> content-type: application/json; charset=UTF-8
> content-length: 513

> {
  "name" : "06b42f641b07",
  "cluster_name" : "docker-cluster",
  "cluster_uuid" : "u58I5cGVRgyJm1uRigWJjg",
  "version" : {
    "number" : "7.0.0-SNAPSHOT",
    "build_flavor" : "default",
    "build_type" : "tar",
    "build_hash" : "5eb7040",
    "build_date" : "2018-11-29T17:32:22.811477Z",
    "build_snapshot" : true,
    "lucene_version" : "8.0.0",
    "minimum_wire_compatibility_version" : "6.6.0",
    "minimum_index_compatibility_version" : "6.0.0-beta1"
  },
  "tagline" : "You Know, for Search"
}
```  changing this may impact the java agent.

### PR DESCRIPTION
**Pull request by jalvz**
fixes #1679 

Before and after:

`curl -H "Accept: blabla"  http://localhost:8200`

```
{"build_date":"2019-01-09T10:14:34Z","build_sha":"352dce789ccb9a0237cb5d97d58381ef375ad276","version":"7.0.0"}
```

`curl -H "Accept: blabla"  http://localhost:8200`

```
build_date:"2019-01-09T10:12:19Z"
build_sha:"fe25172f74ba25f58a043bb8685bea2c906ecfb0"
version:"7.0.0"
```

`curl http://localhost:8200`

```
{"ok":{"build_date":"2019-01-09T10:14:34Z","build_sha":"352dce789ccb9a0237cb5d97d58381ef375ad276","version":"7.0.0"}}
```

`curl http://localhost:8200`

```
{
  "build_date": "2019-01-09T10:12:19Z",
  "build_sha": "fe25172f74ba25f58a043bb8685bea2c906ecfb0",
  "version": "7.0.0"
}
```

`curl http://localhost:8200/intake/v2/events`

```
{"accepted":0,"errors":[{"message":"only POST requests are supported"}]}
```
 
`curl http://localhost:8200/intake/v2/events`

```
{
  "accepted": 0,
  "errors": [
    {
      "message": "only POST requests are supported"
    }
  ]
}
```

Alternative: get rid of the text/plain responses altogether, eg. Elasticsearch always return JSON no matter the `Accept` header.

At minimum `apm-integration-testing`, agents and Cloud team need to be aware of this.

- [x] needs changelog

### DIFF
--- a/beater/common_handler.go
+++ b/beater/common_handler.go
@@ -18,6 +18,7 @@
 package beater
 
 import (
+	"bytes"
 	"context"
 	"crypto/subtle"
 	"encoding/json"
@@ -44,7 +45,7 @@ type serverResponse struct {
 	err     error
 	code    int
 	counter *monitoring.Int
-	body    interface{}
+	body    map[string]interface{}
 }
 
 var (
@@ -269,30 +270,26 @@ func sendStatus(w http.ResponseWriter, r *http.Request, res serverResponse) {
 	responseCounter.Inc()
 	res.counter.Inc()
 
-	var msgKey string
-	var msg interface{}
+	body := res.body
 	if res.err == nil {
 		responseSuccesses.Inc()
 		if res.body == nil {
 			w.WriteHeader(res.code)
 			return
 		}
-		msgKey = "ok"
-		msg = res.body
 	} else {
 		responseErrors.Inc()
 
 		logger := requestLogger(r)
-		msgKey = "error"
-		msg = res.err.Error()
-		logger.Errorw("error handling request", "response_code", res.code, "error", msg)
+		body = map[string]interface{}{"error": res.err.Error()}
+		logger.Errorw("error handling request", "response_code", res.code, "error", body)
 	}
 
 	if acceptsJSON(r) {
-		sendJSON(w, map[string]interface{}{msgKey: msg}, res.code)
+		sendJSON(w, body, res.code)
 		return
 	}
-	sendPlain(w, fmt.Sprintf("%s", msg), res.code)
+	sendPlain(w, mapToString(body), res.code)
 }
 
 // requestLogger is a convenience function to retrieve the logger that was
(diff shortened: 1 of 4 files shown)


simitt:
Changing this is a breaking change in the agent <-> server communication (at least for the agents using the healthcheck endpoint). 
Despite us discussing a versionised healtcheck endpoint earlier and deciding against it, I suggest to now introduce a versionised endpoint and deprecate the other healthcheck endpoint, as I don't think this is important enough to force customers updating their agents.

jalvz:
The reason for use the root handler is to align better with Elasticsearch. If we remove the current root handler and make a versioned one, that is a breaking change too. 

Keeping the current root handler and adding a versioned one just to remove the "ok" is not worth at all, so we rather close the issue as won't fix.

graphaelli:
I vote we punt on this and close the issue won't fix.

simitt:
Fine with me. We can revisit with the next breaking change on the API level.

jalvz:
I have the following assumptions:
- Users pre 6.5 will have to update agent versions either case because of v1.
- Users post 6.5 are rather early adopters that want to keep their agents always up to date, and will probably update the agent version either case.
- Most users are pre 6.5 
- Not all agents are using the "ok" string in the root handler to drive behavior. 
- There is almost no friction to update an agent, just bump a version in a dependency file.

Are these assumptions correct?

jalvz:
Following up on offline discussion, confirmed that no Elastic agent handles the response content, so this wouldn't require users to upgrade agent versions.

jalvz:
@nachogiljaldo does any Cloud component check the response body of the apm-server healhcheck endpoint (`/`) ?

If so, this PR (aiming 7.0 only) might break it.

nachogiljaldo:
@jalvz we only check the status code (200) not the response.

Also for the intake endpoints, if my sight's not lying, you only beautify the json, is that right? We do have some integration tests for that (that send data, etc.) so if the response changes they might fail (not if it's just a beautification of course).

PR #1749, opened 2019-01-09
Remove concurrent_requests setting

### LINKED ISSUE #1674: Remove `concurrent_request`option
From `v2` on the `concurrent_requests` config option is only used for creating the [publisher's internal go channel](https://github.com/elastic/apm-server/blob/master/beater/beater.go#L134). This doesn't need to be configurable from the user, and can lead to decreased throughput if misconfigured. 

Instead I propose to use a multiple of `runtime.GOMAXPROCS(0)` as queue size, as [the amount of receivers listening to this channel](https://github.com/elastic/apm-server/blob/master/publish/pub.go#L90) is also defined by the number of available CPUs.

jalvz:
fixes #1674

needs changelog

### DIFF
--- a/docs/configuration-process.asciidoc
+++ b/docs/configuration-process.asciidoc
@@ -7,7 +7,6 @@ Example config file:
 ----
 apm-server:
   host: "localhost:8200"
-  concurrent_requests: 5
   rum:
     enabled: true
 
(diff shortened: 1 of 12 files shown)


docs/configuration-process.asciidoc
```diff
@@ -51,12 +50,6 @@ Defaults to 30 seconds.
 Maximum duration in seconds before releasing resources when shutting down the server.
 Defaults to 5 seconds.
 
-[[concurrent_requests]]
```
simitt: As I learned yesterday, we should add removed blocks to the `redirects.asciidoc` section to ensure all links still work.

Issue #1683, opened 2018-12-19
update apm-server writes to corresponding to 6.6 aliases


graphaelli:
Fields aliased according to https://github.com/elastic/apm-server/blob/6.x/_meta/ecs-migration.yml should instead write to those directly, aliases in reverse are not to be added (no need for 6.x UI to load 7.x index data)

non-indexed fields are not represented there and should be migrated as well, where required.

simitt:
Closing this as we are not using aliases.

PR #1755, opened 2019-01-10
 add top-level "observer" field

graphaelli:
Enables `SkipAgentMetadata` to suppress the top-level agent fields, but does not move context.system.agent there yet.

for #1683

### DIFF
--- a/publish/pub.go
+++ b/publish/pub.go
@@ -28,6 +28,7 @@ import (
 
 	"github.com/elastic/apm-server/transform"
 	"github.com/elastic/beats/libbeat/beat"
+	"github.com/elastic/beats/libbeat/common"
 )
 
 type Reporter func(context.Context, PendingReq) error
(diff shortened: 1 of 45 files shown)


publish/pub.go
```diff
@@ -61,12 +62,22 @@ var (
...
 
 		//       If set >0 `Close` will block for the duration or until pipeline is empty
-		WaitClose:         shutdownTimeout,
+		WaitClose: shutdownTimeout,
+		Fields: common.MapStr{
+			"observer": common.MapStr{
+				"type":         info.Beat,
+				"hostname":     info.Hostname,
+				"version":      info.Version,
+				"id":           info.ID.String(),
```
graphaelli: @ruflin thoughts on including `id` and `ephemeral_id` here?  They're not in ECS under `observer` but I don't think they're appropriate elsewhere.

ruflin: Agree that this is where they belong for apm-server.

PR #1756, opened 2019-01-10
Ecs migration file

jalvz:


jalvz:
good ideas, ill get on that tomorrow. thanks :+1:

_meta/ecs-migration.yml
```diff
@@ -0,0 +1,239 @@
...
+- from: context.request.method
+  to: http.request.method
+  index: true
+
+- from: context.request.http_version
+  to: http.version
+  index: true
+
+- from: context.request.body
+  to: http.request.body.content
```
jalvz: @graphaelli so what do we do here?
keep `body.content` as in the last ECS beta and update 6.6, or go with `body` ?

graphaelli: Since this field isn't indexed there's nothing to do in 6.6, right?  We should be able to move wherever we like in 7.0 and either a reindex of a 6.x index will move it, or the UI will smooth over the difference between the 6.x and 7.0 locations when it is displayed.

graphaelli: I think only the response body is indexed, can you confirm?

graphaelli:
I like that test a lot, we'll have to continue documenting fields there that overwrite those set by libbeat to keep this test working, but I think that's ok.  I also think index should default to `true`, like we do in the [index pattern generator](https://github.com/elastic/apm-server/blob/6261944f046ad43d9d9dfadc288f1f0863154696/vendor/github.com/elastic/beats/libbeat/kibana/fields_transformer.go#L137).

jalvz:
done here f7f61d4fa87f8daade107f1d5c5dc18e00b3e62b

jalvz:
jenkins retest this please

PR #1759, opened 2019-01-10
Remove metrics setting

jalvz:
This is not used any more since https://github.com/elastic/apm-server/pull/1731/files#diff-31f9c87c519544cc8f4fbdae985b0d0eL164

As per https://github.com/elastic/apm-server/pull/1737, it doesn't make sense to keep the option around and should be the agent's call whether to enable/disable.

- [x] Needs changelog
- [x] Might worth clarification note to `metrics-api-changes-65`, to be backported

### DIFF
--- a/docs/redirects.asciidoc
+++ b/docs/redirects.asciidoc
@@ -179,6 +179,11 @@ This configuration option is no longer supported. Please see <<configuration-apm
 
 This configuration option is no longer supported. Please see <<configuration-apm-server>> for current configuration options.
 
+[role="exclude",id="metrics.enabled"]
+=== `metrics.enabled`
+
+This configuration option is no longer supported. Please see <<configuration-apm-server>> for current configuration options.
+
 [role="exclude",id="max_request_queue_time"]
 === `max_request_queue_time`
 
--- a/docs/upgrading-to-65.asciidoc
+++ b/docs/upgrading-to-65.asciidoc
@@ -118,8 +118,13 @@ The data format remains unchanged.
 
 The Metricsets endpoint, previously `/v1/metrics`,
 has also integrated with the new intake endpoint at `/intake/v2/events`.
-In terms of functionality, the only change relates to timestamps.
-See the <<metricset-api, metricset documentation>> for more information.
+There are two changes to be aware of:
+
+* Timestamps are now UTC based and formatted as microseconds since Unix epoch.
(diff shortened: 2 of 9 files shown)


graphaelli:
Good call.  This also points out that metrics could not be disabled in the v2 intake.

docs/upgrading-to-65.asciidoc
```diff
@@ -119,6 +119,7 @@ The data format remains unchanged.
 The Metricsets endpoint, previously `/v1/metrics`,
 has also integrated with the new intake endpoint at `/intake/v2/events`.
 In terms of functionality, the only change relates to timestamps.
+The `metrics.enabled` setting only applies to `v1/metrics`, metricsets can't be disabled in v2.
```
jalvz: @bmorelli25 we had this setting for the metrics v1 endpoint, but in v2 is not used because metrics go through the general intake endpoint and can't be disabled.

We missed it this back in 6.5, trying to fix it now. Is this a good way and the right place to note?

jalvz:
125046055650ad6191678686717efba299199a79 to be backported to 6.x, 6.6, and 6.5

docs/upgrading-to-65.asciidoc
bmorelli25: This is the right place, but I'm concerned that the line above this says "the only change relates to timestamps". I'm thinking the paragraph (L119-123) should be changed to something like this:

```
The Metricsets endpoint, previously `/v1/metrics`,
has also integrated with the new intake endpoint at `/intake/v2/events`.
There are two changes to be aware of:

* Timestamps are now UTC based and formatted as microseconds since Unix epoch.
* The `metrics.enabled` setting has been deprecated and only applies to `v1/metrics`.
Metricsets can't be disabled in v2.

See the <<metricset-api, metricset>> schema for more information.
```

docs/redirects.asciidoc
```diff
@@ -179,6 +179,11 @@ This configuration option is no longer supported. Please see <<configuration-apm
 
 This configuration option is no longer supported. Please see <<configuration-apm-server>> for current configuration options.
 
+[role="exclude",id="metrics.enabled"]
```
bmorelli25: Awesome! Thanks for adding this

docs/upgrading-to-65.asciidoc
jalvz: thank you, updated :+1:

Issue #1255, opened 2018-08-03
Global context values override local transaction context values


roncohen:
As shown in [the tests](https://github.com/elastic/apm-server/blob/master/model/context_test.go#L108-L111), context from the top level values currently override any local values. Local values should take precedence.

alvarolobato:
Related https://github.com/elastic/apm-server/issues/1175

jalvz:
@simitt can we close this now?

alvarolobato:
This is probably fixed, needs checking.

simitt:
Just checked, the `user` is currently the only object that can be set in the `metadata` as well as per event. When merging the user the one sent with the event has precedence of the one sent with the metadata. So this issue can be closed.

PR #1783, opened 2019-01-14
Move user related fields to align with ECS schema.

### LINKED ISSUE #1782: Move `user` related fields according to ECS schema
Follow the [mapping](https://github.com/elastic/apm-server/blob/6.x/_meta/ecs-migration.yml#L124) and move user related data to match ECS schema.

simitt:
Move `context.user.email` to `user.email`
Move `context.user.id` to `user.id`
Move `context.user.username` to `user.name`
Move `context.user.ip` to `client.ip`
Move `context.user.user-agent` to `user_agent`

fixes #1782 
depends on https://github.com/elastic/kibana/pull/28678


docs/data/elasticsearch/generated/transactions.json
```diff
@@ -63,6 +63,9 @@
                     "started": 43
                 },
                 "type": "request"
+            },
+            "user": {
+                "id": "123user"
```
simitt: `user` information from the `metadata` is set if no `transaction.context.user` information was sent.

model/error/event.go
```diff
@@ -208,6 +224,10 @@ func (e *Event) fields(tctx *transform.Context) common.MapStr {
 	e.data = common.MapStr{}
 	e.add("id", e.Id)
 
+	if tctx.Metadata.User != nil {
+		e.add("user_agent.original", tctx.Metadata.User.UserAgent)
```
simitt: the `user_agent` needs to be set somewhere within the `fields`, otherwise the `user_agent` pipeline doesn't work.

model/metadata/metadata.go
```diff
@@ -80,7 +80,10 @@ func (m *Metadata) Merge(fields common.MapStr) common.MapStr {
 	utility.Add(fields, "host", m.System.fields())
 	utility.Add(fields, "process", m.Process.fields())
 	utility.MergeAdd(fields, "service", m.Service.fields())
-	utility.MergeAdd(fields, "user", m.User.fields())
+	utility.AddIfNil(fields, "user", m.User.Fields())
```
simitt: I don't think we should merge `user` information from `metadata` and `context.user` but overwrite it.

simitt:
Kibana Index PR is up https://github.com/elastic/kibana/pull/28678

model/metadata/metadata.go
simitt: @graphaelli the `user_agent` and the `client.ip` are set, even if the `context.user` information is set directly in an event, as they are nested under different root keys and not under `user`. So using `AddIfNil` seems proper here.

model/error/_meta/fields.yml
```diff
@@ -4,16 +4,16 @@
   fields:
     - name: view errors
       type: keyword
-      script: "doc['service.name'].size() > 0 ? doc['service.name'].value : doc['context.service.name'].size() > 0 ? doc['context.service.name'].value : null"
+      script: "doc['service.name'].size() > 0 ? doc['service.name'].value : doc['service.name'].size() > 0 ? doc['service.name'].value : null"
```
graphaelli: the second set is for compatibility with 6.x indexes and shouldn't change

_meta/kibana/6/dashboard/apm-dashboards.json
```diff
@@ -592,7 +592,7 @@
               "type": "top_hits",
               "schema": "metric",
               "params": {
-                "field": "context.service.name",
+                "field": "service.name",
```
graphaelli: how does this work with aliases?  Shouldn't block this pr but I think it needs a follow up to use either `service.name` or `context.service.name`, which is present.

ingest/pipeline/definition.json
```diff
@@ -12,8 +12,8 @@
       },
       {
         "user_agent" : {
-          "field": "context.user.user-agent",
-          "target_field": "context.user.user_agent",
+          "field": "user_agent.original",
+          "target_field": "user_agent_parsed",
```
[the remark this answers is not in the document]
graphaelli: can this go under `user_agent` now also?  I don't think there are any field conflicts.   We should also add the ingest processor fields in fields.yml so they are indexed, right?.

model/error/event.go
```diff
@@ -28,6 +28,8 @@ import (
 	"strconv"
 	"time"
 
+	"github.com/elastic/apm-server/model/metadata"
```
graphaelli: surprised this got through the linter

model/metadata/user.go
```diff
@@ -42,24 +44,49 @@ func DecodeUser(input interface{}, err error) (*User, error) {
...
-		IP:        decoder.StringPtr(raw, "ip"),
 		UserAgent: decoder.StringPtr(raw, "user-agent"),
+		IP:        decoder.StringPtr(raw, "ip"),
+		Name:      decoder.StringPtr(raw, "username"),
+		Email:     decoder.StringPtr(raw, "email"),
+	}
+
+	//id can be string or int
+	tmp := decoder.Interface(raw, "id")
+	fmt.Println(tmp)
```
graphaelli: debugging snuck in

model/transaction/event.go
```diff
@@ -21,6 +21,8 @@ import (
 	"errors"
 	"time"
 
+	"github.com/elastic/apm-server/model/metadata"
```
graphaelli: again strange that this got through `make check`

_meta/kibana/6/dashboard/apm-dashboards.json
simitt: Good point, I created a follow up issue https://github.com/elastic/apm-server/issues/1786.

model/error/_meta/fields.yml
simitt: You are right, that needs more thoughts, I am going to revert the dashboard changes here, as they are unrelated to the `user` field changes anyways.

ingest/pipeline/definition.json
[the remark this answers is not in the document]
graphaelli: ah, missed that one.  How about indexing these?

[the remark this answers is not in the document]
simitt: I mostly saw this as an intermediate state as I was hoping the incompatibility could be resolved before launching, and afaik `user_agent` fields haven't been indexed so far. 
But if you think it makes sense I can index them.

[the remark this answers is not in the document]
graphaelli: I'd think we should get them indexed for 7.0 but can hold off until the corresponding [es issue](https://github.com/elastic/elasticsearch/issues/37329) is resolved.

simitt:
@graphaelli @jalvz is there something left you require for this PR? Otherwise I'd appreciate an approval.

simitt:
jenkins, retest this please.

simitt:
jenkins, retest this

model/metadata/metadata.go
graphaelli: Agreed on both counts.  Think we need an integration test for this too?

jalvz:
Great work! Really thorough testing :100:

model/metadata/metadata.go
simitt: The `MergeAdd` was introduced long time ago, https://github.com/elastic/apm-server/pull/687/files. 

I think it is a bug to mix two potential different `user` informations to one, that is why I changed it to `AddIfNil`, which basically first checks if there was a user sent with the specific event, and if not sets the one from the `metadata`. 

The possibility to send a `user` in `metadata` and the `event` and if this makes sense or not is a completely different topic to discuss. Please raise any concerns about this in this issue https://github.com/elastic/apm-server/issues/1255.

ingest/pipeline/definition.json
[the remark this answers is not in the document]
simitt: Created a follow up issue https://github.com/elastic/apm-server/issues/1805

simitt:
Merging this in now as the failing test is a flaky, unrelated one.

PR #1651, opened 2018-12-14
add kubernetes and container fields

roncohen:
Fixes #1621

PR #1785, opened 2019-01-14
add kubernetes and container fields

### LINKED ISSUE #1621: Allow APM Server to receive k8s/container information from agents
This is the APM Server side of https://github.com/elastic/apm/issues/22 and https://github.com/elastic/apm/issues/21.

Metadata should be extended to allow fields the necessary fields under `system.kubernetes` and `system.container`. We need to take the data from there and put it in the right place under the root in `kubernetes.*` and `docker.container.*` (with ECS this should change to just `container.*`)

graphaelli:
replaces #1651
relies on elastic/beats#10060

fixes #1621


beater/test_approved_es_documents/TestPublishIntegrationErrors.approved.json
```diff
@@ -295,6 +311,16 @@
                     "platform": "darwin"
                 }
             },
+            "kubernetes": {
+                "namespace": "namespace1",
+                "node": {
```
jalvz: out of curiosity, why don't we have same fields as ECS (eg we have node, but not container, labels, etc)?

docs/spec/system.json
```diff
@@ -17,6 +17,48 @@
...
+        },
+        "container": {
+            "properties": {
+                "id" : {
+                    "description": "Container ID",
+                    "type": ["string"],
+                    "maxLength": 1024
+                }
+            },
+            "required": ["id"]
```
jalvz: this is required so as to not have an empty `container`? 
if so, probably i am overthinking this, but it could happen that an `id` is not possible to retrieve in some case, then at some point we add another field under `container` like `name`, but agents can't send  `container.name` without a `container.id`?

graphaelli:
Thanks @jalvz.  The corresponding libbeat update is now approved but can't be merged yet due to CI issues - what do you think about merging this now (after resolving conflicts) and following with a beats update?

beater/test_approved_es_documents/TestPublishIntegrationErrors.approved.json
graphaelli: I think it's because this is the only information we're able to retrieve from within a container.

graphaelli:
jenkins, retest this please

graphaelli:
jenkins, retest this please

graphaelli:
jenkins, retest this please

PR #1797, opened 2019-01-16
[6.6] index the url path only of the sourcemaps (#1661)

jalvz:
Backports the following commits to 6.6:
 - index the url path only of the sourcemaps  (#1661)


jalvz:
jenkins retest this please

PR #1809, opened 2019-01-17
Rename transaction span dropped total

### LINKED ISSUE #1675: Investigate to move `transaction.span_count.dropped.total` to `transaction.span_count.dropped`
Move `transaction.span_count.dropped.total` to `transaction.span_count.dropped` so it's at the same level as transaction.span_count.started and aligns with the v2 intake protocol.

I don't see how we can change that without conflicting with 6.x indices, as the type of the ES field would change.

graphaelli:
to be rebase with master, opening early to test a fix in CI infra

resolves #1675


graphaelli:
jenkins, retest this please

graphaelli:
sync job failed, will follow up with an update as we have a bunch of these lined up.

Issue #1752, opened 2019-01-09
[ECS] Update non indexed fields


jalvz:
For reference: https://github.com/graphaelli/apm-server/pull/1/files

Issue #1804, opened 2019-01-17
Move remaining context fields


jalvz:
Review all field names and agree on them for 7.0, with special attention to avoid possible ECS conflicts in the future

jalvz:
Special attention to `context.db`, `context.custom` and `context.response.finished`

jalvz:
Agreed to move everything out of context and rename `span.http` fields names for ECS alignment if it makes sense (just field rename for consistency, not move it to the top)


cc @roncohen

simitt:
Also remember to move `context.page`, see https://github.com/elastic/apm-server/issues/411#issuecomment-420178858

roncohen:
@jalvz thanks for the ping. Seems to make sense to me to get rid of the last stuff in `context`. Is there an overview somewhere of all the changes we're making?

simitt:
[ECS migration](https://github.com/elastic/apm-server/blob/master/_meta/ecs-migration.yml) lists all the changes so far.

PR #1813, opened 2019-01-18
Update http and url ECS fields

jalvz:
- [x] loosen intake API for URL port to accept strings and numbers
- [x] add more tests and make sure to exercise all the fields

This includes all non indexed fields.

Fixes https://github.com/elastic/apm-server/issues/1752


jalvz:
Missing a few unit tests, putting it up for early review

jalvz:
@elastic/apm-ui you need to be aware for this, non indexed fields root attributes are changing as per ECS:

`{event}.context.request.url` - > `url`
`{event}.context.request` (except `url`) - > `http.request`
`{event}.context.response` - > `http.response`

graphaelli:
what's the plan with `url.port` ? I was expecting we would switch to a long for that in 7.0.

model/context.go
```diff
@@ -0,0 +1,79 @@
...
+func HttpFields(http common.MapStr) common.MapStr {
+	var source = func(dottedKeys string) interface{} {
+		return utility.Get(http, dottedKeys)
+	}
+
+	destination := common.MapStr{}
+	utility.Add(destination, "version", source("request.http_version"))
+	utility.DeepAdd(destination, "request.method", source("request.method"))
+
+	utility.DeepAdd(destination, "request.method", source("request.method"))
```
simitt: This is duplicated. 

> Http request method.
The field value must be normalized to lowercase for querying. See "Lowercase Capitalization" in the "Implementing ECS" section.

Referring to [ECS](https://github.com/elastic/ecs#-http-fields) the `http.request.method` needs to be lowercase.

_meta/ecs-migration.yml
```diff
@@ -35,42 +38,42 @@
...
+  index: false
+
+# Not in ECS
+- from: context.request.headers.cookie
+  to: http.request.headers.cookies.original
+  index: false
+
+- from: context.request.headers.user-agent
+  to: http.request.headers.user-agent
+  index: false
```
simitt: This is also not defined in ECS

_meta/ecs-migration.yml
```diff
@@ -35,42 +38,42 @@
...
+
+- from: context.request.headers.content-type
+  to: http.request.headers.content-type
```
simitt: not defined in ECS

_meta/ecs-migration.yml
```diff
@@ -35,42 +38,42 @@
...
+  index: false
+
+- from: context.request.env
+  to: http.request.env
+  index: false
```
simitt: not defined in ECS

_meta/ecs-migration.yml
```diff
@@ -35,42 +38,42 @@
...
+
+- from: context.request.socket
+  to: http.request.socket
+  index: false
```
simitt: also not defined in ECS

_meta/ecs-migration.yml
```diff
@@ -117,53 +120,44 @@
...
-#
-#- from: context.response.finished
-#  to: http.response.finished
+- from: context.request.url.protocol
+  to: url.scheme
 
-#
-#- from: context.response.status_code
-#  to: http.response.status_code
+- from: context.response.finished
```
simitt: this is not necessarily an `http` response

_meta/ecs-migration.yml
```diff
@@ -117,53 +120,44 @@
...
-#  index: false
+- from: context.response.status_code
+  to: http.response.status_code
+
+- from: context.response.headers.content-type
+  to: http.response.headers.content-type
+  index: false
+
+- from: context.response.headers_sent
+  to: http.response.headers.sent
```
simitt: again, those fields are not defined in ECS

_meta/ecs-migration.yml
```diff
@@ -35,42 +38,42 @@
...
-#  index: false
-#
+- from: context.request.method
+  to: http.request.method
+
+- from: context.request.http_version
+  to: http.version
+
+- from: context.request.body
+  to: http.request.body.content
```
simitt: As we don't restrict this to keywords it should be `http.request.body.content.text`

_meta/ecs-migration.yml
simitt: The standard in ECS is that if the name does not end with `text` then it is a `keyword`. This is an object, so not sure how to bring this into ECS convention.

_meta/ecs-migration.yml
```diff
@@ -185,30 +179,25 @@
...
-#  # not in ECS
-#- from: context.db.user
-#  to: context.db.user
-#  index: false
-#
-#- from: context.user.email
-#  to: user.email
+  # not in ECS
+- from: context.db.user
+  to: context.db.user
```
simitt: Why is this kept under `context` while everything else is moved outside?

_meta/ecs-migration.yml
```diff
@@ -35,42 +38,42 @@
...
+  to: http.request.headers.cookies.parsed
+  index: false
+
+# Not in ECS
+- from: context.request.headers.cookie
+  to: http.request.headers.cookies.original
+  index: false
+
+- from: context.request.headers.user-agent
+  to: http.request.headers.user-agent
```
simitt: ECS defines `user_agent`. While it is a root field there and not here, for the purpose of aligning I suggest to use `http.request.headers.user_agent.original`.

beater/test_approved_es_documents/TestPublishIntegrationErrors.approved.json
```diff
@@ -25,52 +25,6 @@
...
-                        "search": "?query=string"
-                    }
-                },
-                "response": {
-                    "finished": true,
-                    "headers": {
-                        "content-type": "application/json"
-                    },
-                    "headers_sent": true,
-                    "status_code": 200
```
simitt: Where did this `headers_sent` and `status_code` go?

model/context.go
```diff
@@ -0,0 +1,79 @@
...
+	utility.DeepAdd(destination, "request.socket", source("request.socket"))
+	utility.DeepAdd(destination, "response.finished", source("response.finished"))
+	utility.DeepAdd(destination, "response.status_code", source("request.status_code"))
+	utility.DeepAdd(destination, "response.headers.content-type", source("response.headers.content-type"))
+	utility.DeepAdd(destination, "response.headers.sent", source("request.headers_sent"))
+
+	return destination
+}
+
+func UrlFields(http common.MapStr) common.MapStr {
```
simitt: I don't think this should be called `http` as most of the information is just fetched from `context`, I'd rather call this `input` or `fields`.

model/error/event.go
```diff
@@ -192,12 +192,18 @@ func (e *Event) Transform(tctx *transform.Context) []beat.Event {
...
 	utility.Add(fields, "client", e.User.ClientFields())
 	utility.Add(fields, "user_agent", e.User.UserAgentFields())
 	tctx.Metadata.Merge(fields)
 
+	utility.Add(fields, "http", m.HttpFields(e.Context))
+	utility.Add(fields, "url", m.UrlFields(e.Context))
+	delete(e.Context, "request")
+	delete(e.Context, "response")
+
+	utility.Add(fields, "context", e.Context)
```
simitt: If I understand this PR correct, then it aims for getting rid of `context`. What's left here?

model/transaction/event.go
```diff
@@ -170,6 +170,13 @@ func (e *Event) Transform(tctx *transform.Context) []beat.Event {
...
+	utility.Add(fields, "http", m.HttpFields(e.Context))
+	utility.Add(fields, "url", m.UrlFields(e.Context))
+	delete(e.Context, "request")
+	delete(e.Context, "response")
+
+	utility.Add(fields, "context", e.Context)
+
 	events = append(events, beat.Event{Fields: fields, Timestamp: e.Timestamp})
 
 	return events
```
simitt: What happens with the `context` in `span`?

_meta/ecs-migration.yml
jalvz: @graphaelli do you have any opinion on this?

_meta/ecs-migration.yml
jalvz: Apparently it only set here: https://github.com/elastic/apm-agent-nodejs/blob/8b66e282044a7a27a0e204503d3b10ba4b73a19e/lib/parsers.js#L142-L145

So should be fine

beater/test_approved_es_documents/TestPublishIntegrationErrors.approved.json
jalvz: this was a typo in the mapping, fixed now. good catch :+1:

_meta/ecs-migration.yml
jalvz: actually, in the intake `body` is either a string or an object (naturally), so this is the same case as `request.env` 
i am leaning towards leaving it as is

model/transaction/event.go
jalvz: https://github.com/elastic/apm-server/issues/1804 open to discuss

model/context.go
jalvz: Another good catch, implemented in a56d3f5ba827a8625283b1b56916ee298c37d8e0

_meta/ecs-migration.yml
simitt: Thanks for checking that 👍

_meta/ecs-migration.yml
simitt: fine with me, thanks for looking into it

simitt:
You need to (manually) update the approval files for the `system-tests`.

_meta/ecs-migration.yml
webmat: The recommendation is to have all fields as `keyword`. If you need `text` datatype, you should have it as a multi-field. This is the reverse of the ES default.

Since this field is not in ECS, you're free to do what you want here.

_meta/ecs-migration.yml
```diff
@@ -35,42 +38,46 @@
...
+  index: false
+
+  # Not in ECS
+- from: context.request.headers.cookie
+  to: http.request.headers.cookies.original
+  index: false
+
+  # Not in ECS
+- from: context.request.headers.user-agent
+  to: http.request.headers.user-agent.original
```
webmat: Actually `user_agent` is in ECS: https://github.com/elastic/ecs#user_agent

_meta/ecs-migration.yml
webmat: And about things that aren't in ECS (like `http.request.env`), they're not automatically a problem. If we don't add it to ECS, there will never be a conflict. And in this case, it's perfectly fine for you to define this field there.

`http.request.env` is the variables defined in the application's context? If that's the case, I can guarantee it'll never be in ECS (as it has nothing to do with the HTTP protocol), and would not named like this (as we try to use full words as much as possible).

So no conflict here 👍

_meta/ecs-migration.yml
webmat: I don't see a risk of conflict here either 👍

_meta/ecs-migration.yml
webmat: If it stays there, just remove the entry. This file is for what's being migrated :-)

_meta/ecs-migration.yml
jalvz: what is the reason for that? the actual header name is User-Agent

graphaelli: I think this should stay as `user-agent` since it's nested under `http.request.headers` but I'd also be ok to switch if everyone else prefers that

webmat: In ECS it's at the top level, so not really a conflict with how you normalize header names as field names (if you decide to move it at the top level).

If you decide to leave it nested here because you prefer to list it as one of the header entries, then I'd opt to leave as is, to keep the consistent usage of '-' with other headers.

My actual suggestion would be to move this to ECS, however (and I thought I had seen another entry in this file to do so).

Moving this field to the ECS name will ensure the user agents recorded by APM can be found easily by any integration or person that supports the common schema, without having to create an exception for APM. That's the whole point of ECS, after all :-)

jalvz: Yes, as you noted there is another entry in this file for that. We have 2 user agents: the one in the request to the apm-server (which is moved to the ECS field at the root) and this one, which is in a request possibly captured by an APM agent.

jalvz:
@sqren this updates the `elasticsearch/generated` files again.

`request.method` is also lower case now, in case is relevant

_meta/ecs-migration.yml
webmat: Are these present in the same event at the same time? If not, then both should break down the user agent in the ECS field, `user_agent` at the top level.

If they can both be present in the same event, then that's a more complicated question :-)

_meta/fields.common.yml
```diff
@@ -20,6 +20,90 @@
...
+    - name: url
+      type: group
+      description: >
+        A complete Url, with scheme, host and path.
+      fields:
+
+       - name: scheme
+         type: keyword
+         description: >
+           The protocol of the request, e.g. "https:".
```
graphaelli: the example should be "https" - ecs is surprisingly specific on this - https://github.com/elastic/ecs#-url-fields.

model/context.go
```diff
@@ -0,0 +1,82 @@
...
+			portInt = &p
+		}
+	} else if portStr, ok := port.(string); ok {
+		if p, err := strconv.Atoi(portStr); err == nil {
+			portInt = &p
+		}
+	}
+	utility.Add(destination, "port", portInt)
+	utility.Add(destination, "original", source("request.url.raw"))
+	utility.Add(destination, "scheme", source("request.url.protocol"))
```
graphaelli: what do you think about trimming `:` here?

model/error/generated/schema/error.go
```diff
@@ -186,7 +186,7 @@ const ModelSchema = `{
                     "maxLength": 1024
                 },
                 "port": {
-                    "type": ["string", "null"],
+                    "type": ["string", "integer","null"],
```
graphaelli: good idea

_meta/ecs-migration.yml
jalvz: Yes, they are present at the same time.

model/context.go
simitt: addressed in https://github.com/elastic/apm-server/pull/1836

PR #1814, opened 2019-01-18
Update beats framework to 54d9923

jalvz:


_beats/libbeat/_meta/fields.common.yml
```diff
@@ -25,3 +24,13 @@
           type: keyword
           description: >
             Error type.
+
+    - name: beat.name
+      type: alias
+      path: host.name
+      migration: true
+
+    - name: beat.hostname
```
graphaelli: This is a good reminder to review all of these aliases that we inherit and decide if  they should be kept (they're not new)

Issue #1673, opened 2018-12-19
Remove dashboards, remove config options and docs for index patterns,


simitt:
We discussed that we will remove everything Kibana and dashboard related from `apm-server.yml` and the docs, as the preferred way of loading ~~dashboards and~~ index patterns is via Kibana. 

~~Since the index pattern still needs to be created from libbeat logic, and the UI solution doesn't allow for configuration yet, we will still keep the functionality around for expert users but clean up the configuration file.~~

docs changes: elastic/apm-server#1715 

UPDATE:
We will remove the documentation and config options around Kibana index pattern and dashboards. 
We will remove bundling dashboards with Kibana and APM Server. 
(updated the title)

simitt:
Update: We should completely remove the functionality to import dashboards from the APM Server and remove bundling the dashboards with APM Server for 7.0.

PR #1791, opened 2019-01-15
[docs] Remove dashboard references in beats

bmorelli25:
* Removes remaining references to "dashboards" in the APM docs.
* Removes old `deprecate_dashboard_loading` attribute and replaces with `no_dashboards` (used by xpack and filebeat) where possible, and `no_dashboards_apm` where not possible.
* Cleans up a couple of the conditionals.
* Related to #1715.

Once this is approved I'll open a PR in the beats repo and make the persistent changes there.

PR #1815, opened 2019-01-21
Remove dashboards from APM Server.

simitt:
From 7.0 on APM dashboards and visualisations will no longer be bundled directly with Kibana or APM Server.

depends on #https://github.com/elastic/kibana/pull/29043
depends on https://github.com/elastic/beats/pull/10277/

closes elastic/apm-server#1673

cc @sqren 

UPDATE:
Updating to the latest libbeat changes that allow to manage a commands flags, also pulled in some more changes on the kibana index pattern. This PR now depends on merging the index pattern changes in to Kibana https://github.com/elastic/kibana/pull/29501


cmd/root.go
```diff
@@ -72,4 +72,15 @@ func init() {
...
 	})
+	// remove dashboard from export commands
+	for _, cmd := range RootCmd.ExportCmd.Commands() {
+		if cmd.Name() == "dashboard" {
+			RootCmd.ExportCmd.RemoveCommand(cmd)
+		}
+	}
+	// remove dashboards and ml-jobs from setup commands
+	setup := RootCmd.SetupCmd
+	setup.ResetFlags()
```
jalvz: Running `./apm-server setup` prints
```Loaded index template
Loading dashboards (Kibana must be running and reachable)
Skipping loading dashboards, No directory /home/juan/go/src/github.com/elastic/apm-server/kibana/6
```
which might confuse users (looks a bit like an error, when it is not)

Also `./apm-server help` references dashboards.

simitt: I will investigate, I am sure this worked when I tried it out.

simitt: I checked again, you are right that it still shows up in the description, it was only removed in the flags listed in the help. I'll push a follow up.

simitt: depends on https://github.com/elastic/beats/pull/10277/

simitt:
jenkins, run package tests please

simitt:
jenkins, retest this please

jalvz:
forgot to mention: i guess we can exclude the sync of `dev-tools/cmd/dashboards` too

simitt:
Thanks for the thorough review @jalvz ! 

* `make dashboards` is only developer utility, so I didn't change anything, but you're right I should. I pushed a change. 

* Regarding the docs: there is a PR https://github.com/elastic/apm-server/pull/1791 where @bmorelli25 removes everything dashboard related. If something is overlooked, please note it there. I didn't update any docs. 

* There is an open PR https://github.com/elastic/beats/pull/10138/files#diff-9f2f32ca2002718c833fab82f71c64ce to remove the `--setup` flag. Afterwards I expect the setup not to show up for the beats help at all, so I didn't bother to change that. 

* I now excluded the `dev-tools/cmd/dashboards` from being synced and removed it. Great finding!

> Unrelated to this PR, just an observation i noticed: is a bit weird that if you pass any gibberish to the export command then it shows the help, but if you pass it to the setup command it is ignored

This is because the `export` cmd actually expects information what to export, eg. `./apm-server export template`, but the `setup` cmd just sets up everything if you don't pass additional information. If you add the  _gibberish_ after the full cmd (`export template abc`), it behaves the same.

simitt:
jenkins, run package tests please

jalvz:
Oh, sorry for the comment on the docs, i thought that PR was merged. Ignore it then.

PR #1825, opened 2019-01-23
Index error.exception as array of objects

### LINKED ISSUE #1817: Index error.exception as an array of objects
In order to support chained exceptions, we'll move towards storing exceptions as an array of objects. Each object in the array will describe the items in an exception chain.

We don't need to change the mapping, just need to change the events we generate to be a slice. Initially, it'll just be a slice/array with one object; we'll extend this later when we update the intake API to add `error.exception.cause`.

axw:
In order to support chained exceptions in the future,
we store exceptions as an array of objects. For now,
the array (if present) will have a single object.
Later, the array will have multiple elements, with
each exception object in the array corresponding to
the preceding exception object's cause.

Closes #1817

### DIFF
--- a/processor/stream/package_tests/span_attrs_test.go
+++ b/processor/stream/package_tests/span_attrs_test.go
@@ -41,7 +41,7 @@ func spanProcSetup() *tests.ProcessorSetup {
 
 func spanPayloadAttrsNotInFields() *tests.Set {
 	return tests.NewSet(
-		"span.stacktrace",
+		tests.Group("span.stacktrace"),
 		tests.Group("context.db"),
 		"context.http",
 		"context.http.url",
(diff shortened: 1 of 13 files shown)


axw:
This requires https://github.com/elastic/kibana/issues/28991 to not break the UI.

axw:
I seem to have broken the test_backend_error integration test, investigating.

processor/stream/package_tests/span_attrs_test.go
```diff
@@ -41,7 +41,7 @@ func spanProcSetup() *tests.ProcessorSetup {
 
 func spanPayloadAttrsNotInFields() *tests.Set {
 	return tests.NewSet(
-		"span.stacktrace",
+		tests.Group("span.stacktrace"),
```
jalvz: why is this changed?

axw: This is due to the change in tests.flattenMapStrStr, to iterate over `[]common.MapStr` fields. Previously there would be an entry for "span.stacktrace", but not for any of the fields in the objects within it.

PR #1831, opened 2019-01-23
[docs] Document metrics

### LINKED ISSUE #1060: Document application metrics

bmorelli25:
* ~WIP~
* ~Needs https://github.com/elastic/kibana/pull/28621 to be merged first.~

Adds metrics documentation, and makes the APM Data Model easier to understand. Closes elastic/apm-server#1060

### DIFF
--- a/docs/guide/apm-data-model.asciidoc
+++ b/docs/guide/apm-data-model.asciidoc
@@ -1,64 +1,32 @@
 [[apm-data-model]]
 == Data Model
 
-Elastic APM agents capture different types of information from within their instrumented applications - namely `transactions`, `spans`, and `errors`. 
+Elastic APM agents capture different types of information from within their instrumented applications - namely `spans`, `transactions`, `errors`, and `metrics`. 
 
-*<<errors,`Errors`>>* contain information about the error or exception that was captured.
-
-*<<transaction-spans,`Spans`>>* contain information about a specific code path that has been executed.
-They measure from the start to end of an activity,
-and they can have a parent/child relationship with other spans. 
-
-*<<transactions,`Transactions`>>* are a special kind of span that have extra metadata associated with them.
-You can think of transactions as the highest level of work you're measuring within a service.
-For example, serving an HTTP request or running a specific background job.
-
-[[transactions]]
-=== Transactions
-
-A transaction describes an event captured by an Elastic APM agent instrumenting a service. Some examples of a transaction might be:
-
-* An HTTP request
-* A background job
-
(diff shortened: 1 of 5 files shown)


docs/guide/apm-data-model.asciidoc
```diff
@@ -1,59 +1,33 @@
 [[apm-data-model]]
 == Data Model
 
-Elastic APM agents capture different types of information from within their instrumented applications - namely `transactions`, `spans`, and `errors`. 
+Elastic APM agents capture different types of information from within their instrumented applications - namely `spans`, `transactions`, `errors`, and `metricsets`.
```
nehaduggal: Metricsets is not very intuitive. Should we just call it metrics?

bmorelli25: Good point. I'm looking back at the design document, and it actually says they should be called "application metrics" in documentation, but that may be out of date. Tagging the team that worked on the design doc: @axw @jalvz or @graphaelli. Thoughts on calling it `metrics` vs `metricsets` vs `application metrics` in documentation?

axw: +1 on "metrics"; "metricsets" is an implementation detail.

jalvz: Generally agreed, I'm fine either case. We however document implementation details (ie. Intake API), in which it is called `metricsets`.

bmorelli25: Great, will make the change. Thank you!

bmorelli25:
Updated docs from `metricsets` to `metrics` per feedback.

docs/guide/apm-data-model.asciidoc
```diff
@@ -90,24 +58,79 @@ Unforseen errors may cause spans to go missing.
 Because the agent notifies the server about how many spans there should be,
 the number of missing spans is able to be calculated and shown in the UI.
 
+[[transactions]]
+=== Transactions
```
jalvz: interesting that you move transactions after spans (so you can say they are a special kind of span, i guess)
nitpick: i'd say extra or additional attributes, not `metadata` (we are overusing a bit this term)

bmorelli25: @nehaduggal - curious to hear your thoughts on this

watson:
It's not clear to me after this update that we require the metric event object to be named `metricset`. I would expect it to be documented here:

https://www.elastic.co/guide/en/apm/server/master/metricset-api.html

PR #1844, opened 2019-01-25
Define page object on Intake API for RUM.

### LINKED ISSUE #1842: Define context.page on Intake API for transactions and errors
So far the RUM agent has been sending `context.page` entry for `errors` and `transactions` without this being defined in the API. Since we haven't locked down the context data (see #411 ) yet, this information has been stored, but not indexed. 

The `page` entry should be properly defined on the Intake API. 
```
{
  page: { referer: "",  url: "" }
}
```

TODO: 
- [ ] clarify if `page.referer` or `page.url` need to be indexed, and if so define a `keyword` limitation of 1024 on the Intake API.

simitt:
fixes #1842


simitt:
jenkins, test this again

model/transaction/event_test.go
```diff
@@ -80,20 +79,34 @@ func TestTransactionEventDecode(t *testing.T) {
...
+				ParentId:  &parentId,
+				TraceId:   traceId,
+				Duration:  duration,
+				Timestamp: timestampParsed,
+				Marks:     marks,
+				Sampled:   &sampled,
 				SpanCount: SpanCount{Dropped: &dropped, Started: &started},
 				User:      &metadata.User{Id: &userId, Name: &name, IP: &userIp, Email: &email},
 				Labels:    common.MapStr{"foo": "bar"},
+				Page:      &model.Page{Url: &url, Referer: &referer},
```
jalvz: i don't understand why page and context.page are both set in this test... as a matter of fact, it confuses me a bit that any  `Event` can have fields defined twice: as attribute, and inside `context` (this is ofc not introduced in this pr)

PR #1848, opened 2019-01-28
Move beat.version to observer.version

### LINKED ISSUE #1824: update beat.version throughout docs/config/code
following on from #1816, `beat.version` is used in many places throughout apm-server and should be changed to `observer.version`, even in cases where there is no event context to extract that from like [here](https://github.com/elastic/apm-server/blob/56f55c05a3b5ca6ea89950cedbaad150ba7561c4/beater/config.go#L186)

simitt:
fixes elastic/apm-server#1824 

Once this is approved, I can update the docs in beats for the change.


_meta/fields.common.yml
```diff
@@ -373,11 +373,34 @@
...
       type: group
       fields:
+
         - name: listening
           type: keyword
           description: >
             Address the server is listening on.
 
+        - name: hostname
+          level: core
```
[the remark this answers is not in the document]
simitt: nope, left over from copy&pasting this from beats.

docs/copied-from-beats/outputconfig.asciidoc
```diff
@@ -252,10 +252,10 @@ to set the index:
 ------------------------------------------------------------------------------
 output.elasticsearch:
   hosts: ["http://localhost:9200"]
-  index: "%\{[fields.log_type]\}-%\{[beat.version]\}-%\{+yyyy.MM.dd}\" <1>
+  index: "%\{[fields.log_type]\}-%{[{beat_version_key}]}-%\{+yyyy.MM.dd}\" <1>
 ------------------------------------------------------------------------------
 
-<1> We recommend including `beat.version` in the name to avoid mapping issues
+<1> We recommend including `{beat_version_key}` in the name to avoid mapping issues
```
bmorelli25: Are you expecting this attribute to be conditionally rendered? Right now we're showing both `observer.version` and `{beat_version_key}`. As an example:

screen shot 2019-01-28 at 10 55 57 am

bmorelli25:
```suggestion
<1> We recommend including
"+{beat_version_key}+"
in the name to avoid mapping issues when you upgrade.
```

docs/copied-from-beats/outputconfig.asciidoc
```diff
@@ -231,7 +231,7 @@ you need to configure the `setup.template.name` and `setup.template.pattern` opt
 (see <<configuration-template>>). You also must set the default index configuration
 in the `apm-server.yml` file.
 
-NOTE: `beat.version` is a field managed by Beats that is added to every document.
+NOTE: `{beat_version_key}` is a field managed by Beats that is added to every document.
```
bmorelli25:
```suggestion
NOTE: +{beat_version_key}+ is a field managed by Beats that is added to every document.
```

docs/copied-from-beats/outputconfig.asciidoc
```diff
@@ -213,15 +213,15 @@ for more information about the environment variables.
 
 ifndef::apm-server[]
 The index name to write events to. The default is
-+"{beatname_lc}-%\{[beat.version]\}-%\{+yyyy.MM.dd\}"+ (for example,
++"{beatname_lc}-%{[beat_version_key]}-%\{+yyyy.MM.dd\}"+ (for example,
```
bmorelli25: This line should not be changed. This is documentation for other beats and not `apm-server`. Suggestion to revert to original code:
```suggestion
+"{beatname_lc}-%\{[beat.version]\}-%\{+yyyy.MM.dd\}"+ (for example,
```

docs/copied-from-beats/outputconfig.asciidoc
simitt: oh, I missed that one. Yes I'd expect it to show up as `observer_version` every where. Some background: the beats team is using `agent.version` from 7.0, and we use `observer.version`. My intention is to avoid having yet another `ifndef` block in the beats docs, and make `agent.version` default but overridable.

docs/copied-from-beats/outputconfig.asciidoc
```diff
@@ -304,10 +304,10 @@ contains the specified string:
 output.elasticsearch:
   hosts: ["http://localhost:9200"]
   indices:
-    - index: "warning-%{[beat.version]}-%{+yyyy.MM.dd}"
+    - index: "warning-%{[{beat_version_key}]}-%{+yyyy.MM.dd}"
```
bmorelli25: These two attributes do not substitute as-is. Line 302 needs to be changed to allow substitution:
```
// old
["source","yaml"]
// new
["source","yaml",subs="attributes"]
```

docs/copied-from-beats/outputconfig.asciidoc
bmorelli25: Perfect, just making sure. Looks like there are four instances of incorrect rendering. I've made comments below on how to fix each. This should allow for easy substitution in the beats docs without any new conditionals.

docs/copied-from-beats/outputconfig.asciidoc
simitt: The reason I also changed this is that beats changed the default from `beat.version` to `agent.version`. But the beats team probably needs to do a larger rework here for all their docs so it is fine with me to not change this here.

simitt:
jenkins, retest this please

bmorelli25:
PR opened in beats repo: https://github.com/elastic/beats/pull/10416

PR #1850, opened 2019-01-28
Split span.types if no subtype and action are given (#1843)

simitt:
forward port of #1843


simitt:
jenkins, retest this please

Issue #1802, opened 2019-01-16
Mark indices as APM


graphaelli:
Since index names may be endlessly customized, make it possible to consistently identify an APM index.  This will also help the migration assistant know which indices need to be / have already been migrated.

https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-meta-field.html seems to be the right way to do this.

simitt:
@jalvz isn't this done already?

jalvz:
Yes, apparently https://github.com/elastic/apm-server/pull/1862#issuecomment-459277204 didn't close the issue.
Thanks

PR #1857, opened 2019-01-29
add observer.version_major

graphaelli:
for identifiying which major version of apm-server created docs

relates to #1802


publish/pub.go
```diff
@@ -70,11 +70,12 @@ func NewPublisher(info beat.Info, pipeline beat.Pipeline, shutdownTimeout time.D
...
 			"observer": common.MapStr{
-				"type":         info.Beat,
-				"hostname":     info.Hostname,
-				"version":      info.Version,
-				"id":           info.ID.String(),
-				"ephemeral_id": info.EphemeralID.String(),
+				"type":          info.Beat,
+				"hostname":      info.Hostname,
+				"version":       info.Version,
+				"version_major": 7,
```
jalvz: why this is hardcoded and not derived from info.Version?

simitt: I assumed it was for performance reasons.

jalvz: performance of `strings.Split("info.Version", ".")[0]`?
ok, was just curious. i guess this code won't be here by 8.0 :+1:

graphaelli: This was to avoid any chance of a runtime error.  A test was added so we don't forget to change this for 8.0.

PR #1858, opened 2019-01-29
[docs] Update RUM compatibility

bmorelli25:
Updates the server/agent compatibility matrix with new RUM version released today. All other agent updates are OK. @vigneshshanmugam or @jahtalab - can either of you confirm that the change below is correct?

Also, I alphabetized the list, because why not?

### DIFF
--- a/docs/guide/agent-server-compatibility.asciidoc
+++ b/docs/guide/agent-server-compatibility.asciidoc
@@ -1,7 +1,27 @@
 [[agent-server-compatibility]]
 == Agent/Server Compatibility
 
-Below is a chart that outlines the compatibility between different versions of the APM agents and the APM Server. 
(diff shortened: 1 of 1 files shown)


docs/guide/agent-server-compatibility.asciidoc
```diff
@@ -65,5 +65,5 @@ Below is a chart that outlines the compatibility between different versions of t
 |Agent Version |APM Server Version
 |0.x |6.3-6.4
 |1.x |6.4
-|2.x |>= 6.5
+|2.x, 3.x |>= 6.5
```
bmorelli25: This is the the change I added for RUM. 
screen shot 2019-01-29 at 2 03 34 pm

PR #1859, opened 2019-01-29
update transaction.name and error.culprit types

### LINKED ISSUE #1856: transaction.name and error.culprit type updates
from @webmat:

`error.culprit` is `text`, should it be?  If so, switch to `keyword` and add a multifield at `error.culprit.text`

`transaction.name` is `text`.  Switch to `keyword` and add a multifield at `transaction.name.text`

No obligation to do this but it would align with the ECS approach generally.

graphaelli:
closes #1856


docs/spec/errors/error.json
```diff
@@ -46,7 +46,8 @@
                 },
                 "culprit": {
                     "description": "Function call which was the primary perpetrator of this event.",
-                    "type": ["string", "null"]
+                    "type": ["string", "null"],
+                    "maxLength": 1024
```
simitt: This is a breaking change.

jalvz: whatever you both prefer, but imo this is ok - it is very unlikely that this causes issues in practical terms, considering the big picture

simitt: I suggest to at least check with the agent devs if this could be a problem (e.g. verbose class and method names in java) etc.

graphaelli: https://github.com/elastic/apm/issues/43 opened for discussion

simitt:
I am fine with this if the agent devs are 👍

webmat:
Thanks folks!

Issue #1807, opened 2019-01-17
Write a reindex script for documents from 6.x to 7.0 indices


simitt:
Write a script to reindex data (v1, v2) from 6.x to 7.0 indices.

related:
- [x] Move `span.hex_id` to `span.id` https://github.com/elastic/apm-server/issues/1677
- [x] Move `transaction.span_count.dropped.total` to `transaction.span_count.dropped` https://github.com/elastic/apm-server/issues/1675
- [x] Index error.exception as array of objects elastic/apm-server#1825
- [ ] ~splitting types into type, subtype and action (https://github.com/elastic/apm-server/issues/1837#issuecomment-457541992) (not sure splitting is possible in the migration script)~

graphaelli:
The reindex script will be deployed with kibana for 7.0, though we may want to consider exposing it elsewhere as well, to facilitate use in ingest node pipelines.

https://github.com/elastic/apm-integration-testing/pull/277 is currently used for developing the script, instructions included there for using it.  That should be migrated to a kibana api call eventually to avoid duplication/getting out of sync.  https://github.com/elastic/kibana/pull/29516 is pending for adding the reindex api, another change is expected to follow shortly after with the apm specifics.

Currently known outstanding:
- [ ] span type split does seem to be impossible without regexes for now (https://github.com/elastic/elasticsearch/issues/20952) - follow up with es team to see if that's getting in, current target is 6.7
- [ ] set `observer.version_major` once that's merged - #1857
- [ ] trim `error.culprit` at 1024 chars? - #1859 
- [ ] handle `context.url`, drop it? try to move to `span.http.url.original`? - https://github.com/elastic/apm-agent-python/issues/393
- [ ] resolve remaining failure from test - `http.request.headers` and ``http.response.headers` are currently failing
- [ ] test v1 documents
- [ ] test RUM documents as they're are currently excluded due to unpredictable timestamps
- [ ] guards against unusable docs (manually inserted, etc)
- [ ] test docs created without index template?

graphaelli:
Per https://github.com/elastic/apm-server/issues/1175#issuecomment-461058384, the `service.framework` override will be reinstated and the test script will remove the exception for skipping that comparison.

roncohen:
> * handle `context.url`, drop it? try to move to `span.http.url.original`?

we decided to simply drop this field because it was never supposed to be there in the first place

PR #1866, opened 2019-02-01
Fetch headers case insensitive

simitt:
Appearantly some agents are sending up headers without lower casing keys, which passes the json spec, but not the manual decoding process. This PR adds to check for upper and lower case keys for header entries. 

fixes issue from https://github.com/elastic/apm-server/issues/1807:
> resolve remaining failure from test - http.request.headers and http.response.headers are currently failing


model/context.go
```diff
@@ -34,21 +34,30 @@ func HttpFields(fields common.MapStr) common.MapStr {
...
 		utility.DeepAdd(destination, "request.method", strings.ToLower(method))
 	}
 	utility.DeepAdd(destination, "request.body.original", source("request.body"))
-	utility.DeepAdd(destination, "request.headers.cookies.parsed", source("request.cookies"))
-	utility.DeepAdd(destination, "request.headers.cookies.original", source("request.headers.cookie"))
-	utility.DeepAdd(destination, "request.headers.user-agent.original", source("request.headers.user-agent"))
-	utility.DeepAdd(destination, "request.headers.content-type", source("request.headers.content-type"))
 	utility.DeepAdd(destination, "request.env", source("request.env"))
 	utility.DeepAdd(destination, "request.socket", source("request.socket"))
+	utility.DeepAdd(destination, "request.headers.cookies.parsed", caseInsensitive("request.", "cookies"))
```
jalvz: so did you find as well `request.Cookies` ?

simitt: No, that was to be on the safe side.

PR #1870, opened 2019-02-04
Align context handling with other fields.

simitt:


model/context.go
```diff
@@ -18,107 +18,264 @@
...
+	Fragment *string
 }
 
 type Page struct {
 	Url     *string
 	Referer *string
 }
 
-func DecodePage(input interface{}, err error) (*Page, error) {
+type Label common.MapStr
```
graphaelli: `Labels`, to match `Headers` ?

model/context.go
```diff
@@ -18,107 +18,264 @@
...
+	}
+	if scheme := decoder.StringPtr(inpUrl, "protocol"); scheme != nil {
+		trimmed := strings.TrimSuffix(*scheme, ":")
+		url.Scheme = &trimmed
+	}
+	if decoder.Err != nil {
+		return nil, decoder.Err
+	}
+
+	if url.Port = decoder.IntPtr(inpUrl, "port"); url.Port != nil {
```
graphaelli: I'd prefer an `else if` here to group these two operations on `url.Port` together for future maintenance.

model/context.go
```diff
@@ -18,107 +18,264 @@
...
+		Domain:   decoder.StringPtr(inpUrl, "hostname"),
+		Path:     decoder.StringPtr(inpUrl, "pathname"),
+		Query:    decoder.StringPtr(inpUrl, "search"),
+		Fragment: decoder.StringPtr(inpUrl, "hash"),
+	}
+	if scheme := decoder.StringPtr(inpUrl, "protocol"); scheme != nil {
+		trimmed := strings.TrimSuffix(*scheme, ":")
+		url.Scheme = &trimmed
+	}
+	if decoder.Err != nil {
```
graphaelli: why check here and not before `scheme`, or omit altogether and return `&url, decoder.Err` after decoding the port?

model/context.go
```diff
@@ -18,107 +18,264 @@
...
+		headers := Headers(decoder.MapStr(inpResp, "headers"))
+		http.Response = &Resp{
+			Finished:    decoder.BoolPtr(inpResp, "finished"),
+			StatusCode:  decoder.IntPtr(inpResp, "status_code"),
+			HeadersSent: decoder.BoolPtr(inpResp, "headers_sent"),
+			Headers:     &headers,
+		}
+	}
+
+	if decoder.Err != nil {
```
graphaelli: We're inconsistent about this but I'd probably have written `return http, decoder.Err` to eliminate the nil check and put a bit more burden on the caller to not use `http` in case of an error

model/context.go
```diff
@@ -18,107 +18,264 @@
...
+	if err != nil {
+		return nil, err
 	}
 	decoder := utility.ManualDecoder{}
 	pageInput := decoder.MapStr(raw, "page")
-	if decoder.Err != nil || pageInput == nil {
+	if decoder.Err != nil {
 		return nil, errorw.Wrapf(decoder.Err, "fetching Page")
 	}
+	if pageInput == nil {
```
graphaelli: since having `page` is unlikely, I think we should check if `raw["page"]` exists upfront, before creating a decoder and attempting any decoding.

model/context.go
simitt: Decoding scheme could cause an error in the `decoder`, that's why I am not checking it before `scheme`. If the `port` is a string, then decoding to an `IntPtr` first would add an error in the `decoder`, so I cannot check the `decoder.Err` afterwards. I could create a separate decoder for the port, but I didn't think it is worth it.

simitt:
@graphaelli thanks for the feedback, I addressed your comments and added some context specific tests, please have another look.

PR #1886, opened 2019-02-06
bump master to 8.0

graphaelli:


jalvz:
there is no changelog juggling here? (setting new head etc)

graphaelli:
We have a lot of changelog juggling still to do across branches, i thought we'd handle that separately, ok?

PR #1887, opened 2019-02-06
[docs] Add OpenTracing bridge

bmorelli25:
Adds a section on OpenTracing to the APM Overview. I think it makes more sense for the specifics on this to live in the agent docs, so this is really just an arrow pointing there with a small amount of information. @felixbarny, would you mind taking a look at this and letting me know if you think there's anything else that should be included here?


bmorelli25:
Thanks Felix!

PR #1889, opened 2019-02-06
clarify source of kubernetes metadata

### LINKED ISSUE #1888: kubernetes exported fields ambiguous about data origin
A user was confused by this [kubernetes exported fields](https://www.elastic.co/guide/en/apm/server/current/exported-fields-kubernetes-processor.html) doc, thinking it meant apm-server collected that data itself rather than the agents. Perhaps some note on that would be helpful?

graphaelli:
closes #1888


jalvz:
this also needs backport to 6.7 and 6.6, right?

graphaelli:
This information is relevant there but the backport would be non-trivial as `overwrite` for fields wasn't backported to 6.x in libbeat (intentionally).  We could simply modify our vendored copies of `_beats/libbeat/processors/add_kubernetes_metadata/_meta/fields.yml` there - wdyt?

jalvz:
oh, then i don't know. don't have a strong preference

PR #1901, opened 2019-02-07
Allow overriding `service` information from `context`.

### LINKED ISSUE #1175: Allow overriding of meta data in intake V2
In Java, you can have multiple applications in one JVM and multiple frameworks in one application. Therefore, I propose that the intake API should allow that specific payload attributes of service can be overridden. Related Java agent issue: https://github.com/elastic/apm-agent-java/issues/136

The [payload](https://github.com/elastic/apm-server/blob/master/docs/spec/transactions/payload.json) JSON schema defines meta-data which is valid for all the transactions and spans it contains. Instead of allowing only some attributes like `service.name` to be overridden in a `Transaction`, add the [`service`](https://github.com/elastic/apm-server/blob/master/docs/spec/service.json) object to `Transaction`, `Span` and `Error`. This allows default values to be set in the `Payload` "header" but allows specific objects to override the defaults. In languages where there can only ever be one `service.name`, this requires no changes and no additional overhead.

One problem with that approach is that there are required attributes in [`service`](https://github.com/elastic/apm-server/blob/master/docs/spec/service.json) like `agent` and `name`. These should not have to be repeated. If possible with JSON schema, it would be nice to remove the `required` part from the service definition and only add it where necessary whenever the `service` is used.

simitt:
fixes #1175


beater/test_approved_es_documents/TestPublishIntegrationTransactions.approved.json
```diff
@@ -214,7 +214,7 @@
                 "name": "1234_service-12a3",
                 "runtime": {
                     "name": "node",
-                    "version": "8.0.0"
+                    "version": "7.0"
```
graphaelli: I think this will fail on CI

docs/spec/user.json
```diff
@@ -1,7 +1,6 @@
 {
     "$id": "docs/spec/user.json",
     "title": "User",
-    "description": "Describes the authenticated User for a request.",
```
graphaelli: 👍

utility/map_str_enhancer.go
```diff
@@ -26,101 +26,88 @@ import (
...
+		}
 		return
 	}
 
 	switch value := val.(type) {
 	case *bool:
 		if value != nil {
 			m[key] = *value
-		} else {
+		} else if remove {
```
graphaelli: are these elses reachable given the check on line 46?

beater/test_approved_es_documents/TestPublishIntegrationTransactions.approved.json
graphaelli: super confused, that didn't fail

graphaelli: oh 🤦‍♂️ that's the runtime

simitt:
I added some tests showing that data from the `metadata.service` cannot be _deleted_ but only overriden if set in `context.service`. This means if there is `null` set in the event information this does _not_ override the information from the metadata. 
One can set an empty string though, in `event.context.service` as well as in `metadata.service`. Since we are treating zero values as values, I would not call this _deleting_. To be consistent and less error prone I think we should either allow zero values on both places or on none. 

For `users` it behaves differently. Since we are not merging the user information but either take the whole information from `metadata` or from `context` one could kind of delete information on an event. 

I tried to explain this here, but I'll probably have to rephrase as it doesn't seem to be clear enough:

**user**:
> Describes the correlated user for this event. If user data are provided here, all user related information from metadata is ignored, otherwise the metadata's user information will be stored with the event.

**service**:
> Service related information can be sent per event. Provided information will override the more generic information from metadata, non provided fields will be set according to the metadata information.

utility/map_str_enhancer.go
simitt: Line 46 is just a shortcut for cases where the `val` itself is `nil`. The `else` checks are for the cases where the interface `val` is not nil, meaning that it holds a type information and a pointer to the data. The data themselves can be nil again though. 

I found https://blog.golang.org/laws-of-reflection#TOC_3 very helpful when I ran into this.

graphaelli: Ah, that makes sense, thanks

docs/spec/metadata.json
```diff
@@ -5,7 +5,16 @@
     "type": ["object"],
     "properties": {
         "service": {
-            "$ref": "service.json"
+            "$ref": "service.json",
+            "type": "object",
+            "properties.agent.required": ["name", "version"],
```
graphaelli: I didn't know you could do this, neat

utility/map_str_enhancer.go
```diff
@@ -148,7 +135,7 @@ func Add(m common.MapStr, key string, val interface{}) {
 		v := reflect.ValueOf(val)
 		switch v.Type().Kind() {
 		case reflect.Slice, reflect.Array:
-			if v.Len() == 0 {
+			if v.Len() == 0 && remove {
```
graphaelli: does that mean we'll store an empty slice/array if remove is not set? is that what we want?

eyalkoren:
This is not backported to 6.x, right?

simitt:
@eyalkoren confirm, it is not backported.

eyalkoren:
Users with agents that support that upgrading the APM server to 7.x may get their views and aggregations completely changed. Is this documented somewhere?

simitt:
No, you're right, this is missing a changelog entry. I think this should go into the [APM wide release notes](https://www.elastic.co/guide/en/apm/get-started/current/apm-release-notes.html) and also into the server release notes.

PR #1917, opened 2019-02-12
add empty dashboard list. Fixes #1905

### LINKED ISSUE #1905: Prevent APM Server from trying to load dashboards.
As reported by @marius-dr the APM Server throws an error and stops working when the configuration file contains `setup.dashboards.enabled: true`. 
While all dashboard related options have been removed from the APM Server cmds and documentation and `yml` files have been updated to not mention dashboards at all in combination with APM Server, the logic to try and load the dashboards is still inherited from beats. 

This leeds to APM Server throwing an error as no dashboards are bundled with the server any longer. 

Having a quick look, I don't see an option to prevent this from happening in APM Server itself. This would require changes in libbeat. Right now we could only add some hardcoded check to not load dashboards when beat is apm. @urso do you think this would qualify for another _supporter_ similar to index setup? 

Generally my approach is that it is ok to fail if users use some undocumented, unsupported functionality that is simply inherited from beats. This case is different though, as users will still have this option set in their yml files, as it is supported up until 7.0.

roncohen:
This lets APM Server start with the (now deprecated) setting `setup.dashboards.enabled: true`:
This is a quick fix while we work on a better long term solution. 

Will look to contribute the beats changes back to beats.

```
019-02-12T13:00:06.557+0100	INFO	instance/beat.go:273	Setup Beat: apm-server; Version: 7.0.0
2019-02-12T13:00:06.557+0100	INFO	elasticsearch/client.go:165	Elasticsearch url: http://localhost:9200
2019-02-12T13:00:06.557+0100	INFO	[publisher]	pipeline/module.go:113	Beat name: Rons-MacBook-Pro-133.local
2019-02-12T13:00:06.557+0100	INFO	[beater]	beater/beater.go:81	No pipeline callback registered
2019-02-12T13:00:06.558+0100	INFO	kibana/client.go:118	Kibana url: http://localhost:5601/fzq
2019-02-12T13:00:06.639+0100	INFO	kibana/client.go:118	Kibana url: http://localhost:5601/fzq
2019-02-12T13:00:09.150+0100	INFO	instance/beat.go:727	Kibana dashboards successfully loaded.
2019-02-12T13:00:09.150+0100	INFO	instance/beat.go:383	apm-server start running.
```

### DIFF
--- a/_beats/libbeat/scripts/unpack_dashboards.py
+++ b/_beats/libbeat/scripts/unpack_dashboards.py
@@ -6,6 +6,8 @@ import argparse
 
 def transform_data(data, method):
     for obj in data["objects"]:
+        if "attributes" not in obj:
+            continue
         if "uiStateJSON" in obj["attributes"]:
             obj["attributes"]["uiStateJSON"] = method(obj["attributes"]["uiStateJSON"])
 
--- a/_meta/kibana/7/dashboard/apm-dashboards.json
+++ b/_meta/kibana/7/dashboard/apm-dashboards.json
@@ -1,0 +1,4 @@
+{
+    "objects": [{}], 
+    "version": "7.0.0-beta1"
+}


_beats/libbeat/scripts/unpack_dashboards.py
```diff
@@ -6,6 +6,8 @@
 
 def transform_data(data, method):
     for obj in data["objects"]:
+        if "attributes" not in obj:
+            continue
```
simitt: This change will be overwritten with the next `make update-beats`. Are you putting up a PR in beats with it? Otherwise we should ensure to adapt the `update-beats` script to either not overwrite this file or print out a warning if it changed and would be overwritten.

roncohen: yes, that is what this was trying to say:

> Will look to contribute the beats changes back to beats.

roncohen: filed here: https://github.com/elastic/beats/pull/10701

PR #1924, opened 2019-02-13
Update beats framework to 28e0a0b

graphaelli:


graphaelli:
verified docs build looks good.  sync test passes locally

PR #1929, opened 2019-02-14
Uncomment  Beta1 changelog

bmorelli25:


graphaelli:
ignore the build failure, fix is pending

PR #1932, opened 2019-02-15
Update changelog and release state.

simitt:
We need to update [APM release notes](https://www.elastic.co/guide/en/apm/get-started/7.0/apm-release-notes.html) for the beta-1 release.
@elastic/apm-ui @elastic/apm-agent-devs please let me know what we should add for your sections. 

We should get this out asap.

### DIFF
--- a/changelogs/7.0.asciidoc
+++ b/changelogs/7.0.asciidoc
@@ -12,6 +12,10 @@
 https://github.com/elastic/apm-server/compare/v6.7.0\...v7.0.0[View commits]
 
 [float]
+==== Breaking Changes
+- Move fields in ES to be ECS compliant {pull}1766[1766],{pull}1783[1783],{pull}1813[1813],{pull}1836[1836],{pull}1838[1838],{pull}1844[1844],{pull}1848[1848],{pull}1849[1849],{pull}1863[1863],{pull}1870[1870]
+
+[float]
 ==== Added
 
 - Update Elastic APM Go agent to v1.1.2 {pull}1711[1711], {pull}1728[1728].
--- a/changelogs/head.asciidoc
+++ b/changelogs/head.asciidoc
@@ -8,4 +8,4 @@
 [float]
 ==== Removed
 
-https://github.com/elastic/apm-server/compare/v7.0.0-alpha2...master[View commits]
+https://github.com/elastic/apm-server/compare/v7.0.0-beta1...master[View commits]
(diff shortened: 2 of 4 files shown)


formgeist:
I reckon the biggest the change for APM UI is the move to ECS and filtering out 6.x data. Perhaps a changelog that reads like this?

---

*APM UI*

- Breaking change: Moving to new data endpoints. We are filtering out 6.x or lower so if you want to keep your old data, migrate it using the Kibana Migration Assistant (in the Kibana Management section).
- Added support for frozen indices.

@elastic/apm-ui anything else that's critical to mention here?

simitt:
@formgeist I think your suggestions are great for the 7.0.0 release, but I would not suggest that users use the migration script of the beta release to migrate their production data. They should wait for the 7.0.0 release for that. 

We can either add a disclaimer that the script should only be tried out on non-production data or not mention the migration script at all for now. WDYT? (cc @roncohen @graphaelli ).

gregkalapos:
I'm not sure whether we should add this, and in case we add this how to formulate, but:
- We shipped the first preview release of the .NET APM Agent
- (the bad part:) If I remember correctly in Server 7.0 we removed intake v2, so the .NET preview agent won't work with it. I explicitly added to the agent blog/video that the agent is only tested with server 6.5 and 6.6, and 7.0 beta won't work with the first .NET preview agent.

formgeist:
@simitt You're absolutely right, it makes sense for users to wait until GA to migrate their data. I can change the copy to simply add as a disclaimer that APM UI will only display 7.0 data or higher, but users can migrate their data using the migration assistant, but we recommend not doing so on production data. How does that sound?

simitt:
WDYT about something along the lines of:
> We offer a Kibana Migration Assistant (in the Kibana Management section) to help you migrate your data. Be aware that this script is **not** supposed to be run on production data at this stage, as this is a beta release.

formgeist:
@simitt Sounds good

simitt:
@gregkalapos interesting. As you pointed out the agent will not work with 7.0 for now, so I suggest to not add this as it probably creates confusion.

gregkalapos:
> @gregkalapos interesting. As you pointed out the agent will not work with 7.0 for now, so I suggest to not add this as it probably creates confusion.

👍 sounds good! I already work on migrating to intake v2, once that done, I think we can add it to a next release note.

PR #1935, opened 2019-02-15
[7.0] add tests for fixing cmd `apm-server setup template`

simitt:
* backports adding tests for setup cmd.


tests/system/test_setup.py
```diff
@@ -0,0 +1,78 @@
...
+        super(BaseTest, self).setUp()
+
+        self.elasticsearch_url = self.get_elasticsearch_url()
+        self.es = Elasticsearch([self.elasticsearch_url])
+
+        try:
+            self.es.indices.delete_template('apm-*')
+        except:
+            pass
+        assert self.es.indices.exists_template(name='apm-*') == False
```
graphaelli: not a big deal, but `assert not self.es.indices.exists_template(name='apm-*')` would be more idiomatic imo

tests/system/test_setup.py
```diff
@@ -0,0 +1,78 @@
...
+
+        shutil.copy(self.beat_path + "/_meta/beat.yml",
+                    os.path.join(self.working_dir, "apm-server.yml"))
```
graphaelli: I think subclassing `test_export.SubCommandTest` would eliminate the need for this.

simitt: I didn't want to derive from `ServerSetUpBaseTest` as this calls `wait_until_startup()` in the `setup()` command, which again expects to see _Starting apm-server_ in the logs, which won't be there. I did not want to go down the path of overwriting this.

graphaelli: `SubCommandTest` already implements `wait_until_startup` to cover that.  I wouldn't insist except that I'm worried about having the config generation in multiple places.

simitt: You're right! I'll change accordingly.

PR #1937, opened 2019-02-15
Change read_timeout to 45s to fix a race condition (Closes #1936)

### LINKED ISSUE #1936: Race Condition by having same metrics interval reporting and read_timeout default values
Having a default HTTP read_timeout the same value as default metrics interval leads to a race condition on the HTTP connection.

The problem scenario:

1. An TCP connection is established due to a previous HTTP request from an agent
2. Exactly 30s passes with no requests on this TCP stream
3. A client tries HTTP POST to /intake/v2/events to report metrics
4. At the same time but a few nano seconds later the server decides to timeout the connection and send a TCP FIN packet
5. The client never gets an HTTP response and can't re-transmit it due to how the HTTP protocol works: https://tools.ietf.org/html/rfc2616#section-8.1.4
6. An exception happens on the client as it can't double POST not knowing whenever the server got it or not

You can see a wireshark packet capture of this scenario here:
https://github.com/elastic/apm-agent-python/issues/409#issuecomment-463012525

If in point 4 happens a few nanoseconds sooner than 3 a new TCP connection get established with a SYN packet.

If I changed the read_timeout to 45 seconds the metrics requests always happen over the same TCP connection with no disconnects and FIN packets from the apm server side.

There is an investigation from the amp python agent side:
https://github.com/elastic/apm-agent-python/issues/409

reveil:


simitt:
Thanks @reveil for all the work you put into investigating the issue on the python agent side already.

The default timeout on the server is aligned with all the agents. Before changing a default value, I'd want to investigate if this is only an issue with the python agent or if this also happens with default settings on other agents. Depending on the outcome we should change defaults on the server or rather on the agent.

reveil:
Please note other users also reproduced this issue on older agent versions that don't send metrics at all. So it is possible (but very hard I have not managed it yet) to trigger this with normal requests that just happen to align a POST request with a server TCP FIN packet at a bad time. Metrics make it much more likely because they just happen at same interval.

If on the other hand the server timeout is larger than agent metrics interval the TCP connection does not get closed at all as a result of metrics sending. So instead of causing exceptions metrics can now help mitigate them also in other cases.

So to sum up it does not matter which one you change but server timeout should be larger than agent metrics interval as leaving a TCP connection is technically superior.

simitt:
@reveil I suggest to introduce a dedicated `IdleTimeout` instead and set this to `45s` where the [Server gets created](https://github.com/simitt/apm-server/blob/master/beater/server.go#L45). No need to make this configurable at the moment.

kuisathaverat:
jenkins run the tests please

simitt:
Closing this in favor of #2122.

Issue #1915, opened 2019-02-11
[docs][meta] APM Overview improvements


bmorelli25:
Meta issue for tracking improvements to the APM Overview documentation:

- [x] Create a directory that links to all available APM documentation, repositories, and discuss forum (#1780)
- [x] Improve agent installation table
- [x] Format agent/server compatibility better
- [x] Move release notes to their own pages like we have in the APM Server #1940     
- [x] Change release notes to release highlights elastic/apm-server#1940
- [x] Use Elastic style (elastic/apm-dev/issues/444)
- [ ] ~Add section on observability~ To be added in the redesign
- [x] Discuss what agents do/don't do -- This was added to each Agent's overview
- [x] Use `beta[]` tag instead of "beta" (also applies to the server and Kib: https://github.com/elastic/kibana/issues/32900)

Asciidoctor tagged regions:
- [ ] ~The Overview currently links out to Kibana and Elasticsearch. Explore the idea of using custom include statements to bring documentation from other books into the Overview.~
- [ ] ~Similar to the above point, consider either moving or including the APM UI documentation in the Overview instead of linking to it.~

EDIT: APR, 2020 - We won't be doing any of the crossed items. They will be fixed in the documentation redesign.

Issue #1900, opened 2019-02-07
[docs] Document 7.0 reindexing


bmorelli25:
Related to: https://github.com/elastic/kibana/pull/29845
See also: https://github.com/elastic/kibana/pull/29510

Where?
Breaking change in the [release notes](https://www.elastic.co/guide/en/apm/get-started/master/apm-release-notes.html) with links to more information.

alvarolobato:
Related issue about migration instructions from Kibana APM UI. https://github.com/elastic/kibana/issues/30274

bmorelli25:
Closed in #1940

Issue #1803, opened 2019-01-17
Reflect ECS changes in documentation


jalvz:
* Udpate `context.asciidoc` etc.
* Add field changes to breaking changes (like [beats did](https://www.elastic.co/guide/en/beats/libbeat/master/breaking-changes-7.0.html))
* APM Overview breaking changes should include information on ECS changes and reindexing per https://github.com/elastic/kibana/pull/29845

bmorelli25:
Closed in #1940.

PR #1940, opened 2019-02-15
[Docs] 7.0 Breaking changes, Release highlights, Migration assistant, ECS changes

bmorelli25:
This PR accomplishes the following:
* Blurb on migration assistant / 7.0 reindexing #1900 
* Reflects ECS changes in documentation #1803 
   * Adds `renamed_fields.py` to generate a changed fields table 
* Fixes/reworks APM Overview release highlights/breaking changes #1915
* Adds breaking changes and release notes for 7.0 (see below for details)

---
Here's a general idea of what I'm thinking for release notes and breaking changes moving forward.

1. There was way too much happening on the old [Overview RN page](https://www.elastic.co/guide/en/apm/get-started/6.6/apm-release-notes.html). I moved each release to its own page.

2. I think it makes sense to add a separate breaking changes section to the APM Overview like we have in the APM Server, and other Elastic stack repos.

3. A concern I have (and this isn't anything new), is that we have similar information available in many different places.
   * APM Server: Release notes, Breaking changes
   * APM Overview: Release notes, Breaking changes
   * APM UI: Kibana Release notes, Kibana Breaking changes
   * APM agents: some have their own release note sections

   To "fix" this, I think the APM Overview's _Release notes_ section should be renamed to _Release highlights_. I understand this may seem like a subtle change (and it is), but it's an important distinction to make. We're currently treating this section as a hybrid Notes/Highlights section, which has led to some duplication of content. On the bright side, this is an easy change to make. I can focus on expanding and explaining changes here, then link in/out to the pertinent release notes. Similarly, I will link release notes to this section for more information.

4. I like that beats is listing all of the ECS field changes in their 7.0 breaking changes ([here](https://www.elastic.co/guide/en/beats/libbeat/master/breaking-changes-7.0.html)) and think that we should do the same. I've got a template up on the breaking changes page, just need to fill in the details.

5. Added in placeholders for the next few releases to make the process _slightly_ easier.

Thoughts?
cc @roncohen


roncohen:
sounds good to me @bmorelli25! thanks for looking into this. I guess there's a script from beats to export the changes fields?

changelogs/7.0.asciidoc
```diff
@@ -17,6 +27,7 @@ https://github.com/elastic/apm-server/compare/v6.7.0\...v7.0.0[View commits]
 [float]
 ==== Breaking Changes
 - Move fields in ES to be ECS compliant {pull}1766[1766],{pull}1783[1783],{pull}1813[1813],{pull}1836[1836],{pull}1838[1838],{pull}1844[1844],{pull}1848[1848],{pull}1849[1849],{pull}1863[1863],{pull}1870[1870]
+// ** [More information link would go here (linking to APM overview breaking changes)]
```
simitt: I am not sure if it shouldn't be the other way round, to give a highlight in the apm guide, and then link to the exhaustive changes list in the dedicated repositories. 

Apart from that I like the suggested changes.

bmorelli25:
Good call @roncohen - Ruflin created a `renamed_fields.py` script in https://github.com/elastic/beats/pull/10405. I've tweaked it for our needs and added it to this PR

changelogs/7.0.asciidoc
bmorelli25: @simitt - I might've done a bad job explaining. I think the exhaustive changes should live in the dedicated repos (where they currently live). 

The biggest issue for me was determining whether the link you highlight here should link to the _APM **Overview** breaking changes_ or the _APM **Server** breaking changes_. Ultimately I went we the **Overview** so that users would also be able to easily see all APM breaking changes in one place. I still added a brief note in the Server breaking changes. 

I've added some more commits to this PR that should make my vision clearer. Take a look and let me know your thoughts.

simitt: +1 on your proposal and changes. 
One minor note, I think the link from the apm-server to the general APM breaking changes is broken, at least I got a 404 when trying to follow it.

bmorelli25: Yeah, this is one of the limitations we'll run into when making doc changes in two different books (server & overview) at the same time. Since we can only build one book locally at a time, the new pages in the APM overview will not be built, and the links will 404. I've gone through and manually looked at all the URLs to make sure they look correct, and they do, so I'm hopeful there will be no broken links when this PR is merged 🤞

bmorelli25:
Removed WIP tag and rebased - how do we feel about merging this soon 👍/ 👎?

PR #1941, opened 2019-02-18
Specify minimal events on Intake API and ES level.

### LINKED ISSUE #1812: Document / Specify minimal elasticsearch documents
That is, with all the "required" fields (semantically) and none of the optional ones.
This could be documented as approved fields that run through the libbeat processors.

Thanks @simitt for the input.

cc @sqren

simitt:
Add testfiles with minimal events on Intake API and run integration tests to create minimal ES documents. Some events have more than one minimal spec.

fixes elastic/apm-server#1812

@elastic/apm-ui you could make use of how [minimal events](https://github.com/elastic/apm-server/compare/master...simitt:1812-minimal-events?expand=1#diff-59a7edd512bcf2b1c3e52e2d785017f7) look like in your automated tests (as discussed with @sqren some time ago). Please be aware that there are some event types (error, span) that have multiple possibilities how a minimal event can look like. E.g. an error event either has a `log` or an `exception` and if it has an `exception` then it either has an `exception.type` or an `exception.message`. All possibilities of how a minimal event can look like are covered with the examples.

### DIFF
--- a/beater/integration_test.go
+++ b/beater/integration_test.go
@@ -107,6 +107,7 @@ func TestPublishIntegration(t *testing.T) {
 		{payload: "metricsets.ndjson", name: "Metricsets"},
 		{payload: "spans.ndjson", name: "Spans"},
 		{payload: "transactions.ndjson", name: "Transactions"},
+		{payload: "minimal.ndjson", name: "MinimalEvents"},
 	} {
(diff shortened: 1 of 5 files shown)


sqren:
Thanks, this is great!

PR #1942, opened 2019-02-18
Update config yml to reflect proper default vals.

simitt:
Since the configuration yml is not automatically updated on beats updates (as we differ in some sections), I manually compared and updated the config files, especially the default values.

### DIFF
--- a/_meta/beat.yml
+++ b/_meta/beat.yml
@@ -247,7 +247,7 @@ output.elasticsearch:
   #compression_level: 0
 
   # Optional protocol and basic auth credentials.
-  #protocol: "http"
+  #protocol: "https"
   #username: "elastic"
   #password: "changeme"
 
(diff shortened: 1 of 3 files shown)


simitt:
@graphaelli I looked into the ssl handling in libbeat. It seems by default no ssl/tls config is loaded if nothing is commented in in the config file. So I'd say by default ssl is off. 
As soon as some ssl config is enabled though, the tls config is loaded. 
I changed the yml file to show `ssl.enabled:false`.

Let me know what you think.

graphaelli:
Thanks for digging into it, I'm very surprised by this finding and think it's very confusing for users.  I think `ssl.enabled` should only be relevant where the protocol doesn't indicate that tls is desired (logstash/kafka vs https).  Given that changing that would involve a beats update, I think here we should a comment saying something like `# enable custom SSL settings - ssl.enabled: false`

graphaelli:
I agree with you what you pointed out offline - the default is actually true since `nil/true` both enable this setting.  Revised proposal:

```
# enable custom SSL settings.  Set to false to ignore custom SSL settings for secure communication
# ssl.enabled: true
```

simitt:
Thanks for the suggestion @graphaelli , I've update it.

simitt:
Closing this for now, as it needs some more investigation.

PR #1953, opened 2019-02-25
Clarify read_timeout

roncohen:
Clarifying read_timeout to make it clear that TCP keep-alive is unrelated.

### DIFF
--- a/docs/configuration-process.asciidoc
+++ b/docs/configuration-process.asciidoc
@@ -36,7 +36,7 @@ Defaults to 1048576 Bytes (1 MB).
 [[read_timeout]]
 [float]
 ==== `read_timeout`
-Maximum permitted duration for reading an entire request.
+Maximum permitted duration for reading an entire request. This does not relate to TCP keep-alive, but to HTTP request completion time.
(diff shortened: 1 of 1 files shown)


roncohen:
Will forward port on approval

docs/configuration-process.asciidoc
```diff
@@ -36,7 +36,7 @@ Defaults to 1048576 Bytes (1 MB).
 [[read_timeout]]
 [float]
 ==== `read_timeout`
-Maximum permitted duration for reading an entire request.
+Maximum permitted duration for reading an entire request. This does not relate to TCP keep-alive, but to HTTP request completion time.
```
roncohen: as it turns out, ReadTimeout actually does related to keep-alive if "IdleTimeout" isn't set, which we don't

roncohen:
needs some more appropriate wording or potentially letting people set "IdleTimeout". I'm going to close this suggestion for now

simitt:
Just wanted to mention that this does play together. That's probably also why https://github.com/elastic/apm-server/pull/1937 can happen.

PR #1954, opened 2019-02-25
[Docs] Backport docs to 7.0

bmorelli25:
Backports the following commits to 7.0:

* [Docs] 7.0 Breaking changes, Release highlights, Migration assistant, ECS changes (#1940)
* [docs] Add apm_user role (#1916)
* docs: add opentracing bridge (#1887)
* [docs] make update-beats-docs (#1958)

PR #1958, opened 2019-02-25
[docs] make update-beats-docs

bmorelli25:
Updating Beats documentation. Most of these are small grammar or formatting changes.

I'm curious about the new roles added in the table to the `securing-beats.asciidoc` page. Are these relevant to APM Server:

https://github.com/elastic/apm-server/blob/30ba8f0d9227d331364ee707966c183f35b994bd/docs/copied-from-beats/security/securing-beats.asciidoc#L55-L84

If someone could take a look when they have time, that'd be 🔥

I'd like to backport these to `7.0` (#1954) and `7.x` (#1955)

### DIFF
--- a/docs/copied-from-beats/command-reference.asciidoc
+++ b/docs/copied-from-beats/command-reference.asciidoc
@@ -36,7 +36,7 @@ ifdef::has_ml_jobs[]
 endif::[]
 
 ifdef::no_dashboards[]
-:setup-command-short-desc: Sets up the initial environment, including the ES index template.
+:setup-command-short-desc: Sets up the initial environment, including the ES index template
 endif::no_dashboards[]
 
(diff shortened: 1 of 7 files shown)


graphaelli:
The `_security` url updates are 👍.

The "load dashboard" and "ilm" roles don't make sense for apm-server.  The read indices is covered by the built-in apm_user role.  The remaining entries, including the dashboard-viewing role, make sense to me me to keep.

bmorelli25:
Great, thanks @graphaelli! We had almost everything correct then. I re-added `View APM Server dashboards in Kibana`, and changed `Read indices created by APM Server` to the new `apm_user` role. Here's what that page renders as now:
screen shot 2019-02-25 at 2 49 43 pm

PR #1961, opened 2019-02-27
Add config option to allow for experimental fields.

simitt:
For easier enabling and indexing experimental features during development, agent developers can now make use of a newly added config option `apm-server.environment`. When set to `experimental` arbitray additional information can be set up within `context.experimental` that gets dynamically indexed.

How can this be used?
In `apm-server.yml` set the flag `apm-server.environment: experimental` and send up additional information nested within `context.experimental` for spans, transactions or errors. 
In a follow up we can enable this by default for the integration testing. 

**Do not use this in production environment!** This is solely added for the purpose of enabling building POCs and trying out different approaches during development. 

cc @felixbarny


simitt:
related https://github.com/elastic/kibana/pull/32185

simitt:
Not sure why it is not picking up the PR [elastic/kibana#32185](https://github.com/elastic/kibana/pull/32185) in the test.

simitt:
@graphaelli the PR in Kibana ensuring that those index pattern changes would also be merged to the Index Pattern, let's other test runs fail, as they also test against this PR now. I'd appreciate a review on this PR to make progress with it and get our CI back to green.

beater/config.go
```diff
@@ -48,6 +49,7 @@ type Config struct {
 	SelfInstrumentation *InstrumentationConfig `config:"instrumentation"`
 	RumConfig           *rumConfig             `config:"rum"`
 	Register            *registerConfig        `config:"register"`
+	Environment         Environment            `config:"environment"`
```
graphaelli: Can we avoid the term `environment` since it's so overloaded?  How about `mode`?  Are there other envs we're considering or could `experimental` be top level instead?

simitt:
@graphaelli thanks for the input. I changed the contract for decoding to also pass in a config option, and now we can immediately drop any left overs in context if they should not be stored.

tests/system/config/certs/ca.crt.pem
```diff
@@ -0,0 +1,24 @@
+-----BEGIN CERTIFICATE-----
```
graphaelli: supposed to be here?

beater/config.go
```diff
@@ -110,6 +112,23 @@ type InstrumentationConfig struct {
...
 
+//Environment enumerates the APM Server env
+type Mode uint8
+
+const (
+	ModeProduction Mode = iota
+	ModeExperimental
+)
+
+func (env *Mode) Unpack(s string) error {
```
graphaelli: missed one rename

beater/route_config.go
```diff
@@ -22,6 +22,8 @@ import (
 	"net/http"
 	"regexp"
 
+	"github.com/elastic/apm-server/model"
```
graphaelli: since I'm already nitpicking, combine with other apm-server imports

model/context.go
```diff
@@ -112,15 +113,20 @@ func DecodeContext(input interface{}, err error) (*Context, error) {
...
 		Url:     url,
 		Labels:  labels,
 		Page:    page,
 		Custom:  custom,
 		User:    user,
 		Service: service,
-	}, err
+	}
+
+	if experimental, set := ctxInp["experimental"]; set {
```
graphaelli: Does this only need to be set if exists / Can't this be moved into the literal above?

model/error/event.go
```diff
@@ -142,6 +141,10 @@ func DecodeEvent(input interface{}, err error) (transform.Transformable, error)
 		TransactionType:    decoder.StringPtr(raw, "type", "transaction"),
 	}
 
+	if cfg.Experimental {
```
graphaelli: Did you consider pushing this decision to include `experimental` in the context all the way down to `DecodeContext`?  That would at least unify the decision making for transaction and errors but doesn't help for spans.

model/metricset/event.go
```diff
@@ -22,6 +22,8 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/elastic/apm-server/model"
```
graphaelli: again nitpick with combining imports from apm-server

simitt:
@graphaelli can I get another round of review on this?

PR #1966, opened 2019-02-28
Setting user_agent as root key for all events, and canonicalize http headers. 

### LINKED ISSUE #1960: Move user_agent information derived by user agent pipeline to root key.
Currently when applying the user agent pipeline, the user agent information sent by backend agents is handled differently than the user agent information derived from the RUM agent. In both cases the user agent information is retrieved from the headers. 

For the RUM agent the information is directly fetched from the http request from the agent to the server when enriching the events with additional meta data. This information is then stored as root level key `user_agent` when applying the user agent pipeline.

Backend agents send the user agent information as part of the `context.request.headers` information. Currently applying the user agent pipeline leads to store the information under `context.request.headers.user_agent`. The `headers` should not be manipulated by the agents or the server, but reflect the original headers monitored by the agents. 

In both cases the `user_agent` information should end up in the same place and be indexed, when applying the user agent pipeline.

simitt:
Copies user-agent information from `context.request.headers` information
to root level user_agent key, and apply ua pipeline on this field.

fixes elastic/apm-server#1960

This should be backported as a bugfix to 7.x and 7.0 branches.


simitt:
@watson @felixbarny @jahtalab after some offline discussions I adapted the current handling for enabling user-agent pipelines, to ensure the user-agent information is always indexed within the top level key `user_agent`.

model/context.go
```diff
@@ -150,6 +162,18 @@ func (http *Http) Fields() common.MapStr {
 	return fields
 }
 
+func (http *Http) UserAgent() *string {
+	if http == nil || http.Request == nil || http.Request.Headers == nil {
+		return nil
+	}
+	for _, ua := range []string{"user-agent", "User-Agent"} {
```
graphaelli: I'm a bit torn on whether this is good enough or if we should use the same approach the [stdlib does](https://golang.org/src/net/http/header.go?s=1410:1448#L36).  Another option is to decode headers specially and pull the user agent one out while we're decoding anyway, thoughts?

simitt: I like the ideas, investigating a bit.

graphaelli: I really like where you've taken this.  One side effect of this change is that all headers are canonicalized where they weren't before (right?).  I think that's a great change (appearance and opt-in searchability wise should a user choose to index any of these) but should be documented - can we add `http.request.headers` to the fields doc?

 Also, what do you think about having the UI display these specially?  As a user, I'd expect single valued headers not to be displayed as lists.

image

simitt: If I understand your suggestion regarding showing the headers right, then this would involve work from @elastic/apm-ui. 

I think it is ok to label canonicalizing headers as a bugfix. However, I was thinking if we also need to address this in the migration script? From my perspective, we don't have to, as the data would still be readable from the UI. Would appreciate your thoughts on this. 

I also added the restrictions on the Intake API to the `context.response.headers` as it belongs to both.

graphaelli: > If I understand your suggestion regarding showing the headers right, then this would involve work from @elastic/apm-ui.

Yes, exactly.  This work could follow in any future release.

> I think it is ok to label canonicalizing headers as a bugfix.  However, I was thinking if we also need to address this in the migration script?  From my perspective, we don't have to

Agreed on bugfix and also that the migration script does not need (or even could, well) to handle this.

sqren: This sound good to me. @formgeist do you have any opinion on this?

formgeist: The values would be displayed as cells when we merge https://github.com/elastic/kibana/pull/32199, correct? I think that's the right way to go.

simitt: Currently they would be displayed as Lists yes. I second @graphaelli 's suggestion to better display as 
```
Array: foo
Array: bar
Array: baz
User-Agent: Mozilla Chrome Edge
```
I don't see this urgent, but rather as an UI improvement.

sqren: @formgeist This is how it looks after the change we did:

screenshot 2019-03-05 at 13 22 59

This is what they want:

screenshot 2019-03-05 at 13 25 13

formgeist: Right, I didn't see it was another nested level. I reckon we should allow for that, it's nicer presentation.

PR #1971, opened 2019-02-28
Update versions to strings in dashboards.

simitt:
Related to https://github.com/elastic/kibana/pull/30036


simitt:
For the one failing test we first need to get https://github.com/elastic/beats/pull/10996 merged and then do another round of beats update.

PR #1975, opened 2019-03-05
[7.x] Update beats framework to 70d813d

graphaelli:


simitt:
This adds new fields to the index pattern, so you also need to update the Kibana Index pattern (again).

PR #1985, opened 2019-03-12
[6.7] Remove ilm-policy from export and setup cmd.

### LINKED ISSUE #1983: Remove `ilm-policy` from export cmd as it is not suported
When running `apm-server export --help` `ilm-policy` should not show up as it is not supported in 7.0. Running `apm-server export ilm-policy` should redirect to the help page.

simitt:
fixes elastic/apm-server#1983

In 6.6 beats introduced ILM as beta feature. APM Server does not document, nor actively support it, but inherits the logic. This means, someone _could_ configure to run the server with ILM enabled and use the `setup` and `export` cmds for `ilm-policy`. This is not a great experience, for several reasons, one of them is that the implemented beta logic does not work well with current APM index strategies. For 6.x there is currently no way of restricting a users configuration, but we can remove the `ilm-policy` from the `setup` and `export` cmd.

Proposal:
Keep the logic unchanged for 6.6 as it is already released and ILM is in _beta_ in beats, and not documented in APM Server.
Remove the `setup` and `export` cmd for `ilm-policy` in 6.7.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -72,4 +72,24 @@ func init() {
 			DefaultUsername: "apm_system",
 		},
 	})
+
+	// remove ilm-policy from export commands
+	for _, cmd := range RootCmd.ExportCmd.Commands() {
+		if cmd.Name() == "ilm-policy" {
+			RootCmd.ExportCmd.RemoveCommand(cmd)
+		}
+	}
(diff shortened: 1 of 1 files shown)


simitt:
Failing test is the same as currently failing in 6.7 (related to index pattern).

PR #1995, opened 2019-03-12
complete go 1.10.8 update

graphaelli:


graphaelli:
will follow up with changelog if a 6.6.3 release happens

PR #1996, opened 2019-03-13
[6.7] Update beats framework to 8844f10

### LINKED ISSUE #1980: Documenting non indexed fields by adding to fields.yml
Follow up on https://github.com/elastic/apm-server/pull/1961. 

In 6.6 fields that are not indexed in ES have been added to the `fields.yml` to document also non-indexed fields. [Documentation](https://www.elastic.co/guide/en/apm/server/6.6/exported-fields-apm.html#_context_fields) shows a _Field is not indexed_ if `index: false` is set in the base yml. For objects and groups this field is missing, as only `enabled: false` is set. 

Ensure to forward port documenting the fields to 6.7 and 7.x

graphaelli:
fixes #1980


graphaelli:
jenkins, test this please

PR #1999, opened 2019-03-15
Fail setup dashboard sooner

### LINKED ISSUE #1997: bad request loading dashboards in 7.0
While dashboard loading has been deprecated since 6.4, `setup.dashboards.enabled=true` still triggers loading.  https://github.com/elastic/apm-server/issues/1905 worked around part of this by adding empty dashboards.  Unfortunately, this errors when apm-server tries to load the empty dashboards:

```
  error loading .../kibana/7/dashboard/apm-dashboards.json: returned 400 to import file: <nil>. Response: {"statusCode":400,"error":"Bad Request","message":"request body is required: [parse_exception] request body is required"}
```

This may unnecessarily trip people up when upgrading from 6.x to 7.0.

graphaelli:
closes #1997 

```
$ ./apm-server test config -E setup.dashboards.enabled=true
Exiting: setting 'setup.dashboards' has been removed

$ ./apm-server test config -E setup.dashboards.enabled=false -e
...
2019-03-15T16:42:28.306-0400	WARN	[beater]	beater/beater.go:72	setting 'setup.dashboards' has been removed
```

related to #1905

### DIFF
--- a/beater/beater.go
+++ b/beater/beater.go
@@ -18,6 +18,7 @@
 package beater
 
 import (
+	"errors"
 	"net"
 	"net/http"
 	"net/url"
@@ -33,6 +34,7 @@ import (
 	"github.com/elastic/apm-server/pipelistener"
 	"github.com/elastic/apm-server/publish"
 	"github.com/elastic/beats/libbeat/beat"
+	"github.com/elastic/beats/libbeat/cfgfile"
 	"github.com/elastic/beats/libbeat/common"
 	"github.com/elastic/beats/libbeat/logp"
 	"github.com/elastic/beats/libbeat/outputs/elasticsearch"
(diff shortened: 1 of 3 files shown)


beater/beater.go
```diff
@@ -46,9 +48,40 @@ type beater struct {
...
+	setupDashboardRemoved = errors.New("setting 'setup.dashboards' has been removed")
+)
+
+// checkConfig verifies the global configuration doesn't use unsupported settings
+func checkConfig(logger *logp.Logger) error {
+	cfg, err := cfgfile.Load("", nil)
+	if err != nil {
+		// responsibility for failing to load configuration lies elsewhere
+		// this is not reachable after going through normal beat creation
+		return nil
```
jalvz: if it is not reachable, maybe we could do without this `if` block. 
but even the case `err != nil` is covered already in L68 as `Unpack` would fail?

graphaelli: Yeah, L68 would fail but I don't think we should rely on the behavior of `cfg` in case of an error.  This is reachable during tests so we definitely should keep the check.

graphaelli:
package tests ran without issue

PR #2007, opened 2019-03-18
Fail setup dashboard sooner (#1999)

graphaelli:
Backports the following commits to 7.0:

* Fail setup dashboard sooner (#1999)


graphaelli:
@elastic/apm-server should we just completely refuse to start if this is set in the next major?

jalvz:
SGTM

PR #2010, opened 2019-03-18
enable pipeline registration for setup --pipelines

### LINKED ISSUE #2008: apm-server setup --pipelines doesn't load pipelines
Discovered by @jalvz - On 7.0:

This doesn't load pipelines:
```
$ apm-server setup --pipelines -e
...
2019-03-18T14:03:37.315-0400	INFO	instance/beat.go:276	Setup Beat: apm-server; Version: 7.0.0
2019-03-18T14:03:37.317-0400	INFO	[beater]	beater/beater.go:114	No pipeline callback registered
```

But this does:
```
$ apm-server setup --pipelines -E apm-server.register.ingest.pipeline.enabled=true -e
...
2019-03-18T14:06:11.762-0400	INFO	[pipelines]	pipeline/register.go:49	Pipeline successfully registered: apm_user_agent
2019-03-18T14:06:11.762-0400	INFO	[pipelines]	pipeline/register.go:54	Registered Ingest Pipelines successfully.
```

I'm not sure if this was discussed / is working as intended - it does not feel like a good user experience.

graphaelli:
setup pipelines if explicitly directed to or doing `setup --pipelines` and the config is not set at all

fixes #2008

### DIFF
--- a/beater/beater.go
+++ b/beater/beater.go
@@ -75,7 +75,11 @@ func New(b *beat.Beat, ucfg *common.Config) (beat.Beater, error) {
 		logger:  logger,
 	}
 
-	if isElasticsearchOutput(b) && beaterConfig.Register.Ingest.Pipeline.isEnabled() {
+	// setup pipelines if explicitly directed to or setup --pipelines and config is not set at all
+	shouldSetupPipelines := beaterConfig.Register.Ingest.Pipeline.isEnabled() ||
+		(b.InSetupCmd && beaterConfig.Register.Ingest.Pipeline.Enabled == nil)
+
+	if isElasticsearchOutput(b) && shouldSetupPipelines {
 		logger.Info("Registering pipeline callback.")
 		err := bt.registerPipelineCallback(b)
 		if err != nil {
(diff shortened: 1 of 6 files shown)


beater/beater.go
```diff
@@ -75,7 +75,11 @@ func New(b *beat.Beat, ucfg *common.Config) (beat.Beater, error) {
 		logger:  logger,
 	}
 
-	if isElasticsearchOutput(b) && beaterConfig.Register.Ingest.Pipeline.isEnabled() {
+	// setup pipelines if explicitly directed to or setup --pipelines and config is not set at all
+	shouldSetupPipelines := beaterConfig.Register.Ingest.Pipeline.isEnabled() ||
+		(b.InSetupCmd && beaterConfig.Register.Ingest.Pipeline.Enabled == nil)
```
jalvz: no one will ever do `setup --pipelines -E register.ingest.pipeline.enabled=False`, but in that case behaviour here would be wrong, right?

graphaelli: In that case `beaterConfig.Register.Ingest.Pipeline.Enabled` would not be `nil` and pipelines would not be loaded - would you expect them to be loaded no matter what instead?  There is a test (in python) for this case.

jalvz: right, never mind my boolean brain... :man_facepalming:

graphaelli:
I think this is a bug and should be backported, will open that shortly

PR #2023, opened 2019-03-19
note auto index creation troubleshooting

graphaelli:
image


graphaelli:
As far as convenient as this has always been true

PR #2028, opened 2019-03-20
[7.0] Update beats framework to 4920cd6

graphaelli:


vendor/github.com/elastic/beats/libbeat/cmd/instance/beat.go
```diff
@@ -110,6 +110,9 @@ type beatConfig struct {
 	// elastic stack 'setup' configurations
 	Dashboards *common.Config `config:"setup.dashboards"`
 	Kibana     *common.Config `config:"setup.kibana"`
+
+	// Migration config to migration from 6 to 7
+	Migration *common.Config `config:"migration.6_to_7"`
```
jalvz: does this affect us?

_beats/libbeat/docs/version.asciidoc
```diff
@@ -6,3 +6,4 @@
 :docker: 1.12
 :docker-compose: 1.11
 :branch: 7.0
+:major-version: 7.x
```
jalvz: do you know if we need to do something with this?

graphaelli: I don't this so, we don't use this reference as far as I know.

vendor/github.com/elastic/beats/libbeat/cmd/instance/beat.go
graphaelli: It can minimally - setting `migration.6_to_7.enabled=true` triggers creation of 14 non-conflicting aliases.  We won't document this as there isn't a known reason to do this right now.

Issue #1545, opened 2018-11-14
document ILM in APM Server


simitt:

graphaelli:
# An example ILM configuration

The following configuration will create a new index for spans every day or for every 50gb of spans, which comes first.  Each index will be retained for 10 days.  Additional configuration for error, metricset, transaction retention should be straightforward - policies can even be reused if there is overlap, say between errors and transactions.

To follow along locally, adjust this setting first, as the default ILM interval is only every 10 minutes.  This **value** should not be part of the official documentation, but probably good to link to https://www.elastic.co/guide/en/elasticsearch/reference/6.x//using-policies-rollover.html#using-policies-rollover where it is described.

## Step 0. Make ILM run much more often than the 10 minute default
```
PUT /_cluster/settings
{
  "transient": {
    "indices.lifecycle.poll_interval": "10s"
  }
}
```

## Step 1. Setup the apm-$version-* index template
```
./apm-server setup --template
```

## Step 2. Set up a policy for spans

To follow along locally, add a `"max_docs": 50` to the rollover actions to see them trigger without waiting very long.

```
PUT _ilm/policy/apm_span_policy
{
  "policy": {
    "phases": {
      "hot": {
        "actions": {
          "rollover": {
            "max_age": "1d",
            "max_size": "50gb"
          }
        }
      },
      "delete": {
        "min_age": "10d",
        "actions": {
          "delete": {}
        }
      }
    }
  }
}
```

## 3. Set up the ILM index template
```
PUT _template/apm-6.6.0-span-ilm
{
  "order": 2,
  "index_patterns": ["apm-6.6.0-span-*"],
  "settings": {
    "number_of_shards": 1,
    "number_of_replicas": 1,
    "index.lifecycle.name": "apm_span_policy",
    "index.lifecycle.rollover_alias": "apm-6.6.0-span"
  }
}
```

## 4. Create the index + alias
```
PUT apm-6.6.0-span-000001
{
  "aliases": {
    "apm-6.6.0-span":{
      "is_write_index": true 
    }
  }
}
```

## 4b. Verify ILM index template was applied
```
GET apm-*-span/_settings
```

## 5. Start apm-server with modified configuration

Trim off the date template from each branch, so that apm-server is always writing to the same place per event type.  That place is the `is_write_index` alias created previously.

```
--- apm-server.yml
+++ apm-server.ilm.yml
@@ -410,23 +410,23 @@
       when.contains:
         processor.event: "sourcemap"
 
[...]

PR #2033, opened 2019-03-20
[docs] Add manual ILM instructions

bmorelli25:
Here's my first crack at manual ILM setup for APM Server. Closes https://github.com/elastic/apm-server/issues/1545.

Huge shoutout to @graphaelli for creating an awesome [walkthrough](https://github.com/elastic/apm-server/issues/1545#issuecomment-446817453) of the process.

Note: The asciidoc for this change is really ugly due to our `//console` tags not jiving with ordered lists.

### DIFF
--- a/docs/ilm-setup.asciidoc
+++ b/docs/ilm-setup.asciidoc
@@ -1,11 +1,16 @@
+[role="xpack"]
+[[manual-ilm-setup]]
 == Manual index lifecycle management
 
-APM Server allows users to manually set up an index lifecylce managment (ILM) policy.
-This setup is made possible because of Elasticsearch's
-{ref}/using-policies-rollover.html#using-policies-rollover[index rollover] action.
+APM Server can take advantage of the index lifecycle management (ILM) capabilities of Elasticsearch.
+ILM enables you to automate how you want to manage your indices over time.
+You can base actions on factors such as shard size and performance requirements. 
 
-Follow the guide below to set up a custom ILM policy for span indices.
+The guide below will help you set up a custom ILM policy for span indices.
 You can repeat the actions for any other indices.
+
+NOTE: If you're migrating from an existing setup,
+any indices present before ILM was configured will need to be managed manually.
 
 . Set up the default index template
 +
@@ -22,11 +27,13 @@ This is accomplished by running the <<setup-command,`setup --template`>> command
 
 . Create a policy for spans.
 +
-Rollover actions are triggered when one or more criteria are met.
-In this example, we're creating a policy named `apm_span_policy`.
+Index lifecycle management will manage an index based on its defined policy.
+Policies only need to be created once, and will persist through version upgrades.
+Let's create a policy named `apm_span_policy`.
 +
-This policy rolls data over to a new index after _1 day_, or _50gb_, whichever comes first.
-It also tells the old indexes to delete after _10 days_.
+This policy defines two rollover criteria: `"max_age": "1d"`, and `"max_size": "50gb"`.
+When one or more of these criteria are met, the policy rolls data over into a new index.
+`apm_span_policy` also tells the old indexes to delete after _10 days_.
 All of these values can be customized to your specific needs.
 +
 --
@@ -66,6 +73,9 @@ To use the index lifecycle management policy created in the previous step,
 you need to specify it in the index template used to create the indices.
 The following template associates `apm_span_policy` with indices created from the +apm-{stack-version}-span-ilm+ template.
 +
+NOTE: Because we're utilizing the current stack-version ({stack-version}) in this step,
+this action will need to be performed as a part of each version upgrade.
++
 --
 ["source","js",subs="attributes"]
 -----------------------
@@ -74,8 +84,6 @@ PUT _template/apm-{stack-version}-span-ilm
   "order": 2,
   "index_patterns": ["apm-{stack-version}-span-*"], <1>
   "settings": {
-    "number_of_shards": 1,
-    "number_of_replicas": 1,
     "index.lifecycle.name": "apm_span_policy", <2>
     "index.lifecycle.rollover_alias": "apm-{stack-version}-span"
   }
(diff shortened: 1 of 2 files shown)


docs/ilm-setup.asciidoc
```diff
@@ -0,0 +1,189 @@
...
+    - index: "apm-{stack-version}-span"
+      when.contains:
+        processor.event: "span"
+-----------------------
+// CONSOLE
+--
+
+. Start apm-server.
++
+Your ILM policy should now be up and running!
```
bmorelli25: @graphaelli - am I using the term _ILM policy_ correctly here? Would _ILM configuration_ be better? Something else?

docs/ilm-setup.asciidoc
```diff
@@ -0,0 +1,189 @@
+== Manual index lifecycle management
+
```
graphaelli: How about an intro sentence about automatically cleaning / limiting unbounded storage consumption first?

docs/ilm-setup.asciidoc
graphaelli: Yeah, ILM configuration makes more sense to me.

docs/ilm-setup.asciidoc
```diff
@@ -0,0 +1,189 @@
...
++
+--
+["source","js",subs="attributes"]
+-----------------------
+PUT _template/apm-{stack-version}-span-ilm
+{
+  "order": 2,
+  "index_patterns": ["apm-{stack-version}-span-*"], <1>
+  "settings": {
+    "number_of_shards": 1,
```
graphaelli: Let's remove `number_of_shards` and `number_of_replicas` and leave those to the index template loaded by apm-server

docs/ilm-setup.asciidoc
```diff
@@ -0,0 +1,189 @@
...
+      when.contains:
+        processor.event: "span"
+-----------------------
+// CONSOLE
+--
+
+. Start apm-server.
++
+Your ILM policy should now be up and running!
+
```
graphaelli: How about a note here about Migrating from existing setup to ILM configuration?  Namely, that any indices present before ILM was configured will need to be managed manually.

docs/ilm-setup.asciidoc
```diff
@@ -0,0 +1,189 @@
...
+    }
+  }
+}
+-----------------------
+--
+
+. Modify APM Server's configuration.
++
+You'll need to modify the default index configuration in <<apm-server-configuration,`apm-server.yml`>>.
+Trim off the date template from each index you are setting up ILM for,
```
graphaelli: Users will have to configure each index even if not using ILM - that is, this overwrites the default rather than being merged.  For example, if only spans were indexed into `apm-{stack-version}-span`. all other events would end up in `apm-{stack-version}-%{+yyyy.MM.dd}` since that's the default.  I suggest we include something about this here.

docs/ilm-setup.asciidoc
```diff
@@ -0,0 +1,189 @@
...
+
+. Create a policy for spans.
++
+Rollover actions are triggered when one or more criteria are met.
+In this example, we're creating a policy named `apm_span_policy`.
++
+This policy rolls data over to a new index after _1 day_, or _50gb_, whichever comes first.
+It also tells the old indexes to delete after _10 days_.
+All of these values can be customized to your specific needs.
++
```
graphaelli: For symmetry with the other sections, we could point that this only needs to be done once, rather than as part of each version upgrade.

docs/ilm-setup.asciidoc
```diff
@@ -0,0 +1,189 @@
...
+// CONSOLE
+<1> This template applies to all indices with the prefix +apm-{stack-version}-span-+
+<2> Associates `apm_span_policy` with all indices created with this template
+--
+
+. Create the index and alias.
++
+Now we can create the first index: +apm-{stack-version}-span-000001+.
++
+--
```
graphaelli: Let's mention that each of the commands including `{stack-version}` need to be performed as part of each upgrade.

docs/ilm-setup.asciidoc
```diff
@@ -0,0 +1,189 @@
...
+        "version" : {
+          "created" : "7000099"
+        }
+      }
+    }
+  }
+}
+-----------------------
+--
+
```
graphaelli: We could mention that the configuration can be verified / tweaked through the kibana ui via management.

docs/ilm-setup.asciidoc
graphaelli: Do we need to point out that this requires xpack?

docs/ilm-setup.asciidoc
bmorelli25: Commented in slack as well. Wondering if this means this would be a better code example:
```
  indices:
    - index: "apm-8.0.0-span"
      when.contains:
        processor.event: "span"

    - index: "apm-%{[observer.version]}-sourcemap"
      when.contains:
        processor.event: "sourcemap"
    
    - index: "apm-%{[observer.version]}-error-%{+yyyy.MM.dd}"
      when.contains:
        processor.event: "error"
    
    - index: "apm-%{[observer.version]}-transaction-%{+yyyy.MM.dd}"
      when.contains:
        processor.event: "transaction"
    
    - index: "apm-%{[observer.version]}-metric-%{+yyyy.MM.dd}"
      when.contains:
        processor.event: "metric"
    
    - index: "apm-%{[observer.version]}-onboarding-%{+yyyy.MM.dd}"
      when.contains:
        processor.event: "onboarding"
```

docs/ilm-setup.asciidoc
bmorelli25: Added at the top

bmorelli25:
Great feedback, thank you!

docs/ilm-setup.asciidoc
graphaelli: Yes, that's more what I had in mind, though we can still expand observer.version for span.  We want to encourage users to configure ILM for all indices - perhaps just setting one of those to non-ILM would be best?  I would use transaction in that case.

bmorelli25: Alright, I think I've come up with a good fix for this. I added a step to "repeat" the first four actions (policy creation, index template, index/alias, and verification). I then included all indices in apm-server.yml in their ILM form

bmorelli25: Screen Shot 2019-03-21 at 4 07 04 PM

graphaelli: Excellent solution!

PR #2039, opened 2019-03-25
Add labels to metadata

### LINKED ISSUE #2037: Add labels to metadata
As a result of https://github.com/elastic/apm/issues/48 we need to add the option to provide labels on in the metadata. 

Labels in metadata should be added automatically to all transactions, metrics and errors but not on spans.

jalvz:
Metadata labels are injected in transactions, errors and metricsets.

Fixes elastic/apm-server#2037


model/context.go
```diff
@@ -92,7 +92,7 @@ func DecodeContext(input interface{}, cfg Config, err error) (*Context, error) {
 	}
 	raw, ok := input.(map[string]interface{})
 	if !ok {
-		return nil, errors.New("Invalid type for fetching Context fields")
+		return nil, errors.New("invalid type for fetching Context fields")
```
graphaelli: it would help review a lot if these changes were made in a separate commit, or even a separate PR, as it's not clear initially if it's directly related to the goals fo the changeset.

tests/fields.go
```diff
@@ -85,16 +85,24 @@ func (ps *ProcessorSetup) EventFieldsMappedToTemplateFields(t *testing.T, eventF
 	for _, val := range eventFields.Array() {
 		var f = val.(string)
 		for _, m := range mappings {
-			if strings.HasPrefix(f, m.Template) {
-				f = strings.Replace(f, m.Template, m.Mapping, -1)
+			template := m.Template
```
graphaelli: It took me awhile to come around but I like this, certainly better than the alternative of not checking that the field is present at all.

tests/system/test_integration.py
```diff
@@ -129,7 +129,7 @@ def check_docs(self, approved, received, doc_type):
                             assert v["version"].startswith(str(v["version_major"]) + ".")
                             continue
 
-                        self.assert_docs(v, appr[k])
+                        self.assert_docs(v, appr.get(k))
```
graphaelli: `k` must be in `appr` right (except while fixing tests)?  Could this introduce a situation where `v is None` when it shouldn't be without triggering the test to fail?

tests/fields.go
jalvz: it took me quite a while to figure out why the test was even failing... i also think is better checking than not, but implementation is awfully `if`-fy. If you have any better ideas, happy to hear...

model/context.go
jalvz: true, sorry for that :ok_hand:

tests/system/test_integration.py
jalvz: I changed it because if the test fails, the failure is an exception rather than a diff mismatch.
I actually don't know if that could happen, so I'll change it back

tests/fields.go
graphaelli: This is the alternative I had considered, removing labels.* from the check altogether.  I like your solution better.  We have discussed overhauling these tests because of this kind of thing, I'm happy to leave this as is and tackle it later during that effort.

```patch
diff --git a/processor/stream/package_tests/metadata_attrs_test.go b/processor/stream/package_tests/metadata_attrs_test.go
index 0c45338b..2876a9aa 100644
--- a/processor/stream/package_tests/metadata_attrs_test.go
+++ b/processor/stream/package_tests/metadata_attrs_test.go
@@ -92,7 +92,7 @@ func getMetadataEventAttrs(t *testing.T, prefix string) *tests.Set {
        contextMetadata := metadata["metadata"]
 
        eventFields := tests.NewSet()
-       tests.FlattenMapStr(contextMetadata, prefix, nil, eventFields)
+       tests.FlattenMapStr(contextMetadata, prefix, tests.NewSet(tests.Group("labels")), eventFields)
        t.Logf("Event fields: %s", eventFields)
        return eventFields
 }
@@ -100,7 +100,6 @@ func getMetadataEventAttrs(t *testing.T, prefix string) *tests.Set {
 func TestMetadataPayloadAttrsMatchFields(t *testing.T) {
        setup := metadataProcSetup()
        eventFields := getMetadataEventAttrs(t, "")
-
        var mappingFields = []tests.FieldTemplateMapping{
                {Template: "system.container.", Mapping: "container."},             // move system.container.*
                {Template: "system.container", Mapping: ""},                        // delete system.container
@@ -115,7 +114,6 @@ func TestMetadataPayloadAttrsMatchFields(t *testing.T) {
                {Template: "service.agent", Mapping: "agent"},
                {Template: "user.username", Mapping: "user.name"},
                {Template: "process.argv", Mapping: "process.args"},
-               {Template: "labels.*", Mapping: "labels"},
        }
        setup.EventFieldsMappedToTemplateFields(t, eventFields, mappingFields)
 }
@@ -123,8 +121,8 @@ func TestMetadataPayloadAttrsMatchFields(t *testing.T) {
 func TestMetadataPayloadMatchJsonSchema(t *testing.T) {
        metadataProcSetup().AttrsMatchJsonSchema(t,
                getMetadataEventAttrs(t, ""),
-               tests.NewSet(tests.Group("labels")),
                nil,
+               tests.NewSet(tests.Group("labels")),
        )
 }
```

tests/system/test_integration.py
graphaelli: Yeah, that's what I meant by "while fixing tests".  Up to you, but we could keep the `get` change and add an additional assertion about `v` directly.

PR #2043, opened 2019-03-25
[docs] Add deb/rpm running instructions

### LINKED ISSUE #2017: Update running instructions
https://www.elastic.co/guide/en/apm/server/current/setting-up-and-running.html doesn't describe how to run a deb/rpm packaged apm-server.

 Those installations create an `apm-server` user (https://github.com/elastic/apm-server/tree/69f4df1f881bfc1477ff7ce7e80148ea665a7669/packaging/files/linux) so apm-server should be run as follows:

`sudo -u apm-server apm-server -e [FLAGS]`

maybe it also makes sense to document where files live:

yml config file in `/etc/apm-server`
logs in `/var/log/apm-server`
pipeline definition, readme, and such in `/usr/share/apm-server`

Those are standard, but not everyone might be familiar with

bmorelli25:
Fixes #2017.

Note: Decided to link to the directory layout instead of documenting where files live on this page.
Screen Shot 2019-03-25 at 12 10 40 PM

### DIFF
--- a/docs/setting-up-and-running.asciidoc
+++ b/docs/setting-up-and-running.asciidoc
@@ -8,6 +8,7 @@ similar to how you run Elasticsearch.
 You _can_ run it on the same machines as Elasticsearch,
 but this is not recommended,
 as the processes will be competing for resources.
+
 To start APM Server, run:
 
 [source,bash]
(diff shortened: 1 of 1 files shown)


docs/setting-up-and-running.asciidoc
```diff
@@ -24,6 +25,21 @@ You can change the defaults by supplying a different address on the command line
...
+[float]
+[[running-deb-rpm]]
+=== Debian Package / RPM
+
+The Debian package and RPM installations of APM Server create an `apm-server` user.
+To start the APM Server in this case, run:
+
+[source,bash]
+----------------------------------
+sudo -u apm-server apm-server -e [FLAGS]
```
jalvz: I got the example in the issue wrong, the right notation would be 
`sudo -u apm-server apm-server [<argument...>]`
you might want to scan trough docs and make sure we use the notation correctly everywhere, and consistently.

ref: http://docopt.org/

bmorelli25: Fixed. I might ping you tomorrow to have a quick chat about this. Are you saying `[<argument...>]` should replace `[FLAGS]` everywhere in the docs?

bmorelli25:
Thanks as always Juan

PR #2044, opened 2019-03-25
[docs] 6.7 release-state

bmorelli25:
In preparation for the `6.7` release,this PR updates the release state to "released".  I'll merge this when the time is right. 

@graphaelli / @jalvz  - can you verify the `6.7` release notes below look good to go:

https://github.com/elastic/apm-server/blob/56b4ebe0088c9a15ea5eb5f4d491da19f08c1bc6/changelogs/6.7.asciidoc#L6-L22


graphaelli:
Changelog looks accurate, however `removed` should be the same size heading as `added`, it's currently one smaller.

bmorelli25:
Aye, good catch. Fixed!

PR #2047, opened 2019-03-26
bump 6.7.1

graphaelli:


graphaelli:
test failure set to be fixed in #2046

Issue #2021, opened 2019-03-19
Update pipeline config docs


jalvz:
User agent parsing is not distributed as a plugin anymore, but as a ES processor (available by default): https://www.elastic.co/guide/en/elasticsearch/reference/master/user-agent-processor.html

https://www.elastic.co/guide/en/apm/server/7.0/configuring-ingest-node.html should be updated accordingly

PR #2061, opened 2019-03-29
[docs] Update ingest node documentation

bmorelli25:
Closes https://github.com/elastic/apm-server/issues/2021.

Also took this opportunity to clean up/rewrite the ingest node documentation.

### DIFF
--- a/docs/configuring-ingest.asciidoc
+++ b/docs/configuring-ingest.asciidoc
@@ -10,12 +10,13 @@ to pre-process documents before indexing them in Elasticsearch.
 A pipeline is a definition of a series of processors that operate on your data.
 For example, a pipeline can define one processor to remove a field, and another to rename a field.
 Using pipelines involves two steps:
+
 . First, you need to <<register-pipelines,register a pipeline>> in Elasticsearch.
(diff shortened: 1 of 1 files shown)


docs/configuring-ingest.asciidoc
```diff
@@ -56,20 +55,19 @@ For example, consider the following pipeline in a file named `pipeline.json`:
 }
 ------------------------------------------------------------------------------
 
-To register the pipeline run:
+To register this pipeline, run:
 
 [source,shell]
 ------------------------------------------------------------------------------
 curl -H 'Content-Type: application/json' -XPUT 'http://localhost:9200/_ingest/pipeline/test-pipeline' -d @pipeline.json
```
graphaelli: maybe not in this change, but what do you think of changing all of the elasticsearch sample endpoints to `http://elasticsearch:9200` ?

bmorelli25: At some point, yes. Beats still uses localhost > elasticsearch in most places, so I think it makes sense for us to wait until more of the shared docs use elasticsearch.

PR #2076, opened 2019-04-04
[DOCS] Adds tagged region for notable breaking changes

lcawl:
This PR adds a tagged region in the APM Overview > Breaking Changes page (https://www.elastic.co/guide/en/apm/get-started/master/apm-breaking-changes.html), such that content can be re-used in the Installation and Upgrade Guide (https://www.elastic.co/guide/en/elastic-stack/master/elastic-stack-breaking-changes.html).

At least one tagged region must exist to prevent build errors in the Installation and Upgrade Guide, though as is the case now for V8.0.0 it can be an empty tagged region. 

At this time, the Installation and Upgrade Guide for version X only lists the breaking changes specific to that version. Since this APM page contains information about multiple versions, we either need to (1) ensure that only the items that are specific to the appropriate version are tagged as "notable" in each branch, or (2) make the tagged regions version-specific (e.g. tag::notable-v8-breaking-changes[]).  I'm open to whichever implementation is easier for you to maintain.

Related to https://github.com/elastic/docs/pull/788 and https://github.com/elastic/stack-docs/pull/267

### DIFF
--- a/docs/guide/apm-breaking-changes.asciidoc
+++ b/docs/guide/apm-breaking-changes.asciidoc
@@ -10,9 +10,9 @@ Also see <<apm-release-notes>>.
 
 //NOTE: The notable-breaking-changes tagged regions are re-used in the
 //Installation and Upgrade Guide
-// tag::notable-breaking-changes[]
+// tag::notable-v8-breaking-changes[]
 
-// end::notable-breaking-changes[]
+// end::notable-v8-breaking-changes[]
 
 [[breaking-7.0.0-beta1]]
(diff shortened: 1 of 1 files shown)


bmorelli25:
🎉Exciting! Thanks @lcawl.

My vote is for option 2:
> make the tagged regions version-specific (e.g. tag::notable-v8-breaking-changes[])

lcawl:
Thanks for the feedback!  I've updated the name of the tagged region so it's version-specific.

PR #2077, opened 2019-04-04
[7.0]  Update beats framework to aa6953a

graphaelli:
Major changes from libbeat:
* Fix index template always being overwritten - elastic/beats#11671
* Perform Basic license check on Elasticsearch connect - elastic/beats#11296 + elastic/beats#11649


_beats/libbeat/tests/system/beat/beat.py
```diff
@@ -220,6 +220,9 @@ def start_beat(self,
 
     def render_config_template(self, template_name=None,
                                output=None, **kargs):
+
+        print("render config")
```
graphaelli: need to remove this upstream

PR #2095, opened 2019-04-11
[ACM] Agent remote configuration management - Main functionality

jalvz:
- [x] needs changelog
- [x] needs tests
- [x] needs feature docs
- [x] new Kibana config to add in yml file and docs
- [ ] needs sync

#### Notes

- Service environment is optional, and treated like any other value (i.e. a service is unequivocally identified by its name). This is the lowest common denominator, as an agent can always send an extra request for a default:
```
if GET config?service.name=X&service.environment=Z == 404
then GET config?service.name=X
```
ofc we can introduce an option to save the extra round trip.

#### Steps to test

1. Checkout https://github.com/elastic/kibana/pull/36031, and start Elasticsearch and Kibana. 

2. Create the CM index with defined mapping:

```
PUT .apm-cm
{
  "mappings": {
    "properties": {
      "service": {
        "properties": {
          "environment": {
            "type": "keyword",
            "ignore_above": 1024
          },
          "name": {
            "type": "keyword",
            "ignore_above": 1024
          }
        }
      },
      "settings": {
        "properties": {
          "sample_rate": {
            "type": "float"
          }
        }
      },
      "timestamp": {
        "properties": {
          "us": {
            "type": "long"
          }
        }
      }
    }
  }
}
```

3. Index some config docs

```

PUT .apm-cm/_doc/1
{
    "service": { 
      "name": "opbeans-lisp",
      "environment": "lambda"
    },
    "@timestamp" : "2019-03-01T12:12:12",
    "settings": {
      "sampling_rate": 0.1
    }
}

PUT .apm-cm/_doc/2
{
    "service": { 
      "name": "opbeans-fortran",
      "environment": "rocket-launchpad"
    },
    "@timestamp" : "2019-04-01T12:12:12",
    "settings": {
      "sampling_rate": 0.4
    }
}

PUT .apm-cm/_doc/3
{
    "service": { 
      "name": "opbeans-fortran",
      "environment": "rocket-launchpad"
    },
    "@timestamp" : "2019-07-01T12:12:12",
    "settings": {
      "sampling_rate": 0.7
    }
}

PUT .apm-cm/_doc/4
{
    "service": { 
      "name": "opbeans-fortran",
      "environment": "testbed"
    },
    "@timestamp" : "2019-10-01T12:12:12",
    "settings": {
      "sampling_rate": 1.0
    }
}

PUT .apm-cm/_doc/5
{
    "service": { 
      "name": "opbeans-fortran"
    },
    "@timestamp" : "2019-12-01T12:12:12",
    "settings": {
      "sampling_rate": 0.12
    }
}
```

Alternatively, create a config directly in the UI by visiting `/app/apm#/settings/new`.
You should have created some transactions first.

4. Start APM Server with `-E apm-server.kibana.path` properly set and send requests to the config endpoint:
`curl http://localhost:8200/config?service.name=apm-agent-js`
```
{
  "sampling_rate": "0.4"
}
```


jalvz:
`make check-full` fails with "struct field tag `json:settings` not compatible with reflect.StructTag.Get: bad syntax for struct tag value", which I have no clue what is about.

branch compiles and runs fine otherwise

jalvz:
@simitt @graphaelli it would be great if you can have a look

graphaelli:
Took a look and the api and response make sense to me

jalvz:
@graphaelli is there something missing/wrong here preventing merge, other than golint errors?

graphaelli:
I expected we were going to settle on the storage format before merging

jalvz:
Worst case I would just have to remove this function: https://github.com/elastic/apm-server/pull/2095/files#diff-b6b7a562b4bbb8af1d07c8eaf1fda8b0R41, which is trivial. Or am I missing something else?

I'd really avoid PRs open for more than 1 month if possible

beater/agent_config_handler.go
```diff
@@ -0,0 +1,106 @@
...
+			send(nil, http.StatusNotModified)
+
+		case upstreamEtag != "":
+			w.Header().Set("Cache-Control", "max-age=0")
+			w.Header().Set("Etag", upstreamEtag)
+			fallthrough
+
+		default:
+			send(cfg, http.StatusOK)
+		}
```
simitt: nit: would you mind removing the empty lines in the switch/case statement? I find that really hard to read.

beater/beater.go
```diff
@@ -192,7 +194,15 @@ func (bt *beater) Run(b *beat.Beat) error {
 		return nil
 	}
 
-	bt.server = newServer(bt.config, tracer, pub.Send)
+	var kb *kibana.Client
+	if bt.config.Kibana.Enabled() {
+		kb, err = kibana.NewKibanaClient(bt.config.Kibana)
+		if err != nil {
+			bt.logger.Error(err.Error())
```
simitt: Why not return the error here?

With the current logic, `kb` should be set to nil if `err != nil`, to ensure you aren't passing a faulty or half initialized Kibana client on to `newServer`.

beater/route_config.go
```diff
@@ -35,6 +37,8 @@ import (
 var (
 	rootURL = "/"
 
+	agentConfigURL = "/config"
```
simitt: Have you considered nesting under an agent namespace, e.g. `/agent/config`, and also have you thought about versioning? 
In case we are introducing a `config` endpoint to fetch the servers config, would that fit into `/config`?

beater/agent_config_handler.go
```diff
@@ -0,0 +1,106 @@
...
+	"github.com/elastic/apm-server/convert"
+
+	"github.com/pkg/errors"
+
+	"github.com/elastic/beats/libbeat/kibana"
+
+	"github.com/elastic/apm-server/agentcfg"
+)
+
+func agentConfigHandler(client *kibana.Client, secretToken string) http.Handler {
```
simitt: nit: in some places you use `kb *kibana.Client` and here `client *kibana.Client`. It'd make it easier to grasp if we used the same variable names all along (I know we haven't focused on this so far, but newly introduced logic is always a good place to start improving).

beater/route_config.go
jalvz: 1. - I didn't thing about the endpoint name at all. I can change it to anything. @graphaelli are you good with the suggestion? where should the version go? `/agent/v1/config`?
2. - don't understand the question, this endpoint is to expose config, not fetching it - why would I need an endpoint to expose the server config?

beater/beater.go
jalvz: An error will also be returned if config is valid but Kibana is not reachable.
Are you proposing to stop the server, then? I can do that. Although that'd diverge from output behaviour.

_meta/beat.yml
```diff
@@ -155,6 +155,55 @@ apm-server:
...
+
+    # Optional protocol and basic auth credentials.
+    #protocol: "https"
+    #username: "elastic"
+    #password: "changeme"
+
+    # Optional HTTP path
+    #path: ""
+
+    # Use SSL settings for HTTPS. Default is true.
```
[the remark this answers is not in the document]
jalvz: This description is copied verbatim from the Kibana configuration section of a beat, you are saying that I should copy it from the elasticseach section instead?

beater/agent_config_handler.go
jalvz: really spacing hurts your readability more than all lines crammed together without any breathing? 
hmm, ok. i am totally the opposite, then

simitt: I like to use spaces to group things together, in the switch/case block I find it harder to read. But as I said that's a nit, if you prefer otherwise it is fine with me.

beater/beater.go
simitt: ok, that's a good point. Is there any possibility to distinguish between those errors? I agree we should not stop the server when Kibana is not reachable.

beater/route_config.go
simitt: Fetching the current APM Server config might be interesting for the environment in which the server is running (cloud, k8s, ..), but I mainly used it to prove my point. When I have a server exposing an endpoint `config` I would expect it to return the servers config at least if nothing else is requested via query params.

beater/agent_config_handler.go
jalvz: i could argue that context is different, too. a critical function in a big package can benefit from more explicit naming, like
`beater.newMuxer(beaterConfig *Config, client *kibana.Client, report publish.Reporter`

whether a small function in a narrow self-describing package can benefit more from concise naming, like `agentcfg.Fetch(kb *kibana.Client, q Query, err error)`

but I get your point

_meta/beat.yml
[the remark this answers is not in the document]
simitt: Yes, as we diverged with the description from libbeat to make it more clear what is optional/required and what the default setting is. 
E.g. reading _Use SSL settings for HTTPS. Default is true._ I would expect that `ssl` is turned on by default, which is not the case (at least not for `output.elasticsearch`). For turning `ssl` on you need to set `protocol: https`. 
I suggest to quicky verify for Kibana, but I'd expect it to behave the same as for the outputs.

beater/beater.go
jalvz: apm-server does never stop when any other config is malformed/faulty, right? (elasticsearch host, max_bulk_size, etc) 
if we want the kibana setting be an exception, i guess we can add support in libbeat for that

beater/agent_config_handler.go
simitt: Up to you if you want to change it or not. I see this really as nitpicking. 

I didn't intend to start a long discussion around this, but following up on your explanation:
I don't see how `kb` is _concise_, it is an abbreviation for the package `kibana` and doesn't address that it actually holds the `client` from this package. Following your argument about being explicit and concise I would head into the direction of:
`beater.newMuxer(beaterConfig *Config, kbClient *kibana.Client, report publish.Reporter)`: in the beater package you might have multiple different clients, so adding the package abbrev might be helpful
`agentcfg.Fetch(client *kibana.Client, q Query, err error)` not necessary to add the package info in the name, however it is still a `client` entity.

beater/beater.go
simitt: Right, fine as is then.

beater/route_config.go
simitt: I suggest `/v1/agent/configs`, the `agent` being the namespace under which the `config` resource is nested, and aligned with how we introduced versioning for the intake API.

beater/config.go
```diff
@@ -258,6 +259,7 @@ func defaultConfig(beatVersion string) *Config {
 						filepath.Join("ingest", "pipeline", "definition.json")),
 				}},
 		},
-		Mode: ModeProduction,
+		Mode:   ModeProduction,
+		Kibana: nil,
```
simitt: Why are you setting a nil default?

agentcfg/model.go
```diff
@@ -0,0 +1,72 @@
...
+	Source Source `json:"_source"`
+}
+
+// Source represents the elasticsearch _source field of a document
+type Source struct {
+	Settings Settings `json:"settings"`
+}
+
+// Settings hold agent configuration
+type Settings map[string]string
```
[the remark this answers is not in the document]
jalvz: is needed for json unmarshalling

beater/config.go
jalvz: Because a default of `common.Config()` would imply enabled=true.
That means that after upgrading, users would be welcome with a new requirement (Kibana) or an error in the logs.

_meta/beat.yml
```diff
@@ -183,6 +183,56 @@ apm-server:
   #ilm:
     #enabled: false
 
+  #kibana:
```
roncohen: has there been a discussion around using the existing kibana settings instead of introducing a new one?

jalvz: Yes, it has been discussed.

This is the same setting, just in a different namespace. 

The beats setting is under `setup` which is not correct in our case; this is not for a setup command.

Also, should we use that, the server would refuse to start if Kibana is not reachable, which is not what we want, and not coherent eg. with Elasticsearch output.

roncohen: Gotcha 👍

beater/config.go
simitt: Ok. I'd be more explicit about it by either setting ```common.NewConfigFrom(`enabled: false`)``` or adding a dedicated test to the `config_test.go` checking that `Kibana.Enabled() == false`. Otherwise it could easily happen that this gets removed in the future without anything failing.

vendor/github.com/elastic/beats/libbeat/kibana/client.go
```diff
@@ -176,15 +176,14 @@ func (conn *Connection) Request(method, extraPath string,
 	return resp.StatusCode, result, retError
```
simitt: Please remove this file before merging.

jalvz:
Superseded by https://github.com/elastic/apm-server/pull/2289

Issue #1290, opened 2018-08-16
Support for ILM in APM Server


simitt:
Use Index Life Cycle Management for automate rollover for APM documents. 

**Update:**
https://github.com/elastic/beats/pull/10347 allows to implement our own index and ilm supporter, which are handling everything around index setup, ilm and template loading. 
The supporter have full access to the raw config, which means we are also free to define our own configuration for ILM. 

We need to come up with a draft on how ILM config should look like in APM Server. 

**Outdated information:**

Things to discuss:
* Is it still the best setup to have daily indices or is it better to have one index and create a size based rollover strategy?
* Could we allow and change the default to having an index per service if we can get rid of daily indices?

Related Work to ILM:
https://github.com/elastic/beats/issues/7935
https://github.com/elastic/elasticsearch/issues/29823

TODO:
- ~[ ] allow one template per index https://github.com/elastic/apm-server/issues/1483~
- ~[ ] update to ILM implementation from beats and adapt APM Server settings accordingly.~
- ~[ ] document ILM and change to ILM in APM Server https://github.com/elastic/apm-server/issues/1545~

makwarth:
Notes from IRL convo with Shane:

- ILM will be supported in Beats in 6.6, but he said we should check if the support is in libbeat or just Filebeat
- ILM will most likely be supported in Kibana in 6.6
- There's a build we can try when we're ready to implement and test

simitt:
We created a doc about [ILM beats and APM Server](https://docs.google.com/document/d/1kk9ECrvxzkhhkACV2TXHOxtBDT1Lo3YvXgkJ_T6xVsU/edit) integration a while ago, that links to ES and ongoing beats implementation and summarizes the open points we have to clarify.

simitt:
Notes from discussions with @ruflin and @elastic/apm-server team:
- Focus on solving the template per index problem
- Neglect the flexibel index creation, and find an easy solution for APM e.g. index aliases defined per event type up front
- Implement a basic version for the first implementation

Next step: 
- Investigate which changes would be necessary to allow one template per index.

simitt:
Waiting for PR comments on https://github.com/elastic/beats/pull/9616, as recent discussions showed that the current config approach might indicate more flexibility than what it actually allows (cc @urso ).

simitt:
PR https://github.com/elastic/beats/pull/10002 following beats core team suggestions around how to build indices has been closed again, after discussions that introducing this new config option seems rushed.

[...]

PR #2099, opened 2019-04-12
Add ILM support to APM Server

simitt:
This PR adds ILM support to the APM Server,  see https://github.com/elastic/apm-server/issues/1290#issuecomment-472392069.

It introduces an `apm-server.ilm.enabled` setting, that is set to `false` by default, so that no breaking change is introduced. 

The command `apm-server setup --template` gets deprecated with this PR, in favor of `apm-server setup --index-management`. 

When `apm-server setup`, or `apm-server setup --index-management` is run, one apm template gets created defining all of the fields, dynamic mappings, etc. Additionally one minimal template per event type is created, that only holds information about `index lifecycle management`. When ILM is disabled, no ilm policy and write alias is set, when ILM is enabled, the default ILM policy and write alias is linked in the specific templates, and the policies and write aliases are created. The templates per event only match against the specific event index. 

Note that the order of setting up policies, templates and write aliases is important. The templates need to be created before write aliases are created, as this creates an index, and templates will not be applied after index creation.  

The current implementation does not support any configuration options for ILM (policies, pattern, etc). As soon as ILM is enabled, configured `output.elasticsearch.index` and `output.elasticsearch.indices` settings, as well as `setup.template.name`, `setup.template.pattern` settings are ignored. 

The general logic for the `setup` commands are inherited from libbeat. `setup.template.enabled` and `apm-server.ilm.enabled` configs are respected when running the `setup` command. This is important, as otherwise a distinction whether or not ILM should be set up could not be made. 

~~Since this PR depends on unmerged changes in `libbeat`, it first pulls in the current libbeat master and then adds the changes from the [open libbeat PR](https://github.com/elastic/beats/pull/11777).~~

TODO:
- [x] adapt changelog
- [x] pull in libbeat master and remove direct changes of `vendor` folder. 
- [x] discuss and potentially revisit ILM policies
- [x] check with @nachogiljaldo for whitelisting `apm-server.ilm.enabled` settings


idxmgmt/ilm/policy.go
```diff
@@ -0,0 +1,97 @@
...
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package ilm
+
+var eventPolicies = map[string]map[string]interface{}{
```
graphaelli: I wasn't sure about using a map here since it's static so tried it out with a struct - it's not any better IMO.  details below in case you want to consider it still.

<details>

```go
type policy map[string]interface{}

type EventPolicies struct {
	error       policy
	span        policy
	transaction policy
	metric      policy
}

var eventPolicies = EventPolicies{
	error:       PolicyKeepMedium,
	span:        PolicyKeepMedium,
	transaction: PolicyKeep,
	metric:      PolicyKeep,
}

func (e *EventPolicies) get(eventType string) (policy, error) {
	switch eventType {
	case "error":
		return e.error, nil
	case "span":
		return e.span, nil
	case "transaction":
		return e.transaction, nil
	case "metric":
		return e.metric, nil
	default:
		return nil, fmt.Errorf("unknown event type: %s", eventType)
	}
}
```

</details>

idxmgmt/ilm/policy.go
```diff
@@ -0,0 +1,97 @@
...
+	"policy": map[string]interface{}{
+		"phases": map[string]interface{}{
+			"hot": map[string]interface{}{
+				"actions": map[string]interface{}{
+					"rollover": map[string]interface{}{
+						"max_size": "50gb",
+						"max_age":  "7d",
+					},
+					"set_priority": map[string]interface{}{
+						"priority": 100,
```
graphaelli: I didn't know about index priorities 👍

idxmgmt/ilm/policy.go
```diff
@@ -0,0 +1,97 @@
...
+					},
+				},
+			},
+			"warm": map[string]interface{}{
```
graphaelli: is a warm readonly phase needed if it's going to be frozen immediately anyway?  I might be misunderstanding how this policy works

idxmgmt/ilm/supporter_factory.go
```diff
@@ -18,14 +18,67 @@
...
+		Pattern: pattern,
+	}
+
+	return libilm.NewDefaultSupport(log, libilm.ModeEnabled, alias, policy, false, true), nil
+}
+
+func applyStaticFmtstr(info beat.Info, fmt *fmtstr.EventFormatString) (string, error) {
+	return fmt.Run(&beat.Event{
+		Fields: common.MapStr{
+			// beat object was left in for backward compatibility reason for older configs.
```
graphaelli: would this be removed in the next major then?

idxmgmt/ilm/supporter_factory_test.go
```diff
@@ -0,0 +1,62 @@
...
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package ilm
+
+import (
+	"testing"
+
+	"github.com/elastic/hey-apm/_vendor-20190409111253/github.com/stretchr/testify/require"
```
graphaelli: another guess at strange auto import

idxmgmt/ilm/template_test.go
```diff
@@ -0,0 +1,44 @@
...
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package ilm
+
+import (
+	"testing"
+
+	"github.com/elastic/hey-apm/_vendor-20190409111253/github.com/stretchr/testify/assert"
```
graphaelli: one more

apm-server.docker.yml
```diff
@@ -146,6 +146,12 @@ apm-server:
     # Overwrites existing pipeline definitions in Elasticsearch. Defaults to true.
     #overwrite: true
 
+  # When ilm is set to true the APM Server will take care of ILM setup.
+  # Configurations in output.elasticsearch.index and output.elasticsearch.indices sections will be ignored
+  # ilm can only be enabled for output.elasticsearch
+  ilm:
+    enabled: true
```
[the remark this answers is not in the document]
simitt: Somehow the docker.yml always needs two runs to properly update.. I'll look into it.

idxmgmt/indices_test.go
```diff
@@ -0,0 +1,49 @@
...
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package idxmgmt
+
+import (
+	"testing"
+
+	"github.com/elastic/hey-apm/_vendor-20190409111253/github.com/stretchr/testify/assert"
```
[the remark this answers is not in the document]
simitt: Yeah very strange, I didn't bother to look into it as I'll have to update all the dependencies anyways when updating to libbeat. But I'll keep it in mind.

idxmgmt/ilm/policy.go
simitt: You are right, I misunderstood the `min_age` param to always be relative to the last phase, but as stated in the [docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/_timing.html) it actually refers to creation or rollover time:
> min_age is usually the time elapsed from the time the index is created. If the index is rolled over, then min_age is the time elapsed from the time the index is rolled over. The intention here is to execute following phases and actions relative to when data was written last to a rolled over index.

I'll update accordingly.

idxmgmt/ilm/supporter_factory.go
simitt: Yes, I'd definitly remove then. I was a bit torn whether I should add it first place or not as ILM is newly introduced, but thought I align with the rest of the libbeat code for this.

idxmgmt/indices_test.go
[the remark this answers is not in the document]
graphaelli: got it, I didn't actually notice until I tried to run this locally and had to update it

simitt:
Closing for now, as some more changes will come through libbeat changes. Reopen once the issues regarding `setup` and `export` cmds are addressed in its core.

simitt:
jenkins, test this please

simitt:
@elastic/apm-server is adding the ILM related behaviour and is ready for review now.

idxmgmt/supporter.go
```diff
@@ -147,56 +135,149 @@ func (s *supporter) BuildSelector(cfg *common.Config) (outputs.IndexSelector, er
...
-		fields := m.assets.Fields(m.supporter.info.Beat)
-		if err := m.clientHandler.Load(tmplCfg, m.supporter.info, fields, m.supporter.migration); err != nil {
-			return fmt.Errorf("error loading Elasticsearch template: %+v", err)
+		// (3) load event type specific template respecting index lifecycle information
+		if templateComponent.load {
+			templateCfg = ilm.Template(ilmComponent.enabled, templateComponent.enabled, overwriteTemplate, alias, policy)
+			if err = m.clientHandler.Load(templateCfg, m.supporter.info, nil, m.supporter.migration); err != nil {
+				return errors.Wrapf(err, "error loading template %+v", templateCfg.Name)
+			}
+			log.Infof("Additional template loaded for %s", templateCfg.Name)
```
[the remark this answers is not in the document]
simitt: Good point, I updated the logged info to contain `Finished template setup for ..`

simitt:
Thanks @graphaelli! I had one more change - removed the `force_merge` from the policy, after some more digging into it, as this could be a heavy operation better suited to be scheduled by the users.

PR #2101, opened 2019-04-12
set user/group under systemd

### LINKED ISSUE #2096: systemd file missing user/group info
By default, the apm-server systemd file in 7.0.0 is missing the user and group information. This causes the process to attempt startup as root, which fails.

I've temporarily created an override file that adds

```
[System]
User=apm-server
Group=apm-server
```

The system starts properly, now

graphaelli:
fixes #2096 

upstream fix in: elastic/beats#11789


graphaelli:
upstream was approved, merging to get changelog in and simplifying backports (no need for beats update immediately)

Issue #1584, opened 2018-11-27
Make span.stacktrace.lineno optional


gregkalapos:
Follow up from [apm#18](https://github.com/elastic/apm/issues/18).

We agreed to make _lineno_ optional, since most of the agents can not guarantee to always have it.

gregkalapos:
I heard in the server meeting that @simitt brought up the question if this is just simply making it optional, which’d be a breaking change or should we make a formalized rule out of this?

From the .NET perspective I also think that some formalized rules that fit all languages is even better, especially if simply “making line numbers optional” is problematic from the servers perspective.

My original point was that in .NET we can surly not satisfy the line number required field and as it turns out other agents also have problems with it - so some of the agents fake it with 0 or -1. On the other hand those agents have other fields that we can rely on. Currently the [stacktrace_frame](https://github.com/elastic/apm-server/blob/master/docs/spec/stacktrace_frame.json) looks very JS specific to me, so if we can make a set of rules that fits all languages, then even better.

Fyi: for UI we agreed on not showing line number if it’s not supplied by the agent. For that, a prerequisite is that the agent has the option to not sending it - I guess for that it doesn’t matter if it’s just optional or part of a rule.

Hope this helps.

simitt:
Thanks for picking this up @gregkalapos. I think there was a misunderstanding though. Making this attribute optional is not going to be a breaking change. In case we want to introduce some more logic _afterwards_, e.g. one of multiple attributes must be set, this would be a breaking change. That is why I mentioned that if we need more logic on it we would need to introduce it now and not later.

The agreement was that we only make it optional, so no breaking change.

gregkalapos:
Ach, ok, got it.

PR #2105, opened 2019-04-15
make stacktrace lineno optional

jalvz:
fixes https://github.com/elastic/apm-server/issues/1584

- [x] needs changelog

- [x] ping agent devs


jalvz:
In the past it was determined that changing the grouping keys in minor is not a breaking change. 
Someone let me know if that is not the case any more, or if we should still consider the line number number for calculating grouping keys when present.

Issue #1972, opened 2019-03-01
Enable ingest pipelines by default


simitt:
From the [ingest user-agent docs](https://www.elastic.co/guide/en/elasticsearch/plugins/master/ingest-user-agent.html):
>The user_agent processor is no longer distributed as a plugin, but is now a module distributed by default with Elasticsearch. See the User Agent processor for more details.

This will be true for ES versions >= 6.7, and would allow us to enable the user [agent processor](https://www.elastic.co/guide/en/elasticsearch/reference/master/user-agent-processor.html) by default from 7.0 on. We could also add a check if the used ES versions is >= 6.7 and only turn on in case this is true.

simitt:
pinging @makwarth here as this might be interesting for you.

makwarth:
@simitt ah, I just mentioned this in the geo-ip issue 👍

simitt:
linking to the geo-ip issue for clarity #1283

jalvz:
this doesn't seem to be a big undertaking, any reason for not aiming 7.1?

jalvz:
Re-labelling this as per https://github.com/elastic/apm-server/issues/1283#issuecomment-476632253

graphaelli:
Update title to reflect addition of geoip processor alongside the user agent processor as recommended defaults.

Issue #1283, opened 2018-08-14
Add geo ip pipeline to default pipeline definitions in APM Server


simitt:
Investigate if it makes sense (and implement if so) to add geo-ip pipeline to default pipelines in APM
https://www.elastic.co/guide/en/elasticsearch/plugins/6.3/ingest-geoip.html.

makwarth:
FYI https://github.com/elastic/elasticsearch/pull/36898

simitt:
Update from [Ingest geoip](https://www.elastic.co/guide/en/elasticsearch/plugins/master/ingest-geoip.html) (>= 6.7): 
> The geoip processor is no longer distributed as a plugin, but is now a module distributed by default with Elasticsearch. See the GeoIP processor for more details.

This means that we _could_ make use of the geo-ip processor by default, as long as an ES version >= 6.7 is used.

makwarth:
@simitt I think we should definitely do that, and I'd include the [user-agent module](https://www.elastic.co/guide/en/elasticsearch/plugins/master/ingest-user-agent.html). What's the best way to move forward with this?

simitt:
@makwarth thanks for the feedback. I'll bring it to some internal discussions for how to prioritize this.

alvarolobato:
We checked that the new APM Server 7.0 will only need to support ES from 6.7 to 7.x, so we can enable this without doing a version check. 
We can do this task then for 7.1 instead of 7.2

graphaelli:
Add a second pipeline with both user_agent and geoip processors, update commented out default in configuration file.  Verify that upgrade from single pipeline in 7.1 to multiple in 7.2 works as expected / users the new pipeline by default.

jahtalab:
Recently in our sync meeting, it was mentioned that having ASN (Autonomous System Number) data available would help with debugging, e.g. knowing ASN is registered for a mobile service provider.

It seems that this data is available in a different database (GeoLite2 ASN). I'm not sure if we can work around this and have the fields from both the "GeoLite2 City" database as well as the "GeoLite2 ASN" database.

If adding ASN makes the implementation too complicated for this iteration, I'm ok with postponing it.

graphaelli:
It's not especially complicated, but I think we should hold off inclusion of the ASN information until the fields are [standardized in ECS](https://github.com/elastic/ecs/pull/341) and the elasticsearch processor supports those fields directly.  We can document usage in the meantime for users who want this functionality immediately.

[...]

PR #2106, opened 2019-04-15
Enable pipelines by default

jalvz:
related: https://github.com/elastic/apm-server/issues/1283

- [x] needs changelog
- [ ] needs docs update
- [x] consider deprecating `setup pipelines` cmd

### DIFF
--- a/beater/beater.go
+++ b/beater/beater.go
@@ -108,8 +108,7 @@ func New(b *beat.Beat, ucfg *common.Config) (beat.Beater, error) {
 		logger:  logger,
 	}
 
-	shouldSetupPipelines := beaterConfig.Register.Ingest.Pipeline.isEnabled() || b.InSetupCmd
-
+	shouldSetupPipelines := beaterConfig.Register.Ingest.Pipeline.isEnabled()
 	if isElasticsearchOutput(b) && shouldSetupPipelines {
 		logger.Info("Registering pipeline callback.")
 		err := bt.registerPipelineCallback(b)
--- a/beater/beater.go
+++ b/beater/beater.go
@@ -108,9 +108,7 @@ func New(b *beat.Beat, ucfg *common.Config) (beat.Beater, error) {
 		logger:  logger,
 	}
 
-	// setup pipelines if explicitly directed to or setup --pipelines and config is not set at all
-	shouldSetupPipelines := beaterConfig.Register.Ingest.Pipeline.isEnabled() ||
-		(b.InSetupCmd && beaterConfig.Register.Ingest.Pipeline.Enabled == nil)
+	shouldSetupPipelines := beaterConfig.Register.Ingest.Pipeline.isEnabled() || b.InSetupCmd
 
 	if isElasticsearchOutput(b) && shouldSetupPipelines {
 		logger.Info("Registering pipeline callback.")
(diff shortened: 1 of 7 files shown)


beater/beater.go
```diff
@@ -108,9 +108,7 @@ func New(b *beat.Beat, ucfg *common.Config) (beat.Beater, error) {
 		logger:  logger,
 	}
 
-	// setup pipelines if explicitly directed to or setup --pipelines and config is not set at all
-	shouldSetupPipelines := beaterConfig.Register.Ingest.Pipeline.isEnabled() ||
-		(b.InSetupCmd && beaterConfig.Register.Ingest.Pipeline.Enabled == nil)
+	shouldSetupPipelines := beaterConfig.Register.Ingest.Pipeline.isEnabled() || b.InSetupCmd
```
simitt: With this change pipeline would always be setup when in `setup` cmd, ignoring the actual setting of `register.ingest.pipeline.enabled`.

simitt:
> consider deprecating setup pipelines cmd

Why would you want to deprecate this? Afaik the setup cmd followed by a specific option should _only_ set this up.

beater/beater.go
jalvz: yes, is it wrong?

jalvz:
> Why would you want to deprecate this

Because if we default to true when starting apm-server, it becomes redundant I guess. And it is simpler to just provide one way of doing things.

graphaelli:
Logstash/Kafka output users would still need a setup command, I do not think we should deprecate.

beater/beater.go
simitt: From my understanding yes. The `setup` command should respect the configuration file, and when `register.ingest.pipeline.enabled` is set to `false`, then why should they be set up?

simitt:
This is my understanding from the `setup` cmd:
APM Server currently has 4 setup options:
- `apm-server setup`: should run all setup cmds that are available, according to the config
- `apm-server setup --template`: only concerns what needs to be set up for templates
- `apm-server setup --ilm-policy`: only concerns what needs to be set up for using ilm and policies
- `apm-server setup --pipelines`: only concerns what needs to be set up for ingest pipelines

beater/beater.go
jalvz: i believe the fix is like e9c50c240b913977e244fd8eefebc5bf40430f31 then

simitt:
This PR fixes https://github.com/elastic/apm-server/issues/1972 (Only the geo-ip pipeline task has been linked so far).

jalvz:
> Since we are trying to move away from having default settings only set in the yml file, I suggest to check if it is possible to enable in code

Good point, doesn't look that it is possible because lists are not merged when unpacking config. So options are:
1 - enable only in config file
2 - document explicitly that if user wants to add another pipeline, they need to enable "apm_user_agent" as well
3 - park this until we have better support for this

simitt:
After taking some time to look into the pipeline configuration, what do you think about going into the same direction as we did with the `idxmgmt.Supporter`? I would not directly couple it to the `idxmgmt` handling as it only concerns output pipelines for ES. Conceptually we could aim for introducing a pipeline supporter with a `BuildSelector` interface in libbeat though, see [current build selector handling in libbeat](https://github.com/elastic/beats/blob/master/libbeat/outputs/elasticsearch/elasticsearch.go#L219). This would allow for implementing our own pipeline handling, similar to the [`idxmgmt.BuildSelector` implementation](https://github.com/elastic/apm-server/blob/master/idxmgmt/supporter.go#L111). If you're up for it we should talk to the beats core team.

jalvz:
Closing this for now, if we decide to not use the supporter approach, I'll reopen again

PR #2110, opened 2019-04-16
Return a response body

### LINKED ISSUE #2094: Always return number of accepted events

jalvz:
Fixes elastic/apm-server#2094 

Also:

- returns always content-length
- gets rid of custom map marshaller and calls `sendPlain` always the same way (now we call it either like https://github.com/elastic/apm-server/blob/c519b033621fe10f8d24dc4d8b0e59a78b652584/beater/common_handler.go#L292 or like https://github.com/elastic/apm-server/blob/7062c73aacbddc611215ff5b305030edfde3c168/beater/intake_handler.go#L94
- fallback to plain text response if json parsing fails
- makes stream.Result to implement Error instead of String interface.

Let me know if anything here is wrong and I'll revert the appropriate commit. Also let me know if I should rename the `verbose` variable

- [x] needs changelog
- [x] needs ping agent devs


beater/test_approved_stream_result/TestRequestIntegrationSuccess.approved.json
```diff
@@ -0,0 +1,3 @@
+{
+    "accepted": 4
+}
```
simitt: We should keep one test ensuring the current result stays the same, by not having the query param set.

simitt: yes perfect, I overlooked that!

PR #2119, opened 2019-04-17
typo fix and minor edition to yml file

jalvz:


graphaelli:
Good question, looking like there is something wrong with the makefile deps

simitt:
I created an issue https://github.com/elastic/apm-server/issues/2123, as the question also came up in https://github.com/elastic/apm-server/pull/2099#discussion_r276059639

Issue #2121, opened 2019-04-17
Introduce explicit Server idle timeout


simitt:
At the moment the server only has a `ReadTimeout` and no explicit `IdleTimeout`. 
To avoid deriving the idle timeout from the read timeout, introduce an explicit setting with default value 45s.

PR #2122, opened 2019-04-17
Add Idle Timeout to Server config.

simitt:
Introduce a dedicated idle timeout for the apm server, to avoid the timeout being derived from the read timeout. 

Related to https://github.com/elastic/apm-agent-python/issues/409.

Closing https://github.com/elastic/apm-server/pull/1937 in favor of this PR. 

implements elastic/apm-server#2121


beater/config_test.go
```diff
@@ -40,6 +40,7 @@ func TestConfig(t *testing.T) {
 			config: []byte(`{
         "host": "localhost:3000",
         "max_header_size": 8,
+		"idle_timeout": 4s,
```
graphaelli: odd that this whitespace didn't cause intake issues

jalvz:
needs changelog and docs

beater/config_test.go
reveil: Is the idle_timeout configurable? If not the the ability to set a large idle_timeout (previously same read_timeout) is lost and therefore you can't have an uninterrupted TCP stream. This is a serious feature loss imho as HTTP over an uninterrupted TCP connection is the only thing that can make a hard guarantee not to cause the race condition in #1936. In earlier versions I could at lest work around it with configuration even if defaults were problematic.

Issue #2111, opened 2019-04-16
[Docs] Create article on how to add labels, user and custom metadata to transactions


formgeist:
# Summary

In the APM UI, we're making some layout changes to the metadata (contextual information) that sits on the transaction. It's displayed as a tab next to Timeline, and will from https://github.com/elastic/kibana/pull/35150 be consolidated into a single tab, "Metadata". For the section of information that could be user customized (labels, user and custom) we had help messages that linked to the appropriate agent API information on how to set context for each part. I'd like to consolidate this information in an article that we can link to and manage separately, so there's just a single link to a help article to get people started. Perhaps with more information beyond just linking to the agent documentation itself.

![01 DT - Transaction detail - Context (show all).png](https://zube.io/files/elastic/cd8f3cbb1b9dd39722062ad7f61f1f27-01-dt-transaction-detail-context-show-all.png)

@bmorelli25 let me know what you think about this

bmorelli25:
So, as an example for myself, currently with a transaction sample from the node agent I'm seeing links to the api reference:

* **User** - You can configure your agent to add contextual information about your users.
  [Learn more in the documentation.](https://www.elastic.co/guide/en/apm/agent/nodejs/2.x/agent-api.html#apm-set-user-context) _(set-user-context)_
* **Labels** - You can configure your agent to add filterable tags on transactions.
  [Learn more in the documentation.](https://www.elastic.co/guide/en/apm/agent/nodejs/2.x/agent-api.html#apm-set-tag) _(set-tag)_
* **Custom** - You can configure your agent to add custom contextual information on transactions. 
  [Learn more in the documentation.](https://www.elastic.co/guide/en/apm/agent/nodejs/2.x/agent-api.html#apm-set-custom-context) _(set-custom-context)_

New consolidated link will be: "**How to add labels and other data**"

Cool!

Quick question: Can you clarify what _article_ means to you? Would this article live in a blog post, documentation book, somewhere else?

My main concern with linking to something new (versus the documentation) is it's another thing that needs to be maintained. With that being said, we currently have to maintain the Kib -> agent links individually (keep the version current) so perhaps this change isn't a big deal.

I think it makes sense to have a landing page somewhere in the docs (apm overview maybe) that explains centrally what labels, user, and custom are. We can then link out to each agent from there. Similar to what we did with the dropped spans links from Kibana.

formgeist:
[...]

PR #2125, opened 2019-04-17
[docs] Add metadata landing page

bmorelli25:
Creates a placeholder landing page/link for https://github.com/elastic/apm-server/issues/2111.
Unblocks the docs from https://github.com/elastic/kibana/pull/35150.

The link from Kibana will be:
```js
// formatted
{config.docs.base_url}guide/en/apm/get-started/{config.docs.version}/metadata.html
// which points here in master
https://www.elastic.co/guide/en/apm/get-started/master/metadata.html
```

The placeholder in master looks like this. It is not public facing, and will not be accessible unless users click on the link from the master branch in Kibana.
Screen Shot 2019-04-17 at 11 56 20 AM

Note: I've decided to call the anchor "metadata". If anyone thinks a different name is more appropriate, please say so.

### DIFF
--- a/docs/guide/index.asciidoc
+++ b/docs/guide/index.asciidoc
@@ -26,3 +26,5 @@ include::./agent-server-compatibility.asciidoc[]
 include::./apm-breaking-changes.asciidoc[]
 
 include::./apm-release-notes.asciidoc[]
+
+include::./redirects.asciidoc[]
--- a/docs/guide/redirects.asciidoc
+++ b/docs/guide/redirects.asciidoc
@@ -1,0 +1,9 @@
+["appendix",role="exclude",id="redirects"]
+= Deleted pages
+
+The following pages do not exist. They may have moved, been deleted, or have not been created yet.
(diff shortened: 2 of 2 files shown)


jalvz:
this metadata has nothing to do with https://www.elastic.co/guide/en/apm/server/current/metadata-api.html, right?

bmorelli25:
@jalvz - with the intake API metadata endpoint, no. With the actual data, yes there will be overlap

jasonrhodes:
Actually @bmorelli25 the apm-data-model slug already exists, and doesn't mention "Metadata" -- is that URL correct?

bmorelli25:
I pasted the wrong link. My bad. It should end in /metadata. I edited the full link above.

bmorelli25:
I'm going to merge this so the link from Kibana works.

@jalvz - Hope you're enjoying the easter holiday! If you have concerns with naming when you get back, we can always update the link next week.

PR #2129, opened 2019-04-19
[docs] Add labels context and user info

bmorelli25:
Closes https://github.com/elastic/apm-server/issues/2111.

### DIFF
--- a/docs/guide/apm-data-model.asciidoc
+++ b/docs/guide/apm-data-model.asciidoc
@@ -97,8 +97,8 @@ which are points in time relative to the start of the transaction with some labe
 In addition, agents provide options for users to capture custom <<metadata, metadata>>.
 Metadata can be indexed - <<labels-fields,`labels`>>, or not-indexed - <<custom-fields,`custom`>>.
 
-TIP: Most agents limit keyword fields (e.g. `transaction.name`) to 1024 characters,
-and non-keyword fields (e.g. `labels`) to 10,000 characters.
+TIP: Most agents limit keyword fields (e.g. `labels`) to 1024 characters,
+and non-keyword fields (e.g. `span.db.statement`) to 10,000 characters.
 
 Transactions are stored in {apm-server-ref-v}/transaction-indices.html[transaction indices].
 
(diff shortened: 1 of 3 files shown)


docs/guide/apm-data-model.asciidoc
```diff
@@ -144,4 +157,100 @@ For a full list of tracked metrics, see the relevant agent documentation:
...
+* {apm-ruby-ref-v}/metrics.html[Ruby]
+
+// This heading is linked to from the APM UI section in Kibana
+[[metadata]]
+=== Metadata
+
+Metadata can enrich your events and make application performance monitoring even more useful.
+Let's explore the different types of metadata that Elastic APM offers.
+
+==== Quick Reference table
```
bmorelli25: Probably just going to move this information into the sections below. I fear people will skip over the main sections with this here.

docs/guide/apm-data-model.asciidoc
```diff
@@ -144,4 +157,100 @@ For a full list of tracked metrics, see the relevant agent documentation:
...
+Python: {apm-py-ref}/api.html#api-tag[`tag`]
+Ruby: {apm-ruby-ref}/api.html#api-agent-set-tag[`set_tag`]
+Rum: {apm-rum-ref}/api.html#apm-add-tags[`addTags`]
+
+|Custom
+|No
+v|<<transactions>>
+<<errors>>
+v|Go: {apm-go-ref}/api.html#slug-here[`???`]
+Java: {apm-java-ref}/public-api.html#slug-here[`???`]
```
bmorelli25: @felixbarny - The java agent has no custom context?

docs/guide/apm-data-model.asciidoc
```diff
@@ -144,4 +157,100 @@ For a full list of tracked metrics, see the relevant agent documentation:
...
+
+[[labels-fields]]
+==== Labels
+
+Labels are used to add *indexed* information to transactions, spans, and errors.
+Indexed means the data is searchable and aggregatable in Elasticsearch.
+Multiple labels can be defined with different key-value pairs.
+
+* Elasticsearch type: {ref}/object.html[object]
+* Elasticsearch field: `labels` (previously `context.tags`)
```
bmorelli25: Need to version this

docs/guide/apm-data-model.asciidoc
felixbarny: No, not yet, but as we resolved  elastic/apm#7, we can add it now: https://github.com/elastic/apm-agent-java/issues/604

docs/guide/apm-data-model.asciidoc
```diff
@@ -82,12 +85,17 @@ A transaction contains:
...
 * Other relevant information depending on the agent. Example: The JavaScript RUM agent captures transaction marks,
 which are points in time relative to the start of the transaction with some label.
 
-In addition, the agents provide some settings for users to capture customized information. This data is stored as not-indexed in a `custom` object.
-Searchable information is stored as `labels` instead.
+In addition, agents provide options for users to capture custom <<metadata, metadata>>.
+Metadata can be indexed - <<labels-fields,`labels`>>, or not-indexed - <<custom-fields,`custom`>>.
 
 TIP: Most agents limit keyword fields (e.g. `transaction.name`) to 1024 characters,
 and non-keyword fields (e.g. `labels`) to 10,000 characters.
```
felixbarny: labels are keyword fields and should be limited to 1024 by agents. There are actually only a few fields which have the 10k char limit but that not yet completely aligned: https://github.com/elastic/apm-dev/issues/302#issuecomment-432570637

docs/guide/apm-data-model.asciidoc
```diff
@@ -144,4 +157,101 @@ For a full list of tracked metrics, see the relevant agent documentation:
...
+Non-indexed means the data is not searchable or aggregatable in Elasticsearch,
+and you cannot build dashboards on top of the data.
+However, non-indexed information is useful for other reasons,
+like providing contextual information to help you quickly debug performance issues or errors.
+
+* Elasticsearch type: {ref}/object.html[object]
+* Elasticsearch fields: `transaction.custom` | `error.custom`
+
+IMPORTANT: Setting a circular object, large object, or a non JSON serializable object can lead to errors.
+
```
felixbarny: Maybe add a note that mapping explosions are not a concern here because it does not add fields to the mapping? At least I think that this is the case.

docs/guide/apm-data-model.asciidoc
bmorelli25: Ahh, thanks for the clarification here

docs/guide/apm-data-model.asciidoc
bmorelli25: Added above

docs/guide/apm-data-model.asciidoc
```diff
@@ -144,4 +157,100 @@ For a full list of tracked metrics, see the relevant agent documentation:
...
+Indexed means the data is searchable and aggregatable in Elasticsearch.
+Multiple labels can be defined with different key-value pairs.
+
+* Indexed: Yes
+* Elasticsearch type: {ref}/object.html[object]
+* Elasticsearch field: `labels` (previously `context.tags` in <v.7.0)
+* Applies to <<transactions>> | <<transaction-spans>> | <<errors>>
+
+Label values can be a string, boolean, or number.
+Note that some agents only support string values at this time.
```
felixbarny: We should note that number and boolean labels were only introduced in APM Server 6.7+. Using this API in combination with an older APM Server versions leads to validation errors.

docs/guide/apm-data-model.asciidoc
```diff
@@ -144,4 +157,100 @@ For a full list of tracked metrics, see the relevant agent documentation:
...
+Because labels are stored in the same place in Elasticsearch, each value must have the same data type.
```
eyalkoren:
```suggestion
Because labels for a given key are stored in the same place in Elasticsearch, all label values of a given key must have the same data type.
```

In addition, I am not sure what "stored in the same place in Elasticsearch" means.

bmorelli25: Good call on this @eyalkoren, thanks! "Stored in the same place" is confusing. It's meant that the key "foo" from multiple agents is stored in the same place in ES. I'll make this more clear

docs/guide/apm-data-model.asciidoc
```diff
@@ -144,4 +157,98 @@ For a full list of tracked metrics, see the relevant agent documentation:
...
+[[labels-fields]]
+==== Labels
+
+Labels are used to add *indexed* information to transactions, spans, and errors.
+Indexed means the data is searchable and aggregatable in Elasticsearch.
+Multiple labels can be defined with different key-value pairs.
+
+* Indexed: Yes
+* Elasticsearch type: {ref}/object.html[object]
+* Elasticsearch field: `labels` (previously `context.tags` in <v.7.0)
```
graphaelli: drop `v.` from `7.0`.  maybe `pre-7.0` instead of `<` ?

docs/guide/apm-data-model.asciidoc
```diff
@@ -144,4 +157,98 @@ For a full list of tracked metrics, see the relevant agent documentation:
...
+* Applies to <<transactions>> | <<transaction-spans>> | <<errors>>
+
+Label values can be a string, boolean, or number in APM Server v.6.7+.
```
graphaelli: drop `v.` from `6.7`

bmorelli25:
I think this is good to go now! Going to merge... Thanks all for the  👀

Issue #2083, opened 2019-04-05
Support shipping directly to monitoring cluster


graphaelli:
elastic/apm-server#2082 brings over new libbeat capabilities for shipping directly to a monitoring cluster from elastic/beats#9260.  apm-server can follow elastic/beats#11677 a a template for introducing this new functionality.  

https://github.com/elastic/beats/pull/11678 can be pulled in for docs.

thanks @ycombinator!

Issue #1957, opened 2019-02-25
[docs] Remove Kibana endpoint from Beats docs


bmorelli25:
Remove Kibana endpoint from Beats docs that slipped through: https://github.com/elastic/apm-server/pull/1864/commits/7044a9b1e6ce5bdc51b024aa41c5a00d4092bc33

bmorelli25:
This section needs an ifndef to be remove from APM documentation. Will fix on next beats docs update.

bmorelli25:
`ifndef` added for new `no_kibana` attribute: https://github.com/elastic/beats/pull/11823

PR #2131, opened 2019-04-22
[docs] Update documentation copied from beats

bmorelli25:
Updating some docs copied from beats. Everything is pretty straight forward, though I have two comments:

* ~My main concern is with the changes to `monitoring-beats.asciidoc`. Looks like https://github.com/elastic/beats/pull/9260 introduced a new class of settings named `monitoring.*`, which I can't get to work in APM Server. Can a dev confirm these doc changes should be left out for now?~ edit: Just realized I was trying this in `7.0`, not `master`. This change works in master. Should we also update `apm-server.yml` to use the new settings name?
* A separate build issue was causing my `make update-beats-docs` command to fail. To try and fix it, I updated the build_docs script to use the dockerized build command. This seems to work fine. Might make sense to update now that the `.pl` version has reached EOL. 

Related https://github.com/elastic/beats/pull/11823
Closes https://github.com/elastic/apm-server/pull/1957
Closes https://github.com/elastic/apm-server/issues/2083

### DIFF
--- a/apm-server.yml
+++ b/apm-server.yml
@@ -742,14 +742,15 @@ output.elasticsearch:
 # reporting is disabled by default.
 
 # Set to true to enable the monitoring reporter.
-#xpack.monitoring.enabled: false
+#monitoring.enabled: false
 
-# Uncomment to send the metrics to Elasticsearch. Most settings from the
-# Elasticsearch output are accepted here as well. Any setting that is not set is
-# automatically inherited from the Elasticsearch output configuration, so if you
-# have the Elasticsearch output configured, you can simply uncomment the
-# following line.
-#xpack.monitoring.elasticsearch:
+# Most settings from the Elasticsearch output are accepted here as well.
+# Note that these settings should be configured to point to your Elasticsearch *monitoring* cluster.
+# Any setting that is not set is automatically inherited from the Elasticsearch
+# output configuration. This means that if you have the Elasticsearch output configured,
+# you can simply uncomment the following line.
+#monitoring.elasticsearch:
+
   # Optional protocol and basic auth credentials.
   #protocol: "https"
   #username: "apm_system"
--- a/_beats/script/build_docs.sh
+++ b/_beats/script/build_docs.sh
@@ -37,5 +37,5 @@ do
     params="$params --resource=${resource_dir}"
   fi
 
-  $docs_dir/build_docs.pl $params --doc "$index" -out "$dest_dir"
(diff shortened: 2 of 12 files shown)


bmorelli25:
Side note - I wonder if I'm running into the same thing @jalvz did here: https://github.com/elastic/apm-server/pull/2119#issuecomment-484008445.
`make update` seems to only update `_meta/beat.yml` -> `apm-server.yml`
I have to run `make update` again to update `apm-server.docker.yml` from `apm-server.yml`. Is this expected behavior?

apm-server.yml
```diff
@@ -742,14 +742,15 @@ output.elasticsearch:
...
 # Set to true to enable the monitoring reporter.
-#xpack.monitoring.enabled: false
-
-# Uncomment to send the metrics to Elasticsearch. Most settings from the
-# Elasticsearch output are accepted here as well. Any setting that is not set is
-# automatically inherited from the Elasticsearch output configuration, so if you
-# have the Elasticsearch output configured, you can simply uncomment the
-# following line.
-#xpack.monitoring.elasticsearch:
+#monitoring.enabled: false
```
simitt: Sorry for the late response here. With the latest changes from libbeat, both options - `xpack.monitoring` and `monitoring` are supported, but `xpack.monitoring` will be deprecated from `7.1.0` on. Afaik we usually still document deprecated settings, so I suggest to document them both, but clearly mark the deprecated setting. 
Also, when describing the new setting, could you please ensure to fully remove `xpack` also from the header and the description.

bmorelli25: We do typically document deprecated settings, but I chose to follow Beats lead by removing it from yml files and docs: elastic/beats/pull/11677+ elastic/beats/pull/11678. We can document the deprecated settings, but it will mean adding more complexities to the docs as this is beats documentation - which isn't necessarily a bad thing - just want you to be aware.

bmorelli25: I'm asking to see if there's any rhyme or reason as to why xpack was left in the yml files, or if that was an oversight.

simitt: Thanks for checking @bmorelli25. 

Up to you if you want to leave the yml+docs as they are, just wanted to ensure it wasn't removed by mistake and is streamlined with documenting other/future deprecated features and settings.

bmorelli25: Cool beans - I appreciate you bringing this up! I'm very curious to hear the thought process for why beats did this the way they did. Based on their answers, I'll figure out the path forward and determine what I need to or don't need to do.

PR #2133, opened 2019-04-22
add staticcheck

graphaelli:
depends on #2132 

Of the original staticcheck issues:
```
$ staticcheck ./...
beater/beater.go:56:2: error var setupDashboardRemoved should have name of the form errFoo (ST1012)
beater/intake_handler.go:117:2: redundant return statement (S1023)
beater/integration_test.go:56:5: should use time.Since instead of time.Now().Sub (S1012)
beater/integration_test.go:67:7: this value of err is never used (SA4006)
beater/tracing_test.go:90:8: this value of err is never used (SA4006)
beater/tracing_test.go:172:20: this value of err is never used (SA4006)
idxmgmt/supporter.go:188:21: error strings should not be capitalized (ST1005)
idxmgmt/supporter.go:193:21: error strings should not be capitalized (ST1005)
model/error/event.go:113:25: error strings should not be capitalized (ST1005)
model/metadata/metadata_test.go:39:8: field inpErr is unused (U1000)
model/sourcemap/payload_test.go:56:2: this value of sourcemap is never used (SA4006)
model/sourcemap/payload_test.go:105:11: this value of err is never used (SA4006)
model/sourcemap/payload_test.go:119:11: this value of err is never used (SA4006)
model/span/event.go:176:25: error strings should not be capitalized (ST1005)
model/span/event.go:233:11: this value of err is never used (SA4006)
model/span/event_test.go:330:18: this value of err is never used (SA4006)
model/stacktrace.go:70:2: should merge variable declaration with assignment on next line (S1021)
model/stacktrace.go:86:14: should omit value from range; this loop is equivalent to `for errMsg := range ...` (S1005)
model/transaction/event.go:91:25: error strings should not be capitalized (ST1005)
processor/asset/sourcemap/sourcemap.go:79:6: should write s := raw["sourcemap"] instead of s, _ := raw["sourcemap"] (S1005)
processor/asset/sourcemap/sourcemap.go:88:10: should use fmt.Errorf(...) instead of errors.New(fmt.Sprintf(...)) (S1028)
processor/stream/processor_test.go:41:6: func validMetadata is unused (U1000)
processor/stream/processor_test.go:134:18: this value of err is never used (SA4006)
processor/stream/processor_test.go:177:18: this value of err is never used (SA4006)
sourcemap/accessor_test.go:62:11: this value of err is never used (SA4006)
sourcemap/cache.go:31:2: only the first constant in this group has an explicit type (SA9004)
sourcemap/error.go:40:9: the argument is already a string, there's no need to use fmt.Sprintf (S1025)
tests/approvals.go:80:10: should use fmt.Errorf(...) instead of errors.New(fmt.Sprintf(...)) (S1028)
tests/fields.go:125:10: should omit value from range; this loop is equivalent to `for k := range ...` (S1005)
tests/json_schema.go:157:3: should write cond := condRequiredKeys[key] instead of cond, _ := condRequiredKeys[key] (S1005)
tests/json_schema.go:339:9: assigning the result of this type assertion to a variable (switch m := m.(type)) could eliminate the following type assertions:
	/Users/gil/.local/go/src/github.com/elastic/apm-server/tests/json_schema.go:341:6 (S1034)
tests/json_schema.go:391:2: field from is unused (U1000)
tests/json_schema.go:392:2: field to is unused (U1000)
tests/json_schema.go:418:3: unnecessary nil check around range (S1031)
tests/set.go:81:9: should omit value from range; this loop is equivalent to `for k := range ...` (S1005)
tests/set.go:102:9: should omit value from range; this loop is equivalent to `for k := range ...` (S1005)
tests/set.go:130:9: should omit value from range; this loop is equivalent to `for k := range ...` (S1005)
utility/data_fetcher.go:34:2: error var FetchErr should have name of the form ErrFoo (ST1012)
utility/data_fetcher.go:34:23: error strings should not be capitalized (ST1005)
utility/data_fetcher_test.go:28:2: var strPtr is unused (U1000)
utility/data_fetcher_test.go:28:10: var strPtr2 is unused (U1000)
utility/data_fetcher_test.go:30:2: var intPtr is unused (U1000)
utility/data_fetcher_test.go:30:10: var intPtr2 is unused (U1000)
validation/validator.go:42:20: error strings should not be capitalized (ST1005)
```

one is still not addressed:

```
model/span/event.go:235:11: this value of err is never used (SA4006)
```

Ensuring this fails on CI before addressing that.


model/span/event_test.go
```diff
@@ -29,15 +29,16 @@ import (
...
 	m "github.com/elastic/apm-server/model"
 	"github.com/elastic/apm-server/model/metadata"
 	"github.com/elastic/apm-server/sourcemap"
 	"github.com/elastic/apm-server/transform"
-	"github.com/elastic/beats/libbeat/common"
 )
 
 func TestDecodeSpan(t *testing.T) {
-	spanTime, _ := time.Parse(time.RFC3339, "2018-05-30T19:53:17.134Z")
+	spanTime := time.Date(2018, 5, 30, 19, 53, 17, 134*1e6, time.UTC)
```
jalvz: honestly, i find L40 more readable

processor/asset/sourcemap/sourcemap.go
```diff
@@ -85,7 +82,7 @@ func (p *sourcemapProcessor) Validate(raw map[string]interface{}) error {
 
 	_, err := parser.Parse("", []byte(smap))
 	if err != nil {
-		return errors.New(fmt.Sprintf("Error validating sourcemap: %v", err))
+		return errors.Wrap(err, "error validating sourcemap")
```
jalvz: did you check how this reads? sometimes these `wrap` calls feel very unnatural and redundant, and what matters for the user at the end is that the sourcemap hasn't been generated correctly, as the user can't do nothing with a parsing error other than generating it again.

model/span/event_test.go
graphaelli: I was equal on both but am won over by the compile time failure over runtime/ignoring the failure, even if it is test code.

processor/asset/sourcemap/sourcemap.go
graphaelli: Yeah, the error looks the same, like: `error validating sourcemap: unexpected end of JSON input`

PR #2134, opened 2019-04-22
Include changelogs in beats updates

graphaelli:
per discussion IRL, thsi should help highlight changes inherited in beats updates that should be included in the apm-server changelog.


jalvz:
can we curate these a little bit somehow? (eg only include libbeat's)

graphaelli:
@jalvz were you thinking we'd parse them out automatically?   manually?  I expect the diffs themselves will be small, this just being the initial import, but I'm open to improvements.

jalvz:
automatically i guess. why do you expect the diffs to be small? 
most of it is not libbeat for any given version, so totally unrelated. It would also be nice to not go further back than 6.2, we have enough stuff we don't need in the code base already, imo

graphaelli:
> why do you expect the diffs to be small?

I looked at the diffs before putting this up and saw there are only 1 or 2 lines changed in most cases.  

<details>
<summary>beats> git log --stat --pretty="%H" -- CHANGELOG*</summary>

```
 1 file changed, 1 insertion(+)
 1 file changed, 1 insertion(+)
 1 file changed, 1 insertion(+)
 1 file changed, 1 insertion(+)
 1 file changed, 4 insertions(+)
 1 file changed, 1 insertion(+)
 1 file changed, 1 insertion(+)
 1 file changed, 1 insertion(+)
 1 file changed, 1 insertion(+)
 1 file changed, 1 insertion(+)
 1 file changed, 2 deletions(-)
 1 file changed, 1 insertion(+)
 1 file changed, 4 insertions(+)
 1 file changed, 3 insertions(+)
 1 file changed, 2 insertions(+)
 1 file changed, 1 insertion(+)
 1 file changed, 1 insertion(+)
 1 file changed, 1 insertion(+)
 1 file changed, 2 insertions(+)
 1 file changed, 2 insertions(+)
 1 file changed, 1 insertion(+)
 1 file changed, 4 insertions(+)
 2 files changed, 3 insertions(+), 2 deletions(-)
 1 file changed, 2 insertions(+)
 1 file changed, 1 insertion(+)
 1 file changed, 1 insertion(+)
 1 file changed, 1 insertion(+)
 1 file changed, 1 insertion(+)
 1 file changed, 1 insertion(+)
 1 file changed, 1 insertion(+), 1 deletion(-)
 1 file changed, 2 insertions(+)
 1 file changed, 1 insertion(+)
 1 file changed, 2 insertions(+)
```
</details>

> automatically i guess

Looking at the details of the changelogs,  I think it would be difficult to do this automatically.  

> It would also be nice to not go further back than 6.2

I was not planning to go back at all and just have this going forward on 7.x and master.  I share your concern that we already have a lot of unnecessary content in the repo but I think this set will actually add some value.

jalvz:
>  I think it would be difficult to do this automatically

Fair enough, but do you think is worth a try? or do it manually then? 

(im fine with this btw, just don't see much value in having eg. auditbit changelog in apm-server)

Issue #2124, opened 2019-04-17
intake errors not logged at log level error


graphaelli:
Upon sending a request to apm-server started with:
```
apm-server -E logging.level=warning -E apm-server.max_event_size=100
```
a client receives an http 400 with `event exceeded the permitted size` but nothing is logged server side.  This makes debugging intake issues on cloud challenging as `logging.level=error` is set there.

This issue is present in [7.0.0](https://github.com/elastic/apm-server/blob/v7.0.0/beater/intake_handler.go#L96) and [6.7.1](https://github.com/elastic/apm-server/blob/v6.7.1/beater/v2_handler.go#L96)

PR #2139, opened 2019-04-23
Log errors on error log level.

simitt:
Request failures for Intake API should be logged on log level `error` instead of log level `info`. 

implements #2124


simitt:
jenkins, retest this please

PR #2147, opened 2019-04-24
Fix numeric user id decoding

### LINKED ISSUE #2127: Metadata.User.id must be json encoded string
Howdy!

# Preamble
I ran into a either deserialisation or specification problem. The spec define the `id` field of the `Metadata.user` [schema](https://www.elastic.co/guide/en/apm/server/6.7/metadata-api.html#metadata-user-schema) that it can be of type `"type": ["string", "integer", "null"]`.

# Problem Description 
The `metadata.user.id` serialised as number represented json object (e.g. `"id": 123`) without the quotation marks as defined by the [JSON specification](https://tools.ietf.org/html/rfc7159#page-6) renders the apm-server to respond with  a `500: Internal Server Error`. But, when the `id` is serialized as a string (e.g. `"id": "123"`) the apm-server responds with `200`.

### Correct
```
"{"metadata": "service": {"name":"demo", "agent":{"name":"i-am-an-agent", "version":"6.7.0"}},"user":{"id":"12345","username":"foobar","email":"email@acme.com"}}}
```

### Failing
```
"{"metadata": "service": {"name":"demo", "agent":{"name":"i-am-an-agent", "version":"6.7.0"}},"user":{"id":12345,"username":"foobar","email":"email@acme.com"}}}
```

# APM Server Version
```
$ ./apm-server version
apm-server version 6.7.1 (amd64), libbeat 6.7.1 [5665711214eb3fcb956e7d1c61795e427bf97097 built 2019-04-02 14:55:06 +0000 UTC]
```

graphaelli:
already allowed by intake spec

backports part of #1783 

closes #2127

### DIFF
--- a/changelogs/6.7.asciidoc
+++ b/changelogs/6.7.asciidoc
@@ -3,7 +3,7 @@
 
 https://github.com/elastic/apm-server/compare/6.6\...6.7[View commits]
 
-//* <<release-notes-6.721>>
+//* <<release-notes-6.7.2>>
 * <<release-notes-6.7.1>>
 * <<release-notes-6.7.0>>
 
--- a/changelogs/6.7.asciidoc
+++ b/changelogs/6.7.asciidoc
@@ -3,8 +3,21 @@
 
 https://github.com/elastic/apm-server/compare/6.6\...6.7[View commits]
 
+//* <<release-notes-6.721>>
 * <<release-notes-6.7.1>>
 * <<release-notes-6.7.0>>
+
+////
+[[release-notes-6.7.2]]
+=== APM Server version 6.7.2
+
+https://github.com/e2astic/apm-server/compare/v6.7.1\...v6.7.2[View commits]
+
+[float]
+==== Bug fixes
+
+- Fix numeric user id decoding {pull}2147[2147].
+////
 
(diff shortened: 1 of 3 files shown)


model/metadata/user.go
```diff
@@ -42,12 +43,23 @@ func DecodeUser(input interface{}, err error) (*User, error) {
...
+
+	//id can be string or int
+	tmp := decoder.Interface(raw, "id")
+	if tmp != nil {
+		if t, ok := tmp.(json.Number); ok {
+			id := t.String()
+			user.Id = &id
+		} else if t, ok := tmp.(string); ok && t != "" {
+			user.Id = &t
+		}
```
graphaelli: @elastic/apm-server I've only backported what is on master for the moment but believe this should add a decoding error if `Id` is not a `json.Number` or `string`, even if that isn't possible via intake.

changelogs/6.7.asciidoc
```diff
@@ -3,9 +3,22 @@
 
 https://github.com/elastic/apm-server/compare/6.6\...6.7[View commits]
 
+//* <<release-notes-6.721>>
```
simitt: typo above

PR #2156, opened 2019-04-30
[docs] What happens when x is down

bmorelli25:
Closes: https://github.com/elastic/apm/issues/74

Screen Shot 2019-05-01 at 12 49 21 PM

### DIFF
--- a/docs/common-problems.asciidoc
+++ b/docs/common-problems.asciidoc
@@ -244,26 +244,26 @@ APM Agent --> Load Balancer  --> APM Server
 
 [[server-es-down]]
 [float]
-=== What happens to data when APM Server or Elasticsearch is down?
+=== What happens when APM Server or Elasticsearch is down?
 
-==== Elasticsearch down
+*If Elasticsearch is down*
 
 If Elasticsearch goes down, the APM Server will keep data in memory until Elasticsearch is back up,
 or until it runs out of space in its internal in-memory queue. 
 You can <<adjust-queue-size,adjust the internal queue size>> if necessary.
-When the queue becomes full, APM Server will respond with an <<queue-full`HTTP 503: Queue is full`>>,
+When the queue becomes full, APM Server will respond with <<queue-full,`HTTP 503: Queue is full`>>,
 and data will be lost.
 
-==== APM Server down
+*If APM Server is down*
 
 Some agents have internal queues or buffers to temporarily store data in case the APM Server goes down.
 As a general rule of thumb, queues fill up quickly. Assume data will be lost if APM Server goes down.
 Adjusting these queues/buffers can increase the overhead of the agent, so use caution when updating default values.
 
 * Go Agent - `ELASTIC_APM_BUFFER_SIZE`/`ELASTIC_APM_REQUEST_SIZE`?
-* Java Agent - {}/config-reporter.html#config-max-queue-size[`max_queue_size`]
+* Java Agent - {apm-java-ref}/config-reporter.html#config-max-queue-size[`max_queue_size`]
 * Node.js Agent - No internal queue. Data is lost
-* Python Agent - {}/tuning-and-overhead.html#tuning-queue[Transaction Queue]
-* Ruby Agent - {}/configuration.html#config-api-buffer-size[`api_buffer_size`]
+* Python Agent - {apm-py-ref}/tuning-and-overhead.html#tuning-queue[Transaction Queue]
+* Ruby Agent - {apm-ruby-ref}/configuration.html#config-api-buffer-size[`api_buffer_size`]
 * RUM Agent - ?
 // * .NET Agent - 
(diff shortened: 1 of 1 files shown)


docs/common-problems.asciidoc
```diff
@@ -240,3 +241,29 @@ For example:
...
+Some agents have internal queues or buffers to temporarily store data in case the APM Server goes down.
+As a general rule of thumb, queues fill up quickly. Assume data will be lost if APM Server goes down.
+Adjusting these queues/buffers can increase the overhead of the agent, so use caution when updating default values.
+
+* Go Agent - `ELASTIC_APM_BUFFER_SIZE`/`ELASTIC_APM_REQUEST_SIZE`?
+* Java Agent - {}/config-reporter.html#config-max-queue-size[`max_queue_size`]
+* Node.js Agent - No internal queue. Data is lost
+* Python Agent - {}/tuning-and-overhead.html#tuning-queue[Transaction Queue]
+* Ruby Agent - {}/configuration.html#config-api-buffer-size[`api_buffer_size`]
+* RUM Agent - ?
```
bmorelli25: @jahtalab / @vigneshshanmugam - can one of you help me with this?

docs/common-problems.asciidoc
```diff
@@ -240,3 +241,29 @@ For example:
...
+When the queue becomes full, APM Server will respond with an <<queue-full`HTTP 503: Queue is full`>>,
+and data will be lost.
+
+==== APM Server down
+
+Some agents have internal queues or buffers to temporarily store data in case the APM Server goes down.
+As a general rule of thumb, queues fill up quickly. Assume data will be lost if APM Server goes down.
+Adjusting these queues/buffers can increase the overhead of the agent, so use caution when updating default values.
+
+* Go Agent - `ELASTIC_APM_BUFFER_SIZE`/`ELASTIC_APM_REQUEST_SIZE`?
```
bmorelli25: @axw - can you help me understand these settings in this context?

docs/common-problems.asciidoc
```diff
@@ -240,3 +241,29 @@ For example:
...
+* Java Agent - {}/config-reporter.html#config-max-queue-size[`max_queue_size`]
+* Node.js Agent - No internal queue. Data is lost
+* Python Agent - {}/tuning-and-overhead.html#tuning-queue[Transaction Queue]
+* Ruby Agent - {}/configuration.html#config-api-buffer-size[`api_buffer_size`]
+* RUM Agent - ?
+// * .NET Agent -
```
bmorelli25: @gregkalapos - not super important yet, but has this been determined in the .NET agent yet?

gregkalapos: @bmorelli25  Currently it's the same as in Node.js: `No internal queue. Data is lost`. At the moment we don't have a plan for this, so I'd say this will remain so as long as this does not became an important problem. - Let me know if we should rethink this.

bmorelli25: I'm probably not the right person to ask, but based on what I've read in other agent issues and on Slack, I think that's ok 👍.

Thanks for your answer!

docs/common-problems.asciidoc
axw: It's a bit complicated :)

https://www.elastic.co/guide/en/apm/agent/go/current/configuration.html#config-api-buffer-size
https://www.elastic.co/guide/en/apm/agent/go/current/configuration.html#config-api-request-size

There's actually four levels of queuing/buffering:
 - a queue (channel in Go parlance) between the application code and the background "tracer" goroutine. The size of this queue is not configurable, but data will never stay on there for long anyway
 - a circular buffer whose size is controlled by `ELASTIC_APM_BUFFER_SIZE`. When the tracer receives an event from the channel, it will add it to the buffer. If the buffer is full, then old events will be dropped until the new event can be added
 - a zlib compression buffer. The size is not configurable, data just goes into the buffer until there's enough to compress, or is flushed to the request buffer
 - a request buffer, which holds the ND-JSON encoded and zlib compressed payload. The size is configured with `ELASTIC_APM_REQUEST_SIZE`

Increasing `ELASTIC_APM_BUFFER_SIZE` means that more events can be kept in memory, e.g. because the server is unreachable, or because the application is sending events at a faster rate than the agent can send them to the server. 

Increasing `ELASTIC_APM_REQUEST_SIZE` means that more events may be sent to the APM Server in a single request. I think it would be rare to need to change this.

docs/common-problems.asciidoc
jahtalab: @bmorelli25,  We don't retry on failures so if the APM server is down the data is lost. However, if APM server is up but ES is down then APM server will keep events in a queue until that queue becomes full and then starts with request failures and from that point on the data is lost.

bmorelli25: Thank you Hamid!

docs/common-problems.asciidoc
bmorelli25: oooo that is complicated. Thanks for the thorough explanation!

If I'm understanding correctly, then it makes sense to me to mention the circular buffer and its config variable: `ELASTIC_APM_BUFFER_SIZE` here then.

bmorelli25:
@jalvz or @simitt - can one of you 👍/ 👎the APM Server portion of this when you have a sec?

jalvz:
lgtm @bmorelli25 , thanks!

PR #2165, opened 2019-05-03
[ci] add files for JJBB jobs

jonahbull:
Add JJBB job definitions to the repo so they can be managed here.

JJBB is a wrapper around JJB which aims to simplify CI job
configuration. The main benefit is that it moves the job configuration
files out of the Infra repo and into each product team's repo(s). Take
a look at the [README](https://github.com/elastic/infra/blob/master/flavortown/jjbb/README.md) for more information.

`check_paths_for_matches` is a utilty script that lets us work around
some undesirable behavior by the git plugin, which requires an
existing workspace to check the included/excluded-regions directive.

We'll configure a multijob that runs this script on every push to master
for the repo and when the script detects a match in the changed paths
(default is to look for paths that begin with `^.ci/(jobs|views)`),
it'll launch the JJBB job to update job definitions. Otherwise the JJBB
job would get triggered every push to master, which is undesirable for a
variety of reasons.


.ci/jobs/elastic+apm-server+6.7+beats-updated.yml
```diff
@@ -0,0 +1,40 @@
+---
+- job:
+    name: elastic+apm-server+6.7+beats-updated
```
graphaelli: can the name be parameterized too?  Ideally this check would run against 7.1 also

kuisathaverat: @graphaelli we've to replace this job with a multibranch pipeline to avoid to have to modify the job with each new branch.

kuisathaverat: also probably is time to talk about how to automate the beats update process.

PR #2170, opened 2019-05-07
[docs] APM Server install and run structure

bmorelli25:
You currently have to jump around a bit in order to _get up and running with APM Server_. This PR aims to rework the "Getting Started" and "Setting up and running" sections of the APM Server documentation to better guide users through the setup process. It also includes "cloud first" language.

* Similar changes were implemented in the APM Overview in https://github.com/elastic/apm-server/pull/2171.
* Must be persisted in the Beats repository
* Closes https://github.com/elastic/apm-dev/issues/545.

Proposed new structure:
* Overview
* Getting Started With APM Server
  * Step 1: Install APM Server
  * Step 2: Configure APM Server
  * Step 3: Start APM Server
  * Step 4: Next steps
  * Repositories for APT and YUM
  * Running APM Server on Docker
* Setting up APM Server
  * Directory layout
  * Secrets keystore
  * Command reference
  * High Availability
  * APM Server and systemd
* Configuring
  * ...
* ...
* Troubleshooting
  * Common problems
  * Debug
  * Get help
* ...

---
**Important:** I've temporarily set `:release-state:` to `released` for conditional download sections to appear in the documentation preview. This must be reset to `unreleased` before merging.


bmorelli25:
@jalvz / @simitt -- I'd like some feedback from you on important steps when installing APM Server. Right now I have this:

Step 1: Install APM Server
Step 2: Configure APM Server
Step 3: Start APM Server
Step 4: Next steps

Does it make sense to also include a step for "Load the index template in Elasticsearch"? Beats does this, and it looks something like this:

Step 1: Install APM Server
Step 2: Configure APM Server
Step 3: Load the index template in Elasticsearch
Step 4: Start APM Server
Step 5: Next steps

Thoughts? Beats also includes a step for loading dashboards, which doesn't make sense in our case, but are there other steps that you think are important that would be good to include here?

jalvz:
I don't have any strong opinion on step 3 (server should do that already on startup), but maybe is a good idea to make it explicit so that people are aware of it (as a downside: it is one step more which might feel awkward for people using eg systemd)

Maybe is sensible (instead of step 3?) to add an step after "start server" of verifying that everything looks fine in logs (elasticsearch connected, pipeline registered, template loaded, rum enabled (if it was configured), etc.

It looks fine to me either way

bmorelli25:
Thanks @jalvz, that's helpful feedback!

bmorelli25:
This PR is finally ready for review. Since this is a bigger PR, it'd be great to feedback from both of you!

docs/getting-started-apm-server.asciidoc
```diff
@@ -0,0 +1,285 @@
...
+The location of this file varies by platform -- see <<directory-layout>> for help with locating it.
+
+All available configuration options are outlined in
+{apm-server-ref-v}/configuring-howto-apm-server.html[configuring APM Server].
+
+[float]
+[[index-template-config]]
+==== Index template configuration
+
+In Elasticsearch, {ref}/indices-templates.html[index templates] are used to define settings
```
jalvz: A template does a lot of things and summarizing it as "determine how fields should be analyzed" doesn't do it justice, so I'd just leave it as "... are use to define field mappings and settings".

docs/getting-started-apm-server.asciidoc
```diff
@@ -0,0 +1,285 @@
...
+and mappings that determine how fields should be analyzed.
+
+The recommended index template file for {beatname_uc} is installed by the {beatname_uc} packages.
```
jalvz: I'd said before I don't have a strong opinion about this part, but given how many issues users are facing because of this, I changed my mind. 
I think this is good. 99% of times I prefer concision, but here I'd add some points about:

- The APM UI relies on specific mappings, and changing them might break it.
- Users can run `apm-server -e setup template` to force a template overwrite; and `apm-server export template` to see what is the template bundled with apm-server and compare it with the actual template in their elasticsearch instance.
- Note that index templates are not applied retrospectively, but only to indices created from then on. For APM, this usually means one day, or whatever indicates the ILM policy (if there is one).
- Consider a troubleshooting section in Kibana mentioning that certain errors happen because of having the wrong template installed, linking back to this section and ILM manual setup. Feel free to sync up with the apm-ui team for this.

docs/setting-up-and-running.asciidoc
```diff
@@ -1,115 +1,24 @@
 
 [[setting-up-and-running]]
-== Setting up and running APM Server
+== Setting up APM Server
```
jalvz: I'd consider renaming this section, as generally (and vaguely) "set up" implies install + configure, which is already covered. The best I could come up with is "Operating APM Server", which I don't think is too bad; but maybe you have better ideas (or are happy with "Setting up"!)

docs/storage-management.asciidoc
```diff
@@ -151,9 +151,17 @@ deleting data collected for specific services or customers,
...
+{curator-ref-current}[Curator], the {ref}/docs-delete-by-query.html[Delete By Query API],
+or in the {kibana-ref}/managing-indices.html[Kibana Index Management UI].
+
+[float]
+[[delete-data-ilm]]
+==== Delete data with ILM
+
+Index Lifecycle management (ILM) enables you to automate how you want to manage your indices over time.
+You can base actions on factors such as shard size and performance requirements.
+Learn more in <<ilm>>.
```
jalvz: Learn more _about_ ?

docs/getting-started-apm-server.asciidoc
```diff
@@ -0,0 +1,285 @@
+[[getting-started-apm-server]]
+== Getting Started With APM Server
+
+// This section was copied from the APM Overview
+// Switch to tagged regions when available
+The easiest way to get started with APM Server is by using our
```
jalvz: :100: !

docs/getting-started-apm-server.asciidoc
```diff
@@ -0,0 +1,285 @@
...
+
+["source","sh",subs="attributes"]
+------------------------------------------------
+curl -L -O {downloads}/apm-server-{version}-darwin-x86_64.tar.gz
+tar xzvf apm-server-{version}-darwin-x86_64.tar.gz
+------------------------------------------------
+
+endif::[]
+
+[[linux]]
```
jalvz: I'd reconsider the order of the sections to `deb`, `rpm`, `linux`, `mac`, `windows`, and `Running on Docker`:

Maybe is a good idea also to rename `linux` to `other linux`, and change casing to `Linux`, `Mac`, `Windows`, and maybe `RPM` and `DEB` too.

If you change anything here, make sure it is consistent elsewhere... I am sort of sensing that you have your good reasons to use lower case there, but eg. in http://apm-server_2170.docs-preview.app.elstc.co/guide/en/apm/server/master/running-with-systemd.html is `RPM` and `DEB` :)

Speaking of which: http://apm-server_2170.docs-preview.app.elstc.co/guide/en/apm/server/master/running-with-systemd.html shouldn't be embedded in or linked to the running section?

docs/common-problems.asciidoc
```diff
@@ -51,7 +52,7 @@ As a result, Elasticsearch must be configured to allow {ref}/docs-index_.html#in
 === HTTP 400: Data decoding error / Data validation error
 
 The most likely cause for this is that you are using incompatible versions of agent and APM Server.
-For instance, APM Server 6.2.0 changed the Intake API spec and requires a minimum version of each agent.
+For instance, APM Server 6.2 and 6.5 changed the Intake API spec and requires a minimum version of each agent.
```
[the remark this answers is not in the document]
bmorelli25: great find

docs/getting-started-apm-server.asciidoc
```diff
@@ -0,0 +1,285 @@
...
+// *******************************************************
+
+[[installing]]
+=== Step 1: Install APM Server
+
+NOTE: *Before you begin*: If you haven't installed the {stack}, do that now.
+See {stack-gs}/get-started-elastic-stack.html[Getting started with the {stack}].
+
+To download and install {beatname_uc}, use the commands below that work with your system.
+If you use Apt or Yum, you can <<setup-repositories,install {beatname_uc} from our repositories>>
```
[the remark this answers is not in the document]
bmorelli25: Good catch

docs/getting-started-apm-server.asciidoc
bmorelli25: These are good points. I've made some updates.

From what I can tell, the _correct_ capitalization is `deb` and `RPM`. The only reason I had everything capitalized the other way was because that's how Beats does it... in other words, not a very good reason. I'll update these on our end, then chat with DeDe about updating them in all of the shared Beats docs as well.

docs/setting-up-and-running.asciidoc
bmorelli25: This section is really tough because it's kind of a mishmash of topics.

**Directory layout** and **Secrets Keystore** could really be at the beginning of the "Configuring APM Server" section. And **APM Server and Systemd** could be in the "Getting Started" section. I understand why **High Availability** was put here initially, but I don't think it really makes sense in this section.

I'm not opposed to the name "Operating APM Server", but I wonder if there's more we can do to clean this up.

docs/getting-started-apm-server.asciidoc
bmorelli25: Added in your recommendations, thanks!

> Consider a troubleshooting section in Kibana mentioning that certain errors happen because of having the wrong template installed, linking back to this section and ILM manual setup. Feel free to sync up with the apm-ui team for this.

I'm going to leave the last bullet point for when I tackle https://github.com/elastic/apm-server/issues/2470

docs/setting-up-and-running.asciidoc
jalvz: Much agreed. Feel free to move those sections, in this or other PR

docs/configuration-process.asciidoc
```diff
@@ -10,11 +10,27 @@ apm-server:
   rum:
     enabled: true
 
+output:
+  elasticsearch:
+    hosts: ElasticsearchAddress:9200
```
simitt: It would be more consistent to align this with [L29](https://github.com/elastic/apm-server/pull/2170/files#diff-7b74f374359d31b837afa311b6f055efR29) ` hosts: ["ElasticsearchAddress:9200"]`.

docs/getting-started-apm-server.asciidoc
```diff
@@ -0,0 +1,296 @@
...
+
+// *******************************************************
+// STEP 2
+// *******************************************************
+
+[[apm-server-configuration]]
+=== Step 2: Configure APM Server
+
+If you've installed APM Server yourself,
+you can edit the `apm-server.yml` configuration file to customize APM Server to your needs.
```
simitt: Could you link to how to override settings in the hosted version, I am missing that here.

docs/getting-started-apm-server.asciidoc
```diff
@@ -0,0 +1,296 @@
...
+This means you need to ensure the correct index template is loaded from day one.
+
+More information is available in <<load-template-auto, configure template loading>>.
+
+// *******************************************************
+// STEP 3
+// *******************************************************
+
+[[apm-server-starting]]
+=== Step 3: Start APM Server
```
simitt: Sorry for the late response here, but I do actually think that describing to run `apm-server setup` before starting the APM Server is a good idea. One can apply config values the same way as when starting the server, but might see config issues or failing requirements early on. It can be helpful to ensure everything is setup properly before starting data ingestion.

docs/getting-started-apm-server.asciidoc
bmorelli25: 🤦‍♂ Yup, good call.

docs/getting-started-apm-server.asciidoc
```diff
@@ -0,0 +1,314 @@
...
+image::images/apm-architecture-cloud.png[Install Elastic APM on cloud]
+
+https://www.elastic.co/cloud/elasticsearch-service/signup[Try out the {es} Service for free],
+and then begin configuring your own user settings right in the Elasticsearch Service Console.
+Any changes are automatically appended to the `apm-server.yml` configuration file for your instance.
+A full list of configuration options is available in Elastic Cloud's
+{cloud}/ec-manage-apm-settings.html[Add APM user settings].
+
+[float]
+=== Install the stack yourself
```
simitt: _Install_ and/or _Run_ ?

docs/getting-started-apm-server.asciidoc
bmorelli25: In my first attempt at this, I've decided to turn step 2 into "Setup and Configure". I've then used a tagged region to pull in the `setup` command from the command reference page. You can see the changes in 860696c or here: http://apm-server_2170.docs-preview.app.elstc.co/guide/en/apm/server/master/apm-server-configuration.html

docs/getting-started-apm-server.asciidoc
```diff
@@ -0,0 +1,314 @@
...
+
+// This section was copied from the APM Overview
+// Switch to tagged regions when available
+The easiest way to get started with APM Server is by using our
+https://www.elastic.co/cloud/elasticsearch-service[hosted {es} Service] on
+Elastic Cloud. The {es} Service is available on AWS, GCP, and Azure,
+and automatically configures APM Server to work with {es} and {kib}.
+
+[float]
+=== Hosted Elasticsearch Service
```
simitt: Shouldn't this be _Hosted Elastic Service_?

docs/getting-started-apm-server.asciidoc
```diff
@@ -0,0 +1,314 @@
...
+
+// *******************************************************
+// STEP 2
+// *******************************************************
+
+[[apm-server-configuration]]
+=== Step 2: Setup and configure
+
+[float]
+==== Elastic Cloud
```
simitt: I suggest to use the same terminology as above [Hosted Elastic Service](https://github.com/elastic/apm-server/pull/2170/files#diff-231fc508dffaf5b3009bd25f92cc2dc7R12); same for the on-prem alternative.

docs/getting-started-apm-server.asciidoc
bmorelli25: I'll double-check with marketing, but based on https://www.elastic.co/cloud/, I think the actual product name is "Elasticsearch Service"

docs/getting-started-apm-server.asciidoc
bmorelli25: I'm okay with changing this, but if we do, I'll need to change it in APM Overview as well.

What about "Install and manage the stack yourself"?

simitt: SGTM

docs/getting-started-apm-server.asciidoc
simitt: LGTM

docs/getting-started-apm-server.asciidoc
bmorelli25: Confirmed we should stick with "Hosted Elasticsearch Service"

docs/getting-started-apm-server.asciidoc
bmorelli25: Changed

docs/getting-started-apm-server.asciidoc
```diff
@@ -0,0 +1,317 @@
...
+
+[float]
+==== Self installation
+
+Once you've installed {beatname_uc}, you can set up and configure it according to your needs.
+
+[float]
+[[index-template-config]]
+===== Set up
+
```
simitt: While the level you use is the right one, the visual representation does not clearly show that _Set up_ is nested under _Self installation_.  I am wondering if this additional _Setup_ and _Configure_ headers are really needed or you could just remove them and having everything directly under _Self installation_. 

Screenshot 2019-09-11 at 11 40 31

PR #2177, opened 2019-05-09
add geoip + default pipeline definitions

### LINKED ISSUE #1283: Add geo ip pipeline to default pipeline definitions in APM Server

graphaelli:
Adds two new pipeline definitions without changing the existing `apm_user_agent`:
* `apm_user_geo` - only GeoIP resolution of `client.ip` to `client.geo`
* `apm` - both `apm_user_agent` and `apm_user_geo`

updates the default pipeline suggested in apm-server.yml

closes #1283

### DIFF
--- a/_meta/beat.yml
+++ b/_meta/beat.yml
@@ -306,7 +306,7 @@ output.elasticsearch:
   # (1) ensure pipelines are registered in Elasticsearch, see `apm-server.register.ingest.pipeline`
   # (2) enable the following:
   #pipelines:
-  #- pipeline: "apm_user_agent"
+  #- pipeline: "apm"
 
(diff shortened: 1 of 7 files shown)


simitt:
I agree, fwiw there are already some [pipeline registration ](https://github.com/elastic/apm-server/blob/master/tests/system/test_integration.py#L560) and [pipeline setup ](https://github.com/elastic/apm-server/blob/master/tests/system/test_setup.py#L9) system tests that you could extend.

PR #2189, opened 2019-05-20
sync changelogs/release notes to 7.0

graphaelli:
A bit mixed on adding 6.8.0 release notes to 7.0 since it was released after 7.0 but went for it as most users would probably expect them to be there.

### DIFF
--- a/changelogs/6.8.asciidoc
+++ b/changelogs/6.8.asciidoc
@@ -1,0 +1,16 @@
+[[release-notes-6.8]]
+== APM Server version 6.8
+
+https://github.com/elastic/apm-server/compare/6.7\...6.8[View commits]
+
+* <<release-notes-6.8.0>>
+
+[[release-notes-6.8.0]]
+=== APM Server version 6.8.0
+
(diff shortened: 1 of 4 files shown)


changelogs/6.8.asciidoc
```diff
@@ -0,0 +1,16 @@
...
+== APM Server version 6.8
+
+https://github.com/elastic/apm-server/compare/6.7\...6.8[View commits]
+
+* <<release-notes-6.8.0>>
+
+[[release-notes-6.8.0]]
+=== APM Server version 6.8.0
+
+https://github.com/e2astic/apm-server/compare/v6.7.1\...v6.8.0[View commits]
```
jalvz: 6.7.2

graphaelli: Good point, we'll have to fix this is in multiple branches.  What do you think about comparing the tag with the previous branch instead?  Like https://github.com/elastic/apm-server/compare/6.7...v6.8.0?

jalvz: Well, if you click "View commits" on a patch release, you just want to see the content of that patch, don't you?

graphaelli: true

PR #2192, opened 2019-05-22
Version bump to 7.3

simitt:
Wait with merging until `7.2` is branched out of `7.x`.


graphaelli:
All unit/integration tests passed but packaging tests failed due to CI system running out of disk space.  Packaging tests pass locally and don't appear to be generating much larger artifacts that usual, so merging now and will follow up on CI issues.

PR #2209, opened 2019-05-23
Check if ILM is supported when configured.

### LINKED ISSUE #2208: [ILM] check if ILM is supported in client handler
Run a check to see if ILM is supported by the configured ES before loading any templates. This allows to fail early and ensure a clean state after starting the server or running the setup cmd, and it ensures a potential error message is clear on the reason why the setup/connection fails. 

(This does not include changing behaviour to panic and stop the server immediately if ILM is configured but not supported. In such cases the server will try to reconnect. The behaviour is inherited from libbeat and the only possibility to change within APM Server would be to panic.)

simitt:
Raise ILM not supported error early before any templates are loaded
to ensure clean state.

fixes elastic/apm-server#2208

### DIFF
--- a/idxmgmt/supporter.go
+++ b/idxmgmt/supporter.go
@@ -183,12 +183,12 @@ func (m *manager) Setup(loadTemplate, loadILM libidxmgmt.LoadMode) error {
 	//(4) load write alias per event type AFTER the template has been created,
 	//    as this step also automatically creates an index, it is important the matching templates are already there
 
-	type ilmStruct struct {
+	type ilmHandler struct {
 		manager   libilm.Manager
 		supporter libilm.Supporter
 	}
 	var (
-		ilmInfo           []ilmStruct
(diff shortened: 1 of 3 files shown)


idxmgmt/supporter.go
```diff
@@ -176,21 +176,52 @@ func (m *manager) Setup(loadTemplate, loadILM libidxmgmt.LoadMode) error {
...
 
 	//setup index management:
+	//(0) check if setup is supported
 	//(1) load general apm template
 	//(2) load policy per event type
 	//(3) create template per event respecting lifecycle settings
 	//(4) load write alias per event type AFTER the template has been created,
 	//    as this step also automatically creates an index, it is important the matching templates are already there
 
+	type ilmStruct struct {
```
graphaelli: how about `ilmHandler` like in the comments?

PR #2210, opened 2019-05-23
run approvals and fix apm-server version in tests

jalvz:
7.1 CI failure is legit: https://apm-ci.elastic.co/blue/organizations/jenkins/apm-server%2Fapm-server-mbp/detail/7.1/6/pipeline


simitt:
Should we maybe blacklist the `version` in the approval files? There is a possibility to define keys that that have `XXX` as values.

jalvz:
i don't mind the version, is not a big annoyance. thanks!

Issue #1750, opened 2019-01-09
Panic with bad SSL config


jalvz:
apm-server panics at start up if ssl is enabled but no certificate/key path is given. It should fail gracefully with a meaningful error message instead

jalvz:
This now fails with `ERROR	instance/beat.go:877	Exiting: open : no such file or directory`
There is no indication at all about what file can't be opened and why, but at least it doesn't panic.

Issue #1906, opened 2019-02-08
configurable protocols and cipher suites for apm agents to apm-server communication


n0othing:
It'd be useful to be able to restrict certain SSL/TLS procols or cipher suites for apm agent to apm-server communications (similar to apm-server to Elasticsearch[1]).

[1] https://www.elastic.co/guide/en/apm/server/current/configuration-ssl.html#_literal_supported_protocols_literal

PR #2224, opened 2019-05-24
Add config options for Server/Agent communication via TLS

simitt:
This PR introduces more granularity for configuring agent/server communication via TLS. 
It adds support to configure:
* certificate authorities to use for verifying client certificates
* supported protocols
* cipher suites and curve types for ECDHE based cipher suites
* whether or not to require client authentication

TODO:
- [x] changelog
- [x] fix warnings for client auth created by the onboarding doc
- [x] ~find proper handling for tracer when `ssl.enabled`~ works already out of the box
- [x] remove skipped flaky go tests, and replace with python integration tests
- [x] fix tests

fixes to elastic/apm-server#1906


beater/server.go
```diff
@@ -70,25 +79,22 @@ func run(server *http.Server, lis net.Listener, config *Config) error {
...
 
-	ssl := config.SSL
-	if ssl.isEnabled() {
-		cert, err := outputs.LoadCertificate(&config.SSL.Certificate)
-		if err != nil {
-			return err
-		}
-		server.TLSConfig = &tls.Config{Certificates: []tls.Certificate{*cert}}
+	if server.TLSConfig != nil {
+		logger.Info("SSL enabled.")
```
joshbressers: Can we s/SSL/TLS/

beater/server.go
```diff
@@ -70,25 +79,22 @@ func run(server *http.Server, lis net.Listener, config *Config) error {
...
 		return server.ServeTLS(lis, "", "")
 	}
 	if config.SecretToken != "" {
 		logger.Warn("Secret token is set, but SSL is not enabled.")
+	} else {
+		logger.Info("SSL disabled.")
```
joshbressers: s/SSL/TLS/

beater/beater.go
```diff
@@ -184,33 +187,56 @@ func (bt *beater) Run(b *beat.Beat) error {
...
-
 	bt.mutex.Lock()
 	if bt.stopped {
 		defer bt.mutex.Unlock()
 		return nil
 	}
 
-	bt.server = newServer(bt.config, tracer, pub.Send)
+	bt.server, err = newServer(bt.config, tracer, pub.Send)
+	if err != nil {
```
graphaelli: 👍 been meaning to do this for ages

beater/beater.go
```diff
@@ -184,33 +187,56 @@ func (bt *beater) Run(b *beat.Beat) error {
...
+	// following an example from https://golang.org/pkg/net/
+	// dial into tcp connection to ensure listener is ready, send get request and read response,
+	// in case tls is enabled, the server will respond with 400,
+	// as this only checks the server is up and reachable errors can be ignored
+	conn, err := net.DialTimeout("tcp", bt.config.Host, timeout)
+	if err != nil {
+		return false
+	}
+	fmt.Fprintf(conn, "GET / HTTP/1.0\r\n\r\n")
+	bufio.NewReader(conn).ReadString('\n')
```
graphaelli: iirc, you said we need to read something on the conn for an error to be thrown at `Close` - if so, can we read just a single byte (`ReadByte()`) instead?  Does apm-server log anything if we only read part of the response?

beater/server.go
```diff
@@ -70,25 +79,22 @@ func run(server *http.Server, lis net.Listener, config *Config) error {
...
 		logger.Info("RUM endpoints enabled!")
 	case false:
-		logger.Info("RUM endpoints disabled")
+		logger.Info("RUM endpoints disabled.")
 	}
 
 	if config.MaxConnections > 0 {
 		lis = netutil.LimitListener(lis, config.MaxConnections)
-		logger.Infof("connections limit set to: %d", config.MaxConnections)
+		logger.Infof("Connections limit set to: %d", config.MaxConnections)
```
graphaelli:
```suggestion
		logger.Infof("Connection limit set to: %d", config.MaxConnections)
```

beater/server.go
graphaelli: won't this log the same as line 91?  I think it can be removed

apm-server.ssl.yml
```diff
@@ -0,0 +1,872 @@
+################### APM Server Configuration #########################
```
jalvz: out of curiosity, what is this one for?

beater/server.go
simitt: 91 is logging `SSL enabled`, here I log `SSL disabled` to ensure both cases are covered in the logs.

apm-server.ssl.yml
simitt: hah! You caught my, that was my config I played around with during dev, will remove.

simitt:
@graphaelli and @joshbressers I mainly used _SSL_ instead of _TLS_ to avoid confusion, as all the settings are nested under `ssl`. If you think it is preferable I can switch to use TLS instaed, but I suggest to use both, e.g. `SSL/TLS`. WDYT?

I was also thinking about introducing the option `tls`, completely switching to use _TLS_ in terminology, but for backwards compatibility still supporting `ssl` as it would be just an alias to all the configs under `tls`. But we do have the same concerns for all other ssl options (e.g. in the `output` sections), so I refrained from that for now.

jalvz:
re TLS, i believe https://github.com/elastic/dev/issues/670 is still current

joshbressers:
> @graphaelli and @joshbressers I mainly used _SSL_ instead of _TLS_ in the `apm-server.yml` to avoid confusion, as all the settings are nested under `ssl`. If you think it is preferable I can switch to use TLS instaed, but I suggest to use both, e.g. `SSL/TLS`. WDYT?
> 
> I was also thinking about introducing the option `tls`, completely switching to use _TLS_ in terminology, but for backwards compatibility still supporting `ssl` as it would be just an alias to all the configs under `tls`. But we do have the same concerns for all other ssl options (e.g. in the `output` sections), so I refrained from that for now.

Yeah, for now staying consistent is probably more important. I withdraw my suggestion :)

beater/server.go
graphaelli: ah 🤦‍♂

graphaelli:
does this also resolve #1750 ?

simitt:
@graphaelli , no it doesn't, this needs to be resolved in libbeat though.

Issue #2136, opened 2019-04-23
Document ILM support in APM Server


simitt:
Document newly added ILM support in APM Server, implemented with https://github.com/elastic/apm-server/pull/2099.

bmorelli25:
Coordinate with #2099 before merging

simitt:
Please also document the additional privileges the [APM Server `write_user`](https://www.elastic.co/guide/en/apm/server/7.0/beats-basic-auth.html) needs when making use of ILM. 
See the privileges [defined in the integration tests](https://github.com/elastic/apm-integration-testing/pull/451/files). 
Get inspired by the [beats docs](https://www.elastic.co/guide/en/beats/filebeat/current/beats-basic-auth.html)

bmorelli25:
Related to https://github.com/elastic/apm-server/pull/2228

PR #2231, opened 2019-05-29
[docs] ILM documentation

bmorelli25:
Closes https://github.com/elastic/apm-server/issues/2136
Needs https://github.com/elastic/apm-server/pull/2228 (xpack security improvements from Beats)

~~Rough draft up for @simitt and I to discuss on Monday. Lot's of work to do here still, don't get too excited 😄~~

Conversation with Silvia on Jun 4:
* Do not document ability to change ILM policy. This is because it can take up to seven days for the changes to take effect.
* Keep the manual ILM instructions around for now. For users that want to customize ILM, point them at this solution.
* Encourage the user to use the setup command
* Privileges required for APM Server:
  * cluster: manage_ilm
  * indices: manage, manage_ilm

### DIFF
--- a/docs/ilm.asciidoc
+++ b/docs/ilm.asciidoc
@@ -2,25 +2,28 @@
 [role="xpack"]
 == Index lifecycle management (ILM)
 
-Use the {ref}/getting-started-index-lifecycle-management.html[index lifecycle management]
+Use the {ref}/getting-started-index-lifecycle-management.html[index lifecycle management (ILM)]
 feature in {es} to manage your APM Server indices as they age.
-For example, instead of creating daily indices where index size can vary based on the number of agents and number of events sent,
-use an index lifecycle policy to automate a rollover to a new index when the existing index reaches a specified size or age.
+ILM enables you to automate how you want to manage your indices over time,
+by automating rollovers to a new index when the existing index reaches a specified size or age.
 
 You can view and edit the index lifecycle policies in the *Index lifecycle policies* UI in {kib}.
 For more information about working with the UI,
 see {kibana-ref}/index-lifecycle-policies.html[Index lifecyle policies].
 
+APM Server currently offers two ways to get started with ILM:
+
+1. <<ilm-default,**ILM default policy**>> - Get up and running with a default index lifecycle management policy as quickly as possible.
+2. <<manual-ilm-setup,**Manual ILM**>> - Customize and manage your own ILM policies.
+
 NOTE: If you're migrating from an existing setup,
 any indices present before ILM was configured will need to be managed manually.
 
-[float]
-[[ilm-setup]]
-=== ILM default configuration
+[[ilm-default]]
+=== ILM default policy
 
-IMPORTANT: The default index lifecycle configuration cannot be changed until after it has been applied.
-
-The default configuration only applies hot and warm phase definitions.
+Index lifecycle management will manage an index based on its defined policy.
+The default ILM configuration applies hot and warm phase policies.
 Cold and delete phases are not defined.
 Because errors and spans lose information value faster than metrics and transactions do,
 there are two different policies defined:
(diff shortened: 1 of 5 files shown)


docs/ilm.asciidoc
```diff
@@ -0,0 +1,190 @@
+[[ilm]]
+[role="xpack"]
+== Index lifecycle management (ILM)
+
+Use the {ref}/getting-started-index-lifecycle-management.html[index lifecycle management]
+feature in {es} to manage your APM Server indices as they age.
+For example, instead of creating daily indices where index size can vary based on the number of agents and number of events sent,
+use an index lifecycle policy to automate a rollover to a new index when the existing index reaches a specified size or age.
```
simitt: While technically proper, we do use daily rollover without defining a `delete` phase in the current state. You describe later on _why_ we chose this policy, but reading this statement might still be confusing at the current stage.

docs/ilm.asciidoc
```diff
@@ -0,0 +1,190 @@
...
+see {kibana-ref}/index-lifecycle-policies.html[Index lifecyle policies].
+
+NOTE: If you're migrating from an existing setup,
+any indices present before ILM was configured will need to be managed manually.
+
+[float]
+[[ilm-setup]]
+=== ILM default configuration
+
+IMPORTANT: The default index lifecycle configuration cannot be changed until after it has been applied.
```
simitt: Once the indices are created, changes to templates and policies are not going to be applied to existing indices. 

With ILM in APM being relatively strict and not offering config options at the moment, I'd generally head into the direction of _In case you'd like to manage your own ILM policies refer to our manual setup guide_.

docs/ilm.asciidoc
```diff
@@ -0,0 +1,190 @@
...
+
+**Error & Span events**
+_14 days to look at data in a very fast way_
+HOT
+* A new index is created every 50gb or 1 day (whichever comes first). This makes it very easy to delete indices day by day, and ensures no index gets too large.
+* After seven days, hot indices are moved to the warm phase.
+WARM
+* Warm indices are read-only.
+* After another seven days, warm indices are moved to the cold phase.
+COLD
```
simitt: Yes, the cold phase info is outdated.

docs/ilm.asciidoc
```diff
@@ -0,0 +1,190 @@
...
+
+**Transactions & Metrics**
+_Compare the current month with your last month without any performance reduction_
+HOT
+* A new index is created every 50gb or 7 days (whichever comes first). This makes it very easy to delete indices week by week, and ensures no index gets too large.
+* After one month (31 days), hot indices are moved to the warm phase.
+WARM
+* Warm indices are read-only.
+* After two more months (62 days), warm indices are moved to the cold phase.
+COLD
```
simitt: Cold phase is also outdated here.

docs/ilm.asciidoc
```diff
@@ -0,0 +1,190 @@
...
+// END OLD INFO
+
+[float]
+[[ilm-setup]]
+=== ILM setup
+
+To set up index lifecycle management, set `ilm.enabled` to `true` in apm-server.yml.
+ILM can only be enabled for `output.elasticsearch`.
+When enabled, configurations defined for `output.elasticsearch.index` and `output.elasticsearch.indices` will be ignored.
+ILM clears all index settings and sets up a predefined set of indices and policies that the user cannot initially configure.
```
simitt: > ILM clears all index settings

I can't follow what you mean by this.

docs/ilm.asciidoc
```diff
@@ -0,0 +1,190 @@
...
+
+It is recommended to setup index lifecycle management (ILM) before starting APM Server.
+This prevents indices from being created before policies are changed, and allows policies to be applied to indices at creation time.
```
simitt: > This prevents indices from being created before policies are changed, and allows policies to be applied to indices at creation time.

Unfortunately not true any more

docs/ilm.asciidoc
```diff
@@ -0,0 +1,190 @@
...
+Index lifecycle management will manage an index based on its defined policy.
+Run the <<setup-command>> with the ` --index-management` flag to set up the default ILM policy:
+
+[source,js]
+-----------------------
+./apm-server setup --index-management
+-----------------------
+// CONSOLE
+
+//Policies only need to be created once, and will persist through version upgrades.
```
simitt: Policies are versioned, so while they are persisted, they need to be setup for every version.

docs/ilm.asciidoc
```diff
@@ -0,0 +1,190 @@
...
+[[ilm-edit]]
+=== Editing ILM defaults
+
+Once you've enabled index lifecycle management and setup the ILM policies, you can view and edit the policies in the Index lifecycle policies UI in Kibana. For more information about working with the UI,
+see {kibana-ref}/index-lifecycle-policies.html[*Index lifecycle policies*].
+
+Index policies and templates are prefixed with version numbers, for example: `apm-7.2.0-metric-000001`.
+This means if you use APM Server, modify the policy, and upgrade, you'll need to do two things:
+1. Set `ilm.enabled` to false and configure your indices to the old ILM indices in the yml file.
+2. Update the policies manually when you upgrade. When APM Server is started for the first time,
```
simitt: No check for policies when `ilm.enabled` is set to false. There are additional `template`s created per event type holding minimal ILM information, if ilm is disabled in the server config, but should be used with a manual setup, the user has to take care of creating those templates upfront and setting `setup.template.overwrite: false` (default).

docs/ilm.asciidoc
```diff
@@ -0,0 +1,190 @@
...
+APM Server checks to see if the policies are there for this version, if they're not, it creates them.
+
+
+// NOTES BELOW...
+// **************
+
+// True?? If you do not run the setup command, policies will will not be applied immediately??
```
simitt: false, they should always get applied. The setup cmd is just a cleaner separation between setting everything up and ingesting data.

docs/ilm.asciidoc
```diff
@@ -0,0 +1,190 @@
...
+
+// We want users to have the option to define policies and prefixes for the indices, not the whole config.
+
+When you start the APM Server, you'll see in the logs index-management.
+The template gets set up the same as before.
+The ILM policy gets setup.
+Load template error-ilm template (for every type of event) template, policy, alias
```
simitt: yes, but template name changed to `apm-server-<version>-<event-type>`

docs/ilm.asciidoc
```diff
@@ -0,0 +1,190 @@
...
+
+---
+GET _template/apm*
+---
+
+Used to just be the `apm-{observer.version}` template.
+Now, there are four additional templates, one for each event. Each has an order (`"order": 2`). If multiple templates contain the same key, ES prefers the template with the lower number. Each event template sets the index lifecycle.
```
simitt: `order: 2` was a bug, it is actually `order: 1`. 
The higher number overriding the lower number order, see [multiple templates](https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-templates.html#multiple-templates)

docs/ilm.asciidoc
bmorelli25: fixed

docs/ilm.asciidoc
bmorelli25: I'm wondering if we should include this information (and some of the info in https://github.com/elastic/apm-server/issues/2253) in the troubleshooting section instead? I don't want to muddy the water too much, but if you think there will be a lot of users enabling/disabling/enabling/disabling, then we probably should include it here.

bmorelli25:
The docs ci failure is unrelated to this PR

docs/ilm.asciidoc
```diff
@@ -0,0 +1,126 @@
...
+<1> Move to warm phase after _31 days_
+<2> Rollover after _50gb_
+<3> Rollover after _7 days_
+<4> Priority for recovering your indices after a node restart. Higher priorities are recovered first.
+
+Your indices are now configured to use index lifecylce management. Go ahead and <<setting-up-and-running,run APM Server>>.
+
+// [float]
+// [[ilm-upgrade]]
+// ==== ILM default policy upgrades
```
simitt: Having a paragraph declaring default policies are subject to change in future releases is a good idea, please add it.

docs/ilm.asciidoc
```diff
@@ -0,0 +1,126 @@
...
+|=======================================================================
+
+IMPORTANT: Changes to the default index lifecycle policy do not take effect until the current index has rolled over.
+If you'd like to manage a custom policy, see <<manual-ilm-setup,Manual ILM>>.
+
+[float]
+[[ilm-setup]]
+==== ILM default policy setup
+
+To set up index lifecycle management, set `ilm.enabled` to `true` in apm-server.yml.
```
simitt: I'd also mention that in case users switch between ilm enabled/disabled, they should set `setup.template.overwrite` to `true` to ensure complete setup.

docs/ilm.asciidoc
bmorelli25: 👍

bmorelli25:
Rebased, which should fix the failing docs build.

Thank you so much @simitt for all of your help!

docs/ilm.asciidoc
bmorelli25: 👍

bmorelli25:
Had to rebase the branch. Everything looks good on this PR! Will merge once tests go green.

PR #2265, opened 2019-06-07
Add top level span and transaction to metricset.

### LINKED ISSUE #2086: Top-level tags/labels for metrics
For https://github.com/elastic/apm/issues/70, we have the need for top-level `tags`/`labels` for metrics. In particular, these are  `transaction.name`, `transaction.type`, `span.type` and `span.subtype`. This is needed for a unified query bar experience. For example, filtering for a particular `transaction.name` should filter the related metrics as well as they should be used on the service detail view, powering the breakdown graphs.

Some ideas on how to enable top-level labels:
 - APM Server knows about "well-known" tags like `transaction_name` and `transaction_type` and brings them to top-level
 - Extend the metricset spec (https://github.com/elastic/apm-server/blob/master/docs/spec/metricsets/metricset.json) to allow for top-level key/value pairs (dots in the keys have to be allowed)

simitt:
Metricsets can now store information about which transaction and span they belong to, allowing for better query and grouping experience.

Closes elastic/apm-server#2086

cc @felixbarny


felixbarny:
Is this waiting for another approval or can it be merged?

simitt:
We have to wait until https://github.com/elastic/kibana/pull/38411 is approved and merged.

PR #2267, opened 2019-06-10
run package install tests too

graphaelli:


graphaelli:
jenkins, run the tests please

kuisathaverat:
I think we can make it in a separate step to avoid mix logs of different stuff, then it would be easy to know what failed, I've made a PR with my suggestion and a cosmetic change to add labels to all script steps https://github.com/elastic/apm-server/pull/2268

graphaelli:
sounds great as long at the install test will always run on the same host / right after the package step since it depends on the files it builds, thanks!

superseded by #2268

PR #2269, opened 2019-06-11
[docs] Update redirect page for dashboards

bmorelli25:
Adds info about `apm-contrib` to the dashboards redirect page

### DIFF
--- a/docs/redirects.asciidoc
+++ b/docs/redirects.asciidoc
@@ -258,5 +258,5 @@ Loading dashboards from APM Server is no longer supported. Please see the {kiban
 
 Loading Kibana dashboards from APM Server is no longer supported.
 Please use the {kibana-ref}/xpack-apm.html[Kibana APM UI] instead.
-As an alternative, select dashboards and visualizations are available in the
+As an alternative, a small number of dashboards and visualizations are available in the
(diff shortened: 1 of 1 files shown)


docs/redirects.asciidoc
```diff
@@ -256,4 +256,7 @@ Loading dashboards from APM Server is no longer supported. Please see the {kiban
 [role="exclude",id="load-kibana-dashboards"]
 === Dashboards
 
-Loading dashboards from APM Server is no longer supported. Please see the {kibana-ref}/xpack-apm.html[Kibana APM UI] documentation.
\ No newline at end of file
+Loading Kibana dashboards from APM Server is no longer supported.
+Please use the {kibana-ref}/xpack-apm.html[Kibana APM UI] instead.
+As an alternative, select dashboards and visualizations are available in the
```
simitt: typo: _select dashboards_

bmorelli25: Ahh, so in this case, `select` is being used an adjective, not a verb. This makes it synonymous with "chosen", for example.

Let's update this to something a bit less confusing though. I'm thinking: `As an alternative, a small number of dashboards and visualizations are available in the...`. Whatever we go with, I want to encourage use of the UI (when possible).

bmorelli25: Going to merge this for now, but happy to change it later if you have a better suggestion

PR #2271, opened 2019-06-12
Process all user-agent entries from headers.

### LINKED ISSUE #2252: Process multiple user-agent header keys
The APM Server processes only the first value per unique key of the `user-agent` information in the request header, additional values are ignored. 

Multiple values might be set, e.g. when using proxies, which can lead to unexpected and non-deterministic behavior. 

The suggested way of dealing with this is to combine multiple values into one string, separated by a comma (see [RFC2616 4.2](https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2))

simitt:
Process all user-agent entries from headers.

Combine all user-agent values from header information to one string, separated by `;` to not lose information.

fixes elastic/apm-server#2252


utility/data_fetcher.go
```diff
@@ -257,6 +258,19 @@ func (d *ManualDecoder) Headers(base map[string]interface{}) http.Header {
 	return httpHeader
 }
 
+// UserAgentHeader fetches all `user-agent` values from a given header and combines them into one string.
+// Values are separated by `;`.
+func (d *ManualDecoder) UserAgentHeader(header http.Header) string {
+	var ua string
```
graphaelli:
```suggestion
	return strings.Join(header[textproto.CanonicalMIMEHeaderKey("User-Agent")], "; ")
```

graphaelli: rfc 2616/7230 say to use comma:

> A recipient MAY combine multiple header fields with the same field
   name into one "field-name: field-value" pair, without changing the
   semantics of the message, by appending each subsequent field value to
   the combined field value in order, separated by a comma.

Issue #2201, opened 2019-05-23
[self instrumentation] Enhance Logging


graphaelli:
Audit logging selectors, log levels, and messages with a goal of making error logging useful on its own, and making it clear which selectors should be enabled in debug mode.

graphaelli:
Just came across one issue - not all apm-server http requests are logged in the same way:

`/` and `/intake/v2/events` happy paths include `response_code`
```
INFO	[request]	beater/common_handler.go:184	handled request	{"request_id": "6f3e2013-ebd2-405f-ad1f-a4f02e252109", "method": "GET", "URL": "/", "content_length": 0, "remote_address": "127.0.0.1", "user-agent": "Go-http-client/1.1", "response_code": 200}
INFO	[request]	beater/common_handler.go:184	handled request	{"request_id": "2a98868c-6b44-403d-951d-c82922250e0c", "method": "POST", "URL": "/intake/v2/events", "content_length": 3585, "remote_address": "127.0.0.1", "user-agent": "curl/7.54.0", "response_code": 202}
```

`/intake/v2/events` errors do not:
```
ERROR	[request]	beater/intake_handler.go:93	error handling request	{"request_id": "c35b6689-c617-46e1-8cb2-1cb4fce940db", "method": "POST", "URL": "/intake/v2/events", "content_length": 3585, "remote_address": "127.0.0.1", "user-agent": "curl/7.54.0", "error": "queue is full"}
[...]

PR #2278, opened 2019-06-12
abort server startup on sourcemap configuration error

graphaelli:
found while working on #2201, it impacts a few operations that may fail:

https://github.com/elastic/apm-server/blob/51942e8429f3073191988e438cd98e31cba99396/beater/route_config.go#L145

https://github.com/elastic/apm-server/blob/51942e8429f3073191988e438cd98e31cba99396/beater/route_config.go#L193


graphaelli:
needs changelog

PR #2281, opened 2019-06-13
Make client_auth optional by default.

simitt:
This PR changes `ssl.client_authentication` from `required` to `optional`,  meaning that client certificates are verified if available. 

Client authentication is not considered to be a requirement for secure communication between APM agents and APM server, as an additional _token_ can be configured and provided for authentication. None of the agents currently support client authentication, so enforcing it as default does not seem reasonable. Changing the default would also mean a breaking change.

### DIFF
--- a/tests/system/test_access.py
+++ b/tests/system/test_access.py
@@ -165,7 +165,7 @@ class TestSSLEnabledNoClientVerificationTest(TestSecureServerBaseTest):
 
 
 class TestSSLEnabledOptionalClientVerificationTest(TestSecureServerBaseTest):
-    # no ssl_overrides necessary as `required` is default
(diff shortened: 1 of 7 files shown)


tests/system/test_access.py
```diff
@@ -165,8 +165,7 @@ def test_https_server_validation_fails(self):
 
 
 class TestSSLEnabledOptionalClientVerificationTest(TestSecureServerBaseTest):
-    def ssl_overrides(self):
-        return {"ssl_client_authentication": "optional"}
+    # no ssl_overrides necessary as `required` is default
```
graphaelli:
```suggestion
    # no ssl_overrides necessary as `optional` is default
```

simitt:
jenkins, run the tests please

PR #2284, opened 2019-06-13
[6.8] Make client_auth optional by default. (#2281)

simitt:
Backports the following commits to 6.8:

* Enable overwriting ingest pipelines by default. (#2281)


simitt:
jenkins, run the tests please

PR #2285, opened 2019-06-13
[6.8] Update go version to 1.12.4

simitt:
Requiring go version `1.12.4` instead of `1.10` as `1.10` is no longer supported.


graphaelli:
@elastic/beats any concerns over this change?  libbeat 6.8 is on [1.10.8](https://github.com/elastic/beats/blob/6.8/.go-version)

urso:
> @elastic/beats any concerns over this change? libbeat 6.8 is on 1.10.8 

We normally try to keep go versions for older Beats releases on older Go releases. E.g. updating to go1.12 might enable some TLS ciphers/settings that didn't exist in the past. 

All in all thanks to Go compatiblity agreement, I have no objections (as long as CI is still green).

ph:
@graphaelli @urso 

Just a note here, I think the only major things in dropping 1.10 is losing support for Windows XP, and since its ancient, I don't see any problem to update. Concerning 1.12,  assuming that beats follow that move and update 6.8 to 1.12, I think one or two tests related to TLS will fails. Because of the range of supported TLS 1.0 vs 1.1.

Also, one thing to consider, **we are doing a minor bump of Golang in a patch release of apm-server.**

graphaelli:
Good point on the XP support, but I also agree that's not critical as we don't actually support that OS.  I don't think there are any guarantees broken by changing go even in a minor version thanks to, as @urso mentioned, the go 1 compatibility promise.  We don't have reason to expect performance related issues.

@simitt Did you consider updating the go tools from `release-branch.go1.10` as well?  They should not affect apm-server directly so I'm fine to skip them, just want to make sure it's intentional.

simitt:
Closing for now as this needs more investigation in case we want to update.

PR #2291, opened 2019-06-17
Add integration tests for TLS cipher suites.

simitt:
Add tests to reflect default cipher suites and possible cipher suites to choose from for APM Server.


simitt:
Closing for now, need to fix the openssl version requirements on Jenkins, so the tests pass.

PR #2292, opened 2019-06-17
[docs] 7.2 release highlights

bmorelli25:
Add release highlights and breaking changes for `7.2`

Will merge into `master` and `7.x` today. Will merge into `7.2` on day of release.


lcawl:
Can we add a tagged region in the highlights page so that we can start pulling content into the Installation and Upgrade Guide?  For now it can just be an empty tagged region (like this one at the top of your breaking changes), for example:

> //NOTE: The notable-highlights tagged regions are re-used in the
//Installation and Upgrade Guide
// tag::notable-v8-highlights[]
>
>// end::notable-v8-highlights[]

bmorelli25:
@lcawl - added an empty tagged region in 8bc7098. Can you explain why you don't want any content in the tagged region yet?

bmorelli25:
Ahh, I see, this is for 8.0

PR #2301, opened 2019-06-18
Enable APM pipeline by default.

### LINKED ISSUE #1972: Enable ingest pipelines by default

simitt:
Register and apply geo-ip and user-agent pipeline by default.

closes elastic/apm-server#1972 

* [x] update docs - @bmorelli25 please have a look at the changes. Once this is merged I can port them to libbeat.
* [x] changelog
* [x] notify cloud team 
* [x] manually verify that apm pipeline is registered and applied for output ES 
* [x] manually verify that apm pipeline is not registered and applied, but events properly ingested when using logstash output

### DIFF
--- a/apm-server.yml
+++ b/apm-server.yml
@@ -157,24 +157,19 @@ apm-server:
     # Remote apm-servers' secret_token
     #secret_token:
 
-  # A pipeline is a definition of processors applied to documents when writing them to Elasticsearch.
+  # A pipeline is a definition of processors applied to documents when ingesting them to Elasticsearch.
   # Using pipelines involves two steps:
   # (1) registering a pipeline
-  # (2) applying a pipeline during data ingestion (see `output.elasticsearch.pipelines`)
+  # (2) applying a pipeline during data ingestion (see `output.elasticsearch.pipeline`)
   #
-  # You can manually register pipelines, or use this configuration option to ensure
-  # pipelines are loaded and registered at the configured Elasticsearch instances.
-  # Automatic pipeline registration requires
-  # * `output.elasticsearch` to be enabled and configured.
-  # * having the required Elasticsearch Processor Plugins installed.
-  #   APM Server default pipelines require you to have the `Ingest User Agent Plugin` installed.
-  #   Find the default pipeline configuration at `ingest/pipeline/definition.json`.
-  #
+  # You can manually register a pipeline, or use this configuration option to ensure
+  # the pipeline is loaded and registered at the configured Elasticsearch instances.
+  # Find the default pipeline configuration at `ingest/pipeline/definition.json`.
+  # Automatic pipeline registration requires the output.elasticsearch` to be enabled and configured.
   #register.ingest.pipeline:
-    # Registers pipeline definitions in Elasticsearch on APM Server startup. Defaults to false.
-    #enabled: false
-
-    # Overwrites existing pipeline definitions in Elasticsearch. Defaults to true.
+    # Registers APM pipeline definition in Elasticsearch on APM Server startup. Defaults to true.
+    #enabled: true
+    # Overwrites existing APM pipeline definition in Elasticsearch. Defaults to true.
     #overwrite: true
 
   # When ilm is set to true the APM Server will take care of ILM setup.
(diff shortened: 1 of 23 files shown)


apm-server.yml
```diff
@@ -384,13 +379,11 @@ output.elasticsearch:
...
-  # Pipelines are disabled by default. To make use of them you have to:
-  # (1) ensure pipelines are registered in Elasticsearch, see `apm-server.register.ingest.pipeline`
-  # (2) enable the following:
-  #pipelines:
-  #- pipeline: "apm"
+  # A pipeline is a definition of processors applied to documents when ingesting them to Elasticsearch.
+  # APM Server comes with a default pipeline definition, located at `ingest/pipeline/definition.json`, which is
+  # loaded to Elasticsearch by default (see `apm-server.register.ingest.pipeline`).
+  # APM pipeline is enabled by default. For disabling set `pipeline: _none`.
+  #pipeline: "apm"
```
bmorelli25: For my knowledge, this can still be an array, right? If I wanted to register multiple pipelines

docs/configuration-process.asciidoc
```diff
@@ -111,13 +111,14 @@ Disabled by default.
 [[register.ingest.pipeline.enabled]]
 [float]
 ==== `register.ingest.pipeline.enabled`
-Loads pipeline definitions to Elasticsearch on APM Server startup.
-Defaults to false.
+Loads default pipeline definition to Elasticsearch on APM Server startup.
```
bmorelli25:
```suggestion
Loads the default pipeline definition to Elasticsearch on APM Server startup.
```

docs/configuration-process.asciidoc
```diff
@@ -111,13 +111,14 @@ Disabled by default.
...
+Defaults to true.
 
 [[register.ingest.pipeline.overwrite]]
 [float]
 ==== `register.ingest.pipeline.overwrite`
-Overwrites existing pipeline definitions in Elasticsearch. Defaults to true.
+Overwrites existing APM pipeline definition in Elasticsearch.
```
bmorelli25:
```suggestion
Overwrites the existing APM pipeline definition in Elasticsearch.
```

docs/copied-from-beats/outputconfig.asciidoc
```diff
@@ -492,6 +492,9 @@ output.elasticsearch:
 With this configuration, all events with `processor.event: transaction` are sent to a pipeline
 named `transaction_pipeline`. Similarly, all events with `processor.event: error` are sent to a
 pipeline named `error_pipeline`.
+
+The pipeline defaults to `apm`, adding user agent and geo ip information to events.
+For disabling any pipeline you can configure `output.elasticsearch.pipeline: _none`.
```
bmorelli25: What if we updated these two lines to:
```
The default pipeline is `apm`. It adds user agent and geo ip information to events.
To disable this, or any other pipeline, set `output.elasticsearch.pipeline: _none`. 
```

docs/copied-from-beats/outputconfig.asciidoc
```diff
@@ -626,6 +629,8 @@ output.elasticsearch:
 With this configuration, all events with `processor.event: transaction` are sent to a pipeline
 named `transaction_pipeline`, all events with `processor.event: error` are sent to a
 pipeline named `error_pipeline`, etc.
+
+If any pipelines are defined, the default `apm` pipeline will be deactivated.
```
bmorelli25:
```suggestion
NOTE: Defining any pipeline will deactivate the default `apm` pipeline.
```

apm-server.yml
simitt: Multiple pipelines would be registered in `pipelines`, since we only have one I changed the default to use `pipeline`.

simitt:
Aligning the pipeline descriptions in a follow up PR sounds good. 
To give some context: the default `apm` pipeline consists of the [geo-ip](https://www.elastic.co/guide/en/elasticsearch/reference/master/geoip-processor.html#using-ingest-geoip) and the [user-agent](https://www.elastic.co/guide/en/elasticsearch/reference/master/user-agent-processor.html) processors and will be enabled by default. I don't think we need to go into too much detail explaining them, but linking to the ES reference might help.

beater/config.go
```diff
@@ -268,11 +271,14 @@ func defaultConfig(beatVersion string) *Config {
...
+					Overwrite: &pipeline,
 					Path: paths.Resolve(paths.Home,
 						filepath.Join("ingest", "pipeline", "definition.json")),
 				}},
 		},
-		Mode:   ModeProduction,
-		Kibana: common.MustNewConfigFrom(map[string]interface{}{"enabled": "false"}),
+		Mode:     ModeProduction,
+		Kibana:   common.MustNewConfigFrom(map[string]interface{}{"enabled": "false"}),
+		pipeline: "apm",
```
jalvz: ideally the pipeline name is a const (also in beater_test)

publish/pub.go
```diff
@@ -40,7 +40,7 @@ type Reporter func(context.Context, PendingReq) error
 // queue size. As the publisher is not waiting for the outputs ACK, the total
 // number requests(events) active in the system can exceed the queue size. Only
 // the number of concurrent HTTP requests trying to publish at the same time is limited.
-type publisher struct {
+type Publisher struct {
```
jalvz: why is this public?

simitt: Because it is returned in [NewPublisher](https://github.com/elastic/apm-server/pull/2301/files/7472c1f587def4e3b62d8d0a8964d8e05602e349#diff-f2dc32e949f35561380c5c85bc282679R73) which is a public method. Returned types should also be public.

PR #2232, opened 2019-05-31
require clean golint

graphaelli:
The first run should fail due to intentional introduction of lint failure, expecting:

```
 make check-full
...
.../github.com/elastic/apm-server/utility/data_fetcher.go:275:1: exported function FooBar should have comment or be unexported
make: *** [golint] Error 1
```

closes elastic/apm-server#2221

PR #2313, opened 2019-06-21
[7.x] require clean golint (#2232)

graphaelli:
 - require clean golint  (#2232)


jalvz:
Are you doing this out of fear of conflicts in the Makefile or because you see value of linting backports?

simitt:
`master` is reflecting features for 8.0. Maybe some features will only be part of `7.x` but never make it to master, or backports need some changes, as not everything is backported. I think having the linter in place for such cases is a nice-to-have.

PR #2287, opened 2019-06-17
Merge metricset data instead of overwrite them.

simitt:
fixes elastic/apm-server#2286

PR #2315, opened 2019-06-24
Adds fields to support breakdown graphs in the UI

### LINKED ISSUE #2306: Define fields supporting breakdown graphs
Specifically, as described in https://github.com/elastic/apm/issues/78

That is not final, so we can only follow and adapt as reqs change

jalvz:
fixes elastic/apm-server#2306 

- [x] needs changelog
- [x] needs kibana sync
- [x] needs manual testing


docs/spec/metricsets/metricset.json
```diff
@@ -20,6 +20,61 @@
...
+                                    "properties": {
+                                        "count": {
+                                            "type": "number"
+                                        },
+                                        "sum.us": {
+                                            "type": "number"
+                                        }
+                                    }
+                                },
+                                "breakdown.count": {
```
jalvz: i guess this is not ok, and I need to create a `breakdown` object with `count` inside, right?

model/transaction/_meta/fields.yml
```diff
@@ -16,6 +16,17 @@
               description: >
                 Total duration of this transaction, in microseconds.
 
+            - name: count
```
jalvz: so these fields (and others) correspond to the `transaction` type as it comes in the `metricsets`, but they are not passed in the `transaction` events.  is this the right place to put the mapping?

jalvz: i think only `breakdown.count` is strictly required by UI now, so I can remove the others

model/transaction/_meta/fields.yml
```diff
@@ -16,6 +16,17 @@
...
+              type: long
+              count: 1
```
jalvz: i couldn't find what is this `count`, just copied

model/transaction/_meta/fields.yml
simitt: I'd either define it in the `_meta/fields.common.yml` or introduce a `transaction`  namespace in the `metricset` package. It shouldn't really change anything, as in the end it's all combined in the `fields.yml`, but I think it is easier to find again.

model/transaction/_meta/fields.yml
simitt: I think you can safely remove them from these field mappings.
If I recall correctly they are written to the Kibana Index Pattern in the index-pattern/template creation process and we only added it for some fields where it was necessary for the dashboards and visualizations.

docs/spec/metricsets/metricset.json
simitt: Can you try it out with sample requests using `breakdown.count:` and `breakdown: { count: }` notation?

model/metricset/event.go
```diff
@@ -206,21 +218,34 @@ func (me *Metricset) Transform(tctx *transform.Context) []beat.Event {
...
+
 	fields["processor"] = processorEntry
 	tctx.Metadata.Set(fields)
 
 	// merges with metadata labels, overrides conflicting keys
 	utility.DeepUpdate(fields, "labels", me.Labels)
-	utility.DeepUpdate(fields, transactionKey, me.Transaction.fields())
-	utility.DeepUpdate(fields, spanKey, me.Span.fields())
+	utility.DeepUpdate(fields, transactionKey, transactionFields)
+	utility.DeepUpdate(fields, spanKey, spanFields)
```
simitt: Why are the changes in the transformation method necessary? I see you are changing the order of merging the data, but I don't see any specific reason for one or the other. Apart from that I don't see why those changes would be necessary and the code gets more complicated than before.

jalvz: Not sure I understand your question. 
`transaction` and `span` fields inside `samples` need to be pulled up one level, so I collect them in L227 and L229 and move them in L237 and L240, for the result that you can see in the approved files.
Other ideas of how to achieve the same?

> you are changing the order of merging the data

Where am I doing that?

simitt: Current master already pulls out span and transaction information from samples and merges it with top level span and transaction information. That has been done in PR #2265 and #2287 . Therefore I don't understand why this method needs to be changed at all.
From my understanding what was missing is the validation of defined metric attributes on the Intake API level and the mappings in the templates. 
Please correct me if I'm wrong, I might overlook something, but that's my understanding from this. 


>>you are changing the order of merging the data

>Where am I doing that?

so far 
L211: `if _, err := fields.Put(sample.Name, sample.Value); err != nil ` has put span and transaction info from sample to root span and transaction keys
And then information that had been sent top level has been deep merged to them:
L222: `utility.DeepUpdate(fields, transactionKey, me.Transaction.fields())`
L223: `utility.DeepUpdate(fields, spanKey, me.Span.fields())`

Now you first take the root level span and transaction fields and then deep update with the information extracted from samples. 

I don't think the order matters, but as it shouldn't matter here I would go with the existing simpler solution.

jalvz: > Current master already pulls out span and transaction information from samples

I missed that, then. Ill have another look, thanks

Jenkinsfile
```diff
@@ -367,7 +367,7 @@ pipeline {
         expression { return params.kibana_update_ci }
       }
       steps {
-        withGithubNotify(context: 'Synk Kibana') {
+        withGithubNotify(context: 'Sync Kibana') {
```
simitt: 👍

docs/fields.asciidoc
```diff
@@ -16,6 +16,8 @@ grouped in the following categories:
 * <<exported-fields-apm-error>>
 * <<exported-fields-apm-sourcemap>>
 * <<exported-fields-apm-span>>
+* <<exported-fields-apm-span>>
+* <<exported-fields-apm-transaction>>
 * <<exported-fields-apm-transaction>>
```
simitt: Are spans and transactions listed twice now?

processor/stream/package_tests/transaction_attrs_test.go
```diff
@@ -54,13 +54,17 @@ func transactionFieldsNotInPayloadAttrs() *tests.Set {
...
 		"transaction.marks.*.*",
 		tests.Group("observer"),
 		tests.Group("user"),
 		tests.Group("client"),
 		tests.Group("url"),
 		tests.Group("http"),
 		tests.Group("span"),
+		tests.Group("transaction.self_time"),
+		tests.Group("transaction.breakdown"),
+		tests.Group("transaction.duration.sum"),
```
simitt: You shouldn't need to exclude `transaction.*` attributes that we expect to be indexed. If that's necessary then this uncovers a real problem. Same for spans above.

model/metricset/_meta/fields.yml
```diff
@@ -78,3 +78,58 @@
             format: bytes
             description: >
               The Resident Set Size. The amount of memory the process occupied in main memory (RAM).
+
+
+- key: apm-transaction
```
simitt: This is not nested under `transaction`, same for `span`. They are both top-level. That's the reason for [the failing test](https://apm-ci.elastic.co/blue/organizations/jenkins/apm-server%2Fapm-server-mbp%2FPR-2315/detail/PR-2315/3/pipeline/).

processor/stream/package_tests/transaction_attrs_test.go
jalvz: as per offline discussion: keeping these as they are not in the corresponding `testdata` event

docs/fields.asciidoc
jalvz: thanks :man_facepalming: added to common fields instead

PR #2316, opened 2019-06-24
docs: add .NET agent links and information

bmorelli25:
Adds the .NET agent to the APM Overview documentation. 

Closes https://github.com/elastic/apm-agent-dotnet/issues/313.

Needs to be backported to `7.x` and `7.2`

### DIFF
--- a/docs/guide/install-and-run.asciidoc
+++ b/docs/guide/install-and-run.asciidoc
@@ -114,7 +114,7 @@ NOTE: Check the <<agent-server-compatibility,Agent/Server compatibility matrix>>
 |{apm-java-ref}/configuration.html[Config]
 
 .2+|.NET
-3+|The .NET agent automatically instruments ASP.NET Core, while the Full framework can be instrumented with the public API.
+3+|The .NET agent automatically instruments ASP.NET Core applications, while .NET Framework applications can be instrumented with the public API.
 |{apm-dotnet-ref-v}/intro.html[Introduction]
 |
 |{apm-dotnet-ref-v}/configuration.html[Config]
--- a/docs/guide/install-and-run.asciidoc
+++ b/docs/guide/install-and-run.asciidoc
@@ -114,7 +114,7 @@ NOTE: Check the <<agent-server-compatibility,Agent/Server compatibility matrix>>
 |{apm-java-ref}/configuration.html[Config]
 
 .2+|.NET
-3+|The .NET agent comes with ASP.NET Core support out of the box.
+3+|The .NET agent automatically instruments ASP.NET Core, while the Full framework can be instrumented with the public API.
 |{apm-dotnet-ref-v}/intro.html[Introduction]
 |
 |{apm-dotnet-ref-v}/configuration.html[Config]
--- a/docs/guide/install-and-run.asciidoc
+++ b/docs/guide/install-and-run.asciidoc
@@ -88,7 +88,7 @@ please see the full {apm-server-ref-v}/index.html[APM Server documentation].
 === Install and configure APM agents
 
 Agents are written in the same language as your service.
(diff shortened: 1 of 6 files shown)


docs/guide/install-and-run.asciidoc
```diff
@@ -113,6 +113,12 @@ NOTE: Check the <<agent-server-compatibility,Agent/Server compatibility matrix>>
 |{apm-java-ref}/supported-technologies-details.html[Supported technologies]
 |{apm-java-ref}/configuration.html[Config]
 
+.2+|.NET
+3+|The .NET agent comes with ASP.NET Core support out of the box.
```
gregkalapos: I usually say we have "auto instrumentation for ASP.NET Core". Because the agent API also works with Full Framework (the old stuff), it's just we don't have atm any auto instrumentation capability for that... So it'd be nice to somehow hint that the agent works on every .NET Flavor, but autoinstrumentation is currently ASP.NET Core only.

docs/guide/install-and-run.asciidoc
```diff
@@ -88,7 +88,7 @@ please see the full {apm-server-ref-v}/index.html[APM Server documentation].
 === Install and configure APM agents
 
 Agents are written in the same language as your service.
-Currently Elastic APM has agents for Go, Java, Node.js, Python, Ruby, and JavaScript RUM.
+Currently Elastic APM has agents for Go, Java, .NET, Node.js, Python, Ruby, and JavaScript RUM.
```
gregkalapos: 🎉

docs/guide/install-and-run.asciidoc
bmorelli25: Gotcha. What about something like:

> The .NET agent automatically instruments ASP.NET Core, while the Full framework can be instrumented with the public API.

bmorelli25: Pushed a commit with that change ⬆️

I'm going to merge this once CI passes so I can get the backports merged in time for the release. If there are still changes you'd like to see, just let me know! I have no problem opening up a new PR with your suggestions.

gregkalapos: > Gotcha. What about something like:
> 
> > The .NET agent automatically instruments ASP.NET Core, while the Full framework can be instrumented with the public API.

This sounds good to me 👍

SergeyKleyman: The official terminology is just `.NET Framework` and `.NET Core`. Here is [an example in Microsoft docs](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.trace.write?view=netframework-4.7.2#applies-to). So I wonder if we should use `Full framework` term. Instead of

>The .NET agent automatically instruments ASP.NET Core, while the Full framework can be instrumented with the public API.

I would say something like

> The .NET agent automatically instruments ASP.NET Core applications, while .NET Framework ASP.NET applications can be instrumented with the public API.

bmorelli25: Oooo good catch @SergeyKleyman. We should probably use the official terminology.

gregkalapos: +1 for using `.NET Framework`, but would not say "while .NET Framework ASP.NET applications can", because that not only applies to ASP.NET... it can be anything. 

So I'd propose:
>  The .NET agent automatically instruments ASP.NET Core applications, while .NET Framework  applications can be instrumented with the public API.

(or "applications running on top of .NET Framework")

SergeyKleyman: This sounds good to me 👍

bmorelli25: Boom! Look at that. Who says GitHub can't be synchronous 😂. Thanks to both of you 😃

PR #2325, opened 2019-06-25
Allow the Docker image to be run as a random user id

barkbay:
Hi,

I'm working on the compatibility of ECK with Kubernetes and Openshift.
On secured Kubernetes environments the user ID used to run a container is "random", you can't really predict it advance. Consequently we can't start the APM server container because it expects to be run with the user 1000 or 0 (root)

This PR brings some compatibility with such environments, based on the fact that on secured Kubernetes clusters and on Openshift the only thing you know is that the user is always a member of the root group.

You can find more details here: https://docs.openshift.com/container-platform/3.11/creating_images/guidelines.html#openshift-specific-guidelines

Thank you

### DIFF
--- a/_beats/dev-tools/packaging/templates/docker/Dockerfile.tmpl
+++ b/_beats/dev-tools/packaging/templates/docker/Dockerfile.tmpl
@@ -17,6 +17,7 @@ LABEL \
 
 ENV ELASTIC_CONTAINER "true"
 ENV PATH={{ $beatHome }}:$PATH
+ENV BEAT_STRICT_PERMS "false"
 
 COPY beat {{ $beatHome }}
 COPY docker-entrypoint /usr/local/bin/docker-entrypoint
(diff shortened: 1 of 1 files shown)


graphaelli:
This file is actually owned by the beats platform and will be overridden on the next beats update.  We could add a custom Dockerfile template to support this but I'm not comfortable with using this for our generally distributed docker container images as I don't think strict perms shouldn't be relaxed by default.  

This is something we could solve in collaboration with @elastic/beats as they will/have likely run into the same issues.

@barkbay Can you please open an issue describing the environment targeted more in depth so we can understand the issue better and help reach the right solution?

graphaelli:
Also, we don't have this in our docs but perhaps it applies here too? https://www.elastic.co/guide/en/beats/filebeat/master/running-on-kubernetes.html#_red_hat_openshift_configuration

barkbay:
I will open an issue in the `beats` project, but I guess it's fine to consider that the beats need some privileged permissions as they need to read/gather some system metrics.
The document you mentioned run the container as the `root` user on Openshift, is there any reason to do that with the APM server ?

barkbay:
Issue has been opened here: https://github.com/elastic/beats/issues/12686

barkbay:
Closed in favor of https://github.com/elastic/beats/pull/12905

PR #2326, opened 2019-06-25
[7.x] Update beats framework to 51566ec

graphaelli:
needed for backporting #2281 to 7.x

PR #2327, opened 2019-06-25
[7.x] Add config options for Server/Agent communication via TLS (#2224)

graphaelli:
 - Add config options for Server/Agent communication via TLS (#2224)
 - Make client_auth optional by default. (#2281)

depends on elastic/apm-server#2326 to fix tests


changelogs/head.asciidoc
```diff
@@ -0,0 +1,11 @@
...
+== APM Server version HEAD
+
+[float]
+==== Added
+- Support more SSL config options for agent/server communication {pull}2224[2224].
+
+[float]
+==== Removed
+
+https://github.com/elastic/apm-server/compare/7.2\...master[View commits]
```
simitt: The changelog entry should probably go into `7.2` and not `head`.

beater/config_test.go
```diff
@@ -84,7 +87,11 @@ func TestConfig(t *testing.T) {
 				WriteTimeout:    4000000000,
 				ShutdownTimeout: 9000000000,
 				SecretToken:     "1234random",
-				SSL:             &SSLConfig{Enabled: &truthy, Certificate: outputs.CertificateConfig{Certificate: "1234cert", Key: "1234key"}},
+				TLS: &tlscommon.ServerConfig{
+					Enabled:     &truthy,
+					CAs:         []string{"./ca.cert.pem"},
+					Certificate: outputs.CertificateConfig{Certificate: "1234cert", Key: "1234key"},
+					ClientAuth:  4}, //4=RequireAndVerifyClientCert
```
simitt: This needs to be changed to `0`, as libbeat changed the default back to `none`.

changelogs/head.asciidoc
jalvz: fwiw, on my duped version of this pr, i chosen to remove the head entry in the conflict resolution, with the intention of doing the usual "fix changelogs" PR later on

graphaelli: moved to 7.2

beater/config_test.go
graphaelli: libbeat now updated

simitt: 7.x is not on the latest libbeat version, that's why `ClientAuth: 4` is ok here. Once we update to the latest, it should be 0 again.

PR #2337, opened 2019-06-26
Add caching to Agent Remote Configuration handling

### LINKED ISSUE #2220: [ACM] Optimization / caching
Goal: reduce requests to Kibana as much as possible. 
Main risk comes from RUM agent. 
This task should include a small benchmark.

Requires https://github.com/elastic/apm-server/pull/2095

simitt:
closes elastic/apm-server#2220

* [x] changelog
* [x] wait for all agent devs to agree on https://github.com/elastic/apm-server/issues/2220#issuecomment-505861445
* [x] update config name depending on feature name
* [x] change headers according to discussions

I did look at a couple of different go caching libraries, but decided to go for the same one that we already use in the sourcemap handling, [patrickmn/go-cache](https://github.com/patrickmn/go-cache), to avoid adding new dependencies. Since this library hasn't changed in the last 2 years, we might want to consider moving to a more maintained one at some point. Before doing so we should have proper integration testing in place so we can run a macro benchmark rather than a micro benchmark, taking into account different cache expiration strategies of underlying cache implementations.


_meta/beat.yml
```diff
@@ -178,6 +178,10 @@ apm-server:
   #ilm:
     #enabled: false
 
+  # When using agent remote configuration, information fetched from Kibana will be cached in memory for some time.
+  # Specify cache key expiration via this setting. Default is 30 seconds.
+  #remote_config.agent.cache.expiration: 30s
```
jalvz: have you considered _not_ introducing a user-facing setting?

agentcfg/cache.go
```diff
@@ -0,0 +1,81 @@
...
+)
+
+type cache struct {
+	logger  *logp.Logger
+	exp     time.Duration
+	gocache *gocache.Cache
+}
+
+func newCache(logger *logp.Logger, exp time.Duration) *cache {
+	if logger == nil {
```
jalvz: this seems an unnecessary complication, rather pass the logger in `cache_test/newCacheSetup`

agentcfg/cache.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+
+	// add resource to cache
+	c.add(id, doc)
+
+	return doc, err
+}
+
+func (c *cache) add(id string, doc *Doc) {
+	c.gocache.Set(id, doc, c.exp)
```
jalvz: I'm curious about why do you need to store the expiration the `cache` struct, and why you call `Set` instead of `SetDefault`? It is because of "if-in-the-future..."?

agentcfg/cache.go
```diff
@@ -0,0 +1,81 @@
...
+		gocache: gocache.New(exp, cleanupInterval)}
+}
+
+func (c *cache) fetchAndAdd(q Query, fn func(Query) (*Doc, error)) (doc *Doc, err error) {
+	id := q.id()
+
+	// return from cache if possible
+	doc, found := c.fetch(id)
+	if found {
+		return doc, err
```
jalvz: in case unintended: when you use named return parameters, a simple `return` will suffice (same bellow)

jalvz:
> Since this library hasn't changed in the last 2 years, we might want to consider moving to a more maintained

Either is fine by me, but you can also flip this statement on its head and say that this one is "stable" rather than "unmaintained" - not all software has to keep changing forever.

OTOH, with the number of dependencies that we already have, one more or one less is not an issue.

jalvz:
> Since this library hasn't changed in the last 2 years, we might want to consider moving to a more maintained

Either is fine by me, but you can also flip this statement on its head and say that this one is "stable" rather than "unmaintained" - not all software has to keep changing forever.
OTOH, with the number of dependencies that we already have, one more or one less is not an issue.  

Good job so far, thorough testing as always :100:

agentcfg/cache.go
simitt: I don't understand, passing in the logger here has nothing to do with any testing logging. I just pass the logger from the fetcher to the cache.

_meta/beat.yml
simitt: Yes I have. I am a bit twisted on whether or not it is necessary, but went for more flexibility with it. I am also ok with removing this if you don't think it is needed. 
@graphaelli any opinion?

agentcfg/cache.go
simitt: The used library requires to specify a default expiration when instantiating the cache. Since we agreed on using the same expiration for all keys (services) it is not necessary to pass it in again with every `add`. 
But you're right I can use `SetDefault` instead, using `Set` is a relict from my first version, where I kept the flexibility to set the key expiration dynamic per service.

agentcfg/cache.go
jalvz: sure, i just meant the `if nil` check

simitt: The `logger` is a pointer, so it could be `nil`. I know, calling the function it is under our control, so one can argue it is not necessary as we always pass a logger in.. but I don't think it hurts and I rather avoid panics than being lazy with nil checks.

_meta/beat.yml
graphaelli: might as well keep it

simitt:
jenkins, run the tests please

agentcfg/cache.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+
+	// add resource to cache
+	c.add(id, doc)
+	return
+}
+
+func (c *cache) add(id string, doc *Doc) {
+	c.gocache.SetDefault(id, doc)
+	c.logger.Debugf("Cache size %v. Added ID %v.", c.gocache.ItemCount(), id)
```
graphaelli: Thoughts on wrapping this in `c.logger.IsDebug()` ?  This shouldn't be a very hot path but `ItemCount` is relatively expensive, taking a lock.  I think we do the same for sourcemaps and was thinking the same treatment should go there too.

simitt: Thanks for the suggestion, that's a great catch.

_meta/beat.yml
graphaelli: revisiting, can we drop the remote from this - how about `agent_config.cache.expiration`

simitt: I changed to `agent.config.cache.expiration` in case we need to nest additional configs in the future.

Issue #2336, opened 2019-06-26
Adapt TLS default setting to recent changes in libbeat


simitt:
Beats recently changed to set `client_auth` to default values `required` in case `CA` is configured for TLS server config. 
APM Server sets `client_auth` to `optional` by default if nothing is configured. This needs to be adapted to only set it to optional if no `CA` is configured, otherwise the check in beats does not get applied.

PR #2340, opened 2019-06-27
[TLS] Align client_auth handling to beats changes and add description for verification_mode

simitt:
This PR adapts default settings for `client_authentication` when TLS is configured to latest changes in libbeat. It also documents and adds tests for the `verification_mode` setting. 

* [x] changelog entry
* [x] upgrade libbeat  

implements elastic/apm-server#2336

### DIFF
--- a/beater/config_test.go
+++ b/beater/config_test.go
@@ -167,39 +167,13 @@ func TestConfig(t *testing.T) {
 	}
 	for idx, test := range cases {
 		cfg, err := yaml.NewConfig(test.config)
-		assert.NoError(t, err)
-
+		require.NoError(t, err)
 		var beaterConfig Config
 		err = cfg.Unpack(&beaterConfig)
-		assert.NoError(t, err)
+		require.NoError(t, err)
+
 		msg := fmt.Sprintf("Test number %v failed. Config: %v, ExpectedConfig: %v", idx, beaterConfig, test.expectedConfig)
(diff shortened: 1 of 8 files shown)


beater/config_test.go
```diff
@@ -304,3 +278,94 @@ func TestPipeline(t *testing.T) {
...
+			},
+			"DefaultRequiredByCA": {
+				config: map[string]interface{}{"ssl": map[string]interface{}{
+					"certificate_authorities": []string{"./path"}}},
+				tls: &tlscommon.ServerConfig{ClientAuth: 4},
+			},
+			"ConfiguredWithCA": {
+				config: map[string]interface{}{"ssl": map[string]interface{}{
+					"certificate_authorities": []string{"./path"}, "client_authentication": "none"}},
+				tls: &tlscommon.ServerConfig{ClientAuth: 0},
```
graphaelli: This `ClientAuth` comes from [these constants](https://golang.org/pkg/crypto/tls/#ClientAuthType) right?  Can we use those instead?

simitt: Actually [tlsClientAuth](https://github.com/elastic/beats/blob/master/libbeat/common/transport/tlscommon/types.go#L122) is an unexported type from the libbeat package. I cannot use go's constants here.

graphaelli: I figured it was something strange like that, thanks.

PR #2343, opened 2019-06-27
ci(jenkins): notify first time contributor github check

v1v:
Enable `githubNotifyFirstTimeContributor` parameter to notify the PR with a new GitHub check whether the checkout failed related with the first time contributor validation.

## Highlights
- Rather than creating a new GitHub check for each pipeline let's use the shared library to drive that particular behavior.
- Being explicit behavior rather than a default one. 

## How does it look like?

image


v1v:
jenkins run the tests please

Issue #2241, opened 2019-06-03
[ILM] Enable ILM by default when possible


simitt:
To furthermore leverage ILM support, enable it whenever possible, while ensuring changing the default doesn't break anything in current setups. 

This includes 
* checking whether or not the configured ES instance supports ILM
* and respecting configured `output.elasticsearch.index` and `output.elasticsearch.indices`

Issue #2240, opened 2019-06-03
[ILM] Make policies more flexible


simitt:
As a follow up on having basic ILM support added to APM, policies should be more flexible to support advanced user needs. 
Find a good way of making policies more flexible, e.g. by 
* providing multiple pre-defined policies
* ship a policy json file which can be changed by users
* allow to define phases and actions per event type in the `apm-server.yml` config file
* ..

graphaelli:
Perhaps a quick alternative would be to bundle multiple named policies that users can use as a baseline and them customize via the UI.

PR #2356, opened 2019-07-01
[ILM] Enable ILM by default when supported by configured ES output.

simitt:
This PR changes default value for `apm-server.ilm.enabled` from `false` to `auto`.

On startup or setup cmd the server determines if preconditions are fulfilled for enabling ILM.
ILM will be enabled when supported by configured ES, and no custom index and indices settings are configured for `output.elasticsearch`.

When `apm-server.ilm.enabled` is set to `false`, ILM is turned off. 

When `apm-server.ilm.enabled` is set to `true`, the server checks if custom `output.elasticsearch.index` or `output.elasticsearch.indices` configuration is set. If set, the server refuses to start up and fails with according error message. 
Otherwise the server also checks if ES supports ILM. Since the configured ES might not immediately be available (temporary connection loss, currently starting up, ..) the server cannot produce a hard failure. Once the server retrieves a connection to ES, it logs a warning and automatically falls back to ordinary index setup if ILM is not supported. 

* [x] changelog
* [x] notify cloud
* [ ] adapt docs (@bmorelli25 could you change the ILM docs to reflect these changes?)
* [ ] add testcases to testplan
* [x] adapt system tests to changed default

implements elastic/apm-server#2241 

====== 
Special testcases:
 * `./apm-server -e -E apm-server.ilm.enabled=true -E setup.template.name=custom -E setup.template.pattern=custom -E output.elasticsearch.index=custom` 
~refuses to startup with message 
```Exiting: `apm-server.ilm.enabled:true` is not allowed when custom `output.elasticsearch` index names are configured```~
Update: logs a message, enables ILM and ignores `index` configurtaion as otherwise it would be a breaking change to what users could have set up upgrading from `7.2`.
 * `./apm-server -e -E apm-server.ilm.enabled=true` with ES not supporting ILM: 
starts but falls back to ordinary index management and is logging: 
`ERROR	[index-management]	idxmgmt/manager.go:97	automatically disabled ILM as not supported by configured output`
 * `./apm-server -e  -E apm-server.ilm.enabled=auto -E setup.template.name=custom -E setup.template.pattern=custom -E output.elasticsearch.index=custom` 
falls back to ordinary index management and is logging: 
```WARN	[index-management]	idxmgmt/manager.go:97	Automatically disabled ILM as custom index settings configured.```
 * `./apm-server -e  -E apm-server.ilm.enabled=auto` with ES not supporting ILM 
falls back to ordinary index management and is logging: 
 `WARN	[index-management]	idxmgmt/manager.go:94	Automatically disabled ILM as not supported by configured output.`

Same behavior when testing with `setup` command.

### DIFF
--- a/_meta/beat.yml
+++ b/_meta/beat.yml
@@ -172,11 +172,21 @@ apm-server:
     # Overwrites existing APM pipeline definition in Elasticsearch. Defaults to true.
     #overwrite: true
 
-  # When ilm is set to true the APM Server will take care of ILM setup.
-  # Configurations in output.elasticsearch.index and output.elasticsearch.indices sections will be ignored
-  # ilm can only be enabled for output.elasticsearch
-  #ilm:
-    #enabled: false
+  # When ilm is set to `auto`, the APM Server checks a couple of preconditions:
+  # If a different output than Elasticsearch is configured, ILM will be disabled.
+  # If Elasticsearch output is configured, but specific `index` or `indices` settings are configured, ILM will be
+  # disabled, as it only works with default index settings.
+  # If the configured Elasticsearch instance is not eligible for ILM, ILM will also be disabled.
+  # If all preconditions are met, ILM will be enabled.
+  #
+  # When ILM is set to `true`, the APM Server ignores any configured index settings.
+  # For ILM to be applied, The configured output must be set to Elasticsearch and the instance
+  # needs to support ILM. Otherwise APM Server falls back to ordinary index management without ILM.
+  #
+  # Disable ILM by setting it to `false`.
+  #
+  # Defaults to `auto`.
+  #ilm.enabled: "auto"
 
   #kibana:
     # For agent remote configuration, enabled must be true
--- a/apm-server.docker.yml
+++ b/apm-server.docker.yml
@@ -172,11 +172,21 @@ apm-server:
     # Overwrites existing APM pipeline definition in Elasticsearch. Defaults to true.
     #overwrite: true
 
-  # When ilm is set to true the APM Server will take care of ILM setup.
-  # Configurations in output.elasticsearch.index and output.elasticsearch.indices sections will be ignored
-  # ilm can only be enabled for output.elasticsearch
-  #ilm:
-    #enabled: false
+  # When ilm is set to `auto`, the APM Server checks a couple of preconditions:
+  # If a different output than Elasticsearch is configured, ILM will be disabled.
+  # If Elasticsearch output is configured, but specific `index` or `indices` settings are configured, ILM will be
+  # disabled, as it only works with default index settings.
+  # If the configured Elasticsearch instance is not eligible for ILM, ILM will also be disabled.
+  # If all preconditions are met, ILM will be enabled.
+  #
+  # When ILM is set to `true`, the APM Server ignores any configured index settings.
+  # For ILM to be applied, The configured output must be set to Elasticsearch and the instance
+  # needs to support ILM. Otherwise APM Server falls back to ordinary index management without ILM.
+  #
+  # Disable ILM by setting it to `false`.
+  #
+  # Defaults to `auto`.
+  #ilm.enabled: "auto"
 
   #kibana:
     # For agent remote configuration, enabled must be true
--- a/apm-server.yml
+++ b/apm-server.yml
@@ -172,11 +172,21 @@ apm-server:
     # Overwrites existing APM pipeline definition in Elasticsearch. Defaults to true.
     #overwrite: true
 
-  # When ilm is set to true the APM Server will take care of ILM setup.
-  # Configurations in output.elasticsearch.index and output.elasticsearch.indices sections will be ignored
-  # ilm can only be enabled for output.elasticsearch
-  #ilm:
-    #enabled: false
+  # When ilm is set to `auto`, the APM Server checks a couple of preconditions:
+  # If a different output than Elasticsearch is configured, ILM will be disabled.
+  # If Elasticsearch output is configured, but specific `index` or `indices` settings are configured, ILM will be
+  # disabled, as it only works with default index settings.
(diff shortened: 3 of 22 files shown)


simitt:
Yes I can confirm it is behaving properly. You need to pass in `apm-server.ilm.enabled` instead of `ilm.enabled`.

jalvz:
Great, thanks for confirming :man_facepalming: 

Some more observations, probably not very important:
- With ilm auto or true, _and_ no policies exist, logs say "Existing template will be overwritten, as overwrite is enabled" which might make someone believe _their_ template is overwritten

- With auto _and_ index or indices overwrite, there is a log line saying "Automatically disabled ILM", however with auto _and_ another output configured I couldn't find that line

- You said that `./apm-server -e -E apm-server.ilm.enabled=true -E setup.template.name=custom -E setup.template.pattern=custom -E output.elasticsearch.index=custom` should refuse to start, however that doesn't seem to be the case for me: it says "Custom index configuration ignored when ILM is enabled" and moves on

- `apm-server -e -E apm-server.ilm.enabled=true` with ES not supporting ILM does also fallback to ilm disabled (like auto)

simitt:
@jalvz thanks for the feedback! Adressing your observations:

> With ilm auto or true, and no policies exist, logs say "Existing template will be overwritten, as overwrite is enabled" which might make someone believe their template is overwritten

There is dedicated log lines for `templates` and for `policies`. The line `Existing template will be overwritten, as overwrite is enabled.` is always printed when the template has `overwrite:true` set (which is the case when a new policy gets created), no matter if it exists or not. Right after this log line the next log line is more concrete about the template. I see how this can create confusion, changes in libbeat would be required for changing this though. I'll add it to my _improvement list_ for a follow up.

> With auto and index or indices overwrite, there is a log line saying "Automatically disabled ILM", however with auto and another output configured I couldn't find that line

Exactly, only when `output.elasticsearch.index` or `output.elasticsearch.indices` is customized in the config file, ILM will be disabled to respect those custom index settings. When ILM is set to `true` the index settings will be ignored. Would you expect different behavior?

> You said that ./apm-server -e -E apm-server.ilm.enabled=true -E setup.template.name=custom -E setup.template.pattern=custom -E output.elasticsearch.index=custom should refuse to start, however that doesn't seem to be the case for me: it says "Custom index configuration ignored when ILM is enabled" and moves on

Good catch, I updated the description. During implementation I realized this would be a breaking change and adapted handling to reflect `7.2` behavior: enable ILM and ignore index config.

> apm-server -e -E apm-server.ilm.enabled=true with ES not supporting ILM does also fallback to ilm disabled (like auto)

Yes. That is what we agreed on last week. It logs an error and continues with ILM disabled. Otherwise the server would be in a state where it is running but cannot connect to ES which would be a problematic scenario on cloud.

jalvz:
Thanks for clarifications. Happy to see that overwriting template/index does not stop the server. Regarding second point:

> With auto and index or indices overwrite, there is a log line saying "Automatically disabled ILM", however with auto and another output configured I couldn't find that line

I meant that with `auto`, there is always a log line saying how `auto` is resolved; eg. if index is overwritten, log says "ILM disabled".
However when setting another output (eg logstash), I know that `auto` also resolves to false but I didn't find any "ILM disabled" in the logs (if i did it right).

simitt:
> I meant that with auto, there is always a log line saying how auto is resolved; eg. if index is overwritten, log says "ILM disabled".
However when setting another output (eg logstash), I know that auto also resolves to false but I didn't find any "ILM disabled" in the logs (if i did it right).

Good point. The reason for that is though, that the whole `index-management` supporter and manager are not called by the framework in such cases.

OrangeDog:
The documentation was not updated to mention the additional templates this now creates
https://www.elastic.co/guide/en/apm/server/7.3/configuration-template.html

As far as I can tell, it's no longer possible to control which ILM policy is used - there's no way to customise these additional templates?

It's very unfortunate that you added a whole extra system (`apm-server.ilm`) conflicting with the beats standard (`setup.ilm`). That was probably a bad idea.

OrangeDog:
The documentation was not updated to mention the additional templates this now creates
https://www.elastic.co/guide/en/apm/server/7.3/configuration-template.html

As far as I can tell, it's no longer possible to control which ILM policy is used - there's no way to customise these additional templates?

It's unfortunate that you added a whole extra system (`apm-server.ilm`) conflicting with the beats standard (`setup.ilm`). That was probably a bad idea.

simitt:
There is an [ILM specific documentation](https://www.elastic.co/guide/en/apm/server/7.3/manual-ilm-setup.html) showing how ILM can be manually set up, in case the default ILM policies are not suitable.

OrangeDog:
> ILM can be manually set up

Indeed. But all other beats (and apm-server when `apm-server.ilm` is disabled) allow doing it via the `setup` config.

simitt:
There's an open Issue for making policies configurable https://github.com/elastic/apm-server/issues/2240, which you are welcome to follow and participate in finding a solution.

APM Server certainly overlaps with other beats for many features. It does also have a lot of dedicated config options nested under `apm-server.*` though. One of the main reasons for using a dedicated ILM setting in APM Server is that the behavior is a bit different than in other beats, mainly because APM Server creates multiple different indices by default, with different needs for policies, whereas other beats don't. Therefore requirements for config options can easily drift apart and sharing the same config name for different options and behavior seems confusing.

OrangeDog:
I would argue having apm-server suddenly start doing its own ILM, even though `setup.ilm.enabled: false` was equally (if not more) confusing.

simitt:
I will look into how we can improve the documentation around the chosen ILM handling to avoid confusion around this in the future (cc @bmorelli25). 

APM Server did never actively support `setup.ilm.enabled` nor was this option documented. When ILM was introduced in APM Server it was introduced with the current config key `apm-server.ilm.*`

PR #2359, opened 2019-07-01
consolidate logging selectors

graphaelli:
Extracted some of the logging changes I've been kicking around:
* List all logging selectors in `github.com/elastic/apm-server/log`
* Start reducing logger creations
* Log status code for errored requests
* Propogate error on pipeline registration failure

for elastic/apm-server#2201

### DIFF
--- a/tests/system/config/apm-server.yml.j2
+++ b/tests/system/config/apm-server.yml.j2
@@ -144,7 +144,11 @@ output.elasticsearch:
 
 ############################# Logging #########################################
 
+{% if logging_json or logging_level %}
 logging:
+{% else %}
+#logging:
+{% endif %}
   # Send all logging output to syslog. On Windows default is false, otherwise
   # default is true.
   #to_syslog: true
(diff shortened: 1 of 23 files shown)


simitt:
Not introduced by this PR but I just saw the `agentcfgHandler` wraps the [`authHandler` around the `logHandler`](https://github.com/elastic/apm-server/blob/master/beater/agent_config_handler.go#L61). Since this PR cleans up logging, would you mind changing it here?

tests/system/config/apm-server.yml.j2
```diff
@@ -151,8 +160,15 @@ output.elasticsearch:
   # Enable debug output for selected components.
   #selectors: []
 
+{% if logging_json %}
+  # Set to true to log messages in json format.
+  json: {{ logging_json|default("false") }}
```
simitt: Since this is enclosed by the `{% if logging_json %}` I don't think the `default("false")` is ever reached. Same for the `logging_level` below.

graphaelli: true, i'll update, thanks

graphaelli:
jenkins, run the tests please

PR #2360, opened 2019-07-02
[7.x] Sourcemap only log info for correct log level (#2358)

simitt:
 - Sourcemap only log info for correct log level  (#2358)


simitt:
Failing Kibana test has not been introduced by this PR.

PR #2366, opened 2019-07-02
[docs] Agent configuration documentation

bmorelli25:
WIP Agent configuration documentation for elastic/apm#86.

**Must be merged and backported simultaneously with https://github.com/elastic/kibana/pull/40084.**

Open questions:
* ~What's the script to sync the yml files?~
* ~Chat about these `apm-server.kibana.*` config options -- not sure they're all correct.~
* Impact of Cloud? Are `apm-server.kibana.*` settings blacklisted/whitelisted?
  * tbd in https://github.com/elastic/cloud/issues/35680
* ~Figure out if we want to use the beats `configure-kibana-endpoint.asciidoc` file or maintain our own.~ Going to use this for now. There are too many differences.
* How will different/older server versions impact this?

### DIFF
--- a/apm-server.yml
+++ b/apm-server.yml
@@ -198,7 +198,7 @@ apm-server:
   #agent.config.cache.expiration: 30s
 
   #kibana:
-    # For agent remote configuration, enabled must be true
+    # For APM Agent configuration in Kibana, enabled must be true
     #enabled: false
 
     # Scheme and port can be left out and will be set to the default (http and 5601)
--- a/docs/agent-configuration.asciidoc
+++ b/docs/agent-configuration.asciidoc
@@ -1,0 +1,54 @@
+[[agent-configuration-api]]
+== Agent configuration API
+
+APM Server exposes an API endpoint that allows agents to query the server for configuration changes.
+More information on this feature is available in the {kib-ref}/agent-configuration.html[APM Agent configuration in Kibana] documentation.
+
+[[agent-config-endpoint]]
+[float]
+=== Agent configuration endpoint
+Send an `HTTP GET` request to the agent configuration endpoint.
+`service.name` is a required query string parameter.
+
+[source,bash]
+------------------------------------------------------------
+http(s)://{hostname}:{port}/config/v1/agents?service.name=SERVICE_NAME
+------------------------------------------------------------
+
+[[agent-config-api-response]]
+[float]
+==== Responses
+
+200 - Successful
+400 - If `service.name` is not specified
+403 - Endpoint is disabled
+503 - "no configured Kibana Client: provide apm-server.kibana.* settings"
+
+If a <<secret-token,`secret_token`>> has been configured, it should also apply to this endpoint.
+
+[[agent-config-api-example]]
+[float]
+==== Example
+
+Example Agent configuration request including the service name "test-service":
+
+["source","sh",subs="attributes"]
+---------------------------------------------------------------------------
+curl http://127.0.0.1:8200/config/v1/agents?service.name=test-service
+---------------------------------------------------------------------------
+
+A sample response to the above curl request:
+
+["source","sh",subs="attributes"]
+---------------------------------------------------------------------------
+HTTP/1.1 200 OK
+Cache-Control: max-age=30, must-revalidate
+Content-Type: application/json
+Etag: 5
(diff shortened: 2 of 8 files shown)


jalvz:
Needs documenting extra setting introduced in https://github.com/elastic/apm-server/pull/2337

apm-server.yml
```diff
@@ -184,7 +184,7 @@ apm-server:
     #enabled: false
 
   #kibana:
-    # For agent remote configuration, enabled must be true
+    # For APM Agent configuration in Kibana, enabled must be true
```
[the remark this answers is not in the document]
bmorelli25: Thank you!

docs/configure-kibana-endpoint.asciidoc
```diff
@@ -102,3 +105,15 @@ apm-server.kibana.ssl.key: "/etc/pki/client/cert.key
...
+=== Agent Config configuration options
+
+You can specify the following options in the `apm-server.agent.config` section of the
++{beatname_lc}.yml+ config file:
+
+[float]
+==== `agent.config.cache.expiration`
+
+When using APM Agent configuration, information fetched from Kibana will be cached in memory.
+This setting specifies the time before cache key expiration. Defaults to 30 seconds.
```
jalvz: @graphaelli @simitt I don't know if we should document what this is used for:
- if we don't, users might think that higher value is better and cause a negative effect (imagine them setting it to 1 day)

- if we do, then we are documenting agent behaviour in the server and dictating how agents should behave:
The goal with Cache-Control was to *enable* agents to use it at their will or ignore it (the same as etag) - it was simply an attempt to return an appropriate response and provide mechanisms for optimization, should they want it.
Agents decided to use it as a poll interval for now. If we document that here, then we are forcing them to keep that behaviour, which was never my intention. They will not be able to replace their logic without introducing a breaking change.

docs/configure-kibana-endpoint.asciidoc
```diff
@@ -1,12 +1,15 @@
...
-Starting with {beatname_uc} 7.3.0, we've introduced support for
-{kib-ref}/agent-configuration.html[APM Agent configuration in Kibana] via the Kibana API.
-This requires a Kibana endpoint configuration.
+Starting with {beatname_uc} 7.3.0, we've introduced a new Kibana API for APM Server.
+Configuring this endpoint is required for 
+{kib-ref}/agent-configuration.html[APM Agent configuration in Kibana].
 Once the endpoint is configured,
 you'll be able to fine-tune certain agent configurations directly in Kibana without having to redeploy your agents.
 
+IMPORTANT: Kibana must be running and healthy before starting APM Server.
```
jalvz: After my suggestion, we discussed and prioritized https://github.com/elastic/apm-server/issues/2371. Keep an eye on it, hopefully this will have to be removed again (sorry for the back and forth)

docs/configure-kibana-endpoint.asciidoc
```diff
@@ -0,0 +1,119 @@
+[[setup-kibana-endpoint]]
+== Configure the Kibana endpoint
+
+Starting with {beatname_uc} 7.3.0, we've introduced a new Kibana API for APM Server.
+Configuring this endpoint is required for 
+{kib-ref}/agent-configuration.html[APM Agent configuration in Kibana].
+Once the endpoint is configured,
+you'll be able to fine-tune certain agent configurations directly in Kibana without having to redeploy your agents.
```
jalvz: Approved already, but forgot to mention!!: somewhere over here could note that the API between agent and server for cofig is secured by the same secret_token used for other endpoints, if it is configured.

Also forgotten that we have a whole API section (`intake-api`) for which we have a new `/config/v1/agents` entry that could be named Agent Config API, i guess?

docs/configure-kibana-endpoint.asciidoc
bmorelli25: Not a problem at all 😃

bmorelli25:
The docs-build ci will be red as this needs to be merged simultaneously with the Kibana changes.

docs/agent-configuration.asciidoc
```diff
@@ -0,0 +1,54 @@
+[[agent-configuration-api]]
```
bmorelli25: @jalvz - Can you review this API documentation when you return from vacation?

PR #2369, opened 2019-07-03
[7.x] Update beats framework to cecb997

simitt:


_beats/dev-tools/generate_notice.py
```diff
@@ -114,11 +114,15 @@ def has_license(folder):
 
     There are two cases accepted:
```
graphaelli: three ;)

_beats/libbeat/scripts/Makefile
```diff
@@ -51,7 +51,7 @@ GOIMPORTS=goimports
 GOIMPORTS_REPO?=github.com/elastic/beats/vendor/golang.org/x/tools/cmd/goimports
 GOIMPORTS_LOCAL_PREFIX?=github.com/elastic
 GOLINT=golint
-GOLINT_REPO?=github.com/golang/lint/golint
+GOLINT_REPO?=golang.org/x/lint/golint
```
graphaelli: I wonder why?  These should be the same

_beats/reviewdog.yml
```diff
@@ -1,3 +0,0 @@
-runner:
```
graphaelli: this doesn't seem to break our golint target like I expect it should.

_beats/dev-tools/generate_notice.py
simitt: _what_ comments ;)

_beats/libbeat/scripts/Makefile
simitt: APM Server overrides [GOLINT_REPO](https://github.com/elastic/apm-server/blob/4aa57602fb2fc7e0bd887705a17a915d0007cb73/Makefile#L10) to [vendor/github.com/golang/lint/golint](https://github.com/elastic/apm-server/tree/4aa57602fb2fc7e0bd887705a17a915d0007cb73/vendor/github.com/golang/lint/golint), which is still vendored appropriately.

_beats/reviewdog.yml
simitt: The reason why it didn't break is because the actual reviewdog cmd is run with the `-f` param: 

_beats/libbeat/scripts/Makefile:
```
REVIEWDOG?=reviewdog -conf ${ES_BEATS}/reviewdog.yml
```

Makefile:
```
GOLINT_COMMAND=$(shell $(GOLINT) ${GOLINT_TARGETS} | $(REVIEWDOG) -f=golint -diff="git diff 
```

However, I assume we want to use the config file, so I'll add it again. This file is still present in beats. It is not present in APM Server master. This is because it is missing in the `update_beats.sh` script. It looks like it had been added once and with the next beats update it was automatically removed again.

simitt: I pushed an update that brings back this file.

_beats/libbeat/scripts/Makefile
graphaelli: agreed, just curious

PR #2370, opened 2019-07-03
Fix failing system-test

simitt:
Remove testing for events with a hardcoded variable daily index to fix system tests.


simitt:
Merging this in as failed test is unrelated.

PR #2373, opened 2019-07-03
[7.3] Fix failing system-test (#2370)

simitt:
Backports the following commits to 7.3:
 - Fix failing system-test  (#2370)


simitt:
Failing test on Kibana objects - unrelated

PR #2374, opened 2019-07-03
fix: ignore fails on tar test results files

kuisathaverat:
enclose `tar` steps into `catchError` to ignore failures when it is not possible to grab test results files because of permissions errors.


kuisathaverat:
build error is not related

PR #2268, opened 2019-06-11
add package install test, add labels to script steps

kuisathaverat:
* add package install test

Those are moved to other PRs

* add labels to script steps - elastic/apm-server#2378
* add Tool Docker container with the Golang version used and tools needed - elastic/apm-server#2375
* use plain Go or mage commands inside the Tool Docker container - elastic/apm-server#2375
* fix Release build trigger, it was trigger always not only with beats updates. - elastic/apm-server#2379
* move benchmark stage to run in the tool Docker container - elastic/apm-server#2391

PR #2375, opened 2019-07-03
fix: build inside a Golang Docker container

kuisathaverat:
The `build Linux` stage is broken, so to avoid this kind of issues related to the way the Makefile manage the tools installation, we will use a Golang Docker container for run Go, it is a reproducible and controlled build environment for builds. Also, we will use plain mage commands when it is possible and we would go away from Makefile magic that the CI does not need (KISS).

relates to https://github.com/elastic/apm-server/pull/2268


Jenkinsfile
```diff
@@ -454,3 +456,12 @@ pipeline {
     }
   }
 }
+
+def golang(Closure body){
+  def golangDocker = docker.build("golang-mage", "--build-arg GO_VERSION=${GO_VERSION} ${BASE_DIR}/.ci/docker/golang-mage")
+   withEnv(["HOME=${WORKSPACE}", "GOPATH=${WORKSPACE}", "SHELL=/bin/bash"]) {
```
v1v:
```suggestion
  withEnv(["HOME=${WORKSPACE}", "GOPATH=${WORKSPACE}", "SHELL=/bin/bash"]) {
```

Jenkinsfile
```diff
@@ -454,3 +456,12 @@ pipeline {
...
+     golangDocker.inside(' -v /usr/bin/docker:/usr/bin/docker -v /var/run/docker.sock:/var/run/docker.sock'){
```
v1v:
```suggestion
     golangDocker.inside('-v /usr/bin/docker:/usr/bin/docker -v /var/run/docker.sock:/var/run/docker.sock'){
```

Jenkinsfile
kuisathaverat: the white space police 😄

PR #2386, opened 2019-07-04
Fix config handler

### LINKED ISSUE #2383: [ACM] fix response errors
- [ ] If kibana is disabled, the config endpoint should return a `4xx` (eg `403`) instead of `5xx`
- [ ] if kibana is not reachable or returns `5xx`, server should return `502`

jalvz:
fixes elastic/apm-server#2383

### DIFF
--- a/agentcfg/fetch.go
+++ b/agentcfg/fetch.go
@@ -33,7 +33,7 @@ import (
 )
 
 const (
-	endpoint = "/api/apm/settings/cm/search"
+	endpoint = "/api/apm/settings/agent-configuration/search"
 )
 
 // Fetcher holds static information and information shared between requests.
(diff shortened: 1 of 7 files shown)


jalvz:
sure, but https://github.com/elastic/apm-server/pull/2386/files#diff-ac43bed278b2a08583ad0db77c48b00bR155 is not enough?

simitt:
Sure this test is great for the actual behavior. What I meant was adding a logging test, checking for Kibana disabled, similar to https://github.com/elastic/apm-server/blob/ed9fd78d99e890ef3ecbb7e791670b5f51c0778c/tests/system/test_integration_logging.py#L86. With those tests we ensure that e.g. we don't lose information in the logs.

PR #2387, opened 2019-07-04
Refactor http header strings. Add "X-Content-Type-Options: nosniff" header

simitt:
Set `nosniff` for all http responses against intake API. Move http header keys as constants to a central file.


simitt:
jenkins, run this test again please.

Issue #2394, opened 2019-07-05
Use controlled versions for Go packages


kuisathaverat:
We use a few Go packages on the latest versions from GitHub directly, it is a bad habit, we will move them to our vendors forks. 

```
 git clone https://github.com/magefile/mage \
  && cd mage \
  && go run bootstrap.go \
  && mage -version \
  && cd /tools \
  && go get golang.org/x/tools/cmd/benchcmp \
  && go get github.com/t-yuki/gocover-cobertura \
  && go get github.com/jstemmer/go-junit-report \
  && go get github.com/kardianos/govendor \
```

PR #2395, opened 2019-07-05
Vendor check/test tools

graphaelli:
vendor tools used for checks & tests.  ~Doesn't do `github.com/kardianos/govendor` for now.~ does now

for elastic/apm-server#2394


kuisathaverat:
@Gil What about [govendor tool](https://github.com/kardianos/govendor)? Should we add it to vendors?

graphaelli:
Yes certainly, as the beats update check uses it.  I'll add it shortly

PR #2396, opened 2019-07-08
Fix changelog from 7.2 on.

simitt:
Needs backport to `7.3` branch.


simitt:
Yes, just wanted to ensure we also have them accordingly in `master`.

PR #2397, opened 2019-07-08
Remove unused properties from metricset Intake API

simitt:
@axw you discussed some validation in https://github.com/elastic/apm/issues/78#issuecomment-507148033. Please let me know if you are good with not having any validation on the properties and turning back to the more generic json spec for the metricset's `transaction` and `span` sample properties. 

- [ ] Needs backport to 7.x and 7.3
- [x] Add PR number to changelog entry


graphaelli:
lgtm though needs some `go generate`

axw:
Fine by me, thanks for checking!

PR #2403, opened 2019-07-08
Fix timing issue for pipeline system test

simitt:


graphaelli:
Thanks for bringing this one up, it has been very flaky lately.  What do you think of using retries instead?  Something like `self.wait_until(lambda: self.es.ingest.get_pipeline(id=pipeline_id))` for each pipeline.  We could even make `wait_until` return the result of `cond()` and avoid an extra call, no obligation to.

simitt:
@graphaelli thanks for your suggestion, ofc a way better way of fixing the test. Since it is only tests and I followed your idea, I am merging this without awaiting approval to be able to rebase other PRs onto it.

PR #2407, opened 2019-07-09
Enclose etag and IfNoneMatch header in double quotes.

### LINKED ISSUE #2404: Etag header should be double-quoted
As described in https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag (and the relevant RFC), Etag values should be double-quoted. The server currently does not do this:

```sh
$ curl -v http://localhost:8200/config/v1/agents?service.name=foo
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 8200 (#0)
> GET /config/v1/agents?service.name=foo HTTP/1.1
> Host: localhost:8200
> User-Agent: curl/7.58.0
> Accept: */*
> 
< HTTP/1.1 200 OK
< Cache-Control: max-age=30, must-revalidate
< Etag: bxyV1WsBXaU8VEI7_C0d
< Date: Tue, 09 Jul 2019 07:15:10 GMT
< Content-Length: 25
< 
{
  "sample_rate": "1"
}
```

This could be problematic if a caching proxy were placed in front.

`If-None-Match` values should be double-quoted too.

simitt:
fixes elastic/apm-server#2404

needs backport to `7.x` and `7.3`.

### DIFF
--- a/beater/agent_config_handler.go
+++ b/beater/agent_config_handler.go
@@ -46,6 +46,7 @@ func agentConfigHandler(kbClient *kibana.Client, enabled bool, config *agentConf
 
 		query, requestErr := buildQuery(r)
 		cfg, upstreamEtag, internalErr := fetcher.Fetch(query, requestErr)
+		etag := fmt.Sprintf("\"%s\"", upstreamEtag)
 
 		var resp interface{}
 		var state int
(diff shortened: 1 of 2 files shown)


simitt:
Tests pass except for the Integration Tests, where failure is unrelated.

simitt:
Jenkins, run the tests again please

simitt:
@elastic/apm-agent-devs please be aware of this change.

watson:
Note for other agent devs who might be about to do the same as I was about to do: If you or your http framework doesn't strip the double quotes from `Etag` header value, they should remain in place, and so when you send the value back to the APM Server later using the `If-None-Match` header, it should be correctly encoded.

Issue #2260, opened 2019-06-05
[APM-CI] Add changelog sync to CI


bmorelli25:
@graphaelli has created a ~~[script](https://gist.github.com/graphaelli/92fd6b5ff08d69600e880df2a234ac35)~~ that checks if the APM Server changelogs are synced across branches. We'd like to get this check added to CI.

Script now lives in apm-server repo @ [`script/check_changelogs.py`](https://github.com/elastic/apm-server/blob/8daeaefb7816e7691ae669a0c812163ccb0edca4/script/check_changelogs.py)

PR #2418, opened 2019-07-10
feat: backport pipeline from #2260 [6.8]

v1v:
backport pipeline from https://github.com/elastic/apm-server/issues/2260

### DIFF
--- a/script/jenkins/check-changelogs.sh
+++ b/script/jenkins/check-changelogs.sh
@@ -1,0 +1,7 @@
+#!/usr/bin/env bash
+set -exuo pipefail
+
+docker run --rm \
+          -v "$PWD":/usr/src/myapp \
+          -w /usr/src/myapp python:3.6.8-alpine \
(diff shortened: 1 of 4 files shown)


script/jenkins/check-changelogs.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+set -exuo pipefail
+
+docker run --rm \
```
kuisathaverat: I'd provision the docker container in the Jenkinsfile to not force to have docker to run the thing

v1v: Not sure though, I'd rather prefer the script to be simple and then able to run it locally

kuisathaverat: I've created https://github.com/elastic/apm-server/issues/2442

v1v:
After revisiting this PR I realize this is not required though as we the script itself should be branch independent as the list of branches are defined within the script. Therefore I'll close this PR now

PR #2419, opened 2019-07-10
feat: backport pipeline from #2260 [7.1]

v1v:
backport pipeline from https://github.com/elastic/apm-server/issues/2260


v1v:
After revisiting this PR I realize this is not required though as we the script itself should be branch independent as the list of branches are defined within the script. Therefore I'll close this PR now

PR #2421, opened 2019-07-10
Ensure Kibana client reconnects.

### LINKED ISSUE #2371: Agent Configuration unavailable without Kibana available at startup
If Kibana is configured but not reachable, apm-server will start normally without retrying again.
This requires users to start and health-check Kibana before starting apm-server, which might be inconvenient.

simitt:
Implement Kibana client wrapper that takes care of obtaining a valid
Kibana client, in case client cannot be obtained on startup.
Align returned response body to whether or not the user sends a secret token.
Log full error messages. 

fixes elastic/apm-server#2371

### DIFF
--- a/beater/agent_config_handler.go
+++ b/beater/agent_config_handler.go
@@ -40,13 +40,14 @@ const (
 	errMsgKibanaVersionNotCompatible = "not a compatible Kibana version"
 	errMsgNoKibanaConnection         = "unable to retrieve connection to Kibana"
 	errMsgInvalidQuery               = "invalid query"
+	errMsgKibanaDisabled             = "kibana config disabled"
 )
 
 var (
 	minKibanaVersion = common.MustNewVersion("7.3.0")
 )
 
-func agentConfigHandler(kbClient kibana.Client, enabled bool, config *agentConfig, secretToken string) http.Handler {
+func agentConfigHandler(kbClient kibana.Client, config *agentConfig, secretToken string) http.Handler {
 
 	authErrMsg := func(errMsg, logMsg string) string {
 		if secretToken == "" {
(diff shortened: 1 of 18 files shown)


kibana/connecting_client_test.go
```diff
@@ -0,0 +1,137 @@
...
+	})
+}
+
+type rt struct {
+	resp *http.Response
+}
+
+var (
+	mockCfg     = common.MustNewConfigFrom(`{"enabled": "false", "host": "non-existing"}`)
+	mockBody    = ioutil.NopCloser(convert.ToReader(`{"response": "ok"}`))
```
graphaelli: this is the first I've seen `ToReader` - it should probably just return a `ioutil.NopCloser` since we're wrapping that everywhere it's used.

beater/agent_config_handler.go
```diff
@@ -117,12 +139,31 @@ func buildQuery(r *http.Request) (query agentcfg.Query, err error) {
...
 
-func wrap(w http.ResponseWriter, r *http.Request) func(interface{}, int) {
-	return func(body interface{}, code int) {
+func wrap(w http.ResponseWriter, r *http.Request) func(interface{}, int, string, string) {
+	return func(body interface{}, code int, cacheControl string, errMsg string) {
+		w.Header().Set(headers.CacheControl, cacheControl)
+
+		if code >= http.StatusBadRequest {
+			requestLogger(r).Errorw("error handling request", "response_code", code,
+				"body", body, "error", errMsg)
```
graphaelli: Can we drop body, assuming it's always a part of the `errMsg`?

beater/common_handler.go
```diff
@@ -337,11 +337,17 @@ func sendPlain(w http.ResponseWriter, body interface{}, statusCode int) int {
...
+	var b []byte
+	var err error
+	if bStr, ok := body.(string); ok {
+		b = []byte(bStr + "\n")
+	} else {
+		b, err = json.Marshal(body)
+		if err != nil {
+			b = []byte(fmt.Sprintf("%+v", body))
+		}
+		b = append(b, "\n"...)
```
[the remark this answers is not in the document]
simitt: totally agree

simitt:
> Those double quoted, non json bodies don't seem right, or is that what we agreed upon for now?

That's actually the same behavior in `master`, I haven't touched anything related to how the messages are rendered. I can also fix it within the PR. What do you think about e.g `{ "error": "no config found" }`?

kibana/connecting_client_test.go
simitt: I only moved this, so didn't check in detail. But there are other places where we don't wrap it e.g. [agentcfg/fetch.go#L63](https://github.com/elastic/apm-server/blob/master/agentcfg/fetch.go#L63).

PR #2441, opened 2019-07-12
[ACM] Create Etag from settings struct.

### LINKED ISSUE #2434: Central config: Etag not updated when updating configuration
The Etag does not change when changing the configuration. For example, set the sample rate to 0.1, and query the endpoint. Now set the sample rate to 0.2 and query again. The Etag is the same.

Similarly, when providing the Etag in a `If-None-Match` header, the server responds with `304 Not Modified` when updating the configuration as opposed to responding with a 200.

simitt:
To ensure Etag changes when config settings are updated, include config settings when calculating ID for config object.

fixes elastic/apm-server#2434

- [x] needs changelog entry

backport to `7.x` and `7.3`. 

fyi @felixbarny , @axw

### DIFF
--- a/agentcfg/model.go
+++ b/agentcfg/model.go
@@ -21,7 +21,9 @@ import (
 	"crypto/md5"
 	"encoding/json"
 	"fmt"
+	"hash"
 	"sort"
+	"strings"
 )
 
 // Doc represents an elasticsearch document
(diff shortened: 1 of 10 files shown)


agentcfg/model.go
```diff
@@ -18,32 +18,51 @@
...
+type tmpDoc struct {
+	ID     string `json:"_id"`
+	Source struct {
+		Settings json.RawMessage `json:"settings"`
+	} `json:"_source"`
+}
+
+func (t *tmpDoc) hash() string {
+	h := md5.New()
+	h.Write([]byte(t.ID))
```
graphaelli: why include the ID in the hash?

agentcfg/model.go
```diff
@@ -18,32 +18,51 @@
...
+	h.Write([]byte(t.Source.Settings))
```
graphaelli: are the settings guaranteed to be in the same order every time they are queried?

agentcfg/query.go
```diff
@@ -0,0 +1,56 @@
...
+}
+
+// NewQuery creates a Query struct
+func NewQuery(name, env string) Query {
+	return Query{Service{name, env}}
+}
+
+// ID returns the unique id for the query
+func (q Query) ID() string {
+	var str strings.Builder
```
graphaelli: I realize this was just moved code but we could avoid the `strings.Builder` with

```go
if q.Service.Environment == "" {
  return q.Service.Name
}
return q.Service.Name + "_" + q.Service.Environment
```

agentcfg/model.go
simitt: So it is unique across different services. This is not necessary for the workflow we have, but it feels weird that services with the same config result in the same Etag. Adding the ID is not a concerning overhead. At the moment there is no `version` field or `timestamp` field that changes when the config is updated, so this seemed the best solution.

agentcfg/query.go
simitt: Why would you rather use string concatenation than a string builder here?

agentcfg/model.go
simitt: Good question, I expected everything within `_source` field to be the unchanged stored json document and therefore to be returned in the same order for the same stored version. The server is not messing with it before the hash sum is calculated. @ogupte can you confirm that the `_source.settings` is always returned in the same order?
Since there is no version field and the timestamp doesn't get updated on editing, I don't see any other solution atm.

graphaelli: Another solution would be to unmarshal the setttings, and loop over the keys alphabetically, adding the key and value to the hash along the way.  That would be and issue for nested keys but I"m not sure if we're planning on those at any point?

agentcfg/model.go
graphaelli: I'm more worried about the reverse, deleting and re-adding the same settings for a service would result in two different etags for the same configuration.

simitt: I don't think that would be a big issue, but I can change it to not consider the ID.

agentcfg/query.go
graphaelli: I'm expecting less allocations and less operations, will put together a benchmark.  This is hardly a significant contribution to either processing time or memory usage/gc pressure.

agentcfg/model.go
```diff
@@ -34,43 +35,34 @@ type Settings map[string]string
...
+	var keys []string
+	for k := range tmp.Source.Settings {
+		keys = append(keys, k)
 	}
-	doc.ID = tmp.hash()
-	return &doc, nil
+	sort.Strings(keys)
+	out := make(map[string]string)
+	for _, k := range keys {
+		out[k] = fmt.Sprintf("%v", tmp.Source.Settings[k])
```
graphaelli: any concerns about nested keys here?

simitt: Thanks for raising the question again. I checked back with the agent devs, and they expect nested keys to be returned in the format `"a.b.c": "val"` rather then `"a": "{b: {c: val}}"`.
I now updated the handling accordingly, for a config of form:
```
"settings" : {
  "sample_rate" : 0.7,
  "nested_1" : {
    "k1" : [1,2,3],
    "k2" : "test",
    "nested_2" : {
      "k3" : 56
    } 
  }
}
```
we  now return:
```
{
  "nested_1.k1": "1,2,3",
  "nested_1.k2": "test",
  "nested_1.nested_2.k3": "56",
  "sample_rate": "0.7",
}
```

The adapted handling also ensures order on nested keys for calculating the hash sum

agentcfg/model.go
```diff
@@ -33,36 +35,69 @@ type Doc struct {
...
+	var localkey string
 	for _, k := range keys {
-		out[k] = fmt.Sprintf("%v", tmp.Source.Settings[k])
-		h.Write([]byte(fmt.Sprintf("%s_%v", k, tmp.Source.Settings[k])))
+		if rootKey == "" {
+			localkey = k
+		} else {
+			localkey = fmt.Sprintf("%s.%s", rootKey, k)
+		}
+		if inpMap, ok := inp[k].(map[string]interface{}); ok {
```
graphaelli: better as a type switch

agentcfg/model.go
```diff
@@ -39,7 +39,7 @@ type Settings map[string]string
 func NewDoc(inp []byte) (*Doc, error) {
 	var doc Doc
 	settings, err := unmarshal(inp)
-	if settings == nil || err != nil {
+	if err != nil {
 		return &doc, err
```
axw: should that just be `return nil, err` now?

agentcfg/model_test.go
```diff
@@ -0,0 +1,76 @@
...
+		assert.Empty(t, d)
+	})
+
+	t.Run("EmptyInput", func(t *testing.T) {
+		d, err := NewDoc([]byte{})
+		require.NoError(t, err)
+		expectedDoc := Doc{
+			Settings: map[string]string{},
+			ID:       fmt.Sprintf("%x", md5.Sum([]byte{}))}
+		assert.Equal(t, &expectedDoc, d)
```
simitt: @axw as shown by this test I changed the logic to return an empty `{}` and a md5 sum over an empty `[]byte` for the ID.

axw: Thank you!

agentcfg/model.go
```diff
@@ -18,67 +18,89 @@
...
+	for k := range inp {
+		keys = append(keys, k)
 	}
-	return str.String()
+	sort.Strings(keys)
+	var localkey string
+	for _, k := range keys {
+		localkey = dotKey(rootKey, k)
+
+		switch inp[k].(type) {
```
graphaelli: The result of the type assertion can be saved at `switch` and then used in each case, like:

```suggestion
		switch in := inp[k].(type) {
```

PR #2439, opened 2019-07-11
Agent configuration integration tests

graphaelli:
Starts kibana for integration tests now, alongside previously started elasticsearch.

Includes (commented out) failing tests/todos for elastic/apm-server#2434 and #2438.

closes elastic/apm-server#2219

PR #2443, opened 2019-07-12
[ACM] Only allow cache expiration values in full seconds.

### LINKED ISSUE #2438: Central config: non-integer cache-control headers
With `apm-server.agent.config.cache.expiration=1010ms`, the config endpoint returns:

```
Cache-Control: max-age=1.01, must-revalidate
```

I propose rejecting the configuration as invalid rather than rounding to 1. be predictable 2. avoid the < 1s edge case.

simitt:
fixes elastic/apm-server#2438

- [x] changelog entry

backport to `7.x` and `7.3`.


graphaelli:
Per discussion, we can update the integration test introduced in #2439 to check this.

simitt:
I generally agree regarding the `Unpack` but we also use the same `Cache` struct for the sourcemapping cache and wouldn't want to introduce limitation there just now (might be a breaking change).

PR #2474, opened 2019-07-18
 Update beats framework to 7bba6ec 

graphaelli:
primarily to fix packaging via https://github.com/elastic/beats/pull/12957

will follow with go 1.12.7 update after this is merged


graphaelli:
#2475 fill fix the acm test failures encountered here

vendor/vendor.json
```diff
@@ -2476,14 +2477,6 @@
...
 			"versionExact": "v1.3.0"
 		},
-		{
-			"checksumSHA1": "F3J0rky8CLcyEl18/lVc9wNvLRw=",
-			"path": "go.elastic.co/apm/internal/radix",
-			"revision": "9f89e36c6e534fd548810428cad1bd55cd57a5d2",
-			"revisionTime": "2019-01-04T03:47:35Z",
-			"version": "v1.3.0",
-			"versionExact": "v1.3.0"
-		},
```
simitt: Not sure why this got removed, as it generally doesn't look like we changed anything for this dependency, but doesn't seem required anywhere so looks fine.

PR #2475, opened 2019-07-18
ignore unavailable agent configuration index during setup.

graphaelli:
kibana creates that index - when testing without kibana it will not exist.  When testing with kibana, tests that depend on that index will fail if it is missing, no need to fail during setup


graphaelli:
It appears this is newly needed because kibana used to create the `.apm-agent-configuration` index on startup, now it is created on demand when a configuration is added.

PR #2484, opened 2019-07-19
[7.3] Update beats framework to a081d96

simitt:


_beats/vendor/vendor.json
```diff
@@ -1359,44 +1373,44 @@
...
-			"checksumSHA1": "tuhGcluN3UtoiFBovqsep6aPx3s=",
+			"checksumSHA1": "REELpodvqmrKDkIMKRSrZd9g7sE=",
 			"path": "github.com/elastic/gosigar",
-			"revision": "99ed9cf55303a9d3936cb656b9a86a4a6e67b30a",
-			"revisionTime": "2019-05-27T11:32:19Z",
-			"version": "v0.10.3",
-			"versionExact": "v0.10.3"
+			"revision": "f48d9dc84bc636d361c33fab2d7d753b705fd373",
+			"revisionTime": "2019-07-15T18:29:51Z",
+			"version": "HEAD",
```
graphaelli: `HEAD` seems strange, shouldn't the version just be omitted in that case?

simitt: Fetching with govendor with `govendor fetch xyz@head` sets the version in the `vendor.json` file to `HEAD`. That's also how it is in beats `master` + `7.3`.

graphaelli: Ok, thanks for clarifying

PR #2488, opened 2019-07-22
Disable sourcemapping config

### LINKED ISSUE #2357: Provide config option to disable applying sourcemaps to stacktraces
In some use-cases there are no sourcemaps used for the application, e.g. [this one](https://github.com/elastic/apm-agent-rum-js/issues/293). In this case the APM server would constantly log errors into the stacktraces stored in the ES. Adding a config option to disable applying sourcemaps would resolve this issue. If applying sourcemaps is disabled the stack trace should be stored as it was received from the agent.

graphaelli:
via `rum.sourcemapping.enabled=false`

closes #2357

### DIFF
--- a/changelogs/head.asciidoc
+++ b/changelogs/head.asciidoc
@@ -6,7 +6,8 @@
 
 [float]
 ==== Added
-- Upgrade Go to 1.12.7 {pull}XXX[XXX].
+- Upgrade Go to 1.12.7 {pull}2483[2483].
+- Add config option to disable sourcemapping {pull}2488[2488].
 
 [float]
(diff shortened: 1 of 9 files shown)


changelogs/head.asciidoc
```diff
@@ -6,7 +6,8 @@
 
 [float]
 ==== Added
-- Upgrade Go to 1.12.7 {pull}XXX[XXX].
+- Upgrade Go to 1.12.7 {pull}2483[2483].
```
graphaelli: ok to change this here as backporting changelogs always involves some manual updates anyway

PR #2490, opened 2019-07-24
ci: remove the documentation stage

kuisathaverat:
Since the documentation team has elastic-ci/docs build in place for all our PRs, our Documentation stage lose sense, we are doing the same and we have to maintain it. Because of that, we will start to remove this stage from all projects.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -311,36 +311,6 @@ pipeline {
       }
     }
     /**
-    Build the documentation and archive it.
-    Finally archive the results.
-    */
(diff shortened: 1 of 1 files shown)


simitt:
How do we see if a PR breaks the docs then? Can we auto-link to the docs build that gets triggered for the PR?

kuisathaverat:
Our stage was replaced by the build on `elasticsearch-ci/docs` you can see the following check

Screenshot 2019-07-24 at 19 52 13

Issue #2489, opened 2019-07-23
Rework beater package


simitt:
The `beater` package deals with a couple of technical debts at the moment. This is a meta issue to discuss issues and how to address them. 

The `logging` and `monitoring` handling is spread all over the package, leading to errors and not fully logged and monitored requests (e.g. ACM requests increase counter for requests, but not for responses). 

The response writer logic needs some refactoring to have a unique and more idiomatic way of writing responses and errors. Response handling should be separated from increasing monitoring counters. 

After some research the suggested solution is to introduce a `Context` struct pool. A `Context` is passed around between `Handlers` within one request, abstracting the `http.ResponseWriter` and the `*http.Request`. For every new request the context pool takes care of reusing an idle context, and resetting it to the current request.

Introducing a `Context` allows for better separation of concerns for the single middleware handlers, while reusing structs to increase performance.  

Since this is a larger effort, the changes will be introduced in separate PRs, directed against a feature branch. Once all changes are in and all tests are accordingly adapted, the feature branch will be merged against master. 

Separating the effort into following tasks:
- [x] Introducing `Context` and centralizing `logging` (elastic/apm-server#2492 )
- [x] Centralized `monitoring` handling (elastic/apm-server#2493 )
- [x] Rework response writer logic to be more idiomatic and less error prone (elastic/apm-server#2494 )
- [x] Restructure beater package. Move handlers to middleware package and add unit tests per handler, while keeping integration tests for routes. (https://github.com/elastic/apm-server/pull/2518)
- [x] Change unintuitive abbreviations. (https://github.com/elastic/apm-server/pull/2518)
- [x] Change current log handling to get rid of additional `error` and unify used terminology (elastic/apm-server#2201 )
- [x] Introduce proper monitoring handling for ACM. Counters for responses are missing. (#2142   )
- [x] Discuss and change monitoring handling for `root` and `asset` routes. (elastic/apm-server#2142 )

PR #2492, opened 2019-07-24
[beater] Add context handlers and adapt log handling

simitt:
This is the first of a series of PRs for refactoring the `beater` handling wrt logging, monitoring and writing response, more infos in the meta issue #2489 . 

This PR introduces a `Context` struct abstracting the `http.ResponseWriter` and the `*http.Request` and a `ContextPool` that takes care of reusing idle `Context` instances. 

This PR aims for minimal changes to keep reviews easier. It contains _TODOs_ that will be addressed in follow up PRs before merging back to `master`. 

Partially implements #2201

### DIFF
--- a/beater/log_handler.go
+++ b/beater/log_handler.go
@@ -1,0 +1,65 @@
+// Licensed to Elasticsearch B.V. under one or more contributor
+// license agreements. See the NOTICE file distributed with
+// this work for additional information regarding copyright
+// ownership. Elasticsearch B.V. licenses this file to you under
+// the Apache License, Version 2.0 (the "License"); you may
+// not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package beater
+
+import (
+	"net/http"
+
+	"github.com/gofrs/uuid"
+
+	"github.com/elastic/beats/libbeat/logp"
+
+	"github.com/elastic/apm-server/beater/headers"
+	"github.com/elastic/apm-server/beater/request"
+	logs "github.com/elastic/apm-server/log"
+	"github.com/elastic/apm-server/utility"
+)
+
+func logHandler(h Handler) Handler {
+	logger := logp.NewLogger(logs.Request)
+
+	return func(c *request.Context) {
+		requestCounter.Inc()
+		reqID, err := uuid.NewV4()
+		if err != nil {
+			sendStatus(c, internalErrorResponse(err))
+		}
+
+		reqLogger := logger.With(
+			"request_id", reqID,
+			"method", c.Req.Method,
+			"URL", c.Req.URL,
+			"content_length", c.Req.ContentLength,
+			"remote_address", utility.RemoteAddr(c.Req),
+			"user-agent", c.Req.Header.Get(headers.UserAgent))
+
+		h(c)
+
+		keysAndValues := []interface{}{"response_code", c.StatusCode()}
+		if c.StatusCode() >= http.StatusBadRequest {
+			keysAndValues = append(keysAndValues, "error", c.Error())
+			if c.Stacktrace() != "" {
+				keysAndValues = append(keysAndValues, "stacktrace", c.Stacktrace())
+			}
+			reqLogger.Errorw("error handling request", keysAndValues...)
+			return
+		}
+
+		reqLogger.Infow("handled request", keysAndValues...)
+	}
+}
--- a/beater/log_handler.go
+++ b/beater/log_handler.go
@@ -42,19 +42,19 @@ func logHandler(h Handler) Handler {
 
 		reqLogger := logger.With(
 			"request_id", reqID,
-			"method", c.Req.Method,
-			"URL", c.Req.URL,
-			"content_length", c.Req.ContentLength,
-			"remote_address", utility.RemoteAddr(c.Req),
-			"user-agent", c.Req.Header.Get(headers.UserAgent))
+			"method", c.Request.Method,
+			"URL", c.Request.URL,
+			"content_length", c.Request.ContentLength,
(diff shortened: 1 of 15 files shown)


beater/request/context.go
```diff
@@ -0,0 +1,165 @@
...
+	mimeTypeApplicationJSON = "application/json"
+)
+
+var (
+	mimeTypesJSON = []string{mimeTypeAny, mimeTypeApplicationJSON}
+)
+
+// Context abstracts request and response information for http requests
+type Context struct {
+	Req *http.Request
```
graphaelli: I agree with not embedding but think `Request` should be spelled out
```suggestion
	Request *http.Request
```

beater/request/context.go
```diff
@@ -0,0 +1,165 @@
...
+	stacktrace string
+
+	w http.ResponseWriter
+}
+
+// Reset allows to reuse a context by removing all request specific information
+func (c *Context) Reset(w http.ResponseWriter, r *http.Request) {
+	c.Req = r
+
+	c.w = w
```
graphaelli: suggest resetting these in the same order they are defined so it's easier to ensure everything is reset if members are added

beater/request/context.go
```diff
@@ -0,0 +1,165 @@
...
+	c.stacktrace = ""
+}
+
+// Header returns the http.Header of the context's writer
+func (c *Context) Header() http.Header {
+	return c.w.Header()
+}
+
+// StatusCode returns the context's status code
+func (c *Context) StatusCode() int {
```
graphaelli: These getters feel non-idiomatic to me, are we worried about something specific here?

beater/request/context.go
```diff
@@ -0,0 +1,165 @@
...
+func (c *Context) Stacktrace() string {
+	return c.stacktrace
+}
+
+// AddStacktrace sets a stacktrace for the context
+func (c *Context) AddStacktrace(stacktrace string) {
+	c.stacktrace = stacktrace
+}
+
+//TODO: All of the below methods will be changed during the response handling refactoring
```
graphaelli: I didn't see where that's happening, will find you to clear it up

beater/request/context.go
```diff
@@ -0,0 +1,165 @@
...
+
+// WriteHeader sets status code in context and call the http.ResponseWriter method
+func (c *Context) WriteHeader(statusCode int) {
```
graphaelli: I relized `WriteHeader` is in the stdlib like this but `SetStatusCode` makes more sense IMO, feel free to disagree and ignore.

beater/request/context.go
```diff
@@ -0,0 +1,165 @@
...
+	return c.err
+}
+
+// Stacktrace returns the context's stacktrace
+func (c *Context) Stacktrace() string {
+	return c.stacktrace
+}
+
+// AddStacktrace sets a stacktrace for the context
+func (c *Context) AddStacktrace(stacktrace string) {
```
graphaelli: `SetStacktrace`?  again, getter/setter feels a bit foreign

beater/request/context.go
```diff
@@ -0,0 +1,165 @@
...
+	buf, err := json.MarshalIndent(body, "", "  ")
+	if err != nil {
+		logp.NewLogger(logs.Response).Errorf("Error while generating a JSON error response: %v", err)
+		c.sendPlain(body, statusCode)
+		return
+	}
+
+	buf = append(buf, "\n"...)
+	n, _ := c.w.Write(buf)
+	c.w.Header().Set(headers.ContentLength, strconv.Itoa(n))
```
graphaelli: isn't the body already written at this point?

beater/request/context.go
```diff
@@ -0,0 +1,165 @@
...
+		c.sendJSON(body, statusCode)
+	} else {
+		c.sendPlain(body, statusCode)
+	}
+}
+
+func (c *Context) sendJSON(body interface{}, statusCode int) {
+	c.Header().Set(headers.ContentType, "application/json")
+	c.WriteHeader(statusCode)
+	buf, err := json.MarshalIndent(body, "", "  ")
```
graphaelli: any reason not to encode directly onto the response writer instead of allocating this buffer?

beater/request/context.go
simitt: Yes, that is just moved code, as this PR doesn't intend to change or improve anything related to writing response. It only introduced the `context` and centralizes the `logging`. Everything else will follow.

beater/request/context.go
simitt: Out of scope for this PR, in the comings already https://github.com/elastic/apm-server/pull/2494/files#diff-f6d7c55b6aa52771874eb56711422526R157

beater/log_handler.go
```diff
@@ -0,0 +1,65 @@
...
+			"method", c.Req.Method,
+			"URL", c.Req.URL,
+			"content_length", c.Req.ContentLength,
+			"remote_address", utility.RemoteAddr(c.Req),
+			"user-agent", c.Req.Header.Get(headers.UserAgent))
+
+		h(c)
+
+		keysAndValues := []interface{}{"response_code", c.StatusCode()}
+		if c.StatusCode() >= http.StatusBadRequest {
```
graphaelli: maybe the context itself should judge the response, so change `c.StatusCode() >= http.StatusBadRequest` to `c.BadStatus()` (or something better)

beater/request/context.go
simitt: That's not happening in this PR, as this PR only focuses on two things: introducing a `Context` and centralizing logging. 
As discussed I tried to keep PRs small and focused and have one PR per topic(logging, [monitoring](https://github.com/elastic/apm-server/pull/2493), [response writing](https://github.com/elastic/apm-server/pull/2494) and [package restructure](https://github.com/simitt/apm-server/pull/5)). The PRs all build on top of each other.

If you prefer I can also combine all of them into one PR, only focusing on restructuring and refactoring and not introducing logical changes (e.g. more monitoring on ACM or changing wording for logging). This would be easier to maintain for me, but result in a much bigger PR).

beater/request/context.go
simitt: Will be changed in a follow up, would like to leave as is for now.

beater/request/context.go
simitt: You are right, it doesn't feel idiomatic. The intention was to avoid messing with the statusCode or Err outside the context. But I am also not satisfied with them, so changing for now and then we see how it comes together with the following changes.

Issue #2142, opened 2019-04-23
Rework request and response counters


simitt:
At the moment the request and response counters, as well as the log entries for requests and responses are spread between the various `handler` files in the `beater` directory.

I suggest to move the logic to a central place and wrap all requests with it. We already have a `logHandler` wrapped around requests. ~This log handler should be adapted so that it also has access to errors. See if [recordingResponseWriter](https://github.com/elastic/apm-server/blob/master/utility/handlers.go) can be adapted and used for this.~

For the counters we should add something similar, by adding a dedicated `counterHandler` that ensures all the counters are set properly and removes the clutter from the beater package.

~While on it, we could add system tests to check that logs actually contain expected log messages, as the testing for logs is rather sparse.~

Update: 
This issue has been split into 2 issues. Everything concerning improvements of the log handling has been moved to https://github.com/elastic/apm-server/issues/2201.

jalvz:
oh sorry, wrong link :(

PR #2493, opened 2019-07-24
[beater] Introduce middleware logic and dedicated monitoring middleware

simitt:
Follow up PR on top of #2492 focusing on following changes:
* introducing dedicated middleware handling
* refactoring monitoring handling, without changing any logic. This PR aims to make it easier to fix bugs in the current monitoring handling by centralizing the handling, and introducing dedicated tests.

Bugfixes for current monitoring handling should occur in follow up PRs once the refactoring is done to keep the restructuring separate from it. 

related to #2489 

partially implements #2142


beater/middleware.go
```diff
@@ -0,0 +1,31 @@
...
+package beater
+
+import (
+	"github.com/elastic/apm-server/beater/request"
+)
+
+type middleware func(request.Handler) request.Handler
+
+func withMiddleware(h request.Handler, m ...middleware) request.Handler {
+	for i := len(m) - 1; i >= 0; i-- {
```
graphaelli: I was surprised that no tests failed when flipping the order the other direction

simitt: Yes we are lacking some integration tests that actually test the middleware. I am working on adding them in the next PR.

PR #2494, opened 2019-07-24
[beater] Improve Response Writing

simitt:
As part of the `beater` refactoring in #2489 this PR improves how response bodies are written, without changing the actual logic to not break between minor versions. 
This will help improving response writing in the future. 

There is one change in how a 404 body is returned if the path does not exist: Currently only a plain/text response was given, ignoring the `accept` header. This has been changed to return a json body if accepted.

It also introduces a `beatertest` package offering some shared test setup handling.

As a preparation of splitting up the logic into sub packages of `beater` some methods and structs have been exported. 

~Running some local benchmarks with `hey-apm` shows slightly improved (not significant) `index-rate` for this branch compared to `master`.~
update: New benchmarks need to be run after more refactoring before merging back to master, but this PR is only against a feature branch.

### DIFF
--- a/beater/agent_config_handler_integration_test.go
+++ b/beater/agent_config_handler_integration_test.go
@@ -1,0 +1,110 @@
+// Licensed to Elasticsearch B.V. under one or more contributor
+// license agreements. See the NOTICE file distributed with
+// this work for additional information regarding copyright
+// ownership. Elasticsearch B.V. licenses this file to you under
+// the Apache License, Version 2.0 (the "License"); you may
+// not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package beater
+
+import (
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/elastic/beats/libbeat/common"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/elastic/apm-server/beater/beatertest"
+	"github.com/elastic/apm-server/beater/headers"
+	"github.com/elastic/apm-server/beater/request"
+	"github.com/elastic/apm-server/tests"
+)
+
+func TestAgentConfigHandler_RequireAuthorizationMiddleware(t *testing.T) {
+	t.Run("Unauthorized", func(t *testing.T) {
+		cfg := cfgEnabledACM()
+		cfg.SecretToken = "1234"
+		rec := requestToACMHandler(t, cfg)
+
+		assert.Equal(t, http.StatusUnauthorized, rec.Code)
+		tests.AssertApproveResult(t, acmApprovalPath(t.Name()), rec.Body.Bytes())
+	})
+
+	t.Run("Authorized", func(t *testing.T) {
+		cfg := cfgEnabledACM()
+		cfg.SecretToken = "1234"
+		h, err := agentConfigHandler(cfg, beatertest.NilReporter)
+		require.NoError(t, err)
+		c, rec := beatertest.DefaultContextWithResponseRecorder()
+		c.Request.Header.Set(headers.Authorization, "Bearer 1234")
+		h(c)
+
+		assert.Equal(t, http.StatusServiceUnavailable, rec.Code)
+		tests.AssertApproveResult(t, acmApprovalPath(t.Name()), rec.Body.Bytes())
+	})
+}
+
+func TestAgentConfigHandler_KillSwitchMiddleware(t *testing.T) {
+	t.Run("Off", func(t *testing.T) {
+		rec := requestToACMHandler(t, DefaultConfig(beatertest.MockBeatVersion()))
+
+		assert.Equal(t, http.StatusForbidden, rec.Code)
+		tests.AssertApproveResult(t, acmApprovalPath(t.Name()), rec.Body.Bytes())
+
+	})
+
+	t.Run("On", func(t *testing.T) {
+		rec := requestToACMHandler(t, cfgEnabledACM())
+
+		assert.Equal(t, http.StatusServiceUnavailable, rec.Code)
+		tests.AssertApproveResult(t, acmApprovalPath(t.Name()), rec.Body.Bytes())
+	})
+}
+
+func TestAgentConfigHandler_PanicMiddleware(t *testing.T) {
+	h, err := agentConfigHandler(DefaultConfig(beatertest.MockBeatVersion()), beatertest.NilReporter)
+	require.NoError(t, err)
+	rec := &beatertest.WriterPanicOnce{}
+	c := &request.Context{}
+	c.Reset(rec, httptest.NewRequest(http.MethodGet, "/", nil))
+	h(c)
+	assert.Equal(t, http.StatusInternalServerError, rec.StatusCode)
+	tests.AssertApproveResult(t, acmApprovalPath(t.Name()), rec.Body.Bytes())
+}
+
+func TestAgentConfigHandler_MonitoringMiddleware(t *testing.T) {
+	beatertest.ClearRegistry(serverMetrics, ACMResultIDToMonitoringInt)
+	requestToACMHandler(t, DefaultConfig(beatertest.MockBeatVersion()))
+	equal, result := beatertest.CompareMonitoringInt(map[request.ResultID]int{request.IDRequestCount: 1}, ACMResultIDToMonitoringInt)
+	assert.True(t, equal, result)
+}
+
+func requestToACMHandler(t *testing.T, cfg *Config) *httptest.ResponseRecorder {
+	h, err := agentConfigHandler(cfg, beatertest.NilReporter)
+	require.NoError(t, err)
+	c, rec := beatertest.DefaultContextWithResponseRecorder()
+	h(c)
+	return rec
+}
+
+func cfgEnabledACM() *Config {
+	cfg := DefaultConfig(beatertest.MockBeatVersion())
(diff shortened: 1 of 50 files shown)


beater/agent_config_handler_integration_test.go
```diff
@@ -0,0 +1,110 @@
...
+	rec := &beatertest.WriterPanicOnce{}
+	c := &request.Context{}
+	c.Reset(rec, httptest.NewRequest(http.MethodGet, "/", nil))
+	h(c)
+	assert.Equal(t, http.StatusInternalServerError, rec.StatusCode)
+	tests.AssertApproveResult(t, acmApprovalPath(t.Name()), rec.Body.Bytes())
+}
+
+func TestAgentConfigHandler_MonitoringMiddleware(t *testing.T) {
+	beatertest.ClearRegistry(serverMetrics, ACMResultIDToMonitoringInt)
```
jalvz: Having to call `ClearRegistry` before a test is not great; this calls a bit for a `TestMain` with a setup function that clears the registry before each unit test, and then having each unit test needing a clean registry in that file.

beater/beatertest/config.go
```diff
@@ -0,0 +1,21 @@
...
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package beatertest
+
+// MockBeatVersion returns a beat version for testing
+func MockBeatVersion() string { return "8.0.0" }
```
jalvz: I can see where it is used, but I can't grasp how it happened that we need this now...

beater/beatertest/result.go
```diff
@@ -0,0 +1,23 @@
...
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package beatertest
+
+import "fmt"
+
+// ResultErrWrap wraps given input into the expected result error string
+func ResultErrWrap(s string) string { return fmt.Sprintf("{\"error\":\"%+v\"}\n", s) }
```
jalvz: there are a lot of shallow files in `beatertest`, can't these functions be packed together?

beater/smap_handler_integration_test.go
```diff
@@ -0,0 +1,125 @@
...
+	rec := &beatertest.WriterPanicOnce{}
+	c := &request.Context{}
+	c.Reset(rec, httptest.NewRequest(http.MethodGet, "/", nil))
+	h(c)
+	assert.Equal(t, http.StatusInternalServerError, rec.StatusCode)
+	tests.AssertApproveResult(t, assetApprovalPath(t.Name()), rec.Body.Bytes())
+}
+
+func TestAssetHandler_MonitoringMiddleware(t *testing.T) {
+	t.Run("Default", func(t *testing.T) {
```
jalvz: why `t.Run` here?

beater/smap_handler_integration_test.go
```diff
@@ -0,0 +1,125 @@
...
+	t.Run("On", func(t *testing.T) {
+		rec := requestToSmapHandler(t, cfgEnabledSmap())
+
+		assert.Equal(t, http.StatusBadRequest, rec.Code)
+		tests.AssertApproveResult(t, assetApprovalPath(t.Name()), rec.Body.Bytes())
+	})
+}
+
+func TestAssetHandler_LogMiddleware(t *testing.T) {
+	//TODO: How to test?
```
jalvz: maybe just with integration testing is enough?

beater/request/result.go
```diff
@@ -160,20 +120,128 @@ type Result struct {
...
+}
+
+func (r *Result) set(id ResultID, body interface{}, err error) {
+	if r == nil {
 		return
 	}
+	statusCode := http.StatusInternalServerError
+	keyword := KeywordResponseErrorsInternal
+
+	switch id {
```
jalvz: instead of having separate constants for everything and manually matching them here, why not have a map of `<ResultId, Pair(code, keyword)>` and replace the whole `switch` with a map lookup?

beater/beatertest/config.go
simitt: So far we used e.g. `defaultConfig("7.0.0")` all over the place in testing. I think it is nicer to introduce a dedicated mock version that can be used everywhere. This also allows for easier upgrade the version in the future.

beater/beatertest/result.go
simitt: I try to avoid having a file `utility` or similar where it is hard to keep overview and easy to just stuff things in. If you don't feel strongly about this, I'd rather have shallow files here.

beater/smap_handler_integration_test.go
simitt: Yes I fear we need to test this with the `system_tests`, tried to cover as much as possible in the go testing, but haven't found a good way to test the `logp.Logger` yet.

beater/request/result.go
simitt: You are right, I'll change that.

beater/agent_config_handler_integration_test.go
simitt: I prefer keeping the integration test files logically together, but I change the monitoring testing method accordingly, so that `clear` does not need to be called from outside.

simitt:
Thanks for the critical feedback @jalvz , much appreciated. I  addressed the smaller comments directly in code, and try to address some of the higher-level concepts here:

> I don't have a strong opinion on request.Context versus http.Request+httpResponse handling, but my biggest criticism is that the way it is implemented here is too much state-based imo.
    The problem of "response writing is all over the place" is not much solved because a pointer to the context is passed to many helper functions down the line, so it is still going to be relatively easy to eg. write the same header twice without noticing, and things like that.

I personally think it is fine and idiomatic to still keep the logic of gathering a response struct in the middleware and api handlers. The logic that does get abstracted is calling the actual `Write` and `WriteHeader` methods of the `http.ResponseWriter`. This is now all done when calling `context.Write()`, in the proper order.  But I am definitly open for suggestions, do you have an idea on how to improve this?

> I also feel that some things are more complicated now than before, eg. RequireAuthorization and SetAuthorization do something similar modifying the state of the context in a specific way, and you quite need to glue a few different bits to understand how it all comes together; when the older authHandler was fairly trivial to read.

The reason for introducing a dedicated `SetAuthorization` handler was to avoid passing the `secretToken` to more handlers and places than necessary. It is not a good pattern to pass sensitive information around in several places. The authorization middleware now takes care of everything related to the token. 

> request.Context has a large surface too - it doesn't guide me to do things the right way. I can for instance do c.Result.SetWithBody(...) in one handler, then c.Result.Err = x in other, and at the end c.Result.Failure() would be False and the error swallowed.

That's a fair point, I'll review this part again. 

> I don't think we are using Go interfaces in the way they are meant to be used, and I am seeing more of them now (eg asset.Processor, kibana.Client).
    Using an interface just so to mock something in a test is not the best pattern, and ideally those cases should be scoped just to external effects and be as narrow as possible (like RoundTrip)...

I have not introduced `asset.Processor` nor `kibana.Client` with this rework. I am going to introduce one interface with the next refactoring PR but will comment on that separately then. Therefore I suggest to keep this discussion out of this context. Would you mind opening a discuss issue and linking to some references to get the discussion around this started?

> I agree that the new WithMiddleware stuff reads nicer than before, but at the cost of turning something explicit into something implicit, namely order of computation. I can make peace with that, but I wouldn't go further down that route.

Why is the order of computation implicit now? The `WithMiddleware` ensures the middleware gets processed in the exact order of how you pass in the `middleware` elements. 

> Unrelated, but after reading much of this code again, I still don't get used to many of the unconventional abbreviations used everywhere (like "smapper"; and now "acm": it might makes perfect sense for us now, but eg. the new joiner in 1'5 years could have some issues with it). This refactor initiative wouldn't be the worse occasion to go over those and replace them with the full terms.

I added this as a point to the meta issue https://github.com/elastic/apm-server/issues/2489. This PR is against a feature branch, so I'd like to keep the renaming out of this PR, as I think it is big enough. I suggest to start collecting abbreviations you find unintuitive or you would like to see being changed in the meta issue and I follow up on that before merging back to master.

jalvz:
> all done when calling context.Write(),

And that is great, but what happens in `Write` depends on the object state which every handler can manipulate at will, thus introducing an implicit dependency between them.

An idea to improve it is an API so that each handler can only add stuff to the context, and not overwrite/remove what others did.
Imo it would be simpler if each handler only has access to data that they need to do their job and nothing else, but I think to achieve that you would have to go back to arguments.

> It is not a good pattern to pass sensitive information around in several places

I find that argument very brittle, the token has the same in-memory representation in one stack-frame than another, and it is not more or less vulnerable because of it?

> The authorization middleware now takes care of everything related to the token

I think before was also the case and it was only 1 handler instead of 2, but keep it as you find it better.

> Why is the order of computation implicit now?

This is actually related to above comments, but as said, highly subjective and not necessarily introduced in this PR!:

In `fA(fB(fC(), extra_arg))` I know exactly in which order functions run, what they take as arguments, and what do they return.
If a new joiner wants to add another handler, they just insert it wherever they want in the chain along with extra args it might require, and done.

In `WithMiddleware(fC, [fA, fB])` one needs to go read `WithMiddleware` implementation to know in which order arguments are evaluated, look up what is their signature, and then read each middleware to see what extra arguments do they need (encapsulated in the context).
If a new joiner wants to add another handler, they first need to read `WithMiddleware` (the function comment confuses me a bit, btw), then figure out how to pass extra arguments, arrive eventually at `request.Context`, add an attribute there, and hope that no one else touches it down the road...

You ofc skip all this mental process because you know the code already :)

simitt:
>It is not a good pattern to pass sensitive information around in several places

>>I find that argument very brittle, the token has the same in-memory representation in one stack-frame than another, and it is not more or less vulnerable because of it?

It was passed to multiple handlers that then decided what to do with the token. The token should not be internally logged, passing it around means one needs to verify this in multiple places. Decisions if a request is authorized or not was only made in the auth handler but also e.g. in the root handler and the agentconfig handler. 

I now added a check that the `Write` method is only called once. If it is called multiple times then a warning is logged and the result is not written again. 

@jalvz  if you are fine with these changes and agree that overall the PR introduced improvements that we want to get merged, I suggest to merge and move on to the package restructuring.

jalvz:
It's approved already

PR #2499, opened 2019-07-26
Log warning for RUM allow_origins setting '*'.

simitt:
cc @jahtalab @vigneshshanmugam


beater/server.go
```diff
@@ -79,6 +79,12 @@ func run(logger *logp.Logger, server *http.Server, lis net.Listener, config *Con
 	switch config.RumConfig.isEnabled() {
 	case true:
 		logger.Info("RUM endpoints enabled!")
+		for _, s := range config.RumConfig.AllowOrigins {
+			if s == "*" {
+				logger.Warn("CORS related setting `apm-server.rum.allow_origins` contains `*`. Consider more restrictive setting for production use.")
```
graphaelli: I'd suggest either `s/contains/is/` or something like:

```suggestion
				logger.Warn("CORS related setting `apm-server.rum.allow_origins` allows all origins. Consider more restrictive setting for production use.")
```

PR #2528, opened 2019-08-07
[logging] improve logging

simitt:
This builds on top of the logging restructuring and changes to use `result.Keyword` as logging keyword. 
It also logs errors and stacktraces in case they are set but status code is <400 (might be the case when recovering from a panic). 

implements elastic/apm-server#2201


simitt:
jenkins, run the tests again please

simitt:

simitt:

PR #2530, opened 2019-08-07
ci(jenkins): add more resilience to reduce any temporary glitches

### LINKED ISSUE #2529: [Automation][ci] more resilience to environmental issues
Some broken builds were related to some issues when accessing to dockerhub, in order to minimize this particular glitches we will retry twice with sleep time, it might take a bit longer the builds but ideally, it might reduce the broken builds when there are issues related to accessing dockerhub for instance.

An example of the typical environmental issue is:

image

v1v:
Closes #2529

## Highlights
- Retry twice in case of any errors when building or testing
- Sleep between 5-10 seconds for each iteration.
- This is a raw implementation to reduce the build errors which are related to an environmental issue.
- For sure it's not the cleverest way to provide more stability in the CI but it will eventually reduce the manual actions to rebuild when there is a temporary issue with third party systems.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -176,11 +176,8 @@ pipeline {
             withGithubNotify(context: 'Unit Tests', tab: 'tests') {
               deleteDir()
               unstash 'source'
-              dir(BASE_DIR){
-                retry(2) { // Retry in case there are any errors to avoid temporary glitches
-                  sleep randomNumber(min: 5, max: 10)
-                  sh(label: 'Run Unit tests', script: './script/jenkins/unit-test.sh')
-                }
+              dir("${BASE_DIR}"){
+                sh(label: 'Run Unit tests', script: './script/jenkins/unit-test.sh')
               }
             }
           }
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -170,8 +176,11 @@ pipeline {
             withGithubNotify(context: 'Unit Tests', tab: 'tests') {
               deleteDir()
               unstash 'source'
-              dir("${BASE_DIR}"){
-                sh(label: 'Run Unit tests', script: './script/jenkins/unit-test.sh')
+              dir(BASE_DIR){
+                retry(2) { // Retry in case there are any errors to avoid temporary glitches
+                  sleep randomNumber(min: 5, max: 10)
+                  sh(label: 'Run Unit tests', script: './script/jenkins/unit-test.sh')
+                }
```
simitt: I suggest to only add the retry for the `linux build` and `windows build` for now, as those were the stages where I mainly saw issues, and automatically retrying e.g. system tests would increase build times quite a bit in case real failures are found.

v1v: That's indeed a consequence of using this approach, sure thing, I'll then keep this strategy for those builds.

Issue #2502, opened 2019-07-29
ECS: Set host.name along with host.hostname in exported data


simianhacker:
APM server should be shipping `host.name` along with `host.hostname` per ECS. `host.name` should default to `host.hostname` but allow the user to override it with their host identifier if it's different. This is necessary to ensure cross-linking between Infra UI and APM is consistent.

roncohen:
Related: https://github.com/elastic/ecs/issues/498#issuecomment-512210825

roncohen:
We should probably set this in agents instead of in APM Server

graphaelli:
In what situation does cross-linking happen based on `host.name`?

simianhacker:
Everything except APM. In the Infra UI we have to put exceptions in for APM because they using `host.hostname` instead of `host.name`. As a company, we really need to standardize on one field across everything to represent an `ID` for a host. For Kubernetes Pods and Docker Containers we are all using the same `ID` fields. Although, it looks like there is `host.id` in the [ECS documentation](https://www.elastic.co/guide/en/ecs/current/ecs-host.html) but Metricbeat doesn't use it for some reason.  

Work Around Examples: 

- https://github.com/elastic/kibana/pull/42197/files#diff-0f965401975f16849fcc929d348aeebcR221-R227
- https://github.com/elastic/kibana/blob/master/x-pack/legacy/plugins/infra/public/components/waffle/node_context_menu.tsx#L49-L57

@elastic/beats @ruflin Any reason why we are not setting `host.id` in Metricbeat/Filebeat? Seems like that should be the blessed identifier field for Hosts instead of `host.name`

simitt:
Linking to some context around discussions of `host.name` vs. `host.hostname`: https://github.com/elastic/apm-server/pull/1609#discussion_r240687711. Haven't followed changes on `host` since then, but it seemed fair to leave the field empty if not user overwritten (which the agents don't support yet afaik).

graphaelli:
Thanks for digging that up @simitt, I had forgotten some of that context.

We can certainly copy `host.hostname` to `host.name` for APM, and let agents override that value.  We'd still need to keep some workaround in place as the curated UIs should fall back to linking based on `host.hostname` if `host.name` is not present, to support older 7.x data.

`host.id` would be great if multiple processes could arrive at the same value for it independently - that's why kubernetes and docker links work so well - but I don't think there is a portable way to do that well.

simitt:
[...]

PR #2540, opened 2019-08-08
Add new hostname information on Intake API and in ES.

simitt:
Deprecate existing `system.hostname` and add `system.detected_hostname`
and `system.configured_hostname` to the Intake API and map accordingly to `host.hostname` and `host.name` on ES level.

The `host.name` is quite an important information for linking between apps.

follow up on elastic/apm-server#2502

cc @elastic/apm-agent-devs this implements https://github.com/elastic/apm-server/issues/2502#issuecomment-519477402 to support setting ECS aligned `host.name` and `host.hostname` information in the future, while not breaking any current behavior.


simitt:

graphaelli:
commented on the linked issue, can we use `name` instead of `configured_hostname`?  Is there a set of tests with various input combinations and expected es documents?   I think it's a bit confusing to understand what will end up in es so wrote this up, think it's worth adding here or should approval tests cover this?


<details><summary>decode + transform system info</summary>

<p>

```go
func TestSystemDecodeAndTransform(t *testing.T) {
	for _, test := range []struct {
		input  map[string]interface{}
		output common.MapStr
	}{
		{
			input: map[string]interface{}{
				"hostname": "host",
			},
			output: common.MapStr{
				"hostname": "host",
				"name":     "host",
			},
		},
		{
			input: map[string]interface{}{
				"hostname":            "host",
				"configured_hostname": "configured",
			},
			output: common.MapStr{
				"hostname": "host",
				"name":     "configured",
			},
		},
		{
			input: map[string]interface{}{
				"hostname":          "host",
				"detected_hostname": "detected",
			},
			output: common.MapStr{
				"hostname": "detected",
				"name":     "detected",
			},
		},
		{
			input: map[string]interface{}{
				"hostname":            "host",
				"configured_hostname": "configured",
				"detected_hostname":   "detected",
			},
			output: common.MapStr{
				"hostname": "detected",
				"name":     "configured",
			},
		},

	} {
		sys, err := DecodeSystem(test.input, nil)
		require.NoError(t, err)
		assert.Equal(t, test.output, sys.fields())
	}
}
```
</p>
</details>

this doesn't even include the kubernetes permutations.

Should the intake reject payloads containing both `hostname` and the new fields?

simitt:
@graphaelli thanks for the review. 

There is the [metadata/system_test.go](https://github.com/elastic/apm-server/blob/fef24149b5ff0049b2ed563f5d4b3cac642fe278/model/metadata/system_test.go) file that tests the `decoding` for different variants (`hostname`, `configured_hostname`, `detected_hostname`) and the `transforming` including the Kubernetes variants. Did you miss those tests or do you suggest additional ones? IMO test coverage should be good.

>Should the intake reject payloads containing both hostname and the new fields?

I decided against it, as I don't really see value in being that strict here, rather just use the _newer_ values when given. No strong opinion though, if you think we should be more strict, I can change it. 

No strong feelings for the naming, also left a comment on the linked issue.

graphaelli:
> There is the metadata/system_test.go file ... Did you miss those tests or do you suggest additional ones?

The snippet I pasted into my comment was basically code from `system_test.go`.  I found it difficult to line up which cases were tested since the logic is split between decoding and transforming (as it should be) and couldn't tell if approval tests fully cover the cases, so I'm proposing either a new set of tests or additional approval tests that clarify intake -> persistence logic

simitt:
@graphaelli I updated the tests to use the approvals tool and run the `decode` and `transform` in separate test runs but on the same test set.

model/metadata/system_test.go
```diff
@@ -18,165 +18,181 @@
...
-		assert.Equal(t, test.err, err)
+
+		t.Run("Decode"+name, func(t *testing.T) {
+			system, err := DecodeSystem(test.input, test.inputErr)
+			assert.Equal(t, test.s, system)
+			assert.Equal(t, test.err, err)
+		})
+
+		if test.err == nil {
+			t.Run("Transform"+name, func(t *testing.T) {
```
graphaelli: can this be done within the `Decode` run to avoid decoding twice?  Not a huge deal if there is a reason to separate - would you just add that reason in a comment as this is surprising, to me at least, as is.

model/metadata/system_test.go
```diff
@@ -18,165 +18,181 @@
...
-				},
+		"empty ip": {
+			input: map[string]interface{}{"ip": ""},
+			s:     &System{IP: &empty},
+		},
+		"hostname": {
+			input: map[string]interface{}{"hostname": host},
+			s:     &System{DetectedHostname: &host},
+		},
+		"detected hostname": {
```
graphaelli: This shouldn't happen in practice right?  I like testing the path, just want to make sure I understand.

model/metadata/test_approved_system/transform_ignored_hostname.approved.json
```diff
@@ -0,0 +1,3 @@
+{
+    "name": "custom hostname"
```
graphaelli: why doesn't this get `hostname` set to the `hostname` provided?  This case shouldn't happen in practice right?

model/metadata/system_test.go
```diff
@@ -18,165 +18,181 @@
...
+		"k8s namespace with configured hostname": {
+			input: map[string]interface{}{
+				"kubernetes":          map[string]interface{}{"namespace": namespace},
+				"detected_hostname":   detected,
+				"configured_hostname": configured,
+			},
+			s: &System{Kubernetes: &Kubernetes{Namespace: &namespace}, DetectedHostname: &detected, ConfiguredHostname: &configured},
 		},
-		{
+		"k8s empty": {
```
graphaelli: this one covers it currently but for the sake of regression, might be worth adding a case for just `detected_hostname` and `configured_hostname` set

graphaelli:
The new tests are great, thanks for adding.

model/metadata/system_test.go
simitt: right, this shouldn't happen, but the server doesn't prevent it. I'll add a comment.

model/metadata/test_approved_system/transform_ignored_hostname.approved.json
simitt: The intake API accepts 3 fields relate to hostname: `hostname`, `detected_hostname` and `configured_hostname`. 

`hostname` is deprecated, while the other two fields are newly introduced. Agent devs are expected to not use `hostname` any more once they implement the new fields. Therefore, as soon as at least one of the new fields is sent, the deprecated one is ignored, as it is not supposed to be sent anymore. See also https://github.com/elastic/apm-server/issues/2502#issuecomment-519487367

simitt:
jenkins, run test again please

simitt:

PR #2545, opened 2019-08-12
automate Kibana index pattern syncs

### LINKED ISSUE #2354: Automate kibana saved objects sync
Something similar to updating beats, but for the kibana saved objects.

jalvz:
fixes #2354

### DIFF
--- a/script/update_kibana_objects.py
+++ b/script/update_kibana_objects.py
@@ -13,13 +13,12 @@ def exec(cmd):
     """
     Executes the given command and returns its result as a string
     """
-    proc = subprocess.Popen(shlex.split(cmd), stdout=subprocess.PIPE, stderr=subprocess.PIPE)
-    out, err = proc.communicate()
-    code = proc.returncode
-    if err:
-        print(err)
-    if code > 0:
-        sys.exit(code)
+    try:
+        out = subprocess.check_output(shlex.split(cmd))
+    except subprocess.CalledProcessError as e:
+        print(e)
+        sys.exit(1)
+
     return out.decode("utf-8")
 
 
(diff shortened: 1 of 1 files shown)


script/update_kibana_objects.py
```diff
@@ -0,0 +1,87 @@
...
+    """
+    Updates the index pattern in kibana according to the current checked out branch in apm-server
+
+    NOTES:
+    - HOME environment variable must be set
+    - apm-server can't be in detached head state
+    - `make update` must have been run previously
+    """
+
+    apm_dir = os.path.join(os.environ['HOME'], "go/src/github.com/elastic/apm-server")
```
graphaelli:
```suggestion
    apm_dir = os.path.abspath(os.path.join(os.path.dirname(__file__), "..")))
```

script/update_kibana_objects.py
```diff
@@ -0,0 +1,87 @@
...
+    os.chdir(apm_dir)
```
graphaelli: does this need to change dir (so that config dir is in the same dir?) or can the full path to `apm-server` be used instead?

script/update_kibana_objects.py
```diff
@@ -0,0 +1,87 @@
...
+        del index_pattern["attributes"]["title"]
+        kibana_file.seek(0)
+        kibana_file.write(json.dumps(index_pattern, indent=2, sort_keys=True))
+        kibana_file.truncate()
+
+    call("git add " + kibana_file_path)
+    call('git commit -m "update apm index pattern"')
+    call("git push origin update-apm-index-pattern-" + branch)
+
+    shutil.rmtree(dirpath)
```
graphaelli: where does `dirpath` come from?  Is that supposed to be the temporary kibana `path`

script/update_kibana_objects.py
```diff
@@ -0,0 +1,87 @@
...
+import json
+import tempfile
+import shlex
+import shutil
+import subprocess
+import sys
+import os
+
+
+def exec(cmd):
```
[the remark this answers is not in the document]
jalvz: afaics, `check_output` doesn't tell me if the command failed (exit code)

script/update_kibana_objects.py
jalvz: yeah, renaming leftover :+1:

script/update_kibana_objects.py
jalvz: either should work. im curious what is the downside of `chdir`?

script/update_kibana_objects.py
jalvz: im ok with that, but i usually test python in the shell and the `__file__` trick doesn't work there, which is annoying.
since everyone should have `$HOME` set any ways, i thought it was a better way

script/update_kibana_objects.py
graphaelli: I think it just makes refactoring simpler.  If the wd is really important for a group I'd suggest using a context manager for the block where the path is important so something like:

```python
# wd doesn't matter to code above this
with working_dir():
  do()
  ...
# wd doesn't matter to code below this
```

script/update_kibana_objects.py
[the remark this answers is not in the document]
graphaelli: it should raise a `CalledProcessError`

PR #2565, opened 2019-08-16
speed up integration tests

### LINKED ISSUE #2539: optimize build times
Right now build is taking ~30 minutes and some Python tests are excruciatingly slow (+15 secs), we should see how to speed it up.

jalvz:
This change reduces `INTEGRATION_TESTS=1 make system-tests` from ~8:30 mins on my machine to ~4:30 mins

The issue is that beats takes more than 0.1 seconds to index data on elasticsearch, so the refresh call was happening before the docs were inserted.

This change makes the refresh call actually effective.

~~Still, the whole suite runs locally in about 10 minutes; but the build on CI however takes easily 30 minutes, so we should look a bit into the Jenkins pipeline imo.~~

~~@elastic/observablt-robots any pointers on that above would be great. Are you aware of any points of contention? I inspected some pipeline logs top to bottom without reaching conclusion, but I feel there is a lot of scaffolding (git commands, make commands, docker downloads, etc)~~

On another note: I also feel that we are over-testing, we should be more frugal creating and waiting on documents; but probably doesn't make a lot of sense to invest more time on it now. Just a note for the future.

fixes elastic/apm-server#2539

### DIFF
--- a/script/jenkins/linux-test.sh
+++ b/script/jenkins/linux-test.sh
@@ -11,16 +11,6 @@ jenkins_setup
 #}
 #trap cleanup EXIT
 
-make test-deps testsuite
-
-export COV_DIR="build/coverage"
-
-for i in "full.cov" "integration.cov" "system.cov" "unit.cov"
-do
-  name=$(basename ${i} .cov)
-  if [ -f "${COV_DIR}/${i}" ]; then 
-    go tool cover -html="${COV_DIR}/${i}" -o "${COV_DIR}/coverage-${name}-report.html"
-    gocover-cobertura < "${COV_DIR}/${i}" > "${COV_DIR}/coverage-${name}-report.xml"
-  fi
-done
-exit 0
+make update
+make system-tests-environment
+make stop-environment
--- a/tests/system/apmserver.py
+++ b/tests/system/apmserver.py
@@ -81,6 +81,7 @@ class ServerSetUpBaseTest(BaseTest):
 
     def config(self):
         return {"ssl_enabled": "false",
+                "queue_flush": 0,
                 "path": os.path.abspath(self.working_dir) + "/log/*"}
 
     def setUp(self):
@@ -149,7 +150,7 @@ class ElasticTest(ServerBaseTest):
         cfg.update({
             "elasticsearch_host": self.get_elasticsearch_url(),
             "file_enabled": "false",
-            "kibana_host": self.get_kibana_url(),
+            "kibana_enabled": "false",
         })
         cfg.update(self.config_overrides)
         return cfg
@@ -208,13 +209,13 @@ class ElasticTest(ServerBaseTest):
         assert r.status_code == 202, r.status_code
 
         # make sure template is loaded
-        self.wait_until(
-            lambda: self.log_contains("Finished loading index template"),
-            max_timeout=max_timeout)
-        self.wait_until(lambda: self.es.indices.exists(query_index))
+        # self.wait_until(
+        #    lambda: self.log_contains("Finished loading index template"),
+        #    max_timeout=max_timeout)
+        #self.wait_until(lambda: self.es.indices.exists(query_index))
         # Quick wait to give documents some time to be sent to the index
         # This is not required but speeds up the tests
-        time.sleep(0.1)
+        time.sleep(2)
         self.es.indices.refresh(index=query_index)
 
         self.wait_until(
@@ -269,6 +270,7 @@ class ClientSideBaseTest(ServerBaseTest):
     def config(self):
         cfg = super(ClientSideBaseTest, self).config()
         cfg.update({"enable_rum": "true",
+                    "kibana_enabled": "false",
                     "smap_cache_expiration": "200"})
         cfg.update(self.config_overrides)
         return cfg
@@ -301,6 +303,8 @@ class ClientSideBaseTest(ServerBaseTest):
                                 'bundle_filepath': bundle_filepath,
                                 'service_name': service_name
                                 })
+        time.sleep(2)
+        self.es.indices.refresh()
         return r
 
 
--- a/tests/system/apmserver.py
+++ b/tests/system/apmserver.py
@@ -208,12 +208,7 @@ class ElasticTest(ServerBaseTest):
                               headers={'content-type': 'application/x-ndjson'})
         assert r.status_code == 202, r.status_code
 
-        # make sure template is loaded
-        # self.wait_until(
-        #    lambda: self.log_contains("Finished loading index template"),
-        #    max_timeout=max_timeout)
-        #self.wait_until(lambda: self.es.indices.exists(query_index))
-        # Quick wait to give documents some time to be sent to the index
+        # Wait to give documents some time to be sent to the index
         # This is not required but speeds up the tests
         time.sleep(2)
         self.es.indices.refresh(index=query_index)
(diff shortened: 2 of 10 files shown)


jalvz:
jenkins retest this please

jalvz:
jenkins retest this please

jalvz:
While the improvement on CI doesn't seem to be as effective as locally (+50% reduction with latest changes), this modifies the stage in the pipeline to only run `make system-tests-environment` instead of `make testsuite`, as unit test already run as part of another stage. It also removes the cover reporting, as for Python tests is empty.

Overall improvement is ~40-50% in this stage, https://github.com/elastic/apm-server/pull/2614 should help with the whole build.

script/jenkins/linux-test.sh
```diff
@@ -11,16 +11,6 @@ jenkins_setup
...
-export COV_DIR="build/coverage"
-
-for i in "full.cov" "integration.cov" "system.cov" "unit.cov"
-do
-  name=$(basename ${i} .cov)
-  if [ -f "${COV_DIR}/${i}" ]; then 
-    go tool cover -html="${COV_DIR}/${i}" -o "${COV_DIR}/coverage-${name}-report.html"
-    gocover-cobertura < "${COV_DIR}/${i}" > "${COV_DIR}/coverage-${name}-report.xml"
-  fi
-done
```
simitt: Maybe I overlooked it, but I couldn't find a coverage stage or report anymore with this being removed.

tests/system/apmserver.py
```diff
@@ -208,13 +209,13 @@ def load_docs_with_template(self, data_path, url, endpoint, expected_events_coun
...
 
         # make sure template is loaded
-        self.wait_until(
-            lambda: self.log_contains("Finished loading index template"),
-            max_timeout=max_timeout)
-        self.wait_until(lambda: self.es.indices.exists(query_index))
+        # self.wait_until(
+        #    lambda: self.log_contains("Finished loading index template"),
+        #    max_timeout=max_timeout)
+        #self.wait_until(lambda: self.es.indices.exists(query_index))
```
simitt: Please remove if no longer needed

tests/system/apmserver.py
```diff
@@ -208,13 +209,13 @@ def load_docs_with_template(self, data_path, url, endpoint, expected_events_coun
...
         # Quick wait to give documents some time to be sent to the index
         # This is not required but speeds up the tests
-        time.sleep(0.1)
+        time.sleep(2)
```
simitt: Can you explain how this speeds up tests? Now every time a template is loaded this for sure waits 2 seconds.

tests/system/apmserver.py
```diff
@@ -301,6 +303,8 @@ def upload_sourcemap(self, file_name='bundle_no_mapping.js.map',
                                 'bundle_filepath': bundle_filepath,
                                 'service_name': service_name
                                 })
+        time.sleep(2)
```
simitt: same question as above

tests/system/test_requests.py
```diff
@@ -109,10 +109,6 @@ def test_ok(self):
         r = self.request_intake()
         assert r.status_code == 202, r.status_code
 
-    def test_sourcemap_upload(self):
-        r = self.upload_sourcemap(file_name='bundle.js.map')
-        assert r.status_code == 202, r.status_code
```
simitt: Why is this removed, is it covered somewhere else?

script/jenkins/linux-test.sh
```diff
@@ -11,16 +11,6 @@ jenkins_setup
...
-exit 0
+make update
+make system-tests-environment
+make stop-environment
```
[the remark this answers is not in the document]
jalvz: yes, that is intentional. the `integration` build just runs the unit tests in a provided docker env as far i can see. i think it is redundant since we have a separate unit test stage. if you think it is important, we could update that stage to use the docker env _instead_, but run them twice is a waste imo
i also don't see the point at all of dockerizing go unit tests

tests/system/test_requests.py
jalvz: yeah, in many places, eg https://github.com/elastic/apm-server/blob/502f15587e048ef508f32e41a3b78cb2ea8cfe10/tests/system/test_integration.py#L332

tests/system/test_setup_index_management.py
```diff
@@ -2,6 +2,7 @@
 from nose.plugins.attrib import attr
 import unittest
 import logging
+import time
```
jalvz: leftover, will remove

script/jenkins/linux-test.sh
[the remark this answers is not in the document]
simitt: I see your point with the docker env, and I agree that we should not run the same tests more often than necessary. 

My concerns refer to `go test -tags=integration` not being run anywhere anymore. Which, again is not an issue at the moment, we just have to ensure to add it in case we do make use of the golang [build constrain](https://golang.org/pkg/go/build/#hdr-Build_Constraints) for integration tests in the future. 

So these changes are fine with me, just wanted to be explicit about implications.

tests/system/apmserver.py
jalvz: apm-server+libbeat needs 1-2 secs (even with the `flush` change) to index a single event in elasticsearch. If we do not wait, the next call (index refresh) will happen _before_ the indexing happens, thus making the refresh ineffective.

Without an index refresh, you need ~4-5 seconds to query a document and "see" it. This way we put a higher lower bound (2 secs), but it almost guarantees that the refresh happens _after_ the index operation, so the following query (`wait_until`...) returns immediately.

This was made worse by libbeat batching events and taking more time to index them.... some tests were taking 30 seconds just to index and query 3 or 4 documents! 

That said, in my local ~0.5-1 seconds sleep seems to give optimal results, but in Jenkins everything is slower (likely due shared resources) so I think I am going to parametrize this to further speed it up locally.  

I hope this makes sense

script/jenkins/linux-test.sh
jalvz: we don't have (and never had) any coverage for python tests, do we? 
we should still have them for unit tests, they are generated in the corresponding stage.

tests/system/apmserver.py
simitt: It sure does, thanks for the details

script/jenkins/linux-test.sh
simitt: I am probably just blind, but I really can't find where the code coverage for the unit tests should be living now.

jalvz: Ok, where do you look them up? I confess I never been interested in them, and I was expecting to be generated here: https://github.com/elastic/apm-server/blob/master/script/jenkins/unit-test.sh

But now that you mention it, I don't see any `go tool cover` there, just the test run report. So if that it is important I think that is the right place. Makes sense?

simitt: Yes I agree that the unit tests would be the right place to create the coverage report now. I don't look at them quite often, but sometimes on a PR the code coverage stage failed, and I assumed that was done by the now removed code. 
If you look at the last stage of the `system tests`, e.g. [here](https://apm-ci.elastic.co/blue/organizations/jenkins/apm-server%2Fapm-server-mbp/detail/PR-2621/2/pipeline/156) it gives you a link to navigate to the code coverage overview.

jalvz: Good point, thanks.

jalvz: fixed now https://apm-ci.elastic.co/job/apm-server/job/apm-server-mbp/job/PR-2565/30/Coverage-Sourcecode-Files/

PR #2567, opened 2019-08-16
Docs: Asciidoctor compatibility

nik9000:
Fixes all build errors I found when trying to build the docs using
asciidoctor.

PR #2570, opened 2019-08-16
Docs: Asciidoctor compatibility (7.2) (#2567)

nik9000:
Fixes all build errors I found when trying to build the docs using
asciidoctor in the master branch.


nik9000:
Note: the docs don't actually successfully build with asciidoctor in 7.2, but I want to backport #2567 as far as possible and then pick up other failures from there.

PR #2578, opened 2019-08-16
[docs][7.2] Fix fields.asciidoc for asciidoctor

bmorelli25:
This makes 7.2 build in asciidoctor! Needs to be backported.

Closes https://github.com/elastic/apm-server/issues/2577. Related beats PR: https://github.com/elastic/beats/pull/12242.

PR #2585, opened 2019-08-19
[docs][7.2] Sync generated fields

bmorelli25:
Related to https://github.com/elastic/apm-server/pull/2578. Syncs `generate_fields_docs.py` with Beats.


bmorelli25:
@nik9000 can you look at the HTML diff for this change? Looks like every page changed, but I'm not sure why. http://apm-server_2585.docs-preview.app.elstc.co/diff

nik9000:
> @nik9000 can you look at the HTML diff for this change? Looks like every page changed, but I'm not sure why. http://apm-server_2585.docs-preview.app.elstc.co/diff

It picked up a change from the docs repo. It should be safe. You can rerun it if you need a clean diff. Here is the [real](https://github.com/elastic/built-docs/commit/44a7b7b54ec237603b551e374b8c06a75c2740cb) diff.

bmorelli25:
Ahh, that makes sense! Thank you.

PR #2588, opened 2019-08-19
ES fields: Index socket.remote_address as client.ip

### LINKED ISSUE #2262: Apply Geo-IP pipeline also on http.request.socket.remote_address
Currently the geo-ip pipeline is defined to be applied only on RUM agent based events on the [`client.ip`](https://github.com/elastic/apm-server/blob/master/ingest/pipeline/definition.json#L43) field. 

~We could also apply it to process the `host.ip` information and store into `host.geo`.~

Update:
The `host.ip` is the ip parsed from the requests remote address, which is always the host from which the request is sent to the server. 
The agents can send up the `remote_ip` of the original request though, which is stored in `http.request.socket.remote_address`. It would make sense to apply the geo-ip pipeline on this field.

simitt:
closes elastic/apm-server#2262 

@graphaelli and @roncohen I copy the value from `http.request.socket.remote_address to client.ip`. There will be no distinction between `client.ip` collected by RUM or other apm agents. 
If only a certain set of data should be used, one could make use of additional information, e.g. `agent.name`. This is also the reason I dediced to not add an extra value to `client`, indicating whether or not data come from a real end user.


simitt:

PR #2595, opened 2019-08-19
[docs][7.1] Remove arbitrary python expressions

bmorelli25:
For https://github.com/elastic/docs/pull/1083. A few more problems that weren't showing up on the local build.

In Asciidoctor, the old python expressions with `or` all evaluate to `true`. This PR updates them to `ifdef` or `ifndef` directives.


docs/copied-from-beats/command-reference.asciidoc
```diff
@@ -82,31 +82,31 @@ endif::[]
 [options="header"]
 |=======================
 |Commands |
-ifeval::[("{beatname_lc}"=="functionbeat")]
+ifdef::functionbeat[]
```
nik9000: Single equality *should* work in asciidoctor.

bmorelli25: Good catch. They do work. I've removed them from this PR.

PR #2609, opened 2019-08-21
auto-generate exception parent and remove property from the intake

jalvz:
- [x]  update changelog

### DIFF
--- a/processor/stream/test_approved_es_documents/testIntakeIntegrationErrors.approved.json
+++ b/processor/stream/test_approved_es_documents/testIntakeIntegrationErrors.approved.json
@@ -107,7 +107,7 @@
                     },
                     {
                         "message": "on top of it, internet doesn't work",
-                        "parent": 0,
+                        "parent": 1,
                         "type": "ConnectionError"
                     }
                 ],
--- a/beater/test_approved_es_documents/TestPublishIntegrationErrors.approved.json
+++ b/beater/test_approved_es_documents/TestPublishIntegrationErrors.approved.json
@@ -116,7 +116,7 @@
                     },
                     {
                         "message": "on top of it, internet doesn't work",
-                        "parent": 0,
+                        "parent": 1,
                         "type": "ConnectionError"
                     }
                 ],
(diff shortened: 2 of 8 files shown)


processor/stream/test_approved_es_documents/testIntakeIntegrationErrors.approved.json
```diff
@@ -104,7 +104,7 @@
                     },
                     {
                         "message": "on top of it, internet doesn't work",
-                        "parent": 0,
+                        "parent": 1,
```
simitt: From how I understand the discussions I would assume that the exception with message `"something wrong writing a file"` is the parent for both other exceptions. The exception `"disk spinning way too fast"` doesn't need to set a parent, as it directly follows in the array, but `"on top of it, internet doesn't work"` needs to set it. 
I read `parent: 1` as exception with message `disk spinning way too fast` is the parent, especially with the comment above that indexing is zero based (https://github.com/elastic/apm-server/pull/2609/files#diff-52137911864c236a97b14516397a04c6R448)
So my expectation how it should be and what the example indicates do not meet. Could you please clarify.

jalvz: https://github.com/elastic/apm-server/pull/2609/files/5570b3b64b6d14bba268a9c94309969954323b65#diff-6efe590e6f69aa2f41fce1abcf1b1b37R2 prettified and with only relevant attributes looks like this:
```
{
   "error":{
      "id":"0123456789012345",
      "timestamp":1494342245999999,
      "exception":{
         "message":"The username root is unknown",
         "type":"DbError",
         "cause":[
            {
               "type":"InternalDbError",
               "message":"something wrong writing a file",
               "cause":[
                  {
                     "type":"VeryInternalDbError",
                     "message":"disk spinning way too fast"
                  },
                  {
                     "type":"ConnectionError",
                     "message":"on top of it, internet doesn't work"
                  }
               ]
            }
         ],
      },
      "context":{
        ...
      }
   }
}
```
After array serialization:
```
[{"type":"DbError",...},
 {"type":"InternalDbError",...},
 {"type":"VeryInternalDbError",...},
 {"type":"ConnectionError",..., "parent":1}]
```
The top exception `DbError` is the one with index (`posId`) 0, its single cause has index 1 and contains 2 causes; the second of which needs an explicit `parent`. Otherwise would be wrongly linked to its "sibling" (the precedent exception in the list).

Does it make sense?

simitt: Totally, I overlooked the `DbError` exception 👍

PR #2614, opened 2019-08-22
ci(jenkins): build and test stages are now merged.

v1v:
Closes https://github.com/elastic/apm-server/issues/2615

## Highlights
-  Warn in case of errors when running the Windows stage.
- Build and Test stages are now merged to run more in parallel and reduce the sequential stages to have a quick cycle.

## Test cases
- When the windows stage failed and other stages are still running 

image

- When the windows stage failed and the pipeline also failed

image

- More parallel branches

image

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -101,7 +101,7 @@ pipeline {
         }
       }
     }
-    stage('Build'){
+    stage('Build and Test'){
       failFast false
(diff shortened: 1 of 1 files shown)


jalvz:
Can you also move the "Check Kibana obj. updated" to run in parallel with the rest of the bunch?

PR #2635, opened 2019-08-26
Update beats framework and go version

simitt:
* Update beats framework to 9870f90
* Update go version to 1.12.9


simitt:
@jalvz as just discussed offline, would appreciate if you could take over integrating the `mage` changes, as you are more familiar with it and I don't think I find time for digging into it right now.

vendor/vendor.json
```diff
@@ -2,6 +2,18 @@
...
+			"revision": ""
+		},
+		{
+			"path": "appengine_internal",
+			"revision": ""
+		},
+		{
+			"path": "appengine_internal/base",
+			"revision": ""
+		},
```
simitt: Actually this looks weird, I have to investigate.

PR #2636, opened 2019-08-26
Update license text for repo.

simitt:
PR for updating license text in beats: https://github.com/elastic/beats/pull/13347


jalvz:
fyi beats/#13347 was (at least temporarily) reverted, they might come up with a different version.

not sure what do you want to do here, i don't have any strong arguments so i will leave it as is

PR #2644, opened 2019-08-27
ci(jenkins): update timeout and when regexp comparator

v1v:
## Highlights
- As there are more parallel stages/branches, let's reduce the timeout
- Update when using the regexp comparator.

Depends on https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/351


v1v:
Apparently, the comparator only works for the tags rather than branches, I'll raise a PR to the upstream to support this particular specs. So I'll keep this particular PR open until the upstream plugin supports it.

simitt:
@v1v any updates on that? Could you link the upstream PR here?

v1v:
It's now supported in the upstream:
- https://github.com/jenkinsci/pipeline-model-definition-plugin/releases/tag/pipeline-model-definition-1.4.0

We might need to wait for the upcoming upgrade of our current CI environment to be able to use this feature.

v1v:
jenkins run the tests please

v1v:
We are good to go, current build issues are not relevant to this particular PR though

PR #2657, opened 2019-08-29
[7.x] remove empty folders (deleted deps) after beats update (#2656)

jalvz:
 - remove empty folders (deleted deps) after beats update  (#2656)


jalvz:
also adds missing `check-changelogs`

PR #2658, opened 2019-08-29
[7.x] Update beats

jalvz:


jalvz:
Note: 7174251bdedabd463a25762a09c9d6ffad089f1f will be invalidated again with https://github.com/elastic/beats/pull/13405, but need it for now for the build to pass

PR #2659, opened 2019-08-29
Update beats

jalvz:


jalvz:
Note: 42692cbc97716d4b066a638af307542bf4efb8a8 will be invalidated again with https://github.com/elastic/beats/pull/13405, but need it for now for the build to pass

PR #2671, opened 2019-09-03
[Docs] .NET - rename Tags to Labels

gregkalapos:
From https://github.com/elastic/apm-agent-dotnet/issues/423 

Similar to Java, we kept the url to `public-api.html#api-transaction-tags` - let us know in case we'd like to tackle that, but atm just to be on the safe side we only rename the text, the url remans.

### DIFF
--- a/docs/guide/apm-data-model.asciidoc
+++ b/docs/guide/apm-data-model.asciidoc
@@ -19,7 +19,7 @@ They measure from the start to end of an activity,
 and they can have a parent/child relationship with other spans.
 
 Agents automatically instrument a variety of libraries to capture these spans from within your application.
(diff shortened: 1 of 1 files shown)


docs/guide/apm-data-model.asciidoc
```diff
@@ -19,7 +19,7 @@ They measure from the start to end of an activity,
 and they can have a parent/child relationship with other spans.
 
 Agents automatically instrument a variety of libraries to capture these spans from within your application.
-In addition, you can use the Agent API for ad hoc instrumentation of specific code paths.
```
gregkalapos: auto format spotted the empty space there 😇

PR #2690, opened 2019-09-06
[DOCS] Correcting a few typos to match the Style Guide

Titch990:
As discussed with @bmorelli25.

Correcting "open-source" to "open source", "APM UI" to "APM app" according to our Style Guide. 

In fact, the Style Guide says that "app" can be omitted completely, if appropriate, but in this case, since APM has several components, I feel "APM app" is better for disambiguation.

However, I think replacing "Kibana UI" with just" Kibana" instead of "Kibana app" is OK.

Refs for interest:
https://wiki.elastic.co/display/MAR/Terms+quick+reference ("open source")
https://wiki.elastic.co/display/MAR/Term+usage+guide (use of "app)


bmorelli25:
Thanks @Titch990! As mentioned in Slack, I'm going to hold off on merging this for now. I need to tackle this across all 9 APM books as a part of a bigger effort.

PR #2706, opened 2019-09-12
make the rate limiter reusable

jalvz:
The RUM endpoint for central config needs rate limiting, but it can't be reused because now is implemented inside the intake handler. This extracts it to its own middleware.

Needs some more testing...


beater/middleware/rate_limit_middleware.go
```diff
@@ -0,0 +1,38 @@
...
+
+package middleware
+
+import (
+	"github.com/elastic/apm-server/beater/api/ratelimit"
+	"github.com/elastic/apm-server/beater/config"
+	"github.com/elastic/apm-server/beater/request"
+)
+
+const burstMultiplier = 3
```
simitt: If making this generic then the `burstMultiplier` should also be a param instead of a const.

beater/middleware/rate_limit_middleware.go
```diff
@@ -0,0 +1,38 @@
...
+
+// RateLimitMiddleware adds a rate limiter per IP to the request context
+func RateLimitMiddleware(cfg *config.EventRate) Middleware {
+	cache, err := ratelimit.NewLRUCache(cfg.LruSize, cfg.Limit, burstMultiplier)
```
simitt: This is an error that occurs when creating and registering the API handlers. Such an error should not be carried to every single request that is wrapped by this middleware, but rather let the Server startup fail (or recover in whichever other way).

beater/middleware/rate_limit_middleware.go
```diff
@@ -0,0 +1,38 @@
...
+import (
+	"github.com/elastic/apm-server/beater/api/ratelimit"
+	"github.com/elastic/apm-server/beater/config"
+	"github.com/elastic/apm-server/beater/request"
+)
+
+const burstMultiplier = 3
+
+// RateLimitMiddleware adds a rate limiter per IP to the request context
+func RateLimitMiddleware(cfg *config.EventRate) Middleware {
```
simitt: Providing a `RateLimiterMiddleware` I would expect the middleware to take care of the rate limiting already and not only of setting a rate limit manager. The current rate Intake API doesn't really allow that, as the rate limiting occurs depending on the amount of events. 

For something like e.g. ACM for RUM we will probably be able to apply an easier rate limiting strategy as it doesn't need to throttle processing of events, but can simply be an allow or deny strategy (similar to the killswitch middleware but asking a rate limiter). I'd assume this will be doable in a rate limit middleware without carrying logic over to the handler. 

I assume the refactoring is based on the rate limiter needs for RUM ACM. As pointed out I don't think that this will require the same rate limiting handling as for Intake API, so I don't see a lot of value in this concrete middleware. If you still want to move forward with it, then please rename to something that makes it clear that this only _sets_ a rate limiter but doesn't apply it.

beater/middleware/rate_limit_middleware.go
jalvz: Exactly, that is how it works unless I did something wrong

beater/middleware/rate_limit_middleware.go
jalvz: why?

beater/middleware/authorization_middleware.go
```diff
@@ -41,19 +41,19 @@ func RequireAuthorizationMiddleware(token string) Middleware {
 
 			c.Authorized = true
 			h(c)
-		}
+		}, nil
```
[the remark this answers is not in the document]
jalvz: I disagree, and prefer to use state at least as possible. Also, if I understand correctly, if I were to do that, I would be returning an error on each request instead of once during initialization.

beater/middleware/rate_limit_middleware.go
jalvz: It will still require a rate limiter per IP, and like this we can reuse as much logic as possible. But since everything else is a middleware also, it also makes sense to treat rate limiting in the same way. It is not so different eg. than the logging middleware, which puts some stuff in the context and it gets used later on

beater/middleware/middleware_test.go
```diff
@@ -33,22 +33,29 @@ func TestWithMiddleware(t *testing.T) {
...
+	h, e := Wrap(h, m1, m2)
+	require.NoError(t, e)
+	h(c)
 	require.NotNil(t, c.Result.Keyword)
 	assert.Equal(t, "[m1.1][m2.1][h][m2.2][m1.2]", c.Result.Keyword)
 }
+
+func Apply(m Middleware, h request.Handler) request.Handler {
+	r, _ := m(h)
+	return r
```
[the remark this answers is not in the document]
jalvz: I don't see why a middleware can't return an error if it fails to initialize itself - as a matter of fact I think this makes the general handling more flexible. Even more so, I think the `request.Handler` should also return an error to be propagated downstream, instead of maintaining some state in the context - but didn't want to make more changes than strictly necessary.
(And this is just a test helper by the way)

beater/middleware/rate_limit_middleware.go
simitt: 🤦‍♀ right, I misread as part of the actual request handling, fine as is.

beater/middleware/authorization_middleware.go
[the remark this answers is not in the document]
simitt: see https://github.com/elastic/apm-server/pull/2706#discussion_r324140270

beater/middleware/rate_limit_middleware.go
simitt: Because it makes the cache creation more flexible and we might not want to use the same burst multiplier for different rate limiters.

beater/middleware/rate_limit_middleware.go
simitt: The purpose of a logging middleware is to log something and not to decline a request, so I don't think the comparison is appropriate. 

I do agree that when having an allow/deny policy for a rate limiter we should definitly add this as a middleware. That would be a different one though (as pointed out above). 
As said before, not against it, just don't see a lot of value in it and definitely would rename to something more clear.

beater/middleware/middleware_test.go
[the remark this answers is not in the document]
simitt: As mentioned https://github.com/elastic/apm-server/pull/2706#discussion_r324140270 I was misreading the level of where the error was returned. My comment was referring to that error and not meant as a general comment on a middleware not returning an error.

beater/middleware/rate_limit_middleware.go
jalvz: So are you suggesting to make a config option for that? I'd rather don't overwhelm users with more settings until proof they are actually needed, and tackle that problem when we have it. This is for now an implementation detail.

If anything, it _could_ be relevant a different rate limit per this endpoint, but that is out of scope here.

beater/middleware/rate_limit_middleware.go
jalvz: > just don't see a lot of value in it

The value is code reuse, and consistency: extract functionality that can extend the capabilities of any handler, without knowledge of the handler internals. 
Would you prefer to hold onto this until see how is applied to the new endpoint?

> rename to something more clear

"RateLimiterManagerMiddleware"? :thinking:

simitt: `SetRateLimiterMiddleware`?

beater/middleware/rate_limit_middleware.go
simitt: Definitly not suggesting a config option. If you feel so strongly against making this a parameter to the middleware, leave it as is.

beater/middleware/rate_limit_middleware.go
jalvz: Having a verb there reads weird to me. The comparison with the logging middleware was because it creates a logger, sets it in the context, and then it is read/used from handlers (_and_ also from the middleware, yes). Yet it is not called `SetLoggerMiddleware`.

The reason why names are confusing to me is the "manager" bit. The `RateLimitManager` is effectively a rate limiter per IP, I don't understand why `"per IP" == "Manager", the fact that there is a cache + logic that _manages_ something  is an unimportant detail, the important detail is the "per IP" part.
So why not call the whole thing `IPRateLimiter`, and then this `IPRateLimitMiddleware`? 

Ideally it should also be dealing with the "per IP" part, moving this logic https://github.com/elastic/apm-server/blob/0ccad2049ec81008528074d67fe112aeb2e5d6aa/beater/api/intake/handler.go#L57 (and the `rateLimit` implementation) to the middleware, and setting in the context the specific `rate.Limiter` for that request. 

It looks like your confusion comes from the fact that the middleware doesn't do much, would that help?
Again, I didn't want to make this bigger than strictly needed but I am happy to do that just here or in another PR.

simitt: The log middleware is setting the logger, but it also takes care of all the logging during the request handling. The only reason it also sets the logger in the context is to ensure the edge case of a response not being able to be written is covered. I still do not agree that this comparison is a good one. 
You can compare it with the [SetAuthorizationMiddleware](https://github.com/elastic/apm-server/blob/master/beater/middleware/authorization_middleware.go#L50). Two authorization middlewares are available, one for only setting flags and one for rejecting the whole request.

Let's focus on the middleware naming first: My point is, this middleware doesn't do anything other than setting a value. From the name of the middleware I expect the middleware to also apply the rate limiting, which it doesn't. This needs to be clarified.

Regarding your second point:
The `RateLimiterManager` currently is an interface with one method `RateLimiter`. It is not concerned _at all_ with what the key for rate limiting is. And it shouldn't. The IP part is handled by the API [handler](https://github.com/elastic/apm-server/blob/master/beater/api/intake/handler.go#L155). Now when reusing the whole rate limiting logic, it might make sense to build some IPRateLimiter logic that composes the existing rate limiter logic to one logic and then use this everywhere. But it should not be mixed into the current rate limiter manager. Let's keep concerns separated.

simitt: > Again, I didn't want to make this bigger than strictly needed but I am happy to do that just here or in another PR.

+1 on not making this bigger as necessary here. I am not saying you should do more refactoring, only advocating for a different name.

beater/api/mux.go
```diff
@@ -131,26 +128,22 @@ func rumHandler(cfg *config.Config, reporter publish.Reporter) (request.Handler,
...
+	return middleware.Wrap(h, rumMiddleware(cfg)...)
 }
 
 func rumMiddleware(cfg *config.Config) []middleware.Middleware {
 	return append(apmMiddleware(intake.MonitoringMap),
 		middleware.KillSwitchMiddleware(cfg.RumConfig.IsEnabled()),
 		middleware.RequestTimeMiddleware(),
-		middleware.CORSMiddleware(cfg.RumConfig.AllowOrigins))
+		middleware.CORSMiddleware(cfg.RumConfig.AllowOrigins),
+		middleware.SetRateLimitMiddleware(cfg.RumConfig.EventRate))
```
simitt: Nit: I'd put the rate limit middleware right behind the killswitch one. In this case it doesn't make a difference since none of the other middlewares use it, but that should be an implementation detail.

beater/middleware/middleware.go
```diff
@@ -22,13 +22,18 @@ import (
...
 	for i := len(m) - 1; i >= 0; i-- {
-		h = m[i](h)
+		var e error
+		h, e = m[i](h)
+		if err == nil && e != nil {
+			err = e
+		}
 	}
-	return h
+	return h, err
```
simitt: I wonder why you don't immediately `return nil, err` when an error occurs?

beater/middleware/rate_limit_middleware.go
```diff
@@ -0,0 +1,38 @@
...
+
+import (
+	"github.com/elastic/apm-server/beater/api/ratelimit"
+	"github.com/elastic/apm-server/beater/config"
+	"github.com/elastic/apm-server/beater/request"
+)
+
+const burstMultiplier = 3
+
+// SetRateLimitMiddleware adds a rate limiter per IP to the request context
```
simitt: The description does not fit, as the rate limiter itself could also be used to e.g. rate limit per secret token or anything else, not bound to IP.

beater/request/context.go
```diff
@@ -54,7 +56,7 @@ func (c *Context) Reset(w http.ResponseWriter, r *http.Request) {
 	c.Logger = nil
 	c.TokenSet = false
 	c.Authorized = false
-
+	// TODO i guess i should not do reset the rate limit manager ??
```
simitt: It needs to be reset. The middleware ensures the rate limiter is set for every request. The context is reused for any other request. What if it is reused for a request to a different endpoint with a different rate limiter?

jalvz: ok. i don't get it, thou. the rate limiter is stateful, if i reset it every request then is not the case that nothing will ever be rate limited?

simitt: You do create the rate limiter instance once https://github.com/elastic/apm-server/pull/2706/files#diff-7b23454a544752fbe51da51555207091R30. 
For every request that is wrapped in the same middleware the same rate limiter is then set https://github.com/elastic/apm-server/pull/2706/files#diff-7b23454a544752fbe51da51555207091R34.

jalvz: right, thanks

beater/middleware/rate_limit_middleware.go
jalvz: I was trying to document something useful. I guess I can leave it to "SetRateLimitMiddleware sets a rate limiter" :man_shrugging:

beater/api/mux.go
jalvz: out of curiosity, if it doesn't make a difference, why higher in the list is better?

simitt: IMO: Theoretically a rate limiter should be set as soon as possible, so it can be applied as soon as possible. Now the way we use it in the Intake API is special, so it doesn't make a difference. But what if the `CORSMiddleware` used it for some check already to not even handle a preflight request if rate limit is exceeded? 

I know, it's not important now, as it is just theoretical.. that's why I said _NIT_ and didn't require to change..

beater/api/ratelimit/rate_limit_lru_cache.go
```diff
@@ -15,7 +15,7 @@
 // specific language governing permissions and limitations
```
simitt: minor detail, since you moved it to `ratelimit` package, I think it is a leftover that the file is called `rate_limit_lru_cache` instead of `lru_cache`.

beater/middleware/rate_limit_middleware.go
jalvz: I can put another example: all that `RequestTimeMiddleware` does is set `time.Now()` in the request context, yet it is not called `SetRequestTimeMiddleware`.

My point about names is that the should convey _what_ is done, not _how_ something is done. Setting a variable is an implementation detail (_how_), and someone new to the codebase wouldn't find `SetXXXMiddleware` any more enlightening.

I generally don't complaint about names unless they lead me in the wrong direction, which is the case for this suggestion: `SetXXX` is a widespread naming convention for setters, which is not what the middleware is. Functions called `SetXXX` that return something are always confusing to me...

Similar thing goes for `RateLimiterManager`, that "Manager" suffix doesn't help to understand _what_ is happening inside, it is not meaningful. Interface should be named `RateLimiter`, define the method `RateLimit`, and concrete implementation should be `IPRateLimiter` (see eg. in https://talks.golang.org/2014/names.slide#13).

Having several not-so-clear names in the same workflow (like "RateLimiterManager" and then "SetRateLimitMiddleware") undermines a bit the understanding of it.

PR #2709, opened 2019-09-13
update beats

jalvz:
I had to sync manually one dep which did not have the (right) origin set.


jalvz:
jenkins retest this please

script/update_beats.sh
```diff
@@ -28,6 +28,7 @@ rsync -crpv --delete \
     --exclude="dev-tools/jenkins_intake.sh" \
     --exclude="dev-tools/packaging/preference-pane/***" \
     --exclude="dev-tools/mage/***" \
+    --exclude="libbeat/processors/script/***" \
```
simitt: nit: would you mind moving this down to the `libbeat/processors` section for better readability?

jalvz: I think excludes need to go before includes to resolve conflicts properly, so it is better to group all the excludes at the top to avoid risking wrong order

PR #2711, opened 2019-09-16
Make timestamp for metricset optional.

### LINKED ISSUE #2700: [RUM] metricsets schema @timestamp should be created on server to avoid clock drift
+ When sending the metricsets events from the agents side, we use the current system clock time and send it along with `@timestamp` field in the metricset schema.
+ This value is susceptible to clock drift or users clock, Instead the server must be creating the value when the metricsets are sent from the agents. 

There are workarounds other agents are doing like https://github.com/watson/relative-microtime/blob/e91dbc903809d5b5e1228e6a28b84f0032871b9a/index.js#L4-L12 , Instead we would like this to be changed in the server atleast for the RUM agent. 

**NOTE: This is a blocker for RUM agent**

simitt:
fixes #2700

needs backport to `7.x` and `7.4`.


changelogs/7.4.asciidoc
```diff
@@ -28,3 +28,4 @@ https://github.com/elastic/apm-server/compare/v7.3.1\...v7.4.0[View commits]
 - Add support for host.name, adapt Intake API to accept detected and configured host information {pull}2531[2531], {pull}2540[2540].
 - Index `http.request.socket.remote_address` as `client.ip` {pull}2588[2588].
 - Upgrade Go to 1.12.9 {pull}2635[2635].
+- Make metricset timestamp optional {pull}2711[2711].
```
jalvz: you could even say this is a bug

simitt: Yeah I was in between on where to put it. However, so far all agents sending metricsets sent up the timestamp, so it is now loosening the requirements for a new integration with the RUM agent, and therefore I decided with not labeling it as a bug. If you feel strongly I am not against moving it to the bug section though.

Issue #2692, opened 2019-09-09
[docs] Switch to shared docs versioning


bmorelli25:
Currently we update the documentation version in `version.asciidoc` for each release. Example [here](https://github.com/elastic/apm-server/pull/2684/files#diff-bb1c954f8a5c98b5e555ebe570a0f4a1R1). It's not time consuming, but it is prone to errors.

With the addition of https://github.com/elastic/docs/pull/1147 and https://github.com/elastic/docs/pull/1148, there's now an easy way to let the docs repo (and the docs release manager) handle this. Lisa's already done this for the stack-docs repo in https://github.com/elastic/stack-docs/pull/493.

I think it makes sense to do the same for APM Server.

Relevant discussion in https://github.com/elastic/docs/issues/804.

---

Implementation:

- [x] Add `current` file for Agents -- https://github.com/elastic/docs/pull/1198
- [x] Add APM Server attribute to docs -- https://github.com/elastic/docs/pull/1187
- [x] Tell APM Server to use shared version file -- https://github.com/elastic/apm-server/pull/2741
- [x] Tell Agents to use "current" file (after 7.4) -- https://github.com/elastic/apm/issues/150
- [x] Remove Agent attributes from APM Server (after 7.4) -- https://github.com/elastic/apm-server/pull/2741
- [x] Update internal [documentation](https://github.com/elastic/apm-dev/blob/master/docs/apm-documentation.md#updating-an-apm-agent-to-a-new-version) 
- [ ] ~Clean up/remove deprecated `*-v` and hardcoded attributes in the docs repo -- Maybe?
https://github.com/elastic/docs/pull/1187/commits/3519dd934de87a1ba3c8649136b359f36762a788~

graphaelli:
Agreed, I don't see any reason not to do this.

bmorelli25:
This PR requires two steps:

Step 1 is to update the doc repo. Changes have been proposed in https://github.com/elastic/docs/pull/1187. I didn't realize it when I created this issue, but this change also allows us to finally stop linking to `current` versions of the APM Agent documentation. I've also added that change to this PR. Once merged, I'll start step 2 which involves changes in apm-server.

bmorelli25:
✅Done in the PRs linked in the description.

PR #2712, opened 2019-09-16
docs: fix versioning and attribute references

bmorelli25:
For #2692. 

Edit: Changing the scope of this PR. Unfortunately there is no solution for Agents yet. I've reverted those changes and will use this PR to solely add APM Server to the Elastic Stack's shared versioning files. Original description is below.

---

When I first joined Elastic, we had a difficult time linking APM Server/Overview documentation to APM Agent documentation because they are versioned differently.

At the time, the best solution was to create a new attribute, `*-v` (for "versioned"), for each of the Agent links. These links are then built with attributes that describe the current Agent version for that version of the Elastic stack:

https://github.com/elastic/apm-server/blob/862035f4edaeed0e9c6dc41eed4efbb2006f9a32/docs/version.asciidoc#L16-L33

This solution works, but has its limitations. For example, links from elsewhere in the stack (namely Kibana) could not take advantage of these versioned links.

~With recent changes in the documentation build, we can now move these links, and the APM Server versioning into the documentation repo. This simplifies our docs release process, and allows us to use versioned links across the APM stack.~

~This PR versions the `*-ref` attributes and removes the `*-v` attributes to maintain naming consistency across the stack.~

### DIFF
--- a/docs/version.asciidoc
+++ b/docs/version.asciidoc
@@ -1,33 +1,9 @@
-// stack-version can be:    7.0.1, 8.0.0-alpha1, 8.1.0, etc.
-// major-version can be:    8.x, 7.x, 6.x, etc.
-// doc-branch can be:       master, 8.0, 8.1, etc.
-// release-state is only:   released | prerelease | unreleased
 
-:stack-version: 8.0.0
-:major-version: 8.x
-:doc-branch: master
-:branch: {doc-branch}
+// Docs copied from Beats use stack-version and doc-branch instead of version
+:stack-version: {version}
+:doc-branch: {version}
+
 :go-version: 1.12.9
-:release-state: unreleased
 :python: 2.7.9
 :docker: 1.12
 :docker-compose: 1.11
-
-// APM Agent versions
-:server-branch: {doc-branch}
-:go-branch: 1.x
-:java-branch: 1.x
-:rum-branch: 4.x
-:node-branch: 2.x
-:py-branch: 5.x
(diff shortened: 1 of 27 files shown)


docs/version.asciidoc
```diff
@@ -1,33 +1,11 @@
-// stack-version can be:    7.0.1, 8.0.0-alpha1, 8.1.0, etc.
-// major-version can be:    8.x, 7.x, 6.x, etc.
-// doc-branch can be:       master, 8.0, 8.1, etc.
-// release-state is only:   released | prerelease | unreleased
+// Use shared version file
+include::{asciidoc-dir}/../../shared/versions/stack/{source_branch}.asciidoc[]
+
+// Docs copied from Beats use stack-version and doc-branch instead of version
+:stack-version: {version}
```
bmorelli25: These two lines need to be updated to `{apm_server_version}` before this is merged. I'm using `{version}` for now to make the docs ci point out broken links.

bmorelli25:

docs/version.asciidoc
bmorelli25: After a bit more digging. This needs to stay as-is right now. I can only update to `{apm_server_version}` after https://github.com/elastic/docs/pull/1187 is merged.

bmorelli25:
Standard build is green. Going to run a **rebuild** to make sure nothing slips through the cracks here.

@elasticmachine, run elasticsearch-ci/docs rebuild

PR #2716, opened 2019-09-17
[docs] make rum-allow-origins format clear

simitt:
There was some confusion around the _comma separated list_, and since in yml it would also be valid to specify in the following format, it makes sense to just describe _a list_.
```
allow_origins:
  - "foo"
  - "bar"
```
cc @nachogiljaldo

### DIFF
--- a/_meta/beat.yml
+++ b/_meta/beat.yml
@@ -89,7 +89,7 @@ apm-server:
 
     #-- General RUM settings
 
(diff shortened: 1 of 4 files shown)


simitt:
great point @bmorelli25 , updated it also.

simitt:

simitt:

PR #2732, opened 2019-09-18
Refactor rate limit logic

jalvz:
This PR tries to exemplify and materialize some observations I brought up in discussions during https://github.com/elastic/apm-server/pull/2706 and before:

* Avoid interface pollution: This removes one interface not needed. Doing so, the code is easier to navigate, and we test the actual code and not a mock.

* Follow Go idiomatic naming conventions: avoid non obvious acronyms and vague names like "managers" and such, which are more of a Java thing.
 
* Remove complex stateful functions: `handler.rateLimit` was making a rate limit call, plus returning a rate limiter if that was ok. There was also some redundant nil checking going on. Calling the limiter is now separated from getting one.

Incidentally, this is also less code to do the same thing and less dependencies per package (imports), but I still need to double check that I didn't change behaviour involuntarily...

### DIFF
--- a/beater/api/intake/handler_test.go
+++ b/beater/api/intake/handler_test.go
@@ -30,7 +30,6 @@ import (
 	"github.com/pkg/errors"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
-	"golang.org/x/time/rate"
 
 	"github.com/elastic/apm-server/beater/beatertest"
 	"github.com/elastic/apm-server/beater/config"
@@ -44,6 +43,8 @@ import (
 	"github.com/elastic/apm-server/tests/loader"
 	"github.com/elastic/apm-server/transform"
 )
+
+var rateLimit, _ = ratelimit.NewStore(1, 0, 0)
 
 func TestIntakeHandler(t *testing.T) {
 
@@ -63,9 +64,9 @@ func TestIntakeHandler(t *testing.T) {
 			code: http.StatusBadRequest, id: request.IDResponseErrorsValidate,
 		},
 		"RateLimit": {
-			path: "errors.ndjson",
-			rlc:  &mockBlockingRateLimiter{},
-			code: http.StatusTooManyRequests, id: request.IDResponseErrorsRateLimit,
+			path:      "errors.ndjson",
+			rateLimit: rateLimit,
+			code:      http.StatusTooManyRequests, id: request.IDResponseErrorsRateLimit,
 		},
 		"BodyReader": {
 			path: "errors.ndjson",
@@ -141,8 +142,8 @@ func TestIntakeHandler(t *testing.T) {
 		// setup
 		tc.setup(t)
 
-		if tc.rlc != nil {
-			tc.c.RateLimitManager = tc.rlc
+		if tc.rateLimit != nil {
+			tc.c.RateLimiter = tc.rateLimit
 		}
 		// call handler
 		h := Handler(tc.dec, tc.processor, tc.reporter)
@@ -173,7 +174,7 @@ type testcaseIntakeHandler struct {
 	r         *http.Request
 	dec       decoder.ReqDecoder
 	processor *stream.Processor
-	rlc       ratelimit.Manager
+	rateLimit *ratelimit.Store
 	reporter  func(ctx context.Context, p publish.PendingReq) error
 	path      string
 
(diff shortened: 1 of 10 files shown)


jalvz:
Added some more things:

* A test with rate limits per IP
* Consolidate middleware

Middlewares should be applied always in the same order. This is hard to maintain if we have one middleware function per handler. Instead we should compose them more (like done with `apmMiddleware`). 
Doing this, adding now a rum handler for the config endpoint is literally 4 lines of code.

Changes:
`RequestTimeMiddleware` is now added to all the handlers.
`KillSwitchMiddleware` always runs the last.

beater/api/mux.go
```diff
@@ -155,29 +133,49 @@ func sourcemapHandler(cfg *config.Config, reporter publish.Reporter) (request.Ha
...
-func rootHandler(cfg *config.Config, _ publish.Reporter) (request.Handler, error) {
-	return middleware.Wrap(root.Handler(), rootMiddleware(cfg)...)
+func rumMiddleware(cfg *config.Config, m map[request.ResultID]*monitoring.Int) []middleware.Middleware {
+	return append(apmMiddleware(m),
+		middleware.SetRateLimitMiddleware(cfg.RumConfig.EventRate),
+		middleware.CORSMiddleware(cfg.RumConfig.AllowOrigins),
+		middleware.KillSwitchMiddleware(cfg.RumConfig.IsEnabled()))
+}
+
+func sourcemapMiddleware(cfg *config.Config) []middleware.Middleware {
```
simitt: Why not also passing in the monitoring map here?

beater/api/intake/handler_test.go
```diff
@@ -63,9 +64,9 @@ func TestIntakeHandler(t *testing.T) {
 			code: http.StatusBadRequest, id: request.IDResponseErrorsValidate,
 		},
 		"RateLimit": {
-			path: "errors.ndjson",
-			rlc:  &mockBlockingRateLimiter{},
-			code: http.StatusTooManyRequests, id: request.IDResponseErrorsRateLimit,
+			path:      "errors.ndjson",
+			rateLimit: rateLimit,
```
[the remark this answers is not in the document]
jalvz: i can't do that because `NewStore` returns 2 values, but i can move it inside the function

beater/api/mux.go
jalvz: because it is not needed

beater/api/mux.go
```diff
@@ -47,7 +47,7 @@ const (
 	RootPath = "/"
 
 	// ConfigAgent defines the path to query for agent config management
-	ConfigAgent = "/config/v1/agents"
+	AgentConfigPath = "/config/v1/agents"
```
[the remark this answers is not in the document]
jalvz: i only changed it because of readability, we always call this "agent config" and not "config agent" (which doesn't read very naturally)... the actual string value is not so important

beater/api/intake/handler_test.go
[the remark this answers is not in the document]
simitt: good enough

As a side note, you could work around that by
```
rateLimit: func() *ratelimit.Store { 
  rateLimit, _ = ratelimit.NewStore(1, 0, 0)
  return rateLimit
}()
```

beater/api/mux.go
```diff
@@ -155,29 +133,49 @@ func sourcemapHandler(cfg *config.Config, reporter publish.Reporter) (request.Ha
...
 		middleware.RequireAuthorizationMiddleware(cfg.SecretToken))
 }
 
-func rootHandler(cfg *config.Config, _ publish.Reporter) (request.Handler, error) {
-	return middleware.Wrap(root.Handler(), rootMiddleware(cfg)...)
+func rumMiddleware(cfg *config.Config, m map[request.ResultID]*monitoring.Int) []middleware.Middleware {
+	return append(apmMiddleware(m),
+		middleware.SetRateLimitMiddleware(cfg.RumConfig.EventRate),
+		middleware.CORSMiddleware(cfg.RumConfig.AllowOrigins),
+		middleware.KillSwitchMiddleware(cfg.RumConfig.IsEnabled()))
```
simitt: Actually IMO the `KillSwitchHandler` should be the first middleware to run. Why do you run it at the end and waste resources when the endpoint is disabled?

jalvz: Mostly because of convenience, and because it is easy to be consistent about it. I don't expect anyone hammering a disabled endpoint and wasting resources, more likely someone will notice it after a one or few queries and enable it.

simitt: > Mostly because of convenience, and because it is easy to be consistent about it. 

I don't see the inconvenience or inconsistency if the order is changed. 

In case of an attack it would waste resources though, so I suggest to change the order.

PR #2746, opened 2019-09-30
Add service.node.configured_name to Intake API and transform into service.node.name for ES

### LINKED ISSUE #2696: Add new service field to Intake API + ES
A new field `service.node.configured_name` will be sent from the agents. The server needs to add it to the Intake API as part of the metadata. 

The server then needs to populate the new ECS field `service.node.name` with following logic (copied from https://github.com/elastic/apm/issues/141#issue-488124341):

>     If `service.node.configured_name` is set - use it
>     Otherwise, if `system.container.id` is set - use it
>     Otherwise, use host name (apply same logic applied today to `host.name`)

(edited 2019/09/16: move logic how to derive value of `service.node.name` also to apm server)

Detailed information in https://github.com/elastic/apm/issues/141.

Things to clarify:

* [x] Exact field name for Intake API, see https://github.com/elastic/apm/issues/141#issuecomment-530280227
will be `service.node.configured_name`
* [x] wait for https://github.com/elastic/ecs/issues/538 to settle on the ES field
settled on `service.node.name`
* [x] `service.name` only allows values following the pattern `^[a-zA-Z0-9 _-]+$`, do we need to apply the same restrictions to the new field?
not necessary here, as the concerns where for cases where the service name should be part of the index, which is not discussed for this field
* [x] As far as I understand the issue, the information should be used for aggregations, so indexing will be necessary. I suggest to index as a `keyword` with a length limit of 1024 on the Intake API. @eyalkoren do you agree?
see https://github.com/elastic/apm-server/issues/2696#issuecomment-531717630

simitt:
Add service.node.configured_name to Intake API and add support for
retrieving and indexing service.node.name in ES.

fixes #2696

- [x] add changelog entries
- [x] resolve blocking questions https://github.com/elastic/apm-server/issues/2696#issuecomment-536558682


model/metadata/metadata.go
```diff
@@ -99,3 +102,10 @@ func (m *Metadata) SetMinimal(fields common.MapStr) common.MapStr {
 	utility.Set(fields, "service", m.Service.MinimalFields())
 	return fields
 }
+
+func fetch(m common.MapStr, key string) string {
```
jalvz: `fetch` is usually associated with a remote call, this is just a get / lookup. and it could go in `utility` pkg

simitt: Changed the name; I prefer to keep it local and unexported, we can revisit once this is needed more often.

Issue #2646, opened 2019-08-27
Provide Access-Control-Allow-Origin header on the response to RUM ACM endpoint


jahtalab:
Requests made to the /config/v1/agents by the RUM agent will fail if the `Access-Control-Allow-Origin` header is not present on the response.

The value of this header should be generated using the same algorithm used for the `/intake/v2/rum/events` endpoint.

jahtalab:
@elastic/apm-server , Is it possibility to provide the CORS headers on all endpoints by default?

jalvz:
On all endpoints? What is the use case eg for `/intake/v2/events` ?

jahtalab:
I'm basically looking for a solution to make sure we don't forget about CORS headers for the endpoints that are going to be used by the RUM agent. 

We can consider other solutions, if you think having the CORS headers for all endpoints has 
drawbacks. Either way, I don't have a strong opinion on the solution itself but I think we should find a way to make sure these headers are added for all RUM endpoints in the future.

jalvz:
got it, thanks!

Issue #2630, opened 2019-08-23
[ACM] Provide user feedback to Kibana


jalvz:
Agent configuration should indicate whether it was applied remotely or not.

This entails removing the logic around `Etag` and serialization of Kibana responses. Instead, Kibana will calculate `Etags` and apm-server will pass it down to the agents. apm-server will also forward the `etag` sent from the agents to Kibana, so it can know if the agent is up to date or not. 

Kibana ticket: https://github.com/elastic/kibana/issues/43354

PR #2747, opened 2019-09-30
agent config GA

### LINKED ISSUE #2694: [ACM] Implement RUM endpoint
The RUM endpoint will provide the same functionality as as the backend endpoint with following differences:

- No secret token.
- Rate limit.
- CORS handler.

The existing rate limit and allowed origin settings used for the intake endpoint will be applied here.

___

As pointed out by @graphaelli , this endpoint could be abused to expose backend configuration data. To mitigate that, the UI will include a `rum-enabled` checkbox for users to confirm that a given service is instrumented by the RUM agent. 
apm-server will receive that value from Kibana as a boolean, and it will only return configuration if it is true.
____

Currently backend agents derive their polling interval from the `Access-Control-Max-Age` header. We can have a different value for the RUM endpoint.

jalvz:
fixes #2694, fixes elastic/apm-server#2630 , fixes elastic/apm-server#2646 

depends on https://github.com/elastic/kibana/pull/46995

required by https://github.com/elastic/apm-agent-rum-js/pull/439

### DIFF
--- a/agentcfg/cache.go
+++ b/agentcfg/cache.go
@@ -26,58 +26,36 @@ import (
 )
 
 const (
-	cleanupInterval time.Duration = 60 * time.Second
+	cleanupInterval = 60 * time.Second
 )
 
 type cache struct {
 	logger  *logp.Logger
-	exp     time.Duration
 	gocache *gocache.Cache
 }
 
 func newCache(logger *logp.Logger, exp time.Duration) *cache {
-	if logger == nil {
-		logger = logp.NewLogger("agentcfg")
-	}
-	logger.Infof("Cache creation with default expiration %v.", exp)
+	logger.Infof("Cache creation with expiration %v.", exp)
 	return &cache{
 		logger:  logger,
-		exp:     exp,
 		gocache: gocache.New(exp, cleanupInterval)}
 }
 
-func (c *cache) fetchAndAdd(q Query, fn func(Query) (*Doc, error)) (doc *Doc, err error) {
-	id := q.ID()
-
+func (c *cache) fetch(query Query, fetch func() (Result, error)) (Result, error) {
 	// return from cache if possible
-	doc, found := c.fetch(id)
-	if found {
-		return
+	value, found := c.gocache.Get(query.id())
+	if found && value != nil {
+		return value.(Result), nil
 	}
-
-	// call fn to retrieve resource from external source
-	doc, err = fn(q)
+	// retrieve resource from external source
+	result, err := fetch()
 	if err != nil {
-		return
+		return result, err
 	}
+	c.gocache.SetDefault(query.id(), result)
 
-	// add resource to cache
-	c.add(id, doc)
-	return
-}
-
-func (c *cache) add(id string, doc *Doc) {
-	c.gocache.SetDefault(id, doc)
-	if !c.logger.IsDebug() {
-		return
+	if c.logger.IsDebug() {
+		c.logger.Debugf("Cache size %v. Added ID %v.", c.gocache.ItemCount(), query.id())
 	}
-	c.logger.Debugf("Cache size %v. Added ID %v.", c.gocache.ItemCount(), id)
-}
-
-func (c *cache) fetch(id string) (*Doc, bool) {
-	val, found := c.gocache.Get(id)
-	if !found || val == nil {
-		return nil, found
-	}
-	return val.(*Doc), found
+	return result, nil
 }
--- a/agentcfg/fetch.go
+++ b/agentcfg/fetch.go
@@ -23,6 +23,8 @@ import (
 	"net/http"
 	"time"
 
+	"github.com/elastic/apm-server/utility"
+
 	"github.com/pkg/errors"
 
 	"github.com/elastic/beats/libbeat/logp"
@@ -34,7 +36,6 @@ import (
 // Error Messages used to signal fetching errors
 const (
 	ErrMsgSendToKibanaFailed = "sending request to kibana failed"
-	ErrMsgMultipleChoices    = "multiple configurations found"
 	ErrMsgReadKibanaResponse = "unable to read Kibana response body"
 )
 const endpoint = "/api/apm/settings/agent-configuration/search"
@@ -42,45 +43,32 @@ const endpoint = "/api/apm/settings/agent-configuration/search"
 // Fetcher holds static information and information shared between requests.
 // It implements the Fetch method to retrieve agent configuration information.
 type Fetcher struct {
-	docCache *cache
-	logger   *logp.Logger
-	kbClient kibana.Client
+	*cache
+	logger *logp.Logger
+	client kibana.Client
 }
 
 // NewFetcher returns a Fetcher instance.
-func NewFetcher(kbClient kibana.Client, cacheExp time.Duration) *Fetcher {
+func NewFetcher(client kibana.Client, cacheExpiration time.Duration) *Fetcher {
 	logger := logp.NewLogger("agentcfg")
 	return &Fetcher{
-		kbClient: kbClient,
-		logger:   logger,
-		docCache: newCache(logger, cacheExp),
+		client: client,
+		logger: logger,
+		cache:  newCache(logger, cacheExpiration),
 	}
 }
 
 // Fetch retrieves agent configuration, fetched from Kibana or a local temporary cache.
-func (f *Fetcher) Fetch(q Query, err error) (map[string]string, string, error) {
-	req := func(query Query) (*Doc, error) {
-		resultBytes, err := f.request(convert.ToReader(query), err)
-		if err != nil {
-			return nil, err
-		}
-		return NewDoc(resultBytes)
+func (f *Fetcher) Fetch(query Query) (Result, error) {
+	req := func() (Result, error) {
+		return newResult(f.request(convert.ToReader(query)))
 	}
-
-	doc, err := f.docCache.fetchAndAdd(q, req)
-	if err != nil {
-		return nil, "", err
-	}
-
-	return doc.Settings, doc.ID, nil
+	result, err := f.fetch(query, req)
+	return sanitize(query.IsRum, result), err
 }
 
-func (f *Fetcher) request(r io.Reader, err error) ([]byte, error) {
-	if err != nil {
-		return nil, err
-	}
-
-	resp, err := f.kbClient.Send(http.MethodPost, endpoint, nil, nil, r)
+func (f *Fetcher) request(r io.Reader) ([]byte, error) {
+	resp, err := f.client.Send(http.MethodPost, endpoint, nil, nil, r)
 	if err != nil {
 		return nil, errors.Wrap(err, ErrMsgSendToKibanaFailed)
 	}
@@ -91,13 +79,28 @@ func (f *Fetcher) request(r io.Reader, err error) ([]byte, error) {
 	}
 
 	result, err := ioutil.ReadAll(resp.Body)
-	if resp.StatusCode == http.StatusMultipleChoices {
-		return nil, errors.Wrap(errors.New(string(result)), ErrMsgMultipleChoices)
-	} else if resp.StatusCode > http.StatusMultipleChoices {
+	if resp.StatusCode >= http.StatusBadRequest {
 		return nil, errors.New(string(result))
 	}
 	if err != nil {
 		return nil, errors.Wrap(err, ErrMsgReadKibanaResponse)
 	}
 	return result, nil
+}
+
+func sanitize(isRum bool, result Result) Result {
+	if !isRum {
+		return result
+	}
+	hasRumData := utility.Contains(result.Source.Agent, RumAgent) || result.Source.Agent == ""
+	if !hasRumData {
+		return zeroResult()
+	}
+	settings := Settings{}
+	for k, v := range result.Source.Settings {
+		if utility.Contains(k, RumSettings) {
+			settings[k] = v
+		}
+	}
+	return Result{Source: Source{Etag: result.Source.Etag, Settings: settings}}
 }
--- a/agentcfg/model.go
+++ b/agentcfg/model.go
@@ -18,88 +18,81 @@
 package agentcfg
 
 import (
-	"crypto/md5"
 	"encoding/json"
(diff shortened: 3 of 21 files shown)


jalvz:
failing test can't be fixed until Kibana side is merged, as Kibana returns `400` upon unknown attributes (in this case `etag`)
cc @sqren

jalvz:
depends on https://github.com/elastic/kibana/pull/46995

agentcfg/cache.go
```diff
@@ -31,53 +33,38 @@ const (
...
-	c.add(id, doc)
-	return
-}
-
-func (c *cache) add(id string, doc *Doc) {
-	c.gocache.SetDefault(id, doc)
-	if !c.logger.IsDebug() {
-		return
+	if !authorized(q.IsRum, result.Source.Agent) {
+		return ZeroResult(), nil
```
simitt: With this implementation non authorized RUM results are not cached. IMO this should be changed.

agentcfg/fetch.go
```diff
@@ -91,9 +77,7 @@ func (f *Fetcher) request(r io.Reader, err error) ([]byte, error) {
 	}
 
 	result, err := ioutil.ReadAll(resp.Body)
-	if resp.StatusCode == http.StatusMultipleChoices {
```
simitt: Did this change in Kibana so the check is not necessary anymore?

beater/api/mux.go
```diff
@@ -167,6 +177,7 @@ func backendMiddleware(cfg *config.Config, m map[request.ResultID]*monitoring.In
 
 func rumMiddleware(cfg *config.Config, m map[request.ResultID]*monitoring.Int) []middleware.Middleware {
 	return append(apmMiddleware(m),
+		middleware.SetRumFlagMiddleware(),
 		middleware.SetRateLimitMiddleware(cfg.RumConfig.EventRate),
 		middleware.CORSMiddleware(cfg.RumConfig.AllowOrigins),
 		middleware.KillSwitchMiddleware(cfg.RumConfig.IsEnabled()))
```
simitt: It still bothers me that all of these middleware functions are run if someone has RUM disabled; as a user I'd expect the `KillswitchMiddleware` to be run before the CORS and rate limiter middleware at least.

beater/middleware/rate_limit_middleware.go
```diff
@@ -31,7 +31,7 @@ func SetRateLimitMiddleware(cfg *config.EventRate) Middleware {
 
 	return func(h request.Handler) (request.Handler, error) {
 		return func(c *request.Context) {
-			c.RateLimiter = store
+			c.RateLimiter = store.ForIP(c.Request)
```
simitt: Can you change the name to include `IP` now.

agentcfg/model.go
```diff
@@ -18,88 +18,79 @@
...
-		default:
-			out[localkey] = fmt.Sprintf("%+v", val)
-			h.Write([]byte(fmt.Sprintf("%s_%v", localkey, val)))
-		}
+// UnmarshalJSON overrides default method to convert any JSON type to string
+func (s Settings) UnmarshalJSON(b []byte) error {
+	in := make(map[string]interface{})
+	err := json.Unmarshal(b, &in)
+	for k, v := range in {
+		s[k] = fmt.Sprintf("%v", v)
```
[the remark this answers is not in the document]
jalvz: because it is not needed, we don't need a hash of the contents anymore

beater/middleware/rate_limit_middleware.go
jalvz: ha! i was about to do that, but then i thought you will reply with something like this: https://github.com/elastic/apm-server/pull/2706#discussion_r324647299

not sure what is the criteria now, but will change as I agree

beater/api/mux.go
jalvz: i actually think we should do the CORS dance properly and return to the browsers the response they expect, otherwise the browser might show an error in the lines of "domain not allowed" when the root cause is other.

agentcfg/fetch.go
jalvz: i don't think it was ever needed?

agentcfg/cache.go
jalvz: I don't think so. If then the same request comes trough the right endpoint, the result would be a wrong empty result.

beater/middleware/rate_limit_middleware.go
simitt: I don't follow. The logic that the rate limiter limits by IP is now _in the middleware_. It was not before. So now it makes sense to reflect it in the middleware name.

agentcfg/fetch.go
simitt: I think it was there from the beginning, so I expected Kibana could respond with it. If not needed, fine to remove.

agentcfg/cache.go
simitt: Kibana is responding the same way to this query, no matter who requests it, right? So I don't see why caching the `result` here would lead to wrong results.
Ofc I did not mean to cache the `ZeroResult` but the `result`.

jalvz: in that case i don't know why i would want to cache something that i am not even returning, but im fine either way

agentcfg/fetch.go
jalvz: i don't recall either. the only thing i can think of is that the code is 300, so a check like `code < 300` tells you if a response is successful.

agentcfg/cache.go
simitt: My reasoning is the following:
What is the purpose of caching here? 
To avoid unnecessary calls to Kibana based on legitimate requests from agents. But as soon as those requests can be triggered from requests to unsecured RUM endpoint, also to prevent overloading Kibana on any kind of attacks.
The `authorized` check would only return `false` if the agents had a bug or for malicious requests. In the case of malicious requests it makes sense to also avoid sending the same requests over and over to Kibana. 

However, if you add it to the cache, you would probably also need to size limit the cache, as otherwise you open up another attack vector.

I don't see why the decision in which cases something should be added to the cache is in direct relation on whether or not you are returning this information.

simitt: Not sure you find this useful, but I gave this some thought when introducing the caching https://github.com/elastic/apm-server/issues/2220#issuecomment-504935447.

beater/middleware/rate_limit_middleware.go
jalvz: As clarified offline, this is a comment on L29, not L34. 
Changed in 07888478290b0d9154e92a80c6e823b2b1b002b1

jalvz:
> You explained that GA changes will be breaking. Does that mean that the agents using the beta implementation would break when talking to APM Server or would they just stop applying ACM?


they will just not get any config updates, so they won't apply it

beater/api/mux.go
simitt: I lack knowledge about what is the standard behavior related to CORS and browsers. If the advantage is worth the potential additional costs I'm ok with it.

agentcfg/cache.go
jalvz: Thanks. 

I am changing the caching key to be service name + env, instead of Etag.
When there are many RUM agents, they will not send an `ifnonematch` arg until they get one from a previous call, which means that all of them would have to hit Kibana at least one.
Caching the service name solves this problem. 

On the other side, I don't think is necessary to have a cache limit.

jalvz: cd40ddbdc4af56a6cf941c9a377cccb5c3f05ea2

jalvz:
> You call sanitize before adding the value to the cache

Not true, as far I can see.

> The authorized check occurs only for non-cached information

Right, will update.

> the cache can potentially grow infinite with requests for non-existing services

Lets talk about that offline

simitt:
>>You call sanitize before adding the value to the cache

> Not true, as far I can see.

Within the `cache.fetch` you call [`fetch()`](https://github.com/elastic/apm-server/pull/2747/files#diff-1629cb97d00ee13644214527e7c86c77R53), which leads to running the [`sanitize`](https://github.com/elastic/apm-server/pull/2747/files#diff-dd9441da5960be4ce0008ddc0da10d23R65). The returned, sanitized result is then [added to the cache](https://github.com/elastic/apm-server/pull/2747/files#diff-1629cb97d00ee13644214527e7c86c77R57).

jalvz:
This test https://github.com/elastic/apm-server/pull/2747/commits/8bbae34278729c722176c9e0620868ca0df7181f#diff-6756162984ecc44fbead19b809440eaeR305 covers that scenario and passes locally, I have to see what's up

jalvz:
it always worked because each handler has its own cache, added another test completeness.
pushed another change because i didn't like important logic buried in a cache method any ways

tests/system/test_integration_acm.py
```diff
@@ -324,6 +324,28 @@ def test_backend_not_ok(self):
         assert r2.status_code == 200, r2.status_code
         assert r2.json() == {"transaction_sample_rate": "0.3"}, r2.json()
 
+    def test_rum_after_backend(self):
```
simitt: nit: the previous issue would only be uncovered if the order is changed `test_backend_after_rum`.

jalvz: that's also exercised here if you are referring to that https://github.com/elastic/apm-server/pull/2747/files#diff-6756162984ecc44fbead19b809440eaeR313-R325

agentcfg/fetch.go
```diff
@@ -91,13 +81,28 @@ func (f *Fetcher) request(r io.Reader, err error) ([]byte, error) {
...
+	hasRumData := utility.Contains(result.Source.Agent, RumAgent) || result.Source.Agent == ""
+	if !hasRumData {
+		return zeroResult()
+	}
+	settings := Settings{}
+	for k, v := range result.Source.Settings {
+		if utility.Contains(k, RumSettings) {
+			settings[k] = v
+		}
+	}
```
simitt: nit: why not deleting settings that should not be contained, instead of creating a new settings object?

agentcfg/model.go
```diff
@@ -18,88 +18,81 @@
...
+	Etag    string  `json:"etag"`
+	IsRum   bool    `json:"-"`
+}
 
-	h := md5.New()
-	var out = map[string]string{}
-	if err := parse(settings, out, "", h); err != nil {
-		return nil, err
-	}
+func (q Query) toString() string {
```
simitt: I'd prefer calling this `ID` as that's better describing what it does, and `toString` is usually used for a nicer presentation of an instance.

tests/system/test_integration_acm.py
```diff
@@ -9,13 +9,17 @@
...
+class AgentConfigurationTest(ElasticTest):
     config_overrides = {
         "logging_json": "true",
         "kibana_enabled": "true",
         "acm_cache_expiration": "1s",
     }
 
+    @classmethod
+    def setUpClass(cls):
+        super(AgentConfigurationTest, cls).setUpClass()
```
simitt: Why is this necessary?

tests/system/test_integration_acm.py
```diff
@@ -43,7 +49,10 @@ def create_service_config(self, settings, name, env=None, _id="new"):
         )
 
     def update_service_config(self, settings, _id, name, env=None):
-        return self.create_service_config(settings, name, env, _id=_id)
+        return self.create_service_config(settings, name, agent="python", env=env, _id=_id)
```
simitt: nit: No need for setting `agent="python"`, it reflects the default.

tests/system/test_integration_acm.py
```diff
@@ -254,3 +274,91 @@ def test_log_kill_switch_active(self):
...
+
+
+class RumAgentConfigurationIntegrationTest(AgentConfigurationTest):
+    config_overrides = {
+        "enable_rum": "true",
+    }
+
+    def test_rum(self):
+        service_name = "rum-service"
+        create_config_rsp = self.create_service_config({"transaction_sample_rate": 0.2}, service_name, agent="rum-js")
```
simitt: following up on your abbreviation preferences, don't use `rsp` as abbrev for `response` (I'd be totally fine with `resp` as that's widely used).

tests/system/test_integration_acm.py
```diff
@@ -254,3 +274,91 @@ def test_log_kill_switch_active(self):
...
+        r1 = requests.get(self.rum_agent_config_url,
+                          params={"service.name": service_name},
+                          headers={"Content-Type": "application/json"})
+
+        assert r1.status_code == 200
+        assert r1.json() == {'transaction_sample_rate': '0.2'}
+        etag = r1.headers["Etag"]
+
+        r2 = requests.get(self.rum_agent_config_url,
+                          params={"service.name": service_name, "ifnonematch": etag.replace('"', '')},
```
simitt: why is it necessary to remove surrounding `"` of etag? It should be possible to just use the etag that is returned by the server.

jalvz: because the rum agent will actually send it without double quotes

agentcfg/fetch.go
jalvz: because i don't want to modify the argument

tests/system/test_integration_acm.py
jalvz: yes, but this is copied over and the same abbreviation is all over the place. i didn't suggest to remove all abbreviations from the code base in one go.

tests/system/test_integration_acm.py
simitt: But shouldn't the server be able to process with or without quotes? And why is the RUM agent sending without quotes (I probably missed that discussion somewhere).

simitt: I approved the PR, as I haven't followed discussions about this. Please ensure then RUM agent properly removes the quotes from the response they get if you don't change it in the server.

jalvz:
Revisiting this question:

> You explained that GA changes will be breaking. Does that mean that the agents using the beta implementation would break when talking to APM Server or would they just stop applying ACM?


What do you actually meant with "break"? 

I bumped the min version from 7.3 to 7.5, so the agents would get an error response code, and the should handle it, therefore, not applying any config

PR #2764, opened 2019-10-03
Make ILM policies configurable.

### LINKED ISSUE #2240: [ILM] Make policies more flexible

simitt:
* ILM policies can now be configured in the `apm-server.yml`. 
* Supports configuring policy names referring to policies created outside the APM Server.
* Shows default policies in apm-server.yml for guidance of users when creating their own policies.

fixes elastic/apm-server#2240

UPDATE 11/10/2019:
This PR cahnges how ILM and template setup behaves for some edge cases. There is a new setting group introcuded `apm-server.ilm.setup.*` which isolates everything ILM related from `setup.template.*` settings. So far `setup.template.overwrite` was necessary to overwrite event type specific templates, to set and unset ILM for APM indices.

Detailled explanation in the [comment](https://github.com/elastic/apm-server/pull/2764#issuecomment-54109018)


simitt:
Not sure why CI shows red, if you go on details, the tests have passed..

_meta/beat.yml
```diff
@@ -247,6 +239,82 @@ apm-server:
     # Configure curve types for ECDHE based cipher suites.
     #ssl.curve_types: []
 
+  #---------------------------- APM Server - ILM Index Lifecycle Management ----------------------------
+
+  # Index Lifecycle Mangement (ILM)
+  #ilm:
+    # `ilm.enabled: "auto"`:
```
jalvz:
```suggestion
    # `enabled: "auto"`
```
same for all others bellow: remove `ilm` duplicated and extra `:` at the end

idxmgmt/ilm/config.go
```diff
@@ -18,34 +18,117 @@
...
+			return errPolicyFmt
+		}
+		(*p)[k] = policy(inpP)
+	}
+	return nil
+}
+
+//prepare ensures maps are in the format elasticsearch expects for policy bodies,
+//it replaces nil values with an empty map
+func prepare(bb map[string]interface{}) map[string]interface{} {
```
jalvz: I struggle to understand the need for this. Can't you do all the error handling you want in the `Unpack`?
As a matter of fact, how do I craft an `apm-server.yml` so that the result is different with and without the `prepare` function?

simitt: When defining something like `delete: {}` in the yml file, the way it is resolved in code is `delete: nil`. This is not an accepted format by Elasticsearch, and does not reflect what is configured, so I need to ensure it is properly mapped to have an empty hash instead of nil. 
I followed your suggestion and moved it into the `Unpack` method.

simitt:
@jalvz I changed the behavior to consider `apm-server.ilm.overwrite` AND `setup.template.overwrite` when deciding whether or not the ilm templates should be overwritten. 

Now that I introduced `apm-server.ilm.overwrite` it would be cleaner to keep the `setup.template.*` settings completely separate from ILM, but I consider this a breaking change, so I kept the existing logic in.

idxmgmt/manager_test.go
```diff
@@ -157,41 +158,59 @@ func TestManager_Setup(t *testing.T) {
...
+		},
+		"OverwriteTemplate": {
+			cfg:   common.MapStr{"setup.template.overwrite": true},
+			tLoad: libidxmgmt.LoadModeEnabled, ilmLoad: libidxmgmt.LoadModeEnabled,
+			tCt: 5, tWithILMCt: 4, pCt: 2, aCt: 3,
+		},
+		"OverwriteTemplateAndILM": {
+			cfg:   common.MapStr{"setup.template.overwrite": true, "apm-server.ilm.overwrite": true},
+			tLoad: libidxmgmt.LoadModeEnabled, ilmLoad: libidxmgmt.LoadModeEnabled,
+			tCt: 5, tWithILMCt: 4, pCt: 4, aCt: 3,
```
jalvz: I really would like to start using abbreviations as an exception rather than as a rule, here and everywhere else. I don't know what is `tCt`, `pCt`, `aCt`, and such. 
Also, this part of the code is quite complicated already, things eg like `supporter.st`, `esIdxCfg` elsewhere don't make it any easier to follow.

idxmgmt/manager_test.go
```diff
@@ -405,7 +453,8 @@ func (h *mockClientHandler) CreateAlias(alias libilm.Alias) error {
 }
 
 func (h *mockClientHandler) HasILMPolicy(name string) (bool, error) {
-	return name == existingILM, nil
+	fmt.Println(name)
```
jalvz: letfover

idxmgmt/manager.go
```diff
@@ -134,8 +135,17 @@ func (m *manager) templateFeature(loadMode libidxmgmt.LoadMode) feature {
...
+	// Do not use configured `m.supporter.ilmConfig.Mode` to check if ilm is enabled.
+	// The configuration might be set to `true` or `auto` but preconditions are not met,
+	// e.g. ilm support by Elasticsearch
+	// In these cases the supporter holds an internal state `m.supporter.st.ilmEnabled` that is set to false.
+	// The originally configured value is preserved allowing to collect warnings and errors to be
+	// returned to the user.
+
+	// m.supporter.st.ilmEnabled.Load() only returns true for cases where
+	// ilm mode is configured `auto` or `true` and preconditions to enable ilm are true
+	if enabled := m.supporter.st.ilmEnabled.Load(); enabled {
```
jalvz: if this returns a bool, shouldn't be named `Loaded()` or `IsLoaded()`?

idxmgmt/ilm/policy.go
```diff
@@ -17,69 +17,75 @@
 
 package ilm
 
-import "github.com/elastic/beats/libbeat/common"
+type m map[string]interface{}
```
jalvz: We should start considering something like https://github.com/elastic/hey-apm/blob/master/types/types.go, instead of repeating this all over the place

idxmgmt/ilm/config.go
```diff
@@ -18,34 +18,117 @@
...
+		}
+	}
+
+	for event, policyName := range policyMappings {
+		policy, ok := policies[policyName]
+		if !ok {
+			if c.RequirePolicy {
+				return Config{}, errors.Errorf("policy '%s' not configured for ILM setup", policyName)
+			}
+			policy = nil
```
jalvz: isn't this redundant?

idxmgmt/ilm/config.go
```diff
@@ -18,34 +18,117 @@
...
+		return errPolicyFmt
+	}
+	*p = policies{}
+
+	for k, v := range inp {
+		inpP, ok := v.(map[string]interface{})
+		if !ok {
+			return errPolicyFmt
+		}
+		(*p)[k] = prepare(policy(inpP))
```
jalvz:
```suggestion
		(*p)[k] = prepare(inpP)
```

_meta/beat.yml
```diff
@@ -137,61 +183,7 @@ apm-server:
...
-  # If the configured Elasticsearch instance is not eligible for ILM, ILM will also be disabled.
-  # If all preconditions are met, ILM will be enabled.
-  #
-  # When ILM is set to `true`, the APM Server ignores any configured index settings.
-  # For ILM to be applied, The configured output must be set to Elasticsearch and the instance
-  # needs to support ILM. Otherwise APM Server falls back to ordinary index management without ILM.
-  #
-  # Defaults to "auto". Disable ILM by setting it to `false`.
-  #ilm.enabled: "auto"
+  #---------------------------- APM Server - Agent Configuration ----------------------------
```
jalvz: Did you think how are we going to maintain these headers when configs can serve multiple purposes? eg, if we add another feature requiring Kibana, how this header is going to be named?

_meta/beat.yml
```diff
@@ -247,6 +239,82 @@ apm-server:
...
+      #  policy: "rollover-1-day"
+      #- event_type: "transaction"
+      #  policy: "rollover-7-days"
+      #- event_type: "metric"
+      #  policy: "rollover-7-days"
+
+    # Configured policies are added to all pre-defined default policies.
+    # If a policy with the same name as a default policy is configured, the configured policy overwrites the default policy.
+    #policies:
+      #rollover-7-days:
```
jalvz: mixing values as keys in config dictionaries is always quite anti-intuitive to me, 
and we don't gain much compared with traditional `"name": "rollover-7-days"` approach. 
Elasticsearch does this a lot as well so i guess its ok... but I believe this is the only case of this in our own configuration, which makes it a bit awkward and not consistent. For instance, `indices` config reads like
```
  - index: "apm-%{[observer.version]}-error-%{+yyyy.MM.dd}"
      when.contains:
        processor.event: "error"
```

and not like

```
- "apm-%{[observer.version]}-error-%{+yyyy.MM.dd}"
      when.contains:
        processor.event: "error"
```

idxmgmt/ilm/config.go
```diff
@@ -18,34 +18,117 @@
...
-	Event      string                    `config:"event"`
+	Mode          libilm.Mode `config:"enabled"`
+	Overwrite     bool        `config:"overwrite"`
+	RequirePolicy bool        `config:"require_policy"`
+	Policies      []Policy
 }
 
-// Enabled indicates whether or not ILM should be enabled
+//Policy binds together an ILM policy's name and body with an event type
+type Policy struct {
```
jalvz: can we name this `EventPolicy` or something like that to distinguish from the inner `Policy` attribute?

idxmgmt/ilm/config.go
```diff
@@ -18,34 +18,117 @@
...
+		//update policy name per event according to configuration
+		for _, entry := range cfg.Mapping {
+			if _, ok := policyMappings[entry.Event]; !ok {
+				return c, errors.Errorf("event_type '%s' not supported for ILM setup", entry.Event)
+			}
+			policyMappings[entry.Event] = entry.Policy
+		}
+	}
+
+	for event, policyName := range policyMappings {
```
jalvz: as counterpoint to https://github.com/elastic/apm-server/pull/2764/files#diff-b9f51268216554aa54cd5bdb34d4ecb1R210, this reads well, much better that if you were to do `for k, v :=...`

idxmgmt/manager.go
```diff
@@ -201,11 +213,15 @@ func (m *manager) loadPolicy(ilmFeature feature, ilmSupporter libilm.Supporter)
 		return false, nil
 	}
 	policy := ilmSupporter.Policy().Name
+	if ilmSupporter.Policy().Body == nil {
```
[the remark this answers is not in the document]
simitt: I don't think this is worth bloating the `ilm.Supporter` interface in `libbeat`.  

The name `IsPolicyLoaded` would also not be quite appropriate here, as this check decides wheter or not the `manager` responsible for loading policies should be called or not. We do use the standard ILM supporter, as all the loading logic is identic, and I don't want to diverge from that, just because of this check.

idxmgmt/ilm/config.go
```diff
@@ -18,34 +18,117 @@
...
-func ModeString(m libilm.Mode) string {
-	switch m {
-	case libilm.ModeAuto:
-		return "auto"
-	case libilm.ModeEnabled:
-		return "true"
-	default:
-		return "false"
+//NewConfig returns an ILM config, where default configuration is merged with user configuration
+func NewConfig(ucfg *common.Config) (Config, error) {
```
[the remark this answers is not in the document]
simitt: change terms a bit

idxmgmt/manager.go
simitt: `m.supporter.st.ilmEnabled` is an [atomic.Bool](https://github.com/uber-go/atomic/blob/master/atomic.go#L232).

idxmgmt/ilm/policy.go
simitt: I don't think that this should be a public type. You argue for not having abbreviations, having a public type `M` or similar is worse IMO.

idxmgmt/ilm/config.go
simitt: Why would that be redundant?

If the default configuration configures a policy for event_type x, but the user chooses to configure it referring to a policy that is not loaded via apm server, we have to ensure that no policy is loaded for this event type.

_meta/beat.yml
simitt: We can simply rename the header if Kibana is used for something else in the future. Not sure I get your point.
With all the long configurations, I think it helps readability grouping them together.

_meta/beat.yml
simitt: We discussed removing `name` and `body` in this case and using the name as key, during one of our offline meetings to keep the config as short as possible. Therfore I changed to this format.
@graphaelli are you also in favor of changing back and introducing a `name` and `policy` approach again?

idxmgmt/ilm/policy.go
jalvz: I'm arguing for readability, and I think we saw enough of those to know what it means without going to the type definition. If it is public it can be called `Map` (a better name than `m`), but if it is private it can't be called `map`.

So why is worse to declare it once and reuse it everywhere than declaring it each time of use?

idxmgmt/ilm/config.go
jalvz: maybe i'm missing something, but if a map lookup fails it returns (`nil`, `false`), so in the `if !ok` branch it is not needed to do `policy = nil` because it is `nil` already?

_meta/beat.yml
jalvz: i think removing `body` is a good idea either case

idxmgmt/manager.go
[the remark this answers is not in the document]
jalvz: if i would have to answer the question "how to know if an ilm policy is loaded" i don't think that "check that `ilmSupporter.Policy().Body` is `nil`" would cross my mind, but i'm fine with that. 
hopefully we don't have to do this check in many places

_meta/beat.yml
jalvz: (unrelated) i also think that default policy names should start with `apm-` prefix just like index names

idxmgmt/ilm/config.go
```diff
@@ -18,34 +18,117 @@
...
+			}
+			policyMappings[entry.Event] = entry.Policy
+		}
+	}
+
+	for event, policyName := range policyMappings {
+		policy, ok := policies[policyName]
+		if !ok {
+			if c.RequirePolicy {
+				return Config{}, errors.Errorf("policy '%s' not configured for ILM setup", policyName)
```
jalvz: in the error message you could mention "set ilm.require_policy to false" or something like that

simitt:
I received some offline feedback from @jalvz that the newly introduced behavior is not intuitive, as `apm-server.ilm.enabled`, `apm-server.ilm.overwrite`, `setup.template.enabled` and `setup.template.overwrite` are entangled in a quite complex way. With introducing an ILM specific flag deciding whether or not ILM setup should be overwritten, it would be a very unexpected behavior and might even be considered a bug that `setup.template.*` flags are influencing whether or not ILM specific templates are getting loaded. 

An invalid setup of ILM puts the index management for APM on risk, as indices might grow too big and do not rollover as expected, or are not managed at all. To avoid broken setups based on non-intuitive configuration, we decided to isolate the ILM setup completely from all `setup.template.*` config options. 

From this PR on setting up ILM will behave as follows:

 apm-server. ilm. enabled | apm-server. ilm.setup. enabled  |   | create policy | overwrite policy | create template | overwrite template | create alias
-- | -- | -- | -- | -- | -- | -- | --
auto/true | true |   | true | true | true | true | true
auto/true | false |   | false | false | false | false | false
false | true |   | false | false| true | true | false
false | false |   | false | false | false | false | false

With `apm-server.ilm.enabled: auto` and `apm-server.ilm.setup.enabled: true` being the default values.

_meta/beat.yml
simitt: I got rid of it alltogether.

jalvz:
Let me see if I get this right:
Inside the `apm-server.ilm` section there is an `enabled` attribute (bool) and a nested `setup` section, which also has an `enabled` attribute. 

If the innermost (`apm-server.ilm.setup.enabled`) is `false`, doesn't matter what the outermost is (`apm-server.ilm.enabled`), nothing will happen - second and last rows are the same.

If the innermost (`apm-server.ilm.setup.enabled`) is `true`, and the outermost (`apm-server.ilm.enabled`) is `false`, then only the template is written, and not the policies/aliases.

Is that correct?

simitt:
@jalvz that's the behavior.

jalvz:
thanks for confirming, will check it out

_meta/beat.yml
```diff
@@ -247,6 +239,88 @@ apm-server:
...
+        #  policy_name: "rollover-1-day"
+        #- event_type: "transaction"
+        #  policy_name: "rollover-7-days"
+        #- event_type: "metric"
+        #  policy_name: "rollover-7-days"
+
+      # Configured policies are added to all pre-defined default policies.
+      # If a policy with the same name as a default policy is configured, the configured policy overwrites the default policy.
+      #policies:
+        #name: "rollover-7-days"
```
jalvz: use array syntax `# -name: "rollover-7-days"`

same for others bellow

simitt:
> one last note: the line "Existing template will be overwritten, as overwrite is enabled." is a bit confusing as there is no overwrite setting for ilm templates

The template loading functionality is reused from libbeat. When preparing the templates `overwrite:true` is set, as otherwise templates wouldn't be loaded if they exist. While I agree the log message can be confusing, I don't think it is incorrect, as the templates _will_ be overwritten, even though the user doesn't have a config option to disable this other than `ilm.setup.enabled`.

PR #2769, opened 2019-10-04
ci(codecov): resolve PR coverage measures errors

kuisathaverat:
## What does this PR do?

We detected that Codecov does not always return accurate coverage measures on PRs, so after digging a little in the Codecov issues we found a configuration that works better for us.

## Why is it important?

After applying this configuration, we'd not see wrong coverage measures in the PRs.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,16 @@
+codecov:
+  notify:
+    require_ci_to_pass: no
+comment:
+  layout: "header, changes, diff, files"
+  behavior: default
+coverage:
+  status:
+    project:
+      default:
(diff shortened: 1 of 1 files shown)


codecov.yml
```diff
@@ -0,0 +1,16 @@
...
+  notify:
+    require_ci_to_pass: no
+comment:
+  layout: "header, changes, diff, files"
+  behavior: default
+coverage:
+  status:
+    project:
+      default:
+        target: 80%
```
kuisathaverat: we use an 80% for cobertura target, we can change to another target

PR #2770, opened 2019-10-04
Update beats framework to e40fbdd

jalvz:


_beats/testing/environments/latest.yml
```diff
@@ -26,7 +26,7 @@ services:
       - ./docker/logstash/pki:/etc/pki:ro
 
   kibana:
-    image: docker.elastic.co/kibana/kibana:7.2.0
+    image: docker.elastic.co/kibana/kibana:7.3.2
```
simitt: Shouldn't that be `7.4.0` now (I know it is inherited, but might be worth a manual change until updated).

jalvz: :( good catch

PR #2771, opened 2019-10-04
Copy `client.ip` to `source.ip`

### LINKED ISSUE #2766: Copy client.ip address to source.ip address for SIEM integration
Copy the value of `client.ip` to `source.ip` when available for future APM SIEM integration. 

@MikePaquette I am not a 100% sure about the semantic difference between `client` and `source` for APM related data. Would you agree that we can just copy the value and treat it the same?
cc @SergeyKleyman as you are more familiar with the semantics of this value (it corresponds to `context.request.socket.remote_address` on intake API).

simitt:
Ensure to populate the `source.ip` field whenever possibel in alignment
with ECS.

fixes elastic/apm-server#2766


changelogs/head.asciidoc
```diff
@@ -4,5 +4,6 @@
 [float]
 ==== Added
 - Add `service.node.configured_name` to Intake API and transform to `service.node.name` for ES output {pull}2746[2746].
+- Index value from `client.ip` in `source.ip`for SIEM integration {pull}2771[2771].
```
graphaelli: for ECS compatibility ?

graphaelli:
Did you decide we don't want to geoip process this in `source.geo`?

simitt:
Afaik it is not needed for any visualization, so I suggest to not enable it and duplicate data (for now).

PR #2772, opened 2019-10-04
Parse IP and only store valid IPs in client.ip

simitt:
The GeoIP pipeline requires IP addresses to be in valid format. Only
store valid IP addresses in `client.ip`.

fixes 2765


model/context.go
```diff
@@ -396,3 +403,28 @@ func (s *Socket) fields() common.MapStr {
...
+	return m
+}
+
+func parseIP(i interface{}) string {
+	input, ok := i.(string)
+	if !ok {
+		return ""
+	}
+	var ip net.IP
+	if ip = net.ParseIP(input); ip != nil {
```
graphaelli: Is a [net.SplitHostPort](https://golang.org/pkg/net/#SplitHostPort) on all input before parsing an option instead?

simitt: Not on all input, as it expects a port to be present, but I can definitely use it for the second part, thank you.

model/error/event.go
```diff
@@ -194,7 +194,12 @@ func (e *Event) Transform(tctx *transform.Context) []beat.Event {
 	tctx.Metadata.Set(fields)
 	// then add event specific information
 	utility.Update(fields, "user", e.User.Fields())
-	utility.DeepUpdate(fields, "client", e.Http.ClientFields(e.User.ClientFields()))
+
+	clientFields := e.User.ClientFields()
```
jalvz: This logic is duplicated. It can be solved with interfaces alla Go ("accept an interface, return a concrete type"):

```diff
diff --git a/model/error/event.go b/model/error/event.go
index 7f2081172..cd87608ef 100644
--- a/model/error/event.go
+++ b/model/error/event.go
@@ -195,11 +195,7 @@ func (e *Event) Transform(tctx *transform.Context) []beat.Event {
        // then add event specific information
        utility.Update(fields, "user", e.User.Fields())
 
-       clientFields := e.User.ClientFields()
-       if clientFields == nil {
-               clientFields = e.Http.ClientFields()
-       }
-       utility.DeepUpdate(fields, "client", clientFields)
+       utility.DeepUpdate(fields, "client", m.ClientFields(e))
        utility.DeepUpdate(fields, "user_agent", e.User.UserAgentFields())
        utility.DeepUpdate(fields, "service", e.Service.Fields(emptyString, emptyString))
        utility.DeepUpdate(fields, "agent", e.Service.AgentFields())
@@ -474,3 +470,11 @@ func flattenExceptionTree(exception *Exception) []Exception {
        }
        return recur(*exception, 0)
 }
+
+func (e *Event) Usr() metadata.User {
+       return *e.User
+}
+
+func (e *Event) HTTP() m.Http {
+       return *e.Http
+}
diff --git a/model/thing.go b/model/thing.go
index 8b5379070..bbccfd8c7 100644
--- a/model/thing.go
+++ b/model/thing.go
@@ -1 +1,19 @@
 package model
+
+import (
+       "github.com/elastic/apm-server/model/metadata"
+       "github.com/elastic/beats/libbeat/common"
+)
+
+func ClientFields(e IEvent) common.MapStr {
+       clientFields := e.Usr().ClientFields()
+       if clientFields == nil {
+               clientFields = e.HTTP().ClientFields()
+       }
+       return clientFields
+}
+
+type IEvent interface {
+       Usr() metadata.User
+       HTTP() Http
+}
diff --git a/model/transaction/event.go b/model/transaction/event.go
index 1b5832e41..3a8e8e3cb 100644
--- a/model/transaction/event.go
+++ b/model/transaction/event.go
@@ -182,11 +182,7 @@ func (e *Event) Transform(tctx *transform.Context) []beat.Event {
 
        // then merge event specific information
        utility.Update(fields, "user", e.User.Fields())
-       clientFields := e.User.ClientFields()
-       if clientFields == nil {
-               clientFields = e.Http.ClientFields()
-       }
-       utility.DeepUpdate(fields, "client", clientFields)
+       utility.DeepUpdate(fields, "client", m.ClientFields(e))
        utility.DeepUpdate(fields, "user_agent", e.User.UserAgentFields())
        utility.DeepUpdate(fields, "service", e.Service.Fields(emptyString, emptyString))
        utility.DeepUpdate(fields, "agent", e.Service.AgentFields())
@@ -203,3 +199,11 @@ func (e *Event) Transform(tctx *transform.Context) []beat.Event {
 
        return events
 }
+
+func (e *Event) Usr() metadata.User {
+       return *e.User
+}
+
+func (e *Event) HTTP() m.Http {
+       return *e.Http
+}
```

Ignoring naming and pointers, you get the idea. Possibly it can be done for all metadata or ecs common fields. 
Feel free to ignore.

beater/test_approved_es_documents/TestPublishIntegrationTransactions.approved.json
```diff
@@ -335,7 +335,10 @@
             },
             "http": {
                 "request": {
-                    "method": "post"
+                    "method": "post",
+                    "socket": {
+                        "remote_address": "192.0.1"
```
jalvz: I am confused. Isn't this an illegal IP?

model/error/event.go
simitt: I agree, the code duplication should be removed. But there is a lot more duplication between transaction and errors, and I believe even some in spans. I prefer to get rid of the duplication all together, to not introduce more confusion, and since the major part of it does not have anything to do with sanitizing IP addresses, I prefer doing this in a separate PR.

Created https://github.com/elastic/apm-server/issues/2776 to follow up with that.

beater/test_approved_es_documents/TestPublishIntegrationTransactions.approved.json
simitt: Yes it is, but the `http.request.socket.remote_address` stores the original value. The sanitization is only applied to the `client.ip`, which is an ECS field, indexed, and used for further processing, such as the GeoIP pipeline.

PR #2774, opened 2019-10-04
ci(jenkins): reuse top level agent

v1v:
### Highlights
- Let's reuse the top-level agent to reduce the number of fo requested agents and speed up just a bit the builds.
- Reset labels to be `linux && immutable` to help with what kind of OS is required to be used.
- Time improvements per Stage:
  - From the very beginning to the build Linux stage used to take 7 mnts approx and 3 workers and now less than 5 minutes and just one worker.

### Test Cases
- Release

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -350,6 +350,20 @@ pipeline {
         GOPATH = "${env.WORKSPACE}"
         SNAPSHOT="true"
       }
+      when {
+        beforeAgent true
+        allOf {
+          anyOf {
+            branch 'master'
+            branch "\\d+\\.\\d+"
+            branch "v\\d?"
+            tag "v\\d+\\.\\d+\\.\\d+*"
+            expression { return params.Run_As_Master_Branch }
+            expression { return env.BEATS_UPDATED != "false" }
+          }
+          expression { return params.release_ci }
+        }
+      }
       steps {
         withGithubNotify(context: 'Release') {
           deleteDir()
@@ -360,6 +374,7 @@ pipeline {
           golang(){
             dir("${BASE_DIR}"){
               sh(label: 'Build packages', script: './script/jenkins/package.sh')
+              sh(label: 'Test packages install', script: './script/jenkins/test-install-packages.sh')
             }
           }
         }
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -350,20 +350,6 @@ pipeline {
         GOPATH = "${env.WORKSPACE}"
         SNAPSHOT="true"
       }
-      when {
```
kuisathaverat: This stage only run on master branch and tags

Jenkinsfile
```diff
@@ -374,20 +360,13 @@ pipeline {
...
               sh(label: 'Build packages', script: './script/jenkins/package.sh')
-              sh(label: 'Test packages install', script: './script/jenkins/test-install-packages.sh')
             }
           }
         }
       }
       post {
         success {
           echo "Archive packages"
-          googleStorageUpload(bucket: "gs://${JOB_GCS_BUCKET}/snapshots",
```
kuisathaverat: we have to store the packages in GCP

Jenkinsfile
```diff
@@ -374,20 +360,13 @@ pipeline {
           golang(){
             dir("${BASE_DIR}"){
               sh(label: 'Build packages', script: './script/jenkins/package.sh')
-              sh(label: 'Test packages install', script: './script/jenkins/test-install-packages.sh')
```
kuisathaverat: Why do we remove the test?

Jenkinsfile
kuisathaverat: forget, I've seen it is only for testing

Jenkinsfile
kuisathaverat: forget, I've seen it is only for testing

Jenkinsfile
kuisathaverat: but it is easy to add an expression to the when condition that it is always false, something like

```
// Disabled temporarily for testing
expresion { return false }
```

Jenkinsfile
v1v: For the record, I just wanna be sure by reusing the same agent between stages, even when running some docker magic steps the permissions are intact, therefore no surprises for those stages which are only executed in the master branch.

Issue #2338, opened 2019-06-27
Rely on Elasticsearch API Keys instead of secret token for authentication


roncohen:
The `secret_token` was something we did before [API Keys](https://www.elastic.co/guide/en/elasticsearch/reference/current/security-api.html#security-api-keys) existed in th Elasticsearch. Now that it's here, we should investigate using them instead of a secret token. 

There are multiple advantages, for instance:
* allow for multiple tokens at the same time
* add/remove tokens without restarting APM Server
* centralized access control across the stack instead of at the edge

We could create a special [application privilege](https://www.elastic.co/guide/en/elastic-stack-overview/7.2/security-privileges.html#application-privileges) that is required by APM Server for a token to be valid. APM Server can use the [has-privileges API call](https://www.elastic.co/guide/en/elasticsearch/reference/current/security-api-has-privileges.html) to ensure that the token supplied has the correct application privileges. 

Using the framework of application privileges we can also provide much more granular access for example, we can lock down tokens only to be used for RUM or tokens only to be used for specific services, tokens that will only allow error tracking etc etc.

cc @graphaelli who kindly helped investigate

simitt:
@graphaelli created a [POC](https://github.com/elastic/apm-server/compare/master...graphaelli:api-key-auth) for API Key integration. 

Using API keys requires every service to be registered with agent privileges. API keys can be set to expire.

axw:
Does this issue cover providing an API for authenticated agents to request tokens? This could be used to provide a time-limited token for authenticating RUM agents. See https://github.com/elastic/apm-server/issues/1718#issuecomment-519763885.

simitt:
@axw that's an interesting idea, but from my understanding not covered by this issue. As a start it would be up to the users to use existing ES APIs to create and retrieve tokens and forward to the monitored services.

graphaelli:
Agreed that's interesting and a likely next step.  For now, let's (of course) include careful consideration of the privilege setup to ensure only certain API keys can create the short-lived API keys but save the apm-server endpoint for creating those for a follow on effort.

jalvz:
Tested the following 2 scenarios:

**A** 
- Run `./apm-server apikey create --expiration 20s`
- Let the the key expire.
- Ingest an event with that key.
Result is `401/unauthorized`, as expected.

**B**
- Run `./apm-server apikey create --expiration 20s`
- Ingest an event with that key.
- Result is `202/accepted`, as expected.
- Let the the key expire.
[...]

PR #2793, opened 2019-10-11
DRAFT: API key support

simitt:
This is a first draft of supporting API Keys in the APM Agent to APM Server communication. 

The API Key support is in addition to the current secret_token approach, if both are configured then whatever is sent in the request header will be used for verifying the request. Allowing both at the same time is more user-friendly than forcing to update all agents at the same time. 

The branch uses the new [go-elasticsearch client](https://github.com/elastic/go-elasticsearch) for synchronous communication with Elasticsearch. There is a separate config section to configure the ES cluster that should server the API key requests, if not configured, the configuration from the ES output section is used. Only a subset of the currently used config options of the ES output are supported, since not all of them seem important for sync. requests (e.g. retry or backoff). Options regarding proxies, ssl etc are supported. 

Every time a request with an API Key is made to the server, the api key needs to be verified. This is done in three steps:
- request local cache: simple in-memory map structure, gets filled the first time an authorization query is made for the request. It is filled either from the global cache, or ES.
- global cache: fixed size, auto expiring cache, that is shared per APM Server between all incoming requests from secured endpoints. In case this ever gets extended to RUM, we need to ensure RUM endpoints share a _different_ cache. The global cache stores entries per API Key + service name. The entry gets filled the first time a request with this API Key + service combination is made, after fetching the information from ES. In case multiple requests with the same key/service combination occur at the same time, multiple queries to ES are made, potentially overwriting each others entries in the cache. This is not a problem, as the underlying cache is a threadsafe cache implementation [patrickmn/go-cache](github.com/patrickmn/go-cache). 
Since secured requests to backend endpoints now involve queries to caches and external sources instead of doing a constant time comparison of two tokens we need to safeguard the server, and especially the external source from being overloaded. This could be done by some kind of rate limiter, which would need some specific handling to allow a high number of benign requests while having a low tolerance for unauthorized requests. To avoid additional complexity, instead of introducing such a rate limiter, we use a fixed cache size, avoiding overly many requests to ES. If the cache is full, no more new tokens can be served. Since the cache is auto-expiring (by default after 5mins), this is a temporary situation and new tokens will be accepted after expiration. The cache uses a cleanup routine on a regular schedule (currently 60s) to check for expired keys and deleting them serial. The cache also allows to register an `OnEviction` function which is used here. When a key is deleted, a check occurs if the entry was a valid one (storing abilities that were allowed), and in case it was a valid entry, a re-fetch to ES is initiated to update the abilities for the token+service and the entry is pushed to the cache again. This ensures that in case an attack is happening, valid tokens that have been used already stay in the cache and requests using those tokens can still be served. There are more optimizations that can be done here (e.g. avoid hitting ES with requests from evicted tokens every minute). 
- Elasticsearch: the `hasPrivilege` endpoint is called, the first time an authorization method is called for a request. It requests privileges for all known and supported APM Server privileges and for all known applications for this request at once, to avoid repeated requests. Results are pushed into the global and the local cache. 

This PR only supports API keys that are authorized for all services. The API Keys need privileges for application `apm-backend`. The minimum privilege a key for a secured endpoint must have is the `action:access` privilege. Endpoints can then require more specific privileges, e.g. `action:intake`, `action:sourcemap` and `action:config` are required for the respective endpoints. For convenience reasons only applying a privilege `action:full` to an API key is also supported. The implementation is already built in a way that the authorization queries can also be made from within every handler for future support of API Keys per service. 
Application names for API keys only support a subset of characters allowed by APM for service names. The users would need to change their service names to not contain any non-supported characters.

This is only a draft PR for now as more extensive manual and automatic testing is required. 

TODO:
- [ ] changelog entry
- [ ] thorough manual testing
- [x] more automated unit tests (authorization middleware, muxer integration tests only use bearer token so far)
- [ ] system tests with ES that has security enabled
- [ ] sharing a recommended test setup (script) for easier testing
- [ ] discuss and review the on-eviction handling for auto-expired cache keys
- [ ] manually copied license of go-elasticsearch client to root folder for license check. Check if this can be improved.
- [ ] update go-elasticsearch client to latest version

# implements elastic/apm-server#2338


elasticsearch/client.go
```diff
@@ -0,0 +1,151 @@
...
+)
+
+const (
+	prefixHttp       = "http"
+	prefixHttpScheme = prefixHttp + "://"
+	esDefaultPort    = 9200
+)
+
+var (
+	invalidHosts = errors.New("invalid host configuration, must at least contain one hostname")
```
karmi: Perhaps change the word order to `must contain at least one hostname`?

simitt:
This PR is going to be split up into:
* integrating go-elasticsearch library with the dedicated `elasticsearch` package and config changes
* support secured ES and Kibana in system tests
* dedicated API Key support (blocked for now)

PR #2794, opened 2019-10-14
[7.4] Parse IP and only store valid IPs in client.ip (#2772)

simitt:
Backports #2772 to 7.4


changelogs/head.asciidoc
```diff
@@ -0,0 +1,10 @@
+[[release-notes-head]]
```
jalvz: remove this file altogether

PR #2795, opened 2019-10-14
[7.x] Parse IP and only store valid IPs in client.ip (#2772)

### LINKED ISSUE #2765: Ensure only valid IP addresses are stored in `client.ip`
Ensure to only index information that can be handled by the GeoIP processor in `client.ip`.  As uncovered in a [discuss forum entry](https://discuss.elastic.co/t/apm-7-4-error-with-default-pipeline-apm/201976) invalid information can prevent events from being properly ingested to Elasticsearch. 

The information is sent up by agents in the `context.request.socket.remote_address` field. Agents do not currently sanitize this field, so it can hold information other than IP addresses.  In case the value of the field is not in a valid IP address format, either don't store the information or remove additional information (e.g. port).

simitt:
Backports following commits to 7.x:

>The GeoIP pipeline requires IP addresses to be in valid format. Only
>store valid IP addresses in `client.ip`.

> fixes #2765


jalvz:
lgtm

simitt:
failing CI is unrelated

PR #2799, opened 2019-10-14
enable xpack security in system tests

graphaelli:
Start Elasticsearch with xpack security using file realm.  Kibana and apm-server access es using user/pass authentication.

All user passwords set to `changeme`.  The `apm_user` also serves as a kibana user due until the kibana user can be injected into the testing environment, requiring changes upstream in `_beats/libbeat/scripts/Makefile`.


graphaelli:
Now that the tests are passing I think we should break this up a bit.  This changeset should include just the bits needed for enabling xpack security.

graphaelli:
To be merged after #3014, probably easier to review at that point also.

PR #2814, opened 2019-10-17
Adding one more level of hierarchy for referenced doc

eyalkoren:
This doesn't render well in documentation. Not sure if that's the reason, but if so- this is just an example, it is all over the event docs at this level


eyalkoren:
Right.. 🤦‍♂ 
Rushed with it 😊

PR #2877, opened 2019-10-30
[ILM] Add ilm.setup.overwrite config option.

simitt:
To avoid overwriting existing setups on every startup, a dedicated
config option is added. This is more consistent with other settings and
aligns better with cloud deployment.

related to elastic/apm-server#2240

cc @nachogiljaldo 

TODO: 
* [x] Add changelog entry

### DIFF
--- a/idxmgmt/manager_test.go
+++ b/idxmgmt/manager_test.go
@@ -46,64 +46,64 @@ func TestManager_VerifySetup(t *testing.T) {
 		ok   bool
 		warn string
 	}{
-		//"SetupTemplateDisabled": {
-		//	loadTemplate: libidxmgmt.LoadModeEnabled,
-		//	warn:         "Template loading is disabled",
-		//},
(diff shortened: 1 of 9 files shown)


idxmgmt/manager_test.go
```diff
@@ -45,84 +46,86 @@ func TestManager_VerifySetup(t *testing.T) {
...
-				"output.elasticsearch.enabled": false,
-				"output.logstash.enabled":      true},
-			warn: "automatically disabled ILM",
-		},
-		"EverythingEnabled": {
-			templateEnabled: true, loadTemplate: libidxmgmt.LoadModeEnabled,
-			ilmSetupEnabled: true, loadILM: libidxmgmt.LoadModeEnabled,
-			ok: true,
-		},
+		//"LoadILMDisabled": {
```
jalvz: these should be uncommented?

PR #2895, opened 2019-11-13
Update beats framework to b110c7a

jalvz:
Fixes Kibana version falling out of sync


vendor/github.com/elastic/beats/libbeat/cmd/instance/beat.go
```diff
@@ -1037,3 +1042,10 @@ func initPaths(cfg *common.Config) error {
 	}
 	return nil
 }
+
+func setUmaskWithSettings(settings Settings) error {
+	if settings.Umask != nil {
+		return setUmask(*settings.Umask)
+	}
+	return setUmask(0027) // 0640 for files | 0750 for dirs
```
simitt: Didn't look into that and I think `0027` should be fine for APM Server also, but afaik with this change it is more restrictive than before. Maybe worth testing (or at least keeping it in mind if we run into any issues).

PR #2908, opened 2019-11-14
[7.5][docs] 7.5.0 release notes and breaking changes

bmorelli25:
**Do not merge until `7.5.0` release**
Needs to be cherrypicked to `7.x` and `master`.
[Live demo here](http://apm-server_2908.docs-preview.app.elstc.co/diff).

---

This PR updates the following documentation:
* [APM Overview Release Notes](https://www.elastic.co/guide/en/apm/get-started/7.5/apm-release-notes.html)
* [APM Overview Breaking Changes](https://www.elastic.co/guide/en/apm/get-started/7.5/apm-breaking-changes.html)
* [APM Server Breaking Changes](https://www.elastic.co/guide/en/apm/server/7.5/breaking-changes.html)


docs/guide/apm-release-notes.asciidoc
```diff
@@ -25,6 +26,17 @@ For a full list of changes, see the
 
 // end::notable-v8-highlights[]
 
+[[release-highlights-7.5.0]]
```
jalvz: here or somewhere else: agent config goes GA, with new settings, RUM, support for "All services" (in Kibana).
GA breaks with beta, so a 7.5 apm-server will require 7.5 kibana and viceversa.

docs/guide/apm-breaking-changes.asciidoc
```diff
@@ -22,6 +23,16 @@ Also see <<apm-release-notes>>.
 
 // end::notable-v8-breaking-changes[]
 
+[[breaking-7.5.0]]
+=== 7.5.0 APM Breaking changes
+
+// tag::notable-v75-breaking-changes[]
+
+to do -- needs more information
```
simitt: How about something like:
```
- Introduce dedicated `apm-server.ilm.setup.*` flags allowing to customize ILM behavior from within the APM Server configuration. As a side effect `setup.template.*` settings will be ignored for ILM related templates per event type. 
- By default ILM policies will not be versioned anymore, and all event types switched to use the new default policy to rollover after 30 days or when reaching a size 50gb. 
```

PR #2918, opened 2019-11-18
Add the `Access-Control-Expose-Headers` header to the agent config response

### LINKED ISSUE #2915: RUM config endpoint should have Access-Control-Expose-Headers: ETag header in its response
`Access-Control-Expose-Headers: ETag` should be present on the RUM config endpoint. This is required so that the agent would be able to read this header.

Related discussion: https://github.com/elastic/apm-server/issues/2694#issuecomment-533518378

jalvz:
This enables caching on the RUM agent.

Fixes #2915


beater/api/config/agent/handler.go
```diff
@@ -97,6 +97,7 @@ func Handler(client kibana.Client, config *config.AgentConfig) request.Handler {
 		// configuration successfully fetched
 		c.Header().Set(headers.CacheControl, cacheControl)
 		c.Header().Set(headers.Etag, fmt.Sprintf("\"%s\"", result.Source.Etag))
+		c.Header().Set("Access-Control-Expose-Headers", headers.Etag)
```
simitt: nit: could you add the `"Access-Control-Expose-Headers"` to the [headers/keys](https://github.com/elastic/apm-server/blob/master/beater/headers/keys.go), for reuse (it is also used in the CORS middleware)

PR #2926, opened 2019-11-19
Returns a sentinel value as Etag when there is no config in Kibana.

### LINKED ISSUE #2916: Config endpoint should avoid setting empty value on Etag header
Currently, if no configuration exist for a given service name the APM server response from the config endpoint includes `Etag: ""`. While this doesn't create issues, it's better to just avoid sending this header if no configuration exist.

jalvz:
This is so agents can cache empty configurations, as they don't
necessarily inspect the contents of a response body to decide
whether they should process it, and just make a decision based on
the response code (304 vs 200).

Fixes elastic/apm-server#2916


jalvz:
Thanks. 
An empty Etag would mean that they would get a 304 on the first request (because of the removal of the `if etag == ""` check), but that probably would have been fine indeed (albeit a bit more unconventional).

PR #2935, opened 2019-11-20
Parse valid IP from event payload headers

### LINKED ISSUE #2867: Parse headers for client IP address and use remote_address only as fallback.
With https://github.com/elastic/apm/issues/157 the agents are streamlining on sending the remote_address in `context.request.socket.remote_address` without trying to parse any original IP from headers. 
Users can configure agents to not send up headers, by default request headers are sent to the APM Server. When headers are sent, the APM Server can parse them to try to collect information about the source of the request and store that in `client.ip` and `source.ip`. Those values are used for SIEM integration and the GeoIP processor. 

For RUM events the APM Server already has an algorithm in place to [derive the IP address from headers if available](https://github.com/elastic/apm-server/blob/master/utility/remoteaddr.go#L40). I suggest to reuse this logic also for backend events to collect the `client.ip` and `source.ip` address.

simitt:
Streamline behaviour for populating client.ip and source.ip fields, by reusing RUM event logic to parse IP address for backend events. This is a follow up on agents changing their logic on what information they are sending as socket.remote_address.
Parse `Forwarded`, `X-Real-IP` and `X-Forwarded-For` headers in this order, if no valid IP can be parsed, parse `RemoteAddr` from request.

Contains a change to how data are enriched. So far the first remote address parsed from the headers or from the request has been returned for enriching event data, added as `user.ip` (RUM) or `system.ip` (backend). This PR changes behavior in the way that the first _valid IP_ is returned instead. 
E.g. if  the`Forwarded` header contains a value from which no IP address can be parsed, we now try to extract the information from the other headers. @axw since you originally contributed the logic for this parsing, I'd appreciate your feedback. 

fixes elastic/apm-server#2867

This is necessary for APM/SIEM integration.


axw:
@simitt I only looked at the ExtractIP bit - your change looks reasonable to me.

PR #2959, opened 2019-11-25
Fix: respect setting for overwriting ILM setup

### LINKED ISSUE #2958: [Bugfix] respect `apm-server.ilm.setup.overwrite` setting
Ensure that `apm-server.ilm.setup.overwrite` setting is respected when starting APM Server.

simitt:
When running APM Server `apm-server.ilm.setup.overwrite` was ignored and always set to the same value as `apm-server.ilm.setup.enabled`. 

fixes #2958

### DIFF
--- a/idxmgmt/ilm/supporter_factory.go
+++ b/idxmgmt/ilm/supporter_factory.go
@@ -60,7 +60,7 @@ func MakeDefaultSupporter(
 		}
 
 		supporter := libilm.NewStdSupport(log, mode, libilm.Alias{Name: alias, Pattern: pattern},
-			libilm.Policy{Name: p.Name, Body: p.Policy}, ilmConfig.Setup.Enabled, true)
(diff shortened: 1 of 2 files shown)


blookot:
Hi @simitt 
is this fix in 7.5 really? it seems a customer is still experiencing issues with template overwriting: after a few days, their custom ILM template is overwritten by an empty one although they didn't configure the `overwrite` key to true...
thanks

simitt:
@blookot the fix was part of the `7.5.0` release (that's why you don't find it in the changelog, as it was introduced and fixed in the same release). 

There's an example in the docs for how to [enable ILM and disable ILM setup](https://www.elastic.co/guide/en/apm/server/current/ilm-reference.html). Could you please check back with the developers to ensure they are configuring the correct setting: `apm-server.ilm.setup.overwrite: false`. If so, I'd appreciate if you could open a new bug report with some more details so I can try to reproduce.

blookot:
Sure
Thank you very much @simitt I'll follow up in a separate issue if needed

PR #2968, opened 2019-11-27
approvals: use go-cmp instead of gojsondiff

axw:
While working on https://github.com/elastic/apm-server/pull/2839, I found that large maps with many differences would lead to gojsondiff causing tests to churn CPU and eventually time out. This PR replaces use of gojsondiff with go-cmp, which has no such issues.

There is one functional difference: if the approval file is missing, we now compare the received JSON with `null` rather than `{}`. This is to cater for some uses of ApproveEvents with nil maps. Whereas `gojsondiff` will produce no diff when comparing nil and empty maps, go-cmp will. This is why some approval files have been added.


simitt:
Failing tests are all unrelated, feel free to merge any time and please backport to 7.x

PR #2973, opened 2019-11-27
ci(jenkins): run release stage in parallel

v1v:
This might reduce the execution time in about 10 minutes, the average time for the release stage is about 25 minutes. And the current parallel stages take up to 15 minutes or so. 

This approach might require some validation from the apm-server team to ensure they are happy with this change.

## UI when skipped

image

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -314,6 +314,59 @@ pipeline {
           }
         }
         /**
+          build release packages.
+        */
+        stage('Release') {
+          agent { label 'linux && immutable' }
+          options { skipDefaultCheckout() }
+          environment {
+            PATH = "${env.PATH}:${env.WORKSPACE}/bin"
+            HOME = "${env.WORKSPACE}"
+            GOPATH = "${env.WORKSPACE}"
+            SNAPSHOT="true"
+          }
+          when {
+            beforeAgent true
+            allOf {
+              anyOf {
+                branch 'master'
+                branch "\\d+\\.\\d+"
+                branch "v\\d?"
+                tag "v\\d+\\.\\d+\\.\\d+*"
+                expression { return params.Run_As_Master_Branch }
+                expression { return env.BEATS_UPDATED != "false" }
+              }
+              expression { return params.release_ci }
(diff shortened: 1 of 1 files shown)


kuisathaverat:
we do not have the failFast enabled so you would make the release packages even do tests fail, also we will mix test and release, I do not really like it.

mdelapenya:
> we do not have the failFast enabled so you would make the release packages even do tests fail, also we will mix test and release, I do not really like it.

I agree, release stage should come if and only if tests pass, so I'd move it after that

v1v:
IIUC, the release stage is not a release but something else. Releases are managed within the unified release process. Maybe the release stage might mislead and could be renamed to something else, as an idea. That's the reason I initially thought to run more things in parallel, no strong opinion just wanted to gather feedback on whether this approach was helpful. Let\s see what the apm-server team think also :)

simitt:
The _release_ stage tests building and installing the release packages, I think the naming is fair, but we could also rename to e.g. _packaging_.
I agree that technically those tests only need to be run when all the other tests passed, and I would not want to prolong the parallel test execution phase, as this is where issues usually get uncovered, the _release_ stage rarely fails. 
Would it be possible to enable a _fail_fast_ strategy only on the release stage and run it in parallel with the other tests? This way it would reduce overall waiting time in the success case, but stop whenever one or multiple of the other test steps has failed?

kuisathaverat:
The Release stage build, test the built packages, and copy then to a GCP bucket, let's move it to the parallel execution and rename it to `Packaging`, for the fails fast on that stage let's do it in another PR after thinking on the implications, it is not directly supported by the parallel step so we have to think how to implement it..

simitt:
IMO this should only be moved to the parallel test phase when there is some kind of failing fast enabled. 
For a regular PR most commonly tests fail at the _system_test_ part. If moving the `packaging` also to the parallel tests we will always have to wait until those are also done, before the whole test is labeled as failed. I don't think this would be an improvement.

v1v:
So, I'll close this PR for the time being and raise an issue with the requirements to see whether we can apply another approach.

PR #2985, opened 2019-11-29
Upgrade go version to 1.13.4

simitt:
Following up on https://github.com/elastic/apm-server/pull/2984, where libbeat go version was upgraded to 1.13.4.


changelogs/head.asciidoc
```diff
@@ -14,3 +14,4 @@ https://github.com/elastic/apm-server/compare/7.5\...master[View commits]
 - Use go-elasticsearch client for fetching sourcemaps from Elasticsearch {pull}2897[2897].
 - Try to extract IP address from headers before using socket remote_address for `client.ip` and `source.ip` {pull}2935[2935].
 - Only use extracted hostname when valid IP for enriching events {pull}2935[2935].
+- Upgrade Go to 1.13.4 {pull}2984[2984].
```
jalvz: I'm not sure this deserves a changelog entry. I mean, most of the times we don't add it, at least the last updates.

simitt: I did a quick grep for go upgrade changelog entries and found the last one in the `7.5` changelog. I think adding this one is consistent. But the main motivation was that this is a minor version upgrade and not just a patch version upgrade. 

Btw, I expect us to not backport this change to 7.x, we usually only backport patch version upgrades in minor upgrades.

PR #2988, opened 2019-11-29
ci(jenkins): add the update beats stage

kuisathaverat:
## What does this PR do?

Add a new stage that runs the update-beats script.

## Why is it important?

This would help to detect libbeat incompatibilities in early stages. It is part of the parallel test process so it should not add time to the build.

**Notes**
this step was enabled in the past, it was disabled because it never success until a week or so before the release and a few weeks after the release, so it would be marked as unstable when it fails and the build would still a success, the GitHub check will show that there is a problem.

Screenshot 2019-11-29 at 13 14 42

Screenshot 2019-11-29 at 15 55 26

## Related issues
Closes #13371

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -36,6 +36,7 @@ pipeline {
     booleanParam(name: 'bench_ci', defaultValue: true, description: 'Enable benchmarks')
     booleanParam(name: 'release_ci', defaultValue: true, description: 'Enable build the release packages')
     booleanParam(name: 'kibana_update_ci', defaultValue: true, description: 'Enable build the Check kibana Obj. Updated')
+    booleanParam(name: 'beats_update_ci', defaultValue: true, description: 'Test Beats update')
     booleanParam(name: 'its_ci', defaultValue: true, description: 'Enable async ITs')
   }
   stages {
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -328,24 +329,40 @@ pipeline {
...
+                  sh(label: 'Update Beats script', script: './script/jenkins/update-beats.sh')
                 }
               }
-        }*/
+            }
+          }
+          post {
+            always {
+              catchError(buildResult: 'SUCCESS', message: 'Failed to grab test results tar files', stageResult: 'SUCCESS') {
+                tar(file: "update-betas-system-tests-linux-files.tgz", archive: true, dir: "system-tests", pathPrefix: "${BASE_DIR}/build")
```
mdelapenya: Is there a typo here?

```diff
- update-betas-system-tests
+ update-beats-system-tests
```

kuisathaverat:
```suggestion
                tar(file: "update-beats-system-tests-linux-files.tgz", archive: true, dir: "system-tests", pathPrefix: "${BASE_DIR}/build")
```

graphaelli:
This seems to be in the wrong in ci?  We want to know when changes to libbeat will break apm-server, the other way around doesn't seem useful as it shouldn't gate changes in apm-server.

kuisathaverat:
>This seems to be in the wrong in ci? We want to know when changes to libbeat will break apm-server, the other way around doesn't seem useful as it shouldn't gate changes in apm-server.

this stage is designed to warning about problems, it does not mark the build as failed, it only mark its stage as unstable, so you check it with every merge to master if we move the thing to beats-ci would make that you have to go to the Beats-ci to check if the job is broken, indeed the job exists and seems it is not useful.

graphaelli:
The problem https://github.com/elastic/beats/issues/13371 is trying to address is that changes are made in libbeat that end up breaking apm-server without the beats developer knowing it.

It's simply not useful to flag this on the apm-server side at arbitrary times - we'll know the beats update breaks apm-server when we attempt that beats update - it might even be fixed between when it is flagged here and when the beats update happens.

I'm -1 on this change.

kuisathaverat:
In that case, the correct change it's the one that @v1v open in August https://github.com/elastic/infra/pull/14132

kuisathaverat:
we close this one in favor of elastic/infra#14132

PR #2994, opened 2019-12-02
Update beats framework to fa63445

### LINKED ISSUE #2858: Produce Docker images for RedHat Universal Base Image
**Describe the enhancement:**
This feature request is a request to produce APM Server images based on the [RedHat Universal Base Image (UBI)](https://www.redhat.com/en/blog/introducing-red-hat-universal-base-image).

**Describe a specific use case for the enhancement or feature:**
Producing APM Server images based on the official RHEL UBI would be beneficial for anyone using the official RHEL UBI for the base image of their docker containers.

graphaelli:
To pull in https://github.com/elastic/beats/pull/14796, for building UBI based images.

Closes #2858

### DIFF
--- a/_beats/dev-tools/packaging/packages.yml
+++ b/_beats/dev-tools/packaging/packages.yml
@@ -173,6 +173,11 @@ shared:
     extra_vars:
       repository: 'docker.elastic.co/beats'
 
+  - &docker_ubi_spec
+    extra_vars:
+      image_name: '{{.BeatName}}-ubi7'
+      from: 'registry.access.redhat.com/ubi7/ubi:7.7'
+
   #
   # License modifiers for Apache 2.0
   #
@@ -335,6 +340,14 @@ specs:
       types: [docker]
       spec:
         <<: *docker_spec
+        <<: *elastic_docker_spec
+        <<: *elastic_license_for_binaries
+
+    - os: linux
+      types: [docker]
+      spec:
+        <<: *docker_spec
+        <<: *docker_ubi_spec
         <<: *elastic_docker_spec
         <<: *elastic_license_for_binaries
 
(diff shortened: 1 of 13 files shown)


_beats/dev-tools/packaging/packages.yml
```diff
@@ -173,6 +173,11 @@ shared:
     extra_vars:
       repository: 'docker.elastic.co/beats'
 
+  - &docker_ubi_spec
+    extra_vars:
+      image_name: '{{.BeatName}}-ubi7'
+      from: 'registry.access.redhat.com/ubi7/ubi:7.7'
```
jmlrt: Hi @graphaelli,
The base image to use is `registry.access.redhat.com/ubi7/ubi-minimal:7.7` unless there is technical issue. Did you try with this one before using `registry.access.redhat.com/ubi7/ubi:7.7`?

graphaelli: @mikemadden42 mentioned that ubi-minimal was missing some things we need like `useradd` and `groupadd`, anything else worth mentioning?

jmlrt: Logstash and Elasticsearch are using minimal image which is our requirement so I guess you can use useradd and groupadd or have a workaround if needed. However Beats should switch to minimal image for 7.6.0. If you're dependant of beats image for APM, that's fine as long as you can switch when Beats image will switch.

graphaelli: Since beats needs to do the same, we can make the change there and then pull it into apm-server.  Elasticsearch simply installs `shadow-utils` per https://github.com/elastic/elasticsearch/blob/8c2dda90c0f01c7aa92ab82f0329e7b273c14fae/distribution/docker/src/docker/Dockerfile#L16-L19.  @mikemadden42 mind making the changes needed?

PR #3003, opened 2019-12-04
Add mage packaging build and type filters

graphaelli:
eg , build just docker images using existing binaries with:

```
SKIP_BUILD=true SNAPSHOT=true PLATFORMS="linux/amd64" TYPES=docker mage -v package
```

### DIFF
--- a/magefile.go
+++ b/magefile.go
@@ -22,16 +22,19 @@ package main
 import (
 	"context"
 	"fmt"
+	"log"
 	"os"
 	"path/filepath"
+	"strings"
 	"time"
 
 	"github.com/magefile/mage/mg"
 	"github.com/magefile/mage/sh"
 	"github.com/pkg/errors"
 
-	"github.com/elastic/apm-server/beater/config"
 	"github.com/elastic/beats/dev-tools/mage"
+
+	"github.com/elastic/apm-server/beater/config"
 )
 
 func init() {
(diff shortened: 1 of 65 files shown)


magefile.go
```diff
@@ -115,18 +118,53 @@ func dockerConfigFileParams() mage.ConfigFileParams {
...
+func filterPackages(types string) {
+	var packages []mage.OSPackageArgs
+	keep := keepPackages(strings.Split(types, " "))
+	for _, p := range mage.Packages {
+		for _, t := range p.Types {
+			if _, ok := keep[t]; !ok {
+				continue
+			}
+			packages = append(packages, p)
+			break
```
jalvz: why the `break`? can't keep 2 types of the same package or am i reading wrong?

graphaelli:
> maybe add the example of your comment above to the readme, or as a code comment

Something other than https://github.com/elastic/apm-server/pull/3003/files#diff-3db01cb68c42f27f6422c65fa1175e23R151-R152?

magefile.go
graphaelli: right, there are many types per package but you want just one package once the type matches.  The consequence is that, in our case, if you pick `rpm` you also get `deb`.  I thought about filtering that down further but it didn't seem work it when really all we're after for now is 'build docker images' or 'build linux distro packages'

graphaelli:
transient errors pulling from github, merging

graphaelli:
fyi @andrewkroh

Issue #2987, opened 2019-11-29
Provide a setup command to create API keys


jalvz:
At minimum apm-server should be able to create API keys and apm privileges, and invalidate them.

Optionally/future: 
- list existing keys and privileges 
- validate keys (`has_privileges` call) 
- batch creation (eg. for several services)
- reference docs

jalvz:
Depends on: https://github.com/elastic/apm-server/issues/2338

simitt:
## The following behavior was tested and works as expected:

**Testing configuration settings**
* API Key command works even when `apm-server.api_key.enabled: false`: Running `./apm-server apikey create -E apm-server.api_key.enabled=false` still creates api keys.
* Defaulting back to configured `output.elasticsearch.*` if no dedicated elasticsearch is configured for API Key
* Overwriting user configured in `output.elasticsearch.username` by passing in `apm-server.api_key.elasticsearch.username`

**Testing user privileges**
* Trying to create an API Key with a non-existing user results in a 401: `./apm-server apikey create -E apm-server.api_key.elasticsearch.username=foo -E apm-server.api_key.elasticsearch.password=changeme --json`
* Trying to create an API Key with an existing user, that does not have apm privileges results in an according error
* Trying to create an API Key with an existing user, that has APM privileges but not the `manage_security` privilege results in a `403`. 

**Testing API Key `create`**
* Creating API Key without any information returns key valid for all APM privileges that never expires: `./apm-server apikey create`
* Creating API Key with dedicated name: `./apm-server apikey create --name test`
* Creating API Key with expiration date - API Key is not valid after expiration period: `./apm-server apikey create --expiration 1m` 
* Creating API Key only valid for certain APM privileges: 
```
./apm-server apikey create --name simi-apikey --ingest
```
`curl -i -H 'Authorization: ApiKey createdApiKey' "http://localhost:8200/config/v1/agents?service.name=123"` returns `401` 
but
`curl -i -k -H "Content-type: application/x-ndjson" -H "Authorization: ApiKey dy1tdzBtOEI5NnZwcjk5aTdYX2E6cUd6b1pxQ0FTT2ladlVmVWduOTI0UQ" --data-binary @events.ndjson http://localhost:8200/intake/v2/events` returns `202` 

**Testing API Key `info`**
* `./apm-server apikey info --name simi-apikey --json` returns information about all API Keys with this name
* `./apm-server apikey info --id foo --json` returns information about API Key with this ID
[...]

PR #3004, opened 2019-12-04
Add API Key support

### LINKED ISSUE #2338: Rely on Elasticsearch API Keys instead of secret token for authentication

simitt:
With this PR support is added for authorizing requests between APM
agents and the Server via API keys. The API keys need to have privileges
for the generic resource "_".

Used `privileges`:
* `/config/v1/agents`:  `config_agent:read`
* `/intake/v2/events`:  `event:write`
* `/assets/v1/sourcemaps`:  `sourcemap:write`
* `/intake/v2/profile`:  `event:write` (theoretical as not supported yet)
* `/`: any of the above

Used `application`: `apm`

Used `resource`: `-`

API Keys for this implementation phase should be of format:
```
application: "apm"
privileges: "*" OR "config_agent:read","event:write", "sourcemap:write"`
resources: "*"
```

closes elastic/apm-server#2338

---
The main purpose for opening the PR without having integration tests is to 
* provide information for #2987 for the `application`, `privileges` and `resources` as early as possible
* get a code review on time 

What will follow is 
* [ ] Integration tests (depending on https://github.com/elastic/apm-server/pull/2799)
* [x] Changelog entry
* [ ] more manual testing, once there is also a branch available for #2987 to test them together.

---

# Rough outline of Test scenarios

## Default (disabled API Key)
Requests are accepted, independent of auth header (test with Bearer, with API Key, with random Auth method and without Auth header) 

## Set Bearer token via `apm-server.secret_token`
* Requests to backend endpoints with valid Bearer auth header are accepted, all others are denied
* Requests to RUM endpoint do not need authorization, if an auth header is sent, it is not used

### Secret token is not set
* Requests to backend endpoint with valid API Key auth header are accepted, all others are denied
* Requests to RUM endpoint do not need authorization, if an auth header is sent, it is not used

### Also set secret token 
* Requests to backend endpoint with valid API Key auth header or valid Bearer header are accepted, all others are denied
* Requests to RUM endpoint do not need authorization, if an auth header is sent, it is not used

### Change Limit `apm-server.api_key.limit`
Check that limit is respected: requests with new API Keys are not processed, requests with used API Keys are processed unless cached key expires and cache is filled up with other API Keys

### Test Elasticsearch
* test with configured `apm-server.api_key.elasticsearch`
* test with configured `output.elasticsearch`
* test when both ES config options are set, configuration from `apm-server.api_key.elasticsearch` should be used
* test that configured `username` and `password` is not used for actual request to ES (as the privileges need to be derived from the API Key)

#### API Key valid for request to any of the backend endpoints
```
{
  "application":[
    {
      "application": "apm",
      "privileges": ["*"],
      "resources": ["*"]
    }
  ]
}
```

```
{
  "application":[
    {
      "application": "*",
      "privileges": ["*"],
      "resources": ["*"]
    }
  ]
}
```

```
{
  "application":[
    {
      "application": "apm",
      "privileges": ["config_agent:read","sourcemap:write","event:write"],
      "resources": ["*"]
    }
  ]
}
```

```
{
  "application":[
    {
      "application": "apm",
      "privileges": ["*"],
      "resources": ["-"]
    }
  ]
}
```

```
{
  "application":[
    {
      "application": "apm",
      "privileges": ["config_agent:read","sourcemap:write","event:write"],
      "resources": ["-"]
    }
  ]
}
```

#### API Key valid for request to agent config + root
(resource can be "*" or "-")

```
{
  "application":[
    {
      "application": "apm",
      "privileges": ["config_agent:read""],
      "resources": ["*"]
    }
  ]
}
```

#### API Key valid for request to intake API + root
(resource can be "*" or "-")
```
{
  "application":[
    {
      "application": "apm",
      "privileges": ["event:write"],
      "resources": ["*"]
    }
  ]
}
```

#### API Key valid for request to sourcemap + root
(resource can be "*" or "-")
```
{
  "application":[
    {
      "application": "apm",
      "privileges": ["sourcemap:write"],
      "resources": ["*"]
    }
  ]
}
```

#### API Key invalid for request to backend endpoints
```
{
  "application":[
    {
      "application": "apm",
      "privileges": ["*"],
      "resources": ["service_a"]
    }
  ]
}
```

```
{
  "application":[
    {
      "application": "apm",
      "privileges": ["config_agent:read","sourcemap:write","event:write"],
      "resources": ["service_a"]
    }
  ]
}
```

```
{
  "application":[
    {
      "application": "apm",
      "privileges": ["foo"],
      "resources": ["*"]
    }
  ]
}
```

```
{
  "application":[
    {
      "application": "foo",
      "privileges": ["*"],
      "resources": ["*"]
    }
  ]
}
```

### DIFF
--- a/beater/authorization/apikey.go
+++ b/beater/authorization/apikey.go
@@ -1,0 +1,149 @@
+// Licensed to Elasticsearch B.V. under one or more contributor
+// license agreements. See the NOTICE file distributed with
+// this work for additional information regarding copyright
+// ownership. Elasticsearch B.V. licenses this file to you under
+// the Apache License, Version 2.0 (the "License"); you may
+// not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package authorization
+
+import (
+	"encoding/json"
+	"net/http"
+	"strings"
+	"time"
+
+	"github.com/pkg/errors"
+
+	"github.com/elastic/apm-server/beater/headers"
+	"github.com/elastic/apm-server/elasticsearch"
+)
+
+const (
+	application = "apm"
+	sep         = `","`
+
+	cleanupInterval = 60 * time.Second
+)
+
+type apikeyBuilder struct {
+	esClient        elasticsearch.Client
+	cache           *privilegesCache
+	anyOfPrivileges []string
+}
+
+type apikeyAuth struct {
+	*apikeyBuilder
+	key string
+}
+
+type hasPrivilegesResponse struct {
+	Applications map[string]map[string]privileges `json:"application"`
+}
+
+func newApikeyBuilder(client elasticsearch.Client, cache *privilegesCache, anyOfPrivileges []string) *apikeyBuilder {
+	return &apikeyBuilder{client, cache, anyOfPrivileges}
+}
+
+func (a *apikeyBuilder) forKey(key string) *apikeyAuth {
+	return &apikeyAuth{a, key}
+}
+
+// IsAuthorizationConfigured will return true if a non-empty token is required.
+func (a *apikeyAuth) IsAuthorizationConfigured() bool {
+	return true
+}
+
+// AuthorizedFor checks if the configured api key is authorized.
+// An api key is considered to be authorized when the api key has the configured privileges for the requested resource.
+// Privileges are fetched from Elasticsearch and then cached in a global cache.
+func (a *apikeyAuth) AuthorizedFor(resource string) (bool, error) {
+	//fetch from cache
+	if allowed, found := a.fromCache(resource); found {
+		return allowed, nil
+	}
+
+	if a.cache.isFull() {
+		return false, errors.New("api_key limit reached, " +
+			"check your logs for failed authorization attempts " +
+			"or consider increasing config option `apm-server.api_key.limit`")
+	}
+
+	//fetch from ES
+	privileges, err := a.queryES(resource)
+	if err != nil {
+		return false, err
+	}
+	//add to cache
+	a.cache.add(id(a.key, resource), privileges)
+
+	allowed, _ := a.fromCache(resource)
+	return allowed, nil
+}
+
+func (a *apikeyAuth) fromCache(resource string) (allowed bool, found bool) {
+	privileges := a.cache.get(id(a.key, resource))
+	if privileges == nil {
+		return
+	}
+	found = true
+	allowed = false
+	for _, privilege := range a.anyOfPrivileges {
+		if privilegeAllowed, ok := privileges[privilege]; ok && privilegeAllowed {
+			allowed = true
+			return
+		}
+	}
+	return
+}
+
+func (a *apikeyAuth) queryES(resource string) (privileges, error) {
+	query := buildQuery(PrivilegesAll, resource)
+	statusCode, body, err := a.esClient.SecurityHasPrivilegesRequest(strings.NewReader(query),
+		http.Header{headers.Authorization: []string{headers.APIKey + " " + a.key}})
+	if err != nil {
+		return nil, err
+	}
+	defer body.Close()
+	if statusCode != http.StatusOK {
+		// return nil privileges for queried apps to ensure they are cached
+		return privileges{}, nil
+	}
+
+	var decodedResponse hasPrivilegesResponse
+	if err := json.NewDecoder(body).Decode(&decodedResponse); err != nil {
+		return nil, err
+	}
+	if resources, ok := decodedResponse.Applications[application]; ok {
+		if privileges, ok := resources[resource]; ok {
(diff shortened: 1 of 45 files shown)


beater/config/api_key.go
```diff
@@ -0,0 +1,65 @@
...
+func (c *APIKeyConfig) setup(log *logp.Logger, outputESCfg *common.Config) error {
+	if c == nil || !c.Enabled || c.ESConfig != nil {
+		return nil
+	}
+	c.ESConfig = elasticsearch.DefaultConfig()
+	if outputESCfg == nil {
+		return nil
+	}
+	log.Info("Falling back to elasticsearch output for API Key usage")
+	if err := outputESCfg.Unpack(c.ESConfig); err != nil {
```
jalvz: `outputESCfg` comes from `output.elasticsearch`, right? 
From yml doc ("API keys need to be fetched from Elasticsearch. If nothing is configured, configuration settings from the output section will be reused"), I was expecting to find here an attempt to unpack `api_key.elasticsearch` and fallback to the output if that matches the default: is that happening somewhere else, or is that logic implemented differently?

beater/request/result.go
```diff
@@ -88,7 +88,7 @@ var (
 		IDResponseErrorsInvalidQuery:       {Code: http.StatusBadRequest, Keyword: "invalid query"},
 		IDResponseErrorsDecode:             {Code: http.StatusBadRequest, Keyword: "data decoding error"},
 		IDResponseErrorsValidate:           {Code: http.StatusBadRequest, Keyword: "data validation error"},
-		IDResponseErrorsMethodNotAllowed:   {Code: http.StatusMethodNotAllowed, Keyword: "only POST requests are supported"},
+		IDResponseErrorsMethodNotAllowed:   {Code: http.StatusMethodNotAllowed, Keyword: "method not supported"},
```
jalvz: similarly to this, L85 https://github.com/elastic/apm-server/pull/3004/files#diff-f5ea74122372bd20b6b1d9543d5123e6R85 message could be replaced with a "Not authorized" or similar, since we will be using tokens and/or keys

beater/authorization/apikey.go
```diff
@@ -0,0 +1,149 @@
+// Licensed to Elasticsearch B.V. under one or more contributor
```
jalvz: since we will also need a bunch of `apikey` functionality for the subcommand, it might make sense to move this file to a top level `apikey` package.

beater/request/context.go
```diff
@@ -55,11 +57,11 @@ type Context struct {
 func (c *Context) Reset(w http.ResponseWriter, r *http.Request) {
 	c.Request = r
 	c.Logger = nil
-	c.TokenSet = false
-	c.Authorized = false
-	c.IsRum = false
 	c.RateLimiter = nil
+	c.Authorization = &authorization.AllowAuth{}
```
jalvz: trying to connect the dots: does https://github.com/elastic/apm-server/pull/3004/files#diff-3f6ee316627b52f99975b03690d74fb1R49 need to be run after this for every request? is that happening?

beater/config/api_key.go
simitt: In [L45](https://github.com/elastic/apm-server/pull/3004/files/ee039e923d345bcfcab0564c17653ec42d677d42#diff-a6cfe6c325e26db069b78efc06eff45aR45) I check if `c.ESConfig != nil`, that means the `output.elasticsearch` config is only unpacked if it is not already set.
Setting `APIKeyConfig.ESConfig` happens automatically (same as for all other unpacked values), see [test](https://github.com/elastic/apm-server/pull/3004/files/ee039e923d345bcfcab0564c17653ec42d677d42#diff-47b43fcaa2c5dad1a43178f36b7a0245R153)

beater/request/context.go
simitt: No, [`handler.AuthorizedFor`](https://github.com/elastic/apm-server/pull/3004/files#diff-3f6ee316627b52f99975b03690d74fb1R89) is called for every request, as the actual auth method and the api_key can only be determined on a per request base.

You see the call in the [middleware.authorization](https://github.com/elastic/apm-server/pull/3004/files#diff-854a750f79bb08e98ddc18772ff72be1R34)

beater/authorization/apikey.go
simitt: Not sure why that would be a reason to move it to top level. I assume you are adding the tooling to the `cmd` package or as a sub-package `cmd/apikey`. There is no harm in importing the `beater` package from there (it is already imported for the other commands).

simitt:
I manually tested the main scenarios, and they worked as expected.

PR #3014, opened 2019-12-05
Update system tests

graphaelli:
A collection of testing updates extracted from #2799.  The most substantial change introduced makes tests wait for pipelines to load before running.  This uncovered a race in one test that is skipped for now.

### DIFF
--- a/tests/system/apmserver.py
+++ b/tests/system/apmserver.py
@@ -105,6 +105,14 @@ class ServerSetUpBaseTest(BaseTest):
         self.apmserver_proc = self.start_beat(**self.start_args())
         self.wait_until_started()
 
+        # try make sure APM Server is fully up
+        cfg = self.config()
+        # pipeline registration is enabled by default and only happens if the output is elasticsearch
+        if not getattr(self, "register_pipeline_disabled", False) and \
+            cfg.get("elasticsearch_host") and \
+            cfg.get("register_pipeline_enabled") != "false" and cfg.get("register_pipeline_overwrite") != "false":
+            self.wait_until_pipelines_registered()
+
     def start_args(self):
         return {}
 
--- a/tests/system/apmserver.py
+++ b/tests/system/apmserver.py
@@ -196,20 +196,28 @@ class ElasticTest(ServerBaseTest):
         self.kibana_url = self.get_kibana_url()
 
         # Cleanup index and template first
-        self.es.indices.delete(index="apm*", ignore=[400, 404])
-        for idx in self.indices:
-            self.wait_until(lambda: not self.es.indices.exists(idx))
+        if self.es.indices.get("apm*"):
+            self.es.indices.delete(index="apm*", ignore=[400, 404])
+            for idx in self.indices:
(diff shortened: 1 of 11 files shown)


tests/system/apmserver.py
```diff
@@ -181,20 +200,28 @@ def setUp(self):
...
-            self.wait_until(lambda: not self.es.indices.exists(idx))
-
-        self.es.indices.delete_template(name="apm*", ignore=[400, 404])
-        for idx in self.indices:
-            self.wait_until(lambda: not self.es.indices.exists_template(idx))
+        if self.es.indices.get("apm*"):
+            self.es.indices.delete(index="apm*", ignore=[400, 404])
+            for idx in self.indices:
+                self.wait_until(lambda: not self.es.indices.exists(idx), name="index {} to be deleted".format(idx))
+        assert all(idx.startswith("apm") for idx in self.indices), "not all indices prefixed with apm, cleanup assumption broken"
```
simitt: Should this assertion be moved up before the actual `delete` action?

tests/system/apmserver.py
```diff
@@ -181,20 +200,28 @@ def setUp(self):
...
 
         # truncate, don't delete agent configuration index since it's only created when kibana starts up
-        self.es.delete_by_query(self.index_acm, {"query": {"match_all": {}}},
-                                ignore_unavailable=True, wait_for_completion=True)
-        self.wait_until(lambda: self.es.count(index=self.index_acm, ignore_unavailable=True)["count"] == 0)
+        if self.es.count(index=self.index_acm, ignore_unavailable=True)["count"] > 0:
+            self.es.delete_by_query(self.index_acm, {"query": {"match_all": {}}},
+                                    ignore_unavailable=True, wait_for_completion=True)
+            self.wait_until(lambda: self.es.count(index=self.index_acm, ignore_unavailable=True)["count"] == 0,
+                            max_timeout=30, name="acm index {} to be empty".format(self.index_acm))
```
simitt: It might speed up the tests if we put all the delete actions first and then put all the `wait_until`?

tests/system/apmserver.py
```diff
@@ -14,8 +15,9 @@
 
 sys.path.append(os.path.join(os.path.dirname(__file__), '..',
                              '..', '_beats', 'libbeat', 'tests', 'system'))
-from beat.beat import TestCase, TimeoutError
+from beat.beat import INTEGRATION_TESTS, TestCase, TimeoutError
 
+integration_test = unittest.skipUnless(INTEGRATION_TESTS, "integration test")
```
simitt: 💯

tests/system/apmserver.py
graphaelli: Good idea

tests/system/apmserver.py
graphaelli: I hadn't considered that.  I'd like some data to support that change before we do it, as I think we'd tradeoff debugging simplicity (everything in lockstep) with that change

simitt: That's a good point, haven't considered that. Let's keep it as is then.

graphaelli:
The test failure is surprising but unrelated to this change.

PR #3018, opened 2019-12-09
ci(jenkins): workaround gitCheckout issues

v1v:
## Current issue

As far as I see the shallow cloning is disabled in the Jenkinsfile but the logs are not saying so:

image

Likely the JJBB is forcing that particular behavior, what it seems awkward. 

## What does this PR do?

Workaround the current issues when cloning the repo, as the JJBB seems to get more precedence.

## Why is it important?

Fix the pipeline in the master branch.

### DIFF
--- a/.ci/jobs/apm-server-mbp.yml
+++ b/.ci/jobs/apm-server-mbp.yml
@@ -30,8 +30,8 @@
           after: true
           before: true
         prune: true
-        shallow-clone: true
-        depth: 3
+        shallow-clone: false
+        depth: 100
         do-not-fetch-tags: true
         submodule:
           disable: false
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -12,6 +12,7 @@ pipeline {
     CODECOV_SECRET = 'secret/apm-team/ci/apm-server-codecov'
     GITHUB_CHECK_ITS_NAME = 'APM Integration Tests'
     ITS_PIPELINE = 'apm-integration-tests-selector-mbp/master'
+    PIPELINE_LOG_LEVEL = 'DEBUG'
   }
   options {
     timeout(time: 2, unit: 'HOURS')


.ci/jobs/apm-server-mbp.yml
```diff
@@ -30,8 +30,8 @@
           after: true
           before: true
         prune: true
-        shallow-clone: true
-        depth: 3
+        shallow-clone: false
+        depth: 100
```
v1v: Let's grab the last 100 commits which should be sufficient for the time being.

Jenkinsfile
```diff
@@ -12,6 +12,7 @@ pipeline {
     CODECOV_SECRET = 'secret/apm-team/ci/apm-server-codecov'
     GITHUB_CHECK_ITS_NAME = 'APM Integration Tests'
     ITS_PIPELINE = 'apm-integration-tests-selector-mbp/master'
+    PIPELINE_LOG_LEVEL = 'DEBUG'
```
v1v: Add debug logs for the shared library, let's keep them for the time being.

simitt:
Can you please update to latest master version.

.ci/jobs/apm-server-mbp.yml
kuisathaverat: this should only affect the default checkout

kuisathaverat:
>Can you please update to latest master version.

the build always makes a merge with the master is there are changes.

kuisathaverat:
I think that error about `git diff` command it is not related to shallow cloning, it is the git ref `origin/632afadbf0811b4d7280ef279a28799feb8184c1` it is wrong, the following command will fail

`git diff --name-only origin/632afadbf0811b4d7280ef279a28799feb8184c1...6c03abc5d9bd523c6ad75dd525c1d021ba39b4c4`

and the following one will return the correct list of files changed

`git diff --name-only 632afadbf0811b4d7280ef279a28799feb8184c1...6c03abc5d9bd523c6ad75dd525c1d021ba39b4c4`

kuisathaverat:
we can close this one I've made a PR to the library to fix the issue.

simitt:
> we can close this one I've made a PR to the library to fix the issue.

Would you mind linking to it, so we can follow along with the change.

kuisathaverat:
it is linked from the observability-dev repo

v1v:
This quick workaround is not required anymore

PR #3019, opened 2019-12-09
Add API Key Integration tests

simitt:
Add system integration tests for API Key authorization to test
privileges, resources and application for API Keys usage.

related to elastic/apm-server#2338


beater/authorization/apikey.go
```diff
@@ -30,6 +30,9 @@ import (
 )
 
 const (
+	//DefaultResource for apm backend enabled API Keys
+	DefaultResource = "-"
```
graphaelli: is this our convention or something special for elasticsearch?

tests/system/apmserver.py
```diff
@@ -59,7 +60,7 @@ def get_elasticsearch_url(user="", password=""):
         host = os.getenv("ES_HOST", "localhost")
 
         if not user:
-            user = os.getenv("ES_USER", "elastic")
+            user = os.getenv("ES_USER", "admin")
```
graphaelli: do we rely on this fallback?

tests/system/apmserver.py
```diff
@@ -167,7 +171,7 @@ def wait_until_ilm_setup(self):
 
     def wait_until_pipelines_registered(self):
         self.wait_until(lambda: self.log_contains("Registered Ingest Pipelines successfully"),
-                                                  name="pipelines registered")
+                        name="pipelines registered")
```
graphaelli: we should sync the team on these settings.  I like the way this looks better but our automation doesn't seem to flag these and my editor wants to change them back.

tests/system/test_access.py
```diff
@@ -5,7 +5,11 @@
 import os
 import subprocess
 import shutil
-import unittest
+import json
+import random
+import string
+import base64
```
graphaelli: sort new imports

tests/system/test_access.py
```diff
@@ -16,44 +20,316 @@
...
+        name = "apm-{}".format(random_str)
+        payload = json.dumps({
+            "name": name,
+            "role_descriptors": {
+                name+"role_desc": {
+                    "applications": [
+                        {"application": application, "privileges": privileges, "resources": resources}]}}})
+        resp = requests.post(url, data=payload, headers=headers(content_type='application/json'))
+        assert resp.status_code == 200, resp.status_code
+        cls.created_api_keys += [resp.json()["id"]]
```
graphaelli:
```suggestion
        cls.created_api_keys.append(resp.json()["id"])
```

tests/system/test_access.py
```diff
@@ -16,44 +20,316 @@
...
+        payload = json.dumps({"application": [{
+            "application": cls.application,
+            "privileges": [cls.privilege_config, cls.privilege_sourcemap, cls.privilege_intake],
+            "resources": cls.resource_backend}]})
+        resp = requests.post(url_user_privileges, data=payload, headers=headers(content_type=content_type))
+        assert resp.status_code == 200, resp.status_code
+        assert "has_all_requested" in resp.json(), resp.content
+        assert resp.json()["has_all_requested"] == True, resp.json()
+
+        cls.created_api_keys = []
```
graphaelli: Why is this class level?  It appears to be deleted one at a time in `tearDownClass` - I would expect it as instance level and handled via `setUp` and `tearDown` instead

tests/system/test_access.py
```diff
@@ -16,44 +20,316 @@
...
+
+    def test_agent_config(self):
+        """
+        Test authorization logic for backend Agent Configuration endpoint
+        """
+        url = self.agent_config_url
+
+        for token in self.authorized_keys+[self.api_key_privilege_config]:
+            resp = requests.get(url, headers=headers(token, content_type="application/json"))
+            assert resp.status_code != 401,  "token: {}, status_code: {}".format(token, resp.status_code)
```
graphaelli: Why is this `! 401`?  Will these be `404`?  Does it make sense to create configs and ensure they are returned?

tests/system/test_access.py
```diff
@@ -16,44 +20,316 @@
...
+                                       })
+            return resp
+
+        for token in self.unauthorized_keys+[self.api_key_privilege_config, self.api_key_privilege_intake]:
+            resp = upload(token)
+            assert resp.status_code == 401, "token: {}, status_code: {}".format(token, resp.status_code)
+
+        for token in self.authorized_keys+[self.api_key_privilege_sourcemap]:
+            resp = upload(token)
+            assert resp.status_code != 401, "token: {}, status_code: {}".format(token, resp.status_code)
```
graphaelli: These should 202?

beater/authorization/apikey.go
simitt: This is our convention. At the moment the API keys can be created with `resource: *` , but as soon as we support service specific API Keys, this resource needs to be enabled for the early key validation.

tests/system/apmserver.py
simitt: No, but in the APM Server setup as well as in the Integration testing we create a superuser `admin` and not `elastic`. For local testing, it is therefore more convenient to use this, as we don't have to explicitly set the env variable then.

tests/system/apmserver.py
simitt: Oh, I was sure it got auto-changed when I ran `make fmt`.

tests/system/test_access.py
simitt: The API Keys are immutable and do not need to change between the tests. Therefore I decided to create them once upfront, use in all tests and then destroy them again. 

What would be the advantage of creating/deleting on an instance level for this?

tests/system/test_access.py
simitt: I did not intend to test the actual functionality of the endpoint here as it adds more coupling. Currently only the middleware checks for the API Key token, so I think the current approach is efficient. Once we also move a service name based check inside the handlers, it probably makes sense to also test handler functionality.

tests/system/test_access.py
simitt: If RUM is enabled. But I can change to that.

tests/system/apmserver.py
simitt: re-checked, running `make fmt` does change to what you see in the PR.

tests/system/test_access.py
simitt: The `Authorization middleware` runs _before_ the `KillSwitchHandler middleware`. Without enabling `RUM` this would result in a `403` as it hits the kill switch _after_ the authorization. 

I can still enable the RUM endpoints and ensure test runs through.

simitt:
@graphaelli I followed along with your suggested changes, please have another look.

graphaelli:
jenkins, run the tests please

PR #3024, opened 2019-12-10
Wait until agent config index is created 

jalvz:
This is needed because with Kibana NP that index is created asyncrously.


jalvz:
jenkins retest this please

PR #3028, opened 2019-12-10
capture es diagnostics per DIAGNOSTIC_INTERVAL

graphaelli:
Extracted from experiments in #3003.  This disables idx mgmt setup teardown as a result of that investigation.


graphaelli:
The single test failure seems flaky #3030 opened to track.

PR #3031, opened 2019-12-11
Retry connection when Kibana version is not compatible.

### LINKED ISSUE #3022: [Cloud] Agent configuration does not work when upgrading to 7.5.0
When using APM in Elastic Cloud, upgrading from an older deployment (in my case 7.4.1) to 7.5.0, the server's agent configuration endpoint continues to return error responses:

```
{
  "error": "not a compatible Kibana version: min version 7.5.0, configured version 7.4.1"
}
```

Forcefully restarting the APM Server fixes this.

### Steps to reproduce:

1. Create a 7.4.1 deployment
2. Upgrade deployment to 7.5.0
3. Upgrade APM to 7.5.0
4. Upgrade Kibana to 7.5.0
5. `curl -H "Authorization: Bearer <secret-token>" https://<apm-server>/config/v1/agents`

jalvz:
Fixes #3022

### DIFF
--- a/kibana/connecting_client.go
+++ b/kibana/connecting_client.go
@@ -50,14 +50,14 @@ type Client interface {
 	// Connected indicates whether or not a connection to Kibana has been established
 	Connected() bool
 	// SupportsVersion compares given version to version of connected Kibana instance
-	SupportsVersion(v *common.Version) (bool, error)
+	SupportsVersion(*common.Version, bool) (bool, error)
 }
 
 // ConnectingClient implements Client interface
 type ConnectingClient struct {
 	client *kibana.Client
 	cfg    *common.Config
-	m      sync.Mutex
+	m      sync.RWMutex
 }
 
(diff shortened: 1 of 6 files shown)


jalvz:
This is a temporary fix for 7.5.1
Will be moved to libbeat as per https://github.com/elastic/beats/issues/14495

kibana/connecting_client.go
```diff
@@ -98,22 +99,44 @@ func (c *ConnectingClient) Send(method, extraPath string, params url.Values,
...
 		return common.Version{}, errNotConnected
 	}
 	return c.client.GetVersion(), nil
 }
 
 // Connected checks if a connection has been established
-func (c *ConnectingClient) Connected() bool { return c.client != nil }
+func (c *ConnectingClient) Connected() bool {
+	c.m.Lock()
+	defer c.m.Unlock()
```
axw: this should probably also be RLock/RUnlock, as c.client is only read

PR #3040, opened 2019-12-12
[7.x] Retry connection when Kibana version is not compatible. (#3031)

jalvz:
 - Retry connection when Kibana version is not compatible.  (#3031)


simitt:
Merging this in as I assume it should be part of `7.6`.

PR #3062, opened 2019-12-16
[7.x] Add span.context.destination to intake (#3053)

axw:
 - Add span.context.destination to intake  (#3053)


axw:
jenkins run the tests please

PR #3066, opened 2019-12-17
[Jaeger] Add otel consumer converting batches to Elastic APM events

simitt:
Add consumer converting incoming otel batches to Elastic APM format.
Add integration tests covering incoming gRPC requests being transformed to beat events.

partially implements #3307

This PR does not include error handling.


processor/otel/consumer.go
```diff
@@ -58,3 +74,437 @@ func (c *Consumer) ConsumeTraceData(ctx context.Context, td consumerdata.TraceDa
...
+		k := replaceDots(kDots)
+		switch v := v.Value.(type) {
+		case *tracepb.AttributeValue_BoolValue:
+			utility.DeepUpdate(labels, k, v.BoolValue)
+		case *tracepb.AttributeValue_DoubleValue:
+			utility.DeepUpdate(labels, k, v.DoubleValue)
+		case *tracepb.AttributeValue_IntValue:
+			switch kDots {
+			case "http.status_code":
+				//v.IntValue is int64, considering the semantics of the field converting to an int should be fine
```
axw: I think it's safe to remove this comment, code seems pretty logical to me and we do the same thing in several other places.

processor/otel/consumer.go
```diff
@@ -58,3 +74,437 @@ func (c *Consumer) ConsumeTraceData(ctx context.Context, td consumerdata.TraceDa
...
+				http.URL = &url
+				isHTTPSpan = true
+			case "http.method":
+				method := truncate(v.StringValue.Value)
+				http.Method = &method
+				isHTTPSpan = true
+			case "sql.query":
+				// NOTE(axw) in theory we could override
+				// the span name here with our own method
+				// based on the SQL query.
```
axw: I thought about this and later decided that we should just keep whatever span name is given to us. If you agree, let's just delete the note.

processor/otel/consumer.go
```diff
@@ -58,3 +74,437 @@ func (c *Consumer) ConsumeTraceData(ctx context.Context, td consumerdata.TraceDa
...
+				isDBSpan = true
+			case "db.user":
+				db.UserName = &v.StringValue.Value
+				isDBSpan = true
+			case "peer.address":
+				val := truncate(v.StringValue.Value)
+				destination.Address = &val
+			case "peer.service":
+				val := truncate(v.StringValue.Value)
+				event.DestinationService = &model_span.DestinationService{Name: &val}
```
axw: We'll have to make sure this doesn't break service maps, because at the moment the agents are required to send all or none of `destination.service.{name,type,resource}`.

processor/otel/consumer.go
```diff
@@ -58,3 +74,437 @@ func (c *Consumer) ConsumeTraceData(ctx context.Context, td consumerdata.TraceDa
...
+	for key, val := range td.Resource.GetLabels() {
+		md.Labels[key] = truncate(val)
+	}
+}
+
+func parseTransaction(span *tracepb.Span, hostname string, event *transaction.Event) {
+	labels := make(common.MapStr)
+
+	if count := span.GetChildSpanCount(); count != nil {
+		labels["span_count.started"] = count.Value
```
axw: I'm not sure if this is a good idea, as I believe the semantics are different. IIANM, `span_count.started` is meant to record how many (child, grandchild, etc.) spans were started within this transaction, whereas Jaeger/OpenTelemetry just tells us the number of direct child spans started.

_Maybe_ we could redefine `span_count.started` to be a lower bound on the number of started spans.

processor/otel/consumer.go
```diff
@@ -58,3 +74,437 @@ func (c *Consumer) ConsumeTraceData(ctx context.Context, td consumerdata.TraceDa
...
+		subtype := "http"
+		event.Subtype = &subtype
+		event.HTTP = &http
+	case isDBSpan:
+		event.Type = "db"
+		event.Subtype = db.Type
+		event.DB = &db
+	default:
+		event.Type = "custom"
+		event.Subtype = &component
```
axw: component might be "", should this be conditional?

processor/otel/consumer.go
simitt: I only added it as labels, but can remove overall to avoid confusion.

processor/otel/consumer.go
simitt: Good point, I will remove the logic and only add to labels, as it doesn't make sense to store the information here when not usable for service maps.

beater/api/jaeger/testdata/batch_1.approved.json
```diff
@@ -0,0 +1,627 @@
...
+            "trace": {
+                "id": "000000000000000065b4115c7ebe6014"
+            }
+        },
+        {
+            "@timestamp": "2019-12-17T14:22:30.559111+01:00",
+            "agent": {
+                "ephemeral_id": "78dd6e017f54a2c1",
+                "name": "Jaeger",
+                "version": "Go-2.20.1"
```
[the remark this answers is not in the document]
simitt: oh yes I actually forgot to update this behavior - I went through most of the jaeger-tracing repositories and the values are all of the form `<language>-<version>`, which leads to `Jaeger-<language>-<version>` after the otel conversion. Changed this now.

beater/api/jaeger/testdata/batch_1.approved.json
```diff
@@ -0,0 +1,627 @@
...
+                "sampler_param": true,
+                "sampler_type": "const"
+            },
+            "processor": {
+                "event": "transaction",
+                "name": "transaction"
+            },
+            "service": {
+                "language": {
+                    "name": "\u003cunknown\u003e"
```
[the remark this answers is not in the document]
simitt: I changed to `unknown` in alignment with what the Java agent sets when values are missing (the only agent from which I knew they were setting unknown values).

simitt:
@axw this is ready for another review.

processor/otel/consumer.go
axw: Oops, I didn't read that properly. I think removing it's the right way to go.

beater/api/jaeger/testdata/batch_1.approved.json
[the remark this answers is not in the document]
axw: I see now that the OpenTelemetry translator code doesn't extract the language from the `jaeger.version` tag. I think we should do this ourselves for now, and then upstream it.

processor/otel/consumer.go
```diff
@@ -58,3 +74,430 @@ func (c *Consumer) ConsumeTraceData(ctx context.Context, td consumerdata.TraceDa
...
+		// version is of format `Jaeger-<agentlanguage>-<version>`, e.g. `Jaeger-Go-2.20.0`
+		versionParts := strings.SplitN(td.Node.GetLibraryInfo().GetExporterVersion(), "-", 3)
+		if v := versionParts[len(versionParts)-1]; v != "" {
+			md.Service.Agent.Version = &v
+		} else {
+			jaegerVersion := "unknown"
+			md.Service.Agent.Version = &jaegerVersion
+		}
+		agentName := agentNameJaeger
+		if md.Service.Language.Name != nil {
```
axw: The OTel Jaeger translator doesn't set `td.Node.Language`, so I think we should do that ourselves while parsing the version above. Then we can upstream the code. WDYT?

simitt: SGTM, will add.

PR #3067, opened 2019-12-17
Ignore "should have comment" golint errors

jalvz:
Lets put the responsibility of good documentation on the team instead of a linter and show we can do a good job.


graphaelli:
`golint` does a lot more than just enforcing documentation of public methods.  I'm ok with ignoring just those but -1 on removing automated `golint` from ci.

jalvz:
Good idea @simitt.

From https://github.com/golang/lint

> The suggestions made by golint are exactly that: suggestions. Golint is not perfect, and has both false positives and false negatives. Do not treat its output as a gold standard. We will not be adding pragmas or other knobs to suppress specific warnings, so do not expect or require code to be completely "lint-free". In short, this tool is not, and **will never be, trustworthy enough for its suggestions to be enforced automatically**, for example as part of a build process

Bolds mine. A pre-commit hook would be a good alternative for whoever wants to run it, for me the IDE does a pretty good job already. WDYT @graphaelli ?

axw:
What prompted this PR? In general I think it's good to have linting in CI, so my initial reaction is -1; but presumably it's causing you some pain. What is that? Is there another way we can address it?

jalvz:
This is forcing us bad documentation practices: most of times, public types/methods/functions are trivial, implementation of interfaces, or simply need to be (un)marshalled. I truly hate bad/obvious/redundant documentation. 

In practical terms, the linter only ever fails on documentation, so we don't get any other value out of it.
Given we never had issues with code style ("style" as understood by the linter, nevertheless) and documentation suggestions are more misses than hits, let me ask again:

Why a pre-commit hook or similar is not good enough? If style won't be perfect with or without linter, why engineer best judgement can't be trusted?

graphaelli:
>  trustworthy enough for its suggestions to be enforced automatically, for example as part of a build process

True, but we're relying on it as a gate, not automatically changing code.  We've been running it on CI since June 2019 and I'm just not seeing the pain - the gains in consistent method receivers, error returns, use of `range`, etc all seem valuable to me.

Again, I think we should just disable that one check and not the entire linter.

jalvz:
> True, but we're relying on it as a gate

Well, I guess I don't see the difference between a "gate" and an enforcement. Amended.

simitt:
Can we please not force-push on PRs (at least on non-draft PRs), especially when it has discussion/review going on. All the history is lost and it is hard to impossible to follow along after a force push.

jalvz:
@elastic/apm-server can this be reviewed again?

PR #3083, opened 2019-12-19
beater: disable auth for self-instrumentation

### LINKED ISSUE #3051: Enabling API Key without additional secret token breaks self instrumentation
When using self instrumentation via the elastic go-apm-agent the agent can either talk to a remote server or to localhost via a in-process listener.
When the APM Server is configured to require a secret token, the agent uses the configured token for the server communication. With introducing API Key support the server requires a valid API Key if no additional secret token is configured. 

We can either require the user to configure a valid API Key within the APM Server instrumentation configuration and use that as soon as the agent supports API Keys or investigate if we can disable the API Key requirement for the loopback listener.

axw:
When the server is configured to send trace data
to itself, it will now create an entirely separate
muxer and HTTP server for the purpose.

This tracer muxer is configured to not require any
auth, and is not instrumented, so we can simplify
the "doNotTrace" request ignorer.

Closes #3051

### DIFF
--- a/beater/beater.go
+++ b/beater/beater.go
@@ -24,12 +24,10 @@ import (
 	"net"
 	"net/http"
 	"net/url"
-	"os"
 	"sync"
 	"time"
 
 	"go.elastic.co/apm"
-	"go.elastic.co/apm/transport"
 	"golang.org/x/sync/errgroup"
 
 	"github.com/elastic/beats/libbeat/beat"
@@ -41,7 +39,6 @@ import (
 	"github.com/elastic/apm-server/beater/config"
 	"github.com/elastic/apm-server/ingest/pipeline"
 	logs "github.com/elastic/apm-server/log"
-	"github.com/elastic/apm-server/pipelistener"
 	"github.com/elastic/apm-server/publish"
 )
 
@@ -50,11 +47,12 @@ func init() {
 }
 
 type beater struct {
-	config  *config.Config
-	mutex   sync.Mutex // guards server and stopped
-	server  *http.Server
-	stopped bool
-	logger  *logp.Logger
+	config   *config.Config
+	mutex    sync.Mutex // guards server and stopped
+	server   *http.Server
+	stopping chan struct{}
+	stopped  bool
+	logger   *logp.Logger
 }
 
 var (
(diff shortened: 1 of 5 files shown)


tests/system/test_access.py
```diff
@@ -359,6 +360,36 @@ def upload(token):
...
+class TestSelfInstrumentationWithAPIKeys(BaseAPIKeySetup):
+
+    def setUp(self):
+        super(TestSelfInstrumentationWithAPIKeys, self).setUp()
+        self.api_key = self.create_api_key([self.privilege_intake], self.resource_any)
+
+    def config(self):
+        cfg = super(TestSelfInstrumentationWithAPIKeys, self).config()
+        cfg.update({"api_key_enabled": True, "enable_rum": True,
+                    "kibana_enabled": "true", "kibana_host": self.get_kibana_url(),
```
simitt: You only test the event ingestion, setting `enable_rum`, `kibana_enabled` or `kibana_host`  should not be needed here.

tests/system/test_access.py
```diff
@@ -259,7 +259,8 @@ def setUp(self):
     def config(self):
         cfg = super(TestAccessWithAuthorization, self).config()
         cfg.update({"secret_token": "1234", "api_key_enabled": True, "enable_rum": True,
-                    "kibana_enabled": "true", "kibana_host": self.get_kibana_url()})
+                    "kibana_enabled": "true", "kibana_host": self.get_kibana_url(),
+                    "instrumentation_enabled": "true"})
```
simitt: Why did you enable instrumentation here (you are not checking anything related to it afaics).

beater/beater.go
```diff
@@ -159,14 +158,17 @@ func (bt *beater) listen() (net.Listener, error) {
...
+	tracer, tracerServer, err := initTracer(b.Info, bt.config, bt.logger)
 	if err != nil {
 		return err
 	}
-	if traceListener != nil {
-		defer traceListener.Close()
+	if tracerServer != nil {
+		go func() {
+			defer tracerServer.stop()
+			defer tracer.Close()
```
simitt: Do I understand this correct that the `tracer` is now only closed for in-process listener (== `tracerServer!= nil`)? Could you explain why?

beater/beater.go
```diff
@@ -159,14 +158,17 @@ func (bt *beater) listen() (net.Listener, error) {
...
+			<-bt.stopping
```
simitt: nice!

tests/system/test_access.py
axw: Copy-pasta :sweat_smile:
Thanks, fixed.

tests/system/test_access.py
axw: I started updating this suite, and then decided to create a new suite - just forgot to revert this, which I've done now. In case you're wondering: I wanted the test to fail without the changes to beater, but it would have passed on account of the secret_token set in this suite's config.

beater/beater.go
axw: Oops, that was a mistake - thanks, fixed.

PR #3085, opened 2019-12-20
[jaeger] Convert Timeevents to errors

### LINKED ISSUE #3007: [Jaeger] Implement event processor
Implement transformation of Jaeger events to Elastic APM events. Validate input and decide on a strategy for invalid input: figure out which attributes are required for UI and how to fill them if missing, or log event and drop it. 

Part of https://github.com/elastic/apm-server/issues/2886

Starting point: implementation from https://github.com/elastic/apm-server/compare/7.x...axw:otel-7.x

simitt:
Parse TimeEvents from Jaeger spans and convert to elastic error events if they describe an error. Set as much contextual information as available from transactions and spans. 

This PR changes the test input files in the beater package to also contain errors as TimeEvents. To avoid any manual merge issue and reflect realistic data the whole testdata files were replaced.

Fixes #3007


simitt:
jenkins run the tests please

processor/otel/consumer.go
```diff
@@ -408,6 +418,71 @@ func parseSpan(span *tracepb.Span, event *model_span.Event) {
...
+				switch v := v.Value.(type) {
+				case *tracepb.AttributeValue_StringValue:
+					vStr := v.StringValue.Value
+					switch k {
+					case "event":
+						logMessage = vStr
+					case "level":
+						isError = vStr == "error"
+					case "error":
+						exMessage = vStr
```
axw: should this also set `isError`? or is `level` always present?

processor/otel/consumer.go
```diff
@@ -408,6 +418,71 @@ func parseSpan(span *tracepb.Span, event *model_span.Event) {
...
+	var errors []*model_error.Event
+	for _, log := range otelSpan.GetTimeEvents().GetTimeEvent() {
+		var isError bool
+		var logMessage, exMessage string
+		for k, v := range log.GetAnnotation().GetAttributes().GetAttributeMap() {
+			if source == sourceFormatJaeger {
+				switch v := v.Value.(type) {
+				case *tracepb.AttributeValue_StringValue:
+					vStr := v.StringValue.Value
+					switch k {
```
axw: what about "message"?
https://github.com/opentracing/specification/blob/master/semantic_conventions.md#log-fields-table

processor/otel/consumer.go
axw: Maybe also "error.kind" for exception.type?

processor/otel/consumer.go
axw: Maybe also "error.kind" for exception.type?

simitt: I handled all the fields that were set in the examples I saw. But I will give it another thought of how to make this more comprehensive.

processor/otel/consumer.go
simitt: In the examples I saw it was always present, but doesn't hurt to also set in this case.

processor/otel/consumer.go
axw: This is good enough for me now, I think we could extend it later.

processor/otel/consumer.go
```diff
@@ -408,6 +418,93 @@ func parseSpan(span *tracepb.Span, event *model_span.Event) {
...
+						hasMinimalInfo = true
+						isError = true
+					case "level":
+						isError = vStr == "error"
+					}
+				}
+			}
+		}
+		if !isError || !hasMinimalInfo {
+			continue
```
axw: I'm a little bit concerned that the "!hasMinimalInfo" case will be hard to debug if we don't know the attributes. Maybe we should log (debug or trace level) when we encounter such a record?

Issue #2580, opened 2019-08-19
Ambiguous kafka and logstash config documentation


jalvz:
In https://www.elastic.co/guide/en/apm/server/master/kafka-output.html and https://www.elastic.co/guide/en/apm/server/master/logstash-output.html (for every release version) it says that the default value for `enabled` is true, which it is ambiguous at best.

AFAIK those `enabled` are true "by default" only **if** (the whole `output.elasticsearch` sections is commented out **or** `output.elasticsearch.enabled` is false) **and** the corresponding section is uncommented (`output.kafka` or `output.logstash`). 
Since by default that is not the case, you can't really say that `output.kafka.enabled` and `output.logstash.enabled` are on by default.

IMO we should be explicit and document it as I said above, or event better just mention that  kafka / logstash `enabled` settings are disabled by default and add `enabled: true` in the example snippets.

bmorelli25:
Closed in #3093.

PR #3093, opened 2019-12-30
[docs] Fix ambiguous *.output documentation

bmorelli25:
For https://github.com/elastic/apm-server/issues/2580.

Requires a PR in the Beats repo to persist the changes: https://github.com/elastic/beats/pull/15272.

---

This PR adds a paragraph to each output page explaining that the es output must be disabled first:
```
To use this output, edit the {beatname_uc} configuration file to disable the {es}
output by commenting it out...
```

It also clarifies that for apm-server, Logstash and Kafka `enabled` defaults to `false`. This is not the case for beats:
```
ifndef::apm-server[]
The default value is `true`.
endif::[]
ifdef::apm-server[]
The default value is `false`.
endif::[]
```

Finally, there is no redis, file, or console section in the `.yml` configuration files, so I've fixed wording that indicated there was.

### DIFF
--- a/docs/copied-from-beats/outputs/console/docs/console.asciidoc
+++ b/docs/copied-from-beats/outputs/console/docs/console.asciidoc
@@ -7,6 +7,11 @@
 
 The Console output writes events in JSON format to stdout.
 
+To use this output, edit the {beatname_uc} configuration file to disable the {es}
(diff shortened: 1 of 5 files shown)


bmorelli25:
Going to merge this. Let me know if you see anything after the fact that needs fixing.

PR #3094, opened 2019-12-30
[docs] Match Beats directory structure for copied docs

bmorelli25:
Beats is moving certain documentation to live closer to its code. See https://github.com/elastic/beats/pull/14724 as an example.

This breaks `copy-docs.py`, because the current script expects the APM Server `copied-from-beats/` docs to mimic the directory structure in Beats' `libbeat/docs/`.

We could write a new script with different logic to copy Beats documentation over, but there are documentation files in different directories with the same name, which complicates things. I think the easiest, and probably best solution, would just be to continue to match the beats directory structure in `copied-from-beats`. This has the added benefit of simplifying the `libbeat` attributes we use -- in most cases, we'll be able to use the Beats attributes as is.

This PR is _mostly_ just moved files. There are a few changes that were necessary to keep the docs building.

A visualization:

Before:
```
docs
├── copied-from-beats
│   ├── monitoring
│   │   └── *.asciidoc
│   ├── outputs
│   │   └── *.asciidoc
│   ├── security
│   │   └── *.asciidoc
│   └── *.asciidoc
```

After:
```
docs
├── copied-from-beats
│   ├── docs
│   │   ├── monitoring
│   │   │   └── *.asciidoc
│   │   ├── outputs
│   │   │   └── *.asciidoc
│   │   ├── security
│   │   │   └── *.asciidoc
│   │   ├── *.asciidoc
│   ├── outputs
│   │   ├── es
│   │   ├── logstash
│   │   └── ...
```


docs/copied-from-beats/docs/outputs-list.asciidoc
```diff
@@ -0,0 +1,38 @@
...
+
+include::{libbeat-outputs-dir}/redis/docs/redis.asciidoc[]
+
+include::{libbeat-outputs-dir}/fileout/docs/fileout.asciidoc[]
+
+include::{libbeat-outputs-dir}/console/docs/console.asciidoc[]
+
+include::outputs/output-cloud.asciidoc[]
+
+//REVIEWERS: Should the cloud ID topic live somewhere under libbeat/outputs. If
```
bmorelli25: You can ignore these comments, they're from https://github.com/elastic/beats/pull/14724.

dedemorton:
> The structure is a little bit confusing, e.g. docs/copied-from-beats/docs/outputs and docs/copied-from-beats/outputs/es/docs/..

Agreed. Now that the folder contains a single file, it doesn't make sense to have a `docs/outputs` folder under libbeat. I will open a PR to remove it to avoid confusion in the future.

PR #3095, opened 2019-12-31
intake: Add rows_affected to span db information

### LINKED ISSUE #2802: Add `rows_affected` for database queries in spans
* [ ] Add `span.context.db.rows_affected` to Intake API as optional field
* [ ] Clarify if field needs to be indexed.
* [ ] Decide where to store the field in ES. There is an issue in [ECS](https://github.com/elastic/ecs/issues/514) to standardize on sql fields.

simitt:
As layed out in https://github.com/elastic/apm/issues/112 we want to include the number of rows affected by a sql database statement manipulating db data in the available database information of a span.

~As this is not currently used in any aggregation, and described as additional information to be shown to users, the field is not indexed. @SergeyKleyman please let me know if you had in mind that this field should be search or aggregatable.~

Update: the new field is indexed as [Elasticsearch long](https://www.elastic.co/guide/en/elasticsearch/reference/current/number.html#number), according to https://github.com/elastic/apm-server/pull/3095#issuecomment-570137434.

closes #2802


SergeyKleyman:
@simitt Thank you for letting me know. Regarding your question:
>  if you had in mind that this field should be search or aggregatable.

I think this field should be searchable and aggregatable to allow users for example to find out how many DB rows transactions affect on average and which transactions affect more DB rows than 95th percentile, etc.

PR #3096, opened 2019-12-31
intake: Add classname as stacktrace frame attribute

### LINKED ISSUE #3006: Add `classname` to stack trace frame fields
As discussed in https://github.com/elastic/kibana/issues/49467 the server should accept `classname` attributes on `stacktrace frames` and require agents to either send a `classname` or a `filename`.

Agent devs need to be aware that stop sending `filename` information is breaking with older versions of the APM Server.

simitt:
The current format of stacktrace frames does not well reflect native stacktraces for all languages. Especially .NET and Java are missing a `classname` attribute. More detailed discussions can be found at https://github.com/elastic/kibana/issues/49467. This PR adds support for `classname` as a property of a stacktrace frame.

**Intake API:**
Adds support of `classname` as a valid property of a stacktrace frame to the Intake API. Requirements are changed, so that now either a `classname` or  a `filename` is required for a stacktrace frame.

**Calculating Error Grouping Key:**
In case only `classname` is given, it will be used for calculating the grouping key, otherwise the calculation is not changed. 

**Source mapping**
The RUM agent does not make use of `classname`, no changes are required for source mapping logic.

closes #3006


model/error/event.go
```diff
@@ -266,7 +266,12 @@ func (e *Event) updateCulprit(tctx *transform.Context) {
 	if fr == nil {
 		return
 	}
-	culprit := fmt.Sprintf("%v", fr.Filename)
+	var culprit string
+	if fr.Filename != nil {
+		culprit = fmt.Sprintf("%v", *fr.Filename)
```
axw:
```suggestion
		culprit = *fr.Filename
```

model/error/event.go
```diff
@@ -266,7 +266,12 @@ func (e *Event) updateCulprit(tctx *transform.Context) {
...
+	} else if fr.Classname != nil {
+		culprit = fmt.Sprintf("%v", *fr.Classname)
```
axw:
```suggestion
		culprit = *fr.Classname
```

model/error/event_test.go
```diff
@@ -791,13 +809,16 @@ func TestCulprit(t *testing.T) {
 		},
 	}
 	for idx, test := range tests {
-		tctx := &transform.Context{
-			Config: test.config,
-		}
+		t.Run(string(idx), func(t *testing.T) {
```
axw: `string(idx)` is going to create strings with runes '\u0000', '\u0001', etc. I guess you meant `fmt.Sprint(idx)`?

Issue #2708, opened 2019-09-13
APM Server monitoring using apm_system role fails


axw:
**Describe the bug**

The [Monitoring APM Server](https://www.elastic.co/guide/en/apm/server/7.2/monitoring.html) section in the docs describes how to configure the server to send monitoring data to Elasticsearch, and recommends using the built-in `apm_system` user/role.

In 7.2 this fails with an error in the logs like this:

> Failed to publish events: 403 Forbidden: {"error":{"root_cause":[{"type":"security_exception","reason":"action [indices:data/write/bulk] is unauthorized for user [apm_system]"}],"type":"security_exception","reason":"action [indices:data/write/bulk] is unauthorized for user [apm_system]"},"status":403}

**To Reproduce**

1. Create a 7.2 stack with security enabled. Set a password for the `apm_system` user
2. Configure apm-server.yml according to https://www.elastic.co/guide/en/apm/server/7.2/monitoring.html, using the password from step 1.
3. Start the APM Server
4. Check the logs, see there's an error

**Expected behavior**

Monitoring should work using the apm_system user/role.

axw:
This appears to be a regression in 7.2, relating to changes in the underlying Beats infrastructure.

The apm_system role has the same cluster-level privileges as beats_system, but the latter has some additional index-level privileges. These were added to beats_system in https://github.com/elastic/elasticsearch/pull/40876; apm_system should presumably have been updated at the same time.

cachedout:
I have opened https://github.com/elastic/elasticsearch/pull/47302 to start work on this issue.

simitt:
@cachedout can you confirm this issue was fixed with https://github.com/elastic/elasticsearch/pull/47302#issuecomment-539020207?

cachedout:
@simitt I am still waiting for reviews on this: https://github.com/elastic/elasticsearch/pull/47668

cachedout:
Hi again @simitt and @axw.

I have merged https://github.com/elastic/elasticsearch/pull/47668 into master in Elasticsearch.

Backports have been opened to [7.5](https://github.com/elastic/elasticsearch/pull/49913) and [7.x](https://github.com/elastic/elasticsearch/pull/49912). Please let me know if you have any concerns. Thanks!

simitt:
LGTM

axw:
Backports are all done, so closing this out. Thanks @cachedout!

simitt:
[...]

PR #3111, opened 2020-01-07
tests/system: basic smoke-test for monitoring

axw:
Add a system test that verifies the server can connect to the X-Pack Monitoring endpoint using the built-in "apm_system" user.

See https://github.com/elastic/apm-server/issues/2708

### DIFF
--- a/tests/system/test_monitoring.py
+++ b/tests/system/test_monitoring.py
@@ -1,0 +1,40 @@
+from apmserver import integration_test
+from apmserver import ElasticTest
+
+import urlparse
+
+
+@integration_test
+class Test(ElasticTest):
+    def config(self):
+        cfg = super(Test, self).config()
+
+        # Extract username/password from the URL, and store
+        # separately from elasticsearch_host. This is necessary
+        # because Beats will use the userinfo from the URL in
+        # preference to specified values.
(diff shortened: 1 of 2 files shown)


tests/system/test_monitoring.py
```diff
@@ -0,0 +1,40 @@
...
+        if url.password:
+            cfg["elasticsearch_password"] = url.password
+        bare_netloc = url.netloc.split('@')[-1]
+        url = list(url)
+        url[1] = bare_netloc
+
+        # Set a password for the built-in apm_system user, and use that for monitoring.
+        monitoring_password = "changeme"
+        self.es.xpack.security.change_password(username="apm_system",
+                                               body='{"password":"%s"}' % monitoring_password)
```
simitt: By default tests are run with a non-admin user. This request probably requires admin rights. You could run it using something along the lines of
```
admin_user = os.getenv("ES_SUPERUSER_USER", "admin")
admin_password = os.getenv("ES_SUPERUSER_PASS", "changeme")
admin_es = Elasticsearch([self.get_elasticsearch_url(admin_user, admin_password)])
```

axw: Thanks, fixed.

PR #3114, opened 2020-01-08
tests/system: system test for Jaeger Thrift/HTTP

### LINKED ISSUE #3010: [Jaeger] Add system tests for http thrift integration with Jaeger agent
Add system tests sending spans via a Jaeger client (http) to the APM Server.

See if we can leverage python opentelemetry-api for the tests. Be aware that it is only supported for python 3.

Part of elastic/apm-server#2886

axw:
Add a system test that exercises the Jaeger HTTP endpoint. Sends a pre-recorded (Thrift-encoded) payload using requests, similar to what we do for the intake API.

Closes #3010


axw:
Test data generated with this:

```go
package main

import (
        opentracing "github.com/opentracing/opentracing-go"

        jaegercfg "github.com/uber/jaeger-client-go/config"
        jaegerlog "github.com/uber/jaeger-client-go/log"
)

func main() {
        cfg, err := jaegercfg.FromEnv()
        if err != nil {
                panic(err)
        }
        cfg.Sampler = &jaegercfg.SamplerConfig{Type: "const", Param: 1}

        logger := jaegerlog.StdLogger
        closer, err := cfg.InitGlobalTracer("test_service", jaegercfg.Logger(logger))
        if err != nil {
                panic(err)
        }
        defer closer.Close()

        tracer := opentracing.GlobalTracer()
        tracer.StartSpan("test_span").Finish()
}
```

axw:
jenkins run the tests please

tests/system/apmserver.py
```diff
@@ -164,9 +164,14 @@ class ServerSetUpBaseTest(BaseTest):
     sourcemap_url = "{}/{}".format(host, 'assets/v1/sourcemaps')
     expvar_url = "{}/{}".format(host, 'debug/vars')
 
+    jaeger_http_host = "localhost:14268"
```
simitt: `localhost:14268` is the default host. To test that host configuration is properly picked up by the APM Server I would configure a different port. 
For testing with default values, you can remove this.

tests/system/test_integration.py
```diff
@@ -102,6 +102,20 @@ def test_load_docs_with_template_and_add_error(self):
...
+    def test_jaeger_http(self):
+        """
+        This test sends a Jaeger span in Thrift encoding over HTTP, and verifies that it is indexed.
+        """
+        jaeger_span_thrift = self.get_testdata_path('jaeger', 'span.thrift')
+        self.load_docs_with_template(jaeger_span_thrift, self.jaeger_http_url, 'transaction', 1,
+                                     extra_headers={"content-type": "application/vnd.apache.thrift.binary"})
+        self.assert_no_logged_warnings()
+
+        # compare existing ES documents for errors with new ones
```
simitt: %s/errors/transactions

tests/system/test_integration.py
```diff
@@ -102,6 +102,20 @@ def test_load_docs_with_template_and_add_error(self):
 
         self.check_backend_error_sourcemap(self.index_error, count=4)
 
+    def test_jaeger_http(self):
```
simitt: What do you think about creating a dedicated `test_jaeger.py` file where we bundle all jaeger related tests? (Similar to how we separate jaeger from the Intake API in the go code).

axw: Yeah I was just being lazy, I'll do that.

tests/system/test_integration.py
axw: I'm just going to delete the comment if you don't have any objection. It's self-evident from the code IMO.

tests/system/apmserver.py
axw: The test setup now configures `apm-server.jaeger.http.host: "localhost:0"`, and parses the dynamically assigned port from the log file.

axw:
jenkins run the tests please

axw:
jenkins run the tests please

PR #3124, opened 2020-01-08
[test] Use apm_server_user for API Key tests.

simitt:
To run code with the least possible privileges, change
tests using API Keys for data ingestion to create the
API Keys with the apm_server_user instead of the superuser.


testing/docker/elasticsearch/roles.yml
```diff
@@ -1,8 +1,12 @@
 apm_server:
-  cluster: ["manage_ilm"]
+  cluster: ['manage_ilm','manage_security','manage_api_key']
   indices:
     - names: ['apm-*']
       privileges: ['write','create_index','manage','manage_ilm']
+  applications:
```
jalvz: why this?

simitt: This role is applied to the `apm_server_user`, which with these changes has the privileges necessary for API Key creation, so that this user can be used for the API Key tests, instead of the superuser.

jalvz: IIUC all needed for `apm_server_user` to create API keys are the `cluster` privileges added above, the `applications` section is not needed.

jalvz: ah, scratch that. it is

jalvz:
So, if I run this, I can't see the `apm_server_user` in http://localhost:5601/app/kibana#/management/security/users?_g=()
Why is that?

simitt:
Because we are using [file based roles](https://www.elastic.co/guide/en/elasticsearch/reference/current/defining-roles.html#roles-management-file) here. From the docs:
> You cannot view, edit, or remove any roles that are defined in roles.yml by using the role management UI or the role management APIs.

PR #3130, opened 2020-01-09
Merge jaeger feature branch into master

axw:


vendor/vendor.json
```diff
@@ -1873,11 +1915,36 @@
...
+		},
+		{
+			"checksumSHA1": "BAVJhQm6l/IS8B1zFSaaihkq+pc=",
 			"path": "github.com/gogo/protobuf/proto",
-			"revision": "43891d9df6604fb88ab1e69e33166f1634646446",
-			"revisionTime": "2020-01-02T12:26:20Z"
+			"revision": "5628607bb4c51c3157aacc3a50f0ab707582b805",
+			"revisionTime": "2019-10-14T06:15:17Z",
+			"version": "v1.3",
+			"versionExact": "v1.3.1"
```
simitt: We never discussed this before, but we are actually using a different `protobuf/proto` version now than used by `libbeat`. Not quite sure if this could cause some issues.

axw: Was that intentional? I can try rolling back to the beats vendored version.

simitt: afair there were conflicts, as some libraries needed to have a fixed version.

Issue #3137, opened 2020-01-10
Document output.elasticsearch.api_key option


simitt:
With https://github.com/elastic/beats/pull/14324 libbeat introduced API Key support when writing to Elasticsearch. APM Server inherits support, but is missing documentation about it.

@bmorelli25 I think this should also be documented in https://www.elastic.co/guide/en/apm/server/7.5/elasticsearch-output.html

simitt:
Checked the deployed API Key docs for version `7.6`, therefore applying `test-plan-ok` label.

PR #3138, opened 2020-01-10
Document output.elasticsearch.api_key in yml file

simitt:
related to #3137


simitt:
Good point!
* `monitoring.elasticsearch` was also updated in libbeat, so it is supposed to be supported. 
* `apm-server.sourcemap` automatically supports it via https://github.com/elastic/apm-server/blob/master/elasticsearch/config.go#L56.
* I didn't add it to `apm-server.api_key` - theoretically an api_key could be used to create api_keys, but I think documenting that would be more confusing than helping.

_meta/beat.yml
```diff
@@ -264,8 +264,11 @@ apm-server:
         # IPv6 addresses should always be defined as: `https://[2001:db8::1]:9200`.
         # hosts: ["localhost:9200"]
 
-        # Optional protocol and basic auth credentials.
+        # Protocol - either `http` (default) or `https`.
         #protocol: "https"
+
+        # Authentication credentials - either API key or username/password.
+        #api_key: "id:api_key"
```
graphaelli: is this the base64 encoded version or really two strings with colon?

simitt: It is really two strings with colon, the libbeat elasticsearch client takes care of the base64 encoding. 
We should be consistent with this behaviour also for sourcemaps, so I am changing the `elasticsearch/client` behavior accordingly. 
Unfortunately we currently do not have system tests for the case that an `apm-server.source_mapping.elasticsearch` is configured, created an issue for following up on that https://github.com/elastic/apm-server/issues/3166.

graphaelli: great, thanks for clarifying

PR #3157, opened 2020-01-14
Add apikey subcommand

axw:
Continuation of https://github.com/elastic/apm-server/pull/3063


testing/docker/elasticsearch/roles.yml
```diff
@@ -8,7 +8,7 @@ apm_server:
       privileges: ['sourcemap:write','event:write','config_agent:read']
       resources: '*'
 beats:
-  cluster: ['manage_index_templates','monitor','manage_ingest_pipelines','manage_ilm']
+  cluster: ['manage_index_templates','monitor','manage_ingest_pipelines','manage_ilm', 'manage_security','manage_api_key']
```
simitt: Why is adding the cluster roles to `beats` role necessary?

axw: It's used by a system test that tests what happens when the user can create API keys, but doesn't have the privilege themself.

I'm not sure if beats_user has those privileges normally (I didn't add that bit). If not, I think maybe we should create a user dynamically for this test. In a follow up, preferably.

PR #3175, opened 2020-01-15
[test] Improve system tests related to API Keys

simitt:
Following changes have been made to improve the system tests and reduce flakyness:

**Refactor setup of APIKeyBase class for sending requests against api_key secured APM Server**
* delete privileges and wait until they are deleted to ensure privileges for current user can succesfully be created
* use same name for privileges as API Key CLI does

**API Key setup command**
* unify terminology `apikey` and `api_key` within these tests
* use `apm_server_user` consistently: so far the default `ES_USER` or `admin` user was used to create privileges and api keys, but the test `test_info_by_id` expected it to be created for `apm_server_user`. This only worked if the `ES_USER` was set to `apm_server_user`, which is the case on CI but not when running `make system_tests` locally. By default tests should also work without setting env variables. 
* use `setUp` instead of `tearDown`: clean up is necessary before each test (e.g. there could be artifacts from other API Key tests that ran before), also previous flaky tests showed that the `tearDown` commands potentially impact the next tests.

**Flaky self-monitoring test**
* `wait_until` calls were missing `lambda: ` leading to immediately evaluating the function and passing in the return value to `wait_until` instead of passing in the function (thanks @beniwohli for additional explanations regarding `lambda`)
* move in-memory listener self-instrumentation test to the `test_instrumentation` file


tests/system/test_access.py
```diff
@@ -92,30 +92,36 @@ def setUp(self):
...
+        # clean setup:
+        # delete all existing api_keys of current user
         requests.delete(self.api_key_url,
                         data=json.dumps({'name': self.api_key_name}),
                         headers=headers(content_type='application/json'))
         self.wait_until(lambda: self.api_keys_invalidated(), name="delete former api keys")
+        # delete all existing application privileges to ensure they can be created for current user
+        for privilege in ["sourcemap", "intake", "config"]:
+            requests.delete("{}/{}/{}".format(self.privileges_url, self.application, privilege))
+        self.wait_until(lambda: self.privileges_deleted(self.application))
```
axw: privileges_deleted now returns a tuple, so won't this return immediately?

simitt: Ouch, yes that doesn't make sense.

tests/system/test_apikey.py
```diff
@@ -8,18 +8,20 @@
 
 
 class APIKeyBaseTest(BaseTest):
-    api_key_name = "apm_integration_key"
+    apikey_name = "apm_integration_key"
```
simitt: Have consistent naming `apikey` within one file

beniwohli:
@simitt to avoid the same problem happening again (forgetting to wrap the argument in a lambda), you could add a check here: https://github.com/elastic/apm-server/blob/2b02570bb37d70e40fdc0a1ccc9190789071253e/tests/system/apmserver.py#L269-L276

Something like

    assert callable(cond), "First argument of wait_until must be a function"

tests/system/test_access.py
```diff
@@ -73,59 +71,64 @@ def test_backend_intake(self):
...
+        payload = {self.application: {}}
+        for name, action in self.privileges.items():
+            payload[self.application][name] = {"actions": [action]}
+        resp = requests.put(self.privileges_url,
+                            data=json.dumps(payload),
+                            headers=headers(content_type=content_type))
         assert resp.status_code == 200, resp.status_code
-        assert "has_all_requested" in resp.json(), resp.content
-        assert resp.json()["has_all_requested"] == True, resp.json()
+        apm_privilges = resp.json()[self.application]
```
axw:
```suggestion
        apm_privileges = resp.json()[self.application]
```

tests/system/test_access.py
```diff
@@ -142,13 +145,13 @@ def api_keys_invalidated(self):
     def api_key_exists(self, id):
         resp = requests.get("{}?id={}".format(self.api_key_url, id))
         assert resp.status_code == 200, resp.status_code
-        return len(resp.json()["api_keys"]) == 1
+        return len(resp.json()["api_keys"]) == 1, resp.json()
```
axw: IIANM this one will also cause problems for wait_until

simitt:
Failing because of anothher known flaky test.

PR #3180, opened 2020-01-15
[7.6] Update beats framework to 1d0dab5

simitt:


axw:
`_beats/dev-tools/vendor/.keep` got deleted (by update_beats.sh?)
It's needed for generating fields.yml.

PR #3196, opened 2020-01-16
Fix script and update beats framework to 3ed9a54

simitt:


graphaelli:
jenkins, run the tests please

PR #3202, opened 2020-01-17
tests/system: attempt to speed up tests

axw:
- always set setup.template.settings:
  - index.number_of_replicas: 0
  - index.refresh_interval: 100ms
- remove explicit index refreshes
- remove sleeps


axw:
`test_export_config` needs fixing, but I'd like feedback before I invest any more time.

tests/system/config/apm-server.yml.j2
```diff
@@ -170,15 +170,14 @@ apm-server:
...
 setup.template.pattern: {{ override_template }}*
+{% endif %}
+
 setup.template.settings:
   index.number_of_shards: 1
   index.number_of_replicas: 0
-  index.refresh_interval: 1ms
-
-{% endif %}
+  index.refresh_interval: 100ms
```
jalvz: isn't this still too low?

axw: I set it to this because that's the default interval for checks in `wait_until`. I'll bump them both up to 250ms, and see how that fares.

axw: I did that and noticed that `wait_until` is sleeping after the condition is met, which is slowing down the tests quite a bit.

PR #3204, opened 2020-01-17
beater: rewrite flaky test

### LINKED ISSUE #3203: TestNotifyUpServerDown test is flaky
https://apm-ci.elastic.co/job/apm-server/job/apm-server-mbp/job/PR-3202/1/testReport/(root)/beater/Build_and_Test___Unit_Test___TestNotifyUpServerDown/

axw:
- Rewrite TestNotifyUpServerDown to use "setupServer", simplifying the test and removing an intermittent panic.
- Rename to TestOnboarding. I have no idea what the old name was meant to indicate.
- Rewrite onboarding to use publisher.

Closes #3203

### DIFF
--- a/beater/onboarding.go
+++ b/beater/onboarding.go
@@ -18,6 +18,7 @@
 package beater
 
 import (
+	"context"
 	"time"
 
 	"github.com/elastic/beats/libbeat/beat"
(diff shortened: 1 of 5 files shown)


beater/onboarding.go
```diff
@@ -26,16 +27,28 @@ import (
...
+	"github.com/elastic/apm-server/publish"
+	"github.com/elastic/apm-server/transform"
 )
 
-func notifyListening(config *config.Config, pubFct func(beat.Event)) {
+func notifyListening(ctx context.Context, config *config.Config, reporter publish.Reporter) {
 	logp.NewLogger(logs.Onboarding).Info("Publishing onboarding document")
-	event := beat.Event{
-		Timestamp: time.Now(),
+	reporter(ctx, publish.PendingReq{
```
jalvz: why not pass `context.Background()` directly to the `reporter` call instead of making it an argument of `notifyListening`?

axw: `notifyListening` is no longer called in the background (i.e. in its own goroutine). Passing it in from the caller means `notifyListening` doesn't make any assumptions about its calling context.

It doesn't matter much here as it's unlikely we'd ever change this code substantially, but in general it's best to stick to the pattern of instantiating a background context at the top of the stack, passing it down. Then you can add spans to a trace, for example.

PR #3206, opened 2020-01-17
[7.x] beater: rewrite flaky test (#3204)

axw:
 - beater: rewrite flaky test  (#3204)


simitt:
IMO this should also be backported to 7.6 since that's also an active branch.

PR #3217, opened 2020-01-20
Improve system tests

simitt:


tests/system/apmserver.py
```diff
@@ -155,26 +158,10 @@ def dump_diagnotics(self, interval=2):
...
-            except AttributeError as ex:
-                raise ex
-            except:
-                pass
-            time.sleep(poll_interval)
-        raise TimeoutError("Timeout waiting for '{}' to be true. ".format(name) +
-                           "Waited {} seconds.".format(max_timeout))
+        wait_until(cond, max_timeout=max_timeout, poll_interval=poll_interval, name=name)
+
+# If you implement a Subclass of ServerSetUpBaseTest ensure to implement a `tearDown` function taking care of
```
axw: We could also use https://docs.python.org/3/library/unittest.html#unittest.TestCase.addCleanup, which gets called regardless of whether `setUp` succeeds

tests/system/helper.py
```diff
@@ -0,0 +1,139 @@
+from datetime import datetime, timedelta
+import time
+from elasticsearch import Elasticsearch, NotFoundError
+from beat.beat import TimeoutError
+
+apm = "apm"
+apm_prefix = "{}*".format(apm)
+
+
+def wait_until(cond, max_timeout=10, poll_interval=0.25, name="cond"):
```
axw: :100:

tests/system/helper.py
```diff
@@ -0,0 +1,139 @@
...
+        except AttributeError as ex:
+            raise ex
+        except Exception as ex:
+            pass
+        time.sleep(poll_interval)
+    raise TimeoutError("Timeout waiting for '{}' to be true. ".format(name) +
+                       "Waited {} seconds.".format(max_timeout))
+
+
+def cleanup(es, indices, templates, policies, indices_truncate, skip_pipelines=True):
```
axw: I'd suggest inverting `skip_pipelines` to `clean_pipelines`, so we don't have no stinkin' double negatives. Maybe it's worth making them all optional, and all verbs? e.g.

```suggestion
def cleanup(es, delete_indices=[], delete_templates=[], delete_policies=[], truncate_indices=[], delete_pipelines=[apm_prefix]):
```

And then use keyword args everywhere in callers, so it's clear what the args are.

tests/system/apmserver.py
simitt: Working on such an alternative at the moment.

tests/system/apmserver.py
```diff
@@ -27,43 +30,47 @@ class BaseTest(TestCase):
...
+        self.diagnostic_thread.daemon = True
+        self.diagnostic_thread.start()
+
+    def cleanup_diagnostics(self):
+        self.running = False
+        self.diagnostic_thread.join(timeout=30)
+
+    def dump_diagnotics(self, interval=2):
+        while self.running:
+            time.sleep(interval)
```
axw: Can you please either replace this `time.sleep` with a [threading.Timer](https://docs.python.org/3/library/threading.html#timer-objects), or add a TODO? This existing code will slow tearDown, as the sleep won't be interrupted.

tests/system/test_requests.py
```diff
@@ -14,7 +14,6 @@
     from io import StringIO
 
 
-@integration_test
```
axw: why did you remove the annotations?

simitt: These tests do not require Elasticsearch or Kibana, therefore they do not need the annotation.

axw: I must be missing something. Don't they all use Elasticsearch by virtue of inheriting from ServerBaseTest and ClientSideBaseTest? Or will they pass even if Elasticsearch isn't running?

simitt: `Elasticsearch` is only created in `ElasticTest` which inherits from `ServerBaseTest`. So everything inheriting from `ElasticTest` and `ClientSideElasticTest` needs to be marked `integration_test`, while others don't.

tests/system/apmserver.py
simitt: Thanks for the suggestion. This code hasn't changed here, so I only added a TODO for now. Using `threading.Timer` would be a bit more complicated from what I saw and I'd like to move on with these changes.

PR #3218, opened 2020-01-21
Introduce elasticsearch.Error

axw:
Elasticsearch operations that use "doRequest" will now
return an *elasticsearch.Error if the server responds
with a status code > 299. The error type can be used
for special handling of certain status codes, such as
401 Unauthorized when performing _has_privileges checks.

The API Key auth code is updated to cache 401 responses
using this type and `errors.As`, checking for 401s.
Similarly, the apikey CLI subcommand now checks for 404s
specifically for ignoring failed privilege deletions.

Fixes https://github.com/elastic/apm-server/issues/3215

PR #3220, opened 2020-01-21
[7.x] Introduce elasticsearch.Error (#3218)

axw:
 - Introduce elasticsearch.Error  (#3218)


axw:
Unrelated system test failure.

PR #3241, opened 2020-01-23
[7.x] decrease API Key cache time to one minute (#3238)

simitt:
 - decrease API Key cache time to one minute (#3238)


simitt:
jenkins run the tests

Issue #3184, opened 2020-01-15
[self instrumentation] instrument Elasticsearch clients


axw:
Instrument the Elasticsearch clients using the Go agent's `module/apmelasticsearch`, so we can see how queries (e.g. for sourcemaps, and API Key auth) are performing.

We'll need to add context to the functions in the `apm-server/elasticsearch` package, and plumb context through from requests.

graphaelli:
Pretty excited about this, though obv limited initially since it's not used for bulk ingestion.  This could enable:
* calc of distributiuon es query duration and rate
* calc of distribution of es query request and response sizes
* response status codes
* es comm error reporting

PR #3254, opened 2020-01-27
Add dedicated event models to each handler

jalvz:
~Tests are not done yet, but I'd like to get a comment on whether this is the right way to go before spending much time on them.~

~This PR adds an  `EventModel` to each intake handler. An `EventModel` has a decoder function and a jsonschema validator. This will make possible in the future to have different decoders and validators for RUM and backend (eg: to override or not event timestamp, changing the schema, etc).~

~While at it, it also removes the context object containing the request time, metadata, and various configs: the request time can be added better during decoding, the metadata can be stored in the event, and configs can be passed directly from publisher to the `Transform`~


processor/stream/processor.go
```diff
@@ -82,43 +76,21 @@ func (s *srErrorWrapper) Read() (map[string]interface{}, error) {
 	return v, err
 }
 
+type EventModel struct {
+	Name   string
+	Schema *jsonschema.Schema
+	Decode func(interface{}, time.Time, metadata.Metadata) (transform.Transformable, error)
```
axw: Is there benefit in keeping these separate? Can we make the `Decode` function/method responsible for validation too? By doing that, and removing Name (see further comments), we could simplify this to an interface:

```go
type Decoder interface {
    Decode(interface{}, time.Time, metadata.Metadata) (transform.Transformable, error)
}
```

Doing this would enable the decoder to optimise validation/decoding to avoid duplicate work, which I think Silvia is investigating now.

processor/stream/processor.go
```diff
@@ -82,43 +76,21 @@ func (s *srErrorWrapper) Read() (map[string]interface{}, error) {
...
+}
+
 type Processor struct {
-	Tconfig      transform.Config
-	Mconfig      model.Config
+	Models       []EventModel
```
axw: How about making this a map, with the key being the model name? Then we can get rid of EventModel.Name.

processor/stream/processor.go
```diff
@@ -165,15 +137,15 @@ func (p *Processor) readMetadata(reqMeta map[string]interface{}, reader StreamRe
 }
 
 // HandleRawModel validates and decodes a single json object into its struct form
-func (p *Processor) HandleRawModel(rawModel map[string]interface{}) (transform.Transformable, error) {
+func HandleRawModel(rawModel map[string]interface{}, models []EventModel, requestTime time.Time, metadata metadata.Metadata) (transform.Transformable, error) {
```
axw: Why is this exposed more than it was already? Can we unexport it?

transform/transform.go
```diff
@@ -19,26 +19,16 @@ package transform
...
 
 type Transformable interface {
-	Transform(*Context) []beat.Event
-}
-
-type Context struct {
-	RequestTime time.Time
-	Config      Config
-	Metadata    metadata.Metadata
+	Transform(Config, *sourcemap.Store) []beat.Event
```
axw: :100: on removing `transform.Context`, but I think we should go further and remove the remaining config and sourcemap params. Reason being that these do not differ per event.

How about this as an alternative?

If we follow with through the `Decoder` interface idea, then we can have concrete types like `StacktraceDecoder`. That type could have a field for storing a `sourcemap.Store`, and the `Stacktrace` objects returned by `Decode` would keep an internal reference to it (either directly or by referencing the `StacktraceDecoder`). i.e. something like this:

```go
type StacktraceDecoder struct {
    SourcemapStore *sourcemap.Store
    LibraryPattern *regexp.Regexp
    ExcludeFromGrouping *regexp.Regexp
}

type Stacktrace struct {
    decoder *StacktraceDecoder
    Frames []*StacktraceFrame
}

func (s *Stacktrace) Transform(context.Context) []beat.Event {
    // Apply source maps using s.decoder.SourcemapStore.
}
```

Additional `ErrorDecoder` and `SpanDecoder` types would each have `StacktraceDecoder`-type fields. Rather than constructing all of these things globally, they would be constructed in `beater/api/mux.go`, passing in config at construction time.

Note I added a context.Context param to the Transform method. This is not necessary for this PR, but it will be needed for https://github.com/elastic/apm-server/issues/3184 and if we're changing them all then it might be a good opportunity to add it in.

jalvz: It feels a bit creepy to me to have the sourcemap thing (basically global mutable state) as an attribute hidden down in some data type, i rather pass it explicitly as function argument top level.
That said I don't like neither to pass arguments that are not needed in most implementations... I'll explore the `Decoder` idea.

processor/stream/processor.go
jalvz: yeah, definitely

processor/stream/processor.go
jalvz: I guess the benefit is write it only once, and not on each `Decode` implementation.

model/error/event.go
```diff
@@ -113,20 +133,17 @@ type Log struct {
...
-	}
-	if input == nil {
-		return nil, errMissingInput
-	}
-
+func (d decoder) Decode(input interface{}, requestTime time.Time, metadata metadata.Metadata) (m.Transformable, error) {
 	raw, ok := input.(map[string]interface{})
 	if !ok {
 		return nil, errInvalidType
 	}
```
axw: Would it be reasonable to instead always assume the input type is an object, and do the type assertion on the outside and change the `Decode` signature to accept `map[string]interface{}`? Then we could get rid of a little bit of boilerplate.

Alternatively, maybe we could make this a naked type assertion straight after a successful call to Validate, as it will have internally checked the type. i.e.

```go
if err := validation.Validate(input, cachedModelSchema); err != nil {
    return nil, err
}
raw := input.(map[string]interface{}) // safe because Validate already checked the type
```

model/event.go
```diff
@@ -15,15 +15,20 @@
...
 
-type Processor interface {
-	Validate(map[string]interface{}) error
-	Decode(map[string]interface{}) (*metadata.Metadata, []transform.Transformable, error)
-	Name() string
+type Decoder interface {
+	Decode(interface{}, time.Time, metadata.Metadata) (Transformable, error)
+}
+
+type Transformable interface {
```
axw: Nice, I'll be glad to see the back of the transform package.

model/error/event.go
```diff
@@ -91,6 +107,10 @@ type Event struct {
 
 	Experimental interface{}
 	data         common.MapStr
+
+	Metadata metadata.Metadata
+
+	decoder decoder
```
axw: I think this should be `*decoder`, and `func (d decoder) Decode(...)` should instead be a method on the pointer type. Otherwise there'll be a lot of unnecessary copying.

axw: I had forgotten that this Event type is also constructed directly by processor/otel, so perhaps referencing decoder itself here is not the right thing to do after all. It's not really a problem at the moment, as RUM data doesn't go through processor/otel, but it's still a bit ugly.

Another couple of options:
 1. Have Decoder wrap Event with another type that (RUMEvent?) that has the sourcemap-related bits embedded.
 2. Introduce middleware for post-processing Transformables. The middleware would encapsulate the sourcemap/stacktrace configuration, and the Transformable interface would remain simple so we wouldn't have to keep plumbing the config through all the different layers.

jalvz: I had thought of `1`. 
Can you elaborate on the middleware idea? With post-processing you mean to work on the `beat.Event` types?

model/error/event.go
jalvz: Sure, happy to go with a naked type assertion!

beater/api/asset/sourcemap/handler.go
```diff
@@ -26,10 +26,9 @@ import (
 
 	"github.com/elastic/apm-server/beater/request"
 	"github.com/elastic/apm-server/decoder"
-	"github.com/elastic/apm-server/processor/asset"
+	"github.com/elastic/apm-server/processor/asset/sourcemap"
 	"github.com/elastic/apm-server/publish"
-	"github.com/elastic/apm-server/transform"
-	"github.com/elastic/apm-server/utility"
+	sourcemap2 "github.com/elastic/apm-server/sourcemap"
```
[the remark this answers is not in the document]
jalvz: he, yeah

model/error/event.go
axw: It could be done like that (it would basically be a pipeline processor?), or we could have something like this:

In `package publish`:
```go
package publish

type TransformerMiddleware func(model.Transformable)
```

(Publisher could be configured with a TransformerMiddleware, calling it for each Transformable before finally calling the Transform method.)

In the `model/error` package you'd have something like this:

```go
package error

func NewErrorStacktraceMiddleware(sourcemapStore *sourcemap.Store) func(model.Transformable) model.Transformable {
    return func(t model.Transformable) {
        switch t := t.(type) {
        case *Event:
            // Use sourcemapStore etc. to update the event
        }
    }
}
```

It seems kinda dirty to me how `Transform` modifies the Stacktrace objects. Perhaps it would be better to operate on the generated `beats.Event` output instead.

jalvz: Ok. I think that this solution will spread the number of places where data is juggled, and add another thing you have to "remember" when adding a new event type (ie another switch case). I feel it is not worth it...

I like the idea of injecting the config settings in the type (either as wrapper with the configs + the actual event, or the fields directly embedded in the event); but for the sourcemap I think the option less bad is to pass it as argument to `Transform` from the publisher.

If we later have the need for more caches (ie. api keys per service - can't picture how implementation will look like, but could need something like it), we then can embed them all in a big `state`/`context` object (the word "context" has being seized by go's package, but semantically is accurate: something that affects event transformation other than the events themselves). wdyt?

axw: I agree that the second middleware approach isn't great, and shunting this work off to a "pipeline processor" isn't great either, but I don't want to give up yet! ;)

Context is for carrying request/call-specific context. In this case, sourcemap config is not specific to the individual call context. It's configuration that applies to _all_ calls, hence why I don't like having to pass it around all the time.

So, one more idea: we stop making the model object types responsible for transforming themselves, and introduce a separate Transformer. This is the thing that would be configured, rather than attaching transformation config to the model object.

With this approach the model packages would only be concerned with decoding and providing a common object representation based on the intake API. Reducing to these responsibilities, we could more easily auto-generate the struct types and decoding from the/a spec, and we would just have to hand write the transformation to events.

(If you want to get back to your original change, I can also take a stab at this later.)

jalvz: ok... i ended implementing half of the idea (actually i implemented all of it, and then reverted the last part), i found it a bit weird that the decoding happens on each event (multiple packages) and the transformation all in one place.

I also got back to the original decoder functions - I don't see the point of having 1 interface + 5 additional types (1 per implementation) when it can be done with normal functions...

axw: For bystanders: I had a play with the branch, and came up with this: https://github.com/jalvz/apm-server/compare/lay-rum-ground...axw:lay-rum-ground

simitt: The last couple of days I have been looking into alternatives regarding decoding and validations allowing us to clean up the currently used decoding and schema validation logic. While I did not touch the overall beater/API handling structure, this might still be interesting in the context of your discussion here. I make sure to send a wrap up until next week.

simitt: > Reducing to these responsibilities, we could more easily auto-generate the struct types and decoding from the/a spec, and we would just have to hand write the transformation to events.

This is exactly what I've looked into.

model/span/event_test.go
```diff
@@ -363,7 +354,6 @@ func TestSpanTransform(t *testing.T) {
 				"labels":      common.MapStr{"label.a": 12, "label.b": "b", "c": 1},
 				"processor":   common.MapStr{"event": "span", "name": "transaction"},
 				"service":     common.MapStr{"name": serviceName, "environment": env},
-				"timestamp":   common.MapStr{"us": int64(float64(timestampUs) + start*1000)},
```
simitt: `timestamp` shouldn't be removed but moved to the `decode` test now.

processor/otel/consumer.go
```diff
@@ -108,39 +100,54 @@ func (c *Consumer) convert(td consumerdata.TraceData) (metadata.Metadata, []tran
...
 				ParentId:  parentID,
 				TraceId:   traceID,
 				Timestamp: startTime,
 				Duration:  duration,
 				Name:      &name,
+				Metadata:  md,
+			}
+			parseTransaction(otelSpan, hostname, &tx)
+			if tx.Timestamp.IsZero() {
+				tx.Timestamp = time.Now()
```
simitt: It's a detail, but this sets the timestamp to a different time for every event within the same request. I'd prefer keeping the current logic of setting the same timestamp for all events within one request.

processor/otel/consumer.go
```diff
@@ -611,3 +618,21 @@ func truncate(s string) string {
...
+}
+
+type transformableSpan struct {
+	*span.Event
+}
+
+func (ts transformableSpan) Transform() []beat.Event {
+	// TODO(axw) provide different methods on Span for transforming RUM vs. non-RUM errors?
+	return ts.Event.Transform(nil, nil, nil)
+}
```
simitt: Why duplicating the logic here instead of making the `transformableError` and `transformableSpan` at https://github.com/elastic/apm-server/pull/3254/files#diff-9544c869a4bceb063ad78365bfa27c88 public?

publish/pub.go
```diff
@@ -49,17 +50,12 @@ type Publisher struct {
...
+	Transformables []Transformable
 	Trace          bool
 }
 
-// PublisherConfig is a struct holding configuration information for the publisher,
-// such as shutdown timeout, default pipeline name and beat info.
-type PublisherConfig struct {
-	Info            beat.Info
-	ShutdownTimeout time.Duration
-	Pipeline        string
```
simitt: Removing the configuration and instead making the `publish` package aware of the `config.config`  is not a necessary change for this refactoring, and I don't think it is an improvement so please revert.

publish/pub.go
```diff
@@ -162,8 +157,8 @@ func (p *Publisher) processPendingReq(req PendingReq) {
 	}
 
 	for _, transformable := range req.Transformables {
-		span := tx.StartSpan("Transform", "Publisher", nil)
-		events := transformable.Transform(req.Tcontext)
+		span := tx.StartSpan("transform", "Publisher", nil)
```
simitt: Why changing from `Transform` to `transform`?

simitt:
This PR became really big and it holds some opinionated refactoring that is not necessary for the RUM changes, e.g. renamings, restructuring, etc. 
I fear this is going to open up discussions unrelated to the goal you try to achieve and therefore suggest to do such refactorings in dedicated PRs. IMO the part necessary for the RUM changes is mainly changes related to introducing the `RUMProcessor` and `BackendProcessor` (https://github.com/elastic/apm-server/pull/3254/files#diff-01c58acd54a543ec500d1ebf9e7bbe94R47). 

LMK if I am missing something that requires to have all these changes in this PR.

processor/otel/consumer.go
jalvz: ok. why that would be preferable?

processor/stream/processor.go
```diff
@@ -164,32 +171,24 @@ func (p *Processor) readMetadata(reqMeta map[string]interface{}, reader StreamRe
...
-			tr, err := model.modelDecoder(entry, p.Mconfig, err)
-			if err != nil {
-				return nil, err
-			}
-			return tr, nil
+// handleRawModel validates and decodes a single json object into its struct form
+func handleRawModel(rawModel map[string]interface{}, decoders map[string]EventDecoder, requestTime time.Time, metadata metadata.Metadata) (publish.Transformable, error) {
+	for name, decoder := range decoders {
+		if entry, ok := rawModel[name]; ok {
+			return decoder.Decode(entry, requestTime, metadata)
```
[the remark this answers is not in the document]
simitt: I agree that this can be done in a follow up.

tests/json_schema.go
```diff
@@ -24,23 +24,25 @@ import (
...
+	"github.com/elastic/apm-server/tests/loader"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
-	"github.com/elastic/beats/libbeat/beat"
 	"github.com/elastic/beats/libbeat/mapping"
 )
 
-type TestProcessor interface {
```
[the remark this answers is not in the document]
jalvz: While refactoring processors, I deemed necessary to actually test those real processors, as I didn't necessarily trusted the mock processors. 
As a matter of fact, testing the production code directly did catch a couple of issues.
Making the production code testable by itself without needing lots of custom logic is a good thing IMO, but in this case if forced some additional changes, like eg. https://github.com/elastic/apm-server/pull/3254#discussion_r376847765

model/onboarding/onboarding.go
```diff
@@ -15,40 +15,25 @@
...
-	listenAddr string
+type Doc struct {
+	ListenAddr string
 }
 
-func (o onboardingDoc) Transform(tctx *transform.Context) []beat.Event {
+func (d Doc) Transform() []beat.Event {
 	return []beat.Event{{
-		Timestamp: tctx.RequestTime,
+		Timestamp: time.Now(),
```
simitt: Theoretically there could be a larger gap in pushing an event into the queue for transformation and actually transforming events. We should aim for aligning on the `Timestamp` meaning for the events (some are now moved to the decode phase, whereas for the onboarding it is moved to the transform phase). Just recently there was a pipeline processor introduced adding the ingestion timestamp, so I think settling for the request timestamp (as you have done for the other events) makes more sense. Then I would suggest to change this back to also reflect the scheduling timestamp.

model/span/event.go
```diff
@@ -256,33 +253,38 @@ func (d *DestinationService) fields() common.MapStr {
...
+	}
+
 	event := Event{
 		Name:          decoder.String(raw, "name"),
-		Start:         decoder.Float64Ptr(raw, "start"),
+		Start:         start,
 		Duration:      decoder.Float64(raw, "duration"),
 		Sync:          decoder.BoolPtr(raw, "sync"),
-		Timestamp:     decoder.TimeEpochMicro(raw, "timestamp"),
+		Timestamp:     decoder.TimeEpochMicro(raw, "timestamp", requestTime.Add(offset)),
```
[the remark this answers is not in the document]
jalvz: i would have argued that that logic is not trivial as it depends on order of execution; so having it duplicated it is particularly dangerous.

processor/stream/processor.go
```diff
@@ -83,42 +116,16 @@ func (s *srErrorWrapper) Read() (map[string]interface{}, error) {
...
 type Processor struct {
-	Tconfig      transform.Config
-	Mconfig      model.Config
-	MaxEventSize int
-	bufferPool   sync.Pool
+	Decoders      map[string]EventDecoder
+	MaxEventSize  int
+	bufferPool    sync.Pool
+	stoppingError error
+	errors        []error
```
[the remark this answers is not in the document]
jalvz: This was done for the sake of testability. I could have added the `limitedAdd` logic here directly, but I'm 99% sure that this wouldn't have any practical consequence

jalvz:
Superseded by #3297

processor/otel/consumer.go
simitt: that makes sense, thanks for the explanation.

tests/json_schema.go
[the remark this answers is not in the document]
simitt: I see you closed this PR, but your comment above about issues you caught worries me:
> As a matter of fact, testing the production code directly did catch a couple of issues.

Please create issues for them and link them here.

processor/stream/processor.go
```diff
@@ -48,6 +44,43 @@ var (
...
+	transformer := transformer.Transformer{
+		Experimental: experimental,
+	}
+	return &Processor{
+		Decoders: map[string]EventDecoder{
+			"transaction": DecoderFunc(transformer.DecodeTransaction),
+			"span":        DecoderFunc(transformer.DecodeSpan),
+			"error":       DecoderFunc(transformer.DecodeError),
+			"metricset":   DecoderFunc(transformer.DecodeMetricset),
+		},
```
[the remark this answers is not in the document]
jalvz: Something like that, yes. Each implementation/model will have its own `DecodeXXX` function as needed

tests/json_schema.go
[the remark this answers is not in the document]
axw: @jalvz sorry, I was unclear. I didn't mean to suggest they weren't needed at all, but rather that they could be separated into a separate PR. Because there's a lot of changes involved, it would help to review them independently. Same with the other similar comments I made.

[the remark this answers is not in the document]
jalvz: > I didn't mean to suggest they weren't needed at all, but rather that they could be separated into a separate PR

Sure. It just felt a bit creepy to leave the tests for another PR. Most of the changes are related to the `Transform` and `Validate` change, thou.

> your comment above about issues you caught worries me

I meant about implementation changes in this PR, not existing code from before.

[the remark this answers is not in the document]
simitt: ah I misunderstood then, thanks for clarification.

PR #3276, opened 2020-02-03
update beats framework to 

simitt:
* Update beats framework to 7aa2f7a

* update TLS config for elasticsearch client

* temporary validation fix for self instrumentation `nonzero` check on `hosts`.


elasticsearch/config.go
```diff
@@ -73,19 +73,20 @@ func (h Hosts) Validate() error {
...
-	var dial, tlsDial transport.Dialer
 	var addrs []string
 	proxy, err := httpProxyURL(config)
 	if err == nil {
 		addrs, err = addresses(config)
 	}
-	if err == nil {
-		dial, tlsDial, err = dialer(config)
+	if err != nil {
+		return &http.Transport{}, nil, nil
```
axw: This is changing behaviour, so errors from the above functions aren't returned. Was that intentional? I suspect not, and that instead this should be:

```suggestion
		return nil, nil, err
```

(I'd also like it if we made this code a bit less clever, and change the `if err == nil` above to standard `if err != nil`, even if it does lead to duplicated code...)

simitt: I wanted to keep the changes minimal, but happy to switch the error checks to standard behavior.

PR #3277, opened 2020-02-03
strip trailing slashes from kibana host configuration

### LINKED ISSUE #2647: Remove trailing slash for apm-server.kibana.host configuration option
If the value of `apm-server.kibana.host` config option contains a trailing slash, e.g. `http://kibana:5601/` the APM server fails to connect to Kibana. We can however check the presence of a trailing slash and remove it.

graphaelli:
closes #2647 

Can remove some duplication / risk of getting out of sync with beats framework if  elastic/beats#16042 is accepted

### DIFF
--- a/beater/api/mux.go
+++ b/beater/api/mux.go
@@ -181,7 +181,7 @@ type middlewareFunc func(*config.Config, *authorization.Handler, map[request.Res
 func agentConfigHandler(cfg *config.Config, authHandler *authorization.Handler, middlewareFunc middlewareFunc) (request.Handler, error) {
 	var client kibana.Client
 	if cfg.Kibana.Enabled {
-		client = kibana.NewConnectingClient(common.MustNewConfigFrom(cfg.Kibana.Config))
+		client = kibana.NewConnectingClient(common.MustNewConfigFrom(cfg.Kibana))
 	}
 	h := agent.Handler(client, cfg.AgentConfig)
 	msg := "Agent remote configuration is disabled. " +
--- a/beater/api/mux.go
+++ b/beater/api/mux.go
@@ -22,6 +22,7 @@ import (
 	"net/http"
 	"regexp"
 
+	"github.com/elastic/beats/libbeat/common"
 	"github.com/elastic/beats/libbeat/monitoring"
 
 	"github.com/elastic/beats/libbeat/logp"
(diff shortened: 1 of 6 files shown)


beater/config/config.go
```diff
@@ -42,6 +44,45 @@ var (
 	regexObserverVersion = regexp.MustCompile("%.*{.*observer.version.?}")
 )
 
+type kibanaConfig struct {
+	Enabled bool                `config:"enabled"`
+	Config  kibana.ClientConfig `config:",inline"`
```
axw: Can `kibana.ClientConfig` be embedded? It would be nice if we could turn (for example) `cfg.Kibana.Config.Host` into `cfg.Kibana.Host`.

beater/config/config.go
```diff
@@ -42,6 +44,45 @@ var (
...
+}
+
+func (k *kibanaConfig) Unpack(c interface{}) error {
```
axw: Would it make sense to make this `Unpack(cfg *ucfg.Config)` instead? i.e.

```go
func (k *kibanaConfig) Unpack(cfg *ucfg.Config) error {
    if err := cfg.Unpack(&k.Config); err != nil {
        return err
    }
    k.Enabled = cfg.Enabled()
    k.Config.Host = strings.TrimRight(k.Config.Host, "/")
    return nil
}
```

IIANM, that'll save a round-trip from Config -> interface{} -> Config

beater/config/config.go
graphaelli: Great idea.  I was trying to make `type kibanaConfig kibana.ClientConfig` originally and just embedding didn't even occur to me.

beater/config/config.go
graphaelli: @axw Another great suggestion, thanks.  I had to use `common.Config` to get the `Enabled()` method - would you double check 99decba9649b831ae79aec3464bdb8826eca8b01 to see if it looks right?

beater/api/mux.go
```diff
@@ -179,15 +180,15 @@ type middlewareFunc func(*config.Config, *authorization.Handler, map[request.Res
 
 func agentConfigHandler(cfg *config.Config, authHandler *authorization.Handler, middlewareFunc middlewareFunc) (request.Handler, error) {
 	var client kibana.Client
-	if cfg.Kibana.Enabled() {
-		client = kibana.NewConnectingClient(cfg.Kibana)
+	if cfg.Kibana.Enabled {
+		client = kibana.NewConnectingClient(common.MustNewConfigFrom(cfg.Kibana))
```
axw: Can you please add a TODO (or just do it) to pass in the `kibana.ClientConfig` directly, rather than another round-trip through `libbeat/common`? Then we can change `apm-server/kibana` to call `libbeat/kibana.NewClientWithConfig`, rather than _yet another_ round-trip :)

robgil:
# [Codecov](https://codecov.io/gh/elastic/apm-server/pull/3277?src=pr&el=h1) Report
> Merging [#3277](https://codecov.io/gh/elastic/apm-server/pull/3277?src=pr&el=desc) into [master](https://codecov.io/gh/elastic/apm-server/commit/0d828d7c5eaed62f5fb537a8217aafcb90ad3e98&el=desc) will **increase** coverage by `0.00%`.
> The diff coverage is `n/a`.



```diff
@@           Coverage Diff           @@
##           master    #3277   +/-   ##
=======================================
  Coverage   79.33%   79.34%           
=======================================
  Files         107      107           
  Lines        5600     5621   +21     
=======================================
+ Hits         4443     4460   +17     
- Misses       1157     1161    +4     
```

beater/config/config.go
```diff
@@ -42,6 +44,40 @@ var (
...
 
+type kibanaConfig struct {
+	Enabled             bool `config:"enabled"`
+	kibana.ClientConfig `config:",inline"`
+}
+
+func (k *kibanaConfig) Unpack(cfg *common.Config) error {
+	if cfg == nil {
+		return nil
+	}
```
axw: Pretty sure this is unnecessary. go-ucfg's `Config.Unpack` will return an error if the config is nil, before calling into the provided value.

beater/config/config.go
```diff
@@ -42,6 +44,40 @@ var (
 	regexObserverVersion = regexp.MustCompile("%.*{.*observer.version.?}")
 )
 
+type kibanaConfig struct {
```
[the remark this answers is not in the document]
graphaelli: done

beater/config/config.go
graphaelli: I noticed you removed it in the last suggestion but hesitated since this is exported.  On second thought, taking the suggestion.

beater/api/mux.go
graphaelli: Good call, just did it

graphaelli:
holding off on merging this for a day or two to see if we can get the upstream changes into beats first

PR #3279, opened 2020-02-04
add event.ingested to APM events

### LINKED ISSUE #2934: Add ingest timestamp to pipeline
To be able to use transforms (and other post-processing pipelines), an ingest timestamp could be added to the events on ingestion:

https://www.elastic.co/guide/en/elasticsearch/reference/7.4/accessing-data-in-pipelines.html#accessing-ingest-metadata

This allows jobs to collect new documents, either for alerts, transforms or other projections of processor events. The current `@timestamp` references the time when the event was recorded, not when it was ingested, and is not a reliable way to sort when trying to collect new documents.

As APM Server currently adds the pipeline and the index template, it should probably handle the configuration of the ingest timestamp as well.

We should figure out how to deal with backwards compatibility. If the user is on Kibana 7.6, but on APM Server 7.1, is it possible for Kibana to add both the processor and update the mapping for the APM indices? If we add `{ ingest: { timestamp: { ... } } }` to the mapping from Kibana, what happens when APM server overrides the template again?

cc @elastic/apm-ui

graphaelli:
closes #2934


ingest/pipeline/definition.json
```diff
@@ -57,5 +62,20 @@
...
   }
+},
+{
+  "id": "apm_ingest_timestamp",
+  "body": {
+    "description": "Add an ingest timestamp for APM events",
+    "processors": [
+      {
+        "set": {
+          "if": "ctx.processor?.event != 'span'",
```
simitt: How about also excluding `metric` and `profile` events until needed?

graphaelli: I had considered that but figured they were low enough volume that it's ok?  I think building data frames on metric and profile data might actually be very useful and didn't want to raise a barrier to it.  Finally, since it's pretty straightforward to change this on the fly I wasn't too worried.  How strongly do you feel about the suggestion?

simitt: Not strong at all, was mainly curious about why only excluding spans.

tests/system/test_integration.py
```diff
@@ -69,6 +69,11 @@ def test_load_docs_with_template_and_add_transaction(self):
         # compare existing ES documents for transactions with new ones
         rs = self.es.search(index=index_transaction)
```
graphaelli: looks like we can change these to `wait_for_events` and consolidate the `event.ingested` munging

graphaelli: @simitt I made some changes after your review to get tests passing - we now disregard the value in `event.ingested` in all approval tests - would peek at b3778dc9422d66443426995f654144d1d3dd866e and ensure you still approve?

tests/system/error.approved.json
```diff
@@ -31,6 +31,9 @@
             "us": 1494342245999000
         },
         "@timestamp": "2017-05-09T15:04:05.999Z",
+        "event": {
+            "ingested": "2017-05-09T15:04:05.998Z"
```
simitt: The timestamps look odd - have you set them once manually? (They should be some date yesterday)

tests/system/test_integration.py
```diff
@@ -67,14 +67,12 @@ def test_load_docs_with_template_and_add_transaction(self):
         self.assert_no_logged_warnings()
 
         # compare existing ES documents for transactions with new ones
-        rs = self.es.search(index=index_transaction)
-        assert rs['hits']['total']['value'] == 4, "found {} documents".format(rs['count'])
-        self.approve_docs('transaction', rs['hits']['hits'])
+        transaction_docs = self.wait_for_events('transaction', 4, index=index_transaction)
```
simitt: 💯

tests/system/error.approved.json
graphaelli: Good eye - yes, I set them manually

jalvz:
does this need a backport?

graphaelli:
yes please!

Issue #3109, opened 2020-01-07
Rename Intake API to API


simitt:
The current structure of the docs section _Intake API_ is misleading. 

I suggest to rename _Intake API_ to _API_ as it contains sections for all APM Server endpoints and the nested _Events API_ to _Events Intake API_. 

![Screenshot 2020-01-07 at 09 27 36](https://user-images.githubusercontent.com/5555349/71880194-0af15200-3130-11ea-8934-2610f2e750ca.png)

PR #3283, opened 2020-02-04
[docs] Rename intake API

bmorelli25:
Closes https://github.com/elastic/apm-server/issues/3109.

* "Intake API" --> "API"
* "Events API" --> "Events intake API"
* Other cleanup as these files haven't been touched in a while.

### DIFF
--- a/docs/events-api.asciidoc
+++ b/docs/events-api.asciidoc
@@ -1,5 +1,9 @@
 [[events-api]]
-== Events API
+== Events Intake API
+
+NOTE: Most users do not need to interact directly with the events intake API.
+
+The events intake API is what we call the internal protocol that APM agents use to talk to the APM Server.
 
 Agents capture different types of information, known as events.
 These events are sent to a single endpoint which then sorts and processes the events.
@@ -9,8 +13,9 @@ Events can be:
 * Spans
 * Errors
 * Metrics
+* Metadata
 
-You can learn more about events in the {apm-overview-ref-v}/apm-data-model.html[APM Data Model]. 
+See the {apm-overview-ref-v}/apm-data-model.html[APM Data Model] to learn more about the different types of events. 
 
 [[events-api-endpoint]]
 [float]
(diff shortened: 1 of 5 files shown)


docs/events-api.asciidoc
```diff
@@ -90,20 +95,20 @@ An example error response might look something like this:
...
-* <<span-api, Spans>>
-* <<error-api, Errors>>
-* <<metricset-api, Metrics>>
-* <<example-intake-events, Example Request Body>>
+* <<metadata-api>>
+* <<transaction-api>>
+* <<span-api>>
+* <<error-api>>
+* <<metricset-api>>
+* <<example-intake-events>>
```
bmorelli25: @simitt We don't have a section here for the new Jaeger endpoints. My gut instinct is that it isn't necessary, but I'm curious to hear your thoughts.

docs/events-api.asciidoc
```diff
@@ -9,8 +13,9 @@ Events can be:
 * Spans
 * Errors
 * Metrics
+* Metadata
```
simitt: Metadata need to be sent as first line of every request, but they are not an event category. The meta information is the same for every event of one request and stored in ES as part of each of these events. 
E.g. information about the monitored `service` or the installed `agent` is stored with every event but for efficiency reasons only sent once per request as `metadata`. 
In summary - please do not list it as event here.

docs/events-api.asciidoc
simitt: I am with you, the Jaeger data are sent via a different protocol and endpoint and is not part of the events intake API.

docs/intake-api.asciidoc
```diff
@@ -21,16 +18,16 @@ The request body looks roughly like this:
 
 With NDJSON, agents can open an HTTP POST request and use chunked encoding to stream events to the APM Server as soon as they are recorded in the agent.
 This makes it simple for agents to serialize each event to a stream of newline delimited JSON.
-
 The APM Server also treats the HTTP body as a compressed stream and thus reads and handles each event independently.
```
simitt: All of the above description is part of the events Intake API and should be moved.

docs/events-api.asciidoc
bmorelli25: Thanks, will remove

PR #3291, opened 2020-02-05
[7.x] strip trailing slashes from kibana host configuration (#3277)

graphaelli:
 - strip trailing slashes from kibana host configuration (#3277)


jalvz:
`./apm-server -e -E apm-server.kibana.host=http://localhost:5601/` connects to Kibana as expected

PR #3305, opened 2020-02-11
Instrument go-elasticsearch client

### LINKED ISSUE #3184: [self instrumentation] instrument Elasticsearch clients

axw:
## Motivation/summary

This PR introduces instrumentation for the go-elasticsearch client, which is currently used for source maps and API Key auth. By capturing Elasticsearch queries, we can better diagnose performance issues involving Elasticsearch.

A couple of notes for reviewers:
 - Only API Key auth queries are fully instrumented at the moment. I will follow up later to propagate context through Transformables, so we can add spans for source mapping.
 - As this is not intended as a user-facing feature, I'm deferring system tests for now.

## How to test these changes

1. Enable the server's self-instrumentation, and API Key auth.
2. Instrument an application with an Elastic APM agent (I used the Go agent, naturally), configuring API Key auth
3. Observe in the APM app that "db.elasticsearch" spans are captured like so:

image

## Related issues

Closes #3184


simitt:
Testing it through works as expected (and nicely shows caching).

PR #3310, opened 2020-02-11
Refactor decoder.ReqDecoder bits

axw:
This is a pure refactoring change, there should be no functional differences.

We introduce two new middlewares, SystemMetadataMiddleware and UserMetadataMiddleware, which extract request-level user/system metadata. These replace decoder.DecodeSystemData and decoder.DecodeUserData respectively, which are consequently removed.

The middlewares update the new request.Context.RequestMetadata field, which is a map. To avoid panics due to nil map insertions, we introduce request.NewContext and update all existing struct literals to use this instead.

Handlers no longer receive a decoder.ReqDecoder, apart from sourcemap which requires one for decoding _sourcemap_ data. The decoder.ReqDecoder type is moved to beater/api/asset/sourcemap,
which is the only remaining user. Also, we move the sourcemap decoder implementation to the same package, which is the only user.

Finally, some dead code is removed:
 - decoder.ReqReader
 - decoder.DecodeLimitJSONData, and some package-internal code that only it used


beater/api/mux.go
```diff
@@ -152,7 +143,7 @@ func sourcemapHandler(cfg *config.Config, builder *authorization.Builder, report
 	if err != nil {
 		return nil, err
 	}
-	h := sourcemap.Handler(systemMetadataDecoder(cfg, decoder.DecodeSourcemapFormData), psourcemap.Processor, *tcfg, reporter)
+	h := sourcemap.Handler(sourcemap.DecodeSourcemapFormData, psourcemap.Processor, *tcfg, reporter)
```
jalvz: why pass `sourcemap.DecodeSourcemapFormData` as an argument to the handler?
it would be simpler to unexport it, and call it directly from the handler in https://github.com/elastic/apm-server/blob/58ef039ae47c897c863429e6eaf9f2f3670209dd/beater/api/asset/sourcemap/handler.go#L55-L54

`ReqDecoder` would not be needed either.

beater/api/mux.go
```diff
@@ -196,20 +187,30 @@ func apmMiddleware(m map[request.ResultID]*monitoring.Int) []middleware.Middlewa
...
 
-func backendMiddleware(_ *config.Config, auth *authorization.Handler, m map[request.ResultID]*monitoring.Int) []middleware.Middleware {
-	return append(apmMiddleware(m),
-		middleware.AuthorizationMiddleware(auth, true))
+func backendMiddleware(cfg *config.Config, auth *authorization.Handler, m map[request.ResultID]*monitoring.Int) []middleware.Middleware {
+	backendMiddleware := append(apmMiddleware(m),
+		middleware.AuthorizationMiddleware(auth, true),
+	)
+	if cfg.AugmentEnabled {
+		backendMiddleware = append(backendMiddleware, middleware.SystemMetadataMiddleware())
```
jalvz: Having `middleware.SystemMetadataMiddleware()` take a boolean argument would push complexity down and keep this more streamlined and also in line of how other middlewares work. Something like: 
```golang
backendMiddleware := append(apmMiddleware(m),
		middleware.AuthorizationMiddleware(auth, true),
                middleware.SystemMetadataMiddleware(cfg.AugmentEnabled)
	)
```

```golang
func SystemMetadataMiddleware(enabled bool) Middleware {
	return func(h request.Handler) (request.Handler, error) {
		if !enabled {
			return h, nil
		}
                // stuff
```

Same for others.

beater/api/mux.go
axw: I personally agree, but I wanted to leave that change (which I thought might be contentious) to another PR. Doing it in this PR would require non-trivial changes to the tests, which currently rely on not having to know the details of the decoding.

beater/api/mux.go
axw: To me it seems strange to install middleware when you know it will _never_ have any use. This configuration is static, not based on any property of the request.

If you know that the middleware isn't even installed, you can immediately put it out of your mind what it does. If it's installed but takes in config, then you still need to keep in mind its behviour.

> also in line of how other middlewares work

I can see two middlewares which take boolean params, but they don't mean enable/disable.
- AuthorizationMiddleware _always_ creates an Authorization object for the request, but optionally enforces the authorization up front
- KillSwitchMiddleware either runs the request or returns an error if the _route_ is disabled

jalvz: > If it's installed but takes in config, then you still need to keep in mind its behviour

fair enough :+1:

PR #3311, opened 2020-02-11
[7.x] Instrument go-elasticsearch client (#3305)

axw:
 - Instrument go-elasticsearch client  (#3305)


jalvz:
Worked as expected:

image

It also shows up in the service map

image

Issue #2854, opened 2019-10-23
[RUM] Payload optimisation for the RUM agent


vigneshshanmugam:
In the RUM agent, we are experimenting with optimising the payload that is sent to the APM server. Currently we have done some prototypes which you can find in the repo here - https://github.com/vigneshshanmugam/rum-agent-payload

We have planned to do these changes in two steps 
### Phase 1 
+ change from NDJSON to JSON format
+ Having JSON format would allow us to do some of the optimisations that are explained in the comparison table - https://github.com/vigneshshanmugam/rum-agent-payload#comparision
+ Details around each optimisation can be found here - https://rum-agent-payload.vigneshh.now.sh/

### Phase 2
+ Since we don't have the flexibility of doing compression on the browser side, we are learning towards shortening field names and optimising the structure of payload even further. One example would be to rename `transaction` to `t` in the top level and constructing spans as tries. 

We are working on a proposal doc at the moment, but the repo already contains some numbers for the phase 1 which would already get close to 50% improvement from the current release. 

### Update 12/03 (Juan)

- We will shorten all fields that RUM can potentially use.
- We won't change the format from NDJSON to JSON, but embed spans inside transactions, or build logic around order of events instead.
- We possibly won't pursue the trie approach (is a bit complicated unfortunately).
- This can't be done backwards compatible from server side.
 
### Update 17/04 (Juan)

Recap of some relevant design decisions taken:

- All marks and metricset fields are shortened. Any field not in the spec will be ignored (not indexed, but not rejected either).
  
- Spans are embedded inside transactions. `trace_id` and `transaction_id` have been removed from spans. Instead of `parentId`, there is a "parentIndex" (shortened to `pi`) which refers to the index of the parent span in the span list. If missing, the parent is the transaction.

 

### Issues

- [x] [Shorten field names](https://github.com/elastic/apm-server/issues/3403)
- [x] [Remove redundant fields](https://github.com/elastic/apm-server/issues/3404)
- [x] [Require minimum version](https://github.com/elastic/apm-server/issues/3512)
- [x] [Add metricsets](https://github.com/elastic/apm-server/issues/3475)
- [x] [Decode marks](https://github.com/elastic/apm-server/issues/3473)
- [x] [Exhaustive approval tests](https://github.com/elastic/apm-server/issues/3476)
- [x] [Manually test not defined (yet processed) fields](https://github.com/elastic/apm-server/issues/3481)

vigneshshanmugam:
[...]

PR #3328, opened 2020-02-13
Rum V3 spec and handler

### LINKED ISSUE #2905: Use a separate spec for RUM intake endpoint
This is a prerequisite for https://github.com/elastic/apm-server/issues/2854

While at it, adapt specs as needed: eg. make `timestamp` required for backend, remove it completely from RUM

jalvz:
Closes #2905


processor/stream/processor.go
```diff
@@ -149,6 +149,28 @@ func RUMProcessor(cfg *config.Config, tcfg *transform.Config) *Processor {
...
+				schema:       transaction.RUMV3Schema,
+				modelDecoder: transaction.DecodeRUMV3Event,
+			},
+			"span": {
+				schema:       span.RUMV3Schema,
+				modelDecoder: span.DecodeRUMV3Event,
+			},
+			"error": {
+				er.RUMV3Schema,
+				er.DecodeRUMV3Event,
```
simitt: Please be consistent with naming the fields on initializing the structs

jalvz:
The main ticket is this one: https://github.com/elastic/apm-server/issues/2854

There is a bit more context in the first commit message:
"RUM V3 aims at optimizing payload sizes. This commit only copies over the existing specs; they will be updated progressively"

Basically, on this commit the new endpoint behaves exactly like the old one. Forgot to mention that this only copies the main top level specs (not `context`, etc), and that it does not include the `metricset` processor.  As a matter of fact, I'd like to remove the `metricset` processor from the existing endpoint too, if possible.

PR #3344, opened 2020-02-17
[7.x] Add new http.response fields for transactions and spans (#3327)

simitt:
 - Add new http.response fields for transactions and spans (#3327)


axw:
Tested and it works as expected, see https://github.com/elastic/apm-server/pull/3327#issuecomment-606415960

PR #3355, opened 2020-02-18
Merge ES default values to Sourcmeap configuration.

### LINKED ISSUE #3350: Respect default values for ES instance when configuring for Sourcemaps
When enabling Elasticsearch output some default values are configured (e.g. the host), that do not need to be explicitly set by the user. This should be the same behavior when customizing Elasticsearch configuration apm-server.rum.source_mapping.elasticsearch.*.

Currently no default values are set as soon as anything is configured via this setting.
This is similar to the issue handled in #3341.

simitt:
## Motivation/summary
When configuring an Elasticsearch instance for Sourcemap usage, Elasticsearch default values should be merged with the custom config, to avoid having to configure all settings.

At the moment default values are not merged as soon as any of the `apm-server.rum.source_mapping.elasticsearch.*` options is configured, which is unexpected behavior and not aligned with how other configuration options are merged with default values. 

This PR changes the behavior to use default options where no dedicated configuration option is provided. 

## Checklist

## How to test these changes

Configure `apm-server.rum.source_mapping.elasticsearch.*` without defining a `host`. It should be merged with the default host. 
Default values are `hosts: []string{"localhost:9200"}, protocol: "http", timeout: 5s`. 

Also configure with a different host than the default host and ensure it is applied. 

Without these changes, the default values would not be set and an error would occur. 

## Related issues

fixes #3350


beater/config/rum.go
```diff
@@ -140,20 +140,37 @@ func replaceVersion(pattern, version string) string {
...
+		return errors.Errorf("error unpacking sourcemapping config: %w", err)
+	}
+	*s = SourceMapping(cfg)
+	if inp.HasField("elasticsearch") {
+		s.esConfigured = true
+	}
+	return nil
+}
+
+type tmpSourceMapping SourceMapping
```
axw: Perhaps move this inside the Unpack method, and comment it to make it clear that it exists just to remove the custom "Unpack" method?

PR #3356, opened 2020-02-18
[7.6] Merge default values with custom API Key Elasticsearch settings (#3342)

simitt:
Backports the following commits to 7.6:
 - Merge default values with custom API Key Elasticsearch settings (#3342)


simitt:
Closing this as it would require backporting a breaking change in beats. The fix will be available in the next minor (7.7.).

PR #3367, opened 2020-02-21
Attempt to fix flaky API Key test

### LINKED ISSUE #3298: Fix flaky APIKeyTest test_invalidate_by_id
Looks like there is a new flaky test showing up:

**Build and Test / System and Environment Tests / test_apikey.APIKeyTest.test_invalidate_by_id** (from nosetests)

**Error Message**
```
{u'invalidated_api_keys': [], u'error_count': 0}
```

**Stacktrace**
```
Traceback (most recent call last):
  File "/usr/lib/python2.7/unittest/case.py", line 329, in run
    testMethod()
  File "/go/src/github.com/elastic/apm-server/tests/system/test_apikey.py", line 90, in test_invalidate_by_id
    assert invalidated.get("invalidated_api_keys") == [apikey["id"]], invalidated
AssertionError: {u'invalidated_api_keys': [], u'error_count': 0}
```

simitt:
## Motivation/summary

This PR aims for fixing two flaky system tests related to API Key creation via `apikey` command. In the current test implementation mostly the response from the command is checked, which is generally fair for testing the command. It can lead to potential issues though with following tests, when the command returned but the API Key change was not yet finished in Elasticsearch. 

This PR refactors current test structure a bit to move basic `wait_until_...` functionality from the auth tests to the apikey command tests and make it reusable. With the changes the command tests ensure to wait until an API Key is really created or deleted in ES. 

Since the PR involves moving logic around, the parts where logic is only moved is kept in dedicated commits for easier review. 

The PR also moves TLS related tests to a dedicated file to keep test files cleaner. 

## Checklist

## Related issues
Fixes #3298


jalvz:
> when the command returned but the API Key change was not yet finished in Elasticsearch

If i understand correctly, the assumption is that a read after a write to the security API might not "see" the write? Has been confirmed with the elasticsearch team that this can happen?

simitt:
https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-refresh.html

> If your application workflow indexes documents and then runs a search to retrieve the indexed document, we recommend using the index API's refresh=wait_for query parameter option. This option ensures the indexing operation waits for a periodic refresh before running the search.

simitt:
The system tests use a refresh interval of 250ms https://github.com/elastic/apm-server/blob/master/tests/system/config/apm-server.yml.j2#L198

tests/system/test_apikey_cmd.py
```diff
@@ -16,11 +65,12 @@ def config(self):
         }
 
     def setUp(self):
-        super(APIKeyBaseTest, self).setUp()
+        super(APIKeyCommandBaseTest, self).setUp()
         self.user = os.getenv("ES_USER", "apm_server_user")
         password = os.getenv("ES_PASS", "changeme")
         self.es_url = self.get_elasticsearch_url(self.user, password)
         self.kibana_url = self.get_kibana_url()
+        self.base = APIKeyBase(self.es_url)
```
jalvz: this is an uncommon pattern, a comment explaining why is needed would help

simitt:
> I see. I guess I didn't expect that to apply to the security API.

That's a good point, I'll follow up on that.

simitt:
@jalvz you are right, that the referenced refresh logic doesn't apply to the creation/invalidation of API Keys. I do not see how these tests can occasionally fail under these circumstances.  

What do you think about still merging this PR in, as the additional `wait_until` calls should not add a lot of overhead when changes are immediately available. They would help to pinpoint the failure to the `creation` or the `invalidation` cmd though.

jalvz:
Sure

tests/system/test_apikey_cmd.py
simitt: It is using _composition over inheritance_ to avoid multiple inheritance. Happy to add a comment, and change the variable name as I can see how _base_ might be misleading.

simitt:
Merging this in then, as failing tests are unrelated.

PR #3368, opened 2020-02-22
Overhaul Makefile (WIP)

axw:
## Motivation/summary

The existing Makefile:
 - has accumulated unused variables and build targets over several years
 - mixes build and system test variables, making it difficult to understand why things are required (which leads to the above)
 - uses an Inversion of Control which is very complicated by virtue of having many users (i.e. the various beats)

This PR overhauls the Makefile, making the following key changes:
 - removed reliance on libbeat/scripts/Makefile (no more inversion of control, targets are all defined in the one Makefile)
 - now using mage in a few more places
 - build tools are installed into "./bin",    to ensure consistency and avoid polluting the environment
 - using proper Make targets/dependencies rather than having everything be .PHONY, reducing unnecessary rebuilds
 - removed various unused targets, and renamed some others to be clearer (e.g. "system-tests-environment" is now "docker-system-tests", which simply runs "make system-tests" in docker-compose)
 - introduced "config.mk", which can be used for overriding variable defaults, e.g. for GOLINT_UPSTREAM

There are some collateral changes:
 - System test configuration ($ES_URL, etc.) is now defined in the top-level docker-compose.yml. You can now run "docker-compose" commands directly, without having to go through the Makefile.)
 - The system tests have been updated in a couple of places to use the same defaults as defined in the top-level docker-compose.yml. The idea is to optimise for use of the docker-compose.yml in the apm-server repo, and not what's in apm-integration-testing.
 - Some of the build-related scripts are removed where unused, and simplified in other cases.

## Checklist

~- [ ] I have signed the [Contributor License Agreement](https://www.elastic.co/contributor-agreement/).~
- [x] My code follows the style guidelines of this project (run `make check-full` for static code checks and linting)
- [x] I have rebased my changes on top of the latest master branch
- [x] I have made corresponding changes to the documentation
~- [ ] I have added tests that prove my fix is effective or that my feature works~
~- [ ] New and existing [**unit** tests](https://github.com/elastic/apm-server/blob/master/TESTING.md) pass locally with my changes~
- [ ] I have updated [CHANGELOG.asciidoc](https://github.com/elastic/apm-server/blob/master/CHANGELOG.asciidoc)

## How to test these changes

Run through CI.

## Related issues

None.

PR #3372, opened 2020-02-24
magefile: use mage to build fields assets

axw:
## Motivation/summary

This is part of the Makefile rewrite.

Stop using the common "update" target and move more logic into "mage update". In particular, build fields.yml, include/fields.go, and docs/fields.asciidoc from magefile. This is in line with how other beats do it now.

Since we're not using the common update target, we won't create _meta/kibana.generated any more. AFAICS this isn't needed, as we don't ship dashboards with apm-server, and index patterns are generated dynamically. Hence, update packaging to exclude this directory.

## Checklist
~- [ ] I have updated [CHANGELOG.asciidoc](https://github.com/elastic/apm-server/blob/master/CHANGELOG.asciidoc)~ (no user-facing changes)

## How to test these changes

- `make update release`

## Related issues

Pulled out of #3368

### DIFF
--- a/script/jenkins/windows-build.ps1
+++ b/script/jenkins/windows-build.ps1
@@ -43,6 +43,12 @@ New-Item -ItemType directory -Path build\coverage | Out-Null
 New-Item -ItemType directory -Path build\system-tests | Out-Null
 New-Item -ItemType directory -Path build\system-tests\run | Out-Null
 
+choco install python -y -r --no-progress --version 3.8.1.20200110
+refreshenv
+$env:PATH = "C:\Python38;C:\Python38\Scripts;$env:PATH"
+$env:PYTHON_ENV = "$env:TEMP\python-env"
+python --version
+
 echo "Building fields.yml"
 exec { mage fields }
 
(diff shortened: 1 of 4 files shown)


script/jenkins/windows-build.ps1
```diff
@@ -43,6 +43,12 @@ New-Item -ItemType directory -Path build\coverage | Out-Null
 New-Item -ItemType directory -Path build\system-tests | Out-Null
 New-Item -ItemType directory -Path build\system-tests\run | Out-Null
 
+choco install python -y -r --no-progress --version 3.8.1.20200110
```
jalvz: why is this duplicated now?

axw: Just because I don't want to refactor this script too. The two scripts already are mostly overlapping - would be nice to clean that up, but I'd prefer to tackle it later.

axw: I think I missed the point of your question. It's needed here because the targets moved around a little bit. "make fields" now creates docs/fields.asciidoc, which requires the use of a Python script.

PR #3374, opened 2020-02-24
Add fingerprint calculation for servicemaps to spans.

### LINKED ISSUE #3335: Add fingerprint processor for use in service maps
Currently Kibana is using a composite aggregation on 5 fields of spans for collecting trace ids that it should inspect to build a service map. These fields are:

- `service.name`
- `service.environment`
- `destination.address`
- `span.type`
- `span.subtype`

Fingerprinting these fields on ingest (i.e., some kind of serialization) would allows to either use a composite agg with one source, which is seemingly faster, or a terms agg, which has even better performance plus removes some limitations around composite aggregations (like nesting it under a sampler aggregation).

APM Server could install this fingerprinting processor as it already controls the `apm` pipeline.

Additionally the field in which the fingerprinted value is stored should be added to the apm-* template mapping.

I've previously used `connection.hash`, but there's probably a better field name. Here's an example of the processor (again, probably room for improvement):

```json
{
  "apm_connection_hash" : {
    "description" : "sets a connection hash based on various fields",
    "processors" : [
      {
        "script" : """
      def address = ctx.destination?.address;
      
      if (address == null || address == "") {
        return;
      }
      
      def hash = new ArrayList();
      def fieldsToHash = new HashMap();
      fieldsToHash['span.type'] = ctx.span?.type;
      fieldsToHash['span.subtype'] = ctx.span?.subtype;
      fieldsToHash['service.name'] = ctx.service?.name;
      fieldsToHash['service.environment'] = ctx.service?.environment;
      fieldsToHash['destination.address'] = address;
      
      for (entry in fieldsToHash.entrySet()) {
        if (entry.getValue() != null) {
          hash.add("{"+entry.getKey()+":'"+entry.getValue()+"'}");
        }
      }
      
      ctx.connection = new HashMap();
      ctx.connection.hash = hash.join('/');
    """
      }
    ]
  }
}
```

simitt:
## Motivation/summary
Calculate span fingerprint based on `service`, `destination`, `type` and `subtype` information and index as `span.servicemap.fingerprint`.

Adding a dedicated field calculated from the above mentioned fields allows for more performant processing of the data for the new servicemaps feature in Kibana. See detailed motivation description in #3335 .

## Checklist

## How to test these changes
Ingest spans and ensure a `span.servicemap.fingerprint` is stored accordingly: 
* The fingerprint is the same for events containing the same information for 
`service.name`, `service.environment`, `span.type`, `span.subtype`, `span.destination`
* No fingerprint value is set for spans where no `span.destination.address` is given.
* for spans where `service` information is overwritten on a per event base, the overwritten service information is used for the calculation.

The testfile [intake-v2/spans.ndjson](https://github.com/elastic/apm-server/blob/master/testdata/intake-v2/spans.ndjson) contains data covering scenarios with and without destination information and with and without overriding metadata service information.

## Related issues
Closes #3335


model/span/event.go
```diff
@@ -447,5 +447,37 @@ func (e *Event) fields(tctx *transform.Context) common.MapStr {
...
+		add(*e.Service.Name)
+	} else if tctx.Metadata.Service != nil && tctx.Metadata.Service.Name != nil {
+		add(*tctx.Metadata.Service.Name)
+	}
+	if e.Service != nil && e.Service.Environment != nil {
+		add(*e.Service.Environment)
+	} else if tctx.Metadata.Service != nil && tctx.Metadata.Service.Environment != nil {
+		add(*tctx.Metadata.Service.Environment)
+	}
+	return base64.StdEncoding.EncodeToString(buf.Bytes())
```
axw: What's the reason behind base64-encoding? I can see that it would be necessary if there were any binary values involved, but there aren't any.

I also don't understand why we're not hashing - that would produce smaller strings. By using a fast, non-cryptographic hash like xxhash64, overhead should be pretty minimal, will avoid allocations and copying, and results in a compact value.

simitt: This is a valid point, I changed the logic to make use of the xxhash algorithm, which was leading to only one additional allocation per event, which is negligible, and comparable processing time. 

The logic is now basically exactly what is done in the [fingerprint processor](https://github.com/elastic/beats/blob/master/libbeat/processors/fingerprint/fingerprint.go) in beats. I still went with implementing it directly in the transformation function to avoid the need of overriding the [MakeDefaultObserverSupport](https://github.com/elastic/beats/blob/master/libbeat/publisher/processing/default.go#L85) for ensuring to always set the processor in code.Some logic would need to be exported, and I think the current code is simple enough to duplicate instead of making other required changes.

docs/fields.asciidoc
```diff
@@ -1512,6 +1512,18 @@ type: long
 
 --
 
+
+
+*`span.servicemap.hash.xxhash`*::
```
axw: Do consumers care about the hash details? What was the reasoning behind changing from `span.servicemap.fingerprint`? IMO we don't need to explain the implementation details here - it's just an opaque hash.

simitt: I generally agree, but tried to be ECS compliant with the naming https://github.com/elastic/ecs/blob/master/docs/field-details.asciidoc#hash-fields, which lead to me this name, now that we are using a hash algo. 
The point where I was uncertain is whether or not `64-bit` information should be contained, but I decided against it as this could be seen the default.

simitt:
Failing tests are also failing in master.

docs/fields.asciidoc
axw: Hmm. In the ECS cases they're not opaque hashes, though - they're being used for integrity checks. Also, there's `event.hash` which doesn't specify the type of hash.

I suppose it doesn't matter _too_ much as Kibana and APM Server release together, but I still think it'd be better to not identify the implementation details. That would allow us to change the algorithm used later, without affecting the UI.

graphaelli: +1 on opaque approach

dcode: Concur with @axw. The other hashes in ECS are either for integrity or unique identification of an object (e.g. a file). The `event.hash` is a hash of the entirety of the original event. This fingerprint approach for hashing a couple of fields together is common in the infosec community too for easier pivoting in Kibana. For this usecase, it seems `fingerprint` might be more appropriate.

simitt: Thanks for all the input, I'll change accordingly.

model/span/event.go
```diff
@@ -18,10 +18,14 @@
...
 
 import (
+	"encoding/hex"
+	"fmt"
 	"net"
 	"strings"
 	"time"
 
+	"github.com/cespare/xxhash"
+
```
graphaelli:
```suggestion
```

model/span/event_test.go
```diff
@@ -18,12 +18,15 @@
...
 import (
+	"encoding/hex"
 	"encoding/json"
 	"errors"
 	"fmt"
 	"testing"
 	"time"
 
+	"github.com/cespare/xxhash"
+
```
graphaelli:
```suggestion
```

model/span/event.go
```diff
@@ -447,5 +449,31 @@ func (e *Event) fields(tctx *transform.Context) common.MapStr {
...
+func (e *Event) servicemapHash(tctx *transform.Context) string {
+	if e.Destination == nil || e.Destination.Address == nil || *e.Destination.Address == "" {
+		return ""
+	}
+	h := xxhash.New()
+	fmt.Fprintf(h, "%s|%s", *e.Destination.Address, e.Type)
+	if e.Subtype != nil {
+		fmt.Fprintf(h, "|%s", *e.Subtype)
+	}
+	if e.Service != nil && e.Service.Name != nil {
```
graphaelli: can this service name/environment be consolidated somewhere?  can it be done just once before transform?

simitt: I changed the solution a bit. 

Regarding your comment that it should be done just once per transform - this is only run once per event transform. Am I missing something here?

model/span/event.go
```diff
@@ -447,5 +449,46 @@ func (e *Event) fields(tctx *transform.Context) common.MapStr {
...
+		fmt.Fprintf(h, "|%s", name)
+	}
+	if env := serviceVal("env", e.Service, tctx.Metadata.Service); env != "" {
+		fmt.Fprintf(h, "|%s", env)
+	}
+
+	return hex.EncodeToString(h.Sum(nil))
+}
+
+func serviceVal(key string, services ...*metadata.Service) string {
```
graphaelli: per discussion, split into two functions rather than using the key.  Looking forward to results of whether return should be `string` or `*string`

model/span/event.go
graphaelli: As we discussed IRL, i was specifically referring to the pick the first filled in service name or environment.  The change you already made is mostly what I meant.

PR #3359, opened 2020-02-19
Instrument kibana client with apm go agent

simitt:
## Motivation/summary
This adds elastic apm instrumentation to the kibana client, and adds context passing to all client methods.

This change allows to gain more insights into outgoing requests to Kibana, as part of the Agent Configuration endpoint. For the moment it adds instrumentation to the APM Server code only, but once [the instrumentation is also added to beats](https://github.com/elastic/beats/pull/16427), the `apmhttp` auto-instrumentation can be leveraged for outgoing calls. 

closes #3185

## Checklist

~- [ ] I have signed the [Contributor License Agreement](https://www.elastic.co/contributor-agreement/).~
- [x] My code follows the style guidelines of this project (run `make check-full` for static code checks and linting)
- [x] I have rebased my changes on top of the latest master branch

~- [ ] I have made corresponding changes to the documentation~
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing [**unit** tests](https://github.com/elastic/apm-server/blob/master/TESTING.md) pass locally with my changes

- [x] I have updated [CHANGELOG.asciidoc](https://github.com/elastic/apm-server/blob/master/CHANGELOG.asciidoc)

## How to test these changes

Start APM Server with `rum` and `kibana` enabled, and start the Elastic Stack. Make a request to the agent config endpoint, e.g. `curl -i "http://localhost:8200/config/v1/agents?service.name=123"` and ensure the outgoing request to Kibana is recorded. 

Screenshot 2020-02-19 at 17 14 28

## Related issues

closes #3185

PR #3386, opened 2020-02-25
[7.x] Instrument kibana client with apm go agent (#3359)

simitt:
 - Instrument kibana client with apm go agent (#3359)


jalvz:
Works as expected 
image

PR #3393, opened 2020-02-26
Makefile: rewrite all the things

axw:
## Motivation/summary

This is a ~complete rewrite of our Makefile. Key points:

- The main aim is to reduce its complexity by removing reliance on libbeat/scripts/Makefile.
- We're now using `mage` in a few more places, e.g. creating the python-venv, and will continue relying on beats/dev-tools/mage for the foreseeable future.
- We're using proper Make targets/dependencies in more places now, rather than using .PHONY targets which cause unnecessary rebuilds
- Build tools used by the Makefile are now installed into the "build" directory to ensure consistency and to avoid polluting the environment.

For clarity of purpose, I have renamed "system-tests-environment" to "docker-system-tests"~; I've also changed it to remove containers after running, so explicitly stopping the environment afterwards is no longer necessary~. The "start-environment" and "stop-environment" targets no longer exist -- use docker-compose commands directly as needed.

Finally, the Makefile includes "config.mk" if it exists, which can be used for overriding variable defaults such as GOLINT_UPSTREAM.

## Checklist
~- [ ] I have updated [CHANGELOG.asciidoc](https://github.com/elastic/apm-server/blob/master/CHANGELOG.asciidoc)~ (no user-facing changes)

## How to test these changes

- `make`
- `make fmt`
- `make docker-system-tests`
- `make check-approvals`

## Related issues

None.


axw:
Thanks for the review @simitt!

> Running make docker-system-tests runs successfully, but according to your description, containers should be stopped and removed afterwords, only the container running the tests is stopped and removed though (ES, Kibana are still running).

My mistake. I forgot that `--rm` only stops/removes the specified service, and not dependencies. There's no change from the old Makefile in that respect then - I'll update the description.

> I also had to manually remove the build folder as all following make commands resulted in this error:
>
> ./build/mage: ./build/mage: cannot execute binary file

Oops, that'll be because I'm installing tools into the build dir, which is shared with the Docker container. Since you're on macOS, that'll be a problem. Pretty sure the following problem you listed is the same, just in the opposite direction. Will fix before merging.

Not sure what "Failed building wheel" is about, I'll look into it.

Makefile
```diff
@@ -1,220 +1,293 @@
...
-	printf "version: '2.3'\nservices:\n beat:\n  build:\n   args: [UID=%d]" $(shell id -u) > $@
-system-tests-environment: docker-compose.override.yml
-build-image: docker-compose.override.yml
+##############################################################################
+# Rules for creating and installing build tools.
+##############################################################################
+
+# $GOBIN must be set to use "go get" below. Once we move to modules we
+# can just use "go build" and it'll resolve all dependencies using modules.
+export GOBIN=$(CURDIR)/build
```
graphaelli: I changed this to `export GOBIN=$(addprefix $(CURDIR)/,$(GOOSBUILD))` (though I'm not sure that's the best change) to resolve:

```
$ make
go get ./vendor/github.com/magefile/mage
make: ./build/darwin/mage: No such file or directory
make: *** [include/fields.go] Error 1
```

axw:
@graphaelli 

> Does the version embedding still happen for releases?

Yes. `make release` just calls `mage package`, which does not use anything from the Makefile: it's all one way now, Make -> mage. The `make apm-server` target is just a convenience, and I removed the version/time to avoid rebuilds during development.

> Can the default target be sped up? configs and fields.yml seem to be build every time. Not a dealbreaker as go build is not much more typing.

I had originally managed to get rebuilds to a minimum, but some dependencies crept back in. I'll have a look and see if I can fix that.

Makefile
axw: I missed pushing a commit - it's `export GOBIN=$(abspath $(GOOSBUILD))` now.

axw:
@simitt sorted the failures, had to install wheel in the Python venv. Updated the Dockerfile to do that. Not sure why it only affects the Docker env.

@graphaelli I removed the update dependency from apm-server, so `make` is now fast again. While I was at it, I modified the Makefile to run "mage -compile" to compile our magefile.go into a static binary, rather than rebuilding it each time "mage" is run. Should make some other things faster.

simitt:
Changes LGTM, I tested again and the listed issues are fixed.

PR #3394, opened 2020-02-26
beater/jaeger: introduce "auth_tag" configuration

### LINKED ISSUE #3339: Add support for Jaeger auth handling
The APM Server Jaeger gRPC endpoint supports TLS. If apm-server.ssl is configured, the settings are automatically applied to the APM Server's Jaeger gRPC endpoint. The implemented HTTP endpoint does not support TLS or using a secret token for now. 

Investigate possibilities and limitations for supporting auth handling for Jaeger spans. 

Considerations from [issue#2886](https://github.com/elastic/apm-server/issues/2886#issue-519237545):
> One thing that is yet to be solved is auth. A limited subset of the Jaeger clients (Java and C#) support the JAEGER_AUTH_TOKEN config, which can be used to send a bearer token along with requests to the collector. This only works when using HTTP. The Jaeger agent only supports client certificate auth. We do support this, but it might also be worth a lighter-weight option, such as ferrying a secret/bearer token through tags, using the --jaeger.tags CLI flag.

axw:
## Motivation/summary

We introduce support for configuring the server to look for a specified Process tag in Jaeger batches, and use the value for authorization. The value is expected to have the same format as HTTP authorization headers that the Elastic APM agents support. By default Jaeger authorization is disabled, as there is no standard auth tag name.

This enables token-based auth as an alternative to client certificate auth.

Note for reviewers: an alternative approach to configuration would be to have a hard-coded auth tag, like "elastic_auth", and have that enabled by default (or enabled with a boolean). Since token-based auth isn't a native feature of Jaeger clients/agents, I thought it might be a bit non-intuitive, hence why I made it opt-in. Let me know what you think.

## Checklist

## How to test these changes

1. Configure `apm-server.jaeger.auth_tag: "authorization"`
1. Configure `apm-server.secret_token: "foo"`
1. Run Jaeger Agent without any tags, run HotROD, and check the agent shows "not authorized" errors and nothing is indexed
1. Run Jaeger Agent with `--agent.tags authorization=foo`, run HotROD
1. Check it all works, and there is no "labels.authorization" field in any docs

## Related issues

Closes #3339


beater/jaeger/server.go
```diff
@@ -79,15 +94,15 @@ func NewServer(logger *logp.Logger, cfg *config.Config, tracer *apm.Tracer, repo
...
+		api_v2.RegisterCollectorServiceServer(srv.grpc.server, grpcCollector{auth, traceConsumer})
 	}
 	if cfg.JaegerConfig.HTTP.Enabled {
 		// TODO(axw) should the listener respect cfg.MaxConnections?
 		httpListener, err := net.Listen("tcp", cfg.JaegerConfig.HTTP.Host)
 		if err != nil {
 			return nil, err
 		}
-		httpMux, err := newHTTPMux(traceConsumer)
+		httpMux, err := newHTTPMux(auth, traceConsumer)
```
simitt: Do we actually want to support auth handling for HTTP, since this does not support TLS? 

If so, I think the docs should be aligned, currently they only discuss auth handling for gRPC.

axw: Good point. I was thinking it might still be useful for folks using a service mesh, stunnel, or some other way of securing comms without specific TLS support in the server/client. Pretty tenuous though, so I'll remove it for now.

axw: I renamed `apm-server.jaeger.auth_tag` to `apm-server.jaeger.grpc.auth_tag`, changed the code to only check auth in gRPC.

axw:
Failing system tests are the ones I fixed in another PR - merging.

axw:
@bmorelli25 forgot to ask you to review the docs, sorry. Can you please take a look, and feel free to make adjustments.

PR #3401, opened 2020-02-27
[7.x] tests/system: fix agent config tests (#3400)

axw:
 - tests/system: fix agent config tests (#3400)

### DIFF
--- a/_beats/testing/environments/snapshot-oss.yml
+++ b/_beats/testing/environments/snapshot-oss.yml
@@ -3,7 +3,7 @@
 version: '2.3'
 services:
   elasticsearch:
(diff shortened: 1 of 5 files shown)


axw:
Failing because the Kibana backport was merged yesterday, and hasn't made it into a new snapshot Docker image yet.

axw:
Oh, actually, it has I think - it's just that the 7.x branch is stuck using 7.6.0-SNAPSHOT still.

graphaelli:
@axw I was just looking into why these weren't failing when I found this.  #16699 is up for pushing this change to beats.  I pushed that to this branch as well since he next beats update will require these changes anyway for a passing build.

axw:
Thanks @graphaelli!

PR #3408, opened 2020-03-02
Add a pipeline definition to remove spans fields.

### LINKED ISSUE #3304: Make metadata configurable
Currently we set only "minimal" metadata for span documents, in order to minimise storage costs:
https://github.com/elastic/apm-server/blob/92434dd5ee39e07a31950fd6f9d21c21fa072ebc/model/metadata/metadata.go#L100-L106

This leaves out Kubernetes and container metadata, for example. There may be use cases for including these. We should consider making their inclusion configurable, e.g. including the fields in events, but then by default dropping them in an ingest node pipeline processor, which users could customise.

### LINKED ISSUE #3403: [RUM v3] Shorten field names in the intake spec
as per: https://github.com/vigneshshanmugam/rum-agent-payload/blob/master/examples/shortening-fields/script.js

jalvz:
This was done in code before. By moving it to a pipeline users can configure what fields to remove and what to keep.
     
 Closes #3304 

- [x] needs changelog
- [x] needs docs issue


ingest/pipeline/definition.json
```diff
@@ -77,5 +82,22 @@
       }
     ]
   }
+},
+{
+  "id": "apm_span",
+  "body": {
+    "description": "Removes span fields usually not needed to save storage",
```
simitt: nit: I would go a bit more into detail which fields and why it is not needed to store them, e.g. something like `Removes detailed metadata information which is available on the span's transaction, to save storage`.

model/span/event.go
```diff
@@ -388,7 +388,7 @@ func (e *Event) Transform(tctx *transform.Context) []beat.Event {
 	}
 
 	// first set the generic metadata
-	tctx.Metadata.SetMinimal(fields)
+	tctx.Metadata.Set(fields)
 
 	// then add event specific information
 	utility.DeepUpdate(fields, "service", e.Service.MinimalFields())
```
simitt: This should be updated to `utility.DeepUpdate(fields, "service", e.Service.Fields())`

jalvz: good catch :+1

jalvz:
Depends on https://github.com/elastic/apm-server/pull/3413

PR #3413, opened 2020-03-03
Use approved files from system tests for docs.

jalvz:
This PR contains 3 changes, related to each other:

1. Adds a approval system-tests for metricset.
2. Uses approved files from system-tests for docs. They are more faithful and useful for users because they go trough any ingest node pipelines. This is particularly relevant for #3408 
3. Bring back the old code used for copying event examples to the docs folder. The new way has a few disadvantages:
  - Doesn't detect removal of files, or changes in the source file to be copied.
  - Adding a file requires 2 changes in 2 different places.
  - CI can not validate that the docs files have been synced.
  - Different syntax/behaviour in Linux/MacOS (`install -D`)
The benefits of the new way (avoid syncing when nothing changed) are negligible in this case.


tests/system/apmserver.py
```diff
@@ -372,6 +371,8 @@ def approve_docs(self, base_path, received):
         # The first tuple element exists to sort IDs before timestamps.
         def get_doc_id(doc):
             doc_type = doc['processor']['event']
+            if doc_type == 'metric':
+                return len(doc)
```
axw: What's the idea behind using the length of the doc as its ID, rather than timestamp? If you do need a separate case for metrics, can you please keep this in line with the code below, and return a tuple to sort them all?

Makefile
```diff
@@ -132,37 +132,25 @@ endif
 ##############################################################################
 
 .PHONY: docs
-docs: $(docs_data_files)
+docs:
```
axw: maybe make `copy-docs` a prereq?

Makefile
```diff
@@ -132,37 +132,25 @@ endif
...
-docs/data/intake-api/generated/events.ndjson: testdata/intake-v2/events.ndjson
-docs/data/intake-api/generated/rum_v3_events.ndjson: testdata/intake-v3/rum_events.ndjson
-docs/data/intake-api/generated/sourcemap/bundle.js.map: testdata/sourcemap/bundle.js.map
-docs/data/elasticsearch/generated/errors.json: processor/stream/test_approved_es_documents/testIntakeIntegrationErrors.approved.json
-docs/data/elasticsearch/generated/spans.json: processor/stream/test_approved_es_documents/testIntakeIntegrationSpans.approved.json
-docs/data/elasticsearch/generated/transactions.json: processor/stream/test_approved_es_documents/testIntakeIntegrationTransactions.approved.json
-docs/data/elasticsearch/generated/metricsets.json: processor/stream/test_approved_es_documents/testIntakeIntegrationMetricsets.approved.json
-docs/data/elasticsearch/generated/rum_v3_transactions.json: processor/stream/test_approved_es_documents/testIntakeRUMV3Transactions.approved.json
-docs/data/elasticsearch/generated/rum_v3_errors.json: processor/stream/test_approved_es_documents/testIntakeRUMV3Errors.approved.json
+.PHONY: copy-docs
```
axw: thanks/sorry - I think during the rewrite I originally had intended the "update" target to be non-PHONY, but it didn't work out due to magefile

tests/system/apmserver.py
jalvz: my bad, i thought that metricsets came without timestamp - it was just the timestamp was not set in the sample file. reverted this

jalvz:
@bmorelli25 fyi: example elasticsearch documents are now more faithful: the top level "events" field is removed, and it includes all the post-processing (fields added by pipelines and such)

PR #3414, opened 2020-03-03
Shorten field names in the RUM V3 spec

### LINKED ISSUE #3403: [RUM v3] Shorten field names in the intake spec

jalvz:
~~Some pending decisions:~~
~~- What to do with fields not used by RUM currently (leave them / remove them / shorten them~~ ~~anyways). Ill come up with the list of fields, and upload also a file with the mappings.~~
~~- What to spec in `marks.json`, and weather to make a separate spec for RUM V3.~~

Fields removed in this PR (not used by RUM):

`response.finished`
`response.headers_sent`
`request.body`
`request.socket`
`request.cookies`
`ephemeral_id`
`service.node`
`span.db`
`stacktrace.vars`
`stactrace.library_frame`

Closes elastic/apm-server#3403


jalvz:
fyi @vigneshshanmugam

docs/spec/rum_v3_context.json
```diff
@@ -0,0 +1,298 @@
...
+                            "maxLength": 1024
+                        },
+                        "ephemeral_id": {
+                            "description": "Free format ID used for metrics correlation by some agents",
+                            "type": ["string", "null"],
+                            "maxLength": 1024
+                        }
+                    }
+                },
+                "framework": {
```
vigneshshanmugam: we might start sending these field since we already have framework specific information. The reason why they are not sent right now its mainly bcoz of payload size. Can we also shorten these fields

docs/spec/rum_v3_context.json
```diff
@@ -0,0 +1,298 @@
...
+                    "type": ["string", "null"],
+                    "pattern": "^[a-zA-Z0-9 _-]+$",
+                    "maxLength": 1024
+                },
+                "environment": {
+                    "description": "Environment name of the service, e.g. \"production\" or \"staging\"",
+                    "type": ["string", "null"],
+                    "maxLength": 1024
+                },
+                "runtime": {
```
vigneshshanmugam: these could be prefilled from user agent since that is the runtime for the RUM agent.

docs/spec/spans/rum_v3_span.json
```diff
@@ -175,39 +182,35 @@
...
                     "minItems": 0
                 },
-                "sync": {
+                "sy": {
                     "type": ["boolean", "null"],
                     "description": "Indicates whether the span was executed synchronously or asynchronously."
                 }
             },
-            "required": ["duration", "name", "type", "id","trace_id", "parent_id"]
+            "required": ["d", "n", "t", "id","trace_id", "parent_id"]
```
vigneshshanmugam: `trace_id` and `parent_id` must be optional right?

jalvz: yeah, that is job for another pr :)

jalvz:
Here are some fields not shortened so far:

`r.finished`

`r.headers_sent`

`r.headers`

`q.body`

`q.headers`

`q.http_version`

`q.socket`

`q.url[raw|protocol|full|hostname|port|pathname|search|hash]`

`q.cookies`

`se.ephemeral_id`

`se.framework`

`se.environment`

`se.runtime`

`se.node`

`st.context_line`

`st.classname`

`st.library_frame`

`st.module`

`st.post_context`

`st.pre_context`

`st.vars`

`y.action`

You mentioned framework, what about the rest? I am quite sure that some can be dropped, but wanted to hear from you

vigneshshanmugam:
### Fields that RUM would never use it in for the foresable future. 
+ r.finished
+ r.headers_sent
+ q.body (Expensive to read stream in RUM, Might end up in memory leak if not done correctly) 
+ q.socket
+ se.ephemeral_id
+ se.node
// Hard to get these below fields unless browsers exposes these (not going to happen for now)
+ st.post_context
+ st.pre_context
+ st.vars
+ st.library_frame

### Needs changes
+ `se.environment` - `se.e`
+ `q.headers, r.headers` - RUM agent currently sets `server-timing` info on response headers. so if we can have specific fields and also accept wildcards, we can set headers to `q.h` and `r.h` and `r.h.st` for server-timing info. 
+ `se.runtime` - can be autofilled from User agent  by server
+ `q.url` - RUM agent currently sends `context.http.url` which is same as the request URL. We can delegate the work to APM server and can do the job of parsing the URL if its required. 
+ `q.cookies` - Possible but totally depends on Same site cookie spec. Also this could be large and might pollute our payload. so we should try to not send this at all. 
+ `y.action` - Optional right now and we don't set it for all spans at the moment. we can shorten it to `y.a`. We might send it as `external.http.post` or `external.http.get` in future.

jalvz:
Awesome, thanks! I'll investigate the `runtime` bit, that would have to happen in a pipeline.

processor/stream/test_approved_es_documents/testIntakeRUMV3Errors.approved.json
```diff
@@ -1,111 +1,44 @@
...
+                "culprit": "test/e2e/general-usecase/app.e2e-bundle.min.js?token=secret",
+                "custom": {
+                    "testContext": "testContext"
                 },
+                "grouping_key": "d41d8cd98f00b204e9800998ecf8427e",
+                "id": "3661352868c17c78b773d2f1beae6d41",
                 "page": {
-                    "referer": "http://localhost:8000/test/e2e/",
-                    "url": "http://localhost:8000/test/e2e/general-usecase/"
+                    "referer": "h://localhost:8000/test/e2e/",
```
jalvz: argh

processor/stream/test_approved_es_documents/testIntakeRUMV3Events.approved.json
```diff
@@ -0,0 +1,512 @@
...
+            "transaction": {
+                "custom": {
+                    "testContext": "testContext"
+                },
+                "duration": {
+                    "us": 295000
+                },
+                "id": "ec2e280be8345240",
+                "marks": {
+                    "a": {
```
jalvz: ill create a ticket to define these in a follow-up pr

processor/stream/processor.go
```diff
@@ -142,32 +146,34 @@ func RUMProcessor(cfg *config.Config, tcfg *transform.Config) *Processor {
...
 				schema:       span.RUMV3Schema,
 				modelDecoder: span.DecodeRUMV3Event,
 			},
-			"error": {
+			"e": {
 				schema:       er.RUMV3Schema,
 				modelDecoder: er.DecodeRUMV3Event,
 			},
 		},
+		metadataSchema: metadata.RUMV3ModelSchema(),
```
jalvz: The `metricset` model is missing here, I'll file a ticket for that to follow up in a separate PR

processor/stream/processor_test.go
```diff
@@ -214,7 +214,7 @@ func TestRUMV3(t *testing.T) {
 		name string
 	}{
 		{path: "rum_errors.ndjson", name: "RUMV3Errors"},
-		{path: "rum_transactions_spans.ndjson", name: "RUMV3Transactions"},
+		// {path: "rum_events.ndjson", name: "RUMV3Events"},
```
jalvz: note for self: this works, uncomment

model/metadata/service.go
```diff
@@ -67,35 +69,36 @@ type node struct {
...
-func DecodeService(input interface{}, err error) (*Service, error) {
+func DecodeService(input interface{}, hasShortFieldNames bool, err error) (*Service, error) {
 	if input == nil || err != nil {
 		return nil, err
 	}
 	raw, ok := input.(map[string]interface{})
 	if !ok {
 		return nil, errors.New("invalid type for service")
 	}
+	field := fields.Mapper(hasShortFieldNames)
```
simitt: You could initialize the decoder with the mapping information `decoder(fields.Mapper(hasShortFieldNames))`, allowing to abstract away the mapper inside the decoding.

processor/stream/test_approved_es_documents/testIntakeRUMV3Errors.approved.json
```diff
@@ -1,111 +1,44 @@
...
-                            "function": "\u003canonymous\u003e",
-                            "line": {
-                                "column": 18,
-                                "number": 1
-                            }
-                        }
-                    ]
+                "culprit": "test/e2e/general-usecase/app.e2e-bundle.min.js?token=secret",
+                "custom": {
+                    "testContext": "testContext"
```
simitt: Where did the `stacktrace` get lost? Looking at https://github.com/elastic/apm-server/pull/3414/files#diff-ed8e5b35bb28351f66974257d6d24873R2 one of the errors sends a stacktrace.

testdata/intake-v3/rum_events.ndjson
```diff
@@ -1,4 +1,14 @@
...
+{"y":{"id":"a3c043330bc2015e","x_id":"ec2e280be8345240","parent_id":"ec2e280be8345240","trace_id":"286ac3ad697892c406528f13c82e0ce1","n":"POST h://localhost:8003/fetch","t":"external","su":"h","sy":false,"s":119.93500008247793,"d":15.949999913573265,"c":{"h":{"mt":"POST","url":"h://localhost:8003/fetch","sc":200},"dt":{"se":{"n":"h://localhost:8003","rc":"localhost:8003","t":"external"},"ad":"localhost","po":8003}}}}
+{"y":{"id":"bc7665dc25629379","x_id":"ec2e280be8345240","parent_id":"ec2e280be8345240","trace_id":"286ac3ad697892c406528f13c82e0ce1","n":"Fire \"DOMContentLoaded\" event","t":"hard-navigation","su":"browser-timing","s":120,"d":2}}
+{"me":{"x":{"n":"general-usecase-initial-p-load","t":"p-load"},"sa":{"xdc":{"v":1},"xds":{"v":295},"xbc":{"v":1}}}}
+{"me":{"x":{"n":"general-usecase-initial-p-load","t":"p-load"},"y":{"t":"Request"},"sa":{"ysc":{"v":1},"yss":{"v":1}}}}
+{"me":{"x":{"n":"general-usecase-initial-p-load","t":"p-load"},"y":{"t":"Response"},"sa":{"ysc":{"v":1},"yss":{"v":1}}}}
+{"me":{"x":{"n":"general-usecase-initial-p-load","t":"p-load"},"y":{"t":"Processing"},"sa":{"ysc":{"v":1},"yss":{"v":124}}}}
```
simitt: One of the spans should contain a `stacktrace`. It also has been tested so far with the [rum_transactions+span.ndjon](https://github.com/elastic/apm-server/pull/3414/files#diff-e1b3950e4b04b62f51bd12d0829ab2a5L3)

docs/data/intake-api/generated/rum_v3_events.ndjson
```diff
@@ -1,4 +1,14 @@
...
+{"y":{"id":"a3c043330bc2015e","x_id":"ec2e280be8345240","parent_id":"ec2e280be8345240","trace_id":"286ac3ad697892c406528f13c82e0ce1","n":"POST h://localhost:8003/fetch","t":"external","su":"h","sy":false,"s":119.93500008247793,"d":15.949999913573265,"c":{"h":{"mt":"POST","url":"h://localhost:8003/fetch","sc":200},"dt":{"se":{"n":"h://localhost:8003","rc":"localhost:8003","t":"external"},"ad":"localhost","po":8003}}}}
+{"y":{"id":"bc7665dc25629379","x_id":"ec2e280be8345240","parent_id":"ec2e280be8345240","trace_id":"286ac3ad697892c406528f13c82e0ce1","n":"Fire \"DOMContentLoaded\" event","t":"hard-navigation","su":"browser-timing","s":120,"d":2}}
+{"me":{"x":{"n":"general-usecase-initial-p-load","t":"p-load"},"sa":{"xdc":{"v":1},"xds":{"v":295},"xbc":{"v":1}}}}
+{"me":{"x":{"n":"general-usecase-initial-p-load","t":"p-load"},"y":{"t":"Request"},"sa":{"ysc":{"v":1},"yss":{"v":1}}}}
+{"me":{"x":{"n":"general-usecase-initial-p-load","t":"p-load"},"y":{"t":"Response"},"sa":{"ysc":{"v":1},"yss":{"v":1}}}}
+{"me":{"x":{"n":"general-usecase-initial-p-load","t":"p-load"},"y":{"t":"Processing"},"sa":{"ysc":{"v":1},"yss":{"v":124}}}}
```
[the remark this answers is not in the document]
jalvz: `parent_id` and `trace_id` will be removed. Did I miss something?

model/fields/rum_v3_mapping.go
```diff
@@ -0,0 +1,132 @@
...
+
+func Mapper(shortFieldNames bool) func(string) string {
+	return func(s string) string {
+		shortField, ok := rumV3Mapping[s]
+		if ok && shortFieldNames {
+			return shortField
+		}
+		return s
+	}
+}
```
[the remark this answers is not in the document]
jalvz: I think is more complex and the performance impact is negligible. I can benchmark it if you think is worth it, tho.

model/metadata/service.go
jalvz: I tried that, but there are a lot of small functions (`decodeHTTP`, `decodeService`, `decodeStacktrace`, etc) that would need to be adapted to be methods instead, meaning that they would not be reusable (eg. duplicated `metadata.DecodeService`,  `transaction.DecodeService`, `span.DecodeService`, etc.). 
At the end thought it was not worth it and this produced the smallest diff.

docs/spec/errors/rum_v3_error.json
```diff
@@ -16,7 +16,7 @@
                     "type": ["string", "null"],
```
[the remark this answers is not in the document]
jalvz: I don't have a strong preference, but mixing 2 taxonomies could be more confusing... eg. `rum_v3_transactions` could still fit in either `transactions` or `rum_v3`, not sure why one would make more sense over the other...

processor/stream/test_approved_es_documents/testIntakeRUMV3Errors.approved.json
jalvz: great catch!

docs/spec/errors/rum_v3_error.json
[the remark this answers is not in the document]
simitt: It's the first time the endpoint and version are part of the spec name, therefore I think the suggested grouping makes sense. Just a suggestion though, up to you how to finally organize it.

docs/data/intake-api/generated/rum_v3_events.ndjson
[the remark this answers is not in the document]
simitt: These two and `transaction_id` - wasn't aware they are going to be removed. Ignore the comment then.

model/fields/rum_v3_mapping.go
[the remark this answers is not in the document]
simitt: I expect it to only be a couple of nanoseconds difference per call, when `shortFieldNames == false`, but it is called for every decoded field, which is potentially a couple hundreds for larger events. This won't mage a big difference, especially as no allocations are involved, but the suggested solution is fairly straight forward, so I would rather go with that and avoid unnecessary access of the map.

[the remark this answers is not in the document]
jalvz: all things equal ("equal" meaning +-200ns), I rather chose the simplest implementation if you don't mind (or avoid unnecessary complexity, put it another way).

jalvz:
>  This might be a rather theoretical concern, so I am not asking for any related changes in this PR, but would like to get a discussion going about possible implications.


So, for my understanding, what are those implications, exactly?

model/context.go
```diff
@@ -125,7 +125,7 @@ func DecodeContext(input interface{}, cfg Config, err error) (*Context, error) {
 	}
 
 	decoder := utility.ManualDecoder{}
-	field := fields.Mapper(cfg.HasShortFieldNames)
+	field := field.Mapper(cfg.HasShortFieldNames)
```
simitt: The var `field` is shadowing the package name now. How about changing the variable to `mapper`?

simitt:
Created https://github.com/elastic/apm-server/issues/3481 to bundle the discussion started above and not block this PR.

model/context.go
jalvz: Yeah... amended

Issue #3441, opened 2020-03-06
Document `apm_remove_span_metadata` pipeline


jalvz:
This pipeline will do something that before was hardcoded, meaning that now the metadata fields to exclude from spans are configurable.

One important caveat is that users with an output other than Elasticsearch need to run `setup` against Elasticsearch to load the new pipeline, otherwise span documents will start taking much more space.

ref: https://github.com/elastic/apm-server/pull/3408

jalvz:
Consider caveat in https://github.com/elastic/apm-server/pull/3446#issuecomment-605875117

cc @bmorelli25

PR #3446, opened 2020-03-09
[7.x] Add a pipeline definition to remove spans fields. (#3408)

jalvz:
 - Add a pipeline definition to remove spans fields. (#3408)


simitt:
Tested the following:

## Fresh install (ESS, on-prem): 
New pipeline is loaded and succesfully applied. Same meta data as in `7.6` are stored for spans. Removing the pipeline from the apm pipeline and loading the pipelines again via APM Server leads to indexing all metadata also for spans. This works as expected.

## Upgrade from 7.6 to 7.7 (ESS, tar file on-prem): 
New pipeline is loaded into Elasticsearch, but NOT applied as `apm-server.register.ingest.pipeline.overwrite: false` by default. Users can overwrite `apm` pipeline to include the new pipeline by setting `overwrite` to `true`, but the default behavior on upgrade changes. Not sure if this works as expected, as it is not defined in the issue. 

The default behavior between a fresh installation and an upgrade diverges, as the exsting APM pipeline is not overwritten. More data are indexed for upgraded APM Servers than for freshly installed ones, even when the used `apm-server.yml` configuration is identical.  
I am not certain this is expected behavior, and would consider to also disable the new pipeline by default, leading to same behavior for new+upgraded installations, all indexing more span data. At the very least this new behavior should be highlighted in the docs.

jalvz:
I opened a docs issue, but forgot to label it for 7.7: https://github.com/elastic/apm-server/issues/3441

I don't have a strong opinion on what the behaviour should be... Perhaps we should avoid this kind of scenarios in the future.

simitt:
As agreed on offline we keep the logic as is and ensure to have good docs around the upgrade implications.

PR #3454, opened 2020-03-10
Fix CI Dockerfile paths for go get

simitt:
## Motivation/summary
Since switching to go modules there is no vendor folder,
adapt go get commands accordingly for CI Dockerfile.

This is supposed to fix failing CI on `master`

## Checklist

~- [ ] I have signed the [Contributor License Agreement](https://www.elastic.co/contributor-agreement/).~
~- [ ] My code follows the style guidelines of this project (run `make check-full` for static code checks and linting)~
- [x] I have rebased my changes on top of the latest master branch

~- [ ] I have made corresponding changes to the documentation~
~- [ ] I have added tests that prove my fix is effective or that my feature works~
- [ ] New and existing [**unit** tests](https://github.com/elastic/apm-server/blob/master/TESTING.md) pass locally with my changes

~- [ ] I have updated [CHANGELOG.asciidoc](https://github.com/elastic/apm-server/blob/master/CHANGELOG.asciidoc)~

## Related issues
[failing tests in master](https://apm-ci.elastic.co/blue/organizations/jenkins/apm-server%2Fapm-server-mbp/detail/master/382/pipeline)

PR #3456, opened 2020-03-10
Install modules into build directory for CI

simitt:
## Motivation/summary
Adapt to golang-mage Dockerfile to install modules into build directory and use accordingly in Jenkinsfile. 

## Checklist

~- [ ] My code follows the style guidelines of this project (run `make check-full` for static code checks and linting)~
~- [ ] I have rebased my changes on top of the latest master branch~

## How to test these changes
CI tests need to pass

## Related issues
follow up on https://github.com/elastic/apm-server/pull/3454/files#r390167891

### DIFF
--- a/.ci/docker/golang-mage/Dockerfile
+++ b/.ci/docker/golang-mage/Dockerfile
@@ -1,25 +1,11 @@
 ARG GO_VERSION=1.13.8
 FROM golang:${GO_VERSION}
 
-WORKDIR /tools
-
-RUN echo "Building from golang:${GO_VERSION}"
-RUN git clone https://github.com/magefile/mage \
-  && cd mage \
-  && go run bootstrap.go \
-  && mage -version \
-  && cd /tools \
-  && go get golang.org/x/tools/cmd/benchcmp \
-  && go get github.com/t-yuki/gocover-cobertura \
-  && go get github.com/jstemmer/go-junit-report \
-  && go get github.com/kardianos/govendor \
-  && go get github.com/elastic/go-licenser \
-  && go get github.com/pierrre/gotestcover \
-  && go get github.com/stretchr/testify/assert \
-  && go get golang.org/x/tools/cmd/goimports
+ENV TOOLS=/tools
+WORKDIR $TOOLS
+COPY go.mod .
+RUN go mod download
 
 RUN apt-get update -y -qq \
   && apt-get install -y -qq  python3 python3-pip python3-venv \
   && rm -rf /var/lib/apt/lists/*
-
-WORKDIR /go
--- a/Makefile
+++ b/Makefile
@@ -98,10 +98,6 @@ docker-system-tests: docker-compose.override.yml
 docker-compose.override.yml:
 	printf "version: '2.3'\nservices:\n beat:\n  build:\n   args: [UID=%d]" $(shell id -u) > $@
 
-.PHONY: test-packages-install
-test-packages-install:
-	@$(MAGE) -v testPackagesInstall
-
 ##############################################################################
 # Rules for updating config files, fields.yml, etc.
 ##############################################################################
--- a/Makefile
+++ b/Makefile
@@ -98,6 +98,10 @@ docker-system-tests: docker-compose.override.yml
 docker-compose.override.yml:
 	printf "version: '2.3'\nservices:\n beat:\n  build:\n   args: [UID=%d]" $(shell id -u) > $@
 
+.PHONY: test-packages-install
+test-packages-install:
+	@$(MAGE) -v testPackagesInstall
+
 ##############################################################################
 # Rules for updating config files, fields.yml, etc.
 ##############################################################################
@@ -142,7 +146,7 @@ update-beats-docs: $(PYTHON)
 	@$(PYTHON) script/copy-docs.py
 
 .PHONY: copy-docs
-copy-docs: 
+copy-docs:
 	@mkdir -p docs/data/intake-api/generated/sourcemap
 	@cp testdata/intake-v2/events.ndjson docs/data/intake-api/generated/
 	@cp testdata/intake-v3/rum_events.ndjson docs/data/intake-api/generated/rum_v3_events.ndjson
--- a/script/jenkins/build.sh
+++ b/script/jenkins/build.sh
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
 set -xeuo pipefail
 
-mage -debug build
+make apm-server
--- a/script/jenkins/package.sh
+++ b/script/jenkins/package.sh
@@ -5,4 +5,4 @@ set -euox pipefail
 # to prevent jenkins_release.sh from adding more PLATFORMS
 export PLATFORMS="${PLATFORMS:-+linux/amd64}"
 
-mage -debug package
+make release
--- a/script/jenkins/test-install-packages.sh
+++ b/script/jenkins/test-install-packages.sh
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
 set -xeuo pipefail
 
-make test-packages-install
+./build/linux/mage -v testPackagesInstall
--- a/script/jenkins/test-install-packages.sh
+++ b/script/jenkins/test-install-packages.sh
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
 set -xeuo pipefail
 
-mage -v -debug testPackagesInstall
+make test-packages-install
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -419,7 +419,7 @@ def golang(Closure body){
   def golangDocker
   retry(3) { // Retry in case there are any errors when building the docker images (to avoid temporary glitches)
     sleep randomNumber(min: 2, max: 5)
-    golangDocker = docker.build('golang-mage', "--build-arg GO_VERSION=${GO_VERSION} -f .ci/docker/golang-mage/Dockerfile ${BASE_DIR}")
+    golangDocker = docker.build('golang-mage', "--build-arg GO_VERSION=${GO_VERSION} -f  ${BASE_DIR}/.ci/docker/golang-mage/Dockerfile ${BASE_DIR}")
   }
   withEnv(["HOME=${WORKSPACE}", "GOPATH=${WORKSPACE}", "SHELL=/bin/bash"]) {
      golangDocker.inside('-v /usr/bin/docker:/usr/bin/docker -v /var/run/docker.sock:/var/run/docker.sock'){
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -419,7 +419,7 @@ def golang(Closure body){
   def golangDocker
   retry(3) { // Retry in case there are any errors when building the docker images (to avoid temporary glitches)
     sleep randomNumber(min: 2, max: 5)
-    golangDocker = docker.build('golang-mage', "--build-arg GO_VERSION=${GO_VERSION} ${BASE_DIR}/.ci/docker/golang-mage")
+    golangDocker = docker.build('golang-mage', "--build-arg GO_VERSION=${GO_VERSION} -f .ci/docker/golang-mage/Dockerfile ${BASE_DIR}")
   }
   withEnv(["HOME=${WORKSPACE}", "GOPATH=${WORKSPACE}", "SHELL=/bin/bash"]) {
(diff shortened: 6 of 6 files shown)


.ci/docker/golang-mage/Dockerfile
```diff
@@ -1,25 +1,11 @@
...
-  && go get github.com/jstemmer/go-junit-report \
-  && go get github.com/kardianos/govendor \
-  && go get github.com/elastic/go-licenser \
-  && go get github.com/pierrre/gotestcover \
-  && go get github.com/stretchr/testify/assert \
-  && go get golang.org/x/tools/cmd/goimports
+ENV TOOLS=/tools
+WORKDIR $TOOLS
+COPY go.mod .
+RUN go mod download
```
kuisathaverat: mage install is still needed

```suggestion
RUN go mod download
RUN git clone https://github.com/magefile/mage \	RUN go mod download
  && cd mage \	
  && go run bootstrap.go \	
  && mage -version
```

simitt: We wanted to avoid a global install, instead resolved the mage issues by using the make commands. (Thanks @axw for your help on this).

kuisathaverat: This container is used by the CI, it forces the CI to use fixed tools, do it should install mage. This container is a repeatable environment where we can build the apm-server. You can have Makefile magic to install mage if you want to help people to prepare the environment, but introduce those tools installed on the build process make it difficult to track and introduces errors not related to the build.

Makefile
```diff
@@ -98,6 +98,10 @@ docker-system-tests: docker-compose.override.yml
 docker-compose.override.yml:
 	printf "version: '2.3'\nservices:\n beat:\n  build:\n   args: [UID=%d]" $(shell id -u) > $@
 
+.PHONY: test-packages-install
```
axw: This is such a specific thing to have in the Makefile. How would you feel about reverting this, and changing `script/jenkins/test-install-packages.sh` as suggested below?

script/jenkins/test-install-packages.sh
```diff
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
 set -xeuo pipefail
 
-mage -v -debug testPackagesInstall
+make test-packages-install
```
axw:
```suggestion
./build/linux/mage test-packages-install
```

I'm not too keen on adding something like this to the Makefile, unless it's something we run regularly in day-to-day development. I didn't even know it existed until now, but I don't know - do you run this regularly yourself?

script/jenkins/build.sh
```diff
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
 set -xeuo pipefail
 
-mage -debug build
+make apm-server
```
kuisathaverat: we preferred to use mage build it is simple and does not have any no-build stuff

script/jenkins/package.sh
```diff
@@ -5,4 +5,4 @@ set -euox pipefail
 # to prevent jenkins_release.sh from adding more PLATFORMS
 export PLATFORMS="${PLATFORMS:-+linux/amd64}"
 
-mage -debug package
+make release
```
kuisathaverat: we preferred to use mage release it is simple and does not have any no-build stuff

script/jenkins/test-install-packages.sh
kuisathaverat: we preferred to use mage test-packages it is simple and does not have any no-build stuff

.ci/docker/golang-mage/Dockerfile
axw: @kuisathaverat we're using Go modules now, and we have mage as one of the entries in `go.mod`. When we build mage (as part of the Make command), it builds a specific version.

script/jenkins/build.sh
axw: What does "no-build stuff" mean?

axw: BTW, the previous Makefile was hideously complex, so I can understand why you wouldn't have wanted to rely on it. Take a look at the new one, it's a lot simpler now.

simitt: I am not sure I understand your concern, the `make` targets to which I changed to the same as the formerly used `mage` targets.

script/jenkins/test-install-packages.sh
simitt: I don't.

script/jenkins/build.sh
kuisathaverat: the concern is about that we move away from `make` to avoid make stuff not related to build/compile/test, the previous Makefile made some checks about tools install those tools and finally starts to compile/build/test, we want to avoid this and manage the things isolated.

.ci/docker/golang-mage/Dockerfile
kuisathaverat: I do not see the changes in the makefile here, so I guess that I have to take a look, but from my perspective have a `make` command that calls a `mage` command sounds a little complex, knowing that both are build tools with the same purpose.

script/jenkins/build.sh
axw: OK. The Makefile does setup tools, but it does so in a repeatable and deterministic manner, based on on the versions specified in `go.mod`. So the end result is the same, whether you install the tools first and then run `mage` directly, or run `make` which installs the tools and then runs `mage`.

I'd rather have one single build procedure that both CI and devs use. Please take a look at the new Makefile and let us know if you still have concerns.

.ci/docker/golang-mage/Dockerfile
simitt: Please take a look and let me know what you think. We tried to be in line with the go modules changes, but can go back to install mage globally for CI if it's important.

kuisathaverat: The new Makefile is much simpler. What makes me think about why we do not move everything to `mage` and remove the `Makefile`? this makes `mage` a pre-requisite but wi will have only one build definition `magefile.go` and only one build tool `mage`

kuisathaverat: I saw that `go mod download` does not install mage binary

```
root@c3553c27f088:/go/src/github.com/elastic/apm-server# find /go -name mage
/go/pkg/mod/github.com/magefile/mage@v1.9.0/mage
/go/pkg/mod/github.com/elastic/beats/v7@v7.0.0-alpha2.0.20200312020002-9735c6d51540/dev-tools/mage
/go/pkg/mod/github.com/elastic/beats/v7@v7.0.0-alpha2.0.20200312020002-9735c6d51540/metricbeat/scripts/mage
/go/pkg/mod/github.com/elastic/beats/v7@v7.0.0-alpha2.0.20200312020002-9735c6d51540/winlogbeat/scripts/mage
/go/pkg/mod/github.com/elastic/beats/v7@v7.0.0-alpha2.0.20200312020002-9735c6d51540/filebeat/scripts/mage
/go/pkg/mod/github.com/elastic/beats/v7@v7.0.0-alpha2.0.20200312020002-9735c6d51540/heartbeat/scripts/mage
/go/pkg/mod/github.com/elastic/beats/v7@v7.0.0-alpha2.0.20200312020002-9735c6d51540/packetbeat/scripts/mage
/go/pkg/mod/github.com/elastic/beats/v7@v7.0.0-alpha2.0.20200312020002-9735c6d51540/auditbeat/scripts/mage
/go/pkg/mod/github.com/elastic/beats/v7@v7.0.0-alpha2.0.20200312020002-9735c6d51540/x-pack/functionbeat/scripts/mage
/go/pkg/mod/github.com/elastic/beats/v7@v7.0.0-alpha2.0.20200306124118-bbf9d6697f4a/dev-tools/mage
/go/pkg/mod/github.com/elastic/beats/v7@v7.0.0-alpha2.0.20200306124118-bbf9d6697f4a/metricbeat/scripts/mage
/go/pkg/mod/github.com/elastic/beats/v7@v7.0.0-alpha2.0.20200306124118-bbf9d6697f4a/winlogbeat/scripts/mage
/go/pkg/mod/github.com/elastic/beats/v7@v7.0.0-alpha2.0.20200306124118-bbf9d6697f4a/filebeat/scripts/mage
/go/pkg/mod/github.com/elastic/beats/v7@v7.0.0-alpha2.0.20200306124118-bbf9d6697f4a/heartbeat/scripts/mage
/go/pkg/mod/github.com/elastic/beats/v7@v7.0.0-alpha2.0.20200306124118-bbf9d6697f4a/packetbeat/scripts/mage
/go/pkg/mod/github.com/elastic/beats/v7@v7.0.0-alpha2.0.20200306124118-bbf9d6697f4a/auditbeat/scripts/mage
/go/pkg/mod/github.com/elastic/beats/v7@v7.0.0-alpha2.0.20200306124118-bbf9d6697f4a/x-pack/functionbeat/scripts/mage
/go/pkg/mod/github.com/elastic/go-txfile@v0.0.7/dev-tools/lib/mage
/go/pkg/mod/cache/download/github.com/magefile/mage
```

kuisathaverat: There is a make target to build `mage`

```
BIN_MAGE=$(GOOSBUILD)/bin/mage

# BIN_MAGE is the standard "mage" binary.
$(BIN_MAGE): go.mod
	go build -o $@ github.com/magefile/mage
```

script/jenkins/build.sh
kuisathaverat: I saw it, we download the modules, then we run a  `make` target to build `mage`, and then we can use `mage`, complex form my POV

.ci/docker/golang-mage/Dockerfile
kuisathaverat:
```suggestion
RUN go mod download
RUN go build github.com/magefile/mage
```

script/jenkins/build.sh
axw: To be fair, the `go mod download` isn't necessary; it's an optimisation to avoid re-downloading each time the container is run. We could remove that and the rest would work, just more slowly.

Apart from that, the idea is to have everything encapsulated within the Make command: installing and running the required tools, in order to build the targets. This takes care of bootstrapping the build environment, to minimise the number of tools required in the environment to begin with. You need Go, Python, Make (`build-essential`), and that's about it.

(From the other comment thread)

> The new Makefile is much simpler. What makes me think about why we do not move everything to mage and remove the Makefile? this makes mage a pre-requisite but wi will have only one build definition magefile.go and only one build tool mage

A couple of reasons:
 - Make is a standard (and very mature) development tool in Linux, Mage is not.
 - If we require Mage to be installed globally, it makes it more difficult to rely on a specific version. It's true that we don't rely on a specific version of _Make_, but again, that's vastly more mature.

We can of course control the version in CI, but that doesn't help on developers' laptops.

Anyway, I do not see what the issue is here since we would be maintaining the Makefile either way, but if it's really a problem for you/CI then we can globally install and directly call mage just in CI. My concern is that the more you diverge CI from day-to-day development, the more likely it is that issues will escape dev and show up in CI.

kuisathaverat: >If we require Mage to be installed globally, it makes it more difficult to rely on a specific version. It's true that we don't rely on a specific version of Make, but again, that's vastly more mature.

You require it there are make targets that does not work without mage

>Anyway, I do not see what the issue is here since we would be maintaining the Makefile either way, but if it's really a problem for you/CI then we can globally install and directly call mage just in CI. 

you can make whatever you want with the makefile, mage file, python, ... 

>My concern is that the more you diverge CI from day-to-day development, the more likely it is that issues will escape dev and show up in CI.

Give us the target we have to execute from the CI to build and test the project, we will manage to execute them based on our need for the CI.

script/jenkins/test-install-packages.sh
```diff
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
 set -xeuo pipefail
 
-mage -v -debug testPackagesInstall
+./build/linux/mage -v testPackagesInstall
```
kuisathaverat: we can use directly `mage` if we merge https://github.com/elastic/apm-server/pull/3456/files#r392167309

kuisathaverat:
Let's move on if we include `mage` on the `golang-mage` container I'd agree with this PR

simitt:
> Let's move on if we include mage on the golang-mage container I'd agree with this PR

That's what we tried to avoid, as it would globally install mage.

kuisathaverat:
this `golang-mage` mage container is made on a daily basis and stored on the worker's cache, this means that we do not build the container on every build. So when `mage` is inside we do not need to download and build mage, the container in the cache has it.

PR #3457, opened 2020-03-10
[7.x] Fix CI Dockerfile paths for go get (#3454)

simitt:
 - Fix CI Dockerfile paths for go get (#3454)


axw:
Should we wait for vendor changes to be backported before merging?

axw:
Oh, but we have to have this at the same time, otherwise the backport can't build. Merging.

PR #3463, opened 2020-03-11
[7.x] add event.ingested to APM events (#3279)

jalvz:
 - add event.ingested to APM events (#3279)


jalvz:
jenkins retest this please

simitt:
Tested the following:

## Fresh install (ESS, on-prem): 
New pipeline is loaded and succesfully applied. This works as expected.

## Upgrade from 7.6 to 7.7 (ESS, tar file on-prem): 
New pipeline is loaded into Elasticsearch, but NOT applied as `apm-server.register.ingest.pipeline.overwrite: false` by default. Users can overwrite `apm` pipeline to include the new pipeline by setting `overwrite` to `true` and restarting the server or by running `./apm-server setup --pipelines -e -E apm-server.register.ingest.pipeline.overwrite=true`. By default the new pipeline is not added to the `apm` pipeline and not applied to ingested events, to not mess with potentially adapted pipeline setup from users. Therefore this is expected behavior. 

We should ensure highlighting the newly available pipeline in the release docs, and ensure the APM UI can handle missing `event.ingested` fields also when APM Server version >= 7.7.0.

dgieselaar:
We are not using event.ingested right now so it should be fine for the time
being.

Op ma 30 mrt. 2020 11:19 schreef Silvia Mitter <notifications@github.com>:

Issue #3458, opened 2020-03-10
[docs] Put the "reference" back into reference docs


bmorelli25:
As a prerequisite to the Observability documentation overhaul, we should begin to:

* Separate non-reference content from reference content
* Improve TOC headings to be easier to scan
* Sort lists in a consistent manner (alphabetically)

## Beats prereqs

- [ ] Make update beats docs
- [ ] Simplify Beats documentation headers (and use consistent wording) https://github.com/elastic/beats/pull/16825

## APM Server work (more to be added)

- [ ] Break configuration section into two different sections: `Configuration` & `How to`
- [ ] Sort lists alphabetically where feasible
- [ ] Improve scanability of the TOC by adding shorter nav titles with title abbreviations to APM specific topics:
```
== Long descriptive page title for SEO
++++
<titleabbrev>Short TOC header</titleabbrev>
++++
```

## Related Issues
* Beats Reference content: https://github.com/elastic/beats/issues/16821
* Beats restructure and refactor: https://github.com/elastic/beats/pull/16825

PR #3492, opened 2020-03-16
[docs] Reference structure and heading changes

bmorelli25:
## Motivation/summary
As a prerequisite to the Observability documentation overhaul, this PR begins to:

* Separate non-reference content from reference content -- This PR moves _guide_ (non-reference) content under a new **How to guides** section.
* Improve TOC headings to be easier to scan -- Abbreviated titles make content easier to scan. Better yet, Asciidoctor allows us to abbreviate titles in the TOC, while maintaining longer headings on individual pages for SEO purposes.

## Live preview

A live preview of this PR is available **[here](http://apm-server_3492.docs-preview.app.elstc.co/diff)**.

## TOC/ Structure changes

Current --> Proposed
Screen Shot 2020-03-16 at 2 35 06 PM

The new **How to guides** section would contain:
Screen Shot 2020-03-16 at 3 23 49 PM

## To do

~Reference config still needs to be removed from sections like Jaeger, ILM, and maybe others?~
~The _manage storage_ section is still a mix of reference/guide~

## Related issues
For https://github.com/elastic/apm-server/issues/3458.


simitt:
The new overview of the structure looks great!

bmorelli25:
This PR is ready for a review. Please click around the [live preview](http://apm-server_3492.docs-preview.app.elstc.co/diff) if you can. I'm especially interested in feedback on the separation of reference from non-reference content in the ILM and Jaeger how-tos. Because this how to content still lives in the APM Server reference, _why_ the content is separated may not be clear to users. If you feel the same way, I can revert the commits in this PR where I made those changes (6f51370 / f240b0a), and we can just move forward with the structure and heading changes.

jalvz:
Hey, take my comment here with a grain of salt:

> Improve TOC headings to be easier to scan 

I do like the new headings more, and also like that there are less entries at the top level.


> Separate non-reference content from reference content

I think that is a noble goal, _but_...:
- The how-to guides are still in the middle of the "APM Server reference".
- They still describe configuration, so not sure how guide-like they are, really. They are more verbose because the touch on more complicated stuff, but there isn't anything there like real-word user stories, how to leverage apm-server for use case X, or things like that.
- Ultimately, if the goal to help users to navigate better the docs, I am not sure this will really stand out.
 
I don't know what other changes are planned, so maybe this PR makes more sense in the future. All said, some ideas:

- API and Exported fields are pure reference-like sections, quite boring and not particularly useful to most readers. I'd be happy for those to be buried under _somewhere_.

- Maybe make the How to guides stand out a bit more. Like putting it at the end, renaming it to HOWTO (all uppercase), or re-purposing it as a FAQ, or Examples... I think in this case all the important content should still be described in the Configure section, even if briefly; and then expanded/elaborated here.  

An unrelated note (while at it): Does the upgrade section needs an entry for 6.5 and 7.0 upgrades on each stack version? 
Unless I miss something, those sections could removed from all docs since 7.1, and add quick note with a link maybe in the Install section, or in the release notes section: "If you are upgrading from 6.x, check out..."
The breaking changes subsection could be moved as well to the release notes, and ditch the whole Upgrade section.


I am not stopping you from moving forward with this, thou :) Feel free to merge and keep iterating!

bmorelli25:
Really great feedback, thank you both!

@axw ...

> I'm not sure if all of those how-to guides fit in the APM Server docs, but perhaps some of them?

Agreed. In fact, the long-term goal is for _none_ of these guides to live in the APM Server reference. They'll be separated based on whether they are admin/analyst tasks, and live in a much easier to find location (see this [internal link](https://docs.google.com/document/d/1jOTSdgFH_qStDC5p0Plolc8Rv1WU4ZXuEJgzouLXuOk/edit#) for the proposed structure).

> The Jaeger one in particular I think should be at a higher level, as being buried in the APM Server docs requires some understanding of our architecture first. Particularly if/when we have support in ESS where the APM Server is less prominent, I think having to know to look in the server docs requires too much knowledge of the details. Having said that, I don't know where it should live. Maybe in APM Server for now, with a view to relocate to a higher level page later?

Agreed. Similar answer as before: this wont be the final location of these guides. This is more of a triage location for our content as we adapt what we have into how-to guides, and begin to create more task-focussed documentation. As a short-term solution, maybe we could start to move this how-to content to the APM Overview?

> Perhaps "Use environment variables in the configuration" should just go under the Configure reference docs, as it's not really outcome oriented?

Fair point.

> Does "Enable Real User Monitoring (RUM)" deserve its own how-to page in the APM Server? I think "Upload source Maps" (weird capitalisation btw?) is a good one though – particularly if we expand it to be a more thorough walkthrough of uploading a source map, and then showing the outcome of it being applied.

As-is, it probably doesn't. I was thinking it'd make sense to have a high level overview of RUM and then have child guides underneath it, but we don't really have the content yet.

Great idea with the source map improvements. I'll open an issue.

@jalvz ...

> The how-to guides are still in the middle of the "APM Server reference".
They still describe configuration, so not sure how guide-like they are, really. They are more verbose because the touch on more complicated stuff, but there isn't anything there like real-word user stories, how to leverage apm-server for use case X, or things like that.
Ultimately, if the goal to help users to navigate better the docs, I am not sure this will really stand out.
I don't know what other changes are planned, so maybe this PR makes more sense in the future.

I don't disagree with anything you've said. I think I could've done a better job explaining the context around this PR. The long-term goal is for _none_ of these guides to live in the APM Server reference. They'll be separated based on whether they are admin/analyst tasks, and live in a much easier to find location (see this [internal link](https://docs.google.com/document/d/1jOTSdgFH_qStDC5p0Plolc8Rv1WU4ZXuEJgzouLXuOk/edit#) for the proposed structure). This is an attempt at step one: creating a location for this kind of content to live, and adapting the content we already have, where possible, to fit this mold.

> Maybe make the How to guides stand out a bit more. Like putting it at the end, renaming it to HOWTO (all uppercase), or re-purposing it as a FAQ, or Examples... I think in this case all the important content should still be described in the Configure section, even if briefly; and then expanded/elaborated here.

Hmmm, I wonder if moving it up the list might help it stand out. Like:
* Overview
* Get started
* Set up
* How to

I'm hesitant to repurpose content as this is supposed to be a step towards our end-goal for documentation.

> An unrelated note (while at it): Does the upgrade section needs an entry for 6.5 and 7.0 upgrades on each stack version? Unless I miss something, those sections could removed from all docs since 7.1, and add quick note with a link maybe in the Install section, or in the release notes section: "If you are upgrading from 6.x, check out..." The breaking changes subsection could be moved as well to the release notes, and ditch the whole Upgrade section.

I think the argument for keeping the upgrade guides around is that google heavily favors our `current` documentation. So users are likely to end up on the 7.6 docs, regardless of the version they use. If we only include upgrade content in certain branches, we may be making it too difficult to find, which could lead to an increase in the number of problems people have. With that being said... maybe you're on to something with moving the upgrade/release/breaking changes content around. Let me play with it.

axw:
> Agreed. Similar answer as before: this wont be the final location of these guides. This is more of a triage location for our content as we adapt what we have into how-to guides, and begin to create more task-focussed documentation. As a short-term solution, maybe we could start to move this how-to content to the APM Overview?

In that case, I'm not bothered. I don't think it matters too much if it lives here or in APM Overview in the interim.

bmorelli25:
Sounds like we're all on board. I'll merge this tomorrow unless anyone objects.

Issue #3498, opened 2020-03-18
client_authentication is not working as described in reference


insukcho:
I'm trying to set up a self-managed Elastic APM Server in my machine for the test.

When we check the reference [here](https://www.elastic.co/guide/en/apm/server/current/agent-server-ssl.html#_client_authentication), we can see below comments.

**client_authentication**
_This configures what types of client authentication are supported. The valid options are none, optional, and required. The default is optional. If certificate_authorities has been specified, this setting will automatically change to required._
- none - Disables client authentication.
- optional - When a client certificate is given, the server will verify it.
- required - Requires clients to provide a valid certificate.

So, if a client certificate is given, `optional` and `required` should be working as same, right?

But when I send data from APM Python Agent (Django/Flask), APM Java Agent on Tomcat, only `optional` was working but `required` give me `TLS handshake error` like below.
```
http: TLS handshake error from 127.0.0.1:54321: remote error: tls: unsupported extension
```

You can reproduce this with below steps.

1) Create self-assigned certificates (PEM format) with [elasticsearch-certutil](https://www.elastic.co/guide/en/elasticsearch/reference/current/certutil.html#certutil)
```
$ vi ./instance.yml
instances:
  - name: 'client'
    dns: [ 'localhost' ]

$  bin/elasticsearch-certutil cert ca --pem --in ./instance.yml --out ./certs.zip
```

2) Unzip ./certs.zip file and use copy and paste under APM Server home directory. Put SSL configuration in apm-server.yml like below.
```
apm-server.ssl.enabled: true
apm-server.ssl.key: "./certs/client/client.key"
apm-server.ssl.certificate: "./certs/client/client.crt"
apm-server.ssl.certificate_authorities: ['./certs/ca/ca.crt']
apm-server.ssl.secret_token: ${SECRET_TOKEN}
 
# Currently, this setup is required. If not setting up, it will fail for SSL hand shaking.
apm-server.ssl.client_authentication: 'optional'
 
# For Java Agents, we need to limit supported protocol to avoid using TLSv1.3
supported_protocols: [TLSv1.1, TLSv1.2]
```

3) Copied same certificates to each Agent's home directory. and change configuration like below.
- Java Agent on Tomcat
Need to change CA certificate to DER, and then store it in JKS(Java KeyStore). Set it up in `$TOMCAT_HOME/bin/setenv.sh`
```
[...]

Issue #3494, opened 2020-03-17
[docs] Clarify client certificate related TLS config options


simitt:
There are two server side TLS configuration options related to verifying client certificates:
* `apm-server.ssl.certificate_authorities`
* `apm-server.ssl.client_authentication`   

We've seen some confusion around these configuration options. As requested by @beniwohli 
these options should be moved to a dedicated section, where it becomes clear that they should only be customized when the agent provides a client certificate. None of the backend agents currently supports this. This is only useful for the RUM agent, when a client certificate is configured to be presented in the browser.

beniwohli:
See also https://github.com/elastic/apm-agent-python/issues/758

simitt:
Update: the option can also be used by Jaeger agents.

Issue #3442, opened 2020-03-06
Avoid requesting certification on the RUM endpoint


jahtalab:
When the `client_authentication` is set to `optional` (the default value) the APM server seems to request client certification on all endpoints. When the request is made from the RUM agent, requesting a certification would result in the browser to prompt the user to choose a certification to present to the server. The request still succeeds if the user cancels this prompt, since presenting certification is optional, but if the user chooses an invalid certification the request would fail.

I propose to avoid requesting certification on the RUM endpoint when `client_authentication` is set to `optional`.

simitt:
I can confirm that the server sends a certificate request when `ssl.client_authentication: optional`. If no certificate is presented, it doesn't lead to an error though.

simitt:
The `golang` implementation does request a client certificate when `ClientAuth >= RequestClientCert`, see [crypto/tls/handshake_server.go](https://github.com/golang/go/blob/master/src/crypto/tls/handshake_server.go#L486) (same for the TLS 1.3 implementation).

Valid options for `ClientAuth` are: 
	* `NoClientCert`
	* `RequestClientCert`
	* `RequireAnyClientCert`
	* `VerifyClientCertIfGiven`
	* `RequireAndVerifyClientCert`

If setting the server config to `NoClientCert` a provided certificate would never be verified, this is the behavior when configuring `apm-server.ssl.client_authentication: "none"`. This is not an option for `apm-server.ssl.client_authentication: "optional"`, as given certs need to be verified.

We could switch the default configuration to `apm-server.ssl.client_authentication: "none"`. As I see this as a breaking change, I suggest to change in 8.0 and document this caveat for the RUM agent until then.

wolframhaussig:
Why not use the option `VerifyClientCertIfGiven` you mentioned? Don't ask for a certificate as it is optional but verify it if it is given?

simitt:
That's actually what happens: `apm-server.ssl.client_authenticaion: "optional"` is translated to `VerifyClientCertIfGiven`. As linked above, a client certificate is requested though for every option except for `NoClientCert`
```
if c.config.ClientAuth >= RequestClientCert {
    // Request a client certificate
    ...
}
```

simitt:
Tested with Firefox and Chrome and could only observe Chrome to print for a certificate. There is no change in behavior testing for `7.6` and `6.8`.

wolframhaussig:
[...]

Issue #3198, opened 2020-01-16
Clarify `apm-server.ssl.certificate_authorities` 


simitt:
Explicitly mention that `apm-server.ssl.certificate_authorities` should only be set when the agent is sending a client certificate, and cross reference to [client_authentication](https://www.elastic.co/guide/en/apm/server/current/agent-server-ssl.html#_client_authentication) being required, if it is set.

Also add a sentence for clarification in the [apm-server.yml](https://github.com/elastic/apm-server/blob/master/apm-server.yml#L97)

karmi:
I'd add something like this to the `apm-server.yml`:

```
# Configure a list of root certificate authorities for verifying client certificates.
# Enable ONLY when you want to use and verify client certificates.
#certificate_authorities: []
```

insukcho:
> If certificate_authorities has been specified, this setting will automatically change to required.

@bmorelli25 This comment is from https://www.elastic.co/guide/en/apm/server/current/agent-server-ssl.html#_client_authentication.

Could you please add this in [apm-server.yml](https://github.com/elastic/apm-server/blob/master/apm-server.yml#L125) something like below? 

```
    # Configure which type of client authentication is supported.
    # Options are `none`, `optional`, and `required`. Default is `optional`.
    # If certificate_authorities has been specified, this setting will 
    # automatically change to `required`.
    #client_authentication: "optional"
```

simitt:
closed with #3500

PR #3500, opened 2020-03-18
Changing TLS client_authentication default and cleaning up docs

simitt:
## Motivation/summary
This PR covers a couple of changes related to APM Server's SSL settings for communicating with the agents. 

It changes the default value for `apm-server.ssl.client_authentication` from `optional` to `none`. Sending client certificates is not supported by any of the backend agents. It can be used by the RUM agent when a client certificate is configured in the browser. It can also be used by Jaeger agents when connecting via gRPC, which is an experimental feature so far. 
The default being set to `optional` does not prevent anyone from connecting to the server without providing a certificate. If users want to rely on client certification verification the value should be changed to `required`. For more details see https://github.com/elastic/apm-server/issues/3442#issuecomment-600628214. 

The PR also aims to clarify the documentation around the client certificate related configuration options (`apm-server.ssl.client_authentication` and `apm-server.ssl.certificate_authorities`) in the `apm-server.yml` and the docs. See https://github.com/elastic/apm-server/issues/3498 and https://github.com/elastic/apm-server/issues/3494.

This PR also removes the configuration option `apm-server.ssl.verification_mode` from the `apm-server.yml` and the docs. The [`VerificationMode`](https://github.com/elastic/beats/blob/master/libbeat/common/transport/tlscommon/server_config.go#L85) gets internally translated to [`tls.Config InsecureSkipVerify`](https://github.com/elastic/beats/blob/master/libbeat/common/transport/tlscommon/tls_config.go#L99). 
From the [golang/src/crypto/tls/common.go `InsecureSkipVerify`](https://github.com/golang/go/blob/master/src/crypto/tls/common.go#L548):
> InsecureSkipVerify controls whether a client verifies the
> server's certificate chain and host name.
> If InsecureSkipVerify is true, TLS accepts any certificate
> presented by the server and any host name in that certificate.
> In this mode, TLS is susceptible to man-in-the-middle attacks.
> This should be used only for testing.

Configuring the verification mode on the server does not have any effect. 

## Checklist

## How to test these changes

Start APM Server with TLS and make a request from the browser (Chrome). Ensure no prompt for chosing a certificate appears.

## Related issues
https://github.com/elastic/apm-server/issues/3198
https://github.com/elastic/apm-server/issues/3494
https://github.com/elastic/apm-server/issues/3498
https://github.com/elastic/apm-server/issues/3442


beater/config/config_test.go
```diff
@@ -404,32 +404,6 @@ func TestTLSSettings(t *testing.T) {
 		}
 	})
 
-	t.Run("VerificationMode", func(t *testing.T) {
```
axw: I'm on board with removing docs/references to the config, as it's only meant for testing purposes, but I think we should keep the tests since the config will still take effect if someone sets it IIANM.

axw: Disregard, I didn't look closely enough. As you said, it's not relevant for `apm-server.ssl.*`

simitt:
Jenkins, run the tests please

PR #3513, opened 2020-03-19
[RUM V3] Require minimum version

### LINKED ISSUE #3512: [RUM V3] Require minimum version.
RUM V3 PR's are being backported normally, but the handler won't work in 7.7.
We should make that clear by explicitly disabling it.

jalvz:
Closes #3512


jalvz:
I rather do not diverge the branch from master more than what is strictly necessary. Besides, this returns a more user-friendly error to a 7.7 customer, instead of a generic `404`.

jalvz:
I agree, but also think we should help users, if it is at our hand. The middleware doesn't need to be removed after release for it to work. 
Will merge this, thanks.

simitt:
The middleware does not _need_ to be removed, but it should, as it doesn't provide any value once released, but adds another layer every request needs to go through.

PR #3518, opened 2020-03-20
[7.x] Require a minimum version in the RUM V3 handler. (#3513)

jalvz:
 - Require a minimum version in the RUM V3 handler. (#3513)

### DIFF
--- a/beater/api/mux.go
+++ b/beater/api/mux.go
@@ -19,8 +19,12 @@ package api
 
 import (
 	"expvar"
+	"fmt"
 	"net/http"
 	"regexp"
+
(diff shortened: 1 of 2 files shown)


simitt:
## Tested (on-prem, ESS):
The endpoint is disabled in `7.7` as expected, but the received error message does not contain the upgrade version. 
```
{
  "error": "forbidden request: RUM V3 endpoint is disabled in this version. Update to  or higher"
}
```

simitt:
@jalvz maybe even change this to not suggest upgrading to a concrete version (returning a concrete version is basically committing to the feature in 7.8).

PR #3519, opened 2020-03-22
[7.x] all: propagate context through Transformables (#3515)

axw:
 - all: propagate context through Transformables (#3515)


axw:
jenkins run the tests please

Issue #3473, opened 2020-03-12
[RUM v3] Adapt marks


jalvz:
Field mappings are defined in https://github.com/elastic/apm-server/pull/3414/files#diff-d366d2c620a50e51efced6c70769c248

We need to build the logic in the decoders to transform field names, and update the spec accordingly if needed / desired.

PR #3526, opened 2020-03-23
Spec out transaction marks for RUM V3

### LINKED ISSUE #3475: [RUM v3] Add metricset handler
Currently metricsets are not implemented for RUM v3. We need to put them in.

Field name mappings are already defined in https://github.com/elastic/apm-server/pull/3414/files#diff-d366d2c620a50e51efced6c70769c248

jalvz:
Closes elastic/apm-server#3473

### DIFF
--- a/docs/spec/transactions/rum_v3_mark.json
+++ b/docs/spec/transactions/rum_v3_mark.json
@@ -1,0 +1,107 @@
+{
+    "$id": "docs/spec/transactions/rum_v3_mark.json",
+    "type": ["object", "null"],
+    "description": "A mark captures the timing in milliseconds of a significant event during the lifetime of a transaction. Every mark is a simple key value pair, where the value has to be a number, and can be set by the user or the agent.",
+    "properties": {
+        "a": {
+            "type": ["object", "null"],
+            "description": "agent",
+            "properties": {
+                "dc": {
+                    "type": ["number", "null"],
+                    "description": "domComplete"
+                },
+                "di": {
+                    "type": ["number", "null"],
+                    "description": "domInteractive"
+                },
+                "ds": {
+                    "type": ["number", "null"],
+                    "description": "domContentLoadedEventStart"
+                },
+                "de": {
+                    "type": ["number", "null"],
+                    "description": "domContentLoadedEventEnd"
+                },
+                "fb": {
+                    "type": ["number", "null"],
+                    "description": "timeToFirstByte"
+                },
+                "fp": {
+                    "type": ["number", "null"],
+                    "description": "firstContentfulPaint"
+                },
+                "lp": {
+                    "type": ["number", "null"],
+                    "description": "largestContentfulPaint"
+                }
+            }
+        },
+        "nt": {
+            "type": ["object", "null"],
+            "description": "navigation-timing",
+            "properties": {
+                "fs": {
+                    "type": ["number", "null"],
+                    "description": "fetchStart"
+                },
+                "ls": {
+                    "type": ["number", "null"],
+                    "description": "domainLookupStart"
+                },
+                "le": {
+                    "type": ["number", "null"],
+                    "description": "domainLookupEnd"
+                },
+                "cs": {
+                    "type": ["number", "null"],
+                    "description": "connectStart"
+                },
+                "ce": {
+                    "type": ["number", "null"],
+                    "description": "connectEnd"
+                },
+                "qs": {
+                    "type": ["number", "null"],
+                    "description": "requestStart"
+                },
+                "rs": {
+                    "type": ["number", "null"],
+                    "description": "responseStart"
+                },
+                "re": {
+                    "type": ["number", "null"],
+                    "description": "responseEnd"
+                },
+                "dl": {
+                    "type": ["number", "null"],
+                    "description": "domLoading"
+                },
+                "di": {
+                    "type": ["number", "null"],
+                    "description": "domInteractive"
+                },
+                "ds": {
+                    "type": ["number", "null"],
+                    "description": "domContentLoadedEventStart"
+                },
+                "de": {
+                    "type": ["number", "null"],
+                    "description": "domContentLoadedEventEnd"
+                },
+                "dc": {
+                    "type": ["number", "null"],
+                    "description": "domComplete"
+                },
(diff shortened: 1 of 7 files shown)


simitt:
The referenced issue describes `metricsets` but this PR concerns `transaction.marks`. Have you linked the right Issue?

model/transaction/event.go
```diff
@@ -91,7 +91,75 @@ type SpanCount struct {
...
+	event, _ := transformable.(*Event)
+	raw, _ := input.(map[string]interface{})
+
+	decoder := &utility.ManualDecoder{}
+	fieldName := field.Mapper(cfg.HasShortFieldNames)
+	marks := fieldName("marks")
+
+	agent := fieldName("agent")
+	agentMarks := decoder.MapStr(raw, agent, marks)
+	agentMarks["domComplete"] = decoder.Float64Ptr(raw, fieldName("domComplete"), marks, agent)
```
simitt: This can be made more DRY to avoid repeating every key 3 times. Something similar to
```
	decoder := &utility.ManualDecoder{}
	fieldName := field.Mapper(cfg.HasShortFieldNames)

	resolveMarks := func(m common.MapStr, key string) {
		mappedKey := fieldName(key)
		m[key] = decoder.Float64Ptr(m, mappedKey)
		delete(m, mappedKey)
	}

	marks := fieldName("marks")

	agent := fieldName("agent")
	agentMarks := decoder.MapStr(raw, agent, marks)
	resolveMarks(agentMarks, "domComplete")
	resolveMarks(agentMarks, "domInteractive")
	resolveMarks(agentMarks, "domContentLoadedEventStart")
	resolveMarks(agentMarks, "domContentLoadedEventEnd")
```

docs/spec/transactions/rum_v3_mark.json
```diff
@@ -0,0 +1,108 @@
+{
+    "$id": "docs/spec/transactions/rum_v3_mark.json",
+    "type": ["object", "null"],
+    "description": "A mark captures the timing in milliseconds of a significant event during the lifetime of a transaction. Every mark is a simple key value pair, where the value has to be a number, and can be set by the user or the agent.",
+    "patternProperties": {
+        "^[^.*\"]*$": {
+            "type": ["number", "null"]
+        }
+    },
+    "properties": {
```
simitt: These defined properties don't behave as expected, the testdata do not match against them. You can try it out by removing the above `patternProperties` section and sending the testdata - they are invalid. The testdata match the `patternProperties` but not the newly defined attributes.  

Looking at [rum_v3_transactions](https://github.com/elastic/apm-server/blob/7d5cfbf872f6df6decc7ca7979db3104578e99b0/docs/spec/transactions/rum_v3_transaction.json#L79) , `marks` are defined with 
```
               "k": {
                    "type": [
                        "object",
                        "null"
                    ],
                    "description": "A mark captures the timing of a significant event during the lifetime of a transaction. Marks are organized into groups and can be set by the user or the agent.",
                    "patternProperties": {
                        "^[^.*\"]*$": {
                            "$ref": "rum_v3_mark.json"
                        }
                    },
                    "additionalProperties": false
```
Everything defined in `rum_v3_mark.json` is therefore nested inside these properties, which is one nesting level too deep, when specifying the mark grouping (`agent`, `navigation_timing`) here.

The reason it was defined as is, was to match all marks without further specification, allowing the RUM agent devs to just add marks on the fly. The marks are defined as `objects` (first patternProperties definition in the transaction file) containing `numbers` (the second patternProperties in the marks file). 

From the changes you made I assume you want to keep the current behavior but additionally spec out the well known shortened names for easier guidance for RUM agent devs. If the assumption is right, you could solve it with something similar to https://github.com/jalvz/apm-server/compare/rumv3-marks...simitt:json-schema-v3-marks?expand=1

jalvz: Thanks, I noticed something was off indeed, but didn't came as far yesterday. I'll pick the necessary changes from your diff and later create a separate PR for cleaning up the schemas (removing the `allOf`, etc).

jalvz: Pushed a fix. I think we want to allow extra fields bellow `marks` ("k"), `agent` ("a"), and `navigation-timing` ("nt"), but I'll double check.
The fix is incomplete, thou. Next problem is that those fields can be strings too. I am not sure how flexible json-schema's `anyOf` is when it comes to merging rules....

simitt: The proposed solution did allow for additional attributes nested under `marks`,  `agent` and `navigation-timing`, as long as they were matched by the defined properties or the `patternProperties` and being `numbers`. The `additionalProperties: false` ensured that attributes not matching the defined or pattern properties are rejected.

Transaction marks being allowed to be `string` values is a new requirement then, so far they were restricted to being `numbers`. But that should be easily solvable by adding `string` to the allowed types for all the properties.

jalvz: > The proposed solution did allow for additional attributes nested under marks

ok, i think i didn't manage to get that working locally, will check again tomorrow.

> Transaction marks being allowed to be string values is a new requirement

no, i meant it accepts strings when it shouldn't (so it is wrong)

simitt: It occurred to me that with the changes for the RUM Intake it might not be a good idea to process _undefined_ marks.
When the RUM agent sends additional marks with shortened keys in future versions, the server will only be able to map to something meaningful in an upcoming version. But older server receiving additional marks will just store them. The additional marks won't be easily readable being indexed with shortened keys, and they also won't be comparable with marks processed by newer servers, as the keys will be different. 
Therefore my question, what is the intention of still processing and indexing unknown marks and wouldn't it make more sense to allow but ignore them?

jalvz: Clarified that with them: allow & ignore.

docs/spec/transactions/rum_v3_mark.json
```diff
@@ -0,0 +1,107 @@
+{
```
jalvz: @jahtalab @vigneshshanmugam fyi there is a json-schema spec for RUM V3

Issue #3536, opened 2020-03-24
Add script verifying libbeat dependency can be upgraded


simitt:
In the past we had CI scripts running checking that libbeat can be updated to the latest according version (`master`, `7.x`). The script being removed in https://github.com/elastic/apm-server/pull/3371 seems unintentional. The usage was commented out in the `Jenkinsfile`, suggesting the script not being used anymore.
However, it is still run from https://beats-ci.elastic.co/job/elastic+apm-server+7.x+beats-update/ and https://beats-ci.elastic.co/job/elastic+apm-server+master+beats-update/

The test runs still have the same value than before the cleanup and switching to go modules.

axw:
Oops, sorry. I intentionally removed it thinking it was unused.

graphaelli:
I think https://github.com/elastic/beats/pull/17218 resolves this issue.

PR #3371, opened 2020-02-24
Simplify build scripts

axw:
## Motivation/summary

- Remove use of "gotestcover" in our build scripts (still used in the libbeat Makefile). This tool has been made redundant by recent versions of Go, which are capable of running tests for multiple packages while capturing coverage.
- Remove a couple of unused build scripts ("ci.sh" and "update-beats.sh"), and remove the commented-out update-beats.sh block from Jenkinsfile. That can be added back later if needed.
- Simplify build_apm_docs.sh, and update to stop assuming use of $GOPATH. This will enable migration to modules.

## Checklist

- [x] I have signed the [Contributor License Agreement](https://www.elastic.co/contributor-agreement/).
- [x] My code follows the style guidelines of this project (run `make check-full` for static code checks and linting)
- [x] I have rebased my changes on top of the latest master branch
~- [ ] I have made corresponding changes to the documentation~
~- [ ] I have added tests that prove my fix is effective or that my feature works~
~- [ ] New and existing [**unit** tests](https://github.com/elastic/apm-server/blob/master/TESTING.md) pass locally with my changes~
~- [ ] I have updated [CHANGELOG.asciidoc](https://github.com/elastic/apm-server/blob/master/CHANGELOG.asciidoc)~ (no user-facing changes)

## How to test these changes

- `make apm-docs`
- Run through CI

## Related issues

Pulled out of #3368 
Relates #3296

PR #3537, opened 2020-03-24
fix: restore beats-update.sh script

kuisathaverat:
## Motivation/summary

It restores a deleted script on #3371, this script is used on beats-ci to test updates on the APM Server, so for the moment, it is not possible to remove it.

## Related issues

Closes https://github.com/elastic/apm-server/issues/3536 https://github.com/elastic/beats/issues/16129

### DIFF
--- a/script/update_beats.sh
+++ b/script/update_beats.sh
@@ -1,0 +1,87 @@
+#!/usr/bin/env bash
+
+set -ex
+
+BEATS_VERSION="${BEATS_VERSION:-master}"
+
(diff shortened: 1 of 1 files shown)


graphaelli:
I don't think this will work on master or 7.7 - can the caller use `make update-beats` instead?

kuisathaverat:
I guess so, if it makes the same thing

kuisathaverat:
I am testing the make beats-update,  we can close this one

Issue #3567, opened 2020-03-27
Proposal: model restructure


axw:
We currently have a `model` package which serves two purposes: schema validation and decoding of events from Elastic APM agents into our object model; and transforming our object model into beat.Events.

I propose that we split these apart into two distinct packages, and move the schema validation and decoding logic into `processor/stream`, or as a sub-package thereof. The `model` package would continue to hold our object model types, as well as the beat.Event transformation logic.

axw:
Additionally:
 - add Metadata fields to model types; decoder will be responsible for setting them to the merged stream/event metadata. Currently we send the metadata through to the Transform method - rather circuitous
 - calculate event timestamps from the request time in the decoder. The model types should have correct timestamps by the time we get to transformation. The RequestTime logic only makes sense for RUM, so let's isolate it there

axw:
We should also move some of the repeated logic in decoders up into processor/stream: https://github.com/elastic/apm-server/pull/3570#discussion_r400187492

axw:
Package and/or functions need renaming: https://github.com/elastic/apm-server/pull/3614#pullrequestreview-387168482

Not sure how yet. I'm thinking we should create a `Decoder` type, and make all of these methods of it: Decoder.DecodeMetadata, Decoder.DecodeTransaction, etc.

If we rename `processor/stream` to `input/elasticapm`, it would probably make sense to move the Decoder type directly into that package.

PR #3570, opened 2020-03-29
Introduce model.Input for decoding

axw:
## Motivation/summary

Introduce the model.Input type to pass additional context to decoders, containing:
 - the raw decoded JSON input
 - the "request time", i.e. the time at which the input was received
 - decoding configuration -- temporarily, until we have a configurable Decoder type

RequestTime has been removed from transform.Context, and the decoders are now responsible for setting the Timestamp on the model types according to the input semantics.

In a follow-up, I would do the same thing for Metadata: add a Metadata field to model.Input, which would be initialised to the stream metadata value, and remove transform.Context's Metadata field.

The motivation behind these changes is to make the model types complete and self contained, and creating a clear separation of logic between decoding and transformation:
 - after decoding (from Elastic APM agents, Jaeger, ...), the model types should be complete, with the exception of post-processing enrichment such as source mapping
 - transformation would only be responsible for translating to `beat.Events` (perhaps still performing enrichment inline)

## Checklist

## Related issues

See https://github.com/elastic/apm-server/issues/3567

### DIFF
--- a/model/span/event.go
+++ b/model/span/event.go
@@ -273,23 +273,23 @@ func (d *DestinationService) fields() common.MapStr {
 	return fields
 }
 
-func DecodeRUMV3Event(input interface{}, cfg m.Config, err error) (transform.Transformable, error) {
-	return DecodeEvent(input, cfg, err)
+func DecodeRUMV3Event(input m.Input, err error) (transform.Transformable, error) {
+	return DecodeEvent(input, err)
 }
 
 // DecodeEvent decodes a span event.
-func DecodeEvent(input interface{}, cfg m.Config, err error) (transform.Transformable, error) {
+func DecodeEvent(input m.Input, err error) (transform.Transformable, error) {
 	if err != nil {
 		return nil, err
 	}
-	if input == nil {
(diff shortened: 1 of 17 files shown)


model/span/event.go
```diff
@@ -273,23 +273,23 @@ func (d *DestinationService) fields() common.MapStr {
...
+func DecodeEvent(input m.Input, err error) (transform.Transformable, error) {
 	if err != nil {
 		return nil, err
 	}
-	if input == nil {
+	if input.Raw == nil {
 		return nil, errMissingInput
 	}
-	raw, ok := input.(map[string]interface{})
+	raw, ok := input.Raw.(map[string]interface{})
```
jalvz: Maybe move all this logic duplicated on each event to somewhere around https://github.com/elastic/apm-server/pull/3570/files#diff-01c58acd54a543ec500d1ebf9e7bbe94R191 (ie. let `input.Raw` be a non nil `map[string]interface{}`

axw: Sounds good, I will do that in a follow-up.

Issue #3485, opened 2020-03-15
Introduce transaction histogram metrics


axw:
## Motivation / summary

Currently the way Elastic APM works is by recording every single transaction as an individual document - including "unsampled transactions". In order to compute statistics, we use Elasticsearch's aggregations framework to query over the documents in real time.

This is a simple approach, and permits aggregation over arbitrary dimensions (filtering criteria), but also comes with some downsides such as higher storage cost and aggregation/query performance.

See also: https://github.com/elastic/apm/issues/104.

We will introduce an option to record pre-aggregated transaction duration histograms, using the [histogram field type](https://www.elastic.co/guide/en/elasticsearch/reference/current/histogram.html) introduced in Elasticsearch 7.6.

## Approach

The APM Server will take responsibility for aggregating and producing these histogram metrics. We may later also support agents producing the metrics in order to avoid sending unsampled transactions to the APM Server, but this is initially out of scope. To support that we would have the agents set a flag on events indicating that they have already been aggregated into a histogram.

Enabling transaction histogram metrics will not initially cause unsampled transactions to be dropped, so there will be no storage reduction -- only improved query performance. We will introduce a separate option for dropping unsampled transactions, and in a future major version (e.g. 8.0) we may start dropping unsampled transactions by default.

The histogram field will be used to power most if not all aggregations used in the APM UI, when the search bar is _not_ in use. When the search bar is in use, we will fall back to the existing approach of querying over the individual documents. For the case where the histogram fields are used, the histogram metric documents must also include the context fields used by the default filters in the APM UI.

To support identifying trace groups (i.e. root transaction groups), we will flag the documents relating to root transactions with `transaction.root: true`.

Thus, the APM Server must record histograms for each observed combination of the following fields:

 - agent.name
 - service.name
 - service.version (for deployment annotations)
 - service.environment
 - transaction.name
 - transaction.type
 - transaction.result
 - transaction.root
 - host.hostname
 - container.id
 - kubernetes.pod.name

The histogram field itself will be called `transaction.duration.histogram`. The exact algorithm and parameters to be used is TBD, but following suit with Elasticsearch and using HDRHistogram is likely.

### RUM-specific support

For RUM, the server will also need to perform GeoIP lookup and User-Agent parsing, and include their results to power RUM-specific visualisations. Specifically we would also need these fields:

 - client.geo.country_iso_code
 - user_agent.name

[...]

PR #3573, opened 2020-03-30
all: add metadata directly to model types

axw:
## Motivation/summary

Remove `transform.Context.Metadata`, and add Metadata fields directly to the top-level event model types (`model/transaction.Event.Metadata`, etc.). These fields are now set during decoding, rather than metadata being sent through the transformation to be added to the final `beat.Event`. With this change, `Transformable`s are responsible only for enrichment (sourcemapping) and transforming to `beat.Event`s.

The motivation behind this change is to ensure that decoded model objects are complete, and self-contained. This is important for https://github.com/elastic/apm-server/issues/3485, which requires metadata for aggregation purposes. I think it also clarifies the responsibilities of decoder and transformation.

Note to reviewers: please ignore 7101d72fe45664bdc75364668b751b1ecd13e774, this PR is based on top of https://github.com/elastic/apm-server/pull/3570 which is not yet merged.

## Checklist

## How to test these changes

`make test`

## Related issues

Requires https://github.com/elastic/apm-server/pull/3570
Relates to https://github.com/elastic/apm-server/issues/3567


model/config.go
```diff
@@ -19,6 +19,8 @@ package model
 
 import (
 	"time"
+
+	"github.com/elastic/apm-server/model/metadata"
```
simitt: I find the filename `config.go` confusing, as it defines the `Input` struct holding real event data. Can you rename to `model/input.go`?

axw: Yes, good call. I thought about it and didn't because I intended to move this to another package altogether - but I'll rename it anyway, as that might not happen for a while.

PR #3595, opened 2020-04-01
[RUM v3] add metricsets to RUM V3

### LINKED ISSUE #3475: [RUM v3] Add metricset handler

jalvz:
Closes elastic/apm-server#3475 

- [x] Changelog
- [x] Approval tests


model/metricset/event.go
```diff
@@ -144,8 +155,8 @@ func (md *metricsetDecoder) decodeSamples(input interface{}) []*Sample {
 		}
 
 		samples[i] = &Sample{
-			Name:  name,
-			Value: md.Float64(sampleMap, "value"),
+			Name:  inverseFieldName(name),
+			Value: md.Float64(sampleMap, fieldName("value")),
```
simitt, line 159: The `fieldName("value")` can be derived once before the loop.

model/field/rum_v3_mapping.go
```diff
@@ -130,3 +130,17 @@ func Mapper(shortFieldNames bool) func(string) string {
 		return s
 	}
 }
+
+func InverseMapper(shortFieldNames bool) func(string) string {
+	m := map[string]string{}
+	for k, v := range rumV3Mapping {
+		m[v] = k
+	}
```
simitt, line 138: The inverse mapping is done every time `InverseMapper` is called, which happens for every sample event. Please make the mapping static instead.

model/field/rum_v3_mapping.go
```diff
@@ -131,13 +133,17 @@ func Mapper(shortFieldNames bool) func(string) string {
 	}
 }
 
+var once sync.Once
+var inverseMap = map[string]string{}
```
simitt, line 137: The `rumV3Mapping` and the `inverseMap` are conceptually the same, the one is just an inversion of the other. Why would it be important that the `inverseMap` is lazy initialized on the first call, instead of also defining it as global variable on loading?

For the sake of consistency and simplicity I suggest to instead do a simple initialization, such as 
```
var inverseMap = func() map[string]string {
	m := make(map[string]string)
	for k, v := range rumV3Mapping {
		m[v] = k
	}
	return m
}()
```

For naming consistency I further suggest to use `rumV3InverseMapping`.

jalvz: sure

PR #3608, opened 2020-04-02
[Doc] Improvements for APM fields

renshuki:
## Motivation/summary

Docs says:
> Fields are optional tags that can be added to any output.

https://www.elastic.co/guide/en/apm/server/7.6/configuring-output.html#libbeat-configuration-fields

Which makes me expect an APM server configuration like the following:

```yml
#-------------------------- Elasticsearch output --------------------------
output.elasticsearch:
  hosts: ["localhost:9200"]
  fields: {project: "myproject", instance-id: "574734885120952459"}
```

But actually field cannot be placed under a specific output, otherwise `fields` don't get send to it.
Instead, it's necessary to define `fields` under the root configuration like the following:

```yml
fields: {project: "myproject", instance-id: "574734885120952459"}
#-------------------------- Elasticsearch output --------------------------
output.elasticsearch:
  hosts: ["localhost:9200"]
``` 

Which gives the expected output:

```json
"fields": {
      "instance-id": "574734885120952459",
      "project": "myproject"
    },
```

I'm open to suggestions for better wording as well.
Thanks.

### DIFF
--- a/docs/configuring-output-after.asciidoc
+++ b/docs/configuring-output-after.asciidoc
@@ -12,7 +12,7 @@ See <<config-sourcemapping-elasticsearch>> for more details.
 [float]
 === `fields`
 
-Fields are optional tags that can be added to any output.
+Fields are optional tags that can be added to all outputs.
 Fields can be scalar values, arrays, dictionaries, or any nested combination of these.
 By default, the fields that you specify here will be grouped under a `fields` sub-dictionary in the output document.
 
@@ -21,6 +21,9 @@ Example:
 [source,yaml]
 ------------------------------------------------------------------------------
 fields: {project: "myproject", instance-id: "574734885120952459"}
+#-------------------------- Elasticsearch output --------------------------
+output.elasticsearch:
+  hosts: ["localhost:9200"]
 ------------------------------------------------------------------------------
 
 To store the custom fields as top-level fields, set the `fields_under_root` option to true.
--- a/docs/configuring-output-after.asciidoc
+++ b/docs/configuring-output-after.asciidoc
@@ -17,7 +17,7 @@ They are defined at the top-level in your configuration file, and will apply to 
 Fields can be scalar values, arrays, dictionaries, or any nested combination of these.
 By default, the fields that you specify here will be grouped under a `fields` sub-dictionary in the output document.
 
-Example:
+Example using the Elasticsearch output:
 
 [source,yaml]
 ------------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


graphaelli:
@renshuki would retarget this to `master` please? We can backport the changes to the relevant branches from there

docs/configuring-output-after.asciidoc
```diff
@@ -12,7 +12,7 @@ See <<config-sourcemapping-elasticsearch>> for more details.
 [float]
 === `fields`
 
-Fields are optional tags that can be added to any output.
+Fields are optional tags that can be added to all outputs.
```
bmorelli25, line 15: Since there is no ability to specify the output, I think it makes more sense so say that the are added to all outputs

```suggestion
Fields are optional tags that are added to all outputs.
```

docs/configuring-output-after.asciidoc
```diff
@@ -21,6 +21,9 @@ Example:
 [source,yaml]
 ------------------------------------------------------------------------------
 fields: {project: "myproject", instance-id: "574734885120952459"}
+#-------------------------- Elasticsearch output --------------------------
+output.elasticsearch:
+  hosts: ["localhost:9200"]
```
bmorelli25, line 26: With the suggestion above, I'm not sure a change is needed here. What do you think @renshuki ?

bmorelli25: If you think the additional content is helpful, we can keep it in the example, but should specify that this example is for the ES output. Something like:

```
Example using the Elasticsearch output:

[source,yaml]
fields: {project: "myproject", instance-id: "574734885120952459"}
output.elasticsearch:
  hosts: ["localhost:9200"]
```

docs/configuring-output-after.asciidoc
renshuki: Looks good to me, thanks @bmorelli25

docs/configuring-output-after.asciidoc
renshuki: I just want to make sure users understand they have to put the `fields` setting outside of the `output`. I think that adding:
> Example using the Elasticsearch output:

for clarification is the best option. Thanks.

bmorelli25: Great. That sounds good to me.

renshuki: Appreciate the help, I added the change.

renshuki:
> @renshuki would retarget this to `master` please? We can backport the changes to the relevant branches from there

@graphaelli not sure how to do it. Should I just change the base branch to `master` (via **Edit**) or I need to cherry-pick the commits and create a new PR targetting `master` branch instead? Thanks.

renshuki:
I've cherry-picked the commits, re-applied them onto master and changed the base branch for this PR to master. Hopefully it's all good now.

bmorelli25:
Thank you @renshuki! I'll get this backported

PR #3623, opened 2020-04-07
model: merge stream and per-event metadata

axw:
## Motivation/summary

Merge event-specific service and user metadata
with stream metadata at decoding time. This
avoids allocations, and enables us to have a
single source of truth post-decoding time for
identifying the service and user for an event.
The latter point is important to enable aggregation
of events in apm-server.

Labels continue to be merged at transformation
time for now. We might change that later.

NOTE: there is a subtle functional change here:
when decoding `transaction.context` and
`error.context`, we set `source.ip` based on the
`user.ip` if provided – but only for event-specific
user metadata. I have changed it to use stream
user metadata instead, if event-specific user
metadata is not provided.

## Checklist

## How to test these changes

`make test`

## Related issues

### DIFF
--- a/beater/test_approved_es_documents/TestPublishIntegrationErrors.approved.json
+++ b/beater/test_approved_es_documents/TestPublishIntegrationErrors.approved.json
@@ -334,7 +334,6 @@
                 "scheme": "https"
             },
             "user": {
-                "email": "bar@example.com",
                 "id": "99",
                 "name": "foo"
             },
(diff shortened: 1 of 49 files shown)


axw:

axw:

simitt:
Regarding merging or replacing `user` information, we had some discussions about it in https://github.com/elastic/apm-server/issues/1175. 
I do not think any agent is currently sending user information in the metadata and in the event. Afaik the RUM agent sends `user` data in the `metadata` while all other agents send the data in the `event`. 

What would be the logical reason for merging two user objects though? My assumption when implementing was that the user per event differs from the default user rather than having more information available on the same user. In this case we would rather replace than merge the user.

axw:
@simitt you're right, there's no good logical reason for doing it. I was going for consistency, but on second thoughts that's the wrong thing. I'll revert that change, and add a comment stating the intended behaviour.

Issue #3195, opened 2020-01-16
Histogram metric support


axw:
Currently, the metrics API only supports recording single value metrics: counters and gauges. We would like to extend the API to also support ingesting pre-aggregated histograms, storing as the [histogram field type](https://github.com/elastic/elasticsearch/pull/48580).

- [ ] Define intake API changes (https://github.com/elastic/apm-server/issues/4884)
~- [ ] Investigate how to implement dynamic field mapping for histograms~

For dynamically mapping histograms, we are currently waiting for https://github.com/elastic/elasticsearch/issues/61939.

axw:
For dynamic field mapping, we could use `[match_mapping_type](https://www.elastic.co/guide/en/elasticsearch/reference/master/dynamic-templates.html#match-mapping-type): object`.

This might be a bit too lax, as it would assume _all_ objects we index are histograms, and might prevent adding support for [aggregate_metric](https://github.com/elastic/elasticsearch/pull/49830) later. An extension of Elasticsearch's `match_mapping_type: object` to also match on the types of object properties might be helpful here. i.e. also require "values" and "counts" fields, with the appropriate types.

Alternatively we could enforce a specific field naming, e.g. prefix all histograms with "histogram.". That feels like a cop out to me though, forcing implementation details onto the user.

simitt:
> Alternatively we could enforce a specific field naming, e.g. prefix all histograms with "histogram.". That feels like a cop out to me though, forcing implementation details onto the user.

Based on the fixed format for histogram fields, we might also be able to detect histograms when processing the input in the apm-server and then add a specific prefix, so the convention is not on the user but auto-applied. In case users search for their specific metric name this might still be an issue. Maybe nesting all of those metrics under a histogram key might be an option.

axw:
I wasn't very clear in my comment, but that's what I meant by that bit you quoted. i.e. store everything with "histogram" as the root key. I think this would be the simplest option working with what we've got, but it means that users need to prefix everything with "histogram." in their queries.

Again I wasn't clear, but that's what I meant by "forcing implementation details onto the user". Application developers would record a metric with one name, but then it ends up getting this "histogram." prefix because of implementation details/constraints. Not the end of the world, but I think it would be nice if we could avoid it.

axw:
Opened an ES feature request: https://github.com/elastic/elasticsearch/issues/51341

axw:
[...]

PR #3639, opened 2020-04-09
model: enhancements to support transaction aggregations

axw:
## Motivation/summary

There are a few smallish changes here bundled together, in support of transaction duration aggregations (https://github.com/elastic/apm-server/issues/3485).

Summary:
 - export `metadata.System.Hostname`, so its value can be used in aggregation keys
 - add `metricset.Transaction.{Result,Root}` fields, to use in aggregation keys
 - add `metricset.Sample.{Counts,Values}` fields, for indexing histogram metrics
 - use non-pointer fields for `metricset.Metricset` and related types to simplify usage, and reduce allocations

Note that the added fields are only added to the model types, and not the intake API. That will come later, as part of https://github.com/elastic/apm-server/issues/3195.

## Checklist

## How to test these changes

`make test`

## Related issues

### DIFF
--- a/model/metricset/event.go
+++ b/model/metricset/event.go
@@ -77,7 +77,23 @@ type Sample struct {
 	Name string
 
 	// Value holds the metric value for single-value metrics.
+	//
+	// If Counts and Values are specified, then Value will be ignored.
 	Value float64
+
+	// Values holds the bucket values for histogram metrics.
+	//
+	// These values must be provided in ascending order.
+	Values []float64
+
+	// Counts holds the bucket counts for histogram metrics.
+	//
+	// These numbers must be positive or zero.
+	//
+	// If Counts is specified, then Values is expected to be
+	// specified with the same number of elements, and with the
+	// same order.
+	Counts []int64
 }
 
 // Transaction provides enough information to connect a metricset to the related kind of transactions.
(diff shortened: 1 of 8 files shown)


model/modeldecoder/metricset.go
```diff
@@ -86,74 +84,28 @@ type metricsetDecoder struct {
...
-			Name:  inverseFieldName(name),
-			Value: md.Float64(sampleMap, value),
-		}
-		if md.Err != nil {
-			return nil
-		}
-		i++
+	valueFieldName := fieldName("value")
+	for name, s := range input {
+		sampleObj, _ := s.(map[string]interface{})
```
simitt, line 93: Why are you removing the error check here?

model/metricset/event.go
```diff
@@ -77,7 +77,23 @@ type Sample struct {
...
 	Value float64
+
+	// Values holds the bucket values for histogram metrics.
+	//
+	// These values must be provided in ascending order.
+	Values []float64
+
+	// Counts holds the bucket counts for histogram metrics.
+	//
+	// These numbers must be positive or zero.
```
simitt, line 91: (Where) do you plan to add checks for the restrictions mentioned in the comments?

axw: I wasn't planning on adding any checks beyond what Elasticsearch does, I just added that as a reminder/documentation.

model/modeldecoder/metricset.go
axw: It's guaranteed to be that type, by virtue of having passed JSON Schema validation. Since I moved schema validation into the exported Decode methods, you can no longer get here without passing validation.

model/metricset/event.go
axw: In case it's not clear: if/when we add support for histogram metrics coming from agents, we would validate these constraints in modeldecoder. In the immediate future these fields will only be set by the metrics computed by the server, so we'll be in full control of the data.

Issue #3404, opened 2020-02-27
[RUM v3] Optimize fields


jalvz:
This includes:
- Derive `parent_id`, `trace_id` and `transaction_id` in the `span` from its `transaction`
- Make required booleans optional, defining absence as `false` ( `sync`)

- ~~Make `start` and `duration` int~~.
- ~~Remove unneeded fields: `transaction.marks.navigationTiming`, `http.url`~~

axw:
> Make start and duration int.

They're currently "number", which allows the agent to send an integer as well as a float. What's the reason for making the schema stricter?

jalvz:
> They're currently "number", which allows the agent to send an integer

Good point, I haven't noticed. 
But while looking at how to implement this, I changed my mind about the approach: it's probably simpler for us that the agent embeds the spans inside their transactions.

There is no easy way that we can take fields out of a transaction decoder and pass them as arguments to the spans decoder.

I think we would need an specific `HandlerRawModel` for this, or a flag inside the processor, and it would break the uniformity of the work flow (decoder signatures, etc).

wdyt?

axw:
If the RUM agent will always have spans within transactions, then I think we wouldn't need or want a separate span decoder. The RUM transaction decoder would return a Transformable which internally combines info from the transaction and embedded span objects to produce the transaction and span docs.

jalvz:
All boolean fields are optional, so they can default to false when missing.

I think that with https://github.com/elastic/apm-server/pull/3659 this can be closed.

PR #3648, opened 2020-04-14
[RUM V3] Embed spans inside transactions

jalvz:
## Motivation/summary

This moves RUM V3 spans inside their parent transactions. This allows the RUM agent to not send `parent_id`, `transaction_id`, and `trace_id`. APM Server will take those fields from the transaction.

`parent_idx` is introduced, optional: when missing, the transaction is the parent. This is often the case in RUM. When the parent of an span is another span, RUM will send up the field indicating the index (position in the array) of the span.

## Checklist

## How to test these changes

The RUM agent does not support V3 yet, so the way to tests this change is sending a handcrafted payload and checking in Elasticsearch that all spans are indexed with the correct `parent_id`, `transaction_id`, and `trace_id`.

## Related issues

Implements the first point of #3404


docs/spec/spans/rum_v3_span.json
```diff
@@ -10,22 +9,9 @@
...
-                "trace_id": {
-                    "description": "Hex encoded 128 random bits ID of the correlated trace.",
-                    "type": "string",
-                    "maxLength": 1024
-                },
-                "parent_id": {
-                    "description": "Hex encoded 64 random bits ID of the parent transaction or span.",
-                    "type": "string",
+                "pid": {
+                    "description": "Hex encoded 64 random bits ID of the parent span. Absent when the parent is a transaction.",
```
axw, line 13: Did we consider making this an index into the spans list, like in the v1 protocol? That would be even more compact.

model/transaction/event.go
```diff
@@ -70,6 +72,11 @@ type Event struct {
 	Experimental interface{}
 }
 
+type RUMV3Event struct {
```
axw, line 75: I think this is OK for the short term (expedient), but long term I don't think we should have a distinct "RUMV3Event" model type. This is a result of us treating the intake events as 1:1 with model objects. Instead, we should change processor/stream to stop assuming 1:1, and instead allow the decoders to produce >1 model objects.

axw: Not for now, but another way I've been thinking about going about this is something like:

- move all model types into the `model` package (i.e. collapse `model`, `model/transaction`, `model/span`, etc.)
- introduce a type like:

```go
package model

type Batch struct {
    Transaction	[]Transaction
    Span	[]Span
    Metricset	[]Metricset
    ...
}
```

The `DecodeFoo` methods would decode into a batch, by appending to the appropriate field(s).

docs/spec/spans/rum_v3_span.json
jalvz: No, good idea :+1:

model/transaction/event.go
jalvz: SGTM

model/modeldecoder/span.go
```diff
@@ -58,15 +58,17 @@ func decodeSpan(input Input, schema *jsonschema.Schema) (transform.Transformable
...
+		Metadata:  input.Metadata,
+		Name:      decoder.String(raw, fieldName("name")),
+		Start:     decoder.Float64Ptr(raw, fieldName("start")),
+		Duration:  decoder.Float64(raw, fieldName("duration")),
+		Sync:      decoder.BoolPtr(raw, fieldName("sync")),
+		Timestamp: decoder.TimeEpochMicro(raw, fieldName("timestamp")),
+		Id:        decoder.String(raw, fieldName("id")),
+		ParentId:  decoder.StringPtr(raw, "parent_id"),
+		// ParentIdx comes from RUM V3 payloads only, and used to populate ParentId
+		ParentIdx:     decoder.IntPtr(raw, fieldName("parent_idx")),
```
axw, line 70: Can we just use "pi" here, rather than going through field mapping? It's only ever going to be relevant to RUMv3 isn't it?

jalvz: I considered it, but then I thought it would be convenient to have all the gibberish defined in the same place, so it works as a single-source-of truth (ie. documentation) for the rum agent folks. (the json-schema files are not a great documentation source anymore IMO)

PR #3651, opened 2020-04-15
Introduce transaction metrics aggregation

axw:
## Motivation/summary

Introduce transaction metrics aggregation, for producing transaction duration histogram metrics (https://github.com/elastic/apm-server/issues/3485). The metrics documents are published periodically, and grouped by various transaction attributes (name, type, result, ...) which enable the APM UI's filters.

TODO (probably in follow-ups):

- [ ] Monitoring
- [ ] Telemetry
- [ ] Docs
- [ ] Update apm-server.yml (after finalising the configuration attribute names)
- [ ] Add support for RUM maps (requires geoIP lookup in apm-server)
- [ ] Introduce config for dropping non-sampled transactions

## Checklist

- [ ] I have made corresponding changes to the documentation

## How to test these changes

1. Enable aggregation (`apm-server.aggregation.enabled: true`)
2. Send transactions with varying names and durations
3. Compare percentiles over raw transaction docs to percentiles over pre-aggregated histograms

```
GET /apm-8.0.0-*/_search
{
  "size": 0,
  "query": {
    "range": {
      "@timestamp": {
        "gte": "now-1d"
      }
    }
  }, 
  "aggs": {
    "by_transaction_name": {
      "terms": {
        "field": "transaction.name",
        "size": 1000
      }, 
      "aggs": {
        "transaction_percentiles": {
          "percentiles": {
            "field": "transaction.duration.us",
            "percents": [1, 5, 25, 50, 75, 95, 99]
          }
        },
        "histogram_percentiles": {
          "percentiles": {
            "field": "transaction.duration.histogram",
            "percents": [1, 5, 25, 50, 75, 95, 99]
          }
        },
        "p75_error": {
          "bucket_script": {
              "buckets_path": {
                  "transaction": "transaction_percentiles[75.0]", 
                  "histogram": "histogram_percentiles[75.0]"
              },
              "script": "(params.histogram / params.transaction) - 1"
          }
        },
        "p90_error": {
          "bucket_script": {
              "buckets_path": {
                  "transaction": "transaction_percentiles[90.0]", 
                  "histogram": "histogram_percentiles[90.0]"
              },
              "script": "(params.histogram / params.transaction) - 1"
          }
        },
        "p95_error": {
          "bucket_script": {
              "buckets_path": {
                  "transaction": "transaction_percentiles[95.0]", 
                  "histogram": "histogram_percentiles[95.0]"
              },
              "script": "(params.histogram / params.transaction) - 1"
          }
        },
        "p99_error": {
          "bucket_script": {
              "buckets_path": {
                  "transaction": "transaction_percentiles[99.0]", 
                  "histogram": "histogram_percentiles[99.0]"
              },
              "script": "(params.histogram / params.transaction) - 1"
          }
        }
      }
    }
  }
}
```

## Related issues

Part of https://github.com/elastic/apm-server/issues/3485

Closes https://github.com/elastic/apm-server/issues/3664


axw:
This probably needs to move under x-pack, since it is reliant on the histogram data type, which is available only with Basic+ subscription.

axw:
TODO:
 - [ ] Merge master, move aggregation bits to x-pack
 - [ ] Ensure we use fields.all.yml for syncing kibana saved objects

aggregation/txmetrics/aggregator.go
```diff
@@ -0,0 +1,364 @@
...
+	}
+}
+
+type transactionMetrics struct {
+	// TODO(axw) consider vendoring or forking hdrhistogram,
+	// since it has been archived in GitHub.
+	histogram *hdrhistogram.Histogram
+}
+
+func (m *transactionMetrics) buckets() (counts []int64, values []float64) {
```
simitt, line 343: The method name `buckets()` is a bit confusing here, as you already use `buckets` for the `map` variable holding `transactionMetrics` (https://github.com/elastic/apm-server/pull/3651/files#diff-2f01ba44a03d252c4b885c9693a9c3c5R284).

aggregation/txmetrics/aggregator.go
```diff
@@ -0,0 +1,364 @@
...
+	return &metricset
+}
+
+func (a *Aggregator) updateTransactionMetrics(key transactionAggregationKey, duration time.Duration) bool {
+	a.mu.RLock()
+	defer a.mu.RUnlock()
+
+	m := a.active
+	m.mu.Lock()
+	defer m.mu.Unlock()
```
simitt, line 220: So far on publishing we only use a [read lock](https://github.com/elastic/apm-server/blob/master/publish/pub.go#L135) (except for when the server stops). With this addition a `Lock` is created for every transaction before being pushed to the `pendingRequests` channel, holding the lock while running the histogram related aggregation. Afaics this potentially creates a bottleneck for processing transaction events. 
This is additionally concerning since this logic runs while the APM agent still waits for a server response. 

We might want to consider something like pushing transactions into a temporary channel, do the aggregation when reading from the channel and only then pushing the transaction (+metricset) into the `pendingRequest` channel. Another option would be to do some refactoring allowing to do the metrics aggregation on transactions when reading from the `pendingRequest` channel before running the `transformation` on the event. 
I didn't look into any performance numbers, so might be good enough for now, but wondering if you measured this with hey-apm for concurrent requests?

aggregation/txmetrics/aggregator.go
```diff
@@ -0,0 +1,364 @@
...
+	minDuration time.Duration = 0
+	maxDuration time.Duration = time.Hour
+)
+
+// Aggregator aggregates transaction durations, periodically publishing histogram metrics.
+type Aggregator struct {
+	report              publish.Reporter
+	interval            time.Duration
+	maxBuckets          int
+	hdrHistogramSigfigs int
```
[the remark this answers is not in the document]
axw: Even though it's documented in `AggregatorConfig`? I might just add that type as a field, rather than duplicating each field. i.e.

```go
type Aggregator struct {
    config AggregatorConfig

    mu               sync.RWMutex
    active, inactive *metrics
}
```

WDYT?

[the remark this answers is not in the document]
simitt: sounds good

aggregation/txmetrics/aggregator.go
axw: Noted, will rename this one. Perhaps histogramFields()? Or just distribution()?

beater/config/aggregation.go
```diff
@@ -0,0 +1,71 @@
...
+		return nil
+	}
+	if c.Interval <= 0 {
+		c.Interval = defaultAggregationInterval
+	}
+	if c.MaxTransactionGroups <= 0 {
+		c.MaxTransactionGroups = defaultAggregationMaxTransactionGroups
+	}
+	if c.HDRHistogramSignificantFigures <= 0 {
+		c.HDRHistogramSignificantFigures = defaultAggregationHDRHistogramSignificantFigures
```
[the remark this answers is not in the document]
axw: Oops, I just didn't know about min/max validators :)
I'll update to use them.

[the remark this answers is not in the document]
axw: Will need to wait for a fix to land in go-ucfg: https://github.com/elastic/go-ucfg/pull/163

aggregation/txmetrics/aggregator.go
simitt: both SGTM

aggregation/txmetrics/aggregator.go
axw: I haven't yet tested it with hey-apm, but will do so after I've rebased this on the other in-flight changes. My expectation is that it's good enough, but we'll see. ICYMI, there's a parallel benchmark in the txmetrics package:

```
BenchmarkAggregateTransaction            6073344               190 ns/op               0 B/op          0 allocs/op
BenchmarkAggregateTransaction-2          5032620               223 ns/op               0 B/op          0 allocs/op
BenchmarkAggregateTransaction-4          5218090               233 ns/op               0 B/op          0 allocs/op
BenchmarkAggregateTransaction-8          4581522               276 ns/op               0 B/op          0 allocs/op
BenchmarkAggregateTransaction-16         4438081               265 ns/op               0 B/op          0 allocs/op
```

I intend to revise this code later to minimise contention, following the approach taken for breakdown metrics in the Go agent: https://github.com/elastic/apm-agent-go/blob/839d3cc2699fdcc73b5524916519845724ec9d2d/breakdown.go#L225.

The idea would be to only take an exclusive lock for the whole map when recording a new group. We would still need to take an exclusive lock on the histogram/transaction group metrics though, unless we fork codahale/hdrhistogram and add a method that uses atomic increments.

axw:
@simitt 

> This also adds a new field to the transaction.duration, that probably should be part of a standardized metrics ECS concept. I assume we will need to put it under transaction first though as we do not have ECS metrics draft or work ongoing yet afaik.

Right. There's this, and also `transaction.root: <boolean>`. Any objection to going ahead with the current naming? Or just add to the TODO pile?

simitt:
I'd say it is fine to move on with the current field names; just pointed ECS out to kind of upvote for an ECS metrics concept; don't think it is realistic to finish the ECS metrics before we want to use them here.

aggregation/txmetrics/aggregator.go
axw: I ended up renaming the other one to "groups" (as in aggregation groups), and this one to "histogramBuckets".

axw:

aggregation/txmetrics/aggregator.go
axw: I prototyped the changes necessary to remove the exclusive lock, and forked hdrhistogram to add atomic update (turns out to be pretty straightforward).

```
BenchmarkAggregateTransaction            6636734               172 ns/op
BenchmarkAggregateTransaction-2          7426639               159 ns/op
BenchmarkAggregateTransaction-4          9878884               115 ns/op
BenchmarkAggregateTransaction-8         14216102                86.3 ns/op
BenchmarkAggregateTransaction-16        13393129                88.9 ns/op
```

There's a clear pattern there: previously adding goroutines decreased throughput (due to lock contention), now it improves.

I also ran both a hey-apm test, and a separate simple load test with multiple clients, and I saw no noticeable change with/without aggregation enabled. What do you think about going ahead with the current implementation, and then iterate on the implementation?

simitt: SGTM - thanks for the investigation

beater/config/aggregation.go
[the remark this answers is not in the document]
axw: Rather than wait for it, I think we should land with the incorrect error message and follow up with a beats update & sync.

x-pack/apm-server/aggregation/txmetrics/aggregator.go
```diff
@@ -0,0 +1,339 @@
...
+	hostname           string
+	kubernetesPodName  string
+	serviceEnvironment string
+	serviceName        string
+	serviceVersion     string
+	transactionName    string
+	transactionResult  string
+	transactionType    string
+	userAgent          string
+}
```
graphaelli, line 283: Premature to support it now but is there a path to add more+custom dimensions to the aggregation key?

x-pack/apm-server/aggregation/txmetrics/aggregator.go
```diff
@@ -0,0 +1,339 @@
...
+			Counts: counts,
+			Values: values,
+		}},
+	}
+	return out
+}
+
+type metrics struct {
+	// TODO(axw) investigate optimising the map for concurrent
+	// updates, along the same lines as what we do in th Go agent
```
graphaelli, line 256:
```suggestion
	// updates, along the same lines as what we do in the Go agent
```

tests/system/transaction_histogram_metrics.approved.json
```diff
@@ -0,0 +1,177 @@
...
+        },
+        "user_agent": {
+            "original": "Mozilla Chrome Edge",
+            "name": "Other",
+            "device": {
+                "name": "Other"
+            }
+        }
+    }
+]
```
graphaelli, line 177: sort keys if making additional changes to this pr

log/selectors.go
```diff
@@ -19,21 +19,22 @@ package logs
...
+	Otel               = "otel"
+	Pipelines          = "pipelines"
+	Request            = "request"
+	Response           = "response"
+	Server             = "server"
+	Sourcemap          = "sourcemap"
+	Stacktrace         = "stacktrace"
+	Tracing            = "tracing"
+	Transform          = "transform"
+	TransactionMetrics = "txmetrics"
```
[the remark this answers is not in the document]
axw: > Now that you are using the tags, this explicit setup function is redundand right?

Ah indeed, I didn't understand how the default stuff worked before. Removed the setup method.

> nit: should be moved above Transform

Fixed, thanks.

x-pack/apm-server/aggregation/txmetrics/aggregator.go
axw: Yes, and this will be made easier with my followup to use a custom hash map structure. We would just use a custom equality method, rather than using the compiler-generated struct equality.

axw: BTW @roncohen and I briefly discussed making this a libbeat processor, so it could be used by other beats.

The rough idea would be:
 - make the grouping field names, and metric field, configurable
 - output libbeat events with the grouping fields + aggregated metric field

I don't intend to do that until there's a need for it though.

axw:
jenkins run the tests please

axw:
The system test failures reproduce on master, so I'm going to merge this. I've opened https://github.com/elastic/apm-server/issues/3688 to investigate the failures.

simitt:
As discussed offline, percentile error rate comparing `transaction.duration.us` and `transaction.duration.histogram` looks good.

When shutting down the APM Server, the current `metrics` document is lost. I'd expect the server to send the last metrics and wait until this document is processed before finishing to shutdown. Not considering this as a blocker though.

PR #3663, opened 2020-04-18
Collapse model packages

axw:
## Motivation/summary

Move all model types into the `model` package. This removes stuttering like `metadata.Metadata`, and renames `span.Event` to `model.Span` to clarify the split between input events and output model objects.

This will make it possible to introduce a `model.Batch` type which would hold a collection of top-level model objects ("events"), for transformation, etc. This would enable us to pass around concrete structs, avoiding premature conversion to the `Transformable` interface, simplifying the implementation of processing specific objects, e.g. transaction metrics aggregation.

I've moved `model/field` to `model/modeldecoder/field`, since it's only relevant to decoding the Elastic APM agent events.

Some of the type names could do with renaming, or combining, due to the shared namespace:
 - `model.Http` (used by transactions and errors) and `model.HTTP` (used by spans)
 - `model.Sample` (was `metricset.Sample`) should probably have "Metric" in its name, e.g `model.MetricSample".

(I haven't done that yet, as I'd like to keep this PR to mostly just moving things around.)

## Checklist

## How to test these changes

`make test`

## Related issues

Closes https://github.com/elastic/apm-server/issues/3567


model/transaction.go
```diff
@@ -15,38 +15,33 @@
...
-	emptyString        = ""
+	transactionProcessorName = "transaction"
+	transactionDocType       = "transaction"
 )
 
 var (
-	Metrics         = monitoring.Default.NewRegistry("apm-server.processor.transaction")
-	transformations = monitoring.NewInt(Metrics, "transformations")
-	processorEntry  = common.MapStr{"name": processorName, "event": transactionDocType}
+	TransactionMetrics         = monitoring.Default.NewRegistry("apm-server.processor.transaction")
```
graphaelli, line 38: I recognize this is just maintaining the current visibility but can these (`ErrorMetrics`, `SpanMetrics`, etc) not be exported now?

axw: Done for all but SourcemapMetrics, which is used by processor/asset. I've left a TODO for that.

PR #3676, opened 2020-04-21
[7.x] beater: introduce NewCreator and RunServer(Func) (#3668)

axw:
 - beater: introduce NewCreator and RunServer(Func) (#3668)


axw:
jenkins run the tests please

PR #3677, opened 2020-04-21
Add appliedByAgent flag to agentcfg query.

simitt:
## Motivation/summary
When using central config in combination with Jaeger for sampling, the config never shows as having been applied in the Kibana UI. This PR adds a query option when sending a request to Kibana, allowing the APM Server to indicate that agent configuration is applied by the agent.

## Checklist

~- [ ] I have signed the [Contributor License Agreement](https://www.elastic.co/contributor-agreement/).~
- [x] My code follows the style guidelines of this project (run `make check-full` for static code checks and linting)
- [x] I have rebased my changes on top of the latest master branch

~- [ ] I have made corresponding changes to the documentation~
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing [**unit** tests](https://github.com/elastic/apm-server/blob/master/TESTING.md) pass locally with my changes

- [x] I have updated [CHANGELOG.asciidoc](https://github.com/elastic/apm-server/blob/master/CHANGELOG.asciidoc)

## How to test these changes

* Test together with https://github.com/elastic/kibana/pull/63967; right now that means check out the PR branch and [start Kibana local](https://github.com/elastic/kibana/blob/master/CONTRIBUTING.md#setting-up-your-development-environment)
* Create agent configuration in Kibana
* Run a jaeger agent pointing to the APM Server as collector e.g.`./jaeger-agent --reporter.grpc.host-port=localhost:14250`
* Start apm-server with kibana and jaeger grpc enabled: `./apm-server -e -E apm-server.kibana.enabled=true -E apm-server.jaeger.grpc.enabled=true`
* send a sampling request to the jaeger agent, e.g. `curl localhost:5778/sampling?service=myservice`
* ensure agent config in Kibana is shown as applied after making requests from the hotrod app

## Related issues
fixes #3579

PR #3497, opened 2020-03-18
[7.x] [Jaeger] Add gRPC sampling endpoint (#3490)

simitt:
Backports the following commits to 7.x:
 - [Jaeger] Add gRPC sampling endpoint (#3490)

PR #3686, opened 2020-04-23
[7.x] Add appliedByAgent flag to agentcfg query. (#3677)

simitt:
 - Add appliedByAgent flag to agentcfg query. (#3677)


jalvz:
Note for tester: when testing this, consider https://github.com/elastic/apm-server/pull/3689

axw:
Verified.

I repeated the test run in https://github.com/elastic/apm-server/pull/3497#issuecomment-606472459, and checked that the mark next to the service in the "Agent remote configuration" UI went green after being applied.

Issue #3688, opened 2020-04-24
tests/system: .apm-agent-configuration/_delete_by_query resulting in 409 conflicts


axw:
https://apm-ci.elastic.co/job/apm-server/job/apm-server-mbp/job/PR-3651/16/testReport/

> elasticsearch: WARNING: POST http://elasticsearch:9200/.apm-agent-configuration/_delete_by_query?ignore_unavailable=true&wait_for_completion=true [status:409 request:0.020s]
elasticsearch: DEBUG: > {"query":{"match_all":{}}}
elasticsearch: DEBUG: < {"took":17,"timed_out":false,"total":3,"deleted":2,"batches":1,"version_conflicts":1,"noops":0,"retries":{"bulk":0,"search":0},"throttled_millis":0,"requests_per_second":-1.0,"throttled_until_millis":0,"failures":[{"index":".apm-agent-configuration","id":"eF4GqnEByINrVAt-Ypdx","cause":{"type":"version_conflict_engine_exception","reason":"[eF4GqnEByINrVAt-Ypdx]: version conflict, required seqNo [19], primary term [1]. current document has seqNo [20] and primary term [1]","index_uuid":"m6i98aghSzWcZS9j2oaq5w","shard":"0","index":".apm-agent-configuration"},"status":409}]}

Not sure what's up here. Could be related to recent Jaeger remote config changes to immediately mark config as applied?

PR #3690, opened 2020-04-24
tests/system: use Kibana REST API to cleanup ACM

axw:
## Motivation/summary

Use the Kibana APM app's [Agent Configuration API](https://www.elastic.co/guide/en/kibana/master/agent-config-api.html) to delete agent config instead of deleting documents in the Elasticsearch index directly.

## Checklist

## How to test these changes

`make system-tests`

## Related issues

Fixes https://github.com/elastic/apm-server/issues/3688

### DIFF
--- a/tests/system/kibana.py
+++ b/tests/system/kibana.py
@@ -1,0 +1,82 @@
+#import json
+from urllib.parse import urljoin
+
+import requests
+
+class Kibana(object):
+    def __init__(self, url):
+        self.url = url
+
+    def create_agent_config(self, service, settings, agent=None, env=None):
+        resp = self._upsert_agent_config(service, settings, agent=agent, env=env)
+        result = resp.json()["result"]
+        assert result == "created", result
+        return resp
+
+    def create_or_update_agent_config(self, service, settings, agent=None, env=None):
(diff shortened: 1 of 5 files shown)


tests/system/kibana.py
```diff
@@ -0,0 +1,82 @@
...
+        assert result == "created", result
+        return resp
+
+    def create_or_update_agent_config(self, service, settings, agent=None, env=None):
+        resp = self._upsert_agent_config(service, settings, agent=agent, env=env, overwrite=True)
+        result = resp.json()["result"]
+        assert result in ("created", "updated"), result
+        return resp
+
+    def _upsert_agent_config(self, service, settings, agent=None, env=None, overwrite=False):
```
simitt, line 22: Not sure if this `_` is an oversight or on purpose, but in case it is used to signal internal usage only, please do the same for `list_agent_config` and `delete_agent_config` and group them together. 
IMO we should just remove the underscore here though.

axw: `_` is copied across from test_integration_acm, and is intended to signal internal usage; it exists only to avoid repeating code in the above callers

I think `list_agent_config` and `delete_agent_config` could be useful outside of this class, as they're part of the ACM API. They're not used _yet_, but I don't see harm in exposing them.

PR #3691, opened 2020-04-24
aggregation/txmetrics: optimise for concurrent writers

axw:
## Motivation/summary

Optimise txmetrics.Aggregator for concurrent writers,
in order to avoid impacting stream processing throughput.

We achieve this by minimising use of exclusive locks during
updates, enabled by two major changes:

 - Support for atomic updates in elastic/go-hdrhistogram (https://github.com/elastic/go-hdrhistogram/commit/aac925fdf3816d2056884dc4e3679d30483555a5)
 - A custom hash table using "separate chaining", with a fixed-size backing array

When updating, we first take a shared lock with the existing
hash table, and attempt to find an entry with the given key;
if we find it then we atomically update the histogram. If we
do not find a matching key, then we take an exlusive lock,
check again in any newly added entries (there could have been
an intervening, concurrent update), and update if found,
otherwise add a new entry if the maximum has not yet been reached.

## Checklist

## How to test these changes

`make test`

## Related issues


axw:
```
AggregateTransaction       192ns ± 9%     222ns ± 2%  +15.64%  (p=0.008 n=5+5)
AggregateTransaction-2     216ns ± 8%     192ns ± 3%  -11.30%  (p=0.008 n=5+5)
AggregateTransaction-4     293ns ± 9%     188ns ± 1%  -35.88%  (p=0.016 n=5+4)
AggregateTransaction-8     320ns ± 6%     147ns ± 3%  -54.22%  (p=0.008 n=5+5)
```

We're in the nanoseconds here, but it's important to note that with the old code, increased concurrency led to increased time per operation. This is due to lock contention, which could lead to more expensive context switching.

With the new code, increased concurrency leads to decreased time per operation, which implies less contention/greater parallelism. There's an increase in time for the single-CPU case, which may not be unusual (e.g. consider an apm-server local to each service) but again, this is in the nanoseconds. Also, we can bring this down by switching to maphash when we upgrade to Go 1.14+. At least, my tests showed it was faster in this case.

PR #3695, opened 2020-04-27
[RUM V3] Derive transaction information for metricset

jalvz:
## Motivation/summary

For RUM v3 we have embedded spans inside transactions, so to not make the agent send `transactionId`, `traceId`, and `parentId`. Insteads, the server derives those fields.

Metricsets also contain transaction fields that can be derived in the server. However, embedding metricsets inside transactions can be risky because it could reach the event size limit more easily.

This PR removes transaction data from the RUM V3 metricset spec, and takes it from the transaction, without embedding metricsets. 

This is achieved by propagating a context trough all the decoding functions, which now makes it possible to backtrack the span approach and do it in the same way for spans.

Depending how big RUM events can be, there are several options:
- Embed spans in transactions, but not metricsets.
- Embed both.
- Embed neither.

The downside of not embedding events is that the code is more complicated, and order-dependent.

There would be also several alternatives to handle the event size limit, which currently is configurable and defaults to 300kb. We can e.g. introduce a new value for RUM V3, either hardcoded, configurable, or as a hardcoded multiplier of the existing setting.

## Checklist

- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works

## How to test these changes

`go test ./...`

## Related issues

RUM V3 meta: https://github.com/elastic/apm-server/issues/2854


jalvz:
> Perhaps we could allow transaction breakdown metrics to be embedded in transactions

Yeah, that makes sense if the events don't grow too large, or if we use a different value for RUM V3 max event size. I guess the difference with spans is that transaction name can be arbitrarily verbose. But maybe we could embed them in the spans instead?

> this use of context feels a bit subtle to me. If we must have it at the top level

Isn't that the idiomatic way to handle Context (top level)? What would be the non subtle way?

axw:
>>Perhaps we could allow transaction breakdown metrics to be embedded in transactions
>>
> Yeah, that makes sense if the events don't grow too large, or if we use a different value for RUM V3 max event size. I guess the difference with spans is that transaction name can be arbitrarily verbose.

True, though we should be discouraging that anyway as it reduces the utility of aggregations.

Embedding the breakdown metrics won't be worse than embedding spans, so I think we should either embed neither or embed both. I'm inclined to embed both, to keep it simple.

> But maybe we could embed them in the spans instead?

I don't think so. There may be multiple spans with the same type/subtype in a single transaction, but we should only send one metricset for each distinct type/subtype pair.

>>this use of context feels a bit subtle to me. If we must have it at the top level
>>
> Isn't that the idiomatic way to handle Context (top level)? What would be the non subtle way?

Not sure we're talking about the same thing now. When I said "If we must have it at the top level", the "it" I was referring to was metricsets. i.e. if we need to report metricsets at the top level of a stream, vs. embedding in a transaction, then using context as a means of accumulating state for the stream seems a bit subtle to me.

jalvz:
> When I said "If we must have it at the top level", the "it" I was referring to was metricsets

ah :)

jalvz:
I'm closing this in favour of the alternative approach of embbeding metricsets in transactions

PR #3701, opened 2020-04-28
[7.x] aggregation/txmetrics: optimise for concurrent writers (#3691)

axw:
 - aggregation/txmetrics: optimise for concurrent writers (#3691)


axw:
I'm removing this from test-plan, as it's just an optimisation of https://github.com/elastic/apm-server/pull/3699

PR #3713, opened 2020-04-29
Fix typo in spec ids

jalvz:
Bad ids are confusing the PHP schema validator: https://github.com/swaggest/php-json-schema

This fixes it.


jalvz:
cc @SergeyKleyman

PR #3733, opened 2020-05-04
[RUMV3] allow metricsets inside transactions

jalvz:
## Motivation/summary
RUM V3 spec will allow metricsets inside transactions, in addition to top level events. 

Either case, the RUM agent will not report transaction fields. When inside the transaction (breakdown metrics), the apm-server will derive and fill the fields from the transaction.

## Checklist

## How to test these changes

`make test`


testdata/intake-v3/rum_events.ndjson
```diff
@@ -1,6 +1,3 @@
...
+{"me": {"y": {"t": "Processing"},"sa": {"ysc": {"v": 1},"yss": {"v": 124}}}}
```
simitt, line 3: Is that a valid case that a metricset belongs to a span but not to a transaction?

jalvz: Yeah, probably doesn't make much sense, but I wanted to keep a top level event to test they are ingested. RUM doesn't send system metrics yet (they will soon) so there aren't  "real" examples we can use.

docs/data/intake-api/generated/rum_v3_events.ndjson
```diff
@@ -1,6 +1,3 @@
...
+{"me": {"y": {"t": "Processing"},"sa": {"ysc": {"v": 1},"yss": {"v": 124}}}}
```
vigneshshanmugam, line 3: is this a valid metricset? We can send metricset inside transaction or even in the same events as of transaction?

vigneshshanmugam: Asking because how do you differentiate them from custom metricsets like CPU, Memory etc.

jalvz: https://github.com/elastic/apm-server/pull/3733/#discussion_r419396153

:see_no_evil:

jalvz: I don't think it is needed to differentiate, we just accept both, but if it comes inside the transactions then we fill in the transaction info

jalvz: maybe you (and @simitt ) were thinking about having different, separated specs for system and breakdowns, but I think that would just complicate things (more types internally, etc).

vigneshshanmugam: Cool, thanks for the details Juan.

PR #3735, opened 2020-05-04
ci: aggregate downstream test results

v1v:
## Motivation/summary

The GitHub PR comment has been enabled recently, so there is only one place now to get a highlevel view what the CI build looks like. In order to enhance the build status, let's see if we can aggregate the results from the APM-ITs within the main pipeline.

## What

Run the `apm-integration-testing` test selector pipeline in the parallel stage rather than asynchronous. What does it mean? Pipeline needs to wait for the outcome from the `apm-integration-testing` to be able to aggregate the results in the PR comment

## Related issues

Depends on https://github.com/elastic/apm-integration-testing/pull/835

## Tests

From 3150 tests (See [latest build](https://apm-ci.elastic.co/job/apm-server/job/apm-server-mbp/job/master/lastBuild/testReport/))

image

To 3356 tests (See [this build](https://apm-ci.elastic.co/job/apm-server/job/apm-server-mbp/job/PR-3735/3/testReport/))

image

Test agents reports:

image

Test server reports:

image

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -10,7 +10,6 @@ pipeline {
     JOB_GCS_BUCKET = credentials('gcs-bucket')
     JOB_GCS_CREDENTIALS = 'apm-ci-gcs-plugin'
     CODECOV_SECRET = 'secret/apm-team/ci/apm-server-codecov'
-    GITHUB_CHECK_ITS_NAME = 'APM Integration Tests'
     ITS_PIPELINE = 'apm-integration-tests-selector-mbp/master'
     DIAGNOSTIC_INTERVAL = "${params.DIAGNOSTIC_INTERVAL}"
     ES_LOG_LEVEL = "${params.ES_LOG_LEVEL}"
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -10,7 +10,6 @@ pipeline {
     JOB_GCS_BUCKET = credentials('gcs-bucket')
     JOB_GCS_CREDENTIALS = 'apm-ci-gcs-plugin'
     CODECOV_SECRET = 'secret/apm-team/ci/apm-server-codecov'
-    GITHUB_CHECK_ITS_NAME = 'APM Integration Tests'
     ITS_PIPELINE = 'apm-integration-tests-selector-mbp/master'
```
v1v, line 13:
```suggestion
    ITS_PIPELINE = 'apm-integration-tests-selector-mbp/PR-835'
```

Jenkinsfile
```diff
@@ -10,8 +10,7 @@ pipeline {
     JOB_GCS_BUCKET = credentials('gcs-bucket')
     JOB_GCS_CREDENTIALS = 'apm-ci-gcs-plugin'
     CODECOV_SECRET = 'secret/apm-team/ci/apm-server-codecov'
-    GITHUB_CHECK_ITS_NAME = 'APM Integration Tests'
-    ITS_PIPELINE = 'apm-integration-tests-selector-mbp/master'
+    ITS_PIPELINE = 'apm-integration-tests-selector-mbp/PR-835'
```
v1v, line 13: TODO: rollback
```suggestion
    ITS_PIPELINE = 'apm-integration-tests-selector-mbp/master'
```

v1v:
@jalvz, good points, I actually didn't mention any of them, thanks for this. Just to highlight the question regarding reporting easily with faster builds might be it's not easy to implement.

I've moved the `apm-its` stage to run in parallel with the other ones to avoid the overhead slowing down the overall build. Downside, more infra requirements for breaking builds.

If we move those stages to be more sequential, then it will give us the confidence that no waste of resources.

If the matter is about easy to debug/troubleshoot builds within the GitHub PR comment then, this particular approach will help to spot all the test failures for all the stages even when those stages triggered another build (in this case the `apm-integration-testing`). This is not only for us who know how the CI works but also for the first time contributors or people who are not familiar with the CI.

What do you think?

jalvz:
sounds good, go for it! :+1:

v1v:
Jenkins run the tests please

PR #3721, opened 2020-04-30
docs/spec: require duration >= 0

axw:
## Motivation/summary

Until now the APM Server would accept negative transaction and span durations, which does not make sense. We should reject negative durations, so we don't record junk data.

Recording negative durations would impact our ability to use HDRHistogram in percentile aggregations: https://github.com/elastic/kibana/pull/64758.

## Checklist

- [x] I have signed the [Contributor License Agreement](https://www.elastic.co/contributor-agreement/).
- [x] My code follows the style guidelines of this project (run `make check-full` for static code checks and linting)
- [x] I have rebased my changes on top of the latest master branch
~- [ ] I have made corresponding changes to the documentation~
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing [**unit** tests](https://github.com/elastic/apm-server/blob/master/TESTING.md) pass locally with my changes
- [x] I have updated [CHANGELOG.asciidoc](https://github.com/elastic/apm-server/blob/master/CHANGELOG.asciidoc)

## How to test these changes

I hacked the Go agent to allow it to send negative durations, set the duration to -1, and observed a schema validation error. Normally the Go agent does not allow this; it uses negative values as a sentinel for setting duration based on `now-start`.

## Related issues

Closes https://github.com/elastic/kibana/pull/64758

PR #3741, opened 2020-05-05
[7.x] docs/spec: require duration >= 0 (#3721)

axw:
 - docs/spec: require duration >= 0 (#3721)


simitt:
tested for requests against
`intake/v3/rum/events`
`intake/v2/rum/events`
`intake/v2/events`

PR #3702, opened 2020-04-28
Configuration for dropping non-sampled transactions

axw:
## Motivation/summary

Introduce configuration to drop non-sampled transactions. Typically this should only be used in conjunction with transaction aggregation. Off by default for backwards compatibility.

TODO:

~- [ ] Docs~ (will document later, when we have UI support)
- [x] System test
- [x] Log warning if aggregation is not enabled
- [x] Monitoring -- increment counter if non-sampled transaction docs are dropped
- [x] Changelog

## Checklist

- [x] I have signed the [Contributor License Agreement](https://www.elastic.co/contributor-agreement/).
- [x] My code follows the style guidelines of this project (run `make check-full` for static code checks and linting)
- [x] I have rebased my changes on top of the latest master branch
~- [ ] I have made corresponding changes to the documentation~
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing [**unit** tests](https://github.com/elastic/apm-server/blob/master/TESTING.md) pass locally with my changes
- [x] I have updated [CHANGELOG.asciidoc](https://github.com/elastic/apm-server/blob/master/CHANGELOG.asciidoc)

## How to test these changes

1. Configure an agent with <100% sampling rate
2. Configure server with `apm-server.sampling.keep_non_sampled=false`
3. Create some transactions
4. Observe that only sampled transactions are recorded

## Related issues

Relates to https://github.com/elastic/apm-server/issues/3485

PR #3742, opened 2020-05-05
[7.x] Configuration for dropping non-sampled transactions (#3702)

axw:
 - Configuration for dropping non-sampled transactions (#3702)


simitt:
Tested that only sampled transactions are recorded and a warning is logged, in case aggregation is not configured but unsampled transactions are dropped.

PR #3753, opened 2020-05-06
Clarify error message when unauthorized Kibana connection

### LINKED ISSUE #3568: Missleading error message
To reproduce:

- Start elasticsearch with username and password eg. via `apm-integration-testing`
- Start apm-server with the `-E apm-server.kibana.enabled=true` option, but without passing `apm-server.kibana.username` or `apm-server.kibana.password`
- Query agent configuration for a service: `curl "http://localhost:8200/config/v1/agents?service.name=my-service"` 

The result of that query is:
```
{
  "error": "service unavailable"
}
```

But the real cause is in the logs
```
2020-03-27T15:37:07.832+0100	ERROR	[request]	middleware/log_middleware.go:95	service unavailable	{"request_id": "878592a1388b31af", "method": "GET", "URL": "/config/v1/agents?service.name=my-service", "content_length": 0, "remote_address": "127.0.0.1", "user-agent": "curl/7.58.0", "trace.id": "878592a1388b31af04a06b17edc8e518", "transaction.id": "878592a1388b31af", "response_code": 503, "error": "{\"statusCode\":401,\"error\":\"Unauthorized\",\"message\":\"Unauthorized\"}"}
```

The underlying cause should be better surfaced to users.

jalvz:
Fixes elastic/apm-server#3568

### DIFF
--- a/beater/api/config/agent/handler.go
+++ b/beater/api/config/agent/handler.go
@@ -188,8 +188,9 @@ func extractInternalError(c *request.Context, err error, withAuth bool) {
 		keyword = agentcfg.ErrMsgReadKibanaResponse
 
 	case strings.Contains(msg, agentcfg.ErrUnauthorized):
-		fullMsg := "APM Server is not authorized to contact Kibana. " +
-			"Please configure apm-server.kibana.username and apm-server.kibana.password"
+		fullMsg := "APM Server is not authorized to query Kibana. " +
+			"Please configure apm-server.kibana.username and apm-server.kibana.password, " +
+			"and ensure the user has the necessary privileges."
 		body = authErrMsg(fullMsg, agentcfg.ErrUnauthorized, withAuth)
 		keyword = agentcfg.ErrUnauthorized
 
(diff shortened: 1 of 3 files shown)


beater/api/config/agent/handler.go
```diff
@@ -187,6 +187,12 @@ func extractInternalError(c *request.Context, err error, withAuth bool) {
 		body = authErrMsg(msg, agentcfg.ErrMsgReadKibanaResponse, withAuth)
 		keyword = agentcfg.ErrMsgReadKibanaResponse
 
+	case strings.Contains(msg, agentcfg.ErrUnauthorized):
+		fullMsg := "APM Server is not authorized to contact Kibana. " +
+			"Please configure apm-server.kibana.username and apm-server.kibana.password"
```
axw, line 192: Would this error occur if the server had username/password configured, but the username didn't have the right privileges? If so, perhaps reword a bit, like:

"APM Server is not authorized to query agent configuration via Kibana. Please configure apm-server.kibana.username and apm-server.kibana.password, and ensure the user has the necessary privileges."

PR #3764, opened 2020-05-06
[7.x] Clarify error message when unauthorized Kibana connection (#3753)

jalvz:
 - Clarify error message when unauthorized Kibana connection (#3753)

### DIFF
--- a/agentcfg/fetch.go
+++ b/agentcfg/fetch.go
@@ -37,9 +37,9 @@ import (
 
 // Error Messages used to signal fetching errors
 const (
-	ErrMsgSendToKibanaFailed = "sending request to kibana failed"
-	ErrMsgReadKibanaResponse = "unable to read Kibana response body"
-
+	ErrMsgSendToKibanaFailed   = "sending request to kibana failed"
+	ErrMsgReadKibanaResponse   = "unable to read Kibana response body"
+	ErrUnauthorized            = "Unauthorized"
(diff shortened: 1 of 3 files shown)


axw:
Will need to wait for a new BC to test this one.

axw:
If I don't have any kind of auth enabled in apm-server, then with invalid/missing kibana auth I get this response when querying agent config:

```
$ curl http://localhost:8200/config/v1/agents?service.name=foo

{                                       
  "error": "Unauthorized"         
}
```

I'm not sure that behaviour is quite right, but it's existing behaviour. What I'm not sure about is whether we should be suppressing the error message when the server doesn't have any auth enabled. It certainly makes sense to suppress it when it's enabled but not supplied.

If I enable secret token auth in apm-server, and send a secret token in the query, I see the full message:

```
$ curl -H "Authorization: Bearer hunter2" http://localhost:8200/config/v1/agents?service.name=foo

{
  "error": "APM Server is not authorized to query Kibana. Please configure apm-server.kibana.username and apm-server.kibana.password, and ensure the user has the necessary privileges."
}
```

axw:
@jalvz is that behaviour expected?

jalvz:
Yes, I followed the same logic from other cases:
https://github.com/elastic/apm-server/blob/6270692644f0cb75e75ee0617fc96b3d7ce2146a/beater/api/config/agent/handler.go#L177 (`authErrMsg`)

Issue #3318, opened 2020-02-12
Move "system" tests to Go


jalvz:
Explore the Docker API, https://github.com/testcontainers/testcontainers-go, or a similar library to write integration/system tests in Go, and gradually move existing Python tests.

Now is a good time for this, since Python 2 just reached its End Of Life.

bsideup:
Just out of curiosity:

Why not testcontainers-java, the most feature complete solution out there? :)

Although I should mention that testcontainers-go is well developed and they try to catch up feature wise with the Java version :)

axw:
@bsideup the primary reason is that this project is written in Go. We're trying to simplify things. The main change we're discussing here is removing our dependency on Python, and consolidating on Go.

axw:
Not completely migrated, but we'll tackle them one by one over time. New system tests will be written in Go.

PR #3790, opened 2020-05-13
Introduce Go system tests

axw:
## Motivation/summary

Introduce new Go packages for system tests, to reduce the number of tools we use in development. The goal is to simplify writing, running, and maintaining system tests.

 - `systemtest/apmservertest`, which provides an API similar to [net/http/httptest](https://golang.org/pkg/net/http/httptest) for running apm-server
 - `systemtest`, which holds system tests. I have translated a subset of the Python-based system tests over, but the Python tests haven't yet been removed.

These packages live within a new Go Module called `github.com/elastic/apm-server/systemtest`, which provides a clear delineation between the server code and system tests. This also means that running `go test ./...` at the top level of the repo will not run the system tests. To do that, explicitly change into the `systemtest` directory and run `go test ./...` from there.

One significant difference with the Python implementation is that we now run the proper `apm-server` binary, rather than `apm-server.test` (compiled test binary). This means we can't get code coverage from the system tests, but it means that we could run the system tests with the apm-server binary that we distribute. For now we always build the `apm-server` binary, but we might change this later.

The tests continue to rely on `docker-compose` for running Elasticsearch and Kibana. This enables us to easily keep Elasticsearch and Kibana running between test runs, to avoid repeatedly waiting for them to initialise.

## Checklist

~- [ ] documentation~
~- [ ] automated tests (add tests for the code changes, all [**unit** tests](https://github.com/elastic/apm-server/blob/master/TESTING.md) pass locally)~
~- [ ] telemetry~

## How to test these changes

```
cd systemtest
go test ./...
```

## Related issues

https://github.com/elastic/apm-server/issues/3318


jalvz:
Yes, I also found that more often than not some test hangs, but from the logs I can't really grok why.

axw:
Thanks for the feedback! I just pushed a fix, which I think may address the hangs you were seeing.

axw:
> Test logs are currently only created when running the server. It would also be nice to store logs for running the server CLI, e.g. using the apikey command.

I started looking into this, and it's a little bit complicated to do it nicely. I'd defer this unless you think it's going to be absolutely necessary.

The other concern I have is that running commands should ideally not require logging, as they're short-lived operations. Important information should be surfaced in stderr by default.

> nit: it would be nice to add a helper function similar to wait_until in our python tests.

I've added request options to the ES client which should help here. In particular, you can pass `estest.WithCondition(conditionFunc)` into `Do()`. A condition function is a function with the signature `func (*esapi.Response) bool`. The request will be retried until it returns true or a timeout passes, in which case an error is returned. e.g.

```go
var result estest.SearchResult
_, err := systemtest.Elasticsearch.Search("apm-*").WithQuery(estest.TermQuery{
        Field: "processor.event",
        Value: "onboarding",
}).Do(context.Background(), &result, estest.WithCondition(result.Hits.NonEmptyCondition()))
require.NoError(t, err)
```

This will repeatedly search `apm-*` until there's at least one onboarding doc. There's a default delay of 100ms which can be overridden by passing in `estest.WithInterval`, and a default timeout of 10s which can be overridden by passing in `estest.WithTimeout`.

script/jenkins/linux-test.sh
```diff
@@ -12,3 +12,6 @@ cleanup() {
 trap cleanup EXIT
 
 make update docker-system-tests
+
+# TODO(axw) make this part of the "system-tests" target
```
axw, line 16: I would do this when we remove the Python system tests, removing the `docker-system-tests` target and making `system-tests` target run the Go tests in the systemtest module.

axw:
@simitt good call, I'll do that. I was intending to leave it for a bit to build confidence. We can always revert if there's a problem though, so no need to be so conservative.

CI will already run these tests as part of the `System and Environment Tests` stage. See `./script/jenkins/linux-test.sh`.

axw:
I've removed the old Python tests that have been reimplemented in Go. The only one that might be contentious is the Jaeger gRPC one, which does not check doc contents. I'd prefer to merge this and do that later, since we already do approval checking in beater/jaeger.

systemtest/apikeycmd_test.go
```diff
@@ -0,0 +1,114 @@
...
+	assert.Equal(t, []interface{}{attrs["id"]}, result["invalidated_api_keys"])
+	assert.Equal(t, float64(0), result["error_count"])
+}
+
+func decodeJSONMap(t *testing.T, r io.Reader) map[string]interface{} {
+	var m map[string]interface{}
+	err := json.NewDecoder(r).Decode(&m)
+	require.NoError(t, err)
+	return m
+}
```
simitt, line 114: API Key commands are only testing the console output, but not that the API Keys are actually created, invalidated, etc. in ES.
This has been implicitly tested in the python tests by using the `wait_until` functions.

systemtest/apmservertest/command.go
```diff
@@ -0,0 +1,185 @@
...
+		return "", err
+	}
+	repoRoot := filepath.Clean(strings.TrimSpace(string(output)))
+	abspath := filepath.Join(repoRoot, "apm-server")
+	if runtime.GOOS == "windows" {
+		abspath += ".exe"
+	}
+
+	log.Println("Building apm-server...")
+	cmd := exec.Command("go", "build", "-o", abspath, "./x-pack/apm-server")
```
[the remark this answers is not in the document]
axw: Will add a TODO. We don't have any system tests for OSS today, right?

systemtest/apmservertest/config.go
```diff
@@ -0,0 +1,229 @@
...
+	data, err := json.Marshal(cfg)
+	if err != nil {
+		return nil, err
+	}
+
+	m := make(map[string]interface{})
+	for k, v := range extra {
+		m[k] = v
+	}
+	if err := json.Unmarshal(data, &m); err != nil {
```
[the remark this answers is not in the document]
axw: That's what I would expect too. I think there was a reason why I had it in this order, but I'm not sure why now. I'll switch the order for now, and rename to "base" if we need to switch it back.

systemtest/apikeycmd_test.go
axw: Good catch. I've added explicit "Authenticate" requests using the generated keys. Also added code to clean up API Keys for these tests.

PR #3791, opened 2020-05-13
[7.7] bump version through beats update

simitt:
## Motivation/summary
Update beats framework to bump version to `7.7.1`


NOTICE.txt
```diff
@@ -565,7 +565,7 @@ Contents of "LICENSE":
 
 --------------------------------------------------------------------
 Dependency: github.com/gofrs/uuid
-Version: v3.2.0
+Version: v3.3.0
```
graphaelli, line 568: any concerns here?

simitt: no, we only use the `uuid` for log entries, so any changes wouldn't be concerning; and the [diff](https://github.com/gofrs/uuid/compare/v3.2.0...v3.3.0) indicates they only implemented the `fmt.Formatter`.

graphaelli: thanks for checking

PR #3817, opened 2020-05-26
[7.x] Correctly process tags in RUMV3 metricsets. (#3816)

jalvz:
 - Correctly process tags in RUMV3 metricsets. (#3816)


axw:
Removing this from the test plan, as it is not user-facing. As mentioned on the original PR, RUM does not currently populate any metricset tags.

PR #3823, opened 2020-05-27
[7.x] Update index pattern path (#3822)

jalvz:
 - Update index pattern path (#3822)


jalvz:
Turns out that the corresponding Kibana backport was closed without merging - this is not needed for now.

PR #3820, opened 2020-05-26
Support custom response headers for RUM endpoints

jalvz:
## Checklist

- [x] I have signed the [Contributor License Agreement](https://www.elastic.co/contributor-agreement/).
- [x] I have updated [CHANGELOG.asciidoc](https://github.com/elastic/apm-server/blob/master/CHANGELOG.asciidoc)

I have considered changes for:
- [x] documentation
- [ ] logging (add log lines, choose appropriate log selector, etc.)
- [ ] metrics and monitoring (create issue for Kibana team to add metrics to visualizations, e.g. [Kibana#44001](https://github.com/elastic/kibana/issues/44001))
- [x] automated tests (add tests for the code changes, all [**unit** tests](https://github.com/elastic/apm-server/blob/master/TESTING.md) pass locally)
- [ ] telemetry
- [ ] Elasticsearch Service (https://cloud.elastic.co)
- [ ] Elastic Cloud Enterprise (https://www.elastic.co/products/ece)
- [ ] Elastic Cloud on Kubernetes (https://www.elastic.co/elastic-cloud-kubernetes)

## How to test these changes

Configure some headers in apm-server.rum.response_headers, send a request to any/all rum endpoints, and observe those headers in the response. Should work for both strings and lists

## Related issues

Closes #3696
Docs issue #3821

PR #3825, opened 2020-05-27
[7.x] Support custom response headers for RUM endpoints (#3820)

jalvz:
 - Support custom response headers for RUM endpoints (#3820)

### DIFF
--- a/_meta/beat.yml
+++ b/_meta/beat.yml
@@ -245,6 +245,10 @@ apm-server:
     # "Content-Encoding", and "Accept"
     #allow_headers : []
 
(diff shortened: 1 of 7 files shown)


axw:
Verified with BC2.

Configured apm-server with `-E apm-server.rum.response_headers={key: 'value'}`, and confirmed that each of the RUM endpoints responded with an additional header, "Key: value":
 - /config/v1/rum/agents
 - /intake/v2/rum/events
 - /intake/v3/rum/events

Also confirmed that non-RUM endpoints do _not_ respond with the header. This includes the root endpoint.

PR #3836, opened 2020-06-02
New tracer

jalvz:
## Checklist

- [ ] documentation
- [ ] telemetry

## Related issues

https://github.com/elastic/beats/pull/18861/


beater/tracing.go
```diff
@@ -43,23 +43,23 @@ func init() {
...
-// initTracer configures and returns an apm.Tracer for tracing
+// initLegacyTracer configures and returns an apm.Tracer for tracing
 // the APM server's own execution. If the server is configured
 // to send tracing data to itself, it will return a tracerServer
 // that can be used for receiving trace data.
-func initTracer(info beat.Info, cfg *config.Config, logger *logp.Logger) (*apm.Tracer, *tracerServer, error) {
+//
+// NOTE: this reads configuration from apm-server.instrumentation.* namespace
+// In 8.0 this will be removed, and the tracer will only be initialized by libbeat (configuration in instrumentation.*)
+func initLegacyTracer(info beat.Info, cfg *config.Config, logger *logp.Logger) (*apm.Tracer, *tracerServer, error) {
```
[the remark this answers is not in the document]
jalvz: SGTM

jalvz:
Notes:

* `make update` after beats update failed with

```
  File "script/generate_notice.py", line 354, in <module>
    modules = gather_modules(args.main_package, args.build_tags)
  File "script/generate_notice.py", line 118, in gather_modules
    print("WARNING: Unknown license for {}: {}".format(modpath, os.path.join(root, filename)))
NameError: name 'root' is not defined
```

`python ./script/generate_notice.py` seemed to fix, but don't know why there is a huge diff.

* `TestServerTracingExternal` still failing. But i'd say this should be moved to libbeat. Thoughts?

axw:
>  make update after beats update failed with
>>   File "script/generate_notice.py", line 354, in <module>
>>     modules = gather_modules(args.main_package, args.build_tags)
>>   File "script/generate_notice.py", line 118, in gather_modules
>>     print("WARNING: Unknown license for {}: {}".format(modpath, os.path.join(root, filename)))
>> NameError: name 'root' is not defined

There's an error in the script; like it says, `root` is an undefined variable - my mistake. However, that line is only reached when there's an unknown license found... so that needs to be fixed. I fixed the script locally and it's triggered by https://github.com/elastic/elastic-agent-client/blob/master/LICENSE.txt. We'll just need to add detection of the Elastic License. I'll look into this.

> python ./script/generate_notice.py seemed to fix, but don't know why there is a huge diff.

If you run the script directly, it's not going to be run the same as when you run `make update`. See the Makefile:

https://github.com/elastic/apm-server/blob/3d62feeff46cf27438546e09829cdc117c3db27d/Makefile#L130

I suppose the error is only occurring through `make update` because it includes x-pack packages, which will import elastic-agent-client.

beater/beater.go
```diff
@@ -130,16 +132,23 @@ type beater struct {
...
-	defer tracer.Close()
 
-	runServer := runServer
-	if tracerServer != nil {
-		runServer = runServerWithTracerServer(runServer, tracerServer, tracer)
+	tracer := b.Instrumentation.Tracer()
+	useLegacyTracer := common.MustNewVersion(version.GetDefaultVersion()).LessThan(&common.Version{Major: 8, Minor: 0})
+
+	if !tracer.Active() && useLegacyTracer {
+		tracer, tracerServer, err = initLegacyTracer(b.Info, bt.config, bt.logger)
```
axw, line 145: Instead of holding onto that old code, can we instead dynamically rename `apm-server.instrumentation` to `instrumentation`, if the latter doesn't exist in the config?

I think we can do this by defining a ConfigOverride in libbeat/cmd/instance.Settings, whose Check method checks for existing "instrumentation", and if not specified sets it to "apm-server.instrumentation"

beater/beater.go
```diff
@@ -130,16 +132,23 @@ type beater struct {
 // Run runs the APM Server, blocking until the beater's Stop method is called,
 // or a fatal error occurs.
 func (bt *beater) Run(b *beat.Beat) error {
-	tracer, tracerServer, err := initTracer(b.Info, bt.config, bt.logger)
+
+	tracerServer, err := newTracerServer(bt.config, b.Instrumentation.Listener())
```
axw, line 136: Shouldn't we only be creating a tracerServer if `b.Instrumentation.Listener()` returns non-nil? i.e. when instrumentation is enabled with no hosts specified

_meta/beat.yml
```diff
@@ -38,38 +38,8 @@ apm-server:
...
-      #cpu:
-        # Set to true to enable CPU profiling.
-        #enabled: false
-        #interval: 60s
-        #duration: 10s
-      #heap:
-        # Set to true to enable heap profiling.
-        #enabled: false
-        #interval: 60s
+  # DEPRECATION NOTICE: the `apm-server.instrumentation` configuration block is deprecated, and has been moved to
```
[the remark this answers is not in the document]
jalvz: My thinking was to not force the users onto docs, if possible. I imagine that if a setting dissappears when upgrading a minor they would be surprised/confused.
I'm fine with changing, don't have strong opinions.

axw:
> TestServerTracingExternal still failing. But i'd say this should be moved to libbeat. Thoughts?

That sounds good to me.

beater/beater.go
jalvz: Hmm how would that exactly work?
```
{
			Check: func(cfg *common.Config) bool {
				return cfg.HasField("instrumentation")
			},
			Config: ???
```

`Config` is just a config, not a function, so how do I copy `apm-server.instrumentation`?

axw: I meant make `Check` do the modification. It would be a bit of a hack, so it should really be replaced with something more fit for purpose. But in the mean time...

```go
Check: func(cfg *common.Config) bool {
    if !cfg.HasField("instrumentation") {
        // Check if cfg has "apm-server.instrumentation", and rename it to "instrumentation"
    }
    return true
}
```

jalvz: ah, uh. that's quite a hack, yes.. 
ok

simitt: How about not using `Check` but changing `libbeatConfigOverride` to be a function and taking care of the overwrite there?

jalvz: But how do I pass to a function the original config I need to copy?
Or you mean to change the whole thing in libbeat?

jalvz:
So tests relying on `beater_tests.newInstrumentation` fail now, I'm kinda stuck with those.
I think those tests should actually be moved system tests (i mean the ones in Python).
Anyone against?

beater/beater.go
axw: I presume @simitt meant changing libbeat. I think that's the right long-term solution. IMO, `ConditionalOverride` should be an implementation of an interface like `ConfigTransformer`:

```go
type ConfigTransformer interface {
    // TransformConfig transforms the provided config, returning a possibly modified config.
    TransformConfig(*common.Config) (*common.Config, error)
}

type ConditionalOverride struct {
    Check OverrideChecker
    Config *common.Config
}

func (o *ConditionalOverride) TransformConfig(in *common.Config) (*common.Config, error) {
    if !o.Check(in) {
        return in, nil
    }
    return common.MergeConfigs(in, o.Config)
}
```

We could then provide an alternative implementation of that interface which does what we need.

beater/tracing.go
```diff
@@ -19,112 +19,30 @@ package beater
...
-	listener  net.Listener
-	transport transport.Transport
+	cfg      *config.Config
+	logger   *logp.Logger
+	server   *http.Server
+	listener net.Listener
 }
 
-func newTracerServer(cfg *config.Config) (*tracerServer, error) {
+func newTracerServer(cfg *config.Config, listener net.Listener) (*tracerServer, error) {
```
axw, line 45:
```suggestion
func newTracerServer(cfg *config.Config, listener net.Listener) (*tracerServer) {
```

(AFAICS this function can't return any errors any more - right?)

axw:
> So tests relying on beater_tests.newInstrumentation fail now, I'm kinda stuck with those.
> I think those tests should actually be moved system tests (i mean the ones in Python).
> Anyone against?

I'm OK with that. It feels a bit heavy, but it's not code that needs to change often so I guess unit tests aren't really needed.

BTW, the issue with those tests is that you created a tracer but didn't configure its transport to send to the listener. If you do stick with the unit tests, maybe use `libbeat/instrumentation.New` rather than duplicating logic.

cmd/root.go
```diff
@@ -39,34 +39,62 @@ const (
...
-				},
-				"_source": map[string]interface{}{
-					"enabled": true,
 				},
 			},
 		},
+	}),
+},
+	{
+		Check: func(cfg *common.Config) bool {
```
axw, line 76: Can you please add a TODO here to update libbeat, adding a more fit for purpose API?

cmd/root_test.go
```diff
@@ -0,0 +1,47 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/elastic/beats/v7/libbeat/cfgfile"
+)
+
+func TestOverrideLegacyInstrumentationConfig(t *testing.T) {
+	cfg, err := cfgfile.Load("tests/legacy_instrumentation.yml", libbeatConfigOverrides)
```
axw, line 30: It's conventional to add test-related data files in a directory called "testdata", which the `go` tool has a special case to ignore when listing packages and such. Can you please rename?

PR #3849, opened 2020-06-05
Strip leading 0s from traceIds and spanIds if the span source is Jaeger

### LINKED ISSUE #3837: [Jaeger] store 64-bit trace IDs compactly
**Is your feature request related to a problem? Please describe.**

Attempting to correlate logs with traces produced by Jaeger fails by default when using 64-bit trace IDs. This is due to the fact that APM Server treats the trace ID as 128-bit and records the trace ID left-padded with zeros (32 hex digits), while Jaeger clients stringify trace IDs left-padded with zeros to either 16 hex digits (64-bit) or 32 hex digits (128-bit)

**Describe the solution you'd like**

When receiving trace data from Jaeger, if the high 64-bits of a trace ID are zero then treat as a 64-bit trace ID and write only the low 64-bits in hex encoding. e.g. if the trace ID received is `0000000000000000a7890b687652000e`, then store as `a7890b687652000e`.

**Describe alternatives you've considered**

Remove all leading zeros from the encoded trace ID. This could also lead to issues with log correlation, in the case of 64-bit trace IDs with the highest byte(s) set to zero.

**Additional context**
See https://discuss.elastic.co/t/apm-jaeger-and-trace-id/234978/2

tobiasstadler:
Strip leading 0s from traceIds and spanIds if the span source is Jaeger. This enables log correlation if the the application uses jaeger tracer.

Should fix #3837


tobiasstadler:
Could anyone start a CI run, please?

processor/otel/consumer.go
```diff
@@ -81,14 +82,43 @@ func (c *Consumer) convert(td consumerdata.TraceData) *model.Batch {
...
-		var parentID *string
 		root := len(otelSpan.ParentSpanId) == 0
-		if !root {
-			str := fmt.Sprintf("%x", otelSpan.ParentSpanId)
-			parentID = &str
+
+		var parentID, spanID, traceID string
+		if td.SourceFormat == sourceFormatJaeger {
+			if !root {
+				jaegerParentSpanID, err := tracetranslator.BytesToUInt64SpanID(otelSpan.ParentSpanId)
```
axw, line 90: Could you please extract this and the conversion of `otelSpan.SpanId` below into a separate function? Something like:

```go
func formatJaegerSpanID(id []byte) string {...}
```

And in that function, please add a comment with a link to https://github.com/jaegertracing/jaeger-client-go/blob/v2.23.1/span_context.go - or more canonical documentation, if you're aware of any.

processor/otel/consumer.go
```diff
@@ -81,14 +82,43 @@ func (c *Consumer) convert(td consumerdata.TraceData) *model.Batch {
...
+				if err != nil {
+					parentID = fmt.Sprintf("%x", otelSpan.ParentSpanId)
+				} else {
+					parentID = fmt.Sprintf("%x", jaegerParentSpanID)
+				}
+			}
+
+			jaegerTraceIDHigh, jaegerTraceIDLow, err := tracetranslator.BytesToUInt64TraceID(otelSpan.TraceId)
```
axw, line 98: As above, but `formatJaegerTraceID` of course

axw:
jenkins run the tests please

tobiasstadler:
> Could you please update the changelog? https://github.com/elastic/apm-server/blob/master/changelogs/head.asciidoc

Should I also add it to breaking changes?

axw:
> Should I also add it to breaking changes?

I think what you listed is fine. Jaeger support is still labelled experimental, so it's subject to changes like this.

axw:
Thanks again for your contribution @tobiasstadler, much appreciated!

PR #3855, opened 2020-06-08
[7.x] Strip leading 0s from traceIds and spanIds if the span source is Jaeger (#3849)

axw:
 - Strip leading 0s from traceIds and spanIds if the span source is Jaeger (#3849)


axw:
Verified with BC2. Sent some Jaeger spans, confirmed that they did not have leading 0s.

PR #3857, opened 2020-06-08
Index page URL as ECS field

jalvz:
## Motivation/summary

See:

Closes elastic/apm-server#3827

## Checklist

- [x] I have signed the [Contributor License Agreement](https://www.elastic.co/contributor-agreement/).
- [x] I have updated [CHANGELOG.asciidoc](https://github.com/elastic/apm-server/blob/master/CHANGELOG.asciidoc)

I have considered changes for:
- [ ] documentation
- [ ] logging (add log lines, choose appropriate log selector, etc.)
- [ ] metrics and monitoring (create issue for Kibana team to add metrics to visualizations, e.g. [Kibana#44001](https://github.com/elastic/kibana/issues/44001))
- [x] automated tests (add tests for the code changes, all [**unit** tests](https://github.com/elastic/apm-server/blob/master/TESTING.md) pass locally)
- [ ] telemetry
- [ ] Elasticsearch Service (https://cloud.elastic.co)
- [ ] Elastic Cloud Enterprise (https://www.elastic.co/products/ece)
- [ ] Elastic Cloud on Kubernetes (https://www.elastic.co/elastic-cloud-kubernetes)

## How to test these changes

Index a RUM event with something in `context.page`, and see it indexed as ECS fields in `url` and `http.request.referrer`.

Note: if agent sends `context.request.url`, that will be used for the `url` field, and not the page

PR #3865, opened 2020-06-10
[7.x] Index page URL and Referer as ECS fields (#3857)

jalvz:
 - Index page URL and Referer as ECS fields (#3857)


axw:
Verified with BC2. Created a barebones webpage instrumented with RUM:

```html
<html>
<head>
<script src="https://unpkg.com/@elastic/apm-rum/dist/bundles/elastic-apm-rum.umd.min.js" crossorigin></script>
<script>
elasticApm.init({
  serviceName: 'webapp',
  serverUrl: 'http://localhost:8200'
})
</script>
</head>
<body>
hello!
<a href="/">back</a>
</body>
```

Clicking the button takes you to the same page with referer set. Confirmed that `transaction.page.url` is exploded into `url.*`, and `transaction.page.referer` is copied to `http.request.referrer`.

image

PR #3866, opened 2020-06-10
Include field name in decoding errors

jalvz:
## Motivation/summary

Troubleshooting decoding errors is currently very hard because the error doesn't include the field name, and there are many call sites (and calls) that can return an error.

This adds the field name to the decoding error message.

## Checklist

- [ ] documentation
- [ ] telemetry

## How to test these changes

`make test`


axw:
Can you give an example of an error where this would have helped? I'm wondering why we're getting these "fetch" errors at all - shouldn't JSON Schema validation have prevented them?

jalvz:
@axw they are relevant basically for development. One example for instance when you get it is if you call the wrong decoder method (eg. use `decoder.String` instead of `decoder.StringPtr`, etc)

axw:
> @axw they are relevant basically for development. One example for instance when you get it is if you call the wrong decoder method (eg. use decoder.String instead of decoder.StringPtr, etc)

OK. So in this case it would be like the schema said something is optional, but in the code we made it required by using the wrong method. Fair enough.

PR #3884, opened 2020-06-13
Support for message_bus.destination in the Jaeger processor

tobiasstadler:
A Jaeger transaction/span with attribute message_bus.destination should be mapped to the Elastic APM type messaging


axw:
jenkins run the tests please

PR #3889, opened 2020-06-16
fix: check that docker is installed

kuisathaverat:
## What does this PR do?

It checks that Docker is installed before run it.

## Why is it important?

Some worker types do not have Docker installed this breaks the packer build.

### DIFF
--- a/.ci/packer_cache.sh
+++ b/.ci/packer_cache.sh
@@ -19,8 +19,9 @@ docker.elastic.co/beats-dev/golang-crossbuild:${GO_VERSION}-ppc
 docker.elastic.co/beats-dev/golang-crossbuild:${GO_VERSION}-s390x
 golang:${GO_VERSION}
 "
-
-for image in ${DOCKER_IMAGES}
-do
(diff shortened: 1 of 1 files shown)


.ci/packer_cache.sh
```diff
@@ -19,8 +19,9 @@ docker.elastic.co/beats-dev/golang-crossbuild:${GO_VERSION}-ppc
...
-for image in ${DOCKER_IMAGES}
-do
-(retry 2 docker pull ${image}) || echo "Error pulling ${image} Docker image, we continue"
-done
+if [ -x "$(command -v docker)" ]; then
+  for image in ${DOCKER_IMAGES}
+  do
+  (retry 2 docker pull ${image}) || echo "Error pulling ${image} Docker image, we continue"
+  done
+fi
```
simitt, line 27: What happens if docker is not installed? Will the CI just finish and pass?

kuisathaverat: for those packer images that do not have Docker installed packer will skip the docker pull, doing that, the build success. This does not affect to the APM server build, we do not use those images for builds, they are used for testing OS.

PR #3918, opened 2020-06-25
RUM V3: add context page fields

### LINKED ISSUE #3657: [RUM] Extend context.page for allowing new metrics as part of transaction
Currently RUM agent sends page specific information for all managed transactions as part of the transaction context. eg : 

```js
transaction.context.page =  {
    referer: "blah"
    url: "http://example.com"
}
```

some of the metrics like `Memory`, `Network information(4g, 3g)`, `Hardware concurrency(whether the user is on low/high end device)` can only be added to `context.custom` which feels not so intuitive. I would like to propose we find a common schema for adding these navigation level metrics as part of the transaction. It would help with some ideas like

+  Creating RUM dashboards and filtering for users on low end mobiles with limited memory coming from an `search` page. 

These are some of the metrics I thought we could send as part of the page-context. 

```js
{
  page: {
      referer: document.referrer,
      url: window.location.href,
      networkInfo: {
        effectiveType: '4g',
        rtt: 50,
        downlink: 10,
        saveData: false,
        downlinkMax: Infinity,
        type: 'wifi'
      },
      hardwareConcurrency: 4, // cores
      deviceMemory: 4 (float), // A full list of possible values is as follows: 0.25, 0.5, 1, 2, 4, 8
      servedViaServiceWorker: "yes"  // ["yes", "no", "unsupported/not applicable"] 
    }
} 
```

Does allowing * in the page context solve the use case for RUM since its already used only by the RUM agent ?

jalvz:
## Motivation/summary

Collect more data for RUM analytics.

## Checklist

- [ ] documentation
- [ ] telemetry

## How to test these changes

Send RUM V3 data with new fields, and see them indexed. 
Send RUM V3 data without new fields, and see that they are not rejected.

## Related issues

Closes elastic/apm-server#3657


model/context.go
```diff
@@ -107,10 +107,37 @@ func truncate(s string) string {
...
+func (ni *NetworkInfo) fields() common.MapStr {
+	fields := common.MapStr{}
+	if ni == nil {
+		return fields
+	}
+	utility.Set(fields, "type", ni.Type)
+	utility.Set(fields, "downlink", ni.Downlink)
+	utility.Set(fields, "downlinkMax", ni.DownlinkMax)
+	utility.Set(fields, "effectiveType", ni.EffectiveType)
+	utility.Set(fields, "roundTripTime", ni.RoundTripTime)
```
axw, line 128: Maybe just "rtt"? It's a pretty well known acronym.

model/context.go
```diff
@@ -107,10 +107,37 @@ func truncate(s string) string {
...
+	utility.Set(fields, "saveData", ni.SaveData)
```
axw, line 129: I don't see "savedata" in the [Network Information API](https://wicg.github.io/netinfo/#networkinformation-interface) spec. Seems to me it's more a property of the request (page) than the network information.

@vigneshshanmugam where does this property come from?

model/context.go
```diff
@@ -107,10 +107,37 @@ func truncate(s string) string {
 	return s
 }
 
+type NetworkInfo struct {
+	EffectiveType *string
+	RoundTripTime *int64
+	Downlink      *int64
+	DownlinkMax   *int64
```
axw, line 114: Downlink could be fractional, so should be a float64:

> The downlink attribute represents the effective bandwidth estimate in megabits per second, rounded to nearest multiple of 25 kilobits per second

Same for downlinkMax.

model/modeldecoder/context.go
```diff
@@ -218,14 +218,39 @@ func decodePage(raw common.MapStr, hasShortFieldNames bool, err error) (*model.P
...
+func decodeNetworkInfo(raw map[string]interface{}, hasShortFieldNames bool) (*model.NetworkInfo, error) {
+	decoder := utility.ManualDecoder{}
+	fieldName := field.Mapper(hasShortFieldNames)
+	networkInfo := &model.NetworkInfo{
+		EffectiveType: decoder.StringPtr(raw, fieldName("effectiveType")),
+		RoundTripTime: decoder.Int64Ptr(raw, fieldName("roundTripTime")),
+		Downlink:      decoder.Int64Ptr(raw, fieldName("downlink")),
+		DownlinkMax:   decoder.Int64Ptr(raw, fieldName("downlinkMax")),
+		SaveData:      decoder.BoolPtr(raw, fieldName("saveData")),
+		Type:          decoder.StringPtr(raw, fieldName("referer")),
```
axw, line 249:
```suggestion
		Type:          decoder.StringPtr(raw, fieldName("type")),
```

model/context.go
```diff
@@ -107,10 +107,37 @@ func truncate(s string) string {
...
 // Page consists of URL and referer
 type Page struct {
-	URL     *URL
-	Referer *string
+	URL                    *URL
+	Referer                *string
+	NetworkInfo            *NetworkInfo
+	Cores                  *int
+	Memory                 *int
+	ServedViaServiceWorker *string
```
axw, line 140: in https://github.com/elastic/apm-server/issues/3657#issuecomment-615128282 @vigneshshanmugam said we're still exploring whether this should be a boolean or keyword. Has that been clarified since?

model/context.go
```diff
@@ -107,10 +107,37 @@ func truncate(s string) string {
...
+
 // Page consists of URL and referer
 type Page struct {
-	URL     *URL
-	Referer *string
+	URL                    *URL
+	Referer                *string
+	NetworkInfo            *NetworkInfo
+	Cores                  *int
+	Memory                 *int
```
axw, line 139: This should be a float64: https://developer.mozilla.org/en-US/docs/Web/API/Navigator/deviceMemory

axw, line 139: NetworkInfo, Cores, and Memory don't really feel to me like they belong in "Page". They're properties of the device, not the page.

model/context.go
```diff
@@ -203,6 +230,10 @@ func (page *Page) Fields() common.MapStr {
 		utility.Set(fields, "url", page.URL.Original)
 	}
 	utility.Set(fields, "referer", page.Referer)
+	utility.Set(fields, "system.cpu.cores", page.Cores)
+	utility.Set(fields, "system.memory.total", page.Memory)
+	utility.Set(fields, "servedViaServiceWorker", page.ServedViaServiceWorker)
+	utility.Set(fields, "networkInfo", page.NetworkInfo.fields())
```
axw, line 236: As mentioned on https://github.com/elastic/apm-server/issues/3657#issuecomment-615058295, networkInfo feels to me like extensions of the "network" ECS fieldset. Do you see any problems with that?

e.g. networkInfo.type -> network.physical (or something like that; there's precedent in "network.transport")

model/context.go
```diff
@@ -107,10 +107,37 @@ func truncate(s string) string {
...
+}
+
+func (ni *NetworkInfo) fields() common.MapStr {
+	fields := common.MapStr{}
+	if ni == nil {
+		return fields
+	}
+	utility.Set(fields, "type", ni.Type)
+	utility.Set(fields, "downlink", ni.Downlink)
+	utility.Set(fields, "downlinkMax", ni.DownlinkMax)
```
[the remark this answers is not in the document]
jalvz: ok - but note we haven't been doing that for RUM so far (`timeToFirstByte`, `requestStart`, etc), changing gears now can be more confusing for people dealing with those fields down the line.

model/context.go
jalvz: maybe is not 100% known (particularly for non native speakers)?  what's the harm in `roundTripTime` ?

model/context.go
jalvz: my understand was that rum would not send decimal number anyways. will change..

model/context.go
jalvz: In https://github.com/elastic/apm-server/issues/3657#issuecomment-644658711 I understood that these fields will be queried/used in the context of a page load, so I thought it would be intuitive to set them there, following the original proposal.  
Having a `device` field or similar on the `transaction` doesn't feel any more semantically appropriate to me (a device is not a property of a transaction), and using metricsets as I suggested in the issue would not help querying.

I guess it could be shoehorned in transaction context - just because context means pretty much anything...
Do you have any suggestion in particular?

model/context.go
vigneshshanmugam: You are right, the spec network Information API got updated and it no longer has `saveData` but its supported by chrome and might be added in future so we can spec it and remove usage if not required. 
Screenshot 2020-06-26 at 10 47 47

vigneshshanmugam: Preview is here - https://s3.amazonaws.com/pr-preview/WICG/netinfo/saveData.html

model/context.go
vigneshshanmugam: It will be string with predefined values. Updated the description with more context.

model/context.go
vigneshshanmugam: Good point Andrew, I am also inclined more towards having it as part of `device` that page. But would it affect/confused with the user_agent.device in any way?

docs/spec/rum_v3_context.json
```diff
@@ -153,6 +153,81 @@
...
+                            "maxLength": 1024
+                        },
+                        "rtt": {
+                            "description": "roundTripTime",
+                            "type": [
+                                "integer",
+                                "null"
+                            ]
+                        },
+                        "dlk": {
```
vigneshshanmugam, line 179: any reason for three letter combination instead of just `d` ? 
```
downlink -> d
downlinkMax -> dm
saveData -> sd
```

that would align with the camelCasing of the spec.

jalvz: d and sd are all already used for duration and started fields, so we would know which one refers too. we could potentially resolve ambiguity by looking at the parent field, but that would entail more complexity than what i am willing to introduce...

model/context.go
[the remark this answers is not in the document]
axw: The motivation behind my request was to move these fields closer to something that we could add to ECS. These network-related fields are more general than timeToFirstByte etc., which are pretty RUM-specific. I'm fine with the latter being non-ECS compliant for the foreseeable future.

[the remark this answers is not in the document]
jalvz: fare enough, then ill stick basically to those 2 (`downlink_max`, `effective_type`), doesn't look like anything else can be ECS-ified.

model/context.go
jalvz: no, no problem. but i understand right `network` should still be under `transaction.device`, and not as top level field?

docs/spec/rum_v3_context.json
vigneshshanmugam: Fair enough, not a big of a deal. I forgot that we dont take depth of parent in to consideration. Let keep it as it is for now.

model/context.go
axw: There's nothing wrong with "roundTripTime"; I'm arguing that it's more common to write "rtt" than it is "round-trip time" in technical settings. It is of course not known by 100% of people, just as the phrase wouldn't be; but it is a standard technical phrase and acronym in our industry: https://en.wikipedia.org/wiki/Round-trip_delay_time.

Looking through beats, I didn't find any fields with "roundTripTime" or "round_trip_time" in the name. There are several occurrences of fields with "rtt" in the name though:
- https://www.elastic.co/guide/en/beats/filebeat/current/exported-fields-googlecloud.html#_vpcflow_2
- https://www.elastic.co/guide/en/beats/heartbeat/current/exported-fields-tls.html#_rtt_6 (repeats for other protocols)

model/context.go
axw: > Having a device field or similar on the transaction doesn't feel any more semantically appropriate to me (a device is not a property of a transaction), and using metricsets as I suggested in the issue would not help querying.

Fair point that it's not a property of the transaction. What I was getting at is that this information is not necessarily restricted to page-loads, so let's not overfit to this one requirement. These fields may also be applicable for serverless/lambda.

Here's a more concrete suggestion:

 1. extend our `system` model type with optional fields to report the the number of CPU cores and total memory
 2. add a `network` model type which can be reported in stream metadata
 3. allow `transaction.context.system` which should be merged with the system from stream metadata, like we do with `transaction.context.service`
 4. same for `transaction.context.network`

We could skip 3-4 if we want these properties to be associated with spans, error and metrics as well as transactions.

model/context.go
axw: No, where I talked about "device" I was just talking concepts. The recorded fields should be top-level.

model/context.go
```diff
@@ -107,12 +108,42 @@ func truncate(s string) string {
 	return s
 }
 
+type NetworkInfo struct {
+	EffectiveType *string
+	RoundTripTime *int64
+	Downlink      *float64
+	DownlinkMax   *float64
+	SaveData      *bool
```
axw, line 116: Can we please separate `SaveData` out? It's not a property of the network like all the other things, it's both network-related and specific to the page-load request.

Maybe this one can be a direct field of `Page`.

model/context.go
```diff
@@ -107,12 +108,42 @@ func truncate(s string) string {
...
 type Page struct {
 	URL     *URL
 	Referer *string
 }
 
+type Device struct {
+	NetworkInfo            *NetworkInfo
+	Cores                  *int
+	Memory                 *float64
+	ServedViaServiceWorker *string
```
axw, line 144: Please move `ServedViaServiceWorker` to `Page`. This is not related to the device, and is page-load specific.

docs/spec/rum_v3_network.json
```diff
@@ -0,0 +1,48 @@
...
+            ]
+        },
+        "dlkm": {
+            "description": "downlink_max",
+            "type": [
+                "number",
+                "null"
+            ]
+        },
+        "ph": {
```
vigneshshanmugam, line 39: To what field, this maps to the Network Info API?

axw:
Thanks for the updates @jalvz, it's looking much better to me now.

I'd like to hold off on this change for the moment, and probably defer to 7.10. I'm concerned that if we push this, we may end up getting the field names wrong (with respect to ECS), and have to rework them and deal with all of the backwards compatibility issues that entails. As far as I'm aware we don't strictly need this for 7.9, so I'd prefer to take a more measured approach.

To take this forward, I propose we:
 1. Move this PR into Draft
 2. Replace https://github.com/elastic/apm-server/issues/3657 with an elastic/apm issue using the [New Metadata/Context Fields issue](https://github.com/elastic/apm/issues/new?labels=agents%2C+poll&template=new-fields-issue.md) template, using the current state of this PR as the basis
 3. Once we're agreed upon that, raise one or more [RFCs](https://github.com/elastic/ecs/tree/master/rfcs) for ECS

jalvz:
> using the current state of this PR as the basis

Just the basis indeed, as the fields have only been spec'ed for RUM V3

docs/spec/rum_v3_network.json
jalvz: short for "physical layer", equivalent to "type".

vigneshshanmugam:
@axw @jalvz Should we hold on to the implementation from the RUM agent side? Suggestions?

axw:
@vigneshshanmugam I think it would be best not to merge anything yet.

axw:
This is very stale, closing until there's some resolution of adding network details to ECS: https://github.com/elastic/apm-server/issues/3657#issuecomment-788538219

PR #3926, opened 2020-07-01
[7.x] aggregation/txmetrics: aggregate user_agent.name (#3886)

axw:
 - aggregation/txmetrics: aggregate user_agent.name (#3886)


axw:
Note to self: let's test transaction duration histogram metrics using TSVB histogram support.

PR #3927, opened 2020-07-01
[7.x] Support for message_bus.destination in the Jaeger processor (#3884)

axw:
 - Support for message_bus.destination in the Jaeger processor (#3884)


axw:
See https://github.com/elastic/apm-server/pull/3928#issuecomment-662917047

PR #3897, opened 2020-06-17
Support for peer.service in the Jaeger processor

tobiasstadler:
According to https://opentracing.io/specification/conventions/ 'peer.service' is similar to 'span.destination.service.name'

PR #3923, opened 2020-06-30
[7.x] New tracer (#3836)

jalvz:
## Motivation/summary

* Backports https://github.com/elastic/apm-server/pull/3836\

* Also includes https://github.com/elastic/apm-server/pull/3909/commits/4bed7bb64eb05feed6aa70dad37e78c68b7aeb30 and a Beats update

## Checklist

- [x] I have signed the [Contributor License Agreement](https://www.elastic.co/contributor-agreement/).
- [ ] I have updated [CHANGELOG.asciidoc](https://github.com/elastic/apm-server/blob/master/CHANGELOG.asciidoc)

I have considered changes for:
- [x] documentation
- [ ] logging (add log lines, choose appropriate log selector, etc.)
- [ ] metrics and monitoring (create issue for Kibana team to add metrics to visualizations, e.g. [Kibana#44001](https://github.com/elastic/kibana/issues/44001))
- [ ] automated tests (add tests for the code changes, all [**unit** tests](https://github.com/elastic/apm-server/blob/master/TESTING.md) pass locally)
- [ ] telemetry
- [ ] Elasticsearch Service (https://cloud.elastic.co)
- [ ] Elastic Cloud Enterprise (https://www.elastic.co/products/ece)
- [ ] Elastic Cloud on Kubernetes (https://www.elastic.co/elastic-cloud-kubernetes)

## How to test these changes

- Try both the new configuration and old instrumentation config settings and see that they both work, but the new one takes precedence.
- See transactions from both libbeat and apm-server in the UI (or Elasticsearch index if you prefer)
- Test that is disabled by default
- Test it with explicitly setting `instrumentation.enabled=false`, and see no transactions in the UI.

PR #3928, opened 2020-07-01
[7.x] Support for peer.service in the Jaeger processor (#3897)

axw:
 - Support for peer.service in the Jaeger processor (#3897)

### DIFF
--- a/processor/otel/consumer.go
+++ b/processor/otel/consumer.go
@@ -321,6 +321,7 @@ func parseSpan(span *tracepb.Span, event *model.Span) {
 	var message model.Message
 	var db model.DB
 	var destination model.Destination
+	var destinationService model.DestinationService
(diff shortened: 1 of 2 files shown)


axw:
Verified with BC2, along with #3923 

```go
tx := tracer.StartSpan("tx")                                                              
span := tracer.StartSpan("frozzlet",
        opentracing.ChildOf(tx.Context()),
        opentracing.Tag{Key: "peer.service", Value: "peer_service"},
        opentracing.Tag{Key: "message_bus.destination", Value: "message_bus_destination"},
        opentracing.Tag{Key: "peer.hostname", Value: "peer_hostname"},
)
span.Finish()                                                                             
tx.Finish()                                                                               
```

Checked in Discover: `destination.address`, `span.destination.service.name`, and `span.message.queue.name` are all set to the specified values.

image

PR #3903, opened 2020-06-18
Use more inclusive language

axw:
## Motivation/summary

Remove use of "whitelist" and "blacklist".

I have not changed the testdata (sourcemaps), nor the docs copied from beats. The latter should be changed upstream.

## Checklist

- [x] I have signed the [Contributor License Agreement](https://www.elastic.co/contributor-agreement/).
~- [ ] I have updated [CHANGELOG.asciidoc](https://github.com/elastic/apm-server/blob/master/CHANGELOG.asciidoc)~

I have considered changes for:
- [x] documentation
~- [ ] logging (add log lines, choose appropriate log selector, etc.)~
~- [ ] metrics and monitoring (create issue for Kibana team to add metrics to visualizations, e.g. [Kibana#44001](https://github.com/elastic/kibana/issues/44001))~
- [x] automated tests (add tests for the code changes, all [**unit** tests](https://github.com/elastic/apm-server/blob/master/TESTING.md) pass locally)
~- [ ] telemetry~
~- [ ] Elasticsearch Service (https://cloud.elastic.co)~
~- [ ] Elastic Cloud Enterprise (https://www.elastic.co/products/ece)~
~- [ ] Elastic Cloud on Kubernetes (https://www.elastic.co/elastic-cloud-kubernetes)~

## How to test these changes

`make test`

## Related issues

#3896

PR #3930, opened 2020-07-01
[7.x] Use more inclusive language (#3903)

axw:
 - Use more inclusive language (#3903)


axw:
unrelated CI failure

PR #3951, opened 2020-07-07
Fix legacy tracer initialization

### LINKED ISSUE #3950: Legacy tracer config not recognized when passed as command line arg
While `apm-server.instrumentation.enabled: true` works as expected if set in `apm-server.yml`, the same is not true when passed as argument with `-E`.

This is because command line config settings are merged after the overrides hook is called from libbeat, which is what we are using to copy `apm-server.instrumentation` to `instrumentation`.

Caused by #3836

jalvz:
## Motivation/summary

Change the approach to init `apm-server.instrumentation`: handle it in `beater.go` instead of relying on the config overrides logic

**Update**: additionally, enable the profile endpoint always, without checking the configuration. this wasn't working neither when using the new configuration from libbeat.  

## Checklist

- [ ] documentation
- [ ] telemetry

## How to test these changes

- Check that both `apm-server.instrumentation.*` and `instrumentation.*` settings work when reading config from a file or passing it with `-E` command args.

- Check that `instrumentation.profile.*` settings work as well.

- Check that `hosts` config works when explicitly set.

## Related issues

Fixes #3950


jalvz:
jenkins run hey-apm tests please

beater/api/mux.go
```diff
@@ -91,15 +91,7 @@ func NewMux(beaterConfig *config.Config, report publish.Reporter) (*http.ServeMu
...
-	}
-
-	// Profiling is currently experimental, and intended for profiling the
-	// server itself, so we only add the route if self-profiling is enabled.
-	if beaterConfig.SelfInstrumentation.IsEnabled() {
-		if beaterConfig.SelfInstrumentation.Profiling.CPU.IsEnabled() ||
-			beaterConfig.SelfInstrumentation.Profiling.Heap.IsEnabled() {
-			routeMap = append(routeMap, route{ProfilePath, profileHandler})
-		}
+		{ProfilePath, profileHandler},
```
simitt, line 94: nit: could you still keep a comment that the profiling is currently beta

beater/beater.go
```diff
@@ -217,6 +218,51 @@ func (bt *beater) registerPipelineCallback(b *beat.Beat) error {
...
+}
+
+// initLegacyTracer exists for backwards compatibility and it should be removed in 8.0
+// it does not instrument the beat output
+func initLegacyTracer(info beat.Info, cfg *config.Config) (*apm.Tracer, net.Listener, error) {
+	selfInstrumentation := cfg.SelfInstrumentation
+	if selfInstrumentation == nil || !selfInstrumentation.IsEnabled() {
+		return apm.DefaultTracer, nil, nil
+	}
+	conf := common.MustNewConfigFrom(cfg.SelfInstrumentation)
```
simitt, line 244: `MustNewConfigFrom` panics in case an error occurs. So far we use it in tests or when the whole input config has fixed values. I don't think we shoul be using it with user provided input and suggest to instead use `NewConfigFrom`  with the additional error check.

tests/system/test_instrumentation.py
```diff
@@ -15,6 +15,8 @@
 os.environ["ELASTIC_APM_API_REQUEST_TIME"] = "1s"
 
 
+# Exercises the DEPRECATED apm-server.instrumentation.* config
+
```
simitt, line 19: The tests in `test_libbeat_instrumentation` are copies of this file. I assume you didn't DRY the tests based on the fact that the support for `apm-server.instrumentation`  will be removed in 8.0.?
Please add a comment in both, to make future changes to the test cases in both files, and to delete this file for 8.0.

PR #3965, opened 2020-07-13
script: vendor gimme

axw:
## Motivation/summary

Vendor https://github.com/travis-ci/gimme (v1.5.4), rather than fetching it from GitHub. This should help improve resilience to GitHub flakiness.

## Checklist

~- [ ] documentation~
~- [ ] automated tests (add tests for the code changes, all [**unit** tests](https://github.com/elastic/apm-server/blob/master/TESTING.md) pass locally)~
~- [ ] telemetry~

## How to test these changes

`./script/run_with_go_ver go version`

## Related issues

None.

### DIFF
--- a/script/common.bash
+++ b/script/common.bash
@@ -34,20 +34,6 @@ get_go_version() {
   fi
 }
 
-# install_gimme
-# Install gimme to HOME/bin.
-install_gimme() {
-  # Install gimme
-  if [ ! -f "${HOME}/bin/gimme" ]; then
-    mkdir -p ${HOME}/bin
-    curl -sL -o ${HOME}/bin/gimme https://raw.githubusercontent.com/travis-ci/gimme/v1.1.0/gimme
-    chmod +x ${HOME}/bin/gimme
-  fi
-
-  GIMME="${HOME}/bin/gimme"
-  debug "Gimme version $(${GIMME} version)"
-}
-
 # setup_go_root "version"
 # This configures the Go version being used. It sets GOROOT and adds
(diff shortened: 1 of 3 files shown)


axw:
> Looks good to me, got some concerns for using such a big script, wondering if we should support gimme by default in the workers to avoid maintaining this script. It should be a quite easy feature to be added, and should be available in the PATH, happy to do it as similar done for the gvm what do you think?

It's 28KB, which is nothing compared to what we used to how much Go source code we used to vendor. I'd prefer to vendor for simplicity, as it also means we have a consistent method for developer and CI environments.

kuisathaverat:
or we can use [gvm](https://github.com/andrewkroh/gvm) that it is installed on all workers, and it is the procedure we use on other projects

axw:
@kuisathaverat I'm happy to consolidate on that for CI/release-manager, but it appears that beats uses gimme in release-manager:

https://github.com/elastic/beats/blob/29b2d27bde28258c6ec645d35e5d9857cfc0eed8/dev-tools/run_with_go_ver#L16
https://github.com/elastic/beats/blob/29b2d27bde28258c6ec645d35e5d9857cfc0eed8/dev-tools/common.bash#L55-L65

AFAICS, that's _only_ used by release-manager. So I think we should probably ensure gvm is installed in the release-manager VMs, and then consolidate on gvm.

I'll merge this now, as there's a time sensitive release-manager issue (caused by caching GitHub failures) that this PR will resolve. Let's do the gvm stuff separately.

Issue #2342, opened 2019-06-27
[Automation] Lint stage


kuisathaverat:
the intake stage invokes the `check-full` target here https://github.com/elastic/apm-server/blob/9f9eb1913e93c06d6936a95107cbc61920651958/script/jenkins/intake.sh#L13, which in turn invokes the linter here https://github.com/elastic/apm-server/blob/bdd4550116fa50a288f37b8ff6a46f4a25d2b2fd/Makefile#L116

it would be to take the linter out of `check-full` in the make file and run it alone in a separate stage after the integration tests

cachedout:
Closing after discussion in https://github.com/elastic/apm-server/pull/3966

PR #3966, opened 2020-07-13
Move lint to separate stage in Jenkins

cachedout:
This removes linting from the `check-full` Makefile target and instead relies on a separate stage in Jenkins to lint.

Closes https://github.com/elastic/apm-server/issues/2342


cachedout:
jenkins run the tests please

Makefile
```diff
@@ -65,7 +65,7 @@ SYSTEM_TEST_TARGET?=./tests/system
 NOSETESTS_OPTIONS?=--process-timeout=90 --with-timer -v --with-xunit --xunit-file=build/TEST-system.xml
 
 .PHONY: check-full
-check-full: update check golint staticcheck
+check-full: update check staticcheck
```
axw, line 68: I'm not super keen on removing targets from check-full -- it's no longer "full" then, adding another thing for developers to remember to run.

Can we either run golint twice (probably no big deal), or explicitly list the steps to run in CI in the non-lint stage?

Jenkinsfile
```diff
@@ -481,6 +481,17 @@ pipeline {
         }
       }
     }
+    stage('Lint'){
```
axw, line 484: I'm not sure if I'm reading the Jenkinsfile properly, but I guess this runs Lint after everything else - at least that's what was proposed in https://github.com/elastic/apm-server/issues/2342. What's the rationale for placing it late in the pipeline?

cachedout: TBH, I had this same question. I picked this issue up out of the backlog and so I went with the assumption that this point had been previously discussed. If this was not previously discussed, I agree that it should be moved earlier.

axw: Hmm, but now I'm wondering why are we moving it out of "Intake" at all :confused: 
@kuisathaverat do you recall the reasoning for splitting this out in https://github.com/elastic/apm-server/issues/2342?

kuisathaverat: to have fast feedback on lint errors.

axw: OK, thanks. I'd suggest we make it the first stage then. I'm not super convinced this is worthwhile though, since Intake takes ~4m30s.

cachedout: I just picked this up out of our backlog, to be honest, so I'm not too committed to any course of action here.

It sounds to me like there isn't much appetite for this change and I agree that it isn't buying much here. I'm going to close this PR and the associated issue and if this concern arises again, we can start a new discussion.

axw: SGTM, thanks @cachedout!

PR #3969, opened 2020-07-13
Use peer.hostname for destination.address if peer.address is not given

tobiasstadler:


axw:
jenkins run the tests please

PR #3970, opened 2020-07-14
[7.x] Use peer.hostname for destination.address if peer.address is not given (#3969)

axw:
 - Use peer.hostname for destination.address if peer.address is not given (#3969)


axw:
See https://github.com/elastic/apm-server/pull/3928#issuecomment-662917047

PR #3975, opened 2020-07-15
Use peer.address for DestinationService.Resource when converting a Jaeger span

tobiasstadler:


axw:
jenkins run the tests please

PR #4012, opened 2020-07-22
Prep golang env for package build test

cachedout:
This corrects an error discovered in the newly provisioned pipeline to smoke test the package installation for apm-server:

https://apm-ci.elastic.co/job/apm-server/job/apm-server-check-packages/1/console

### DIFF
--- a/.ci/check-packages.groovy
+++ b/.ci/check-packages.groovy
@@ -60,12 +60,14 @@ pipeline {
               deleteDir()
               unstash 'source'
               dir("${BASE_DIR}"){
-                sh(label: 'make batch',
-                  script: """#!/bin/bash
-                    echo "beats_url_base: ${BEATS_URL_BASE}" > run-settings-jenkins.yml
-                    echo "apm_url_base: ${APM_URL_BASE}" >> run-settings-jenkins.yml
-                    echo "version: ${VERSION}" >> run-settings-jenkins.yml
-                    RUN_SETTINGS=jenkins make batch""")
+                withGoEnv(){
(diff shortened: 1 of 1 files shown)


.ci/check-packages.groovy
```diff
@@ -60,12 +60,14 @@ pipeline {
               deleteDir()
               unstash 'source'
               dir("${BASE_DIR}"){
-                sh(label: 'make batch',
-                  script: """#!/bin/bash
-                    echo "beats_url_base: ${BEATS_URL_BASE}" > run-settings-jenkins.yml
-                    echo "apm_url_base: ${APM_URL_BASE}" >> run-settings-jenkins.yml
-                    echo "version: ${VERSION}" >> run-settings-jenkins.yml
-                    RUN_SETTINGS=jenkins make batch""")
+                withGoEnv(){
```
v1v, line 63: neat! I even forget about this :)

mdelapenya: Would this step provide all the Go variables? If that's correct, that would simplify the pipeline for the e2e tests a lot!

kuisathaverat: `withGoEnv` install go and set the environment, there is another one that install also mage `withMageEnv`, at the moment is important to use it before change the folder to the source code, if not, you could modify the mod.go file, we have to add the read-only option for the mod.go.

```
withGoEnv(){
  dir('src'){
...
  }
}
```

mdelapenya: Got it, I'll use the go step, taking into consideration your advice about the DIR usage

PR #4028, opened 2020-07-27
take ownership of version

graphaelli:
Since moving to go modules APM Server has been unable to bump versions independently of beats (before `vendor/` could be edited directly).


magefile.go
```diff
@@ -66,7 +66,9 @@ func Build() error {
 // GolangCrossBuild build the Beat binary inside of the golang-builder.
 // Do not use directly, use crossBuild instead.
 func GolangCrossBuild() error {
-	return mage.GolangCrossBuild(mage.DefaultGolangCrossBuildArgs())
+	args := mage.DefaultGolangCrossBuildArgs()
+	args.WinMetadata = false
```
graphaelli, line 70: This works but probably isn't the best solution

magefile.go
```diff
@@ -45,7 +45,7 @@ import (
 
 func init() {
 	mage.SetBuildVariableSources(&mage.BuildVariableSources{
```
axw, line 47: Seems we need to use absolute paths, because the cross-build happens from inside the x-pack directory. (Or alternatively maybe we should change something in dev-tools to stop cd'ing into x-pack? Not sure why that's being done.)

```go
repoInfo, err := mage.GetProjectRepoInfo()
if err != nil {
        panic(err)
}
mage.SetBuildVariableSources(&mage.BuildVariableSources{
        BeatVersion: filepath.Join(repoInfo.RootDir, "version", "version.go"),
        GoVersion:   filepath.Join(repoInfo.RootDir, ".go-version"),
        DocBranch:   filepath.Join(repoInfo.RootDir, "docs/version.asciidoc"),
})
```

magefile.go
graphaelli: Agreed. Switched to the repo info, approach, thanks for the pointer

graphaelli:
I hadn't even considered that, pushed up that change.

PR #4046, opened 2020-07-31
processor/otel: set destination.* for Jaeger

axw:
## Motivation/summary

Extend the handling of Jaeger semantic conventions for setting `destination.*` fields to include HTTP spans which do not include `peer.*` attributes.

- We now set `destination.{address,port}` by parsing `http.url` if `peer.*` have not been specified.
- We set `destination.service.resource` to the same value as `destination.service.name` if `peer.service` has been specified. Otherwise, we set both by parsing `http.url`.
- If any `destination.service.*` fields are set, we set `destination.service.type` to the span type.

Note that there is no specific handling for database spans, because the DBMS is not always known. Instead, we rely on the user to set `peer.service` to identify the downstream service.

## Checklist

~- [ ] documentation~
~- [ ] telemetry~

## How to test these changes

- Run apm-server with Jaeger gRPC, and jaeger-agent
- Run https://gist.github.com/axw/0452ba180b5ca39d938771804a041fb2
- curl http://localhost:1234

image

## Related issues

None


processor/otel/consumer.go
```diff
@@ -428,12 +443,55 @@ func parseSpan(span *tracepb.Span, event *model.Span) {
...
+				}
+			}
+
+			// Set destination.{address,port} from the HTTP URL,
+			// if no peer.* attributes were specified.
+			if destination.Address == nil {
+				destination.Address = &hostname
+			}
+			if destination.Port == nil && port > 0 {
+				destination.Port = &port
```
axw, line 469: We should use the scheme default port here, if port is unspecified.

processor/otel/consumer.go
```diff
@@ -397,27 +399,40 @@ func parseSpan(span *tracepb.Span, event *model.Span) {
...
+			case "peer.hostname", "peer.ipv4", "peer.ipv6":
+				val := truncate(v.StringValue.Value)
+				destination.Address = &val
 			case "peer.service":
-				destinationService.Name = &v.StringValue.Value
+				val := truncate(v.StringValue.Value)
+				destinationService.Name = &val
+				if destinationService.Resource == nil {
+					// Prefer using peer.address for resource.
+					destinationService.Resource = &v.StringValue.Value
```
simitt, line 431: Why are you not also using the truncated value here?

processor/otel/consumer.go
```diff
@@ -428,12 +443,50 @@ func parseSpan(span *tracepb.Span, event *model.Span) {
...
+			}
+
+			// Set destination.{address,port} from the HTTP URL,
+			// if no peer.* attributes were specified.
+			if destination.Address == nil {
+				destination.Address = &hostname
+			}
+			if destination.Port == nil && port > 0 {
+				destination.Port = &port
+			}
```
simitt, line 465: This might lead to `destination.Address` being set by `peer.*`  and `destination.Port` being derived from `http.URL` and vice versa - not sure if this is a real world concern or not, but might be confusing.

processor/otel/consumer.go
axw: Accidental, thanks for picking up - will fix

processor/otel/consumer.go
axw: This is mostly to cater for taking the scheme default port. I suppose we could always set both `destination.address` and `destination.port` from `http.url` if provided. Do you think that would be less confusing?

axw: As that would be in line with what we do in our own agents, I'll go ahead and make that change.

Issue #3334, opened 2020-02-14
APM Server Telemetry


jalvz:
Investigate how can we start reporting Telemetry metrics, and which metrics would be interesting for us to report.

Libbeat gives support through the `monitoring` package:
https://github.com/elastic/beats/blob/master/libbeat/publisher/pipeline/module.go#L42
 
We want a registry in the `state` namespace, and then use it eg. like:
`monitors.NewString(telemetry, "key").Set("value")`

jalvz:
Maybe for inspiration: https://github.com/elastic/kibana/issues/50757

graphaelli:
next step is to come up with list of telemetry that we need to track

jalvz:
Initial proposal:

- How many are using custom pipelines
`register.ingest.pipeline.overwrite`

- How many are using API keys
`api_key.enabled`

- How many are using RUM (although I think is possible to track this already?):
`rum.enabled`

- How many are potentially using remove config 
`kibana.enabled`

- How many are using custom ILM policies
`ilm.setup.overwrite`

- How many are configuring custom ILM policies from apm-server
`ilm.setup.require_policy`

- How many are using Jaeger/GRPC
`jaeger.grpc.enabled`

- How many are using Jaeger/HTTP
`jaeger.http.enabled`

- How many are tuning for performance
`queue.mem.events != 4096`
`queue.mem.flush.min_events != 2048`

- How many are using custom templates
`setup.template.overwrite`

- How many are using the Elasticsearch output
`output.elasticsearch.enabled`

- How many are using the Logstash output
`output.logstash.enabled`

- How many are using the Kafka output
`output.kafka.enabled`

axw:
> queue.mem.events != 4096
> queue.mem.flush.min_events != 2048

Any reason why we should not just record the values?

I would also add `ilm.enabled`

axw:
Sample "state" monitoring doc, based on existing libbeat monitoring:

```json
          "timestamp" : "2020-06-26T07:50:16.744Z",
          "type" : "beats_state",
          "beats_state" : {
            "beat" : {
              "type" : "apm-server",
              "version" : "8.0.0",
              "name" : "goat",
              "host" : "goat",
              "uuid" : "0ddccead-2600-47c2-b488-9704600e82bb"
            },
            "state" : {
              "host" : {
                "architecture" : "x86_64",
                "os" : {
                  "kernel" : "5.4.0-28-generic",
                  "codename" : "focal",
[...]

PR #4055, opened 2020-08-06
Add telemetry for apm-server configuration

jalvz:
## Motivation/summary

As title

## Checklist

- [ ] documentation
- [x] telemetry

## Related issues

#3334 

Needs Kibana work and mappings in the telemetry cluster indices


beater/telemetry.go
```diff
@@ -0,0 +1,118 @@
...
+	setupTemplateEnabled      *monitoring.Int
+	setupTemplateOverwrite    *monitoring.Int
+	setupTemplateAppendFields *monitoring.Int
+	ilmEnabled                *monitoring.Int
+	ilmSetupEnabled           *monitoring.Int
+	ilmSetupOverwrite         *monitoring.Int
+	ilmSetupRequirePolicy     *monitoring.Int
+	jaegerGRPCEnabled         *monitoring.Int
+	jaegerHTTPEnabled         *monitoring.Int
+	sslEnabled                *monitoring.Int
```
axw, line 47: What's the reason for making these ints, and not bools? In libbeat, the `management.enabled` field is a bool: https://github.com/elastic/beats/blob/7ec32e4d2e1dcd9dcfb9b61e49fff794deeb2c2a/libbeat/cmd/instance/beat.go#L322

beater/telemetry.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+var configMonitors = &configTelemetry{
+	rumEnabled:                monitoring.NewInt(configRegistry, "rumEnabled"),
```
axw, line 51: What's the reason for diverging from the config names we use? Why `rumEnabled` and not `rum.enabled`?

beater/telemetry.go
```diff
@@ -0,0 +1,118 @@
...
+	"github.com/elastic/apm-server/beater/config"
+	"github.com/elastic/apm-server/idxmgmt"
+	"github.com/elastic/beats/v7/libbeat/beat"
+	"github.com/elastic/beats/v7/libbeat/common"
+	"github.com/elastic/beats/v7/libbeat/idxmgmt/ilm"
+	"github.com/elastic/beats/v7/libbeat/monitoring"
+)
+
+var apmRegistry = monitoring.GetNamespace("state").GetRegistry().NewRegistry("apm-server")
+var configRegistry = apmRegistry.NewRegistry("configuration")
```
axw, line 30: Can we do remove this "configuration" namespace? The libbeat state does not have such a namespace, and I'd like to be consistent if possible.

beater/telemetry_test.go
```diff
@@ -0,0 +1,92 @@
...
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package beater
+
+import (
+	"testing"
+
+	"gotest.tools/assert"
```
axw, line 23: stretchr/testify?

beater/telemetry.go
jalvz: consistent with what? Other beats derive registers from the state registry for telemetry too:

https://github.com/elastic/beats/blob/a6f180e273cb7de74a8ae5b704e85b076498a913/libbeat/cmd/instance/beat.go#L181
https://github.com/elastic/beats/blob/a6f180e273cb7de74a8ae5b704e85b076498a913/libbeat/monitoring/report/elasticsearch/elasticsearch.go#L357
etc

In this case, the reason is that if we later on add telemetry for other (dynamic) variables, we can separate them better and not have all counters mixed up at the same level.

beater/telemetry_test.go
jalvz: doh

beater/telemetry.go
jalvz: Well, Ill argue that this is using the config names. `camelCase` instead of `dots` to avoid mapping to object in ES.

beater/telemetry.go
jalvz: I don't think that *exactly one* use case sets any precedence :)
It is because ints might be easier to aggregate than bools, and give more granular information. 
IIUC, if a cluster deploys eg. 10 apm-servers, with ints we can see how many of those have rum enabled, but with bools we can only answer "is there at least one rum enabled?" or "do all have rum enabled?" questions.
Even if it turns out we only want to answer at-least-one questions, I don't see any drawbacks to ints - they just give more flexibility.

beater/telemetry.go
axw: > to avoid mapping to object in ES

... but why? Case in point: `management.enabled`

beater/telemetry.go
axw: Consistent with the structure that libbeat produces:

```
    "management" : {
        "enabled" : false
    },
    "queue" : {
        "name" : "mem"
    },
    "output" : {
        "name" : "elasticsearch"
    },
```

Why can't we do the same, like this?

```
    "apm-server": {
        "rum": {
            "enabled": true
        },
        "kibana": {
            "enabled": true,
            "version": "7.8.0" (for example)
        },
        ...
    }
```

beater/telemetry.go
axw: > I don't think that exactly one use case sets any precedence :)

In libbeat this is the only occurrence, but there are plenty of others in the telemetry schema: https://github.com/elastic/kibana/blob/master/x-pack/plugins/telemetry_collection_xpack/schema/xpack_plugins.json

(There are a couple of fields called "enabled" and "disabled" with type `long`; those are referring to number of rules enabled/disabled.)

In the event that there's just one other occurrence (for this or other changes), I would prefer that we do not take a second approach unless there's a very good reason to.

> It is because ints might be easier to aggregate than bools, and give more granular information.
>
> IIUC, if a cluster deploys eg. 10 apm-servers, with ints we can see how many of those have rum enabled, but with bools we can only answer "is there at least one rum enabled?" or "do all have rum enabled?" questions.

We can use a filter aggregation for this, with doc_count giving the same answer.

jalvz: > I would prefer that we do not take a second approach unless there's a very good reason to

I don't think using one or other type for a variable qualifies as a "second approach" - ints instead of bools is not like breaking the molds precisely. Still the reason was anyways easier aggregation.

Similarly, regarding your other (related) comments, I imagined that if we start adding more data, and index it like:
```
{ state: {
    configuration: {
        rum_enabled: 1,
        apikeys_enabled: 1,
        ...
   },
   events: {
   ...
   },
   agents: {
   ...
   },
}
```    

it will be more readable than an endless bag of:
```
{state: {
    key: {
        nested_key: {
             nested_key: value
        },
    }
{
```

without any order or structure. 
   

I don't find that to be less consistent (there isn't anything documented/prescribed I am ware of), but more trying to do what fits best.

Just answering questions, thou. I am not strongly opinionated about and will go trough suggestions

beater/telemetry.go
```diff
@@ -0,0 +1,88 @@
...
+	configMonitors.apiKeysEnabled.Set(apmCfg.APIKeyConfig.IsEnabled())
+	configMonitors.kibanaEnabled.Set(apmCfg.Kibana.Enabled)
+	configMonitors.jaegerHTTPEnabled.Set(apmCfg.JaegerConfig.HTTP.Enabled)
+	configMonitors.jaegerGRPCEnabled.Set(apmCfg.JaegerConfig.GRPC.Enabled)
+	configMonitors.sslEnabled.Set(apmCfg.TLS.IsEnabled())
+	configMonitors.pipelinesEnabled.Set(apmCfg.Register.Ingest.Pipeline.IsEnabled())
+	configMonitors.pipelinesOverwrite.Set(apmCfg.Register.Ingest.Pipeline.ShouldOverwrite())
+	indexManagementCfg, err := idxmgmt.NewIndexManagementConfig(info, rootCfg)
+	if err != nil {
+		return
```
axw, line 78: Sorry, I missed this before: please return the error and have the caller check it. If we get an error here something has gone badly wrong, and we should make some noise about it.

beater/telemetry_test.go
```diff
@@ -0,0 +1,92 @@
...
+	assert.Equal(t, configMonitors.ilmEnabled.Get(), true)
+	assert.Equal(t, configMonitors.ilmSetupEnabled.Get(), true)
+	assert.Equal(t, configMonitors.ilmSetupOverwrite.Get(), true)
+	assert.Equal(t, configMonitors.ilmSetupRequirePolicy.Get(), false)
+	assert.Equal(t, configMonitors.jaegerGRPCEnabled.Get(), true)
+	assert.Equal(t, configMonitors.jaegerHTTPEnabled.Get(), true)
+	assert.Equal(t, configMonitors.sslEnabled.Get(), false)
+}
+
+func resetCounters() {
```
axw, line 76: Nit: now that we're using booleans and setting the values unconditionally in `recordConfigs`, I think we could get rid of `resetCounters`.

beater/telemetry.go
```diff
@@ -0,0 +1,88 @@
...
+
+var configMonitors = &configTelemetry{
+	rumEnabled:                monitoring.NewBool(apmRegistry, "rum.enabled"),
+	apiKeysEnabled:            monitoring.NewBool(apmRegistry, "api_key.enabled"),
+	kibanaEnabled:             monitoring.NewBool(apmRegistry, "kibana.enabled"),
+	pipelinesEnabled:          monitoring.NewBool(apmRegistry, "register.ingest.pipeline.enabled"),
+	pipelinesOverwrite:        monitoring.NewBool(apmRegistry, "register.ingest.pipeline.overwrite"),
+	setupTemplateEnabled:      monitoring.NewBool(apmRegistry, "setup.template.enabled"),
+	setupTemplateOverwrite:    monitoring.NewBool(apmRegistry, "setup.template.overwrite"),
+	setupTemplateAppendFields: monitoring.NewBool(apmRegistry, "setup.template.append_files"),
```
axw, line 57:
```suggestion
	setupTemplateAppendFields: monitoring.NewBool(apmRegistry, "setup.template.append_fields"),
```

beater/telemetry.go
```diff
@@ -0,0 +1,88 @@
...
+	ilmEnabled:                monitoring.NewBool(apmRegistry, "ilm.enabled"),
+	ilmSetupEnabled:           monitoring.NewBool(apmRegistry, "ilm.setup.enabled"),
+	ilmSetupOverwrite:         monitoring.NewBool(apmRegistry, "ilm.setup.overwrite"),
+	ilmSetupRequirePolicy:     monitoring.NewBool(apmRegistry, "ilm.setup.require.policy"),
```
axw, line 61:
```suggestion
	ilmSetupRequirePolicy:     monitoring.NewBool(apmRegistry, "ilm.setup.require_policy"),
```

beater/telemetry.go
jalvz: > we should make some noise about it.

What do you mean with "noise"?
All other errors returned to `beater` result in the process stopped. Is telemetry recording critical? If so, why? It not, why do you want to stop the process on a non critical error?

If you want to propagate it, what other subroutines should be aborted because of telemetry not working?
If you just want to log it, why it needs to be returned? More so: that error is checked elsewhere (at the point where is actually critical), so in best case the error would be logged twice. 
With current logic the process will stop before even getting to that, so it would be dead code.

beater/telemetry_test.go
jalvz: Probably, but if a test messes with global state I prefer to make it more obvious, so I rather leave it like that if you don't mind.

beater/telemetry.go
simitt: I agree that the error should not be swallowed here, maybe logging it would be good enough though. How this error is handled here should not depend on whether or not some completely unrelated code part experiencing the same error would already stop the whole APM Server; it's completely independent from a logical point of view. Such implicit dependencies are dangerous (usually no tests would fail when anything changes and it wouldn't be anything obvious to change telemetry if some general program logic order changes).

axw: @simitt thank you, you explained my concern perfectly.

Silently swallowing errors is never OK in production code. Code changes, and assumptions made at the time consequently change.

@jalvz I see a few ways forward:

 - Bubble up the error. That _should_ never happen, but if it does happen (e.g. because of code changes) we'll find out about it quickly and have an opportunity to fix it.
 - Log the error and continue. Again, should never happen. If it does, maybe we'll find out about it if we see errors in the logs.
 - Refactor the code to eliminate the possibility of an error. Ideally we shouldn't be parsing the config twice, which is the cause of the issue here. Any errors here are not (currently) caused by telemetry failing, but due to config validation. If we get here and config is invalid, there's a logic error in the code that needs fixing.

Removing the possibility of an error would be ideal, but it doesn't look at all straightforward due to libbeat taking control of setting up index management.

beater/telemetry_test.go
axw: Fair enough. Maybe in the future we can get rid of the global (e.g. pass in a telemetry object), but this is fine for now.

axw (approved):
LGTM, thank!

PR #4084, opened 2020-08-19
systemtest: add tests for aggregation and sampling

axw:
## Motivation/summary

Migrate aggregation and sampling system tests to Go.

As part of this we introduce "approval" test support.
The implementation is a little different to the Python
version: the Python version ignores various dynamic
fields when computing the diff, whereas in this version
we replace dynamic fields before comparing, so the
approved files don't change in unrelated ways every
time there is an expected change, such as a new field.

## Checklist

~- [ ] documentation~
~- [ ] telemetry~

## How to test these changes

cd systemtest && go test -v

## Related issues

None.


systemtest/aggregation_test.go
```diff
@@ -0,0 +1,113 @@
...
+
+	var result estest.SearchResult
+	_, err = systemtest.Elasticsearch.Search("apm-*").WithQuery(estest.BoolQuery{
+		Filter: []interface{}{
+			estest.ExistsQuery{Field: "transaction.duration.histogram"},
+		},
+	}).Do(context.Background(), &result,
+		estest.WithCondition(result.Hits.NonEmptyCondition()),
+	)
+	require.NoError(t, err)
```
simitt, line 112: Should we also check the search result here?

systemtest/apmservertest/filter.go
```diff
@@ -0,0 +1,116 @@
...
+// EventMetadata holds event metadata.
+type EventMetadata struct {
+	System  *model.System
+	Process *model.Process
+	Service *model.Service
+	Labels  model.IfaceMap
+}
+
+// EventMetadata holds event metadata.
+type EventMetadataFilter interface {
```
simitt, line 46: No strong feelings, but is a `Filter` supposed to manipulate data? The way I know filters is that they either let data go through or not but don't change the input data. What do you think about naming this something like `metadataComposer` or `metadataOperator` instead?

systemtest/approvals.go
```diff
@@ -0,0 +1,171 @@
...
+
+// ApproveEvents compares the _source of the search hits with
+// the contents of the file in "systemtest/approvals/<name>".
+//
+// Dynamic fields (@timestamp, observer.id, etc.) are replaced
+// with a static string for comparison. Integration tests elsewhere
+// use canned data to test fields that we do not cover here.
+//
+// If the events differ, then the test will fail.
+func ApproveEvents(t testing.TB, name string, hits []estest.SearchHit) {
```
simitt, line 51: Why are you not using the existing approvals code?

systemtest/apmservertest/filter.go
axw: I used this name to be in line with filter APIs available in a few of the agents, e.g. [Node.js](https://www.elastic.co/guide/en/apm/agent/nodejs/current/agent-api.html#apm-add-filter), [Ruby](https://www.elastic.co/guide/en/apm/agent/ruby/current/api.html#api-agent-add-filter), [.NET](https://www.elastic.co/guide/en/apm/agent/dotnet/current/public-api.html#filter-api). I might otherwise have used "processor".

simitt: understood; the code comment suggests to move parts of this to the go agent and make it reusable, therefore aligning the naming makes sense.

systemtest/approvals.go
axw: I did start going down that route. I ended up creating a new thing in order to keep a clear separation between the systemtest module and the main apm-server module.

systemtest/aggregation_test.go
axw: Hmm, my intention was just to check that there is a result (non error implies the NonEmptyCondition was satisfied). However, it's probably a good idea to check that the result is correct in the shutdown case. I'll update it.

systemtest/approvals.go
axw: Discussed live: I'll look into creating another module in the repo which provides a shared package.

axw:
It's going to take a bit to shuffle the approvals code around, so I'll move the PR back into draft for now.

simitt (approved):

PR #4103, opened 2020-08-23
[docs] SSL documentation backport

eyalkoren:
Sorry for contaminating your master with multiple commits, didn't squash 🤦‍♂️ 
I cherry-picked the relevant commits and squashed before backporting.
Please verify it's OK and merge if it is.
Thanks.


axw:
Thanks!

axw:
Ahh hmm, wait. @bmorelli25 it looks like there's a few links dropped in the docs. Is it safe to backport this to 7.9 as-is, or are some changes needed?

eyalkoren:
I first backported the corresponding PR in the Java agent, so links are available in the current version docs, as @bmorelli25 suggested.
This should allow to use them in the current server docs. The build produces docs preview and those seem to work properly.

axw:
For current I think that makes sense - I'm not sure if it does for existing releases, where there may be blogs and what not linking in. I thought we didn't remove links from existing release docs as a policy, so I'll leave it to Brandon.

PR #4112, opened 2020-08-29
Introduce approvaltest module

axw:
## Motivation/summary

Move tests/approvals and tests/scripts/approvals.go into a new module,
github.com/elastic/apm-server/approvaltest. This new module will be used
by both apm-server (in integration tests) and the systemtest module,
enabling us to have those two modules not import from one another.

We also reduce the approvals API surface area to two functions: ApproveJSON,
and ApproveEventDocs. The ApproveJSON function can be used for approving
arbitrary JSON-encoded values, while ApproveEventDocs is specifically intended
for approving JSON-encoded Elasticsearch event documents.

To enable the API reduction, we introduce beatertest.EncodeEventDoc(s), which
encode events using the libbeat Elasticsearch output. In some tests we previously
used the libbeat JSON codec directly, which led to `@metadata` being included in
diffs. The output removes `@metadata`, hence it no longer features in our approval
diffs.

For dynamic fields like `profile.id`, we now replace the value with a known string
("dynamic") rather than ignoring the field completely. This minimises the diffs of
`.approved.json` files, and also allows us to detect added/removed fields.

## Checklist

~- [ ] documentation~
~- [ ] telemetry~

## How to test these changes

`make test`

## Related issues

See https://github.com/elastic/apm-server/pull/4084#discussion_r476218406


simitt (approved):
Nice, thank you for pulling out in a dedicated module. 

I can confirm that the bug, that led to not removing `.received` files, is also fixed with these improvements.

PR #4128, opened 2020-09-01
[7.x] Span metrics (#4077)

jalvz:
 - Span metrics (#4077)


axw:
Tested with 7.10-BC1, using go.elastic.co/apm@master. I ran the following program:

```go
package main

import (
        "context"
        "fmt"
        "net/http"

        "go.elastic.co/apm"
        "go.elastic.co/apm/module/apmhttp"
)

func main() {
        tracer := apm.DefaultTracer
        client := apmhttp.WrapClient(http.DefaultClient)

        for i := 0; i < 100; i++ {
                fmt.Println(i)

                req, _ := http.NewRequestWithContext(ctx, "GET", "http://testing.invalid", nil)
                resp, err := client.Do(req)
                if err == nil {
                        panic("expected error")
                }

                req, _ = http.NewRequestWithContext(ctx, "GET", "https://elastic.co", nil)
                resp, err = client.Do(req)
                if err != nil {
                        panic(err)
                }
                resp.Body.Close()

                tx.End()
        }
        tracer.Flush(nil)
}
```

- Confirmed that there are multiple metrics docs, one for each of testing.invalid and elastic.co
- The docs include `event.outcome`
- The `span.destination.service.response_time.count` field matches the expected number (100)

I also tested with ELASTIC_APM_TRANSACTION_SAMPLE_RATE=0.1, and confirmed that the count and sum are approximately the same.

PR #4142, opened 2020-09-03
Fix metadata.service.* fields requirements in json schema

### LINKED ISSUE #4141: Validation for metadata.service not working
Turns out that the current way of writing the [validation rules](https://github.com/elastic/apm-server/blob/master/docs/spec/metadata.json#L11:L20) for `metadata.service.*` fields in the json-schema is not working properly. 
Instead of writing e.g. ` "properties.agent.required": ["name", "version"]` one would need to expand to objects like 
```
            "properties": {
                "agent": {
                    "type": "object",
                    "required": [
                        "name",
                        "version"
                    ]
                }
             }

```
or probably even better, just copy & paste the whole `service`  definition and adapt the required fields. 

Unfortunately this has also been missed in the `package_tests` as tests for metadata are very sparse. 

Following fields should be changed to the key being `required` and value not being allowed to be `null`: 
* `metadata.service.name` (currently key required but value allowed to be `null`) , 
* `metadata.service.agent` (currently key required but value allowed to be `null`), * `metadata.service.agent.name`,`metadata.service.agent.version`,`metadata.service.runtime.name`,`metadata.serice.runtime.version`, `metadata.service.language.name` (currently key not required and value allowed to be `null`)

Technically this is a bug and we should fix it; but it would potentially break agents that might not send the required fields.

I suggest to create a fix and ask agents to run their tests against the fixed json schema to see if any current agent versions would break when fixing this. 

The bug exists since `7.0`.

simitt:
## Motivation/summary

Validations on metadata.service.* fields were not run correctly. This PR changes the schema to properly run validations for `v2` and `rum-v3`:
* `metadata.service.name`,`metadata.service.agent.name`: key required, value must be non-empty string , 
* `metadata.service.agent.version`: key must be sent, value must not be null
* `metadata.service.runtime.name`,`metadata.service.runtime.version`: keys must be sent, values must not be null if `metadata.service.runtime` is sent
* `metadata.service.language.name`: keys must be sent, values must not be null if `metadata.service.language` is sent

The PR also fixes an issue where `metadata.service.cloud.*` fields that are not required were not allowed to be sent as `null` values. 

## Checklist

~- [ ] documentation~
~- [ ] telemetry~
- [x] get elastic APM agent devs involved to check if the fix breaks any of the current agents.

## How to test these changes
Sending payloads where `metadata.service.agent.name`,`metadata.service.agent.version`,`metadata.service.runtime.name`,`metadata.serice.runtime.version`,`metadata.service.language.name` are missing or set to `null` and ensure payload is invalid. 

## Related issues
fixes #4141


axw (approved):
LGTM, but we should probably wait until we've got a resolution on https://github.com/elastic/apm-server/issues/4141#issuecomment-686452739

simitt:
Update: I added a `minLen=1` requirement for the `metadata.service.name` and `metadata.service.agent.name` - since sending an empty string should also be considered a bug.

simitt:
@elastic/apm-agent-devs as just discussed please check the requirements for the current agent versions and let me know if this bugfix causes issues, otherwise I'll merge by the end of the week.

axw (approved):
LGTM, but I'm wondering why we make `version` required but don't also require it to be non-empty. Seems like all fields should be either optional or required and non-empty.

simitt:
I didn't want to introduce too many changes with this as it will be deployed as bugfix; the reason I went for adding `minLen` to `service.name` and `service.agent.name` is that they are more crucial to the APM solution.

eyalkoren:
Java agent should be OK with this change.
* `metadata.service.name` - there should be fallback always, including when setting an empty string.
* `metadata.service.agent.name`: hardcoded "java" 
* `metadata.service.agent.version`: either the valid version or "unknown"
* `metadata.service.runtime.name`: hardcoded "Java"
* `metadata.service.runtime.version`: the value of the `java.version` system property, should never be null
* `metadata.service.language.name`: hardcoded "Java"

As far as I can see, it is the case since GA.
Since the only configurable option is `service.name`, I think it shouldn't break anything.

SergeyKleyman (approved):

vigneshshanmugam (approved):
Looks good from RUM agent side.

PR #4143, opened 2020-09-03
feat: limit sourcemap line length

clyfish:
## Motivation/summary

My coordinating node got OOM when query documents with huge lone lines in the sourcemap, so I want limit the line length of the sourcemap.

## Checklist

- [ ] documentation
- [ ] telemetry

### DIFF
--- a/sourcemap/mapper.go
+++ b/sourcemap/mapper.go
@@ -24,6 +24,7 @@ import (
 )
 
 const sourcemapContentSnippetSize = 5
+const sourcemapLineLimit = 256
 
 // Map sourcemapping for given line and column and return values after sourcemapping
 func Map(mapper *sourcemap.Consumer, lineno, colno int) (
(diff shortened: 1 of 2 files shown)


sourcemap/mapper.go
```diff
@@ -24,6 +24,7 @@ import (
 )
 
 const sourcemapContentSnippetSize = 5
+const sourcemapLineLimit = 256
```
axw, line 27: I think this would be classified as a breaking change. To avoid that, we would need to have it configurable and unlimited by default.

axw (changes requested):
Thanks for the PR!

I agree in principle that we should provide a way of limiting the line length, but it should be configurable (at least initially) to avoid breaking existing users.

I can suggest two options to proceed:
 1. Add configuration, e.g. `apm-server.rum.source_mapping.line_limit`. This would need to be unlimited by default to maintain backwards compatibility.
 2. Abort this change, create a feature request issue, and use an [Ingest pipeline](https://www.elastic.co/guide/en/elasticsearch/reference/current/ingest.html) in the mean time

sourcemap/mapper.go
clyfish: https://github.com/elastic/apm-server/issues/4149
It's too hard for me, so I've created a feature request issue.

axw: OK, thanks all the same. If you have any trouble setting up an Ingest pipeline as an alternative, please open a topic at https://discuss.elastic.co/c/apm and someone will try to help out.

clyfish: Thank you, I've deployed some coordinating nodes to avoid crash the data nodes.
I think the OOM happend in the gather phase.

PR #4159, opened 2020-09-07
[7.x] tests: fix docker-system-tests (#4123)

axw:
 - tests: fix docker-system-tests (#4123)


axw:
Oops, too hasty. Needs approvaltest first.

PR #4267, opened 2020-10-06
Update to elastic/beats@bad38abf9473 and bump version to 7.11.0

jalvz:
## Checklist

- [ ] I have signed the [Contributor License Agreement](https://www.elastic.co/contributor-agreement/).
- [ ] I have updated [CHANGELOG.asciidoc](https://github.com/elastic/apm-server/blob/master/CHANGELOG.asciidoc)

I have considered changes for:
- [ ] documentation
- [ ] logging (add log lines, choose appropriate log selector, etc.)
- [ ] metrics and monitoring (create issue for Kibana team to add metrics to visualizations, e.g. [Kibana#44001](https://github.com/elastic/kibana/issues/44001))
- [ ] automated tests (add tests for the code changes, all [**unit** tests](https://github.com/elastic/apm-server/blob/master/TESTING.md) pass locally)
- [ ] telemetry
- [ ] Elasticsearch Service (https://cloud.elastic.co)
- [ ] Elastic Cloud Enterprise (https://www.elastic.co/products/ece)
- [ ] Elastic Cloud on Kubernetes (https://www.elastic.co/elastic-cloud-kubernetes)

PR #4272, opened 2020-10-06
docs: Update field ref with ECS icon

bmorelli25:
## Motivation/summary

I updated `generate_fields_docs.py` in https://github.com/elastic/beats/pull/21359 to include an ECS icon next to ECS fields in the field reference documentation. This PR makes use of that new script to create APM Server's field reference:

* Updates to elastic/beats@fac0347
* Uses `make update` to update the field reference

~Unfortunately, I can't get `make update` to make use of that change. I still have to use a local clone of Beats to get things to work:~

~go mod edit -replace=github.com/elastic/beats/v7=../beats~

~Is this because APM Server needs to update to a later version of Beats before we can make use of this change? If so, does it make more sense to wait on this PR until after that happens?~

## Related issues

Closes elastic/apm#276.
Relates to elastic/beats#20729.


graphaelli:
> Is this because APM Server needs to update to a later version of Beats before we can make use of this change? If so, does it make more sense to wait on this PR until after that happens?

Sounds right.  You could bundle a `make update-beats` with this pr to test the theory.

bmorelli25:
> Sounds right. You could bundle a make update-beats with this pr to test the theory.

Neat, thanks. That worked!

axw (approved):
Nice, I like it!

bmorelli25:
It looks like Beats has been updated twice since last night: First in https://github.com/elastic/apm-server/pull/4266, and then again in https://github.com/elastic/apm-server/pull/4267. I will remove that from this PR.

PR #4282, opened 2020-10-07
[7.10] use beats-patched badger

graphaelli:
Created with 

```
go mod edit -replace github.com/dgraph-io/badger/v2=github.com/elastic/badger/v2@v2.2007.2-beats
go build
make notice
```

relates to https://github.com/elastic/beats/pull/21613

PR #4283, opened 2020-10-07
use beats-patched badger

graphaelli:
Same as #4282

```
go mod edit -replace github.com/dgraph-io/badger/v2=github.com/elastic/badger/v2@v2.2007.2-beats
go build
make notice
```

Had I done this in the right order this pr would have come first and that one would have been a backport.

relates to elastic/beats#21613


axw:
Failure appears unrelated, merging.

PR #4286, opened 2020-10-08
decoder v2: redact input information from decode error

simitt:
Avoid logging any input information on decode error.

## Motivation/summary

APM Server should not log APM agent data on decode errors, it should log as much information as possible to identify the root cause for the failure though. 

https://github.com/json-iterator/go adds input information as context to the error message, which should not be logged. Error information is added hierarchically, but errors are not wrapped, so there is no good way of unwrapping and only keeping the _outer part_ of the error message. 
A possibility I looked into was to use `jsoniter.RegisterTypeDecoderFunc(string, func(ptr unsafe.Pointer, iter *Iterator)` for registering a decoder wrapper for every custom struct, reusing the default json-iter struct decoder, while being able to access and manipulate the `iter.Error` after decoding. Unfortunately I did not find a good way to do that, as the default decoders are not exported.

Every error is registered via calling the `iterator.ReportError(operation string, msg string)` method, which always wraps the error information in a fixed error message format. This PR removes any input information by using a regexp to match against this error wrapper message. This is a bit fragile as it depends on some internas of the `json-iterator`, but at least it is only one place where the information is set. Changes in the `json-iterator` behavior should be covered by our approval tests containing error messages. 

## Checklist

`- [ ] documentation~
~- [ ] telemetry~

## How to test these changes
start APM Server and send documents leading to a decoding error; check the server logs

## Related issues
#3551

### DIFF
--- a/model/modeldecoder/v2/decoder.go
+++ b/model/modeldecoder/v2/decoder.go
@@ -42,11 +42,11 @@ type DecodeError struct {
 	err error
 }
 
-var jsoniterErrRegexp = regexp.MustCompile(`expect.* but found .*error found in .* bigger context.*`)
+var jsoniterErrRegexp = regexp.MustCompile(`but found .*error found in .* bigger context.*`)
 
 func newDecodeErrFromJSONIter(err error) DecodeError {
 	if jsoniterErrRegexp.MatchString(err.Error()) {
-		err = errors.New(jsoniterErrRegexp.ReplaceAllString(err.Error(), "invalid"))
+		err = errors.New(jsoniterErrRegexp.ReplaceAllString(err.Error(), ""))
 	}
 	return DecodeError{err}
 }
--- a/model/modeldecoder/v2/decoder.go
+++ b/model/modeldecoder/v2/decoder.go
@@ -42,7 +42,7 @@ type DecodeError struct {
 	err error
 }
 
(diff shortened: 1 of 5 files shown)


simitt:

model/modeldecoder/v2/decoder.go
```diff
@@ -41,6 +42,15 @@ type DecodeError struct {
 	err error
 }
 
+var jsoniterErrRegexp = regexp.MustCompile(`expect.* but found .*error found in .* bigger context.*`)
```
axw, line 45: What do you think about keeping the "expect <...>" bit intact? That could be helpful to understand why the field is invalid, and is based on the target field type rather than the input.

simitt: yes, sure we can keep that

model/modeldecoder/v2/decoder.go
```diff
@@ -41,6 +42,15 @@ type DecodeError struct {
 	err error
 }
 
+var jsoniterErrRegexp = regexp.MustCompile(`but found .*error found in .* bigger context.*`)
```
axw, line 45:
```suggestion
var jsoniterErrRegexp = regexp.MustCompile(` but found .*error found in .* bigger context.*`)
```

axw (approved):

simitt:

PR #4308, opened 2020-10-12
[CI] run everything for branches/tags

v1v:
## Motivation/summary

Branches and tags will run all the stages even when the merged changes are only related to docs


simitt (approved):

v1v:
Steps failures are not relevant since the retry/sleep happened

Issue #4185, opened 2020-09-10
Feature: tail-based sampling


axw:
Add support to for tail-based sampling: sampling based on properties of the trace known only after the trace completes, such as the root transaction duration and outcome. This will enable users to retain traces based on how interesting they are, i.e. preferring to keep slower traces, or traces with a failure outcome.

The solution we plan to implement will be primarily based in apm-server, in order to avoid network transfer costs and adding significant load on Elasticsearch. Servers will individually reservoir-sample root transactions, and synchronise their sampling decisions in order to index any other events that are part of sampled traces.

- [x] implement local event storage elastic/apm-server#4108
- [x] implement publish/subscribe for synchronising sampled trace IDs elastic/apm-server#4139
- [x] implement reservoir sampling elastic/apm-server#4182
- [x] implement dynamic trace group sampling policies (sample trace group X with percentage Y, ...) elastic/apm-server#4320
- [ ] implement an efficient codec for storing events elastic/apm-server#4120
- [x] monitoring, logging #4346 
- [x] telemetry https://github.com/elastic/apm-server/pull/4360
- [ ] documentation

axw:
Closing this meta issue, issues for remaining things have been opened and can be tracked independently.

PR #4320, opened 2020-10-14
sampling: introduce policies

axw:
## Motivation/summary

Revise tail-sampling configuration, removing the
default sampling rate and introducing configurable
sampling policies.

Each policy defines criteria to match root transactions,
and a sampling rate. Policies are evaluated in the
order specified, comparing the specified attributes
against incoming root transactions. In the future policies
could also match non-root transactions or even spans,
at some additional cost.

Attributes that can be used for matching are:
 - service.name
 - service.environment
 - trace.name (root transaction name)
 - trace.outcome (root transaction outcome)

Policies may omit any of the attributes. With the
exception of the service name, omitting an attribute
has the effect of grouping all matching traces for
the purposes of tail-sampling. For example, if only
service name is specified, then all traces originating
at that service will be sampled uniformly.

If service name is not specified, then that policy may
match any service. There is a currently hard-coded
limit of 1000 service names that may be matched in
this way. If the limit is reached, traces from any additional
services will not be sampled.

If a root transaction matches none of the policies, then
it will be dropped and the trace will not be sampled.

## Checklist

~- [ ] documentation~ (later)
~- [ ] metrics and monitoring (create issue for Kibana team to add metrics to visualizations, e.g. [Kibana#44001](https://github.com/elastic/kibana/issues/44001))~ (will follow up with monitoring, e.g. number of active sampling groups)
~- [ ] telemetry~ (will follow up with reporting whether it's enabled, and number of policies)
- [x] Elastic Cloud Enterprise (https://www.elastic.co/products/ece)
- [x] Elastic Cloud on Kubernetes (https://www.elastic.co/elastic-cloud-kubernetes)

## How to test these changes

1. Run apm-server with the following config:

```yaml
apm-server:
  sampling:
    tail:
      enabled: true
      interval: 10s
      policies:
      - trace.name: many
        sample_rate: 0.5
      - trace.name: few
        sample_rate: 0.1
```

2. Send 1000 transactions with the name "many", and 1000 with the name "few".
3. Wait ~10 seconds
4. Check that there are 500 "many" transactions, and 100 "few" transactions, indexed into `apm-*-transaction`:

```
GET /apm-8.0.0-transaction/_search
{
  "size": 0,
  "aggs": {
    "by_name": {
      "terms": {
        "field": "transaction.name",
        "size": 100
      }
    }
  }
}

{
  "took" : 0,
  "timed_out" : false,
  "_shards" : {
    "total" : 1,
    "successful" : 1,
    "skipped" : 0,
    "failed" : 0
  },
  "hits" : {
    "total" : {
      "value" : 600,
      "relation" : "eq"
    },
    "max_score" : null,
    "hits" : [ ]
  },
  "aggregations" : {
    "by_name" : {
      "doc_count_error_upper_bound" : 0,
      "sum_other_doc_count" : 0,
      "buckets" : [
        {
          "key" : "many",
          "doc_count" : 500
        },
        {
          "key" : "few",
          "doc_count" : 100
        }
      ]
    }
  }
}
```

## Related issues

Part of #4185


simitt (commented):
Didn't go into every detail, but from a high-level review LGTM

beater/config/sampling.go
```diff
@@ -90,7 +108,6 @@ func defaultTailSamplingConfig() TailSamplingConfig {
 	return TailSamplingConfig{
 		Enabled:               false,
 		ESConfig:              elasticsearch.DefaultConfig(),
-		DefaultSampleRate:     0.5,
```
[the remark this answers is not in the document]
axw: Good question! I left out a default because I'm not sure that there's a good one-size-fits-most default. Our agents don't sample anything by default, so it seems reasonable to me to require users to define policies that make sense for their application.

x-pack/apm-server/sampling/config.go
```diff
@@ -36,14 +36,16 @@ type LocalSamplingConfig struct {
...
-	// group will dropped.
-	MaxTraceGroups int
+	// Once MaxDynamicServices is reached, root transactions from a service that
+	// does not have an explicit policy defined may be dropped.
+	MaxDynamicServices int
 
-	// DefaultSampleRate is the default sample rate to assign to new trace groups.
-	DefaultSampleRate float64
+	// Policies holds local tail-sampling policies. Policies are matched in the
+	// order provided. Policies should therefore be ordered from most to least
```
jalvz, line 46: can/should we follow up with a task to order them ourselves from most to least specific, and not bother users with ordering?

x-pack/apm-server/sampling/groups.go
```diff
@@ -16,41 +16,99 @@ import (
...
-		maxGroups:               maxGroups,
-		groups:                  make(map[traceGroupKey]*traceGroup),
+	groups := &traceGroups{
+		ingestRateDecayFactor: ingestRateDecayFactor,
+		maxDynamicServices:    maxDynamicServices,
+		staticGroups:          make(map[string]serviceGroups),
+		dynamicGroups:         make(map[string]serviceGroups),
+	}
+	for _, policy := range policies {
+		if policy.ServiceName == "" {
```
jalvz, line 93: why empty service name is a special case, can't this logic be added to `match`?

x-pack/apm-server/sampling/config.go
axw: Great question!

I started out trying to make policies order-insensitive by defining an partial ordering (e.g. compare service name, then trace name, etc.).

In the end, particularly because everything is optional, I felt it would be too complex and subtle for users to understand how the policies would be applied, and it would be more straightforward to require explicit ordering. My expectation is that typical users will have a handful of policies, so ordering them would not be onerous.

Final point: we intend to develop a tail-sampling policy UI, which could take care of automatically ordering policies and could give some visual indications of precedence to reduce the subtlety.

In summary, I would like to wait and see.

x-pack/apm-server/sampling/groups.go
axw: That would be technically possible, but this approach:

enables the invariant that a serviceGroup (serviceGroupKey) is always associated with a specific service: either one defined statically (in the policy criteria) or dynamically (no service defined in criteria). This invariant is useful to ensure we can't group multiple services for the purposes of sampling (particularly the weighted random sampling bit), which would rarely if ever make much sense.
simplifies keeping track of how many "dynamic service groups" have been created, i.e. unique groups matching a policy without an explicit service name specified. this is used to enforce the max groups config.

beater/config/sampling.go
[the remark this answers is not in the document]
simitt: fair point

x-pack/apm-server/sampling/config.go
jalvz: Fair enough :+1:

x-pack/apm-server/sampling/groups.go
jalvz: Ok, maybe put that on a comment :)

jalvz (approved):

simitt (approved):

simitt:
Tested with following payload that is sent 1000 times per test run:
```
{"metadata":{"service":{"name":"simple","agent":{"version":"1.10.0","name":"java"}}}}
{"span":{"type":"external","id":"1234567890aaaade","trace_id":"axz","parent_id":"spanparent","name":"GET users-authenticated","duration":3.781912,"start":100}}
{"transaction":{"name":"something","type":"http","id":"123455555","trace_id":"aaaa","span_count":{"started":17,"dropped":0},"duration":32.592981}}
{"error": {"id": "xFoaabb123FFFFFF", "timestamp": 1533826745999000,"log": {"message": "no user found", "stacktrace": [{"classname": "User::Special"}]}}}
{"span":{"type":"external","id":"123-with-transaction","transaction_id":"4341a8e0df1906ecbfa9","trace_id":"0acd456789abcdef0123456789abcdef","parent_id":"spanparent","name":"GET users-authenticated","duration":3.781912,"start":100}}
{"transaction":{"name":"mytransaction","type":"http","id":"4341a8e0df1906ecbfa9","trace_id":"0acd456789abcdef0123456789abcdef","span_count":{"started":17,"dropped":0},"duration":32.592981}}
```

(0) Sampling disabled
```
apm-server:
  sampling:
    tail:
      enabled: false
```

  | #errors | #spans | #spans (belong to transaction mytransaction) | # transaction (something) | # transaction (mytransaction)
-- | -- | -- | -- | -- | --
Expected | 1000 | 1000 | 1000 | 1000 | 1000
Observed | 1000 | 1000 | 1000 | 1000 | 1000

(1) Sampling enabled with `trace.name` policy
```
apm-server:
  sampling:
    tail:
      enabled: false
      interval: 10s
      policies:
      - trace.name: mytransaction
        sample_rate: 0.5
```


  | #errors | #spans | #spans (belong to transaction mytransaction) | # transaction (something) | # transaction (mytransaction)
-- | -- | -- | -- | -- | --
Expected | 1000 | ? | 500 | 0 | 500
Actual | 0 | 0 | 0 | 0 | 0



Receiving following APM Server response:
```
{
  "accepted": 0,
  "errors": [
    {
      "message": "no matching policy"
    }
  ]
}
```

(2) Sampling enabled with `service.name` policy
```
apm-server:
  sampling:
    tail:
      enabled: false
      interval: 10s
      policies:
      - service.name: simple
        sample_rate: 0.5
```


  | #errors | #spans | #spans (belong to transaction mytransaction) | # transaction (something) | # transaction (mytransaction)
-- | -- | -- | -- | -- | --
Expected | 1000 | ? | 500 | 500 | 500
Actual | 1000 | 0 | 1000 | 1000 | 1000

Haven't tested any further (other config options). @axw could you have a look whether I misconfigured something, I'd actually be surprised if the whole payload was rejected when no policy is met, and the sampling rate wasn't respected as we do have tests for this.

axw:
> I'd actually be surprised if the whole payload was rejected when no policy is met

I think that's fair.

We don't want to be creating a document for every sampled trace by default though. When we introduce a UI to configure this, we should enforce a catch-all policy so this cannot happen in practice. i.e. require the user to specify a default sample rate, and set it to something relatively low by default (say 10%). I've opened https://github.com/elastic/apm-server/issues/4583

Ideally we would also push config down to the APM Agents so they can determine whether or not a trace can skip tail-sampling altogether.

> (2) Sampling enabled with service.name policy
> 
> ```
> apm-server:
>   sampling:
>     tail:
>       enabled: false
>       interval: 10s
>       policies:
>       - service.name: simple
>         sample_rate: 0.5
> ```

Did you really have `enabled: false`, or is that a typo?

axw:
I've been doing some testing and there appears to be some duplication going on.

```
GET /apm-7.11.0-transaction/_search
{
  "size": 0,
  "query": {
    "term": {
      "service.name": {
        "value": "service_1"
      }
    }
  },
  "aggs": {
    "transactions": {
      "cardinality": {
        "field": "transaction.id"
      }
    }
  }
}

{
  "took" : 9,
  "timed_out" : false,
  "_shards" : {
    "total" : 1,
    "successful" : 1,
    "skipped" : 0,
    "failed" : 0
  },
  "hits" : {
    "total" : {
      "value" : 200,
      "relation" : "eq"
    },
    "max_score" : null,
    "hits" : [ ]
  },
  "aggregations" : {
    "transactions" : {
      "value" : 100
    }
  }
}
```

Notice there's 200 transaction docs, but only 100 unique transaction IDs. This might explain what you've seen.

axw:
I've opened https://github.com/elastic/apm-server/issues/4584 to track that issue.

axw:
@simitt sorry, I forgot to mention in the instructions that it's currently necessary to manually create the index/template for `apm-sampled-traces`. See https://github.com/elastic/apm-server/issues/4584#issuecomment-750746095

simitt:
> Did you really have enabled: false, or is that a typo?

I had it enabled - just retested to ensure; updated the comment now.

simitt:
> @simitt sorry, I forgot to mention in the instructions that it's currently necessary to manually create the index/template for apm-sampled-traces. 

Actually I do have the template in there, but I am getting the same results.

simitt:
As @axw pointed out in an offline conversation my testing method was the problem as I was sending the same root traces multiple times. Switched to a different [testing setup](https://gist.github.com/simitt/7ece7a741ed8fb35c59447659f88cbba) now. 
* Sampling with only a `sample_rate` specified works as expected. 
* Sampling with a `trace.name` and a `sample_rate` specified works as expected. 
* Accidentially specified `transaction.name` instead of `trace.name` -> non-deterministic amount of transactions was indexed. It is an invalid policy configuration so it's ok that it doesn't work, but a future config validation would be great. 
* Sampling with a `trace.name` and a `sample_rate` specified works as expected if events with the same `trace.name` are sent. Sending events with different `trace.name` attributes leads to non-deterministic results. 
* Sampling with two policies - one specifying a `trace.name` the other one a `service.name` and sending events that match both policies. I'm not certain what the expected outcome should be here, but the results looked random. This might not be allowed in the first place; again a config validation would be nice long term.

axw:
> Sampling with two policies - one specifying a trace.name the other one a service.name and sending events that match both policies. I'm not certain what the expected outcome should be here, but the results looked random. This might not be allowed in the first place; again a config validation would be nice long term.

The policies should be matched in the order specified -- first matching one applies. It certainly should not be random. Do you still have the test app, and can you test that again to check?

simitt:
Tested again with different configs. The transaction events were created via a small go app using the go APM agent, see: https://gist.github.com/simitt/7ece7a741ed8fb35c59447659f88cbba#file-testapp-go

## Following test instructions
works as expected

<details>
  <summary>Test Details</summary>

Config:
```
apm-server:
  sampling:
    tail:
      enabled: true
      interval: 10s
      policies:
      - trace.name: no_spans
        sample_rate: 0.5
      - trace.name: with_spans
        sample_rate: 0.1
```

Outcome:
```  
  "aggregations" : {
    "transactions" : {
      "doc_count_error_upper_bound" : 0,
      "sum_other_doc_count" : 0,
      "buckets" : [
        {
          "key" : "no_spans",
          "doc_count" : 500
        },
        {
          "key" : "with_spans",
          "doc_count" : 100
        }
      ]
    }
  }
```
</details>

## Multiple policies - policy order:
The policy specifying a `service.name` seems to overrule the policy specifying a `trace.name`.

<details>
  <summary>Test Details</summary>
Sending 1000 transactions with `transaction.name: no_spans`; all transactions matching `service.name: goagent`.

Config:
```
apm-server:
  sampling:
    tail:
      enabled: true
      interval: 10s
      policies:
      - trace.name: no_spans
        sample_rate: 0.5
      - service.name: goagent
        sample_rate: 0.1
```

Outcome:
```
  "aggregations" : {
    "transactions" : {
      "doc_count_error_upper_bound" : 0,
      "sum_other_doc_count" : 0,
      "buckets" : [
        {
          "key" : "no_spans",
          "doc_count" : 100
        }
      ]
    }
  }
```

Config:
```
apm-server:
  sampling:
    tail:
      enabled: true
      interval: 10s
      policies:
      - service.name: goagent
        sample_rate: 0.1
      - trace.name: no_spans
        sample_rate: 0.5
```

Outcome:
```
  "aggregations" : {
    "transactions" : {
      "doc_count_error_upper_bound" : 0,
      "sum_other_doc_count" : 0,
      "buckets" : [
        {
          "key" : "no_spans",
          "doc_count" : 100
        }
      ]
    }
  }
```
</details>



## Multiple policies - different events:
Generally seems to work fine, but occasionally sampling was not precise. 

<details>
  <summary>Test Details</summary>
Sending 1000 transactions with transaction.name: `no_spans` and 1000 transactions with `transaction.name: with_spans`; all transactions matching `service.name: goagent`.

Config:
```
apm-server:
  sampling:
    tail:
      enabled: true
      interval: 10s
      policies:
      - service.name: goagent
        sample_rate: 0.1
      - trace.name: no_spans
        sample_rate: 0.5
```

Outcome:
```  
"aggregations" : {
    "transactions" : {
      "doc_count_error_upper_bound" : 0,
      "sum_other_doc_count" : 0,
      "buckets" : [
        {
          "key" : "no_spans",
          "doc_count" : 100
        },
        {
          "key" : "with_spans",
          "doc_count" : 100
        }
      ]
    }
  }
```

```
  "aggregations" : {
    "transactions" : {
      "doc_count_error_upper_bound" : 0,
      "sum_other_doc_count" : 0,
      "buckets" : [
        {
          "key" : "with_spans",
          "doc_count" : 103
        },
        {
          "key" : "no_spans",
          "doc_count" : 97
        }
      ]
    }
  }
```
```
  "aggregations" : {
    "transactions" : {
      "doc_count_error_upper_bound" : 0,
      "sum_other_doc_count" : 0,
      "buckets" : [
        {
          "key" : "with_spans",
          "doc_count" : 113
        },
        {
          "key" : "no_spans",
          "doc_count" : 87
        }
      ]
    }
  }
```
</details>

simitt:
Tested with 2 APM Servers and interaction between them worked fine.


<details>
  <summary>Test Details</summary>
Sending 1000 transactions with transaction.name: `no_spans` and 1000 transactions with `transaction.name: with_spans`; all transactions matching `service.name: goagent`. 

Config:
```
apm-server:
  sampling:
    tail:
      enabled: true
      interval: 10s
      policies:
      - trace.name: with_spans
        sample_rate: 0.1
      - trace.name: no_spans
        sample_rate: 0.5
```

Outcome:
```  
  "aggregations" : {
    "transactions" : {
      "doc_count_error_upper_bound" : 0,
      "sum_other_doc_count" : 0,
      "buckets" : [
        {
          "key" : "no_spans",
          "doc_count" : 501
        },
        {
          "key" : "with_spans",
          "doc_count" : 100
        }
      ]
    }
  }
```
</details>

axw:
> The policy specifying a service.name seems to overrule the policy specifying a trace.name.

Good catch. This is a bug, I'll open an issue.

axw:
https://github.com/elastic/apm-server/issues/4616 -- assigned to 7.12.

PR #4324, opened 2020-10-15
[7.10] fix: ensure ES index names are lowercase for ILM (#4322)

simitt:
Backports the following commits to 7.10:
 - fix: ensure ES index names are lowercase for ILM (#4322)


simitt:
seems to be an unrelated test failure

PR #4329, opened 2020-10-19
fix: nest metricset's span and transaction attributes

### LINKED ISSUE #4326: Metricset spec should have transaction and span properties
**APM Server version** (`apm-server version`): master branch

the metricset.json spec includes a `$ref` to `span_type.json` and `transaction_type.json` in `allOf`: 

https://github.com/elastic/apm-server/blob/24ece9434ca500109fc03c2401d12e8707a13b01/docs/spec/metricsets/metricset.json#L5-L10

Both span_type.json and transaction_type,json contain a `type` property:

https://github.com/elastic/apm-server/blob/24ece9434ca500109fc03c2401d12e8707a13b01/docs/spec/span_type.json#L5-L10

https://github.com/elastic/apm-server/blob/24ece9434ca500109fc03c2401d12e8707a13b01/docs/spec/transaction_type.json#L5-L10

From what I understand, JSON schema is not clear about what should happen in this scenario, and it's possible that some JSON schema validator implementations may reject this as an invalid schema.

Issue above aside, chatting to @axw, it sounds like the schema may not be correct here anyway - the properties related to transaction should be nested under a `"transaction"` property, and those related to span, under a `"span"` property, looking at the `Metricset` structure

https://github.com/elastic/apm-server/blob/acd41cc94bc38e0f04cf83abaefb73f08590145e/model/metricset.go#L48-L81

simitt:
## Motivation/summary
Fix span and transaction references in metricsets:
* `subtype` and `type` should have been nested under `span`
* `name` and `type` should have been nested under `transaction`

For easier readability include the properties directly and get rid of the
shared spec files.

## Checklist

~- [ ] documentation~
~- [ ] telemetry~

## How to test these changes
Send metricset docs with (in-)valid transaction and span events against the server. 

## Related issues
fixes #4326


docs/spec/metricsets/metricset.json
```diff
@@ -22,11 +20,61 @@
                     },
                     "additionalProperties": false
                 },
+                "span": {
+                    "type": [
+                        "object",
+                        "null"
+                    ],
+                    "properties": {
+                        "sub_type": {
```
axw, line 29:
```suggestion
                        "subtype": {
```

docs/spec/metricsets/metricset.json
```diff
@@ -22,11 +20,61 @@
...
+                            "description": "A further sub-division of the type (e.g. postgresql, elasticsearch)",
+                            "type": [
+                                "string",
+                                "null"
+                            ],
+                            "maxLength": 1024
+                        },
+                        "type": {
+                            "description": "Keyword of specific relevance in the service's domain (eg: 'db.postgresql.query', 'template.erb', etc)",
```
axw, line 38:
```suggestion
                            "description": "Keyword of specific relevance in the service's domain (eg: 'db', 'template', etc)",
```

docs/spec/metricsets/metricset.json
axw: Maybe we should continue to use span_subtype.json etc. though, but nested under the `span` object? `span.json` uses it; I think it would be great if we could maintain a single definition.

axw (changes requested):
Looks good, but I think I'd prefer if we continued using the sub-schema files, or don't use them anywhere.

docs/spec/metricsets/metricset.json
simitt: The main reason to get rid of the sub-schema json files here was easier readability. The `span_type.json` and `transaction_type.json` defined `"type": "string"`, which is aligned with requirements for `span.json` and `transaction.json` but when reusing for `metricsets` the type needs to also include `null` as the fields are optional. Referencing the files and changing the types of the deep nested attributes can easily lead to errors (as I've seen on `service.*` related fields just recently) and makes the schemar harder to read. Since all of this is subject to change soon anyways, I'd go with some duplication and easier readability right now.

axw: Solid points. Sounds good, thanks!

axw (approved):

axw:
Tested with 7.11.0 BC1. Sending a metricset with either an invalid `transaction.type` or invalid `span.type` causes an error. I achieved this by sending `testdata/intake-v2/metricsets.json`, modified to set those fields to numeric values.

Example error:

```
{
  "accepted": 2,
  "errors": [
    {
      "message": "decode error: data read error: v2.metricsetRoot.Metricset: v2.metricset.Span: v2.metricsetSpanRef.Type: ReadString: expects \" or n,",
      "document": "{\"metricset\": { \"samples\": { \"transaction.breakdown.count\":{\"value\":12}, \"transaction.duration.sum.us\":{\"value\":12}, \"transaction.duration.count\":{\"value\":2}, \"transaction.self_time.sum.us\":{\"value\":10}, \"transaction.self_time.count\":{\"value\":2}, \"span.self_time.count\":{\"value\":1},\"span.self_time.sum.us\":{\"value\":633.288}, \"byte_counter\": { \"value\": 1 }, \"short_counter\": { \"value\": 227 }, \"integer_gauge\": { \"value\": 42767 }, \"long_gauge\": { \"value\": 3147483648 }, \"float_gauge\": { \"value\": 9.16 }, \"double_gauge\": { \"value\": 3.141592653589793 }, \"dotted.float.gauge\": { \"value\": 6.12 }, \"negative.d.o.t.t.e.d\": { \"value\": -1022 } }, \"tags\": { \"some\": \"abc\", \"code\": 200, \"success\": true }, \"transaction\":{\"type\":\"foo\",\"name\":\"GET /\"},\"span\":{\"type\":123,\"subtype\":\"mysql\"},\"timestamp\": 1496170422281000 }}"
    }
  ]
}
```

Issue #4031, opened 2020-07-29
Document transaction metrics


axw:
We have added support in APM Server for:
 - aggregating transaction duration metrics (https://github.com/elastic/apm-server/issues/3485)
 - dropping unsampled transactions (https://github.com/elastic/apm-server/pull/3702)

These can be controlled with the following config options:

- `apm-server.sampling.keep_unsampled: true|false` (currently defaults to `true`, for backwards compatibility)
  - Controls whether or not non-sampled transactions documents are recorded
- `apm-server.aggregation.transactions.enabled: true|false` (currently defaults to `false`)
  - Controls whether or not metrics are gathered and published
- `apm-server.aggregation.transactions.interval: <duration>` (currently defaults to `1m`)
  - Controls how frequently the metrics are published
- `apm-server.aggregation.transactions.max_groups: [1,+Inf]` (currently defaults to 1000)
  - Maximum number of transaction groups to keep track of. Once this is exceeded, we devolve into recording a metrics document for each transaction not in one of the transaction groups being tracked.
- `apm-server.aggregation.transactions.hdrhistogram_significant_figures: [1,5]` (currently defaults to 2)
  - Transaction durations are aggregated using [HDRHistogram](http://hdrhistogram.org). This option controls the number of significant figures to maintain for recorded values, which controls relative error. See also https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-metrics-percentile-aggregation.html#_hdr_histogram
- `apm-server.aggregation.transactions.rum.user_agent.lru_size: [1,+Inf]` (currently defaults to 5000)
  - For "page-load" RUM transactions we aggregate on the user-agent name, which requires user-agent parsing. To avoid parsing every user-agent, we maintain an cache of user-agent strings. This option controls the cache size.

There's a UI proof of concept underway in progress: ~https://github.com/elastic/kibana/pull/66871~ https://github.com/elastic/kibana/pull/73953. I think we probably want to wait for the UI side to be in its shippable state before we document these options, otherwise they'll just be confusing.

PR #4345, opened 2020-10-22
docs: Transaction metrics reference

bmorelli25:
## Summary

Alphabetizes the config reference and adds reference docs for transaction metrics.

For a number of good reasons, this is still opt-in. How much do we want to publicize this? Better worded: do we need more than just reference documentation for this feature yet?

## Related issues

Closes https://github.com/elastic/apm-server/issues/4031


docs/transaction-metrics.asciidoc
```diff
@@ -0,0 +1,93 @@
+[x-pack]
+[[transaction-metrics]]
+== Configure transaction metrics
+
+++++
+<titleabbrev>Transaction metrics</titleabbrev>
+++++
+
+When enabled, {beatname_uc} produces the transaction histogram metrics that are used to power the APM app.
+Shifting this responsibility from APM app to APM Server results in improved query performance.
```
axw, line 10:
```suggestion
Shifting this responsibility from APM app to APM Server results in improved query performance and removes the need to store unsampled transactions.
```

Not sure how much details we want to get into here. Storage reductions are the main driver behind this, query time may not be affected much.

docs/transaction-metrics.asciidoc
```diff
@@ -0,0 +1,93 @@
...
+
+Default: `5000`.
+
+[float]
+[[configuration-sampling]]
+=== Configuration options: `apm-server.sampling.*`
+
+[[sampling-keep_unsampled]]
+[float]
+==== `keep_unsampled`
```
axw, line 85: I'm not sure if we should document this one. I think this will end up going away in favour of the updating the agents to stop sending unsampled transactions. Also, I suspect this config might be incompatible with RUM/User Experience monitoring, where we probably want to record individual requests without a full sampled trace.

docs/transaction-metrics.asciidoc
```diff
@@ -0,0 +1,93 @@
...
+Once exceeded, APM Server devolves into recording a metrics document for each transaction that is not in one
+of the transaction groups being tracked.
+
+Default: `1000`.
+
+[[transactions-hdrhistogram_significant_figures]]
+[float]
+==== `hdrhistogram_significant_figures`
+
+The fixed, worse-case percentage error (specified as a number of significant digits)
```
axw, line 61: wors**t**-case?

docs/transaction-metrics.asciidoc
```diff
@@ -0,0 +1,93 @@
+[x-pack]
+[[transaction-metrics]]
+== Configure transaction metrics
+
```
axw, line 4: The feature is currently experimental. Should we have a flag here?

axw (commented):
Thank you!

dgieselaar:
This feature is also opt-in in Kibana. Do we need docs for that as well?

docs/transaction-metrics.asciidoc
bmorelli25: If we do remove this config from the docs, do the `apm-server.aggregation.transactions.*` configs still result in storage reduction? I want to make sure your previous suggested change still holds true if we remove this.

docs/transaction-metrics.asciidoc
```diff
@@ -0,0 +1,95 @@
...
+Default: `1000`.
+
+[[transactions-hdrhistogram_significant_figures]]
+[float]
+==== `hdrhistogram_significant_figures`
+
+The fixed, worst-case percentage error (specified as a number of significant digits)
+to maintain for recorded metrics.
+Supported values are `1` through `5`.
+See {ref}/search-aggregations-metrics-percentile-aggregation.html#_hdr_histogram[HDR histogram] for more information.
```
bmorelli25, line 66: req'd to fix doc failure

```suggestion
See {ref}/search-aggregations-metrics-percentile-aggregation.html#_hdr_histogram_2[HDR histogram] for more information.
```

bmorelli25:
> This feature is also opt-in in Kibana. Do we need docs for that as well?

Yes. Is it `xpack.apm.searchAggregatedTransactions`? I can add a note to these docs with a link to https://www.elastic.co/guide/en/kibana/current/apm-settings-in-kibana.html and vice versa.

Also, question: When I enable this, how do I know that it's working? What can I look for?

docs/transaction-metrics.asciidoc
axw: Ahh good point. Indeed, storage reduction relies on us discarding the unsampled transaction docs.
I guess we should keep it then, and just mark it as experimental.

axw:
> Also, question: When I enable this, how do I know that it's working? What can I look for?

With the example config

```yaml
apm-server:
  aggregation:
    transactions:
      enabled: true
      interval: 1m
  sampling:
    keep_unsampled: false
```

Configure the agent to sample, and you should end up with a distribution chart that looks roughly correct. If Kibana were not configured with `searchAggregatedTransactions`, then the charts would show numbers based on only the sampled transactions.

bmorelli25:
Perfect. Thanks for your help, @axw.

RobDTech:
Hi, this is a fantastic feature and just what we need. 
We don't appear to be able to drop the unsampled docs as we still get documents with "transaction.sampled: false". 

We are on v7.13.2.

We have configured the following

apm-server:
  host: "0.0.0.0:8200"
  aggregation:
    transactions:
      enabled: true
      interval: 1m
  sampling:
    keep_unsampled: false

Log entry suggests the aggregation is working:
{"file.name":"apm-server/main.go","file.line":57},"message":"creating transaction metrics aggregation with config: {Enabled:true Interval:1m0s MaxTransactionGroups:10000 HDRHistogramSignificantFigures:2}","ecs.version":"1.6.0"}

I can't see anything around keep unsampled in the logs.

Is this enabled in v7.13.2 or what could we be doing wrong please? 

Many thanks,

Rob

cyrille-leclerc:
@bmorelli25 I searched how to create a graph from Transaction as Histograms and didn't find the docs searching on Google. Do we have an SEO question here?

bmorelli25:
> @bmorelli25 I searched how to create a graph from Transaction as Histograms and didn't find the docs searching on Google. Do we have an SEO question here?

That's a result of only adding reference documentation for this feature. There's no how-to guide with a Googlable H1. This plays into the thread on Slack... Would you mind opening an issue?

cyrille-leclerc:
FYI I created https://github.com/elastic/observability-docs/issues/1335

PR #4346, opened 2020-10-25
sampling: add logging and monitoring

axw:
## Motivation/summary

Log rate-limited warnings when the maximum number of
dynamic services is reached, causing root transactions
(indirectly, related trace events) to be dropped as a result.

Also, add some monitoring: record the number of events
(transactions and spans) that the tail-sampler has processed,
stored (locally), or explicitly dropped. Report the current LSM
and value log size for Badger. Report the current number of
dynamic service groups.

## Checklist

~- [ ] documentation~
- [x] metrics and monitoring (create issue for Kibana team to add metrics to visualizations, e.g. [Kibana#44001](https://github.com/elastic/kibana/issues/44001))
~- [ ] telemetry~ (config telemetry coming in a followup)
- [x] Elastic Cloud Enterprise (https://www.elastic.co/products/ece)
- [x] Elastic Cloud on Kubernetes (https://www.elastic.co/elastic-cloud-kubernetes)

## How to test these changes

- Run server with monitoring and tail-sampling enabled, with a simple catch-all policy (e.g. only `sample_rate` defined)
- Send 1002 transactions each with a unique service name
- Observe that a single warning is logged due to the hard-coded limit of 1000 services being reached
- Observe reported metrics (query the index) includes the metrics documented in the description. One of those should be `dynamic_service_groups`, which should be 1000. Bear in mind that this depends on timing, as the groups may be reset between monitoring intervals.

## Related issues

https://github.com/elastic/apm-server/issues/4185

### DIFF
--- a/x-pack/apm-server/sampling/processor.go
+++ b/x-pack/apm-server/sampling/processor.go
@@ -143,22 +143,23 @@ func (p *Processor) ProcessTransformables(ctx context.Context, events []transfor
 		return nil, ErrStopped
 	}
 	for i := 0; i < len(events); i++ {
-		var drop, stored bool
+		var report, stored bool
 		var err error
 		switch event := events[i].(type) {
 		case *model.Transaction:
 			atomic.AddInt64(&p.eventMetrics.processed, 1)
-			drop, stored, err = p.processTransaction(event)
+			report, stored, err = p.processTransaction(event)
 		case *model.Span:
 			atomic.AddInt64(&p.eventMetrics.processed, 1)
-			drop, stored, err = p.processSpan(event)
+			report, stored, err = p.processSpan(event)
 		default:
 			continue
 		}
 		if err != nil {
 			return nil, err
 		}
-		if drop {
+		if !report {
+			// We shouldn't report this event, so remove it from the slice.
 			n := len(events)
 			events[i], events[n-1] = events[n-1], events[i]
 			events = events[:n-1]
@@ -166,11 +167,12 @@ func (p *Processor) ProcessTransformables(ctx context.Context, events []transfor
 		}
 		if stored {
 			atomic.AddInt64(&p.eventMetrics.stored, 1)
-		} else if drop {
+		} else if !report {
 			// We only increment the "dropped" counter if
-			// we didn't also store the event, so we can
-			// track how many events are definitely dropped
-			// and indexing isn't just deferred until later.
+			// we neither reported nor stored the event, so
+			// we can track how many events are definitely
+			// dropped and indexing isn't just deferred until
+			// later.
 			//
 			// The counter does not include events that are
 			// implicitly dropped, i.e. stored and never
@@ -181,19 +183,20 @@ func (p *Processor) ProcessTransformables(ctx context.Context, events []transfor
 	return events, nil
 }
 
-func (p *Processor) processTransaction(tx *model.Transaction) (drop, stored bool, _ error) {
+func (p *Processor) processTransaction(tx *model.Transaction) (report, stored bool, _ error) {
 	if tx.Sampled != nil && !*tx.Sampled {
 		// (Head-based) unsampled transactions are passed through
 		// by the tail sampler.
-		return false, false, nil
+		return true, false, nil
 	}
 
 	traceSampled, err := p.storage.IsTraceSampled(tx.TraceID)
 	switch err {
 	case nil:
-		// Tail-sampling decision has been made, index or drop the transaction.
-		drop := !traceSampled
-		return drop, false, nil
+		// Tail-sampling decision has been made: report the transaction
+		// if it was sampled.
+		report := traceSampled
+		return report, false, nil
 	case eventstorage.ErrNotFound:
 		// Tail-sampling decision has not yet been made.
 		break
@@ -204,7 +207,7 @@ func (p *Processor) processTransaction(tx *model.Transaction) (drop, stored bool
 	if tx.ParentID != "" {
 		// Non-root transaction: write to local storage while we wait
 		// for a sampling decision.
-		return true, true, p.storage.WriteTransaction(tx)
+		return false, true, p.storage.WriteTransaction(tx)
 	}
 
 	// Root transaction: apply reservoir sampling.
@@ -216,7 +219,7 @@ Tail-sampling service group limit reached, discarding trace events.
 This is caused by having many unique service names while relying on
 sampling policies without service name specified.
 `[1:])
-		return true, false, nil
+		return false, false, nil
 	} else if err != nil {
 		return false, false, err
 	}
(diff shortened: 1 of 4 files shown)


x-pack/apm-server/sampling/processor.go
```diff
@@ -112,45 +164,61 @@ func (p *Processor) ProcessTransformables(ctx context.Context, events []transfor
...
 	default:
-		return false, err
+		return false, false, err
 	}
 
 	if tx.ParentID != "" {
 		// Non-root transaction: write to local storage while we wait
 		// for a sampling decision.
-		return true, p.storage.WriteTransaction(tx)
+		return true, true, p.storage.WriteTransaction(tx)
```
simitt, line 207: Why is `drop: true` in this case? I see that in https://github.com/elastic/apm-server/pull/4346/files#diff-23ea3f45d5b6898582f132034e90a9e3da30d7345b4cbe4c5799237094e418cfR169 you are only using the `drop` information if `stored` is `false` but I would find it more intuitive to only set `drop: true` once it is really dropped. Maybe I am missing something, but I think changing it here would not have any side effects.

x-pack/apm-server/sampling/processor.go
```diff
@@ -160,27 +228,29 @@ func (p *Processor) processTransaction(tx *model.Transaction) (bool, error) {
...
 		// transactions, we do not propagate this to other APM Servers.
-		return true, p.storage.WriteTraceSampled(tx.TraceID, false)
+		return true, false, p.storage.WriteTraceSampled(tx.TraceID, false)
 	}
 
 	// The root transaction was admitted to the sampling reservoir, so we
 	// can proceed to write the transaction to storage and then drop it;
 	// we may index it later, after finalising the sampling decision.
-	return true, p.storage.WriteTransaction(tx)
+	return true, true, p.storage.WriteTransaction(tx)
```
simitt, line 237: Actually reading this comment - what is the meaning of `dropped` in this context?

simitt (commented):
Overall looks good, but I am confused by the `drop` counter (see my comments below).

x-pack/apm-server/sampling/processor.go
axw: "dropped" has a double meaning, which is a bit confusing. Here we're saying that we should drop the transaction from the list of events that are to be returned from ProcessTransformables, to prevent them from being indexed immediately.

Perhaps it would be more understandable if I inverted this return variable? i.e. instead of telling ProcessTransformables to drop the event, tell it if we want it to report the event immediately. Then we'll increment the dropped counter if and only if we're not reporting immediately, and we also did not store the event for later.

simitt: If you don't mind changing, I believe it would make things more clear.

axw: I've pushed a change, please let me know if it's clear enough now.

simitt (approved):

x-pack/apm-server/sampling/processor.go
simitt: Thank you - the changes make it easier to understand

simitt:
Tested with BC1 - observed unexpected non-consistent behavior.

```
apm-server:
  sampling:
    tail:
      enabled: true
      interval: 10s
      policies:
      - sample_rate: 0.1
```

Sending 1 request with 1 error (gets indexed as expected) and 1005 transactions all with unique `context.service.name`, `trace.id` and `transaction.id`, for example:
```
{"transaction":{"context":{"service":{"name":"11"}},"name":"mytransaction","type":"http","id":"11","trace_id":"11","span_count":{"started":17,"dropped":0},"duration":32.592981}}
```

Observed:
* nothing is logged
* observed metrics:
```
               "sampling" : {
                  "tail" : {
                    "dynamic_service_groups" : 0,
                    "storage" : {
                      "lsm_size" : 44840,
                      "value_log_size" : 4293131
                    },
                    "events" : {
                      "dropped" : 2,
                      "stored" : 1000,
                      "processed" : 1005
                    }
                  },
                  "transactions_dropped" : 0
                },
```
* 3 transactions are indexed:
```
GET /apm*transaction*/_search
{
  "size": 0,
  "aggs": {
    "by_name": {
      "terms": {
        "field": "service.name",
        "size": 1005
      }
    }
  }
}
```
```
{
  "took" : 0,
  "timed_out" : false,
  "_shards" : {
    "total" : 1,
    "successful" : 1,
    "skipped" : 0,
    "failed" : 0
  },
  "hits" : {
    "total" : {
      "value" : 3,
      "relation" : "eq"
    },
    "max_score" : null,
    "hits" : [ ]
  },
  "aggregations" : {
    "by_name" : {
      "doc_count_error_upper_bound" : 0,
      "sum_other_doc_count" : 0,
      "buckets" : [
        {
          "key" : "216",
          "doc_count" : 1
        },
        {
          "key" : "426",
          "doc_count" : 1
        },
        {
          "key" : "477",
          "doc_count" : 1
        }
      ]
    }
  }
}

```

@axw since I also observed unexpected behavior for other sampling testing, this might be related; I assume I might be doing something wrong.

simitt:
Also tested with only 500 events per request and observed the same behavior (same 3 events as above were stored), sending 1 error + 505 transactions with IDs between 500 and 1005 none of the transactions were stored, but monitoring index shows following:
```
                "sampling" : {
                  "transactions_dropped" : 0,
                  "tail" : {
                    "storage" : {
                      "value_log_size" : 7936255,
                      "lsm_size" : 44740
                    },
                    "events" : {
                      "processed" : 507,
                      "dropped" : 0,
                      "stored" : 507
                    },
                    "dynamic_service_groups" : 507
                  }
                }
```

axw:
@simitt I'm not sure why you're getting the same 3 events in two different runs, or any events at all for that matter. Since you've set the sample rate to 0.1 and there's only 1 transaction for each service name, we should not be recording anything. The warning _should_ still be logged though.

I just tested with sample rate set to 0.999 (can't currently set to 1.0, and ran the following program and it logged the message as expected. The number of events recorded isn't right though, per the other issue.

```go
package main

import (
        "fmt"
        "sync"
        "time"

        "go.elastic.co/apm"
)

func main() {
        var g sync.WaitGroup
        for i := 0; i < 1005; i++ {
                g.Add(1)
                tracer, _ := apm.NewTracer(fmt.Sprintf("service_%d", i+1), "")
                go func() {
                        defer g.Done()
                        defer tracer.Close()
                        defer tracer.Flush(nil)
                        tx.Duration = time.Second
                        tx.End()
                }()
        }
        g.Wait()
}
```

simitt:
Similar to https://github.com/elastic/apm-server/pull/4320#issuecomment-750823379 I was testing with artifical data, which had a flaw. Switching to test with a proper test app, I can confirm behavior as expected.
Tested for BC2.

PR #4382, opened 2020-11-04
docs: Alphabetize exported field ref

bmorelli25:
## Motivation/summary

I was chatting with @jalvz this morning and he pointed out "APM system metrics fields" aren't sorted alphabetically in the field reference. This PR changes the "APM System Metrics" `key` to be consistent with other APM field keys.

I'm not sure if this `key` is used outside of the docs, so my apologies if this is a dumb idea and breaks things in APM Server 🙈 . The other option would be to keep the `key` the same and change the title of this section from "APM System Metrics" to "System Metrics".

## Screenshots

// removed


axw (commented):
Looks better, but I'm wondering if we shouldn't change the title from "APM System Metrics" to "System Metrics" instead. These are a subset of the system metrics captured by Metricbeat. Similar to the process metrics below, which we don't call "APM Process Metrics".

bmorelli25:
We can definitely do that; I wasn't sure whether a name change or key change made more sense. I'll make the update.

axw (approved):
Thank you!

jalvz (approved):
Thanks!

PR #4395, opened 2020-11-09
fix: set default value for ONLY_DOCS env var

kuisathaverat:
## Motivation/summary

There is no default value for ONLY_DOCS, so only on PR has a value, this makes no stage is run. This PR fix the issue by setting a default value for ONLY_DOCS

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -16,6 +16,7 @@ pipeline {
     DOCKER_SECRET = 'secret/apm-team/ci/docker-registry/prod'
     DOCKER_REGISTRY = 'docker.elastic.co'
     DOCKER_IMAGE = "${env.DOCKER_REGISTRY}/observability-ci/apm-server"
+    ONLY_DOCS = "false"
   }
   options {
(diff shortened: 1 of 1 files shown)


mdelapenya (approved):

v1v (approved):
♥️

simitt (approved):

Jenkinsfile
```diff
@@ -16,6 +16,7 @@ pipeline {
     DOCKER_SECRET = 'secret/apm-team/ci/docker-registry/prod'
     DOCKER_REGISTRY = 'docker.elastic.co'
     DOCKER_IMAGE = "${env.DOCKER_REGISTRY}/observability-ci/apm-server"
+    ONLY_DOCS = "false"
```
cachedout, line 19: Nit: This seems to be a string whereas on L82 we set it to a bool when we use the value from `isGitRegionMatch`. Then on L106 it seems like we eval it as a string again.

kuisathaverat: Environment vars are stored as Strings no matter the input type.

PR #4398, opened 2020-11-10
fix: correct package stage when condition

kuisathaverat:
## Motivation/summary

* Trigger package stage on master and release branches
* Trigger package stage on PRs that updates libbeat
* Trigger package stage on PRS when there is a comment like `jenkins run package tests`

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -410,8 +410,11 @@ pipeline {
               expression { return params.release_ci }
               expression { return env.ONLY_DOCS == "false" }
               anyOf {
-                expression { return env.BEATS_UPDATED != "false" }
+                branch 'master'
+                branch pattern: '\\d+\\.\\d+', comparator: 'REGEXP'
+                expression { return isPR() && env.BEATS_UPDATED != "false" }
                 expression { return env.GITHUB_COMMENT?.contains('package tests') }
+                expression { return params.Run_As_Master_Branch }
               }
             }
           }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -412,6 +412,7 @@ pipeline {
               anyOf {
                 branch 'master'
                 branch pattern: '\\d+\\.\\d+', comparator: 'REGEXP'
+                tag pattern: 'v\\d+\\.\\d+\\.\\d+.*', comparator: 'REGEXP'
                 expression { return isPR() && env.BEATS_UPDATED != "false" }
                 expression { return env.GITHUB_COMMENT?.contains('package tests') }
                 expression { return params.Run_As_Master_Branch }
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -410,7 +410,9 @@ pipeline {
               expression { return params.release_ci }
               expression { return env.ONLY_DOCS == "false" }
               anyOf {
-                expression { return env.BEATS_UPDATED != "false" }
+                branch 'master'
+                branch comparator: 'REGEXP', pattern: '\\d\\.\\d'
```
simitt, line 414: shouldn't this be `'\\d+\\.\\d+'`?

simitt (changes requested):
I might be misunderstanding something, but I believe a lot of the checks in https://github.com/elastic/apm-server/blob/fae7121c62693f153adbd94bcfffe214d793b798/Jenkinsfile#L441-L446 are redundant since https://github.com/elastic/apm-server/blob/fae7121c62693f153adbd94bcfffe214d793b798/Jenkinsfile#L412-L416 is checked for the parent stage already; having the checks in there might make the setup more error prone and confusing; e.g. the `Publish` stage should be run when `params.Run_As_Master_Branch` evaluates to `true` - but that is not even an option in the parent check.

Jenkinsfile
kuisathaverat: yep, 7.10 is not covered

kuisathaverat:
> I might be misunderstanding something, but I believe a lot of the checks in
> 
> https://github.com/elastic/apm-server/blob/fae7121c62693f153adbd94bcfffe214d793b798/Jenkinsfile#L441-L446
> 
> are redundant since
> https://github.com/elastic/apm-server/blob/fae7121c62693f153adbd94bcfffe214d793b798/Jenkinsfile#L412-L416
> 
> is checked for the parent stage already; having the checks in there might make the setup more error prone and confusing; e.g. the `Publish` stage should be run when `params.Run_As_Master_Branch` evaluates to `true` - but that is not even an option in the parent check.

yes, if we group stages on a parent stage, we could make a refactor of the conditions, but right now there is not that parent stage, that change is not the change made in this PRs, we are changing the when condition on the packaging stage

simitt:
I don't believe that's true - there is two stages called `Package`. The one where this PR adds conditions [Stage Package:L398](https://github.com/elastic/apm-server/blob/fae7121c62693f153adbd94bcfffe214d793b798/Jenkinsfile#L398) is the parent stage to [Stage Package:421](https://github.com/elastic/apm-server/blob/fae7121c62693f153adbd94bcfffe214d793b798/Jenkinsfile#L421) and [Stage Publish:437](https://github.com/elastic/apm-server/blob/fae7121c62693f153adbd94bcfffe214d793b798/Jenkinsfile#L437).

You can do the refactor in a separate PR to ensure we get this fix in quickly; but wanted to point out that some of the conditions in the `Publish` stage will never even be reached as they are not part of the parent `Package` stage conditions - so that might be an actual bug (e.g. when run _as_master_).

simitt (approved):
Good to merge and do the refactoring of the conditions in a follow up. 

Thanks for the fix!

kuisathaverat:
> I don't believe that's true - there is two stages called `Package`. The one where this PR adds conditions [Stage Package:L398](https://github.com/elastic/apm-server/blob/fae7121c62693f153adbd94bcfffe214d793b798/Jenkinsfile#L398) is the parent stage to [Stage Package:421](https://github.com/elastic/apm-server/blob/fae7121c62693f153adbd94bcfffe214d793b798/Jenkinsfile#L421) and [Stage Publish:437](https://github.com/elastic/apm-server/blob/fae7121c62693f153adbd94bcfffe214d793b798/Jenkinsfile#L437).
> 
> You can do the refactor in a separate PR to ensure we get this fix in quickly; but wanted to point out that some of the conditions in the `Publish` stage will never even be reached as they are not part of the parent `Package` stage conditions - so that might be an actual bug (e.g. when run _as_master_).

I was thinking on refactor all when conditions, I didn't realize that package and publish are serial stages in the parallel block

Screenshot 2020-11-10 at 12 28 04

yep, for the package and publish we can refactor part of the when condition

simitt (approved):

Issue #4378, opened 2020-11-03
Implement the new indexing strategy


jalvz:
This is for integration with Elastic Agent, part of https://github.com/elastic/apm-server/issues/4004

The new indexing strategy is defined [here](https://docs.google.com/document/d/1y56a9fjkLi6Zen5qGC_JKYM9ljYpBA5W0fgdWilcwYc/edit?ts=5f6c1aab#).

### TL;DR
Index names will have 3 variable "chunks": `type` and `service name`, that come from ingested data; and `namespace`, that comes from Elastic Agent. 
`service name` and `namespace` are user-defined, while `type` is well known in advance and can be `logs`, `metrics`, or `traces`. 
We need to route the right events to the right indices based on those variables.

### Requires  

For `service name`:
https://github.com/elastic/package-spec/issues/64

For `traces` type:
https://github.com/elastic/package-spec/issues/68
https://github.com/elastic/kibana/issues/81451
https://github.com/elastic/elasticsearch/issues/64047

axw:
For posterity, code for agent/fleet config: https://github.com/elastic/apm-server/compare/master...jalvz:integration-model?expand=1

axw:
Existing integrations add the fields `data_stream.type`, `data_stream.dataset` and `data_stream.namespace` as fields to docs. We should do the same; this will also simplify the rest of this task, since we can define index selection rules based on these fields.

axw:
Still TODO:

- [ ] system tests
- [ ] set `apm-server.data_streams.enabled` when in Fleet mode
- [ ] make `data_stream.namespace` configurable using config received from Elastic Agent
- [ ] improve config validation, to ensure things like `output.elasticsearch.indices` cannot be set in conjunction `apm-server.data_streams.enabled`

axw:
Punting docs to 7.12+: https://github.com/elastic/apm-server/issues/4603

PR #4409, opened 2020-11-11
Add initial support for indexing to data streams

axw:
## Motivation/summary

This PR introduces initial support for indexing events into data streams, with the new `apm-server.data_streams.enabled` config. When this is set, the server will:
 - add `data_stream.{type,dataset,namespace}` fields to each published event
 - use these fields to route the events to the appropriate data stream (`$type-$dataset-$namespace`)
 - disable ILM setup by default

Still TODO, in followups:
 - system tests
 - set `apm-server.data_streams.enabled` when in Fleet mode
 - make `data_stream.namespace` configurable (using config received from Elastic Agent)
 - improve config validation, to ensure things like `output.elasticsearch.indices` cannot be set in conjunction `apm-server.data_streams.enabled`

## Checklist

~- [ ] documentation~ (later maybe; more likely this will be undocumented and set indirectly by using Fleet)
- [x] telemetry

## How to test these changes

- Run apm-server with `-E apm-server.data_streams.enabled=true`
- Send some transactions, spans, errors, and runtime metrics
- Confirm that transactions and spans go into `traces-*`, errors into `logs-*`, and runtime metrics into `metrics-*`.

## Related issues

Partially addresses #4378


idxmgmt/supporter.go
```diff
@@ -71,48 +88,51 @@ type ilmIndexSelector struct {
...
+	if cfg.Output.Name() != esKey || cfg.ILM.Mode == libilm.ModeDisabled {
+		disableILM = true
+	} else if cfg.ILM.Mode == libilm.ModeAuto {
+		// ILM is set to "auto": disable if we're using data streams,
+		// or if we're not using data streams but we're using customised,
+		// unmanaged indices.
+		if cfg.DataStreams || unmanagedIdxCfg.Customized() {
+			disableILM = true
+		}
+	}
```
simitt, line 119: This indicates that when `apm-server.ilm.enabled=true` and `apm-server.data_streams.enabled=true` the ILM config is stronger than the data streams. While I understand that data streams is early stage and not enabled by default, it is the new strategy so would flip this and make data streams overrule ILM if configured. 

Looking at https://github.com/elastic/apm-server/pull/4409/files#diff-257c35f49365729cf758ca1de95ac8957ae2b337b0bd85ab21e7a992a925e94bR168 this also doesn't seem aligned with the `BuildSelector` that is returned - as soon as `cfg.DataStream` is true it will return the data stream supporter. 
So what I could observe is that if `apm-server.ilm.enabled=true` and `apm-server.data_streams.enabled=true` data are indexed to the new data stream indices, but the ILM indices are still created.

idxmgmt/supporter.go
```diff
@@ -63,6 +66,20 @@ type indexState struct {
...
+	fmtstr, err := fmtstr.CompileEvent(datastreams.IndexFormat)
+	if err != nil {
+		return nil, err
+	}
+	expr, err := outil.FmtSelectorExpr(fmtstr, "", outil.SelectorLowerCase)
+	if err != nil {
+		return nil, err
+	}
+	return outil.MakeSelector(expr), nil
+}
```
simitt, line 81: What if ES does not support data streams? Then the events would still be ingested in the new way, but there would not be any base templates or data stream setup. Since it is very unlikely someone turns this on at this stage (since other pieces are missing) we can ignore it for now, but might lead to problems in the future. When introducing ILM we added covered for ES versions not supporting ILM.

changelogs/head.asciidoc
```diff
@@ -22,3 +22,4 @@ https://github.com/elastic/apm-server/compare/7.10\...master[View commits]
 * Log warnings in aggregation of transaction metrics when grouping limit is reached {pull}4313[4313]
 * Configurable tail-based sampling policies {pull}4320[4320]
 * Monitoring and telemetry for tail-based sampling {pull}4346[4346] {pull}4360[4360]
+* Support for data streams {pull}4409[4409]
```
simitt, line 25: should we put a note that this is experimental (or is it experimental?)

simitt (commented):
Generally looks good to me, left a couple of comments. 

I realized that `apm-*` templates are still set up when data streams are enabled, while they wouldn't be applied any more. Doesn't have to be changed in this PR but we could disable template setup in this case. 


Out of curiosity - Is there a Kibana/APM issue already to adapt the UI to the data stream changes? (right now it is ofc broken when data streams are enabled)

datastreams/servicename.go
```diff
@@ -0,0 +1,39 @@
...
+func NormalizeServiceName(s string) string {
+	s = strings.ToLower(s)
+	s = strings.Map(replaceReservedRune, s)
+	return s
+}
+
+func replaceReservedRune(r rune) rune {
+	switch r {
+	case '\\', '/', '*', '?', '"', '<', '>', '|', ' ', ',', '#', ':':
+		return '_'
```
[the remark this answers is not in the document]
axw: Whoops! Thank you, good catch - it should indeed be included here.

changelogs/head.asciidoc
axw: Yep, reworded to "Experimental support"

idxmgmt/supporter.go
axw: I expect we won't document the new `apm-server.data_streams.enabled` config -- it's meant as an internal config to be switched on when using Fleet/Agent, which implies data streams are supported.

When using Fleet the server will be configured with minimal privileges, so checking for data stream support won't even be possible in that mode AFAIK.

idxmgmt/supporter.go
axw: This code here is only meant to be about changing the default behaviour, when the user hasn't set anything in config. I was planning to create a follow up PR with additional config validation that would prevent setting `apm-server.ilm.*`, `output.index`, and `output.indices` when `apm-serverdata_streams.enabled=true`. Would that address your concern, and are you OK with it coming in a follow up?

axw:
> I realized that apm-* templates are still set up when data streams are enabled, while they wouldn't be applied any more. Doesn't have to be changed in this PR but we could disable template setup in this case.

Existing integrations disable this by injecting config, e.g. https://github.com/elastic/beats/blob/b6896ee6892da8ce4567d63ee0fa962512fb8701/x-pack/elastic-agent/spec/filebeat.yml#L3. I don't mind disabling it by default when enabling data streams though. Probably a bit neater.

> Out of curiosity - Is there a Kibana/APM issue already to adapt the UI to the data stream changes? (right now it is ofc broken when data streams are enabled)

Not yet AFAIK. It's broken by default, but you can load a template and update the APM app settings in Kibana to fetch transactions and spans from `traces-*` instead of `apm-*`, etc.

datastreams/servicename.go
```diff
@@ -0,0 +1,45 @@
...
+import "strings"
+
+// NormalizeServiceName translates serviceName into a string suitable
+// for inclusion in a data stream name.
+//
+// Concretely, this function will lowercase the string and replace any
+// reserved characters with "_".
+func NormalizeServiceName(s string) string {
+	s = strings.ToLower(s)
+	s = strings.Map(replaceReservedRune, s)
```
jalvz, line 29: There are consequences to this, I think if it a replacement actually happens we should log a warn. And maybe create a doc issue / note specifically for this?

docs/fields.asciidoc
```diff
@@ -38,6 +38,39 @@ Fields common to various APM events.
 
 
 
+*`data_stream.type`*::
++
+--
+Data stream type, e.g. logs, metrics, or traces.
```
jalvz, line 44: I'd be a bit more explicit and state that can only be one of those values, and not an arbitrary string

model/transaction.go
```diff
@@ -111,7 +112,13 @@ func (e *Transaction) fields() common.MapStr {
 func (e *Transaction) Transform(_ context.Context, _ *transform.Config) []beat.Event {
 	transactionTransformations.Inc()
 
+	// Transactions are stored in a "traces" data stream along with spans.
+	dataset := datastreams.NormalizeServiceName(e.Metadata.Service.Name)
```
jalvz, line 116: The assumption that we don't need to differenciate spans from transactions is strong, but what if we are wrong? Isn't cheap and harmless to add an `apm.span` or `apm.transaction` dataset prefix, just in case?

model/profile.go
```diff
@@ -66,6 +67,11 @@ func (pp PprofProfile) Transform(ctx context.Context, _ *transform.Config) []bea
 		profileID = fmt.Sprintf("%x", uuid)
 	}
 
+	// Profiles are stored in their own "metrics" data stream, with a data
+	// set per service. This enables managing retention of profiling data
+	// per-service, and indepedently of lower volume metrics.
+	dataset := fmt.Sprintf("%s.profiling", datastreams.NormalizeServiceName(pp.Metadata.Service.Name))
```
jalvz, line 73: any chance we can do this `datastreams.NormalizeServiceName` call only once for all the events?

model/metricset.go
```diff
@@ -180,6 +185,18 @@ func (me *Metricset) Transform(ctx context.Context, _ *transform.Config) []beat.
 		fields["timeseries"] = common.MapStr{"instance": me.TimeseriesInstanceID}
 	}
 
+	// Metrics are stored in a "metrics" data stream. Metrics that include well-defined
+	// transaction/span fields (i.e. breakdown metrics, transaction and span metrics)
+	// will be stored separately from application and runtime metrics.
+	dataset := datastreams.NormalizeServiceName(me.Metadata.Service.Name)
```
jalvz, line 191: I thought the conclusion was that all apm documents would have an `apm` dataset prefix?

idxmgmt/supporter.go
simitt: That sounds totally fine.

idxmgmt/supporter.go
simitt: makes sense;

model/transaction.go
axw: > Isn't cheap and harmless to add an apm.span or apm.transaction dataset prefix, just in case

It's not necessarily harmless, as more datasets will lead to more shards. We can always revisit this in the future.

docs/fields.asciidoc
axw: Reworded to clarify that it will be one of those values.

datastreams/servicename.go
axw: I expect it will be pretty common, given that people can and do use hyphens in their service names. Logging warnings would suggest that they should not hyphenate, which is unreasonable.

The UI is probably a more appropriate place to call attention to this anyway, rather than expecting people to go digging in logs. We can do that by looking for reserved characters in `service.name` values if necessary.

We should certainly document the behaviour. I've added a TODO list item to https://github.com/elastic/apm-server/issues/4378.

model/profile.go
axw: It would be possible by having each event define a dataset prefix, and then defining a script processor. It will be more convoluted, which is why I went this way.

I haven't thought of any other better options - any ideas?

model/metricset.go
axw: Yes, I just hadn't updated the PR yet. I've updated it now.

axw:
jenkins run the tests please

axw:
jenkins run the tests please

simitt (approved):

model/transaction.go
jalvz: So how much it would hurt?
I guess its fine, but a similar argument can be done the other way around: by separating spans and transactions in different indices, it is likely that some queries in the APM UI can be changed to faster alternatives. 
We would not get this benefit if we revisit this decision in the future because the UI would have to query for both index name and index content... 
The opposite is not true: if we go with separated spans and transactions to start with, and later we find they don't add value (performance or user-flexibility wise), we would be able to optimize for less shards without any drawbacks.

model/profile.go
jalvz: Not now, was just curious

jalvz (approved):

model/transaction.go
axw: > So how much it would hurt?

We won't know for sure until we test.

This was discussed in the proposal, and the discussion concluded with this comment (from me):

"It seems to me that there's no clear answer here. In order to make progress I propose that we start out conservatively w.r.t. sharding, and have a combined data stream as described.
This will go out as experimental to start with, and we can adjust as necessary. We can replay some real data through the options to compare performance."

> I guess its fine, but a similar argument can be done the other way around: by separating spans and transactions in different indices, it is likely that some queries in the APM UI can be changed to faster alternatives.

Span docs are currently used in two places in the UI, AFAIK:
 1. displaying an individual trace waterfall
 2. walking traces for the service map

I'm confident (1) is going to be cheap either way. We know (2) is a sub-optimal approach to building the service map, and we're investigation options to move away from it.

> We would not get this benefit if we revisit this decision in the future because the UI would have to query for both index name and index content...

Not sure I follow. We don't query spans by themselves; only ever spans and transactions together. If we did want to query only spans in the future, then we could split the data sets and make `processor.event` a `constant_keyword`; that would mean only the spans data streams would be considered in the query. This can be done without changing the index pattern, and without breaking backwards compatibility.

Anyway, as mentioned above: let's go with this for now and test it at scale. We can use an ingest node processor to split the transactions and spans into separate streams for testing.

axw:
jenkins run the tests please

axw:
Something very wrong is going on in the integration tests...

axw:
(I'm going to try reproducing this locally, so I can debug...)

axw:
Same error is occurring in https://github.com/elastic/apm-server/pull/4445 ...

axw:
The integration tests have been failing for a while, and are unrelated to the server as far as I can see. I've captured my findings in https://github.com/elastic/apm-integration-testing/issues/987.

simitt:
Tested with BC1 - sent some event data for errors, spans, transactions and metrics - everything ended up in the expected data streams.

Issue #4004, opened 2020-07-21
Integrate with Elastic Agent


axw:
### Things to investigate/consider

- [x] elastic-agent configuration
- [x] elastic-agent program spec
- [x] indexing strategy (e.g. include service name in data stream name? specific data sets for transaction metrics, breakdown metrics, etc.?), constant_keyword fields, reducing flexibility in configuration
- [ ] how to read/write non-timeseries data (tail-based sampling, API Keys, source maps)
- [x] ~how to integrate with cloud-on-k8s (depends on working group)~ (WIP, tracked separately)
- [x] ~upgrade plan to 8.0~ (will be tracked separately)

______

**Update 03/11, 08/12 (Juan)**

### Main lines of work required

- [x] Implement the [Indexing Strategy](https://docs.google.com/document/d/1y56a9fjkLi6Zen5qGC_JKYM9ljYpBA5W0fgdWilcwYc/edit?ts=5f6c1aab#) (#4378, #4449)

- [x] Automate / Semi-automate APM Package: this includes splitting `fields.yml` files, generate docs, tests, other assets, document development workflow, etc. (#4432)

- [x] ~Prepare and test minimal templates, default ILM policies if any~

- [x] Move forward with spec proposed changed (linked bellow)

_____

### Issues

This list is non exahustive. Check comments for links to PRs done in relation to the APM package.

#### Required
- [x] https://github.com/elastic/package-spec/issues/68
- [x] https://github.com/elastic/package-spec/issues/70
- [x] https://github.com/elastic/kibana/issues/82599

#### Closed issues
- [x] ~https://github.com/elastic/kibana/issues/81451~
- [x] ~https://github.com/elastic/elasticsearch/issues/64047~

jalvz:
@axw you mentioned that you had some details you wanted to add here, right? or did I misinterpret?

axw:
@jalvz I had previously added a list, but I think I forgot to save and it was lost. I'll add things as I remember them :sweat_smile:
Feel free to add things you think we should look at too.

jalvz:
- In Fleet / Ingest Manager, the top level entity in an **integration** is the **data stream**. Each integration can have 1 or more data streams. Each data stream has a **type**. Currently there are 2 types defined in the spec: [logs and metrics](https://github.com/elastic/package-spec/blob/e7102f8cbd392d977bb2a21b12674dbe2ae8b985/versions/1/data_stream/manifest.spec.yml#L93-L95).

[...]

PR #4432, opened 2020-11-17
Generate fields for APM package

jalvz:
As title. For more details, head to the diff in the `Readme.md` file.

## Checklist

- [x] documentation
- [ ] telemetry

## How to test these changes

- run `make update`
- check the contents of `apmpackage/apm`

## Related issues

#4004


apmpackage/genfields.go
```diff
@@ -0,0 +1,149 @@
...
+func GenerateFields(ecsDir string) {
+	// TODO remove all field files before generate them
+
+	// TODO get this from GH directly
+	ecsFlatFields := loadECSFields(ecsDir)
+
+	inputFieldsFiles := map[string][]string{
+		"logs":    {"model/error/_meta/fields.yml"},
+		"metrics": {"model/metricset/_meta/fields.yml", "model/profile/_meta/fields.yml"},
+		"traces":  {"_meta/fields.common.yml", "model/transaction/_meta/fields.yml", "model/span/_meta/fields.yml"},
```
simitt, line 18: Why is the `_meta/fields.common.yml` only used for traces? It contains fields that are also relevant for errors and some of them for metricsets.

apmpackage/model.go
```diff
@@ -0,0 +1,77 @@
...
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package apmpackage
+
+type FieldDefinition struct {
```
simitt, line 20: Could you unexport this? (I believe it is only used within this package)

apmpackage/model.go
```diff
@@ -0,0 +1,77 @@
...
+	Required    *bool  `yaml:"required,omitempty"`
+	Type        string `yaml:"type,omitempty"`
+	Format      string `yaml:"format,omitempty"`
+	Description string `yaml:"description,omitempty"`
+	Release     string `yaml:"release,omitempty"`
+	Alias       string `yaml:"alias,omitempty"`
+	Path        string `yaml:"path,omitempty"`
+	Footnote    string `yaml:"footnote,omitempty"`
+	// Example is not consistent in ECS schema (either single field or array)
+	//Example     string             `yaml:"example,omitempty"`
```
simitt, line 35: Is there anything left to do with `Example`? Should it be an `interface{}`?

apmpackage/model.go
```diff
@@ -0,0 +1,77 @@
...
+		Format:      f.Format,
+		Description: f.Description,
+		Release:     f.Release,
+		Alias:       f.Alias,
+		Path:        f.Path,
+		Footnote:    f.Footnote,
+		IgnoreAbove: f.IgnoreAbove,
+		Fields:      nil,
+		MultiFields: f.MultiFields,
+		IsECS:       f.IsECS,
```
simitt, line 73: If the slice elements are expected to be copied you should use `copy(dst, src)`  (but looking at how it is used, it's probably fine as is)

simitt (commented):
Overall LGTM on a first review. My main question is how this integrates with any update workflow - will it be part of running `make update`? 

Do you plan on addressing the TODOs before labeling this ready for review? (I believe fetching ECS info from github rather than expecting local installation would make sense).

apmpackage/model.go
jalvz: This was copied over from the beats migration script, that probably can be removed, yes

apmpackage/model.go
jalvz: right, this copies everything except `Fields`

jalvz:
> Overall LGTM on a first review. My main question is how this integrates with any update workflow - will it be part of running make update?

Exactly, it should be part of the target that generates the big fields.yml file.

> Do you plan on addressing the TODOs before labeling this ready for review? (I believe fetching ECS info from github rather than expecting local installation would make sense).

Thanks for validating, I wasn't sure (I copied the `loadECSFields` method from the integrations repo, that expects a local install, that's why) So yeah, will do that

apmpackage/apm/0.1.0/manifest.yml
```diff
@@ -0,0 +1,29 @@
+format_version: 1.0.0
+name: apm
+title: Elastic APM
+version: 0.1.0
```
axw, line 4:
```suggestion
version: 7.11.0
```

?

I expect we'll want to keep the package version in sync with the stack.

apmpackage/cmd/gen-package/main.go
```diff
@@ -0,0 +1,50 @@
...
+	"io/ioutil"
+	"os"
+
+	"github.com/elastic/apm-server/apmpackage"
+)
+
+var packageVersion string
+
+func main() {
+	flag.StringVar(&packageVersion, "packageVersion", "0.1.0", "Package version")
```
axw, line 31: Maybe add a TODO to get the default version from `cmd/version.go` (or move that into a package and make it an exported variable)?

apmpackage/gendocs.go
```diff
@@ -0,0 +1,116 @@
...
+		"Trim": strings.TrimSpace,
+	}).ParseFiles(docsTemplateFilePath)
+	if err != nil {
+		panic(err)
+	}
+	path := DocsFilePath(version)
+	file, err := os.OpenFile(path, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0644)
+	if err != nil {
+		panic(err)
+	}
```
axw, line 51:
```suggestion
	}
	defer file.Close()
```

apmpackage/gendocs.go
```diff
@@ -0,0 +1,116 @@
...
+	"encoding/json"
+	"io/ioutil"
+	"os"
+	"path"
+	"sort"
+	"strings"
+	"text/template"
+)
+
+func GenerateDocs(inputFields map[string]Fields, version string) {
```
axw, line 30: Unless you plan to call this from some other code in the near future, can you please move everything into `apmpackage/cmd/gen-package`, and unexport all symbols?

Makefile
```diff
@@ -105,7 +110,7 @@ docker-compose.override.yml:
 # Rules for updating config files, fields.yml, etc.
 ##############################################################################
 
-update: fields go-generate add-headers copy-docs notice $(MAGE)
+update: fields gen-package go-generate add-headers copy-docs notice $(MAGE)
```
axw, line 113:
```suggestion
update: fields go-generate add-headers copy-docs gen-package notice $(MAGE)
```

`gen-package` uses the results of `copy-docs`

apmpackage/genfields.go
```diff
@@ -0,0 +1,166 @@
...
+			nonECSCopy.Fields = append(nonECSCopy.Fields, nonECSChild)
+		}
+	}
+	return ecsCopy, nonECSCopy
+}
+
+// adapted from https://github.com/elastic/integrations/tree/master/dev/import-beats
+
+func loadECSFields() map[string]interface{} {
+	resp, err := http.Get("https://raw.githubusercontent.com/elastic/ecs/master/generated/ecs/ecs_flat.yml")
```
axw, line 115: Should we be grabbing this from a tag, depending which version of ECS the server/libbeat is using?

apmpackage/model.go
```diff
@@ -0,0 +1,81 @@
...
+
+func (f field) isNonECSLeaf() bool {
+	return f.Type != "group" && !f.IsECS
+}
+
+type Fields []field
+
+func (f Fields) Len() int           { return len(f) }
+func (f Fields) Less(i, j int) bool { return f[i].Name < f[j].Name }
+func (f Fields) Swap(i, j int)      { f[i], f[j] = f[j], f[i] }
```
axw, line 50: AFAICS the Fields type exists only for implementing sort.Interface. It _might_ be neater to use `sort.Slice` instead, seeing as this is not performance-critical code.

apmpackage/model.go
```diff
@@ -0,0 +1,81 @@
...
+		Path:        f.Path,
+		Footnote:    f.Footnote,
+		IgnoreAbove: f.IgnoreAbove,
+		Fields:      nil,
+		MultiFields: f.MultiFields,
+		IsECS:       f.IsECS,
+		HasECS:      f.HasECS,
+		HasNonECS:   f.HasNonECS,
+	}
+}
```
axw, line 81:
```suggestion
func copyFieldRoot(f field) field {
	f.Fields = nil
	return f
}
```

apmpackage/model.go
```diff
@@ -0,0 +1,81 @@
...
+	Type        string                 `yaml:"type,omitempty"`
+	Format      string                 `yaml:"format,omitempty"`
+	Description string                 `yaml:"description,omitempty"`
+	Release     string                 `yaml:"release,omitempty"`
+	Alias       string                 `yaml:"alias,omitempty"`
+	Path        string                 `yaml:"path,omitempty"`
+	Footnote    string                 `yaml:"footnote,omitempty"`
+	IgnoreAbove *int                   `yaml:"ignore_above,omitempty"`
+	MultiFields []multiFieldDefinition `yaml:"multi_fields,omitempty"`
+	Fields      []field                `yaml:"Fields,omitempty"`
```
axw, line 36: Is this one's YAML tag intentionally title-cased? Why?

apmpackage/model.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+type Fields []field
+
+func (f Fields) Len() int           { return len(f) }
+func (f Fields) Less(i, j int) bool { return f[i].Name < f[j].Name }
+func (f Fields) Swap(i, j int)      { f[i], f[j] = f[j], f[i] }
+
+type multiFieldDefinition struct {
+	Name         string `yaml:"Name,omitempty"`
```
axw, line 53: Is this one's YAML tag intentionally title-cased? Why?

tests/system/apmserver.py
```diff
@@ -1,3 +1,8 @@
+from kibana import Kibana
+from es_helper import index_smap, index_span, index_error, apm_prefix
+from es_helper import cleanup, default_pipelines
+from helper import wait_until
+from beat.beat import INTEGRATION_TESTS, TestCase, TimeoutError
```
axw, line 5: This is going to break, it needs to be after the `sys.path.append`. I guess this change came from autopep8?

go.mod
```diff
@@ -21,7 +21,8 @@ require (
 	github.com/elastic/go-hdrhistogram v0.1.0
 	github.com/elastic/go-licenser v0.3.1
 	github.com/elastic/go-sysinfo v1.4.0 // indirect
-	github.com/elastic/go-ucfg v0.8.3
+	github.com/elastic/go-ucfg v0.8.4-0.20200415140258-1232bd4774a6
+	github.com/elastic/package-registry v0.13.0
```
axw, line 25: Nothing is importing this, have you run `go mod tidy`?

axw (changes requested):

axw:
A couple of comment which I forgot to add:
 - I wonder if there's value in checking in the generated files, vs. just checking in the script and static files. We can generate into the build directory, and copy from there into the integrations repo.
 - It would be nice to cache the (versioned) `ecs_flat.yml` files in the repo, so we don't have to download them every time we run `make update`.

tests/system/apmserver.py
jalvz: yes, this comes from `make fmt`. do i have the wrong version of something?

apmpackage/model.go
jalvz: no, good catch

apmpackage/apm/0.1.0/manifest.yml
jalvz: no other integration follows the stack version  (not even endpoint), any reason we should be the odd one out?

jalvz:
> I wonder if there's value in checking in the generated files

I think there is. Eg. when new fields are added is easy to see in the diff the table generated for the readme file and check it is rendered properly.

I don't have a strong opinion thou.

>  It would be nice to cache the (versioned) ecs_flat.yml files in the repo, so we don't have to download them every time we run make update.

I can add a TODO for that, but not sure that is worth the trouble, tbh. It takes millisecs and we need and internet connection to push the changes anyways.

apmpackage/apm/0.1.0/data_stream/logs/fields/ecs.yml
```diff
@@ -0,0 +1,297 @@
+- name: url
```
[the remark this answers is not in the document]
jalvz: nope!  🤦

apmpackage/apm/0.1.0/manifest.yml
axw: Fair enough. No reason I suppose - I just wasn't aware.

tests/system/apmserver.py
axw: I don't know, it happens to me too (I don't often run `make fmt` -- I just rely on vim/goimports). I've opened https://github.com/elastic/apm-server/issues/4454 to fix it.

apmpackage/gendocs.go
jalvz: awful excuse i guess, but i went this way because my IDE debugger doesn't like functions in `main` package 🤷

axw:
> I think there is. Eg. when new fields are added is easy to see in the diff the table generated for the readme file and check it is rendered properly.
>
> I don't have a strong opinion thou.

Same, so keep it for now and maybe we'll revisit it later.

> I can add a TODO for that, but not sure that is worth the trouble, tbh. It takes millisecs and we need and internet connection to push the changes anyways.

When we're updating the package, sure. I run `make update` all the time though, and I would prefer that it didn't reach out to the internet unless needed (most of the time it would not). Please add a TODO -- I'm happy to do the work since it may only annoy me ;)

apmpackage/apm/0.1.0/manifest.yml
jalvz: btw don't know the reason for that exactly, but i assume it is because in most releases there won't be new fields, which would result in many (most?) package versions actually identical.

Makefile
jalvz: why? 🤔

apmpackage/genfields.go
jalvz: yes, thought about it but didn't know how... you made me realize how now 👍

Makefile
axw: Make will evaluate the dependencies in the order given. If we run `gen-package` before `copy-docs` then we may end up with stale docs in the package.

apmpackage/cmd/gen-package/main.go
jalvz: done, lmk if i got it right

jalvz:
Wrote some actual docs in 2411077021aa1c6729ea634e3f3d12af7acf77f3. Ill hand that off to Brandon for polishing and completing. I guess he will move that somewhere else and link instead; but it would be great a review now to make sure they are accurate and understandable.

apmpackage/cmd/gen-package/genfields.go
```diff
@@ -0,0 +1,168 @@
...
+	"github.com/elastic/ecs/code/go/ecs"
+)
+
+func generateFields(version string) map[string][]field {
+
+	ecsFlatFields := loadECSFields()
+
+	inputFieldsFiles := map[string][]field{
+		"logs":    concatFields("model/error/_meta/fields.yml"),
+		"metrics": concatFields("model/metricset/_meta/fields.yml", "model/profile/_meta/fields.yml"),
```
axw, line 36:
```suggestion
		"metrics": concatFields("model/metricset/_meta/fields.yml", "model/profile/_meta/fields.yml", "x-pack/apm-server/fields/_meta/fields.yml"),
```

apmpackage/cmd/gen-package/genfields.go
```diff
@@ -0,0 +1,168 @@
...
+	resp, err := http.Get(url)
+	if err != nil {
+		panic(err)
+	}
+	fields, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		panic(err)
+	}
+	var ret map[string]interface{}
+	err = yaml.Unmarshal(fields, &ret)
```
axw, line 126:
```suggestion
	defer resp.Body.Close()
	var ret map[string]interface{}
	err = yaml.NewDecoder(resp.Body).Decode(&ret)
```

axw (approved):
LGTM with the additional x-pack fields. Thanks!

I haven't been over it with a fine-tooth comb, but it looks good and safe - so we can adjust over time as we test if needed.

simitt:
@jalvz not sure this needs additional manual testing - please feel free to remove the labels in case you don't think it brings value.

simitt:
Loooked into how to test this and decided to add a `test-plan-skip` label as everything that has been generated via `make update` has been committed and therefore approved already.

PR #4446, opened 2020-11-19
Duplicate common fields

jalvz:
## Motivation/summary

This copies all the fields in `fields.common.yml` to the `_meta/fields.yml` files of each type. 
This is needed for the APM integration, particularly for docs as they will be generated out of  the`fields.yml` files of each type.
This results in duplicated fields.

## Checklist

- [ ] documentation
- [ ] telemetry

## How to test these changes

Run `setup template` and check that is unchanged.


model/error/_meta/fields.yml
```diff
@@ -2,6 +2,767 @@
...
+
+    - name: data_stream.dataset
+      type: keyword
+      description: Data stream dataset name.
+      example: backend_service
+
+    - name: data_stream.namespace
+      type: keyword
+      description: User-defined data stream namespace.
+      example: production
```
simitt, line 18: Not a big thing, but as soon as we add the `data_stream` fields here they will be in the template and show up at the docs. Do we really want this at this point? Why do we need the fields here already?

processor/stream/package_tests/metadata_attrs_test.go
```diff
@@ -73,7 +73,7 @@ func metadataProcSetup() *tests.ProcessorSetup {
 			intakeTestProcessor{Processor: stream.Processor{MaxEventSize: lrSize}}},
 		SchemaPath: "../../../docs/spec/v2/metadata.json",
 		TemplatePaths: []string{
-			"../../../_meta/fields.common.yml",
+			"../../../model/transaction/_meta/fields.yml",
```
simitt, line 76: Can you please add a NOTE here for why transaction fields are used and add a TODO comment for when we refactor the package tests.

processor/stream/package_tests/span_attrs_test.go
```diff
@@ -50,7 +49,6 @@ func spanPayloadAttrsNotInFields() *tests.Set {
 	)
 }
 
-// fields in _meta/fields.common.yml that are shared between several data types, but not with spans
 func spanFieldsNotInPayloadAttrs() *tests.Set {
 	return tests.Union(
 		tests.NewSet(
```
simitt, line 54: Since the common fields are not used any more I believe you could remove some of the fields from this exception set (as indicated by the comment you are removing).

model/error/_meta/fields.yml
jalvz: This comes from merging @axw changes, so I'll summon him to answer this questions :)

(I personally think it is ok thou, as that is always the case for any new fields added - they show up in the master docs before the actual release)

processor/stream/package_tests/span_attrs_test.go
jalvz: Correct me if Im wrong, but I think is more like the comment was outdated? All the metadata fields are added to spans now: https://github.com/elastic/apm-server/blob/09951bac1edfb9bb45ba50b2f85081ec988318f3/model/span.go#L206

model/error/_meta/fields.yml
simitt: Right, we already have the fields in the templates. I got confused with having the fields in the dedicated [base-field.yml in the other PR](https://github.com/elastic/apm-server/pull/4432/files#diff-bb85cbda3ca728d641efb8225953c02e871930ebcd87a4ca7529fc2f309a1b99). 

Will take this to an offline discussion about why we introduced the fields.

processor/stream/package_tests/metadata_attrs_test.go
jalvz: I added a comment, but wouldn't know what to say in a TODO. Feel free to add a GH comment-suggestion thingy.

processor/stream/package_tests/span_attrs_test.go
simitt: right!

processor/stream/package_tests/metadata_attrs_test.go
simitt: How about something along the lines of:
```
TODO: move metadata package tests into event specific tests when refactoring package tests
```

simitt (approved):

processor/stream/package_tests/metadata_attrs_test.go
jalvz:
```suggestion
			// TODO: move metadata package tests into event specific tests when refactoring package tests
			"../../../model/transaction/_meta/fields.yml",
```

model/error/_meta/fields.yml
jalvz: > I got confused with having the fields in the dedicated base-field.yml in the other PR.

Totally reasonable!

axw: I probably should not have added those fields as they're really only meaningful when running under Agent. We can remove them in a followup, as there will probably be some fallout in tests.

axw (approved):

jalvz:
cc @bmorelli25 field reference docs will look different because of this, there is no longer a "common fields" section, instead all fields are duplicated for each type (transaction, span, etc) - Like beats does. Let me know if you spot something odd!

axw:
I ran `apm-server export template` for 7.10 and 7.11 BC1, and compared them. The differences were:
 - various ECS field additions
 - various ECS fields changed from keyword to wildcard
 - removal of the accidental "tracing" ECS fields
 - `labels_string`, `labels_boolean`, and `labels_*` dynamic templates are repeated. This seems undesirable. I'll open an issue for that, otherwise looks good.

axw:
Opened #4576

PR #4447, opened 2020-11-19
chore: use testcontainers to control compose's life cycle in system tests

mdelapenya:
## Motivation/summary
Let's use the library to start the compose file.

We are also adding support to destroy it if the `DESTROY_STACK` environment variable is set. Otherwise (default) will keep existing behaviour.

## Checklist

- [ ] documentation
- [ ] telemetry

## How to test these changes

```shell
$ cd systemtest
# will destroy the stack after tests run
$ DESTROY_STACK=true go test ./...
# won't destroy the stack after tests run
$ DESTROY_STACK= go test ./...
```

## Follow-ups
I'm not adding the new variable to Jenkins (maybe in at `script/common.bash`) because it will cause the containers to be destroyed before the `post` section in the Jenkins pipeline catures it to archive them in Jenkins UI.

### DIFF
--- a/systemtest/containers.go
+++ b/systemtest/containers.go
@@ -72,6 +72,22 @@ func StartStackContainers() error {
 	return waitKibanaContainerHealthy(ctx)
 }
 
+// ShutdownStackContainers stops the stack if the "DESTROY_STACK"
+// environment variable is set
+func ShutdownStackContainers() {
+	if "" == os.Getenv("DESTROY_STACK") {
+		log.Println("We leave Elasticsearch and Kibana running, to avoid slowing down iterative development and testing. Use docker-compose to stop services as necessary")
+		return
+	}
+
+	execError := compose.Down()
+	err := execError.Error
+	if err != nil {
+		log.Fatalf("Could not destroy the stack: %v\n", err)
+	}
+	return
+}
+
 // NewUnstartedElasticsearchContainer returns a new ElasticsearchContainer.
 func NewUnstartedElasticsearchContainer() (*ElasticsearchContainer, error) {
 	// Create a testcontainer.ContainerRequest based on the "elasticsearch service
--- a/systemtest/containers.go
+++ b/systemtest/containers.go
@@ -24,7 +24,6 @@ import (
 	"net"
 	"net/url"
 	"os"
-	"os/exec"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


v1v (approved):

cachedout (approved):

axw (changes requested):
Can you please add a motivation for these changes?

At the moment I don't see any benefit to using testcontainers for running docker-compose. There's more code to maintain which is a -1 for me; so what do we gain by adding it?

Also, what is the benefit of having `DESTROY_STACK` over directly running `docker-compose down` afterwards?

mdelapenya:
> At the moment I don't see any benefit to using testcontainers for running docker-compose. There's more code to maintain which is a -1 for me; so what do we gain by adding it?
> 
> Also, what is the benefit of having `DESTROY_STACK` over directly running `docker-compose down` afterwards?

I added it with the initial intention of controlling the containers started in tests directly from the test runner (TestMain), but I did not wanted to change the current behaviour of keeping containers running, which is used by developers. I was tempted to add the variable only for Jenkins to keep it, but then I realised that if destroying the containers alongside the tests finish, would remove the ability to archive docker logs in the Jenkins UI, because it relies on running containers (we are brainstorming about this too).

I could simply and completely remove the shutdown method and its call, which would simplify the code (less code to maintain), still using testcontainers to spin up the compose file. I'd like to use the library instead of custom (but simple!) code because we are planning to add a WaitFor strategy for compose (see https://github.com/testcontainers/testcontainers-go/pull/261), which would help removing the waitForKibana method (again less code 😊). We can wait for that PR to be released preparing the code for it, or simply keeping it as is, what ever you see more convenient.

axw:
> we are planning to add a WaitFor strategy for compose (see testcontainers/testcontainers-go#261), which would help removing the waitForKibana method (again less code blush).
> We can wait for that PR to be released preparing the code for it

Thanks for the additional context. Sounds like a good change, but let's wait for that feature before switching over please.

Re container destruction: even if we can automatically destroy containers at the end of the run, this can already be done by simply calling `docker-compose down` from the Jenkins script. What does this function+environment variable improve on?

mdelapenya:
> > we are planning to add a WaitFor strategy for compose (see [testcontainers/testcontainers-go#261](https://github.com/testcontainers/testcontainers-go/pull/261)), which would help removing the waitForKibana method (again less code blush).
> > We can wait for that PR to be released preparing the code for it
> 
> Thanks for the additional context. Sounds like a good change, but let's wait for that feature before switching over please.

👍 

> Re container destruction: even if we can automatically destroy containers at the end of the run, this can already be done by simply calling `docker-compose down` from the Jenkins script. What does this function+environment variable improve on?

I added the variable as a way to keep the previous behavior. But I'm not married to this 😄, so I'm going to close it until we have the waitFor in the library side, and then we can discuss again about it.

Thank you @axw for your feedback!

PR #4449, opened 2020-11-20
Data streams improvements

axw:
## Motivation/summary

Separate data streams support from existing index management and selector, and extend config validation to prevent mixing data streams and existing index management config (ILM, pipelines, template setup, index names, etc.)

When the server is running under Agent, it is now necessary for data streams to be enabled. We should inject the config (`-E apm-server.data_streams.enabled`) from the program spec.

Finally, we introduce system tests to show that with data streams enabled, the server works when given only limited privileges like those given by Fleet.

## Checklist

~- [ ] documentation~ (later)
~- [ ] telemetry~

## How to test these changes

- Run the server with `apm-server.data_streams.enabled=true`, should index into `traces-*` etc., with no pipeline specified at index time.
- Run `apm-server setup` with `apm-server.data_streams.enabled=true`, should fail.
- Run `apm-server setup --pipelines` with `apm-server.data_streams.enabled=true`, should fail.

## Related issues

https://github.com/elastic/apm-server/issues/4378


jalvz:
> When the server is running under Agent, it is now necessary for data streams to be enabled

https://github.com/elastic/beats/pull/22689

systemtest/elasticsearch.go
```diff
@@ -90,11 +90,18 @@ func newElasticsearchConfig() elasticsearch.Config {
...
+		apmMetricsPrefix = "metrics-apm.*"
+		apmLogsPrefix    = "logs-apm.*"
+	)
 	requests := []estest.Request{
-		esapi.IndicesDeleteRequest{Index: []string{prefix}},
-		esapi.IngestDeletePipelineRequest{PipelineID: prefix},
-		esapi.IndicesDeleteTemplateRequest{Name: prefix},
+		esapi.IndicesDeleteRequest{Index: []string{legacyPrefix}},
+		esapi.IndicesDeleteDataStreamRequest{Name: apmTracesPrefix},
+		esapi.IndicesDeleteDataStreamRequest{Name: apmMetricsPrefix},
```
simitt, line 102: What about `apmLogsPrefix`?

simitt (commented):
`apm-server.ilm.*` configurations (including `apm-server.ilm.setup.enabled=false` and 
`apm-server.ilm.enabled=false`) are not allowed when using datastreams - but configuring `setup.template.*` is allowed and ignored. 
From a migration perspective it might be easier for users if their configurations for `ILM` and `templates` were still allowed but ignored (and a warning is logged). 

Apart from that - changes look good and on a quick test run work as expected.

systemtest/elasticsearch.go
axw: Nice catch, thanks.

axw:
> apm-server.ilm.* configurations (including apm-server.ilm.setup.enabled=false and
apm-server.ilm.enabled=false) are not allowed when using datastreams - but configuring setup.template.* is allowed and ignored.

Oops, that was an oversight.

> From a migration perspective it might be easier for users if their configurations for ILM and templates were still allowed but ignored (and a warning is logged).

That's a good point, I suppose it's a bit friendlier to log warnings instead. I'll update to do that.

axw:
I broke things by moving the ConditionalOverride stuff, looking into it now.

idxmgmt/supporter_factory.go
```diff
@@ -56,50 +61,111 @@ func MakeDefaultSupporter(log *logp.Logger, info beat.Info, configRoot *common.C
...
+}
+
+// defaultTemplateConfig customises template.DefaultConfig() with index settings.
+// This is only meaningful when data streams are not in use.
+func defaultTemplateConfig() template.TemplateConfig {
+	cfg := template.DefaultConfig()
+	cfg.Settings = template.TemplateSettings{
+		Index: map[string]interface{}{
+			"codec": "best_compression",
+			"mapping.total_fields.limit": 2000,
```
simitt, line 165:
```suggestion

			"mapping": map[string]interface{}{
				"total_fields": map[string]interface{}{
					"limit": 2000,
				},
			},
```

simitt: When running `./apm-server export template` two `mapping` entries would be produced otherwise

simitt (commented):
I believe the two failing system tests actually uncover a bug in the current behavior - the APM Server injected changes should not show up as config changes, but they should be part of the exported template (as they are now) - so changing the expectations to fix the tests would be fine.

jalvz:
should we update `docs/data/elasticsearch/generated` with the new dataset fields?

axw:
> I believe the two failing system tests actually uncover a bug in the current behavior - the APM Server injected changes should not show up as config changes, but they should be part of the exported template (as they are now) - so changing the expectations to fix the tests would be fine.

Yes indeed, you make a good point. I should have thought about that a bit harder. I managed to get it to work with config overrides, but it would be better if these defaults don't show up. I'm not sure how we can stop the logging overrides from showing up there. I'll dig a bit more...

axw:
> should we update docs/data/elasticsearch/generated with the new dataset fields?

@jalvz IMO we should hold off for now. The initial release will be very experimental, so probably better not to advertise so widely.

axw:
> Yes indeed, you make a good point. I should have thought about that a bit harder. I managed to get it to work with config overrides, but it would be better if these defaults don't show up. I'm not sure how we can stop the logging overrides from showing up there. I'll dig a bit more...

I'm going to pull this out into a separate PR.

idxmgmt/supporter_factory.go
axw: Ah thanks, I hadn't noticed that. I flattened it because `cfg.Unpack` is panicking when using a map. I'll dig into why that's happening.

axw: https://github.com/elastic/go-ucfg/issues/167

Deeper into the rabbit hole we go...

beater/beater.go
```diff
@@ -250,13 +244,40 @@ func checkConfig(logger *logp.Logger) error {
...
 func (bt *beater) registerPipelineCallback(b *beat.Beat) error {
+	if !hasElasticsearchOutput(b) {
+		bt.logger.Info("Output is not Elasticsearch: pipeline registration disabled")
+		return nil
+	}
+
+	if bt.config.DataStreams.Enabled {
+		bt.logger.Info("Data streams enabled: pipeline registration disabled")
+		b.OverwritePipelinesCallback = func(esConfig *common.Config) error {
+			return errors.New("index pipeline setup must be performed externally when using data streams")
```
jalvz, line 270: can this message be a bit more explicit? ie mention to install the apm package

cmd/root.go
```diff
@@ -52,6 +53,33 @@ var libbeatConfigOverrides = []cfgfile.ConditionalOverride{{
 				"rotateeverybytes": 10 * 1024 * 1024,
 			},
 		},
+	}),
+}, {
+	Check: func(cfg *common.Config) bool {
+		if eflag := flag.CommandLine.Lookup("E"); eflag != nil {
```
jalvz, line 59: can we add a todo and an issue in libbeat to add support for this (maybe eg via a `Transform(*common.Config) *common.Config`) ?

idxmgmt/datastreams.go
```diff
@@ -0,0 +1,71 @@
...
+// VerifySetup always returns true and an empty string, to avoid logging
+// duplicate warnings.
+func (dataStreamsManager) VerifySetup(template, ilm idxmgmt.LoadMode) (bool, string) {
+	// Just return true to avoid logging warnings. We'll error out in Setup.
+	return true, ""
+}
+
+// Setup will always return an error, in response to manual setup (i.e. `apm-server setup`).
+func (dataStreamsManager) Setup(template, ilm idxmgmt.LoadMode) error {
+	return errors.New("index setup must be performed externally when using data streams")
```
jalvz, line 70: can this message be a bit more explicit? ie mention to install the apm package

jalvz (approved):

cmd/root.go
axw: I'm going to revert this bit, but if I come up against it again then I will.

beater/beater.go
axw: I added `, by installing the 'apm' integration package` to the end of the error message.

idxmgmt/datastreams.go
axw: updated same as above

axw:
jenkins run the tests please

simitt:
Tested with BC1
* `export`, `setup` behave as described
* no pipelines are setup
* if `apm-server.data_streams.enabled: true` and `apm-server.ilm.setup: true` a message is logged:
```
{"log.level":"warn","@timestamp":"2020-12-23T21:26:42.393+0100","log.logger":"index-management","log.origin":{"file.name":"idxmgmt/supporter_factory.go","file.line":146},"message":"`apm-server.ilm` specified, but will be ignored as data streams are enabled","ecs.version":"1.6.0"}
```

Issue #4454, opened 2020-11-24
`make fmt` breaks tests/system code


axw:
`make fmt` likes to reorder imports in Python code, and in doing so breaks code in tests/system. See https://github.com/elastic/apm-server/pull/4432#discussion_r529235597

PR #4456, opened 2020-11-24
feat(ci): include checking Go/Python formats in the Intake stage of the pipeline

mdelapenya:
## Motivation/summary
We'd like to capture lint errors the soonest, as we have detected the code is not format-compliant with current tooling (see eea4611267520a381c7cb4485844d469547e033f: this commit applies the format suggested by `make fmt`).

~~For that reason we are adding a stage to run the lint process, which will invoke the existing tools (make fmt) and will fail the build if the process changed files (Go or python files).~~
For that reason we are reusing the `Intake` stage to run the lint goal, which will invoke the existing tools (make fmt) and will fail the build if the process changed files (Go or python files).

The new Make goal (`make lint`) will exit with 1 or 0 depending on the state of the `git status --porcelain` command.

## Checklist

- [ ] documentation
- [ ] telemetry

## How to test these changes
Run `make lint`. It will return 1 or 0 depending if the format process found unformatted files.

## Related issues


Jenkinsfile
```diff
@@ -86,6 +86,30 @@ pipeline {
...
+          expression { return params.intake_ci }
+          expression { return env.ONLY_DOCS == "false" }
+        }
+      }
+      steps {
+        withGithubNotify(context: 'Lint') {
+          deleteDir()
+          unstash 'source'
+          dir("${BASE_DIR}"){
+            sh(label: 'Run Lint', script: './.ci/scripts/lint.sh')
```
v1v, line 108: NIT:

What do you think a make goal with the script content? Something like the below snippet?

```
ci-lint: lint
  git status --porcelain || (echo "Lint detected files not following project's format. Please run 'make fmt' locally to fix this error"; exit 1) && echo 'Lint passed'
```

v1v (approved):

Jenkinsfile
mdelapenya: Sent in 5fd4f3d

kuisathaverat (approved):

cachedout (approved):

Jenkinsfile
```diff
@@ -104,7 +104,8 @@ pipeline {
         withGithubNotify(context: 'Lint') {
           deleteDir()
           unstash 'source'
-          golang(){
+          setEnvVar('GO_VERSION', readFile(file: "${BASE_DIR}/.go-version").trim())
```
mdelapenya, line 107: Is this still required in this stage? I see the env var already defined for another one here: https://github.com/elastic/apm-server/pull/4456/commits/8d090edcf7ef07a8af342b9fb669d8cfc0781b43#diff-e6ffa5dc854b843b3ee3c3c28f8eae2f436c2df2b1ca299cca1fa5982e390cf8L68

Makefile
```diff
@@ -230,6 +231,14 @@ gofmt: $(GOIMPORTS) add-headers
 autopep8: $(MAGE)
 	@$(MAGE) pythonAutopep8
 
+.PHONY: lint
+lint: fmt
+	@if [ -n "$(GIT_STATUS)" ]; then\
```
v1v, line 236: is GIT_STATUS set before running the goal?

v1v: I said nothing

mdelapenya: I think it's lazy evaluated, right?

mdelapenya:
I wonder if the modified imports are affecting how the tests behave... 🤔

axw:
> I wonder if the modified imports are affecting how the tests behave... thinking

Yes:  https://github.com/elastic/apm-server/issues/4454

Makefile
axw: `make check` (which calls `mage check`) already checks for uncommitted changes. This is redundant. I think what's missing is the `fmt` dependency.

Makefile
```diff
@@ -65,7 +66,7 @@ SYSTEM_TEST_TARGET?=./tests/system
 PYTEST_OPTIONS?=--timeout=90 --durations=20 --junit-xml=build/TEST-system.xml
 
 .PHONY: check-full
-check-full: update check golint staticcheck
+check-full: update check golint staticcheck lint
```
axw, line 69: This is going to break developers' workflows.
Currently, check-full will succeed as long as all code is OK. With this change, the code needs to be checked in as well. The developer then needs to check in the code before they can tell if it it's valid, which is backwards.

axw (changes requested):
I'm not really keen on having `make check-full` make modifications due to running `make fmt`. Preferably, `make check-full` would just check if running `make fmt` would require changes, without actually making them. Similar to `make check-goimports` in the Go Agent: https://github.com/elastic/apm-agent-go/blob/10035b7cb990c85ef0e61fc3aa6a7505b2556c5a/Makefile#L16-L17

Makefile
mdelapenya: I'll remove the lint target then, delegating it to check

tests/system/apmserver.py
```diff
@@ -1,3 +1,8 @@
+from kibana import Kibana
+from es_helper import index_smap, index_span, index_error, apm_prefix
+from es_helper import cleanup, default_pipelines
+from helper import wait_until
+from beat.beat import INTEGRATION_TESTS, TestCase, TimeoutError
```
axw, line 5: I've put up a fix for this import-reordering issue in https://github.com/elastic/apm-server/pull/4459

Makefile
mdelapenya: I added the dependency on `fmt` to `check`. Is that correct?

tests/system/apmserver.py
mdelapenya: I reverted my changes here in favour of yours, thanks!

mdelapenya:

Makefile
axw: I would prefer if "check" was non-modifying, but I think this is OK for now as it's fixing a hole we have at the moment. Ultimately I'd like to do something like the Go Agent does, checking if formatting is required without actually doing it: https://github.com/elastic/apm-agent-go/blob/master/scripts/check_goimports.sh)

axw (approved):

axw:
I've put up https://github.com/elastic/apm-server/pull/4654 to replace this PR, introducing non-mutating targets `check-fmt`, `check-gofmt`, and `check-autpep8`. `make check` will run the `check-fmt` target, which will exit with an error if there are differences.

Assuming my change goes in, this PR will be redundant. Closing it optimistically, will reopen as needed

PR #4434, opened 2020-11-17
systemtest.CleanupElasticsearch: refresh indices

axw:
## Motivation/summary

Prior to deleting the ILM policy, refresh indices to ensure the most recent changes are visible.

See https://github.com/elastic/apm-server/issues/4366#issuecomment-728676137. A little bit of a shot in the dark.

## Checklist

- [x] I have signed the [Contributor License Agreement](https://www.elastic.co/contributor-agreement/).
~- [ ] I have updated [CHANGELOG.asciidoc](https://github.com/elastic/apm-server/blob/master/CHANGELOG.asciidoc)~

I have considered changes for:
~- [ ] documentation~
~- [ ] logging (add log lines, choose appropriate log selector, etc.)~
~- [ ] metrics and monitoring (create issue for Kibana team to add metrics to visualizations, e.g. [Kibana#44001](https://github.com/elastic/kibana/issues/44001))~
- [x] automated tests (add tests for the code changes, all [**unit** tests](https://github.com/elastic/apm-server/blob/master/TESTING.md) pass locally)
~- [ ] telemetry~
~- [ ] Elasticsearch Service (https://cloud.elastic.co)~
~- [ ] Elastic Cloud Enterprise (https://www.elastic.co/products/ece)~
~- [ ] Elastic Cloud on Kubernetes (https://www.elastic.co/elastic-cloud-kubernetes)~

## How to test these changes

`cd systemtest && go test -v`

## Related issues

https://github.com/elastic/apm-server/issues/4366

PR #4501, opened 2020-12-09
[7.x] systemtest.CleanupElasticsearch: refresh indices (#4434)

jalvz:
 - systemtest.CleanupElasticsearch: refresh indices (#4434)


axw:
The test failure is because an older fix hasn't been backported, I'm on it

PR #3829, opened 2020-05-28
Ensure ECS compliant logging when enabled.

simitt:
## Motivation/summary
Move towards ECS compliant logging within the APM Server code by changing the log format and switching `logging.ecs` on by default. 

Log files created within the libbeat code might still not be ECS compliant. 

** Do not merge this PR until we are sure we want to backport to 7.9 **

## Checklist

~- [ ] I have signed the [Contributor License Agreement](https://www.elastic.co/contributor-agreement/).~
- [x] I have updated [CHANGELOG.asciidoc](https://github.com/elastic/apm-server/blob/master/CHANGELOG.asciidoc)

I have considered changes for:
- [x] documentation
- [x] logging (add log lines, choose appropriate log selector, etc.)
~- [ ] metrics and monitoring (create issue for Kibana team to add metrics to visualizations, e.g. [Kibana#44001](https://github.com/elastic/kibana/issues/44001))~
- [x] automated tests (add tests for the code changes, all [**unit** tests](https://github.com/elastic/apm-server/blob/master/TESTING.md) pass locally)
~- [ ] telemetry~
- [x] Elasticsearch Service (https://cloud.elastic.co)
~- [ ] Elastic Cloud Enterprise (https://www.elastic.co/products/ece)~
~- [ ] Elastic Cloud on Kubernetes (https://www.elastic.co/elastic-cloud-kubernetes)~

## How to test these changes
Start APM Server and observe log lines. Following changes should be applied by default:
* `logging.ecs` is enabled, therefore [ECS logs MVP fields](https://github.com/elastic/ecs-logging/tree/master/spec#minimum-viable-product) are added 
* `error` -> `error.message`
* `stacktrace` -> `error.stacktrace`
* `url` -> `url.original`
* `request_id` -> `http.request.id`
* `method` -> `http.request.method`
* `content_length` -> `http.request.body.bytes`
* `remote_address` -> `source.address`
* `user-agent` -> `user_agent.original`

## Related issues
implements https://github.com/elastic/apm-server/issues/3796

PR #4536, opened 2020-12-15
[7.x] Ensure ECS compliant logging when enabled. (#3829)

simitt:
 - Ensure ECS compliant logging when enabled. (#3829)


simitt:
package step failing on all PRs - looking into it in a separate step as it seems unrelated.

PR #4548, opened 2020-12-15
Set overwrite=true for various fields.

simitt:
This is required since the `fields.yml` files are now duplicated per event type because of the Elastic agent package restructure,
The index pattern generator requires to explicitly set overwrite:true for
duplicate fields.


docs/fields.asciidoc
```diff
@@ -43,6 +43,8 @@ Processor name.
 
 type: keyword
 
+{yes-icon} {ecs-ref}[ECS] field.
```
jalvz, line 46: The logic for discerning ECS fields in the docs is wrong, would you mind create an issue for that when you get a change?

jalvz (approved):
As a follow up we could add a test to ensure all fields have `overwrite:true`

simitt:
> As a follow up we could add a test to ensure all fields have overwrite:true

If they aren't the `areKibanaObjectsUpdated` test step actually fails - we just ignored it.

docs/fields.asciidoc
simitt: great catch!
https://github.com/elastic/apm-server/issues/4554

Issue #4503, opened 2020-12-10
Native OTLP intake


axw:
Add native [OpenTelemetry Protocol](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/protocol/otlp.md) (OTLP) intake support to APM Server.

We will piggyback on the work done for https://github.com/elastic/apm-server/issues/3984 to serve on the same port as used for receiving data from Elastic APM agents. Initially we will support OTLP/gRPC, later we can consider extending this to OTLP/HTTP.

The APM Server must be capable of receiving at least traces and metrics over OTLP, as we do with the [elastic opentelemetry-collector exporter](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/master/exporter/elasticexporter) and we should also consider extending this to receiving logs.

axw:
I think we're going to have to vendor `go.opentelemetry.io/collector` in the apm-server tree, renaming the packages (e.g. `apm-server/internal/otel_collector/...`). The gRPC services are all hidden in internal packages (`go.opentelemetry.io/collector/internal/data/...`).

axw:
Opened https://github.com/open-telemetry/opentelemetry-collector/issues/2395 to propose adding a few simple functions to opentelemetry-collector to keep things simple.

PR #4585, opened 2020-12-24
processor/otel: update tests to use Jaeger model

axw:
## Motivation/summary

Update the current tests to use the Jaeger data model, and use the OpenTelemetry Collector's translation code in the tests. This is a precursor updating the opentelemetry-collector module dependency.

This means the tests no longer have to reproduce certain translation behaviour implemented in OpenTelemetry Collector for Jaeger-specific functionality. e.g. the formatting for the "exporter version", which we parse to extract the language name and version. When we update the dependency, we will no longer be able to rely on "SourceFormat" to tell that data came from Jaeger. Instead, we can rely on properties set by the translator.

## Checklist

## How to test these changes

N/A. This is a non-functional change.

## Related issues

Required for #4503


jalvz (approved):

axw:
Test failure unrelated, fixed by https://github.com/elastic/apm-server/pull/4587

Issue #2975, opened 2019-11-27
Ensure that the Elasticsearch client in 8.0 can work with a 7.x elasticsearch instance


jalvz:
apm-server 8.0 needs to be compatible with the last 7.x.

Now that we have our own elasticsearch client, we need to ensure we use the right dependency in function of the stack version.

https://github.com/elastic/apm-server/pull/2974 needs a follow up so that, when a v8 agent is created, we ping Elasticsearch to retrieve the version, and if it is 7.x, we tear down the client, make a 7.x connection, and return it as an interface.

This should have no impact in the code other than the `elasticsearch` package.

graphaelli:
* Can go-elasticsearch support this directly?
* What happens during rolling upgrades, what version does the ES cluster report?
* How does libbeat handle this today?
* How do we test this?

jalvz:
> Can go-elasticsearch support this directly

Probably no, but will summon @karmi for confirmation

> What happens during rolling upgrades, what version does the ES cluster report

If there no breaking changes concerning the APIs we use, nothing happens.
If there are breaking changes regarding uploading sourcemaps or creating API keys, we would recommend users to not do those operations during a rolling upgrade.
If there are breaking changes in the `has_privileges` API, we could recommend to upgrade APM Server before Elasticsearch, and detect or recover from that specific error:
eg. if the path changes, the error would be a `404`, and we could trigger a client reconnect upon seeing it to get a fresh version. 
If the response format changes, the error would be an empty struct after JSON unmarshalling, and we could do the same. 

However I don't think that the `has_privileges` will break in 8.0, and even if it does I would tackle this problem when it comes.

> How do we test this

If we migrate Python tests to Go we will be in a better position to test that automatically.

simitt:
APM Server 8.0 is not supposed to work with an ES of version 7.x. Therefore this is not a concern and the issue can be closed.

PR #4587, opened 2020-12-24
fix: invalidating API Keys against ES master

simitt:
## Motivation/summary
The ES InvalidateApiKey API does not support param `id` anymore. Switch to using `ids` instead.

Related https://github.com/elastic/elasticsearch/pull/66671

Do NOT backport this to `7.x` as ES only introduced the `ids` parameter in `7.10`.
Ideally we would derive the used Elasticsearch version and adapt the parameter accordingly. There exists already an issue to [ensure ES client 8.0 can work with 7.x ES](https://github.com/elastic/apm-server/issues/2975), which would be a bigger change and needs more discussion how to tackle this. For now this fixes functionality and system tests in master.

## Checklist

## How to test these changes

* Run `apm-server apikey invalidate --id "123"` cmd against an `>=7.10` or `8.0` ES -> expected to work
* Run `apm-server apikey invalidate --id "123"` cmd against an `<7.10` ES -> expected to throw an error

### DIFF
--- a/cmd/apikey.go
+++ b/cmd/apikey.go
@@ -116,7 +116,7 @@ If neither of them are, an error will be returned.`,
 				// TODO(axw) this should trigger usage
 				return errors.New(`either "id" or "name" are required`)
 			}
-			return invalidateAPIKey(client, &id, &name, json)
+			return invalidateAPIKey(client, id, name, json)
 		}),
 	}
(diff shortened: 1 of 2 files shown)


simitt:
Merging this in to fix tests. @elastic/apm-server let me know in case you are not happy with the solution.

Issue #3796, opened 2020-05-14
beater/middleware: update fields to be ECS-compliant


axw:
We should update the logging middleware to use ECS-compliant field names and values:

https://github.com/elastic/apm-server/blob/c0f47670caa543889907501d1c26820311912991/beater/middleware/log_middleware.go#L34

"method" -> "http.method", "URL" -> "http.url", etc.

simitt:
Depends on https://github.com/elastic/beats/pull/18820

PR #4590, opened 2020-12-28
Switch to JSON logging by default

simitt:
## Motivation/summary
This PR changes the default for `logging.json=true`.

With https://github.com/elastic/apm-server/pull/3829 we switched to ECS compliant logging by default. Since the [ecs-logging-go-zap](https://github.com/elastic/ecs-logging-go-zap) is built on top of the JSON formatter, using it only makes sense in combination with logging in JSON. 

We should consider this a bugfix for an unreleased bug and backport to `7.11`. 

## Checklist

## How to test these changes
Start APM Server and observe logs are in JSON format by default. 

## Related issues
related to #3796


jalvz (approved):
I don't have a strong opinion, but this doesn't look like a bug fix to me

axw:
I think it's reasonable to consider this a bug, seeing as the original change won't work according to user expectations without this fix.

simitt:
Thanks @axw and @jalvz. As described above:
> We should consider this a bugfix for an unreleased bug and backport to 7.11.

I added it as point in the _Change_ section, as I consider it a bugfix for #3829 which will be also released with `7.11`.

jalvz:
Simply started with and without `-E logging.json=false`, it works as expected

PR #4602, opened 2021-01-11
processor/otel: update to go.opentelemetry.io

axw:
## Motivation/summary

Replace `github.com/opentelemetry/opentelemetry-collector` with `go.opentelemetry.io/collector`, and update to the new collector API. This is necessary for adding OTLP support.

A couple of notes on non-trivial changes:

- The boolean "error" attribute is no longer added to labels, as the translation from Jaeger to the internal data model drops the attribute and sets the span status.
- The "message" span log attribute will be removed during translation from Jaeger, and used to set the span event name instead.
- There's no longer any "SourceFormat" passed in, so we now rely on the exporter name to determine whether we're dealing with Jaeger data.

## Checklist

## How to test these changes

N/A (non-functional change)

## Related issues

Required for https://github.com/elastic/apm-server/issues/4503


processor/otel/metadata.go
```diff
@@ -0,0 +1,130 @@
...
+	})
+
+	if strings.HasPrefix(exporterVersion, "Jaeger") {
+		// version is of format `Jaeger-<agentlanguage>-<version>`, e.g. `Jaeger-Go-2.20.0`
+		const nVersionParts = 3
+		versionParts := strings.SplitN(exporterVersion, "-", nVersionParts)
+		if out.Service.Language.Name == "" && len(versionParts) == nVersionParts {
+			out.Service.Language.Name = versionParts[1]
+		}
+		if v := versionParts[len(versionParts)-1]; v != "" {
```
simitt, line 83: should this be `len(nVersionParts)-1` ?

axw: Nope.
 - `nVersionParts` is a constant (did you mean `nVersionParts-1?), and I believe this code is intended to cater for both `Jaeger-<version>` and `Jaeger-<language>-<version>`
 - This is existing code, just moved from consumer.go

simitt: yes I actually meant `nVersionParts-1`; 
theoretically if `exporterVersion` is `Jaeger` we would set the `out.Service.Agent.Version=Jaeger` - but since it is existing code, I assume that is theoretically and ok to keep as is.

simitt (approved):

PR #4618, opened 2021-01-13
 beater: run gmuxed gRPC server by default

### LINKED ISSUE #3984: Investigate serving Elastic APM and Jaeger inputs on the same port
Currently the server can be configured to listen for requests from Elastic APM agents and Jaeger agents, but only on separate ports. To enable simplified deployment, we should consider serving them on a single port -- at least as an option. To do that, we could use something like https://github.com/soheilhy/cmux

axw:
## Motivation/summary

Use gmux to run a gRPC server on the same port as the Elastic APM HTTP server, and register the Jaeger services by default.

When auth is enabled, it is necessary to configure Jaeger to send an agent tag called "elastic-apm-auth".
This cannot be changed.

## Checklist

~- [ ] Documentation has been updated~ Let's do this in a followup, after we have a discussion about the UX around auth tags.

## How to test these changes

- Run APM Server with default config
- Run `jaeger-agent` with `--reporter.grpc.host-port=localhost:8200`
- Exercise an app instrumented with the Jaeger client
- Check that Jaeger spans are indexed

## Related issues

Closes #3984


systemtest/apmservertest/server.go
```diff
@@ -418,7 +479,7 @@ func (s *Server) GetExpvar() *Expvar {
 	return expvar
 }
 
-func makeHTTPURLString(host string) string {
-	u := url.URL{Scheme: "http", Host: host}
+func makeHTTPURLString(scheme, host string) string {
```
simitt, line 482: nit: rename to `makeURLString`?

beater/server.go
```diff
@@ -89,6 +96,10 @@ func newServer(logger *logp.Logger, cfg *config.Config, tracer *apm.Tracer, repo
 	if err != nil {
 		return server{}, err
 	}
+	grpcServer, err := newGRPCServer(logger, cfg, tracer, reporter, httpServer.TLSConfig)
+	if err != nil {
+		return server{}, err
+	}
 	jaegerServer, err := jaeger.NewServer(logger, cfg, tracer, reporter)
```
simitt, line 103: Could we add a comment here and/or in the `jaeger` package that this is deprecated? Might be confusing for other devs to come accross the muxed gRPC server and the deprecated separate grpc server logic.

simitt (commented):
Looks great! Only left nit comments.

axw:
Thanks for the review @simitt! I've addressed your comments, and this PR is now ready for final review.

_meta/beat.yml
```diff
@@ -388,6 +388,12 @@ apm-server:
 
   # When enabling Jaeger integration, APM Server acts as Jaeger collector. It supports jaeger.thrift over HTTP
   # and gRPC. This is an experimental feature, use with care.
+  #
+  # WARNING: This configuration is deprecated, and will be removed in a future release.
```
simitt, line 392: @axw the plan is to remove these config options in `8.0` right?

axw: Yes. I've updated the comment to state that.

changelogs/head.asciidoc
```diff
@@ -14,3 +14,4 @@ https://github.com/elastic/apm-server/compare/7.11\...master[View commits]
 
 [float]
 ==== Added
+* Jaeger gRPC is now served over the same port as the Elastic APM agent protocol {pull}4618[4618]
```
simitt, line 17: Should we start putting a `Deprecated` section in the changelogs? It is implicit by this message, but might be more user friendly to point it out directly.

simitt (approved):

changelogs/head.asciidoc
axw: I think that's a good idea, will do.

axw: Done.

@bmorelli25 FYI I've added a "Deprecated" section in changelogs/head.asciidoc, please let us know if you have any concerns about that.

jalvz:
Tested with `jaeger-agent --reporter.grpc.host-port=localhost:8200` and a dummy jaeger app. Traces are shown in Kibana like our own!

PR #4635, opened 2021-01-19
script: simplify generate_notice.py by using go-licence-detector.

### LINKED ISSUE #3516: Update generate_notice.py to use go-licence-detector
The generate_notice.py script could be used by other Go projects - we should consider pulling it out into a separate repo to facilitate that.

I think we should also consider rewriting it in Go to remove the Python dependency (Go is naturally required). If we did that, it might make sense to contribute it to https://github.com/elastic/go-licenser.

axw:
## Motivation/summary

Outsource the heavy lifting to https://github.com/elastic/go-licence-detector, which is now used by Beats. We mostly use the beats-provided templates, but there is a small amount of customisation:
 - change product name from "Elastic Beats" to "Elastic APM Server"
 - add Beats itself to the list of licence overrides

Also, unlike Beats we do not use `go list -m all`. Instead, we (continue to) list only the packages which are linked into the final binary, and include their modules. This excludes numerous modules which are included only for tests.

## How to test these changes

- Run `make notice`, and check NOTICE.txt looks sensible
- Run `python3 scripts/generate_notice.py --csv=dependencies.csv`, and check `dependencies.csv` looks sensible (compare to previous version)

## Related issues

Closes #3516


simitt (commented):
Not sure that it is needed but I realized that with this change we are not showing the content of the dependencies' _NOTICE_ files anymore. Apart from that, I checked some samples of th the licenses - LGTM.

axw:
> Not sure that it is needed but I realized that with this change we are not showing the content of the dependencies' NOTICE files anymore. Apart from that, I checked some samples of th the licenses - LGTM.

I believe that I added that in error, and what ECK and Beats are doing is fine.

axw:
jenkins run the tests please

simitt (approved):

PR #4640, opened 2021-01-20
Revert "script: simplify generate_notice.py by using go-licence-detec…

simitt:
## Motivation/summary
 Revert "script: simplify generate_notice.py by using go-licence-detector. (#4635)"

This reverts commit d64e333 as the packaging seems to be broken.


jalvz (approved):
guess needs backport too?

PR #4646, opened 2021-01-21
[7.11] script: simplify generate_notice.py (#4644)

axw:
Backports the following commits to 7.11:
 - script: simplify generate_notice.py (#4644)


axw:
Don't merge until we're confident that #4644 hasn't broken release manager again.

Issue #4474, opened 2020-12-01
Use _doc_count data type when indexing histograms 


dgieselaar:
In the APM app, we are currently using `value_count` aggregations for transactions, as transaction metric documents use histogram fields, meaning `doc_count` is no longer representative. ES now also has a `_doc_count` data type that allows us to specify how many documents a transaction metric document represents. ES will then use this field in its bucketing aggregations, removing the need for an additional `value_count` aggregation, which should speed up some of our queries and remove some complexity from our queries.

https://www.elastic.co/guide/en/elasticsearch/reference/master/mapping-doc-count-field.html

axw:
I think this should be straightforward. I thought we would need to update betas and fields.yml; that is not the case.

To whomever implements this: I think we'll just need to add a field describing the doc count to `model.Metricset`, and use that to add a `_doc_count` field to the metric event docs that we index.

jalvz:
Checked `_doc_count` field in metrics matched the number of docs running the aggregation:

```
GET apm-7.12.0-transaction/_search
{
  "aggs": {
    "txnames": {
      "terms": { "field": "transaction.name" } 
    }
  }
}
```

PR #4647, opened 2021-01-21
Add a _doc_count field for single-histogram metricsets

axw:
## Motivation/summary

If a metricset is published with a single histogram, add a `_doc_count` field which is the sum of the histogram's counts. This will enable us to remove some `value_count` aggregations in the UI to determine the transaction count.

## Checklist

## How to test these changes

1. Enable transaction metrics aggregation
2. Send some transactions
3. Check that the docs have `_doc_count` in them
4. Perform aggregations over the docs (e.g. terms agg on `transaction.name`), and check that the `doc_count` value in each bucket matches the number of transaction events aggregated.

## Related issues

Closes https://github.com/elastic/apm-server/issues/4474


simitt (commented):
Code looks good and I tested the change - everything works as expected against ES `master`. 

Since `_doc_count` support was recently introduced, I also tested behavior against `7.9` ES. Writing the docs works fine, but running aggregations doesn't take the values into account. Since this impacts histogram functionality, I assume that Kibana would need to check for the ES version/indices when aggregating?

Screenshot 2021-01-25 at 15 24 13

Screenshot 2021-01-25 at 15 24 21

dgieselaar:
@simitt Kibana is only expected to work against ES versions of the same minor/major, so it's not expected to support ES 7.9 when running Kibana 7.12.

simitt:
Right! Perfectly fine then.

simitt (approved):

axw:
Sorry about the force-push, forgot I was working on a non-draft PR. About to merge anyway.

axw:
jenkins run the tests please

axw:
jenkins run the tests please

PR #4656, opened 2021-01-27
sampling: fix TestStorageGC

### LINKED ISSUE #4651: [test] Rework and reenable`TestStorageGC` test
The [TestStorageGC](https://github.com/elastic/apm-server/blob/master/x-pack/apm-server/sampling/processor_test.go#L447) test has been disabled with https://github.com/elastic/apm-server/pull/4626#issuecomment-766809925 as it was failing when running tests with `go test -race -v ./...`.

Re-enable a refactore test that is more resiliant when new fields are going to be added to events in the future.

axw:
## Motivation/summary

Fix TestStorageGC by disabling GC (by setting a long interval) while storing events to drive up the vlog size. Enable it only after multiple vlog files have been written. This ensures that GC doesn't remove files while we're still writing, causing the test to never have multiple vlog files.

Also, make the Badger value log file size configurable, so tests can override it with a smaller size.

## How to test these changes

`go test -race ./x-pack/apm-server/sampling`

## Related issues

Closes #4651

### DIFF
--- a/x-pack/apm-server/sampling/config.go
+++ b/x-pack/apm-server/sampling/config.go
@@ -76,6 +76,10 @@ type StorageConfig struct {
 	// TTL holds the amount of time before events and sampling decisions
 	// are expired from local storage.
 	TTL time.Duration
+
+	// ValueLogFileSize holds the size for Badger value log files.
+	// If unspecified, then the default value of 128MB will be used.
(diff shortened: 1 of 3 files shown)


simitt (approved):
Thanks for fixing that test!

PR #4658, opened 2021-01-27
Sourcemaps are datasets

jalvz:
## Checklist


apmpackage/apm/0.1.0/data_stream/sourcemaps/manifest.yml
```diff
@@ -0,0 +1,3 @@
+title: APM sourcemaps
+type: error
```
axw, line 2: Why "error"? Sourcemaps apply to span stacktraces too, and conceivably more in the future (e.g. logs).
However, this reminds me: IIANM, type must be one of "logs", "metrics", or "traces". Sourcemaps are none of these.

PR #4659, opened 2021-01-27
fix make fmt

jalvz:


axw (approved):
Thanks for fixing. One day I'll remember about macOS.

PR #4669, opened 2021-02-01
Rename dataset to avoid ambiguous index pattern generation

jalvz:
## Motivation/summary

Solves the issue described in https://github.com/elastic/kibana/issues/88307#issuecomment-770913428

While we could maybe find a way out with priority ordering, that would introduce unnecessary complexity. 

Happy to replace "user" with something more appropriate if there are better ideas. 

## How to test these changes

Wait for package 0.5.0 release

### DIFF
--- a/apmpackage/apm/0.1.0/data_stream/app_metrics/manifest.yml
+++ b/apmpackage/apm/0.1.0/data_stream/app_metrics/manifest.yml
@@ -1,5 +1,5 @@
 title: APM application metrics
 type: metrics
-dataset: apm.user
+dataset: apm.app
 dataset_is_prefix: true
-ilm_policy: metrics-apm.user_metrics-default_policy
+ilm_policy: metrics-apm.app_metrics-default_policy
--- a/apmpackage/apm/0.1.0/data_stream/app_metrics/manifest.yml
+++ b/apmpackage/apm/0.1.0/data_stream/app_metrics/manifest.yml
@@ -1,5 +1,5 @@
 title: APM application metrics
 type: metrics
-dataset: apm
+dataset: apm.user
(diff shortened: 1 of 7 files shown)


apmpackage/apm/0.1.0/data_stream/app_metrics/manifest.yml
```diff
@@ -1,5 +1,5 @@
 title: APM application metrics
 type: metrics
-dataset: apm
+dataset: apm.user
```
axw, line 3:
```suggestion
dataset: apm.app
```

The title of the stream is "app metrics" after all :)

axw (changes requested):
Thanks, I'd forgotten about that one.

Please delete the checklist section in the description.

simitt:
This PR fixes the apm issue; we still need to find a more generic answer to the question around how to avoid such conflicts right?

apmpackage/apm/0.1.0/data_stream/app_metrics/manifest.yml
jalvz: ha, I was 50/50 between "app" and "user"

axw (approved):

jalvz:
> This PR fixes the apm issue; we still need to find a more generic answer to the question around how to avoid such conflicts right?

There is no much about it, if one dataset name is prefix of another, it is going to be ambiguous. Maybe it can be validated in the package registry.

axw:
Confirmed with BC2 and epr-snapshot. All APM index templates have a prefix of `apm.` on the dataset:

image

Application metrics are being put into `metrics-apm.app.<service>-*` data streams:

image

Issue #4573, opened 2020-12-21
[Fleet] Support for APM Agent Central Config with zero configuration


axw:
Agent central config depends on APM Server querying Kibana, and having privileges as laid out in https://www.elastic.co/guide/en/apm/server/current/privileges-agent-central-config.html.

When running under Fleet, APM Server will by default not have the necessary privileges to query central config. We will have to either inject secondary credentials for querying Kibana, or somehow extend the privileges of the primary API Key provided by Fleet to APM Server.

Users should not need to care about configuring API Keys for this; they should rather be concerned about enabling or disabling the central config feature, which implicitly updates privileges.

**Alternatively:** we can reimplement APM Agent Central Config to use Fleet. We would need to:
 - Update the APM app in Kibana to store agent config in (and retrieve from) Fleet policies
   - This config would be pushed down to APM Server, which would store it in memory.
   - Updating APM Agent config must not cause APM Server to restart.
 - Add a means for APM Server to record status for each APM Agent config block (so we can see if the config has been applied)

simitt:
Created a Kibana issue to track work for this for Fleet https://github.com/elastic/kibana/issues/89311

axw:
Reopening and changing the description so we don't lose sight of enabling this feature without manually injecting API Keys.

axw:
> We will have to either inject secondary credentials for querying Kibana, or somehow extend the privileges of the primary API Key provided by Fleet to APM Server.

Fleet provides extension points for the package policy create and update APIs.

Information courtesy of Paul Tavares:

> API hooks available for Package policies (aka: integration policies): one for Create and another for Update. To use them, a dependency must be set on your kibana Plugin to Fleet, which will then expose `FleetStartContract` during the `start` phase of the plugin's lifecycle. This interface includes a method named `registerExternalCallback()` which allows you to register a callback function for the following:
> 
> fleetStart.registerExternalCallback('packagePolicyCraete', async () => {});
> and 
> fleetStart.registerExternalCallback('packagePolicyUpdate', async () => {});
> The Types for the callbacks supported are defined here: https://github.com/elastic/kibana/blob/a99ccc27d7a9d2ad7be162c21f545e8eebf59c69/x-pack/plugins/fleet/server/plugin.ts#L132-L148 

So what we should be able to do is have the APM app hook into these API calls, and inject one or more additional API Keys. This could be done without elastic/kibana#89311.

[...]

PR #4670, opened 2021-02-01
Wire kibana config from fleet

jalvz:
## Motivation/summary

Adds support for central config and sourcemaps in managed mode.

## Checklist

- [X] Update [CHANGELOG.asciidoc](https://github.com/elastic/apm-server/blob/master/CHANGELOG.asciidoc)
- [X] Tests

## How to test these changes

In the Kibana UI create an APM integration without (or with a wrong API key), and check that `curl http://localhost:8200/config/v1/agents\?service.name=foo` returns "Unauthorized". Then create a valid api key and update the apm policy with it. Repeat the same curl command, and this time should respond with service-does-not-exist error.

For sourcemaps: index a sourcemap, and check that APM Server can read it back when ingesting errors, provided the right API key has been configured in the policy editor.
 

## Related issues

Requires https://github.com/elastic/beats/pull/23856

Closes https://github.com/elastic/apm-server/issues/4573


jalvz:
Still missing tests, but opening this already for review

beater/config/config.go
```diff
@@ -40,18 +40,20 @@ const (
 )
 
 type KibanaConfig struct {
-	Enabled             bool `config:"enabled"`
+	Enabled             bool   `config:"enabled"`
+	APIKey              string `config:"api_key"`
 	kibana.ClientConfig `config:",inline"`
 }
 
 func (k *KibanaConfig) Unpack(cfg *common.Config) error {
```
axw, line 48: I think we can simplify this method to:

```go
func (k *KibanaConfig) Unpack(cfg *common.Config) error {
        type kibanaConfig KibanaConfig
        if err := cfg.Unpack((*kibanaConfig)(k)); err != nil {
                return err
        }
        k.Enabled = cfg.Enabled()
        k.Host = strings.TrimRight(k.Host, "/")
        return nil
}
```

That way we rely on the config tag for unpacking into APIKey.

beater/beater.go
```diff
@@ -280,7 +285,7 @@ type sharedServerRunnerParams struct {
 }
 
 func newServerRunner(ctx context.Context, args serverRunnerParams) (*serverRunner, error) {
-	cfg, err := config.NewConfig(args.RawConfig, elasticsearchOutputConfig(args.Beat))
+	cfg, err := config.NewConfig(args.RawConfig, args.KibanaConfig, elasticsearchOutputConfig(args.Beat))
 	if err != nil {
 		return nil, err
 	}
```
axw, line 291:
```suggestion
	cfg, err := config.NewConfig(args.RawConfig, elasticsearchOutputConfig(args.Beat))
	if err != nil {
		return nil, err
	}
	if args.KibanaConfig != nil {
		args.Kibana = *args.KibanaConfig
	}
```

Seeing as this is the only place we pass in the arg, can we just replace the Kibana config afterwards and avoid forcing all other callers to pass in `nil`? Alternatively we could not unpack `Fleet.Kibana` in the integration config, and merge it into the config in `serverCreator.Create`.

apmpackage/apm/0.1.0/agent/input/template.yml.hbs
```diff
@@ -2,3 +2,6 @@ apm-server:
     host: {{host}}
     secret_token: {{secret_token}}
     rum.enabled: {{enable_rum}}
+    kibana:
+        enabled: true
```
axw, line 6:
```suggestion
```

`enabled: true` is implied, it's not necessary to explicitly specify it.

axw (changes requested):
I'm pretty confused about what this PR is meant to be doing.

We discussed that the API Key passed used by Fleet would not have sufficient privileges, i.e. would not have the APM space privileges required for querying central config. IIANM that's why you added `apm-server.kibana` section to the integration package.

So why is https://github.com/elastic/beats/pull/23856 needed, and why is the server still looking at the Fleet config?

jalvz:
> We discussed that the API Key passed used by Fleet would not have sufficient privileges

And this doesn't take the `access_api_key` from Fleet indeed, just the Kibana connection information (host, etc). We need that, right?

The API Key comes from user input in the package, as we talked about.

apmpackage/apm/0.1.0/agent/input/template.yml.hbs
jalvz: Sure, I just liked to be explicit.

apmpackage/apm/0.1.0/manifest.yml
```diff
@@ -41,6 +41,11 @@ policy_templates:
         required: true
         show_user: true
         default: false
+      - name: kibana_api_key
+        type: string
```
simitt, line 45: should this be `type: password`?

axw:
> And this doesn't take the access_api_key from Fleet indeed, just the Kibana connection information (host, etc). We need that, right?

Ahhhh sorry, I see. Thanks for the explanation.

If https://github.com/elastic/beats/pull/23856 does go ahead: is it not possible to have the spec extract "fleet.kibana" and merge it into "apm-server.kibana"? Is that the bit you said you couldn't get to work? I suppose this approach is fine though.

Anyway, I see there's some contention on the Beats issue so let's wait until that's resolved?

jalvz:
is it not possible to have the spec extract "fleet.kibana" and merge it into "apm-server.kibana"? Is that the bit you said you couldn't get to work?

I tried a few things, I believe the problem with that one is that it is not possible to merge dicts on conflict; so we would lose either the api_key or the other fields.

apmpackage/apm/0.1.0/manifest.yml
jalvz: it appears there is no such a thing...

axw (approved):
LGTM. A couple of things I'd like to see:

 - A test for the new APIKey behaviour in kibana.ConnectingClient. I think this should be done before merging.
 - A system test for agent central config when running under Fleet, using the injected `fleet.kibana` config, as a followup.

apmpackage/apm/0.1.0/manifest.yml
simitt: Am I looking at the wrong place here https://github.com/elastic/package-spec/blob/a8fe73d82b25806f6b96582b99993a63b67ef130/versions/1/data_stream/manifest.spec.yml#L53?

axw:
I tested this with 7.12 BC2, and it doesn't seem to be working. I started the stack with apm-integration-testing (with https://github.com/elastic/apm-integration-testing/pull/1069 applied):

`./scripts/compose.py start 7.12 --bc=37f40745 --apm-server-managed --with-elastic-agent --package-registry-snapshot`

Then looking at the APM Server logs in the Fleet UI, I see logs like:

> [elastic_agent.apm_server][info] Kibana url: http://localhost:5601
> [elastic_agent.apm_server][error] failed to obtain connection to Kibana: fail to get the Kibana version: HTTP GET request to http://localhost:5601/api/status fails: fail to execute the HTTP GET request: Get "http://localhost:5601/api/status": dial tcp 127.0.0.1:5601: connect: connection refused. Response: .

jalvz:
sorry @axw , Beats backports not merged yet

axw:
Tested again with BC3, works well!

Issue #4492, opened 2020-12-07
[Fleet] Add service name to datasets


jalvz:
- ~Depends on [spec definition](https://github.com/elastic/package-spec/issues/64)~
- Depends on [kibana support](https://github.com/elastic/kibana/issues/88307)
- Should rely on [existing code to sanitize service name](https://github.com/elastic/apm-server/blob/master/datastreams/servicename.go)

axw:
Confirmed with BC2 and epr-snapshot, with opbeans-python and opbeans-go running:

`apm-integration-testing$ ./scripts/compose.py start 7.12 --bc=latest --apm-server-managed --with-elastic-agent --package-registry-url=https://epr-snapshot.elastic.co --with-opbeans-python --with-opbeans-go`

The "-" in the service names, which is a reserved character (data stream separator) is replaced with "_".

![image](https://user-images.githubusercontent.com/843579/109102005-d50f9280-7762-11eb-9b9f-41159754ea5f.png)

PR #4674, opened 2021-02-02
Add service name to dataset field

jalvz:
## Motivation/summary

Implement the indexing strategy for APM data

## Checklist

## How to test these changes

Install APM package, send data to apm-server, and observe that data streams are created containing the ingested service names, and no legacy indices are created

## Related issues

Closes https://github.com/elastic/apm-server/issues/4492
Requires https://github.com/elastic/kibana/pull/89870


model/metricset.go
```diff
@@ -199,15 +199,16 @@ func (me *Metricset) Transform(ctx context.Context, cfg *transform.Config) []bea
...
 		if isInternal {
 			// Metrics that include well-defined transaction/span fields
 			// (i.e. breakdown metrics, transaction and span metrics) will
 			// be stored separately from application and runtime metrics.
-			fields[datastreams.DatasetField] = InternalMetricsDataset
-		} else {
-			fields[datastreams.DatasetField] = AppMetricsDataset
+			dataset = InternalMetricsDataset
 		}
+		dataset += "." + datastreams.NormalizeServiceName(me.Metadata.Service.Name)
```
simitt, line 210: nit: Could you also use `fmt.Sprintf` here as everywhere else; it would only create one new string, instead of two.

simitt (approved):

PR #4675, opened 2021-02-02
Add note about tags vs labels to json schemas

basepi:
Tags were renamed to Labels on the agent side years ago, but it's not super obvious that labels go to `"tags"` in the intake API.

I'm open to better ways to word this addition -- the important thing is that now if you do a Find for `labels`, something actually shows up. :)

### DIFF
--- a/docs/spec/v2/error.json
+++ b/docs/spec/v2/error.json
@@ -507,7 +507,7 @@
           }
         },
         "tags": {
-          "description": "Tags are a flat mapping of user-defined tags. Allowed value types are string, boolean and number values. Tags are indexed and searchable.",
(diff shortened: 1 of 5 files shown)


simitt:
Thanks @basepi for the contribution, I like the idea. 
The `docs` are actually generated. You would need to actually add the description to the `Tags` definitions in the [modeldecoder](https://github.com/elastic/apm-server/blob/master/model/modeldecoder/v2/model.go) and then run `make update` which generates the docs. I am happy to help with that if you prefer.

basepi:
I'll do it! Beni mentioned that they were autogenerated but I wasn't sure from where. Thanks!

simitt (approved):
Thanks for the contribution!

There's one flaky test you ran into, feel free to merge once CI is green. After merging please create a backport PR for `7.x`.

PR #4698, opened 2021-02-08
[DOCS] Make tagged sections for breaking changes unique

lcawl:
This PR updates renames the tagged section for the 7.10 breaking changes, since otherwise it appears in both 7.10 and 7.11 docs.


bmorelli25 (approved):
Thanks! 🥇

PR #4710, opened 2021-02-10
Makefile: always use .go_version

axw:
## Motivation/summary

The Makefile runs various Go commands as part of variable expansions, for example to identify the beats module and to determine $GOOS. This only works if Go is installed on the machine already.

To better support machines that do not have Go installed, and to avoid running builds with mixed versions of Go, run `./script/run_with_go_ver` in the Makefile to force the use of the Go version defined in .go_version for all targets, shell invocations, and subprocesses. This ensures also ensures we can build on machines that do not have Go installed already.

## How to test these changes

`make`

## Related issues

None.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,12 @@
 # Enforce use of modules.
 export GO111MODULE=on
 
-GOOSBUILD=./build/$(shell go env GOOS)
+# Ensure the Go version in .go_version is installed and used.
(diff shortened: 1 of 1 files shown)


simitt (commented):
LGTM

axw:
Adding this to the test plan to verify that subsequent release-manager builds are built with the expected version of Go.

jalvz:
Checked printing the `runtime.BuildVersion` is a delve debugging session, works as expected.

Issue #4678, opened 2021-02-03
[OTLP] Document support for OTLP/gRPC on the standard port


axw:
When elastic/apm-server#4503 is fully implemented, APM Server will serve OTLP gRPC over the same port as the Elastic APM agent protocol (i.e. port 8200 by default).

~Like in https://github.com/elastic/apm-server/issues/4621, I think we should first have a discussion about how we want the UX to look around auth tags, and again I'll mark this issue blocked until we have a decision.~ Let's stick with what we've got for now, we can always introduce separate tags later.

cyrille-leclerc:
@bmorelli25 can we work together on this improvement of OTLP docs? I have prepared elements for this.

bmorelli25:
👍 Sure! Fair warning: this issue wasn't in our 7.12 docs planning, so it hasn't been prioritized. I'd like to wrap @EamonnTP into the conversations as well in case he is able to pick it up before me.

bmorelli25:
Hmmm. Ignore that comment. I mentioned it in https://github.com/elastic/observability-dev/issues/1279#issuecomment-749772370, but it never made it into the planning doc for some reason. Oh well. Between @EamonnTP and I, I'm sure we can get it done.

EamonnTP:
@bmorelli25 I can pick this up. @cyrille-leclerc I'll put some time in your calendar to discuss more.

PR #4711, opened 2021-02-10
processor/otel: extend support for OpenTelemetry semantic conventions

axw:
## Motivation/summary

Extend support for OpenTelemetry semantic conventions to cover what we support for traces in the OpenTelemetry Collector exporter, minus support for "exception" span events. These will be covered in a followup.

A couple of functional changes for the existing Jaeger support:
 - If the span type is not inferred, it will now be "app" rather than "custom". This aligns with the OpenTelemetry Collector exporter, and works better with the "time spent by span type" breakdown charts.
 - We no longer assume an outcome/result of "success" if the span didn't fail. We now default to "unknown", and only set "succcess" or "failure" if one of those outcomes is known, like we do in the OpenTelemetry Collector exporter and our own agents.

## Checklist

~- [ ] Documentation has been updated~ https://github.com/elastic/apm-server/issues/4678

## How to test these changes

1. Start apm-server
2. Run an OpenTelemetry-instrumented application, configured to export OTLP to APM Server's standard host:port.
3. Repeat step 2, exporting to an OpenTelemetry Collector with the elastic exporter.
4. Compare the data received directly by APM Server to the data sent by the exporter.

## Related issues

#4503


jalvz (approved):
Didn't run it, but LGTM. That's going to be a lot of fields for testing...

simitt:
Tested with BC 2 - works as expected.

Tested with the otel `elastic` exporter, otel `otel/elastic` exporter and directly exporting from a go app instrumented with the otel exporter. Stored information on transactions was the same.

PR #4722, opened 2021-02-11
Add support for consuming OTLP/gRPC metrics

### LINKED ISSUE #4503: Native OTLP intake

axw:
## Motivation/summary

Add support for consuming basic OTLP/gRPC metrics,
like we support in the OpenTelemetry Collector exporter.

We do not yet support distributions or summaries; these
will be added once the Elasticsearch enhancements we
are dependent on are available. If these are received, the
server will ignore them and increment a monitoring counter
to indicate that an unsupported metric was received and
dropped.

## Checklist

~- [ ] Documentation has been updated~ https://github.com/elastic/apm-server/issues/4678

## How to test these changes

1. Start apm-server with monitoring enabled
2. Run an application that is instrumented with OpenTelemetry to collect metrics (including some distributions and summaries), and configured to export OTLP to APM Server's standard host:port.
3. Repeat step 2, exporting to an OpenTelemetry Collector with the elastic exporter.
4. Compare the data received directly by APM Server to the data sent by the exporter.
5. Check the beats monitoring doc for apm-server, and check the value for `apm-server.otlp.grpc.metrics.consumer.unsupported_dropped` corresponds to the number of distribution and summary metrics sent

## Related issues

Closes #4503


processor/otel/consumer.go
```diff
@@ -65,9 +66,29 @@ const (
...
 }
 
+// ConsumerStats holds statistics about data consumption.
+type ConsumerStats struct {
+	// UnsupportedMetricsDropped records the number of unsupported metrics
+	// that have been dropped by the consumer.
+	UnsupportedMetricsDropped int64
+}
+
+type consumerStats struct {
```
jalvz, line 81: I am confused about this, why both `consumerStats` and `ConsumerStats` are needed?

processor/otel/metrics.go
```diff
@@ -0,0 +1,235 @@
...
+		c.convertInstrumentationLibraryMetrics(instrumentationLibraryMetrics.At(i), metadata, out)
+	}
+}
+
+func (c *Consumer) convertInstrumentationLibraryMetrics(in pdata.InstrumentationLibraryMetrics, metadata model.Metadata, out *model.Batch) {
+	var ms metricsets
+	otelMetrics := in.Metrics()
+	var unsupported int64
+	for i := 0; i < otelMetrics.Len(); i++ {
+		if !c.addMetric(otelMetrics.At(i), &ms) {
```
jalvz, line 84: instead of having a method that both mutates and argument and returns a value, can't we say `unsupported += otelMetrics.Len() - ms.Len()`?

processor/otel/metrics.go
```diff
@@ -0,0 +1,235 @@
...
+	}
+	if unsupported > 0 {
+		atomic.AddInt64(&c.stats.unsupportedMetricsDropped, unsupported)
+	}
+}
+
+func (c *Consumer) addMetric(metric pdata.Metric, ms *metricsets) bool {
+	switch metric.DataType() {
+	case pdata.MetricDataTypeIntGauge:
+		dps := metric.IntGauge().DataPoints()
```
jalvz, line 100: Maybe dumb question, but can't we make all these data types conform an interface defining `LabelsMap()`, `Timestamp()` and `Value()`?
 All these cases are the same

processor/otel/metrics.go
axw: No, because two metrics may end up in the same metricset. (`ms` is a collection of metricsets)

processor/otel/consumer.go
axw: `consumerStats` might change at any time, `ConsumerStats` is a snapshot. I'll add a comment.

processor/otel/metrics.go
axw: Yes, but:
 - That won't work when we add support for distributions and summaries
 - Ideally we would also support distinguishing doubles from ints, storing the latter in a `long` field type

I'd prefer to leave it as it is for now (this is also how it looks in opentelemetry-collector-contrib), and refactor it later if it's a pain.

axw:
jenkins run the tests please

jalvz (approved):

simitt:
Tested with BC 2; metrics are accepted and ingested as expected.

Regarding testing of distributions or summaries - how did you test that when implementing? I was instrumenting a go app, but the [otel go metrics implementation](https://github.com/open-telemetry/opentelemetry-go/blob/main/metric/metric.go) only supports counters so far.

axw:
@simitt it does support histograms and summaries. You need to define an aggregator in the metrics controller, and then use a ValueRecorder meter. Frankly speaking their metrics API is obtuse, so you can't be faulted for missing that :)

Here's how I tested histograms:

https://github.com/elastic/apm-server/blob/d0624d4a8785b11ec47df1ff91087ca704007110/systemtest/otlp_test.go#L174-L193

I believe you need to change `NewWithHistogramDistribution` to `NewWithInexpensiveDistribution` to aggregate the recordings as a summary instead.

simitt:
I was missing that one could use `simple.NewWithInexpensiveDistribution()`  - thanks.

Tested with BC3: 
* compared metrics for data sent via otel exporter and elastic exporter - no differences found
* works as expected when using the otel exporter to directly send data to APM Server
* when using the elastic exporter, no `beats_stats.metrics.apm-server.otlp.grpc.metrics.consumer.unsupported_dropped` are recorded (probably expected that this is not supported, but mentioning it due to the testing instructions above)
* `otlp` data are not indexed, as they are missing in the [template definition](https://github.com/elastic/elasticsearch/blob/master/x-pack/plugin/core/src/main/resources/monitoring-beats.json). This is probably fine as no aggregations are currently done on the data.

PR #4725, opened 2021-02-11
Use the apm-server version everywhere*

### LINKED ISSUE #4420: APM Server version not hooked properly
The version that ends in `beat.Info` comes from libbeat, not apm-server. Possibly a bug.

axw:
## Motivation/summary

We have been using the libbeat version in the root handler,
which may not be the same as the apm-server version since
we maintain our own version now. This PR injects the apm-server
version into the root handler so it can be returned.

In the publisher code, we had hard-coded the major version
number. Less problematic, but I have changed this to parse the
full version to extract the major version number.

In the Elasticsearch client we still use the libbeat version for
determining the major version number. I think this is OK because
we should never have different *major* version of libbeat in
APM Server, for the time being at least.

Finally: instead of using the libbeat version in integration tests,
use a hard-coded version of "1.2.3" so we don't have to change
approval files each time we bump the version. This should enable
us to automate the version bump.

## Checklist

~- [ ] Update [CHANGELOG.asciidoc](https://github.com/elastic/apm-server/blob/master/CHANGELOG.asciidoc)~ should have no visible impact on users, as we have kept libbeat up to date.

## How to test these changes

1. Run apm-server
2. Query http://localhost:8200/
3. Check that it responds with the apm-server version

## Related issues

Fixes #4420


axw:
> In the Elasticsearch client we still use the libbeat version for determining the major version number. I think this is OK because we should never have different major version of libbeat in APM Server, for the time being at least.

Having said that, it would be better if we passed the version in. That way we could better unit test the behaviour.

jalvz (approved):

PR #4728, opened 2021-02-14
Support for Node.js profiles

dgieselaar:
Implements support for Node.js profiles, by normalizing sample.count to samples.count and adding wall.microseconds as a supported value type (Node.js pprof profiles don't support CPU time, only wall time).

Maybe a little premature but seems like a low-risk change?

Would appreciate any guidance on what tests should be added (and where).

## Checklist


model/profile.go
```diff
@@ -58,7 +58,14 @@ func (pp PprofProfile) Transform(ctx context.Context, cfg *transform.Config) []b
 	valueFieldNames := make([]string, len(pp.Profile.SampleType))
 	for i, sampleType := range pp.Profile.SampleType {
 		sampleUnit := normalizeUnit(sampleType.Unit)
-		valueFieldNames[i] = sampleType.Type + "." + sampleUnit
+		// Go profiles report samples.count, Node.js profiles report sample.count.
+		// We use samples.count for both so we can aggregate on one field.
```
dgieselaar, line 62: Convention says that the first sample type is always the number of samples collected. However, for heap profiles this field is `alloc_objects.count`, which is explicitly aggregated on in fetcher.go (and passed to the pprof UI).

model/profile/_meta/fields.yml
```diff
@@ -619,6 +619,15 @@
               count: 1
               description: >
                 Amount of CPU time profiled, in nanoseconds.
+        
+        - name: wall
+          type: group
+          fields:
+            - name: us
```
axw, line 626: It would be nice to have a single base unit -- on-CPU time is recorded as `profile.cpu.ns`. In theory if we stored in nanoseconds then we could run into overflows when aggregating over a large cluster (hundreds) of service instances, over a significant time period (months to years).

Probably something to consider later, when we firm all of this up.

axw (commented):
Nice, thanks!

> Maybe a little premature but seems like a low-risk change?

Perfectly fine and low-risk indeed.

> Would appreciate any guidance on what tests should be added (and where).

For now I think it would be good enough to modify [`model/profile_test.go`](https://github.com/elastic/apm-server/blob/0c7bd2216fa9572c1c02924ba0199cc1598efcf0/model/profile_test.go#L48-L67), adding `sample.count` and `wall.microseconds` samples and checking they're normalised appropriately.

We can add more extensive functional tests later.

dgieselaar:
jenkins run the tests please

axw (approved):
LGTM! Can you please add an entry to changelogs/head.asciidoc? Then it's good to merge.

axw:
I'm going to remove this from the test plan since it's currently only intended for internal consumption.

PR #4729, opened 2021-02-15
sampling: require a default policy

### LINKED ISSUE #4583: Tail-sampling: require a catch-all/default policy
If you enable tail-sampling and send a transaction that do not match any of the policies, then APM Server will reject the transaction with the error "no matching policy". The reason behind this is to avoid sending every single transaction to Elasticsearch due to bad configuration.

We should instead enforce a catch-all/default policy (i.e. just a sample rate, no criteria). When we have a UI we can have a specific section for this policy, and perhaps add a warning if the user sets the default sample rate to something high.

axw:
## Motivation/summary

When tail-based sampling is enabled, require a default policy to be defined. A default policy is one with no criteria, only a sampling rate which applies to any traces that does not match any other policy.

Without a default policy, the default behaviour is to drop traces that do not match any policy. By requiring a default we ensure that this behaviour is intentional: either drop (sample_rate=0) or not (sample_rate>0).

## Checklist

~- [ ] Documentation has been updated~ (no docs yet)

## How to test these changes

Run apm-server with `-E apm-server.sampling.tail.enabled`, check that it exits with an error because no default policy is defined.

## Related issues

Closes #4583


simitt (approved):

axw:
Looks like an unrelated test failure.

axw:
jenkins run the tests please

PR #4731, opened 2021-02-15
Mage targets for  faster development workflow

jalvz:
## Motivation/summary

Adds script to speed up fleet-related development: 

`run_agent.sh` builds and packages an apm-server binary for Linux, creates a new Elastic Agent container based off the most recent image, mount-binds the apm-server package at `distributions/apm-server-8.0.0-linux-x86_64` in the `install` directory, connects to the `apm-integration-testing` network, and runs it.

This is useful when a developing a feature that needs to be tested while running under Fleet.


magefile.go
```diff
@@ -178,6 +185,168 @@ func Package() {
...
+
+	err = sh.Run("tar", "cvf", buildDir+".tar", buildDir)
+	if err != nil {
+		return err
+	}
+
+	err = sh.Run("gzip", buildDir+".tar")
+	if err != nil {
+		return err
+	}
```
axw, line 221:
```suggestion
	err = sh.Run("tar", "cvzf", buildDir+".tar.gz", buildDir)
	if err != nil {
		return err
	}
```

axw (commented):
Looks useful!

The main thing I'm not too happy about is that this adds dependencies to the main `go.mod` and consequently modifies `NOTICE.txt`. End-users shouldn't really be affected by this.

Could we do this as a couple of bash scripts instead? The `QuickPackage` target should be as simple as this:

```bash
#!/bin/bash
set -xe

version=$(mage QualifiedVersion)
builddir=build/distributions/apm-server-$version-linux-x86_64
mkdir -p $builddir

cp -f LICENSE.txt NOTICE.txt README.md apm-server.yml $builddir
GOOS=linux go build -o $builddir/apm-server ./x-pack/apm-server
tar cvzf $builddir.tar.gz $builddir
```

(Requires the addition of a mage target called "QualifiedVersion" to magefile.go for printing out `mage.BeatQualifiedVersion)`)

The `RunAgent` target looks like it should be doable by invoking the `docker` CLI, but I haven't pored over it so maybe I missed something.

jalvz:
> Could we do this as a couple of bash scripts instead?

That seems so much easier... 😅 🤦

axw (commented):
Much simpler :)
Looks good, please revert NOTICE.txt and go.sum changes.

jalvz:
doh, I thought `go mod tidy` would fix that. thanks

axw (approved):
Not sure why there are whitespace changes to NOTICE.txt -- probably should be reverted? Otherwise LGTM

jalvz:
> Not sure why there are whitespace changes to NOTICE.txt 

I don't know either, but since they are removed it seems more correct like this?

axw:
> I don't know either, but since they are removed it seems more correct like this?

Only if it doesn't break automation (doesn't seem to have? I'm surprised). I'm not bothered otherwise.

PR #4735, opened 2021-02-16
Disable sourcemap upload endpoint when data streams enabled

jalvz:
## Motivation/summary

If we don't do this, users will see a `202  - Accepted` response, but apm-server will fail to index sourcemaps due to lack of permissions. 

## Checklist

## How to test these changes

Run apm-server with data streams enabled and send a POST request to `http://localhost:8200/assets/v1/sourcemaps`.
The request should fail.


beater/api/mux.go
```diff
@@ -220,8 +220,9 @@ func rumMiddleware(cfg *config.Config, _ *authorization.Handler, m map[request.R
 func sourcemapMiddleware(cfg *config.Config, auth *authorization.Handler) []middleware.Middleware {
 	msg := "Sourcemap upload endpoint is disabled. " +
 		"Configure the `apm-server.rum` section in apm-server.yml to enable sourcemap uploads. " +
-		"If you are not using the RUM agent, you can safely ignore this error."
-	enabled := cfg.RumConfig.IsEnabled() && cfg.RumConfig.SourceMapping.IsEnabled()
+		"If you are not using the RUM agent, you can safely ignore this error. " +
+		"If you are running APM Server managed by Fleet, you need to upload Sourcemaps directly to Elasticsearch."
```
axw, line 224: We know whether or not APM Server is managed by Fleet - can we change the error message based on that knowledge? Saves the user trying to figure out whether source maps are explicitly disabled, or implicitly because data streams are enabled.

axw (changes requested):
Nice catch.

Just one thing with the error message. Generally I'd like to minimise cognitive overhead for users by tailoring the error message to the running configuration.

beater/api/mux.go
jalvz: I was just being lazy... you got me

axw (approved):
Thanks :)

axw:
Confirmed with BC2:

```
$ curl http://localhost:8200/assets/v1/sourcemaps -F sourcemap=@app.e2e-bundle.min.js.map -F bundle_filepath=/index.js -F service_name=service_name -F service_version=123
{
  "error": "forbidden request: When APM Server is managed by Fleet, Sourcemaps must be uploaded directly to Elasticsearch."
}
```

axw:
Also confirmed that the old message is returned from a standalone APM Server when RUM is disabled:

```
{
  "error": "forbidden request: Sourcemap upload endpoint is disabled. Configure the `apm-server.rum` section in apm-server.yml to enable sourcemap uploads. If you are not using the RUM agent, you can safely ignore this error."
}
```

PR #4763, opened 2021-02-17
 Update to elastic/beats@1b31c26

jalvz:
## Motivation/summary

Brings ECS 1.8.0


axw:
Can we update beats instead? That'll bump ecs to v1.8.0: https://github.com/elastic/beats/pull/23465. I think we should keep this in sync.

axw (approved):

PR #4623, opened 2021-01-14
beater: add support for config reloading

axw:
## Motivation/summary

Add support for reloading config, to handle Fleet policy updates without restarting the apm-server process.

## Checklist

- [x] Update [CHANGELOG.asciidoc](https://github.com/elastic/apm-server/blob/master/CHANGELOG.asciidoc)

## How to test these changes

1. Install apm integration package, assign to policy/agent
2. Wait for apm-server to be running and available
3. Update the policy (e.g. enable/disable RUM)
4. Check policy update takes effect without restarting anything

## Related issues

Closes #4376

PR #4787, opened 2021-02-18
[7.x] beater: add support for config reloading (#4623)

axw:
 - beater: add support for config reloading (#4623)


axw:
jenkins run the tests please

PR #4841, opened 2021-02-19
Minor updates to package

jalvz:
## Motivation/summary

* Adds changelog
* Update documentation to at least not say lies and mislead the poor documentation team
* Rearrange settings a bit


apmpackage/apm/0.1.0/manifest.yml
```diff
@@ -43,43 +43,36 @@ policy_templates:
         required: false
         show_user: true
         default: false
+      - name: api_key_limit
+        type: int
+        title: Maximum number of API Keys
+        description: Restrict number of unique API Keys per minute, used for auth between APM Agents and Server.
+        required: false
+        show_user: false
```
simitt, line 51: nit: The `api_key_limit` is in the _Advanced Settings_ section (`show_user:false`). I find it easier to read to first group the keys that are always shown and then group the advanced settings together.

simitt (approved):

apmpackage/apm/0.1.0/manifest.yml
jalvz: ah, ok, hadn't noticed that. i was following the logic of putting related things together (so this bellow "enable api key")

PR #4855, opened 2021-02-22
systemtest: rewrite some auth system tests in Go

axw:
## Motivation/summary

Rewrite most of the auth system tests in Go.

## How to test these changes

Non-functional change, run the tests.

## Related issues

None.


axw:
This also has the side benefit of speeding up the tests somewhat. The Python-based auth tests were taking about 15 seconds each, and the new combined one takes 5 seconds in total. I didn't look closely at why -- probably eliminted repeated setup/teardown.

jalvz (approved):

Issue #4664, opened 2021-02-01
Define a default environment to record on agent data


axw:
We would like to provide users with the option of defining the environment (`production`, `dev`, `testing`, etc.) at the APM Server level.

Except where the agent has been configured to specify the environment (e.g. through the `ELASTIC_APM_SERVICE_ENVIRONMENT` environment variable):
 - All data received from agents would have the server-defined environment assigned
 - Agents requesting central config would receive config for the server-defined environment

Issue #4468, opened 2020-11-30
Set default `service.environment` 


sorenlouv:
**Related**
 - https://github.com/elastic/apm-dev/issues/689

APM UI is planning to [remove](https://github.com/elastic/kibana/issues/77747) the "All environments" dropdown option, since mixing data between environments causes more harm than good. We are also going to remove the option "No environment defined". 
Lastly, we will default to showing services from "production". This default will be configurable and there will still be a dropdown in the UI to change environment

These changes means that services without environments won't show up any more. To solve this I suggest that `service.environment` is made required. Either by requiring agents to specify an environment or by filling it out automatically if empty. In this case I suggest defaulting to "production".

axw:
> Either by requiring agents to specify an environment or by filling it out automatically if empty. In this case I suggest defaulting to "production".

I think setting it in the server would be best. I'd like to avoid breaking compatibility with agents in 8.0 if possible, to minimise the pain of upgrading.

sqren:
@axw that sounds good to me 👍 
This would mean that the UI can expect `environment` to always have a value from 8.0, right?

axw:
> This would mean that the UI can expect environment to always have a value from 8.0, right?

@sqren I think at the latest, yes. @AlexanderWert mentioned that there was some ongoing discussion about this, about it possibly breaking central config. I'm not sure where that discussion is happening, or if it has been resolved...

felixbarny:
> This would mean that the UI can expect environment to always have a value from 8.0, right?

A notable exception would be if users have an older version of APM Server with a newer version of Kibana. Although, based on telemetry, most of users update APM Server and Kibana in tandem. Also, ES might contain data from older versions.

Introducing new required fields later on is painful. Any chance you can still treat the `service.environment` as optional?

sqren:
> A notable exception would be if users have an older version of APM Server with a newer version of Kibana

I don't think we officially support running two different majors of the stack (7.11 and 8.0 for instance).

> Any chance you can still treat the service.environment as optional

That's what we are doing currently and would like to move away from. So yes, we can keep doing it but I think we can provide a better user experience, and remove some technical debt by simplifying the environment selector to only concrete values.

sqren:
> Introducing new required fields later on is painful.

Btw. if we add a migration script that automatically fills out `service.environment` users shouldn't feel the pain.

felixbarny:
> I don't think we officially support running two different majors of the stack (7.11 and 8.0 for instance).

[...]

PR #4861, opened 2021-02-22
Introduce a configurable default service environment

axw:
## Motivation/summary

Introduce a configurable default service environment. This is recorded in events which have no service environment defined, and is used in agent central config when no environment is specified. There is no default default, so there is no change in behaviour unless the configuration is set.

## Checklist

## How to test these changes

1. Run apm-server configured with `apm-server.default_service_environment: abc123`
2. Send some transactions with environment set (to something else); the environment should be kept intact, and not overridden
3. Send some transactions with environment unset; the default environment should be set
4. Enable transaction aggregation, ensure all metrics have an environment associated
5. Enable central config, ensure that an agent with no environment set picks up config for the environment "abc123"

## Related issues

Closes https://github.com/elastic/apm-server/issues/4664.


model/modelprocessor/environment.go
```diff
@@ -0,0 +1,58 @@
...
+				t.Metadata.Service.Environment = s.DefaultServiceEnvironment
+			}
+		case *model.Metricset:
+			if t.Metadata.Service.Environment == "" {
+				t.Metadata.Service.Environment = s.DefaultServiceEnvironment
+			}
+		case *model.Error:
+			if t.Metadata.Service.Environment == "" {
+				t.Metadata.Service.Environment = s.DefaultServiceEnvironment
+			}
```
jalvz, line 54: should be exhaustive here? i prefer prefer that something fails loudly if eg. we add a new event type

apmpackage/apm/0.1.0/manifest.yml
```diff
@@ -62,6 +62,12 @@ policy_templates:
         required: false
         description: Enter as <Id>:<API Key>
         show_user: true
+      - name: default_service_environment
```
jalvz, line 65: not sure about this, shouldn't we use the `namespace` instead?

axw: I'd prefer to keep them separate for now, so users have the flexibility to have two dimensions:
 - a namespace is a collection of APM services, e.g. split by business applications, business unit, etc.
 - typical environments for those services: production, testing, development

jalvz: ah ok, that was news to me 👍

simitt: Please also update the [README](https://github.com/elastic/apm-server/tree/master/apmpackage/apm/0.1.0/docs#namespace) accordingly, it refers to using `namespace` for environment.

model/modelprocessor/environment.go
axw: Yeah, good point - it should be and it's not. I forgot to cover profiles.

I _could_ make this panic if it sees a type it doesn't know about, but I'd rather not have this feature cause such a catastrophic failure in case test coverage isn't quite adequate (which evidently it's not).

I think ideally we wouldn't be dealing with a "Transformable" interface here at all, but instead pass around the concrete `model.Batch` struct type. Then the processor can just iterate over the bits it cares about. We could have tests for processors that check the Batch type hasn't changed, if necessary.

I'd like to defer that for, and just add Profile in for now. WDYT?

simitt (approved):

simitt:
Apologies - I just approved, but didn't see your ongoing conversation.

apmpackage/apm/0.1.0/manifest.yml
axw: Just to be clear, setting namespace to environment is still a reasonable choice. Perhaps we should just soften the language a little, to point it out as an option but not necessarily state that it is recommended.

simitt: yeah, that's what I meant, the additional use case for `namespace` tha you explained above, should be added to the README.

model/modelprocessor/environment.go
jalvz: sounds good, thanks (create an issue, maybe?)

jalvz:
One more question, sorry for delay: this intended for 7.13, I suppose?

axw:
@jalvz yes, now labelled accordingly

model/modelprocessor/environment.go
axw: Opened https://github.com/elastic/apm-server/issues/4877

apmpackage/apm/0.1.0/manifest.yml
axw: done

jalvz:
> @jalvz yes, now labelled accordingly

I was going to say that in that case better hold onto merging to allow for package changes for 7.12 😬

axw:
@jalvz I suppose if that's needed we should be working off the 7.12 branch instead. We'll eventually version the package with the stack right? So bump the version on master/7.x to 0.2.0 and have 0.1.0 on 7.12?

axw:
@sqren no plans at the moment. I recall we discussed this (taking away "all" and "none") in the past, and decided that it wasn't doable at the time. I think maybe because of backwards compatibility? What are we going to do for existing data?

sqren:
>  I recall we discussed this (taking away "all" and "none") in the past, and decided that it wasn't doable at the time. I think maybe because of backwards compatibility? What are we going to do for existing data?

Yes, removing them now can be seen as a breaking change. Which is why I'd like to add a default environment asap so we can do it in ~18 months time (or whatever our data guarantees for APM data is).

We could add a migration script that updates documents to ensure they have an environment. But it requires a re-index of all data so I'm not sure how feasible it is.

simitt:
@sqren can you create an issue for this so we can continue the discussion, including what the default env should be set to? I am not certain setting a default on the server would be more meaningful than having a `none` or `all` in the UI.

sqren:
Should I re-open https://github.com/elastic/observability-dev/issues/1118 since it already contains all the background?

axw:
@sqren I think reopen https://github.com/elastic/apm-server/issues/4468, with anything that's changed since it was closed.

PR #4876, opened 2021-02-25
processor/otel: translate OpenTelemetry exceptions

### LINKED ISSUE #4712: [OpenTelemetry] Convert "exception" span events to errors
When processing OpenTelemetry traces, we should convert [exceptions](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/exceptions.md) to Elastic APM errors.

axw:
## Motivation/summary

Add support for translating OpenTelemetry exception span events to errors, like we do with Jaeger error span logs. This includes parsing Java stack traces. We can add support for parsing stack traces from other languages later.

Also, stop adding HTTP *span* context to errors; only do this for transactions, to be consistent with Elastic APM agents.

## Checklist

## How to test these changes

1. Run APM Server
1. Instrument a Java application with OpenTelemetry, point it at APM Server
1. Provoke the application into throwing an exception
1. Check that the exception is recorded as an error in Elasticsearch, and that the stack trace is parsed and recorded as individual frames.

## Related issues

Closes #4712


processor/otel/exceptions_test.go
```diff
@@ -0,0 +1,336 @@
...
+	... 1 more
+	Suppressed: java.lang.ArithmeticException: / by zero
+		at Junk.c(Junk.java:25)
+		... 3 more
+Caused by: LowLevelException
+	at Junk.e(Junk.java:37)
+	at Junk.d(Junk.java:34)
+	at Junk.c(Junk.java:21)
+	... 3 more`[1:],
+		),
```
simitt, line 108: Where did you get the format from? Did you instrument some java app with [open-telemetry/opentelemetry-java-instrumentation](https://github.com/open-telemetry/opentelemetry-java-instrumentation)? 

Not a java expert, so maybe the format is a very stable one, but I have following questions:
* assuming you tested with above mentioned library - it is not yet GA (`0.17.0`) - I believe we should add an end-to-end test with an instrumented JAVA app to ensure we realize any breaking changes. 
* is the exception format the same between JAVA versions?
* depending on the answers for above questions - we should ensure to be clear about what is supported in the docs.

processor/otel/consumer.go
```diff
@@ -640,16 +641,64 @@ func convertSpanEvent(
...
+				exceptionType = v.StringVal()
+			case "exception.escaped":
+				exceptionEscaped = v.BoolVal()
+			}
+		})
+		if exceptionMessage == "" && exceptionType == "" {
+			// `At least one of the following sets of attributes is required:
+			// - exception.type
+			// - exception.message`
+			return
```
simitt, line 677: This gets translated into `error.exception`, for which the elastic apm intake protocol requires a `message`. Below, if no `message` is present you set a default value - I think it is fair, to include as many exceptions from otel as possible. 

I am not clear on why you require a `type` or a `message` though? The requirement for a `message` or `type` on the elastic apm intake protocol only applies to `error.log` - but this is not set here.

processor/otel/exceptions_test.go
axw: The format is based on https://docs.oracle.com/javase/7/docs/api/java/lang/Throwable.html. I'll add a comment to that effect.

> is the exception format the same between JAVA versions?

AFAIK yes, perhaps @elastic/apm-agent-java can comment.

processor/otel/consumer.go
axw: The "At least one of..." comment comes from https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/exceptions.md. I'll be more explicit about that.

simitt (approved):

processor/otel/exceptions_test.go
axw: FWIW I checked the Java 15 docs and I don't see any significant changes.

simitt: thanks for checking

simitt:
tested with BC 4 - works as expected 
Screenshot 2021-05-12 at 15 34 32

PR #4886, opened 2021-03-02
Add error.grouping_name field

### LINKED ISSUE #4726: Keyword field for error group name
Currently we use `error.log.message` or `error.exception.message` to display the title for an error group. When we need a list of error groups, we use a terms aggregation on `error.grouping_key`, and then a `top_hits` aggregation to get `error.log.message` and `error.exception.message`. Using a top_hits aggregation like this is slow, esp if there are a lot of different error groups. The faster alternative, a top_metrics aggregation only works on numeric fields and keywords.

My suggestion is to add `error.grouping_name` as a keyword field, so we A) have a more consistent way of displaying error group names, and B) we can use the top_metrics aggregation where it makes sense.

axw:
## Motivation/summary

Add an ingest pipeline which adds an "error.grouping_name"
field to error documents. This is set based on the value
of "error.log.message", or "error.exception[0].message".
If an error contains both a log message and an exception,
then the log message will win.

## Checklist

## How to test these changes

1. Capture an exception.
2. Check that the error doc has an `error.grouping_name` field equal to the exception message.

## Related issues

Closes #4726


jalvz (approved):

simitt:
tested with BC-4:
* `error.grouping_name` == `error.exception.message` if only `error.exception.message` available
* otherwise `error.grouping_name` == `error.log.message`

PR #4908, opened 2021-03-03
Update packate category

jalvz:
## Motivation/summary

It will show up as `Elastic Stack`, more accurate than `Monitoring`.


axw:
Can it be both? I don't think monitoring is inaccurate.

axw (approved):

PR #4912, opened 2021-03-04
model: add system.network.{cpus,memory,network}

axw:
## Motivation/summary

Extend the system metadata with properties for describing the number of CPU cores, amount of system memory, and network information for end-user devices.

## Checklist

## How to test these changes

TODO

## Related issues

#3657

### DIFF
--- a/docs/spec/v2/metadata.json
+++ b/docs/spec/v2/metadata.json
@@ -394,6 +394,14 @@
             }
           }
         },
+        "cpus": {
+          "description": "CPUCores holds the number of CPU cores of the system on which the monitored service is running.",
+          "type": [
+            "null",
+            "integer"
+          ],
+          "minimum": 1
+        },
         "detected_hostname": {
           "description": "DetectedHostname is the hostname detected by the APM agent. It usually contains what the hostname command returns on the host machine. It will be used as the event's hostname if ConfiguredHostname is not present.",
           "type": [
@@ -466,6 +474,51 @@
                   "maxLength": 1024
                 }
               }
+            }
+          }
+        },
+        "memory": {
+          "description": "Memory holds the memory of the system on which the monitored service is running, in bytes.",
+          "type": [
+            "null",
+            "integer"
+          ],
+          "minimum": 1
+        },
+        "network": {
+          "description": "Network holds information about the network that the system is connected to. This is only relevant for browsers and mobile devices.  See https://developer.mozilla.org/en-US/docs/Web/API/NetworkInformation for more details on how these fields should be set and interpreted for browsers.",
+          "type": [
+            "null",
+            "object"
+          ],
+          "properties": {
+            "downlink": {
+              "description": "Downlink holds the effective bandwidth estimate in megabits per second.",
+              "type": [
+                "null",
+                "integer"
+              ]
+            },
+            "effectivetype": {
+              "description": "EffectiveType describes the effective network type as one of a discrete set of network types, e.g. 'slow-2g', '2g', '3g', or '4g'.",
+              "type": [
+                "null",
+                "string"
+              ]
+            },
+            "rtt": {
+              "description": "RTT holds the estimated effective round-trip time of the current connection, in milliseconds.",
+              "type": [
+                "null",
+                "integer"
+              ]
+            },
+            "savedata": {
+              "description": "SaveData reports whether or not reduced data usage has been enabled on the client device.",
+              "type": [
+                "null",
+                "boolean"
+              ]
             }
           }
         },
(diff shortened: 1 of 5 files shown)


axw:
@vigneshshanmugam here's a start on the new fields for RUM device info.

I started adding a new top-level "browser" field to metadata, but it occurred to me that this is probably also going to be relevant to mobile devices, and in that case there may not be any browsers involved. @bryce-b would we want to capture these properties for mobile?

Still not sure what to put the network info in the Elasticsearch schema. I'd appreciate any thoughts anyone may have.

vigneshshanmugam:
> here's a start on the new fields for RUM device info

This looks really good. 

> I started adding a new top-level "browser" field to metadata, but it occurred to me that this is probably also going to be relevant to mobile devices, and in that case there may not be any browsers involved.

Yeah very good point. Is there any harm keeping them in the top level as `connectioninfo` or `networkinfo`?

docs/spec/v2/metadata.json
```diff
@@ -469,6 +477,51 @@
...
+          "properties": {
+            "downlink": {
+              "description": "Downlink holds the effective bandwidth estimate in megabits per second.",
+              "type": [
+                "null",
+                "integer"
+              ]
+            },
+            "effectivetype": {
+              "description": "EffectiveType describes the effective network type as one of a discrete set of network types, e.g. 'slow-2g', '2g', '3g', or '4g'.  See https://developer.mozilla.org/en-US/docs/Web/API/NetworkInformation for more details.",
```
vigneshshanmugam, line 503: should we move this URL under `network` so all fields can be referenced from there.

axw: Yes, done.

axw:
>> I started adding a new top-level "browser" field to metadata, but it occurred to me that this is probably also going to be relevant to mobile devices, and in that case there may not be any browsers involved.
> 
> Yeah very good point. Is there any harm keeping them in the top level as connectioninfo or networkinfo?

Those are a little bit too generic for my tastes. In theory we might later want to enable backend services to record network information on behalf of their clients. If we did that it would make more sense to nest under `client`, like `client.network.*`

I just had a quick look and there has been a discussion about adding device fields to ECS (https://github.com/elastic/ecs/issues/448), but the conclusion was to use `host` for now.

I'm inclined to record the data as `network.*` in a way that won't collide with any ECS changes later on. We could capitalise the field names to avoid colliding, per https://www.elastic.co/guide/en/ecs/current/ecs-custom-fields-in-ecs.html#_capitalization

Thus in the ES docs we would end up fields like

- `network.EffectiveType`
- `network.RTT`
- `network.Downlink`
- `network.SaveData`
- ...

WDYT?

vigneshshanmugam:
> I'm inclined to record the data as network.* in a way that won't collide with any ECS changes later on. We could capitalise the field names to avoid colliding, per https://www.elastic.co/guide/en/ecs/current/ecs-custom-fields-in-ecs.html#_capitalization

Sounds good to me. We need to come up with a shorthand fields for the v3 spec and should be good to go.

vigneshshanmugam:
@axw Can we arrive on a consensus here and add support for these new RUM related fields now that we have already done the OTEL side https://github.com/elastic/apm-server/pull/5436

axw:
@vigneshshanmugam yes, it's just a matter of prioritisation.

For OTel we decided to stick with lower_snake_case names and so I think we should do the same here. We ended up with `network.connection_type` for OTel; would you be OK with using that name instead of effectivetype?

vigneshshanmugam:
> We ended up with network.connection_type for OTel; would you be OK with using that name instead of effectivetype?

Yes totally we can use the same naming scheme from OTEL. But for RUM, we are going to be using shortened fields for the V3 spec so it should not be a huge concern from the agent perspective. 

>yes, it's just a matter of prioritization.

I don't think we have any plans of adding the fields for 7.14 @drewpost Correct me if am wrong, But probably if we can include them as part of 7.15, that should be a good to include them as part of Exploratory view dashboard.

axw:
> Yes totally we can use the same naming scheme from OTEL. But for RUM, we are going to be using shortened fields for the V3 spec so it should not be a huge concern from the agent perspective.

That's fine, I was just talking about the field name in Elasticsearch. We can still come up with a shortened name for v3.

vigneshshanmugam:
> That's fine, I was just talking about the field name in Elasticsearch. We can still come up with a shortened name for v3.

Sounds good. ++ to keeping it as `network.connection_type` as it resonates pretty well to the https://developer.mozilla.org/en-US/docs/Web/API/NetworkInformation/effectiveType

axw:
This is very stale. It's unclear what the priority is, so let's come back to this when it is clearer. A design doc would be great.

Issue #4484, opened 2020-12-02
Translate "context canceled" errors to more meaningful error messages


axw:
When the server is unable to respond to an agent in a timely manner, it may respond with "context canceled" because the agent has set a request timeout. This is not a meaningful error message, and can be difficult to understand by non-Go developers. We should trap these errors and produce a more meaningful message.

PR #4918, opened 2021-03-04
display timeouts to user as "server timeout"

stuartnelson3:
## Motivation/summary

timeouts configured at the server level result in
the request context being canceled; translate this
to a "server timeout" message before
logging/sending to the client.

~~the actual message being sent and status code
could also be changed, if there's a better way to
communicate what's happening, or a more
appropriate status code, we can change it. i'm not
sure at which point in the request/response
lifecycle it's failing, so it could be a 408 or
503.~~

~~note: this feels like a very broad approach.
reading around online did not provide a concrete
way to check that the request had timed out. an
alternative method, such as setting a
context.WithTimeout on the request instead of the
server's ReadTimeout might give us more visibility
into the cause. my concern with the approach in
this PR is that some un-related error might cause
the request (and context) to be canceled, and then
we report this as a timeout.~~

EDIT:

the translation is now implemented as a middleware that is applied to all endpoints.

the clients receive a 503 service unavailable, with text "request timed out".

fixes https://github.com/elastic/apm-server/issues/4484

## Checklist

## How to test these changes

build this branch locally and run with `./apm-server -c apm-server.secured.yml -e 2>&1 | grep 'timed out'`

a user can build the code in this branch, and run `hey-apm -instances 5`, or whatever is an appropriate `-instances` count to overload the server, causing requests to timeout.

expected log lines should look something along the lines of:

```
{
  "log.level": "error",
  "@timestamp": "2021-03-05T10:56:57.775+0100",
  "log.logger": "request",
  "log.origin": {
    "file.name": "middleware/log_middleware.go",
    "file.line": 60
  },
  "message": "request timed out",
  "url.original": "/intake/v2/events?verbose=",
  "http.request.method": "POST",
  "user_agent.original": "hey-apm",
  "source.address": "127.0.0.1",
  "http.request.id": "f4fbb971-4d88-494f-90b9-bbc31ad3f7df",
  "event.duration": 30561586246,
  "http.response.status_code": 503,
  "error.message": "request timed out",
  "ecs.version": "1.6.0"
}
```


axw:
Looking good so far! Did you miss a change when you pushed? I don't see where you're mapping the error to a status code and message.

> note: this feels like a very broad approach.

I agree. There are at least a couple of reasons why context might be cancelled:
 - the client cancelled the request (e.g. due to a client-side timeout, or browser window being closed)
 - the server timed out due to read_timeout or write_timeout elapsing

It could also be that the server is shutting down, but I don't think we need to be too pedantic about that. On shutdown we cancel some contexts, but if those context cancelled errors do leak then we should just trap them and turn them into "server shutting down" errors.

I think we should return 503 Service Unavailable. It's probably safe to just have the error message say "request timed out". Even though in theory the client may have aborted the request, in practice I expect this would only occur due to a client-side timeout.

I think checking the error should go up higher in the stack, probably in a middleware. The timeouts are applied to all endpoints, even if they're more likely to occur in the event intake. I'd say ideally we would just use https://golang.org/pkg/net/http/#TimeoutHandler, but it doesn't really fit in with how we format responses, log errors, and so on. Maybe it can be used for inspiration though.

stuartnelson3:
@axw updated to use middleware. not sure what's up with CI, it looks like some sort of build failure? https://apm-ci.elastic.co/job/apm-server/job/apm-server-mbp/job/PR-4918/2/console

anyway, let me know how the code looks, or if you want more tests.

stuartnelson3:
ah, goimports ..

```
[2021-03-05T10:32:40.066Z] goimports differs:
[2021-03-05T10:32:40.066Z]  - beater/middleware/timeout_middleware.go
[2021-03-05T10:32:40.066Z]  - beater/middleware/timeout_middleware_test.go
```

beater/middleware/timeout_middleware.go
```diff
@@ -0,0 +1,46 @@
...
+	return func(h request.Handler) (request.Handler, error) {
+		return func(c *request.Context) {
+			h(c)
+
+			err := c.Request.Context().Err()
+			if errors.Is(err, context.Canceled) {
+				id := request.IDResponseErrorsTimeout
+				code := request.MapResultIDToStatus[id].Code
+
+				c.Result.Set(id, code, request.MapResultIDToStatus[id].Keyword, tErr.Error(), tErr)
```
axw, line 42:
```suggestion
				c.Result.SetDefault(request.IDResponseErrorsTimeout)
```

SetDefault internally maps the ID to HTTP status code and message.

axw (approved):
Looks good! Just one very minor issue.

For bigger things I'd want to see some kind of integration test that proves it's wired up correctly (e.g. timeout response set applied before logging), but for this change I think unit testing is sufficient.

axw:
Looks like you need to update `TestDefaultMonitoringMapForRegistry` in beater/request. The number of monitoring variables has increased due to the new result ID.

beater/middleware/timeout_middleware.go
```diff
@@ -0,0 +1,45 @@
...
+	tErr := errors.New("request timed out")
+	return func(h request.Handler) (request.Handler, error) {
+		return func(c *request.Context) {
+			h(c)
+
+			err := c.Request.Context().Err()
+			if errors.Is(err, context.Canceled) {
+				c.Result.SetDefault(request.IDResponseErrorsTimeout)
+				c.Result.Err = tErr
+				c.Result.Body = tErr.Error()
```
axw, line 41: IIANM, Result.SetDefault indirectly calls Result.Set, which sets these:

https://github.com/elastic/apm-server/blob/09cbdea568f8e43b91b197b4fc81e8662c88e9ce/beater/request/result.go#L194-L195

Was something failing that was fixed by making this change?

stuartnelson3: Kinda sorta. The body wasn't being set, and the error was being wrapped. My initial error was no longer being used so compilation failed, and then I read the code further and saw that it wasn't setting the body, and then I opted to set the error explicitly because part of the github issue was removing "context canceled" since it doesn't carry meaning for non-go devs.

https://github.com/elastic/apm-server/blob/09cbdea568f8e43b91b197b4fc81e8662c88e9ce/beater/request/result.go#L160-L162

https://github.com/elastic/apm-server/blob/09cbdea568f8e43b91b197b4fc81e8662c88e9ce/beater/request/result.go#L216-L222

We should probably set the body, otherwise it's `nil`, but we could not set the error and just have `context canceled: request timed out` as the error message. Thoughts?

axw:
@stuartnelson3 could you please create a new PR to update CHANGELOG.asciidoc?

axw:
Tested with 7.13.0-100107e6, looks good. I used apmbench to load up the server, and it eventually logged this error:

> ```{"log.level":"error","@timestamp":"2021-04-30T08:54:28.736Z","log.logger":"request","log.origin":{"file.name":"middleware/log_middleware.go","file.line":60},"message":"request timed out","url.original":"/intake/v2/events","http.request.method":"POST","user_agent.original":"elasticapm-go/1.11.0 go/go1.15.8","source.address":"172.19.0.1","trace.id":"b0a8b46e52277558ab9a680a52e01b93","transaction.id":"b0a8b46e52277558","http.request.id":"b0a8b46e52277558","event.duration":639792198,"http.response.status_code":503,"error.message":"request timed out","ecs.version":"1.6.0"}```

I did not observe any "context canceled" errors.

axw:
I also tested with `ELASTIC_APM_SERVER_TIMEOUT=500ms` in apmbench's environment, which triggered far more of these errors. They looked the same, as expected.

PR #4926, opened 2021-03-08
model: append events on transformation

axw:
## Motivation/summary

Rather than having each model type create a slice only to append it to another slice, append to that slice directly. Cuts out some unnecessary allocations.

Also, I have renamed and unexported the `Transform` methods on the event types to clarify that they are no longer transformed directly, but only ever through a `model.Batch`.

## How to test these changes

Non-functional change.

## Related issues

None.

### DIFF
--- a/model/batch.go
+++ b/model/batch.go
@@ -69,19 +69,19 @@ func (b *Batch) Len() int {
 func (b *Batch) Transform(ctx context.Context, cfg *transform.Config) []beat.Event {
 	events := make([]beat.Event, 0, b.Len())
 	for _, event := range b.Transactions {
-		events = append(events, event.Transform(ctx, cfg)...)
+		events = event.appendBeatEvents(cfg, events)
 	}
 	for _, event := range b.Spans {
-		events = append(events, event.Transform(ctx, cfg)...)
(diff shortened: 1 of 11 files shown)


axw:
Modifying `processor/stream.BenchmarkBackendProcessor` to transform batches:

```
name                                          old time/op    new time/op    delta
BackendProcessor/heavy.ndjson/NoRateLimit-12    21.7ms ±21%    21.3ms ±12%    ~     (p=0.796 n=10+10)

name                                          old speed      new speed      delta
BackendProcessor/heavy.ndjson/NoRateLimit-12  18.6MB/s ±18%  18.9MB/s ±12%    ~     (p=0.781 n=10+10)

name                                          old alloc/op   new alloc/op   delta
BackendProcessor/heavy.ndjson/NoRateLimit-12    17.1MB ± 0%    17.0MB ± 0%  -0.70%  (p=0.000 n=10+10)

name                                          old allocs/op  new allocs/op  delta
BackendProcessor/heavy.ndjson/NoRateLimit-12      164k ± 0%      162k ± 0%  -1.16%  (p=0.000 n=10+10)
```

jalvz (approved):

Issue #3565, opened 2020-03-27
Revise model transformation approach


axw:
Follow-on from https://github.com/elastic/apm-server/issues/3551

In our model transform code, we're currently heavily relying on [utility.Set](https://github.com/elastic/apm-server/blob/a00a2b0acfcbe0056aae078e14ddb47d212d73b3/utility/map_str_enhancer.go#L30) and friends, which has a couple of performance-related issues:
 - blindly copies all maps, even when using a single-use temporary map
 - relies on interfaces/type reflection, which is slower than type-specific calls

We also create a lot of maps even when they're ultimately not used due to being empty.

We should investigate extending our model transform utilities to cut down on heap allocations and unnecessary copying.

I created a quick hack branch to highlight some of the gains we can expect to achieve: https://github.com/axw/apm-server/pull/new/optimise-fields

```
$ benchstat /tmp/old.txt /tmp/new.txt
name                      old time/op    new time/op    delta
TransactionEventDecode-8    18.0µs ± 5%    11.0µs ±11%  -39.16%  (p=0.016 n=4+5)

name                      old alloc/op   new alloc/op   delta
TransactionEventDecode-8    16.8kB ± 0%    10.7kB ± 0%  -36.36%  (p=0.000 n=4+5)

name                      old allocs/op  new allocs/op  delta
TransactionEventDecode-8       125 ± 0%        88 ± 0%  -29.60%  (p=0.008 n=5+5)
```

axw:
What would be particularly nice is if libbeat accepted an interface for `beat.Event.Fields`, along the lines of:

```go
type Fielder interface {
    GetField(string) (interface{}, bool)
    SetField(string, interface{}) (bool)
}

// Recycler is an optional interface for Fielder to implement, to enable recycling memory.
type Recycler interface {
    Recycle()
}
```

Fielders could optionally implement [go-structform/gotype.Folder](https://github.com/elastic/go-structform/blob/73075d12bc3de37594f9eb259c7747b6b30ed70f/gotype/fold.go#L38) to customise libbeat/output encoding.

We could then create a generated ECS type structure which implements this interface, into which we transform our model objects. e.g.

```go
type Fields struct {
    Service ServiceFields
    Transaction TransactionFields
    ...
}

func (f *Fields) GetField(key string) (interface{}, bool) {
    var prefix, suffix string
    if i := strings.IndexRune(key, '.'); i >= 0 {
        prefix, suffix = key[:i], key[i+1:]
    }
[...]

PR #4927, opened 2021-03-08
Move business logic out of model transformation

axw:
## Motivation/summary

Move some business logic out of model transformation
into model processors (hostname, node name) and decoder
(page.* -> url & http). This moves us closer to being
able to generate code for model transformation.

When decoding, copy `page.url` to `url` and `page.referer`
to `http.request.referrer`. By doing this in the decoder,
we avoid further enshrining the page fields on model types.

When decoding span-specific service metadata, overlay it on
the event's Metadata.Service field rather than creating a
separate field and merging at transformation time.

System.DetectedHostname, System.ConfiguredHostname, and
Service.Node.Name are now set by model processors. This
removes the need for the System.Hostname() method.

Many approvals have changed because they do not go through
the full beater pipeline. Importantly though, system tests
and beater integration tests have not changed.

## How to test these changes

Non-functional change.

## Related issues


simitt (approved):
There's a lot going on in this PR, but great changes. I really like the `modelprocessors` you introduced.

axw:
> There's a lot going on in this PR, but great changes.

Indeed, sorry about that - I'll try to better in the future. I started out making small PRs and then got a bit carried away in this one.

> I really like the modelprocessors you introduced.

Glad to hear it!

axw:
Sorry about force-push, I merged badly and had to fix it.

jalvz:
nothing to test here, let me know i missed something

simitt:
@jalvz I was thinking of testing that the `hostname` information is still properly populated. Same applies to the `page.url` fields for which the refactoring has been applied.

PR #4942, opened 2021-03-10
systemtest: fix query for enrollment API Keys

axw:
## Motivation/summary

Fleet has apparently changed how enrollment API Keys should be queried, I suppose because of the move to store them in dedicated indices for Fleet Server. Update system tests to match.

## How to test these changes

cd systemtest && go test -v -run FleetIntegration


axw:
Trivial fix, merging without review.

Issue #4949, opened 2021-03-11
APM Server fails to resolve config variables when loading template setup config


axw:
**APM Server version** (`apm-server version`):

v7.11.0

**Description of the problem including expected versus actual behavior**:

When `setup.template.fields` is explicitly configured to something containing a variable, such as `${path.config}`, the server exits with an error such as:

> ERROR instance/beat.go:971 Exiting: unpacking template config failed: missing field accessing 'fields' (source:'/etc/apm-server/apm-server.yml')

The server should resolve `${path.config}` to the config path (e.g. `/etc/apm-server`) and load fields from there.

**Steps to reproduce**:

 1. Run `apm-server -E 'setup.template.fields="${path.config}/fields.yml"'`
 2. apm-server exits with the error message `Exiting: unpacking template config failed: missing field accessing 'fields' (source:'command line flag')`

PR #4950, opened 2021-03-11
idxmgmt: fix config merging

axw:
## Motivation/summary

Fix a bug related to merging user-defined and default `setup.template` config.

We were merging onto a new config, which meant root config attributes like `path.config` were no longer accessible. The fix is to merge defaults over the root config to ensure the root config attributes are still accessible, and then merge the root config back over that to ensure user-defined config takes precedence.

## Checklist

## How to test these changes

See https://github.com/elastic/apm-server/issues/4949

## Related issues

Fixes https://github.com/elastic/apm-server/issues/4949


axw:
jenkins run the tests please

idxmgmt/supporter_factory.go
```diff
@@ -153,24 +155,21 @@ func (cfg *IndexManagementConfig) logWarnings(log *logp.Logger) {
...
-	return out, nil
+	if configRoot == nil {
+		return defaultConfig, nil
+	}
+	// NOTE(axw) it's important that we merge onto the root config,
+	// due to how config variable resolution works; variables are
+	// resolved using the root of the left-most config in the merge.
+	// We merge the user-defined config back over the defaults to
+	// ensure they take precedence.
+	return common.MergeConfigs(configRoot, defaultConfig, configRoot)
```
simitt, line 174: This looks complicated - looking at the `common.MergeConfigs` code I couldn't figure out why we need the extra `configRoot`.

axw: Agreed, hence the NOTE :)
Is the comment unclear?

axw:
Just found a bug, moving back into draft while I fx.

idxmgmt/supporter_factory.go
axw: I tweaked the comment slightly, hopefully it's a bit clearer now.

simitt (approved):

PR #4954, opened 2021-03-15
beater: add pprof-related config

axw:
## Motivation/summary

Introduce hidden config for development and performance analysis:

 - apm-server.pprof.enabled (default: false)

   Adds the "/debug/pprof" endpoint to the server's
   main HTTP server. This can be used instead of the
   "--httpprof" flag, to enable working with a single
   port.

 - apm-server.pprof.block_profile_rate (default: 0)

   If >0, sets the block profile rate. See
   https://golang.org/pkg/runtime/#SetBlockProfileRate

 - apm-server.pprof.mem_profile_rate (default: 0)

   If >0, sets the memory allocation profile rate. See
   `MemProfileRate` at https://golang.org/pkg/runtime/#pkg-variables

 - apm-server.pprof.mutex_profile_rate (default: 0)

   If >0, sets the mutex profile rate. See
   https://golang.org/pkg/runtime/#SetMutexProfileFraction

These config do not support hot-reloading, and will only work when passed in the initial config. I'm not intending to document these, at least not now; they're not intended for end users.

## How to test these changes

1. Run apm-server without config, `http://localhost:8200/debug/pprof` should 404
2. Run apm-server with `-E apm-server.pprof.enabled=true`, generate load
   - `http://localhost:8200/debug/pprof` should return the pprof index page, and links (cmdline, goroutine, heap, etc.) should be valid
   - `go tool pprof http://localhost:8200/debug/pprof/heap` should yield a non-empty heap profile
   - `go tool pprof http://localhost:8200/debug/pprof/block` should yield an empty block profile, because no block profile rate has been configured
3. Run apm-server with `-E apm-server.pprof.enabled=true -E apm-server.pprof.block_profile_rate=1`, generate load
   - `go tool pprof http://localhost:8200/debug/pprof/block` should yield a non-empty block profile

## Related issues

None specifically.

### DIFF
--- a/beater/api/mux.go
+++ b/beater/api/mux.go
@@ -19,6 +19,7 @@ package api
 
 import (
 	"net/http"
+	"net/http/pprof"
 
 	"github.com/elastic/beats/v7/libbeat/beat"
 	"github.com/elastic/beats/v7/libbeat/logp"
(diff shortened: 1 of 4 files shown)


axw:
jenkins run the tests please

simitt (approved):

simitt:
tested with BC 4 - works as expected

PR #4975, opened 2021-03-22
Set client.ip for iOS agent

### LINKED ISSUE #4935: Apply geo-ip data to mobile trace data
Related: elastic/apm-server#2262
Due to platform constraints, the easiest way to capture geo-ip data is to use the host ip address available in the header data when connecting to the apm server. According to elastic/apm-server#2262 this will be `host.id`. For context, the iOS agent utilizes the otelp interface on the apm server.

axw:
## Motivation/summary

Introduce a new "ClientMetadata" gRPC interceptor which records information about the gRPC client. The logic was extracted from the logging interceptor, which now relies on the client metadata interceptor being installed before it.

Also, move the gRPC auth interceptor after logging and metrics so auth failures are logged and included in metrics.

## Checklist

## How to test these changes

1. Instrument an iOS application with https://github.com/elastic/apm-agent-ios, pointed at APM Server.
2. Check that event documents have a `client.ip` field which matches the IP of the iOS application's device.

## Related issues

Closes #4935


beater/interceptors/metadata.go
```diff
@@ -0,0 +1,83 @@
...
+		ctx = context.WithValue(ctx, clientMetadataKey{}, values)
+		return handler(ctx, req)
+	}
+}
+
+type clientMetadataKey struct{}
+
+// ContextWithClientMetadata returns a copy of ctx with values.
+func ContextWithClientMetadata(ctx context.Context, values ClientMetadataValues) context.Context {
+	return context.WithValue(ctx, clientMetadataKey{}, values)
```
stuartnelson3, line 67: for my own education: using `clientMetadataKey{}` provides us with an identifier to extract the values, but it also doesn't allocate because it's an empty struct?

beater/interceptors/metadata.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+// ClientMetadataFromContext returns client metadata extracted by the ClientMetadata interceptor.
+func ClientMetadataFromContext(ctx context.Context) (ClientMetadataValues, bool) {
+	values, ok := ctx.Value(clientMetadataKey{}).(ClientMetadataValues)
+	return values, ok
```
stuartnelson3, line 73: any reason you're assigning to variables instead of `return ctx.Value(...)`?

stuartnelson3 (approved):

stuartnelson3:
>Instrument an iOS application with https://github.com/elastic/apm-agent-ios, pointed at APM Server.

is this something that's easy to do? or is it requiring running some app in xcode etc etc

beater/interceptors/metadata.go
axw: Because that doesn't compile :)
`return ctx.Value(...)` would compile, but the type assertion requires assignment to intermediate variables.

beater/interceptors/metadata.go
axw: Using an unexported struct type as the key is recommended by the Go docs (https://golang.org/pkg/context/):

>    // A key identifies a specific value in a Context. Functions that wish
>    // to store values in Context typically allocate a key in a global
>    // variable then use that key as the argument to context.WithValue and
>    // Context.Value. **A key can be any type that supports equality;
>    // packages should define keys as an unexported type to avoid
>    // collisions.**

axw:
>> Instrument an iOS application with https://github.com/elastic/apm-agent-ios, pointed at APM Server.
>
> is this something that's easy to do? or is it requiring running some app in xcode etc etc

I understand that it currently requires us to use a mac, I suppose running xcode. This will probably fall on @bryce-b to test.

simitt:
@bryce-b tested this with an instrumented iOS app. https://github.com/elastic/apm-server/issues/5247 was created as a follow up task.

Issue #4972, opened 2021-03-18
Service view with OpenTelemetry traces report empty instance node names and miss dependencies


cyrille-leclerc:
**APM Server version** (`apm-server version`):

7.12-SNAPSHOT

```
apm-server version 7.12.0 (amd64), libbeat 7.12.0 [cc08062860651282026f807dc63ac3ab167a2da8 built 2021-03-16 10:00:17 +0000 UTC]
```

**Description of the problem including expected versus actual behavior**:

The APM Service view when displaying OpenTelemetry traces is 
* missing dependencies even though dependencies are found on the service map and looking at the distributed traces
* missing "time spend by span time" (we are already aware of this)
* missing the name of the instances

![image](https://user-images.githubusercontent.com/459691/111699824-e23b1100-8838-11eb-95d4-e2f63a058f3e.png)

![image](https://user-images.githubusercontent.com/459691/111700416-aa809900-8839-11eb-9b98-ee12515b4dfc.png)

**Steps to reproduce**:

Example application
https://github.com/cyrille-leclerc/my-shopping-cart/tree/open-telemetry

```
{
  "_index": "apm-7.12.0-transaction-000001",
  "_type": "_doc",
  "_id": "IS84R3gB9QfJzqEW8jr3",
  "_version": 1,
  "_score": null,
  "_source": {
    "parent": {
      "id": "fbdd0f9589210eaa"
    },
    "process": {
      "pid": 9387
    },
    "agent": {
      "name": "opentelemetry/java",
      "version": "1.0.1"
    },
    "source": {
      "ip": "127.0.0.1"
    },
    "processor": {
      "name": "transaction",
      "event": "transaction"
    },
    "url": {
      "path": "/api/orders",
      "original": "http://localhost:8080/api/orders",
      "scheme": "http",
      "port": 8080,
      "domain": "localhost",
      "full": "http://localhost:8080/api/orders"
    },
    "labels": {
      "process_runtime_description": "AdoptOpenJDK OpenJDK 64-Bit Server VM 11.0.10+9",
      "thread_id": 36,
      "thread_name": "http-nio-8080-exec-1",
      "os_type": "DARWIN",
      "process_executable_path": "/Library/Java/JavaVirtualMachines/adoptopenjdk-11.jdk/Contents/Home:bin:java",
      "process_runtime_name": "OpenJDK Runtime Environment",
      "os_description": "Mac OS X 10.16",
      "service_namespace": "com-shoppingcart",
[...]

PR #4976, opened 2021-03-22
processor/otel: set default RepresentativeCount

axw:
## Motivation/summary

When translating OpenTelemetry spans to transactions/spans,
set RepresentativeCount to 1. This is incorrect when sampling
is enabled, but it's better to have incorrectly scaled metrics
than none at all, which is what happens when RepresentativeCount
is zero.

Until/unless some tag along the lines of Jaeger's sampler.param
is added, users will have to mentally scale down the metrics
in proportion to the sampling rate, to find the actual event counts.

## Checklist

## How to test these changes

1. Run an OpenTelemetry-instrumented application which sends HTTP requests to another service
2. Configure to send data to APM Server, capture some requests and wait a minute or so (the aggregation interval).
3. Check the service map looks right
4. Check the service overview dependencies list has the expected entries

## Related issues

Closes https://github.com/elastic/apm-server/issues/4972


stuartnelson3 (approved):

simitt:
Tested with BC4; with following setup:
*  `serviceB` : elastic-apm instrumented, calls `service-direct-otel`
* `service-direct-otel`: otel instrumented, calls `auth` service
* `auth` : elastic-apm instrumented, does not make external calls

Observed:
* the service map looks ok
Screenshot 2021-05-12 at 12 13 31

* the dependencies of `service-direct-otel` look ok
Screenshot 2021-05-12 at 12 13 02

* the service map for `service-direct-otel` shows `localhost:4001` instead of the service name `auth` - @axw is that expected?
Screenshot 2021-05-12 at 12 13 50

* `serviceB` shows no dependencies - I believe this is 
Screenshot 2021-05-12 at 12 12 52

axw:
> the service map for service-direct-otel shows localhost:4001 instead of the service name auth - @axw is that expected?

Doesn't seem expected to me. I wouldn't expect the downstream to change based on the starting point for the service map. Maybe a UI bug? Can you share the docs, or code?

> serviceB shows no dependencies - I expected to see the service-direct-otel as dependency

Agreed... can you see any `metricset.name: service_destination` metrics for `service.name: serviceB`?

simitt:
> Agreed... can you see any metricset.name: service_destination metrics for service.name: serviceB?

simitt:
>  Can you share the docs, or code?

That is the relevant bit:
```
tracer := global.Tracer(tracerName)

	// send request to auth service
	if err := tracer.WithSpan(r.Context(), "Check Auth", func(ctx context.Context) error {
		// add some random wait time
		<-time.After(time.Duration((rand.Intn(5)+2)*100) * time.Millisecond)

		client := http.Client{Transport: othttp.NewTransport(http.DefaultTransport)}
		req, _ := http.NewRequestWithContext(ctx, http.MethodGet, "http://localhost:4001/auth", nil)
		req.Header.Set("Authorization", r.Header.Get("Authorization"))
		resp, err := client.Do(req)
		if err != nil {
			return err
		}
		body, err := ioutil.ReadAll(resp.Body)
		if resp.StatusCode >= 400 {
			return errors.New(string(body))
		}
		return err
	}); err != nil {
		http.Error(w, err.Error(), http.StatusUnauthorized)
		return
	}
```

axw:
OK. So `destination.service.resource` will be `http://localhost:4001`, taken from the URL. I don't know why the service map would be different depending on the starting point. Seems likely to be a UI bug to me.

> No not for serviceB

Can you share the code for serviceB?

simitt:
Relevant bits for serviceB:
```
var client = apmhttp.WrapClient(http.DefaultClient)

func handler(w http.ResponseWriter, r *http.Request) {
	req, _ := http.NewRequestWithContext(r.Context(), http.MethodGet, "http://localhost:4002/user", nil)
	req.Header.Set("Authorization", r.Header.Get("Authorization"))
	resp, err := client.Do(req)
	if err != nil {
		apm.CaptureError(r.Context(), err).Send()
		http.Error(w, err.Error(), 400)
		return
	}
	body, _ := ioutil.ReadAll(resp.Body)
	if resp.StatusCode >= 400 {
		err := errors.New(string(body))
		apm.CaptureError(r.Context(), err).Send()
		http.Error(w, err.Error(), resp.StatusCode)
		return
	}
	doSomething()
	w.Write(body)
}
```

axw:
@simitt I can't see anything wrong with that. I just realised that having no service_destination metrics for serviceB can't have anything to do with OpenTelemetry, as you're using the Go Agent for serviceB. So I don't think it's relevant, but it would be good to get to the bottom of it. If you can share the full code I can try and investigate further.

simitt:
Right, that should be unrelated, let's follow up on that offline. The opentelemetry related changes can be considered to be working as expected.

Issue #4138, opened 2020-09-03
Test/investigate Elasticsearch connection backoff improvements


axw:
The Elasticsearch libbeat output has configuration for backing off on network errors. We should implement some tests (perhaps manual, but documented and repeatable) to verify that the server will not cause retry storms when Elasticsearch is unavailable.

graphaelli:
Is the intention to allowlist this on ESS based on the results?  If not, can we include that as a goal?

axw:
My intention is to make the default behave well enough that user intervention should be unnecessary, and make adjustments to code as necessary to get there. We can allow list existing config anyway in the interim.

stuartnelson3:
closed by #4991

PR #4981, opened 2021-03-22
test backoff time across different runs

stuartnelson3:
The idea behind this test is that, given there's
randomness in the jitter, across N requests
and a large enough max time, there should be an
appreciable variance between total time required
to execute N requests because of the jitter.

Whether or not that is a good idea, or if someone
has a better way to test for randomness, is up to
the reader.

The tests were parallelized so that multiple
backoff cycles could be tested at the same time
without hitting the 10min golang test timeout, and
also just because who wants to wait that long.

For reference, the current test setup has
completed full runs in:
- [5m35.752209718s 6m37.949983435s 6m52.242817354s]
- [5m2.881926583s 5m27.0031564s 5m49.786217507s]

## Motivation/summary

cf. https://github.com/elastic/apm-server/issues/4138

## Checklist

## How to test these changes

You can add log statements to the test to verify the different request timings, or run the test.


axw (commented):
The test looks good, but I'm not convinced it'll pay its weight given how long it takes to run. I'm not sure what the alternatives are though.

For any automated tests we add to the apm-server repo, I'd prefer they focus on what we control in apm-server code. Testing of the underlying libbeat bits should live in the beats repo. I'd be happy if we just made this a manual testing exercise.

The things I'm most interested in knowing are:
 - What happens when we vary `output.elasticsearch.worker`? Do we know that they share a client, and the backoff is coordinated? Or independently calculated for each worker?
 - Is `max_retries` being honoured? (I recall @simitt finding that it is not, with the way we use libbeat now.)
 - Do backoff and retries behave the same when the server is inaccessible, as opposed to being accessible but returning errors.
 - Do the non-libbeat output Elasticsearch clients in APM Server honour the configured backoff? These are: API Key auth lookups, tail-based sampling.

stuartnelson3:
>The test looks good, but I'm not convinced it'll pay its weight given how long it takes to run. I'm not sure what the alternatives are though.

yeah, as we both kind of mentioned in standup. i don't think it should be run as an automated thing, but then if it's manual I (pessimistically perhaps) don't think it will ever be run.

>What happens when we vary output.elasticsearch.worker? Do we know that they share a client, and the backoff is coordinated? Or independently calculated for each worker?

each client is independent (as checked by logging each time a new client + backoff gets created per worker), they do not share the underlying backoff. this seems to be in line with go saying that http client's should be reused since the underlying transport stores state, and new clients will be independent. As long as it's a single ES cluster being communicated with, seems like re-using a single client would make sense.

based on https://www.elastic.co/blog/how-to-tune-elastic-beats-performance-a-practical-example-with-batch-size-worker-count-and-more, since multiple clients more or less doubles throughput in the example, it appears there's some sort of internal coordination that prevents a single client from sending multiple requests at once. as an "outsiders" note, letting the go runtime map goroutines to threads seems like a fine way to handle increasing throughput, and manually setting workers to handle that seems like it might be obviating the work handled by the runtime and an unnecessary knob to twiddle. the internal backoff would ideally handle errors to slow sending events and otherwise max out what the system allows it.

Side note:
Knowing nothing about how libbeat handles it (based on configs it seems to be knobs), we could also look at doing something along the lines of TCP's [flow control](https://en.wikipedia.org/wiki/Transmission_Control_Protocol#Flow_control) and [congestion control](https://en.wikipedia.org/wiki/Transmission_Control_Protocol#Congestion_control) for deciding how many bytes to send in a bulk request (but this is not really relevant for now, but just an idea while i'm looking at this stuff)

>Is max_retries being honoured? (I recall @simitt finding that it is not, with the way we use libbeat now.)

I'll have to figure out someway to check this. as of now i've just been handling reconnects from failure, and not http failures when sending events or the sever not being present at all.

>Do the non-libbeat output Elasticsearch clients in APM Server honour the configured backoff? These are: API Key auth lookups, tail-based sampling.

tbd, will look at that today.

axw:
> as an "outsiders" note, letting the go runtime map goroutines to threads seems like a fine way to handle increasing throughput, and manually setting workers to handle that seems like it might be obviating the work handled by the runtime and an unnecessary knob to twiddle. the internal backoff would ideally handle errors to slow sending events and otherwise max out what the system allows it.
>
> Side note:
> Knowing nothing about how libbeat handles it (based on configs it seems to be knobs), we could also look at doing something > along the lines of TCP's flow control and congestion control for deciding how many bytes to send in a bulk request (but this is not really relevant for now, but just an idea while i'm looking at this stuff)

I would dearly love for all of this to be adaptive, and for users to not have to touch a thing -- except maybe setting some limits, like an upper bound on bulk sizes.

stuartnelson3:
>Do the non-libbeat output Elasticsearch clients in APM Server honour the configured backoff? These are: API Key auth lookups, tail-based sampling.

Short answer, no.

The following config options are available in both v7 and v8:

- https://github.com/elastic/go-elasticsearch/blob/master/elasticsearch.go#L47-L50
- https://github.com/elastic/go-elasticsearch/blob/master/elasticsearch.go#L60

none of which are being used in our client config code (which is the only entrance to making an es client in `apm-server` as far as i can see):
https://github.com/elastic/apm-server/blob/master/elasticsearch/client.go#L152-L172

If one of the retry status codes is encountered (502, 503, 504), the client currently spams 3 requests and then fails.

stuartnelson3:
Started respecting `max_retry`, `backoff.init`, and `backoff.max` for the non-libbeat client. We can discuss at next standup what sort of backoff we want, for now I've just thrown in a basic exponential backoff.

stuartnelson3:
closed in favor of https://github.com/elastic/apm-server/pull/4991

Issue #4919, opened 2021-03-04
Visualize JVM health indicators collected by the OpenTelemetry Java Agent


cyrille-leclerc:
**APM Server version** (`apm-server version`):

**Description of the problem including expected versus actual behavior**:

**Steps to reproduce**:

https://github.com/cyrille-leclerc/my-shopping-cart/tree/39d22e715d96e15d5540ea28eca007de9d3edab8

### JVM Runtime Metrics reported by the OpenTelemetry Collector Prometheus Exporter

```
# HELP runtime_jvm_gc_collection Time spent in a given JVM garbage collector in milliseconds.
# TYPE runtime_jvm_gc_collection counter
runtime_jvm_gc_collection{gc="G1 Old Generation"} 0
runtime_jvm_gc_collection{gc="G1 Young Generation"} 22
# HELP runtime_jvm_memory_area Bytes of a given JVM memory area.
# TYPE runtime_jvm_memory_area counter
runtime_jvm_memory_area{area="heap",type="committed"} 5.6623104e+07
runtime_jvm_memory_area{area="heap",type="max"} 4.294967296e+09
runtime_jvm_memory_area{area="heap",type="used"} 1.4552776e+07
runtime_jvm_memory_area{area="non_heap",type="committed"} 4.1222144e+07
runtime_jvm_memory_area{area="non_heap",type="used"} 3.7732072e+07
# HELP runtime_jvm_memory_pool Bytes of a given JVM memory pool.
# TYPE runtime_jvm_memory_pool counter
runtime_jvm_memory_pool{pool="CodeHeap 'non-nmethods'",type="committed"} 2.555904e+06
runtime_jvm_memory_pool{pool="CodeHeap 'non-nmethods'",type="max"} 5.840896e+06
runtime_jvm_memory_pool{pool="CodeHeap 'non-nmethods'",type="used"} 1.288832e+06
runtime_jvm_memory_pool{pool="CodeHeap 'non-profiled nmethods'",type="committed"} 2.555904e+06
runtime_jvm_memory_pool{pool="CodeHeap 'non-profiled nmethods'",type="max"} 1.22908672e+08
runtime_jvm_memory_pool{pool="CodeHeap 'non-profiled nmethods'",type="used"} 1.288448e+06
runtime_jvm_memory_pool{pool="CodeHeap 'profiled nmethods'",type="committed"} 6.750208e+06
[...]

Issue #2631, opened 2019-08-23
Breaking changes for 8.0


graphaelli:
Consolidated list of breaking changes we should consider for the next major.

- [ ] replace `timestamp.us` and `span.start` with @timestamp (from #1678)
- [ ] add [`date_range`](https://www.elastic.co/guide/en/elasticsearch/reference/current/range.html) field for trace events
- [ ] Revisit `error` fields and grouping logic (from #1769)
- [ ] Healthcheck improvements (from meta #2632)
- [ ] bug fixes in `beater` package that are breaking changes (TODO comments in code) 
- [ ] ensure every route only accepts dedicated _METHOD_ and put logic for it directly into the `mux.go` file when creating routes.
- [ ] start filling in `event.*` fields (not breaking)
- [ ] reject events with an empty transaction type (https://github.com/elastic/apm-server/issues/3141)
- [ ] use xxHash for grouping key calculation (https://github.com/elastic/apm-server/issues/3226)
- [ ] consider json-schema changes after separating RUM decoding, (eg: optional timestamp)
- [ ] require `transaction.name` - already happens by convention, should formalize since UI doesn't show no-name transactions
- [x] remove `page.url` and `page.referer` (https://github.com/elastic/apm-server/pull/3857)
- [ ] aligning `db` fields in ECS is potentially a breaking change for `span.db`, see [ECS#514](https://github.com/elastic/ecs/issues/514#issuecomment-663123661) 
- [ ] store some metrics as `aggregate_metric_double` https://github.com/elastic/apm-server/issues/4612
- [ ] create data stream specific ingest pipelines
- [ ] ILM policy to include delete / frozen tier
------------------------------

Discarded for 8.0:

~- [ ] Remove `null` values from Json Spec (from #1410)~
~- [ ] Add `labels` and deprecate `tags` on Intake API (from #1835)~
~- [ ] Move onboarding index to system/dot-prefix or saved object~
~- [ ] Multiple listener support (from #1405)~
~- [ ] Synchronous upload of Sourcemaps or consider to move upload endpoint to Kibana (from #2862 )~
~- [ ] move pipeline configuration into index template / use `final_pipeline`~ (Moving to Fleet.)
~- [ ] restructure apm-server.yml file to follow along with filebeat input structure [as suggested](https://github.com/elastic/apm-server/pull/2976/files#r352960036)~ (Jaeger/OTLP will be served on the same port as the Elastic APM agent protocol)

------------------------------

[...]

PR #4986, opened 2021-03-24
ingest: translate known OpenTelemetry JVM metrics

axw:
## Motivation/summary

Copy well-known JVM metrics from their OpenTelemetry
names and associated labels to their counterparts
produced by the Elastic APM Java agent. This enables
users to visualise the metrics in the Elastic APM
app in Kibana.

Metrics are duplicated to avoid surprising users by
silently dropping the OpenTelemetry names. Users can
drop either one of the duplicates with ingest pipeline
customisation.

## Checklist

## How to test these changes

1. Run apm-server
2. Instrument a Java application with https://github.com/open-telemetry/opentelemetry-java-instrumentation, pointed at apm-server
3. Wait for metrics to be collected and sent to APM Server (not sure how long this is)
4. Navigate to the APM app in Kibana, and check the GC and memory usage graphs are populated

See also https://github.com/elastic/apm-server/issues/4919

## Related issues

Closes https://github.com/elastic/apm-server/issues/4919


axw:
@bmorelli25 I haven't added any docs here, not sure where would be appropriate?

felixbarny (commented):
Why does the ingest pipeline get applied to all data streams? I'd expect it to only be in `/data_stream/app_metrics/elasticsearch/ingest_pipeline/apm_opentelemetry_metrics.json`

bmorelli25:
> @bmorelli25 I haven't added any docs here, not sure where would be appropriate?

As this is a new default ingest pipeline, I'd like to see docs added here: https://github.com/elastic/apm-server/blob/master/docs/configuring-ingest.asciidoc. Could you add a new entry to the table that includes an added badge: `added::[7.13]`.

axw:
> Could you add a new entry to the table that includes an added badge: added::[7.13].

Thanks @bmorelli25, will do.

> Why does the ingest pipeline get applied to all data streams? I'd expect it to only be in /data_stream/app_metrics/elasticsearch/ingest_pipeline/apm_opentelemetry_metrics.json

@felixbarny that would be ideal. With the old way of doing things, we only ever have a single pipeline. While we maintain both the old and new ways in tandem, we're just copying the single pipeline to all data streams in the integration package. In the future it would make sense to be more targeted.

axw:
@bmorelli25 tagging you for review on the docs change. I also added an entry for another recently introduced ingest pipeline. Not sure how much of a pain it would be, but perhaps we should document the specific mapping when we update the OTel docs. Then in the ingest pipeline docs page we can link there.

docs/configuring-ingest.asciidoc
```diff
@@ -40,6 +40,17 @@ and then viewing `ingest/pipeline/definition.json`.
 
 To disable this, or any other pipeline, set `output.elasticsearch.pipeline: _none`.
 
+`apm_error_grouping_name`::
```
bmorelli25, line 43: L38-41 snuck in the middle of the table. Can you move L38-41 below the new content that you've added.

bmorelli25 (commented):
> Not sure how much of a pain it would be, but perhaps we should document the specific mapping when we update the OTel docs. Then in the ingest pipeline docs page we can link there.

👍 Updated the doc issue accordingly.

docs/configuring-ingest.asciidoc
axw: Oops - fixed, thanks.

bmorelli25 (commented):
Docs look good! Thanks.

felixbarny:
> we're just copying the single pipeline to all data streams in the integration package. 

Is that a limitation in APM Server or in data streams?

> In the future it would make sense to be more targeted.

Is there an issue to track that?

axw:
>> we're just copying the single pipeline to all data streams in the integration package.
>
> Is that a limitation in APM Server or in data streams?

APM Server. We're doing this for our own sanity, having a single source of truth for the pipeline.

>> In the future it would make sense to be more targeted.
>
> Is there an issue to track that?

There wasn't, but I've just now added an item to https://github.com/elastic/apm-server/issues/2631.

felixbarny (approved):

axw:
jenkins run the tests please

ingest/pipeline/definition.yml
```diff
@@ -74,3 +77,35 @@ apm_error_grouping_name:
...
+      # Copy `runtime.jvm.gc.count` (OpenTelemetry) to `jvm.gc.count` (Elastic APM).
+      field: jvm.gc.count
+      copy_from: runtime.jvm.gc.count
+      if: ctx.runtime?.jvm?.gc?.count != null
+  - set:
+      # Copy `runtime.jvm.gc.collection` (OpenTelemetry) to `jvm.gc.time` (Elastic APM).
+      # Both are defined in milliseconds. This is the old name for `runtime.jvm.gc.time`.
+      field: jvm.gc.time
+      copy_from: runtime.jvm.gc.collection
+      if: ctx.runtime?.jvm?.gc?.collection != null
```
simitt, line 105: nit: It probably shouldn't happen, but could `runtime.jvm.gc.time` and `runtime.jvm.gc.collection` both be set? If so the value from the _old_ name would override the newer one. I'd probably flip it, or set `override: false` here.

ingest/pipeline/definition.yml
```diff
@@ -74,3 +77,35 @@ apm_error_grouping_name:
...
+  - set:
+      # Copy `labels.gc` (OpenTelemetry) to `labels.name` (Elastic APM), for jvm.gc.{time,count}.
+      field: labels.name
+      copy_from: labels.gc
+      override: false # don't replace existing labels.name field, if any
+      if: ctx.labels?.gc != null
```
simitt, line 111: This is not only applied for `jvm.gc.{time,count}` but to every `labels.gc` right?

apmpackage/apm/0.1.0/data_stream/app_metrics/elasticsearch/ingest_pipeline/apm_opentelemetry_metrics.json
```diff
@@ -0,0 +1,41 @@
+{
+  "description": "Populate Elastic APM metric fields from well-known OpenTelemetry metric counterparts",
+  "processors": [
+    {
+      "set": {
+        "field": "jvm.memory.{{labels.area}}.{{labels.type}}",
+        "copy_from": "runtime.jvm.memory.area",
+        "if": "ctx.runtime?.jvm?.memory?.area != null && ctx.labels?.area != null && ctx.labels?.type != null"
```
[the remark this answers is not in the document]
axw: No, hence the use of the [null safe](https://www.elastic.co/guide/en/elasticsearch/painless/current/painless-operators-reference.html#null-safe-operator) operator. If `labels` is null, then `ctx.labels?.area` will evaluate to `null`.

ingest/pipeline/definition.yml
axw: Fair, I'll move `runtime.jvm.gc.collection` above `runtime.jvm.gc.time`.

ingest/pipeline/definition.yml
axw: Indeed, probably a bit too lax. I'll add `&& ctx.runtime?.jvm?.gc != null` to the condition.

axw:
run elasticsearch-ci/docs

axw:
jenkins run the tests please

simitt:
Tested with BC4:
Instrumented a java test app with the otel instrumentation and waited until metrics documents existed. No metrics show up in the APM UI. 

Screenshot 2021-05-12 at 15 02 26

Screenshot 2021-05-12 at 15 00 11

axw:
> agent.name needs to be java (not opentelemetry/java) so that the JVM tab shows up

Ugh :(
I don't think we should be changing the agent name, so this part will probably need to happen in the UI.

> minimum jvm.memory.heap.used needs to exist for JVM details to be shown.

Hmm. Is there just one metric document? We should be capturing `jvm.memory.heap.used`. Maybe what's happening is that we're splitting the metrics into multiple docs - would that explain this?

felixbarny:
What might also be missing is `agent.ephemeral_id`. IIRC, the UI needs that for downsampling/calculating derivatives for the GC time and count.

simitt:
created https://github.com/elastic/kibana/issues/100253 as a follow up for showing opentelemetry JVM metrics. 

> Hmm. Is there just one metric document? We should be capturing jvm.memory.heap.used. Maybe what's happening is that we're splitting the metrics into multiple docs - would that explain this?

simitt:
created a follow up for the metrics docs https://github.com/elastic/apm-server/issues/5275

axw:
> What might also be missing is agent.ephemeral_id. IIRC, the UI needs that for downsampling/calculating derivatives for the GC time and count.

@felixbarny there's a TypeScript definition for `ephemeral_id`, but no other code in the UI that actually references the field in aggs or otherwise. Does that indicate a UI bug?

felixbarny:
> Does that indicate a UI bug?

Looking at the aggs that Kibana does, I think it doesn't work as I'd expect

https://github.com/elastic/kibana/blob/1f02c48d3cf57dd921cfe64186a695f8375e3954/x-pack/plugins/apm/server/lib/metrics/by_agent/java/gc/fetch_and_transform_gc_metrics.ts#L70-L109

It does a derivative over the max value of gc.count across instances as opposed to summing the derivatives of each instance.

I guess it would be easier for the UI if the Java agent reported deltas for gc.time instead of an incrementing counter. However, it might be very tricky to adjust the OTel metrics to report deltas.
IIRC, there's some setting in OTel that allows to report deltas instead of monotonically incrementing counters. But I'm not sure if that setting also applies for the [GC metrics that are captured via a `longSumObserver`](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/runtime-metrics/library/src/main/java/io/opentelemetry/instrumentation/runtimemetrics/GarbageCollector.java).

simitt:
@felixbarny do you mind moving this over to a new issue? It sounds like it'd be worth a follow up discussion.

felixbarny:
After giving it another thought, it's probably not an issue as that chart is only ever shown in the context of a specific instance. Therefore, the ephemeral_id is not needed.

PR #4991, opened 2021-03-24
add exponential backoff to es clients

stuartnelson3:
## Motivation/summary

This is being spun out of work done in https://github.com/elastic/apm-server/pull/4981, with original motivation coming from https://github.com/elastic/apm-server/issues/4138.

After investigation, our non-libbeat esclients weren't respecting backoff and retry configuration set in `apm-server.yml`. This has been added, and a basic exponential backoff. The included library is [used as an example](https://github.com/elastic/go-elasticsearch/blob/4b40206692088570801280584e614027e6ce818b/_examples/bulk/indexer.go#L86) in `go-elasticsearch`, so it seemed like a decent place to start. I'm not sure what the policy is on bringing in external libraries though, and it's easy enough to create our own (it does have an MIT license, though).

Regarding testing, I just wanted to check that we're hooking up the retry function correctly, and that we're setting the retry count correctly. Any suggestions for simplifications or additions are welcome!

## Checklist

## How to test these changes

1. run a web server that will fail (something like [this](https://gist.github.com/stuartnelson3/ddbf40dea8c49f0d0699f1ac481ac2ac))
2. build and run a command that uses an es client, like `./apm-server apikey create -E output.elasticsearch.backoff.init=3s`

output from the webserver should show the requests are spaced out.


go.mod
```diff
@@ -5,6 +5,7 @@ go 1.13
 require (
 	github.com/akavel/rsrc v0.10.1 // indirect
 	github.com/apache/thrift v0.13.0
+	github.com/cenkalti/backoff v2.2.1+incompatible
```
axw, line 8: should this be github.com/cenkalti/backoff/v4? and v4.1.0 (latest)?

elasticsearch/client.go
```diff
@@ -131,43 +131,66 @@ func NewClient(config *Config) (Client, error) {
...
-		Addresses: addresses,
-		Transport: transport,
-		Header:    headers,
+		APIKey:               apikey,
+		Username:             user,
+		Password:             pwd,
+		Addresses:            addresses,
+		Transport:            transport,
+		Header:               headers,
+		RetryOnStatus:        []int{502, 503, 504, 429},
```
axw, line 190: nit: create a var at the top of the file to share with v7 client?

elasticsearch/backoff.go
```diff
@@ -0,0 +1,55 @@
...
+		Max:  time.Minute,
+	}
+
+	// DefaultBackoff is an exponential backoff configured with default
+	// backoff settings.
+	DefaultBackoff = exponentialBackoff(DefaultBackoffConfig)
+)
+
+func exponentialBackoff(b elasticsearch.Backoff) backoffFunc {
+	retryBackoff := backoff.NewExponentialBackOff()
```
axw, line 44: According to https://github.com/cenkalti/backoff/blob/c2975ffa541a1caeca5f76c396cb8c3e7b3bb5f8/exponential.go#L77-L78, there's a default jitter (RandomizationFactor) and mulitplier is 1.5. Did you change your mind about adding jitter, or should we set that to zero? And should we explicitly set the multiplier to 2 to match libbeat?

elasticsearch/backoff_test.go
```diff
@@ -0,0 +1,94 @@
...
+
+	req, err := http.NewRequest("GET", srv.URL, nil)
+	assert.NoError(t, err)
+	c.Perform(req)
+	assert.True(t, called)
+}
+
+func TestBackoffRetries(t *testing.T) {
+	var (
+		i       int
```
axw, line 70:
```suggestion
		requests       int
```

nit: I'd prefer a more self-descriptive variable name

elasticsearch/backoff.go
```diff
@@ -0,0 +1,55 @@
...
+		Init: time.Second,
+		Max:  time.Minute,
+	}
+
+	// DefaultBackoff is an exponential backoff configured with default
+	// backoff settings.
+	DefaultBackoff = exponentialBackoff(DefaultBackoffConfig)
+)
+
+func exponentialBackoff(b elasticsearch.Backoff) backoffFunc {
```
axw, line 43: can we add a unit test for this?

axw (commented):
Basically LGTM, just a few minor things. Can we use github.com/cenkalti/backoff/v4 rather than github.com/cenkalti/backoff?

elasticsearch/backoff.go
stuartnelson3: Let's explicitly set the multiplier, and while we're in there, we can drop the randomization

elasticsearch/backoff.go
```diff
@@ -0,0 +1,56 @@
...
+	return func(attempts int) time.Duration {
+		// Attempts starts at 1, after there's already been a failure.
+		// https://github.com/elastic/go-elasticsearch/blob/de2391/estransport/estransport.go#L339
+		if attempts == 1 {
+			return b.Init
+		}
+		next := math.Pow(float64(attempts), 2)
+		// Make sure we don't exceed max
+		next = math.Min(next, maxSeconds)
+		return time.Duration(next) * time.Second
```
axw, line 54:
```suggestion
		next := b.Init * (1<<(attempts-1))
		if next > b.Max {
			next = b.Max
		}
		return next
```

We need to take `b.Init` whether or not attempts is 1, and it should be 2^N not N^2 (sorry if you did it this way based on my message in Slack, I wrote that backwards).

elasticsearch/backoff_test.go
```diff
@@ -0,0 +1,109 @@
...
+	backoffCfg := elasticsearch.Backoff{
+		Init: init,
+		Max:  time.Minute,
+	}
+	backoffFn := exponentialBackoff(backoffCfg)
+	assert.Equal(t, init, backoffFn(1))
+	assert.Equal(t, 4*time.Second, backoffFn(2))
+	assert.Equal(t, 9*time.Second, backoffFn(3))
+	assert.Equal(t, 16*time.Second, backoffFn(4))
+	assert.Equal(t, 25*time.Second, backoffFn(5))
```
axw, line 107:
```suggestion
	assert.Equal(t, 4*time.Second, backoffFn(2))
	assert.Equal(t, 8*time.Second, backoffFn(3))
	assert.Equal(t, 16*time.Second, backoffFn(4))
	assert.Equal(t, 32*time.Second, backoffFn(5))
```

axw (changes requested):

elasticsearch/backoff.go
stuartnelson3: Ah, ok. I was thinking about `backoff.init` and wasn't sure if it meant "wait an initial X seconds, and then start an exponential backoff", or "wait X seconds, and then each additional backoff is exponential based off the initial backoff". I read the docs in `apm-server.yml` and it wasn't super clear:

```
After waiting backoff.init seconds, apm-server
tries to reconnect. If the attempt fails, the backoff timer is increased
exponentially up to backoff.max.
```

I'm not sure which way is "better", but wanting to wait an initial 20s, for example, and then starting the backoff from there (t0 = 20s, t1 = 21s, t2 = 22s, t3 = 24s, etc), is something that I could see making sense.

axw: I have seen implementations like that before, where there's an initial delay -- it's not how the [libbeat implementation](https://github.com/elastic/beats/blob/2d88ed0286494c9423011b02784c14e79c3f7b8a/libbeat/common/backoff/equal_jitter.go#L54-L73) works though. There the init value is used to initialise the backoff duration on first failure, and then that duration is exponentially increased up to the max.

I suggest we align with libbeat, excluding jitter for now at least, and consider clarifying the docs (which are copied from beats). I'd be surprised if anyone looks at those docs though tbqh; I expect (and hope) folks mostly rely on the defaults.

stuartnelson3: sounds good :ok_hand:

axw (approved):
thanks for bearing with me :)

PR #4866, opened 2021-02-23
model: convert more *strings to strings

axw:
## Motivation/summary

Simplify the model types by converting `*string` to `string` where we can rely on an empty string being invalid.

## How to test these changes

Non-functional change, run the tests.

## Related issues

None.

PR #4994, opened 2021-03-25
[7.x] model: convert more *strings to strings (#4866)

axw:
 - model: convert more *strings to strings (#4866)


axw:
jenkins run the tests please

PR #4998, opened 2021-03-25
[7.x] Replace publish.Reporter with model.BatchProcessor (#4878)

axw:
 - Replace publish.Reporter with model.BatchProcessor (#4878)


axw:
@Mergifyio update

PR #4882, opened 2021-03-02
Enable transaction aggregation by default

axw:
## Motivation/summary

Enable by default so the UI can set the metrics-based mode to auto or on by default.
We still don't drop unsampled transactions by default.

## Checklist

- [x] Update [CHANGELOG.asciidoc](https://github.com/elastic/apm-server/blob/master/CHANGELOG.asciidoc)
~- [ ] Documentation has been updated~ there are currently no docs for this feature.

## How to test these changes

1. Run apm-server with default config
2. Send some transactions
3. Wait a minute or so
4. Check for transaction metrics docs (should contain `transaction.duration.histogram`)

## Related issues

Closes #4816

PR #5000, opened 2021-03-25
[7.x] Enable transaction aggregation by default (#4882)

axw:
 - Enable transaction aggregation by default (#4882)


axw:
@Mergifyio update

PR #5005, opened 2021-03-25
[7.x] display timeouts to user as "server timeout" (506311f0)

axw:
 - display timeouts to user as "server timeout" (506311f0)


axw:
Ah just realised that the PR was not squashed. Will recreate...

PR #5009, opened 2021-03-26
[7.x] model: de-pointer several integer fields (#4924)

axw:
 - model: de-pointer several integer fields (#4924)


axw:
@Mergifyio update

PR #5030, opened 2021-03-29
Introduce `apm-server.rum.allow_service_names` config

### LINKED ISSUE #4709: Limit cardinality of RUM data streams
The new [APM indexing strategy](https://github.com/elastic/apm-server/issues/4378) incorporates the service name in the data stream's dataset. By doing this we effectively partition data by service name, enabling per-service policies (ILM, security, ...)

Because RUM is typically unauthenticated, this means that the cardinality of RUM data streams would be unbounded if we did not restrict the incoming service names. We should introduce a configuration that rejects data not matching a given service name.

axw:
## Motivation/summary

Introduce `apm-server.rum.allow_service_names` config which, if set, restricts RUM events to those with a service name exactly matching one of the items in the the list. If any event is received with a non-matching service name, that event and any proceeding it will be rejected.

## Checklist

## How to test these changes

1. Run apm-server with default config, send RUM data with a service name set to "evergiven" and show that it is indexed.
2. Run apm-server with `apm-server.rum.allow_service_names: ["evergiven"]`, and again show that RUM events are indexed.
3. Run apm-server with `apm-server.rum.allow_service_names: ["evergreen"]`, show that RUM events are NOT indexed.

## Related issues

Closes #4709


simitt (commented):
Overall LGTM

axw:
@simitt I've just added a var to the integration package since your review, can you please take another look?

simitt (approved):
LGTM, please file an issue to add this to the allowlisted values for the ESS deployments.

simitt:
tested with BC4 - works as described in the test instructions.

PR #5031, opened 2021-03-29
apmpackage: bump version to 0.2.0

axw:
## Motivation/summary

Bump integration package version to 0.2.0

## How to test these changes

In Kibana navigate to Fleet, and find the "Elastic APM" integration. Check that its version is 0.2.0.

## Related issues

None.


jalvz:
Why bump and not just add 0.2.0?

simitt:
I am in favor of bumping - it shows the diff as opposed to when all files were duplicated and the whole `0.2.0` section would show up as newly added.
Also, keeping the latest version in the APM Server and all published versions in the package repo itself seems fine; what would be the benefit of duplicating everything?

jalvz:
Sure, I'm fine with that.

axw:
Silvia summed up what I was thinking (plus more). Given that we're going to align versions with the stack, think of 0.1.0 as a stand-in for 7.12.0, and 0.2.0 as a stand-in for 7.13.0.

simitt (approved):

Issue #4149, opened 2020-09-04
limit sourcemap line length


clyfish:
My coordinating node got OOM when query documents with huge lone lines in the sourcemap, so I want limit the line length of the sourcemap.

I've made a PR, but it's too hard for me to make the limit configurable.
So I closed the PR.
https://github.com/elastic/apm-server/pull/4143

stuartnelson3:
closing this as we haven't been able to reproduce ourselves. If you can provide a way for us to reproduce it we'll take another look.

clyfish:
After doubering the memory limit from `-Xmx3g -Xms3g` to `-Xmx6g -Xms6g`, our cluster is stable without OOM In recent months.

PR #5033, opened 2021-03-29
support truncating sourcemap lines

stuartnelson3:
## Motivation/summary

User-requested feature to limit the line length for returned sourcemaps, as excessively long lines were OOM'ing coordinating nodes.

closes https://github.com/elastic/apm-server/issues/4149

## Checklist

## How to test these changes

Start a copy of APM server with the `source_mapping.max_line_length` set to something small. Confirm that when viewing lines, they've been truncated.

### DIFF
--- a/apm-server.yml
+++ b/apm-server.yml
@@ -240,6 +240,10 @@ apm-server:
       # Sourcemapping is enabled by default.
       #enabled: true
 
+      # Maximum allowed line length in characters. Default maximum line length
+      # is unlimited, indicated by `0`. Minimum allowed value is 16.
+      #max_line_length: 0
+
       # Source maps are always fetched from Elasticsearch, by default using the output.elasticsearch configuration.
       # A different instance must be configured when using any other output.
       # This setting only affects sourcemap reads - the output determines where sourcemaps are written.
--- a/beater/config/rum.go
+++ b/beater/config/rum.go
@@ -59,12 +59,12 @@ type EventRate struct {
 
 // SourceMapping holds sourecemap config information
 type SourceMapping struct {
-	Cache        *Cache `config:"cache"`
-	Enabled      *bool  `config:"enabled"`
-	IndexPattern string `config:"index_pattern"`
-	// TODO: uint? uint64? Or do we not care and just use an int?
-	MaxLineLength uint                  `config:"max_line_length"`
-	ESConfig      *elasticsearch.Config `config:"elasticsearch"`
+	Cache        *Cache                `config:"cache"`
+	Enabled      *bool                 `config:"enabled"`
+	IndexPattern string                `config:"index_pattern"`
+	ESConfig     *elasticsearch.Config `config:"elasticsearch"`
+	// MaxLineLength of 0 indicates no maximum length.
+	MaxLineLength int `config:"max_line_length"`
 	esConfigured  bool
 }
 
--- a/beater/config/rum.go
+++ b/beater/config/rum.go
@@ -59,11 +59,13 @@ type EventRate struct {
 
 // SourceMapping holds sourecemap config information
 type SourceMapping struct {
-	Cache        *Cache                `config:"cache"`
-	Enabled      *bool                 `config:"enabled"`
-	IndexPattern string                `config:"index_pattern"`
-	ESConfig     *elasticsearch.Config `config:"elasticsearch"`
-	esConfigured bool
+	Cache        *Cache `config:"cache"`
+	Enabled      *bool  `config:"enabled"`
+	IndexPattern string `config:"index_pattern"`
+	// TODO: uint? uint64? Or do we not care and just use an int?
+	MaxLineLength uint                  `config:"max_line_length"`
+	ESConfig      *elasticsearch.Config `config:"elasticsearch"`
+	esConfigured  bool
 }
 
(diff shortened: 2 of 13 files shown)


apm-server.yml
```diff
@@ -240,6 +240,10 @@ apm-server:
       # Sourcemapping is enabled by default.
       #enabled: true
 
+      # Maximum allowed line length in characters. Default maximum line length
+      # is unlimited, indicated by `0`. Minimum allowed value is 16.
+      #max_line_length: 0
```
stuartnelson3, line 245: Are these options listed elsewhere and the apm-server.yml files generated? or do I add this to e.g. apm-server.docker.yml as well?

beater/config/rum.go
```diff
@@ -131,9 +133,10 @@ func (s *SourceMapping) Unpack(inp *common.Config) error {
 
 func defaultSourcemapping() *SourceMapping {
 	return &SourceMapping{
-		Cache:        &Cache{Expiration: defaultSourcemapCacheExpiration},
-		IndexPattern: defaultSourcemapIndexPattern,
-		ESConfig:     elasticsearch.DefaultConfig(),
+		Cache:         &Cache{Expiration: defaultSourcemapCacheExpiration},
+		IndexPattern:  defaultSourcemapIndexPattern,
+		ESConfig:      elasticsearch.DefaultConfig(),
+		MaxLineLength: 0,
```
stuartnelson3, line 139: `0` indicates "no maximum"

sourcemap/mapper.go
```diff
@@ -19,23 +19,38 @@ package sourcemap
...
 	"strings"
 
 	"github.com/go-sourcemap/sourcemap"
 )
 
-const sourcemapContentSnippetSize = 5
+const (
+	sourcemapContentSnippetSize = 5
+	// TODO: What value do we want here?
+	minMaxLineLength = 16
```
stuartnelson3, line 31: I'm assuming 16 is much smaller than we want, it was just convenient for testing.

sourcemap/mapper.go
```diff
@@ -66,3 +81,40 @@ func abs(n int) int {
 	}
 	return n
 }
+
+// limitScanLines is a modified form of bufio.ScanLines. It truncates any lines
+// longer than the supplied `max` value.
+func limitScanLines(max int) func([]byte, bool) (int, []byte, error) {
```
stuartnelson3, line 87: This is mostly copy and paste. I'm happy to remove this and just re-slice `scanner.Text()` above, but i figured if we're trying to avoid allocations, why not try to avoid allocations. (`scanner.Text()` allocates a new string)

stuartnelson3:
This is a possible solution, but I'm happy to hear alternate suggestions. I like how this impl is using the api and setting `Split()`, but it is mostly copying `bufio.SplitLines`, which is less cool.

I haven't tested this manually (or written a system test). Maybe I can steal someone for 15min tomorrow after standup and we can go through how to do this (upload a sourcemap and then view it in the UI).

apm-server.yml
axw: Both `apm-server.yml` and `apm-server.docker.yml` are generated from `_meta/beat.yml`. Edit that one and run `make update` (or just `make config`).

axw (commented):
Looks fine overall, though I'd prefer to simplify the truncation code.

I'm a little unclear on what triggered the OOM in the reported issue. Initially I thought it was about the length of lines in the source map itself, but in https://github.com/elastic/apm-server/pull/4143 the author is only limiting the length of the source lines added to stack trace frames -- like you are here.

It would be good if you could try to reproduce the failure by indexing error events with very long source context lines. You might be able to simulate this by using the Go agent, passing in a custom `stacktrace.ContextSetter` to https://pkg.go.dev/go.elastic.co/apm#Tracer.SetContextSetter and capturing an error.

sourcemap/mapper.go
stuartnelson3: Seems kind of silly if someone limits the output to 1 character, for example. But, I'm also happy to delete it.

sourcemap/mapper.go
stuartnelson3: Yeah, it won't decrease the number of allocations, but I guess they'd potentially be smaller (which probably isn't much of a win).

sourcemap/mapper_test.go
```diff
@@ -32,17 +30,17 @@ import (
...
 	// no sourcemapConsumer
-	_, _, _, _, _, _, _, ok := Map(nil, 0, 0)
+	_, _, _, _, _, _, _, ok := Map(nil, 0, 0, 0)
 	assert.False(t, ok)
 }
 
 func TestMapNoMatch(t *testing.T) {
 	m, err := sourcemap.Parse("", []byte(test.ValidSourcemap))
-	require.NoError(t, err)
+	assert.NoError(t, err)
```
simitt, line 39: why changing to `assert`? Running the tests if the parsing failed doesn't make much sense and would probably just lead to confusing errors (hiding the original issue). Similar for other changes below.

stuartnelson3: ah sorry, just my OCD. I'll change them back

stuartnelson3:
closing this for now. it doesn't solve the problem, and I haven't been able to reproduce the issue.

PR #4997, opened 2021-03-25
[7.x] systemtest: fix Elastic Agent enrollment (#4883)

axw:
Backports the following commits to 7.x:
 - systemtest: fix Elastic Agent enrollment (#4883)

PR #5035, opened 2021-03-30
Remove testing/environments

axw:
## Motivation/summary

Remove our dependency on `testing/environments` from beats for running system tests, inlining the definition of elasticsearch and kibana services directly into our `docker-compose.yml`. We only use the snapshot versions of these services, and we customise them heavily; the cost of the dependency outweighs its benefit.

I've further customised the elasticsearch service's healthcheck to wait for the cluster status to be (no worse than) yellow, rather than green, due to the introduction of Fleet indices. These indices are created with a replica. Depending on timing the cluster status will be either green or yellow. Either way the system tests should be fine.

I have introduced a `make check-docker-compose` target (dependency of `make check-full`) to ensure the images are kept up to date, by comparing image tags to the version in `https://snapshots.elastic.co/latest/<branch>.json`.

## How to test these changes

`make docker-system-tests`

## Related issues

https://github.com/elastic/apm-server/pull/4997#issuecomment-806458307


jalvz (approved):
this is fantastic, thanks!

PR #5037, opened 2021-03-31
tests/system: fix system tests

axw:
## Motivation/summary

Fix Python-based system tests due to several recent changes:
 - Kibana now has a different response structure for listing agent config (API only used in system tests, no backwards compatibility issue for APM Server)
 - Add `user_agent.device.type` to approvals, which is now added by the `user_agent` ingest processor

## How to test these changes

`make system-tests`

## Related issues

None.


axw:
Trivial fix that only affects system tests, merging without review to unblock other PRs.

PR #5041, opened 2021-03-31
beater/interceptors: don't reassign shared logger

axw:
## Motivation/summary

Make sure we create a new shadowed "logger" variable,
rather than reassigning to the "logger" param passed
into the interceptor. Otherwise, each request will add
duplicate fields to an ever-growing list of fields.

## How to test these changes

Send multiple gRPC requests to APM Server, with logging enabled. Make sure there are no duplicate fields in any request's log entry. (If you run a load generator, it'll become apparent pretty quickly as the terminal fills up with logs!)

## Related issues

None.


jalvz (approved):

stuartnelson3 (approved):
:100:

simitt:
successfully tested with BC4 when sending requests via a go otel app - no duplicate log entris

PR #5062, opened 2021-04-07
cmd/apmbench: introduce apmbench tool/framework

axw:
## Motivation/summary

Introduce a new `apmbench` command for generating load on a running APM Server, and measuring event throughput and server resource usage. Results are obtained through the `/debug/vars` API, rather than querying Elasticsearch; this enables measuring the performance of the server even when Elasticsearch is not running.

`apmbench` builds on [`testing.Benchmark`](https://golang.org/pkg/testing/#Benchmark) to provide a familiar API for developing benchmarks, and reuses the Go testing benchmark result format. This will enable the use of [gobench](https://github.com/elastic/gobench), [benchstat](https://pkg.go.dev/golang.org/x/perf/cmd/benchstat), and other tools that consume Go benchmark results.

Finally, `apmbench` provides flags (`-cpuprofile`, etc.) for capturing profiles similar to `go test`. These capture profiles from the target server, which must have been configured with `apm-server.pprof.enabled=true`.

## How to test these changes

1. Run apm-server with the console output, and with expvar and pprof both enabled: `./apm-server --strict.perms=false -e -E output.elasticsearch.enabled=false -E output.console.enabled=true -E apm-server.expvar.enabled=true -E apm-server.pprof.enabled=true &>/dev/null`
2. Run apmbench

## Related issues

None.


axw:
I'm going to move this back into draft, and extract the opentelemetry-go bump in systemtest into a separate PR to make it easier to review.

simitt:
I observed is that the provided `benchtime` is not always respected. The benchmarking often stops before the benchtime is reached because it failed. This happens when running under heavy load, my assumption is that the apm server under load doesn't respond in time to the `expvar` endpoint queries, letting the benchmark fail. 
For doing real performance benchmarks it would be helpful that the benchmark doesn't stop after a short/random amount of time.

systemtest/benchtest/flags.go
```diff
@@ -0,0 +1,91 @@
...
+	"os"
+	"regexp"
+	"strconv"
+	"strings"
+	"testing"
+	"time"
+)
+
+var (
+	server        = flag.String("server", "http://localhost:8200", "apm-server URL")
```
simitt, line 33: Could you also fallback to `ELASTIC_APM_SERVER_URL` here, similar to the `secretToken` behavior?

systemtest/benchtest/expvar.go
```diff
@@ -0,0 +1,62 @@
...
+	ActiveEvents int64 `json:"libbeat.output.events.active"`
+	TotalEvents  int64 `json:"libbeat.output.events.total"`
+}
+
+func queryExpvar(out *expvar) error {
+	req, err := http.NewRequest("GET", *server+"/debug/vars", nil)
+	if err != nil {
+		return err
+	}
+	req.Header.Set("Accept", "application/json")
```
simitt, line 54: Can we add an option or hardcode that insecure certificates should be ignored? So basically when setting `ELASTIC_APM_VERIFY_SERVER_CERT=false` for the agent, respecting that also here.

systemtest/benchtest/flags.go
axw: done

systemtest/benchtest/expvar.go
axw: done, hardcoded

axw:
> I observed is that the provided benchtime is not always respected. The benchmarking often stops before the benchtime is reached because it failed. This happens when running under heavy load my assumption is that the apm server under load doesn't respond in time to the expvar endpoint queries, letting the benchmark fail. The stats.Errors.SendStream returns an error.
For doing real performance benchmarks it would be helpful that the benchmark doesn't stop after a short/random amount of time.

I've changed the test to measure the number of transport errors as a benchmark metric (lower is better), rather than causing the test to fail. Let me know if you think that makes sense or not, or if you have an alternative suggestion.

simitt (approved):
Looks great!

stuartnelson3:
Tested with BC2:
server:
```
# docker run --network=host docker.elastic.co/staging/apm-server:7.14.0-e99135ef --strict.perms=false -e -E output.elasticsearch.enabled=false -E output.console.enabled=true -E apm-server.expvar.enabled=true -E apm-server.pprof.enabled=true
```
apmbench:
```
# go run .
Benchmark1000Transactions             24          70089627 ns/op                 0 error_responses/sec       14610 events/sec
 651219 B/op        5432 allocs/op
BenchmarkOTLPTraces                29271             49816 ns/op                 0 error_responses/sec       19663 events/sec
   1648 B/op          14 allocs/op
```

Issue #4713, opened 2021-02-10
[OpenTelemetry] add support for RPC semantic conventions


axw:
Add support for [OpenTelemetry RPC semantic conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/rpc.md). In particular, we should add support for setting destination service context, for service map support.

Acceptance Criteria: a gRPC call is 
* displayed well in the APM Service Map
* and its span attributes are treated as first class citizen, not as `labels.xxx`attributes.
* outcome/result is set appropriately based on the gRPC status code

cyrille-leclerc:
@axw I propose to add the following acceptance criteria: "GRPC calls captured by Otel instrumentations render well in Elastic APM (Service Map...)"

axw:
@cyrille-leclerc sure, sounds good.

cyrille-leclerc:
Example of problem today with GRPC:

source code: https://github.com/cyrille-leclerc/my-shopping-cart/tree/83253fa0fb3629458865ec01db8e431c6b0674ce

Waterfall graph of spans properly showing the GRPC hop
![image](https://user-images.githubusercontent.com/459691/108221761-703bb300-7138-11eb-8084-3bb155ae3010.png)

Service map  NOT showing the GRPC call
![image](https://user-images.githubusercontent.com/459691/108222124-c872b500-7138-11eb-95d3-b1b8b39e6818.png)


Transaction document of the GRPC Checkout Service

```json
{
  "_index": "apm-7.11.0-transaction-000001",
  "_type": "_doc",
  "_id": "aLtzsHcBb3BmC7-Wt9SX",
  "_version": 1,
  "_score": null,
  "_source": {
    "parent": {
      "id": "ae2a65ad03dcd746"
    },
    "agent": {
      "name": "opentelemetry/java",
      "version": "0.15.0"
    },
    "processor": {
      "name": "transaction",
      "event": "transaction"
    },
    "labels": {
      "os_description": "Mac OS X 10.16",
      "service_namespace": "com-shoppingcart",
      "process_command_line": "/Library/Java/JavaVirtualMachines/adoptopenjdk-15.jdk/Contents/Home:bin:java -javaagent:./../.otel/opentelemetry-javaagent-all-0.15.0.jar -Dotel.exporter.otlp.endpoint=http://localhost:4317 -Dio.opentelemetry.auto.slf4j.simpleLogger.defaultLogLevel=info",
[...]

PR #5074, opened 2021-04-12
processor/otel: add support for rpc conventions

axw:
## Motivation/summary

Handle `rpc.*` semantic conventions for OpenTelemetry:
 - `rpc.system` is used to identify RPC spans. The value
   is currently used for span subtype, and dropped for
   transactions as we have nowhere to store it.
 - `rpc.service` is dropped, as it is included in the
   operation name and we have nowhere to store it.
 - `rpc.method` is dropped, as it is included in the
   operation name and we have nowhere to store it.
 - `rpc.grpc.status_code` is used to set transaction result,
   but is dropped for spans.

Dropping fields is clearly not ideal. We should revisit
this and define an RPC schema which works for OpenTelemetry
as well as our own agents, and which fits into ECS.

RPC transactions are given the type "request", and spans
are given the type "external" with the `rpc.system` value
as the subtype.

`span.destination.service` is set from the peer address, like
we do for our own agents.

## Checklist

## How to test these changes

Instrument gRPC-based server and client with OpenTelemetry, collect traces with APM Server and check:
- transaction type should be "request"
- transaction result should be "OK" (or some other gRPC status code's string representation)
- span type should be "external", and subtype should be "grpc"
- client and server should be connected in the service map in Kibana
- there should be no "labels.rpc.*" fields

## Related issues

Closes https://github.com/elastic/apm-server/issues/4713


axw:
Regarding dropped attributes: I have opened an RFC at https://github.com/elastic/ecs/pull/1357 to introduce a new fieldset.

model/client.go
```diff
@@ -25,14 +25,24 @@ import (
...
 // Client holds information about the client of a request.
 type Client struct {
+	// Domain holds the client's domain (FQDN).
+	Domain string
+
 	// IP holds the client's IP address.
 	IP net.IP
+
+	// Port holds the client's IP port.
+	Port int
```
simitt, line 35: Shouldn't the new fields be indexed?

simitt (commented):
I assume there are currently no plans that the Elastic APM agents would also want to send the `client.domain` or `client.port` information? (And therefore fields are not added to `modeldecoder` and ingest spec)

model/client.go
axw: Indeed. I was thinking it would be automatic because they're ECS fields

axw:
> I assume there are currently no plans that the Elastic APM agents would also want to send the client.domain or client.port information? (And therefore fields are not added to modeldecoder and ingest spec)

Correct, no plans to do that for the moment.

model/client.go
axw: added in ba585d166f1393fb07a68b91f3aaf0b6448fecad

axw:
ugh, I guess the test is still flaky

axw:
jenkins run the tests please

axw:
jenkins run the tests please

simitt (approved):

stuartnelson3:
Attempting to test this, but I don't know if my instrumentation setup is correct. The results don't align at all with what the expected results are. I can make my client/server available, if that's helpful.

The document doesn't look how the test description says it should. This was run against `apm-server version 7.13.0 (amd64), libbeat 7.13.0 [0aa1b50a5b680be19556d53f870fa9723bdc9a03 built 2021-04-27 15:36:50 +0000 UTC]`

```json
{
        "_index" : "apm-7.13.0-transaction-000001",
        "_type" : "_doc",
        "_id" : "a8FHPHkBR2aNSogFxxGx",
        "_score" : 1.0,
        "_source" : {
          "parent" : {
            "id" : "4de49650e073c5bd"
          },
          "agent" : {
            "name" : "opentelemetry/go",
            "version" : "0.20.0"
          },
          "processor" : {
            "name" : "transaction",
            "event" : "transaction"
          },
          "labels" : {
            "rpc_method" : "SayHello",
            "rpc_service" : "api.HelloService",
            "rpc_grpc_status_code" : 0,
            "rpc_system" : "grpc"
          },
          "observer" : {
            "hostname" : "slaptop",
            "id" : "299f1635-eeee-4aeb-96f7-5383b6f7ad21",
            "ephemeral_id" : "0053d551-ef40-4036-b318-392111a4df07",
            "type" : "apm-server",
            "version" : "7.13.0",
            "version_major" : 7
          },
          "trace" : {
            "id" : "38e63cbe310b43375c52e684ca99a3c6"
          },
          "@timestamp" : "2021-05-05T11:26:42.360Z",
          "ecs" : {
            "version" : "1.8.0"
          },
          "service" : {
            "node" : {
              "name" : "slaptop"
            },
            "framework" : {
              "name" : "go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc",
              "version" : "semver:0.20.0"
            },
            "name" : "hello-app",
            "language" : {
              "name" : "go"
            },
            "version" : "1.0"
          },
          "host" : {
            "hostname" : "slaptop",
            "ip" : "127.0.0.1",
            "name" : "slaptop"
          },
          "event" : {
            "ingested" : "2021-05-05T11:26:47.723590255Z",
            "outcome" : "success"
          },
          "transaction" : {
            "result" : "OK",
            "duration" : {
              "us" : 50159
            },
            "name" : "api.HelloService/SayHello",
            "span_count" : {
              "dropped" : 0,
              "started" : 0
            },
            "id" : "c0efaa543242814a",
            "type" : "unknown",
            "sampled" : true
          },
          "timestamp" : {
            "us" : 1620214002360770
          }
        }
      }
```

axw:
> I can make my client/server available, if that's helpful.

Yes please. Can you please mark this one `test-plan-regression` and open a bug issue, labelled v7.13.0 and test-plan? If you could share the program there I'll try it out and see what's going on.

stuartnelson3:
Per discussion in https://github.com/elastic/apm-server/issues/5234, this tested successfully against BC4

PR #5085, opened 2021-04-13
systemtest: use Fleet Server

### LINKED ISSUE #5075: Update system tests to use Fleet Server
We'll need to update our system tests to use Fleet Server for 7.13. We can do this before or after feature freeze, depending on when Fleet server support is removed from Kibana.

axw:
## Motivation/summary

In the Fleet integration system test, bootstrap Fleet Server and then run the APM integration on a separate Elastci Agent that enrolls with the Fleet Server.

## Related issues

Closes #5075

### DIFF
--- a/systemtest/containers.go
+++ b/systemtest/containers.go
@@ -43,7 +43,11 @@ import (
 	"github.com/elastic/go-elasticsearch/v7"
 )
 
-const startContainersTimeout = 5 * time.Minute
+const (
+	startContainersTimeout = 5 * time.Minute
+
+	defaultFleetServerPort = "8220"
+)
 
 // StartStackContainers starts Docker containers for Elasticsearch and Kibana.
 //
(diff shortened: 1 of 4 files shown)


axw:
test seems to be flaky, bringing back into draft while I debug...

systemtest/containers.go
```diff
@@ -361,14 +370,27 @@ func (c *ElasticAgentContainer) Start() error {
 	defer cancel()
 
 	// Update request from user-definable fields.
-	if c.FleetEnrollmentToken != "" {
+	if c.FleetServer {
+		c.request.Env["FLEET_SERVER_ENABLE"] = "1"
+		c.request.Env["FLEET_SERVER_HOST"] = "0.0.0.0"
```
simitt, line 375: Shouldn't be necessary anymore afaik.

simitt (approved):
Looks great.

systemtest/containers.go
axw: I just tried removing FLEET_SERVER_HOST and FLEET_SERVER_PORT (and tried just removing each individually), and the server never became accessible. In the elastic-agent log I see "State changed to CONFIG: Re-configuring". Maybe a bug masked by overriding the host/port?

axw: I'm going to merge as it is and come back to this, so we can get our system tests running again

simitt: Yes sure; thanks for trying it out.

PR #5113, opened 2021-04-16
Adding port forwarding data as this is needed to access APM server ou…

kartheekmannepalli:
…tside of docker.

Adding port forwarding data as this is needed to access APM server outside of docker. Without this when running on local APM server cannot be accessed with localhost:8200.

## Checklist

### DIFF
--- a/docs/copied-from-beats/docs/shared-docker.asciidoc
+++ b/docs/copied-from-beats/docs/shared-docker.asciidoc
@@ -244,6 +244,7 @@ ifeval::["{beatname_lc}"=="apm-server"]
 ["source", "sh", subs="attributes"]
 --------------------------------------------
 docker run -d \
+  -p 8200:8200 \
   --name={beatname_lc} \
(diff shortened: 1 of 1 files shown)


axw:
Thanks for the PR! I don't think we can make this change as-is, because the file you've modified is copied from beats: https://github.com/elastic/beats/blob/master/libbeat/docs/shared-docker.asciidoc

@bmorelli25 any suggestions as to where/how to document this?

bmorelli25:
Hi, @kartheekmannepalli and thank you! Like @axw mentioned, this update is a bit complex because the source file lives in a different repository and is copied over to the apm-server repo. Would you be willing to open this same update in that repository? Here's the file that needs to be updated: https://github.com/elastic/beats/blame/master/libbeat/docs/shared-docker.asciidoc#L246

bmorelli25:
Superseded by https://github.com/elastic/beats/pull/25345.

PR #5118, opened 2021-04-19
Fix approvals for system tests

axw:
## Motivation/summary

`user_agent.device.type` is no longer set by default. Update system test approvals to reflect this.

## How to test these changes

N/A, non-functional change. The field was added and then removed in the 7.13 cycle.

## Related issues

None.


axw:
trivial, merging without review

Issue #5119, opened 2021-04-19
[sampling]: rewrite pubsub implementation to work properly with multiple indices and shards


axw:
The current [pubsub](https://github.com/elastic/apm-server/tree/master/x-pack/apm-server/sampling/pubsub) implementation for tail-based sampling is broken when there are multiple shards, as it performs searches on `_seq_no`. We should consider doing something like https://github.com/elastic/fleet-server/pull/200, polling for global checkpoint changes and using those for searching new documents.

PR #5126, opened 2021-04-20
sampling: fix pubsub implementation

axw:
## Motivation/summary

The initial implementation was written as a ~quick hack, with the expectation that it would be replaced by the Changes API.
It was broken due to its ignorance of data streams, and multi-shard indices. Sequence numbers are only comparable within a single shard.

Instead of waiting for the Changes API we could take the same approach as https://github.com/elastic/fleet-server/pull/200, using
a new Elasticsearch API built for Fleet Server. The main issue here is that the API does not (currently) support data streams.

Given the above and that there is no known delivery date for the Changes API, we propose to instead revise the pubsub implementation to address the problems by:

 - enforcing single-shard indices for sampled trace data streams
 - searching (now single-shard) backing indices individually

In addition, we now use global checkpoints to bound searches (ensuring all replicas have committed the documents). Querying underlying indices and global checkpoints will require an additional "monitor" index privilege.

## Checklist

## How to test these changes

1. Run two APM Servers with tail-based sampling enabled
2. Create some transactions
3. Force a [rollover](https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-rollover-index.html) of the sampled traces data stream, to create a new index
4. Create some more transactions
5. Check transactions are indexed according to the sampling config

## Related issues

Closes https://github.com/elastic/apm-server/issues/5119


x-pack/apm-server/sampling/pubsub/checkpoints.go
```diff
@@ -0,0 +1,83 @@
...
+
+// getGlobalCheckpoints returns the current global checkpoint for each index
+// underlying dataStream. Each index is required to have a single (primary) shard.
+func getGlobalCheckpoints(
+	ctx context.Context,
+	client elasticsearch.Client,
+	dataStream string,
+) (map[string]int64, error) {
+	indexGlobalCheckpoints := make(map[string]int64)
+	resp, err := esapi.IndicesStatsRequest{
```
henningandersen, line 29: The default directly against REST API is that it returns all metrics. I think the same is the case for the go client (but did not check). A few of the metrics have some overhead, so I suggest to ask for just the "Get" metric which is very cheap. You always get the sequence numbers out.

x-pack/apm-server/sampling/pubsub/pubsub.go
```diff
@@ -86,114 +93,163 @@ func (p *Pubsub) SubscribeSampledTraceIDs(ctx context.Context, traceIDs chan<- s
...
+
+// searchTraceIDs searches the configured data stream for new sampled trace IDs, sending them to the out channel.
+//
+// searchTraceIDs works by fetching the global checkpoint for each index backing the data stream, and comparing
+// this to the most recently observed sequence number for the indices. If the global checkpoint is greater, then
+// we search through every document with a sequence number greater than the most recently observed, and less than
+// or equal to the global checkpoint.
+//
+// Searches may not return everything up to the global checkpoint, depending on whether the index has been
+// refreshed since the global checkpoint was last updated. As such, we will only progress the locally observed
```
henningandersen, line 119: This still risks not seeing documents. Unfortunately, you have to refresh currently to avoid that. The global checkpoints returned are independent of refresh. The refreshed state represents and old set of documents. Let us assume you find global checkpoint 10. The state could then be:

unrefreshed: 0,1,2,3,4,5,6,7,8,9,10,12,13
refreshed: 0,1,2,3,4,5,6,7,9,10,12

So while you do not consider 12 (since it is above GCP), you did loose document with seqno 8 by only looking at refreshed state.

There are 2 solutions:

1. Do a refresh between the stats call and the search call.
2. Only consider docs with consecutive seq-nos

The problem with 1 is that performance will degrade if you do too many refreshes.

The problem with 2 is that actually seeing the docs is delayed until a poll round after a refresh (something have to refresh, could be just the normal refresh timer).

The best choice depends on the polling rate. If you potentially poll multiple times per second, the second options is probably better. If you poll only every 10 seconds or so, I think doing an explicit refresh is probably best.

x-pack/apm-server/sampling/pubsub/checkpoints.go
axw: Thanks, this is an easy one!

x-pack/apm-server/sampling/pubsub/pubsub.go
axw: Thanks - I'll add an explicit refresh in then. We'll query every 30 seconds by default. I was trying to minimise the load on Elasticsearch, as there will potentially be many APM Servers. Hopefully this is still a temporary measure anyway, and that we can move over to the global checkpoints API eventually.

stuartnelson3 (commented):
Everything looks reasonable, but this is enough out of my realm of knowledge for me to not be able to provide a definitive :+1: / :-1: . I'd be happy to listen to a quick walkthrough of the code, or merge and deal with any potential issues that arise since the current implementation is broken.

x-pack/apm-server/sampling/pubsub/checkpoints.go
```diff
@@ -0,0 +1,86 @@
...
+	}.Do(ctx, client)
+	if err != nil {
+		return nil, errors.New("index stats request failed")
+	}
+	defer resp.Body.Close()
+	if resp.IsError() {
+		switch resp.StatusCode {
+		case http.StatusNotFound:
+			// Data stream does not yet exist.
+			return indexGlobalCheckpoints, nil
```
simitt, line 44: If I understand that right then for an ES version not supporting the new checkpoint API an empty array would be returned here, but the tail based sampling code would still run, trying to apply sampling, but never receiving checkpoint updates. Have you considered checking the ES version instead and disabling sampling if an ES version < `7.13` is detected?

axw: We are not using the new checkpoint API yet. This error case is about the data stream not yet existing, e.g. because no sampled trace IDs have been recorded. In that case we just return an empty map.

simitt (approved):

stuartnelson3:
confirmed with SNAPSHOT

PR #5133, opened 2021-04-21
[7.x] Bump to 7.14 and update beats

jalvz:
## Checklist


axw:
@jalvz can you please update .mergify.yml? We'll need to change the existing 7.x and add a new section for 7.13.

jalvz:
@axw that would be in master, right?

axw:
@jalvz ah yes, sorry - on master.

PR #5171, opened 2021-04-27
automation: use the right variable name

v1v:
## Motivation/summary

PR got merged but there was a typo so the automation cannot happen as expected


axw (approved):

v1v:
This script is not validated in the CI at all, since it runs as part of another daily process. So I'll merge it now

Issue #5018, opened 2021-03-26
Introduce direct agent configuration


axw:
As part of the move to Fleet we will need to move away from fetching agent config directly from Kibana, as the privileges APM Server is given do not cover this. Instead, agent config will be pushed down to APM Server via the server's policy.

In order to make the above possible, we will add configuration to APM Server to directly specify agent config. This will include the criteria, config settings, and an optional Etag value. For example:

```yaml
apm-server:
  agent_config:
  - service.name: ten_percent
    config:
      transaction_sample_rate: 0.1
  - service.environment: production
    etag: abc123 # optional, computed by APM Server if not specified
    config:
      transaction_sample_rate: 0.5
```

If the "etag" property is not set for a config block, the server will compute one by hashing the settings. The property is configurable so that Kibana can inject a value. When creating/updating agent config in Kibana, it will take responsibility for calculating an Etag value and injecting it into the server config; Kibana will later use this for identifying whether config has been applied (more below).

Configuring a list of agent config will disable fetching agent config from Kibana.

Currently when APM Server queries Kibana for agent config, Kibana will mark config as having been applied when the provided Etag is found to be current. As we will no longer be communicating with Kibana, we'll need to write documents to Elasticsearch instead. The server will track whether in a config query from an agent, the agent supplies the current Etag (If-None-Match) for the config block that matches its criteria. When this is first true, the server will index a document including the Etag value indicating that the config has been applied.

Alternatively, we could have agents periodically send an internal agent statistics event which includes the current config Etag value. This would have the benefit of enabling us to see _how many_ and _which_ agents have applied the config.

felixbarny:
> Alternatively, we could have agents periodically send an internal agent statistics event which includes the current config Etag value. This would have the benefit of enabling us to see how many and which agents have applied the config.

I like that idea but I'd suggest treating that as an enhancement and not couple it to the initial implementation. This also assures we're backward compatible with older agents. All agents will need to add an `ephemeral_id` (currently only the Java agent does that IIRC).
Alternatively, APM Server could generate that event whenever it receives a config request from an agent which would require fewer changes in the agents (ephemeral_id would probably still be required (or we could auto-populate the ephemeral_id by hashing the metadata)). 

To show the count of agents that have successfully applied the config, Kibana could query for the unique count of `ephemeral_id`s for a given etag within the last 1-5 minutes. It should not look at all the data as a single service that frequently restarts would skew that number. Also, restricting the time range should lead to faster queries.

[...]

PR #5177, opened 2021-04-28
Direct agent configuration

stuartnelson3:
## Motivation/summary

Add support for direct agent configuration in apm-server

## Checklist

## How to test these changes

Build and run `apm-server` with `agent_configs` defined. `cURL` the server to verify the response makes sense for the incoming request.

`apm-server.yml`
```yaml
  default_service_environment: production
  agent_config:
  - service.name: ten_percent
    config:
      capture_body: off
      capture_body: true
      log_level: info
      recording: true
  - service.environment: production
    config:
      transaction_sample_rate: 0.5
  - service.environment: dev
    config:
      transaction_sample_rate: 0.1

```
verify the default_service_environment is respected
```
# curl 'http://localhost:8200/config/v1/agents?service.name=ten_percent'
{
  "capture_body": "true",
  "log_level": "info",
  "recording": "true",
  "transaction_sample_rate": "0.5"
}
```
verify alternative environments can be requested
```
# curl 'http://localhost:8200/config/v1/agents?service.name=ten_percent&service.environment=dev'
{
  "capture_body": "true",
  "log_level": "info",
  "recording": "true",
  "transaction_sample_rate": "0.1"
}
```
requests with a missing or unrecognized `service.name` return an error
```
# curl 'http://localhost:8200/config/v1/agents?service.name=dne'
{
  "error": "no config found for service.name=dne"
}
```
## Related issues

closes https://github.com/elastic/apm-server/issues/5018


beater/api/config/agent/handler.go
```diff
@@ -61,6 +63,122 @@ var (
 	rumAgents = []string{"rum-js", "js-base"}
 )
 
+// DirectConfigurationHandler returns a request.Handler for replying to agent
+// central configuration requests. The handler parses the incoming request and
+// responds with the config matching the service.name/service.environment pair.
+//
+// TODO: What if no matching service.name/service.environment pair is found?
```
stuartnelson3, line 70: I'm assuming we want to return some form of error, and the agent will continue to run with defaults?

_meta/beat.yml
```diff
@@ -273,6 +273,22 @@ apm-server:
       #index_pattern: "apm-*-sourcemap*"
 
   #---------------------------- APM Server - Agent Configuration ----------------------------
+  # Directly specify agent configuration. If `agent_config` is set, agent
+  # configuration under `kibana` will be ignored.
+  # TODO: Provide a link to documentation? or list all options?
```
stuartnelson3, line 278: ^

stuartnelson3:
I'm using an `md5` hash; I feel no particular affinity for it. If there's a standard / preferred hashing function let me know.

beater/api/config/agent/handler.go
```diff
@@ -61,6 +63,119 @@ var (
...
+			Settings: nameConf.Config,
+			Etag:     nameConf.Etag,
+			Agent:    name,
+		}, nil
+	}
+	merged := merge(nameConf.Config, envConf.Config)
+	m, err := json.Marshal(merged)
+	if err != nil {
+		return nil, err
+	}
```
axw, line 149: There should be no merging and the logic here should match what we do in Kibana, which is to find the first matching config block in descending order:

 - service name and environment both defined in criteria, both match queried/default service environment and name
 - only service name defined in criteria, matches the queried service name
 - only service environment defined in criteria, matches the queried/default service environment

axw (commented):
With the caveat that I haven't looked into this in any great detail, I feel like we should be able to reuse the existing HTTP handler and introduce a new interface that that parameterises the config source and matching. Basically create an interface that is satisfied by https://github.com/elastic/apm-server/blob/35939ff776ae81968f5fc4c07a5cd38a0abf9f3a/agentcfg/fetch.go#L70, and implement a new DirectAgentConfigFetcher (feel free to come up with a better name).

WDYT? Does this sound feasible?

stuartnelson3:
>With the caveat that I haven't looked into this in any great detail, I feel like we should be able to reuse the existing HTTP handler and introduce a new interface that that parameterises the config source and matching.
>WDYT? Does this sound feasible?

I went down that path actually, but then backed out when I saw the `validateClient()` code. I could drop that into fetch I suppose and change the interface args. At the time it felt like enough of a misnomer to have a `Fetch` interface method also do validation on the client that I went the easy route. I'll push it a change and see how it feels.

beater/api/config/agent/handler.go
stuartnelson3: >only service environment defined in criteria, matches the queried/default service environment

Based on the pre-existing code, this isn't allowed:
https://github.com/elastic/apm-server/blob/master/beater/api/config/agent/handler.go#L170-L172

>service name and environment both defined in criteria, both match queried/default service environment and name

So in this situation, whichever is defined first under the `agent_config` (name or environment) is the config block returned?

stuartnelson3: I think my confusion is stemming from `agent_config` example having `service.name` or `service.environment` in the issue, but the actual ui in kibana has you selecting both of these values.

axw: >> only service environment defined in criteria, matches the queried/default service environment
>
> Based on the pre-existing code, this isn't allowed:
https://github.com/elastic/apm-server/blob/master/beater/api/config/agent/handler.go#L170-L172

The agent must specify its service name. What I meant was that in the central config criteria the name is not specified.

> I think my confusion is stemming from agent_config example having service.name or service.environment in the issue, but the actual ui in kibana has you selecting both of these values.

In the UI there is a special value "All" for both name and environment, which is effectively the same as not setting it at all. So for the above scenario, in the existing UI that would be name set to "All" and environment set to something specific.

axw:
> I went down that path actually, but then backed out when I saw the validateClient() code. I could drop that into fetch I suppose and change the interface args. At the time it felt like enough of a misnomer to have a Fetch interface method also do validation on the client that I went the easy route. I'll push it a change and see how it feels.

I think it's OK for the fetcher implementation to internally perform some validation. That's a backend-specific implementation detail.

As for changing the args, not sure if that's necessary. Rather than having `validateClient` directly set the response status code and error message, perhaps we could instead have the fetcher implementation return a specific error type to indicate it's a ServiceUnavailable error? And I think the `withAuth` handling could be handled entirely in the HTTP handler code.

stuartnelson3:
Two things to note:

I updated the documentation but I have no idea what should actually be written.

From the issue:
>The server will track whether in a config query from an agent, the agent supplies the current Etag (If-None-Match) for the config block that matches its criteria. When this is first true, the server will index a document including the Etag value indicating that the config has been applied.

Is there an opinion how we want to handle this (assuming we want to do it in this PR)? The workflow seems straightforward: agent config requests comes in, if it has an etag that matches the etag for a config block in the server, index a document in ES indicating this has been applied. The specifics of which index to use and how the document should look are things I'm not sure of.

axw:
>> The server will track whether in a config query from an agent, the agent supplies the current Etag (If-None-Match) for the config block that matches its criteria. When this is first true, the server will index a document including the Etag value indicating that the config has been applied.
>
> Is there an opinion how we want to handle this (assuming we want to do it in this PR)? The workflow seems straightforward: agent config requests comes in, if it has an etag that matches the etag for a config block in the server, index a document in ES indicating this has been applied. The specifics of which index to use and how the document should look are things I'm not sure of.

I'm not sure on the specifics yet, and we'll need to coordinate with the UI team. Let's leave this to a followup PR? I think this one is getting big enough as it is.

beater/config/config.go
```diff
@@ -92,6 +100,59 @@ type Config struct {
 	DefaultServiceEnvironment string                  `config:"default_service_environment"`
 
 	Pipeline string
+
+	// TODO: wat is with this naming
+	AgentConfigs []ServiceConfig `config:"agent_config"`
+}
+
+// ServiceConfig defines configuration for agents.
+type ServiceConfig struct {
```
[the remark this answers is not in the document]
stuartnelson3: There's already an AgentConfig struct defined, any preference on how to navigate that?

https://github.com/elastic/apm-server/blob/master/beater/config/config.go#L85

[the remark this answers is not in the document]
axw: Ah bugger :( It may be confusing that there's both "agent_config" and "agent.config" in `apm-server.yml`. I'm not sure how to fix this. Current best thought is to rename AgentConfig to KibanaAgentConfig.

agentcfg/fetch.go
```diff
@@ -121,3 +183,56 @@ func containsAnyPrefix(s string, prefixes []string) bool {
...
+					Settings: cfg.Config,
+					Etag:     cfg.Etag,
+					Agent:    name,
+				}}, nil
+			}
+			nameConf = &f.scs[i]
+		}
+		if cfg.Service.Environment == env {
+			envConf = &f.scs[i]
+		}
```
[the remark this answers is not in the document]
stuartnelson3: my kingdom for reasonably modern pattern matching ..

stuartnelson3:
>Let's leave this to a followup PR? I think this one is getting big enough as it is.

that sounds like a good idea to me

stuartnelson3:
jenkins run the tests

agentcfg/fetch_test.go
```diff
@@ -151,3 +155,175 @@ func mockDoc(sampleRate float64) m {
...
+					Name:        "service1",
+					Environment: "production",
+				},
+			},
+			agentConfigs: []config.AgentConfig{
+				{
+					Service: config.Service{Name: "", Environment: "production"},
+					Config:  map[string]string{"key1": "val2", "key2": "val2"},
+					Etag:    "def456",
+				},
```
axw, line 177: maybe also add a block above with only Name set (and matching), showing that both Name & Environment always wins?

beater/config/config.go
```diff
@@ -39,6 +42,11 @@ const (
 	msgInvalidConfigAgentCfg = "invalid value for `apm-server.agent.config.cache.expiration`, only accepting full seconds"
 )
 
+var (
```
axw, line 45: Actually now that I think of it, it's allowed to have neither - sorry. That would match all services, and would have lowest precedence.

beater/config/config.go
```diff
@@ -92,6 +100,57 @@ type Config struct {
 	DefaultServiceEnvironment string                  `config:"default_service_environment"`
 
 	Pipeline string
+
+	// TODO: wat is with this naming
```
axw, line 104: I think it's OK now, and we can remove this? Maybe not ideal, perhaps we can come back to it later

agentcfg/fetch.go
```diff
@@ -121,3 +182,49 @@ func containsAnyPrefix(s string, prefixes []string) bool {
...
+
+	for i, cfg := range f.cfgs {
+		if cfg.Service.Name == name && cfg.Service.Environment == env {
+			nameConf = &f.cfgs[i]
+			break
+		} else if cfg.Service.Name == name && cfg.Service.Environment == "" {
+			nameConf = &f.cfgs[i]
+		} else if cfg.Service.Name == "" && cfg.Service.Environment == env {
+			envConf = &f.cfgs[i]
+		}
```
axw, line 212: As mentioned elsewhere, we'll also need a catchall case: where `cfg.Service.Name == "" && cfg.Service.Environment == ""`. That will have lowest precedence.

beater/api/mux.go
```diff
@@ -179,17 +180,29 @@ func (r *routeBuilder) rumAgentConfigHandler() (request.Handler, error) {
...
+		// with kibana.
+		f := agentcfg.NewDirectFetcher(cfg.AgentConfigs)
+		h = agent.NewHandler(f, cfg.KibanaAgentConfig, cfg.DefaultServiceEnvironment)
+	} else {
+		var client kibana.Client
+		if cfg.Kibana.Enabled {
+			client = kibana.NewConnectingClient(&cfg.Kibana)
+		}
+
+		f := agentcfg.NewFetcher(client, cfg.KibanaAgentConfig.Cache.Expiration)
```
axw, line 196: When we come to fixing up the jaeger bits, we'll probably want to extract this logic into a function in the `beater` package. Then we can pass the fetcher into the places it's needed: `beater/mux`, `beater/jaeger`, and the gRPC server in the `beater` package itself.

agentcfg/fetch.go
```diff
@@ -66,8 +76,59 @@ func NewFetcher(client kibana.Client, cacheExpiration time.Duration) *Fetcher {
...
+func (v *ValidationError) Keyword() string { return v.keyword }
+
+// Body returns the body for the ValidationError.
+func (v *ValidationError) Body() string { return v.body }
+
+// Error() implements the error interface.
+func (v *ValidationError) Error() string { return v.err.Error() }
+
+// Validate validates the currently configured Fetcher.
+func (f *Fetcher) validate(ctx context.Context) *ValidationError {
```
axw, line 96: Doesn't have to be in this PR, but we should rename Fetcher to KibanaFetcher or something similar.

beater/api/config/agent/handler.go
```diff
@@ -34,119 +35,109 @@ import (
...
+}
 
-		result, err := fetcher.Fetch(c.Request.Context(), query)
-		if err != nil {
-			apm.CaptureError(c.Request.Context(), err).Send()
-			extractInternalError(c, err, c.AuthResult.Authorized)
-			c.Write()
-			return
-		}
+type fetcher interface {
```
axw, line 80: Seeing as it crosses package boundaries, can we unexport this please? Maybe define it in `agentcfg`?

axw (commented):
Basically LGTM, but I missed a couple of biggish things in earlier reviews

beater/config/config.go
stuartnelson3: yeah, definitely not a comment that should make it into master :)

agentcfg/fetch.go
stuartnelson3: Wouldn't that result in an empty config? In which case, the initial assignment `result := zeroResult()` would cover it?

beater/api/config/agent/handler.go
stuartnelson3: I'll move it over there, but then wouldn't it have to be exported in order to be listed as an argument?

beater/config/config.go
stuartnelson3: In which case we return the first agent_config?

agentcfg/fetch.go
stuartnelson3: Read your comment [above](https://github.com/elastic/apm-server/pull/5177#discussion_r627238743), will wait for clarification

agentcfg/fetch.go
stuartnelson3: No time like the present

beater/api/mux.go
stuartnelson3: Ended up putting it in `agentcfg`. The logic was entirely based on stuff in the `config` package, and it was returning a `Fetcher`. Happy to put it elsewhere of course.

beater/api/config/agent/handler.go
axw: I guess you figured it out already, but I mistyped - I meant "export" rather than "unexport". Looks good now, thanks

beater/config/config.go
axw: Here's how it should work. Return the first matching AgentConfig in the list, evaluating rules in order of precedence:

1. AgentConfig with both service name and environment set, matching the queried service name and environment
2. AgentConfig with only service name set, matching the queried service name
3. AgentConfig with only service environment set, matching the queried service environment
4. AgentConfig with neither service name nor environment set

If none of those resolve, return an empty result.

That follows what we do in the UI. You can select "Any" for service name and/or environment in the UI, which is equivalent to leaving that property unset in the agent_config entry.

stuartnelson3: Alright, think I (hopefully) got it now

axw (approved):
LGTM! :tada:

axw:
Ran the following tests, with a healthy dose of forcible restarts of APM & Fleet through the cloud console due to https://github.com/elastic/apm-server/issues/5665, the fix for which has not yet been picked up in 7.14-SNAPSHOT.

1. Created a 7.14-SNAPSHOT deployment
2. Ran a program to watch for agent config changes
3. Created agent config for all services/environments, sampling rate 0.5; observed program pick up the change
4. Ran migration to Fleet integration

Policy appears to be created without agent config:

image

5. Update APM agent config, policy is updated:

image

6. Observe the sampling rate is picked up by the watcher program
7. Delete APM agent config, observe change in watcher program
8. Create two new APM agent configs: one for all services again, and one for the watcher program's specific service name; observe config for the latter

I observed some problems in Kibana (I'll open issues), but the server side looks good.

PR #5190, opened 2021-04-30
mergify: fix delete_head_branch condition

axw:
Check that the name of "head" matches, rather than "base".


axw:
jenkins run the tests please

v1v (approved):
Good finding! :)

Issue #4487, opened 2020-12-03
Define units and types for well-defined metric fields


axw:
We should update `fields.yml` with units and types (gauge/counter) for metric fields. See https://github.com/elastic/package-spec/pull/73

Although this metadata shouldn't be needed by the APM app in Kibana, it would (in the future) help with ad-hoc analysis through Lens and Discover, which will (in the future) use the metadata to format and aggregate data appropriately.

axw:
Prerequisite for adding field metadata to ES templates: https://github.com/elastic/beats/pull/25419
We will need to do something for fields defined in the integration package too.

jalvz:
This would be something like

```diff
diff --git a/model/metricset/_meta/fields.yml b/model/metricset/_meta/fields.yml
index 69c4d7cc4..c55a28517 100644
--- a/model/metricset/_meta/fields.yml
+++ b/model/metricset/_meta/fields.yml
@@ -165,7 +165,6 @@
           overwrite: true

         - name: duration
-
           type: group
           fields:
             - name: count
@@ -177,11 +176,12 @@
               fields:
                 - name: us
                   type: long
+                  unit: micros
+                  metric_type: gauge
                   overwrite: true
                   description: Aggregated transaction duration, in microseconds.

         - name: self_time
-
           type: group
           description: >
             Portion of the transaction's duration where no direct child was running
@@ -195,13 +195,14 @@
               fields:
                 - name: us
                   type: long
+                  unit: micros
+                  metric_type: gauge
                   overwrite: true
                   description: >
                     Aggregated transaction duration, excluding the time periods where a
                     direct child was running, in microseconds.

         - name: breakdown
-
           type: group
           fields:
             - name: count
@@ -235,7 +236,6 @@
           overwrite: true

         - name: self_time
-
           type: group
           description: >
             Portion of the span's duration where no direct child was running
@@ -249,6 +249,8 @@
             fields:
               - name: us
                 type: long
+                unit: ms
+                metric_type: gauge
                 overwrite: true
                 description: >
                   Aggregated span duration, excluding the time periods where a
diff --git a/model/profile/_meta/fields.yml b/model/profile/_meta/fields.yml
index 79fd1c453..39e3de3da 100644
--- a/model/profile/_meta/fields.yml
+++ b/model/profile/_meta/fields.yml
@@ -632,8 +632,9 @@
             All samples within a profile will have the same profile ID.

         - name: duration
-
           type: long
+          unit: micros
[...]

PR #5230, opened 2021-05-05
Add metric_type and unit to system metrics

axw:
## Motivation/summary

Add `metric_type` and `unit` to system metric fields, so Kibana will be able to aggregate and format them appropriately.

Also update `apmpackage/cmd/gen-package` to copy across these field properties.

## Checklist

## How to test these changes

1. `apm-server export template`
2. Check the template includes `metric_type` and `unit` field metadata for `system.*` metrics

## Related issues

https://github.com/elastic/apm-server/issues/4487


jalvz (approved):
There is a package changelog too. I'm not sure how do you want to go about it, we can dup relevant lines or remember to do it before publishing.

axw:
@jalvz do you know how/where that is used? In any case I'd prefer to have a single source of truth, so let's update/generate it later.

simitt:
tested with BC2 - as expected

PR #5239, opened 2021-05-06
Add support for dynamic histogram metrics

axw:
## Motivation/summary

Introduce support for specifying metric type and unit.

If these are set for any metrics in a metricset, we will create a `_metric_descriptions` field which is used in a new ingest processor. The ingest processor uses these properties to set `dynamic_templates`. Currently we only understand the type "histogram", and unit is ignored. Later we will add support for "summary" metrics, and for setting field metadata for metric type (gauge, counter), and for units (byte, percent, nanos, etc.)

The OTLP processor is updated to consume histogram metrics accordingly. We will follow up with support for the Elastic APM metrics intake in another PR.

## Checklist

## How to test these changes

1. Run apm-server
2. Instrument an application with OpenTelemetry, capturing metrics configured to be sent as histograms

## Related issues

https://github.com/elastic/apm-server/issues/3195


model/metricset.go
```diff
@@ -96,6 +106,20 @@ type Sample struct {
...
+	// If Type is unspecified or invalid, it will be ignored.
+	Type MetricType
+
+	// Unit holds an optional unit:
+	//
+	// - "percent" (value is in the range [0,1])
+	// - "byte"
+	// - a time unit: "nanos", "micros", "ms", "s", "m", "h", "d"
+	//
+	// If Unit is unspecified or invalid, it will be ignored.
```
simitt, line 120: assuming this will hold true once the `Unit` is actually filled

processor/otel/metrics.go
```diff
@@ -157,6 +179,23 @@ func (c *Consumer) addMetric(metric pdata.Metric, ms *metricsets) bool {
 	return false
 }
 
+func histogramSample(metricName string, bucketCounts []uint64, explicitBounds []float64) model.Sample {
+	values := make([]float64, 0, len(explicitBounds))
+	counts := make([]int64, 0, len(bucketCounts))
+	for i, count := range bucketCounts {
+		if count > 0 {
+			counts = append(counts, int64(count))
+			values = append(values, explicitBounds[i])
```
simitt, line 188: from the otel code:
> 	// The number of elements in bucket_counts array must be by one greater than
	// the number of elements in explicit_bounds array.

Wouldn't that lead to an index-out-of-bound exception for `explicitBounds[i]`?

stuartnelson3 (approved):

stuartnelson3:
Dis-regard my approval, @simitt is bringing her expert knowledge to bear :)

processor/otel/metrics.go
stuartnelson3: The tests should also be updated if this is the case, as the slices representing `bucket_counts` and `explicit_bounds` are equal length

model/metricset.go
axw: Unit is currently always ignored. The unit is added to the doc, but gets removed by the ingest pipeline.

simitt (approved):
left one minor comment, apart from that LGTM

processor/otel/metrics.go
```diff
@@ -162,6 +188,69 @@ func (c *Consumer) addMetric(metric pdata.Metric, ms *metricsets) bool {
...
+	// histogram_quantile function (https://prometheus.io/docs/prometheus/latest/querying/functions/#histogram_quantile)
+	// to achieve consistent percentile aggregation results:
+	//
+	// "The histogram_quantile() function interpolates quantile values by assuming a linear
+	// distribution within a bucket. (...) If a quantile is located in the highest bucket,
+	// the upper bound of the second highest bucket is returned. A lower limit of the lowest
+	// bucket is assumed to be 0 if the upper bound of that bucket is greater than 0. In that
+	// case, the usual linear interpolation is applied within that bucket. Otherwise, the upper
+	// bound of the lowest bucket is returned for quantiles located in the lowest bucket."
+	values := make([]float64, 0, len(explicitBounds))
```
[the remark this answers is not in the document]
axw: indeed, another good catch :) fixed, thanks

simitt:
Tesetd with BC 2 - histogram metrics are ingested, and dynamic mapping is added

Screenshot 2021-07-12 at 11 06 00

PR #5264, opened 2021-05-17
idxmgmt: fix ILM support detection

### LINKED ISSUE #5259: idxmgmt: transient errors while checking ILM enablement cause auto ILM support to disable
https://github.com/elastic/apm-server/blob/2b50e255482c595e8d7ae81a59cd89b15306b157/idxmgmt/supporter.go#L186

Because we swallow these "CheckEnabled" errors, ILM support is disabled and APM Server continues to start up and leads to creation of time-based indices.

axw:
## Motivation/summary

If Elasticsearch is unavailable while checking for ILM support, we must not swallow errors and take them to mean ILM is not supported. This would lead to APM Server writing time-based indices for the remainder of an established connection.

Simplify idxgmmt package, and bring it closer into line with how the libbeat version works. Move ILM support checks into `Manager.Setup`, which will now return ILM check errors.

## Checklist

## How to test these changes

This is a bit hard to test directly, as it depends on an error occurring with Elasticsearch at just the right time when calling the `/_xpack` endpoint to check that ILM is supported and enabled in Elasticsearch.

Perhaps: place a reverse proxy between apm-server and Elasticsearch which intermittently returns an error (not 400! this goes down a different code path) for `/_xpack`, and continuously restart apm-server and send some events, making sure no time-based indices are ever created.

## Related issues

Fixes #5259


simitt (approved):
Haven found time yet to try it out, but code changes LGTM. Great finding and improvement!

jugsofbeer:
crossing fingers this makes it into  v7.13.0

axw:
Out of an abundance of caution, we're going to hold this until 7.13.1, to avoid disrupting the 7.13.0 release with this non-trivial bug fix. In the mean time, the workaround for this bug is to explicitly enable/disable ILM in apm-server.yml.

stuartnelson3:
Tested with BC2. Two indices, `apm-7.14.0-onboarding-2021.07.15` and `apm-7.14.0-onboarding-2021.07.14`, appeared during the course of testing.

initial:
image
after testing:
image


Used this reverse-proxy: https://gist.github.com/stuartnelson3/528efb313bb1aeaeabf72afbb6223846
Ran an infinite loop of:
- Send 4 transactions, wait 5 sec
- Restart server, wait 5 sec

Let this run for 30min or so.

Is this what we're expecting?

axw:
Yes, because "onboarding" docs go into daily indices - they do not use ILM. This issue only applies to transactions, errors, spans, metrics.

stuartnelson3:
Cool, then I'll mark this as "confirmed" since I could not reproduce it.

ppf2:
Switched the label from 7.13.0 to 7.13.1 to indicate that the first version with the fix is 7.13.1.

PR #5267, opened 2021-05-17
Remove/replace `processor/stream/package_tests`

### LINKED ISSUE #4408: Rewrite processor/stream/package_tests as functional tests
https://github.com/elastic/apm-server/tree/master/processor/stream/package_tests tests the coverage of our Elasticsearch mappings, ensuring that the fields in indexed documents are defined in our templates, and the fields defined in the templates are covered by our test data.

In order to do this, the tests mock out several key parts of the server (publish & libbeat pipeline), and reimplement parts of processor/stream (decoding, validating).

Furthermore, the tests are indirectly checking the mappings by loading `fields.yml`. Once we're fully running under Elastic Agent, `fields.yml` will no longer be contained within apm-server.

We should reimplement these tests as functional tests, which send data to a real, running APM Server, and check the results are as expected, e.g. by fetching the index templates from Elasticsearch and comparing to stored documents. This will avoid having to mock and reimplement parts of the server, improving test coverage of the production code paths.

axw:
## Motivation/summary

Remove/replace `processor/stream/package_tests` and associated code in package `tests`.

`AttrsPresence` tests have simply been removed as they were redundant with introduction of `modeldecoder` tests.

`KeywordLimitation` tests have been replaced with a system test that checks the installed template rather than fields.yml.
The test is simplified to check *all* keywords have ignore_above set to 1024, with explicit exceptions specified. At present there is just one ECS field that does not set ignore_above to 1024.

`PayloadAttrsMatchFields` tests have been replaced with a system test that ingests all supported event types, and ensures that mappings are covered by the installed index template.

We do not check that all attributes in the Elasticsearch template are covered by the payloads. This made sense when the input and output schemas were more aligned, but we have continued to diverge. This part has not been replaced.

## How to test these changes

N/A

## Related issues

Closes elastic/apm-server#4408


axw:
jenkins run the tests please

systemtest/template_test.go
```diff
@@ -0,0 +1,194 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/elastic/apm-server/systemtest"
+	"github.com/elastic/apm-server/systemtest/apmservertest"
+	"github.com/elastic/apm-server/systemtest/estest"
+	"github.com/elastic/go-elasticsearch/v7/esapi"
+)
+
+func TestIndexTemplateKeywords(t *testing.T) {
```
simitt, line 39: FYI - I believe the `ignore_above` in the template is set within the libbeat code, and was not the main focus of the `1024` limitation test. 
IMO the important part is to test that the limit is set on the Intake API for fields that are indexed. But as you pointed out in the description, the Intake API and the indexed doc structure diverged quite a bit, so there would be some manual mapping required (which I agree we should get rid of). So in conclusion, I think it is fine to not test this on the Intake API anymore, but we might also open a discussion then if we even want to set the limit on the Intake API, or just align with the index setting - allow indexing a value > `1024`, but ignoring everything above the limit. WDYT?

systemtest/estest/search.go
```diff
@@ -43,6 +43,9 @@ func (es *Client) ExpectMinDocs(t testing.TB, min int, index string, query inter
 	var result SearchResult
 	opts = append(opts, WithCondition(result.Hits.MinHitsCondition(min)))
 	req := es.Search(index)
+	if min > 10 {
+		req = req.WithSize(min)
+	}
```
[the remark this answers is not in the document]
axw: That's the default size for searches. The parameter is the minimum number of to wait for, so for <=10 we wait for at least min and return all. For > 10 we wait for and return min.

systemtest/template_test.go
axw: > I believe the ignore_above in the template is set within the libbeat code

Ah thanks, I didn't realise there was a default set by libbeat. In that case, let's remove this test and also remove ignore_above (where it's 1024) from all of our fields.yml definitions?

> but we might also open a discussion then if we even want to set the limit on the Intake API, or just align with the index setting - allow indexing a value > 1024, but ignoring everything above the limit. WDYT?

My gut feeling is that we should keep the limit as it is (AFAIK we haven't had any complaints?), but worth raising with the agent devs.

systemtest/estest/search.go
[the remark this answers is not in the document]
axw: I'm going to modify ExpectMinDocs to return all hits regardless of the value of min.

systemtest/template_test.go
simitt: Verified that by default `ignore_above: 1024` is set in [libbeat](https://github.com/elastic/beats/blob/master/libbeat/template/processor.go#L265-L271) for `keywords`.

Agreed that we can remove the limits from the `fields.yml` files (maybe in a follow up PR).

simitt (approved):
Thanks for cleaning this up. 

I leave it up to you to remove or merge with the `TestIndexTemplateKeywords`.

PR #5269, opened 2021-05-18
Remove tests/loader

axw:
## Motivation/summary

Remove custom test helpers which do not pay their weight (diff removes more than it adds). Instead, use `ioutil.ReadFile` directly and expect tests to specify paths relative to their package directory.

## How to test these changes

N/A

## Related issues

None.


stuartnelson3 (approved):
:100:

PR #5273, opened 2021-05-18
Move tests.MockKibanaClient to its own package

axw:
## Motivation/summary

Continuing on my crusade against the `tests` package. As it is, the package is a bit of a "utils" package; it has testing functions/types for an assortment of unrelated requirements. Separating them into their own test packages by use-case closer to the code to which they are related, a la "httptest", should clarify their purpose.

## How to test these changes

N/A

## Related issues

None.


axw:
@Mergifyio backport 7.x

PR #5295, opened 2021-05-20
[DOC] Fixing broken box for RUM

insukcho:
I'm not sure how to change RUM's first column background color to gray, but this will fix the broken box with the right columns' order from left to right


bmorelli25 (approved):
Thank you, @insukcho!

Issue #5275, opened 2021-05-18
Create one metrics document with jvm metrics per opentelemetry event


simitt:
Follow up from https://github.com/elastic/apm-server/pull/4986#issuecomment-843050707: when translating JVM metrics one document with all JVM metrics should be created, rather than creating one document per metric, as otherwise JVM metrics do not show up in the UI as expected.

axw:
I think the simplest thing to do will be to remove the ingest pipeline I added for 7.13, and do it in code in processor/otel. That's where we split metrics by labels currently.

PR #5309, opened 2021-05-24
Translate OpenTelemetry JVM metrics in processor/otel

axw:
## Motivation/summary

Remove the ingest pipeline which translates OpenTelemetry Java metrics, and do this in processor/otel instead. This ensures the copied metrics are grouped by their labels/dimensions, like all other metrics.

## Checklist

## How to test these changes

1. Run apm-server
2. Instrument a Java application with https://github.com/open-telemetry/opentelemetry-java-instrumentation, pointed at apm-server
3. Wait for metrics to be collected and sent to APM Server (not sure how long this is)
4. Navigate to the APM app in Kibana, and check the GC and memory usage graphs are populated
5. Check that the apm metrics docs are grouped into docs by labels

See also #4919

## Related issues

Closes https://github.com/elastic/apm-server/issues/5275


processor/otel/metrics.go
```diff
@@ -169,15 +174,66 @@ type stringMapItem struct {
...
+		// Chop off the "runtime." prefix, i.e. runtime.jvm.gc.time -> jvm.gc.time.
+		// OpenTelemetry and Elastic APM metrics are both defined in milliseconds.
+		elasticapmSample := sample
+		elasticapmSample.Name = sample.Name[len("runtime."):]
+
+		// Copy label "gc" to "name".
+		var elasticapmLabels []stringMapItem
+		for _, label := range labels {
+			if label.key == "gc" {
+				elasticapmLabels = []stringMapItem{{key: "name", value: label.value}}
```
simitt, line 228: nit: This changes the previous behavior to not override the `labels.name` field if it exists.

simitt (approved):

processor/otel/metrics.go
axw: I think the old implementation was probably a bit over the top in that regard. The labels come from the OpenTelemetry Java instrumentation which won't set the name label. And even if it did, we should just disregard it. We're creating a new metric here, rather than adding a field to an existing metric doc.

simitt:
Testing with BC2 - jvm memory metrics show up, but no garbage collection metrics, despite metrics documents with gc values exist. 

Screenshot 2021-07-13 at 13 44 03

**Example docs:**
``` 
{
  "runtime": {
    "jvm" : {
      "memory" : {
        "non_heap" : {
          "committed" : 1.53894912E8,
          "used" : 1.50294384E8
        },
        "heap" : {
          "committed" : 6.37534208E8,
          "max" : 1.7179869184E10,
          "used" : 5.5478808E7
        }
      }
    },
    "agent" : {
      "name" : "opentelemetry/java",
      "version" : "1.0.1"
    }
}
```
and garbage collection
```
{
  "runtime" : {
    "jvm" : {
      "gc" : {
        "collection" : 28
      }
    }
  },
  "agent" : {
    "name" : "opentelemetry/java",
    "version" : "1.0.1"
  }
}
```

simitt:
Tested with BC 2 - JVM memory metrics show up as expected, but garbage collection boxes are empty. 
Querying metrics docs shows that garbage collection metrics are collected, so it might be a UI issue. 
Screenshot 2021-07-16 at 09 43 40
Screenshot 2021-07-16 at 09 43 49

simitt:
reported in https://github.com/elastic/kibana/issues/105923

PR #5331, opened 2021-05-25
[6.8] bump version

mdelapenya:
## What does this PR do?
It bumps Beats version.

On CI, it also allows forcing a specific (an older) version of python for the Windows workers.

## Why is it important?
CI builds on windows are failing (See #5324)

## Related issues
- Supersedes #5324


script/jenkins/windows-test.ps1
```diff
@@ -28,7 +28,7 @@ $env:PATH = "$env:GOPATH\bin;C:\tools\mingw64\bin;$env:PATH"
 $env:MAGEFILE_CACHE = "$env:WORKSPACE\.magefile"
 
 # Setup Python.
-exec { choco install python2 -y -r --no-progress --version 2.7.17 }
+exec { choco install python2 -y -r --force --no-progress --version 2.7.17 }
```
mdelapenya, line 31: OTOH, Python 2.7.18 is installed in the workers. We could simply remove the install command and use workers' Python. Thoughts?

cachedout (approved):

script/jenkins/windows-test.ps1
v1v: I guess, we could do so

v1v (approved):

script/jenkins/windows-test.ps1
mdelapenya: Will wait for the @elastic/apm-server team for clarification about the desired Python version. Will remove the installation  if they agree.

jalvz: that should be fine, but going to merge this nevertheless 👍

jalvz (approved):

Issue #4895, opened 2021-03-02
[fleet] Add `apm-server.url` to the APM fleet configuration 


simitt:
When running on cloud or via ECK the publicly reachable APM Server `URL` differs from the server's `host`. 
Since the public URL is not visible anywhere else, we need to ensure it shows up in the APM Server integration in the Fleet UI. 

~The APM Setup Guide currently embeds the APM Server URL + secret token for user convenience when setting up APM agents and will have to retrieve this information from the agent policy (see https://github.com/elastic/kibana/issues/89989).~

The public URL can be injected when the managed policy is created. It doesn't have any direct impact on the APM Server itself.

axw:
@simitt instead of adding another var to the policy template, could we have ESS/ECK/etc. add a "config" value (as opposed to a var defined in the policy template)?

In my prototype branch for injecting agent config into policies, there's an example of adding arbitrary config:

https://github.com/elastic/kibana/blob/6205fbe62eb7e0bfcbb2ef48e8745c538c2b7aee/x-pack/plugins/security_solution/server/endpoint/services/artifacts/manifest_manager/manifest_manager.ts#L370

simitt:
Good idea - I believe this should be possible.

axw:
@simitt if we can do that then there will be no changes required in apm-server or the integration package. The config will effectively be a conduit between ESS and code in Kibana.

Can we close this and reopen if needed?

simitt:
I updated the description accordingly. We downscaled the efforts around the Getting Started guide, which will not show the APM Server URL + secret token for APM Server integrations, but only link to the Fleet UI. 

Config values do not show up in the UI (afaik), therefore I keep my proposal to add a `URL` config option. It should default to the `hosts` value, will be injected in cloud, and can be set for on-prem installations. It only has an informal purpose.

axw:
OK. It might be a bit confusing that you can edit it, but I suppose it's fine to start with. We can revisit it when when implement an APM-specific policy editor.

Issue #4898, opened 2021-03-03
Align package version with Stack version


axw:
We will publish new APM integration packages along with the stack.

Not yet clear if we would need to do this on *every* stack release (e.g. patch releases). The main point is that any new package should be published at the same time as a stack release, and have the same version as that stack release.

simitt:
related https://github.com/elastic/kibana/issues/93393

axw:
Done in https://github.com/elastic/apm-server/commit/0502a9b9076d795c727ee31dc4eab3ab23e96167

PR #5332, opened 2021-05-25
add apm-server url to handlebars tmpl

stuartnelson3:
## Motivation/summary

Display a publicly-reachable url for apm-server when running in cloud or eck

closes https://github.com/elastic/apm-server/issues/4895

## Checklist

## How to test these changes

1. Start apm-server under fleet
2. Go to the apm-server integration in the fleet ui
3. Verify that the url is visible


axw (approved):

jalvz:
shouldn't we be rolling version 0.3.0 ?

axw:
@jalvz we should be bumping to 7.14.0: https://github.com/elastic/apm-server/issues/4898. That can happen independently of this change; 0.2.0 is on the 7.13 branch.

axw:
@stuartnelson3 can you please remove apmpackage/apm/0.2.0? That version lives on the 7.13 branch.

Slightly longer term I think we should probably remove the version from the path in the apm-server repo, and only add it when copying to package storage.

stuartnelson3:
removed :ok_hand:

apmpackage/apm/0.3.0/manifest.yml
```diff
@@ -47,6 +47,12 @@ policy_templates:
             required: true
             show_user: true
             default: localhost:8200
+          - name: url
+            type: text
+            title: URL
+            required: true
+            show_user: true
+            default: localhost:8200
```
axw, line 55:
```suggestion
            default: http://localhost:8200
```

?

axw (approved):

apmpackage/apm/0.3.0/manifest.yml
stuartnelson3: I remember someone saying that it should default to the same value as `host`, but I'm happy to add the `http://`. Not sure what the "?" is asking :)

simitt: @stuartnelson3 I mentioned to default to the `host` value, but am +1 on adding `http://`

axw: that was vague way of saying I'm not 100% sure, but I think it should be a proper URL

simitt (approved):

jalvz:
> Slightly longer term I think we should probably remove the version from 
the path in the apm-server repo, and only add it when copying to package storage

we are runing `elastic-package` thou, which expects a version folder.

axw:
@jalvz which command specifically?

If I run `elastic-package lint` in `apmpackage/apm` I get:

```
int the package
Error: package root not found
```

If I run in the `apmpackage/apm/0.3.0` I get:

```
Lint the package
Done
```

If I rename "0.3.0" to "foobar", it still works; `elastic-package lint` doesn't appear to be traversing up to find a version directory.

axw:
I've observed the URL showing up in the integration editor UI, but I think we should probably test this in ESS/ECE.

simitt:
Tested on ESS with with latest `7.14` snapshot - URL shows up in the UI.
Screenshot 2021-07-12 at 11 11 51

Issue #5094, opened 2021-04-14
[OpenTelemetry] Map OpenTelemetry `messaging.*` semantic conventions to Elastic Common Schema


cyrille-leclerc:
Issue reported by https://github.com/gbanasiak on https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/3093

**Is your feature request related to a problem? Please describe.**

https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/671 introduced translation of `http.*` and `db.*` semantic conventions into Elastic equivalent, i.e. `span.destination.*` fields. The missing bit are `messaging.*` [conventions ](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/messaging.md) which make service map presentation of services using components such as Kafka or RabbitMQ incomplete.

**Describe the solution you'd like**

Expand [existing translation](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.24.0/exporter/elasticexporter/internal/translator/elastic/traces.go#L293-L349) to include `messaging.*` conventions.

**Describe alternatives you've considered**

The alternative would be to use semantic convention for which translation exists (e.g. `db.*`) while manually instrumenting messaging spans, but that's hacky.

cyrille-leclerc:
@axw was it fixed in 7.14 or 7.15?

axw:
@cyrille-leclerc https://github.com/elastic/apm-server/pull/5334 was merged into 7.14.

tiny-dancer:
@axw  minor this is still appearing in the limitations documentation: https://www.elastic.co/guide/en/apm/guide/current/open-telemetry.html#open-telemetry-known-limitations

axw:
@tiny-dancer thanks for the heads up. There's still an issue remaining, which I'll write up and link from the docs soon. The gist is that OTel doesn't provide enough information for us to always correctly classify an OTel span as an Elastic APM transaction/span.

PR #4955, opened 2021-03-15
Translate more OpenTelemetry resource conventions

axw:
## Motivation/summary

Translate more OpenTelemetry resource conventions, focusing on common conventions which already have a clear mapping to our data model. Add fields to metadata model types for missing ECS fields.

## Checklist

- [x] Update [CHANGELOG.asciidoc](https://github.com/elastic/apm-server/blob/master/CHANGELOG.asciidoc)

## How to test these changes

1. Run an OpenTelemetry-instrumented application with rich instrumentation (e.g. Java auto-instrumentation), pointed at apm-server
2. Wait for events to be recorded, check:
   - Labels for untranslated attributes
   - Captured field values are sensible

## Related issues

https://github.com/elastic/apm-server/issues/4714

PR #5334, opened 2021-05-25
translate otel messaging.* to ecs

stuartnelson3:
## Motivation/summary

Translate relevant otel messaging.* semantic conventions to their
elastic-equivalent fields.

closes https://github.com/elastic/apm-server/issues/5094

## Checklist

## How to test these changes

Ingest messaging spans from an otel-instrumented messaging service.

EDIT: https://github.com/open-telemetry/opentelemetry-go-contrib/tree/main/instrumentation/github.com/Shopify/sarama/otelsarama/example

Check the transactions and spans in the kibana UI for the following:

transaction:
- `transaction.type=messaging`
- Message metadata has fields `queue.name`, `system`, and `operation`.

span:
- `span.type=messaging`
- `span.subtype` is the framework name, eg. kafka
- `span.action` is one of `receive`, `send`, or `process`
- Message metadata has fields `queue.name`.


stuartnelson3:
Opening this to start discussion. I'm modeling this after [prior art](https://github.com/elastic/apm-server/pull/4955), but also adding specific transaction/span translation according to https://github.com/elastic/apm/blob/master/specs/agents/tracing-instrumentation-messaging.md, which the PR I was looking at did not have.

The questions I have are:

- Should I be doing the transaction/span translation
- Which of the fields from [the message attributes](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/messaging.md#messaging-attributes) should we be including in the `Message` object
- I'm inferring the `send` operation based on what I saw in the otel spec (linked in previous bullet), but I'm not sure if that's correct

model/message.go
```diff
@@ -29,6 +31,27 @@ type Message struct {
 	Headers   http.Header
 	AgeMillis *int
 	QueueName string
+	// TODO: Do we want to add all these keys to Message? Or should they be
+	// parsed as individual strings and added to the span in
+	// processor/otel/consumer.go?
+	RoutingKey string
+	System     string
+	Operation  string
```
axw, line 39: Can we start out with just the bare minimum, and then expand from there in another PR? The bare minimum being the intersection with what we capture for Elastic APM agents already: queue/topic name.

model/message.go
```diff
@@ -29,6 +31,27 @@ type Message struct {
...
+	return fmt.Sprintf("%s %s %s %s",
+		m.System, strings.ToUpper(m.Operation), m.direction(), m.QueueName,
+	)
+}
+
+func (m *Message) direction() string {
+	if m.Operation == "send" {
+		return "to"
+	}
+	return "from"
```
axw, line 54: Can we keep this logic in `processor/otel` until we need to share it with something else?

model/metadata.go
```diff
@@ -29,6 +29,7 @@ type Metadata struct {
 	UserAgent UserAgent
 	Client    Client
 	Cloud     Cloud
+	Message   Message
```
axw, line 32: Our "Metadata" is roughly equivalent to OpenTelemetry's "Resource" - information about the service, or the environment in which the service is operating. Message details don't belong here, as they're operation-specific.

processor/otel/consumer.go
```diff
@@ -633,6 +671,14 @@ func translateSpan(span pdata.Span, metadata model.Metadata, event *model.Span)
 		event.DB = &db
 	case isMessagingSpan:
 		event.Type = "messaging"
+		event.Subtype = message.System
+		event.Action = message.Operation
+		event.Name = message.Name()
```
axw, line 676: We use the OpenTelemetry span's name everywhere else - I think we should do the same here (and not set `event.Name` here.)

processor/otel/metadata.go
```diff
@@ -118,6 +118,20 @@ func translateResourceMetadata(resource pdata.Resource, out *model.Metadata) {
...
+		case conventions.AttributeMessagingDestination:
+			out.Message.QueueName = truncate(v.StringVal())
+		case conventions.AttributeMessagingOperation:
+			s := v.StringVal()
+			if s == "" {
+				s = "send"
+			}
+			out.Message.Operation = truncate(s)
+		case "messaging.rabbitmq.routing_key":
+			out.Message.RoutingKey = truncate(v.StringVal())
```
axw, line 133:
```suggestion
```

messaging attributes shouldn't exist on resources, only on spans

processor/otel/consumer.go
```diff
@@ -633,6 +671,14 @@ func translateSpan(span pdata.Span, metadata model.Metadata, event *model.Span)
...
+		// TODO: This is what we want to do? Following agent messaging spec
+		// https://github.com/elastic/apm/blob/master/specs/agents/tracing-instrumentation-messaging.md
```
axw, line 678: Is the TODO regarding destination service resource? If so, yes I think so.

processor/otel/consumer.go
```diff
@@ -296,12 +296,31 @@ func translateTransaction(
...
+					// and various examples
+					// https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/messaging.md#apache-kafka-example
+					stringval = "send"
+				}
+				message.Operation = stringval
+				isMessagingSpan = true
+			case conventions.AttributeMessagingSystem:
+				message.System = stringval
+				destinationService.Resource = stringval
+				destinationService.Name = stringval
```
axw, line 322: does this compile? transactions shouldn't have a destination service, only spans should

axw (commented):
The big question for me is: when do we create a transaction or a span? At the moment we only create a transaction for root otel spans, or otel spans with span kind SERVER. I think we should probably also create transactions for span kind CONSUMER... but only in some cases? Like you might consume a message within a transaction, and in that case it should be a span.

processor/otel/consumer.go
stuartnelson3: Ah, I cargo-culted that from the span code and forgot to check if it recompiled (it doesn't, and all of that code has been removed now anyway)

stuartnelson3:
>I think we should probably also create transactions for span kind CONSUMER... but only in some cases? Like you might consume a message within a transaction, and in that case it should be a span.

Wouldn't the current logic hold then? Whether span type is `SpanKindCONSUMER` or not, it being the root span is still valid and will branch correctly between transaction and span.

axw:
> It appears when consuming a message the sarama otel wrapper is trying to pull out a parent span ctx.
>
> Messages sent using the sarama library will have a span context; preserving that relationship would be in line with what otel seems to be doing. If a producer isn't instrumented, but the consumer is, then we'll start a new transaction. Similarly, if a producer's span has no parent, it will also become a transaction.

According to our [messaging spec](https://github.com/elastic/apm/blob/master/specs/agents/tracing-instrumentation-messaging.md):
 - sending messages is always represented as a span within a transaction, e.g. a background task or within an HTTP request handler. For OpenTelemetry these spans get created even when not operating within another span, so in this case I think it's fine if we also create a transaction if there's no parent context. No changes needed for PRODUCER
 - passive and active message reception leads to transactions and spans respectively

For the Sarama instrumentation you pointed at, we're only dealing with passive message reception: you register a callback and it's invoked whenever a new message is received. The "transaction" is the handling of a new message.

For [JMS](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/jms-1.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jms/JmsInstrumenters.java) on the other hand, there are two ways of receiving messages: a "MessageListener" (passive like sarama's consumer) and "MessageConsumer" (active).

> My personal confusion at having a parent/child relationship connected by an async message bus notwithstanding, it seems like we should probably adhere to the "if it's a root span -> new tx; else new span" flow we currently have?

Again following the spec (with a slight adjustment to allow root producer spans -> transactions):

- If it's a root OTel span, we should create a transaction.
- If it's a non-root producer, we should create a span.
- If it's a passively received/processed message, we should create a transaction.
- If it's an actively received message (i.e. something like a channel receive within an HTTP handler), we should create a span.

The challenge is how to distinguish between the last two bullet points. I'm not sure that we can do this with the information we have. I think we would need https://github.com/open-telemetry/opentelemetry-specification/issues/366

processor/otel/consumer.go
```diff
@@ -500,12 +503,28 @@ func translateSpan(span pdata.Span, metadata model.Metadata, event *model.Span)
...
+					// TODO: Is this correct? Inferring that an empty string means "send", based on
+					// only `send`, `process`, and `receive` are allowed
+					// https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/messaging.md#operation-names
+					// `messaging.operation` can only have `process` and `receive` values
+					// cf. table in section https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/messaging.md#apache-kafka-example
+					// and various examples
+					// https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/messaging.md#apache-kafka-example
+					stringval = "send"
+				}
+				messageOperation = stringval
```
stuartnelson3, line 521: From reading the otel messaging spec about `message.operation`:

>The distinction between receiving and processing of messages is not always of particular interest or sometimes hidden away in a framework (see the Message consumption section above) and therefore **the attribute can be left out**.

(emphasis mine)

The span name also "should" be set, with the operation in the last position, but I would guess that has a higher likelihood of being respected. I'm wondering if it might be a good fallback to set the operation to this if `messageOperation=""` when setting `event.Action` below. Thoughts @axw ?

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/messaging.md#span-name

axw: I don't have a strong opinion. I would be inclined to stay away from parsing, and:
 - use `message.operation` if it's set
 - use "send" if the span kind is PRODUCER

If you think it's important to also parse, I'm cool with that too.

stuartnelson3: Sounds reasonable to me. I'd prefer not parsing as well.

processor/otel/consumer.go
```diff
@@ -517,12 +520,21 @@ func translateSpan(span pdata.Span, metadata model.Metadata, event *model.Span)
...
-			// messaging
-			//
-			// TODO(axw) translate OpenTelemtry messaging conventions.
-			case "message_bus.destination":
+			// messaging.*
+			case "message_bus.destination", conventions.AttributeMessagingDestination:
 				message.QueueName = stringval
 				isMessagingSpan = true
+			case conventions.AttributeMessagingOperation:
+				if stringval == "" && span.Kind() == pdata.SpanKindPRODUCER {
```
axw, line 528: Won't the attribute just be omitted altogether, rather than being present with an empty string?

stuartnelson3: ah, yeah. will do the check later in the code.

processor/otel/consumer.go
```diff
@@ -152,7 +152,7 @@ func (c *Consumer) convertSpan(
 	var span *model.Span
 
 	name := otelSpan.Name()
-	if root || otelSpan.Kind() == pdata.SpanKindSERVER {
+	if root || otelSpan.Kind() == pdata.SpanKindSERVER || otelSpan.Kind() == pdata.SpanKindCONSUMER {
```
axw, line 155: probably worth adding a comment here, and an issue for handling active/passive consumption differently

stuartnelson3: Added a [comment](https://github.com/elastic/apm-server/pull/5334/commits/98d4f8503979831fce8709bb926050018df0b5f5) and an issue https://github.com/elastic/apm-server/issues/5373

processor/otel/consumer_test.go
```diff
@@ -518,6 +518,48 @@ func TestRPCSpan(t *testing.T) {
...
+	assert.Empty(t, tx.Labels)
+	assert.Equal(t, &model.Message{
+		QueueName: "myQueue",
+	}, tx.Message)
+}
+
+func TestMessagingSpan(t *testing.T) {
+	span := transformSpanWithAttributes(t, map[string]pdata.AttributeValue{
+		"messaging.system":      pdata.NewAttributeValueString("kafka"),
+		"messaging.operation":   pdata.NewAttributeValueString(""),
```
axw, line 541:
```suggestion
```

axw (approved):

axw:
Verified with 7.14.0-BC2, using the opentelemetry-go-contrib sarama example:

Produce span type/subtype/action and properties:
image

Receive transaction type:
image

Receive transaction properties:
image

Consume span type/action properties:
image

The consume span is missing details, but that's because it's a custom span created by the example app without those details: https://github.com/open-telemetry/opentelemetry-go-contrib/blob/262f81f753760cd1333d1925918300be3f01e77e/instrumentation/github.com/Shopify/sarama/otelsarama/example/consumer/consumer.go#L87-L90

PR #5333, opened 2021-05-25
chore: use gvm to install Go

mdelapenya:
## What does this PR do?
It replaces gimme with gvm, as it's the tool used by Beats developement.

Besides that, we are leveraging APM CI shared library to manage Go env: GOPATH, GOROOT, etc, reading version of Go from `./.go-version` file (from the root directory of the project)

The end goal is fix an issue found while testing #5331, as the merge commit failed the build in the benchmarks stage because of a GOROOT env var not being found.

## Related issues
- Discovered while building https://github.com/elastic/apm-server/pull/5331 on CI: https://apm-ci.elastic.co/blue/organizations/jenkins/apm-server%2Fapm-server-mbp/detail/6.8/49/pipeline/: the benchmark stage fails with the following error:

>[2021-05-25T12:12:59.904Z] + go get -u github.com/elastic/gobench
[2021-05-25T12:12:59.904Z] go: cannot find GOROOT directory: /var/lib/jenkins/workspace/apm-server_apm-server-mbp_6.8/.gvm/versions/go1.10.8
[2021-05-25T12:12:59.904Z] .linux.amd64
script returned exit code 2

## Follow-ups
Apply same approach to the master branch

PR #5348, opened 2021-05-26
chore: use gvm and withMageEnv

mdelapenya:
## Motivation/summary
This PR tries to standardize the CI pipeline, using the built-in steps that provides a well-known Go env. For that, we are replacing gimme with gvm, and using the `withMageEnv` and `withGoEnv` when needed, always within a `dir` step, to make sure they properly ready the `.go-version` file from project's root directory.

## Checklist

## Related issues

- Forward port of #5333


script/common.bash
```diff
@@ -34,18 +34,31 @@ get_go_version() {
...
 
+# install_gvm
+# Install gvm to /usr/local/bin.
+# To read more about installing gvm in other platforms: https://github.com/andrewkroh/gvm#installation
+install_gvm() {
+  # Install gvm
+  if [ ! -f "/usr/local/bin/gvm" ]; then
+    curl -sL -o ~/bin/gvm https://github.com/andrewkroh/gvm/releases/download/v0.3.0/gvm-linux-amd64
+    chmod +x /usr/local/bin/gvm
+  fi
```
axw, line 45: `script/common.bash` gets run, indirectly, as part of our Makefile: https://github.com/elastic/apm-server/blob/73e63c04a5feaba3cfcbe60cd6173fdc517ea593/Makefile#L9

Thus, this is going to break devs: Linux because this needs root, and MacOS because this assumes Linux. Can we install it into ~/.gvm? I would also be much more comfortable if this checked a sha256sum.

axw (changes requested):

axw:
@mdelapenya are you planning on updating this any time soon? If not, let's close it and reopen later please.

mdelapenya:
Yes please, let's close it and continue with it later. I've had this out of my radar for time (sorry about that), but this week I'm on call. So please let's close it. I'll create a ticket to standardise build tools in our repo. Will link this ticket from there

PR #5350, opened 2021-05-26
sampling: persist subscriber position

### LINKED ISSUE #4437: sampling/pubsub: persist subscriber position across restarts
Currently, restarting the server will mean (re)processing all of the documents in the sampled traces index. We should look at recording information about the last observed document in the sampled traces index so we can resume subscription at that position after a server restart.

axw:
## Motivation/summary

Currently, the tail-sampling sampled trace ID subscriber will resume searching from the beginning of each index when the server is restarted. This could be quite expensive in heavy environments. To address this issue, this PR makes the subscriber resumable across restarts.

We introduce `sampling/pubsub.Position`, which holds information about the subscriber's "position" in the sampled traces data stream. Internally this holds a map of indices to observed `_seq_no` values. `pubsub.Pubsub.SubscribeSampledTraceIDs` now accepts a channel to which it will send the subscriber's position when it changes.

The `sampling.Processor` will now persist the subsciber's position to disk in the tail-sampling local storage directory, in a file called "subscriber_position.json". When the processor starts it will read this file and initialise the subscriber with the position so it resumes searches from there.

## Checklist

## How to test these changes

1. Run apm-server with tail-sampling configured
2. Send transactions
3. Check that after ~30 seconds, there's a `subscriber_position.json` file under `<data-dir>/tail_sampling`
4. Restart apm-server
5. Send some more transactions
6. Check that transactions are sampled according to tail-sampling config

For bonus points:

7. Do some long-run testing where there are a lot of tail-sampling decision documents
8. Restart apm-server and ensure it does not churn through all the old documents

## Related issues

Closes #4437

### DIFF
--- a/x-pack/apm-server/sampling/processor_test.go
+++ b/x-pack/apm-server/sampling/processor_test.go
@@ -732,7 +732,6 @@ func cloneBatch(in *model.Batch) *model.Batch {
 // and file info (including modification time).
 func waitFileModified(tb testing.TB, filename string, after time.Time) ([]byte, os.FileInfo) {
 	// Wait for subscriber_position.json to be written to the storage directory.
-	//subscriberPositionFile := filepath.Join(config.StorageDir, "subscriber_position.json")
 	timeout := time.NewTimer(10 * time.Second)
 	defer timeout.Stop()
 	ticker := time.NewTicker(50 * time.Millisecond)
--- a/x-pack/apm-server/sampling/processor_test.go
+++ b/x-pack/apm-server/sampling/processor_test.go
@@ -11,6 +11,7 @@ import (
 	"math/rand"
 	"os"
 	"path"
+	"path/filepath"
 	"sort"
 	"strings"
 	"testing"
@@ -18,6 +19,7 @@ import (
 
 	"github.com/dgraph-io/badger/v2"
 	"github.com/gofrs/uuid"
+	"github.com/pkg/errors"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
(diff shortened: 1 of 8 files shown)


x-pack/apm-server/sampling/pubsub/pubsub.go
```diff
@@ -158,13 +181,16 @@ func (p *Pubsub) searchTraceIDs(ctx context.Context, out chan<- string, observed
 				return err
 			}
 			if maxSeqno > observedSeqno {
+				observedSeqnosMu.Lock()
 				observedSeqno = maxSeqno
+				observedSeqnos[index] = observedSeqno
+				changed = true
+				observedSeqnosMu.Unlock()
```
stuartnelson3, line 188: Nice catch on the map multi-write

x-pack/apm-server/sampling/pubsub/pubsub.go
```diff
@@ -141,9 +162,11 @@ func (p *Pubsub) searchTraceIDs(ctx context.Context, out chan<- string, observed
...
 	}
 	if err := p.refreshIndices(ctx, indices); err != nil {
-		return err
+		return false, err
 	}
 
+	var changed bool
+	var observedSeqnosMu sync.Mutex
 	g, ctx := errgroup.WithContext(ctx)
 	for _, index := range indices {
```
stuartnelson3, line 171: Unrelated to this PR, but while I'm looking: do you have an estimate how many indices will typically be queried? Wondering about how many requests this could trigger `apm-server` to send out at once, if that's a concern.

stuartnelson3 (approved):
Looks good. I left a question and a couple comments, but nothing show-stopping.

x-pack/apm-server/sampling/pubsub/pubsub.go
```diff
@@ -87,23 +88,43 @@ func (p *Pubsub) onBulkIndexerItemFailure(ctx context.Context, item elasticsearc
...
+
+	// Copy pos because it may be mutated by p.searchTraceIDs.
+	pos = copyPosition(pos)
 	for {
 		select {
 		case <-ctx.Done():
 			return ctx.Err()
+		case positionsOut <- pos:
+			// Copy pos because it may be mutated by p.searchTraceIDs.
+			pos = copyPosition(pos)
```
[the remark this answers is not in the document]
axw: > we mutate pos in p.searchTraceIDs, and then update the channel listener by sending a copy of the mutated pos

We mutate pos in p.searchTraceIDs, send pos to the channel, and _then_ copy it. This ensures the position sent over the channel won't be mutated.

> Is it avoiding a potential race between sending the position and having it marshaled/written to disk, and updating the value at the next tick?

Yes, the reason is to avoid a data race between the receiver of the `positions` channel (writing to disk) and `p.searchTraceIDs`. We need to make sure the map inside the position isn't modified after it has been sent over the channel.

x-pack/apm-server/sampling/pubsub/pubsub.go
axw: _Probably_ only ever one or two. We haven't set up ILM for this yet, but the idea is to have a fairly aggressive rollover + deletion policy: rollover the index after (say) an hour, and then delete after another hour. I say probably because we could also have size based rollover, i.e. rollover when the index grows to 30GB. I don't think we'll need that, but we'll see.

Whatever rollover/delete time duration we choose defines the maximum duration for a trace. If an event for a trace comes in a day later, then the sampled trace ID will have been long forgotten, so the event will end up getting dropped.

x-pack/apm-server/sampling/processor_test.go
```diff
@@ -700,3 +726,36 @@ func cloneBatch(in *model.Batch) *model.Batch {
 		Spans:        in.Spans[:],
 	}
 }
+
+// waitFileModified waits up to 10 seconds for filename to exist and for its
+// modification time to be greater than "after", and returns the file content
+// and file info (including modification time).
+func waitFileModified(tb testing.TB, filename string, after time.Time) ([]byte, os.FileInfo) {
+	// Wait for subscriber_position.json to be written to the storage directory.
+	//subscriberPositionFile := filepath.Join(config.StorageDir, "subscriber_position.json")
```
axw, line 735:
```suggestion
```

x-pack/apm-server/sampling/pubsub/pubsub.go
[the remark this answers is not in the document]
axw: Gonna merge, let me know if you have any concerns later.

stuartnelson3:
Confirmed with BC2: After sending transactions, `subscriber_position.json` existed. Restarted the server, and sent 10 transactions with a tail-based sampling rate of 0.2, confirmed that only 2 new transactions were indexed.

PR #5351, opened 2021-05-26
Add additional logging when starting with data_streams.enabled

Mpdreamz:
## Motivation/summary

Setting `data_streams.enabled` on its own is not sufficient. The agent needs to go in managed mode and the `apm` integration has to be installed. 

## How to test these changes

> ./apm-server -e -E logging.json=false -E logging.ecs=false

Should now print an ERROR explaining the misconfiguration but will still allow the server to continue starting up as this is used in some testing scenarios


beater/beater.go
```diff
@@ -116,6 +117,9 @@ func NewCreator(args CreatorParams) beat.Creator {
 			if b.Manager != nil && b.Manager.Enabled() {
 				return nil, errors.New("data streams must be enabled when the server is managed")
 			}
+		} else if bt.config.DataStreams.Enabled && !fleetmode.Enabled() {
+			// not supported only available for development purposes
+			bt.logger.Errorf("Started apm-server with data streams enabled but no active fleet management mode was specified")
 		}
```
Mpdreamz, line 123: I wonder if we should simply set 

```go
bt.config.DataStreams.Enabled = fleetmode.Enabled()
```

And disallows it to be set through the external configuration and document the following for testing purposes.

```
 -E management.enabled=true -E management.mode=x-pack-fleet
```

axw: That might be reasonable - I'm not sure if enabling fleet management mode when not actually running under Elastic Agent / Fleet would pose any problems; if we can then think that would be fine. The main thing is that I don't want to have to run up all the infrastructure just to test data streams.

axw (approved):

beater/beater.go
axw: heh, this change actually trips up a system test which checks that no warnings or errors are logged :)
do you have time to look into whether we can remove this config and just rely on the fleet management config? if not I can look into it some time later

Mpdreamz: >  I'm not sure if enabling fleet management mode when not actually running under Elastic Agent / Fleet would pose any problems; if we can then think that would be fine. 

The elastic-agent-client that the fleet manager uses requires its config in protobuf bytes available on stdIn. So that's a dead end.

> do you have time to look into whether we can remove this config and just rely on the fleet management config? if not I can look into it some time later

Will amend later today!

systemtest/datastreams_test.go
```diff
@@ -25,13 +25,14 @@ import (
 	"testing"
 	"time"
 
+	"github.com/elastic/apm-server/systemtest/estest"
```
axw, line 28: can you move this back please? all `github.com/elastic/...` import should be grouped together
did your IDE do this?

systemtest/datastreams_test.go
```diff
@@ -110,6 +111,9 @@ func TestDataStreamsEnabled(t *testing.T) {
 			// There should be no warnings or errors logged.
```
axw, line 111: comment needs an update :)

axw (approved):
LGTM. Maybe later for unit testing purposes we can have an in-memory elastic-agent shim, and get rid of the extra config.

systemtest/datastreams_test.go
Mpdreamz: It must have done, `make fmt` does not pick up on this ordering. Will see if GoLand can be configured to order imports like we dictate.

axw:
Verified with 7.14.0-BC2, by starting apm-server with `-E apm-server.data_streams.enabled=true` but not in Fleet mode:

> `localtesting_7.14.0_apm-server | {"log.level":"error","@timestamp":"2021-07-09T05:24:36.309Z","log.logger":"beater","log.origin":{"file.name":"beater/beater.go","file.line":128},"message":"Started apm-server with data streams enabled but no active fleet management mode was specified","service.name":"apm-server","event.dataset":"apm-server","ecs.version":"1.6.0"}`

Issue #5034, opened 2021-03-30
Set service.name and event.dataset in logs


felixbarny:
Currently, the APM Server logs don't contain a `service.name` field which makes it hard to find the server logs in the logs UI as users can't filter for `service.name: "apm-server"`.
The `event.dataset` field should also be set to `apm-server.log`. See the [ecs logging spec](https://github.com/elastic/ecs-logging/blob/master/spec/spec.json) for more details. 

This issue depends on https://github.com/elastic/ecs-logging-go-zap/issues/24

axw:
I'll need to double check this later, but I'm pretty sure that Elastic Agent (by way of Filebeat) adds `event.dataset` for APM Server when running under Elastic Agent/Fleet. Seeing as that's the path forward, do you think it's sufficient to rely on Elastic Agent setting that field?

We can/should probably set `service.name` for all Beats as well as APM Server.

felixbarny:
I think the application emitting the logs has the best knowledge on what the `service.name` and the `event.dataset` should be. The idea with ECS loggers is that there's no need to configure Filebeat to add specific fields (besides add metadata). The content of the log files is more or less exactly what gets stored in ES. This lets users have the same Filebeat configuration for all of their apps.

> We can/should probably set service.name for all Beats as well as APM Server.

👍 
I think the best way would be to configure that in the logging setup of each beat.
The `event.dataset` would then be automatically set to `${service.name}.log` by ecs-logging-go-zap, according to the [ECS logging spec](https://github.com/elastic/ecs-logging/blob/master/spec/spec.json).

axw:
> I think the application emitting the logs has the best knowledge on what the service.name and the event.dataset should be.

Yes, fair enough. We'll have to see what happens when we do that -- whether Filebeat overrides the specified value, or if it doesn't, whether doing so has any negative effects on Fleet's agent logs UI.

jalvz:
As I just learnt, service name = beat name. Shouldn't this issue be on Beats repo then?

axw:
I think the apm-server team will be implementing it either way so ultimately it shouldn't matter. Feel free to transfer it though.

PR #5355, opened 2021-05-27
Add service name to logs

jalvz:
Implements part of https://github.com/elastic/apm-server/issues/5034

## Checklist

## How to test these changes

Send some events to apm-server, check that `service.name` is in the request logs

### DIFF
--- a/agentcfg/cache.go
+++ b/agentcfg/cache.go
@@ -18,6 +18,7 @@
 package agentcfg
 
 import (
+	"fmt"
 	"time"
 
 	gocache "github.com/patrickmn/go-cache"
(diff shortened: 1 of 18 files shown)


axw (changes requested):
Explicitly adding `service.name` to every log message is going to be difficult and error-prone to maintain. Moreover, we should be adding the fields to log messages coming from libbeat too.

The service name is the beat name, and that's [already known to logp](https://github.com/elastic/beats/blob/e5b976a3bee54553b96b82be8bbaf1e6835c280d/libbeat/logp/config.go#L29), so can we add the field centrally in logp? It should be possible to do this with a `zap.Fields` option, or with `zap.Logger.WithFields`.

Given that we'll also need to set `event.dataset` as part of https://github.com/elastic/apm-server/issues/5034, maybe we should add an option in ecszap for setting both `service.name` and `event.dataset`?

jalvz:
> The service name is the beat name

I understood it was the apm service name, not the beat name. The this doesn't make sense, closing.

axw:
Tested changed made in https://github.com/elastic/beats/pull/25998, using 7.14 BC1. Example apm-server log record:

```
localtesting_7.14.0_apm-server | {"log.level":"info","@timestamp":"2021-07-06T07:56:51.344Z","log.logger":"request","log.origin":{"file.name":"middleware/log_middleware.go","file.line":63},"message":"request ok","service.name":"apm-server","event.dataset":"apm-server","url.original":"/","http.request.method":"GET","user_agent.original":"curl/7.29.0","source.address":"127.0.0.1","http.request.body.bytes":0,"http.request.id":"bb9ec214-10f0-4c7a-95d9-1c48bcf50928","event.duration":97785,"http.response.status_code":200,"ecs.version":"1.6.0"}
```

Notably, `service.name` and `event.dataset` fields are there.

PR #5357, opened 2021-05-28
Fix gvm installation

axw:
Fix a typo introduced in https://github.com/elastic/apm-server/pull/5333. I took the liberty of changing it to install to `~/.gvm/gvm` rather than `/usr/local/bin/gvm`.

### DIFF
--- a/_beats/dev-tools/common.bash
+++ b/_beats/dev-tools/common.bash
@@ -47,7 +47,8 @@ install_gvm() {
     chmod +x "$GVM"
   fi
 
-  debug "Gvm version $(${GVM} --version)"
+  export PATH="${GVMDIR}:${PATH}"
+  debug "Gvm version $(gvm --version)"
 }
 
 # setup_go_root "version"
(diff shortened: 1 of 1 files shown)


simitt (approved):

_beats/dev-tools/common.bash
```diff
@@ -39,12 +39,14 @@ get_go_version() {
 # To read more about installing gvm in other platforms: https://github.com/andrewkroh/gvm#installation
 install_gvm() {
   # Install gvm
-  if [ ! -f "/usr/local/bin/gvm" ]; then
-    curl -sL -o ~/bin/gvm https://github.com/andrewkroh/gvm/releases/download/v0.3.0/gvm-linux-amd64
-    chmod +x /usr/local/bin/gvm
+  GVMDIR="$HOME/.gvm"
+  GVM="$GVMDIR/gvm"
```
mdelapenya, line 43: If the pipeline makes sure that GVMDIR is in the PATH, then LGTM :)

axw: Oh right, it doesn't. I guess we should just add it to $PATH here?

mdelapenya (approved):

PR #5364, opened 2021-05-28
update agent_config examples

stuartnelson3:
## Motivation/summary

The values are being rendered as strings in kibana
and internally represented as strings anyway, so
we should be consistent

mentioned [here](https://github.com/elastic/kibana/pull/100744#issuecomment-850037068)

## Checklist

## How to test these changes

Non-functional change.

### DIFF
--- a/_meta/beat.yml
+++ b/_meta/beat.yml
@@ -290,11 +290,11 @@ apm-server:
   #     capture_body: true
   #     log_level: info
   #     recording: true
-  #     transaction_sample_rate: 0.1
+  #     transaction_sample_rate: '0.1'
   # - service.name: frontend
   #   agent.name: rum-js
   #   config:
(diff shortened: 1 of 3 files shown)


axw (approved):
Maybe explicitly quote them all? IIANM, YAML will treat each of `off`, `false`, and `true` as booleans. That just leaves `info` as a string, so it's probably clearer just to quote them all at that point.

stuartnelson3:
After our discussion in stand-up, we decided to remove the `agent_config` section entirely, correct? @simitt @axw

axw:
Yes, let's just remove the example from apm-server.yml for now. Maybe it can come back later.

stuartnelson3:
Closed in favor of https://github.com/elastic/apm-server/pull/5366

PR #5380, opened 2021-05-31
fix direct agent configuration

stuartnelson3:
## Motivation/summary

Manual testing with https://github.com/elastic/kibana/pull/100744 revealed
errors in the current apm-server logic. This PR introduces changes so that
apm-server correctly switches to direct agent configuration when running under
managed mode.

depends on https://github.com/elastic/kibana/pull/100744

## Checklist

## How to test these changes

1. Start elasticsearch using apm-integration-testing
2. Start a version of kibana including [this PR](https://github.com/elastic/kibana/pull/100744)
3. Start elastic-agent (not managed by fleet) with a version of `apm-server`
   including this PR
4. From the kibana UI:
  - Add apm-server as an integration to a fleet policy
  - Create an agent config in APM / Settings / Agent Config
  - Verify that the agent_config exists in the fleet policy and is nested under
    `apm-server`
5. Update the running `elastic-agent.yml` with the generated policy from the
   previous step
6. Verify that `cURL`ing the config endpoint returns your agent config
  - e.g. `curl 'http://localhost:8200/config/v1/agents?service.name=all'`


beater/beater.go
```diff
@@ -308,6 +308,11 @@ func newServerRunner(ctx context.Context, args serverRunnerParams) (*serverRunne
 	if cfg.DataStreams.Enabled && args.KibanaConfig != nil {
 		cfg.Kibana.ClientConfig = *args.KibanaConfig
 	}
+	if args.Beat.Manager != nil && args.Beat.Manager.Enabled() {
+		// If we're running in managed mode we do not want to
+		// communicate with kibana.
+		cfg.Kibana.Enabled = false
+	}
```
stuartnelson3, line 315: For some reason I can't find the code again, but our default unmarshal behavior for bool type is to set it to true if it's not found. Even if it is set, apm-server doesn't communicate with kibana if running in managed mode.

stuartnelson3:
Note: The current nesting in https://github.com/elastic/kibana/pull/100744 has `agent_configs` on the same level as `apm-server`. In order for the config to work correctly, `agent_configs` needs to be nested underneath it, ie. `apm-server.agent_configs`.

stuartnelson3:
Note: If we're running in managed mode (`Kibana.Enabled == false`), but no agent configs exist, then this code branch will be executed and attempts to query the agent config will fail: https://github.com/elastic/apm-server/blob/16618f2a00400dcde109c00037be79e24f3938c5/beater/api/mux.go#L200-L206

I could thread some sort of "managed" bool through to help make this decision, but I'm not sure if there might be some better way to handle it. Any suggestions?

beater/beater.go
axw: Are you thinking of the [`libbeat/common.Config.Enabled()`](https://github.com/elastic/beats/blob/27e76c56771103326a62a9e46a58f345ed7eda21/libbeat/common/config.go#L280-L294) method, which returns true when the config block is defined but the "enabled" field is not explicitly specified?

I'm not sure if that's relevant here, since we only set `cfg.Kibana.ClientConfig` above.

axw: Ah, I see the problem. I just created a default apm policy and I get this:

image

The presence of that `kibana: ...` means it is enabled. This config was introduced as a temporary measure to support central config via Kibana; we should remove it in 7.14:

https://github.com/elastic/apm-server/blob/16618f2a00400dcde109c00037be79e24f3938c5/apmpackage/apm/0.2.0/manifest.yml#L62-L67

https://github.com/elastic/apm-server/blob/16618f2a00400dcde109c00037be79e24f3938c5/apmpackage/apm/0.2.0/agent/input/template.yml.hbs#L7-L8

axw: can we remove this change now?

stuartnelson3: removed

axw (approved):

simitt (approved):

stuartnelson3:
jenkins run the tests

axw:
Verified with 7.14.0-BC2. I ran apm-integration-testing with `--apm-server-managed`, and then ran this Go program:

```go
package main                                                                  
                                                                              
import (                                                                      
        "context"                                                             
        "fmt"                                                                 
                                                                              
        "go.elastic.co/apm/apmconfig"                                         
        "go.elastic.co/apm/transport"                                         
)                                                                             
                                                                              
func main() {                                                                 
        var args apmconfig.WatchParams                                        
        args.Service.Name = "main"                                            
        transport, _ := transport.NewHTTPTransport()                          
        changes := transport.WatchConfig(context.Background(), args)          
        for change := range changes {                                         
                fmt.Println("changed:", change)                               
        }                                                                     
}                                                                             
```

(This is using internals of the Go Agent.)

Then setting/updating a central config rule for service "main", I observed the changes in the config-watching program.

There is something wrong with how apm-server is handling config changes generally. On some changes, I observed apm-server would become unresponsive. Then looking inside the container I found the process had exited, but there was no sign of any errors in the logs. I'll continue digging into that separately.

Issue #3876, opened 2020-06-11
model: align HTTP with ECS


axw:
ECS defines the [HTTP](https://www.elastic.co/guide/en/ecs/current/ecs-http.html) fieldset. 

APM stores transaction HTTP fields in compliance with ECS, but span HTTP fields are nested inside the span namespace. We should move these fields to the top level http namespace. Some of the fields are currently indexed; these fields would need to be duplicated to avoid breaking changes.

APM stores additional HTTP fields for transactions and spans that are not yet defined in ECS:
 - `http.response.decoded_body_size`
 - `http.response.encoded_body_size`
 - `http.response.headers`
 - `http.response.transfer_size`
 - `http.request.cookies`

We should compare these to the ECS-defined fields (`http.request.bytes`, etc.) and propose changes to ECS or the APM Server code as necessary.

APM stores the HTTP request body as `http.request.body.original`, whereas ECS defines it as `http.request.body.content`. When upgrading to 7.0 we agreed on keeping original; not sure if we need to revisit.

cyrille-leclerc:
FYI Preparation Notes for the RFC to standardise in ECS: [OpenTelemetry > Semantic conventions for HTTP spans](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/http.md)

PR #5396, opened 2021-06-01
Lift HTTP span fields 1 level

jalvz:
## Motivation/summary

Sets HTTP span fields as top fields for ECS compatibility.  All the fields except response headers are copied for backwards compatibility. 

## Checklist

## Related issues

Implements part of https://github.com/elastic/apm-server/issues/3876


axw (changes requested):

model/span.go
```diff
@@ -254,6 +255,7 @@ func (e *Span) fields(ctx context.Context, cfg *transform.Config) common.MapStr
 
 	fields.maybeSetMapStr("db", e.DB.fields())
 	fields.maybeSetMapStr("http", e.HTTP.fields())
+	fields.delete("http.response.headers")
```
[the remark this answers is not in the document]
jalvz: > I think this is backwards? http.response.headers doesn't exist in ECS, so we should only set it as span.http.response.headers for now

OK, I thought we wanted to move everything up because after 8.0 is going to be weird that some http fields are top level, and others are inside spans. Like, if they end up in ECS we would have them already at the top level, otherwise they would not hurt.

beater/test_approved_es_documents/TestPublishIntegrationEvents.approved.json
```diff
@@ -221,6 +221,15 @@
                     "platform": "Linux"
                 }
             },
+            "http": {
+                "method": "GET",
```
axw, line 225: this is not an ECS field: https://www.elastic.co/guide/en/ecs/current/ecs-http.html

beater/test_approved_es_documents/TestPublishIntegrationEvents.approved.json
```diff
@@ -221,6 +221,15 @@
...
+                "response": {
+                    "status_code": 200
+                },
+                "url": {
+                    "original": "http://localhost:8000"
```
axw, line 230: nor is this

axw (changes requested):
Some of the field names in ECS are different. Please check through the ECS docs to make sure all fields we might add under "http" are ECS-compliant.

I'd suggest at this point that for the fields that have the same _meaning_ but different names (e.g. `http.request.method`), just output the ECS field name. For backwards compatibility, define an alias field in fields.yml, from the old APM-specific name to the new ECS name.

jalvz:
Sorry, I forgot about this. I don't understand this part:

> For backwards compatibility, define an alias field in fields.yml, from the old APM-specific name to the new ECS name

If we still keep the old field unmodified, what is the problem with backwards compatibility?

axw:
> If we still keep the old field unmodified, what is the problem with backwards compatibility?

That comment only made sense in conjunction with the rest: only output the new (ECS) fields (and use an alias for the old ones). Producing both fields maintains backwards compatibility too, it just means we'll be using more storage space.

axw (approved):

axw:
Verified with 7.14.0-BC2, span HTTP fields are elevated to `http.*` and `url.*`, with existing `span.http.*` fields remaining:

image

PR #5401, opened 2021-06-02
Make process title dynamic

jalvz:
This test always fails for me locally (both when running from the IDE and command line), make the conflicting field dynamic.


axw (changes requested):
Instead of ignoring the value in the tests, can you please update apmservertest to set a hard-coded process title?

https://github.com/elastic/apm-server/blob/31371a13f37fd92c17e43abd22646e409cd7f4c4/systemtest/apmservertest/filter.go#L103-L117

That way we can continue to check that the server receives and records the value as expected.

axw (approved):
thanks!

Issue #5372, opened 2021-05-31
Hardcode sizing dependent settings in APM Server


simitt:
Read ENV _CLOUD_APM_CAPACITY_ indicating the cloud deployment size and hardcode size relevant config option default values. 

setting\capacity | 512 | 1024 | 2048 | 4096 | 8192
-- | -- | -- | -- | -- | --
&nbsp; | &nbsp; | &nbsp; | &nbsp; | &nbsp; | &nbsp;
output.elasticsearch.worker | 5 | 7 | 10 | 14 | 20
output.elasticsearch.bulk_max_size | 267 | 381 | 533 | 762 | 1067
queue.mem.events | 2000 | 4000 | 8000 | 16000 | 32000
queue.mem.flush.min_events | 267 | 381 | 533 | 762 | 1067

PR #5402, opened 2021-06-02
Hardcode troughput settings for cloud environment

jalvz:
## How to test these changes

Start apm-server with eg `CLOUD_APM_CAPACITY=512`, and check that the corresponding have the correct values.

## Related issues

Closes https://github.com/elastic/apm-server/issues/5372

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -19,6 +19,7 @@ package cmd
 
 import (
 	"fmt"
+	"os"
 
 	"github.com/spf13/pflag"
 
@@ -37,7 +38,16 @@ import (
 const (
 	beatName        = "apm-server"
 	apmIndexPattern = "apm"
+	cloudEnv        = "CLOUD_APM_CAPACITY"
 )
+
+var cloudMatrix = map[string][]int{
+	"512":  {5, 267, 2000, 267},
+	"1024": {7, 381, 4000, 381},
+	"2048": {10, 533, 8000, 533},
+	"4096": {14, 762, 16000, 762},
+	"8192": {20, 1067, 32000, 1067},
+}
 
 var libbeatConfigOverrides = []cfgfile.ConditionalOverride{{
 	Check: func(_ *common.Config) bool {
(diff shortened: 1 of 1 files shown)


simitt (commented):
Code looks good; would be great to have a test ensuring that the ENV is picked up and that values can be overriden via `apm-server.yml`.

jalvz:
Any idea how to do that? We don't have any visibility on those settings, or I don't know how to read them back. I don't know how to test the behavior neither, so I can't set up and integration test to send some events and _expect_ something.

axw (commented):
> Any idea how to do that? We don't have any visibility on those settings, or I don't know how to read them back. I don't know how to test the behavior neither, so I can't set up and integration test to send some events and expect something.

You could write a system test that checks everything except flush.min_events by looking at stack monitoring docs: https://github.com/elastic/beats/pull/24700. There's already helpers in systemtest for waiting for a stack monitoring doc.

Alternatively, I think it should be possible to unit this by passing a mock `beat.Creator` into `cmd.NewRootCommand`: https://github.com/elastic/apm-server/blob/f32bb45bae196b894715323f025059752e9f8e8f/cmd/root.go#L75

simitt:
How about adding an info log line containing these values when the server is started? This would be helpful information anyways. 
@axw added reporting for these values in libbeat https://github.com/elastic/beats/pull/24700, so you could also create a system test where the monitored values are read back.

jalvz:
I found a way to unit test it. Don't you think is overkill to write a system test just to ensure that an environment variable is picked up?

axw:
> I found a way to unit test it. Don't you think is overkill to write a system test just to ensure that an environment variable is picked up?

Generally speaking, things that can be configured by users should have functional tests. Since we're only translating the environment variable to other config, which is tested elsewhere, I think a unit test in this case is fine. Having no tests is not. I mentioned the system test approach just in case writing a unit test wasn't feasible in the short term without a lot of refactoring.

simitt (approved):
Apart from the minor comment, this looks good. Thanks for adding it!

stuartnelson3:
confirmed with BC2
```
# env CLOUD_APM_CAPACITY=4096 ./apm-server export config > 4096.cfg
# env CLOUD_APM_CAPACITY=1024 ./apm-server export config > 1024.cfg
```

`# diff 4096.cfg 1024.cfg`
```diff
21c21
<     bulk_max_size: 762
---
>     bulk_max_size: 381
27c27
<     worker: 14
---
>     worker: 7
35c35
<     events: 16000
---
>     events: 4000
37c37
<       min_events: 762
---
>       min_events: 381
```

Issue #5002, opened 2021-03-25
Introduce configuration for specifying service-specific source maps


axw:
As part of the move to Fleet we will need to move away from fetching source maps directly from (and indexing source maps into) Elasticsearch, as the privileges APM Server is given does not cover this. Instead, source maps should be uploaded to Kibana, stored as a Fleet Artifact, with references to Fleet Artifact download URLs pushed down to APM Server via the server's policy.

In order to make the above possible, we will add configuration to APM Server to specify zero or more source maps, including:
 - service name
 - service version
 - bundle filepath
 - source map URL (typically a Fleet Server artifact download URL)

For example:

```yaml
apm-server:
  rum:
    source_mapping:
      source_maps:
      - service.name: opbeans-rum
        service.version: 1.2.3
        bundle.filepath: /test/e2e/general-usecase/bundle.js.map
        sourcemap.url: http://somewhere.com/bundle.js.map
```

The server will fetch and cache these like today: fetch on first access, cache in memory with an expiration time. We might make improvements to caching later, but that can be done separately. If the list of source maps is dynamically updated, then the cache will be invalidated for (service name, service version, bundle filepath) tuples that are added, changed, or removed.

Configuring a list of source maps will disable fetching source maps from Elasticsearch. It will be an error to specify Elasticsearch configuration under `apm-server.rum.source_mapping` if the `source_maps` property is non-null.

Fleet Server requires auth for fetching artifacts. If a source map URL points to Fleet Server, then APM Server must pass its Fleet API access key in download requests.

stuartnelson3:
Excepting cache management, is it sufficient to make a request to `sourcemap.url` with the auth header to get the sourcemap for a given key lookup? I'm not finding documentation, but the code seems to indicate that is enough:

https://github.com/elastic/fleet-server/blob/d72fcdd6d9c0707fef43918c2398b741cef77130/cmd/fleet/router.go#L72-L74
https://github.com/elastic/fleet-server/blob/d72fcdd6d9c0707fef43918c2398b741cef77130/cmd/fleet/handleArtifacts.go#L84

axw:
@stuartnelson3 yes that should be sufficient - it should be enough to perform a GET request with the configured URL, with the API Key in the Authorization header.

stuartnelson3:
>If the list of source maps is dynamically updated, then the cache will be invalidated for (service name, service version, bundle filepath) tuples that are added, changed, or removed.

[...]

PR #5410, opened 2021-06-03
Service specific source maps

stuartnelson3:
## Motivation/summary

support configuring source maps from kibana and fetching them from fleet server.

closes https://github.com/elastic/apm-server/issues/5002

## Checklist

## How to test these changes

Run elasticsearch + and elastic-agent in fleet mode.

Start kibana with the code from https://github.com/elastic/kibana/issues/95393 has been completed, and apm-server with the fleet-server sourcemap code.

- Upload a sourcemap via kibana
- Confirm an updated config with the new `source_maps` block has been pushed to apm-server
- Confirm ingesting rum traces for the uploaded sourcemap get mapped appropriately

## Related issues

depends on https://github.com/elastic/kibana/issues/95393


beater/beater.go
```diff
@@ -618,12 +620,31 @@ func newTransformConfig(beatInfo beat.Info, cfg *config.Config) (*transform.Conf
...
+		b := sourcemap.NewESStore(esClient, index, logger)
+
+		return sourcemap.NewStore(b, logger, cfg.Cache.Expiration)
+	}
+
+	// make a copy of the default http client, configure a Transport to be
+	// the same as the esClient, instrument it, and set it on the http
+	// client.
+	c := *http.DefaultClient
+	tr, err := elasticsearch.NewHTTPTransport(cfg.ESConfig)
```
axw, line 639: Communicating with fleet-server may require a different TLS config (e.g. different CA certificate), so I don't think using the elasticsearch transport is the right thing to do. I don't know where we get that config from though, we'll need to investigate and/or ask for advice.

axw (commented):
Approach looks good. The biggest unknown for me is where we get TLS config from for communicating with fleet-server.

beater/config/sourcemapping.go
```diff
@@ -0,0 +1,32 @@
...
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package config
+
+type SourceMapConfig struct {
+	Service   Service   `config:"service"`
```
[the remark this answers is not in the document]
stuartnelson3: works a treat

beater/beater.go
stuartnelson3: Asked folks working on fleet, waiting for a confirmation. Initial response is it might be handed to the child proc during the gRPC handshake?

sourcemap/store.go
```diff
@@ -45,38 +42,75 @@ var (
...
+	if ok {
+		// found an inflight request, wait for it to complete.
+		s.Unlock()
+		<-wait
+		// Try to read the value again
+		// TODO: All waiting Fetch()'s will attempt to re-fetch once
+		// the coordinated backend request has finished. At some point
+		// we need to start failing incoming requests as more start
+		// piling up, if eg. the backend is down.
+		return s.Fetch(ctx, name, version, path)
```
stuartnelson3, line 92: Wanted to call out this TODO. I have no idea what a reasonable value would be for limiting connections, if that's something we want to handle now.

_meta/beat.yml
```diff
@@ -244,9 +244,12 @@ apm-server:
       # Sourcemapping is enabled by default.
       #enabled: true
 
-      # Source maps are always fetched from Elasticsearch, by default using the output.elasticsearch configuration.
+      # Source maps may be fetched from Elasticsearch by using the
+      # output.elasticsearch configuration.
       # A different instance must be configured when using any other output.
       # This setting only affects sourcemap reads - the output determines where sourcemaps are written.
+      # Note: Configuring elasticsearch is not allowed if apm-server is being
```
simitt, line 251: nit: `%s/allowed/supported`

beater/beater.go
```diff
@@ -618,12 +620,26 @@ func newTransformConfig(beatInfo beat.Info, cfg *config.Config) (*transform.Conf
 }
 
 func newSourcemapStore(beatInfo beat.Info, cfg config.SourceMapping) (*sourcemap.Store, error) {
-	esClient, err := elasticsearch.NewClient(cfg.ESConfig)
+	if len(cfg.SourceMapConfigs) == 0 {
```
simitt, line 623: Doesn't that lead to a similar problem we faced with the agent config settings? If the sourcemap gets uploaded after the APM Server was started, APM Server would still try to fetch from the ES store instead of the fleet store.

sourcemap/store.go
```diff
@@ -45,38 +42,75 @@ var (
...
+	} else {
```
simitt, line 93: no need for using `else`, since the `if` case returns

sourcemap/store.go
simitt: How about adding a timeout for waiting on the inflight backend request to return? If the timeout is exceeded, return an error for this request rather than calling `s.Fetch(..)` again. 
Restricting the number of allowed requests would be tricky IMO - it much depends on the setup (for example the rate limiter values).

simitt (commented):
Left a first high-level review.

beater/beater.go
stuartnelson3: I'm not sure. If a sourcemap is uploaded, then the config will be regenerated and pushed to apm-server, right? That will trigger a reload of the server and fresh evaluation of this logic, which should now have the sourcemap config.

sourcemap/store.go
```diff
@@ -44,39 +41,103 @@ var (
...
 	return &Store{
-		cache:   gocache.New(expiration, cleanupInterval(expiration)),
-		esStore: &esStore{client: client, index: index, logger: logger},
+		cache:   gocache.New(cacheExpiration, cleanupInterval(cacheExpiration)),
+		backend: b,
 		logger:  logger,
+		// TODO: Number of requests waiting for a sourcemap to be
+		// fetched. Any suggestion as to where we should set this is
+		// appreciated.
+		waiters:     make(chan struct{}, maxWaiters),
```
axw, line 75: I think protecting against request storms with the inflight map is a good idea, but I'm not sure about this waiters limit. What is it protecting against?

sourcemap/store.go
```diff
@@ -44,39 +41,103 @@ var (
...
+
+		t := time.NewTimer(s.waitTimeout)
+		select {
+		case <-wait:
+			t.Stop()
+		case <-t.C:
+			// Release back to waiters semaphore
+			<-s.waiters
+			return nil, errors.New("sourcemap fetch: timeout")
+		}
```
axw, line 114: we should also have a case for `ctx.Done()`

sourcemap/store.go
```diff
@@ -44,39 +41,103 @@ var (
...
-	cache   *gocache.Cache
-	esStore *esStore
-	logger  *logp.Logger
+	cache       *gocache.Cache
+	backend     backend
+	logger      *logp.Logger
+	waiters     chan struct{}
+	waitTimeout time.Duration
+
+	sync.Mutex
```
axw, line 50:
```suggestion
	mu sync.Mutex
```

so we don't expose the mutex to the world

sourcemap/store.go
stuartnelson3: With the inflight map we'll only send a single request to the fleet server, but if the fleetserver is down or slow, then we'll start piling up requests to apm-server who are waiting for the result of fetch, which could potentially bring down the apm-server. Without having checked any numbers this could definitely fall into the realm of being overly-defensive, so no worries if you want me to drop it.

sourcemap/store.go
stuartnelson3: :100:

sourcemap/store.go
axw: I'd prefer to drop it, or at least remove it from this PR.

Source mapping is (at least for now) initiated by the "publisher" (apm-server/publish), which has a worker per GOMAXPROCS -- one per CPU. Thus there shouldn't ever be a large backlog, unless we were to shuffle source mapping to somewhere else in the process.

Aside from that we already have other safeguards in place: rate-limiting for RUM, and it is possible to set a maximum number of concurrent connections.

sourcemap/store.go
axw: Now that the waiters semaphore is gone, I'm wondering if we need this timeout either. Seems to me like the caller should be in charge of how long it waits for a source map, via the provided context.

beater/beater.go
simitt: That's right. But when the APM Server is in fleet mode and no sourcemaps are configured, the server would still query ES to see if a sourcemap is there, when RUM events are ingested. In the fleet mode, the server should never run queries against ES. Instead, if no sourcemaps are configured, the fetcher simply shouldn't query anything, as there is no entry for the key.

beater/beater.go
axw: OK so after some digging, I found that elastic-agent is sending the fleet config to apm-server along with the apm-server specific bits. Initially I thought it wasn't because it doesn't show up in `elastic-agent inspect`.

Example:

```json
{
  "apm-server": {
    "api_key": {
      "enabled": null,
      "limit": null
    },
    "capture_personal_data": null,
    "default_service_environment": null,
    "host": ":8200",
    "max_event_size": null,
    "rum": {
      "allow_headers": null,
      "allow_origins": null,
      "allow_service_names": null,
      "enabled": null,
      "event_rate": {
        "limit": null,
        "lru_size": null
      },
      "exclude_from_grouping": null,
      "library_pattern": null,
      "response_headers": null,
      "source_mapping": {
        "elasticsearch": {
          "api_key": null
        }
      }
    },
    "secret_token": null
  },
  "data_stream": {
    "namespace": "default"
  },
  "fleet": {
    "access_api_key": "OE5WWkEzb0JhR19YNFNhdUJuajg6eWJIWEdQV25TMGl6VEF0Wjk4YTlLQQ==",
    "agent": {
      "id": "ab5f009b-1497-46ce-b799-6fa505327b41",
      "logging": {
        "level": "info"
      },
      "monitoring": {
        "http": {
          "enabled": false,
          "host": "",
          "port": 6791
        }
      }
    },
    "enabled": true,
    "host": {
      "id": "d32ae199d8deb2d5edeae9b6a8434d89"
    },
    "hosts": [
      "172.20.0.4:8220"
    ],
    "protocol": "https",
    "reporting": {
      "check_frequency_sec": 30,
      "threshold": 10000
    },
    "ssl": {
      "certificate_authorities": [
        "/etc/pki/tls/certs/fleet-ca.pem"
      ],
      "renegotiation": "never",
      "verification_mode": "full"
    },
    "timeout": "10m0s"
  },
  "id": "3d3b4503-5c02-4e12-9715-5acd264205c1",
  "meta": {
    "package": {
      "name": "apm",
      "version": "0.3.0"
    }
  },
  "name": "apm",
  "revision": 1,
  "type": "apm",
  "use_output": "default"
}
```

We'll need to update `beater/config.Fleet` to add the fleet bits we need (`hosts`, `protocol`, `ssl`, `access_api_key`), and probably remove `kibana` while we're there: https://github.com/elastic/apm-server/blob/7f0fafec377e6ac39aa758f681cab785e4d5eeea/beater/config/integration.go#L68-L70

sourcemap/store.go
stuartnelson3: I'm in favor of that

beater/beater.go
stuartnelson3: Cool, sounds good

beater/config/rum.go
```diff
@@ -79,14 +80,13 @@ func (c *RumConfig) setup(log *logp.Logger, dataStreamsEnabled bool, outputESCfg
...
 
-	var apiKey string
-	if c.SourceMapping.esConfigured {
-		if dataStreamsEnabled {
-			// when running under Fleet, the only setting configured is the api key
-			apiKey = c.SourceMapping.ESConfig.APIKey
-		} else {
-			return nil
-		}
+	if c.SourceMapping.esConfigured && len(c.SourceMapping.Metadata) > 0 {
```
stuartnelson3, line 83: This should work once a new package with the changes in `apmpackage/apm/agent/input/template.yml.hbs` is published

apmpackage/apm/agent/input/template.yml.hbs
```diff
@@ -13,6 +13,10 @@ apm-server:
     read_timeout: {{read_timeout}}
     response_headers: {{response_headers}}
     rum:
+        enabled: {{enable_rum}}
+        {{#if sourcemap_api_key}}
+        source_mapping.elasticsearch.api_key: {{sourcemap_api_key}}
+        {{/if}}
```
axw, line 19:
```suggestion
```

And remove the `sourcemap_api_key` var from manifest.yml. We won't need this after introducing the Fleet method.

beater/beater.go
```diff
@@ -622,13 +621,46 @@ func newTransformConfig(beatInfo beat.Info, cfg *config.Config) (*transform.Conf
 	return transformConfig, nil
 }
 
-func newSourcemapStore(beatInfo beat.Info, cfg config.SourceMapping) (*sourcemap.Store, error) {
-	esClient, err := elasticsearch.NewClient(cfg.ESConfig)
+func newSourcemapStore(beatInfo beat.Info, cfg config.SourceMapping, fleetCfg *config.Fleet) (*sourcemap.Store, error) {
+	if fleetmode.Enabled() {
```
axw, line 625: Maybe use `if fleetCfg != nil {` here instead? We should have fleetCfg if and only if fleetmode.Enabled(), and if we don't have it then I believe the sourcemap.NewFleetStore call will panic. I guess by checking that the config is non-nil, unit testing will be a bit easier?

sourcemap/fleet_store.go
```diff
@@ -0,0 +1,140 @@
...
+}
+
+func newFleetStore(
+	c *http.Client,
+	fleetCfg *config.Fleet,
+	cfgs []config.SourceMapMetadata,
+) (fleetStore, error) {
+	// TODO(stn): This is assuming we'll get a single fleet address right now.
+	// If we get more, how do we know which artifact is where? Are they
+	// present on all fleet-servers and we should rr through the addresses?
```
axw, line 77: If we receive more than one address, we should assume that they're all for the same cluster and any one can be used to fetch artifacts. We might have multiple addresses to support multiple different networks, so yeah probably just rr. That can be done in a followup, but if you do defer it would you mind opening an issue please?

sourcemap/fleet_store.go
```diff
@@ -0,0 +1,140 @@
...
+		body, err := ioutil.ReadAll(resp.Body)
+		if err != nil {
+			return "", fmt.Errorf("failure querying fleet: statuscode=%d response=(failed to read body)", resp.StatusCode)
+		}
+		return "", fmt.Errorf("failure querying fleet: statuscode=%d response=%s", resp.StatusCode, body)
+	}
+
+	// Looking at the index in elasticsearch, currently
+	// - no encryption
+	// - zlib compression
```
axw, line 128: It can wait until later, but I think Kibana should probably inject these details into the policy so they end up in SourceMapMetadata: compression algorithm, sha256 hash to verify data integrity.

sourcemap/fleet_store_test.go
```diff
@@ -0,0 +1,77 @@
...
+		AccessAPIKey: apikey,
+		TLS:          nil,
+	}
+
+	cfgs := []config.SourceMapMetadata{
+		{
+			ServiceName:    name,
+			ServiceVersion: version,
+			BundleFilepath: path,
+			SourceMapURL:   "/",
```
axw, line 65: also check the URL path in the handler above?

sourcemap/store_test.go
```diff
@@ -142,6 +153,145 @@ func TestStore_Fetch(t *testing.T) {
...
+			} else {
+				assert.NotNil(t, consumer)
+			}
+
+			wg.Done()
+		}()
+	}
+
+	wg.Wait()
+	assert.Equal(t, int64(1), errs)
```
axw, line 217: I feel like this test could be simplified a bit. Are multiple goroutines necessary? Can't we just make a single Fetch call in the main test function?

sourcemap/store_test.go
```diff
@@ -142,6 +153,145 @@ func TestStore_Fetch(t *testing.T) {
...
+		waitc   = make(chan struct{})
+	)
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		<-waitc
+		// zlib compress
+		wr := zlib.NewWriter(w)
+		defer wr.Close()
+		wr.Write([]byte(test.ValidSourcemap))
+	}))
+	defer ts.Close()
```
axw, line 174: unless we need to test a non-timeout after a timeout, this handler could just be `<-req.Context().Done()`

axw (commented):
This is mostly LGTM, just a handful of small things.

sourcemap/fleet_store.go
stuartnelson3: :+1:

sourcemap/fleet_store.go
stuartnelson3: Sounds good. I'll open the issue since this PR has been open for a long time already.

stuartnelson3: https://github.com/elastic/apm-server/issues/5514

sourcemap/store_test.go
stuartnelson3: which very helpfully pointed out I wasn't attaching the passed-in context to the request :)

stuartnelson3:
@axw let me know if you spot anything else when you get the chance

beater/beater_test.go
```diff
@@ -280,3 +282,27 @@ func TestTransformConfig(t *testing.T) {
 	test(true, false, false)
 	test(true, true, true)
 }
+
+func TestStoreUsesRUMElasticsearchConfig(t *testing.T) {
```
axw, line 286: Can we add a basic unit test that shows the Fleet sourcemap store backend is used when Fleet config is defined?

sourcemap/fleet_store.go
```diff
@@ -0,0 +1,140 @@
...
+	}
+	return newStore(s, logger, expiration)
+}
+
+func newFleetStore(
+	c *http.Client,
+	fleetCfg *config.Fleet,
+	cfgs []config.SourceMapMetadata,
+) (fleetStore, error) {
+	// TODO(stn): This is assuming we'll get a single fleet address right now.
```
axw, line 75: can you please update the comment to resolve the question? maybe link to the issue

sourcemap/fleet_store_test.go
```diff
@@ -0,0 +1,79 @@
...
+		// zlib compress
+		wr := zlib.NewWriter(w)
+		defer wr.Close()
+		wr.Write([]byte(wantRes))
+	}))
+	defer ts.Close()
+
+	fleetCfg := &config.Fleet{
+		Hosts:        []string{ts.URL},
+		Protocol:     "https",
```
axw, line 57: This doesn't look realistic - in particular the protocol is wrong, and hosts just contains host:port pairs without a scheme. See https://github.com/elastic/apm-server/pull/5410#discussion_r650462506 for real values

Can we please adjust the test accordingly?

axw (approved):
just a few more minor comments, otherwise LGTM!

sourcemap/fleet_store_test.go
stuartnelson3: I'm confused, the comment you linked shows protocol as being "https":

image

will update the hosts to just be host:port :+1:

stuartnelson3: Ah, you mean `http` since we're using an http server?

axw: Yup that's right. The comment I linked to was from running a real fleet-server with TLS.

stuartnelson3:
jenkins run the tests

axw:
I tested this with the current Kibana API which requires sending the source map as a string, rather than as a file.

It works, but there's a small issue. I essentially manually ran this system test, but using the Kibana API for uploading the source map: https://github.com/elastic/apm-server/blob/29cfed315a9f20205b9fe0dba829f18489d943f7/systemtest/sourcemap_test.go#L38-L60

The issue is that the "bundle_filepath" is apparently not being cleaned. The path with "e2e/../e2e" in it does not match, but when I manually clean it (i.e. change that substring to "e2e"), then it does match.

The existing APM Server endpoint cleans the URL path before storing it in Elasticsearch: https://github.com/elastic/apm-server/blob/29cfed315a9f20205b9fe0dba829f18489d943f7/beater/api/asset/sourcemap/handler.go#L130

Kibana should probably do the same. Alternatively we could clean it after receiving it in the APM Server, but seems better to store it cleaned in the first place. I'll open a Kibana issue.

axw:
Calling this one working as expected, there's elastic/kibana#105911 open to sort out the issue noted above.

PR #5414, opened 2021-06-04
apmpackage: remove version directory

axw:
## Motivation/summary

Remove the version from the apm integration package directory. It is not necessary, and the version bumping makes retaining git history more difficult. See: https://github.com/elastic/apm-server/commits/master/apmpackage/apm/0.3.0

With these changes, bumping the version should just be a matter of updating `apmpackage/apm/manifest.yml` and adding a changelog entry.

## How to test these changes

Non-functional change.


script/copy_package.py
```diff
@@ -84,7 +81,7 @@ def bump(v):
     # copy over the package and replace version in manifest and pipeline names
     shutil.copytree(src, dst)
     subprocess.check_call('rm -rf {0}'.format(os.path.join(dst, 'README.template.md')), shell=True)
-    cmd = 'find {0} -not -name "*.png" -type f -print0 | xargs -0 sed -i "" "s/{1}/{2}/g"'.format(
+    cmd = 'find {0} -not -name "*.png" -type f -exec sed -i -e "s/{1}/{2}/g" {{}} \\;'.format(
```
axw, line 84: I didn't look too deeply into it, but the script wouldn't run for me as it was -- I guess some discrepancy between the version of find shipped with macOS.

simitt: changes look good - I ran `make update` with the changed code and everything still worked fine on a macOS

simitt (approved):

script/copy_package.py
axw: thanks for testing!

jalvz:
I guess apm-integration-testing's `--package-registry-apm-path` will not work right out of the box when pointing it to `apm-server/apmpackage`; one would need to copy the package elsewhere and stick in the version folder. 
just an fyi

axw:
Ah, I didn't know about that one. Indeed, we'd need to do something like https://github.com/elastic/apm-server/blob/master/testing/docker/package-registry/entrypoint.sh

axw:
Opened https://github.com/elastic/apm-integration-testing/issues/1169

Issue #5377, opened 2021-05-31
Send APM Server configurations to Kibana APM endpoint on startup


simitt:
APM Server needs to send its configuration values to a Kibana APM endpoint on startup when running on ESS/ECE. The request is only made if an `apm-server.kibana.*` section is configured and APM Server is running on Elastic cloud (ENV indicating that apm-server is running on cloud, or `apm-server.kibana` config option, ..).

Ideally the configuration options are already flattened (dotted), and any credentials for ES and Kibana are removed. APM Server startup should not be blocked by the request, the existing ConnectingClient implementation should be reusable for setting up a connecting and retrying.

simitt:
Align with the @elastic/apm-ui team on the concrete endpoint, related to https://github.com/elastic/kibana/issues/100657.

stuartnelson3:
>Ideally the configuration options are already flattened (dotted)

Is there support for this already within apm-server or some other elastic libraries?

stuartnelson3:
>Ideally the configuration options are already flattened (dotted)

Should everything be flattened, or just objects that have single keys?

eg.

```
{parent: {child: 1, other_child: 2}, 'other_parent.child': 3}
// or
{'parent.child': 1, 'parent.other_child': 2, 'other_parent.child': 3}
```

simitt:
Everything should be flattened, so it would just be `key: value` pairs. Beats `common.MapStr` has a `flatten` method https://github.com/elastic/beats/blob/master/libbeat/common/mapstr.go#L269-L291 that you might want to reuse or use for inspiration.

PR #5424, opened 2021-06-09
Send APM Server config to Kibana

stuartnelson3:
## Motivation/summary

Send a sanitized and flattened version of the apm-server config to kibana on
startup.

Closes https://github.com/elastic/apm-server/issues/5377

## Checklist

## How to test these changes

Start up kibana, elasticsearch and apm-server. Verify that the apm-server
config is indexed in elasticsearch.

## Related issues

Depends on https://github.com/elastic/kibana/issues/100657


stuartnelson3:
I'm removing anything involving `ESConfig` or `Kibana`, is there anything else that should be removed? `apm-server.ssl.keyphrase` looks like it should probably be removed if set.

Question regarding the flattened format: Do we want to do any sort of lower camelcasing for the config values? Or just leave them with their default formatting, ie.

```json
{
  "APIKeyConfig.Enabled": false,
  "APIKeyConfig.LimitPerMin": 100,
  "AgentConfigs": null,
  "Aggregation.ServiceDestinations.Enabled": true,
  "Aggregation.ServiceDestinations.Interval": 60000000000,
  "Aggregation.ServiceDestinations.MaxGroups": 10000,
  "Aggregation.Transactions.Enabled": true,
  "Aggregation.Transactions.HDRHistogramSignificantFigures": 2,
  "Aggregation.Transactions.Interval": 60000000000,
  "Aggregation.Transactions.MaxTransactionGroups": 10000,
  "AugmentEnabled": true,
  "DataStreams.Enabled": false,
  "DefaultServiceEnvironment": "",
  "Expvar.Enabled": false,
  "Expvar.URL": "/debug/vars",
  "Host": "localhost:8200",
  "IdleTimeout": 45000000000,
  "JaegerConfig.GRPC.AuthTag": "",
  "JaegerConfig.GRPC.Enabled": false,
  "JaegerConfig.GRPC.Host": "localhost:14250",
  "JaegerConfig.GRPC.TLS": null,
  "JaegerConfig.HTTP.Enabled": false,
  "JaegerConfig.HTTP.Host": "localhost:14268",
  "MaxConnections": 0,
  "MaxEventSize": 307200,
  "MaxHeaderSize": 1048576,
  "Mode": 0,
  "Pipeline": "apm",
  "Pprof.BlockProfileRate": 0,
  "Pprof.Enabled": false,
  "Pprof.MemProfileRate": 0,
  "Pprof.MutexProfileRate": 0,
  "ReadTimeout": 30000000000,
  "Register.Ingest.Pipeline.Enabled": true,
  "Register.Ingest.Pipeline.Overwrite": false,
  "Register.Ingest.Pipeline.Path": "ingest/pipeline/definition.json",
  "ResponseHeaders": null,
  "RumConfig.AllowHeaders": [],
  "RumConfig.AllowOrigins": [
    "*"
  ],
  "RumConfig.AllowServiceNames": null,
  "RumConfig.Enabled": false,
  "RumConfig.EventRate.Limit": 300,
  "RumConfig.EventRate.LruSize": 1000,
  "RumConfig.ExcludeFromGrouping": "^/webpack",
  "RumConfig.LibraryPattern": "node_modules|bower_components|~",
  "RumConfig.ResponseHeaders": null,
  "RumConfig.SourceMapping.Cache.Expiration": 300000000000,
  "RumConfig.SourceMapping.Enabled": true,
  "RumConfig.SourceMapping.IndexPattern": "apm-*-sourcemap*",
  "Sampling.KeepUnsampled": true,
  "Sampling.Tail.Enabled": false,
  "Sampling.Tail.IngestRateDecayFactor": 0.25,
  "Sampling.Tail.Interval": 60000000000,
  "Sampling.Tail.Policies": null,
  "Sampling.Tail.StorageDir": "tail_sampling",
  "Sampling.Tail.StorageGCInterval": 300000000000,
  "Sampling.Tail.TTL": 1800000000000,
  "SecretToken": "",
  "SelfInstrumentation.APIKey": "",
  "SelfInstrumentation.Enabled": false,
  "SelfInstrumentation.Environment": "",
  "SelfInstrumentation.Hosts": null,
  "SelfInstrumentation.Profiling.CPU.Duration": 10000000000,
  "SelfInstrumentation.Profiling.CPU.Enabled": false,
  "SelfInstrumentation.Profiling.CPU.Interval": 60000000000,
  "SelfInstrumentation.Profiling.Heap.Enabled": false,
  "SelfInstrumentation.Profiling.Heap.Interval": 60000000000,
  "SelfInstrumentation.SecretToken": "",
  "ShutdownTimeout": 5000000000,
  "TLS": null,
  "WriteTimeout": 30000000000
}
```

axw:
> I'm removing anything involving ESConfig or Kibana, is there anything else that should be removed? apm-server.ssl.keyphrase, instrumentation.api_key, and instrumentation.secret_token all look like candidates.

None of these can be set in ESS, but to be on the safe side I think it would be a good idea to remove:
 - `apm-server.ssl.*`
 - `apm-server.instrumentation.*`
 - `*.elasticsearch.*`
 - `instrumentation.*`

@simitt will that work?

> Question regarding the flattened format: Do we want to do any sort of lower camelcasing for the config values? Or just leave them with their default formatting, ie.

I think we should be sending the libbeat config input, rather than the apm-server config into which it is unmarshaled. The raw libbeat config is accessible in https://github.com/elastic/apm-server/blob/7f55bc8607c0856089a975be49a086d261fc0b56/beater/beater.go#L135

simitt:
> None of these can be set in ESS, but to be on the safe side I think it would be a good idea to remove:
> 
>     apm-server.ssl.*
>     apm-server.instrumentation.*
>     *.elasticsearch.*
>     instrumentation.*

Following ssl related settings need to be synced: `apm-server.ssl.enabled`, `apm-server.ssl.certificate`, `apm-server.ssl.key`. Removing the other settings is fine. 

> I think we should be sending the libbeat config input, rather than the apm-server config into which it is unmarshaled.

Why is that? On the Kibana side we need to define a list of supported config options (hardcoded).

axw:
> Why is that? On the Kibana side we need to define a list of supported config options (hardcoded).

A couple of reasons:

- If we refactor the config structs (e.g. rename fields), then the JSON will change. This should be stable for the rest of 7.x, right? We could add `json:"..."` struct tags, but we already have a defined stable config format so may as well use that.
- The config that's being marshalled at the moment includes defaults, which I don't think should be transferred?

simitt:
@axw good points; +1

stuartnelson3:
>I think we should be sending the libbeat config input, rather than the apm-server config into which it is unmarshaled. The raw libbeat config is accessible in 
How do we marshal

kibana/send_config.go
```diff
@@ -0,0 +1,110 @@
...
+	"net/http"
+	"strings"
+	"time"
+
+	"github.com/elastic/beats/v7/libbeat/common"
+)
+
+// TODO: Get this value from Oliver
+// https://github.com/elastic/kibana/issues/100657
+const kibanaConfigUploadPath = "/apm/settings/schema"
```
ogupte, line 35: We're arraning fleet-related APIs under `/apm/fleet/` so we'll probabaly use something like:
`/apm/fleet/apm_server_settings`

ogupte:
In the Kibana PR, I'm defining only the configs which map to input vars in the package policy:

https://github.com/elastic/kibana/pull/102682/files#diff-ef993cd3669a97d649974bea6f79a1c2406abbf7ca7f1cc0d86cda4df7d2c5adR231-R254

```json
{
  'apm-server.host': 'host',
  'apm-server.secret_token': 'secret_token',
  'apm-server.rum.enabled': 'enable_rum',
  'apm-server.default_service_environment': 'default_service_environment',
  'apm-server.rum.allow_service_names': 'rum_allow_service_names',
  'apm-server.rum.allow_origins': 'rum_allow_origins',
  'apm-server.rum.allow_headers': 'rum_allow_headers',
  'apm-server.rum.event_rate.limit': 'rum_event_rate_limit',
  'apm-server.rum.event_rate.lru_size': 'rum_event_rate_lru_size',
  'apm-server.max_event_size': 'max_event_bytes',
  'apm-server.capture_personal_data': 'capture_personal_data',
  'apm-server.max_header_size': 'max_header_bytes',
  'apm-server.idle_timeout': 'idle_timeout',
  'apm-server.read_timeout': 'read_timeout',
  'apm-server.shutdown_timeout': 'shutdown_timeout',
  'apm-server.write_timeout': 'write_timeout',
  'apm-server.max_connections': 'max_connections',
  'apm-server.expvar.enabled': 'expvar_enabled',
  'output.elasticsearch.ssl.enabled': 'tls_enabled',
  'output.elasticsearch.ssl.certificate': 'tls_certificate',
  'output.elasticsearch.ssl.key': 'tls_key',
  'output.elasticsearch.ssl.supported_protocols': 'tls_supported_protocols',
  'output.elasticsearch.ssl.cipher_suites': 'tls_cipher_suites',
  'output.elasticsearch.ssl.curve_types': 'tls_curve_types',
}
```
Anything that isn't included in this map will be displayed to the user as unsupported.

stuartnelson3:
@ogupte we've made a couple changes to the input vars but haven't pushed them to the package policy yet. There's one final change for 7.14 in https://github.com/elastic/apm-server/pull/5410, and then we'll pushed an updated version.

The config should look like the example you posted in your comment, ie. with the full path for each config. The example I posted above was from only marshalling the things nested under `apm-server`, but the full raw config will be marshaled.

kibana/send_config.go
```diff
@@ -0,0 +1,110 @@
...
+				continue
+			}
+
+			// Are there other kinds of recoverable errors?
+			return err
+		}
+		// TODO: What sort of response will we get?
+		if resp.StatusCode > http.StatusOK {
+			return fmt.Errorf("bad response %s", resp.Status)
+		}
```
simitt, line 71: Maybe cover the scenario where Kibana < `7.14` and therefore doesn't support the endpoint. (log and stop retrying)

kibana/send_config_test.go
```diff
@@ -0,0 +1,85 @@
...
+      limit: 300
+      lru_size: 1000
+output.elasticsearch:
+  hosts: ["localhost:9200"]
+  enabled: true
+  compression_level: 0
+  protocol: "https"
+  username: "elastic"
+  password: "changeme"
+  worker: 1
```
simitt, line 84: `instrumentation` and `kibana` are not covered here; can you also cover some config option that doesn't start with `apm-server` to check it is sent.

simitt (commented):
generally looks good, left minor comments

simitt:
I just realized that the config options only contain the `apm-server.*` options, and not the `output.*`, `setup.*`, etc related config options, and that the options are not prefixed with `apm-server.*` anymore at this point (for example `apm-server.ssl` is only `ssl` at the point of flattening). 

Since only apm-server options are supported as input vars in the apm integration, I guess that is fair, but we need to add some note to the APM migration UI that none of the other config options is ever supported.  

The config options are currently sent without the `apm-server.*` prefix.  The `flattenAndClean` check assumes that the whole configuration (including `output.*`, `instrumentation.*`), etc. is there and that the server specific settings are prefixed with `apm-server.*` which is not the case. Please change the logic and the testcase accordingly. When sending the options to Kibana, the `apm-server` prefix should be added, to make it straight forward on the Kibana side.

axw:
If we need to, I believe it is still possible to get the parent config: cast the `*common.Config` to `*ucfg.Config`, and use the latter's Parent method: https://pkg.go.dev/github.com/elastic/go-ucfg#Config.Parent

kibana/send_config_test.go
```diff
@@ -82,4 +84,12 @@ output.elasticsearch:
...
   password: "changeme"
   worker: 1
+instrumentation:
+  enabled: false
+  environment: ""
+  hosts:
+  - http://remote-apm-server:8200
+kibana:
+  enabled: true
+  api_key: abc123
```
simitt, line 94: `kibana` should be nested inside `apm-server.`

simitt (approved):
Apart from a minor comment on the testcase, changes LGTM now.

simitt:
Tested this on an ECE instance.

Issue #5173, opened 2021-04-27
Mobile spans don't line up with server side spans


bryce-b:
<img width="1843" alt="Screen Shot 2021-04-26 at 10 44 06 AM" src="https://user-images.githubusercontent.com/75274611/116285620-27f5ce00-a743-11eb-9785-02f536a96063.png">

The timestamps generated on the mobile device can be significantly different than the server-side timestamp. 

It sounds like this issue has already been identified and solved for the rum agent. (per @axw "[Rum uses] the timestamp the event payload was received from the agent[...]") That same logic should be applied to mobile transactions, as well.

A few questions: 
1. Mobile data is sent using the OpenTelemetry Protocol. Will that cause issues with this solution?
2. Will all spans in the mobile transaction need to be "re-tuned" based off the server side timestamp?

Let me know if there is any additional information needed.

felixbarny:
It works for RUM as it sends the spans within the transaction document. Implementing a similar functionality on top of OTLP would be a bit more challenging as events can be interleaved.


Potential issues
* Is it guaranteed that all spans of a single transaction are sent to the same server if there's a load balancer in front of multiple APM Servers? Not quite sure how load balancing works for gRPC - when and how often are new requests started.
* Spans may start after their parent ends (example: button click creates a short span that schedules a background action)


Solution approaches
* APM Server could buffer spans sent by mobile agents until the corresponding transaction is received and then adjust the span and transaction timestamps accordingly.
* An alternative that also accounts for spans that end after their transaction ends: The first span of a trace that's received by APM Server determines the offset (arrival time-span duration). The offset is stored in a map that has the trace id as a key. The key is cleared either if a span has no parent id (root span) or after a timeout.
* The mobile agent buffers spans (on disk or in memory) until all spans of a trace are ended. It then sends all spans in batch to APM Server. The mobile agent could also add a custom attribute to the root span to indicate for how long it has been queued. Open question: can we ensure it's sent to the same Server?
* Send data using the RUM intake API protocol.

axw:
We also have some outstanding issues with how timestamps are dealt with in RUM: https://github.com/elastic/apm-server/issues/1830. Ideally I'd like to come up with a solution which addresses both.

[...]

PR #5433, opened 2021-06-10
processor/otel: adjust timestamps for mobile

axw:
## Motivation/summary

Add support for a custom resource attribute, `telemetry.sdk.elastic_export_timestamp`, representing the time at which the agent sent the event payload to the server.

The server uses this to adjust the timestamps of events, to cater for end-user devices having incorrect system clock settings. When the server receives an OTLP payload, it records its current system time as `receive_timestamp`. Then it calculates `time_delta = receive_timestamp - export_timestamp`, and adds this to the timestamp of each event in the payload.

For example, let's say the client's clock is 5 minutes behind. It starts a span at 12:00, ends it at 12:05, and exports it to the server at 12:10; the server receives it at 13:10. The server records:
 - `receive_timestamp = 13:10`
 - `export_timestamp = 12:10`
 - `time_delta = 13:10-12:10 = 1h`

This is then added to the timestamp of the span, adjusting it to 13:00.

One important thing to note: this change means the network latency will always be included in the timestamp. If the client's clock is correct and network latency is significant, then we could increase error. We may want to later introduce a time delta threshold for the adjustment, i.e. don't apply a delta if it's less than Xms.

## Checklist

## How to test these changes

1. Once https://github.com/elastic/apm-agent-ios/pull/36 is merged, create some distributed traces with the iOS agent and opbeans-*
2. Check that they show up correctly on the waterfall.

## Related issues

Closes https://github.com/elastic/apm-server/issues/5173


axw:
@felixbarny tagged you for review just to get your OK on the approach. Reading the description should be enough :)

felixbarny (approved):
SGTM

processor/otel/consumer_test.go
```diff
@@ -578,6 +578,74 @@ func TestArrayLabels(t *testing.T) {
...
+func TestConsumeTracesExportTimestamp(t *testing.T) {
+	traces, otelSpans := newTracesSpans()
+
+	// The actual timestamps will be non-deterministic, as they are adjusted
+	// based on the server's clock.
+	//
+	// Use a large delta so that we can allow for a significant amount of
+	// delay in the test environment affecting the timestamp adjustment.
+	const timeDelta = time.Hour
+	const allowedError = 5 // seconds
```
stuartnelson3, line 590: I've had success testing time using a helper func to wrap `time.Now()`, so that the helper func can be redefined in tests and return a specific given time, ie.
```go
timeNow = func() time.Time { return time.Now() }
```
and then over-writing it to return a specific time in the tests, if you'd rather do that than have a delta. I'm fine with either.

stuartnelson3 (approved):
small comment, but that's all :+1:

processor/otel/consumer_test.go
axw: I'm not super keen on patching unexported variables, as I'd prefer to test using only the exported API. In the past I've gone all in on having a `Clock` interface which can be passed in. I'd like to avoid this as it's pretty invasive.

Let's stick with the current way and we can see how it goes, happy to come back and reassess.

axw:
jenkins run the tests please

simitt (approved):

simitt:
Tested with BC2 - distributed trace shows up, timeline seems legit 
Screenshot 2021-07-13 at 21 56 32

Issue #5347, opened 2021-05-26
iOS intake support: auth handling and rate limiting


simitt:
The [iOS agent](https://github.com/elastic/apm-agent-ios) uses the opentelemetry Intake support of the APM Server to send apm events, rather than the RUM intake. This raises some questions related to auth handling and rate limiting:

* Whenever `apm-sever.api_key.enabled: true` or an `apm-server.secret_token` is configured, a valid API Key or secret token needs to be sent with the request, also when sent via opentelemetry intake. Investigate how to handle this for requests from the iOS intake.

* Investigate whether some similar rate limiting as currently enabled for RUM endpoints should be enabled for iOS requests.

axw:
My current thinking is to introduce a formal concept of "anonymous access", along the lines of what is supported by Elasticsearch: https://www.elastic.co/guide/en/elasticsearch/reference/current/anonymous-access.html

While we're at it, we could group the existing auth types into a new group:

```yaml
apm-server:
  #auth:
    #secret_token:
    #api_key:
      #enabled: false
      #limit: 100
    #anonymous:
      #allow_rum: false
      #allow_mobile: false
      #allow_service_names: []
      #event_rate:
        #limit: 300
        #cache_size: 1000
```

By setting `allow_rum: true`, we would allow unauthenticated HTTP requests to `/config/v1/rum/agents`, `/intake/v2/rum/events`, and `/intake/v3/rum/events`.

By setting `allow_mobile: true`, we would allow unauthenticated OTLP/gRPC requests as long as they can be identified as originating from an iOS agent.

For both RUM and iOS, requests could be forged so that the client appears to be one of the expected agent types. It is already the case that you can disable/enable the RUM endpoints, which is effectively the same as enabling/disabling unauthenticated access. The same applies to OTLP/gRPC: to combat abuse we can (for example) check for an expected User-Agent string, as opposed to checking the client requested `/intake/v2/rum/events` rather than `/intake/v2/events`.

axw:
Another option would be to make it a bit more generic, and define the allowlist in terms of agent name. e.g.

```yaml
apm-server:
  #auth:
    #anonymous:
    #  allow_agent: ["rum-js", "iOS/Swift"]
    #  allow_service: [] 
```

It's no big deal to have "rum-js" in the list by default since RUM is disabled by default. Having iOS in there requires discussion.

[...]

PR #5444, opened 2021-06-14
[fleet integration] Support all apm-server config options and adapt defaults.

simitt:
## Motivation/summary
* Add missing `apm-server.*` settings that should be supported to the APM Server Integration.
* Adapt default values to be aligned with cloud values where possible.

## Checklist

- [x] Update [CHANGELOG.asciidoc](https://github.com/elastic/apm-server/blob/master/CHANGELOG.asciidoc)

## How to test these changes
- Spin up a local package registry and copy over the `apmpackage` information to a new version `0.3.0`. 
- Spin up an elastic-agent 
- Add an apm-integration and ensure the new config values are there

## Related issues

implements #4528

## Preview
Screenshot 2021-06-14 at 09 59 39
Screenshot 2021-06-14 at 09 59 46

PR #5457, opened 2021-06-15
Introduce `apm-server.auth.*` config

axw:
## Motivation/summary

Introduce the new AgentAuth config structure, which holds API Key and secret token auth. Later we will add "anonymous" auth here too, and deprecate/replace some RUM config (rate limiting and allowed service names).

We also introduce a new YAML naming scheme for the config, `apm-server.auth.*`. The old config is deprecated and copied across to the new config fields.

## Checklist

## How to test these changes

1. Set `apm-server.secret_token` and `apm-server.api_key.*`, make sure they are honoured (e.g. query "GET /" with/out auth)
2. Set `apm-server.auth.secret_token` and `apm-server.auth.api_key.*`, same again.
3. Set both new and old with different values: check that a warning is logged that the old config is ignored, and check that it is possible to auth using the new but not old

## Related issues


docs/configuration-process.asciidoc
```diff
@@ -99,6 +103,8 @@ The token is not used for RUM endpoints. By default, no authorization token is s
 It is recommended to use an authorization token in combination with SSL enabled.
 Read more about <<securing-apm-server, Securing APM Server>> and the <<secret-token, secret token>>.
 
+deprecated::[7.14.0, "In versions prior to 7.14.0, secret token authorization was known as `apm-server.secret_token`. In 7.14.0 this was renamed `apm-server.auth.secret_Token`. The old configuration will continue to work until 8.0.0, and the new configuration will take precedence."]
```
simitt, line 106: typo: `secret_Token`

simitt (commented):
Looks great.

docs/configuration-process.asciidoc
axw: thanks, fixed

axw:
@bmorelli25 would you please take a look at the docs changes? I've renamed some config, and added new deprecation sections for the deprecated config names.

I've also moved "api_key.* configuration options" into the "API keys" section, not sure if it was intentional that it was separate? I can move back if you prefer, but then I'm not sure where the deprecation section should go.

simitt (changes requested):
Can you also change the `apmpackage` please. I suggest we remove the deprecated settings and only support the new ones.

axw:
@simitt will do. I was going to wait for your changes (https://github.com/elastic/apm-server/pull/5444) to land, but I'll just update now for the purposes of review and update again when that lands.

simitt (approved):

bmorelli25 (approved):
Docs look great–thanks!

It looks like there's one [additional change](https://github.com/elastic/apm-server/blame/master/docs/copied-from-beats/docs/command-reference.asciidoc/#L142-L143) that needs to be made to the command reference file. I'll fix that in the Beats repo and copy it over to apm-server before 7.14.

axw:
Failure is related to the beats update, which will be resolved by https://github.com/elastic/apm-server/pull/5471

simitt:
Tested with BC2:
* old settings work as expected
* new settings work as expected
* API Key auth disabled when:
```
apm-server.api_key.enabled: true
apm-server.auth.api_key.enabled: false
```
  * Secret token 'abcd' expected
```
apm-server.secret_token: 'xxx'
apm-server.auth.secret_token: 'abcd'
```

Issue #5466, opened 2021-06-16
[Intake API] make specific destination.service fields optional and deprecate them


simitt:
Based on https://github.com/elastic/apm/pull/436 the fields `span.context.destination.service.name` and `span.context.destination.service.type` should be optional and deprecated at the Intake API level. The information is not used anywhere throughput the APM UI and apm agents are going to stop sending the information.

PR #5468, opened 2021-06-16
deprecate span destination name and type fields

jalvz:
## Checklist

## How to test these changes

Send some events to apm-server with span.destination.service.resource and without span.destination.service.name nor span.destination.service.type.
Events should be accepted.

## Related issues

Closes  https://github.com/elastic/apm-server/issues/5466


simitt (changes requested):
To make the change on the Intake API you need to add a `Deprecated` note in the comment and remove the `required` tag in the [modeldecoder](https://github.com/elastic/apm-server/blob/master/model/modeldecoder/v2/model.go#L702-L707). Running `make update` will update the `docs/spec` accordingly. (Same for rumv3 modeldecoder).

jalvz:
thanks @simitt , didn't know about that

model/modeldecoder/v2/model.go
```diff
@@ -700,14 +700,16 @@ type spanContextDestination struct {
 
 type spanContextDestinationService struct {
 	// Name is the identifier for the destination service,
-	// e.g. 'http://elastic.co', 'elasticsearch', 'rabbitmq'
-	Name nullable.String `json:"name" validate:"required,maxLength=1024"`
+	// e.g. 'http://elastic.co', 'elasticsearch', 'rabbitmq' (
+	// This field is DEPRECATED and will be removed in a future release
```
simitt, line 704: nit: Could you start with the deprecated note so it shows up at the beginning of the Intake API comment?
Similar to https://github.com/elastic/apm-server/blob/f29f10905bde682d833866c0881f282b88c881b3/model/modeldecoder/v2/model.go#L495

simitt (approved):

axw:
Tested with 7.14 BC1. I hacked the Go Agent to not send the destination service type or name, only resource; the server accepted this and record resource and neither of the other empty/missing fields.

Issue #5470, opened 2021-06-17
Periodically report applied agent configuration


axw:
The server should periodically report which agent configuration, injected into the server config via Fleet (https://github.com/elastic/apm-server/issues/5018), has been applied by one or more agents.

See the description of https://github.com/elastic/apm-server/issues/5018 for some thoughts on how we might go about this. Another option not listed there might be to use the [Manager.SetPayload](https://github.com/elastic/beats/blob/4684ac7ed3da20ecc533af492862cdb029f5cb3a/x-pack/libbeat/management/manager.go#L186) method to record information in the integration's status. I don't know if this is possible, just something to look into.

stuartnelson3:
I've tracked metadata like this before via prometheus metrics. What would you think about some sort of gauge vec (metric type is sort of unimportant since we're kind of abusing metrics) where we track etag as label? Should be able to get the instantaneous status, and time to full convergence, and if there are any agents who are stuck on an old version.

axw:
I think that would be a fine approach. It's kinda what I was thinking in #5018:

> Alternatively, we could have agents periodically send an internal agent statistics event which includes the current config Etag value. This would have the benefit of enabling us to see how many and which agents have applied the config.

But I assume you mean we would do it in the server, and not in agents.

stuartnelson3:
I think the agents would be the right place actually

stuartnelson3:
As a first step we'll add it to the server, and evaluate from there

PR #5481, opened 2021-06-17
record per-agent config status

stuartnelson3:
## Motivation/summary

~~Record the clientIP, service name, service
environment, and etag for the applied config, for
each agent request. This can be used to see the
current state of config application, convergence
towards a new config, or if there are services
that are running old config versions.~~

Record the etag for applied agent_configs in elastic search.

closes https://github.com/elastic/apm-server/issues/5470

## Checklist

## How to test these changes

Running apm-server under fleet, start an agent that requests its config from apm-server. Update the config, and verify that the timeseries has been ingested into elasticsearch.


axw (commented):
That's a neat use of the prometheus integration :slightly_smiling_face:

A have a couple of concerns:
 - I think the tracer is disabled unless libbeat instrumentation is enabled, and it's disabled by default. Enabling by default is not an option yet.
 - IIANM, the Prometheus client doesn't automatically forget label values once they've been seen. The agent config API can be called by RUM without any authentication, so we would need to delete label values to avoid memory exhaustion attacks.

As nice as it is to be able to see which agents are on what config version, we don't have that ability today. I think we should pare it back a bit to just do what we today, and we can expand on it later. I believe that today we just need a single dimension, etag. To avoid memory exhaustion we should only maintain and report metrics for currently known etag values.

stuartnelson3:
>IIANM, the Prometheus client doesn't automatically forget label values once they've been seen. The agent config API can be called by RUM without any authentication, so we would need to delete label values to avoid memory exhaustion attacks.

That's a super good point. Even if someone is just in a highly dynamic environment, the number of different (legitimate) clientIPs could become an issue.

stuartnelson3:
>I think the tracer is disabled unless libbeat instrumentation is enabled, and it's disabled by default. Enabling by default is not an option yet.

Should we move to indexing via es client then?

axw:
> Should we move to indexing via es client then?

I think we should use the libbeat publisher for now at least. I don't know exactly how that will look, but for example you could have `agentcfg.DirectFetcher` keep track of which config has been applied (queried with the current Etag value, or MarkAsAppliedByAgent); and then have something periodically scraping that and creating `model.Metricsets` and sending them to the batchProcessor passed to beater.newServer

stuartnelson3:
K, pushed up an idea of how it might look. When you have a chance let me know what you think. If it's looking reasonable, what should I be putting into each `model.Metricset`?

axw (commented):
This approach looks good. We should only report config as "applied" if one of the following is true though:
 - `agentcfg.Query.Etag == agentcfg.Result.Etag` (i.e. we know the agent has previously received the config)
 - `agentcfg.Query.MarkAsAppliedByAgent == true` (for clients that don't/can't use the Etag approach)

https://github.com/elastic/apm-server/blob/f5664d8196599758da0dd1e1c88771d24e0f4870/agentcfg/model.go#L58-L69

> If it's looking reasonable, what should I be putting into each model.Metricset?

Here's what I'd suggest:

```yaml
 - Metricset.Name: "agent_config"
 - Metricset.Labels: {"etag": <etag>}
 - Metricset.Samples:
   - Name: "agent_config_applied"
     Value: 1
```

The constant metric value is not ideal, but good enough to start with.

agentcfg/reporter.go
```diff
@@ -0,0 +1,81 @@
...
+func (r Reporter) Run(ctx context.Context) error {
+	t := time.NewTicker(30 * time.Second)
+	defer t.Stop()
+	for {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case result := <-r.resultc:
+			if _, ok := r.applied[result.Source.Etag]; !ok {
+				r.applied[result.Source.Etag] = struct{}{}
```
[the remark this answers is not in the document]
stuartnelson3: :+1: makes sense

agentcfg/reporter.go
```diff
@@ -0,0 +1,81 @@
...
+)
+
+type Reporter struct {
+	f       Fetcher
+	p       model.BatchProcessor
+	resultc chan Result
+
+	// keeps track of the agent_configs that have been queried and applied
+	// to agents.
+	applied map[string]struct{}
```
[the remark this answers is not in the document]
stuartnelson3: makes sense to me

stuartnelson3:
jenkins run the tests

agentcfg/reporter.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+}
+
+func (r Reporter) Fetch(ctx context.Context, query Query) (Result, error) {
+	result, err := r.f.Fetch(ctx, query)
+	// Only report configs when the query etag == current config etag, or
+	// when the agent indicates it has been applied.
+	if err == nil &&
+		(query.Etag == result.Source.Etag ||
+			query.MarkAsAppliedByAgent != nil && *query.MarkAsAppliedByAgent) {
```
stuartnelson3, line 54: Sort of wonky formatting, wasn't sure if i should collapse it all into a single boolean like this or if two would make it easier to read.

beater/server.go
```diff
@@ -126,24 +129,25 @@ func newServer(
...
-		cfg:          cfg,
-		httpServer:   httpServer,
-		grpcServer:   grpcServer,
-		jaegerServer: jaegerServer,
+		logger:        logger,
+		cfg:           cfg,
+		httpServer:    httpServer,
+		grpcServer:    grpcServer,
+		jaegerServer:  jaegerServer,
+		fetchReporter: fetchReporter,
```
axw, line 150: can we give this a more descriptive name? it fetches and reports what? agentcfgFetchReporter?

agentcfg/reporter.go
```diff
@@ -0,0 +1,97 @@
...
+				applied[result.Source.Etag] = struct{}{}
+			}
+			continue
+		case <-t.C:
+		}
+		batch := new(model.Batch)
+		for etag := range applied {
+			m := &model.Metricset{
+				Name:    "agent_config",
+				Labels:  make(common.MapStr, 1),
```
axw, line 82:
```suggestion
				Labels:  common.MapStr{"etag": etag},
```

agentcfg/reporter.go
```diff
@@ -0,0 +1,97 @@
...
+				Samples: []model.Sample{{Name: "agent_config_applied", Value: 1}},
+			}
+			m.Labels.Put("etag", etag)
```
axw, line 85:
```suggestion
```

axw (changes requested):
Basically LGTM, just need to make sure we select on ctx.Done() and don't do a naked send on the channel - that could block the handler on shutdown.

When we come to https://github.com/elastic/apm-server/issues/5487, we should make sure to check that these metrics are reported in a system test.

agentcfg/reporter.go
```diff
@@ -0,0 +1,97 @@
...
+		f:        f,
+		p:        batchProcessor,
+		interval: interval,
+		logger:   logger,
+		resultc:  make(chan Result),
+	}
+}
+
+func (r Reporter) Fetch(ctx context.Context, query Query) (Result, error) {
+	result, err := r.f.Fetch(ctx, query)
```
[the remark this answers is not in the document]
stuartnelson3: `:facepalm:`

agentcfg/reporter.go
```diff
@@ -0,0 +1,102 @@
...
+	if err != nil {
+		return Result{}, err
+	}
+	// Only report configs when the query etag == current config etag, or
+	// when the agent indicates it has been applied.
+	if err == nil &&
+		(query.Etag == result.Source.Etag || query.MarkAsAppliedByAgent) {
```
axw, line 56:
```suggestion
	if query.Etag == result.Source.Etag || query.MarkAsAppliedByAgent {
```

(we know err == nil due to check above)

axw (approved):

Issue #4829, opened 2021-02-19
[fleet] Support java attacher settings via apm-server.yml 


jalvz:
Mainly: java agent version, and include/exclude rules.

felixbarny:
As a first step, this could be implemented as regular `apm-server.yaml` settings. In a second step, we could expose that part of the settings in the Fleet policy editor (depends on https://github.com/elastic/kibana/issues/89987).

jalvz:
To make it configurable it needs to be in the policy editor, users can't edit apm-server.yml by hand.

axw:
I think what @felixbarny meant (correct me if I'm wrong) is that we can do that later. Support it via `apm-server.yml` to start with would enable us to test the feature without waiting for the policy changes.

simitt:
Updated the title to reflect that his issue is only concerned with adding the include/exclude rules to the `apm-server.yml` and support them in the server.

Issue #4830, opened 2021-02-19
[meta] [fleet] Java agent auto attaching


jalvz:
The goal is to enable users to instrument their Java applications with zero configuration.

This requires a few steps:

- [x]  Bundle the Java agent attacher in apm-server (https://github.com/elastic/apm-server/issues/4824)
- [x] Add settings in the fleet policy editor for the attached
  - [x] https://github.com/elastic/observability-design/issues/77
  - [x] https://github.com/elastic/kibana/issues/108513
- [x] Support java attacher settings via apm-server.yml  (https://github.com/elastic/apm-server/issues/4829)
- [x] Support `--download-agent-version` option https://github.com/elastic/apm-server/issues/5853
- [x] ~Download the appropriate version of the Java agent.~ APM Server asks the Elastic Agent to download the Java Agent binaries in a specific version (the version from the apm-server.yml). Elastic Agent downloads the binaries and returns the path to the binaries to the APM Server. (https://github.com/elastic/apm-server/issues/5477)
- [x] APM Server runs the improved Java agent attacher (prefer `$JAVA_HOME/bin/java` over `java`), which takes care of attaching the binary to all matching JVMs on the same host  (https://github.com/elastic/apm-server/issues/5478)
- [x] Process attacher logs (as part of https://github.com/elastic/apm-server/issues/5478)
- [x] limit Java agent auto attachment to not be available when running in ECE/ESS environments (https://github.com/elastic/apm-server/issues/5479)

felixbarny:
Some additional steps:

- Handling of version updates
  - Download new version
  - Stop attacher
  - Start attacher with path to updated `--agent-jar`
  - Delete old agent jar?
- Handling of config changes (ex include/exclude rules)
  - Stop attacher
  - Restart with new config

simitt:
@felixbarny could you help us define more details for this so we can start working on it?

jalvz:
@simitt we should add this to https://github.com/elastic/apm-server/issues/5388, right?

simitt:
good catch @jalvz , added it.

jalvz:
@felixbarny how should we handle java logs in the end? 
will apm-server just log them as if it were their own? should we handle them separately and have extra settings for it (including rotation / etc)?

simitt:
[...]

PR #5483, opened 2021-06-17
add java attacher config

stuartnelson3:
## Motivation/summary

Adding basic include/exclude and config flags.

Based on [this comment](https://github.com/elastic/apm-server/issues/4830#issuecomment-863207642) I pulled out version as something to be static for a given build of apm-server, which we update when we bundle with a new version of the attacher.

closes https://github.com/elastic/apm-server/issues/4829

## Checklist

## How to test these changes

Update the `apm-server.yml` to include java attacher config fields, and verify they're parsed.

eg.
```yaml
  java_attacher:
    enabled: true
    continuous: true
    include_main:
      - "MyApplication"
      - "my-application.jar"
    exclude_user:
      - "root"
    include_vmargs:
      - "elastic.apm.agent.attach=true"
    config:
      - "service_name=my-cool-service"
      - "server_url=http://localhost:8200"
```


beater/config/java_attacher.go
```diff
@@ -0,0 +1,42 @@
...
+	IncludeAll bool     `config:"include_all"`
+	IncludePID []string `config:"include_pid"`
+
+	IncludeMain  []string `config:"include_main"`
+	IncludeVMArg []string `config:"include_vmarg"`
+	IncludeUser  []string `config:"include_user"`
+
+	ExcludeMain  []string `config:"exclude_main"`
+	ExcludeVMArg []string `config:"exclude_vmarg"`
+	ExcludeUser  []string `config:"exclude_user"`
```
simitt, line 35: How did you define these rules, are they speced out somewhere?

stuartnelson3: I looked at the docs online: https://www.elastic.co/guide/en/apm/agent/java/current/setup-attach-cli.html

I'm not 100% sure about what I included (or didn't include), just going off of the issue description that said 

>Mainly: java agent version, and include/exclude rules.

Since exactly what is happening still sounds a bit loosely defined, I decided to go with this for now and then continue from here with related issues

beater/config/java_attacher.go
```diff
@@ -0,0 +1,42 @@
...
+// specific language governing permissions and limitations
+// under the License.
+
+package config
+
+// JavaAttacherConfig holds configuration information for running a java
+// attacher jarfile.
+type JavaAttacherConfig struct {
+	Enabled    bool `config:"enabled"`
+	Continuous bool `config:"continuous"`
```
simitt, line 24: I'd rather expect that the apm-server would always set `continuous` when starting the attacher; but ultimately leave that to @felixbarny .

simitt (approved):
LGTM! Let's get @felixbarny involved to take a look at the config options.

beater/config/java_attacher.go
stuartnelson3: :+1: will see what he says

felixbarny (commented):
I suppose you're consciously just adding the settings and the impl of these settings comes in a follow up PR?

beater/config/java_attacher.go
```diff
@@ -0,0 +1,41 @@
...
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package config
+
+// JavaAttacherConfig holds configuration information for running a java
+// attacher jarfile.
+type JavaAttacherConfig struct {
```
felixbarny, line 22: I've almost overlooked it..

The config is sensitive to the ordering of the includes and excludes. See also https://www.elastic.co/guide/en/apm/agent/java/current/setup-attach-cli.html#setup-attach-cli-usage.

For example `--exclude-main ".*" --include-main opbeans` will exclude all JVMs, while `--include-main opbeans --exclude-main ".*"` includes only JVMs containing `opbeans` in their jar name.

felixbarny (changes requested):

beater/config/java_attacher.go
stuartnelson3: Given that it's order dependent, should we just accept an `args` string in the config instead of breaking it down into individual include/exclude flags? I'm not sure otherwise how else we would know how to order the flags for the attacher.

simitt: That sounds good to me.

felixbarny: It took me a while to find it again, but there's something similar in Fleet: https://github.com/elastic/beats/pull/23848

Applying the same principle here, the config could look like this:

```yaml
discovery-rules:
- rule: include # fleet uses allow/deny
  main: opbeans
- rule: include
  pid: 42
- rule: exclude
  main: ".*"
```

This would result in these arguments: `--include-main opbeans --include-pid 42 --exclude-main ".*"`

An alternative:

```yaml
discovery-rules:
- include-main: opbeans
- include-pid: 42
- exclude-main: ".*"
```

axw: I'd rather we have some structure to the options like this than accept arbitrary args. That would exclude any shenanigans around running the attacher in unexpected/unsupported ways by manually crafting the policy/config; the server should be strict about the ways in which it will run the attacher program.

felixbarny:
For `config`, instead of 

```yaml
    config:
      - "service_name=my-cool-service"
      - "server_url=http://localhost:8200"
```

Could we do something like this instead?

```yaml
    config:
      service_name: my-cool-service
      server_url: http://localhost:8200
```

beater/config/java_attacher.go
stuartnelson3: @axw / @simitt do you have a preference between @felixbarny's first and second suggestions? I'm leaning towards the second option, but am not opposed to option 1

axw: No strong opinion, either works for me and they will be hidden behind a layer of Kibana anyway - so whichever is simpler (probably 2?)

beater/config/java_attacher.go
```diff
@@ -0,0 +1,41 @@
...
+	Enabled bool `config:"enabled"`
+
+	IncludeAll bool     `config:"include_all"`
+	IncludePID []string `config:"include_pid"`
```
felixbarny, line 26:
```suggestion
	IncludePID []int `config:"include_pid"`
```

beater/config/java_attacher.go
```diff
@@ -0,0 +1,41 @@
...
+
+	IncludeMain  []string `config:"include_main"`
```
felixbarny, line 28: A note for the implementation: All the values will have to be escaped when passing them to the attacher to avoid injection attack vectors.

Example:
```
 - include_main: "foo; rm -rf /"
```

should not lead to the hard drive being purged

felixbarny (approved):

beater/config/java_attacher.go
stuartnelson3: :100:

beater/config/java_attacher.go
felixbarny: IIRC, there's some weirdness around yaml and number types so take that suggestion with a grain of salt and feel free to ignore it.

stuartnelson3:
jenkins run the tests

axw:
I'm removing this from the 7.14 test plan, as the feature can't be used yet. Let's test the config as part of the feature when we deliver it.

PR #5502, opened 2021-06-21
beater: even more refactoring

axw:
## Motivation/summary

- rate limiting middleware is now installed for both RUM and backend agent APIs, but only applies for anonymous clients (currently only RUM)
- rate limiting middleware now performs an initial Allow check at the request level, for consistent request rate limiting of those endpoints that are rate limited. Handlers only need to perform route-specific rate limiting, such as per-event rate limiting.
- agent config now restricts "insecure" (RUM) agents on the basis that they are anonymous, rather than being RUM specifically. The list of insecure agent names (those allowed for anonymous auth) is now passed in to support making this configurable later.

## How to test these changes

The only functional change is the response body for rate-limited RUM intake requests has changed slightly. This is not a problem in practice as the RUM agent does not parse the response.

## Related issues

Part of https://github.com/elastic/apm-server/issues/5347


beater/api/mux.go
```diff
@@ -219,46 +243,39 @@ func apmMiddleware(m map[request.ResultID]*monitoring.Int) []middleware.Middlewa
...
 		"If you are not using the RUM agent, you can safely ignore this error."
 	rumMiddleware := append(apmMiddleware(m),
 		middleware.ResponseHeadersMiddleware(cfg.ResponseHeaders),
 		middleware.ResponseHeadersMiddleware(cfg.RumConfig.ResponseHeaders),
-		middleware.SetRumFlagMiddleware(),
-		middleware.SetIPRateLimitMiddleware(cfg.RumConfig.EventRate),
 		middleware.CORSMiddleware(cfg.RumConfig.AllowOrigins, cfg.RumConfig.AllowHeaders),
 		middleware.AnonymousAuthorizationMiddleware(),
-		middleware.KillSwitchMiddleware(cfg.RumConfig.Enabled, msg),
+		middleware.AnonymousRateLimitMiddleware(ratelimitStore),
```
simitt, line 264: Isn't the order wrong now for Authorization and RateLimiting middleware? (the `AnonymousRateLimitMiddleware` checks `c.AuthResult.Anonymous` which will only be set afterwards)

axw: `middleware.Wrap` calls the Middleware functions in reverse order:

https://github.com/elastic/apm-server/blob/ab082c3f7e8d53f25cba6466797f596fe2af6d34/beater/middleware/middleware.go#L29

So `Wrap(handler, Auth, RateLimit)` leads to `Auth(RateLimit(handler))`

i.e. the first in the chain is outermost, last is innermost after the wrapped handler

simitt: right, I still get confused with the `KillSwitch` not being the first middleware in the chain; looks good then

simitt (approved):

beater/api/mux.go
axw: I found it pretty confusing too. Probably something to come back and clarify later.

PR #5506, opened 2021-06-21
Fix UBI source URL

jmlrt:
## Motivation/summary

This commit fix the source URL for UBI image to ensure that it stays
consistent with the URL generated in
https://artifacts.elastic.co/reports/dependencies/dependencies-current.html

## Related issues

None


axw (approved):
LGTM. Assuming this will need to go into 7.14, can you please label v7.14.0 to kick off an automated backport?

jmlrt:
Thanks @axw 👍🏻 
This should also eventually go into 7.13.3, will add labels.

axw:
@jmlrt perhaps not very intuitively, v7.13.0 is the label to use for backporting to the next 7.13.x release.

PR #5515, opened 2021-06-22
Introduce shared gRPC authorization interceptor

axw:
## Motivation/summary

Introduce a new gRPC authorization interceptor which is used for both OTLP and Jaeger. This pulls authorization up to the interceptor level, and out of individual methods (in the case of Jaeger). This will enable us to introduce another interceptor for rate limiting anonymous agents.

The new authorization interceptor dispatches to method-specific handlers. All OTLP methods use the "authorization" metadata approach, whereas Jaeger has different approaches depending on the method: extract process tags for PostSpans, and anonymous auth only for GetSamplingStrategy.

## How to test these changes

Non-functional change.

## Related issues


simitt (approved):
I really like how clean the handling with the interceptors is; great changes!

PR #5518, opened 2021-06-23
beater/api/intake: fix rate limiting

axw:
## Motivation/summary

Copy the batch processor in the closure before copying, to avoid wrapping it multiple times.

## How to test these changes

Enable RUM, send 200 RUM events over 10 requests - make sure this completes without triggering the rate limiter.

## Related issues

None, fixes bug introduced in https://github.com/elastic/apm-server/pull/5492


simitt (approved):

stuartnelson3:
Confirmed with BC2, did not receive a 429 response

Issue #5488, opened 2021-06-17
[tech debt] Move python tests for sourcemaps to go systemtests


simitt:
Rewrite python integration tests related to sourcemaps to go systemtests for apm-server integration

axw:
Closed by https://github.com/elastic/apm-server/pull/5522

PR #5522, opened 2021-06-23
Copy sourcemap Python tests to Go

jalvz:
Tests not moved over:

`SourcemappingDisabledIntegrationTest`, `SourcemapInvalidESConfig`, `SourcemapESConfigUser``SourcemapESConfigAPIKey`

I think is a bit overkill to run the whole stack just to check that configuration has been applied correctly, that should be covered in unit tests. 

Closes #https://github.com/elastic/apm-server/issues/5488


systemtest/sourcemap_test.go
```diff
@@ -0,0 +1,349 @@
...
+	systemtest.SendRUMEventsPayload(t, srv, "../testdata/intake-v2/errors_rum.ndjson")
+	systemtest.Elasticsearch.ExpectDocs(t, "apm-*-error", nil)
+	expectSourcemapUpdated(t)
+
+	// delete sourcemap index
+	deleteIndex(t, "apm-*-sourcemap*")
+
+	// index error document again
+	systemtest.SendRUMEventsPayload(t, srv, "../testdata/intake-v2/errors_rum.ndjson")
+	systemtest.Elasticsearch.ExpectDocs(t, "apm-*-error", nil)
```
axw, line 201: we should either delete the old error, or use ExectMinDocs with >1 here, otherwise the check would be satisfied by the error indexed from the first SendRUMEventsPayload call won't it?

axw (commented):
Looks good, thanks! Just a few minor comments

systemtest/sourcemap_test.go
jalvz: right, missed that. thanks

systemtest/sourcemap_test.go
```diff
@@ -0,0 +1,344 @@
...
+	systemtest.SendRUMEventsPayload(t, srv, "../testdata/intake-v2/errors_rum.ndjson")
+	result := systemtest.Elasticsearch.ExpectDocs(t, "apm-*-error", nil)
+	expectSourcemapUpdated(t, result)
+
+	// delete sourcemap index
+	deleteIndex(t, "apm-*-sourcemap*")
+
+	// index error document again
+	systemtest.SendRUMEventsPayload(t, srv, "../testdata/intake-v2/errors_rum.ndjson")
+	result = systemtest.Elasticsearch.ExpectMinDocs(t, 1, "apm-*-error", nil)
```
axw, line 200:
```suggestion
	result = systemtest.Elasticsearch.ExpectMinDocs(t, 2, "apm-*-error", nil)
```

systemtest/sourcemap_test.go
```diff
@@ -0,0 +1,344 @@
...
+	resp, err = systemtest.Elasticsearch.Indices.Flush()
+	require.NoError(t, err)
+	resp.Body.Close()
+}
+
+func expectSourcemapUpdated(t *testing.T, result estest.SearchResult) {
+	for _, hit := range result.Hits.Hits {
+		source := Source{}
+		data, _ := hit.RawSource.MarshalJSON()
+		err := json.Unmarshal(data, &source)
```
axw, line 306:
```suggestion
		err := hit.UnmarshalSource(&source)
```

systemtest/sourcemap_test.go
```diff
@@ -0,0 +1,344 @@
...
+	require.NoError(t, err)
+	resp.Body.Close()
+	resp, err = systemtest.Elasticsearch.Indices.Flush()
+	require.NoError(t, err)
+	resp.Body.Close()
+}
+
+func expectSourcemapUpdated(t *testing.T, result estest.SearchResult) {
+	for _, hit := range result.Hits.Hits {
+		source := Source{}
```
axw, line 304:
```suggestion
		var source struct {
			Error Error
		}
```

Then we can get rid of the Source type, which has too general a name.

systemtest/sourcemap_test.go
```diff
@@ -0,0 +1,344 @@
...
+
+		for _, stacktrace := range source.Error.Log.Stacktrace {
+			assert.True(t, stacktrace.Sourcemap.Updated)
+		}
+	}
+}
+
+type Source struct {
+	Error Error
+}
```
axw, line 323:
```suggestion
```

axw (approved):

systemtest/sourcemap_test.go
jalvz: 🤦

PR #5525, opened 2021-06-23
Introduce rate limit gRPC interceptor

axw:
## Motivation/summary

Introduce a gRPC interceptor for rate limiting anonymous clients. It is installed in the OTLP and muxed Jaeger services, but will not be effective until we have support for anonymous clients other than RUM. Rate limiting has not been added to the deprecated Jaeger server.

The per-event rate limiting that was in the intake handler has been moved up a level, and is shared by the intake handler as well as the muxed gRPC server. Now thate the ratelimit package is used by gRPC, I have moved the ratelimit package out of beater/api.

I have removed the fine-grained rate limiting unit tests, and replaced them with a slightly extended system test and unit test for the new shared processor.

## How to test these changes

Non-functional change.

## Related issues

#5347


simitt (approved):
LGTM! Thanks for all the clean step-by-step PRs - these make the review process straight forward.

PR #5539, opened 2021-06-24
kibana expects the apm-config as a json string

stuartnelson3:
## Motivation/summary

The expected format in kibana was clarified to me by @ogupte 

This might change in the future, in which case
this commit can be reverted to the previous
version that sent a serialized json blob

### DIFF
--- a/kibana/send_config.go
+++ b/kibana/send_config.go
@@ -44,7 +44,11 @@ func SendConfig(ctx context.Context, client Client, conf *ucfg.Config) error {
 		return err
 	}
 
-	b, err := json.Marshal(format(flat))
+	formatted, err := format(flat)
+	if err != nil {
+		return err
+	}
+	b, err := json.Marshal(formatted)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


axw (approved):

axw:
We might want to write a system test for this at some stage, to test the apm-server and kibana code together.

stuartnelson3:
Definitely. Since it still is in flux (@ogupte said he's going to see if he can accept arbitrary json instead of a string, like now) I think we can hold off, but once we settle on something :+1:

stuartnelson3:
Was informed that now arbitrary json is accepted, so this is unnecessary. The following should work. I'll open a new PR to change the key from `schemaJson` to `schema`.

```
curl --request POST \
  --url http://localhost:5603/api/apm/fleet/apm_server_schema \
  --header 'Authorization: Basic YWRtaW46Y2hhbmdlbWU=' \
  --header 'Content-Type: application/json' \
  --header 'kbn-xsrf: true' \
  --data '{
	"schema": {
		"apm-server.host": "0.0.0.0:8200",
		"apm-server.rum.enabled": true,
		"output.ssl.enabled": true
	}
}'
```

PR #5540, opened 2021-06-24
Java attacher command

### LINKED ISSUE #5478: [fleet] Run the Java agent attacher to attach binaries to all matching JVMs
APM Server runs the improved Java agent attacher (prefer `$JAVA_HOME/bin/java` over `java`), which takes care of attaching the binary to all matching JVMs on the same host 

Take care of processing attacher logs.

part of https://github.com/elastic/apm-server/issues/4830

### LINKED ISSUE #5479: [fleet] Only support java agent attachment when not running in cloud env
Limit Java agent auto attachment to not be available when running in ECE/ESS environments. 

part of https://github.com/elastic/apm-server/issues/4830

stuartnelson3:
## Motivation/summary

Run the java attacher when not in a cloud environment, and log its output.

closes #5478

## Checklist


stuartnelson3:
I'm encountering weird issues running `fork/exec` from within apm-server. Regardless of the command I'm running, I receive an error:
`fork/exec /usr/bin/java: operation not permitted`

I've confirmed that `echo`, in place of `java`, also has the same issue. Running the same command in a stand-along go program on my machine does not encounter this problem.

stuartnelson3:
example log output (this is a command that just printf's on a timer):

```
{"log.level":"info","@timestamp":"2021-06-24T16:31:03.639+0200","log.logger":"java-attacher","log.origin":{"file.name":"java_attacher/java_attacher.go","file.line":90},"message":"got time 2021-06-24 16:31:03.638571917 +0200 CEST m=+1.000189534","service.name":"apm-server","event.dataset":"apm-server","ecs.version":"1.6.0"}
{"log.level":"info","@timestamp":"2021-06-24T16:31:04.638+0200","log.logger":"java-attacher","log.origin":{"file.name":"java_attacher/java_attacher.go","file.line":90},"message":"got time 2021-06-24 16:31:04.638761077 +0200 CEST m=+2.000378708","service.name":"apm-server","event.dataset":"apm-server","ecs.version":"1.6.0"}
{"log.level":"info","@timestamp":"2021-06-24T16:31:05.639+0200","log.logger":"java-attacher","log.origin":{"file.name":"java_attacher/java_attacher.go","file.line":90},"message":"got time 2021-06-24 16:31:05.638927984 +0200 CEST m=+3.000545608","service.name":"apm-server","event.dataset":"apm-server","ecs.version":"1.6.0"}
{"log.level":"info","@timestamp":"2021-06-24T16:31:06.639+0200","log.logger":"java-attacher","log.origin":{"file.name":"java_attacher/java_attacher.go","file.line":90},"message":"got time 2021-06-24 16:31:06.639071653 +0200 CEST m=+4.000689273","service.name":"apm-server","event.dataset":"apm-server","ecs.version":"1.6.0"}
```

beater/beater.go
```diff
@@ -394,6 +395,19 @@ func (s *serverRunner) run() error {
 	s.acker.Open()
 	pipeline := pipetool.WithACKer(s.pipeline, s.acker)
 
+	if eac := os.Getenv("ELASTIC_AGENT_CLOUD"); eac == "" && s.config.JavaAttacherConfig.Enabled {
+		// Not running on ECE/ESS, start the java attacher
+		go func() {
+			attacher, err := javaattacher.New(s.config.JavaAttacherConfig)
+			if err != nil {
+				s.logger.Errorf("java attacher error: %v", err)
```
jalvz, line 403: shouldn't be returning here or something?

beater/java_attacher/java_attacher.go
```diff
@@ -0,0 +1,120 @@
...
+	cfg    config.JavaAttacherConfig
+	logger *logp.Logger
+}
+
+func New(cfg config.JavaAttacherConfig) (JavaAttacher, error) {
+	if cfg.JavaBin == "" {
+		if jh := os.Getenv("JAVA_HOME"); jh != "" {
+			cfg.JavaBin = filepath.Join(jh, "/bin/java")
+		} else {
+			bin, err := exec.Command("which", "java").Output()
```
jalvz, line 44: does this work on windows? (should it work on windows?)

beater/java_attacher/java_attacher.go
```diff
@@ -0,0 +1,120 @@
...
+		return err
+	}
+
+	if err := cmd.Start(); err != nil {
+		return err
+	}
+
+	// TODO: How do we want to handle logging?
+	// See comment:
+	// https://github.com/elastic/apm-server/issues/4830#issuecomment-863207642
```
jalvz, line 73: something to keep in mind is that we need to ensure that attacher logs are ECS compatible, because we can't mix ECS and non ECS when ECS is enabled (oh well, how many times can i say "ECS" in sentence?)

stuartnelson3: hm, maybe @felixbarny knows?

beater/java_attacher/java_attacher.go
stuartnelson3: I actually have no idea. I guess it probably wouldn't work on windows. How do we handle windows support in apm server?

jalvz: we run automated tests in windows, that's pretty much all. but this `which` is going to be problematic i think. Maybe is ok for now, and just document it.

beater/beater.go
stuartnelson3: yep, whoops :sweat_smile:

beater/java_attacher/java_attacher.go
felixbarny: The attacher logs are ECS compatible.

beater/java_attacher/java_attacher.go
felixbarny: Can we just use plain `java` if `JAVA_HOME` is not specified? That would solve for the case when `java` is within the path of executables.

axw: I assume @stuartnelson3 did this for logging, which could be useful when debugging. As an alternative to which java, there's https://golang.org/pkg/os/exec/#LookPath

beater/java_attacher/java_attacher.go
```diff
@@ -0,0 +1,122 @@
...
+			select {
+			case <-ctx.Done():
+				return
+			case <-donec:
+				// TODO: Is this necessary, or will the scanner
+				// finish when the command exits?
+				return
+			default:
+			}
+			j.logger.Info(scanner.Text())
```
felixbarny, line 90: This might log ECS JSON within ECS JSON.

beater/java_attacher/java_attacher.go
```diff
@@ -0,0 +1,122 @@
...
+			j.logger.Errorf("error scanning attacher logs: %v", err)
+		}
+	}()
+
+	return cmd.Wait()
+}
+
+func (j JavaAttacher) build(ctx context.Context) *exec.Cmd {
+	args := append([]string{"-jar", javaAttacher}, j.formatArgs()...)
+	return exec.CommandContext(ctx, j.cfg.JavaBin, args...)
```
felixbarny, line 102: Does Go make sure to escape the args or do we have to do that manually?
Either way, I think it's worth adding a test case for that.

stuartnelson3: yeah we need to escape the input args still, thanks for the reminder

beater/java_attacher/java_attacher.go
stuartnelson3: @simitt @axw is there a way to write unformatted output via `*logp.Logger`? Just writing to stdout via `fmt` isn't supported under fleet/elastic-agent, if i remember correctly

beater/java_attacher/java_attacher.go
axw: do we? exec.Command doesn't run through a shell, it calls the exec syscalls directly - so the args should be passed through to java exactly as specified

beater/java_attacher/java_attacher.go
axw: I'm not aware of a way to do that. I think we might need to parse and reformat :/
Maybe @simitt has a better idea.

felixbarny: That may not be so bad after all and probably better than logging into a separate file.

beater/java_attacher/java_attacher.go
stuartnelson3: I thought I had figured some way around it but as long as java is the binary invoked (which we control) it appears we'll be fine

beater/config/java_attacher.go
```diff
@@ -23,8 +23,9 @@ import "fmt"
 // attacher jarfile.
 type JavaAttacherConfig struct {
 	Enabled        bool                `config:"enabled"`
-	DiscoveryRules []map[string]string `config:"discovery_rules"`
-	Config         map[string]string   `config:"config"`
+	DiscoveryRules []map[string]string `config:"discovery-rules"`
+	Config         []string            `config:"config"`
```
felixbarny, line 27: Why did you change from `map[string]string` to `[]string`?
I thought we settled on this syntax in the yaml file:

```yaml
  config
    foo: bar
    baz: qux
```

That seems nicer than 
```yaml
  config
    - "foo=bar"
    - "baz=qux"
```

stuartnelson3: you're probably correct, as that's what was there. I'm guessing i saw the example test i had from before and got confused, and backed out the wrong change

beater/java_attacher/java_attacher.go
```diff
@@ -0,0 +1,125 @@
...
+			case <-ctx.Done():
+				return
+			case <-donec:
+				// TODO: Is this necessary, or will the scanner
+				// finish when the command exits?
+				return
+			default:
+			}
+			json.Unmarshal(scanner.Bytes(), &b)
+			j.logger.Info(b.Message)
```
felixbarny, line 93: You'll also probably want to switch on the log level and call `j.logger.*` accordingly.

Not sure how the JSON parser handles dotted/nested fields, but the logs would contain the dotted notation, for example `"log.level": "INFO"`
```go
		b := struct {
			LogLevel string `json:"log.level"`
			Message  string `json:"message"`
		}{}
```

These are the possible values: `FATAL, ERROR, WARN, INFO, DEBUG, TRACE`

stuartnelson3: Officially, apm-server has debug, warn, info, and error levels. We'll not call a fatal level error from within apm-server because the attacher fails, and trace isn't supported. Any preference for how to indicate these in the apm-server logs?

felixbarny: I'd map `FATAL` and `ERROR` to `logger.Error`, and `DEBUG` and `TRACE` to `logger.Debug`. In practice, I think we currently only make use of the  `ERROR`, `INFO` and `DEBUG` levels.

beater/java_attacher/java_attacher.go
```diff
@@ -0,0 +1,141 @@
...
+	return cmd.Wait()
+}
+
+func (j JavaAttacher) build(ctx context.Context) *exec.Cmd {
+	args := append([]string{"-jar", javaAttacher}, j.formatArgs()...)
+	return exec.CommandContext(ctx, j.cfg.JavaBin, args...)
+}
+
+func (j JavaAttacher) formatArgs() []string {
+	args := []string{"--continuous"}
```
felixbarny, line 125: If you can query the log level of APM Server, you could set `--log-level` to one of `<off|fatal|error|warn|info|debug|trace|all>`.

beater/config/java_attacher.go
```diff
@@ -23,8 +23,9 @@ import "fmt"
 // attacher jarfile.
 type JavaAttacherConfig struct {
 	Enabled        bool                `config:"enabled"`
-	DiscoveryRules []map[string]string `config:"discovery_rules"`
+	DiscoveryRules []map[string]string `config:"discovery-rules"`
```
felixbarny, line 26: After giving it another thought, I think we should remove the include `include-pid` and `exclude-pid` rules. I don't think they make sense in the context of globally specifying include/exclude rules. That's because PIDs are not predictable before a process starts and even more so across different hosts.

stuartnelson3: makes sense :ok_hand:

beater/java_attacher/java_attacher.go
stuartnelson3: @axw is there an easy way to do this within the code?

axw: I'm not aware of a way to do that. It may be necessary to update libbeat/logp to expose the level.

felixbarny: If the log level can't be directly read from the logger itself, is it possible to read the `logging.level` config option here?

beater/beater.go
```diff
@@ -394,6 +396,31 @@ func (s *serverRunner) run() error {
 	s.acker.Open()
 	pipeline := pipetool.WithACKer(s.pipeline, s.acker)
 
+	if eac := os.Getenv("ELASTIC_AGENT_CLOUD"); eac == "" && s.config.JavaAttacherConfig.Enabled {
```
stuartnelson3, line 399: @axw @simitt should I disable or delete this block of code so that we can merge the rest of the PR? there are still questions about other parts of the implementation and it's lower priority, so I could see it being a while before someone gets back to this.

beater/java_attacher/java_attacher.go
axw: Perhaps. It's also possible to enable debug level for specific "selectors". For example we could introduce a "java-attacher" selector, and you could enable debug logging for just that by passing `-d java-attacher` to the apm-server command.

What we can do with the existing libbeat/logp API is to check if debug logging for the selector (or the process overall) is enabled: https://github.com/elastic/beats/blob/c3925fcb5435df6397803542492398da3e8ed484/libbeat/logp/logger.go#L111-L112

I think the pragmatic solution here is to check that and pass `--log-level=debug` if enabled, otherwise pass `--log-level=info`. We'll still parse the log records and call the appropriate logger method (Infof, Warnf, Errorf, etc.), so if the server is configured to only log warnings then the info logs will just get dropped after being parsed.

beater/beater.go
axw: Yes let's remove this wiring-up bit until the rest is complete please. (FWIW I'm not a fan of the ELASTIC_AGENT_CLOUD check. We should probably just rely on the fact that seccomp will be enabled in ESS/ECE for security, and on the integration editor UI disallowing the Java attacher options.)

beater/java_attacher/java_attacher.go
stuartnelson3: It's possible to get the `logging.level` setting from the parent config, using the same code from sending the config to kibana:

```go
	cfg := ucfg.Config(*s.rawConfig)
	parentCfg := cfg.Parent()
```

And from there accessing `logging.level`. If it's empty, then we use the default `info` level.

beater/java_attacher/java_attacher.go
```diff
@@ -0,0 +1,155 @@
...
+
+	return append(args, cfg...)
+}
+
+func makeArg(flagName string, args ...string) string {
+	return flagName + " " + strings.Join(args, " ")
+}
+
+func setLogfn(l *logp.Logger, level string) logfn {
+	switch level {
```
felixbarny, line 143: The attacher emits the `log.level` in upper-case
```suggestion
	switch strings.ToLower(level) {
```

stuartnelson3: This value is coming from the (lowercase) settings in apm-server.yml, I'm using that value for the `--log-level` flag when starting the attacher

axw: We should be varying the log function based on the log level emitted by the attacher though. We might pass `--log-level=debug` to the attacher, but that doesn't mean every log entry is debug level.

beater/java_attacher/java_attacher.go
```diff
@@ -0,0 +1,155 @@
...
+		return err
+	}
+
+	donec := make(chan struct{})
+	defer close(donec)
+	go func() {
+		scanner := bufio.NewScanner(stdout)
+		b := struct {
+			Message string `json:"message"`
+		}{}
```
axw, line 92:
```suggestion
		b := struct {
			Message string `json:"message"`
			LogLevel string `json:"log.level"`
		}{}
```

axw: I think we should probably also parse the timestamp and any additional fields. Perhaps another PR.

axw (commented):
Mostly LGTM, but the logging isn't quite right. I think we should also have a unit test for the log proxying - not necessarily in this PR, but before we wire this code up. One way we could do that is to have a [TestMain](https://golang.org/pkg/testing/#hdr-Main) which acts as a fake attacher when run with some environment variable, setting JavaBin to `os.Args[0]` in the test.

beater/java_attacher/java_attacher.go
felixbarny: In fact, we could always pass `--log-level=debug` and then let the server's logger decide whether or not to log the actual level. In theory, it could slow down the attacher a bit by logging and parsing the logs of fine-grained levels unnecessarily. But I don't expect that to be a problem in practice. It may make things a bit easier. Just an idea :)

stuartnelson3: How about we go with `--log-level=debug` for now, and log the lines accordingly as they come. when we actually hook it up we can see if this is an issue or not?

stuartnelson3:
>Mostly LGTM, but the logging isn't quite right. I think we should also have a unit test for the log proxying - not necessarily in this PR, but before we wire this code up. One way we could do that is to have a TestMain which acts as a fake attacher when run with some environment variable, setting JavaBin to os.Args[0] in the test.

sounds good. I'd prefer doing it in another PR so that this can be merged and we can focus on the current release

axw (approved):
LGTM!

PR #5545, opened 2021-06-28
Separate authentication from authorization

axw:
## Motivation/summary

Overhaul beater/authorization (now beater/auth) to provide two primary interfaces: Authenticator and Authorizer.

Authenticator is responsible for authenticating a client (typically but not always an agent), returning authentication details (e.g. username for API Key) and an Authorizer for authorizing specific actions.

The auth middleware and interceptor authenticate and store the returned Authorizer in request context, so it can be used by specific method handlers. Authorize accepts an auth.Action and auth.Resource to authorize. When we add anonymous agent auth, the anonymous Authorizer implementation can check the action and use it to decide which properties of the resource to check (i.e. only service name), and which actions to even allow (i.e. disallow source map upload).

HTTP and gRPC status codes are changed to follow the semantics more correctly. Failure to authenticate leads to 401 (HTTP) and Unauthenticated (gRPC). Failure to authorize leads to 403 (HTTP) and PermissionDenied (gRPC). This can only happen when an agent authenticates using an API Key, but the API Key is not authorized for a particular action.

RUM rate limiting now only applies when either secret token or API Key auth is enabled. Otherwise there's no point, backend agents are (currently) never rate limited.

## Checklist

## How to test these changes

Mostly non-functional changes. Check auth works as before with the exception of HTTP/gRPC status codes.

## Related issues


beater/auth/authenticator.go
```diff
@@ -0,0 +1,193 @@
...
+
+// NewAuthenticator creates an Authenticator with config, authenticating
+// clients with one of the allowed methods.
+func NewAuthenticator(cfg config.AgentAuth) (*Authenticator, error) {
+	b := Authenticator{secretToken: cfg.SecretToken}
+	if cfg.APIKey.Enabled {
+		// do not use username+password for API Key requests
+		cfg.APIKey.ESConfig.Username = ""
+		cfg.APIKey.ESConfig.Password = ""
+		cfg.APIKey.ESConfig.APIKey = ""
```
stuartnelson3, line 146: Should apikey be set to `""` if it's being used by the client?

beater/auth/authenticator.go
```diff
@@ -0,0 +1,193 @@
...
+			return nil, err
+		}
+
+		cache := newPrivilegesCache(cacheTimeoutMinute, cfg.APIKey.LimitPerMin)
+		b.apikey = newApikeyAuth(client, cache)
+	}
+	return &b, nil
+}
+
+// Authenticate authenticates a client given an authentication mtehod and token,
```
stuartnelson3, line 158: `/s/mtehod/method/`

beater/auth/authenticator.go
```diff
@@ -0,0 +1,193 @@
...
+// returning the authentication details and an Authorizer for authorizing specific
+// actions and resources.
+//
+// Authenticate will return ErrAuthFailed (possibly wrapped) if at least one auth
+// method is configured and the, and no valid credentials have been supplied.
```
stuartnelson3, line 163: an errant `and the` from a previous version of the comment?

stuartnelson3 (commented):
read-through with a few comments

beater/auth/apikey.go
```diff
@@ -0,0 +1,202 @@
...
+
+func (c *privilegesCache) get(id string) (*es.HasPrivilegesResponse, bool) {
+	if val, exists := c.cache.Get(id); exists {
+		return val.(*es.HasPrivilegesResponse), true
+	}
+	return nil, false
+}
+
+func (c *privilegesCache) add(id string, privileges *es.HasPrivilegesResponse) {
+	c.cache.SetDefault(id, privileges)
```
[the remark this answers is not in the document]
axw: I'd like to consolidate the caching techniques we use, and probably use a fixed size LRU here. I think we should probably also rate limit auth attempts. That's for another day though -- this is existing code, I just moved it from `beater/authorization/privilege.go`.

beater/auth/authenticator.go
axw: This is also existing code, I guess your question comes from the mismatch with the comment above? I'll clarify the comment. The point is to not use any auth defined by apm-server config, and only rely on the API Key provided to the Authenticate method.

beater/auth/authenticator.go
axw: astute :)

beater/auth/anonymous.go
```diff
@@ -15,19 +15,16 @@
...
 	"context"
-	"testing"
-
-	"github.com/stretchr/testify/assert"
 )
 
-func TestDenyAuth(t *testing.T) {
-	handler := denyAuth{}
+// AnonymousAuth implements the Authorizer interface.
+type AnonymousAuth struct{}
```
jalvz, line 25: this is not used right now, correct?

jalvz (commented):
An idea for next time to make it smaller might be to separate functional changes from refactor from moved code... It is a bit hard to understand what was copy-pasted and what is new code

beater/api/asset/sourcemap/handler_test.go
```diff
@@ -89,6 +90,20 @@ func TestAssetHandler(t *testing.T) {
...
 			code: http.StatusAccepted,
 		},
+		"unauthorized": {
+			authorizer: func(context.Context, auth.Action, auth.Resource) error {
+				return auth.ErrUnauthorized
+			},
+			code: http.StatusForbidden,
+			body: beatertest.ResultErrWrap("unauthorized"),
+		},
+		"auth_unavailable": {
```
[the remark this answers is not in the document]
axw: The point of that test is to point out that errors returned by Authenticate other than ErrAuthFailed do not result in 401. i.e. it's not the client's fault, there's something wrong with the server/backend.

beater/auth/anonymous.go
axw: It's used in the auth middleware when auth isn't required, and in the Jaeger method authenticator for agent config.

axw:
> An idea for next time to make it smaller might be to separate functional changes from refactor from moved code... It is a bit hard to understand what was copy-pasted and what is new code

Yep fair comment. I had figured there wasn't that much code that was just moved to warrant that, but evidently there was enough that it could have helped to separate. Thanks!

simitt (approved):
This separation is another great improvement.

stuartnelson3:
Confirmed a 401 for HTTP and code = Unauthenticated for gRPC with BC2, request was successful when the correct auth was supplied

PR #5551, opened 2021-06-28
[cloud] Set config settings for continuous cloud sizes

simitt:
## Motivation/summary

Do not restrict cloud settings to predefined sizes to support potential sizing changes in the future (for example allow larger deployments than 8GB). 
Pleateau on number of workers at `20` (aligned with current upper limit of cloud apm deployments of 8GB). 

## How to test these changes
* Start withouth `CLOUD_APM_CAPACITY`: should not set any default values for `output` and `queue` 
* Start withouth `CLOUD_APM_CAPACITY` set to one of the previously defined values (`512`, `1024`, `2048`, `4096`, `8192`) should set default values for `output` and `queue` to current hardcoded values
* Start withouth `CLOUD_APM_CAPACITY=xyz` should set default values for `output` and `queue` according to the implemented formula
## Related issues

related to #5372
follow up on https://github.com/elastic/apm-server/pull/5402


cmd/root.go
```diff
@@ -41,65 +43,32 @@ const (
...
-						"mem": map[string]interface{}{
-							"events": cloudMatrix[cap].events,
-							"flush": map[string]interface{}{
-								"min_events": cloudMatrix[cap].minEvents,
-							},
-						},
-					},
-				})
+				m := map[string]interface{}{}
+				cloudValues(m)
```
jalvz, line 68: why `cloudValues` takes as argument an empty map?

jalvz (approved):

axw:
@Mergifyio update

stuartnelson3:
Confirmed locally with BC2 using `apm-server export config`, in cloud before migration (via the data streams migration warning modal which includes `queue.mem.events`). I couldn't complete the migration on QA, with the warning "apm@0.3.0-dev.1 not found", which would be the package that was deleted from SNAPSHOT recently.

@ogupte the next kibana build candidate should be using apmpackage 0.3.0, correct?

axw:
@stuartnelson3 make sure you use the GCP region on QA cloud - sounds like the images still haven't been updated for the AWS region.

stuartnelson3:
confirmed on cloud:

```
GET .monitoring-beats-7-2021.07.19/_search
{
"_source": ["beats_state.state.output.*", "beats_stats.metrics.libbeat.pipeline.queue.*"]
}
```

PR #5559, opened 2021-06-29
systemtest: disable profiling tests

axw:
## Motivation/summary

Disable the TestAPMServerProfiling system test, which intermittently times out. This is an experimental feature; we should fix the test if/when we proceed with the feature.

## Related issues

Closes https://github.com/elastic/apm-server/issues/5322


simitt (approved):

axw:
unrelated test failure...

PR #5565, opened 2021-06-29
agentcfg: fix flaky test

axw:
## Motivation/summary

TestReportFetch is intermittently failing due to multiple applied agent config metrics batches being processed concurrently, leading to the order being non-deterministic. Check the etag values without requiring them to be in a specific order.

While we're in the area, make sure goroutines have exited before Run returns.

## How to test these changes

Non-functional change. Reproduce with `go test -v -failfast -count=100 -run TestReportFetch`

## Related issues

None

### DIFF
--- a/agentcfg/reporter.go
+++ b/agentcfg/reporter.go
@@ -19,6 +19,7 @@ package agentcfg
 
 import (
 	"context"
+	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


simitt (approved):

axw:
unrelated test failure

Issue #5493, opened 2021-06-18
pubsub: Pubsub offers no Close method for closing BulkIndexer


axw:
The tail-based sampling pubsub implementation internally uses a go-elasticsearch BulkIndexer for indexing sampling decisions. This is never closed, as there is no method for closing the pubsub:

https://github.com/elastic/apm-server/blob/7b86cb707cf3456dcdf53da5db9798c721ebc7cc/x-pack/apm-server/sampling/pubsub/pubsub.go#L54-L60

Issue #5480, opened 2021-06-17
Data race in pub/sub implementation


jalvz:
Error:
`http: superfluous response.WriteHeader call from github.com/elastic/apm-server/x-pack/apm-server/sampling/pubsub_test.newRequestResponseWriterServer.func1 (pubsub_test.go:400)`

Details:
https://apm-ci.elastic.co/job/apm-server/job/apm-server-mbp/job/PR-5468/5/testReport/junit/(root)/pubsub/Build_and_Test___ARM_build_test___TestSubscribeSampledTraceIDs/

axw:
I looked into this a bit, and I'm _pretty_ sure (but not 100% sure) that it's an issue in the test rather than the implementation. However, while testing I did find another issue in the implementation: https://github.com/elastic/apm-server/issues/5493

PR #5568, opened 2021-06-29
sampling/pubsub: minor fixes

axw:
## Motivation/summary

Revise the pubsub API so that the PublishSampledTraceIDs method is long-lived and accepts a channel of trace IDs. This brings PublishSampledTraceIDs closer to the API of SubscribeSampledTraceIDs, and ensures we properly close the BulkIndexer used by the publisher when it exits.

Also, fix a race in the pubsub tests. We now write HTTP response from the handler goroutine rather than in another goroutine which waits for the request. This avoids a race in the existing test code, where the client receives the response and closes the connection, causing the server's handler context to be cancelled before the handler returns.

## Related issues

Closes https://github.com/elastic/apm-server/issues/5493
Closes https://github.com/elastic/apm-server/issues/5480

### DIFF
--- a/x-pack/apm-server/sampling/processor.go
+++ b/x-pack/apm-server/sampling/processor.go
@@ -367,8 +367,9 @@ func (p *Processor) Run() error {
 
 	remoteSampledTraceIDs := make(chan string)
 	localSampledTraceIDs := make(chan string)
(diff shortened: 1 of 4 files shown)


simitt (approved):
Changes look good.

PR #5599, opened 2021-07-05
Update go elasticsearch

stuartnelson3:
Opening this PR so others can look at issues updating `go-elasticsearch` v7.

The failure in `TestBackoffRetries` appears to be `go-elasticsearch` client correctly respecting the number of retries now (total expected requests to an endpoint == initial failing request + num retries).

It appears the pubsub test is now receiving an unexpected `POST` request to `testing.invalid/index_name/_search` in x-pack/apm-server/sampling/pubsub/pubsubtest/client.go. Previously this was a `GET`. I'm leaving this here in case someone has context right now, otherwise I'll look into this change when I'm online tomorrow.


go.mod
```diff
@@ -19,8 +19,8 @@ require (
 	github.com/elastic/elastic-agent-client/v7 v7.0.0-20210407144825-cc1c33cfa1d0 // indirect
 	github.com/elastic/elastic-package v0.0.0-20210310173719-3b8f76516ae3
 	github.com/elastic/gmux v0.1.0
-	github.com/elastic/go-elasticsearch/v7 v7.11.0
-	github.com/elastic/go-elasticsearch/v8 v8.0.0-20201007143536-4b4020669208
+	github.com/elastic/go-elasticsearch/v7 v7.13.1
```
axw, line 22: sorry I was a bit unclear earlier, we should be using the 7.14 branch of go-elasticsearch

`go get github.com/elastic/go-elasticsearch/v7@7.14` should do it

axw (commented):
Not sure about the unexpected POST -- I guess related to https://github.com/elastic/go-elasticsearch/pull/267? I think we'll also need to adjust pubsubtest to handle the new Elasticsearch version checks.

go.mod
stuartnelson3: I'm caught in some weird dependency loop, I think because go is seeing `7.14` as being tagged with version `v7.5`:

```
# go get github.com/elastic/go-elasticsearch/v7@7.14
go get: removed github.com/elastic/elastic-package v0.0.0-20210310173719-3b8f76516ae3
go get: downgraded github.com/elastic/go-elasticsearch/v7 v7.9.0 => v7.5.1-0.20210701130713-3c08ee738285

# make update
// elided ..
no required module provides package github.com/elastic/elastic-package; to add it:
        go get github.com/elastic/elastic-package

# go get github.com/elastic/elastic-package@3b8f76516ae3
go get: added github.com/elastic/elastic-package v0.0.0-20210310173719-3b8f76516ae3
go get: upgraded github.com/elastic/go-elasticsearch/v7 v7.5.1-0.20210701130713-3c08ee738285 => v7.9.0
```

When I try to upgrade to the 7.14 branch, it removes `elastic-package`, but when I re-add it, go mod "upgrades" to `v7.9.0`..

@Anaethelion do you have any ideas?

simitt: There's a version conflict: `elastic-package` requires [go-elasticsearch `v.7.9.0`](https://github.com/elastic/elastic-package/blob/master/go.mod#L10); but the `go-elasticsearch 7.14` branch is resolved into `v7.5.1` (with the latest commit, but still the wrong version). The dependency has been resolved similar in the past, but it becomes an issue now that `elastic-package` is a dependency. 

Here comes my wild guess: 
Up until `7.5` the ES client releases were tagged from the `7.x` branch, starting with `7.6` dedicated version branches are created. This means that up until the `7.14` release, this branch (created from `7.x`) is still associated with the `7.5.x` tag. Does that sound reasonable?

Are there any major changes in the `7.14` branch that we need to get into the APM Server for the features that we are using? If not - I suggest to update to the released `7.13` branch instead.

axw: Another couple of options:
 - We could move the elastic-package dep to a separate go.mod; I ended up doing this for the Go Agent: https://github.com/elastic/apm-agent-go/blob/master/scripts/jenkins/jenkins.go.mod 
 - I _think_ we could use a `replace` directive

stuartnelson3: I'm sort of inclined to use officially tagged versions and use `v7.13.1` for now and keep it simple

x-pack/apm-server/sampling/pubsub/pubsubtest/client.go
```diff
@@ -106,20 +106,22 @@ func (rt *channelClientRoundTripper) RoundTrip(r *http.Request) (*http.Response,
 	var handler func(*http.Request, *httptest.ResponseRecorder) error
 	switch r.Method {
 	case "GET":
-		if strings.HasSuffix(r.URL.Path, "/_stats/get") {
+		if strings.HasSuffix(r.URL.Path, "/") {
```
simitt, line 109: I think this should be `r.URL.Path == "/"` , rather than `strings.HasSuffix`, to avoid matching for example, `/_stats/get/`

go.mod
```diff
@@ -91,9 +90,12 @@ replace (
...
-	k8s.io/api => k8s.io/api v0.19.4
-	k8s.io/apimachinery => k8s.io/apimachinery v0.19.4
-	k8s.io/client-go => k8s.io/client-go v0.19.4
+	k8s.io/api => k8s.io/api v0.21.2
+	k8s.io/apimachinery => k8s.io/apimachinery v0.21.2
+	k8s.io/client-go => k8s.io/client-go v0.21.2
+	k8s.io/kubectl => k8s.io/kubectl v0.21.2
 )
 
+replace github.com/elastic/go-elasticsearch/v7 => github.com/elastic/go-elasticsearch/v7 v7.5.1-0.20210701130713-3c08ee738285
```
simitt, line 99: why is sthis still necessary when using the released `7.13` version?

simitt (changes requested):

go.mod
stuartnelson3: It's not, after I had to update some `replace` directives to get new dependencies happy, i thought "well i guess i can just go to 7.14 if I'm already doing `replace` stuff". Happy to just use 7.13.1 though.

x-pack/apm-server/sampling/pubsub/pubsubtest/client.go
stuartnelson3: I remember now why I did that, and everywhere else is doing `strings.HasSuffix`: r.URL.Path is always returning (and I have no idea why) host+path, eg.:

```
testing.invalid/
testing.invalid/traces-sampled-testing/_stats/get
testing.invalid/index_name/_refresh
testing.invalid/index_name/_search
```

We're only using `testing.invalid/`, and I can do that as a direct comparison

go.mod
simitt: got it; this is fine.

simitt (approved):

stuartnelson3:
Unsure what to do about the system test failure:

```
Error Message

requests.exceptions.ConnectionError: ('Connection aborted.', ConnectionResetError(104, 'Connection reset by peer'))
```
https://apm-ci.elastic.co/job/apm-server/job/apm-server-mbp/job/PR-5599/9/testReport/tests.system.test_auth/TestAPIKeyWithInvalidESConfig/Build_and_Test___System_and_Environment_Tests___test_backend_intake/

if someone has any insight it'd be much appreciated

axw:
> if someone has any insight it'd be much appreciated

I think it's related to the new product check in go-elasticsearch. It appears to be making two requests, each of which takes 15 seconds, exceeding the client's request timeout:

https://github.com/elastic/go-elasticsearch/blob/3c08ee738285efe07a22da30d2e83ed000a2de0c/elasticsearch.go#L282-L310

I think that's a bug in go-elasticsearch though? I don't think it should make the second request if the first one fails. Also those sync.Onces are a smell; looks like transient errors will be cached.

stuartnelson3:
Makes sense. The product check was introduced in 7.14. Would you prefer waiting for a fix, or downgrading to 7.13.1?

axw:
https://github.com/elastic/go-elasticsearch/pull/303 should fix the problem. I'd like to wait for that to be merged and backported. If there are any significant delays then we can reassess

go.mod
Anaethelion: @simitt is right, for reference this is the `go mod` doc for [version](https://golang.org/ref/mod#versions), [version resolution strategies](https://golang.org/ref/mod#minimal-version-selection) and a simpler explanation on [jfrog's blog](https://jfrog.com/blog/go-big-with-pseudo-versions-and-gocenter/).

For the future I can offer to create a `v7.x.0-pre` tag upon each branch creation to ease understanding before the release tag. I'll check for incompatibility and get back to the team if it's doable!

Anaethelion:
https://github.com/elastic/go-elasticsearch/pull/303 has been merged and back ported as https://github.com/elastic/go-elasticsearch/pull/304 !

Issue #4941, opened 2021-03-10
CI: automate update-beats


axw:
Currently we need to periodically run `make update-beats` to keep apm-server up to date with libbeat and beats packaging. I'd like this to be automated to avoid prolonged version skew. There's some overlap here with https://github.com/elastic/apm-server/issues/4240.

The way I imagine this working is to have a Jenkins job that:
 1. Runs `make update-beats`
 2. Stages the changes to `go.mod`, `go.sum`, `NOTICE.txt` ~and `testing/environments` (mirrored from libbeat)~
 3. Creates a PR with auto-merge and delete-on-merge bits set

I think (1) and (2) should be taken care of by https://github.com/axw/apm-server/commit/54cbe197056a7fc633bf03c366ffb5f32f5e8b66.

For (3) we should be able to use `githubCreatePullRequest`, but I'd really like this to be auto-merged and branches deleted automatically. I think that might require migrating to use the official GitHub CLI (`gh`, cli.github.com) instead of `hub`, which provides access to auto-merge/delete through `gh pr merge --auto --delete-branch`.

v1v:
> For (3) we should be able to use githubCreatePullRequest, but I'd really like this to be auto-merged and branches deleted automatically. I think that might require migrating to use the official GitHub CLI (gh, cli.github.com) instead of hub, which provides access to auto-merge/delete through gh pr merge --auto --delete-branch.

There is a CI pipeline step for the `gh` -> https://github.com/elastic/apm-pipeline-library/blob/master/vars/gh.txt
`gh(command: 'pr merge --auto --delete-branch')`

v1v:
PRs:
- https://github.com/elastic/apm-server/pull/5690 to create the Job in the Jenkins UI
- Another PR will be created with the whole automation.

The idea to split in two different PRs will help with the development process:
- JJBB are nothing but a change in the UI.
- Pipelines can be tested on a PR basis if the above already exists.

But my local environment is not in a good shape so, I'm iterating with the above approach for now

PR #5610, opened 2021-07-05
update release doc

stuartnelson3:
- Clarify version bumping for beats
- Add command for pulling the latest version of `go-elasticsearch`
- Add step for updating mergify config file


RELEASES.md
```diff
@@ -9,11 +9,17 @@
 
 * Update beats
 
-  run `make update-beats` in the branch from which the new branch will be created before FF to recognize potential issues
+  * Before feature freeze and creating the release branch, update `BEATS_VERSION` to the release version in the top-level Makefile of the release's parent branch
```
axw, line 12: I'm not sure I understand this addition. Maybe some examples would help.

In this case for example, let's say the release is 7.14. I take "the release's parent branch" to mean 7.x. Is that what you meant? We shouldn't need to update `BEATS_VERSION` in the 7.x branch - that's only updated when we branch, not before. This is part of the `update versions in release branch` task below, but it could probably be called out.

I think the idea of the existing comment was that if we (ideally regularly) run `make update-beats` on the 7.x branch, we _should_ avoid surprises when it comes to time to branching and updating beats on the release branch. In the hopefully-not-too-distant-future this should be taken care of by https://github.com/elastic/apm-server/issues/4941

stuartnelson3: ah k. yeah I thought this was essentially a repeat of the below step, where this gets run regularly before a "final"-ish update post-branch.

I'll remove this section then?

axw: Yeah I think so. Maybe call out the BEATS_VERSION change in the below section?

stuartnelson3: cool, updated

axw (approved):

Issue #4958, opened 2021-03-16
RUM: `sourcemap.error` on and `sourcemap.updated` are set overzealously


axw:
We currently store `sourcemap.error` and `sourcemap.updated` on _every_ stack frame in RUM spans and errors if no matching sourcemaps have been uploaded. This will bloat documents considerably. I would like to revisit this decision, and only set these fields under certain conditions. Specifically:

- Only set `sourcemap.error` if a matching sourcemap is found and some error occurred while attempting to match and update frames. Not matching a sourcemap to a frame is not an error. For now we'll continue to set `sourcemap.error` at the per-frame level, but later it might make sense to set at the document level when performing sourcemapping in ingest node -- this would be more natural since we can use an `on_failure` handler.
- Only set `sourcemap.updated` (to true) when a frame is matched and updated. Non-presence of this field means the frame has not been updated.

Issue #4120, opened 2020-08-31
Introduce efficient codec for model events


axw:
Introduce an efficient binary codec for model events (at least transactions and spans), for persisting to local storage. This would be used in tail-based sampling, and perhaps in other processing such as https://github.com/elastic/apm-server/issues/5936.

For the most efficient codec, we should use a statically defined encoding. Barring arguments made for alternatives, let's use protobuf.

In order to have a statically defined encoding we'll need our model to be stable. The events we produce are now substantially ECS-compliant (which is stable), with a few exceptions. For example, we emit `transaction.duration.us` or `span.duration.us`; ideally we would record only the ECS `event.duration` field (https://github.com/elastic/apm-server/issues/5999).

I suggest we split this up into two phases:

**Phase 1**
- Continue to maintain package model by hand, and continue to use this to emitting JSON documents
- Introduce protobuf definitions, generate code in a separate package with protoc
- Generate code to translate between handwritten model types and protoc-generated Go types

**Phase 2**
- Use protobuf to generate JSON encoding code (https://github.com/elastic/apm-server/issues/3565)
- Investigate using protoc-generated Go types everywhere, and removing handwritten model types

For phase 2 we'll need some more changes to remove any remaining logic in the model package, e.g. setting `<event>.duration.us`. This can be done in an ingest pipeline.

axw:
I spent a bit of time looking into this, with https://github.com/elastic/apm-server/compare/master...axw:model-protobuf being the fruits of my labour.

I wrote a generator which parses the model types and generates protobuf definitions, and then used plain old golang/protobuf to generate the code. I was going to use gogo/protobuf but had some probably-PEBKAC issues and ended up abandoned that for this short spike.

For generating protobuf correctly we would probably need to check them in and take the old definitions into account, so we don't accidentally remove fields and break compatibility with previous releases. Another option would be to have hand-written protobuf as the source of truth for our model types and generate the model types instead, using standard protoc tooling. That would be more typical and almost certainly less error prone. It might be painful to migrate though.

While testing I also realised that the model types we have do not have any encoding/json tags, and so we end up encoding all the empty fields. I added some `json:",omitempty"` tags to a bunch of fields to reduce this.

I compared this with the existing JSON codec, but bear in mind that the benchmarks are far from comprehensive. They currently only encode model.Transactions with the TraceID and ID fields set, and nothing else. With a more comprehensive set of fields, we might observe a different outcome. Nevertheless, here's what I found.

```
[...]

PR #5631, opened 2021-07-07
Move source mapping to model processing

axw:
## Motivation/summary

Source mapping transformation is rewritten as a `model.BatchProcessor`, moved out of the model package and into the sourcemap package. Moving such logic out of the model package takes us another step towards generating the model types and `beat.Event` transformation.

One side-effect of the move to a process, worth debating, is that source mapping now occurs in the HTTP handler goroutine rather than in the publisher, which could increase request latency slightly. On the other hand this means that source mapping for more clients than there are CPUs can now happen concurrently (the publisher is limited to the number of CPUs in this regard); and the handlers will block for up to one second anyway if the publisher is busy/queue is full.

If a stack frame cannot be mapped, we no longer set `sourcemap.error` or log anything. Just because RUM and source mapping is enabled, does not mean that all stacks _must_ be source mapped; therefore these "errors" are mostly just noise. Likewise we now only set `sourcemap.updated` when it is true.

I removed the `TestSourcemapCacheExpiration` system test as I believe it is redundant. In the test immediately above that one we prove that caching is effective, and cache expiration is unit tested.

## Checklist

## How to test these changes

1. Enable RUM, upload a source map, check that the source map is applied to indexed events
2. Enable RUM, _don't_ upload a source map, check that events are indexed without `sourcemap.error` or `sourcemap.updated `set

## Related issues

Prerequisite for https://github.com/elastic/apm-server/issues/4120 and https://github.com/elastic/apm-server/issues/3565
Closes https://github.com/elastic/apm-server/issues/4958


axw:
The failures highlight an obvious issue, which is that if Elasticsearch is unavailable it should not prevent ingestion. When we have ditched the old direct-to-Elasticsearch source map implementation this won't be an isssue, but for now we'll need to continue logging and recording these errors in `sourcemap.error`.

axw:
I reinstated the logging/recording of source map fetching errors.

It is possible for the processor to take a significant amount of time due to Elasticsearch search retries, e.g. when the server is unavailable. In a realistic setup I can't see moving this from the publisher to the handler making a practical difference. If the delay occurs in the publisher, then its queue would back up and cause the handler to timeout. I have added a configurable timeout for the sourcemap processor, defaulting to 5s. This is currently ineffective due to https://github.com/elastic/go-elasticsearch/issues/300.

In the future when we're fully on Fleet, the catalogue of source maps will be pushed to APM Server. I think it would then be reasonable for APM Server to eagerly fetch and cache these, rather than fetching them on demand. Then the fetching will not be in the critical path of the handler, and it can just use them if they're in memory.

beater/beater.go
```diff
@@ -618,24 +625,14 @@ func runServerWithTracerServer(runServer RunServerFunc, tracerServer *tracerServ
 	}
 }
 
-func newTransformConfig(beatInfo beat.Info, cfg *config.Config, fleetCfg *config.Fleet) (*transform.Config, error) {
-	transformConfig := &transform.Config{
+func newTransformConfig(beatInfo beat.Info, cfg *config.Config, fleetCfg *config.Fleet) *transform.Config {
```
stuartnelson3, line 628: `fleetCfg` can be removed as an argument, it was only used in `newSourcemapStore`

beater/api/asset/sourcemap/handler.go
```diff
@@ -46,8 +47,14 @@ var (
 	validateError = monitoring.NewInt(registry, "validation.errors")
 )
 
+// AddedNotifier is an interface for notifying of sourcemap additions.
+// This is implemented by sourcemap.Store.
+type AddedNotifier interface {
```
stuartnelson3, line 52: What's the motivation for exporting this interface? It doesn't cost anything to export it, just curious since the interface isn't being referenced as an argument or struct member anywhere else.

stuartnelson3 (approved):
One small change and a question; but LGTM!

beater/api/asset/sourcemap/handler.go
axw: It's used in the Handler function below. Am I misunderstanding your question?

axw: Going to merge as-is, happy to update if you think there's an improvement to be had.

simitt (approved):
LGTM

beater/api/asset/sourcemap/handler.go
stuartnelson3: I don't think it's an improvement, just wondering your personal opinion about exporting or not exporting an interface. In this case, `AddedNotifier` is only referenced once in the project, as an argument in the below function in the same package. A non-exported `addedNotifier` would also have worked, so I was curious if there was a specific motivation behind the choice.

axw: Gotcha. My preference is to always have the interface exported if it is referenced by some exported type/function. IMO you should be able to understand how to use the interface just by clicking through the godoc UI, without delving into source.

stuartnelson3: cool, thanks!

PR #5637, opened 2021-07-08
Stop running python system tests on Windows

axw:
## Motivation/summary

We run the Python system tests on Windows, but almost all of them are skipped because we run them without the stack containers. What remains is a handful of trivial tests, so I don't think it's worth running them at all.


simitt (approved):
Agreed, let's remove them from the windows tests; it just gives a wrong impession of having system test coverage on windows.

PR #5656, opened 2021-07-09
Fix rum rate limit default value

simitt:
## Motivation/summary

Map the cloud default value for `apm-server.rum.rate_limit` to the newer config setting `apm-server.rum.event_rate.limit` for the switch to data streams. 

## How to test these changes
Initiate switch to data streams in the Kibana UI on a cloud deployment and ensure the `apm-server.rum.rate_limit` setting does not show up as unsupported.


stuartnelson3 (approved):

axw:
Tested with 7.14-SNAPSHOT (apm-server reports fb38ac015cae7f8ec0da42568c5a95b5c851c0c9, which includes the backport commit for this.)

The config is still there:

image

Looking at the code, this:

https://github.com/elastic/apm-server/pull/5656/files#diff-9d07c75e6440575e8136aa51bd45f376ef5d9c2dfdc9f199951115a42ff4df7bR117

should probably be reassigning `k` rather than updating `out`, as otherwise we'll set both keys.

simitt:
> should probably be reassigning k rather than updating out, as otherwise we'll set both keys.

It is missing a `continue` after setting the key.

stuartnelson3:
fixed in #5699, added label for filtering purposes

Issue #5595, opened 2021-07-01
[compressed spans] Adapt span metrics calculation


simitt:
To make the compressed feature (https://github.com/elastic/apm/issues/432) work we need to extend the intake API (https://github.com/elastic/apm-server/issues/5485) and adapt the calculation of span metrics.

## Adapt span metrics calculation
Since this feature is folding a set of spans into one single we need to adapt the calculation of span metrics so that the resulting metric values are still correct.

axw:
I _think_ this is a simple matter of updating `model.Span.RepresentativeCount` when decoding "span.composite": https://github.com/elastic/apm-server/blob/06bea7cfe1263c416bb904f455e4e9d436a4f9b0/model/span.go#L80-L85

i.e. if we set that to `1/sample_rate * composite.count`, then the existing span metrics aggregator should do the right thing, IIANM

Issue #5485, opened 2021-06-17
Make intake API support compressed spans


AlexanderWert:
To make the compressed feature (https://github.com/elastic/apm/issues/432) work we need to extend the intake API and adapt the calculation of span metrics (https://github.com/elastic/apm-server/issues/5595).

### Extend data model
We need to introduce a new property `composite` in the span intake data model.
The semantics is the following. 

- If this field is `null`, then the corresponding span is a normal span as before
- If the compressed property is **not** `null`, then it's a compressed span
  - in this case the `duration` represents the sum of all compressed spans
  - the `composite.count` represents the number of spans that have been compressed into this single span
  - `composite.end` is the end timestamp of the last span --> `end - timestamp` is the overall duration of the compressed spans including any delays between the spans
  - `composite.exact_match` is a boolean that indicates whether the compressed spans are exact matches of each other or of the same kind (numerous alternating, fast queries to the same backend)

```
{
   "name": "SELECT FROM product_types WHERE user_id=?",
   "timestamp": 1613575086000000,
   "duration": 3781,
   "composite": {
       	"count": 10,
        "end": 1613575093183900,
        "exact_match": true
    }
}
```

AlexanderWert:
Resolved with https://github.com/elastic/apm-server/pull/5661

PR #5661, opened 2021-07-09
Add support for Composite spans in the intake API

estolfo:
## Motivation/summary

Support a composite object in the span model and in the intake API. The composite object holds metrics on a group of spans represented by a single one.

The composite object has the following format

```
{
   "name": "SELECT FROM product_types WHERE user_id=?",
   "timestamp": 1613575086000000,
   "duration": 3781,
   "composite": {
       	"count": 10,
        "sum": 3592,
        "compression_strategy": "exact_match"
    }
}
```

## Checklist

## How to test these changes

To test these changes specifically, not including the changes to span metrics calculations:

- `make`
- `docker compose up -d`
- Add the ES credentials to `apm-server.yml` (`username: "admin"`, `password: "changeme"`)
- `./apm-server -c apm-server.yml -e -d "*"`
- `curl -H "Content-type: application/x-ndjson" --data-binary @testdata/intake-v2/spans.ndjson http://localhost:8200/intake/v2/events`
- In Kibana Dev Tools, run the following query and verify the span document has the expected `composite` nested object.
```
GET /apm-8.0.0-span/_search
{
  "query": {
    "match": {
      "trace.id": "edcbaf0123456789abcdef9876543210"
    }
  }
}
```
Expected nested document:
```
            "composite" : {
              "count" : 10,
              "sum" : {
                "us" : 359298
              },
              "compression_strategy" : "exact_match"
            }
```

This can also be tested along with https://github.com/elastic/apm-server/issues/5595. 

## Related issues

Related #5485 
Related #5595


model/span.go
```diff
@@ -189,6 +197,19 @@ func (d *DestinationService) fields() common.MapStr {
...
 
+func (c *Composite) fields() common.MapStr {
+	if c == nil {
+		return nil
+	}
+	var fields mapStr
+	if c.Count > 1 {
+		fields.set("count", c.Count)
+	}
+	fields.maybeSetBool("exact_match", &c.ExactMatch)
```
axw, line 208: This only sets the field when it's true. Is that intentional, or should it always be set?

model/span.go
```diff
@@ -189,6 +197,19 @@ func (d *DestinationService) fields() common.MapStr {
 	return common.MapStr(fields)
 }
 
+func (c *Composite) fields() common.MapStr {
+	if c == nil {
+		return nil
+	}
+	var fields mapStr
+	if c.Count > 1 {
```
axw, line 205: Isn't it necessarily >= 2, if it's non-nil?

axw (changes requested):
Nice work! What's there looks good, just a few very minor comments aside from the one about `end`. I'm not sure if we really need that. If we _do_ need it, I'm wondering if we could store it as the ECS field `event.end` instead.

model/modeldecoder/v2/model.go
```diff
@@ -768,6 +771,22 @@ type stacktraceFrame struct {
...
 
+type spanComposite struct {
+	// Count is the number of compressed spans the composite span represents.
+	// The minimum count is 2 as a composite span represents at least two spans.
+	Count nullable.Int `json:"count" validate:"required,min=2"`
+	// End is the end timestamp of the last compressed span.
+	// The net duration of all compressed spans is equal to the composite spans'
+	// duration. The gross duration, including the time between the spans, is
+	// equal to (composite.end - timestamp).
+	End nullable.TimeMicrosUnix `json:"end" validate:"required"`
```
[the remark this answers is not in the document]
estolfo: My understanding is that the duration is the sum of all the span durations but that doesn't necessarily equal composite.end - timestamp, as the end timestamp takes into account the delay between spans.
In other words, (start timestamp + duration) doesn't necessarily equal end timestamp.

model/span.go
estolfo: It looks like from [this code](https://github.com/elastic/apm-server/blob/master/model/mapstr.go#L39-L45) that the boolean value is set as long as it's not `nil`. Am I reading the code incorrectly?
Also, [this test](https://github.com/elastic/apm-server/pull/5661/files#diff-69fa71f505412f98fb040c0923f0ff552c47f73e3315356a26b1eb41938a9a71R123-R165) would fail if it only set the field when it was true.

model/modeldecoder/v2/model.go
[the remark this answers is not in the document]
estolfo: I'll ask about storing it in `event.end`

model/span.go
axw: Oh sorry, I missed that `c.ExactMatch` is a bool field, and you're taking the address of it. Since you're taking the address, and the pointer is always non-nil, it will always be set. In that case, this would be simpler:

```suggestion
	fields.set("exact_match", c.ExactMatch)
```

estolfo: That makes sense, thanks!

model/span.go
```diff
@@ -189,6 +197,19 @@ func (d *DestinationService) fields() common.MapStr {
...
+		fields.set("count", c.Count)
+	}
+	fields.set("exact_match", c.ExactMatch)
+	fields.maybeSetMapStr("end", utility.TimeAsMicros(c.End))
```
estolfo, line 209: @axw The `end` field is required in the composite object, so this can also be simply `fields.set("end", utility.TimeAsMicros(c.End))`, correct?

axw: Yes, unless `c.End` can be zero, in which case `utility.TimeAsMicros(c.End)` will return `nil`. AFAICS we're not currently validating that c.End is > 0.

axw (commented):
Basically LGTM, but I just realised we still need to update https://github.com/elastic/apm-server/blob/master/model/span/_meta/fields.yml. This defines how fields are mapped in Elasticsearch. After that file is updated, you will need to run `make update` to update generated docs and integration package files.

estolfo:
As you're aware @axw, the agent spec is still being discussed and updated. So I'm going to wait until that's settled a bit more, update this PR, then request a review again. Thanks for your review so far!

model/span.go
```diff
@@ -111,6 +112,13 @@ type DestinationService struct {
 	Resource string
 }
 
+// Composite holds details on a group of spans compressed into one.
+type Composite struct {
+	Count               int
+	Sum                 float64
```
axw, line 118: would you mind adding a comment to Sum (at least), noting that it is in milliseconds?

axw (approved):
Looks great, thanks! Just one minor comment. Please go ahead and merge after adding a comment on Composite.Sum. When you do, please add the `backport-7.x` label to have Mergify automatically create the backport.

stuartnelson3:
Confirmed with BC2

Issue #5203, opened 2021-04-30
Add fields for mobile network connectivity attributes


bryce-b:
* The following fields need to be added:
`net.host.connection_type`: `5G`,`4G`,`3G`,`2G`,`1G`, `wifi`,`unavailable` (String)
`net.host.carrier.mcc` : mobile country codes (String)
`net.host.carrier.mnc` : mobile network codes (String) 
`net.host.carrier.name`: Carrier names (String)  (e.g.: `T-Mobile`, `AT&T`, etc)
`net.host.carrier.icc`: 2-character iso country code (String) (e.g.: `US`, `DE`)

* These fields also need to be added into ECS 

Note: Here are related tickets
[OpenTelemetry spec issue](https://github.com/open-telemetry/opentelemetry-specification/issues/1647)
[iOS agent ticket](https://github.com/elastic/apm-agent-ios/issues/23)

bryce-b:
Here's a link to a rudimentary test app for generating iOS data: https://github.com/elastic/ios-integration-testing
It requires MacOS. 
the apm server ip/url can be set in this file : https://github.com/elastic/ios-integration-testing/blob/master/Shared/Model/ios_integration_testingApp.swift
Let me know if you need help getting it set up.

axw:
We should keep https://github.com/elastic/apm-server/pull/4912 in mind when we do this - specifically there's overlap on `net.host.connection_type`.

jalvz:
@bryce-b should any of these fields be required?

bryce-b:
@jalvz what are the implications of being 'required'? Will the traces be ignored if they are not attached? If so, they shouldn't be required.

jalvz:
Exactly, thanks 👍

PR #5671, opened 2021-07-12
Update network connection fields

axw:
## Motivation/summary

The OTel semantic conventions for network connection type have evolved: `net.host.connection_type` is now `net.host.connection.type` and `net.host.connection.subtype`. See https://github.com/open-telemetry/opentelemetry-specification/pull/1759. This PR extends https://github.com/elastic/apm-server/pull/5436 to adapt to these changes.

Support Elastic APM agents sending "network.connection.type" ("n.c.t" in RUMv3) in metadata, which will be added to all events in the stream. This is currently intended for RUM. In the future we may want to support setting this at the even level too, to cater for operation-specific network information.

Finally, the Elasticsearch field names have been changed to `network.connection.type` and `network.connection.subtype`.

## Checklist

## How to test these changes

1. Run the iOS agent
2. Check that the `network.connection.type` and `network.connection.subtype` fields are recorded

## Related issues

https://github.com/elastic/apm-server/issues/5203


model/error/_meta/fields.yml
```diff
@@ -357,7 +357,13 @@
         - name: connection_type
           type: keyword
           description: >
-            Cellular network technology, eg. 4G
+            Network type, eg. "wifi", "cell"
+          overwrite: true
+
+        - name: connection_subtype
```
simitt, line 363: Should we change to `connection.type` and `connection.subtype` on the document level?

simitt (commented):
What is the urgency of this PR being backported to `7.14`? Does the iOS agent not work with the current fields, or would the `network.connection*` fields just be ignored?

model/error/_meta/fields.yml
bryce-b: I haven't made the `.type` & `.subtype` updates to the agent yet, since the spec hasn't been merged yet. So the iOS agent is still sending `net.host.connection_type` and not yet sending anything related to `subtype`.

simitt: In this case I suggest we keep what we have for `7.14` and keep this PR open until the spec is merged.

bryce-b:
@simitt I mentioned to Andrew in slack: we don't need to backport this to 7.14; I've been holding off on updating the agent with these spec changes due to the Otel spec change not being merged yet.

model/error/_meta/fields.yml
axw: I've made this change now.

stuartnelson3 (approved):

simitt:
Cannot yet be tested as the latest version of the iOS agent does not contain the change; it still sends `net.host.connection_type` instead of `net.host.connection.type`, see https://github.com/elastic/apm-agent-ios/blob/main/Sources/Instrumentation/NetworkInfo/NetworkStatusInjector.swift#L28.

@bryce-b any update on when this will be changed in the agent?

bryce-b:
@Silvia These changes will be in this week.

PR #5676, opened 2021-07-12
mergify: change backport trigger labels

axw:
## Motivation/summary

Change the labels used for triggering backports to explicitly specify the target branch name.

This avoids an issue where "v7.14.0" creates backports for 7.x, and then when we update mergify to have "v7.14.0" backport to 7.14 redundant backport PRs are created.

## Related issues

https://github.com/elastic/apm-server/pull/5669
https://github.com/elastic/apm-server/pull/5668
https://github.com/elastic/apm-server/pull/5667
...

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -17,7 +17,7 @@ pull_request_rules:
     conditions:
       - merged
       - base=master
-      - label=v7.15.0
+      - label=backport-7.x
     actions:
       backport:
         assignees:
@@ -29,7 +29,7 @@ pull_request_rules:
     conditions:
       - merged
       - base=master
-      - label=v7.14.0
+      - label=backport-7.14
     actions:
       backport:
         assignees:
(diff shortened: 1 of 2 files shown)


axw:
Ideally we would use a match operator (`~=`) in the label check, and backport according to the branch name in the label. That way we could just have one rule that changes the target branch depending on the label. AFAICT it is not possible to use a regex capturing group to do this.

simitt (approved):
Thanks for the quick turnaround!

v1v:
> Ideally we would use a match operator (~=) in the label check, and backport according to the branch name in the label. That way we could just have one rule that changes the target branch depending on the label. AFAICT it is not possible to use a regex capturing group to do this.

I'll raise a discussion in the mergify project and see if they have any suggestions.

Regarding this PR:
* if it helps we raise PRs in the Beats project to bump the 7.x version when needed. If it helps we could do so. 
* I like this approach, although, it changes the existing backport process with a new label that follows a different semantic compared to what we have done so far. I'm not sure yet if we need to be cautious

axw:
> I'll raise a discussion in the mergify project and see if they have any suggestions.

Thanks @v1v!

> if it helps we raise PRs in the Beats project to bump the 7.x version when needed. If it helps we could do so.

Maybe later, but that wouldn't help with this problem.

> I like this approach, although, it changes the existing backport process with a new label that follows a different semantic compared to what we have done so far. I'm not sure yet if we need to be cautious

Yeah, the existing approach has a problem. I'll elaborate with an example using the existing approach of using `v7.N.0` labels.

During the development of 7.14:
- the 7.x branch has version 7.14.0
- the label used to backport from master to the 7.x branch is v7.14.0

After feature freeze:
- we create branch 7.14
- the 7.x branch's version is bumped to 7.15.0
- the label used to backport from master to the 7.14 branch is v7.14.0
- the label used to backport from master to the 7.x branch is v7.15.0

So the target branch for v7.14.0 changes before/after feature freeze. This leads to Mergify recreating the backports.

v1v:
Thanks for the clarity, I'll also ask in the mergify if we can do something to avoid the recreation, I actually saw it in the past but somehow I thought it was some glitch with the GitHub events or the service but nothing that was always happening. Thanks again


https://github.com/Mergifyio/mergify-engine/discussions/2865 is the discussion I raised, please feel free to amend anything that I said :)

v1v (approved):

PR #5687, opened 2021-07-13
Move more code from model to model processors

axw:
## Motivation/summary

Move more logic out of model transformation, into dedicated model processors:

 - `modelprocessor.SetExcludeFromGrouping` matches stack frames against a regular expression, setting the "exclude_from_grouping" field
 - `modelprocessor.SetLibraryFrame` matches stack frames against a regular expression, setting the "library_frame" field
 - `modelprocessor.SetCulprit` identifies the first source mapped, non-library, stack frame and uses it to set or update the error culprit

The "library_frame" field is now only set if it is true, to simplify code and keep documents minimal. The UI collapses stack frames identified as library frames, but takes omission to mean the same thing as false, i.e. "falsy", so this change has no other practical effects.

The new model processors are only installed for the RUM intake APIs. With these changes, we no longer need to identify events as originating from the RUM endpoint, so we remove the `model.Span.RUM` and `model.Error.RUM` fields.

## How to test these changes

Send an exception event to APM Server, check that (non-)library frames are still differentiated.

## Related issues

Pulling logic out of model transformation for
- https://github.com/elastic/apm-server/issues/4120
- https://github.com/elastic/apm-server/issues/3565


beater/api/mux.go
```diff
@@ -181,18 +184,35 @@ func (r *routeBuilder) rumIntakeHandler(newProcessor func(*config.Config) *strea
...
+			if err != nil {
+				return nil, errors.Wrap(err, "invalid exclude from grouping regex")
 			}
+			batchProcessors = append(batchProcessors, modelprocessor.SetExcludeFromGrouping{Pattern: re})
 		}
-		batchProcessor = batchProcessorWithAllowedServiceNames(batchProcessor, r.cfg.RumConfig.AllowServiceNames)
-		h := intake.Handler(newProcessor(r.cfg), requestMetadataFunc, batchProcessor)
+		if r.sourcemapStore != nil {
+			batchProcessors = append(batchProcessors, modelprocessor.SetCulprit{})
+		}
```
simitt, line 213: Could you add a note that the order of the processors is important?

simitt (commented):
Changes generally look good!

beater/api/mux.go
axw: done

simitt (approved):

PR #5699, opened 2021-07-14
Fix: remove unsupported rum rate limit key

simitt:
## Motivation/summary
Fix: Remove the wrong rum rate limit setting and send the correct one instead. 

## How to test these changes
See #5656

## Related issues
follow up on #5656


axw (approved):

axw:
Tested with 7.14.0-SNAPSHOT (6aad3135bec274366c3454f3aef656732fe73b8f) on QA cloud, looks better now:

image

Issue #5663, opened 2021-07-09
OTel array attributes not recorded as labels


stuartnelson3:
**APM Server version** (`apm-server version`): 
```
# ./apm-server version
apm-server version 7.14.0 (amd64), libbeat 7.14.0 [e369a0ef477ce27654ac3a401773ec7806bd3394 built 2021-07-08 06:28:07 +0000 UTC]
```

**Description of the problem including expected versus actual behavior**:

When sending traces using an attribute array, the values do not appear in the document in elasticsearch. When changing the code to send a single attribute value, the value appears as a label.

single attribute:
![image](https://user-images.githubusercontent.com/1398104/125111057-b7294580-e0e5-11eb-930c-2fab2bda0551.png)

**Steps to reproduce**:

 1. Start elasticsearch + kibana + apm-server build candidates for v7.14
 2. Run this [server](https://gist.github.com/stuartnelson3/61b9e61bf4ab78955dc0a7844d26c626): `EXPORTER_ENDPOINT=localhost:8200 go run main.go`
 3. `curl http://localhost:8080/users/123`
 4. Check the last document when querying elasticsearch `GET apm-7.14.0-span-000001/_search`

The gist linked above has a commented out line at the bottom setting a single string attribute.

axw:
I did it for transactions, but forgot to do it for spans too. Also resource attributes.

stuartnelson3:
Testing against #5704, array attributes at the resource level are correctly handled, but when I add them at the span level they still aren't appearing. Single value attributes added at the span level aren't appearing for me now, either.

version: `apm-server version 7.14.0 (amd64), libbeat 7.14.0 [bea70f612b1d70561dbedd57b1cd21fe56120795 built 2021-07-19 05:24:50 +0000 UTC]`

gist: https://gist.github.com/stuartnelson3/61b9e61bf4ab78955dc0a7844d26c626 (updated to add resource attributes)

axw:
@stuartnelson3 I just ran the latest 7.14-SNAPSHOT image, and pointed your gist at it:

> ```
> $ docker run --rm -p 8200:8200 docker.elastic.co/apm/apm-server:7.14.0-SNAPSHOT -E output.console.enabled=true -E output.elasticsearch.enabled=false
> 
[...]

PR #5704, opened 2021-07-14
processor/otel: record array resource attributes

axw:
## Motivation/summary

If an unhandled array attribute is found in resource attributes, store that as a label like we do with span attributes.

Also, handle array attributes when translating to Elastic spans and not just transactions.

## Checklist

## How to test these changes

Send a trace with OTLP with array attributes at both the resource and span level, and ensure they are recorded as labels.

## Related issues

Fixes https://github.com/elastic/apm-server/issues/5663


stuartnelson3 (approved):

stuartnelson3:
Confirmed against SNAPSHOT

Issue #5689, opened 2021-07-13
Sourcemap filled up the root filesystem in the /tmp directory.


clyfish:
**APM Server version** (`apm-server version`):
7.12.0

**Description of the problem including expected versus actual behavior**:
One of our apm-server instance responses `400 Bad Request` when uploading some large sourcemap files.
I logined in the instance, and found there are many sourcemap files in the /tmp directory filling up the root filesystem.

**Steps to reproduce**:
Upload a large (46MB) sourcemap file to apm-server, the temporary file in the /tmp directory does not get cleared.

**Provide logs (if relevant)**:
```bash
bash-4.2$ df -h /
Filesystem      Size  Used Avail Use% Mounted on
overlay          10G   10G   20M 100% /
bash-4.2$ pwd
/tmp
bash-4.2$ du -sh
10G     .
bash-4.2$ ls -l | head 
total 10465996
-rwx------ 1 root       root      836 Nov 13  2020 ks-script-DrRL8A
-rw------- 1 apm-server root 47006226 Jun 21 08:17 multipart-004292911
-rw------- 1 apm-server root 47676535 Jul  8 12:42 multipart-009546798
-rw------- 1 apm-server root 47676652 Jul 12 09:20 multipart-016420862
-rw------- 1 apm-server root 47454195 Jun 23 06:26 multipart-020602083
-rw------- 1 apm-server root 36134883 Jul  8 06:27 multipart-021437042
-rw------- 1 apm-server root 36134883 Jun 22 10:00 multipart-022302060
-rw------- 1 apm-server root 36134883 Jun 28 06:14 multipart-028398313
-rw------- 1 apm-server root 46708689 Jun 17 07:48 multipart-034270813
```

axw:
Thanks for reporting this @clyfish.

I can reproduce this on the master branch. Go's standard library writes form data to /tmp when it exceeds 32MB, so I just added a 32MB string to https://github.com/elastic/apm-server/blob/master/testdata/sourcemap/bundle.js.map and uploaded it. A "multipart-..." file was left behind in /tmp. It is supposed to be removed automatically when the upload completes, but clearly that's not working properly. I'll look into it.

PR #5718, opened 2021-07-15
beater/request: clear multpart.Form on reset

axw:
## Motivation/summary

After completing a request, immediately reset the context so we don't hold onto memory while the context in the memory pool.

When resetting request context, call mime/multipart.Form.RemoveAll to remove temporary files created for storing uploaded source maps. Although net/http does this at the end of a request, it does so on the original net/http.Request. This does not work when the request is copied (e.g. using Request.WithContext), which we do in various middleware.

## Checklist

## How to test these changes

1. Upload a source map that is greater than 32MB in size
2. Check that no "multipart-*" files are left behind in /tmp

## Related issues

Fixes https://github.com/elastic/apm-server/issues/5689

### DIFF
--- a/beater/request/context_pool.go
+++ b/beater/request/context_pool.go
@@ -42,6 +42,7 @@ func (pool *ContextPool) HTTPHandler(h Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		c := pool.p.Get().(*Context)
 		defer pool.p.Put(c)
+		defer c.Reset(nil, nil)
 		c.Reset(w, r)
 		h(c)
 	})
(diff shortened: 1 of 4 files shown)


beater/request/context_pool.go
```diff
@@ -42,6 +42,7 @@ func (pool *ContextPool) HTTPHandler(h Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		c := pool.p.Get().(*Context)
 		defer pool.p.Put(c)
+		defer c.Reset(nil, nil)
```
simitt, line 45: 👍

simitt (approved):

axw:
@Mergifyio refresh

PR #5719, opened 2021-07-15
processor/otel: handle net.host.* as span attributes

axw:
## Motivation/summary

We were mistakenly handling the new `net.host.*` attributes as resource attributes, whereas they should be span attributes. These fields describe a network connection which will be specific to an operation (span), and not the service (resource) performing that operation.

`net.host.connection_type` has been updated to `net.host.connection.type`, per the spec change.

## Checklist

## How to test these changes

When https://github.com/elastic/apm-agent-ios/pull/46 lands, run a program instrumented with the iOS agent, and check that the information is indexed into `network.*` fields and not as labels. The only field that should show up as a label is `net.host.connection.subtype`, which is not yet mapped.

## Related issues

https://github.com/elastic/apm-server/pull/5436

### DIFF
--- a/processor/otel/consumer.go
+++ b/processor/otel/consumer.go
@@ -65,6 +65,15 @@ const (
 	outcomeSuccess = "success"
 	outcomeFailure = "failure"
 	outcomeUnknown = "unknown"
+
+	// TODO: handle net.host.connection.subtype, which will
+	// require adding a new field to the model as well.
+
+	AttributeNetworkType        = "net.host.connection.type"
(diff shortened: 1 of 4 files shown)


simitt (approved):

simitt:
Running with the latest SNAPSHOT but still seeing the labels
Screenshot 2021-07-16 at 14 38 26

axw:
Probably because https://github.com/elastic/apm-agent-ios/pull/46 hasn't landed yet. I think we should just follow up on this in 7.15, seeing as there's more to do anyway.

PR #5729, opened 2021-07-15
Remove package transform

axw:
## Motivation/summary

Remove package transform, which is no longer needed. Transformation to beat.Events longer requires parameterisation, with all customisable transformations moving to model processing. We could probably also remove the context parameter to the transformable interface, but I've left it there for now as it's not actively causing problems.

Transformation should now be more or less instant, and errors cannot occur, so it is no longer useful to instrument these operations.

## Checklist

## How to test these changes

The only user-visible change is that spans are no longer produced for event transformation. This can be observed by enabling instrumentation (`-E instrumentation.enabled=true`), sending some events to the server, and then checking that no "Transform" spans are produced.

## Related issues


simitt (approved):
🎉

PR #5739, opened 2021-07-16
Restructure model.Metricset

axw:
## Motivation/summary

A few smallish changes to the `model.Metricset` type, for future-proofing. The goal is to make the type suitable for generating transformation to `beat.Events` logic later, while making the type better suited to protobuf definition in the short term.

`model.Metricset.Samples` is now a map, instead of a slice of named metrics, and the metrics fields are added as flat dotted field namess rather than hierarchical objects.

In the future we will send the samples map as-is (with values, type, and unit) to Elasticsearch and perform transformation to top-level metrics fields and dynamic templates completely in the ingest pipeline.

`model.Sample` is now called `model.MetricsetSample`, to clarify its relationship with `model.Metricset`.

We no longer set "_doc_count" in transformation; this is now part of the type and is set in transaction metrics aggregation
for "transaction.duration.histogram", the only metric where it is expected to be set.

## Checklist

## How to test these changes

Non-functional change.

## Related issues

#4120
#3565

### DIFF
--- a/agentcfg/reporter.go
+++ b/agentcfg/reporter.go
@@ -86,9 +86,11 @@ func (r Reporter) Run(ctx context.Context) error {
 		batch := make(model.Batch, 0, len(applied))
 		for etag := range applied {
 			batch = append(batch, model.APMEvent{Metricset: &model.Metricset{
-				Name:    "agent_config",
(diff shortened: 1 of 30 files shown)


simitt (approved):
changes make the coder easier to comprehend - thanks.

Issue #4854, opened 2021-02-22
apmpackage: install a pipeline for migrating pre-datastream indices


axw:
To enable users to view old data in 8.0, we should provide an Ingest node pipeline that can be used for reindexing into data streams, to be used by the Upgrade Assistant (https://github.com/elastic/kibana/issues/91875).

At a minimum the pipeline must route data to the correct data stream, and add `data_stream.*` fields to documents, so the documents end up in the right data stream and schema as produced by APM Server when managed by Fleet.

For 8.0.0 we may also need to add/remove/rename/restructure fields. We are also investigating another option of using runtime fields to querying over data with multiple schema versions.

axw:
Some thoughts on how to go about this.

1. Update [ingest/pipeline/definition.yml](https://github.com/elastic/apm-server/blob/master/ingest/pipeline/definition.yml) to add a new pipeline definition at the end of the file.
2. Test it by running apm-server in standalone mode, and then reindexing events using the new pipeline: https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-reindex.html#reindex-with-an-ingest-pipeline

The new pipeline will need to set `_index` to the appropriate data stream (see https://www.elastic.co/guide/en/elasticsearch/reference/current/ingest.html#access-metadata-fields), and set the `data_stream.*` fields:
 - set `data_stream.namespace` to "migrated" (just an idea, could also just set it to "default")
 - set `data_stream.type` and `data_stream.dataset` according to the value of the `processor.event` field (following logic in apm-server/model):
   - for `processor.event: transaction` and `processor.event: span`, set `data_stream.type: traces` and `data_stream.dataset: apm`
   - for `processor.event: error`, set `data_stream.type: logs` and `data_stream.dataset: apm.error`
   - for `processor.event: profile`, set `data_stream.type: metrics` and `data_stream.dataset: apm.profiling`
   - for `processor.event: metric`, set `data_stream.type: metrics` and set `data_stream.dataset` according to the presence of various fields:
     - if the doc has any of the fields `event`, `transaction`, or `span`, set `data_stream.dataset: apm.internal`
     - otherwise set `data_stream.dataset: apm.app.<normalized_service_name>`, where the `<normalized_service_name>` is the "normalized" value of the `service.name` field following the logic in https://github.com/elastic/apm-server/blob/92f04526831bc471fca92cc189b6f21d1d3159b6/datastreams/servicename.go#L22-L23

[...]

PR #5768, opened 2021-07-20
ingest: introduce apm_data_stream_migration

axw:
## Motivation/summary

Introduce a new ingest pipeline, `apm_data_stream_migration`, which can be used in a reindex operation to set data_stream fields on events and route them to appropriate data streams.

## Checklist

## How to test these changes

1. Run apm-server standalone, send some data of various types
2. Install the "apm" integration (no need to add it to an agent/policy etc., just need the index templates to be created)
3. Reindex `apm-*` using this pipeline, excluding `processor.event: onboarding`
4. Check the events have `data_stream.*` fields added, and they are routed to the correct data streams

## Related issues

Closes https://github.com/elastic/apm-server/issues/4854


stuartnelson3 (approved):
Not experienced with pipelines but the script code makes sense :+1:

stuartnelson3:
Confirmed with BC3

PR #5769, opened 2021-07-20
Update to version 7.13.5

stuartnelson3:
## Motivation/summary

Upgrade to 7.13.5


axw (approved):
I think we can merge this without waiting for CI. It's failing because the stack bump hasn't passed in a while

PR #5770, opened 2021-07-20
[sourcemaps] Support multiple fleet addresses when requesting a sourcemap

### LINKED ISSUE #5514: [sourcemaps] support multiple fleet addresses
The current sourcemap fetching code for fleet assumes that we receive only a single fleet server.

Add support for round-robin requests between the fleet servers. Assume all fleet servers are in the same cluster, ie. an artifact is available on each individual server.

Original comment:
https://github.com/elastic/apm-server/pull/5410#discussion_r655827485

estolfo:
These changes support fetching sourcemaps from one of many fleet servers. It sends the request to all the fleet servers in parallel and returns the first successful response.

I've made some adjustments to the `fleetStore` struct to keep a map of the `(service.name, service.version, bundle.filepath)` combinations to the `sourcemap.url`. I've also added a separate slice to store the list of base fleet server addresses, which represent the configured protocol with the host and port of a given fleet server.

The full url is constructed when the sourcemap is fetched using a base fleet server url and the sourcemap url (path).

Closes #5514 

## How to test these changes

Taking the test plan from https://github.com/elastic/apm-server/pull/5410:

Run elasticsearch and multiple elastic-agents in fleet mode. 2 or 3 should be sufficient for this test.

Start kibana and apm-server with the fleet-server sourcemap code.

- Upload a sourcemap via kibana
- Confirm an updated config with the new source_maps block has been pushed to apm-server
- Confirm ingesting rum traces for the uploaded sourcemap get mapped appropriately

### DIFF
--- a/sourcemap/fleet_store.go
+++ b/sourcemap/fleet_store.go
@@ -24,6 +24,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"sync"
 	"time"
 
 	"github.com/pkg/errors"
@@ -35,10 +36,15 @@ import (
 	logs "github.com/elastic/apm-server/log"
 )
 
+const defaultFleetPort = 8220
+
+var errMsgFleetFailure = errMsgFailure + " fleet"
+
 type fleetStore struct {
-	apikey    string
-	c         *http.Client
-	fleetURLs map[key]string
+	apikey        string
+	c             *http.Client
+	sourceMapURLs map[key]string
+	fleetBaseURLs []string
 }
 
 type key struct {
@@ -71,35 +77,76 @@ func newFleetStore(
 	fleetCfg *config.Fleet,
 	cfgs []config.SourceMapMetadata,
 ) (fleetStore, error) {
-	// TODO(stn): Add support for multiple fleet hosts
-	// cf. https://github.com/elastic/apm-server/issues/5514
-	host := fleetCfg.Hosts[0]
-	fleetURLs := make(map[key]string)
+	sourceMapURLs := make(map[key]string)
+	fleetBaseURLs := make([]string, len(fleetCfg.Hosts))
 
 	for _, cfg := range cfgs {
 		k := key{cfg.ServiceName, cfg.ServiceVersion, cfg.BundleFilepath}
-		u, err := common.MakeURL(fleetCfg.Protocol, cfg.SourceMapURL, host, 8220)
+		sourceMapURLs[k] = cfg.SourceMapURL
+	}
+
+	for i, host := range fleetCfg.Hosts {
+		baseURL, err := common.MakeURL(fleetCfg.Protocol, "", host, defaultFleetPort)
 		if err != nil {
 			return fleetStore{}, err
 		}
-		fleetURLs[k] = u
+		fleetBaseURLs[i] = baseURL
 	}
+
 	return fleetStore{
-		apikey:    "ApiKey " + fleetCfg.AccessAPIKey,
-		fleetURLs: fleetURLs,
-		c:         c,
+		apikey:        "ApiKey " + fleetCfg.AccessAPIKey,
+		fleetBaseURLs: fleetBaseURLs,
+		sourceMapURLs: sourceMapURLs,
+		c:             c,
 	}, nil
 }
 
 func (f fleetStore) fetch(ctx context.Context, name, version, path string) (string, error) {
 	k := key{name, version, path}
-	fleetURL, ok := f.fleetURLs[k]
+	sourceMapURL, ok := f.sourceMapURLs[k]
 	if !ok {
 		return "", fmt.Errorf("unable to find sourcemap.url for service.name=%s service.version=%s bundle.path=%s",
 			name, version, path,
 		)
 	}
 
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	type result struct {
+		sourcemap string
+		err       error
+	}
+
+	results := make(chan result)
+	var wg sync.WaitGroup
+	for _, baseURL := range f.fleetBaseURLs {
+		wg.Add(1)
+		go func(fleetURL string) {
+			defer wg.Done()
+			sourcemap, err := sendRequest(f, ctx, fleetURL)
+			results <- result{sourcemap, err}
+		}(baseURL + sourceMapURL)
+	}
+
+	go func() {
+		wg.Wait()
+		close(results)
+	}()
+
+	var err error
+	for result := range results {
+		err = result.err
+		if err == nil {
+			return result.sourcemap, nil
+		}
+	}
+
+	// All requests resulted in a query failure
+	return "", err
+}
+
+func sendRequest(f fleetStore, ctx context.Context, fleetURL string) (string, error) {
 	req, err := http.NewRequest(http.MethodGet, fleetURL, nil)
 	if err != nil {
 		return "", err
@@ -116,9 +163,9 @@ func (f fleetStore) fetch(ctx context.Context, name, version, path string) (stri
 	if resp.StatusCode != http.StatusOK {
 		body, err := ioutil.ReadAll(resp.Body)
 		if err != nil {
-			return "", fmt.Errorf("failure querying fleet: statuscode=%d response=(failed to read body)", resp.StatusCode)
+			return "", fmt.Errorf(errMsgFleetFailure, ": statuscode=%d response=(failed to read body)", resp.StatusCode)
 		}
-		return "", fmt.Errorf("failure querying fleet: statuscode=%d response=%s", resp.StatusCode, body)
+		return "", fmt.Errorf(errMsgFleetFailure, ": statuscode=%d response=%s", resp.StatusCode, body)
 	}
 
 	// Looking at the index in elasticsearch, currently
--- a/sourcemap/fleet_store.go
+++ b/sourcemap/fleet_store.go
@@ -124,7 +124,7 @@ func (f fleetStore) fetch(ctx context.Context, name, version, path string) (stri
 		wg.Add(1)
 		go func(fleetURL string) {
 			defer wg.Done()
-			sourcemap, err := sendRequest(f, ctx, fleetURL)
+			sourcemap, err := sendRequest(ctx, f, fleetURL)
 			select {
 			case <-ctx.Done():
 			case results <- result{sourcemap, err}:
@@ -152,7 +152,7 @@ func (f fleetStore) fetch(ctx context.Context, name, version, path string) (stri
 	return "", ctx.Err()
 }
 
-func sendRequest(f fleetStore, ctx context.Context, fleetURL string) (string, error) {
+func sendRequest(ctx context.Context, f fleetStore, fleetURL string) (string, error) {
 	req, err := http.NewRequest(http.MethodGet, fleetURL, nil)
 	if err != nil {
 		return "", err
--- a/sourcemap/fleet_store.go
+++ b/sourcemap/fleet_store.go
@@ -148,6 +148,7 @@ func (f fleetStore) fetch(ctx context.Context, name, version, path string) (stri
 	if err != nil {
 		return "", err
 	}
+	// No results were received: context was cancelled.
 	return "", ctx.Err()
 }
 
--- a/sourcemap/fleet_store.go
+++ b/sourcemap/fleet_store.go
@@ -125,7 +125,10 @@ func (f fleetStore) fetch(ctx context.Context, name, version, path string) (stri
 		go func(fleetURL string) {
 			defer wg.Done()
 			sourcemap, err := sendRequest(f, ctx, fleetURL)
-			results <- result{sourcemap, err}
+			select {
+			case <-ctx.Done():
+			case results <- result{sourcemap, err}:
+			}
 		}(baseURL + sourceMapURL)
 	}
 
@@ -142,8 +145,10 @@ func (f fleetStore) fetch(ctx context.Context, name, version, path string) (stri
 		}
 	}
 
-	// All requests resulted in a query failure
-	return "", err
+	if err != nil {
+		return "", err
+	}
+	return "", ctx.Err()
 }
 
 func sendRequest(f fleetStore, ctx context.Context, fleetURL string) (string, error) {
--- a/sourcemap/fleet_store_test.go
+++ b/sourcemap/fleet_store_test.go
@@ -77,4 +77,106 @@ func TestFleetFetch(t *testing.T) {
 	assert.True(t, hasAuth)
 }
 
+func TestMultipleFleetHostsFetch(t *testing.T) {
+	var (
+		hasAuth       bool
+		apikey        = "supersecret"
+		name          = "webapp"
+		version       = "1.0.0"
+		path          = "/my/path/to/bundle.js.map"
+		c             = http.DefaultClient
+		sourceMapPath = "/api/fleet/artifact"
+	)
+
+	h := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		assert.Equal(t, sourceMapPath, r.URL.Path)
+		auth := r.Header.Get("Authorization")
+		hasAuth = auth == "ApiKey "+apikey
+		// zlib compress
+		wr := zlib.NewWriter(w)
+		defer wr.Close()
+		wr.Write([]byte(resp))
+	})
+
+	ts0 := httptest.NewServer(h)
+	defer ts0.Close()
+
+	ts1 := httptest.NewServer(h)
+	defer ts1.Close()
+
+	fleetCfg := &config.Fleet{
+		Hosts:        []string{ts0.URL[7:], ts1.URL[7:]},
+		Protocol:     "http",
+		AccessAPIKey: apikey,
+		TLS:          nil,
+	}
+
+	cfgs := []config.SourceMapMetadata{
+		{
+			ServiceName:    name,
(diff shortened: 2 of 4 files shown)


sourcemap/fleet_store.go
```diff
@@ -71,35 +75,45 @@ func newFleetStore(
...
 	k := key{name, version, path}
-	fleetURL, ok := f.fleetURLs[k]
+	sourceMapURL, ok := f.sourceMapURLs[k]
 	if !ok {
 		return "", fmt.Errorf("unable to find sourcemap.url for service.name=%s service.version=%s bundle.path=%s",
 			name, version, path,
 		)
 	}
 
+	rng := rand.New(rand.NewSource(time.Now().UnixNano()))
```
axw, line 111: Typically you would create a `*rand.Rand` once and reuse it. Did you intentionally choose to create a new one each time? If not, I'd prefer that we create one in `newFleetStore` and store it in the `fleetStore struct. It would need to be protected with a mutex.

Alternatively, we could use the top-level `rand.Intn` function.

sourcemap/fleet_store.go
```diff
@@ -71,35 +75,45 @@ func newFleetStore(
...
+	idx := rng.Int() % len(f.fleetBaseURLs)
+	baseURL := f.fleetBaseURLs[idx]
```
axw, line 113:
```suggestion
	idx := rng.Intn(len(f.fleetBaseURLs))
	baseURL := f.fleetBaseURLs[idx]
```

sourcemap/fleet_store_test.go
```diff
@@ -77,4 +77,62 @@ func TestFleetFetch(t *testing.T) {
...
+	ts1 := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		assert.Equal(t, sourceMapPath, r.URL.Path)
+		auth := r.Header.Get("Authorization")
+		hasAuth = auth == "ApiKey "+apikey
+		// zlib compress
+		wr := zlib.NewWriter(w)
+		defer wr.Close()
+		wr.Write([]byte(resp))
+	}))
+	defer ts1.Close()
```
axw, line 111: seeing as the handler is exactly the same in each case, how about we extract it? i.e. something like:

```go
h := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
	assert.Equal(t, sourceMapPath, r.URL.Path)
	auth := r.Header.Get("Authorization")
	hasAuth = auth == "ApiKey "+apikey
	// zlib compress
	wr := zlib.NewWriter(w)
	defer wr.Close()
	wr.Write([]byte(resp))
})

ts0 := httptest.NewServer(h)
defer ts0.Close()

ts1 := httptest.NewServer(h)
defer ts1.Close()
```

axw: Having said that, we should probably make multiple requests and check which server served the request, to show that the hosts are randomly selected.

sourcemap/fleet_store.go
axw: Another thing: we shouldn't assume that all of the addresses will be accessible to apm-server. In theory there may be a fleet-server in a network that apm-server cannot access.

I think we should probably try each address until we get a success. In that case it might make more sense to use https://pkg.go.dev/math/rand#Perm

axw (commented):
Code looks good overall. I've left a few comments, let me know if you want to sync to discuss.

sourcemap/fleet_store.go
stuartnelson3: Do you want to try each address in series, or in parallel? I'm assuming there will be less than 10 fleet servers to contact, and given the relative infrequency of the lookups, I wouldn't expect resource consumption to be an issue.

axw: Trying them in parallel sounds fine.

Maybe later we can do something a little more elaborate, e.g. keep track of the fastest/most recently available server, and add a small delay before trying others. But we probably don't need to do that initially.

sourcemap/fleet_store.go
```diff
@@ -71,35 +79,74 @@ func newFleetStore(
...
+	sourceMapURL, ok := f.sourceMapURLs[k]
 	if !ok {
 		return "", fmt.Errorf("unable to find sourcemap.url for service.name=%s service.version=%s bundle.path=%s",
 			name, version, path,
 		)
 	}
 
+	idxs := f.rng.Perm(len(f.fleetBaseURLs))
+	respReceived := make(chan string, 1)
+	fetchErrReturned := make(chan error, 1)
```
estolfo, line 118: Does it make sense to limit the channel sizes to 1?

sourcemap/fleet_store_test.go
estolfo: I'm not clear on why it would be necessary to test that a host is selected randomly. I'm using a package function `rand.Perm`, which I can assume works correctly. So if we get a successful response, we can assume we are selecting a host randomly. Let me know if there's something important I'm missing about the way I've written the code.

sourcemap/fleet_store.go
stuartnelson3: In this particular case, I would say it's not necessary, and the way this code is written will actually create channels that can never be drained (and therefore also never GC'd, I think). The channels should also be closed when the function exits.

There are probably more use-cases, but I think I've only used buffered channels when I have a main "coordinating" go-routine that sends off work (eg., a main goroutine sending X messages to a smaller number of worker goroutines) and waits for a response/s.

Since the send on the channels is being done in each goroutine, and therefore not blocking any work, I don't think anything is gained by having a buffered channel, but @axw may have another opinion (or further bits of wisdom).

sourcemap/fleet_store.go
```diff
@@ -71,35 +79,74 @@ func newFleetStore(
...
+	for range idxs {
+		select {
+		case resp := <-respReceived:
+			return resp, nil
+		case err := <-fetchErrReturned:
+			// If the error is not a query failure
+			if !strings.Contains(err.Error(), errMsgFleetFailure) {
+				return "", err
+			}
+		}
```
stuartnelson3, line 142: The way this is written, we'll return the first response, but a later response might be successful.

sourcemap/fleet_store.go
```diff
@@ -71,35 +79,74 @@ func newFleetStore(
...
 		return "", fmt.Errorf("unable to find sourcemap.url for service.name=%s service.version=%s bundle.path=%s",
 			name, version, path,
 		)
 	}
 
+	idxs := f.rng.Perm(len(f.fleetBaseURLs))
+	respReceived := make(chan string, 1)
+	fetchErrReturned := make(chan error, 1)
+
+	for idx := range idxs {
```
stuartnelson3, line 120: Since we're making requests to all the URLs in parallel, is there a reason to use a random number generator?

sourcemap/fleet_store.go
stuartnelson3: Also, if the first response received is successful, we could cancel any remaining requests that are still in-flight.

sourcemap/fleet_store.go
estolfo: I think it's still necessary to use a random number generator: though the go routines are executed concurrently, they are started serially. Over time, I would bet that the tiny difference in start times would become significant and we'd see the go routines started for hosts towards the beginning of the slice finish before those towards the end. I need to research go routines more, but if they are terminated when the surrounding function returns, some fleet hosts with higher slice indexes might never receive a request.
I've also read that go programs don't necessarily use multiple os threads and that the `GOMAXPROCS` default value is 1. So if this were running using a single os thread, the go routines wouldn't be running in parallel and only the fleet hosts with lower index values (or perhaps just the first one) would serve a request.
Let me know if I'm mistaken, I'm still just a go beginner 😃

stuartnelson3: >though the go routines are executed concurrently, they are started serially. Over time, I would bet that the tiny difference in start times would become significant and we'd see the go routines started for hosts towards the beginning of the slice finish before those towards the end.

This could be true! In general, I would assume that the time to start a go routine is a lot lower than the time to finish a network request. I can't find any numbers atm for how long it takes to spawn a go routine, but would assume a network request to take somewhere on the order of milliseconds, and spawning a go routine is probably somewhere in the microseconds realm.

>I need to research go routines more, but if they are terminated when the surrounding function returns, some fleet hosts with higher slice indexes might never receive a request.

Without coordination, go routines exist outside of the lifecycle of the function that starts them (except for the main goroutine, goroutines have no parents or children, in contrast to an os thread or process). In this case, each index will create a goroutine and execute a network request. Since the code below, which reads the result, is exiting as soon as the first response is received (successful or not), the next network request will send to the buffered channel, and any remaining go routines will be stuck forever attempting to send on the full channel.

>I've also read that go programs don't necessarily use multiple os threads and that the GOMAXPROCS default value is 1. So if this were running using a single os thread, the go routines wouldn't be running in parallel and only the fleet hosts with lower index values (or perhaps just the first one) would serve a request.

There may be a single OS thread that performs CPU bound work, but for network i/o the requests will all be started and then the go runtime will most likely context switch to a different, non-waiting go routine (there are a few well-defined places where the go runtime will context switch). Depending on which web request completes first, that go routine will be "woken up" and continue processing.

The default value for `GOMAXPROCS` is the number of CPUs on the system: https://pkg.go.dev/runtime#GOMAXPROCS

axw (commented):
GitHub doesn't want to allow me to continue the thread, so responding to https://github.com/elastic/apm-server/pull/5770#discussion_r676513535 here:

> I'm not clear on why it would be necessary to test that a host is selected randomly. I'm using a package function rand.Perm, which I can assume works correctly. So if we get a successful response, we can assume we are selecting a host randomly. Let me know if there's something important I'm missing about the way I've written the code.

Agreed - we shouldn't be testing that `rand.Perm` is doing its job. Rather we should test the external expectations of this code. Now that we're making the requests in parallel it's not so relevant to test randomisation. If we were issuing them sequentially but in random order, then a test (not necessarily a unit test, could be a functional test) should have picked up the `range` issue I noted, i.e. there was no random ordering at all.

sourcemap/fleet_store.go
```diff
@@ -71,35 +77,81 @@ func newFleetStore(
...
+		err = result.err
+		if err == nil {
+			return result.sourcemap, nil
+		}
+	}
+
+	if err != nil {
+		return "", err
+	}
+	return "", ctx.Err()
```
axw, line 151:
```suggestion
	// No results were received: context was cancelled.
	return "", ctx.Err()
```

sourcemap/fleet_store_test.go
```diff
@@ -77,4 +77,106 @@ func TestFleetFetch(t *testing.T) {
...
+		version       = "1.0.0"
+		path          = "/my/path/to/bundle.js.map"
+		c             = http.DefaultClient
+		sourceMapPath = "/api/fleet/artifact"
+		requestCount  = 0
+	)
+
+	h := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		http.Error(w, "err", http.StatusInternalServerError)
+		requestCount++
```
axw, line 145: this should either use https://pkg.go.dev/sync/atomic#AddInt32, or a mutex; the requests are issued and handled concurrently

sourcemap/fleet_store_test.go
```diff
@@ -77,4 +77,106 @@ func TestFleetFetch(t *testing.T) {
 	assert.True(t, hasAuth)
 }
 
+func TestMultipleFleetHostsFetch(t *testing.T) {
```
axw, line 80: Seeing as we're not testing which server is used, maybe we should fold this into TestFleetFetch above - just update that test to have two servers?

axw (changes requested):
Looking good, thanks!

sourcemap/fleet_store_test.go
```diff
@@ -77,4 +77,106 @@ func TestFleetFetch(t *testing.T) {
...
+	assert.NoError(t, err)
+
+	gotRes, err := f.fetch(context.Background(), name, version, path)
+	require.NoError(t, err)
+
+	assert.Contains(t, gotRes, "webpack:///bundle.js")
+	assert.True(t, hasAuth)
+}
+
+func TestMultipleFleetHostsQueryFailureFetch(t *testing.T) {
```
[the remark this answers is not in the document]
estolfo: ok, I've added a test, if you have a chance to look at it.

sourcemap/fleet_store_test.go
```diff
@@ -77,4 +84,132 @@ func TestFleetFetch(t *testing.T) {
...
+		assert.EqualValues(t, 3, called)
+	case <-time.After(10 * time.Second):
+		t.Fatal("timed out waiting for Fetch to return")
+	}
+
+	select {
+	case <-requestReceived:
+	default:
+		t.Fatal("No second request received by handler")
+	}
```
axw, line 162: I think this is maybe a little bit misleading, and IIANM it's racy.

- since we make the requests in parallel, a second request may received by the handler before the request (context) is cancelled
- even though we wait for `<-fetchReturned` above, we don't wait for all requests to complete in the fetch method if we get a success; so it is possible (if unlikely) that the second request will send to `requestReceived`, and this assertion will fail

I'd suggest changing all the above (line 137+) to:

```go
resp, err := f.fetch(context.Background(), name, version, path)
require.NoError(t, err)
assert.Contains(t, resp, "webpack:///bundle.js")
```

sourcemap/fleet_store_test.go
```diff
@@ -77,4 +84,132 @@ func TestFleetFetch(t *testing.T) {
...
+	select {
+	case <-requestReceived:
+	case <-time.After(10 * time.Second):
+		t.Fatal("timed out waiting for server to receive request")
+	}
+
+	select {
+	case resp := <-fetchReturned:
+		assert.Contains(t, resp, "webpack:///bundle.js")
+		assert.EqualValues(t, 3, called)
```
axw, line 153: there's no guarantee that all request have been received by the time fetch returns, since we return early if we see a successful response from one server

axw (changes requested):

sourcemap/fleet_store_test.go
stuartnelson3: >even though we wait for <-fetchReturned above, we don't wait for all requests to complete in the fetch method if we get a success

Since the successful handler blocks until a read from `requestReceived`, I read this test as verifying that `fetch` returns after it receives its first successful response, even if there are other requests that haven't returned yet. I think the handler is supposed to receive both requests.

> so it is possible (if unlikely) that the second request will send to requestReceived, and this assertion will fail

I think this is what is being tested, that the second request has sent to `requestReceived`. The successful handler is blocked by a send to `requestReceived`, so a single receive on that and then receive from `requestReceived` means `fetch()` returned after its first successful response, with the second request is still waiting within the handler? And the final receive is verifying that both requests were sent to the handler.

The `ctx` in the handler is not the same as the `ctx` in the request, right? So a successful response + `cancel()` of the request ctx won't effect the `ctx` in the server.

estolfo: I agree that this check is unreliable because the request following the successful request might not reach the handler
```go
select {
	case <-requestReceived:
	default:
		t.Fatal("No second request received by handler")
	}
```

But I still don't understand why this is not relevant for testing that the first time the handler is reached, the fetch method has returned successfully?

```go
select {
    case resp := <-fetchReturned:
        assert.Contains(t, resp, "webpack:///bundle.js")
        assert.EqualValues(t, 3, called)
    case <-time.After(10 * time.Second):
        t.Fatal("timed out waiting for Fetch to return")
    }
```

axw: > I think this is what is being tested, that the second request has sent to requestReceived. The successful handler is blocked by a send to requestReceived, so a single receive on that and then receive from requestReceived means fetch() returned after its first successful response, with the second request is still waiting within the handler? And the final receive is verifying that both requests were sent to the handler.
> 
> The ctx in the handler is not the same as the ctx in the request, right? So a successful response + cancel() of the request ctx won't effect the ctx in the server.

@stuartnelson3 right, but the request may never even make it to the 2nd+ servers. Hypothetical ordering:

1. client issues request to ts0, ts1, ts2
2. request is transmitted to ts0
3. ts0 processes request, responds with success
4. client processes response, cancels context
5. because context is cancelled, requests to ts1 and ts2 are never transmitted
6. test fails on second `case <-requestReceived`

It's unlikely, but run it enough times on a slow CI box and it will eventually happen.

> But I still don't understand why this is not relevant for testing that the first time the handler is reached, the fetch method has returned successfully?

@estolfo you're referring to the fetchReturned/timeout bit specifically? The existing `TestFetchContext` test which uses this pattern is doing so because it's testing that asynchronous context cancellation works -- we're not doing that here (at the moment).

I was suggesting the simplification to expedite merging the PR, but I think it would be good to test at this level of detail - it just might take a bit longer. Something like:
 - call fetch in a goroutine (like you're doing)
 - block until all servers have received their request
 - have ts0 respond with failure
 - have ts1 respond with success
 - block until ts2's request is cancelled

axw (approved):

axw:
@Mergifyio backport 7.x

axw:
Tested with BC1, looks good.

- created stack using apm-integration-testing `--with-elastic-agent`
- ran elastic-agent on my laptop with apm integration
- ran a small reverse proxy on two ports, with one returning 500 to each request and the other one passing through to fleet-server
- uploaded a source map to kibana
- send a RUM error to apm-server
- checked that both reverse proxies were queried
- checked that the error was indexed with source mapping applied

Issue #5039, opened 2021-03-31
Introduce support for hot-reloading config without restarting


axw:
At the moment, APM Server will internally restart the HTTP/gRPC server, aggregators, sampler, etc. whenever a configuration change is received. This is implemented by using `libbeat/cfgfile.NewRunnerList`: https://github.com/elastic/apm-server/blob/0e09aa69022be69e15345b21d891ed3ec7c3f067/beater/beater.go#L191

As we progress with integration into Fleet, dynamic reconfiguration will be increasingly common. e.g, support for agent central config (https://github.com/elastic/apm-server/issues/5018) and source maps (https://github.com/elastic/apm-server/issues/5002), as well as future features such as tail-based sampling will all rely on dynamic reconfiguration.

Restarting everything just because agent configuration has changed is clearly not ideal. We should investigate ways to improve this, so we restart only the services that require it. For example, when `apm-server.host` changes, naturally the server needs to restart.

stuartnelson3:
Based on the options from the redesign (https://github.com/elastic/kibana/issues/91412) I've made a list of the config options that should be reloadable. If they require a server restart (such as TLS) that is also noted.

- maximum size of request header
- idle time
- read time
- shutdown time
- write time
- max event size
- custom headers
- max api keys for agent auth
- capture personal data
- service config / default service environment
- golang expvar support (route under /debug/vars)
- simultaneous connections (Triggers restart because it's a wrapper around listener)

RUM:

- enable
- custom headers
- access-control-allow-headers
- custom http response headers
- rate limit per IP
- rate limit cache size
- allowed service names

TLS: should trigger a server restart
- enable
- server cert path
- server cert key path
- supported protocol versions
- cipher suites
- curve types

Not controlled by apm-server config screen:
- agent_config
- sourcemaps

stuartnelson3:
closed by #5911

axw:
Tested in https://github.com/elastic/apm-server/pull/5911.

PR #5787, opened 2021-07-22
add hot-reloading to apm-server

stuartnelson3:
## Motivation/summary

Add hot-reloading for the apm-server. Currently, a full restart happens, which can cause events to be lost. Reload config as much as possible, unless a change requires a restart (eg. modifying TLS).

closes https://github.com/elastic/apm-server/issues/5039

## Checklist


beater/beater.go
```diff
@@ -237,6 +240,7 @@ func (bt *beater) start(ctx context.Context, cancelContext context.CancelFunc, b
 type serverCreator struct {
 	context context.Context
 	args    sharedServerRunnerParams
+	runner  *serverRunner
```
stuartnelson3, line 243: The serverCreator holds a reference to the runner; when it receives a new config it parses that (as it does now), but instead of creating a new runner it will pass the params so it can update it's configuration.

beater/beater.go
```diff
@@ -255,13 +259,20 @@ func (s *serverCreator) Create(p beat.PipelineConnector, rawConfig *common.Confi
...
 		FleetConfig:              &integrationConfig.Fleet,
-	})
+	}
+	if s.runner != nil {
+		// If we've already created a runner, just reconfigure it.
+		return s.runner, s.runner.configure(params)
+	} else {
+		s.runner, err = newServerRunner(s.context, params)
+		return s.runner, err
+	}
```
stuartnelson3, line 275: The serverCreator holds a reference to the runner; when it receives a new config it parses that (as it does now), but instead of creating a new runner it will pass the params so it can update it's configuration.

beater/beater.go
```diff
@@ -310,63 +332,75 @@ type sharedServerRunnerParams struct {
...
+		namespace:              args.Namespace,
+		beat:                   args.Beat,
+		logger:                 args.Logger,
+		tracer:                 args.Tracer,
+		tracerServer:           args.TracerServer,
+		wrapRunServer:          args.WrapRunServer,
+		restartc:               make(chan struct{}),
+		// This is just for testing right now.
+		params: args,
+	}
```
stuartnelson3, line 358: there's a lot of duplication here between setting the initial state and configuring it. ideally this could be minimized. i believe i added it so that configure wouldn't have to do a bunch of nil checks on the incoming config, as well as the struct members on the serverRunner. Could maybe have newServerRunner just return a default serverRunner with default values for all its struct members, which would eliminate needing to pass in arguments.

beater/beater.go
```diff
@@ -310,63 +332,75 @@ type sharedServerRunnerParams struct {
...
+	if args.Logger != nil {
+		*s.logger = *args.Logger
+	}
+	if args.Tracer != nil {
+		*s.tracer = *args.Tracer
+	}
+	if args.TracerServer != nil {
+		*s.tracerServer = *args.TracerServer
+	}
+	s.wrapRunServer = args.WrapRunServer
```
stuartnelson3, line 400: The runServer concept (I expect) still needs to be dealt with -- as @axw mentioned, we might need to have a pre-set chained BatchProcessor and make them configurable as the conf file changes.

beater/beater.go
```diff
@@ -381,31 +415,151 @@ func (s *serverRunner) run() error {
...
+		} else {
+			*s.sourcemapStore = *store
+		}
+	} else {
+		// TODO: We have to remember to disable the sourcemap if it's
+		// no longer enabled
+		s.sourcemapStore = nil
+	}
+
+	if !s.running {
```
stuartnelson3, line 434: This block contains the "one-time" configuration, which potentially should be moved. That might remove the need to have a "running" status for the server.

beater/beater.go
```diff
@@ -381,31 +415,151 @@ func (s *serverRunner) run() error {
...
+}
+
+// Stop begins a timer, after which the server will shutdown. When the server
+// is being restarted, Stop() and then Start() are called, but we cannot
+// differentiate between that and only receiving Stop().
+// Start a timer, and if we receive a message from the Start() method, do not
+// stop the server.
+// The channel in the timer still needs to be drained.
+func (s *serverRunner) Stop() {
+	t := time.NewTimer(10 * time.Second)
```
stuartnelson3, line 505: Reloading is handled by two interfaces:

- There is a struct that "creates" something that can be started and stopped
- There is the thing that starts and stops

I'm assuming we cannot change this interface, so I'm hoping that the start/stop process happens within a reasonable amount of time so we can differentiate between a restart or a shutdown.

I also need to double check that when elastic-agent is shutting down, libbeat only sends a Stop message to the reloadable things it's managing.

@michalpristas can you confirm? When elastic-agent wants to shutdown a managed process, does it simply `Stop()` all of the reloadable items, or does it do something else (like send a unix signal)

beater/beater.go
```diff
@@ -381,31 +415,151 @@ func (s *serverRunner) run() error {
...
+	go func() {
+		select {
+		case <-t.C:
+			s.stopOnce.Do(s.cancelRunServerContext)
+			s.Wait()
```
stuartnelson3, line 510: In the event of a slow restart (ie. the timer elapses) we'll need to recreate the server. We could set some sort of `done` status and have the serverCreator make a new serverRunner.

stuartnelson3: edit: this was added as `runnerStatus`, which we change between an initial, running, and done state. done is set in the event of a stop without a start within the time limit. if a start is sent after this, a new serverRunner is created.

beater/beater.go
```diff
@@ -416,29 +585,51 @@ func (s *serverRunner) run() error {
...
+				if err != nil {
+					fmt.Printf("failed to integration config apm-server.dev.yml! %v\n", err)
+					continue
+				}
+				cfg := integrationConfig.APMServer
+				params := s.params
+				params.RawConfig = cfg
+				s.params = params
+				s.configure(params)
+			case <-s.runServerContext.Done():
```
stuartnelson3, line 613: This exists for local testing of the reload logic within `configure()`

beater/http.go
```diff
@@ -85,24 +114,25 @@ func newHTTPServer(
 		MaxHeaderBytes: cfg.MaxHeaderSize,
 	}
 
+	// TODO: Track TLS settings, if it changes then we need to restart the
+	// server
```
stuartnelson3, line 118: I'm not sure if we need to restart the server or not when changing tls settings in the server. Since we `accept()` on the socket, maybe the tls handshake is done with the current tls config on the server and not a copy of the initial config when starting `server.Serve(listener)`

beater/http.go
```diff
@@ -130,6 +160,7 @@ func (h *httpServer) start() error {
 	}
 
 	if h.cfg.MaxConnections > 0 {
+		// TODO: reload
```
stuartnelson3, line 163: There is some configuration in `httpServer.start()` that is not executed elsewhere, such as how to listen (`Listen` vs `ListenTLS`), which probably would necessitate a server restart.

beater/http.go
stuartnelson3: in `httpServer.start()` there's a branch between `ServeTLS()` and `Serve()`, so turning tls on or off will require a restart, but modifying settings may not.

beater/server.go
```diff
@@ -133,31 +100,34 @@ func newServer(
...
+	// does the serverRunner notice the change and create a new server?
+	// Probably would result in simpler code if the serverRunner handled it.
 	httpServer, err := newHTTPServer(
 		logger, info, cfg, tracer, reporter, batchProcessor, agentcfgFetchReporter, ratelimitStore, sourcemapStore,
 	)
 	if err != nil {
-		return server{}, err
+		return nil, err
 	}
 	grpcServer, err := newGRPCServer(logger, cfg, tracer, batchProcessor, httpServer.TLSConfig, agentcfgFetchReporter, ratelimitStore)
```
stuartnelson3, line 122: grpcServer and jaegerServer still need to be checked out to see what can be reloaded vs. restarted

beater/server.go
```diff
@@ -225,7 +195,59 @@ func newGRPCServer(
...
+		ratelimitStore,
+		sourcemapStore,
+	)
+}
+
+func (s *server) startFetchReporter(ctx context.Context) {
+	s.agentCtx = ctx
+	ctx, s.agentCancel = context.WithCancel(s.agentCtx)
+	go s.agentcfgFetchReporter.Run(ctx)
+}
```
stuartnelson3, line 237: The old fetchReporter needs to be stopped and the new one started when the config is updated because of its internal run loop that batches and sends config updates.

beater/server_test.go
```diff
@@ -513,6 +513,8 @@ func TestServerConfigReload(t *testing.T) {
 	err = reloadable.Reload([]*reload.ConfigWithMeta{{Config: inputConfig}})
 	require.NoError(t, err)
 
+	// TODO: we're not re-listening any more.
+	t.Skip()
```
stuartnelson3, line 517: test needs to be updated because it's testing old behavior

beater/beater.go
axw: The interface between Fleet and libbeat is just one interface: [`libbeat/common/reload.Reloadable`](https://github.com/elastic/beats/blob/master/libbeat/common/reload/reload.go).

We are using a layer on top of this which translates reloads into Stop/Start calls: [`libbeat/cfgfile.Reloader`](https://github.com/elastic/beats/blob/master/libbeat/cfgfile/reload.go). We should probably move off this, and implement Reloadable directly.

axw (commented):
@stuartnelson3 I've only done a very brief look through, as I think this will probably end up changing dramatically.

If we have "beater" implement the Reloadable interface (see other comment) directly, then I imagine we can remove the "creator" and "runner" bits entirely. Beater would be responsible for identifying config that requires a restart of the HTTP and/or gRPC servers, and for reconfiguring processors/auth/RUM enablement/etc.

axw (commented):
I'm having a hard time following the changes because of the shift to the new interface. I think it would be easier to follow (and review) if we split these into two changes: first move to the new Reloadable interface, and then change the reload logic so it doesn't always restart the server. I took the liberty of putting up a draft PR for the first bit: https://github.com/elastic/apm-server/pull/5884

Once we have that, I think it would be helpful to define a struct or something that contains dynamic config, and add a method like `serverRunner.updateDynamicConfig(...) error`. Then in the Reload method we can:
 - extract and remove dynamic config from the args, filling in this struct
 - assume any remaining config is "static" config -- hash it or just encode it as JSON and compare exactly with the existing static config to decide whether to restart

beater/beater.go
```diff
@@ -194,273 +214,363 @@ func (bt *beater) start(ctx context.Context, cancelContext context.CancelFunc, b
...
 			defer closeTracer()
 			if bt.config.ShutdownTimeout > 0 {
 				time.AfterFunc(bt.config.ShutdownTimeout, cancelContext)
 			}
-			inputs.Stop()
+			cancel()
 		}
-		reload.Register.MustRegisterList("inputs", inputs)
-
+		reload.Register.MustRegister("apm-server", bt)
```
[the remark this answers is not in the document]
stuartnelson3: I haven't tested it with fleet, so I would trust your memory

stuartnelson3:
closed in favor of #5897

PR #5803, opened 2021-07-26
correctly set etag when not defined

stuartnelson3:
## Motivation/summary

the calculated etag value wasn't being properly
set on the agent configs. the etag value is handed
to us from kibana so this isn't a big deal, but
when I was testing locally I realized my etag
value was blank, so when I cURL'd the endpoint
without an etag header, they would match (""=="")
and the server would respond with a 304.

## How to test these changes

add agent configs to `apm-server.yml`:

```yml
  agent_config:
  - service.name: ten_percent
    config:
      capture_body: true
      log_level: info
      recording: true
```

curl the server:

```
# curl 'http://localhost:8200/config/v1/agents?service.name=ten_percent'
{
  "capture_body": "true",
  "log_level": "info",
  "recording": "true"
}
```


axw (changes requested):
Could you please add a test that highlights the issue? IIUC, the issue is that we're not calling the `setup` method on the slice element in this code, but rather on a copy:

https://github.com/elastic/apm-server/blob/742e4fa749701da1dbdeb5880e94e428322f6af7/beater/config/config.go#L95-L99

Did you consider just changing that code to index into the slice explicitly, like this?

```go
for i := range c.AgentConfigs {
        if err := c.AgentConfigs[i].setup(); err != nil {
                return nil, err
        }
}
```

stuartnelson3:
I did not! Would that be your preference?

axw:
Unlikely to matter here, but yes I would typically prefer having a slice of non-pointer types if possible - better memory locality. Smaller diff in this case :)

stuartnelson3:
cool, :heavy_check_mark:

axw (approved):

axw:
Tested with BC1, looks good.

I ran apm-server standalone with apm-integration-testing, adding the following args per the description:

```
"-E",
"apm-server.agent_config=[{'service.name':'ten_percent','config':{capture_body:true,log_level:info,recording:true}}]"
```

Then curl'd the config endpoint:

```
$ curl -v 'http://localhost:8200/config/v1/agents?service.name=ten_percent'
> GET /config/v1/agents?service.name=ten_percent HTTP/1.1
> Host: localhost:8200
> User-Agent: curl/7.68.0
> Accept: */*
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< Access-Control-Expose-Headers: Etag
< Cache-Control: max-age=30, must-revalidate
< Etag: "7b534bef04ed8f93f4f936d653b14fb3"
< X-Content-Type-Options: nosniff
< Date: Tue, 24 Aug 2021 10:15:53 GMT
< Content-Length: 75
< 
{
  "capture_body": "true",
  "log_level": "info",
  "recording": "true"
}
```

```
$ curl -v -H "If-None-Match: 7b534bef04ed8f93f4f936d653b14fb3" 'http://localhost:8200/config/v1/agents?service.name=ten_percent
> GET /config/v1/agents?service.name=ten_percent HTTP/1.1
> Host: localhost:8200
> User-Agent: curl/7.68.0
> Accept: */*
> If-None-Match: 7b534bef04ed8f93f4f936d653b14fb3
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 304 Not Modified
< Access-Control-Expose-Headers: Etag
< Cache-Control: max-age=30, must-revalidate
< Etag: "7b534bef04ed8f93f4f936d653b14fb3"
< X-Content-Type-Options: nosniff
< Date: Tue, 24 Aug 2021 10:17:34 GMT
<
```

PR #5854, opened 2021-08-02
Remove sourcemap/test and elasticsearch/estest

axw:
## Motivation/summary

Remove unnecessarily fragmented packages used for testing source maps, moving the code directly into the sourcemap package:

- elasticsearch/estest was only used inside the sourcemap package's tests; remove this package and refactor tests to use an httptest.Server, exercising the code in much the same way as in a real apm-server
- sourcemap/test had two purposes: providing a sourcemaps for tests, and providing elasticsearch.Clients that would respond with canned responses; remove sourcemap/test, canned responses are now handled per above, sourcemaps used for testing are moved into the sourcemap_test package, and the beater test that was using sourcemap/test now uses testdata/sourcemap instead

## Checklist

## How to test these changes

Non-functional change.

## Related issues

Yak shaving related to cherry-picking go-elasticsearch product check changes into https://github.com/elastic/apm-server/pull/5838


stuartnelson3 (approved):
:100: awesome!

PR #5855, opened 2021-08-02
[7.x] sourcemap: fix data race in test (#5842)

axw:
 - sourcemap: fix data race in test (#5842)


axw:
run elasticsearch-ci/docs

PR #5875, opened 2021-08-03
docs: update transaction metrics

axw:
## Motivation/summary

Update apm-server.aggregation.transactions.* docs:

- `enabled` defaults to true since 7.13
- `max_groups` defaults to 10000 since 7.10
- `rum.user_agent.lru_size` was dropped (I forget when)

Also change some of the wording around the purpose of these metrics:
they are primarily about reducing storage costs, not improving query
performance. The latter is expected to eventually be the case, but
has not yet been proven and is not hte primary goal.

### DIFF
--- a/docs/transaction-metrics.asciidoc
+++ b/docs/transaction-metrics.asciidoc
@@ -33,7 +33,7 @@ apm-server:
 ==== `enabled`
 
 Enables the collection and publishing of transaction metrics.
-Enabling this setting the removes the need to store unsampled transactions, reducing storage costs.
+Enabling this setting removes the need to store unsampled transactions, reducing storage costs.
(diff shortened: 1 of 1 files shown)


docs/transaction-metrics.asciidoc
```diff
@@ -33,9 +33,10 @@ apm-server:
 ==== `enabled`
 
 Enables the collection and publishing of transaction metrics.
-This setting improves query performance in the APM app.
+Enabling this setting the removes the need to store unsampled transactions, reducing storage costs.
```
bmorelli25, line 36:
```suggestion
Enabling this setting removes the need to store unsampled transactions, reducing storage costs.
```

bmorelli25 (approved):
LGTM

Issue #5807, opened 2021-07-27
ingest: apm_convert_destination_address is incompatible with Elasticsearch 7.12.0


axw:
In https://github.com/elastic/apm-server/pull/5695 I removed some code that sets `destination.ip` from `destination.address`, and added an ingest node pipeline that does this instead. The pipeline uses `convert`, but support for the IP target field type was only added in 7.13, so this fails with older versions of Elasticsearch.

Issue #5786, opened 2021-07-22
server setup - ingest pipeline apm_error_grouping_name - [value] required property is missing


domanq:
The APM Server 7.13.4 setup fails at "PUT https://...:9200/_ingest/pipeline/apm_error_grouping_name ..."  with Elasticsearch 7.9.3. 
The pipeline definition uses copy_from, which is not supported by this version of Elastic.
According to the compatibility matrix it should be: Elasticsearch 7.9.x -> APM Server 7.0.x-7.13.x.

axw:
Thanks for the bug report -- that was an oversight which we'll fix.

PR #5876, opened 2021-08-03
ingest/pipeline: fix compat with old Elasticsearch

axw:
## Motivation/summary

Fix compatibility with older Elasticsearch versions:

 - don't use set.copy_from, it's not supported before 7.11
 - don't use convert, converting IPs was only added in 7.13

## Checklist

## How to test these changes

Run apm-server with 7.0.0

## Related issues

Closes https://github.com/elastic/apm-server/issues/5786
Closes https://github.com/elastic/apm-server/issues/5807


stuartnelson3 (approved):

simitt:
Cannot be tested with BC3 as a recent change in libbeat breaks compatibility with older ES versions, see https://github.com/elastic/beats/pull/27107/#issuecomment-909322023.

simitt:
Tested with BC3: it requires to set `script.painless.regex.enabled=true` in ES but then the pipelines can be successfully loaded.

PR #5892, opened 2021-08-04
Conditionally restart apm-server

stuartnelson3:
## Motivation/summary

Separate out dynamic and static config options, and indicate whether the server should be restarted based on this.

## Checklist

~~- [ ] Update [CHANGELOG.asciidoc](https://github.com/elastic/apm-server/blob/master/CHANGELOG.asciidoc)~~

## How to test these changes

Non-functional change

## Related issues

#5039


beater/beater.go
```diff
@@ -256,26 +261,84 @@ func (r *reloader) Reload(configs []*reload.ConfigWithMeta) error {
...
+	cfg.DefaultServiceEnvironment = ""
+	cfg.AgentConfigs = nil
+	// Does our static config match the previous static config? If not,
+	// then we should restart.
+	m, err := json.Marshal(cfg)
+	if err != nil {
+		return dynamicConfig{}, false, err
+	}
+
+	shouldRestart := !reflect.DeepEqual(m, r.staticConfig)
```
axw, line 337:
```suggestion
	shouldRestart := !bytes.Equal(m, r.staticConfig)
```

beater/beater.go
```diff
@@ -431,6 +494,34 @@ func (s *serverRunner) run() error {
 	return publisher.Stop(s.backgroundContext)
 }
 
+type dynamicConfig struct {
```
axw, line 497: I think I led you astray: this mirrors beater/config, so probably not very maintainable. How about instead we just zero out the dynamic config, like you're doing already, to check if a restart is needed. If so restart, and otherwise call updateDynamicConfig with the `*beater/config.Config`?

axw (changes requested):

beater/beater.go
stuartnelson3: I think this is a solid improvement :ok_hand:

axw (approved):

PR #5896, opened 2021-08-05
aggregation/spanmetrics: handle composite spans

axw:
## Motivation/summary

Update spanmetrics to take into account composite spans, using the composite span count and duration. Like non-composite spans, these are also multiplied by the "representative count", i.e. the inverse sampling ratio.

## Checklist

## How to test these changes

1. Send some composite spans to the server
2. Check that the span metrics have the expected count and duration sum

## Related issues

Closes https://github.com/elastic/apm-server/issues/5595

### DIFF
--- a/changelogs/head.asciidoc
+++ b/changelogs/head.asciidoc
@@ -19,6 +19,7 @@ https://github.com/elastic/apm-server/compare/7.13\...master[View commits]
 
 [float]
 ==== Added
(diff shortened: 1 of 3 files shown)


stuartnelson3 (approved):

stuartnelson3:
Confirmed with BC2, in conjunction with #5661

PR #5897, opened 2021-08-05
Implement reload

stuartnelson3:
## Motivation/summary

Support config reloading in apm-server

## Checklist

## How to test these changes

Go through [this list](https://github.com/elastic/apm-server/issues/5039#issuecomment-889228676) and verify that modifications to the config file have been picked up and applied.

## Related issues

https://github.com/elastic/apm-server/issues/5039

### DIFF
--- a/beater/beater.go
+++ b/beater/beater.go
@@ -25,9 +25,11 @@ import (
 	"net"
 	"net/http"
 	"os"
+	"os/signal"
 	"runtime"
 	"strings"
 	"sync"
+	"syscall"
 	"time"
 
 	"github.com/elastic/beats/v7/libbeat/common/fleetmode"
@@ -214,6 +216,33 @@ func (bt *beater) start(ctx context.Context, cancelContext context.CancelFunc, b
 		if err := reloader.reload(bt.rawConfig, "default", nil); err != nil {
 			return nil, err
 		}
+		go func() {
+			sigs := make(chan os.Signal, 1)
+			defer close(sigs)
+			signal.Notify(sigs, syscall.SIGUSR1)
+			for {
+				select {
+				case <-sigs:
+					rawcfg, err := common.LoadFile("./apm-server.dev.yml")
+					if err != nil {
+						fmt.Printf("failed to load apm-server.dev.yml! %v\n", err)
+						continue
+					}
+					integrationConfig, err := config.NewIntegrationConfig(rawcfg)
+					if err != nil {
+						fmt.Println("failed to get apm config:", err)
+						continue
+					}
+					apmServerCommonConfig := integrationConfig.APMServer
+					if err := reloader.reload(apmServerCommonConfig, "default", nil); err != nil {
+						fmt.Println("got err while reloading:", err)
+						continue
+					}
+				case <-ctx.Done():
+					return
+				}
+			}
+		}()
 	}
 	return done, nil
 }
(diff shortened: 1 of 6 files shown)


beater/beater.go
```diff
@@ -214,6 +216,33 @@ func (bt *beater) start(ctx context.Context, cancelContext context.CancelFunc, b
...
+					apmServerCommonConfig := integrationConfig.APMServer
+					if err := reloader.reload(apmServerCommonConfig, "default", nil); err != nil {
+						fmt.Println("got err while reloading:", err)
+						continue
+					}
+				case <-ctx.Done():
+					return
+				}
+			}
+		}()
```
stuartnelson3, line 245: for local testing

beater/server.go
```diff
@@ -218,14 +194,72 @@ func newGRPCServer(
...
+	s.agentcfgFetchReporter = agentcfg.NewReporter(agentcfg.NewFetcher(cfg), batchProcessor, 30*time.Second)
+	ratelimitStore, err := ratelimit.NewStore(
+		cfg.AgentAuth.Anonymous.RateLimit.IPLimit,
+		cfg.AgentAuth.Anonymous.RateLimit.EventLimit,
+		3, // burst multiplier
+	)
+	if err != nil {
+		return err
+	}
+	*s.cfg = *cfg
```
axw, line 222: This looks like it'll create data races. What's synchronising access to the config?

axw (commented):
For something that can be hot-reloaded without restarting each of the components (HTTP+gRPC server, agent config fetcher, aggregation, tail-based sampling, ...), we're going to need some synchronisation to avoid data races.

What I'm imagining is something like this:

```go
// ServerParams holds parameters for running the APM Server.
type ServerParams struct {
...

        // Config returns a channel over which the APM Server config can be received.
        //
        // The initial config will be available in the channel immediately, with the possibility
        // of dynamic config updates being received later.
        Config func() <-chan *config.Config
...
}
```

Then we would need to have each of the components call this function to obtain a channel for watching config, and update themselves independently.

Easier said than done though :)

What do you think about paring all of this back, and going back to something _similar_ but a little better than what we were doing before. On config change:
 - start a whole new "server"
 - gracefully shutdown the old one

We should be able to do this with zero downtime by using SO_REUSEPORT (e.g. using https://pkg.go.dev/github.com/libp2p/go-reuseport)

beater/beater.go
stuartnelson3: Is this something we want or think would be valuable? Happy to make the suggested improvements, just not sure if this is something we want in the project.

stuartnelson3:
>We should be able to do this with zero downtime by using SO_REUSEPORT (e.g. using https://pkg.go.dev/github.com/libp2p/go-reuseport). It still wouldn't be ideal as this would mean resetting aggregation and sampling, but we weren't planning to tackle that initially anyway -- we could try doing something more ambitious later.

Happy to go this route, but I originally dismissed it since it only targets (to my knowledge) linux, and while it's been around for a while (since 3.9 it seems https://lwn.net/Articles/542629/), there will be customers running older machines that fail on this :) if that's fine then I say we do it.

beater/beater.go
axw: I'd prefer not to add it right now. If you were planning to remove it before merging then just dismiss my comment

stuartnelson3:
Ah ok, just read the code in https://pkg.go.dev/github.com/libp2p/go-reuseport, looks like it silently fails if SO_REUSEPORT isn't available

axw:
> Happy to go this route, but I originally dismissed it since it only targets (to my knowledge) linux, and while it's been around for a while (since 3.9 it seems https://lwn.net/Articles/542629/), there will be customers running older machines that fail on this :) if that's fine then I say we do it.

Good point about Linux < 3.9, I had forgotten about that. Looking at our [support matrix](https://www.elastic.co/support/matrix), Elastic Agent doesn't support any Linux distros that use such an old kernel -- even if apm-server does. We only need reload support when running under Elastic Agent.

Windows does have `SO_REUSEADDR`, and that does roughly the same thing (and is used by go-reuseport).

stuartnelson3:
closed in favor of #5911

PR #5903, opened 2021-08-08
Move setting destination.ip back into apm-server

### LINKED ISSUE #5898: destination.ip not always an IP when set by pipeline
**APM Server version** 8.0.0-SNAPSHOT ~, likely 7.14.0+~

When ingesting a span containing an `destination.address` that is not an IP apm-server will accept the payload but fail to index with:

```
Cannot index event publisher.Event ... "failed to parse field [destination.ip] of type [ip] ... 'XXX' is not an IP string
```

It appears this is a regression introduced in elastic/apm-server#5876 where the previous processor filtered out addresses that did not parse to IPs.

axw:
## Motivation/summary

Revert the addition of the `apm_convert_destination_address` ingest pipeline, and set the `destination.ip` field during
model transformation. We can move it back to ingest node once we can rely on Elasticsearch supporting the `ip` type in the `convert` processor. While we're at it, move Destination out of Span, and up to APMEvent.

## Checklist

## How to test these changes

Send some spans which result in `destination.address` being set -- one with an IP value, one with a hostname. Both should be indexed, and `destination.address` should also be copied to `destination.ip` in the IP value case.

## Related issues

Fixes #5898


stuartnelson3 (approved):

stuartnelson3:
confirmed with bc3

PR #5911, opened 2021-08-09
Reuse port

stuartnelson3:
## Motivation/summary

Implement zero-downtime config reloading by configuring socket with `SO_REUSEPORT`. On a reload, start a new server, and once we've been signaled it has started, tear down the old server.

## Checklist

## How to test these changes

1. Start the apm-server managed by fleet
2. In a separate terminal tab, curl an apm-server endpoint in an infinite loop, like an agent config endpoint
3. Update the agent config
4. Check that no requests in the infinite loop have failed, and that the response reflects the config change

## Related issues

https://github.com/elastic/apm-server/issues/5039


beater/server.go
```diff
@@ -235,6 +235,7 @@ func (s server) run() error {
 	if s.jaegerServer != nil {
 		g.Go(s.jaegerServer.Serve)
 	}
+	runningc <- struct{}{}
```
axw, line 238: We don't really know that the listener has been created yet. It's possible that `s.httpServer.start` on line 231 hasn't been scheduled yet, or that the listener it creates hasn't been created, by the time we signal `runningc`.

How about we move the listener creation out of `httpServer.start` and into the `newHTTPServer` function? That way when `newBaseRunServer` returns we know the listener has been created.

beater/beater.go
```diff
@@ -261,81 +256,32 @@ func (r *reloader) Reload(configs []*reload.ConfigWithMeta) error {
...
-		return false, err
-	}
-	key, err := json.Marshal(m)
-	if err != nil {
-		return false, err
+	// If the old runner exists, cancel it
+	if r.runner != nil {
+		r.runner.cancelRunServerContext()
+		<-r.runner.done
+		r.runner = nil
```
axw, line 281:
```suggestion
```

axw (changes requested):
Very nice! Looks good, just one issue related to the signalling

beater/beater.go
```diff
@@ -261,81 +256,31 @@ func (r *reloader) Reload(configs []*reload.ConfigWithMeta) error {
 func (r *reloader) reload(rawConfig *common.Config, namespace string, fleetConfig *config.Fleet) error {
 	r.mu.Lock()
 	defer r.mu.Unlock()
-	shouldRestart, err := r.shouldRestart(rawConfig)
+	runningc := make(chan struct{})
+	defer close(runningc)
```
axw, line 260:
```suggestion
```

shouldn't need the channel or the select below any more

axw (approved):
LGTM with runningc removed!

axw:
I tested with BC2, and I don't think it's quite working right.

- Started elastic-agent, added APM integration
- curl'd healthcheck endpoint in a loop
- Updated apm integration policy

When I did this, curl failed with:

```
Recv failure: Connection reset by peer
* Closing connection 0
curl: (56) Recv failure: Connection reset by peer
```

PR #5928, opened 2021-08-11
Add `data_streams.wait_for_integration`

axw:
## Motivation/summary

Add `apm-server.data_streams.wait_for_integration` config. When both this and `apm-server.data_streams.enabled` are true, the server will wait at startup until the APM integration package is installed. To do this, APM Server requires `apm-server.kibana` to be configured, and will query the Fleet API to query the integration package status.

## Checklist

- [ ] Update [CHANGELOG.asciidoc](https://github.com/elastic/apm-server/blob/master/CHANGELOG.asciidoc)
- [x] Documentation has been updated

For functional changes, consider:
- Is it observable through the addition of either **logging** or **metrics**?
- Is its use being published in **telemetry** to enable product improvement?
- Have system tests been added to avoid regression?

## How to test these changes

1. Create a fresh cluster with no integration isntall
2. Run apm-server with `apm-server.data_streams.enable=true` (`wait_for_integration` defaults to true)
3. Observe that apm-server logs that it is waiting for the integration, and it does not start listening on 8200
4. Install the apm integration package in Kibana
5. Observe that apm-server continues, and starts listening on 8200
6. Send some events
7. Check that they are indexed into data streams

## Related issues

https://github.com/elastic/apm-server/issues/5914

Issue #5914, opened 2021-08-10
Introduce standalone data stream support, deprecate Elasticsearch setup


axw:
APM Server is currently responsible for installing Elasticsearch (legacy) index templates, ingest pipelines, and ILM policies. We now also have the Fleet integration which installs variants of these, with the index templates creating `traces-*`, `logs-*`, and `metrics-*` data streams instead of `apm-*` indices.

We intend to increasingly rely on Fleet and Elastic Agent for powering features such as APM agent central config and source map distribution. In the future, we will remove the ability to write to `apm-*` indices and always create data streams.

We should introduce an option for standalone APM Server to start writing data streams, relying on Fleet (Kibana) to set up the integration package. When this option is enabled, APM Server will wait for the integration package to be installed before writing any data, and will write to the data stream names rather than legacy index names.

I propose that we introduce the option no later than 7.last, where it will be off by default. At the same time we will deprecate Elasticsearch setup.
In 8.0 we will _remove_ the ability to setup Elasticsearch, and turn the option on by default (https://github.com/elastic/apm-server/issues/5961). It should still be possible to disable the option for scenarios where APM Server lacks privileges to check the package is installed; in this case the user is responsible for ensuring the package is installed before APM Server is started.

axw:
Related: https://github.com/elastic/apm-server/issues/5950 https://github.com/elastic/apm-server/issues/5949

PR #5934, opened 2021-08-12
Wait for integration using Elasticsearch queries

axw:
## Motivation/summary

Extend `data_streams.wait_for_integration` to use either Kibana or Elasticsearch queries, depending on what is configured.

If Kibana config is defined, and Kibana can be successfully queried, then it will be used. Otherwise if the Elasticsearch output is configured, Elasticsearch queries will be made to check the presence of index templates which imply the integration has been installed.

## Checklist

## How to test these changes

See https://github.com/elastic/apm-server/pull/5928; run with and without `apm-server.kibana` config.

## Related issues

#5914

### DIFF
--- a/beater/waitintegration.go
+++ b/beater/waitintegration.go
@@ -53,6 +53,9 @@ func waitForIntegration(
 	var ticker *time.Ticker
 	for {
 		if ticker == nil {
+			// We start the ticker on the first iteration, rather than
+			// before the loop, so we don't have to wait for a tick
+			// (5 seconds by default) before peforming the first check.
 			ticker = time.NewTicker(interval)
 			defer ticker.Stop()
 		} else {
@@ -63,7 +66,6 @@ func waitForIntegration(
 			}
 		}
 		if kibanaClient != nil {
-			//logger.Errorf("error querying integration package status: %s", err)
 			installed, err := checkIntegrationInstalledKibana(ctx, kibanaClient, logger)
 			if err != nil {
 				logger.Errorf("error querying Kibana for integration package status: %s", err)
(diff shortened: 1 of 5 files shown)


beater/waitintegration.go
```diff
@@ -0,0 +1,148 @@
...
+			defer ticker.Stop()
+		} else {
+			select {
+			case <-ctx.Done():
+				return ctx.Err()
+			case <-ticker.C:
+			}
+		}
+		if kibanaClient != nil {
+			//logger.Errorf("error querying integration package status: %s", err)
```
stuartnelson3, line 66: looks like this one snuck in

beater/waitintegration.go
```diff
@@ -0,0 +1,148 @@
...
+	templates := []string{
+		"traces-apm",
+		"traces-sampled",
+		"metrics-apm.app",
+		"metrics-apm.internal",
+		"logs-apm.error",
+	}
+	// IndicesGetIndexTemplateRequest accepts a slice of template names,
+	// but the REST API expects just one index template name. Query them
+	// in parallel.
```
stuartnelson3, line 127: :+1:

beater/waitintegration.go
```diff
@@ -0,0 +1,148 @@
...
+	for {
+		if ticker == nil {
+			ticker = time.NewTicker(interval)
+			defer ticker.Stop()
+		} else {
+			select {
+			case <-ctx.Done():
+				return ctx.Err()
+			case <-ticker.C:
+			}
```
stuartnelson3, line 63: I paused for a moment to think about why the code is written this way before realizing it's to send a request immediately, and then wait on subsequent loops. Could you write a short comment for Future Stuart:tm:?

stuartnelson3 (approved):
looks good :+1:

beater/waitintegration.go
axw: :+1: good call

stuartnelson3:
Issue filed: https://github.com/elastic/apm-server/issues/6078

stuartnelson3:
Confirmed with BC3

Issue #5707, opened 2021-07-14
Define component templates for APM


simitt:
Fleet creates a custom component template by default for every datastream. Component templates can only overwrite settings and mappings that are not part of the `_index_template`'s settings and mappings. Move all settings that should be changable by users into the component template definition in the apmpackage. 

For example in the [manifest](https://github.com/elastic/package-storage/blob/production/packages/apm/0.2.0/data_stream/traces/manifest.yml) define `dynamic: false` (to avoid dynamic indexing for unmapped fields) and add some index setting
```
elasticsearch:
  index_template:
    settings:
      number_of_replicas: 2
    mappings:
      dynamic: false
```

With the `custom` component, users can then define different settings and additional mappings.

---
POC
* Add the above mentioned `elasticsearch.index_template` to the `traces` manifest
* Start an environment that is consuming your local package (e.g. `./scripts/compose.py start master --with-elastic-agent --package-registry-url=http://host.docker.internal:8080`), set up the apm package and add it to the agent policy
* add a different index setting to the `traces-apm@custom` component template and add some field mapping
* ingest some data (if you ingested data before, call the rollover API to ensure a new index is created `POST traces-apm-default/_rollover`)
* check the index settings and mappings (e.g. `GET .ds-traces-apm-default-2021.07.14-000003/_settings`)

---
related Kibana bug https://github.com/elastic/kibana/issues/105603

Mpdreamz:
Linking this ES feature request for awareness both ways: https://github.com/elastic/elasticsearch/issues/75031

axw:
I started looking into this, but it's unclear to me what the expected outcome is. If you specify index_template like in the description, for `traces-apm-*`, you end up with two additional component templates:

**traces-apm@mappings**

```
{
  "dynamic": false
}
```

**traces-apm@settings**

```
{
  "index": {
    "number_of_replicas": "2"
  }
}
```

- those properties are not defined in the index template anyway, so why are we doing this?
- AFAICT the addition of these component templates does not affect the index template at all, so I don't see how we can "Move all settings that should be changable by users into the component template definition in the apmpackage." without changes to Fleet

What am I missing?

Mpdreamz:
It should create 4 component templates and link them at install time:

https://github.com/elastic/kibana/blob/3864fe1559281b4a2731e3a9439c501125e65669/x-pack/test/fleet_api_integration/apis/epm/install_overrides.ts#L47

[...]

PR #5947, opened 2021-08-16
apmpackage: copy full field hierarchy

axw:
## Motivation/summary

Fix integration package index template mappings by copying the full field hierarchy to the integration package, instead of just flattened leaf fields. Copying only leaf fields means we miss inherited group properties such as `dynamic`.

- set a default `dynamic` property for each data stream mapping
- add `ecs.version` to base fields

## Checklist

## How to test these changes

1. Install the integration
2. Run apm-server with data streams (e.g. `-E apm-server.data_streams.enabled=true`)
3. Send some HTTP transactions with request headers, check that APM Server indexes them successfully
4. Check that the request headers are not indexed

## Related issues

https://github.com/elastic/apm-server/issues/5707

### DIFF
--- a/apmpackage/apm/data_stream/internal_metrics/manifest.yml
+++ b/apmpackage/apm/data_stream/internal_metrics/manifest.yml
@@ -2,3 +2,9 @@ title: APM internal metrics
 type: metrics
 dataset: apm.internal
 ilm_policy: metrics-apm.internal_metrics-default_policy
+elasticsearch:
+  index_template:
+    mappings:
+      # Internal metrics should have all fields strictly mapped;
+      # we are in full control of the field names.
+      dynamic: strict
--- a/apmpackage/apm/data_stream/sampled_traces/manifest.yml
+++ b/apmpackage/apm/data_stream/sampled_traces/manifest.yml
@@ -9,3 +9,7 @@ elasticsearch:
       # global checkpoints as a way of limiting search
       # results.
       number_of_shards: 1
+    mappings:
+      # Sampled traces should have all fields strictly mapped;
+      # we are in full control of the field names.
+      dynamic: strict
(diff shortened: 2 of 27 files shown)


apmpackage/apm/data_stream/internal_metrics/manifest.yml
```diff
@@ -2,3 +2,9 @@ title: APM internal metrics
 type: metrics
 dataset: apm.internal
 ilm_policy: metrics-apm.internal_metrics-default_policy
+elasticsearch:
+  index_template:
+    mappings:
+      # Internal metrics should have all fields strictly mapped;
+      # we are in full control of the field names.
+      dynamic: strict
```
simitt, line 10: That would potentially lead to errors if the APM Server version > package version containing the templates. I'd be careful with using `strict` as long as the automatic package upgrade is not in place.

apmpackage/apm/data_stream/sampled_traces/manifest.yml
```diff
@@ -9,3 +9,7 @@ elasticsearch:
       # global checkpoints as a way of limiting search
       # results.
       number_of_shards: 1
+    mappings:
+      # Sampled traces should have all fields strictly mapped;
+      # we are in full control of the field names.
+      dynamic: strict
```
simitt, line 15: same comment as above

apmpackage/apm/data_stream/internal_metrics/manifest.yml
axw: Discussed offline, we'll plan to require Kibana >= APM Server, and rely on auto-upgrade: https://github.com/elastic/apm-server/issues/5948. If that doesn't pan out then we can revisit this.

simitt (approved):

simitt:
Tested with BC3 - works as expected with package `0.4.0`

PR #5953, opened 2021-08-16
docs: Change API key docs to Kibana UI

bmorelli25:
### Motivation/summary

This PR updates API key documentation to use the Kibana UI instead of ES APIs or APM Server tools.

### Let me see it

* [Create an API key for agent/server comms](https://apm-server_5953.docs-preview.app.elstc.co/guide/en/apm/server/master/api-key.html)
* [Grant access to ES resources with API keys](https://apm-server_5953.docs-preview.app.elstc.co/guide/en/apm/server/master/beats-api-keys.html)

### Related issues

Closes https://github.com/elastic/observability-docs/issues/965.


docs/copied-from-beats/docs/security/api-keys.asciidoc
```diff
@@ -19,71 +19,114 @@ IMPORTANT: Review <<feature-roles>> before creating API keys for {beatname_uc}.
...
-          "names": ["{beat_default_index_prefix}-*"],
-          "privileges": ["view_index_metadata", "create_doc"]
-        }
+    "{beat_default_index_prefix}_writer": {
+        "cluster": ["monitor", "read_ilm"], <1>
+        "index": [ <2>
+            {
+                "names": ["{beat_default_index_prefix}-*"],
+                "privileges": ["view_index_metadata", "create_doc"]
+            }
```
bmorelli25, line 41: Question for reviewers. Are these permissions correct for a **publishing** user? I didn't change anything here, but I did notice that our [**Grant privileges and roles needed for writing events**](https://www.elastic.co/guide/en/apm/server/current/privileges-to-publish-events.html) documentation shows different req'd privs.

docs/copied-from-beats/docs/security/api-keys.asciidoc
```diff
@@ -19,71 +19,114 @@ IMPORTANT: Review <<feature-roles>> before creating API keys for {beatname_uc}.
...
+[source,json,subs="attributes,callouts"]
+----
+{
+    "{beat_default_index_prefix}_monitoring": {
+        "cluster": ["monitor"], <1>
+        "index": [ <2>
+            {
+                "names": [".monitoring-beats-*"],
+                "privileges": ["create_index", "create"]
+            }
```
bmorelli25, line 87: Same question here. What we had is different from what we show in [**Grant privileges and roles needed for monitoring**](https://www.elastic.co/guide/en/apm/server/current/privileges-to-publish-monitoring.html).

docs/secure-communication-agents.asciidoc
```diff
@@ -31,30 +31,37 @@ include::./ssl-input.asciidoc[]
...
-. Make sure <<ssl-setup,SSL/TLS>> is enabled
 . <<configure-api-key,Enable and configure API keys>>
-. <<create-api-key,Create an API key with APM Server>>
-. <<set-api-key,Set the API key in your APM Agents>>
+. <<create-api-key,Create an API key in {kib}>>
+. <<set-api-key,Set the API key in your APM agents>>
 
-NOTE: API Keys are not applicable for the RUM Agent,
-as there is no way to prevent them from being publicly exposed.
+NOTE: API Keys
```
simitt, line 57: leftover?

docs/copied-from-beats/docs/security/api-keys.asciidoc
simitt: Good catch! The linked docs have a NOTE at the beginning that the `setup` role is separated from the `writer` role to minimize the privileges required during writing. That's why the dedicated writer role has less privileges. But even with that, there are still discrepancies.

I'd only put a minimal example here, using the same privileges as documented for the writer role, and add a note that this assumes previous setup and link to the privileges docs for more details on that. Otherwise I am certain we will diverge again. 

I did not explicitly test the privileges again, but remember that we did extensive testing when overhauling the privileges docs, so I think we can rely on them.

docs/copied-from-beats/docs/security/api-keys.asciidoc
simitt: Same as above, let's rely on the privilege docs.

docs/copied-from-beats/docs/security/api-keys.asciidoc
bmorelli25: > I'd only put a minimal example here, using the same privileges as documented for the writer role, and add a note that this assumes previous setup and link to the privileges docs for more details on that. Otherwise I am certain we will diverge again.

Makes sense! I'll get that fixed.

> but remember that we did extensive testing when overhauling the privileges docs, so I think we can rely on them.

💯

Thanks!

PR #5958, opened 2021-08-18
model: add APMEvent.{Observer,ECSVersion}

axw:
## Motivation/summary

Introduce Observer and ECSVersion to APMEvent, and populate them with model processors. This replaces the libbeat processors that were setting these fields.

Add some missing fields to the mappings, which became evident when switching to strict mapping in the integration.

## Checklist

## How to test these changes

Non-functional change.

## Related issues

None


beater/processors.go
```diff
@@ -57,3 +61,37 @@ func rateLimitBatchProcessor(ctx context.Context, batch *model.Batch) error {
...
+	for i := range *b {
+		event := &(*b)[i]
+		event.ECSVersion = ecs.Version
+	}
+	return nil
+}
+
+// newBeatInfoBatchProcessor returns a model.BatchProcessor that sets observer
+// fields from info.
+func newBeatInfoBatchProcessor(info beat.Info) model.ProcessBatchFunc {
```
simitt, line 77: How about calling this `observerBatchProcessor`?

simitt (approved):
Just for my understanding - the additional definitions for indexed fields are required for the apmpackage as fields from libbeat templates are not inherited there right?

axw:
> Just for my understanding - the additional definitions for indexed fields are required for the apmpackage as fields from libbeat templates are not inherited there right?

As discussed live: yes.

beater/processors.go
axw: Yes, that is a better name. Done.

PR #5970, opened 2021-08-19
modelindexer: introduce go-elasticsearch indexer

axw:
## Motivation/summary

Introduce an experimental option to index events using go-elasticsearch, bypassing libbeat. This only works when data streams are enabled. You can enable the experimental indexer with:

`apm-server -E output.elasticsearch.experimental=true -E apm-server.data_streams.enabled=true`

This approach aligns with how we want to evolve apm-server output tuning: moving away from the current libbeat queue and output configuration, simplifying config to be similar to how APM Agents are configured: bulk requests are now executed either after some time elapses (30 seconds by default, but this should probably be lowered) or if a number of bytes is reached (5MB by default).

TODO:

 - [ ] Hook up some metrics to monitor the queue size, number of bulk requests, number of events indexed, etc.
 - [ ] Add unit tests
 - [ ] Add a system test that compares this indexer vs. libbeat
 - [ ] Optimise the indexer by reusing memory buffers (probably can come later; initial testing shows it's already faster than libbeat)

## Checklist

## How to test these changes

1. Run apm-server with `-E output.elasticsearch.experimental=true -E apm-server.data_streams.enabled=true`
2. Send various events, check they are indexed successfully


stuartnelson3:
Confirmed with BC3

Issue #4715, opened 2021-02-10
[OpenTelemetry] Support OpenTelemetry Span Events


cyrille-leclerc:
Elastic APM silently ignores the Otel Span Events.

Specification of OpenTelemetry Span Events: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#add-events

Example of Otel Span Event we want to support:

```java
span.addEvent("order-creation", Attributes.of(
        EcommerceAppOpenTelemetryAttributes.CUSTOMER_ID, customerId,
        EcommerceAppOpenTelemetryAttributes.ORDER_PRICE, orderPrice,
        EcommerceAppOpenTelemetryAttributes.PAYMENT_METHOD, paymentMethod,
        EcommerceAppOpenTelemetryAttributes.SHIPPING_METHOD, shippingMethod,
        EcommerceAppOpenTelemetryAttributes.SHIPPING_COUNTRY, shippingCountry));
```

https://github.com/cyrille-leclerc/my-shopping-cart/blob/08f33761010d268793edb6e1d2101eaa579d830f/frontend-java/src/main/java/com/mycompany/ecommerce/controller/OrderController.java#L96

Source code ignoring the span events: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/91cbf1fd7f4eeea0087fdf782e3ff5b04e3ac90f/exporter/elasticexporter/internal/translator/elastic/traces.go#L88

paulgrav:
Would be good get some movement on this issue. Currently our devs are logging to stdout given that Span Events aren’t supported in either the Elastic Exporter or Elastic APM Server, or worse bunging stuff in attributes.

axw:
@paulgrav apologies for the silence, we've been rather busy. This isn't currently scheduled, but I'm hoping we'll be able to get to it within the next few months. My current thinking is that we'll index span events as logs, such that they'll show up in the trace logs UI:

![image](https://user-images.githubusercontent.com/843579/128819149-164bd03a-36ee-4506-b029-7465eac793ab.png)

Please let me know if that sounds off.

One of reasons why we haven't jumped on this is that APM Server does not create any log indices. We are working on migrating APM to data streams, and as part of that work we'll be introducing logs support. To get span events in sooner, we might support them only when producing data streams. This would mean using the Fleet integration, or waiting for https://github.com/elastic/apm-server/issues/5914.

paulgrav:
Thanks for the update.

[...]

Issue #4410, opened 2020-11-11
model: create a specification for valid event docs


axw:
We currently have hand-written model types, whose Transform methods are responsible for setting fields according to our schema. The schema is defined by fields.yml, and is only partially covered by ECS.

In the Kibana APM app code there are Typescript type definitions for the document `_source`: https://github.com/elastic/kibana/tree/master/x-pack/plugins/apm/typings/es_schemas/raw. In theory we could generate these from our field definitions, and with some annotations to indicate which fields are "required" (expected to always be produced by APM Server) we could avoid having `?:` on every property.

If we had a more complete model specification, we could use that to generate Typescript type definitions as well as the Go code for our own model Transform methods. Furthermore, in the future we might build on this to avoid the need for creating memory-heavy `common.MapStr` objects by generating code following an approach like described in https://github.com/elastic/apm-server/issues/3565#issuecomment-604839280. The main point being that relying on code generation makes the change easier than if we continue hand writing everything.

Finally, with a spec we could produce a more efficient codec for local event storage (i.e. for tail-based sampling): https://github.com/elastic/apm-server/issues/4120.

axw:
One idea I have in mind is to have the model types defined in protobuf, perhaps with some extensions (custom options) to define the JSON mapping. We could then:

 - generate the Go transformation code
 - generate the Typescript type definitions
 - directly implement a more efficient (protobuf) codec for local storage

It might be excessive, but we could then also consider using [CUE](https://cuelang.org/docs/integrations/protobuf/) protobuf extensions to define constraints (e.g. field `zed` is required when the `metricset.name` field value is "foo").

dgieselaar:
@axw for what it's worth, with the Fields API we can generate a typescript type from the index mapping on the fly. But we won't be able to differentiate between required and non-required, or single and multiple value fields.

axw:
@dgieselaar Nice. I suppose in theory we could add field metadata to fill in that gap. Field metadata is a limited resource though, so probably best to go with a statically generated definition still.

PR #5984, opened 2021-08-20
Introduce model.Processor

axw:
## Motivation/summary

Introduce type `model.Processor` and field `model.APMEvent.Processor`, which controls the `processor.name` and `processor.event` fields. These fields are used for identifying the event type, used for routing to the appropriate index/data stream and also for monitoring purposes.

Processed event metrics have been moved to a dedicated `model.BatchProcessor`, and out of the model package; they use the Processor field to form the metric name.

Note: Several unused metrics have been removed: stack trace and stack frame counts.

## Checklist

## How to test these changes

Enable monitoring, ingest some events, check that `apm-server.processor.<event-type>.transformations` metrics are recorded.

## Related issues

elastic/apm-server#4410
elastic/apm-server#4120
elastic/apm-server#4715


model/modeldecoder/rumv3/decoder.go
```diff
@@ -161,20 +161,22 @@ func DecodeNestedTransaction(d decoder.Decoder, input *modeldecoder.Input, batch
...
 
 	for _, m := range root.Transaction.Metricsets {
-		metricset := input.Base
-		mapToMetricsetModel(&m, &metricset)
-		metricset.Metricset.Transaction.Name = transaction.Transaction.Name
-		metricset.Metricset.Transaction.Type = transaction.Transaction.Type
-		*batch = append(*batch, metricset)
+		event := input.Base
+		mapToMetricsetModel(&m, &event)
+		event.Processor = model.MetricsetProcessor
```
stuartnelson3, line 166: It appears this `event.Processor = model.MetricsetProcessor` is also being set in `mapToMetricsetModel`? Is that intentional?

model/modelprocessor/datastream.go
```diff
@@ -39,14 +39,14 @@ func (s *SetDataStream) ProcessBatch(ctx context.Context, b *model.Batch) error
 }
 
 func (s *SetDataStream) setDataStream(event *model.APMEvent) {
-	switch {
-	case event.Transaction != nil || event.Span != nil:
+	switch event.Processor {
+	case model.SpanProcessor, model.TransactionProcessor:
```
stuartnelson3, line 43: :partying_face:

stuartnelson3 (approved):
Big fan of the move to explicitly setting the datastream based on `Processor`. One question where it looks like processor might be getting double set, but that's all.

simitt (approved):

model/modeldecoder/rumv3/decoder.go
axw: good catch, I've removed these ones and consolidated on setting it in mapToFooModel

PR #4401, opened 2020-11-10
Support check packages for release and master branches

v1v:
## Motivation/summary

Support the package testing for release and master branches running only when the apm-server main pipeline is success, then it triggers the check packages for the given commit-id and only validating the apm-server packages no the beats packages which are tested in the beats-ci.

## Related issues

Requires https://github.com/elastic/apm-server/pull/4400

## Important

`BEAT_VERSION` cannot be used for some reason the builds in windows failed as a consequence of setting that env variable:

```
[2020-11-12T11:23:49.814Z] Building 
[2020-11-12T11:23:53.996Z] 2020/11/12 11:23:52 exec: go list -m
[2020-11-12T11:23:53.996Z] >> build: Building apm-server
[2020-11-12T11:23:53.996Z] Error: failed to parse version 'Makefile:242: recipe for target 'build//bin/mage' failed'
[2020-11-12T11:23:53.996Z] github.com/elastic/beats/v7/dev-tools/mage.ParseVersion
[2020-11-12T11:23:53.996Z] 	C:/Users/jenkins/workspace/pm-server_apm-server-mbp_PR-4401/pkg/mod/github.com/elastic/beats/v7@v7.0.0-alpha2.0.20201029160358-633285e75a03/dev-tools/mage/common.go:793
...
[2020-11-12T11:23:53.996Z] failed generating Windows .syso metadata file
[2020-11-12T11:23:53.996Z] github.com/elastic/beats/v7/dev-tools/mage.Build
[2020-11-12T11:23:53.996Z] 	C:/Users/jenkins/workspace/pm-server_apm-server-mbp_PR-4401/pkg/mod/github.com/elastic/beats/v7@v7.0.0-alpha2.0.20201029160358-633285e75a03/dev-tools/mage/build.go:150
[2020-11-12T11:23:53.996Z] main.Build
[2020-11-12T11:23:53.996Z] 	C:/Users/jenkins/workspace/pm-server_apm-server-mbp_PR-4401/src/github.com/elastic/apm-server/magefile.go:67
[2020-11-12T11:23:53.996Z] main.main.func14
[2020-11-12T11:23:53.996Z] 	C:/Users/jenkins/workspace/pm-server_apm-server-mbp_PR-4401/src/github.com/elastic/apm-server/mage_output_file.go:599
[2020-11-12T11:23:53.996Z] main.main.func12.1
...
```

PR #6009, opened 2021-08-23
Disable Google cache after object upload

cachedout:
## Motivation/summary

Attempt to fix failure in branch packaging test

## Checklist

## Related issues

Refs issue in private repo.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -497,6 +497,12 @@ pipeline {
                   sharedPublicly: true,
                   showInline: true)
 
+                // Disable caching
+                gsutil(
+                    command: "-h 'Cache-Control:public, max-age=0' ${BUCKET_URI}",
+                    credentialsId: "${JOB_GCS_CREDENTIALS}",
+                )
+
                 // Copy those files to another location with the sha commit to test them afterward.
                 googleStorageUpload(bucket: "gs://${JOB_GCS_BUCKET}/commits/${env.GIT_BASE_COMMIT}",
                   credentialsId: "${JOB_GCS_CREDENTIALS}",
@@ -504,6 +510,12 @@ pipeline {
                   pattern: "${BASE_DIR}/build/distributions/**/*",
                   sharedPublicly: true,
                   showInline: true)
+
+                // Disable caching
+                gsutil(
+                    command: "-h 'Cache-Control:public, max-age=0' gs://${JOB_GCS_BUCKET}/commits/${env.GIT_BASE_COMMIT}",
+                    credentialsId: "${JOB_GCS_CREDENTIALS}",
+                )
               }
             }
           }


v1v (commented):
IIRC, we don't manage the google storage bucket defined in the `JOB_GCS_BUCKET` credentials but infra.

What's the current issue? In case, it can be solved in the bucket itself?

cachedout:
> What's the current issue? In case, it can be solved in the bucket itself?

Thanks for the question, @v1v 

What's happening here is that when we upload the artifact, Google caches it for 3600s by default. This means that we upload a new hash for the artifact but when we try to download it, we get the previous copy. This then breaks the packaging test because beats-tester fails to get the correct hash.

There were two previous attempts to fix this issue, neither of which were successful:

https://github.com/elastic/beats-tester/pull/204
https://github.com/elastic/beats-tester/pull/205

So the idea here is to upload the artifact and then immediately adjust the meta-data. I did investigate whether or not it is possible to set this on a per-bucket layer and it is not. We could do it at upload-time, but the strategy of using a second-call as this PR does seems to only add minimal overhead.

v1v:
Thanks for the clarification.
There is something that I'm confused, https://github.com/elastic/apm-server/pull/4401 was implemented to support the `commit-id` folder name rather than the `latest` folder name. The idea was to avoid testing the `beats-tester` pipeline when the latest artifacts were uploaded from the `apm-server` by another build, a concurrent issue.

Maybe there is a missing configuration somewhere that somehow is not well implemented? Otherwise, the cache issue should not happen since the `commit-id` is unique.

v1v:
To illustrate the above, I've just looked at:

- https://apm-ci.elastic.co/job/apm-server/job/apm-server-mbp/job/master/1159/ that triggered the beats-tester build -> https://apm-ci.elastic.co/job/apm-server/job/apm-server-check-packages-mbp/job/master/757/

And the `beats-tester.properties` file is read in https://apm-ci.elastic.co/blue/organizations/jenkins/apm-server%2Fapm-server-check-packages-mbp/detail/master/757/pipeline/27#step-31-log-1.

image

And the content can be seen in 

https://apm-ci.elastic.co/job/apm-server/job/apm-server-mbp/job/master/1159/artifact/beats-tester.properties/*view*/

Definitely there is something missing since the `APM_URL_BASE` is not correct 

Though

cachedout:
Could we be falling back here? 

https://github.com/elastic/apm-server/pull/4401/files#diff-62bc844eece79c6e6233131c5452490b5d38b86764b592ed4ec4e764f0547c28R53

v1v:
There was a bug in the current implementation, since the global environment variables map is immutable :/ . https://github.com/elastic/apm-server/pull/6012 should help with this, 🤞

cachedout:
Ah, nice. OK, let's try that. I'll close this in the meantime.

PR #6011, opened 2021-08-23
fail open if snapshot not available

### LINKED ISSUE #5967: Fix check-docker-compose target if no snapshots are available
For the `7.15` branch https://github.com/elastic/apm-server/pull/5965/commits/55f1bf0aa8c07204889d2bd958fc922e1a0060b8 removed the `check-docker-compose` command, as it was failing due to no snapshot being available yet. 

The `check-docker-compose` target was introduced in https://github.com/elastic/apm-server/pull/5035 to ensure that the used docker images are kept in sync. The target should be safe to ignore if no snapshot is yet available for this branch.

stuartnelson3:
## Motivation/summary

check-docker-compose target shouldn't fail if no updated snapshot is available.

according to the issue:

>The target should be safe to ignore if no
snapshot is yet available for this branch.

closes #5967

## How to test these changes

non-functional change

## Related issues

#5967


simitt (commented):
Please also add the `check-docker-compose` command to the `check-full` target again (removed in https://github.com/elastic/apm-server/commit/55f1bf0aa8c07204889d2bd958fc922e1a0060b8)

stuartnelson3:
It seems it was only removed in `7.x`, wouldn't we make that change when backporting?

simitt (approved):
yes, that should be done in `7.x` backport then. 
LGTM

PR #6012, opened 2021-08-23
[CI]: use setEnvVar instead of global env variable

v1v:
## Motivation/summary

It can't override the environment variable defined in the environment `{}` block, therefore let's use a different approach and add more debug traces.

## Checklist

## How to test these changes

In the CI

## Related issues

Supersedes https://github.com/elastic/apm-server/pull/6009


cachedout (approved):

cachedout:
This looks to be the correct fix. Nicely done, @v1v !

Issue #4636, opened 2021-01-19
[meta] APM Server managed by Elastic Agent with Fleet (GA)


simitt:
This meta issue collects work required to make APM Server managed by Elastic Agent with Fleet generally available. 

### >= 8.0 
- [x] Create data stream specific pipelines https://github.com/elastic/apm-server/issues/6001
- [x] Ensure APM Server has enough privileges for tail based sampling https://github.com/elastic/apm-server/issues/5490
- [ ] ~Drop privileges when managed by Elastic Agent https://github.com/elastic/apm-server/issues/4571~ Moved to after GA.
- [x] Configurable throughput settings https://github.com/elastic/beats/issues/26638
- [ ] ~Investigate support for spooling to disk https://github.com/elastic/apm-server/issues/5441~ Not that high priority given the new APM output handling
- [x] Java agent auto-attachment  https://github.com/elastic/apm-server/issues/4830 (work on APM Server side is done)

### 7.16 
- [x] Add `message` field to error docs https://github.com/elastic/apm-server/issues/5971
- [ ] ~[stretch goal] Java agent auto-attachment  https://github.com/elastic/apm-server/issues/4830~
- [x] Make use of component templates in APM and restrict dynamic indexing https://github.com/elastic/apm-server/issues/5707
- [x] Align package version with Stack version https://github.com/elastic/apm-server/issues/4898

###  7.15
- [x] Reload configs without restart https://github.com/elastic/apm-server/issues/5039

#### Stretch
- [x] Supporte multiple fleet-server addresses https://github.com/elastic/apm-server/issues/5514
###  7.14
- [x] Meta issue collecting required steps for `7.14` https://github.com/elastic/apm-server/issues/5388

### Kibana - Stack Monitoring
- [x] Same level of insights needs to be provided to users https://github.com/elastic/kibana/issues/90157
- [x] changes in Elastic Agent to expose metrics

###  7.12
- [x] [Meta issue collecting required steps for `7.12`](https://github.com/elastic/apm-server/issues/4558).

simitt:
Closing this as all the required work on the APM Server side for going GA and upgrading to 8.0 has been done. Single bugs (e.g. stack monitoring issues) are tracked separately.

PR #6043, opened 2021-08-25
Add dynamic_templates to app_metrics data stream

axw:
## Motivation/summary

Add a "histogram" dynamic template to the application metrics (`metrics-apm.app-*`) data stream index template, so we can dynamically map histogram metrics when using data streams.

## Checklist

## How to test these changes

1. Run apm-server with `-E apm-server.data_streams.enabled=true`
2. Send some histogram metrics, e.g. using the Python agent's Prometheus integration, or with OpenTelemetry
3. Check the metrics are indexed as histograms

## Related issues

https://github.com/elastic/apm-server/issues/4636


axw:
This change highlighted an issue with the internal metrics data stream: its template defines `dynamic: strict` under the premise that the server is in full control of the metrics that are indexed.

This is not quite true at the moment, as some metrics (breakdown metrics) come from agents, and we don't currently vet these at all. I propose that we begin doing so. For any metricset that contains `transaction.name`, `span.name`, etc., we will silently discard any unknown metrics, much like we do for unknown fields on transactions, spans, and errors.

stuartnelson3 (approved):

Issue #5910, opened 2021-08-09
[Testing] One Kibana field is duplicated


mdelapenya:
Detected in current master branch: https://apm-ci.elastic.co/blue/organizations/jenkins/apm-server%2Fapm-server-mbp/detail/master/1137/pipeline/171

The build result is not affected (still green/stable/success), although I'm not sure if that should pop-up in any other manner

### Steps to reproduce:
1. Run `make are-kibana-objects-updated`

>/Users/mdelapenya/.gimme/versions/go1.16.6.darwin.amd64/bin/go build -o build/darwin/bin/mage github.com/magefile/mage
./build/darwin/bin/mage -compile=build/darwin/mage
2021/08/09 11:06:41 exec: go list -m
Error generating Index Pattern: field <network.connection.type> is duplicated, remove it or set 'overwrite: true', {Name:type Type:keyword Description:Network connection type, eg. "wifi", "cell"
 Format: Fields:[] MultiFields:[] Enabled:<nil> Analyzer: SearchAnalyzer: Norms:false Dynamic:{Value:<nil>} Index:<nil> DocValues:<nil> CopyTo: IgnoreAbove:0 AliasPath: MigrationAlias:false Dimension:<nil> DynamicTemplate:false Unit: MetricType: ObjectType: ObjectTypeMappingType: ScalingFactor:0 ObjectTypeParams:[] Analyzed:<nil> Count:0 Searchable:<nil> Aggregatable:<nil> Script: Pattern: InputFormat: OutputFormat: OutputPrecision:<nil> LabelTemplate: UrlTemplate:[] OpenLinkInCurrentTab:<nil> Overwrite:false DefaultField:<nil> Path:network.connection.type}, {"aggregatable":true,"analyzed":false,"count":0,"doc_values":true,"indexed":true,"name":"network.connection.type","scripted":false,"searchable":true,"type":"string"}.
make: *** [build/index-pattern.json] Error 1

PR #6058, opened 2021-08-26
Stop updating/checking Kibana index pattern

axw:
## Motivation/summary

We're moving towards the integration package,
and this is no longer needed. We don't need
to list out every field mapping, as the index
pattern refreshes the mapping automatically.

In the unlikely event that we need to update
a Kibana field formatter, we can do so manually.

## Checklist

## How to test these changes

N/A

## Related issues

Closes https://github.com/elastic/apm-server/issues/5910


simitt (approved):
Generally LGTM, but I think we also have to consider removing https://www.elastic.co/guide/en/apm/server/current/exported-fields.html from the docs then.

axw:
@simitt I'll merge, but I'd like to follow up on your comment. I don't understand why we should remove those docs. They're still kept up to date.

PR #6065, opened 2021-08-29
modeldecoder/rumv3: drop standalone metricsets

axw:
## Motivation/summary

When using v3, the RUM agent only sends metricsets for breakdown metrics embedded within transactions, and never as standalone events:

https://github.com/elastic/apm-agent-rum-js/blob/52f9118e587bef1831a743eee3a3166f9eba141a/packages/rum-core/src/common/compress.js#L260

Drop the `metricsetRoot` type from modeldecoder/rumv3. Also, RUM transaction breakdown metricsets never contain tags/labels, so we can drop that field from `transactionMetricset` (nee `metricset`) type.

## Checklist

## How to test these changes

Run apm-integration-testing with opeans-rum

Check that opbeans-rum breakdown metrics still show up in the APM UI, along the lines of:

image

## Related issues

None


axw:
FYI @elastic/apm-agent-rum.

This PR drops the ability to send `{"me":{...}}` (metricsets) at the top level of a v3 payload, and drops the ability to attach tags/labels to metrics. AFAICS the RUM agent only ever sends breakdown metrics (which don't ever have labels), and with v3 they are always embedded within a transaction. Please let me know if I overlooked anything.

stuartnelson3 (approved):
looks good, also a fan of the renaming to show the relationship between metricsets and transactions

vigneshshanmugam (commented):
@axw This would be a breaking change category for the RUM agent. 

- RUM uses api v2 by default - https://www.elastic.co/guide/en/apm/agent/rum-js/current/configuration.html#api-version and includes the metricsets in the root level. 

- On the contrary, `breakdownMetrics` is disabled by default in the RUM config - https://www.elastic.co/guide/en/apm/agent/rum-js/current/configuration.html#breakdown-metrics. We could still release this change and ask users to change the config to V3 if they would like to see breakdown metrics in the UI. 

Or else we would need to wait till 6.x of the RUM agent is released. WDYT?

axw:
@vigneshshanmugam I have only changed the V3 decoder. The V2 decoder still accepts metricsets at the root level. 

The V3 decoder was accepting them embedded in transactions as well as at the root. I removed the ability for them to be at the root.

vigneshshanmugam:
@axw Ah totally misread that. Gotcha all good then 👍🏽 Thanks for the details.

axw:
Thanks for confirming!

stuartnelson3:
Confirmed with BC3

Issue #5084, opened 2021-04-13
[doc] Document server-generated metrics


axw:
Ideally we would have a complete specification for valid documents (https://github.com/elastic/apm-server/issues/4410). Until we get there, we should at least make sure the Elasticsearch docs we produce are documented -- possibly as part of user-facing elastic.co docs, possibly just as developer-focused docs in the git repo.

For example, these docs should describe the fields present in:
 - "transaction" metrics (https://github.com/elastic/apm-server/blob/bb856fcc78a49f3c874688ae85c1762663d9381a/x-pack/apm-server/aggregation/txmetrics/aggregator.go#L358-L401)
 - "service_destination" metrics (https://github.com/elastic/apm-server/blob/bb856fcc78a49f3c874688ae85c1762663d9381a/x-pack/apm-server/aggregation/spanmetrics/aggregator.go#L257-L297)

axw:
@sqren FYI

PR #6111, opened 2021-09-02
Introduce dedicated fields for internal metrics

axw:
## Motivation/summary

Add dedicated fields to model.Transaction and model.Span for well-defined, internal transaction and span metrics. Drop unknown metrics received in metricsets with any transaction or span fields defined.

Internal metrics will be indexed with strict mapping. We already have explicitly defined fields for these metrics. This PR ensures that unknown metrics do not slip through. If an agent is upgraded before the server and some new internal metric starts being produced, that must not be indexed as it would cause other internal metrics to be rejected by Elasticsearch.

Drop the unused `metricset.period` field that was being added to service_destination metrics.

## Checklist

## How to test these changes

- Run apm-server with `-E instrumentation.enabled=true`
- Check that breakdown metrics (Time spent by span type) work correctly in the UI
- Check that service destination metrics (Dependencies) work correctly in the UI

## Related issues

- https://github.com/elastic/apm-server/issues/5084
- https://github.com/elastic/apm-server/issues/4410
- https://github.com/elastic/apm-server/pull/6043#issuecomment-905257392


simitt (approved):

leyao-daily:
This PR said the field metricset.period is removed but why the document still mentioned the field [here](https://www.elastic.co/guide/en/apm/guide/7.17/exported-fields-apm-span-metrics-xpack.html).

PR #4061, opened 2020-08-10
Use service.version for Metadata.Service.Version when converting a Jaeger span

tobiasstadler:

PR #6114, opened 2021-09-02
Use deployment.environment for Metadata.Service.Environment when converting a Jaeger span.

tobiasstadler:
## Motivation/summary

deployment.environment is used in OpenTelemetry and already converted to Metadata.Service.Environment for OpenTelemetry spans. It would be nice if it is also supported for Jaeger spans.

## Checklist
- [ ] ~Documentation has been updated~


axw (commented):
Thanks for the PR @tobiasstadler.

I'm not sure if this is needed. Jaeger has "tracer tags" (aka "process tags"), which will be translated to "resource" in OTel.

Can you set `deployment.environment` as a tracer tag rather than span tag? You should be able to do this with an environment variable (`JAEGER_TAGS=deployment.environment=production`) or in code.

tobiasstadler:
Setting process tags works. Thank You!

axw:
Thanks for confirming.

@tobiasstadler you may recall adding support for `service.version` back in https://github.com/elastic/apm-server/pull/4061. That should probably also be set in tracer tags. How would you feel about us deprecating setting service version in span tags?

tobiasstadler:
I am fine with deprecating it.

Issue #3338, opened 2020-02-17
Process additional Jaeger Span `logs` 


simitt:
At the moment only `error` logs are processed and transformed into Elastic APM error events. Record additional Span Logs according to the [OpenTracing spec](https://github.com/opentracing/specification/blob/master/semantic_conventions.md#log-fields-table).

Pick up previous discussion about investigating if it makes sense to introduce a dedicated Elastic APM convention for extended support.

graphaelli:
Linking to #3723.  I had thought these would turn into transaction marks originally but generalizing them to actual logs would make them even more useful.

Dmitry31:
Any future for this feature? There is no clear statement in the documentation about OpenTracing and OpenTelemetry support. Looking around on github I feel like Jaeger support is no more a priority as OpenTracing merged into OpenTelemetry ... But OpenTelemetry logs specs are in experimental state so no support neither?

axw:
@Dmitry31 Jaeger span logs are equivalent to OpenTelemetry span events, so when we get to https://github.com/elastic/apm-server/issues/4715 we will be able to address this too.

Issue #3873, opened 2020-06-11
model: align labels with ECS


axw:
ECS defines [labels](https://www.elastic.co/guide/en/ecs/current/ecs-base.html) as key-value pairs, where all values are stored as keywords. APM also allows booleans and scaled_float values.

We should propose an extension to ECS.

axw:
I think it's also worth considering extending to allow integers (`long` type in ES), to better support storing OpenTelemetry attributes. This isn't specific to traces/APM.

simitt:
Allowing `long` could be problematic from an implementation point, for cases where the first value of an intended float would be e.g. `24.0` and the type `long` would be auto-derived from it instead of `scaled_float`. We specifically used `scaled_float` in the past to be less ambiguous.

axw:
This might be an unpopular opinion, but I'm starting to wonder if having dynamically-typed label fields is something we should continue. We see that sometimes users have a bad time because they changed the label type in their instrumentation, or they have multiple applications with colliding label names/different types.

Maybe we should split them into separate fields, like `labels`, `double_labels`, `long_labels`, and `boolean_labels`?

simitt:
I don't see how this change would cover the issues arising from using the same lable name with different types in different applications, but otherwise agree that dynamic typing can be very painful here and adding the type to the name might improve the situation.

jalvz:
> split them into separate fields, like labels, double_labels, long_labels, and boolean_labels

That would be the Go way, but APIs might feel very unidiomatic on pretty much every other language, thou.

webmat:
I like the idea of adding labels fields for different types. I'd like to avoid having an explosion of types though (e.g. half_float, float, double, scaled_float). So perhaps I would err on the side of using only the roomier versions of any given type (long and double). Happy to make exceptions if any are needed.

I think it may be acceptable define these new object fields within the existing `labels` object type:

* `labels.*`: unless otherwise noted, are `keyword`
* `labels.long.*`
* `labels.double.*`
* `labels.boolean.*`

We're currently starting to plan the ECS 2.0 breaking changes. This could be part of it.

On the other hand, `labels` sounds like a misnomer for numerics and booleans. We're open to ideas for capturing ad hoc metrics or other kinds of numerics :-)

axw:
@simitt 

> I don't see how this change would cover the issues arising from using the same lable name with different types in different applications

[...]

PR #6122, opened 2021-09-04
processor/otel: index span events as logs

axw:
## Motivation/summary

Produce log events for non-exception OpenTelemetry span events, capturing the span event name as the log message, and all other attributes as labels. We only index logs when data streams are enabled; they are dropped when classic indices are in use.

## Checklist

## How to test these changes

1. Instrument an application with either an OpenTelemetry SDK or Jaeger, capturing a span and span logs/events.
2. Run APM Server with data streams enabled, check that logs are indexed into `logs-apm.app-<namespace>` and show up in the "Logs" panel for a trace sample
3. Run APM Server with data streams disabled, check that logs are not indexed at all

## Related issues

Closes https://github.com/elastic/apm-server/issues/4715
Closes https://github.com/elastic/apm-server/issues/3338


axw:
@bmorelli25 @cyrille-leclerc FYI I've made a few minor edits to the Jaeger and OpenTelemetry docs, removing the caveats related to span events. There were also some outdated caveats listed in the docs related to metrics which I've removed as well.

simitt (approved):
LGTM

cyrille-leclerc:
I love it, which naming convention did you use

I see https://github.com/elastic/apm-server/pull/6122/files#diff-ca87920522e282bbb78ffa10d9ef40c27190cdc609058b12edade698ac3d8526R48
```
 model.DataStream{Type: "logs", Dataset: "apm.app", Namespace: "custom"}
```

Which suggests the stream name will be `logs-apm.app-custom`

Some frequent samples show 

My thoughts are:
* We have to be consistent with what APM will do for transactions, spans, and metrics,
* According to "[Datastream naming scheme](https://www.elastic.co/guide/en/fleet/current/data-streams.html#data-streams-naming-scheme)":
   * Should the `namespace` be [`deployment.environment`](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/semantic_conventions/deployment_environment.md) like rpdocution", "staging" as our docs say _"A user-configurable arbitrary grouping, such as an environment (`dev`, `prod`, or `qa`), a team, or a strategic business unit..."_ rather than the proposed `custom`
   * Should the `dataset` 
      * contain the `service.name`, maybe as an alternative to the proposed `app` of `apm.app`
   "_describes the ingested data and its structure for each index_"
      * be prefixed by `apm.` as you propose? `apm.` refer to a category of tools more than referring to the structure of ingested data
 
Looking at the existing name like `logs-nginx.access-production`, I am thinking of something like 

`logs-[${service.namespace}].${service.name}.trace-${deployment.environment}`

Which would be `logs-ecommerce.frontend.trace-production`

axw:
> Should the namespace be deployment.environment like rpdocution", "staging" as our docs say "A user-configurable arbitrary grouping, such as an environment (dev, prod, or qa), a team, or a strategic business unit..." rather than the proposed custom

The "custom" you see is just a unit test. The namespace is user-defined when they create their agent policy in Fleet, and is the same for all data streams produced by the APM server corresponding to that policy.

The team has discussed in the past the idea of permitting templates in the namespace, e.g. set it to `{{service.environment}}` so it will end up being set to the same value as the service environment. This is not currently possible.

> Should the dataset contain the service.name, maybe as an alternative to the proposed app of apm.app
> "describes the ingested data and its structure for each index"

I don't think it's necessary. The index has the same structure regardless of the application. This is somewhat aspirational: OTel attributes are dynamically mapped as `labels.*`, which means there could be a mapping conflict if two applications map an attribute as different types. We plan to address this in two ways:
 - we will split numeric labels off into a new `numeric_labels` field
 - we will switch these both over to `flattened`, to avoid mapping explosions

See also https://github.com/elastic/apm-server/issues/3873#issuecomment-900966203

> be prefixed by apm. as you propose? apm. refer to a category of tools more than referring to the structure of ingested data

Perhaps we don't need the common prefix. One thing we would need to check is whether the UI has permissions to read all `logs-*` data streams.

> Maybe I could rephrase saying, for an application (service.namespace: ecommerce, service.name: frontend, deployment.environment: production) what are the data streams for transactions, spans, transaction events, span events, logs, metrics?

- `traces-apm.app-<namespace>` (transactions and spans)
- `logs-apm.error-<namespace>` (errors)
- `logs-apm.app-<namespace>` (span events/logs)
- `metrics-apm.app.<service.name>-<namespace>` (application metrics)
- `metrics-apm.internal-<namespace>` (APM-internal metrics)

The reason for having the service name in the application metrics data stream is because in that case only, there are going to be dynamically mapped fields which
- may lead to mapping collisions
- will differ considerably between applications

stuartnelson3:
Confirmed that logs are indexed with datastreams enabled, and not indexed with datastreams disabled. The logs didn't show up in the "logs" tab, though.

stuartnelson3:
issue: https://github.com/elastic/apm-server/issues/6544

stuartnelson3:
Confirmed with BC3

cyrille-leclerc:
@axw could we please have a screenshot with a somewhat realistic example of span event composed of a name and a bunch of attributes?

If you can 
* confirm me that span events appear in the trace/logs tab as lines of "logs" with a message that is composed of  the event name and of its attributes
* tell me what is the formatting of the "Message" column, how we format the attributes, then I'll be able to craft a screenshot

image

axw:
> confirm me that span events appear in the trace/logs tab as lines of "logs" with a message that is composed of the event name and of its attributes

Span events will appear in the trace logs tab.

> tell me what is the formatting of the "Message" column, how we format the attributes, then I'll be able to craft a screenshot

AFAIK, the message column is simply taken from the value of the `message` field. The `message` field is set to the span event name:

https://github.com/elastic/apm-server/blob/793f99ae00af412364913e77c0459136ece07a78/processor/otel/traces.go#L873

cyrille-leclerc:
Many thanks @axw . 

* Can you confirm that the OTel resource attributes like `service.name`or `service.version` are added as fields of the log message in the Elasticsearch storage?
* Would it make sense for you that when visualizing a span events, we should display the following data of the span event (see [Span Event Specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#add-events)):
   * Name of the event
   * Timestamp of the event
   * Span event attributes

My understanding is that this visualization of the span events that would be different from the visualization of raw log messages may require a different renderer for log messages and for span events.

Would that make sense?

axw:
> Can you confirm that the OTel resource attributes like service.nameor service.version are added as fields of the log message in the Elasticsearch storage?

Those fields will be present on the log document, but not all fields will be. Labels will not be.

> Would it make sense for you that when visualizing a span events, we should display the following data of the span event

Message and timestamp 100% belong on the top level, and probably the service name. Attributes will likely get a bit busy. I think they belong in a flyout, which we don't yet have but is tracked in https://github.com/elastic/kibana/issues/111325

> My understanding is that this visualization of the span events that would be different from the visualization of raw log messages may require a different renderer for log messages and for span events.

I'm having trouble parsing this. Can you restate, please?

cyrille-leclerc:
> > My understanding is that this visualization of the span events that would be different from the visualization of raw log messages may require a different renderer for log messages and for span events.
> 
> I'm having trouble parsing this. Can you restate, please?

If we want a special visualization in the logs tab, then we will need to evolve the "logs visualization tab" to have dedicated rendering for "span events" and for "raw log messages".

As you propose for the moment to rely on the flyout, the question of a dedicated renderer is not needed

PR #6125, opened 2021-09-06
Remove package "convert"

axw:
## Motivation/summary

- "convert" is a not very meaningful name
- the code within is trivial
- the code is also dangerous; it ignores errors in "ToReader", but makes no mention of this in its docs

Inline the JSON marshalling/unmarshalling logic into the few tests that use them, and remove the package.

## Checklist

## How to test these changes

Non-functional change.

## Related issues

None.


simitt (approved):
+1 on removing this; thanks for the change!

Issue #5949, opened 2021-08-16
Return 503 while server is in wait-for-integration loop.


Mpdreamz:
Currently the server delays binding HTTP until it knows the downstream dependencies are set up.
To be precise it does this when `apm-server.data_streams.wait_for_integration` is set to true, also defaults to `true` when `apm-server.data_streams.enable` is set.

This has a potential knock on affect on agents trying to connect and timing out their batch operations.

If the server binds immediately but is able to inform the agents its not ready yet through `503` this could be remedied.

graphaelli:
I think this apm-server should queue events until full while waiting for the integration to be installed, just like it would if es was unavailable.

Mpdreamz:
On the fence about queueing, ES being unavailable is something that could potentially resolve itself. 

Here we know ES/Kibana is available but not configured correctly and requiring manual configuration.

In cases where the stack is updated before the server the time to resolve the failure is low and a typical upgrade should have already installed the integration before upgrading standalone server. 
In cases where the server is updated before the stack the time to resolve is high and accepting events during upgrade only to fail later feels trappy.

axw:
We can't know if the integration is installed or not if Elasticsearch is unavailable. I'd prefer to have fewer special cases, so I'm going to go with the enqueuing approach for now. I'm about to put up a PR which just blocks all Elasticsearch client requests (aside from the index template checks) until the integration package is known to be installed. The result of this is that the server will behave as if Elasticsearch is unavailable.

PR #6130, opened 2021-09-08
Add "waitReady" preconditions for Elasticsearch clients

axw:
## Motivation/summary

When configured for standalone+data streams, we wait for the integration to be installed. This is now checked asynchronously, blocking Elasticsearch client requests rather than the server starting up. This allows events to be received, processed, and enqueued, but not indexed until the integration is installed. The tail-based sampling license check has been moved to this new wait loop, and uses the go-elasticsearch client to fetch the license instead of libbeat.

We add a "NewElasticsearchClient" function to beater.ServerParams, which should be used to construct `elasticsearch.Clients` wherever the client may be used to perform indexing operations. The elasticsearch package API has been simplified, and some unit tests removed as a result.

Package beater tests now have the option of publishing events using the standard libbeat elasticsearch output. We use this in the wait-for-integration test to check that events are not sent to Elasticsearch until the preconditions are met. We now deregister the pipeline registration callback when the server stops to fix tests which use the elasticsearch output, as otherwise the callback added in one test bleeds into those proceeding.

The "/" HTTP route handler now reports whether the server is "publish ready", i.e. it is ready to publish events.

## Checklist

## How to test these changes

1. Run apm-server with `-E apm-server.data_streams.enabled=true`
2. Send some events, verify that they are not indexed
3. Install the integration package
4. Send some more events, verify that they are indexed

## Related issues

Closes https://github.com/elastic/apm-server/issues/5949

### DIFF
--- a/beater/waitready.go
+++ b/beater/waitready.go
@@ -1,0 +1,84 @@
+// Licensed to Elasticsearch B.V. under one or more contributor
+// license agreements. See the NOTICE file distributed with
+// this work for additional information regarding copyright
+// ownership. Elasticsearch B.V. licenses this file to you under
+// the Apache License, Version 2.0 (the "License"); you may
+// not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package beater
+
+import (
+	"context"
+	"net/http"
+	"time"
+
+	"go.elastic.co/apm"
+
+	"github.com/elastic/beats/v7/libbeat/logp"
+)
+
+// waitReady waits for preconditions to be satisfied, by calling check
+// in a loop every interval until ctx is cancelled or check returns nil.
+func waitReady(
+	ctx context.Context,
+	interval time.Duration,
+	tracer *apm.Tracer,
+	logger *logp.Logger,
+	check func(context.Context) error,
+) error {
+	logger.Info("waiting for preconditions")
+	tx := tracer.StartTransaction("wait_for_preconditions", "init")
+	ctx = apm.ContextWithTransaction(ctx, tx)
+	var ticker *time.Ticker
+	for {
+		if ticker == nil {
+			// We start the ticker on the first iteration, rather than
+			// before the loop, so we don't have to wait for a tick
+			// (5 seconds by default) before peforming the first check.
+			ticker = time.NewTicker(interval)
+			defer ticker.Stop()
+		} else {
+			select {
+			case <-ctx.Done():
+				return ctx.Err()
+			case <-ticker.C:
+			}
+		}
+		if err := check(ctx); err != nil {
+			logger.Errorf("error checking preconditions: %s", err)
+			continue
+		}
+		return nil
+	}
+}
+
+// waitReadyRoundTripper wraps a *net/http.Transport, ensuring the server's
+// indexing preconditions have been satisfied by waiting for "ready" channel
+// to be signalled, prior to allowing any requests through.
+//
+// This is used to prevent elasticsearch clients from proceeding with requests
+// until the APM integration is installed to ensure we don't index any documents
+// prior to the data stream index templates being ready.
+type waitReadyRoundTripper struct {
+	*http.Transport
+	ready <-chan struct{}
+}
+
+func (c *waitReadyRoundTripper) RoundTrip(r *http.Request) (*http.Response, error) {
+	select {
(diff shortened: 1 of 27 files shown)


beater/waitready.go
```diff
@@ -0,0 +1,84 @@
...
+// until the APM integration is installed to ensure we don't index any documents
+// prior to the data stream index templates being ready.
+type waitReadyRoundTripper struct {
+	*http.Transport
+	ready <-chan struct{}
+}
+
+func (c *waitReadyRoundTripper) RoundTrip(r *http.Request) (*http.Response, error) {
+	select {
+	case <-c.ready:
```
stuartnelson3, line 79: Will this block on an infinite number of requests, or does the es client wait for responses before sending more requests? I don't think this needs to be solved now (or potentially at all), but assuming there's no coordination when sending new requests, will all incoming events (pre-integration install) find their way into a request and potentially wait there forever, and crash the server? If setting up the integration takes arbitrarily long, will the apm-server get OOM killed, restart, and then queue up requests in the hope that (this time) the integration will finish?

stuartnelson3 (approved):
looks awesome. question about the behavior wrt a long wait before the integration installs, but that's all.

beater/waitready.go
axw: > Will this block on an infinite number of requests, or does the es client wait for responses before sending more requests?

It will block an indefinite number of requests, so it depends on how the Elasticsearch client is used. With the exception of the Elasticsearch sourcemap implementation, in all cases I'm aware of we create requests serially:
 - The libbeat elasticsearch output has N workers, which will each have at most one bulk request open at a time.
 - Same as above for the go-elasticsearch bulk indexer (used in tail-based sampling, and my modelindexer draft PR)
 - Tail-based sampling pubsub subscriber executes one request at a time

The Elasticsearch sourcemap implementation won't work with data streams, so I think we're OK to ignore that one. I'll merge, and we can keep an eye on it.

marclop:
Tested using the latest `7.16.0 BC`, works as expected:

1. Create a deployment with `apm-server.data_stream.enabled: true`
2. `curl -H "Authorization: Bearer ${ELASTIC_APM_SECRET_TOKEN}" ${ELASTIC_APM_SERVER_URL}`:
```json
{
  "build_date": "2021-11-03T05:31:56Z",
  "build_sha": "d7c0b56b3ec4722d890232a32528ae202abd76a4",
  "publish_ready": false,
  "version": "7.16.0"
}
```

3. Save this go program as `main.go`:
```go
package main

import (
	"context"
	"log"
	"time"

	"go.elastic.co/apm"
)

func main() {
	tx := apm.DefaultTracer.StartTransaction("apm-server-7.16.0", "waitReady")
	span, ctx := apm.StartSpan(ctx, "capture-error", "waitReady")
	span.Duration = time.Second
	span.End()
	tx.Duration = 2 * time.Second
	tx.End()
	apm.DefaultTracer.Flush(nil)
	log.Println("done")
}
```

4. Run it with `go run main.go`:
```
2021/11/04 11:50:50 sent {Errors:{SetContext:0 SendStream:0} ErrorsSent:0 ErrorsDropped:0 TransactionsSent:1 TransactionsDropped:0 SpansSent:1 SpansDropped:0}
2021/11/04 11:50:50 done
```

5. Navigate to Kibana dev-tools, no documents have been indexed:
```javascript
GET apm*transaction,apm*span,traces*/_search
{
  "query": {
    "match_all": {}
  }
}
```

```json
{
  "took" : 2,
  "timed_out" : false,
  "_shards" : {
    "total" : 2,
    "successful" : 2,
    "skipped" : 0,
    "failed" : 0
  },
  "hits" : {
    "total" : {
      "value" : 0,
      "relation" : "eq"
    },
    "max_score" : null,
    "hits" : [ ]
  }
}
```

6. Install the integration
7. Verify that documents are indexed
```javascript
GET apm*transaction,apm*span,traces*/_search
{
  "_source": [false], 
  "fields": [
    "span.type",
    "transaction.type"
  ], 
  "query": {
    "match_all": {}
  }
}
```

```json
{
  "took" : 3,
  "timed_out" : false,
  "_shards" : {
    "total" : 3,
    "successful" : 3,
    "skipped" : 0,
    "failed" : 0
  },
  "hits" : {
    "total" : {
      "value" : 2,
      "relation" : "eq"
    },
    "max_score" : 1.0,
    "hits" : [
      {
        "_index" : ".ds-traces-apm-default-2021.11.04-000001",
        "_type" : "_doc",
        "_id" : "53UY6XwBPIMMLtjvKObN",
        "_score" : 1.0,
        "_source" : { },
        "fields" : {
          "span.type" : [
            "waitReady"
          ]
        }
      },
      {
        "_index" : ".ds-traces-apm-default-2021.11.04-000001",
        "_type" : "_doc",
        "_id" : "6HUY6XwBPIMMLtjvKObN",
        "_score" : 1.0,
        "_source" : { },
        "fields" : {
          "transaction.type" : [
            "waitReady"
          ]
        }
      }
    ]
  }
}
```

Screen Shot 2021-11-04 at 12 00 23 PM

PR #6148, opened 2021-09-09
Add architecture markdown

simitt:
## Motivation/summary
To improve the onboarding experience a dedicated architecture markdown is added.
This adds the first diagram, outlining the ingest flow.

[Preview](https://github.com/elastic/apm-server/blob/b1b6e6df22f66cd2e3fb0568920e9c0051e72ce8/ARCHITECTURE.md)

Keen on getting feedback!


axw (commented):
Looks good! A few comments:

Jaeger should be added to the main gRPC input.

I wonder if we should combine the middleware and interceptors into "middleware (HTTP) / interceptors (gRPC)". There should be equivalents in each, except for a couple where it doesn't make sense (e.g. CORS).

I'm not convinced it's worth having a separate middleware group for RUM vs. backend either. I'd drop the KillSwitch one from the diagram, and move "RUM specific" to the main middleware group. Should be clear that it only applies to RUM.

"Authorization" middleware should probably be just "Auth". The middleware authenticates, and stores information in the request context for subsequent authorization.

Similar to above, not convinced it's worth having three separate groups of processors. They're essentially the same for OTLP, Jaeger, and Elastic APM agents. For RUM agents we do add source mapping, etc., but as above I think we can just call this out as "RUM specific processors (e.g. source mapping)"?

Perhaps rename "Publisher pipeline" to "output (e.g. Elasticsearch, Kafka)"
Do you think it would be worth mentioning that we rely on libbeat here?

simitt:
Thanks for the suggestions Andrew, I updated the diagram. 

Regarding the Publisher pipeline, I believe it makes sense to signal that the APMEvents go into a pipeline at this point from which they are eventually fetched for sending to the output. I added a note of libbeat. 

What do you think?

bmorelli25:
I'm not sure that I can offer a useful review, but holy guacamole, this is _really_ neat.

axw (approved):
Looks good :)

This is a great way to start the architecture document. We can later fill out the doc with more detail on each of these boxes, cross referencing the diagram.

> Regarding the Publisher pipeline, I believe it makes sense to signal that the APMEvents go into a pipeline at this point from which they are eventually fetched for sending to the output. I added a note of libbeat.

OK, I think it's fine as it is now.

I'm mostly objecting to the word "pipeline", as we don't make use of (or support) libbeat processors now. I think the important things to call out are: we rely on libbeat's queuing and outputs, and this is where things become asynchronous.

Maybe let's see where the modelindexer effort goes, and we can think about revising this bit.

PR #6173, opened 2021-09-15
docs: document internal metrics

### LINKED ISSUE #5084: [doc] Document server-generated metrics

axw:
## Motivation/summary

Document internal metrics: span breakdown, transaction histogram, and service destination.

I copied some content from https://www.elastic.co/guide/en/apm/agent/java/current/metrics.html#metrics-application. I did not document `transaction.duration.*`, as it doesn't look like we're using it anywhere; I have an open question out to see if we can drop it.

I've copied a real metric document produced by opeans-java, excluding cloud metadata for privacy and brevity. The example now looks more like an Elasticsearch document and represents real metrics rather than something contrived. Finally, this means we can get rid of some crufty system tests that were hanging around only for generating these metric docs.

## Checklist

## How to test these changes

N/A

## Related issues

Closes #5084


simitt (approved):
This is very valuable! It will help server devs as well as users trying to leverage metrics for their own visualizations!

bmorelli25 (approved):

PR #6186, opened 2021-09-16
[mergify] add backport policy and default label

v1v:
## Motivation/summary

Notify PR contributors about the backport policy.

This policy will notify what are the different backport options and tag the PR with the `backport-skip`.

We can think about enforcing this rule to do not allow merges but for the time being, what do you think?

## Test

See https://github.com/elastic/apm-pipeline-library/pull/1281#issuecomment-922766344 and https://github.com/elastic/apm-pipeline-library/pull/1281#event-5327050151

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -114,7 +114,6 @@ pull_request_rules:
           * `backport-7./d` is the label to automatically backport to the `7./d` branch. /d is the digit
 
           NOTE: `backport-skip` has been added to this pull request.
-          ```
       label:
         add:
           - backport-skip
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -104,6 +104,7 @@ pull_request_rules:
   - name: notify the backport policy
     conditions:
       - -label~=^backport
+      - -base=master
     actions:
       comment:
         message: |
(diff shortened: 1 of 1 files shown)


kuisathaverat (approved):

axw (approved):
Nice, thanks @v1v 

> We can think about enforcing this rule to do not allow merges but for the time being, what do you think?

I think this is a good step. If the nudge isn't enough, then we can come back to blocking merges.

.mergify.yml
```diff
@@ -100,4 +100,21 @@ pull_request_rules:
         - label=automation
         - head~=^update-beats
     actions:
-      delete_head_branch:
\ No newline at end of file
+      delete_head_branch:
+  - name: notify the backport policy
+    conditions:
+      -label~=^backport
```
v1v, line 106:
```suggestion
      - label~=^backport
```

.mergify.yml
```diff
@@ -100,4 +100,21 @@ pull_request_rules:
         - label=automation
         - head~=^update-beats
     actions:
-      delete_head_branch:
\ No newline at end of file
+      delete_head_branch:
+  - name: notify the backport policy
+    conditions:
+      - label~=^backport
```
v1v, line 106:
```suggestion
      - -label~=^backport
```

.mergify.yml
```diff
@@ -100,4 +100,21 @@ pull_request_rules:
...
+      comment:
+        message: |
+          This pull request does not have a backport label. Could you fix it @{{author}}? 🙏
+          To fixup this pull request, you need to add the backport labels for the needed
+          branches, such as:
+          * `backport-7.x` is the label to automatically backport to the `7.x` branch.
+          * `backport-7./d` is the label to automatically backport to the `7./d` branch. /d is the digit
+
+          NOTE: `backport-skip` has been added to this pull request.
+          ```
```
v1v, line 117:
```suggestion
```

v1v:
There is a missing bit regarding skipping the backports validation

PR #6200, opened 2021-09-21
intake: Handle `transaction.dropped_spans_stats`

### LINKED ISSUE #5850: [spans dropped stats] Extend intake to support stats for dropped spans
As specified in [this APM agents spec](https://github.com/elastic/apm/blob/master/specs/agents/handling-huge-traces/tracing-spans-dropped-stats.md), there might be different reasons for spans being dropped by APM agents. 

In this case agents should collect statistics about the dropped spans in a `transaction` object as follows:
```
{
  ... ,

  "dropped_spans_stats": [
    {
      "type": "external",
      "subtype": "http",
      "destination_service_resource": "example.com:443",
      "outcome": "failure",
      "duration.count": 28,
      "duration.sum.us": 123456
    },
    {
      "type": "db",
      "subtype": "mysql",
      "destination_service_resource": "mysql",
      "outcome": "success",
      "duration.count": 81,
      "duration.sum.us": 9876543
    }
  ]
}
```
 

This issue is about extending the APM server's intake API to accept this kind of data and store it with the transaction documents in ES.

marclop:
## Motivation/summary

Adds a new `transaction.dropped_spans_stats` optional field to the API,
which accepts an array of spans which were due to transaction_max_spans
or exit_span_min_duration being exceeded. Additionally, the metrics
aggregator for the `destination_service` metricset has been updated to
process transaction where `transaction.dropped_spans_stats > 0`.

```javascript
{
  "transaction": {
    "dropped_spans_stats": [
      {
        "type": "external",
        "subtype": "http",
        "destination_service_resource": "example.com:443",
        "outcome": "failure",
        "duration": {
            "count": 28,
            "sum.us": 123456
        }
      },
      {
        "type": "db",
        "subtype": "mysql",
        "destination_service_resource": "mysql",
        "outcome": "success",
        "duration": {
            "count": 81,
            "sum.us": 9876543
        }
      }
    ]
  }
}
```

## Checklist

## How to test these changes

TODO

## Related issues

Closes elastic/apm-server#5850


systemtest/main_test.go
```diff
@@ -24,11 +24,16 @@ import (
...
 	if err := StartStackContainers(); err != nil {
 		log.Fatal(err)
 	}
+	log.Println("INFO: done.")
+	log.Println("INFO: setting up fleet...")
 	if err := Fleet.Setup(); err != nil {
 		log.Fatal(err)
 	}
+	log.Println("INFO: done.")
+	log.Println("INFO: running system tests...")
```
marclop, line 37: Added these logging statements since they looked like nice output to have, but I can remove them from the PR since it's not strictly tied to this change.

model/transaction.go
```diff
@@ -145,3 +156,31 @@ func (m TransactionMark) fields() common.MapStr {
...
 	return out
 }
+
+type DroppedSpanStats struct {
+	Type                       string
+	Subtype                    string
+	DestinationServiceResource string
+	Outcome                    string
+	Count                      *int
+	DurationSumUs              *int // microseconds
```
axw, line 166: It would be nice if we could use the recently introduced `model.AggregatedDuration` hwere. I've asked about changing the fields to accommodate that: https://github.com/elastic/apm-server/issues/5850#issuecomment-924627435

systemtest/approvals.go
```diff
@@ -71,6 +71,7 @@ var apmEventSortFields = []string{
 	"span.id",
 	"error.id",
 	"timeseries.instance",
+	"span.destination.service.resource",
```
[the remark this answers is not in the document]
axw: No worries. I think when we start using time series mode, we can use _tsid instead.

systemtest/huge_traces_test.go
```diff
@@ -0,0 +1,54 @@
...
+	srv.Config.Aggregation = &apmservertest.AggregationConfig{
+		ServiceDestinations: &apmservertest.ServiceDestinationAggregationConfig{
+			Enabled:  true,
+			Interval: time.Second,
+		},
+	}
+	err := srv.Start()
+	require.NoError(t, err)
+
+	systemtest.SendBackendEventsPayload(t, srv, "../testdata/intake-v2/transactions.ndjson")
```
axw, line 43: It would be great if we could isolate only the relevant transactions/spans. Perhaps eventually we should use the Go agent directly, if/when we've implemented support for composite and dropped spans stats there.

systemtest/main_test.go
axw: Looks good. Maybe drop the "done" logs, as they're implied by the immediately following logs.

axw (commented):
Looks great! Let's update the model to use `model.AggregatedSum` now that Felix has agreed

systemtest/huge_traces_test.go
marclop: I was unsure whether to copy the transaction that I added in the file or use the whole file, it does create unnecessary data in Elasticsearch, I could do that and add a comment to switch to the Agent implementation when available.

We could also add a filter on fields to these ingesting functions so that we can filter on the test. I can change the signature to accept a variadic string argument which can be used to filter on "<key>=<val>".

axw: How about we introduce a new testdata file with just one transaction with dropped spans for now (then no need to filter, and some other changes to tests can be reverted I think), and a TODO to switch over to the Go agent when available?

marclop: sounds good, I'll do that then.

x-pack/apm-server/aggregation/spanmetrics/aggregator_test.go
```diff
@@ -251,16 +247,20 @@ func TestAggregateTransactionDroppedSpansStats(t *testing.T) {
 					Subtype:                    "elasticsearch",
 					DestinationServiceResource: "https://elasticsearch:9200",
 					Outcome:                    "success",
-					Count:                      &esDroppedCount,
-					DurationSumUs:              &esDurationSum,
+					Duration: model.AggregatedDuration{
+						Count: 10,
+						Sum:   1_500_000,
```
axw, line 252:
```suggestion
						Sum:   1500 * time.Microsecond,
```

YMMV, but I find it easier to understand if the units are specified.

x-pack/apm-server/aggregation/spanmetrics/aggregator_test.go
```diff
@@ -251,16 +247,20 @@ func TestAggregateTransactionDroppedSpansStats(t *testing.T) {
...
 				{
 					Type:                       "query",
 					Subtype:                    "mysql",
 					DestinationServiceResource: "mysql://mysql:3306",
 					Outcome:                    "unknown",
-					Count:                      &mySQLDroppedCount,
-					DurationSumUs:              &mySQLDurationSum,
+					Duration: model.AggregatedDuration{
+						Count: 2,
+						Sum:   3_000_000,
```
axw, line 262:
```suggestion
						Sum:   3000 * time.Microsecond,
```

beater/tracing_test.go
```diff
@@ -33,11 +33,12 @@ import (
...
-	"4340a8e0df1906ecbfa9": true,
-	"cdef4340a8e0df19":     true,
-	"00xxxxFFaaaa1234":     true,
-	"142e61450efb8574":     true,
+	"945254c567a5417e":                 true,
+	"4340a8e0df1906ecbfa9":             true,
+	"cdef4340a8e0df19":                 true,
+	"00xxxxFFaaaa1234":                 true,
+	"142e61450efb8574":                 true,
+	"ddf109a4c4aa5f2b6e984548ca57774c": true,
```
axw, line 41: I think we can revert this one too?

axw (approved):
Very nice! LGTM, just a couple of suggestions.

x-pack/apm-server/aggregation/spanmetrics/aggregator_test.go
marclop: ah yes, I partially fixed it on the assertion but forgot to do it here, thanks!

changelogs/head.asciidoc
```diff
@@ -32,6 +33,7 @@ https://github.com/elastic/apm-server/compare/7.15\...master[View commits]
 - HTTP server errors (e.g. TLS handshake errors) are now logged {pull}6141[6141]
 - Span documents now duplicate extended HTTP fields, which were previously only under `span.http.*`, under `http.*` {pull}6147[6147]
 - We now record the direct network peer for incoming requests as `source.ip` and `source.port`; origin IP is recorded in `client.ip` {pull}6152[6152]
+- We now collect span destination metrics for transactions with too many spans (for example due to transaction_max_spans or exit_span_min_duration) when collected and sent by APM agents {pull}6200[6200]
```
marclop, line 36: I added 2 entries, but I can keep maybe this one only, I think they're both relevant, but maybe it's too much for this change.

axw: Personally I don't find much value in the "Intake API Changes" section, but I don't have a problem with it either. What you have added is fine.

simitt: The Intake API changes are mainly ment to provide value to agent developers as reference which apm-server version supports which intake fields.

PR #6203, opened 2021-09-21
bump version

stuartnelson3:
## Motivation/summary

bump version

PR #6205, opened 2021-09-21
7.14 update beats

stuartnelson3:
## Motivation/summary

I encountered this problem while bumping the version in https://github.com/elastic/apm-server/pull/6203. I don't know if this is necessary, however, and running `make update-beats` on `7.x` does not encounter this same problem.


marclop (approved):
Looks good, seems to fix the build

PR #6236, opened 2021-09-23
sampling: maintain a single Badger database for process lifetime

axw:
## Motivation/summary

Move Badger outside of the server reload lifecycle, so multiple concurrently existing tail-sampling processors (i.e. overlapping on reload) can access the now process-global, shared, Badger database. The Badger database is opened the first time a tail-sampling processor is created, and is closed when the process exits.

Remove tail-sampling storage directory config; ti is now always "tail_sampling" under the data directory. There is no good reason for it to be configurable, and making it configurable means we can't guarantee a single Badger database for the lifetime of the process.

## Checklist

## How to test these changes

Cannot yet be manually tested. Needs changes to the integration package to test tail-based sampling via Fleet.

## Related issues

None.


x-pack/apm-server/main.go
```diff
@@ -191,12 +222,30 @@ func wrapRunServer(runServer beater.RunServerFunc) beater.RunServerFunc {
 	}
 }
 
+func closeBadger() error {
+	if badgerDB != nil {
+		return badgerDB.Close()
```
stuartnelson3, line 227: does this access need to be coordinated to?

stuartnelson3 (approved):
one question, but that's all :+1:

x-pack/apm-server/main.go
axw: Nope. `closeBadger` is only called at process exit time and never concurrently with `getBadgerDB`. I'll add a comment.

stuartnelson3:
Should we add `test-plan-skip` to this one?

axw:
Yes. Technically it is possible to test via the integration now, but it's not straightforward or supported yet.

Issue #6002, opened 2021-08-20
Create experimental output using go-elasticsearch


simitt:
Introduce an experimental output using go-elasticsearch, avoiding libbeat output. 

see https://github.com/elastic/apm-server/pull/5970

## Goal:
* Get an understanding for involved effort, potential issues and shortcomings
* Get a rough understanding of performance implications

marclop:
# Benchmark results

I benchmarked against an ECE installation comparing the performance of:

* APM standalone (libbeat output) vs APM standalone (experimental output)
* APM standalone vs APM integration

No output settings were tuned.

## Environment

Single node ECE installation on an r5d.2xlarge (64GB of RAM).

Each of the benchmarks was run against a deployment with the following topology:

- 16GB single node Elasticsearch.
- 1GB Kibana instance.
- 4GB APM server.

All benchmarks have been performed running `systemtest/cmd/apmbench` with the benchmark
scenarios defined in: https://github.com/elastic/apm-server/blob/3b228e4e945613125826ea913cee2f340121fc0d/systemtest/cmd/apmbench/main.go#L30-L60

## Summary

### Standalone with libbeat output vs experimental output

There seems to a pretty significant performance gain when running `apm-server` in standalone
mode with the experimental `go-elasticsearch` output enabled vs the default `libbeat` output.

```console
$ benchstat benchresult/7.16.0-standalone.txt benchresult/7.16.0-standalone-experimental.txt
name              old time/op              new time/op              delta
1000Transactions               515ms ± 2%               196ms ± 3%   -61.98%  (p=0.002 n=6+6)
OTLPTraces                     908µs ± 2%               295µs ± 4%   -67.44%  (p=0.002 n=6+6)

name              old events/sec           new events/sec           delta
1000Transactions               1.94k ± 2%               5.11k ± 3%  +163.05%  (p=0.002 n=6+6)
OTLPTraces                     1.09k ± 3%               3.38k ± 4%  +210.20%  (p=0.002 n=6+6)

name              old alloc/op             new alloc/op             delta
1000Transactions               315kB ± 1%               265kB ± 0%   -15.89%  (p=0.002 n=6+6)
OTLPTraces                    1.74kB ± 0%              1.73kB ± 0%    -0.27%  (p=0.004 n=6+5)

name              old allocs/op            new allocs/op            delta
1000Transactions               5.18k ± 0%               5.11k ± 0%    -1.35%  (p=0.002 n=6+6)
[...]

PR #6243, opened 2021-09-26
publish: implement BatchProcessor; remove tracing

axw:
## Motivation/summary

- Modify `publish.Publisher` so that it implements `model.BatchProcessor` directly.
- Remove the tracing instrumentation from publisher. We don't use and don't support libbeat processors, so this is not useful.

## Checklist

## How to test these changes

1. Enable instrumentation
2. Send some events to apm-server
3. Check that no "Publisher" type transactions are traced

## Related issues

Related to https://github.com/elastic/apm-server/issues/6002


beater/beater.go
```diff
@@ -569,8 +569,8 @@ func (s *serverRunner) waitReady(ctx context.Context, kibanaClient kibana.Client
 }
 
 // newFinalBatchProcessor returns the final model.BatchProcessor that publishes events.
-func (s *serverRunner) newFinalBatchProcessor(libbeatReporter publish.Reporter) model.BatchProcessor {
-	return &reporterBatchProcessor{libbeatReporter}
+func (s *serverRunner) newFinalBatchProcessor(p *publish.Publisher) model.BatchProcessor {
+	return p
 }
```
marclop, line 574:
```suggestion
```
Do we need to keep this function? We can just remove it since it doesn't do anything anymore, right? it's just returning the same argument that's passed back to the caller.

marclop (approved):

beater/beater.go
axw: I intend to rebase https://github.com/elastic/apm-server/pull/5970 after this merges, at which point it'll be non-trivial. Depending on the configuration, `newFinalBatchProcessor` will either return a libbeat or go-elasticsearch processor.

marclop:
Tested with `7.16.0 BC2` locally:

1. `./apm-server -E output.elasticsearch.username=admin -E output.elasticsearch.password=changeme -E instrumentation.enabled=true -e`
2. Verify that no results are returned with the following query:
```javascript
GET apm-*,traces-apm*,logs-apm*/_search
{
  "query": {
    "bool": {
      "should": [
        {
          "match": {
            "transaction.type": "Publisher"
          }
        },
        {
          "match": {
            "span.type": "Publisher"
          }
        }
      ]
    }
  }
}
```

PR #6268, opened 2021-09-30
dropped_spans_stats: Remove `type`, `subtype` fields

marclop:
## Motivation/summary

Removes the `dropped_spans_stats.{type,subtype}` fields from the intake
API and sets the `model.APMEvent.Transaction.DroppedSpansStats` to `nil`
after the metricsets have been created for the `dropped_spans_stats`.

Introduces a new `modelprocessor.DroppedSpansStatsDiscarder` since we do
not want to send the `model.APMEvent.Transaction.DroppedSpansStats` to
the published and store it in Elasticsearch, since the only use for the
field is the process the relevant metricsets for spans that are dropped.
To allow any processors before the new processor is run, it is appended
just before the publisher.

## Checklist

- ~[ ] Update [CHANGELOG.asciidoc](https://github.com/elastic/apm-server/blob/master/CHANGELOG.asciidoc)~

## How to test these changes

The system tests cover the base usage, but to test this change:

1. `make && docker compose up -d`
3. Run APM server: `./apm-server -E output.elasticsearch.username=admin -E output.elasticsearch.password=changeme -e`
4. Ingest `testdata/intake-v2/transactions-huge_traces.ndjson `:
    - `curl -H "Content-type: application/x-ndjson" --data-binary @testdata/intake-v2/transactions-huge_traces.ndjson http://localhost:8200/intake/v2/events`

### Run the following query in `Dev Tools > Console`
```
GET /apm*-span,apm*-transaction/_search
{
  "query": {
    "term": {
      "transaction.id": "ddf109a4c4aa5f2b6e984548ca57774c"
    }
  }
}
```

Assert that the document has no `dropped_spans_stats` field under `_source`.

### Verify that `chatty-service` has the two dependencies listed in the `.ndjson` file
Screen Shot 2021-09-30 at 4 22 37 PM

## Related issues

Follow up from #6200 
Part of amended spec elastic/apm#515

### DIFF
--- a/beater/beater.go
+++ b/beater/beater.go
@@ -465,15 +465,19 @@ func (s *serverRunner) run(listener net.Listener) error {
 	}
 	runServer = s.wrapRunServerWithPreprocessors(runServer)
 
-	batchProcessor := s.newFinalBatchProcessor(reporter)
+	batchProcessor := make(modelprocessor.Chained, 0, 3)
 	if !s.config.Sampling.KeepUnsampled {
 		// The server has been configured to discard unsampled
 		// transactions. Make sure this is done just before calling
 		// the publisher to avoid affecting aggregations.
-		batchProcessor = modelprocessor.Chained{
-			sampling.NewDiscardUnsampledBatchProcessor(), batchProcessor,
-		}
+		batchProcessor = append(batchProcessor,
+			sampling.NewDiscardUnsampledBatchProcessor(),
+		)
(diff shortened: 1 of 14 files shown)


axw (approved):
LGTM. I think it could be useful to have transaction-level dropped span stats, but we can always add them back in (i.e. stop dropping) later when we have a specific use case in mind.

Issue #6269, opened 2021-09-30
[tail sampling] Improve monitoring metrics


simitt:
## Current behavior
Three event based metrics are recorded: `processed`, `dropped` and `stored`.

* `processed`: counts all `transaction` and `span` events, independent of them being dropped or indexed eventually. It does not count any other event type (`metric`, `error`, `profile`). If an error occurs during batch processing, then this is the only metric that gets updated.
* `dropped`: events that are neither reported nor stored for later, to track events that are definitely dropped. This metric also includes non trace events (`error`, `metric`, `profile`), despite them not actually being dropped. 
* `stored`: events that are locally stored with a TTL; if  a positive sampling decision is made for the trace to which the events belong, before the TTL is reached, they are indexed, otherwise they are dropped (but do not increase the dropped counter)

Ideally we have some metrics that can be used to verify the tail based sampling algorithm. 

## Suggested behavior
* `processed`: all trace events (stays as is)
* `sampled`: all trace events that are eventually indexed
* `head_unsampled`: all trace events that were not sampled by the head based sampling
* `dropped`: all trace events that are neither reported nor stored for later, to track events that are definitely dropped. 
* `stored`: stays as is; might be interesting to compare to `sampled` to sse how many events are directly indexed and how many are locally stored first
* `unprocessed`: all non-trace events (they are immediately forwarded to ES)

eventually dropped trace events could be calculated by `eventually_dropped = processed - sampled - head_unsampled`

bryce-b:
it might be nice to have "eventually_dropped" as an explicit metric recorded, so it doesn't have to be inferred. Maybe call it "timed_out" ?

axw:
What would we use `unprocessed` for?

simitt:
Good question, given that the sampling logic does not touch these events at all, probably not worth collecting the metrics. I guess I wasn't too sure if they should be captured as they are currently part of the `dropped` metrics (but I think that's just a bug). 
Updated the proposal.

axw:
> Good question, given that the sampling logic does not touch these events at all, probably not worth collecting the metrics. I guess I wasn't too sure if they should be captured as they are currently part of the dropped metrics (but I think that's just a bug).

Yep, I also think that's bug; it should only count processed events that were not stored.

PR #6273, opened 2021-10-01
tail sampling: Change monitoring metrics

simitt:
## Motivation/summary
Introduce `sampled`, `head_unsampled` and `unprocessed` metrics for tail based sampling metrics collection.
Adapt logic for `dropped` to only contain trace events.

## Checklist

## How to test these changes
Start APM Server with tail based sampling configuration and monitoring enabled. Send requests and verify the recorded metrics. 

## Related issues

implements #6269

cc @bryce-b


marclop (approved):
Looks good

x-pack/apm-server/sampling/processor.go
```diff
@@ -164,7 +170,6 @@ func (p *Processor) ProcessBatch(ctx context.Context, batch *model.Batch) error
 			events = events[:n-1]
 			i--
 		}
-		p.updateProcessorMetrics(report, stored)
```
axw, line 167: I think you uncovered a more serious bug. IIANM, ProcessBatch is dropping metrics. I think we should leave this call to updateProcessorMetrics here, and update the switch to just skip over non transaction/span events.

axw (changes requested):

PR #6280, opened 2021-10-04
[Documentation] Update OpenTelemetry docs section on limitation

cyrille-leclerc:
Update OpenTelemetry docs section on limitation

## Motivation/summary

Report on which version of Elastic some problems are fixed.

## Checklist

## How to test these changes

N/A

## Related issues

None

### DIFF
--- a/docs/guide/opentelemetry-elastic.asciidoc
+++ b/docs/guide/opentelemetry-elastic.asciidoc
@@ -377,7 +377,7 @@ Here is an example of AWS Lambda Node.js function managed with Terraform and the
 [[elastic-open-telemetry-traces-limitations]]
 ===== OpenTelemetry traces
 
-* Traces of applications using `messaging` semantics might be wrongly displayed or not shown in the APM UI. You may only see `spans` coming from such services, but no `transaction` https://github.com/elastic/apm-server/issues/5094[#5094]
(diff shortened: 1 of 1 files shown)


axw:
We don't normally list bug fixes throughout the docs. They're already mentioned in the [release notes](https://www.elastic.co/guide/en/apm/server/current/release-notes-7.15.html).

cyrille-leclerc:
Thanks @axw , the problem I would like to fix is that I met few users reporting "Elastic doesn't work with OpenTelemetry Metrics" and I would like to make it clear in our docs that they have to bump to Elastic 7.15.
Would you have recommendations on the way to solve this?

simitt:
+1 on @axw 's point. The docs are versioned, if users have a `7.14` deployment running, they need to refer to the `7.14` docs, where they will find the limitations for their version listed. When a limitation was fixed in `7.15` we should remove it from the `7.15` docs. Describing that the deployment needs to be upgraded to `7.15` in the docs for this version is very confusing. 

In case you are surfacing that it is not obvious to users to select the docs aligned with their deployment version, then that would be a larger problem, that should not be partially handled in one specific section of the docs.

simitt:
pulling in @bmorelli25, as he might have some idea how to better highligt the limitations and progress on specific topics.

bmorelli25:
I agree that it doesn't make sense to add this to the 7.15 docs. If anything, shouldn't we remove those two limitations now that they've been fixed?

We generally don't update unmaintained docs (7.14), but if you think it'd be helpful, we can add a "known issues" section to the release notes that explains that upgrading to 7.15 will fix these problems.

PR #6283, opened 2021-10-04
systemtest: Fix flaky `TestRUMXForwardedFor`

### LINKED ISSUE #6281: [flaky test] systemtest TestRUMXForwardedFor occasionally fails
The systemtest `TestRUMXForwardedFor` occasionally fails when resovling the `geo` information. This might happen when the tests are run on a different instance. If I remember correctly, @stuartnelson3 did some first analysis on this a while back. 

```
[2021-10-04T09:14:16.190Z] === RUN   TestRUMXForwardedFor

[2021-10-04T09:14:16.190Z]     rum_test.go:62:   map[string]interface{}{

[2021-10-04T09:14:16.190Z]           	"events": []interface{}{

[2021-10-04T09:14:16.190Z]           		map[string]interface{}{

[2021-10-04T09:14:16.190Z]           			"@timestamp": string("dynamic"),

[2021-10-04T09:14:16.190Z]           			"agent":      map[string]interface{}{"name": string("rum-js"), "version": string("5.5.0")},

[2021-10-04T09:14:16.190Z]           			"client": map[string]interface{}{

[2021-10-04T09:14:16.190Z]           				"geo": map[string]interface{}{

[2021-10-04T09:14:16.190Z]           					... // 2 identical entries

[2021-10-04T09:14:16.190Z]           					"country_iso_code": string("AU"),

[2021-10-04T09:14:16.190Z]           					"country_name":     string("Australia"),

[2021-10-04T09:14:16.190Z]           					"location": map[string]interface{}{

[2021-10-04T09:14:16.190Z]         - 						"lat": float64(-31.9474),

[2021-10-04T09:14:16.190Z]         + 						"lat": float64(-31.9674),

[2021-10-04T09:14:16.190Z]         - 						"lon": float64(115.8648),

[2021-10-04T09:14:16.190Z]         + 						"lon": float64(115.8621),

[2021-10-04T09:14:16.190Z]           					},

[2021-10-04T09:14:16.190Z]           					"region_iso_code": string("AU-WA"),

[2021-10-04T09:14:16.190Z]           					"region_name":     string("Western Australia"),

[2021-10-04T09:14:16.190Z]           				},

[2021-10-04T09:14:16.190Z]           				"ip": string("220.244.41.16"),

[2021-10-04T09:14:16.190Z]           			},

[2021-10-04T09:14:16.190Z]           			"ecs":   map[string]interface{}{"version": string("dynamic")},

[2021-10-04T09:14:16.190Z]           			"event": map[string]interface{}{"ingested": string("dynamic"), "outcome": string("unknown")},

[2021-10-04T09:14:16.190Z]           			... // 8 identical entries

[2021-10-04T09:14:16.190Z]           		},

[2021-10-04T09:14:16.190Z]           	},

[2021-10-04T09:14:16.190Z]           }

[2021-10-04T09:14:16.190Z]         

[2021-10-04T09:14:16.190Z]         Run `make update check-approvals` to verify the diff

```

marclop:
## Motivation/summary
Changes the systemtest `TestRUMXForwardedFor` to not assert the exact ES
document contents of the ingested RUM transaction, instead, it ignores
the contents of `client.geo.location` but asserts that the field is set,
has "lat" and "lon" properties set, and the value of the IP address.

## How to test these changes

1. `cd systemtest`
2. ` go test -count=10 -v -run ^TestRUMXForwardedFor$ .`

## Related issues

Closes #6281


systemtest/rum_test.go
```diff
@@ -65,7 +66,23 @@ func TestRUMXForwardedFor(t *testing.T) {
...
+	var ip string
+	for _, doc := range result.Hits.Hits {
+		client := doc.Source["client"].(map[string]interface{})
+		ip = client["ip"].(string)
+		geo = client["geo"].(map[string]interface{})
+	}
+	assert.NotZero(t, geo)
+	assert.Contains(t, geo, "location")
+	assert.Contains(t, geo, "region_iso_code")
+	assert.Contains(t, geo, "country_iso_code")
```
marclop, line 84: I chose to assert these fields being set since I feel like they won't likely change the format, but can also just remove the assertion.

systemtest/rum_test.go
```diff
@@ -65,7 +66,23 @@ func TestRUMXForwardedFor(t *testing.T) {
 		"@timestamp", "timestamp.us",
 		// RUM events have the source port recorded, and in the tests it will be dynamic
 		"source.port",
+		// Do not assert the exact contents of the geo field since they can change slightly
+		// depending on the IP lookup.
+		"client.geo",
```
axw, line 71:
```suggestion
		// Do not assert the exact contents of the geo field since they can change slightly
		// depending on the IP lookup.
		"client.geo.location",
```

IMO we should either only disregard `client.geo.location`, or not use approvals at all. The test is only about making sure X-Forwarded-For makes it through to geoIP enrichment.

marclop: I'll amend that. I initially went that route but then thought we might not want to assert the contents at all, I'll revert to what I asserted initially.

systemtest/rum_test.go
```diff
@@ -65,7 +66,21 @@ func TestRUMXForwardedFor(t *testing.T) {
...
+	var location = make(map[string]interface{})
+	var ip string
+	for _, doc := range result.Hits.Hits {
+		client := doc.Source["client"].(map[string]interface{})
+		ip = client["ip"].(string)
+		location = client["geo"].(map[string]interface{})["location"].(map[string]interface{})
+	}
+	assert.Contains(t, location, "lat")
+	assert.Contains(t, location, "lon")
+	assert.Equal(t, ipAddress, ip)
```
axw, line 83: We know that geoIP enrichment has been effective due to the `city_name` etc. being set. We also see the IP address in the approbal. Are these new assertions (lines 74+) testing anything additional?

axw (approved):
LGTM in that it fixes the issue, but I think we probably don't need the additional assertions. Thanks for the speedy fix :)

systemtest/rum_test.go
marclop: Probably not adding anything and just overcasting this test case, I'll remove them since it's clear that we're already asserting the output with the document check.

PR #6305, opened 2021-10-07
processor/otel: don't truncate db.statement

### LINKED ISSUE #6303: db statement is truncated
https://github.com/elastic/apm-server/blob/a275ed9dc843762b42737a108b47533afa1eef70/processor/otel/traces.go#L535

when string is a db.statement the value is truncated, but db.statement is a text type in the ES data. It should not be truncated before knowing if it is a keyword or a text.

Either we should truncate only when we are sure it is a keyword (truncate in every case statement) or for the db statement we use v.StringVal() as a value directly

axw:
## Motivation/summary

Database statements are not keywords, and should not be truncated.

## Checklist

## How to test these changes

Send an OTel span with a `db.statement` longer than 1024 characters, check that the whole statement is stored and is not truncated.

## Related issues

Closes #6303

### DIFF
--- a/changelogs/head.asciidoc
+++ b/changelogs/head.asciidoc
@@ -18,6 +18,7 @@ https://github.com/elastic/apm-server/compare/7.15\...master[View commits]
 
 [float]
 ==== Bug fixes
+- processor/otel: don't truncate `db.statement` {pull}6305[6305]
 
 [float]
 ==== Intake API Changes


cdroulers:
@axw does backport mean there will be a patch release of 7.15?

axw:
@cdroulers there should be one or two 7.15.x releases before 7.16.0 is released. There should be a 7.15.1 release soon -- probably in the next couple of weeks.

Issue #6259, opened 2021-09-29
OpenTelemetry Span attributes mapping for agent bridges


SylvainJuge:
Related specification PR : https://github.com/elastic/apm/pull/516

In order to minimize OpenTelemetry mapping complexity in agents, OTel attributes will be provided as-is in `otel.attributes` property for both span and transaction objects.

The mapping of those attributes should be identical to the one that is applied with OTLP intake.

axw:
Main work was completed in #6308, with some leftover bits being tracked in #6354.

PR #6308, opened 2021-10-07
Otel span mapping

stuartnelson3:
## Motivation/summary

translate incoming OTel attributes from agents into the expected apm transaction/span values using the same mapping applied in otlp intake

## Checklist

## How to test these changes

Ingest transactions/spans with an `"otel": {"span_kind": "kind", "attributes": { // insert common attributes here } }` kv pair in the json. Verify that the values in `attributes` have been correctly translated to their expected fields in the spans and transactions.

## Related issues

requested in https://github.com/elastic/apm-server/issues/6259


axw:
Tested with BC2. Looks good, but I realise now that we never added `span.kind` to the mapping. I think we should add it as a keyword type, I'll open an issue to follow up on that.

PR #6311, opened 2021-10-11
Use transaction/span timestamp in metric bucketing

### LINKED ISSUE #6254: Timestamp of original event is disregarded for transaction/span metrics

7.15

Currently, when APM Server creates metricsets from transactions and exit spans, it will use its own timer to group and timestamp metrics, without taking the timestamp of the original event into account. This means that if data is delayed, it will be indexed into Elasticsearch as if it was actually observed at a (much) later time. This is inaccurate, and it makes it hard to compare results based on transaction events and transaction metrics. Additionally, it means we cannot use APM Server when generating synthetic data which timestamp lies in the past, which would help us in creating more useful test data sets.

Related: https://github.com/elastic/apm-server/issues/6249

**Steps to reproduce**:

- Run https://github.com/sqren/synthbeans-node/ and compare the results of transaction metrics and transaction events, the former should cover a smaller time range.

axw:
## Motivation/summary

Update transaction and service-destination metrics to use the original event timestamp when aggregating, by adding the timestamp truncated to the aggregation interval to aggregation keys.

This ensures that events that are sent with delay or otherwise with timestamps in the past or future (e.g. synthetic test data) will be aggregated into the appropriate time windows and not in the apm-server's current aggregation time range.

## Checklist

## How to test these changes

1. Send some transactions with timestamps in the past and future
2. Check that the server produces metrics for the time ranges according to those events, and not for the real time at which the events were sent

## Related issues

Fixes #6254

### DIFF
--- a/changelogs/head.asciidoc
+++ b/changelogs/head.asciidoc
@@ -19,6 +19,7 @@ https://github.com/elastic/apm-server/compare/7.15\...master[View commits]
 [float]
 ==== Bug fixes
 - processor/otel: don't truncate `db.statement` {pull}6305[6305]
+- Use timestamp of original events for transaction/span metrics {pull}6311[6311]
 
 [float]
 ==== Intake API Changes
--- a/docs/metricset-indices.asciidoc
+++ b/docs/metricset-indices.asciidoc
@@ -94,6 +94,8 @@ You can filter and group by these dimensions (some of which are optional, for ex
 * `kubernetes.pod.name`: The name of the Kubernetes pod running the service that served the transaction
 --
 
+The `@timestamp` field of these documents holds the start of the aggregation interval.
+
 [float]
 ==== Service-destination metrics
 
@@ -117,6 +119,8 @@ You can filter and group by these dimensions:
 * `service.name`: The name of the service that made the request
 * `service.environment`: The environment of the service that made the request
 --
+
+The `@timestamp` field of these documents holds the start of the aggregation interval.
 
 [float]
 [[example-metric-document]]
--- a/systemtest/aggregation_test.go
+++ b/systemtest/aggregation_test.go
@@ -60,9 +60,10 @@ func TestTransactionAggregation(t *testing.T) {
 
 	// Send some transactions to the server to be aggregated.
 	tracer := srv.Tracer()
+	timestamp, _ := time.Parse(time.RFC3339Nano, "2006-01-02T15:04:05.999999999Z") // should be truncated to 1s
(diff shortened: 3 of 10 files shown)


marclop:
Tested with `7.16.0 BC2`, works correctly:

1. `./apm-server -E output.elasticsearch.username=admin -E output.elasticsearch.password=changeme -e`
2. Run this program `go run main.go`
```go
package main

import (
	"context"
	"log"
	"time"

	"go.elastic.co/apm"
)

func main() {
	ts, err := time.Parse(time.RFC3339, "2021-10-01T00:00:00Z")
	if err != nil {
		panic(err)
	}
	tx := apm.DefaultTracer.StartTransactionOptions("apm-server-7.16.0", "past",
		apm.TransactionOptions{
			Start: ts,
		},
	)

	ts, err = time.Parse(time.RFC3339, "2021-10-10T00:00:00Z")
	if err != nil {
		panic(err)
	}
	span, ctx := apm.StartSpanOptions(ctx, "capture-error", "future", apm.SpanOptions{
		Start: ts,
	})
	span.Duration = time.Second
	span.End()
	tx.Duration = 2 * time.Second
	tx.End()
	apm.DefaultTracer.Flush(nil)
	log.Println("done")
}
```

3. Verify the Transaction metrics are ingested with the same `@timestamp`.
```javascript
GET apm-*,traces-apm*,logs-apm*/_search
{
  "_source": [false], 
  "fields": [
    "@timestamp"
  ], 
  "query": {
    "match": {
      "transaction.type": "past"
    }
  }
}
```

```json
{
  "took" : 8,
  "timed_out" : false,
  "_shards" : {
    "total" : 3,
    "successful" : 3,
    "skipped" : 0,
    "failed" : 0
  },
  "hits" : {
    "total" : {
      "value" : 2,
      "relation" : "eq"
    },
    "max_score" : 0.2876821,
    "hits" : [
      {
        "_index" : "apm-7.16.0-metric",
        "_id" : "JbXz6HwBbWqRD_wQ81bf",
        "_score" : 0.2876821,
        "_source" : { },
        "fields" : {
          "@timestamp" : [
            "2021-10-01T00:00:00.000Z"
          ]
        }
      },
      {
        "_index" : "apm-7.16.0-transaction",
        "_id" : "DLXz6HwBbWqRD_wQa1Zm",
        "_score" : 0.2876821,
        "_source" : { },
        "fields" : {
          "@timestamp" : [
            "2021-10-01T00:00:00.000Z"
          ]
        }
      }
    ]
  }
}

```

PR #6287, opened 2021-10-04
docs: APM integration documentation

bmorelli25:
### Summary

// coming soon

### Build it

```
$GIT_HOME/docs/build_docs --doc $GIT_HOME/apm-server/docs/integrations-index.asciidoc --chunk 2 --open
```

### To do

- [ ] Fix broken old redirects
- [ ] Fix `fields.asciidoc` script which was manually overwritten in ab092ac

PR #6318, opened 2021-10-11
make update

stuartnelson3:
## Motivation/summary

upgrade beats and go 1.17.1

## How to test these changes

non-functional change


axw:
@bmorelli25 @simitt in https://github.com/elastic/apm-server/pull/6318/commits/f10e2b19c386fd4d4c31baf53b5b3c9398fb0ce4 I've updated the Makefile to stop regenerating fields.asciidoc.

I did this mostly because of the changes in https://github.com/elastic/apm-server/pull/6287, which mean that the generated file is broken. After 7.16 we won't be updating the old libbeat fields.yml files anymore, so fields.asciidoc will not be updated either. Hence we can stop regenerating it and just use what's checked in now.

Issue #6170, opened 2021-09-14
Default ILM template for APM uses hot/warm


cconboy:
>>  it looks like APM is setting up a hot/warm ILM template out of the box that is used by default. This means that every new deployment in ESS will have a warm tier and it will auto-create itself whenever Autoscaling is enabled. Can we switch this default ILM policy for APM to be hot-only? Otherwise every single deployment will end up with a Warm tier, even if it’s just for a few APM docs.

If I understand correctly, changing to hot only would lead to growth in hot indices. APM is using hot/warm intentionally as the best available strategy, and has been doing this for >1year, although the interaction with autoscaling is newer. I also understand there to be changes coming in 8.0 that indirectly fix this issue.

pcsanwald:
hey @cconboy thanks for opening the issue. The issue we've seen in the past is that customers are surprised when a warm tier is scaled into existence because of an ILM policy, which can happen for relatively small amounts of data. A representative salesforce case is 00766551, although there are plenty more examples.

The reason this is coming up in the context of autoscaling, is that autoscaling is the thing that is creating the warm tier from a customer's perspective, and since we are talking about enabling autoscaling by default for new deployments, we expect to see a good amount of warm tiers getting autoscaled into existence if we keep our ILM policy configuration as it is today. We have two options:

1) Decide to keep our defaults as-is, and accept that this will be a point of confusion with some customers that support will need to deal with.

2) Change our default policy to hot only or something similar.

cconboy:
@graphaelli @Mpdreamz

simitt:
Thanks for bringing this to our attention. The main reason why we chose to introduce hot/warm by default was that data become less interesting over time and therefore older data are not actively indexed and are expected to appear less frequently in searches.
If the default policy is causing issues, and a hot only default policy is preferred, I don't see a reason why we couldn't change that (preferably from 8.0 on). Could you help me understand though why warm tiers are a source of confusion and less desirable than keeping all data in the hot tier? 

> I also understand there to be changes coming in 8.0 that indirectly fix this issue.

From 8.0 on, APM Server will only write to data streams. There are no current plans to revisit the ILM policy, so as soon as apm data are ingested, I don't think the switch to data streams would change anything.

graphaelli:
> There are no current plans to revisit the ILM policy

In that case I misinformed @cconboy on this point, I see now that it is indeed the same with data streams.  I agree changing this with 8.0+/data streams is tidy and users can customize their policies until then.

[...]

PR #6322, opened 2021-10-12
Remove warm phase from default ILM policy

axw:
## Motivation/summary

Remove warm phase from default ILM policy for standalone. Having a warm phase by default causes warm data nodes to spin up after 30 days when autoscaling is enabled; this is surprising to users, and was unintended.

This does not affect existing deployments unless they explicitly opt into overwriting the ILM policy during server initialisation/setup. Users who wish to have a warm phase and have enabled overwriting will need to customise their ILM policy in `apm-server.yml`.

## Checklist

## How to test these changes

1. Run apm-server in a clean cluster
2. Check that the the ILM policy created does not include a warm phase

## Related issues

Closes https://github.com/elastic/apm-server/issues/6170

### DIFF
--- a/_meta/beat.yml
+++ b/_meta/beat.yml
@@ -468,12 +468,6 @@ apm-server:
                     #max_age: "30d"
                   #set_priority:
                     #priority: 100
-              #warm:
-                #min_age: "30d"
-                #actions:
-                  #set_priority:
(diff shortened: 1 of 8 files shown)


axw:
@bmorelli25 FYI there are a few minor docs updates here.

Issue #6354, opened 2021-10-14
Correctly decode OpenTelemetry Span numbers and slices for agent bridges


stuartnelson3:
Address the remaining issues from https://github.com/elastic/apm-server/pull/6308 mentioned in [this comment](https://github.com/elastic/apm-server/pull/6308#discussion_r728757146):

>I believe we enable https://pkg.go.dev/encoding/json#Decoder.UseNumber in our decoder, meaning all numbers will come in as json.Number. Also, I think arrays will be coming in as []interface{} rather than a slice of a specific type.

>We'll need to either identify integers or keep a mapping of attribute names to their expected type and convert accordingly. Otherwise we won't correctly handle semantic conventions like http.status_code, which is expected to be an integer.

>We can land this as it is, but we should address those issues in a followup before 7.16 goes out. I'd like to see a system test for this feature, covering the different types.

PR #6368, opened 2021-10-18
Fix decoding of numeric and array OTel attributes

axw:
## Motivation/summary

Fix decoding of OTel attributes in Elastic APM agent input:
 - Numerical values come in as `json.Number`
 - Arrays come in as `[]interface{}`
 - We need to store some numbers as ints for semantic convention handling. I've introduced a (hard-coded for now) list of attributes which of "double" type; all other numbers are decoded as ints if possible.
 - `APMEvent.Labels` must be initialised before calling the `processor/otel.Translate*` methods

## Checklist

## How to test these changes

1. (Ideally using a real Elastic agent) send a span with `http.status_code` in `otel.attributes`, and check that it is stored in `http.response.status_code` in the resulting Elasticsearch document
2. Send some floating point attributes, check that they are recorded as floating point under `labels`
3. Send some array values, check that they are recorded under `labels`

## Related issues

Closes https://github.com/elastic/apm-server/issues/6354


stuartnelson3:
Confirmed with BC2

PR #6392, opened 2021-10-20
apmbench: Add `-secure` flag to TLS verify

marclop:
## Motivation/summary

Adds a new `-secure` flag which if set, it will verify the remote server
TLS certificate, it is disabled by default and can also be configured by
setting `ELASTIC_APM_VERIFY_SERVER_CERT`.
## How to test these changes

Run `systemtest/cmd/apmbench` against an ECE environment with a self-signed certificate and verify it works.


marclop:
Sorry, I don't know how I missed that.

PR #6401, opened 2021-10-20
[7.16] Update go.sum

graphaelli:
In an attempt to fix the release build.


graphaelli:
sure thing, I just followed your instructions and it worked perfectly.

PR #6407, opened 2021-10-21
Added support for overwriting the service.name for metricset objects (#6388)

### LINKED ISSUE #6388: Allow to overwrite service.name for metricset documents
As described in https://github.com/elastic/apm/issues/65, in some cases (e.g. JBoss in Java) multiple services can exists within one runtime / for a single APM agent. In these cases, the APM agents (in particular the Java agent) currently reports one generic `service.name` in the metadata object and overwrites the `service.name` (with the specific values) within the individual transactions in v2.

Metrics that are reported for individual services by an APM agent need to be able to specify the service.name for a metricset object (similar as they can do today for transaction object). 

This issue is about:
- adding the `service.name` as an optional field to the v2 intake api for metrics
- In case the APM agent provides a `service.name` with a metrics object, use this value for the metric document instead of the generic value from the metadata object.

tobiasstadler:
## Motivation/summary
Allow to overwrite service.name for metricset objects

## Checklist

## How to test these changes
As far as I know this is not yet supported by any agent (but I have changes for the java agent), so testing is currently not possible.

## Related issues

Fixes #6388

### DIFF
--- a/systemtest/cmd/runapm/main.go
+++ b/systemtest/cmd/runapm/main.go
@@ -1,0 +1,166 @@
+// Licensed to Elasticsearch B.V. under one or more contributor
+// license agreements. See the NOTICE file distributed with
+// this work for additional information regarding copyright
+// ownership. Elasticsearch B.V. licenses this file to you under
+// the Apache License, Version 2.0 (the "License"); you may
+// not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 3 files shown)


tobiasstadler:
@axw can I back ported this to 7.16 or should it be 8.0+ only?

axw:
@tobiasstadler we're in feature freeze for 7.16, so 8.0+ only, sorry.

tobiasstadler:
Ok, no problem

PR #5974, opened 2021-08-19
Copy `error.log.message` or `error.exception.message` to `message`

axw:
## Motivation/summary

Copy `error.log.message` or `error.exception.message` to the ECS field `message`. Add `message` to error fields.

## Checklist

- [x] Update [CHANGELOG.asciidoc](https://github.com/elastic/apm-server/blob/master/CHANGELOG.asciidoc)
~- [ ] Documentation has been updated~

## How to test these changes

1. Send errors to apm-server
2. Check that the `message` field is populated
3. Do this above with `apm-server.data_streams.enabled`, and then navigate to the Logs app and check that errors show up

## Related issues

Closes https://github.com/elastic/apm-server/issues/5971

PR #6450, opened 2021-10-28
apmpackage: Add back "message" to error_logs ds

marclop:
## Motivation/summary

Adds back the ecs base "message" field to as an external `ecs` field to
the `error_logs` data stream.

It causes the errors captured by APM to not have a `message` field which
in turn causes the Logs tab to not display the error message.

## How to test these changes

1. Install the APM Integration
2. Save this program as `main.go` and run with `go run main.go`

```go
package main

import (
	"context"
	"errors"
	"log"

	"go.elastic.co/apm"
)

func main() {
	tx := apm.DefaultTracer.StartTransaction("apm-server-7.16.0", "test-plan")
	span, ctx := apm.StartSpan(ctx, "capture-error", "type")
	err := apm.CaptureError(ctx, errors.New("a terrible error"))
	err.Send()
	span.End()
	tx.End()
	apm.DefaultTracer.Flush(nil)
	log.Println("done")
}
```
3. Navigate to the and verify that an error is ingested and under the `Logs` tab `"a terrible error"` is displayed:

Screen Shot 2021-10-28 at 4 58 53 PM

## Related issues

This bug was introduced by #6359.
Discovered while testing #5974


axw:
Verified with BC2 (duplicated because I ran it twice):

image

PR #6493, opened 2021-11-01
upgrade containerd

stuartnelson3:
## Motivation/summary

upgrade containerd to v1.5.7. this was part of looking at the vulnerability in `github.com/buger/jsonparser` (https://github.com/elastic/apm-server/security/dependabot/go.sum/github.com%2Fbuger%2Fjsonparser/open). The dependency is brought in by `github.com/containernetworking/plugins@v0.9.1`, which in turn is brought in by `github.com/containerd/containerd@v1.5.6`. `github.com/containerd/containerd@v1.5.7` unfortunately still has the old version, but the next release will include the fixed `jsonparser`. While I was looking at it, I figured I could bump the version.

## How to test these changes

non-functional change


axw (approved):

stuartnelson3:
@axw should I add a `7.x` backport label to this? it doesn't have any change, security or otherwise, so I don't know if it's necessary

axw:
@stuartnelson3 no I don't think it's needed, but please do add backport-8.0 to keep the 8.0 branch in sync until we hit FF

PR #6537, opened 2021-11-04
systemtest: `queryAgentConfig` retry on `io.EOF`

marclop:
## Motivation/summary

Adds retrying logic when the response from fleet-server returns `io.EOF`
up to 10 times, with a retry timer of 500 Milliseconds.

## How to test these changes

1. `go test -v -count=10 -run TestAgentConfig .`
2. Check that `agentconfig_test.go:127: fleet returned EOF on read, retry 1/10...` appears on the output and the test ultimately pass.

### DIFF
--- a/systemtest/agentconfig_test.go
+++ b/systemtest/agentconfig_test.go
@@ -19,6 +19,8 @@ package systemtest_test
 
 import (
 	"encoding/json"
+	"errors"
+	"io"
 	"net/http"
 	"net/url"
 	"testing"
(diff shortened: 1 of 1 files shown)


systemtest/agentconfig_test.go
```diff
@@ -117,6 +119,15 @@ func queryAgentConfig(t testing.TB, serverURL, serviceName, serviceEnvironment,
 		req.Header.Set("If-None-Match", etag)
 	}
 	resp, err := http.DefaultClient.Do(req)
+
+	maxRetries := 10
+	var retries int
+	for errors.Is(err, io.EOF) && retries < maxRetries {
+		retries++
+		t.Logf("fleet returned EOF on read, retry %d/%d...", retries, maxRetries)
```
axw, line 127:
```suggestion
		t.Logf("apm-server returned EOF on read, retry %d/%d...", retries, maxRetries)
```

These queries are going to apm-server's APM agent config endpoint.

axw (approved):
I don't love this kind of change - I'd much prefer we figure out why there are EOFs and put an end to that. But this has been a bit of a hindrance, so we can come back to that.

systemtest/agentconfig_test.go
marclop: sorry, I had misread the test and thought we were querying the Agents endpoint on Fleet. I agree, that this is not a great fix. I had only resorted to a retry since I thought it was an external dependency that wasn't coming up in time. I've opened an issue to determine the root cause of this and fix it, instead of relying on the retry.

PR #6553, opened 2021-11-08
Document data streams and custom index lifecycle policies

bmorelli25:
### Summary

* Adds documentation for data streams
* Adds documentation for default index lifecycle polices
* Adds documentation for custom index lifecycle policies with component templates
* Moves all "storage" content under the "Manage storage" heading

### Preview

Manage storage bucket: https://apm-server_6553.docs-preview.app.elstc.co/guide/en/apm/guide/current/manage-storage.html 
Data streams: https://apm-server_6553.docs-preview.app.elstc.co/guide/en/apm/guide/current/apm-data-streams.html
ILM/ILP: https://apm-server_6553.docs-preview.app.elstc.co/guide/en/apm/guide/current/ilm-how-to.html

### Related

For https://github.com/elastic/observability-docs/issues/1073
Closes https://github.com/elastic/observability-docs/issues/1223
Closes https://github.com/elastic/observability-docs/issues/1224


docs/how-to.asciidoc
```diff
@@ -4,20 +4,11 @@
...
 
-include::./ilm-how-to.asciidoc[]
-
 include::./jaeger-integration.asciidoc[]
 
 include::./ingest-pipelines.asciidoc[]
-
-include::./manage-storage.asciidoc[]
-
-include::./apm-tune-elasticsearch.asciidoc[]
```
bmorelli25, line 23: Curious what y'all think about moving this content. This PR creates a top-level bucket called, "Manage Storage":
Screen Shot 2021-11-08 at 12 36 15 PM

docs/ilm-how-to.asciidoc
```diff
@@ -1,18 +1,191 @@
...
+Customize the policy to your liking, and when you're done, click **Save policy**.
+
+[discrete]
+[[data-streams-custom-three]]
+=== Step 3: Create a component template
+
+To apply your new index lifecycle policy to a data stream,
+and to ensure the policy continues to be applied in future versions,
+you must create a component template.
+The component template name must start with `.`, follow the data stream naming scheme,
```
axw, line 95: Where did you get the `.` part from? AFAIK they do not have a dot prefix.

docs/ilm-how-to.asciidoc
```diff
@@ -1,18 +1,191 @@
...
+----
+. Continue to **Review** and ensure your request looks similar to the image below.
+If it does, click **Create component template**.
++
+[role="screenshot"]
+image::images/create-component-template.png[Create component template]
+
+[discrete]
+[[data-streams-custom-four]]
+=== Step 4: Clone and modify the existing index template
```
axw, line 132: You shouldn't need to touch any index templates for customising the ILM policy. Just edit the `@custom` component template, which is used by the Fleet-installed index template.

Did you try that and it didn't work?

docs/how-to.asciidoc
axw: I like it. I'm not sure that "Processing and performance" belongs there though, as that's about more than just storage -- it also covers apm-server's processing overhead.

docs/ilm-how-to.asciidoc
bmorelli25: The dot (and these docs) are adapted from a tutorial that @hop-dev wrote.

docs/ilm-how-to.asciidoc
bmorelli25: No, sorry, I didn't try that. I was basing this off of what @hop-dev wrote for other integrations. I can try that too.

docs/data-streams.asciidoc
```diff
@@ -1,4 +1,86 @@
 [[apm-data-streams]]
-== Data streams
+=== Data streams
 
-// to do: fill with content. placeholder for external links for now
+****
+{agent} uses data streams to store append-only time series data across multiple indices
+while giving users a single named resource for requests.
```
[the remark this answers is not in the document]
bmorelli25: It just means that you can send indexing and search requests directly to a data stream -- which will route the request to backing indices that store the stream’s data.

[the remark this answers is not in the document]
bmorelli25: But maybe that's too confusing for these docs. I can clean this up 👍

docs/ilm-how-to.asciidoc
bmorelli25: I think you're right, @axw. Looking back at the original issue, Mark pointed out that integrations that define their own ILM policy can just edit the `@custom` template: https://github.com/elastic/observability-docs/issues/986#issuecomment-914211466

bmorelli25: I've confirmed I can directly edit the `@custom` component template. But that brings up two thoughts:

1. Is this persisted on upgrade?
2. I don't think this method allows the same level of granularity that creating your own index template does. For example, when I create my own index template, I can add an ILM policy that applies only to a specific namespace, like `traces-apm-production*`. With this method, my custom template is only applied to the default index pattern (all namespaces) of `traces-apm-*`. Maybe that's okay though? IDK. That's my question for you 😄

axw: > Is this persisted on upgrade? (not sure how to test this)

Yes

> I don't think this method allows the same level of granularity that creating your own index template does. For example, when I create my own index template, I can add an ILM policy that applies only to a specific namespace, like traces-apm-production*. With this method, my custom template is only applied to the default index pattern (all namespaces) of traces-apm-*. Maybe that's okay though? IDK. That's my question for you smile

Indeed, this is a good point. I'd separate it into use cases, so we don't lead users through unnecessary complexity when they want to do something simple. The basic use case is when you want to customise the ILM policy for a data type (e.g. traces); this can be done just be editing the `@custom` component template. The advanced use case is to customise the ILM policy for a data type and namespace (e.g. production); this requires introducing a new index template with higher priority.

Probably a good idea to document both, but longer term we need to make all of this less manual.

docs/ilm-how-to.asciidoc
hop-dev: @axw Thanks for pointing this out, I think this is a mistake in my original guide, you're right the component templates don't have a dot prefix. 

@joshdover I can't see any record of us discussing having a dot prefix on the namespace specific component templates, I think we may be able to remove it? If so I'll get a PR in for the docs.

bmorelli25:
* Removed `.` prefix
* Updated the tutorial to edit the default `<data-stream-name>@custom` component template
* Added a new section that briefly explains that it is possible to have more granular, namespace-level index lifecycle policies, and links to the [tutorial in the Fleet User Guide](https://www.elastic.co/guide/en/fleet/7.16/data-streams.html#data-streams-ilm-tutorial) for more information.

axw (approved):
Looks great!

PR #6555, opened 2021-11-09
Remove Python system tests

axw:
## Motivation/summary

Remove old Python-based system tests. The remaining tests are either redundant (already have Go equivalents), or are about to be irrelevant as we remove support for index/ingest management.

This change means that sample event documents will be frozen in time. We should come back to these and replace the formerly generated system test documents with documents from ~real applications, like we did for metrics in https://github.com/elastic/apm-server/pull/6173.

## Checklist


marclop (approved):
yay!

axw:
Unrelated test failure

PR #6556, opened 2021-11-09
[DOCS] Recommend manage_own_api_key ES privilege

tvernum:
The `manage_own_api_key` privilege allows the caller to create,
list and invalidate their own API Keys.

The `manage_api_key` privilege allows the caller to create API keys
(for which the caller will be the owner), as well as list all API keys
in the cluster, and invalidate any of them.

For API Keys in APM Server, it is preferable to use the
`manage_own_api_key` cluster privilege instead of `manage_api_key`.


simitt (approved):

bmorelli25 (approved):
Thanks!

Issue #5994, opened 2021-08-20
Stop writing onboarding indices


simitt:
Stop writing onboarding indices. From 8.0 on apm-server only uses data streams, and no privileges for creating indices should be required for the apm-server ES user. 

Related to https://github.com/elastic/kibana/issues/100672#issuecomment-901855583

Issue #5961, opened 2021-08-18
Remove Elasticsearch setup, default to waiting for data streams to be set up


simitt:
Follow up on https://github.com/elastic/apm-server/issues/5914

In `7.x` an option is introduced to wait for data streams to be set up before writing to data streams. The option is disabled by default. When this option is enabled, APM Server waits for the integration package to be installed before writing any data.

In `8.0` we will 
* remove the ability to setup Elasticsearch
* always write to data streams (remove option)
* turn the option to wait for data streams to be set up on by default. It should still be possible to disable the option for scenarios where APM Server lacks privileges to check the package is installed; in this case the user is responsible for ensuring the package is installed before APM Server is started.

PR #6568, opened 2021-11-10
systemtest: data streams everywhere

axw:
## Motivation/summary

- Force the use of data streams in system tests.
- Remove the onboarding test, which will be irrelevant when we stop indexing these documents.
- (Temporarily?) Remove the data stream migration pipeline test.
  If we add the test back, it will require indexing old events directly in the test,
  rather than having apm-server do it.
- Remove the template coverage test; we now explicitly disable dynamic mapping except for
  metrics. Maybe this test can come back later, but at the moment it's causing friction.

## Checklist

## How to test these changes

N/A

## Related issues

https://github.com/elastic/apm-server/issues/5961
https://github.com/elastic/apm-server/issues/5994


simitt (approved):
Looks good!

PR #6569, opened 2021-11-10
package: Update version to 8.0.0

### LINKED ISSUE #6562: Master branch test failures in TestRUMXForwardedFor tests
We saw a pair of test failures for the TestRUMXForwardedFor tests today in the master branch. Specifically, the failures were:

    Build and Test / System and Environment Tests / github.com/elastic/apm-server/systemtest.TestRUMXForwardedFor/data_streams_enabled
    Build and Test / System and Environment Tests / github.com/elastic/apm-server/systemtest.TestRUMXForwardedFor

Additional info can be found in Jenkins: https://apm-ci.elastic.co/job/apm-server/job/apm-server-mbp/job/master/1335/

marclop:
## Motivation/summary

Updates the APM Package version to 8.0.0 since it was not updated when
the 7.16.0 release got pushed, and after we published the apm package
`8.0.0-dev1` to staging, the tests started installing that version of
the package instead of relying on the local package since the remote
version was higher.

## Checklist

## How to test these changes

Run `go test -v -run TestRUMXForwardedFor .`, verify it doesn't fail.

## Related issues

Closes #6562
Introduced by https://github.com/elastic/package-storage/pull/2728

### DIFF
--- a/systemtest/helpers_test.go
+++ b/systemtest/helpers_test.go
@@ -40,5 +40,7 @@ func withDataStreams(t *testing.T, f func(t *testing.T, unstartedServer *apmserv
 		srv := apmservertest.NewUnstartedServer(t)
 		srv.Config.DataStreams = &apmservertest.DataStreamsConfig{Enabled: true}
 		f(t, srv)
+		err = systemtest.Fleet.DeletePackage(systemtest.IntegrationPackage.Name, systemtest.IntegrationPackage.Version)
+		require.NoError(t, err)
 	})
 }
--- a/systemtest/helpers_test.go
+++ b/systemtest/helpers_test.go
@@ -40,7 +40,5 @@ func withDataStreams(t *testing.T, f func(t *testing.T, unstartedServer *apmserv
 		srv := apmservertest.NewUnstartedServer(t)
 		srv.Config.DataStreams = &apmservertest.DataStreamsConfig{Enabled: true}
 		f(t, srv)
-		err = systemtest.Fleet.DeletePackage(systemtest.IntegrationPackage.Name, systemtest.IntegrationPackage.Version)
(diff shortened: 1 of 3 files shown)


marclop:
We should perhaps also bump the package version to `8.1.0` and skip back porting so this doesn't happen again once 8.0.0 and subsequent patches are released.

systemtest/helpers_test.go
```diff
@@ -40,5 +40,7 @@ func withDataStreams(t *testing.T, f func(t *testing.T, unstartedServer *apmserv
 		srv := apmservertest.NewUnstartedServer(t)
 		srv.Config.DataStreams = &apmservertest.DataStreamsConfig{Enabled: true}
 		f(t, srv)
+		err = systemtest.Fleet.DeletePackage(systemtest.IntegrationPackage.Name, systemtest.IntegrationPackage.Version)
```
axw, line 43: This function will be removed in https://github.com/elastic/apm-server/pull/6568, so it's probably not worthwhile making this change. I don't think we can/should attempt to leave Elasticsearch in a pristine state after tests anyway, but happy to hear arguments.

axw (approved):
> We should perhaps also bump the package version to 8.1.0 and skip back porting so this doesn't happen again once 8.0.0 and subsequent patches are released.

SGTM. Perhaps also introduce a `make check-full` target which verifies the package version == apm-server binary version, excluding patch version?

systemtest/helpers_test.go
marclop: sounds good, seems like we're removing the data streams anyway which was why I added this. I will remove.

simitt:
tested with latest `8.0` snapshot

PR #6575, opened 2021-11-10
Remove ingest pipeline registration

axw:
## Motivation/summary

Remove ingest pipeline registration/configuration. Also, stop blocking indexing of logs when data streams are disabled. Along with the removal of explicitly named pipelines, this allows us to simplify the publisher package.

## Checklist

~- [ ] Documentation has been updated~ Docs will need to be overhauled, I'll leave this for now

## How to test these changes

N/A, removing functionality only.

## Related issues

https://github.com/elastic/apm-server/issues/5961


marclop (approved):
🎉

axw:
Ugh, looks like this breaks integration tests. We were going to have to update them for the followup anyway, so I guess I'm doing that now instead.

PR #6588, opened 2021-11-11
ci: Remove legacy system-test junit XML compress

marclop:
## Motivation/summary

Removes the failing compress step that is always run since the legacy
system-test have been removed.

## Checklist


axw (approved):
thanks!

PR #6591, opened 2021-11-11
Query cluster UUID on apm-server startup

stuartnelson3:
## Motivation/summary

stack monitoring ui in cloud depends on the beat module of metricbeat, which requires apm-server to know its `cluster_uuid` in order for the documents to be indexed. The apm-server only gets its cluster_uuid as a callback to making an initial connection with elasticsearch.

When using datastreams, the initial connection isn't made until it attempts to index a document, which means that stack monitoring ui is broken for apm-server until a document is indexed, even though nothing may be wrong with the apm-server.

Copy the request made in libbeat and execute it as a precondition to get the cluster_uuid as soon as the server is ready.

## Checklist

- [ ] ~~Documentation has been updated~~

## How to test these changes

1. start apm-integration with elasticsearch + kibana
2. install the apm-integration
3. start apm-server: `apm-server -e -E apm-server.data_streams.enabled=true -E http.enabled=true`
4. curl the `/state` internal metrics endpoint: `curl -s http://localhost:5066/state | jq '.outputs.elasticsearch'`
5. you should see non-empty output for the cluster_uuid


beater/beater.go
```diff
@@ -969,3 +976,45 @@ func (r *chanReloader) serve(ctx context.Context, reloader reload.Reloadable) er
...
+// TODO: This is copying behavior from libbeat:
+// https://github.com/elastic/beats/blob/b9ced47dba8bb55faa3b2b834fd6529d3c4d0919/libbeat/cmd/instance/beat.go#L927-L950
+// Remove this when cluster_uuid no longer needs to be queried from ES.
+func queryClusterUUID(ctx context.Context, esClient elasticsearch.Client) error {
+	stateRegistry := monitoring.GetNamespace("state").GetRegistry()
+	outputES := "outputs.elasticsearch"
+	elasticsearchRegistry := stateRegistry.GetRegistry(outputES)
+	if elasticsearchRegistry == nil {
+		elasticsearchRegistry = stateRegistry.NewRegistry(outputES)
+	}
```
stuartnelson3, line 989: Being overly defensive here for test purposes. This function is executed without the side-effects from libbeat, which was causing a nil-registry to be found in some tests.

beater/beater.go
```diff
@@ -969,3 +976,45 @@ func (r *chanReloader) serve(ctx context.Context, reloader reload.Reloadable) er
...
+	// TODO: How can I just update a registry value?
+	clusterUUID := "cluster_uuid"
+	if elasticsearchRegistry.Get(clusterUUID) != nil {
+		elasticsearchRegistry.Remove(clusterUUID)
+	}
+	clusterUUIDRegVar := monitoring.NewString(elasticsearchRegistry, clusterUUID)
```
stuartnelson3, line 995: I looked through the api but it wasn't immediately apparent to me how to update a value already existing in a registry.

beater/beater.go
axw: Perhaps we should create the registry in the test instead? Would be nice not to have test-related things in this code

beater/beater.go
axw: There's a `Get` method on the `monitoring.Registry` which I think you can use?

Would the libbeat code have been called yet at this point? I'm wondering though if there's any risk that the libbeat code could kick in concurrently and panic.

axw (commented):
Generally LGTM, but I'm a bit worried about the timing of the var registration. Is there any possibility that the libbeat code would be called after ours, and cause a panic?

beater/beater.go
stuartnelson3: >There's a Get method on the monitoring.Registry which I think you can use?

roger, I have to cast to `*monitoring.String`, but it appears to work (the return value is `Var` interface type, it seems, which only exposes a single `Visit` method).

>Would the libbeat code have been called yet at this point? I'm wondering though if there's any risk that the libbeat code could kick in concurrently and panic.

The registry code is called synchronously, before `*beater.Run`, so it should always be set up. The `Set` method on the `*monitoring.String` could be called at the same time, but it has an internal mutex, so access should be fine.

axw: Cool, thanks. I see you removed the test guards, looks much safer to me now.

axw (approved):
LGTM!

marclop:
Verified with 9bb5e4cdd:

```console
./apm-server -E output.elasticsearch.username=admin -E output.elasticsearch.password=changeme -E apm-server.kibana.host=http://localhost:5601 -E apm-server.kibana.username=admin -E apm-server.kibana.password=changeme -E apm-server.kibana.enabled=true -E apm-server.expvar.enabled=true -E http.enabled=true
...
$ curl -s http://localhost:5066/state | jq '.outputs.elasticsearch'
{
  "cluster_uuid": "nE_NVzmNRAWVnk5BzWVSJw"
}
```

PR #6618, opened 2021-11-15
CI: archive chocolatey.log

v1v:
## Motivation/summary

Recently in apm-server we've been seeing build failures on Windows related to installing Python via Chocolatey, let's store the `chocolatey.log` file to help with the debugging.


axw:
Thanks @v1v. It just occurred to me that we don't actually need Python for the Windows build now, so I think we should close this in favour of https://github.com/elastic/apm-server/pull/6623

simitt:
We still rely on it for `7.15` though - could we fix this only on the `7.15` branch?

axw:
I have a feeling we can remove Python from Windows for 7.15 too, I'll take a quick look.

axw:
Yep, Python installation now removed from 7.15 too: https://github.com/elastic/apm-server/pull/6671

PR #6631, opened 2021-11-16
model/modelindexer: skip unknown response fields

axw:
## Motivation/summary

Fix a bug in modelindexer's bulk response decoder. We need to skip over unknown/unhandled fields, e.g. ingest_took. This field is not specified in the go-elasticsearch bulk response type, so we don't handle it.

## Checklist

## How to test these changes

1. Run apm-server with `-E output.elasticsearch.experimental=true`
2. Send events
3. Check logs for modelindexer errors

## Related issues

None


marclop:
Verified with 9bb5e4cdd, ingested a ton of events testing a few other PRs and did not see any model indexer errors.

PR #6633, opened 2021-11-16
model: Align `labels` field with ECS

### LINKED ISSUE #3873: model: align labels with ECS

marclop:
## Motivation/summary

Updates all data streams to use the ecs defined `labels` for values that
are `string` or `bool`, and adds a new field for `numeric_labels` which
stores all numeric labels as `scaled_float`, using the `scaling_factor`
we were using for indexes.

There's a single exception for the `metrics-apm.internal` data stream,
due to the fact that `dynamic` is set to `strict`, which disables the
externally defined ecs labels dynamic ability. For that data stream, the
field is defined as a standard field.

## Checklist

## How to test these changes

1. Install the integration.
2. Ingest some data with both numeric, string and bool labels.
3. Verify that the string and bool labels are under the `labels` field
4. Verify that numeric (floats and ints) are stored under `numeric_labels`.

## Related issues

Closes #3873


model/apmevent.go
```diff
@@ -62,17 +62,21 @@ type APMEvent struct {
 
 	// Timestamp holds the event timestamp.
 	//
-	// See https://www.elastic.co/guide/en/ecs/current/ecs-base.html
+	// See https://www.elastic.co/guide/en/ecs/current/ecs-base.html#field-timestamp
 	Timestamp time.Time
 
-	// Labels holds labels to apply to the event.
+	// Labels holds labels to apply to the event.  While it's `map[string]interface{}`,
```
axw, line 68: The model types should be closely resemble the document structure, so we can automatically generate the encoding. We should have two fields: `Labels` and `NumericLabels`.

Would it be reasonable to update this PR to split the labels when translating from incoming events to model types? i.e. in modeldecoder and processor/otel.

model/labels.go
```diff
@@ -18,11 +18,74 @@
...
 	"github.com/elastic/beats/v7/libbeat/common"
 )
 
+func filterLabels(labels common.MapStr, fn func(interface{}) interface{}) common.MapStr {
+	result := common.MapStr{}
+	for k, v := range labels {
+		if typeValue := fn(v); typeValue != nil {
+			if slice, ok := typeValue.([]interface{}); ok {
+				for i, elem := range slice {
+					result[fmt.Sprintf("%s_%d", k, i)] = elem
```
axw, line 35: Any particular reason why we don't just store an array of values? https://www.elastic.co/guide/en/elasticsearch/reference/current/array.html

axw (changes requested):

model/apmevent.go
marclop: I had implemented it like you mention at first but then changed it back to just reuse the `common.MapStr` since it required less changes to the codebase. You raise a valid point that models should be as close to the stored documents, so I think it makes sense to change it back to use two fields.

model/labels.go
marclop: Not sure why, but I thought we were going to only allow a single level of key:value in the object, I'm not sure where I got that from.

apmpackage/apm/data_stream/error_logs/fields/ecs.yml
```diff
@@ -76,12 +76,8 @@
...
 - external: ecs
   name: http.version
-# TODO: change labels to be ECS compliant in 8.0 (https://github.com/elastic/apm-server/issues/3873)
-- name: labels
-  type: object
-  description: |
-    A flat mapping of user-defined labels with string, boolean or number values.
-  dynamic: true
+- external: ecs
+  name: labels
```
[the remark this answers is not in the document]
marclop: you're right, I missed that they won't be indexed if we just reference the ecs fields like that.

model/labels.go
axw: I think it would make sense to just index an array value with the specified field name, rather than exploding the labels. I'd say that will align better with user expectations.

apmpackage/apm/data_stream/error_logs/fields/ecs.yml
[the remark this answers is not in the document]
marclop: Opened https://github.com/elastic/package-spec/issues/251

model/modeldecoder/modeldecoderutil/labels.go
```diff
@@ -53,3 +74,56 @@ func NormalizeLabelValues(labels common.MapStr) common.MapStr {
...
+func filterNumberLabels(v interface{}) interface{} {
+	switch v := v.(type) {
+	case float64:
+		return v
+	case int64:
+		return float64(v)
+	case int32:
+		return float64(v)
+	case int:
+		return float64(v)
```
axw, line 118:
```suggestion
```

IIANM the JSON decoder always decodes numbers as json.Number, and we convert those to float64.

processor/otel/metadata.go
```diff
@@ -222,6 +226,32 @@ func ifaceAnyValueArray(array pdata.AnyValueArray) []interface{} {
 
 // initEventLabels initializes an event-specific label map, either making a copy
 // of commonLabels if it is non-nil, or otherwise creating a new map.
-func initEventLabels(commonLabels common.MapStr) common.MapStr {
-	return commonLabels.Clone()
+func initEventLabels(e *model.APMEvent) {
```
axw, line 229: doc comment needs to be updated to not refer to commonLabels

processor/otel/metadata.go
```diff
@@ -222,6 +226,32 @@ func ifaceAnyValueArray(array pdata.AnyValueArray) []interface{} {
...
+			return
+		}
+		switch v[0].(type) {
+		case string:
+			event.Labels[key] = v
+		case float64:
+			event.NumericLabels[key] = v
+		}
+	default:
+		panic(fmt.Sprintf("Unhandled %T, %+v\n", v, v))
```
axw, line 255: I'd rather we didn't panic in this code. If anything, just log a warning. We should probably remove some of the reflection at some point.

axw (commented):
Just a few minor things, othewise LGTM.

I'd like to follow up on this by changing the Labels and NumericLabels fields to be more type-safe. Either

```go
type APMEvent struct {
    Labels map[string][]string
    NumericLabels map[string][]float64
}
```

Or

```go
// Labels/NumericLabels values are either always stored as arrays,
// or store as an array iff len > 1
type APMEvent struct {
    Labels map[string]LabelValue
    NumericLabels map[string]NumericLabelValue
}

type LabelValue struct {
    Value string
    Values []string // If non-nil, use this; otherwise use Value
}

type NumericLabelValue struct {
    Value float64
    Values []float64 // If non-nil, use this; otherwise use Value
}
```

But to be clear, let's do that in a followup.

axw (approved):
LGTM! :tada:

stuartnelson3:
confirmed with 50edba17e7a6e4c5cb58c6258e1a38dc2387bcdb
```go
package main

import (
	"fmt"
	"os"

	"go.elastic.co/apm"
)

func main() {
	version := "undefined"
	if len(os.Args) > 1 {
		version = os.Args[1]
	}
	tx.Context.SetLabel("enabled", true)
	tx.Context.SetLabel("string", "string")
	tx.Context.SetLabel("numbers", 1234.56)
	tx.End()
	apm.DefaultTracer.Flush(nil)
}
```

Issue #5992, opened 2021-08-20
Align default values between standalone and managed apm-server


simitt:
For easier transition from standalone to managed apm-server, align default values of the standalone server to managed.

axw:
We should align the Elasticsearch output compression level, and enable it by default on-prem as well as in Cloud:

https://github.com/elastic/apm-server/pull/6656#discussion_r752927799

PR #6675, opened 2021-11-19
Align standalone managed

stuartnelson3:
## Motivation/summary

bring standalone apm-server defaults closer inline with managed defaults in https://github.com/elastic/cloud-assets/blob/master/stackpack/apm/config/apm-server.yml

## Checklist

## How to test these changes

verify the default values are in line with that's in cloud

## Related issues

closes https://github.com/elastic/apm-server/issues/5992


beater/config/config.go
```diff
@@ -123,10 +123,10 @@ func DefaultConfig() *Config {
 		MaxHeaderSize:   1 * 1024 * 1024, // 1mb
 		MaxConnections:  0,               // unlimited
 		IdleTimeout:     45 * time.Second,
-		ReadTimeout:     30 * time.Second,
+		ReadTimeout:     3600 * time.Second,
```
stuartnelson3, line 126: This is what I'm seeing in https://github.com/elastic/cloud-assets/blob/master/stackpack/apm/config/apm-server.yml#L77, but this seems pretty excessive. Any opinions?

axw: I think we should leave this one. Its value depends on the deployment of apm-server. In the Cloud case, there's a proxy which impose its own timeout.

beater/config/kibana.go
```diff
@@ -42,7 +42,7 @@ func (k *KibanaConfig) Unpack(cfg *common.Config) error {
 
 func defaultKibanaConfig() KibanaConfig {
 	return KibanaConfig{
-		Enabled:      false,
+		Enabled:      true,
```
axw, line 45: Does it make sense to enable Kibana config in all cases? It may not always be available. Under Fleet, is it _ever_ available?

beater/config/auth.go
```diff
@@ -113,6 +113,8 @@ func defaultAnonymousAgentAuth() AnonymousAgentAuth {
 		Enabled:    false,
 		AllowAgent: []string{"rum-js", "js-base"},
 		RateLimit: RateLimit{
+			// TODO: Cloud is using the deprecated rum.rate_limit, should we update it there?
+			// https://github.com/elastic/cloud-assets/blob/master/stackpack/apm/config/apm-server.yml#L94-L99
```
axw, line 117: Yes, we should use `apm-server.auth.anonymous.rate_limit` for 8.0+.

marclop: oops, that slipped through the cracks when I worked on updating the deprecated settings in cloud-assets (https://github.com/elastic/cloud-assets/pull/834). Do you want me to take care of it?

beater/config/kibana.go
stuartnelson3: My understanding was that we communicate less / not at all with kibana, but the stackpack was enabling it for major version > 7: https://github.com/elastic/cloud-assets/blob/master/stackpack/apm/config/apm-server.yml#L121-L126

axw: I think we'll want kibana enabled in 8.0 for situations where users have upgraded from a previous version, and still running standalone. In that case, we'll still want the Kibana connection for agent config and source maps.

For managed mode, this config file is irrelevant.

beater/config/auth.go
stuartnelson3: That'd be good, if you have time

stuartnelson3: https://github.com/elastic/cloud-assets/pull/860

axw: I commented on that PR, but I think we also need to align the rate limit defaults we have in code and the integration. What's in the integration looks wrong. I think I made an error when trying to align with the cloud template, we should probably update the integration to use the defaults we have in code at the moment.

stuartnelson3: https://github.com/elastic/apm-server/pull/6715

marclop (approved):

marclop:
@stuartnelson3 do you have any suggestions on how to test this change?

stuartnelson3:
Hm, i suppose i'd say to just verify that i copied the values over correctly from https://github.com/elastic/cloud-assets/blob/master/stackpack/apm/config/apm-server.yml?

marclop:
verified that standalone defaults match the cloud standalone defaults.

PR #6676, opened 2021-11-19
package: Verify APM server/package versions match

### LINKED ISSUE #6571: Ensure that package version == `apm-server` version
## Description

We should validate and ensure that the `apm` package corresponds with the `apm-server` MAJOR.MINOR numbers. We can add a some validation in `check-full`

marclop:
## Motivation/summary

Adds a check to `check-package` which ensures the apm-server and package
versions are equal, otherwise returns with an error.

## Checklist

## How to test these changes

N/A

## Related issues

Closes #6571

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -108,7 +108,12 @@ endif
 ## get-version : Get the apm server version
 .PHONY: get-version
 get-version:
-	@grep defaultBeatVersion cmd/version.go | cut -d'=' -f2 | tr -d '"'
+	@grep defaultBeatVersion cmd/version.go | cut -d'=' -f2 | tr -d '" '
+
+## get-version : Get the apm package version
+.PHONY: get-package-version
+get-package-version:
+	@grep ^version: apmpackage/apm/manifest.yml | cut -d':' -f2 | tr -d " "
 
 ##############################################################################
 # Documentation.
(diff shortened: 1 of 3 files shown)


apmpackage/apm/changelog.yml
```diff
@@ -1,6 +1,11 @@
 # newer versions go on top
 #
 # change type can be one of: enhancement, bugfix, breaking-change
+- version: "8.1.0"
+  changes:
+    - description: REPLACE ME WITH ACTUAL CHANGE LOG
```
marclop, line 6: I'm not entirely happy with this, but it seems that the `elastic-package` requires an entry here, otherwise:

```
Lint the package
Error: checking package failed: linting package failed: found 1 validation error:
   1. current manifest version doesn't have changelog entry

make: *** [check-package] Error 1
```

Makefile
```diff
@@ -108,7 +108,12 @@ endif
 ## get-version : Get the apm server version
 .PHONY: get-version
 get-version:
-	@grep defaultBeatVersion cmd/version.go | cut -d'=' -f2 | tr -d '"'
+	@grep defaultBeatVersion cmd/version.go | cut -d'=' -f2 | tr -d '" '
+
+## get-version : Get the apm package version
```
simitt, line 113: `get-package-version`

apmpackage/apm/changelog.yml
simitt: Can you create an issue in the package-spec repo to discuss this lint check? Initially the packages were not supposed to be Elastic stack version aligned, so the hard requirement on a changenote made sense. It should be more lenient though for stack version aligned packages. Maybe something like an _empty_ changenote entry could be added?

We can keep the changelog entry for now.

simitt (approved):

apmpackage/apm/changelog.yml
marclop: Opened https://github.com/elastic/package-spec/issues/254

PR #6677, opened 2021-11-19
update beats: avoid failing if no files have changed

v1v:
### What

Fixes

image

### Why

When some files are not updated, do we want to run the automation? Or its an atomic commit?


simitt (commented):
If no files have changed at all, then there is nothing to update.

v1v:
> If no files have changed at all, then there is nothing to update.

so the go.mod/go.sum and NOTICE.md are not relevant in this case?

.ci/scripts/update-beats.sh
```diff
@@ -11,7 +11,8 @@ COMMIT_MESSAGE="Update to elastic/beats@$(go list -m -f {{.Version}} github.com/
 git checkout -b "update-beats-$(date "+%Y%m%d%H%M%S")"
 git add go.mod go.sum NOTICE.txt \
 	.go-version docs/version.asciidoc \
-	docs/fields.asciidoc include/fields.go x-pack/apm-server/include/fields.go
+	docs/fields.asciidoc include/fields.go x-pack/apm-server/include/fields.go || true
```
marclop, line 14: We should probably use the `--ignore-errors` option for `git add` instead of `|| true`

v1v: I like that one! I just reviewed it's supported since https://git-scm.com/docs/git-add/2.3.10 or even earlier. So we are safe to use it

simitt (approved):

PR #6679, opened 2021-11-21
Remove unused code, move/unexport code with one user

axw:
## Motivation/summary

General tidying up of the codebase.

- Remove code (main_test.go) related to the old Python system tests, which ran a version of the apm-server binary with test coverage.
- Remove `tests.IntPtr`, inline in the few places it was used.
- Remove unused scripts and Mage targets.
- Remove `utility.UrlPath` and `utility.CleanUrlPath`, move (unexported) into the one package they are used.
- Remove `utility.UserAgentHeader`, inline in the one place it was used.
- Remove "request time" middleware and utility functions; add Timestamp to the request.Context type.

## Checklist

## How to test these changes

Non-functional change.

## Related issues

None.


beater/api/mux.go
```diff
@@ -264,7 +264,6 @@ func apmMiddleware(m map[request.ResultID]*monitoring.Int) []middleware.Middlewa
 		middleware.TimeoutMiddleware(),
 		middleware.RecoverPanicMiddleware(),
 		middleware.MonitoringMiddleware(m),
-		middleware.RequestTimeMiddleware(),
```
marclop, line 267: Were we resorting to this middleware because previously the APMEvent didn't have a timestamp we could rely on for the timings?

marclop (approved):

beater/api/mux.go
axw: The event struct has always had a timestamp.

There was nothing preventing us from setting the timestamp on the request context before, we (I think I?) just chose to store it in context instead, probably because middleware seemed like the path of least resistance. Storing it directly on the context is simpler.

simitt (approved):

PR #6682, opened 2021-11-22
model: Use typed maps for `Labels|NumericLabels`

marclop:
## Motivation/summary

Changes the internal representation of `model.{Labels,NumericLabels}` to
make them easier / safer to work with to the following:

```go
type APMEvent struct {
    // ...
    Labels map[string]struct{
        Value string
        Values []string // `nil` when the value is not a slice.
    }
    NumericLabels map[string]struct{
        Value float64
        Values []float64 // `nil` when the value is not a slice.
    }
    // ...
}
```

It seems like the previous change introduced a bug that "lost" some of
the labels. This patch will also fix that.

## Checklist

## How to test these changes

1. `docker-compose up -d`
2. `make`
3. Install the APM Integration in Kibana
4. `./apm-server -E output.elasticsearch.username=admin -E output.elasticsearch.password=changeme`
5. `run:`
<details><summary>labels app</summary><p>

```go
package main

import (
	"context"
	"fmt"
	"os"
	"time"

	"go.elastic.co/apm"
)

func main() {
	version := "undefined"
	if len(os.Args) > 1 {
		version = os.Args[1]
	}

	apm.DefaultTracer.Service.Name = "labels"
	apm.DefaultTracer.Service.Version = version
	apm.DefaultTracer.Service.Environment = "test"

	span.Duration = time.Second
	span.Context.SetLabel("string_label", "value")
	span.Context.SetLabel("bool_label", true)
	span.Context.SetLabel("float_label", float64(1234.1234))
	span.Context.SetLabel("int_label", 1234)
	span.End()
	tx.Duration = 2 * time.Second
	tx.End()
	apm.DefaultTracer.Flush(nil)
}
```

</p></details>

6. Verify that the following labels appear in Kibana under `labels` and `numeric_labels`:
image

## Related issues

Follow up from #6633


stuartnelson3 (approved):

model/labels.go
```diff
@@ -18,11 +18,104 @@
...
 )
 
+type Labels map[string]LabelValue
+
+type LabelValue struct {
+	Value  string
+	Values []string
+}
+
+func (l Labels) MaybeSet(k string, v interface{}) bool {
```
axw, line 34: Looks like Labels.MaybeSet and NumericLabels.MaybeSet are only used by modeldecoderutil, and one place in processor/otel which could use Set + strconv. Can we please move the function there? I'd like to limit our use of reflection as much as possible.

model/labels.go
```diff
@@ -18,11 +18,104 @@
...
 	"strings"
 
 	"github.com/elastic/beats/v7/libbeat/common"
 )
 
+type Labels map[string]LabelValue
+
+type LabelValue struct {
+	Value  string
+	Values []string
```
axw, line 31: This could probably do with some documentation.

model/labels.go
```diff
@@ -18,11 +18,104 @@
...
+		}
+	}
+	return sanitizeLabels(result)
+}
+
+type NumericLabels map[string]NumericLabelValue
+
+type NumericLabelValue struct {
+	Value  float64
+	Values []float64
```
axw, line 78: ditto - docs would be good

axw (commented):
This is great, thank you :)

PR #6685, opened 2021-11-22
Update architecture to new ES output handling

simitt:
## Motivation/summary
Reflect the new ES output handling in the architecture diagram.


axw (approved):
LGTM. Do you want to include libbeat still for other outputs?

simitt:
I can if you prefer to, but not certain we should highlight it as long as it is only supported by the standalone binary.

axw:
Good point, let's leave it out for now.

marclop (commented):
We did remove the jaeger dedicated listeners from 8.0 correct? Should we update the diagram to show that they're only available up to 7.16?

marclop (approved):

PR #6690, opened 2021-11-23
Replace package utility with package netutil

axw:
## Motivation/summary

Package netutil is focused on network-related utility functions, currently all related to splitting and parsing addresses, and extracting addresses from common HTTP proxy headers.

Server logs now set `source.address` to the immediate network peer's IP address, without the port. If the request has X-Forwarded-For, etc., then the log message may also include `client.ip` as well. Previously we set `source.address` to either the immediate network peer's IP:port, or the originating client IP based on XFF etc. The change brings server logs into line with what we set in trace events.

## Checklist

## How to test these changes

1. Run apm-server
2. Use curl to send some requests with and without an X-Forwarded-For header
3. Check the `source.address` field matches the IP address of the host
4. Check the `client.ip` field is only included in requests with X-Forwarded-For, and its value matches the header

## Related issues

None


internal/netutil/netutil.go
```diff
@@ -0,0 +1,171 @@
...
+		}
+		key := strings.TrimSpace(field[:eq])
+		value := strings.TrimSpace(field[eq+1:])
+		if len(value) > 0 && value[0] == '"' {
+			var err error
+			value, err = strconv.Unquote(value)
+			if err != nil {
+				// Malformed, ignore
+				continue
+			}
```
marclop, line 123: just a semantic preference, but I find this easier to read:

```suggestion
			if val, err := strconv.Unquote(value); err == nil {
				value = val
			}
			// Malformed, ignore
```

marclop (approved):
looks great

internal/netutil/netutil.go
axw: This code is just moved, so I'm going to leave this one. I don't think your suggestion is _quite_ right anyway, as "Malformed, ignore" is reached from the happy path.

Issue #5836, opened 2021-07-30
Add more detailed test for fetching sourcemap from multiple fleet servers


estolfo:
As per [this](https://github.com/elastic/apm-server/pull/5770#discussion_r679585900) comment, add a more detailed test for fetching sourcemap from multiple fleet servers when some fail and some succeed.

- call fetch in a goroutine
- block until all servers have received their request
- have ts0 respond with failure
- have ts1 respond with success
- block until ts2's request is cancelled

stuartnelson3:
Was taking a look at this, and I don't think there's a way to block within the test code until ts2's request has been canceled. As the fetching code is written, as soon as we get a successful response, we return to the caller. We have no way to know, from the test server's perspective, if the client we're writing to is still reading on the other side. Additionally, even though the fetching code has returned, the test server is still able to successfully write a response.

I added a test which blocks until all test servers have received their requests, and then verifies that the fetching code returns after a successful request, but before the 3rd (and final) test server handler has finished executing. I was able to verify via print statements that the 3rd request is being canceled and the `results` channel being ranged over is successfully closed.

PR #6699, opened 2021-11-23
coordinate successful and failing fetches

stuartnelson3:
## Motivation/summary

make sure that the servers all receive their
requests before confirming that a successful
response from one server cancels the shared
context in the fetch code.

## How to test these changes

non-functional change

## Related issues

closes https://github.com/elastic/apm-server/issues/5836


estolfo (approved):
looks good, thanks for doing this

sourcemap/fleet_test.go
```diff
@@ -123,9 +152,21 @@ func TestFailedAndSuccessfulFleetHostsFetch(t *testing.T) {
...
 
-	consumer, err := f.Fetch(context.Background(), name, version, path)
-	assert.NoError(t, err)
-	assert.NotNil(t, consumer)
+	go func() {
+		consumer, err := f.Fetch(context.Background(), name, version, path)
+		resc <- response{consumer, err}
+	}()
+	// Make sure every server has received a request
+	wg.Wait()
```
marclop, line 160: Can't we remove this wait group? It seems to me that the channels are already making the test wait until we've received all the signals. What is this wait group adding on top of that?

stuartnelson3: The waitgroup is making sure that each handler has received a request before we attempt to read from any of the channels. Without that, we could potentially read from `errc` and `successc` without the third handler ever being invoked.

marclop: got it, feel free to merge, thanks for the explanation

PR #6715, opened 2021-11-24
[apmpackage] use same rate_limit defaults as code

stuartnelson3:
## Motivation/summary

for `apm-server.auth.anonymous.*`, use the same
defaults as are defined in the config code in
beater/config/auth.go

## Checklist


marclop (approved):

apmpackage/apm/manifest.yml
```diff
@@ -70,10 +70,10 @@ policy_templates:
             multi: true
           - name: anonymous_rate_limit_event_limit
             type: integer
-            default: 10
+            default: 300
           - name: anonymous_rate_limit_ip_limit
             type: integer
-            default: 10000
+            default: 1000
```
simitt, line 76: I believe to remember that the idea was to change the in-code defaults to be aligned with the new defaults from the apmpackage in `8.0`. We started with, what we believe are better defaults, in the apmpackage, as it wasn't a breaking change there. For the standalone binary it would have been a breaking change. So I'd be in favor of keeping the values in the package and adapting the in-code defaults for standalone. 

@axw you introduced the new values in https://github.com/elastic/apm-server/pull/5623/files#diff-27399a724d6e4798ba048b909a8a5136448baa1c777830feafeb9a5dabafe4cdR97-R110 - is my above statement alligned with your recollection?

simitt (changes requested):

apmpackage/apm/manifest.yml
axw: > is my above statement alligned with your recollection

Generally speaking, yes, but I think I messed up with these particular config vars. I'm pretty sure I intended to align the new defaults in the integration package with what I _thought_ were the defaults in the cloud-assets templates. Looking at those again, I realise I misread them. So I think we should go with what @stuartnelson3 has here.

simitt (approved):

simitt:
Please backport this to `7.16` though.

PR #6741, opened 2021-11-29
[apmintegration] include java_attacher configuration

stuartnelson3:
## Motivation/summary

enable configuration of java-attacher via apm integration package

## Checklist

## How to test these changes

install the integration and verify that configuration in the UI is reflected in the arguments used to start the java attacher. `log.level=debug` will output the command line arguments used to invoke the attacher.


apmpackage/apm/agent/input/template.yml.hbs
```diff
@@ -27,6 +27,9 @@ apm-server:
     max_header_size: {{max_header_bytes}}
     read_timeout: {{read_timeout}}
     response_headers: {{response_headers}}
+    java_attacher:
+      enabled: {{java_attacher_enabled}}
+      discovery_rules: {{java_attacher_discovery_rules}}
```
stuartnelson3, line 32: @ogupte is this the correct way to interpolate the yml data here?

axw (approved):

marclop:
I've been able to verify that the `apm-server` tries to start the attacher, however, the attacher cannot be run when run in managed mode by the Elastic Agent. @stuartnelson3 Should we consider this as OK?

stuartnelson3:
Since you've confirmed the command is correct from the debug log output, I think the permissions issue is separate to this. I think we can consider this as ok.

Issue #6426, opened 2021-10-26
Use Elasticsearch compatibility header with go-elasticsearch


Mpdreamz:
Elasticsearch `7.10` introduces compatibility headers:

```
Accept:                 application/vnd.elasticsearch+json; compatible-with=7
Content-Type:           application/vnd.elasticsearch+json; compatible-with=7
```

Meaning Elasticsearch takes care of accepting and returning a `7.x` compatible response.

by utilizing this we can support running APM Server 7.16.x running against `Elasticsearch 8.x` with minimal accounting for breaking changes.

Compatible applies to format only, behavioral changes may occur. 

On 7.x the return type is still json:

```
content-type:       application/json; charset=UTF-8
```

On 8.x the return type matches the `Accept` header:

```
content-type:       application/vnd.elasticsearch+json;compatible-with=7
```

https://github.com/elastic/go-elasticsearch is setup to handle this but does not offer a nice configuration option yet (cc @Anaethelion)

axw:
I see the following options:
 1. Enable the compatibility header by default, breaking support for older Elasticsearch versions. IMO this is not a reasonable option -- we explicitly support all 7.x.
 2. Provide users an option to enable the compatibility header, but default to off. Not great, but least complex.
 3. Have apm-server discover the Elasticsearch version, and automatically enable the header. Better for users if it works, but more complex.

I think (2) might even be possible today, by setting `output.elasticsearch.headers`.

simitt:
I tested option (2) to let users decide to configure compatibility with ES 8.0.

With current libbeat behavior I ran into issues where the `Acceot` and `Content-Type` headers were always also set by libbeat, leading to invalid headers. A PR to change the current behavior and support the custom headers is up https://github.com/elastic/beats/pull/29176. 

With https://github.com/elastic/go-elasticsearch/pull/355 the go-elasticsearch client introduced a config setting `enable_compatibility_mode` that can be used to enable the compatibility header when working with the `v7` client. This requires some changes on the APM Server side, to ensure the config option is passed on. The go-es-client does not allow the `accept` and `content-type` headers to be explicitly configured, as it would add the headers leading to invalid headers. PR is up https://github.com/elastic/apm-server/pull/6742

axw:
@simitt can we close this now?

simitt:
I haven't merged the `7.16` backport in libbeat yet to not interfere with `7.16.0` at this late stage. After merging and syncing, this issue can be closed.

PR #6742, opened 2021-11-29
Set es-client config option enable_compatibility_mode for compatible ES responses

simitt:
## Motivation/summary

https://github.com/elastic/go-elasticsearch/pull/355 introduced support for setting
a config option for the 7.x client to request 7.x compatible responses from Elasticsearch.
Reuse the supported config options for libbeat (setting the actual headers) to set this
es-client configuration option. The http headers need to be removed from the client as
the es-client takes care of setting them.

## Checklist

## How to test these change
* Configure es output headers in apm-server.yml 
```
output.elasticsearch:
    headers:
      Content-Type: "application/vnd.elasticsearch+json;compatible-with=7"
      Accept: "application/vnd.elasticsearch+json;compatible-with=7"
```
* Run apm-server `7.16` (not from this PR) and observe ingestion errors
* Run apm-server from this PR and observe that data are properly ingested (you can also add debugging and explicitly check that the `accept` and `content-type` headers are properly set) 

## Related issues

related to #6426


elasticsearch/client.go
```diff
@@ -172,17 +173,30 @@ func newV7Client(
...
+
+	// Set the compatibility configuration if compatibility headers are configured.
+	// Delete headers, as ES client takes care of setting them appropriately.
+	var enableCompatibilityMode bool
+	acceptHeader := headers.Get("Accept")
+	contentTypeHeader := headers.Get("Content-Type")
+	if strings.Contains(acceptHeader, "compatible-with=7") || strings.Contains(contentTypeHeader, "compatible-with=7") {
+		enableCompatibilityMode = true
+		headers.Del("Accept")
+		headers.Del(("Content-Type"))
```
axw, line 185:
```suggestion
		headers.Del("Content-Type")
```

elasticsearch/config.go
```diff
@@ -62,6 +62,8 @@ type Config struct {
 	// with modelindexer; it is otherwise ignored.
 	CompressionLevel int `config:"compression_level" validate:"min=0, max=9"`
 
+	EnableCompatibilityMode bool `config:"enable_compatibility_mode"`
+
```
axw, line 66:
```suggestion
```

AFAICS this isn't being used

elasticsearch/config_test.go
```diff
@@ -137,6 +137,8 @@ func TestBeatsConfigSynced(t *testing.T) {
 
 	// "hosts" is only expected in the local struct
 	delete(localStructFields, "hosts")
+	// "enable_compatibility_mode" is specific to the go-elasticsearch client, not present in libbeat
+	delete(localStructFields, "enable_compatibility_mode")
```
axw, line 141:
```suggestion
```

elasticsearch/client_test.go
```diff
@@ -76,4 +76,26 @@ func TestClientCustomHeaders(t *testing.T) {
...
+		Hosts:   Hosts{srv.URL},
+		Headers: map[string]string{"Content-Type": "application/vnd.elasticsearch+json; compatible-with=7"},
+	}
+
+	client, err := NewClient(&cfg)
+	require.NoError(t, err)
+
+	CreateAPIKey(context.Background(), client, CreateAPIKeyRequest{})
+	// The go-elasticsearch library takes care of setting
+	assert.Equal(t, "application/vnd.elasticsearch+json;compatible-with=7", requestHeaders.Get("Accept"))
```
axw, line 100: should we check Content-Type here too?

simitt: one header check should be enough, as this is done by the client as soon as the config option is set; but I added the second check so it also includes testing the client

axw: fair point!

axw (approved):
LGTM!

marclop:
should we not back port this PR to 8.0?

PR #6767, opened 2021-12-01
datastream: Use internal metrics on empty service.name

### LINKED ISSUE #6762: `agent_config` metrics are sent to application metrics data stream
While digging around in a test cluster, I noticed an unusually named data stream `metrics-apm.app.-default`. Looking at the documents within, all documents are `agent_config` metrics. We should be sending these to the "internal" metrics data stream, as they are not application-specific.

```
      {
        "_index" : ".ds-metrics-apm.app.-default-2021.11.22-000001",
        "_id" : "hiLwV30B0vAOyzkDqflC",
        "_score" : 1.0,
        "_source" : {
          "@timestamp" : "2021-11-25T16:32:09.305Z",
          "metricset.name" : "agent_config",
          "data_stream.dataset" : "apm.app.",
          "data_stream.namespace" : "default",
          "ecs" : {
            "version" : "1.12.0"
          },
          "observer" : {
            "ephemeral_id" : "3cb0aea4-1740-43eb-9d37-8bb6d40cf99e",
            "hostname" : "apm-apm-server-6455b9c6f7-clhf7",
            "id" : "23caee25-bc54-4d97-9ba7-848e16a557f2",
            "type" : "apm-server",
            "version" : "8.1.0",
            "version_major" : 8
          },
          "processor" : {
            "name" : "metric",
            "event" : "metric"
          },
          "agent_config_applied" : 1,
          "data_stream.type" : "metrics",
          "labels" : {
            "etag" : "-"
          }
        }
      },
```

marclop:
## Motivation/summary

Checks that `Service.Name` is not empty before using the field's value
in the datastream `metrics-apm.app.<Service.Name>`.

## Checklist

## How to test these changes

1. `docker-compose up -d`
2. Install APM Integration through Kibana.
3. Create [an APM Agent Configuration](http://localhost:5601/app/apm/settings/agent-configuration/create) setting the max_transactions to `100` for all agents.
4. `make`
5. `./apm-server -E output.elasticsearch.username=admin -E output.elasticsearch.password=changeme -e`
6. Run this Go program
```go
package main

import (
	"context"
	"fmt"
	"os"
	"time"

	"go.elastic.co/apm"
)

func main() {
	version := "undefined"
	if len(os.Args) > 1 {
		version = os.Args[1]
	}
	span.Duration = time.Second
	span.End()
	tx.Duration = 2 * time.Second
	tx.End()
	apm.DefaultTracer.Flush(nil)
	<-time.After(time.Minute)
}
```
7. Verify that the [configuration has been applied by the agent](http://localhost:5601/app/apm/settings/agent-configuration)
8. Navigate to Dev tools, issue the following query and verify that there's at least 1 document:
```javascript
GET metrics-apm.internal*/_search
{
  "query": {
    "term": {
      "metricset.name": {
        "value": "agent_config"
      }
    }
  }
}
```

## Related issues

Closes #6762


systemtest/agentconfig_test.go
```diff
@@ -103,6 +105,14 @@ func TestAgentConfig(t *testing.T) {
 		assert.Equal(t, configured2, settings)
 		assert.NotEqual(t, etags[url], resp.Header.Get("Etag"))
 	}
+
+	t.Log("waiting for the agent_config metrics to be published...")
+	waitForLogMessage(t, srv, "bulk request completed: 2 indexed", 1)
```
axw, line 110: ExpectMinDocs will retry for up to 10 seconds. Is it needed/useful to wait for the metrics to be published first?

axw: Looking closer, I see that we hard code the agentcfg.Reporter to report every 30 seconds. Would it be reasonable to make that configurable (does not need to be documented or advertised), and reduce the interval for tests?

axw (commented):
Looks good overall, but I think the test is going to take at least 30 seconds to pass because of the wait for metrics. It would good if we could avoid slowing down the test

systemtest/agentconfig_test.go
marclop: I thought about making it configurable, but it seemed a little too much for a bug fix. I can make flush period configurable in this PR if you feel that's the best approach.

On the other hand, we're building the agent image to inject an APM server binary into it which will take 15-20s. The flush timer is running in the `apm-server` binary that we run in line 49 and the image building is accounted into that, but making it configurable would most likely reduce the test runtime by almost half.

axw: > I thought about making it configurable, but it seemed a little too much for a bug fix.

Probably a bit too much for backporting to 7.16.

How about we extract the systemtest changes into a followup, making the config change at the same time? That doesn't need to be backported to 7.16 -- particularly since these metrics an't used yet.

> On the other hand, we're building the agent image to inject an APM server binary into it which will take 15-20s.

That time is (should be?) paid only once for all the system tests -- not for every test.

axw (approved):
Thanks!

stuartnelson3:
I followed the steps above using https://github.com/elastic/apm-server/commit/2d1c27052b5ddce37858c85b48a84561752f791d, but did not see the document. I installed apm in managed mode and confirmed the agent received the config:

```json
{"level":"debug","time":"2021-12-22T16:13:18+01:00","message":"central config update: updated transaction_max_spans to 100"}
```
When I executed your query there were no documents, though.

marclop:
I missed to add that  to test this you need to run `make build-package` first, in order to get the latest package changes.
Since `8.0.0-dev4` doesn't contain the latest changes. I plan to publish the package (`8.0.0-dev5`) after https://github.com/elastic/apm-server/pull/6928 is merged to the 8.0 branch.

stuartnelson3:
:+1: double checked with `8.0.0-dev6` and https://github.com/elastic/apm-server/commit/2d1c27052b5ddce37858c85b48a84561752f791d

PR #6768, opened 2021-12-01
otlp: Add beta support to receive logs

### LINKED ISSUE #5491: [OpenTelemetry] add support for receiving logs
APM Server currently supports receiving OpenTelemetry traces and metrics. It would be nice to round that out with support for logs. The [OpenTelemetry Logging](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/overview.md) spec is currently in experimental status; we would provide no greater support guarantees than this, at least until the spec graduates.

marclop:
## Motivation/summary

Adds support to receive OTLP logs on the GRPC receiver. This feature is
_beta_ and should be used sparsely until the OTel spec declares it as
stable.

Currently, it's challenging to test this in the system tests since the
`opentelemetry-go` SDK doesn't expose any APIs for logs.

## Checklist

## How to test these changes

1. `docker-compose up -d`
2. `make`
3. `./apm-server -E output.elasticsearch.username=admin -E output.elasticsearch.password=changeme -e`
4. Run the following go program: (Make sure that you're using `go.opentelemetry.io/collector/model@v0.34.0`):
```go
package main

import (
	"context"
	"time"

	"go.opentelemetry.io/collector/model/otlpgrpc"
	"go.opentelemetry.io/collector/model/pdata"
	semconv "go.opentelemetry.io/collector/model/semconv/v1.5.0"
	"google.golang.org/grpc"
)

func main() {
	conn, err := grpc.Dial("localhost:8200", grpc.WithInsecure(), grpc.WithBlock())
	if err != nil {
		panic(err)
	}
	defer conn.Close()

	client := otlpgrpc.NewLogsClient(conn)
	newLog := func(body interface{}) pdata.Logs {
		logs := pdata.NewLogs()
		resourceLogs := logs.ResourceLogs().AppendEmpty()
		logs.ResourceLogs().At(0).Resource().Attributes().InsertString(
			semconv.AttributeTelemetrySDKLanguage, "go",
		)
		instrumentationLogs := resourceLogs.InstrumentationLibraryLogs().AppendEmpty()
		otelLog := instrumentationLogs.Logs().AppendEmpty()
		otelLog.SetTraceID(pdata.NewTraceID([16]byte{1}))
		otelLog.SetSpanID(pdata.NewSpanID([8]byte{2}))
		otelLog.SetName("doOperation()")
		otelLog.SetSeverityNumber(pdata.SeverityNumberINFO)
		otelLog.SetSeverityText("Info")
		otelLog.SetTimestamp(pdata.NewTimestampFromTime(time.Now()))
		otelLog.Attributes().InsertString("key", "value")
		otelLog.Attributes().InsertDouble("key", 1234)

		switch b := body.(type) {
		case string:
			otelLog.Body().SetStringVal(b)
		case int:
			otelLog.Body().SetIntVal(int64(b))
		case float64:
			otelLog.Body().SetDoubleVal(float64(b))
		case bool:
			otelLog.Body().SetBoolVal(b)
		}
		return logs
	}

	_, err = client.Export(context.Background(), newLog("my log message"))
	if err != nil {
		panic(err)
	}
	_, err = client.Export(context.Background(), newLog(12334))
	if err != nil {
		panic(err)
	}
}
```
5. Install the APM Integration
6. Verify that the two log messages show up in the `Observability > Logs` UI.

image

## Related issues

Closes #5491


processor/otel/logs.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+	if traceID := record.TraceID(); !traceID.IsEmpty() {
+		event.Trace.ID = traceID.HexString()
+	}
+	if spanID := record.SpanID(); !spanID.IsEmpty() {
+		if event.Span == nil {
+			event.Span = &model.Span{}
+		}
+		event.Span.ID = spanID.HexString()
+	}
```
simitt, line 124: We should also set the `ParentID`.

processor/otel/logs.go
```diff
@@ -0,0 +1,133 @@
...
+		}
+	}
+	if traceID := record.TraceID(); !traceID.IsEmpty() {
+		event.Trace.ID = traceID.HexString()
+	}
+	if spanID := record.SpanID(); !spanID.IsEmpty() {
+		if event.Span == nil {
+			event.Span = &model.Span{}
+		}
+		event.Span.ID = spanID.HexString()
```
simitt, line 123: Couldn't this also refer to the `transaction.ID`. Is there any information available that would allow a similar check as https://github.com/elastic/apm-server/blob/master/processor/otel/traces.go#L211-L226?

processor/otel/logs.go
axw: I don't think we need to set `parent.id` for logs do we? If a log is associated with a span, it has a `span.id`; if associated with a transaction, it has `transaction.id`.

processor/otel/logs_test.go
```diff
@@ -0,0 +1,150 @@
...
+		"numeric_key": pdata.NewAttributeValueDouble(1234),
+	})
+
+	switch b := body.(type) {
+	case string:
+		otelLog.Body().SetStringVal(b)
+	case int:
+		otelLog.Body().SetIntVal(int64(b))
+	case float64:
+		otelLog.Body().SetDoubleVal(float64(b))
```
axw, line 141:
```suggestion
		otelLog.Body().SetDoubleVal(b)
```

it's already float64

processor/otel/logs.go
```diff
@@ -0,0 +1,133 @@
...
+) model.APMEvent {
+	event := baseEvent
+	event.Timestamp = record.Timestamp().AsTime().Add(timeDelta)
+	event.Event.Severity = int64(record.SeverityNumber())
+	event.Event.Action = record.Name()
+	event.Log.Level = record.SeverityText()
+	if body := record.Body(); body.Type() != pdata.AttributeValueTypeNull {
+		event.Message = body.AsString()
+		if body.Type() == pdata.AttributeValueTypeMap {
+			body.MapVal().Range(func(k string, v pdata.AttributeValue) bool {
```
axw, line 110: Should we be calling `initEventLabels(&event)` before this, like we do below for attrs? Probably should just be calling it once.

processor/otel/logs.go
axw: I don't think we can know that here, but happy to be told otherwise. We're going to need to do something like https://github.com/elastic/apm/issues/413, renaming `transaction.id` to `span.id`.

I think it might be best to set `span.id` even though it might not be correct, i.e. because it's a transaction. At least then users can search for a transaction with the `span.id` value, and we'll be aligned for a future where `transaction.id` doesn't exist.

@felixbarny WDYT?

felixbarny: SGTM. We could also add a `span.id` to transactions and look for ways how to phase out `transaction.id` going forward.

processor/otel/logs.go
simitt: ok, I wasn't sure if it could be useful in the UI.

processor/otel/logs.go
simitt: sounds good, thanks

processor/otel/logs.go
marclop: that's true, thanks for catching this

model/log.go
```diff
@@ -25,3 +27,17 @@ var (
 	// LogProcessor is the Processor value that should be assigned to log events.
 	LogProcessor = Processor{Name: "log", Event: "log"}
 )
+
+// Log holds information about a log, as defined by ECS.
+//
+// https://www.elastic.co/guide/en/ecs/current/ecs-log.html
+type Log struct {
+	// Level holds the log level of the log event.
+	Level string
```
axw, line 36: should we add this to the new app_logs data stream's ecs.yml?

model/event.go
```diff
@@ -36,10 +36,20 @@ type Event struct {
 
 	// Outcome holds the event outcome: "success", "failure", or "unknown".
 	Outcome string
+
+	// Severity holds the numeric severity of the event for log events.
+	Severity int64
+
+	// Severity holds the action captured by the event for log events.
+	Action string
```
axw, line 44: should we add these to the new app_logs data stream's ecs.yml?

axw (approved):
Looks great, thank you :tada: 

Can you please add the fields to app_logs, unless there's a reason not to?

cristianmad:
This is awesome! Is there a beta build we could try?

Thank you!

stuartnelson3:
confirmed with https://github.com/elastic/apm-server/commit/2d1c27052b5ddce37858c85b48a84561752f791d

PR #6776, opened 2021-12-02
[java attacher] set each space-separated arg individually in cmd

stuartnelson3:
## Motivation/summary

the *Command needs each argument to be space
separated in order to work. The previous code was
grouping flags+arguments as logical units, which
does not get correctly executed by the binary.

the config was also defined in an inconsistent way. Originally it was:

```yml
apm-server:
  java_attacher:
    enabled: true
    download-agent-version: 1.27.0
    discovery-rules:
    - include-main: opbeans
    - exclude-main: ".*"
    config:
      server_url: "http://localhost:8200"

```
It made sense, and simplified the code, to have `config` values be treated the same as `discover-rules`:
```yml
apm-server:
  java_attacher:
    enabled: true
    download-agent-version: 1.27.0
    discovery-rules:
    - include-main: opbeans
    - exclude-main: ".*"
    config:
    - server_url: "http://localhost:8200"
```
(config was a map, now it's a slice of maps)

## Checklist

## How to test these changes

0. Download the java-attacher jar if this isn't being tested with a release that already has it pre-packaged:
  ```
make build/apm-agent-attach-cli-1.27.0-slim.jar
cp build/java-attacher.jar .
```
1. Start apm-server with the above config, and `seccomp.enabled=false` and `log.level=debug`
2. Verify that there's output from the java-attacher

### CODE REVIEW

File: beater/java_attacher/java_attacher.go [hunk 1]
```diff
@@ -55,6 +54,9 @@ func New(cfg config.JavaAttacherConfig) (JavaAttacher, error) {
 		cfg.JavaBin = filepath.FromSlash(cfg.JavaBin)
 	}
 	logger := logp.NewLogger("java-attacher")
+	if _, err := os.Stat(javaAttacher); err != nil {
```
stuartnelson3, line 57: this was an initial error I encountered, the error output from when the attacher was missing was super-unhelpful. The stat error, `failed to start java attacher: stat ./java-attacher.jar: no such file or directory`, is much more helpful.

stuartnelson3:
@eyalkoren discovered this bug initially

@ogupte i've only added discovery-rules to the apm integration package, but now config has the same format. I don't think the config options are order dependent (I'm not seeing it mentioned in [the docs](https://www.elastic.co/guide/en/apm/agent/java/current/setup-attach-cli.html#setup-attach-cli-usage)), but maybe @felixbarny can clarify.

felixbarny:
>  I don't think the config options are order dependent (I'm not seeing it mentioned in the docs), but maybe @felixbarny can clarify.

That's correct. Also, config options can't be set in the Fleet UI.
But it might make more sense to leave config as-is. That's because the same config can't and shouldn't be provided multiple times, such as `--config service_name=foo --config service_name=bar`. By making the Config a map, that gets automatically handled.

But it's not a big deal either way if you prefer `DiscoveryRules` and `Config` to be more consistent.

axw:
> But it might make more sense to leave config as-is. That's because the same config can't and shouldn't be provided multiple times, such as --config service_name=foo --config service_name=bar. By making the Config a map, that gets automatically handled.
>
> But it's not a big deal either way if you prefer DiscoveryRules and Config to be more consistent.

If it's not sensible to pass `--config` multiple times, then I don't think we should structure it that way in the code. Multiple DiscoveryRules can be specified, hence the difference. @stuartnelson3 what's the downside to reverting that bit?

felixbarny:
> If it's not sensible to pass --config multiple times

Sorry, that's not what I was trying to say. `--config` can be specified multiple times, but each config key can only be specified once.

Example:
Illegal: `--config service_name=foo --config service_name=bar`
Legal: `--config service_name=foo --config transaction_sample_rate=0.1`

axw:
@felixbarny thanks for the clarification. Even so, I still think it makes sense to have a single map to make it impossible to render the invalid case.

stuartnelson3:
The downside was verifying rendering multiple config values in the command line arguments correctly. I was sorting the config values to get a consistent result in the args passed to the go `Command`, but maybe we can just say it's good enough as-is and not worry about checking that in a unit test.

felixbarny:
The order in which the `--config` arguments are provided doesn't really matter.

beater/java_attacher/java_attacher_test.go
```diff
@@ -79,7 +85,7 @@ func TestBuild(t *testing.T) {
 	want := filepath.FromSlash("/usr/bin/java -jar ./java-attacher.jar") +
 		" --continuous --log-level debug --download-agent-version 1.25.0 --exclude-user root --include-main MyApplication " +
 		"--include-main my-application.jar --include-vmarg elastic.apm.agent.attach=true " +
-		"--config server_url=http://localhost:8200 --config service_name=my-cool-service"
+		"--config server_url=http://localhost:8200"
 
 	cmdArgs := strings.Join(cmd.Args, " ")
 	assert.Equal(t, want, cmdArgs)
```
felixbarny, line 91: Just an idea that may simplify testing and ordering issues: instead of asserting the whole expected string, you may also assert that certain sub-strings are contained. Such as

```go
assert.Contains(t, "--config server_url=http://localhost:8200", cmdArgs)
assert.Contains(t, "--config service_name=my-cool-service", cmdArgs)
```

felixbarny (approved):

eyalkoren:
With the latest changes, this works:
```yml
apm-server:
  java_attacher:
    enabled: true
    download-agent-version: 1.27.0
    discovery-rules:
    - include-main: opbeans
    - exclude-main: ".*"
    config:
      server_url: "http://localhost:8200"

```
And this doesn't work:
```yml
apm-server:
  java_attacher:
    enabled: true
    download-agent-version: 1.27.0
    discovery-rules:
    - include-main: opbeans
    - exclude-main: ".*"
    config:
    - server_url: "http://localhost:8200"
```

axw:
> The downside was verifying rendering multiple config values in the command line arguments correctly. I was sorting the config values to get a consistent result in the args passed to the go Command, but maybe we can just say it's good enough as-is and not worry about checking that in a unit test.

I see. To get a consistent ordering, I'd usually extract the keys and sort them, and then range over those. Something like:

```go
keys := make([]string, 0, len(config))
for key := range config {
    keys = append(keys, key)
}
sort.Strings(keys)
for _, key := range keys {
    value := config[key]
    ...
}
```

A bit more code, but that way the exported API is minimally complete and doesn't permit invalid configurations.

marclop:
Verified with https://snapshots.elastic.co/8.0.0-b0d1c470/summary-8.0.0-SNAPSHOT.html#apm-server:


```json
{"log.level":"debug","@timestamp":"2022-01-04T15:54:51.383+0800","log.logger":"java-attacher","log.origin":{"file.name":"java_attacher/java_attacher.go","file.line":71},"message":"starting java attacher with command: /opt/homebrew/opt/openjdk@11/bin/java -jar ./java-attacher.jar --continuous --log-level debug --download-agent-version 1.27.1 --include-vmargs elastic.apm.attach=true --exclude-main .*","service.name":"apm-server","ecs.version":"1.6.0"}
"log.level":"debug","@timestamp":"2022-01-04T15:54:56.812+0800","log.logger":"java-attacher","log.origin":{"file.name":"java_attacher/java_attacher.go","file.line":109},"message":"Traversing [/bc-lib] for PGP signature verifier implementation and related dependencies","service.name":"apm-server","ecs.version":"1.6.0"}
{"log.level":"debug","@timestamp":"2022-01-04T15:54:56.819+0800","log.logger":"java-attacher","log.origin":{"file.name":"java_attacher/java_attacher.go","file.line":109},"message":"/bc-lib directory contents: [file:/Users/marclop/repos/elastic/apm-server/agent/1_27_1/lib/bc-fips.jar, file:/Users/marclop/repos/elastic/apm-server/agent/1_27_1/lib/apm-agent-attach-cli-1.27.0-bc-verifier.jar, file:/Users/marclop/repos/elastic/apm-server/agent/1_27_1/lib/bcpg-fips.jar]","service.name":"apm-server","ecs.version":"1.6.0"}
{"log.level":"debug","@timestamp":"2022-01-04T15:54:56.825+0800","log.logger":"java-attacher","log.origin":{"file.name":"java_attacher/java_attacher.go","file.line":109},"message":"co.elastic.apm.attach.bouncycastle.BouncyCastleVerifier class loaded by co.elastic.apm.attach.PgpSignatureVerifierLoader@38f116f6","service.name":"apm-server","ecs.version":"1.6.0"}
{"log.level":"debug","@timestamp":"2022-01-04T15:54:56.914+0800","log.logger":"java-attacher","log.origin":{"file.name":"java_attacher/java_attacher.go","file.line":109},"message":"Requested to download Elastic APM Java agent version 1.27.1","service.name":"apm-server","ecs.version":"1.6.0"}
{"log.level":"debug","@timestamp":"2022-01-04T15:55:00.666+0800","log.logger":"java-attacher","log.origin":{"file.name":"java_attacher/java_attacher.go","file.line":109},"message":"elastic-apm-agent-1.27.1.jar already exists at /Users/marclop/repos/elastic/apm-server/agent/1_27_1","service.name":"apm-server","ecs.version":"1.6.0"}
{"log.level":"debug","@timestamp":"2022-01-04T15:55:00.667+0800","log.logger":"java-attacher","log.origin":{"file.name":"java_attacher/java_attacher.go","file.line":109},"message":"attacher arguments : Arguments{rules=DiscoveryRules{discoveryRules=[include(vmArgs(elastic.apm.attach=true)), exclude(main(.*))]}, config={}, argsProvider='null', help=false, list=false, continuous=true, logLevel=DEBUG, logFile='null', listVmArgs=false, agentJar=/Users/marclop/repos/elastic/apm-server/agent/1_27_1/elastic-apm-agent-1.27.1.jar}","service.name":"apm-server","ecs.version":"1.6.0"}
{"log.level":"error","@timestamp":"2022-01-04T15:55:01.146+0800","log.logger":"java-attacher","log.origin":{"file.name":"java_attacher/java_attacher.go","file.line":98},"message":"error unmarshaling attacher logs: invalid character 'o' looking for beginning of value","service.name":"apm-server","ecs.version":"1.6.0"}
{"log.level":"error","@timestamp":"2022-01-04T15:55:01.166+0800","log.logger":"java-attacher","log.origin":{"file.name":"java_attacher/java_attacher.go","file.line":98},"message":"error unmarshaling attacher logs: invalid character 'o' looking for beginning of value","service.name":"apm-server","ecs.version":"1.6.0"}
{"log.level":"error","@timestamp":"2022-01-04T15:55:01.187+0800","log.logger":"java-attacher","log.origin":{"file.name":"java_attacher/java_attacher.go","file.line":98},"message":"error unmarshaling attacher logs: invalid character 'o' looking for beginning of value","service.name":"apm-server","ecs.version":"1.6.0"}
{"log.level":"debug","@timestamp":"2022-01-04T15:55:02.072+0800","log.logger":"java-attacher","log.origin":{"file.name":"java_attacher/java_attacher.go","file.line":109},"message":"Looking in the following folders for hsperfdata_<user>/<pid> files: [/var/folders/3
```

PR #6779, opened 2021-12-03
systemtest: Remove setup and output config

marclop:
## Motivation/summary

Removes `setup` and `queue` settings from the `systemtest` config since
they've been removed and have no effect on the APM Server.

## Checklist

## How to test these changes

N/A

## Related issues

None

### DIFF
--- a/systemtest/estest/search.go
+++ b/systemtest/estest/search.go
@@ -55,6 +55,16 @@ func (es *Client) ExpectMinDocs(t testing.TB, min int, index string, query inter
 		result.Hits.MinHitsCondition(min),
 		result.Hits.TotalHitsCondition(req),
 	)))
+
+	// Refresh the indices before issuing the search request.
+	refreshReq := esapi.IndicesRefreshRequest{
+		Index:           strings.Split(",", index),
+		ExpandWildcards: "all",
+	}
+	if _, err := refreshReq.Do(context.Background(), es.Transport); err != nil {
+		t.Fatalf("faled refreshing indices: %s: %s", index, err.Error())
+	}
+
 	if _, err := req.Do(context.Background(), &result, opts...); err != nil {
 		t.Fatal(err)
 	}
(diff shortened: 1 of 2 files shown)


marclop:
Noticed while working on some system tests that the `apm-server` the following log message:


```json
{"log.level":"warn","@timestamp":"2021-11-24T15:15:56.358+0800","log.logger":"index-management","log.origin":{"file.name":"idxmgmt/supporter_factory.go","file.line":69},"message":"`setup.template` specified, but was removed in 8.0 and will be ignored: index templates are managed by Fleet","service.name":"apm-server","ecs.version":"1.6.0"}
```

axw (approved):
Thanks, good catch.

We should probably also hard-code `output.elasticsearch.flush_interval=1ms` or something, so the tests aren't all slowed down waiting for the standard 1s flush.

Similarly, we should either refresh indices manually in code (maybe in estest.ExpectDocs/ExpectMinDocs?), or customise the data stream component templates.

systemtest/estest/search.go
```diff
@@ -55,6 +55,16 @@ func (es *Client) ExpectMinDocs(t testing.TB, min int, index string, query inter
...
 		result.Hits.TotalHitsCondition(req),
 	)))
+
+	// Refresh the indices before issuing the search request.
+	refreshReq := esapi.IndicesRefreshRequest{
+		Index:           strings.Split(",", index),
+		ExpandWildcards: "all",
+	}
+	if _, err := refreshReq.Do(context.Background(), es.Transport); err != nil {
+		t.Fatalf("faled refreshing indices: %s: %s", index, err.Error())
```
marclop, line 65: This could just be a warning, or be ignored completely.

axw (approved):

systemtest/estest/search.go
axw:
```suggestion
		t.Fatalf("failed refreshing indices: %s: %s", index, err.Error())
```

axw: I think failing is appropriate. We wouldn't call this function unless we expect some docs.

PR #6786, opened 2021-12-03
Update RUM source map absolute path logic

ysamaali:
Our APM Engineering explained the logic of source map abs_path matching: 
abs_path will match based on the following: 
1. bundle_filepath="http://localhost/static/js/bundle.js" // having full URL matching, if it's not the case, then move to step 2) 
2. bundle_filepath="/static/js/bundle.js" // only the path without URL part, if it's not the case then source map will not be applied correctly

## Checklist


simitt (approved):

simitt:
@ysamaali would you mind targeting the PR against the `7.16` branch and adding a label `backport-7.15` ?

PR #6787, opened 2021-12-03
[DOCS] Procedure to restore a write alias for APM

lucabelluccini:
The procedure is wrong as it is missing the part to stop APM Server. If APM Server is not stopped, after the `DELETE` a new index could be recreated.

# Reviewers

- [ ] Ensure the procedure is valid
- [ ] Review the wording

### DIFF
--- a/docs/legacy/common-problems.asciidoc
+++ b/docs/legacy/common-problems.asciidoc
@@ -318,15 +318,43 @@ GET _cat/recovery/apm*transaction*?s=index&v=true&h=index,stage
 +
 When `stage: done`, you're ready to move on.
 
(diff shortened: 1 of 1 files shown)


bmorelli25:
Thanks as always, @lucabelluccini! Words look good to me. @elastic/apm-server, does this update look correct to you? Preview is [here](https://apm-server_6787.docs-preview.app.elstc.co/guide/en/apm/guide/master/common-problems.html#server-resource-exists-not-alias).

simitt (approved):
Didn't try it out, but that seems reasonable. Thanks for the fix.

bmorelli25 (approved):

Issue #6425, opened 2021-10-26
Introduce version check on installed package


Mpdreamz:
We currently validate that the APM integration is installed while running standalone and assume it is when running managed. 

We don't validate the installed package actually satisfies our `Stack Version >= APM Version` constraint. 

A drift in the installed package could mean that APM Server will introduce new fields with wrong mappings. This can lead to a broken UI experience which can be hard to diagnose. 

Opening this up to discuss if we want to:

* Do nothing. 
* Check and warn
* Check and go into wait loop.
* Check and version our write logic.
   * E.g when introducing new fields only write these if we know the integration supports it.

The first 2 options feel trappy to me, the user won't know they are potentially indexing data the wrong way. This could potentially stay un-diagnosed for a long time.

relates: https://github.com/elastic/beats/issues/28643

axw:
I'm going to go ahead with what I described above -- please let me know if you have any concerns. When #28643 is done we might want to update the "wait for integration" code in apm-server standalone to check the installed version.

PR #6791, opened 2021-12-06
Check APM Server version in ingest pipeline

axw:
## Motivation/summary

We now check `observer.version` represents an APM Server version
no greater than the installed integration package, disregarding
the patch and tag components. This is done in the ingest pipeline,
which is tied to the integration package version.
`observer.version_major` is now set in the ingest pipeline, rather
than in apm-server code.

We introduce an APM-specific application logs data stream. We had
been relying on the base "logs-*" data stream installed by
Elasticsearch. By introducing an APM-specific data stream, we can
define an ILM policy and ingest pipeline which performs the version
check.

## Checklist

## How to test these changes

1. Install integration 8.0.0 or 8.1.0
2. Run a custom apm-server with the version bumped to 9.0.0
3. Send some events to apm-server
4. Check the events are never indexed, no data streams are created
5. Check logs for errors suggesting that the integration should be upgraded

## Related issues

Closes https://github.com/elastic/apm-server/issues/6425


simitt (approved):
LGTM
but one thing I missed in the previous PR:
in https://github.com/elastic/apm-server/blob/master/apmpackage/cmd/genpackage/pipelines.go#L51 you check
```
ctx.observer.version_major > %d || ctx.observer.version_minor > %d
```

what if the APM Integration is `8.0.0` but the APM Server version is `7.16.0` - wouldn't that also fail then? (but shouldn't)

axw:
> what if the APM Integration is 8.0.0 but the APM Server version is 7.16.0 - wouldn't that also fail then? (but shouldn't)

:facepalm: rookie mistake
thanks, will fix

systemtest/main_test.go
```diff
@@ -28,6 +28,13 @@ func TestMain(m *testing.M) {
 	if err := StartStackContainers(); err != nil {
 		log.Fatal(err)
 	}
+	log.Println("INFO: cleaning up Elasticsearch...")
+	if err := cleanupElasticsearch(); err != nil {
+		log.Fatal(err)
+	}
+	if err := InitFleet(); err != nil {
```
marclop, line 35: How come we initialize fleet both here and in line 38?

marclop (approved):

systemtest/main_test.go
axw: because I have not had enough sleep
thanks for catching that, fixing

marclop:
Verified with the `8.1.0` integration and the latest commit in master with the version bumped to `9.1.0`:

1. `make`
2. `./apm-server -E output.elasticsearch.username=admin -E output.elasticsearch.password=changeme -E apm-server.rum.enabled=true -d '*' -e`
```json
{"log.level":"info","@timestamp":"2021-12-21T15:56:25.493+0800","log.logger":"beater","log.origin":{"file.name":"beater/waitready.go","file.line":40},"message":"blocking ingestion until all preconditions are satisfied","service.name":"apm-server","ecs.version":"1.6.0"}
```

3. Ingest some events: `curl -H "Content-type: application/x-ndjson" --data-binary @testdata/intake-v2/transactions_spans_rum.ndjson http://localhost:8200/intake/v2/rum/events`.
4. Check the logs for ingest failures:
```json
{"log.level":"info","@timestamp":"2021-12-21T15:59:38.717+0800","log.logger":"request","log.origin":{"file.name":"middleware/log_middleware.go","file.line":61},"message":"request accepted","service.name":"apm-server","url.original":"/intake/v2/rum/events","http.request.method":"POST","user_agent.original":"curl/7.64.1","source.address":"127.0.0.1","http.request.body.bytes":1155,"http.request.id":"d52c3a00-8eee-42fc-91b1-a5b328d2ad21","event.duration":9869583,"http.response.status_code":202,"ecs.version":"1.6.0"}
{"log.level":"error","@timestamp":"2021-12-21T15:59:39.814+0800","log.logger":"modelindexer","log.origin":{"file.name":"modelindexer/indexer.go","file.line":382},"message":"failed to index event (fail_processor_exception): Document produced by APM Server v9.1.0, which is newer than the installed APM integration (v8.1.0). The APM integration must be upgraded.","service.name":"apm-server","ecs.version":"1.6.0"}
{"log.level":"debug","@timestamp":"2021-12-21T15:59:39.815+0800","log.logger":"modelindexer","log.origin":{"file.name":"modelindexer/indexer.go","file.line":400},"message":"bulk request completed: 0 indexed, 2 failed (0 exceeded capacity)","service.name":"apm-server","ecs.version":"1.6.0"}
```

PR #6802, opened 2021-12-06
Reintroduce integration package generator

axw:
## Motivation/summary

Introduce a new integration package generator/transformer, which takes a partial integration package in apmpackage/apm and performs APM-specific transformations and validations:

 - Inject package version and Kibana version constraint
 - Inject common ingest pipelines
 - Check ilm_policy is present and correct (and fix the
   value in the sampled_traces)
 - Copy fields from traces to rum_traces

We no longer explicitly set `event.ingested` in our ingest pipelines; this is done by the common Fleet-installed pipeline, which is set as the final pipeline in all Fleet-install templates.

## Checklist

## How to test these changes

1. Install the package
2. Check all index templates' ILM policy references are valid

## Related issues

Prereq for https://github.com/elastic/apm-server/issues/6425


apmpackage/cmd/genpackage/main.go
```diff
@@ -0,0 +1,110 @@
...
+var (
+	outputDir = flag.String("o", "", "directory into which the package will be rendered (required)")
+	version   = flag.String("version", "", "integration package version (required)")
+)
+
+func generatePackage(pkgfs fs.FS, version *common.Version) error {
+	// Walk files, performing some APM-specific validations and transformations as we go.
+	//
+	// We assume the target destination does not yet exist.
+	if err := fs.WalkDir(pkgfs, ".", func(path string, d fs.DirEntry, err error) error {
```
marclop, line 42: We can return here directly, no need to handle the error, right?

simitt (approved):
Briefly looked at the generated apmpackage - LGTM

marclop (approved):
LGTM, also nicer to have the checks in the generator itself than parsing things in the Makefile.

apmpackage/cmd/genpackage/main.go
axw: Indeed. I'll do that in my followup PR, in the interest of time.

marclop:
Verified with 9bb5e4cdd and apmpackage `8.0.0-dev4`

PR #6805, opened 2021-12-06
Java attacher fix allowlist

stuartnelson3:
## Motivation/summary

- Fix the allowlist (the docs list `include/exclude-vmarg`, but it's actually `vmargs`)
- Execute java attacher config validation (this was never connected)
- Update error output for invalid commandline flags to include a link to the online documentation

Question: Now that we're validating the java attacher config, a bad config will prevent the server from starting. Is that what we want?

## Checklist

## How to test these changes

1. Start the `apm-server` with an unsupported java attacher flag (such as `include-vmarg`)
2. Verify the bad flag is printed out


simitt:
> Question: Now that we're validating the java attacher config, a bad config will prevent the server from starting. Is that what we want?

simitt (approved):

stuartnelson3:
Created https://github.com/elastic/apm-server/issues/6814

PR #6825, opened 2021-12-07
docs: fix release notes for 7.16.0

bmorelli25:
Moves two changes back to `head.asciidoc` — they'll make it into 7.16.1 instead:

```
- Agent configuration metrics are now sent to the metrics-apm.internal* data stream {pull}6767[6767]
- Improve logged message when APM Server blocks ingestion to wait for the integration to be installed {pull}6796[6796]
```


marclop (approved):
Thanks for the speedy fix!

Issue #6811, opened 2021-12-06
[CI] Failure detected in v8.0.0-beta1 tag


cachedout:
On Nov 9, 2021 3:38 PM CEST a test run was initiated by the tagging of v8.0.0-beta1. However, that run produced a failure:

https://apm-ci.elastic.co/job/apm-server/job/apm-server-mbp/view/tags/job/v8.0.0-beta1/1/console

### Build and Test / System and Environment Tests / github.com/elastic/apm-server/systemtest.TestAPMServerInstrumentation (from github.com_elastic_apm-server_systemtest) 

```
=== RUN   TestAPMServerInstrumentation
SIGQUIT: quit
PC=0x46c341 m=0 sigcode=0

goroutine 0 [idle]:
runtime.futex()
	/var/lib/jenkins/workspace/rver_apm-server-mbp_v8.0.0-beta1/.gvm/versions/go1.17.1.linux.amd64/src/runtime/sys_linux_amd64.s:519 +0x21
runtime.futexsleep(0xc000055000, 0x3a90af3c, 0x0)
	/var/lib/jenkins/workspace/rver_apm-server-mbp_v8.0.0-beta1/.gvm/versions/go1.17.1.linux.amd64/src/runtime/os_linux.go:44 +0x36
runtime.notesleep(0x1eaf630)
	/var/lib/jenkins/workspace/rver_apm-server-mbp_v8.0.0-beta1/.gvm/versions/go1.17.1.linux.amd64/src/runtime/lock_futex.go:160 +0x87
runtime.mPark()
	/var/lib/jenkins/workspace/rver_apm-server-mbp_v8.0.0-beta1/.gvm/versions/go1.17.1.linux.amd64/src/runtime/proc.go:1441 +0x2a
runtime.stopm()
	/var/lib/jenkins/workspace/rver_apm-server-mbp_v8.0.0-beta1/.gvm/versions/go1.17.1.linux.amd64/src/runtime/proc.go:2408 +0x78
runtime.findrunnable()
	/var/lib/jenkins/workspace/rver_apm-server-mbp_v8.0.0-beta1/.gvm/versions/go1.17.1.linux.amd64/src/runtime/proc.go:2984 +0x865
runtime.schedule()
	/var/lib/jenkins/workspace/rver_apm-server-mbp_v8.0.0-beta1/.gvm/versions/go1.17.1.linux.amd64/src/runtime/proc.go:3367 +0x239
runtime.goschedImpl(0xc000001040)
[...]

PR #6826, opened 2021-12-07
Increase system tests timeout

axw:
- Increase the timeout for the system tests to 20 minutes (hardcoded)
- Introduce a `make system-test` target which runs the system tests
- Add an overrideable GOTESTFLAGS var to Makefile, used by `make test` and `make system-test`
- Update CI script to use the new target/var

Closes https://github.com/elastic/apm-server/issues/6811

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,13 +2,14 @@
 # Variables used for various build targets.
 ##############################################################################
 
-# Enforce use of modules.
-export GO111MODULE=on
-
 # Ensure the Go version in .go_version is installed and used.
 GOROOT?=$(shell ./script/run_with_go_ver go env GOROOT)
 GO:=$(GOROOT)/bin/go
 export PATH:=$(GOROOT)/bin:$(PATH)
+
+# By default we run tests with verbose output. This may be overridden, e.g.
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -2,14 +2,15 @@
 # Variables used for various build targets.
 ##############################################################################
 
-# Enforce use of modules.
-export GO111MODULE=on
```
axw, line 6: this has been the default for a while

Makefile
```diff
@@ -45,17 +46,13 @@ apm-server:
 apm-server-oss:
 	@$(GO) build -o $@
 
-.PHONY: apm-server.test
-apm-server.test:
-	$(GO) test -c -coverpkg=github.com/elastic/apm-server/... ./x-pack/apm-server
-
-.PHONY: apm-server-oss.test
-apm-server-oss.test:
-	$(GO) test -c -coverpkg=github.com/elastic/apm-server/...
```
axw, line 54: no longer required, since the removal of python system tests

marclop (approved):
Thanks!

simitt (approved):

PR #6857, opened 2021-12-08
Fix various lint warnings

axw:
## Motivation/summary

Fix various warnings produced by [golangci-lint](https://golangci-lint.run/):

- Remove dead code (e.g. `type beater/config.urls`)
- Check errors. There are more such warnings, but they're e.g. related to `bytes.Buffer.Write` calls, which won't return errors except in pathological cases.

Some code has been simplified at the same time as addressing the warnings. e.g. Publisher is now created only as needed (typically it is not).

## Checklist

## How to test these changes

N/A

## Related issues

None


marclop (approved):
Change looks good, are you planning to add `golangci-lint` in a follow up PR?

axw:
> Change looks good, are you planning to add golangci-lint in a follow up PR?

Maybe. I'll open an issue so we don't forget.

stuartnelson3:
changes to the code look great, thanks!

PR #6903, opened 2021-12-20
[7.17] fix linter deprecations

marclop:
## Motivation/summary

Fixes the failing CI build.


marclop:
Ideally we should exclude the `*_test.go` files from the linting, I don't think we should be enforcing deprecation rules there.

PR #6923, opened 2021-12-22
modelindexer: Fix TooManyRequests stat reporting

marclop:
## Motivation/summary

Fixes a bug where the TooManyRequests (429) response from Elasticsearch
is not correctly reported, since it's returned as an error.

## Checklist

## How to test these changes

* modify `docker-compose.yml` so that the Elasticsearch container has 4Gb of memory assigned to it.
* apply this patch:
```patch
diff --git a/model/modelindexer/indexer.go b/model/modelindexer/indexer.go
index f1ba63815..f12100f0c 100644
--- a/model/modelindexer/indexer.go
+++ b/model/modelindexer/indexer.go
@@ -129,7 +129,7 @@ func New(client elasticsearch.Client, cfg Config) (*Indexer, error) {
 		)
 	}
 	if cfg.MaxRequests <= 0 {
-		cfg.MaxRequests = 10
+		cfg.MaxRequests = 15
 	}
 	if cfg.FlushBytes <= 0 {
 		cfg.FlushBytes = 5 * 1024 * 1024
```
* `./apm-server -E output.elasticsearch.username=admin -E output.elasticsearch.password=changeme -E apm-server.expvar.enabled=true`
* Run this load generator: https://github.com/elastic/apm-server/blob/931cdc43f6a1766fbfed7cf07c9c66cc97dfe1bf/testing/load_generator/main.go. You may need to increase the concurrency numbers to see the stats being reported.
* Navigate to http://localhost:8200/debug/vars and check `libbeat.output.events.toomany` is eventually > 0


model/modelindexer/indexer.go
```diff
@@ -362,12 +362,19 @@ func (i *Indexer) flush(ctx context.Context, bulkIndexer *bulkIndexer) error {
...
 			tx.Outcome = "failure"
 			apm.CaptureError(ctx, err).Send()
 		}
+
+		var errTooMany *errorTooManyRequests
+		// 429 may be returned as errors from the bulk indexer.
+		if errors.As(err, &errTooMany) {
+			atomic.AddInt64(&i.tooManyRequests, int64(n))
+		} else {
+			atomic.AddInt64(&i.eventsFailed, int64(n))
```
simitt, line 376: I interpreted the `eventsFailed` counter to contain all failures. ES sending a `429` is a failure as the events are not indexed. I think it adds value to record these failures in more detail, but I'd see the `tooManyRequests` as subcategory of `eventsFailed` and therefore expected to increase the counter for both.

model/modelindexer/indexer.go
```diff
@@ -368,6 +368,12 @@ func (i *Indexer) flush(ctx context.Context, bulkIndexer *bulkIndexer) error {
 			tx.Outcome = "failure"
 			apm.CaptureError(ctx, err).Send()
 		}
+
+		var errTooMany *errorTooManyRequests
+		// 429 may be returned as errors from the bulk indexer.
+		if errors.As(err, &errTooMany) {
+			atomic.AddInt64(&i.tooManyRequests, int64(n))
+		}
```
simitt, line 376: nit: this whole block could be moved inside the `if err != nil` check above.

simitt (approved):

model/modelindexer/indexer.go
marclop: It already is

PR #6959, opened 2021-12-30
compose: Use service tokens for `fleet-server`

marclop:
## Motivation/summary

This patch updates the current `fleet-server` container to use a custom
entrypoint which creates a service token and uses it. This is necessary
since support to authenticate against Elasticsearch using username /
password authentication has been removed from `8.0` onwards.

Additionally, the docker-compose file has been updated to wait for the
`fleet-server` container to be healthy before exiting.

## Checklist

N/A


simitt (approved):

marclop:
The TBS system test is failing due to https://github.com/elastic/fleet-server/issues/1048

marclop:
There's a pending change on the Elastic Agent that would allow us to continue using user/password in the container, the Elastic Agent itself would generate the service token and pass it down: https://github.com/elastic/beats/pull/29651

simitt:
@marclop the linked PR in beats has been merged. My understanding is that the container setup still needs to use username+password for the setup to be able to receive a service token with correct privileges. Is this change here needed then?

marclop:
@simitt We are good to close, thanks for the ping.

PR #6965, opened 2021-12-30
update docker-compose.yml image version

stuartnelson3:
## Motivation/summary

Update docker images


kuisathaverat (changes requested):
this file is updated automatically with the latest SNAPSHOT

see https://github.com/elastic/apm-server/blob/master/.ci/bump-stack-version.sh#L28
there is a configuration file in the observability-.dev repo .ci/.bump-stack-version.yml#L18-L23

Issue #6814, opened 2021-12-06
[java attacher] bad config should not prevent apm-server startup


stuartnelson3:
> Question: Now that we're validating the java attacher config, a bad config will prevent the server from starting. Is that what we want?

That's not ideal, especially when APM Server is managed via the Elastic Agent, users do not have immediate feedback. I'd rather log it as error in the APM Server and still start the server. Can you create a follow up issue?

_Originally posted by @simitt in https://github.com/elastic/apm-server/issues/6805#issuecomment-986692345_

PR #6991, opened 2022-01-05
[java-attacher] invalid config fails open

stuartnelson3:
## Motivation/summary

Currently, an invalid java-attacher config prevents the server from starting up. Server operation should not depend on the java-attacher having a valid config. Log the error if it exists, and move along.

QUESTION:

Currently, this is being logged as an error. Do we want to use a different log level for this?

## Checklist

## How to test these changes

1. start `apm-server` with the following added to the config:
    ```yaml
    apm-server:
      java_attacher:
        enabled: true
        discovery-rules:
        - disallowed-flag: opbeans
    ```
2. Check the error logs for a message about the invalid flag
3. The apm-server should not exit

## Related issues

closes https://github.com/elastic/apm-server/issues/6814

### DIFF
--- a/beater/config/config.go
+++ b/beater/config/config.go
@@ -114,7 +114,7 @@ func NewConfig(ucfg *common.Config, outputESCfg *common.Config) (*Config, error)
 	}
 
 	if err := c.JavaAttacherConfig.setup(); err != nil {
-		logger.Errorf("failed to setup java-attacher: %v", err)
+		logger.Warnf("failed to setup java-attacher: %v", err)
 		c.JavaAttacherConfig = defaultJavaAttacherConfig()
 	}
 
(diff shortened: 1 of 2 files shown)


beater/config/config.go
```diff
@@ -114,7 +114,8 @@ func NewConfig(ucfg *common.Config, outputESCfg *common.Config) (*Config, error)
 	}
 
 	if err := c.JavaAttacherConfig.setup(); err != nil {
-		return nil, err
+		logger.Errorf("failed to setup java-attacher: %v", err)
```
axw, line 117: I'd probably make it a warning, which (to me at least) implies that it's not going to stop the server from functioning generally.

axw (approved):

beater/config/config.go
stuartnelson3: Makes sense, thanks!

PR #7044, opened 2022-01-12
script: fix latest_snapshot_version

axw:


axw:
Not a fix at all, it was checking the right thing.

PR #7054, opened 2022-01-12
add backport-8.1 to mergify

stuartnelson3:
following instructions from [RELEASES.md](https://github.com/elastic/apm-server/blob/master/RELEASES.md#after-feature-freeze)

Question: We used to always backport to `7.x` and branch on feature freeze, but it appears we're now doing a new branch for each minor version in `8.x`? Is that correct? If so, should I be creating branch `8.1`?


axw (changes requested):
I presume at some point we will set master to 9.0, and create 8.x. We create a specific 8.1 branch when 8.1 FF happens.

stuartnelson3:
>I presume at some point we will set master to 9.0, and create 8.x. We create a specific 8.1 branch when 8.1 FF happens.

So if I'm understanding correctly:
- update the mergify config to use `8.x` instead of `8.1`
- create the `8.x` branch (created from `8.0` when we release it?)
- update master's version to `9.0`
- create the `backport-8.x` label (also when we create `8.x` branch)

I updated the PR based on these assumptions, but we want to wait to actually merge this for when the `8.x` branch exists

axw:
Sorry for being unclear. No, we shouldn't be creating 8.x yet. That'll come some time later, when all the products set their main branch to 9.0. We also shouldn't create 8.1 yet; that'll happen after 8.1 FF.

IMO we should close this and reopen it at 8.1 FF time when we create the 8.1 branch.

stuartnelson3:
:ok_hand: sounds good, thanks for clarifying!

axw (approved):

Issue #5948, opened 2021-08-16
[docs] Document upgrade order when moving asset setup to Kibana


simitt:
When Kibana (Fleet) takes care of the APM package installation and upgrade (installing templates, pipelines, etc), a more strict upgrade order will be required, where ES and Kibana need to >= APM Server version. 
This will be the case for managed APM Server and for standalone APM Server >= `8.0`.

This needs to be reflected in 
* [ ] support matrix
* [x] docs #7103

simitt:
Kibana will not automatically install the APM Package, but it will bundle with the APM Integration in the future (potentially from 8.2 on). The upgrade order for 8.0 is therefore not that strict, but still recommended to first upgrade Elasticsearch and Kibana, upgrade the APM Integration, and then upgrade the APM Server (or the Elastic Agent bundling the APM Server).

bmorelli25:
This shouldn't have been closed -- AFAIK, the product support matrix hasn't been updated.

bmorelli25:
Per your comment above, we should recommend that ES/Kib are upgraded first (which we do in the upgrade guides). Since bundling isn't happening in 8.0, does anything need to change in the support matrix for 8.0?

simitt:
Bundling is going to happen in `8.1`. Nothing to be reflected on the support matrix for `8.0`. We should be good to close this.

PR #7103, opened 2022-01-20
docs: Upgrade to 8.0

bmorelli25:
### Summary

✅ Adds four **8.0** upgrading guides:

- Self-installation standalone (legacy)
- Self-installation APM integration
- Elastic Cloud standalone (legacy)
- Elastic Cloud APM integration

✅ Removes four of the older upgrading guides:

- Upgrade to data streams (no longer relevant)
- Upgrade to 7.7
- Upgrade to 7.0
- Upgrade to 6.5

✅ **In addition, this PR removes all of the changelogs and breaking changes for all 7.x versions.** I'd like to propose that starting in 8.0, we only publish release notes, breaking changes, and upgrading guides for the current major. Why?

- After removing the `setup.*`, `apm-server.ilm.*`, `output.elasticsearch.pipeline`, etc. docs, a lot of our 7.x release notes (in the 8.0 docs) were broken. We had links to a lot of settings and pages that don’t exist anymore.
- This cleans up a lot of the noise in our upgrading and release notes sections. I've added links to the 7.17 docs so users can easily travel back in time when they need to.
- The Elasticsearch and Kibana docs similarly only show info related to the current major.

### Preview this PR

**Click [here](https://apm-server_7103.docs-preview.app.elstc.co/guide/en/apm/guide/master/upgrading-to-8.x.html) for a preview of these changes.**

### Related

Closes https://github.com/elastic/apm-server/issues/5948.


docs/upgrading-to-8.x.asciidoc
```diff
@@ -0,0 +1,233 @@
...
+--
++
+For more details, or for bulk upgrade instructions, see
+{fleet-guide}/upgrade-elastic-agent.html[Upgrade {agent}]
+
+// ********************************************************
+
+[[upgrade-8.0-cloud-standalone]]
+==== Upgrade {ecloud} APM Server standalone to {version}
+
```
bmorelli25, line 184: How does an ESS standalone upgrade work? I expected the APM integration to be installed as a part of this upgrade, but it wasn't—yet my data was now being stored in data streams. 

I confirmed I was still in standalone mode by finding the upgrade option in the APM app.

docs/upgrading-to-8.x.asciidoc
```diff
@@ -0,0 +1,233 @@
+[[upgrading-to-8.x]]
+=== Upgrade to version {version}
+
+Version {version} of Elastic APM offers a number of benefits over previous versions:
+
+* Reduced number of fields per index, better space efficiency, and faster queries
+* More granular data control
+* Errors and metrics data streams are shared with other data sources -- which means better long-term integration with the logs and metrics apps
+* Removes template inheritance for index lifecycle policies and makes use of new {es} index and component templates
```
bmorelli25, line 9: I'm not really sure what to put here. I stuck with the data stream benefits as those do fit.

docs/upgrading-to-8.x.asciidoc
simitt: Which version did you test with? You need to use at least `8.0.0-rc-1`, as the `8.0.0-beta` candidate did not include the changes. When data were written to datastreams after the upgrade, it actually does sound like the apmpackage has been installed; otherwise APM Server would block the ingestion.

docs/upgrading-to-8.x.asciidoc
simitt: We don't have to motivate the changes from legacy to v2 index&component templates here, as this is a change that ES requires. I suggest to keep it brief and link to the ES docs. 
In general, component templates allow to build reusable template blocks that can then be used in multiple index templates. The definition order defines which component overwrites other component's definition for the same settings. It gives a lot of flexibility and ES also offers a [simulation API](https://www.elastic.co/guide/en/elasticsearch/reference/current/simulate-multi-component-templates.html) to test the resulting index template.

docs/upgrading-to-8.x.asciidoc
bmorelli25: I used `v8.0.0-SNAPSHOT`. I'm not sure how to check what that includes. Agreed that it sounds like the apm package was installed—that's why I was confused when I didn't couldn't find it on the policy. I will try again next week.

axw:
> I'd like to propose that starting in 8.0, we only publish release notes, breaking changes, and upgrading guides for the current major.

:100:

axw (approved):

PR #7153, opened 2022-02-01
7.17.1 bump

stuartnelson3:


axw (approved):
LGTM. Should we wait for https://github.com/elastic/beats/pull/29477 and update again after it merges?

stuartnelson3:
Ah good point, yeah that sounds like a good idea to me

PR #7155, opened 2022-02-01
apmpackage: set default values for tbs

simitt:
## Motivation/summary
Add defaults to the apmpackage for TBS

### DIFF
--- a/apmpackage/apm/manifest.yml
+++ b/apmpackage/apm/manifest.yml
@@ -149,8 +149,12 @@ policy_templates:
             multi: true
           - name: tail_sampling_policies
             type: yaml
+            multi: true
+            default: |
+              - sample_rate: 0.1
           - name: tail_sampling_interval
             type: text
+            default: "1m"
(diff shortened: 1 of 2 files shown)


axw (approved):

axw:
I think we need to update systemtest code to parse the multi-YAML default value.

axw:
Oops, I don't think 3e51638a585cc370416f486b466a1a53493bdd6b is the right fix. It's being unmarshalled, but not stored, so it's not clear to me why that would work.

axw:
Tested with apm-integration-testing 8.1.0 BC2.

By default, TBS is disabled as expected:

image

Enabling TBS, a default interval and catch-all sampling policy are set:

image

Rendered policy includes:

```yaml
apm-server:
  ...
  sampling:
    tail:
      enabled: true
      interval: 1m
      policies:
        - sample_rate: 0.1
```

Updating the default config is reflected in the rendered policy.

PR #7190, opened 2022-02-03
systemtest: fix runapm -var flag

axw:
Fix runapm's `-var` flag: initialise the map, so using the flag doesn't panic.


axw:
@Mergifyio backport 8.1

PR #7197, opened 2022-02-04
runapm: make policy name and reinstall flags

axw:
Make the policy name configurable, and make reinstallation of the APM integration package optional, through additional CLI flags.

This makes it possible to run multiple agents, each with their own integration policy. This can be useful when multiple APM Servers are required, e.g. for tail-based sampling.


axw:
@Mergifyio backport 8.1

Issue #7151, opened 2022-01-31
OpenTelemetry missing  the architecture guide


simitt:
**APM Server version** (`apm-server version`): >= 7.16 

**Description of the problem including expected versus actual behavior**:
When restructuring the APM guide, we missed to update the OTEL Elastic Architecture diagram; it is showing a placeholder for the new and the legacy guide. 

<img width="1153" alt="Screenshot 2022-01-31 at 10 39 37" src="https://user-images.githubusercontent.com/5555349/151770810-7fd46e6d-6520-4fbe-8036-60047954c718.png">

<img width="1175" alt="Screenshot 2022-01-31 at 10 39 51" src="https://user-images.githubusercontent.com/5555349/151770822-bc1b694f-82ff-48a9-a622-d5d4970c4fc5.png">

**Expected Behavior**
The diagram should show what we had in the `7.15` version:
<img width="1166" alt="Screenshot 2022-01-31 at 10 41 26" src="https://user-images.githubusercontent.com/5555349/151771014-242d9ae5-33b0-48e9-85e8-0289e839161f.png">

cc @bmorelli25

PR #7218, opened 2022-02-08
docs: Fix OTel img bug

bmorelli25:
Closes https://github.com/elastic/apm-server/issues/7151.


marclop (commented):
am I reading the diff correctly? It seems that the images were duplicated, they're both in `docs/images/...` and `images/...`

bmorelli25:
> am I reading the diff correctly? It seems that the images were duplicated, they're both in docs/images/... and images/...

Yeah, the images looked weird locally so I was trying to use the PR build to see what was going on. Everything looks good now and images have been deduplicated.

Issue #5999, opened 2021-08-20
Replace <event>.duration.us with event.duration


simitt:
Replace `<event>.duration.us` with `event.duration` - align with the UI team.

marclop:
I'm having  a look at this and the description is a little vague, by having a look at the linked PR, we will stop setting individual event `event.{transaction,span,profile}.duration.us` and set the `event.duration` field instead with the underlying Nanosecond that `time.duration` uses, correct?

The APM package index template fields need to be modified to use the `event.duration` field from ECS instead of our field, which needs to be removed.
- profile_metrics
- rum_traces (transaction and span)
- traces (transaction and span)

This needs to then be coordinated with the APM UI team so that the Kibana APM application is modified to instead use `event.duration`.

Is there anything else that needs to be taken care of?

axw:
@marclop aside from the other obvious change needed in apm-server to emit the `event.duration` field instead of the existing ones, the only other required change I can think of is to use `event.duration` in packaged ML jobs.

Changes to the UI and ML jobs would be the most complicated, so we should discuss this with them before embarking. Can you please discuss with the UI team? They may also know whether it's viable to update the ML jobs.

felixbarny:
Is there a corresponding UI issue for this?
The hardest part about it will be backward compatibility. For old data, `event.duration` is not available so the UI should fall back to `<transaction|span>.duration.us`. AFAIK, that's an unresolved issue how to do that in a transparent way.

axw:
> Is there a corresponding UI issue for this?

Doesn't look like it. We had (have?) one for `timestamp.us`, but that turned out to be too difficult to target 8.0.

> The hardest part about it will be backward compatibility. For old data, event.duration is not available so the UI should fall back to <transaction|span>.duration.us. AFAIK, that's an unresolved issue how to do that in a transparent way.

Agreed. If we can do this for 8.0, I think it will depend on us being able to do it without any kind of runtime field magic. Like, querying both fields.

Mpdreamz:
What if starting 8.0 we make `event.{transaction,span,profile}.duration.us` fields all field aliases to `event.duration.us` ? 

That way the producing side can switch over to `event.duration.us` and we push the switch from the consuming side out after the active deprecation period.

axw:
@Mpdreamz not a bad idea, but there are some complications:

[...]

PR #7261, opened 2022-02-11
Set `<event>.duration.us` field in ingest pipeline

axw:
## Motivation/summary

Update package `model` to no longer set `transaction.duration.us` or `span.duration.us`, and instead set `event.duration`. Update our ingest pipelines to use `event.duration` to set the legacy fields, and then remove `event.duration`. This pipeline can be removed once when we are ready and able to update the APM UI to use `event.duration` directly.

I have listed this as a breaking change in the top-level changelog, as it will have visible effects for anyone not using the Elasticsearch output or the packaged ingest pipeline. I have not done this in the integration package changelog, as I assume the ingest pipeline is always in use there.

## Checklist

## How to test these changes

Send transactions and spans with zero and non-zero durations, and ensure they each have `transaction.duration.us` or `span.duration.us` set, and do not have `event.duration` set.

## Related issues

Prerequisite of https://github.com/elastic/apm-server/issues/5999
Part of https://github.com/elastic/apm-server/issues/3565 and https://github.com/elastic/apm-server/issues/4120

### DIFF
--- a/apmpackage/apm/changelog.yml
+++ b/apmpackage/apm/changelog.yml
@@ -3,9 +3,9 @@
 # change type can be one of: enhancement, bugfix, breaking-change
 - version: "8.2.0"
   changes:
-    - description: Placeholder
+    - description: updated traces and rum_traces ingest pipelines to translate `event.duration` to `<event>.duration.us`
       type: enhancement
-      link: https://github.com/elastic/apm-server/pull/123
+      link: https://github.com/elastic/apm-server/pull/7261
 - version: "8.1.0"
   changes:
     - description: Added field mapping for `faas.coldstart` and `faas.trigger.type`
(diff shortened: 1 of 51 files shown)


axw:
The integration tests are failing, so we'll need to update the package before merging this.

simitt (approved):
Change and handling of breaking change LGTM

stuartnelson3:
confirmed with 772e8c1977e36f0c1d06c2af20f7cda893789365

1. start elasticsearch/kibana
2. start apm-server
3. ingest spans/transactions with the following program:
```go
package main

import (
	"context"
	"fmt"
	"os"
	"time"

	"go.elastic.co/apm"
)

func main() {
	version := "undefined"
	if len(os.Args) > 1 {
		version = os.Args[1]
	}

	span.Duration = 0
	span.End()
	tx.Duration = 0
	tx.End()
	<-time.After(time.Second)
	apm.DefaultTracer.Flush(nil)
}
```
4. check the transaction and span documents have `transaction.duration.us` and `span.duration.us` in the console
```
GET traces-apm-default/_search
{
  "sort" : [{ "@timestamp" : "desc" }]
}
```
5. repeat steps 3 and 4, modifying `span.Duration` and `tx.Duration` to be non-zero

PR #7262, opened 2022-02-11
go mod tidy

axw:
On main, `make test` was failing locally with

```
go: updates to go.mod needed; to update it:                                                          
        go mod tidy                                                                                                                                                                                        
```

And I had to add "go-lookslike". Not sure why that didn't happen in CI...


axw:
Closing this, I'm going to update otel-collector so I'll fix this at the same time.

PR #7264, opened 2022-02-11
ci: notify update status by email/slack

v1v:
## Motivation/summary

Proactively sent a notification when something bad happening with the update-beats automation. Notify when the update-beats automation created a Pull Request (maybe this is not needed...)


axw (approved):
Thank you!

PR #7281, opened 2022-02-14
Only overwrite service.version if service.name is also set

tobiasstadler:
## Motivation/summary

For metricsets the `service.version` is only overwritten if also `service.name` is given (see https://github.com/elastic/apm-server/pull/6407#issuecomment-950973237). It should be the same for transactions, errors, ...

## Checklist

## How to test these changes

Ingest a transaction that only contains a `service.version`, but no `service.name`

### DIFF
--- a/beater/test_approved_es_documents/TestPublishIntegrationErrors.approved.json
+++ b/beater/test_approved_es_documents/TestPublishIntegrationErrors.approved.json
@@ -327,8 +327,7 @@
                 "runtime": {
                     "name": "node",
                     "version": "8.0.0"
-                },
-                "version": "5.1.3"
+                }
             },
             "source": {
                 "ip": "12.53.12.1"
--- a/beater/test_approved_es_documents/TestPublishIntegrationEvents.approved.json
+++ b/beater/test_approved_es_documents/TestPublishIntegrationEvents.approved.json
@@ -261,8 +261,7 @@
                 "runtime": {
                     "name": "Java",
                     "version": "10.0.2"
-                },
-                "version": "4.3.0"
+                }
             },
             "source": {
                 "ip": "12.53.12.1"
@@ -400,8 +399,7 @@
                 "runtime": {
                     "name": "Java",
                     "version": "10.0.2"
-                },
-                "version": "4.3.0"
+                }
             },
             "span": {
                 "action": "connect",
@@ -602,8 +600,7 @@
                 "runtime": {
                     "name": "Java",
                     "version": "10.0.2"
-                },
-                "version": "4.3.0"
+                }
             },
             "source": {
                 "ip": "12.53.12.1",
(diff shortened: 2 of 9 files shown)


axw (approved):
LGTM, thanks!

tobiasstadler:
@axw I fixed the tests. `make test` and `make system-test` finished successfully on my machine.

tobiasstadler:
Since this is some kind of bug fix, should it be back ported to 8.1?

axw:
@tobiasstadler we have already frozen 8.1. Although it's arguable a bug fix, I'd prefer that it waits for 8.2. Is it causing any issues for you?

tobiasstadler:
run elasticsearch-ci/docs

tobiasstadler:
Thank You!

stuartnelson3:
testing this was unsuccessful with 772e8c1977e36f0c1d06c2af20f7cda893789365

1. start elasticsearch/kibana
2. start apm-server
3. ingest span + transaction with the following program:
```go
package main

import (
	"context"
	"fmt"
	"os"
	"time"

	"go.elastic.co/apm/v2"
)

func main() {
	version := "undefined"
	if len(os.Args) > 1 {
		version = os.Args[1]
	}

	tracer, err := apm.NewTracer("", "1.0.0")
	if err != nil {
		panic(err)
	}
	span, ctx := apm.StartSpan(ctx, "name", "type")
	span.End()
	tx.End()
	<-time.After(time.Second)
	tracer.Flush(nil)
	fmt.Printf("%s: %+v\n", name, tracer.Stats())
}
```
4. check the transaction document has no `service.version` set. In the console:
```
GET traces-apm-default/_search
{
  "sort" : [{ "@timestamp" : "desc" }]
}
```
`service.name` seems to have been set to the default value (the binary name) from the agent, and therefore `service.version` was also set.
I attempted to injest an ndjson document without a `service.name` set, but the apm server rejected it (`validation error: metadata: service: 'name': validation rule 'minLength(1)' violated`)

axw:
@stuartnelson3 I'm not sure that your tests cover Tobias's changes.

The changes are related to overriding `service.name` and `service.version` at the event level, which the Go agent's API doesn't support. Your test above only affects the `service.name` and `service.version` set at the stream (metadata) level.

I've verified this with 8.2.0-BC1, using https://github.com/elastic/apm-server/blob/main/testdata/intake-v2/transactions_spans.ndjson#L2.

The first transaction overrides `service.name`, but does not specify `service.version`. I confirmed the `service.name` field is set to "serviceabc" but there is no `service.version` field set.

I then modified the document to override `service.version` but not `service.name`, and confirmed that neither are used and that `service.name` and `service.version` from metadata are both used.

Finally, I confirmed that if both `service.name` and `service.version` are specified in the overrides, both are used.

PR #7296, opened 2022-02-15
ingest: don't overwrite <event>.duration.us

axw:
## Motivation/summary

Don't overwrite `transaction.duration.us` or `span.duration.us` if they are set, as otherwise old versions of APM Server cannot send events to newer versions of the ingest pipeline, which is expected to happen.

## Checklist

## How to test these changes

1. Install the integration
2. Run APM Server 8.1.0 (i.e. an older version which sets the `<event>.duration.us` fields directly), send some events
3. Check that the events are indexed with non-zero durations

## Related issues

None. apm-integration-testing has been failing on 8.0 and 8.1 branches, as they are (accidentally?) running the 8.2.0 stack, and hence the 8.2.0 integration. The integration tests were failing because the durations were all coming out as zero.

### DIFF
--- a/apmpackage/cmd/genpackage/pipelines.go
+++ b/apmpackage/cmd/genpackage/pipelines.go
@@ -112,11 +112,17 @@ var clientGeoIPPipeline = []map[string]interface{}{{
 	},
 }}
 
+// This pipeline translates `event.duration` (defaulting to zero if not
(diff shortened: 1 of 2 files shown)


marclop (approved):

stuartnelson3:
confirmed using integration `v8.2.0-dev5` and apm-server `8.1.2`

PR #7298, opened 2022-02-15
ci: use k8s pods

v1v:
PoC regarding k8s pods

### Issues

- https://github.com/elastic/apm-pipeline-library/issues/1552
- NODE_LABELS

### Blockers

- make: not found in the existing docker image, so we might need to build our own docker image with the toolchain

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -62,24 +62,29 @@ pipeline {
                     shallow: false, reference: "/var/lib/jenkins/.git-references/${REPO}.git")
         stash allowEmpty: true, name: 'source', useDefaultExcludes: false
         setEnvVar("GO_VERSION", readFile("${BASE_DIR}/.go-version").trim())
-        script {
-          dir("${BASE_DIR}"){
-            def regexps =[
-              "^_beats.*",
-              "^apm-server.yml",
-              "^apm-server.docker.yml",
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -547,3 +551,13 @@ pipeline {
     }
   }
 }
+
+
+def runInPod(version, Closure body) {
+  container(version.replaceAll('\\.', '-')) {
```
kuisathaverat, line 557: not needed you can set the default container to the Go container

Jenkinsfile
```diff
@@ -107,10 +109,10 @@ pipeline {
       }
       steps {
         withGithubNotify(context: 'Intake') {
-          deleteDir()
-          unstash 'source'
-          dir("${BASE_DIR}"){
-            withGoEnv(){
+          runInPod(env.GO_VERSION) {
```
kuisathaverat, line 112: you can interpolate variables in the pod definition, so you can launch small pods

simitt:
@v1v is this something that you are going to be picking up again or should we close this for now?

v1v:
> @v1v is this something that you are going to be picking up again or should we close this for now?

@simitt , we are working on this but in a best effort, let me close this as I cannot say when I'll pick this up yet

PR #7312, opened 2022-02-16
Typo fix

bZichett:
## Checklist


axw:
Thanks @bZichett! Would you please [sign the CLA](https://www.elastic.co/contributor-agreement)? We can't accept changes without that. If you would prefer, we can make the fix instead.

bZichett:
@axw: You should see my CLA submission now. 

I'd like to be able to contribute in the future, so although this change is a no-op, curious to just get this step out of the way.

axw (approved):
Thanks @bZichett :)

PR #7315, opened 2022-02-17
Revert "[mage] filter .dmg package creation"

graphaelli:
This reverts commit 7a3edfdba03e5fa2e448fa92cbab401b42a08150.

`mage.DMG` is not defined (edit: removed in https://github.com/elastic/beats/pull/30097 and pulled into main since then).  I'm not sure if this was pushed by accident but it doesn't seem to have come via a pull request.  
@stuartnelson3


axw (approved):
Thanks @graphaelli

Issue #4586, opened 2020-12-24
Document tail-based sampling


axw:
Document the tail-based sampling feature https://github.com/elastic/apm-server/issues/4583.

~This should be considered blocked until https://github.com/elastic/apm-server/issues/4583 is fixed.~

PR #7317, opened 2022-02-17
docs: Tail-based sampling

bmorelli25:
### Summary

This PR adds tail-based sampling documentation.

I tested the feature by spinning up an 8.1 instance on ESS and pointing APM integration testing at it:

```
./scripts/compose.py start 8.1 --no-elasticsearch --no-kibana --no-apm-server --with-opbeans-go --with-opbeans-java --with-opbeans-node --apm-server-url=${APM_SERVER_URL} --apm-server-secret-token=${APM_SERVER_TOKEN} --opbeans-elasticsearch-url=${ES_URL}
```

### Preview this PR

* [Transaction sampling](https://apm-server_7317.docs-preview.app.elstc.co/guide/en/apm/guide/master/sampling.html)
  * [Head-based sampling config](https://apm-server_7317.docs-preview.app.elstc.co/guide/en/apm/guide/master/configure-head-based-sampling.html)
  * [Tail-based sampling config](https://apm-server_7317.docs-preview.app.elstc.co/guide/en/apm/guide/master/configure-tail-based-sampling.html)

### Related

* Closes https://github.com/elastic/apm-server/issues/4586
* Closes https://github.com/elastic/observability-docs/issues/916
* Issue to add link to the UI: https://github.com/elastic/kibana/issues/124265


docs/legacy/guide/trace-sampling.asciidoc
```diff
@@ -12,6 +12,9 @@ For example, a sampling value of `.2` indicates a transaction sample rate of `20
 This means that only `20%` of traces will send and retain all of their associated information.
 The remaining traces will drop contextual information to reduce the transfer and storage size of the trace.
 
+TIP: The APM integration supports both head-based and tail-based sampling.
+Learn more <<sampling,here>>.
```
bmorelli25, line 16: Is it true that tail-based sampling is only supported with the APM integration?

docs/sampling.asciidoc
```diff
@@ -1,27 +1,77 @@
...
+**Distributed tracing with tail-based sampling**
+
+With tail-based sampling, all traces are observed and a sampling decision is only made once a trace completes.
+
+In this example, `Service A` initiates four transactions.
+If our sample rate is `.5` (`50%`) for traces with a `success` outcome,
+and `1` (`100%`) for traces with a `failure` outcome,
+the sampled traces would look something like this:
+
+image::./images/dt-sampling-example-3.png[Distributed tracing and tail based sampling example one]
```
bmorelli25, line 71: Is this a good example ([link to img](https://github.com/elastic/apm-server/blob/5fb6cf4e2a4f5d91b721e54dcf9dee4f453a304d/docs/images/dt-sampling-example-3.png))? If not, what would you recommend?

docs/sampling.asciidoc
```diff
@@ -1,27 +1,77 @@
...
+This means all traces will be analyzed against a set of rules, or policies, which will determine the rate at which they are sampled.
 
-The goal of probability sampling is to provide you with a representative set of data that allows
-you to make statistical inferences about the entire group of data.
-In other words, in most cases, you can still find anomalous patterns in your applications, detect outages, track errors,
-and lower MTTR, even when sampling at less than `100%`.
+Tail-based sampling reduces the risk of discarding important data, because the sampling decision is only made _after_
+each trace has been analyzed.
+However, because traces are all initially observed,
+storage and transfer costs may be higher than with head-based sampling.
```
bmorelli25, line 58: Is this true? Are there other downsides?

docs/sampling.asciidoc
```diff
@@ -37,37 +87,25 @@ This means the following data will always accurately reflect *all* of your appli
...
-The initiating service makes the sampling decision in a distributed trace,
-and all downstream services respect that decision.
-
-In each example below, `Service A` initiates four transactions.
-In the first example, `Service A` samples at `.5` (`50%`). In the second, `Service A` samples at `1` (`100%`).
-Each subsequent service respects the initial sampling decision, regardless of their configured sample rate.
-The result is a sampling percentage that matches the initiating service:
-
-image::./images/dt-sampling-example.png[How sampling impacts distributed tracing]
+Regardless of the above, cost conscious customers are likely to be fine with a lower sample rate.
```
bmorelli25, line 103: Please double-check that these are good examples

docs/legacy/guide/trace-sampling.asciidoc
axw: Yes.

docs/sampling.asciidoc
```diff
@@ -1,27 +1,77 @@
...
+[float]
+[[head-based-sampling]]
+==== Head-based sampling
+
+In head-based sampling, the sampling decision for each trace is made when that trace is initiated.
+Each trace has a defined and equal probability of being sampled.
 
 For example, a sampling value of `.2` indicates a transaction sample rate of `20%`.
 This means that only `20%` of traces will send and retain all of their associated information.
 The remaining traces will drop contextual information to reduce the transfer and storage size of the trace.
```
axw, line 24: Since 8.0, we drop non-sampled transactions (except for RUM) in APM Server. Should we update this?

docs/sampling.asciidoc
axw: Kinda. There are pros and cons.

Cons:
 - more data sent from APM agents to APM Server
 - APM Server will use more CPU, memory, and disk (without TBS, APM Server won't hit the disk at all except for maybe swapping)

Pros (aside from the feature itself):
 - less data sent from APM Server to Elasticsearch

I think that pro is pretty important to call out. If you run APM Server close to the instrumented application, then the additional transfer cost can be reduced. Since TBS is implemented in APM Server, data is dropped at APM Server and so less data is transferred to Elasticsearch.

docs/sampling.asciidoc
axw: Looks good :+1:

docs/sampling.asciidoc
```diff
@@ -1,27 +1,77 @@
...
+In this example, `Service A` initiates four transactions.
+If our sample rate is `.5` (`50%`) for traces with a `success` outcome,
+and `1` (`100%`) for traces with a `failure` outcome,
+the sampled traces would look something like this:
+
+image::./images/dt-sampling-example-3.png[Distributed tracing and tail based sampling example one]
 
 [float]
-==== What data is sampled?
+=== Sampled data
```
axw, line 74: As above, we should probably update these docs, as we no longer store non-sampled trace events as of 8.0. (Again, RUM is special.)

docs/sampling.asciidoc
axw: Looks great. Maybe also mention outcome? i.e. failed operations may be of more interest than successful ones.

docs/sampling.asciidoc
```diff
@@ -1,27 +1,77 @@
...
+
+In this example, `Service A` initiates four transactions and has a sample rate of `1` (`100%`).
+Again, the sample rates of `Service B` and `Service C` are ignored.
+
+image::./images/dt-sampling-example-2.png[Distributed tracing and head based sampling example two]
+
 [float]
-==== Why sample?
+[[tail-based-sampling]]
+==== Tail-based sampling
```
axw, line 50: Should we also mention the fact that tail-based sampling uses weighted random sampling? We weight the random sampling of traces based on the root transaction duration, which has the effect of preferring to keep slower traces rather than faster ones.

axw (commented):
Looks great!

docs/sampling.asciidoc
bmorelli25: Yes! I wasn't aware of this. I'll add it in.

docs/sampling.asciidoc
bmorelli25: Very neat! Thanks.

docs/sampling.asciidoc
bmorelli25: Ah, that's right. I'll update.

docs/sampling.asciidoc
axw: Yeah, errors are exempt from sampling. It's just transactions and spans that may be dropped.

docs/sampling.asciidoc
```diff
@@ -1,101 +1,129 @@
...
+Regardless of the sampling decision, all traces retain <<data-model-errors,error>> data.
 
-In each example below, `Service A` initiates four transactions.
-In the first example, `Service A` samples at `.5` (`50%`). In the second, `Service A` samples at `1` (`100%`).
-Each subsequent service respects the initial sampling decision, regardless of their configured sample rate.
-The result is a sampling percentage that matches the initiating service:
+Sampling does not impact aggregated transaction and span <<data-model-metrics,metrics>>.
+This means that the visualizations in the APM app, like latency, throughput, time spent by span type, failed transaction rate,
+errors, error rate, etc., will always accurately reflect *all* of your application's requests,
+regardless of the configured sampling rate.
```
axw, line 87: This might be a bit of an overstatement. Sampling may affect accuracy of latency.

Metrics are based on the sampled traces, and weighted by the inverse sampling rate. e.g. if you are sampling at 5%, then each trace will be counted as 20. Thus, as variance of latency increases, or the sampling rate decreases, your level of error will increase.

docs/sampling.asciidoc
axw: Should we remove "The remaining traces will drop contextual information to reduce the transfer and storage size of the trace."? Some agents will start automatically dropping non-sampled transactions agent-side, when they detect a new enough version of APM Server. In the future this should be the default.

Maybe say they'll be dropped (without specifying where) and link to `Sampled data and visualizations`?

bmorelli25:
Circling back to this now that PTO and Space, Time are over.

docs/sampling.asciidoc
bmorelli25: Hmm. Updated. I'm still a bit confused about this so please let me know if it still needs work (likely).

simitt (approved):
Left a couple minor comments, but otherwise looks good. Feel free to merge after addressing the last bit.

bmorelli25:
@elasticmachine, run elasticsearch-ci/docs

PR #7352, opened 2022-02-21
modelindexer: fix flush-on-close timing issues

axw:
## Motivation/summary

If the background timer goroutine was started, but had not yet added to the errgroup, then Close could return before events enqueued by ProcessEvents were flushed.

We fix this by ensuring the errgroup.Go call is made before ProcessEvents returned. If the timer is stopped (e.g. because Close is called), then we signal to the timer goroutine to flush immediately, instead of calling flushActive(Locked) from multiple code paths.

This should fix the `TestTransactionAggregationShutdown` failures we occasionally see in CI.

## Checklist

## How to test these changes

This is not likely to happen in a real setup. Run `TestTransactionAggregationShutdown` a bunch of times?

## Related issues

None

### DIFF
--- a/model/modelindexer/indexer.go
+++ b/model/modelindexer/indexer.go
@@ -80,12 +80,13 @@ type Indexer struct {
 	available chan *bulkIndexer
 	g         errgroup.Group
 
-	mu       sync.RWMutex
-	closing  bool
-	closed   chan struct{}
-	activeMu sync.Mutex
-	active   *bulkIndexer
-	timer    *time.Timer
+	mu           sync.RWMutex
+	closing      bool
+	closed       chan struct{}
+	activeMu     sync.Mutex
+	active       *bulkIndexer
+	timer        *time.Timer
+	timerStopped chan struct{}
 }
 
 // Config holds configuration for Indexer.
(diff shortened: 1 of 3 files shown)


axw:
> [2022-02-21T06:41:43.797Z] Get "https://docker.elastic.co/v2/": dial tcp 34.68.230.202:443: connect: connection refused

model/modelindexer/indexer.go
```diff
@@ -309,13 +318,7 @@ func encodeMap(v map[string]interface{}, out *fastjson.Writer) error {
...
-func (i *Indexer) flushActive() {
-	i.activeMu.Lock()
-	defer i.activeMu.Unlock()
-	i.flushActiveLocked(context.Background())
-}
-
-func (i *Indexer) flushActiveLocked(ctx context.Context) {
+func (i *Indexer) flushActive(ctx context.Context) error {
 	// Create a child context which is cancelled when the context passed to i.Close is cancelled.
 	flushed := make(chan struct{})
```
marclop, line 323: the `flushed` channel is never closed, and it never receives anything, shouldn't we be sending a `struct{}{}` after the`i.Flush()` call (or `close(flushed)`) so the context is cancelled right away and the goroutine exits? Otherwise we may leak goroutines when we this method is called.

axw: Yep, I just forgot to move the `defer close(flushed)` out of the old goroutine. Thanks, good catch.

axw: Thanks, I've adapted your test a little bit and included it with the fix.

marclop (approved):
Looks great!

PR #7156, opened 2022-02-01
bump version 8.2

stuartnelson3:
TODO:

`make update-beats`
~~update go-elasticsearch~~ skipping for now, updating is resulting in an alpha build?

```
# go get github.com/elastic/go-elasticsearch/v8@8.0
go: downloading github.com/elastic/go-elasticsearch/v8 v8.0.0-alpha.0.20220127111727-5269e7e57568
go: downloading github.com/elastic/elastic-transport-go/v8 v8.0.0-alpha
go get: added github.com/elastic/elastic-transport-go/v8 v8.0.0-alpha
go get: upgraded github.com/elastic/go-elasticsearch/v8 v8.0.0-20210727161915-8cf93274b968 => v8.0.0-alpha.0.20220127111727-5269e7e57568
```

PR #7377, opened 2022-02-22
[8.1] update go elasticsearch

simitt:
## Motivation/summary
Update go-elasticsearch dependency to current `8.1` branch https://github.com/elastic/go-elasticsearch/commit/bc3eb8ef52eac56170cfab53c6074a9714e0d376

This requires some code changes, that have been introduced to `main` in https://github.com/elastic/apm-server/pull/7156/ but not be backported.

### DIFF
--- a/elasticsearch/client.go
+++ b/elasticsearch/client.go
@@ -142,16 +142,15 @@ func newV8Client(
 	fn backoffFunc,
 ) (Client, error) {
 	c, err := esv8.NewClient(esv8.Config{
-		APIKey:               apikey,
-		Username:             user,
-		Password:             pwd,
-		Addresses:            addresses,
-		Transport:            transport,
-		Header:               headers,
-		RetryOnStatus:        retryableStatuses,
-		EnableRetryOnTimeout: true,
-		RetryBackoff:         fn,
-		MaxRetries:           maxRetries,
+		APIKey:        apikey,
+		Username:      user,
+		Password:      pwd,
+		Addresses:     addresses,
+		Transport:     transport,
+		Header:        headers,
+		RetryOnStatus: retryableStatuses,
+		RetryBackoff:  fn,
+		MaxRetries:    maxRetries,
 	})
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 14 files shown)


elasticsearch/client.go
```diff
@@ -142,16 +142,15 @@ func newV8Client(
...
 ) (Client, error) {
 	c, err := esv8.NewClient(esv8.Config{
-		APIKey:               apikey,
-		Username:             user,
-		Password:             pwd,
-		Addresses:            addresses,
-		Transport:            transport,
-		Header:               headers,
-		RetryOnStatus:        retryableStatuses,
-		EnableRetryOnTimeout: true,
```
axw, line 152: Is this the default now?

model/modelindexer/indexer.go
```diff
@@ -442,6 +442,10 @@ func (r *pooledReader) Read(p []byte) (int, error) {
 	return n, err
 }
 
+func (r *pooledReader) Seek(offset int64, whence int) (int64, error) {
```
axw, line 445: Why is there a Seek method now? Does this mean that on io.EOF, the reader may be reused? (See Read above)

elasticsearch/client.go
simitt: No, it has been replaced with `RetryOnError` (https://github.com/elastic/go-elasticsearch/pull/385/files). Given that we already define [status codes for retry](https://github.com/elastic/apm-server/blob/main/elasticsearch/client.go#L34-L38), but haven't yet provided any retry methods, I believe this is what comes closest to the previous behavior.

model/modelindexer/indexer.go
simitt: I believe this is related to this change https://github.com/elastic/go-elasticsearch/pull/315.

axw: I took a look at the code, we only implement it because we're still using the elasticsearch.BulkIndexerItem struct. We have our own BulkIndexer implementation which never calls the Seek method, so we're good.

elasticsearch/client.go
axw: I went back through my memory and found https://github.com/elastic/elastic-transport-go/pull/3, where I reviewed the transport change. It does retry timeouts by default now: https://github.com/elastic/elastic-transport-go/pull/3/commits/89919cddeccea10fb860e17a1f085ac9f237f7cf#r767996155

axw (approved):

PR #7403, opened 2022-02-24
docs: add 7.17.1 release notes

dedemorton:


dedemorton:
@bmorelli25 I didn't see a 7.17 file in main, so I've made the change in 7.17 only. I'm assuming that's intentional, but wanted to mention it in case you want to copy the file to main, too.

axw (approved):

Issue #7216, opened 2022-02-07
APM Server Benchmark POC


simitt:
Build a POC for the new APM Server Benchmark Framework, as a follow up from https://github.com/elastic/apm-server/issues/6549. The POC should demonstrate how the single components play together, but leverage existing logic for the test data generation. 
The benchmark framework will split apm data generation from applying load to the apm-server, and can be configured to send benchmark statistics to a dedicated observability cluster. 

TODO(@marclop): Please insert the overview of the 2.0 Benchmark here.

## Trace Generation
A sample app per language agent generates trace data. For the POC we will reuse the opbeans go app to identify any shortcomings of the app. The apps configure an APM Server endpoint to which the apm data are sent. The configured endpoint will be a new component, the Intake Receiver. It's purpose is to imitate the APM Server's intake API and immediately dump the events to disk, in `ndjson` format. When reusing opbeans, the existing load generator for the opbeans app can be reused for this first POC; adaptions are going to be made as needed. 

## Benchmark Generator
We'll continue to use and further develop the existing `apmbench` tooling. A dedicated `loader` component that takes care of reading the generated `ndjson` files from disk, will be implemented. It passes on the events to the `runner` component, which is responsible for sending batches of requests to the apm-server. The apmbench stats collector collects statistics from the apm-server and ingests them into a dedicated ES cluster for analysis.   

## Setup 
The ecetest tooling is reused to spin up environments containing the apm-server, elasticsearch and kibana. The benchmark generator is supposed to be running on a dedicated environment. Setting up the benchmark environment should be straight forward (not require manual installation of several components on a host). 

## Out Of Scope
* No automation will be set up to schedule regular benchmarks based on this prototype. 
* No involvement of other agent developers for producing benchmark data.

marclop:
Most of the PoC has been implemented at this point and is available to be reviewed. The only remaining point other than addressing the comments that are brought up inn the PR reviews is modifying `gobench` collect and index the additional metrics that have been added to `apmbench`.

marclop:
We're going to be implementing the changes in `gobench` as part of the follow up in #7540

PR #7416, opened 2022-02-28
benchmark: Introduce new binary `intake-receiver`

marclop:
## Motivation/summary

Adds a new binary under `cmd/intake-receiver` which is a simplified
implementation of the APM Server's intake API. Instead of persisting the
events to Elasticsearch, the events are stored in _ndjson_ files. Each
of these files the contains traces received by the APM Agents, keeping
the batches just as they are received to be able to replay them during
the benchmark execution phase. The format for these files is:

    events/<agent_name>-<agent_version>.ndjson

This allows us to classify each version of an APM Agent individually and
possibly benchmark and compare different versions of same application in
the future.

The `intake-receiver` doesn't perform aggregations or produce metrics
from the received traces, it simply stores them in plain-text.

For the PoC phase the `intake-receiver` will be used to store traces as
they are generated by the opbeans apps in the `apm-integration-testing`
repository.

## Checklist

## How to test these changes

1. `cd cmd/intake-receiver && go build .`.
2. `./intake-receiver`
3. `curl -H "Content-type: application/x-ndjson" --data-binary @testdata/intake-v3/rum_events.ndjson http://localhost:8200/intake/v3/rum/events`: Should return an error.
4. `curl -H "Content-type: application/x-ndjson" --data-binary @testdata/intake-v2/transactions.ndjson http://localhost:8200/intake/v2/events`
5. `curl localhost:8200/config/v1/agents` 
6. `curl localhost:8200/` 
```
{"publish_ready":true,"version":"8.2.0"}
```
7. Check there are logs in the intake receiver process such as:
```
2022/02/24 21:26:24 processed request to /intake/v3/rum/events (curl/7.77.0) in 790.5µs
2022/02/24 21:26:35 processed request to /intake/v2/events (curl/7.77.0) in 467.084µs
2022/02/24 21:27:12 processed request to / (curl/7.77.0) in 8.416µs
```
8. Close the intake receiver with an interrupt (ctrl+c) and verify it closes the files it has open:
```
^C2022/02/24 21:28:02 closing http server...
2022/02/24 21:28:02 closed file traces/js-base-4.8.1.ndjson
2022/02/24 21:28:02 closed file traces/elastic-node-3.14.0.ndjson
```
9. Verify that stored files match the source files that were used on the test requests.

## Related issues

Part of #7216


cmd/intake-receiver/main.go
```diff
@@ -0,0 +1,331 @@
...
+
+	var host, version, folder string
+	flag.StringVar(&host, "host", ":8200", "port that the server will listen to")
+	// TODO(marclop) how to keep this version in sync with `../cmd/version.go`?
+	flag.StringVar(&version, "version", "8.2.0", "APM Server version")
+	flag.StringVar(&folder, "folder", "traces", "The path where the received intake date will be stored")
+	flag.Parse()
+
+	// Create a context that will be cancelled when an interrupt is received.
+	ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt, os.Kill)
```
stuartnelson3, line 61: somehow i missed `signal.NotifyContext`; this is awesome

cmd/intake-receiver/main.go
```diff
@@ -0,0 +1,331 @@
...
+	defer stop()
+
+	if folder != "." {
```
stuartnelson3, line 64: is this check necessary? `mkdir -p .` doesn't error or cause issues

cmd/intake-receiver/main.go
```diff
@@ -0,0 +1,331 @@
...
+
+	var err error
+	body := r.Body
+	encoding := r.Header.Get("Content-Encoding")
+	switch encoding {
+	case "deflate":
+		body, err = zlib.NewReader(r.Body)
+	case "gzip":
+		body, err = gzip.NewReader(r.Body)
+	}
```
stuartnelson3, line 167: default case for unsupported content-encoding? might be pedantic, since we're only supporting the agents whose encoding we control

cmd/intake-receiver/main.go
```diff
@@ -0,0 +1,331 @@
...
+		buf.Write([]byte("\n"))
+
+		if decodedMeta {
+			// continue scanning
+			continue
+		}
+		if e := json.Unmarshal(scanner.Bytes(), &meta); e != nil {
+			// TODO(marclop) multierror?
+			err = e
+			// Continue scanning
```
stuartnelson3, line 231: is this the behavior in the apm-server, to keep scanning even if one of the lines is malformed?

cmd/intake-receiver/main.go
```diff
@@ -0,0 +1,331 @@
...
+	}()
+	defer body.Close()
+	scanner := bufio.NewScanner(body)
+	scanner.Buffer(byteBuf, 0)
+	var decodedMeta bool
+	var err error
+	for scanner.Scan() {
+		if err := scanner.Err(); err != nil {
+			return err
+		}
```
stuartnelson3, line 213: should this check be outside the `for` loop? from the docs:

```
It returns false when the scan stops, either by reaching the end of the input or an error.
```
https://pkg.go.dev/bufio#Scanner.Scan

so if `scanner.Scan()` encounters an error, it will return false and exit the loop, but the code won't have the opportunity to check the value of `scanner.Err()`?

stuartnelson3 (commented):
looking good! left some comments

cmd/intake-receiver/main.go
marclop: I will add a check for that case, just in case 😄

cmd/intake-receiver/main.go
marclop: yeah I thought so: https://github.com/elastic/apm-server/blob/4dbe93d223e6098d6f83fc5d1e984570d59cbd39/processor/stream/processor.go#L139-L178

cmd/intake-receiver/main.go
```diff
@@ -0,0 +1,331 @@
...
+	// dropped in as a replacement for APM Server. This means, that all the
+	// config options are ignored.
+	flag.String("e", "", "apm-server compatibility option")
+	flag.String("E", "", "apm-server compatibility option")
+	flag.String("httpprof", "", "apm-server compatibility option")
+
+	var host, version, folder string
+	flag.StringVar(&host, "host", ":8200", "port that the server will listen to")
+	// TODO(marclop) how to keep this version in sync with `../cmd/version.go`?
+	flag.StringVar(&version, "version", "8.2.0", "APM Server version")
```
simitt, line 56: Why does this need to be configurable; shouldn't it always be aligned with the server version of the branch? 

As your comment points out, this should be automatically kept in sync with the apm-server version. How about updating the version with the `make update` command for now? (for example, writing to a `version.go` file and then reading that from here)

cmd/intake-receiver/main.go
simitt: I'm not certain how the RUM agent encodes data; a default option seems reasonable.

cmd/intake-receiver/main.go
```diff
@@ -0,0 +1,331 @@
...
+		agentFileMap: &agentFileMap,
+		basePath:     folder,
+		rootResponse: fmt.Sprintf(`{"publish_ready":true,"version":"%s"}`+"\n", version),
+		bufPool:      sync.Pool{New: func() interface{} { return &bytes.Buffer{} }},
+		bytesBufPool: sync.Pool{New: func() interface{} { return make([]byte, maxScannerBufSize) }},
+	}
+	mux := http.NewServeMux()
+	mux.Handle("/", rh.rootHandler())
+	// TODO(marclop): intake/v2/rum is not supported.
+	for _, p := range []string{"/intake/v2/events", "/intake/v3/rum/events"} {
```
simitt, line 88: Have you tested with the a RUM agent?

simitt (commented):
Haven't had time to try it out yet, but the PR and general direction look good so far. 
The main challenge will be how to keep settings in sync, but so far it is only a handful, so I believe this is good enough.

cmd/intake-receiver/main.go
marclop: Just for the sake of the PR review, we assume that the Content-Encoding is "" in the default case in the APM Server intake: https://github.com/elastic/apm-server/blob/2c36ad76b92f92ea89def677d0925cedf5f4f7e3/decoder/req_decoder.go#L56-L83

cmd/intake-receiver/main.go
marclop: I've tested sending the stored `testdata/intake-v3/...` to the intake receiver and decodes the events well, I will test the RUM agent as well.

marclop:
Should we update the `TESTING.md` doc in this PR? I'm thinking that it may be a little premature since the benchmarking work is still WIP. Happy to do so if you folks think it's best we do it in-line with the changes rather than as a follow up.

simitt:
Agree, let's update the benchmarking part in the `Testing.md` when putting it all together.

simitt (approved):
Looks good!

PR #7478, opened 2022-03-07
[8.1] 8.1.0 release notes

bmorelli25:
Adds release notes and breaking changes for 8.1.0.

* A link to TBS docs will be added in the TBS PR.
* Needs to be cherrypicked to `main`.

### DIFF
--- a/changelogs/head.asciidoc
+++ b/changelogs/head.asciidoc
@@ -1,21 +1,13 @@
 [[release-notes-head]]
 == APM version HEAD
 
-https://github.com/elastic/apm-server/compare/8.0\...main[View commits]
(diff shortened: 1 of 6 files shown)


changelogs/head.asciidoc
```diff
@@ -1,28 +1,22 @@
...
 - Fix missing stack monitoring metrics {pull}7428[7428]
 
 [float]
 ==== Intake API Changes
 
 [float]
 ==== Added
+
+
+// still waiting on these?
```
marclop, line 20: I think so? We're looking at doing some more testing before shipping this feature.

Issue #7461, opened 2022-03-04
API Key user privileges requirements not clearly explained and linked


simitt:
API Key docs are not referencing the user role privileges that are required when generating an APM API Key. 
In the legacy APM Server docs, the privileges are described (https://www.elastic.co/guide/en/apm/guide/current/privileges-api-key.html#privileges-api-key) but also only linked from the alternative [API Key creation workflow via APM Server](https://www.elastic.co/guide/en/apm/guide/current/api-key-legacy.html#create-api-key-workflow-apm-server). 

All of the requirements should be described in the same section and be easier to consume. We could either completely remove it and only link to the requirements described in the [Kibana guide](https://www.elastic.co/guide/en/kibana/current/agent-key-api.html), or closely follow that guide. 

If possible, this should be backported to the `7.17` guide. 

cc @bmorelli25

PR #7544, opened 2022-03-15
docs: clarify API key user privileges

bmorelli25:
### Summary

This PR updates the [APM integration API key docs](https://www.elastic.co/guide/en/apm/guide/current/api-key.html) and [APM Server legacy API key docs](https://www.elastic.co/guide/en/apm/guide/current/api-key-legacy.html). Each guide now includes a step explaining:

1. That API keys can only have the same or lower access rights than the user that creates them.
2. The privileges required to create an API key
3. How to create a role that has the appropriate API key privileges

### Related

* Closes https://github.com/elastic/apm-server/issues/7461.
* Additional context was added to the [Kibana APM API docs](https://www.elastic.co/guide/en/kibana/current/agent-key-api.html) in https://github.com/elastic/kibana/pull/127675


docs/legacy/feature-roles.asciidoc
```diff
@@ -230,7 +230,7 @@ IMPORTANT: {deprecation-notice-config}
 You can configure <<api-key-legacy,API keys>> to authorize requests to APM Server.
 To create an APM Server user with the required privileges for creating and managing API keys:
 
-. Create an **API key role**, called something like `apm_api_key`,
+. Create an **API key role**, called something like `apm_agent_key_role`,
```
axw, line 233: I'm not sure about this renaming. There's no need for an agent's API Key to have `sourcemap:write`. OTOH there's not really any need for a non-agent API Key to have either `config_agent:read` or `event:write`, so the example of adding all three to one API Key probably isn't the best to begin with.

axw (commented):
LGTM in general, just one comment about the API Key name change.

simitt (approved):
+1 on Andrew's comment, otherwise looks good.

docs/legacy/feature-roles.asciidoc
bmorelli25: Sorry, this one fell off of my radar. 👍 makes sense—will fix!

PR #7567, opened 2022-03-17
Always set timestamp on APMEvents for incoming http requests

### LINKED ISSUE #7566: Invalid timestamp for RUM events when capture_personal_data is disabled
APM Server does not record the timestamp from incoming RUM events when `capture_personal_data: false`. The bug manifests in serializing the zero timestamp `@timestamp: "0001-01-01T00:00:00.022Z"` and a negative value for `timestamp.us` when sending data to the output. 

By default the config option is set to `true`. This is happening for current versions `>= 8.0`.

simitt:
## Motivation/summary
APM Server does not record the timestamp from incoming RUM events when `capture_personal_data: false`. This PR ensures the timestamp is always set in the initial `APMEvent`. 
The approval tests are ignoring the concrete timestamp, as it is dynamic and would be ever changing. The PR adds functionality allowing to implement a custom check for dynamic attributes before dumping and comparing them to the approved documents. It is used to ensure that the `@timestamp` is not zero, without checking for the actual value. 

## Checklist

### manual test
* spin up apm-server `>= 8.0`, enable rum and configure `apm-server.capture-personal-data: false` (standalone or managed mode doesn't matter)
* send a rum request to the apm-server 
* observe that the created `@timestamp` is a valid timestamp

When the same test is run on current `main` branch (without this fix) the `@timestamp` value would be the `zero` value. 

### system test
Without the fix, the newly introduced test was failing like this
```
2022/03/17 18:10:53 Built /Users/simitt/coding/go/elastic/apm-server/apm-server
--- FAIL: TestRUMCaptureNoPersonalData (5.79s)
    rum_test.go:255: calling Check() for dynamic field @timestamp failed
    server.go:191: log file: /Users/simitt/coding/go/elastic/apm-server/systemtest/logs/TestRUMCaptureNoPersonalData/apm-server
FAIL
FAIL	github.com/elastic/apm-server/systemtest	30.525s
```

## Related issues
fixes #7566


beater/api/mux.go
```diff
@@ -305,7 +305,9 @@ func firehoseMiddleware(cfg *config.Config, m map[request.ResultID]*monitoring.I
 }
 
 func emptyRequestMetadata(c *request.Context) model.APMEvent {
-	return model.APMEvent{}
+	return model.APMEvent{
+		Timestamp: c.Timestamp,
```
axw, line 309: Should we have a unit test for this?

axw (commented):
I'm not entirely convinced that the extensible `Dynamic` thing is worthwhile - seems a bit complicated to me, and I'd generally prefer to have tests which make more specific assertions than passing in a callback to ApproveEventDocs.

IMO we should:
 - add a unit test which covers the time sync behaviour, checking with/without capture_personal_data
 - maybe add a system test for capture_personal_data, checking the difference when it is/isn't enabled

beater/api/mux.go
simitt: sounds good; I'll update it

simitt:
> maybe add a system test for capture_personal_data, checking the difference when it is/isn't enabled

Refering to your former comment, I ended up not adding a systemtest. As you indicated, the decision about setting personal data is made purely on the muxer level, and therefore a system test is not necessary. Refactored the muxer handling slightly to make it more easily testable.

beater/api/mux_test.go
```diff
@@ -39,6 +41,38 @@ import (
...
+	event := backendRequestMetadataFunc(cfg)(c)
+	assert.Equal(t, tNow, event.Timestamp)
+	assert.Equal(t, model.Host{}, event.Host)
+
+	c.ClientIP = net.ParseIP("127.0.0.1")
+	event = backendRequestMetadataFunc(cfg)(c)
+	assert.Equal(t, tNow, event.Timestamp)
+	assert.NotEqual(t, model.Host{}, event.Host)
+}
+func TestRUMRequestMetadata(t *testing.T) {
```
axw, line 57:
```suggestion

func TestRUMRequestMetadata(t *testing.T) {
```

axw (approved):
LGTM. Probably not worth it in this PR, but perhaps we should take a middleware style approach where a function is passed in. Then the responsibilities can be separated and composed.

simitt:
> Probably not worth it in this PR, but perhaps we should take a middleware style approach where a function is passed in.

Generally agree on a better handling. This is going to be backported as a patch release to `8.1.2`, so I wanted to be cautious about any refactoring.

stuartnelson3:
confirmed with 2ad82dc69c36aecde281dcbcb3e47361d63298c4

PR #7530, opened 2022-03-14
ci: Daily Releasable Artifacts pipeline

v1v:
## Motivation/summary

Daily Releasable Artifacts pipeline for APM Server only on branches

## Issues

Superseedes https://github.com/elastic/apm-server/pull/7529
Requires https://github.com/elastic/apm-server/pull/7531

## Further details

This new pipeline will be triggered automatically after a build in the CI for the active branches has finished successfully.

1) Docker login to fetch the docker image for the `release-manager`
2) Build artifacts like tarballs, Debian packages, RPMs, zips, Docker images, etc. This needs to happen on at least two different platforms, as AArch64 images must be built on an AArch64 host. These should run in parallel.
3) A following stage that combines them and runs a release-manager container to upload them to GCS.
4) Notification in Slack, and email, if (and only if) the pipeline fails.

For clarity, it won't run on a PR basis, though it's possible to manually kick off a build in -> [here](https://apm-ci.elastic.co/job/apm-server/job/apm-server-package-mbp/) I enabled the support for manual builds for PRs to help with this implementation, potentially to be removed, if so, then https://github.com/elastic/apm-server/blob/main/.ci/jobs/apm-server-package-mbp.yml#L17 should be changed to exclude `PR-*`

## Test

I ran a few iterations within this PR to be able to validate it works as expected:

1) [build-4](https://apm-ci.elastic.co/blue/organizations/jenkins/apm-server%2Fapm-server-package-mbp/detail/PR-7530/4/pipeline) generated all the packages for AMD64 and ARM64. (It failed afterwards for some typo)
2) [build-7](https://apm-ci.elastic.co/blue/organizations/jenkins/apm-server%2Fapm-server-package-mbp/detail/PR-7530/7/pipeline) used the above generated packages and run the `release-manager` successfully. (I wanted to confirm the `DRA` staged worked). https://artifacts-snapshot.elastic.co/apm-server/8.2.0-7c2a7da4/summary-8.2.0-SNAPSHOT.html was generated
3) [build-12](https://apm-ci.elastic.co/blue/organizations/jenkins/apm-server%2Fapm-server-package-mbp/detail/PR-7530/4/pipeline) is the one that will validate all the above steps together.

If it works then I'll revert https://github.com/elastic/apm-server/commit/e461a46a9a6df6f82d42410fccac49adbbbc9daa and ask for the review

PR #7584, opened 2022-03-18
ci: support staging for all branches but main

v1v:
### What

Produce snapshot and releasable artifacts, for such:

* it publishes packages after every commit to all active release branch
* it always builds snapshot images off the main branch
* it always builds staging images off other branches

### Implementation details

- Matrix stage to package for `staging` and `snapshot` 
- DRA stage to :
  - First one to combine all the packages/binaries
  - Second one to run the snapshot release workflow
  - Third one, only other branches but main, to run the staging release workflow

### Issue

Follow up from https://github.com/elastic/apm-server/pull/7530

### Question

The existing implementation uses `release-manager-snapshot` -> https://github.com/elastic/apm-server/blob/27d244ea6b88edb69f38165ab86a7a69892f0676/.ci/packaging.groovy#L93 I guess `staging` should not use this but `release-manager-release`, if so:

1) ~~We need to run twice the packaging for the release branches, one with the `snapshot` and another one with `images`, am I right?~~


axw:
> We need to run twice the packaging for the release branches, one with the snapshot and another one with images, am I right?

I think we just start building staging (BC) images as soon as we've created a release branch at feature freeze time. Can you please check with the unified release team?

axw (commented):
Looks good! IIANM we should also remove SNAPSHOT from environment at the top of packaging.groovy

v1v:
> I think we just start building staging (BC) images as soon as we've created a release branch at feature freeze time. Can you please check with the unified release team?

I'll double check with the release team, though I need to figure out how this particular time event can be discovered within the CI

.ci/packaging.groovy
```diff
@@ -127,17 +109,9 @@ pipeline {
...
-              dockerLogin(secret: env.DOCKER_SECRET, registry: env.DOCKER_REGISTRY)
-              script {
-                getVaultSecret.readSecretWrapper {
-                  sh(label: 'release-manager.sh', script: '.ci/scripts/release-manager.sh')
-                }
-              }
+            releaseManager(type: 'snapshot')
+            whenFalse(env.BRANCH_NAME.equals('main')) {
+              releaseManager(type: 'staging')
             }
```
v1v, line 115: Run always `snapshot` for `main`/`8.1`/`8.2` branches and also `staging` for `8.1`/`8.2` branches

.ci/packaging.groovy
```diff
@@ -154,6 +128,61 @@ pipeline {
...
+      credentialsId: "${JOB_GCS_CREDENTIALS}",
+      pathPrefix: "${BASE_DIR}/build/",
+      pattern: "${BASE_DIR}/build/dependencies.csv",
+      sharedPublicly: true,
+      showInline: true)
+  }
+}
+
+def getBucketLocation(type) {
+  return "gs://${JOB_GCS_BUCKET}/${URI_SUFFIX}/${type}"
```
v1v, line 183: Artifacts are not stored in either the `staging` or `snapshot` subfolders, this is quite handy in order to run the same workflow for different type of releases

.ci/packaging.groovy
```diff
@@ -79,6 +78,10 @@ pipeline {
                 name 'PLATFORM'
                 values 'linux && immutable', 'arm'
               }
+              axis {
+                name 'TYPE'
+                values 'snapshot', 'staging'
```
axw, line 83: is it possible to also exclude staging here when on main?

axw (approved):
Looks great! Just one comment - I think it would be best to avoid running `make release-manager-release` on the main branch, as it's quite an expensive target.

.ci/packaging.groovy
v1v: I tried to use https://github.com/elastic/apm-server/pull/7584/commits/eaa81b1b0ddaeb10462e7832c5f186bd78f11ac4 (the declarative form) but somehow the UI representation skipped one of the stages.

image

So I finally decided to use the scripted approach with https://github.com/elastic/apm-server/pull/7584/commits/e46eb9bac19a6baffdb2feeff223be59062a1b83

image

cachedout (approved):

PR #7586, opened 2022-03-18
dra: retry docker pull

v1v:
### What

`docker.elastic.co` might not be available, so let's rely on the cached docker images by default, but let's try to pull the recent docker image too.

### Why

Reduce the environmental issues

### DIFF
--- a/.ci/scripts/release-manager.sh
+++ b/.ci/scripts/release-manager.sh
@@ -35,7 +35,7 @@ done
 
 # ensure the latest image has been pulled
 IMAGE=docker.elastic.co/infra/release-manager:latest
-(retry 3 docker pull "${IMAGE}") || echo "Error pulling ${IMAGE} Docker image, we continue"
+(retry 3 docker pull --quiet "${IMAGE}") || echo "Error pulling ${IMAGE} Docker image, we continue"
 
 # Generate checksum files and upload to GCS
 docker run --rm \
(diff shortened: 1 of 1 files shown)


.ci/scripts/release-manager.sh
```diff
@@ -33,7 +35,7 @@ done
 
 # ensure the latest image has been pulled
 IMAGE=docker.elastic.co/infra/release-manager:latest
-docker pull --quiet $IMAGE
+(retry 3 docker pull "${IMAGE}") || echo "Error pulling ${IMAGE} Docker image, we continue"
```
v1v, line 38:
```suggestion
(retry 3 docker pull --quiet "${IMAGE}") || echo "Error pulling ${IMAGE} Docker image, we continue"
```

axw (approved):

.ci/scripts/release-manager.sh
```diff
@@ -33,7 +35,7 @@ done
 
 # ensure the latest image has been pulled
 IMAGE=docker.elastic.co/infra/release-manager:latest
-docker pull --quiet $IMAGE
+(retry 3 docker pull --quiet "${IMAGE}") || echo "Error pulling ${IMAGE} Docker image, we continue"
```
cachedout, line 38:
```suggestion
(retry 3 docker pull --quiet "${IMAGE}") || echo "Error pulling ${IMAGE} Docker image, we continue"
docker images --filter=reference=$IMAGE
```

Maybe just list out the image IDs so we know which image we're going to be running in order to help with debugging?

cachedout (approved):

marclop (approved):
🎉

PR #7600, opened 2022-03-21
[modelindexer] collect additional metrics

stuartnelson3:
## Motivation/summary

collect metrics on events processed partitioned by
service name and service version.

## Checklist

## How to test these changes

1. start apm-server
2. index events
3. confirm that metrics exist showing the number of events processed by service.name and service.version


stuartnelson3:
Note: this is the first time I've collected custom metrics, so I'm not sure if I'm doing it correctly or not.

model/modelindexer/indexer.go
```diff
@@ -265,6 +295,19 @@ func (i *Indexer) processEvent(ctx context.Context, event *model.APMEvent) error
 	atomic.AddInt64(&i.eventsAdded, 1)
 	atomic.AddInt64(&i.eventsActive, 1)
 
+	if serviceName := event.Service.Name + "." + event.Service.Version; serviceName != "." {
```
stuartnelson3, line 298: when I originally wrote this I was using the concatenation of these two names below, but stopped doing using this value except as a uniqueness key in the map. So, the `.` is unnecessary now, and can be removed.

stuartnelson3:
Also, I added `Version` as a label because that might be interesting. I don't know if users typically run multiple versions of an exporter, but it might be worth having that granularity.

model/modelindexer/indexer.go
```diff
@@ -265,6 +295,19 @@ func (i *Indexer) processEvent(ctx context.Context, event *model.APMEvent) error
...
+		s, ok := i.services[serviceName]
+		if !ok {
+			labels := []apm.MetricLabel{
+				{Name: "service.name", Value: event.Service.Name},
+				{Name: "service.version", Value: event.Service.Version},
+			}
+			s = service{val: 0, labels: labels}
+		}
+		atomic.AddInt64(&s.val, 1)
+		i.services[serviceName] = s
```
axw, line 308: This appears to be growing indefinitely. What's stopping APM Server from exhausting all memory, due to many unique service names?

axw (changes requested):
What's the motivation behind this change? Is there a related issue?

We should have some protections against memory exhaustion.

I'd like to keep modelindexer focused on just indexing, and its stats should just be related to indexing. For an example of capturing metrics related to events, see https://github.com/elastic/apm-server/blob/main/model/modelprocessor/eventcounter.go

model/modelindexer/indexer.go
stuartnelson3: That's a question I don't really have an answer to. In the same way a user can cause a cardinality explosion, if they're ingesting events from an arbitrarily large number of service name/version pairs, the same will happen here. These values could be cleaned out after some amount of inactivity, or use an LRU cache, but I'm not sure which would be the right option here.

stuartnelson3:
>I'd like to keep modelindexer focused on just indexing, and its stats should just be related to indexing. For an example of capturing metrics related to events, see https://github.com/elastic/apm-server/blob/main/model/modelprocessor/eventcounter.go

sounds good

model/modelindexer/indexer.go
axw: Could we take a similar approach to the metrics aggregations, such as txmetrics? i.e.

- have an upper bound on the service cardinality (say, 10000 unique service name/version pairs)
- report deltas only (the number of events seen for the service in the aggregation period), so we can clear them after reporting

model/modelprocessor/servicecounter_test.go
```diff
@@ -0,0 +1,239 @@
...
+	for i := range batch {
+		service := model.Service{Name: genServiceName()}
+		batch[i] = model.APMEvent{Service: service}
+	}
+
+	err := counter.ProcessBatch(context.Background(), &batch)
+	require.NoError(t, err)
+
+	metrics := gatherMetrics(counter)[1:]
+	assert.Less(t, len(metrics), 10000)
```
stuartnelson3, line 216: it seems like there's some sort of limit when gathering. I was receiving ~5000 metrics each time, even though I verified that internally the ServiceCounter was storing up to 10,000 service name/version pairs.

model/modelprocessor/servicecounter.go
```diff
@@ -0,0 +1,97 @@
...
+					continue
+				}
+				labels := []apm.MetricLabel{
+					{Name: "service.name", Value: event.Service.Name},
+					{Name: "service.version", Value: event.Service.Version},
+				}
+				s = service{val: 0, labels: labels}
+			}
+			atomic.AddInt64(&s.val, 1)
+			c.mu.Lock()
```
axw, line 91: If we always take the write mutex, we may as well not use a read-write mutex. There's a couple of options here:
 1. always use a write lock
 2. change the map to `map[serviceKey]*service`; then you can use `atomic.AddInt64` without having to replace the map entry

I would much prefer the second option, as it will reduce lock contention. This approach should look something like:

```go
c.mu.RLock()
s, ok := c.services[key]
if ok {
    defer c.mu.RUnlock()
} else {
    c.mu.RUnlock()
    c.mu.Lock()
    defer c.mu.Unlock()
    s, ok = c.services[key] // check again, c.services may have been updated between RUnlock/Lock
    if !ok {
        s = &service{val: 0, labels: labels}
    }
}
atomic.AddInt64(&s.val, 1)
```

(might be worth splitting the statements in the loop to a separate function, so defer can be used)

model/modelprocessor/servicecounter_test.go
```diff
@@ -0,0 +1,239 @@
...
+			},
+		},
+	}, metrics)
+}
+
+func TestServiceLimit(t *testing.T) {
+	counter := modelprocessor.NewServiceCounter()
+	batch := make(model.Batch, 20000)
+	for i := range batch {
+		service := model.Service{Name: genServiceName()}
```
axw, line 208: No big deal, but do you see a benefit in randomising the service name vs. say using `fmt.Sprintf("service_%d", i)` as the service name?

axw (commented):
We're in a bit of a hard place wrt metrics with dimensions. Naturally I like the use of the Go agent here, but we don't send these metrics anywhere by default. The other option is stack monitoring, but stack monitoring (at least as implemented in libbeat) does not support arbitrary metric dimensions. We'll need to figure out how these are going to be used, and how and where they'll be stored.

model/modelprocessor/servicecounter_test.go
stuartnelson3: Yeah, that was the issue, thanks!

model/modelprocessor/servicecounter_test.go
stuartnelson3: hah, no i don't, your solution is much simpler

stuartnelson3:
@axw what are the next steps for this PR?

axw:
@stuartnelson3 until we have an answer for how to store these kinds of metrics with Agent, I don't think we can proceed with this change. Discussions are just starting to happen regarding this area of Agent now.

I think we should close and reopen it later, when we have the answer to the above.

axw:
Closing per above comment, please reopen if you disagree

PR #6731, opened 2021-11-29
Introduce new Rally track and tooling

axw:
## Motivation/summary

Remove outdated Rally track.

Introduce our own minimal Python requirements.txt. We install Rally (esrally) on-demand.

Introduce a new Rally track and tooling for generating Rally corpora, comprising:
 - `rally/track.json`, a Rally track that references generated and .gitignored corpora in `rally/corpora`.
 - `script/cat_bulk.py`, which runs apm-server configured to point at an in-memory fake Elasticsearch HTTP server that writes bulk requests out to stdout. Run this script, send events to the server as usual, and the resulting Elasticsearch documents will be sent to stdout.
 - `rally/gencorpora`, a Go program which provides a framework for generating Rally corpora using opentelemetry-go and `script/cat_bulk.py`.

The Rally track does not install templates or ingest pipelines. It is currently assumed that the APM integration package is installed before running Rally races.

Two new Makefile targets are added: `make rally/corpora` (generate Rally corpora), and `make rally` (install esrally and run a race).

## Checklist

~- [ ] Update [CHANGELOG.asciidoc](https://github.com/elastic/apm-server/blob/master/CHANGELOG.asciidoc)~
~- [ ] Update [package changelog.yml](https://github.com/elastic/apm-server/blob/master/apmpackage/apm/changelog.yml) (only if changes to `apmpackage` have been made)~
~- [ ] Documentation has been updated~

## How to test these changes

1. `docker-compose up -d`
2. `make apm-server build-package`
3. Install APM integration package
4. `make rally`

## Related issues

See https://github.com/elastic/apm-server/issues/6115

PR #7624, opened 2022-03-23
model/modelindexer: close gzip writer

axw:
## Motivation/summary

While updating #6731 I saw some errors in the cat_bulk.py Python server:

> EOFError: Compressed file ended before the end-of-stream marker was reached                                                              

This happens because the gzip writer was never closed, only flushed, so there was never an end-of-stream marker written to the Elasticsearch bulk request body. I've fixed that and added a unit test for compression level. While adding the test I noticed that the "BytesTotal" statistic is incorrect when compression is enabled; I will open an issue for that.

## Checklist

## How to test these changes

I don't know what the most realistic way of testing this is. It seems that Elasticsearch doesn't care about the end of stream marker. As above, I found the issue while updating #6731, so it may be enough to just run that.

## Related issues

None

### DIFF
--- a/changelogs/head.asciidoc
+++ b/changelogs/head.asciidoc
@@ -8,6 +8,7 @@ https://github.com/elastic/apm-server/compare/8.1\...main[View commits]
 - APM Server now emits events with `event.duration`, and renames the field to `<event>.duration.us` in an ingest pipeline {pull}7261[7261]
 - Remove `observer.version_major` {pull}7399[7399]
 - APM Server no longer ships with SystemV init scripts or the go-daemon wrapper; use systemd {pull}7576[7576]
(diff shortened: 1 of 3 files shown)


marclop (approved):

axw:
## 💚 All backports created successfully

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|8.1|[](https://github.com/elastic/apm-server/pull/7629)|

Note: Successful backport PRs will be merged automatically after passing CI.

### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sqren/backport)

stuartnelson3:
From speaking with @axw, it's ok to check that events are being successfully indexed into elasticsearch. This has been confirmed to work.

PR #7643, opened 2022-03-24
Update magefile for universal Darwin binaries

axw:
See https://github.com/elastic/beats/pull/29585


axw:
unrelated test failure

PR #7704, opened 2022-03-29
Check ElasticSearch version before writing `_doc_count`

### LINKED ISSUE #7170: Only add _doc_count if supported by ES
In `7.12` the APM Server started adding `_doc_count` field to transaction histogram docs (https://github.com/elastic/apm-server/pull/4647). On Elasticsearch versions <= `7.10` this field is dynamically added to the mapping. When upgrading to e newer ES version, this leads to mapping errors, similar to:
```
  {
    "error" : {
      "root_cause" : [
        {
          "type" : "mapper_parsing_exception",
          "reason" : "Field [_doc_count] is defined more than once"
        }
      ],
      "type" : "mapper_parsing_exception",
      "reason" : "Field [_doc_count] is defined more than once"
    },
    "status" : 400
  }
```

APM  Server should add a safe guard, by adding an ES version check and only writing `_doc_count` fields if already supported by ES.

stuartnelson3:
## Motivation/summary

adding `_doc_count` dynamically to elasticsearch < `7.11.0` causes issues when migrating indices to later versions. check the elasticsearch version before writing this field.

closes #7170

## Checklist


stuartnelson3:
Opening up a draft PR for this now because there are some changes that I want feedback on before continuing, namely the refactoring/moving of code pertaining to creating the elasticsearch output client, and querying the elasticsearch for its metadata (`cluster_uuid` and `version`).

With these changes, apm-server startup does currently block briefly waiting on the response from the elasticsearch (if no elasticsearch is reachable it times out after ~10sec). I made it synchronous for now because that was easier, but there's probably some way to communicate that result to the indexer without blocking if that's desired.

Let me know if this seems sane, or if there's a better way to use the elasticsearch version to write/not write `_doc_count`.

axw:
@stuartnelson3 can we target this to the 7.17 branch? I don't think we need it in 8.0+, as we then have a requirement that Elasticsearch is at least as new as APM Server.

stuartnelson3:
sounds good

stuartnelson3:
@axw targeting 7.17 branch and changed the approach to use a model processor, which seems a bit more manageable

model/apmevent.go
```diff
@@ -80,6 +80,9 @@ type APMEvent struct {
 	Metricset     *Metricset
 	Error         *Error
 	ProfileSample *ProfileSample
+
+	// SetDocCount tracks if it should set _doc_count on metricset
+	SetDocCount bool
```
axw, line 85: The model types should not have any logic in them, otherwise it's going to make it difficult to implement automated encoding (https://github.com/elastic/apm-server/issues/3565 etc.)

Do we need this? I think we can just have the model processor set DocCount to zero. It's only set if it's greater than zero.

model/modelprocessor/index_doc_count_field.go
```diff
@@ -0,0 +1,61 @@
...
+// Lowest version of elasticsearch that supports _doc_count being added by
+// apm-server.
+var elasticsearchSupportsDocCount = common.MustNewVersion("7.11.0")
+
+// ProcessBatch sets whether the APMEvent should include _doc_comment on metricsets.
+func (i *IndexDocCountField) ProcessBatch(ctx context.Context, b *model.Batch) error {
+	i.mu.RLock()
+	if i.version == nil {
+		i.mu.RUnlock()
+		return nil
```
axw, line 53: This means we're assuming the Elasticsearch version is new enough until we know for sure. I think that's probably the right default, but it could do with a comment.

beater/beater.go
```diff
@@ -549,7 +549,8 @@ func (s *serverRunner) run(listener net.Listener) error {
 	}
 	runServer = s.wrapRunServerWithPreprocessors(runServer)
 
-	batchProcessor := make(modelprocessor.Chained, 0, 3)
+	batchProcessor := make(modelprocessor.Chained, 0, 4)
```
axw, line 552:
```suggestion
	var batchProcessor modelprocessor.Chained
```

this is not a hot code path, let's simplify a little

beater/beater.go
```diff
@@ -593,7 +594,7 @@ func (s *serverRunner) run(listener net.Listener) error {
 }
 
 // waitReady waits until the server is ready to index events.
-func (s *serverRunner) waitReady(ctx context.Context, kibanaClient kibana.Client) error {
+func (s *serverRunner) waitReady(ctx context.Context, kibanaClient kibana.Client, indexDocCountField *modelprocessor.IndexDocCountField) error {
```
axw, line 597: Can we separate concerns a little bit here? I don't think "waitReady" should know about model processors. Instead, can we have waitReady return the cluster info?

beater/beater.go
```diff
@@ -549,7 +549,8 @@ func (s *serverRunner) run(listener net.Listener) error {
 	}
 	runServer = s.wrapRunServerWithPreprocessors(runServer)
 
-	batchProcessor := make(modelprocessor.Chained, 0, 3)
+	batchProcessor := make(modelprocessor.Chained, 0, 4)
+	batchProcessor = append(batchProcessor, indexDocCountField)
```
axw, line 553: Should we still be adding this processor when using a non-Elasticsearch output?

model/modelprocessor/index_doc_count_field.go
```diff
@@ -0,0 +1,61 @@
...
+// _doc_comment should be indexed on metricsets within a model.APMEvent.
+type IndexDocCountField struct {
+	mu sync.RWMutex
+	// ElasticSearch cluster version
+	version *common.Version
+}
+
+// SetESClusterVersion sets the elasticsearch cluster version, which is used to
+// decide whether _doc_count should be indexed or not.
+func (i *IndexDocCountField) SetESClusterVersion(version *common.Version) {
```
axw, line 38: I guess we shouldn't spend too much time on this, but I think it might be a bit neater if we had a sort of placeholder BatchProcessor which could be swapped with another one at runtime. Then, rather than updating the ES cluster version on this processor (which should only happen once), we would do something like this in beater:

```go
var esDocCountProcessorPlaceholder PlaceholderBatchProcessor
batchProcessor = append(batchProcessor, &esDocCountProcessorPlaceholder) // no-op until/unless set
...
g.Go(func() error {
	clusterInfo, err := s.waitReady(ctx, kibanaClient, indexDocCountField)
	if err != nil {
		return errors.Wrap(err, "error waiting for server to be ready")
	}
	if clusterInfo.Version != nil && clusterInfo.Version.LessThan(elasticsearchSupportsDocCount) {
		esDocCountProcessorPlaceholder.Set(modelprocessor.UnsetIndexDocCountField{})
	}
	return nil
})
```

axw (commented):
@stuartnelson3 looks good overall, just a few details to iron out.

model/apmevent.go
stuartnelson3: ah, clever. yeah that's a good idea

beater/beater.go
stuartnelson3: ah, fair point. I've actually never had to check for that -- how do I best check for that? check to see if `serverRunner.elasticsearchOutputConfig == nil`, and then assume the output is not elasticsearch?

model/modelprocessor/index_doc_count_field.go
stuartnelson3: no worries, added something

stuartnelson3:
@axw do you have a recommendation on how to test if `_doc_count` is being unset if elasticsearch <= 7.11.0? Is there a way to set the elasticsearch version in a systemtest?

axw:
I think we can unit test this bit rather than a full system test. See `beater.TestServerWaitForIntegrationElasticsearch` for an example of setting up beater with a mock Elasticsearch.

PR #7723, opened 2022-03-30
dra: add who to contact with if a DRA or package failure

v1v:
### What

DRA failures are different than package failures, therefore, let's add who to contact with if further analysis is required.

### Why

DRA is a black box so only the system owners know about it, therefore the product teams need to care only for the package generation in case something fails, otherwise, redirect the failures to the system owners.

### Question

Do we want to notify to the system owners by email/slack?

### DIFF
--- a/.ci/packaging.groovy
+++ b/.ci/packaging.groovy
@@ -109,7 +109,8 @@ pipeline {
           }
           post {
             failure {
-              notifyStatus(subject: "[${env.REPO}@${env.BRANCH_NAME}] package failed")
+              notifyStatus(subject: "[${env.REPO}@${env.BRANCH_NAME}] package failed.",
+                           body: 'Contact the Productivity team if you need further assistance')
             }
           }
         }
@@ -134,7 +135,8 @@ pipeline {
             failure {
               notifyStatus(analyse: true,
                            file: "${BASE_DIR}/${env.DRA_OUTPUT}",
-                           subject: "[${env.REPO}@${env.BRANCH_NAME}] DRA failed")
+                           subject: "[${env.REPO}@${env.BRANCH_NAME}] DRA failed.",
+                           body: 'Contact the Release Platform team')
             }
           }
         }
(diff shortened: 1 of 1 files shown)


mdelapenya (approved):

cachedout (approved):

.ci/packaging.groovy
```diff
@@ -109,7 +109,8 @@ pipeline {
           }
           post {
             failure {
-              notifyStatus(subject: "[${env.REPO}@${env.BRANCH_NAME}] package failed")
+              notifyStatus(subject: "[${env.REPO}@${env.BRANCH_NAME}] package failed.",
+                           body: 'Contact the Productivity team if you need further assistance')
```
marclop, line 113: Should we add channel names?

.ci/packaging.groovy
```diff
@@ -134,7 +135,8 @@ pipeline {
             failure {
               notifyStatus(analyse: true,
                            file: "${BASE_DIR}/${env.DRA_OUTPUT}",
-                           subject: "[${env.REPO}@${env.BRANCH_NAME}] DRA failed")
+                           subject: "[${env.REPO}@${env.BRANCH_NAME}] DRA failed.",
+                           body: 'Contact the Release Platform team')
```
marclop, line 139: Same?

marclop (approved):
LGTM. Just thinking whether or not we could add the channel names for easier guided contact points.

.ci/packaging.groovy
v1v: notifyStatus reports by email and slack, so do you prefer the slack channels?

marclop: ah I missed that. Maybe we could include both? but don't let my review block the merging of this PR.

v1v: Let me work on this

v1v: I just added the slack channel -> https://github.com/elastic/apm-server/pull/7723/commits/c8fa511919e3c3dc5c5874bb2b84cfc8b477af60

Let's try with the slack rather than the email to avoid exposing further details, what do you think?

marclop: Sounds good, thank you!

PR #7736, opened 2022-03-30
Add available bulk indexers metric

### LINKED ISSUE #7728: modelindexer: Expose the number of available bulkindexers
## Description

We currently don't have any visibility into what the number of available bulk indexers is and would like to change that. 

I'm unsure what the exact metric name  we should use considering that if we are to also add visualizations for the metric in Stack Monitoring they should work both for the libbeat output and our own model indexer.

## Beat metrics reference

I saw that there's a `libbeat.pipeline.clients` but I am unsure if we can reuse the name for this purpose.

- https://www.elastic.co/guide/en/beats/metricbeat/current/metricbeat-metricset-beat-stats.html

stuartnelson3:
## Motivation/summary

Expose the number of available bulk indexers, to see how the indexer is performing.

## Checklist

## How to test these changes

- start apm-server with `http.enabled: true`.
- curl the http endpoint and check the metric exists:
```
# curl -s localhost:5066/stats | jq .output
{
  "elasticsearch": {
    "bulk_requests": {
      "available": 10,
      "completed": 0
    }
  }
}
 ```
- ingest some events, check that the values change

## Related issues

closes #7728

### CODE REVIEW

File: beater/beater.go [hunk 1]
```diff
@@ -730,6 +730,13 @@ func (s *serverRunner) newFinalBatchProcessor(
 		v.OnKey("total")
 		v.OnInt(stats.Added)
 	})
+	// TODO: Where do we want to put this metric?
+	monitoring.NewFunc(monitoring.Default, "libbeat.output", func(_ monitoring.Mode, v monitoring.Visitor) {
```
marclop, line 734: I'm not sure this will behave as expected.

I had some problems adding metrics when the names of the metrics "clashed" with each other. For example, on line 716, we're adding `libbeat.output.events`, which creates the `libbeat.output` parent. Then on this line, we're adding `libbeat.output` again, which normally wipes the contents of what was previously in `libbeat.output` (`libbeat.output.events....`).

Could you extend the existing test `TestServerElasticsearchOutput` to assert the output of the registry and that it behaves as we expect?

model/modelindexer/indexer.go
```diff
@@ -237,6 +239,8 @@ func (i *Indexer) processEvent(ctx context.Context, event *model.APMEvent) error
 		case <-ctx.Done():
 			return ctx.Err()
 		case i.active = <-i.available:
+			// i.mu.RLock currently held
```
marclop, line 242: was this comment meant to be left here?

model/modelindexer/indexer.go
```diff
@@ -340,6 +344,8 @@ func (i *Indexer) flushActive(ctx context.Context) error {
 	err := i.flush(ctx, bulkIndexer)
 	bulkIndexer.Reset()
 	i.available <- bulkIndexer
+	// i.mu.RLock currently held
```
marclop, line 347: did you mean to leave this comment?

marclop (commented):
Nice find on the unexpected behavior.

model/modelindexer/indexer.go
stuartnelson3: I did, but if you think it's unnecessary, happy to remove

marclop: I was confused because we're just doing an atomic operation on the pointer of that variable, so I was curious to know why that's pointed out. We're not taking the read lock when the `.Stats()` method is called, so I'm curious why you added the comment.

beater/beater.go
```diff
@@ -713,6 +713,8 @@ func (s *serverRunner) newFinalBatchProcessor(
 	})
 	outputType := monitoring.NewString(monitoring.Default.GetRegistry("libbeat.output"), "type")
 	outputType.Set("elasticsearch")
+	availableBulkIndexers := monitoring.NewInt(monitoring.Default.GetRegistry("libbeat.output"), "availableBulkIndexers")
+	availableBulkIndexers.Set(indexer.Stats().AvailableBulkIndexers)
```
marclop, line 717: This will only set the value of the variable once but won't update its value as it changes over time. That's the reason I used the variable var above.

stuartnelson3: whoops!

axw:
Good catch on the race. There may be a better way, but this fixes it:

```patch
diff --git a/model/modelindexer/indexer.go b/model/modelindexer/indexer.go
index 1e4300093..cd9a67be5 100644
--- a/model/modelindexer/indexer.go
+++ b/model/modelindexer/indexer.go
@@ -86,6 +86,7 @@ type Indexer struct {
        closing      bool
        closed       chan struct{}
        activeMu     sync.Mutex
+       activeCond   *sync.Cond
        active       *bulkIndexer
        timer        *time.Timer
        timerStopped chan struct{}
@@ -144,14 +145,16 @@ func New(client elasticsearch.Client, cfg Config) (*Indexer, error) {
        for i := 0; i < cfg.MaxRequests; i++ {
                available <- newBulkIndexer(client, cfg.CompressionLevel)
        }
-       return &Indexer{
+       indexer := &Indexer{
                availableBulkIndexers: int64(len(available)),
                config:                cfg,
                logger:                logger,
                available:             available,
                closed:                make(chan struct{}),
                timerStopped:          make(chan struct{}),
-       }, nil
+       }
+       indexer.activeCond = sync.NewCond(&indexer.activeMu)
+       return indexer, nil
 }
 
 // Close closes the indexer, first flushing any queued events.
@@ -235,6 +238,11 @@ func (i *Indexer) processEvent(ctx context.Context, event *model.APMEvent) error
 
        i.activeMu.Lock()
        defer i.activeMu.Unlock()
+       for i.active != nil && i.active.Len() >= i.config.FlushBytes {
+               // The active bulk indexer is full: wait for it to be
+               // switched out by the background flushActive goroutine.
+               i.activeCond.Wait()
+       }
        if i.active == nil {
                select {
                case <-ctx.Done():
@@ -340,6 +348,7 @@ func (i *Indexer) flushActive(ctx context.Context) error {
        i.activeMu.Lock()
        bulkIndexer := i.active
        i.active = nil
+       i.activeCond.Broadcast()
        i.activeMu.Unlock()
        err := i.flush(ctx, bulkIndexer)
        bulkIndexer.Reset()
diff --git a/model/modelindexer/indexer_test.go b/model/modelindexer/indexer_test.go
index 2586cbefa..4739d45c5 100644
--- a/model/modelindexer/indexer_test.go
+++ b/model/modelindexer/indexer_test.go
@@ -136,7 +136,7 @@ func TestModelIndexerAvailableBulkIndexers(t *testing.T) {
                // TODO: There's a race between hitting the limit `FlushBytes`
                // and switching to the new active bulk indexer, hence the
                // sleep between batches to allow this swap.
-               time.Sleep(50 * time.Millisecond)
+               //time.Sleep(50 * time.Millisecond)
                err := indexer.ProcessBatch(context.Background(), &batch)
                require.NoError(t, err)
        }
```

model/modelindexer/indexer.go
```diff
@@ -473,4 +478,8 @@ type Stats struct {
 
 	// BytesTotal represents the total number of bytes that
 	BytesTotal int64
+
+	// AvailableBulkIndexers represents the number of bulk indexers
+	// available for making bulk index requests.
+	AvailableBulkIndexers int64
```
axw, line 484: "Available bulk indexers" seems like an internal detail of the implementation which I'd rather not expose.

The configuration is "MaxRequests", and there's an existing metric which counts the total number of bulk requests over time. Maybe we should follow that naming? Perhaps "AvailableBulkRequests" (struct field) and "available_bulk_requests" (Elasticsearch field)?

axw (changes requested):
Looks good overall! Just some minor comments.

beater/beater.go
```diff
@@ -713,6 +713,15 @@ func (s *serverRunner) newFinalBatchProcessor(
...
 	outputType.Set("elasticsearch")
+	monitoring.NewFunc(monitoring.Default, "libbeat.output.bulk_requests", func(_ monitoring.Mode, v monitoring.Visitor) {
+		v.OnRegistryStart()
+		defer v.OnRegistryFinished()
+		stats := indexer.Stats()
+		v.OnKey("available")
+		v.OnInt(stats.AvailableBulkRequests)
+		v.OnKey("completed")
+		v.OnInt(stats.BulkRequests)
+	})
```
stuartnelson3, line 724: I couldn't figure out how to add `libbeat.output.bulk_requests_available`, so I nested `available` in a `bulk_requests` object, thinking that in the future we might want to add more metrics to that. I added `completed` because `available` seemed kind of lonely, but it's a duplicate of `events.batches`, so i'm also happy to remove it.

axw: Long term I'd like to stop naming the metrics we have `libbeat.output.*`. I did that initially so the metrics would show up in Stack Monitoring and friends.

Seeing as this metric doesn't exist in libbeat at all, let's not add that prefix please. Maybe `output.elasticsearch.bulk_requests.*`?

axw (commented):
LGTM with the exception of the metric name

marclop (commented):
Looks good. Have you tested if the metric beat `beat` module ingests the documents with the added metrics?
I'm not sure if it will ignore the ones it doesn't know about or if it will collect everything try to index it and fail to since a mapping is missing for that new object and fields.

stuartnelson3:
I'll check it out, that would definitely be good to know

stuartnelson3:
Ok, confirmed that it's just dropped when using the `beat` module, but the events are still being indexed.

marclop (approved):
🎉

PR #7737, opened 2022-03-30
mod: Update go.elastic.co/apm & modules to v2.0.0

marclop:
## Motivation/summary

Updates the `go.elastic.co/apm` dependency to the more up-to-date 2.0.0
and updates the required code paths.

## Checklist

~- [ ] Update [CHANGELOG.asciidoc](https://github.com/elastic/apm-server/blob/main/CHANGELOG.asciidoc)~ not user facing

## How to test these changes

N/A

## Related issues

Depends on elastic/beats#31072


axw:
I haven't yet gotten to the bottom of it, but the system test failures are due to the agent request being cancelled (presumably the agent closed its connection) here: https://github.com/elastic/apm-server/blob/03db2ca860b467822962d34f753a9fed39db7673/systemtest/instrumentation_test.go#L118

axw:
There seems to be a ~10 second delay between libbeat creating the `apm.Tracer`, and the apm-server "beater" code being called. When a Tracer is created it now sends a "GET /" request when its loop starts up. We block all requests until apm-server has started, when we know its listening address, so we can reverse proxy. `net/http/Server` sets a default read timeout of 10 seconds, so the "GET /" request times out before we know the listening address.

I'll update the test code. We should also look into what libbeat is doing in those 10 seconds.

axw:
> We should also look into what libbeat is doing in those 10 seconds.

I wanted to check that we didn't introduce the delay in the agent. Unfortunately we did: https://github.com/elastic/apm-agent-go/issues/1238

axw (approved):

marclop:
Thank you Andrew!

PR #7792, opened 2022-04-05
Remove release tag from APM package

jsoriano:
## Motivation/summary

`release` tag has been deprecated https://github.com/elastic/package-spec/issues/225 in favour of semantic versioning.

It can still be used, but incoherent uses between it and prerelease tags in the version are discouraged. The package registry provides a compatibility layer so packages without `release` continue working with old versions of Fleet. In the API, this tag is filled automatically when not provided.

The APM package is making incoherent use of the `release` tag and its version, `release` is always ga, also in prerelease versions.

## Checklist

## How to test these changes

- APM package should continue working.
- APM package should have a prerelease badge in Fleet UI for prerelease versions.

## Related issues

- Related to https://github.com/elastic/package-spec/issues/225

### DIFF
--- a/apmpackage/apm/manifest.yml
+++ b/apmpackage/apm/manifest.yml
@@ -6,7 +6,6 @@ license: basic
 description: Ingest APM data
 type: integration
 categories: ["elastic_stack", "monitoring"]
-release: ga
 conditions:
   kibana.version: generated
 icons:


jsoriano:
@simitt @axw I am finding problems updating `elastic-package` here. Do you have an idea of what could be happening?
I think this may be related to the specific `go.mod` in the `tools` directory. If I move the entries there to the root-level `go.mod`, `elastic-package` works.

axw:
@jsoriano seems to be a module dependency conflict in tools. That needs to be fixed - we shouldn't have tools in the top level go.mod.

PR #7794, opened 2022-04-05
Fix typo in {span,transaction}.service.origin.id

SylvainJuge:
## Motivation/summary

While working on span/transaction resource destination, I noticed that we already have a few service fields in `span.service.origin`, but there is a typo on one of them, this PR attempts to fix that.

This is already part of the protocol and copied to agent repositories, for example in the Java agent repository: https://github.com/elastic/apm-agent-java/blob/e2ef2453c8a560922dfaabfaf9f9dbd0d3032337/apm-agent-core/src/test/resources/apm-server-schema/current/transaction.json#L596-L602.

This was opened as a draft PR because I'm not very familiar with all the implications of this change, for example changing this will trigger follow-up PRs on every agent repository.

## Checklist

- [ ] Check the actual field usage in APM server
  - [ ]  if not used, then changing this has no effect beyond PRs on all agents (which would not be expected to use it anyway)
  - [ ] if used by APM server, then further testing might be required to avoid any regression.


simitt:
Uh that's a good catch, thanks for opening the PR. You will need to run `make update` which should regenerate the spec and then also run `make test` which should give you more instructions how to update the approved tests.

simitt (approved):
Thanks for your contribution! This is good to merge; I assume agents have not actually sent a `string` field, but the `id` field probably was just ignored. Nevertheless, do you mind mentioning the fix to the other agent devs?

PR #7809, opened 2022-04-06
stream/processor: Limit concurrent decoders to 200

### LINKED ISSUE #7795: Don't run out of memory when the ingest is exponentially higher than the output
**APM Server version** (`apm-server version`): `8.0.0+`

APM Server is now prone to running out of memory when the rate of processed events greatly exceeds the output rate (Elasticsearch indexing). The flow looks like this:

* Server receives a batch from an APM Agent
* Server decodes and validates the first 10 events (batchSize), sends them off to the modelprocessor and eventually in the `modelindexer.ProcessBatch()`, which blocks the batch ingestion until the batch is added to the underlying `bulkIndexer`'s cache.
* The request's context will be cancelled if the whole batch is not read before the server's `ReadTimeout`.
* As long as as the request's ReadTimeout isn't hit, the server will continue decoding events in the batch, 10 at a time.
* The Server's memory will increase until all the memory is consumed, when there are enough APM Agents to trigger this behavior.

marclop:
## Motivation/summary

Adds a new configurable (for now) setting which allows us to limit the
maximum number of requests that are being decoded at any time. This
effectively limits the maximum memory consumption of the intake layer
of the server to `300KB * 10 events * 200 concurrent = 586MB`, of course
the actual maximum heap representation may be more, so adding a 20% of
overhead in this calculation still yields the ~704MB of memory usage for
the decoders, which is the maximum they will ever use. This plus a ~50MB
for the caches of all the bulk indexer (5MB * 10 bulk indexer) is still
below the 1GB mark. Even if Fleet and Elastic Agent were to use 100MB
we're still ~860 MB of memory used, which is below the smallest cgroup
memory limit size in ESS.

I've done some preliminary benchmarking and profiling on this change and
it seems to improve the lock contention on the `indexer.activeMu` since
the number of goroutines trying to acquire or release the lock is lower.

The indexing throughput didn't seem to be affected and the ingested
events/s throughput seems to be the same as it was before, except the
memory consumption is much more contained, only having observed ~750MB
of total RSS consumption. `8.1.2` went all the way up to ~1100MB RSS and
seemed to have slightly lower throughput (most likely due to the extra
contention that could be explained by a higher number of goroutines
trying to acquire the indexer.activeMu).

As it can be observed below, benchmarked with 550 agents and 1200, the
performance was much worse on `8.1.2` and the memory consumption also
much higher (1800-2200MB).

```
BenchmarkAgentGo-550     7346 events/sec  1827118 B/op  5236 allocs/op
BenchmarkAgentNodeJS-550 5452 events/sec  2605381 B/op  4225 allocs/op
BenchmarkAgentPython-550 5529 events/sec  9050536 B/op  22003 allocs/op
BenchmarkAgentRuby-550   6528 events/sec  2887096 B/op	6116 allocs/op
```

Tested with the APM Server standalone running on a `c5d.large` instance,
and a backing 2 node 8GB Elasticsearch cluster running in ESS in the
same region.

## Checklist

## How to test these changes

* Run `apmbench` with a medium to high number of `-agents` 300-800 agents should be enough.
* Verify that the `apm-server`'s memory doesn't exceed 800MB.

## Related issues

Closes #7795 

#### 8.1.2

8.1.2-goroutine

### Patch

patch-goroutine

#### 8.1.2

8.1.2-heap

### Patch

patch-heap


processor/stream/processor.go
```diff
@@ -50,30 +50,35 @@ const (
...
 type Processor struct {
-	MaxEventSize     int
 	streamReaderPool sync.Pool
 	decodeMetadata   decodeMetadataFunc
+	sem              chan struct{}
+	MaxEventSize     int
 }
 
-func BackendProcessor(cfg *config.Config) *Processor {
+func BackendProcessor(cfg *config.Config, sem chan struct{}) *Processor {
```
axw, line 61: Should we create the channel internally using the setting in `cfg`, rather than passing one in?

axw (commented):
This looks great, thanks.

I think the default is good enough for now, but what would be great is if we could calculate the limit from the output's configuration, so we automatically scale but also never produce more than the output can handle (:hattip: to @Mpdreamz for the idea).

e.g. if we know `max_requests * flush_bytes` (50MB by default), then we can calculate an upper bound on max in-flight events as `(max_requests * flush_bytes) / max_event_size` (`50MB/300KB=170` by default).

We don't necessarily want them to be equal, just (obviously) not exponentially more. It may be especially useful to allow some multiple of the output limit when using processors that drop events (TBS, drop-unsampled) before they reach the output.

processor/stream/processor.go
marclop: First, I created a semaphore for each of the backend processors, but that isn't a great solution because each of the stream processors (Backend, RUM, RUMV3) would have its own limit, then the calculation doesn't hold anymore or we need to divide it by 3 which severely would limit the upper bound of concurrent decodes that a single backend processor performs.

Instead, if we choose to have an "intake semaphore" then the limits are shared between all the ingesting processors.

axw: Right, makes sense.

beater/config/config.go
```diff
@@ -77,6 +77,11 @@ type Config struct {
 	// the integration package to be installed, and for checking the
 	// Elasticsearch license level.
 	WaitReadyInterval time.Duration `config:"wait_ready_interval"`
+
+	// MaxConcurrentDecoders sets the limit on the number of concurrent batches
+	// that can be decoded at one time. This effectively limits the amount of
+	// memory consumed by the processors decodeing the incoming intake events.
+	MaxConcurrentDecoders uint `config:"max_concurrent_decoders"`
```
simitt, line 84: We should label this as `beta`.

simitt (commented):
I agree with Andrew that we should derive the semaphore size; but I would not aim for it in this PR to allow us getting a fix out quickly. 
My suggestion is to move forward with this implementation, using a config option as escape hetch, but clearly labeling it `beta` with the expressed intention of removing the config option in an early upcoming version. 

~@marclop can you confirm that the memory analysis only considers one semaphore? If RUM is enabled, we need to expect a higher memory consumption because the limit would affectively be doubled.~
This is not true any more with this implemention. Nice!

beater/config/config.go
marclop: Do we want to document it in `asciidoc`? My guess is no, but just checking

simitt: No, let's not document at all for now.

axw (approved):

PR #7860, opened 2022-04-12
[7.17] 7.17.3 release notes

bmorelli25:
Adds 7.17.3 release notes based on https://github.com/elastic/apm-server/commits/7.17.
There are a few PRs that made it into this version, but none include changelog entries.


axw (approved):

axw:
@bmorelli25 we should probably mention https://github.com/elastic/apm-server/pull/7704, which is missing a changelog entry.

If APM Server 7.17.3 is used with an old (<7.11.0) version of Elasticsearch, it will no longer set `_doc_count` fields. This metadata field was added in Elasticsearch 7.12.0; setting it in earlier versions causes problems on upgrade.

Issue #7752, opened 2022-03-31
docs: Add trouble shooting section to APM Guide


simitt:
The new APM guide is not offering a troubleshooting section common problems anymore. While some of the problems are not applicable to the APM Integration, the majority of the section still has value. We should add a revised version of the [Common Problems](https://www.elastic.co/guide/en/apm/guide/current/common-problems.html): 

* remove the parts that directly run commands via the apm-server CLI, for example `apm-server test config`.
* remove the `queue is full` section, this doesn't happen any more
* reference the [Input APM](https://www.elastic.co/guide/en/apm/guide/current/input-apm.html) section for details on how to configure the APM Server
* keep suggestions to check the APM agent & server logs
* keep section around any [TLS failures](https://www.elastic.co/guide/en/apm/guide/current/common-problems.html#ssl-client-fails) and [common ssl problems](https://www.elastic.co/guide/en/apm/guide/current/common-problems.html#_common_ssl_related_errors_and_resolutions)
* remove the section around [field limit exceeded](https://www.elastic.co/guide/en/apm/guide/current/common-problems.html#field-limit-exceeded) - should not happen anymore; if it does we should be informed about it rather than provide a guide for an unexpected edge case
* keep [i/o timeout](https://www.elastic.co/guide/en/apm/guide/current/common-problems.html#io-timeout)
* change the [If Elasticsearch is down](https://www.elastic.co/guide/en/apm/guide/current/common-problems.html#server-es-down) section to something like the APM Server will eventually deny incoming requests and the Server and the APM agent should be issuing logs accordingly.

marclop:
I opened https://github.com/elastic/apm-server/issues/7692 and they seem to have some overlap, but your issue is more complete than the one I wrote. Should I close the one I opened?

simitt:
It's not quite the same, although it has some overlap. Let's leave both open for now and @bmorelli25 can decide whether to adress them at once or in two separate PRs.

Issue #7692, opened 2022-03-29
Update docs to reflect new model indexer behavior


marclop:
## Description

There are a few mentions of "queue is full" or that the APM Server buffers events into an internal queue to account for network or Elasticsearch unavailability. We should update these mentions to more accurately reflect the behavior of the new model indexer and how it doesn't have an in internal queue anymore.

## References

There may be more, but I was able to find these in the "Legacy APM Reference":

- https://www.elastic.co/guide/en/apm/guide/8.1/high-availability.html
- https://www.elastic.co/guide/en/apm/guide/8.1/common-problems.html

We should probably make a new section on the new APM Guide for the common issues that could be encountered by customers rather than only having these present in the Legacy APM Reference.

Issue #6678, opened 2021-11-20
[docs] Add APM traces indexed by no data in APM UI troublshooting guide


graphaelli:
Inspired most recently by https://discuss.elastic.co/t/apm-agent-logs-shown-on-discover-but-not-on-apm-ui/289706, a [common problems](https://www.elastic.co/guide/en/apm/server/7.15/common-problems.html) entry for data seems to be indexed but the APM UI is not displaying any data would likely be appreciated by users.  While this is resolved with the move to data streams, they are not the default in versions that we will be supporting for a long time, so a write up seems well worth it.

PR #7901, opened 2022-04-18
[docs] New APM troubleshooting / Update legacy version

bmorelli25:
### Summary

This PR addresses three open issues relating to APM troubleshooting:

Per https://github.com/elastic/apm-server/issues/7752:

- [x] Removes apm-server CLI commands, queue is full, and field limit exceeded sections
- [x] Updates configuration links to point to the APM integration input docs
- [x] Keeps and updates TLS, i/o timeout, and ES is down sections

Per https://github.com/elastic/apm-server/issues/6678:

- [x] Adds legacy troubleshooting help for users with incorrectly setup index mappings 

Per https://github.com/elastic/apm-server/issues/7692:

- [x] Updates troubleshooting to reflect the modelindexer output
- [x] Updates a handful of other _queue_ mentions

### Related

* Closes https://github.com/elastic/apm-server/issues/7752.
* Closes https://github.com/elastic/apm-server/issues/6678.
* Closes https://github.com/elastic/apm-server/issues/7692.


docs/legacy/common-problems.asciidoc
```diff
@@ -325,9 +298,10 @@ APM agent --> Load Balancer  --> APM Server
 
 *If Elasticsearch is down*
 
-If Elasticsearch goes down, APM will keep some data in its internal in-memory queue.
-When the queue fills up, APM will respond with <<queue-full-legacy,`HTTP 503: Queue is full`>> and
-block requests until there is space available or until the request timeout is reached.
+APM Server does not have an internal queue to buffer requests,
+but instead leverages an HTTP request timeout to act as back-pressure.
+If Elasticsearch goes down, the APM Server will eventually deny incoming requests.
+Both the APM Server and APM agent(s) will issue logs accordingly.
```
bmorelli25, line 304: 👋 @marclop Is this accurate? Is there a better way to describe the new model indexer behavior? (I took some of your words from https://github.com/elastic/sdh-apm/issues/542#issuecomment-1077818731)

I also used this wording to replace other "queue is full" docs—please see https://github.com/elastic/apm-server/pull/7901/commits/973a63e45736ee16f09b388e888e83de189f4bc0.

docs/common-problems.asciidoc
```diff
@@ -0,0 +1,222 @@
...
+[source,logs]
+----
+[ElasticAPM] APM Server responded with an error:
+"read tcp 123.34.22.313:8200->123.34.22.40:41602: i/o timeout"
+----
+
+To fix this, ensure timeouts are incrementing from the APM agent,
+through your load balancer, to the APM Server.
+
+By default, the agent timeouts are set at 10 seconds, and the server timeout is set at 3600 seconds.
```
bmorelli25, line 179: Did I get this timeout default correct? It's very different from the legacy default of 30s, so I want to make sure I'm not looking at the wrong configuration variable.

docs/api-events.asciidoc
```diff
@@ -88,7 +88,7 @@ An example error response might look something like this:
       "message": "<json-decoding-err>",
       "document": "<ndjson-obj>"
     },{
-      "message": "queue is full" <3>
+      "message": "too many requests" <3>
```
bmorelli25, line 91: Is this a good replacement example for an immediately returning non-event-related error?

marclop: I think they'll now see "request timed out" but I need to double check the exact message.

docs/common-problems.asciidoc
```diff
@@ -0,0 +1,222 @@
...
+[[request-timed-out]]
+[float]
+==== HTTP 503: Request timed out waiting to be processed
+
+This happens when APM Server exceeds the maximum number of requests that it can process concurrently.
+To alleviate this problem, you can try to: reduce the sample rate and/or reduce the collected stack trace information.
+See <<reduce-apm-storage>> for more information.
+
+Another option is to increase processing power.
+This can be done by either migrating your {agent} to a more powerful machine
```
marclop, line 94: @elastic/apm-server This area doesn't mention that it may be because Elasticsearch is undersized or that the number of primary shards for the data streams is the default `1`. Should we mention these on this particular error?

It is something that our customers will see even if they increase processing power and add more APM Servers to the deployment, it may actually make things worse since the number of concurrent index requests will go up and only so many will be processed by Elasticsearch.

docs/common-problems.asciidoc
marclop: @elastic/apm-server seems like Brandon is correct, the default value for the read timeout on the APM package is `3600s` not `30s`: https://github.com/elastic/apm-server/blob/6a3bbb6a72ceaa8e743a09a798f797add29d06c8/apmpackage/apm/manifest.yml#L109-L111

Shouldn't that be set to `30s`? This may actually be something that's causing a lot of the requests that are sent to the APM Server to be queued for longer than necessary.

marclop (commented):
Left some questions for the rest of the team to discuss / answer.

docs/common-problems.asciidoc
bmorelli25: Resolving as this is correct per Slack

simitt (approved):
Left a couple of minor comments; apart from that LGTM!

docs/common-problems.asciidoc
```diff
@@ -0,0 +1,222 @@
...
+
+APM Server does not have an internal queue to buffer requests,
+but instead leverages an HTTP request timeout to act as back-pressure.
+If Elasticsearch goes down, the APM Server will eventually deny incoming requests.
+Both the APM Server and APM agent(s) will issue logs accordingly.
+
+*If APM Server is down*
+
+Some agents have internal queues or buffers that will temporarily store data if the APM Server goes down.
+As a general rule of thumb, queues fill up quickly. Assume data will be lost if APM Server goes down.
```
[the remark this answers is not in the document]
bmorelli25: Thanks. I changed up the wording to make this more clear.

docs/api-events.asciidoc
bmorelli25: Is there a better example then? I'm not sure how to move forward here.

bmorelli25: I'm going to merge and we can address any follow-ups in a new PR.

simitt: Apologies, I missed your comment regarding an example. In this case the APM Agent will just log a timeout because the APM Server doesn't respond in time. So there is no well defined response in this case. I think it would be best to describe the scenario this way (server not responding in time, logging `request timed out`).

Issue #7920, opened 2022-04-20
OTLP: root spans that represent outgoing calls don't count into destination metrics


felixbarny:
Scenario: use the OTEL SDK to create a span for an HTTP client call which is not enclosed within the context of an incoming request. Send this span to APM Server via OTLP.

Expected outcome: destination metrics are created

Actual outcome: as these OTEL spans are modeled as transactions, `processSpans` will never get called. Also, these events will not have a `DestinationService.Resource`.

https://github.com/elastic/apm-server/blob/51c999c3365099a4385036fedd50286830561af9/x-pack/apm-server/aggregation/spanmetrics/aggregator.go#L210-L213

The same issue happens when using OTEL bridge in APM Agents and creating a span that represents an outgoing HTTP request which doesn't have a parent.

A related issue is that these events will not be considered in the service map.

This has an impact on our iOS agent which creates root spans for outgoing HTTP requests if there's no parent span.

Kudos to @eyalkoren for identifying this as a potential issue.

cc @akhileshpok for potential impact on the iOS agent
cc @axw as you implemented both OTLP intake and span destination metrics

PR #7924, opened 2022-04-21
Add support for target service in metric aggregation

### LINKED ISSUE #7745: Backend dependencies granularity: Add `transaction.dropped_spans_stats.service_target_*`
### Intake API
Extend transaction Intake API to accept `transaction.dropped_spans_stats.service_target_type` and `transaction.dropped_spans_stats.service_target_name`.

### Aggregated Metrics Calculation
Extend [`transaction.DroppedSpanStats`](https://github.com/elastic/apm-server/blob/main/model/transaction.go#L149-L153) to contain the `service_target_*` fields. For calculating related span metrics, add the new fields to the [`aggregationKey`](https://github.com/elastic/apm-server/blob/main/model/transaction.go#L149-L153) to consider the new fields in the aggregation. 

### ES data storage
None of the dropped spans stats should be stored in Elasticsearch, the information is only used for calculating metrics.The [`DroppedSpansStatsDiscarder`](https://github.com/elastic/apm-server/blob/main/model/modelprocessor/dropped_spans_stats.go) already takes care of removing the information before sending data to Elasticsearch. No changes required for this part.

lahsivjar:
## Motivation/summary

- `service.target.{type, name}` should be indexed to ES for internal_metrics
- Metric aggregation should offer more granularity by utilising `service.target.{type, name}` fields
- Transaction intake API should accept `service_target_{type, name}` as input for working with dropped span stats 

Full spec list available at elastic/apm#627

## Checklist

## How to test these changes

1. Send API request to transaction intake API with `service_target_{type, name}` fields for the dropped span stats
2. Send API request to intake API with `resource` and `service.target.*` fields
3. Send API request to intake API with `resource` field and without `service.target.*` fields
4. Observe the output metrics to contain the appropriate `service.target.*` fields

## Related issues

Closes #7745


stuartnelson3 (approved):

SylvainJuge (approved):
LGTM, did some manual testing with modified Java agent and new fields are available as expected :tada: .

x-pack/apm-server/aggregation/spanmetrics/aggregator_test.go
```diff
@@ -76,28 +76,36 @@ func TestAggregatorRun(t *testing.T) {
...
+		{serviceName: "service-A", agentName: "java", destination: destinationX, targetType: trgTypeX, targetName: trgNameX, outcome: "success", count: 1},
+		{serviceName: "service-B", agentName: "python", destination: destinationZ, targetType: trgTypeZ, targetName: trgNameZ, outcome: "success", count: 1},
+		{serviceName: "service-A", agentName: "java", destination: destinationZ, targetType: trgTypeZ, targetName: trgNameZ, outcome: "success", count: 1},
+		{serviceName: "service-A", agentName: "java", destination: destinationZ, targetType: trgTypeZ, targetName: trgNameZ, outcome: "success", count: 0},
+		{serviceName: "service-A", agentName: "java", outcome: "success", count: 1},                                             // no destination or service target
+		{serviceName: "service-A", agentName: "java", targetType: trgTypeZ, targetName: trgNameZ, outcome: "success", count: 1}, // no destination
+		// {serviceName: "service-A", agentName: "java", destination: destinationZ, outcome: "success", count: 1},               // no service target, possible?
```
marclop, line 99: I think this scenario may not be possible since the exit spans will always have a target service set. @SylvainJuge for confirmation

marclop (approved):

x-pack/apm-server/aggregation/spanmetrics/aggregator_test.go
SylvainJuge: In theory, every call to an external service must have the `resource` and `service.target` fields properly set, however it can be modified/removed through the agent API (at least in the case of the Java agent), so this scenario is possible, even if quite uncommon.

I think we can assume that if users need to explicitly modify/remove the destination and service target, it is either to customize it to their (niche) needs, or a work-around to override the default values that have been set by automatic instrumentation.

lahsivjar: @SylvainJuge In the case of no service target should the output have `target.type = ""` and `target.name=""` or should it be nil?

lahsivjar: As per the current apm-server codebase, for empty `resource` [the code doesn't process the span](https://github.com/elastic/apm-server/blob/4406146feaaa5c3d900d6edd62d43614b872fc6f/x-pack/apm-server/aggregation/spanmetrics/aggregator.go#L211-L213). For `target.{type, name}` I think we have 3 options:

1. Process them with empty fields
2. Process them with target set as nil
3. Don't process the span if `target.*` not set, similar to what is done with resource

I think 3 might not be a good approach as there is existing issue #7920 which talks about missing destination metrics due to `resource` fields not set in the input with OTEL sdk.

SylvainJuge: I think in the case where no `service.target` fields are provided, they should be `nil` in the apm-server code, and not stored with empty values but absent from the stored document. If I'm not mistaken, this behavior should be consistent with what we currently do for the `span.destination.service.resource`  field, so I would advocate for option 3.

For #7920 , the issue comes from the fact that we don't infer an equivalent `resource` for OTLP intake, which makes the current implementation of metrics aggregation ignore them. I've added a small chapter on Otel compatibility in the specification https://github.com/elastic/apm/blob/6d1eaa24acf1147c2e98a4a327c43a296213ad0d/specs/agents/tracing-spans-service-target.md#otel-and-bridge-compatibility which should cover this case, so implementing this part might also help fix #7920 , let me know if there is anything unclear or not covered in this area.

simitt (approved):
I reviewed the code changes against the spec and you covered adding the new fields to both aggregation keys and adding the new fields to the internal metrics. PR looks great (except for this one minor [comment](https://github.com/elastic/apm-server/pull/7924/files#r873418819) left).

x-pack/apm-server/aggregation/spanmetrics/aggregator_test.go
lahsivjar: @SylvainJuge Thanks for the clarification. For the implementation of the spec for nil `service.target`, if we were to consider point `3` it would be incompatible with older agents which are sending only the `resource`, given this, should we go with approach `2` i.e. setting the target as `nil` if it is not provided?

SylvainJuge: Agents will send spans with either:
1. `resource` for older agents
2. `resource` and `service.target.*` fields for newer agents
3. no `resource` nor `service.target.*` fields when span is not an exit span.

The behavior on apm-server should be respectively to:
1. store `resource` as-is, store `service.target.type` = `""` (empty string and `service.target.name` = `resource`
2. store `resource` and `service.target.*` fields as provided
3. `resource` = nil and `service.target.*` fields should not be present as well.

marclop (approved):
LGTM!

stuartnelson3:
I'm not sure this one is working.

tested with 6256e0ffa6f5d647a1c991d119693592dad321ab

```
# curl -i -H "Content-type: application/x-ndjson" --data-binary @spans.ndjson http://localhost:8200/intake/v2/events
HTTP/1.1 202 Accepted
X-Content-Type-Options: nosniff
Date: Mon, 13 Jun 2022 10:36:37 GMT
Content-Length: 0

```

`spans.ndjson`:
```json
{"metadata": {"user": {"domain": "ldap://abc", "id": "123", "email": "s@test.com", "username": "john"}, "process": {"ppid": 6789, "pid": 1234,"argv": ["node", "server.js"], "title": "node"}, "system": {"platform": "darwin", "hostname": "prod1.example.com", "architecture": "x64", "container": {"id": "container-id"}, "kubernetes": {"namespace": "namespace1", "pod": {"uid": "pod-uid", "name": "pod-name"}, "node": {"name": "node-name"}}}, "labels": {"tag1": "label1"}, "service": {"name": "backendspans", "language": {"version": "8", "name": "ecmascript"}, "agent": {"version": "3.14.0", "name": "elastic-node"}, "environment": "staging", "framework": {"version": "1.2.3", "name": "Express"}, "version": "5.1.3", "runtime": {"version": "8.0.0", "name": "node"}},"cloud":{"account":{"id":"account_id","name":"account_name"},"availability_zone":"cloud_availability_zone","instance":{"id":"instance_id","name":"instance_name"},"machine":{"type":"machine_type"},"project":{"id":"project_id","name":"project_name"},"provider":"cloud_provider","region":"cloud_region","service":{"name":"lambda"}}}}
{"span": {"trace_id": "abcdef0123456789abcdef9876543210", "parent_id": "0000000011111111", "id": "1234abcdef567895", "transaction_id": "ab45781d265894fe", "name": "GET /api/types", "type": "request", "start": 22, "duration": 32.592981, "timestamp": 1532976822281000,"context":{"service":{"target":{"type":"mytype","name":"myname"}},"destination":{"service":{"resource":"myresource"}}}}}
```

I didn't find any metrics datastreams:
image

lahsivjar:
@stuartnelson3 The span in the `spans.ndjson` is missing the `sample_rate` field. I think the aggregator doesn't calculate metrics for zero sample rate. Can you add a `sample_rate: 1` in the span body and check again? (apologies for not having more details in the test description)

Also, the timestamp in the span is quite old so we may have to expand the time range by a few years.

stuartnelson3:
confirmed on cloud while pairing with @lahsivjar

PR #7925, opened 2022-04-21
Backward compatibility with svc target and dest resource

### LINKED ISSUE #7734: Backend dependencies granularity: keep backwards compatibility
Follow up on https://github.com/elastic/apm-server/issues/7715 to also adapt the processing logic to the new intake fields:

## Intake API
Deprecate `span.context.destination.service.resource` on the Intake API. 

## Processing Logic
For backwards compatibility reason, the APM Server should 
* for certain span types, keep populating `span.destination.service.resource`. If agents are sending the new `span.context.service.target.*` fields, the value needs to be derived from the new fields (data compatibility)
* always populate the new `span.service.target.*`. For backwards compatibility with old agents, split the value from `span.destination.service.resource` to derive `span.service.target.*` information. 

A detailed description for when and how to derive the values is provided in https://github.com/elastic/apm/pull/627/files#diff-78ae87d6c58be8cc0ed40d4dcd78faa83a58e462ad0b836912194a5cf8aed336R143

lahsivjar:
## Motivation/summary

1. Deprecate `span.context.destination.service.resource`
2. Infer `span.context.service.target.{type, name}` from `span.context.destination.service.resource` if the former not provided

Full spec list available at https://github.com/elastic/apm/pull/627

## Checklist

## How to test these changes

**Case 1**
1. Send `span.context.service.target.{type, name}` via the intake API
2. Observe the generated span documents to contain `span.service.target.name` and `span.service.target.type` values 

**Case 2**
1. Send `span.context.destination.service.resource` via the intake API along with `span.context.service.target.{type, name}`
2. Observe the generated span documents to contain `span.service.target.name` and `span.service.target.type` values as passed via the intake API

**Case 3**
1. Send `span.context.destination.service.resource` via the intake API WITHOUT `span.context.service.target.{type, name}` in 3 formats via 3 separate requests, example:
    - `postgres/testdb`
    - `postgres`
    - `localhost:8080`
2. Observe the generated span document to contain `span.service.target.name` and `span.service.target.type` values based on the spec. Example based on point 1:
    - `target.type = postgres` and `target.name = testdb`
    - `target.type = postgres`
    - `target.type = ""` and `target.name = localhost:8080`

## Related issues

Closes #7734


stuartnelson3 (changes requested):
looks good! could you add a unit test to verify this is being called and doing the right thing within the mapping code path?

SylvainJuge (commented):
LGTM on my side, definitely seems to fit the spec. +100 on adding automated tests for that as this kind of mapping tend to break easily over time. As I'm not go-fluent and very familiar with apm-server codebase, don´t  block merging this PR on my approval.

I would be able to validate this with an agent once https://github.com/elastic/apm-server/pull/7924 is merged.

SylvainJuge:
In the Case 3, when not sending the new fields, for the last case, on top of checking that `target.name = localhost:8080`, we need to check that `target.type = ""` (empty string). This is important as it allows the UI to query on `target.{name,type}` and use the empty `target.type` as an indicator to use the `resource` for display (might be covered in https://github.com/elastic/kibana/issues/131923)

This should also apply to the general case where the resource can't be parsed to any known format, in this case the field `target.type` should be set to an empty value.

model/modeldecoder/v2/decoder.go
```diff
@@ -1434,3 +1443,19 @@ func mapSpanLinks(from []spanLink, out *[]model.SpanLink) {
...
+			outTarget.Type = submatch[1]
+			outTarget.Name = submatch[2]
+		} else if patternWithOnlyAlphabets.MatchString(res.Val) {
+			outTarget.Type = res.Val
+		} else {
+			outTarget.Name = res.Val
+		}
+	}
+	return outTarget
+}
```
marclop, line 1461: What do you think about simplifying this to a single regex expression, accepting a string, and not returning a pointer to that struct:

```suggestion
func targetFromDestinationResource(resource string) (target model.ServiceTarget) {
	submatch := patternServiceTargetRe.FindStringSubmatch(resource)
	switch len(submatch) {
	case 1:
		target.Type = resource
	case 2:
		target.Type = submatch[1]
		target.Name = submatch[2]
	default:
		target.Name = resource
	}
	return target
}
```

model/modeldecoder/v2/decoder.go
```diff
@@ -67,6 +68,11 @@ var (
 	}
 )
 
+var (
+	patternWithSingleSlash   = regexp.MustCompile(`^([a-z]+)/(\w+)$`)
+	patternWithOnlyAlphabets = regexp.MustCompile(`^[a-z]+$`)
```
marclop, line 73: It should be possible to use a single regex with a non-capturing group for the slash (`/`):

```suggestion
	patternServiceTargetRe = regexp.MustCompile(`^([a-z]+)(?:/)?(\w+)?$`)
```

model/modeldecoder/v2/decoder.go
```diff
@@ -968,6 +974,9 @@ func mapToSpanModel(from *span, event *model.APMEvent) {
 		mapToServiceModel(from.Context.Service, &event.Service)
 		mapToAgentModel(from.Context.Service.Agent, &event.Agent)
 	}
+	if !from.Context.Service.Target.Type.IsSet() {
+		event.Service.Target = inferFromResource(from.Context.Destination.Service.Resource)
```
marclop, line 978: Moving the service destination resource check outside of the function call leaves the code inside that function slightly cleaner, also, we're not immediately escaping that pointer to the heap, although, we are going to anyway in this case.

```suggestion
		if res := from.Context.Destination.Service.Resource; res.IsSet() {
			target := targetFromDestinationResource(res.Val)
			event.Service.Target = &target
		}
```

marclop (commented):
I'm also in favor to adding some tests here to ensure it behaves as expected.

model/modeldecoder/v2/decoder.go
lahsivjar: Hmm, yes this should be possible but we will have to put `\w+` also in the first non-capturing group along with `/` otherwise abcd2343 will also match. Let me think about it.

model/modeldecoder/v2/decoder.go
lahsivjar: > accepting a string, and not returning a pointer to that struct
Sounds cleaner, will do the changes

> What do you think about simplifying this to a single regex expression
Let's continue the discussion in the lines involving regex comment below since it will depend on the conclusion of using a single regex expression

model/modeldecoder/v2/decoder.go
lahsivjar: Using `^([a-z]+)(?:/(\w+))?$` as regex is one possible solution for utilising single regex ([rough draft on go playground](https://go.dev/play/p/6Hkm9tSg1zA)). Readability wise I am a tiny bit more inclined towards using separate regexes. WDYT?

marclop: I don't have a preference between `^([a-z]+)(?:/)?(\w+)?$` and `^([a-z]+)(?:/(\w+))?$` since from what I can see they're identical.

On readability, I don't think having two regexes makes it more readable, since there's more logic to handle on the `if/else` and I prefer to read a single regex that has a nice comment on it (explaining the different groups and what they intend to capture) rather than having an `if/else/else if` combination. Happy to have other members of the team chime in.

I also think that `w` and `[a-z]` are exactly the same so I'm not too worried bout which one to use either.

lahsivjar: > I don't have a preference between ^([a-z]+)(?:/)?(\w+)?$ and ^([a-z]+)(?:/(\w+))?$ since from what I can see they're identical.

With the former, as per my understanding, it is not possible to distinguish between `postgres/123` and `postgres123`. Both of these will lead to submatch array of length 3. On the other hand. the latter expression will not match for `postgres123`

lahsivjar: > I also think that w and [a-z] are exactly the same so I'm not too worried bout which one to use either.

`w` should be equivalent to `[a-zA-Z0-9_]`

simitt: I generally find regexes hard to understand and am a fan of simplifying them for readability. But no strong feelings, whichever one of the two discussed solutions you move forward are fine with me. 
+1 on adding a comment what the regex is intended to capture.

lahsivjar: On thinking about it, I still don't have a strong inclination towards either of the two approaches. The approach involving a single regex doesn't lead to a very complicated regex or logic so I decided to update the PR with the single regex approach.

marclop (approved):
LGTM. Thanks for all the changes

marclop:
Mind adding a changelog entry as well? It seems that's the only thing missing.

lahsivjar:
> Mind adding a changelog entry as well? It seems that's the only thing missing.

@marclop Added changelog entry

stuartnelson3 (approved):
looks good!

stuartnelson3:
tested with d920cdd3487563f235eb40c554cce316a15bfdde

case 1:

```
# cat doc1.json | jq .span
{
  "duration": {
    "us": 32592
  },
  "name": "GET /api/types",
  "id": "1234abcdef567895",
  "type": "request"
}

# cat doc1.json | jq .service.target
{
  "name": "myname",
  "type": "abc123"
}
```
Is this correct? The test case says the data should be available under `span.service.target.*`, but it's actually under `_source.service.target*`. Not sure if you meant `_source` to represent the span document data, or if it's to be nested under `span` within the document.

case 2: same question; assuming this is correct then :ok_hand: 

case 3: same question; assuming this is correct then :ok_hand:

lahsivjar:
@stuartnelson3 You are correct, it should be `_source.service.target.*`, I have updated the summary. Thanks for the tests.

stuartnelson3:
cool, then :+1: confirmed

PR #7966, opened 2022-04-26
benchmark: Adds TF modules, benchmark workspace

### LINKED ISSUE #7854: benchmark: Create APM Server deployment terraform module
## Description

The _benchmarked_ APM Server is going to be running in ESS, because of that, we should write a small Terraform module which allows us to manage that infrastructure efficiently. The Terraform module should allow us to customize the parameters that we need to run our benchmarks, nothing else. These attributes are:

- Deployment region
- Deployment template id
- Version
- APM Server size
- APM Server zones
- Elasticsearch size
- Elasticsearch zones
- Docker image (and/or docker image tag). This setting is only allowed for Elastic employees and is available through the Terraform provider and ESS API.

### LINKED ISSUE #7855: benchmark: Create terraform module that allows the creation of apmbench "workers"
## Description

Since the APM Server we want to benchmark will be running in ESS (generally `gcp-us-west2` or `gcp-europe-west1` since those are the Cloud-first test regions), the `apmbench` binary should be run as close to it as possible. For that purpose, we should create a Terraform module which:

- Creates a VM with a CPU intensive profile (since there's going to be a decent amount of concurrency when running a medium/high number of `-agents` flag). `c6i.4xlarge` or `c6i.2xlarge` have worked well in the tests I've run.
- Uploads a file with all the necessary environment variables exported which is then read or sourced before `apmbench` is executed, so it has the credentials.
- Uploads a compiled `apmbench` which is compatible with the remote VM architecture.
- Enables `expvar` and `pprof` (has to be done through the Fleet API).

marclop:
## Motivation/summary

Adds two new modules under `testing/infra/terraform/modules`:

- `ec_deployment`: Provisions a new deployment in ESS / ECE.
- `benchmark_executor`: Provisions a VM in the configured region and
   uploads a compiled version of `apmbench` to it with an `.envrc` file
   with credentials to index data into a configured deployment.

Most of the codebase has been copied from `ecetest` and modified to work
in a standalone module.

Additionally, a new `testing/benchmark` workspace has been added which
facilitates using the two modules with light automation via `make` and
a few pre-configured settings.

The result is a folder which can be used by devs or automation to create
a configurable deployment (in ESS by default) which can be benchmarked
using `apmbench` running in a remote VM which can also be configured.

## Checklist

## How to test these changes

1. [Read the documentation](https://github.com/marclop/apm-server/tree/f/add-deployment-module-worker-module-benchmark-workspace/testing/benchmark).
2. `cd testing/benchmark && make`
3. `make run-benchmark BENCHMARK_RUN=BenchmarkAgent BENCHMARK_TIME=2m`

## Related issues

* Closes #7854
* Closes #7855


Makefile
```diff
@@ -115,10 +116,14 @@ get-version:
...
 
 .PHONY: docs
-docs:
+docs: tf-docs
 	@rm -rf build/html_docs
 	sh script/build_apm_docs.sh apm-server docs/index.asciidoc build
 
+.PHONY: tf-docs
+tf-docs: $(TERRAFORMDOCS)
+	@$(foreach dir, $(wildcard infra/terraform/modules/*),terraform-docs markdown --hide-empty --header-from header.md --output-file=README.md --output-mode replace $(dir);)
```
axw, line 125:
```suggestion
	@$(foreach dir, $(wildcard infra/terraform/modules/*),$(TERRAFORMDOCS) markdown --hide-empty --header-from header.md --output-file=README.md --output-mode replace $(dir);)
```

I think?

axw: Maybe use a pattern rule?

axw (approved):
LGTM.

Do you anticipate the TF modules being useful outside of benchmarking/performance testing? I wonder if we should move `infra` to `testing/infra`. Perhaps later we should also move `apmbench` so everything is close together.

marclop:
That's a good point. Initially it probably makes sense to move it within the `testing/` tree and if we use it outside of testing / benchmarking then we can consider moving it out of there.

Makefile
marclop: fixed and added a pattern rule to make it more "make"

PR #7974, opened 2022-04-27
Allow benchmarks to specify max rate limit for events

### LINKED ISSUE #7843: apmbench: Allow specifying a target of `events/unit` sent to APM Server
## Description

Allow setting a target of events that are sent to the APM Server during benchmark execution in the `event/unit` format. The scope for this feature is to allow setting the rate of events as sent by `apmbench` rather than always sending as fast as possible.

It may be necessary to modify the existing `benchtest/eventhandler` package to accept a rate limiter which throttle the events sent to make coordination between goroutines possible.

lahsivjar:
## Motivation/summary

Allow benchmarks to specify a max event rate to allow for a constrained load on APM server. Example:

Results without max-event-rate set:
```
$ go run cmd/apmbench/main.go -warmup-events 1000000 -agents 16 -run BenchmarkAgentGo -count 1 -benchtime 1m
BenchmarkAgentGo-16	     198	 331809204 ns/op	         0 error_responses/sec       316.4 errors/sec	     15777 events/sec	       725.8 metrics/sec	     10407 spans/sec	      4328 txs/sec	 1082090 B/op	    2240 allocs/op
```

Results with max-event-rate set:
```
$ go run cmd/apmbench/main.go -warmup-events 1000000 -agents 16 -run BenchmarkAgentGo -count 1 -max-rate 5000eps -benchtime 1m
BenchmarkAgentGo-16	      70	1034792153 ns/op	         0 error_responses/sec	       101.5 errors/sec	      5063 events/sec	       236.5 metrics/sec	      3337 spans/sec	      1388 txs/sec	 1276948 B/op	    2831 allocs/op
```

The approach used is [token bucket](https://en.wikipedia.org/wiki/Token_bucket) which gives us control over throughput to APM server while allowing for bursty traffic similar to real world use cases.

## Checklist

## How to test these changes

1. Run the apmtest benchmark with max events per second set via `max-rate` flag, example:
    ```
    cd systemtest && go run cmd/apmbench/main.go -warmup-events 1000000 -agents 16 -run BenchmarkAgentGo -count 1 -benchtime 1m -max-rate 5000eps
    ```
    NOTE: observed events per second might be slightly higher than the `max-rate` value (particularly when agent count is low) due to bucket size which is initially full and is greater than the max-rate to allow for bursts([ref](https://pkg.go.dev/golang.org/x/time/rate))
2. `apmtest` benchmark can be run without limit by not specifying the `max-rate` flag

## Related issues

Closes #7843


systemtest/benchtest/eventhandler/handler.go
```diff
@@ -128,7 +130,7 @@ func New(p string, t *Transport, storage fs.FS, warmup uint) (*Handler, error) {
 
 // SendBatches sends the loaded trace data to the configured transport. Returns
 // the total number of documents sent and any transport errors.
-func (h *Handler) SendBatches(ctx context.Context) (uint, error) {
+func (h *Handler) SendBatches(ctx context.Context, l *rate.Limiter) (uint, error) {
```
marclop, line 133: Do you think it would be reasonable to keep this signature as is, and instead inject `*rate.Limiter` when the `Handler` is constructed in `event handler.New`? It seems unlikely that we'll use different rate limiters on different calls to `SendBatches`.

systemtest/benchtest/main.go
```diff
@@ -121,6 +125,23 @@ func benchmarkFuncName(f BenchmarkFunc) (string, error) {
 	return name, nil
 }
 
+func getLimiter(epm int) *rate.Limiter {
+	if epm < 0 {
```
marclop, line 129: Should we treat the negative and 0 value equally? Or is there a use case for 0 EPS/EPM?

marclop (commented):
Mostly LGTM. I'm a bit undecided as to whether or not we should accept a context in the benchmark function, particularly since we're passing down a `context.Background()`.

systemtest/benchtest/eventhandler/handler.go
lahsivjar: `WarmupServer` is also defined on the event handler receiver and currently I don't limit warmup events by using a Inf limiter. Having the limiter in the handler but overriding the limiter in warmup server might lead to confusion so I kept limiter as an argument.

If we limit warmup events also then I think having Limiter as part of the handler struct would make more sense but I am not sure if we should limit warmup events. WDYT?

marclop: I see, I am more inclined to have a consistent limit both in warm-up and benchmark execution rather than warm up the server as fast as we can and then rate-limit on the benchmarks.

I can't think of a use-case where we'd want to warm up the APM Server by sending events events at max throughput and then wanting to benchmark a sustained throughput.

Also because of the way the flags are configured, I would expect the apmbench to limit both warm up and each of the benchmark cases.

systemtest/benchtest/main.go
lahsivjar: Hmm, a 0 limit would mean that the bucket is not getting any tokens and would error out after sending 1000 events (the min bucket size). I don't think 0 limit has a good practical use case, we can treat them same as negative. Will update the PR.

lahsivjar:
@marclop For passing the context to the benchmark function, I added it thinking its best to initialize context early on and pass it around so that we can manipulate it, if required, for cancellation or other requirements. Using context.Background for every use-case might make it harder to utilize the context tree as the code grows.

systemtest/benchtest/eventhandler/handler.go
lahsivjar: > I can't think of a use-case where we'd want to warm up the APM Server by sending events events at max throughput and then wanting to benchmark a sustained throughput.

Hmm, how about the case where we want to benchmark a low throughput. Having a lower rate limit for warmup events might take too much time since warmup events have a high count to be useful.

lahsivjar: Looking at the code again, it seems that warmup initialises a separate event handler so promoting rate.Limiter to the handler struct sounds good, I will do the changes. WDYT about above point? Should we have a separate limiter for warmup events to make warmup quickly?

marclop: sounds good, I was thinking if it would make sense to introduce a flag, but I'd rather not unless it becomes a problem. It should be a straightforward enough change to do if we need to, and we can already control the number of warmup events. Does that sound reasonable to you?

lahsivjar: > I was thinking if it would make sense to introduce a flag, but I'd rather not unless it becomes a problem.

I was thinking along the same lines, agreed on introducing it if required. For now I will make the warmup events use the  limiter with same configuration.

lahsivjar: Updated the code with the required changes

marclop (approved):
LGTM. Thanks for the changes. After looking at it again, I think it's not unreasonable to pass down the context, however, we've got no deadline or timeout for cancellation as is.

lahsivjar:
> however, we've got no deadline or timeout for cancellation as is.

Yup, currently we are not utilizing the context. We also have the `benchtime` flag which controls each benchmark's duration. We can probably analyse the requirements as the need arise. WDYT?

lahsivjar:
After discussion with @marclop we have decided to remove context propagation and consider it when and if required by a use-case for apmbench.

marclop (approved):
LGTM!

marclop:
Verified with cf6e6a98:


with `-max-rate`
```
$ apmbench -agents 32 -benchtime=20s -run BenchmarkAgent -max-rate 800eps
BenchmarkAgentAll-32   	       1	23574557977 ns/op	       764.3 events/sec	847483928 B/op	 9631661 allocs/op
BenchmarkAgentGo-32    	       4	6381470610 ns/op	       846.4 events/sec	231386054 B/op	 2188910 allocs/op
BenchmarkAgentNodeJS-32	       9	2472516810 ns/op	       813.7 events/sec	102227351 B/op	 1203862 allocs/op
BenchmarkAgentPython-32	       3	8820205040 ns/op	       795.8 events/sec	358911200 B/op	 4266450 allocs/op
BenchmarkAgentRuby-32  	       6	4608803471 ns/op	       825.7 events/sec	185459880 B/op	 2039004 allocs/op
$ apmbench -agents 32 -benchtime=20s -run BenchmarkAgent -max-rate 3000eps
BenchmarkAgentAll-32   	       4	6550732963 ns/op	      2776 events/sec	744123524 B/op	 8599514 allocs/op
BenchmarkAgentGo-32    	      13	1715088710 ns/op	      3052 events/sec	173370220 B/op	 1845406 allocs/op
BenchmarkAgentNodeJS-32	      33	 695907614 ns/op	      2891 events/sec	88922864 B/op	 1087194 allocs/op
BenchmarkAgentPython-32	       9	2510405707 ns/op	      2796 events/sec	307427632 B/op	 3830232 allocs/op
BenchmarkAgentRuby-32  	      18	1274204312 ns/op	      2945 events/sec	151703845 B/op	 1797515 allocs/op
$ apmbench -agents 32 -benchtime=20s -run BenchmarkAgent -max-rate 100000epm
BenchmarkAgentAll-32   	       2	11658020712 ns/op	      1546 events/sec	761800208 B/op	 8904627 allocs/op
BenchmarkAgentGo-32    	       7	3134118589 ns/op	      1700 events/sec	195095995 B/op	 1947335 allocs/op
BenchmarkAgentNodeJS-32	      20	1185406075 ns/op	      1697 events/sec	92875737 B/op	 1121547 allocs/op
BenchmarkAgentPython-32	       5	4302021394 ns/op	      1632 events/sec	322393289 B/op	 3954861 allocs/op
BenchmarkAgentRuby-32  	       9	2250193715 ns/op	      1668 events/sec	159124680 B/op	 1863438 allocs/op
```

Unbounded

```
$ apmbench -agents 32 -benchtime=20s -run BenchmarkAgent
BenchmarkAgentAll-32   	       5	4138928668 ns/op	      4353 events/sec	702353073 B/op	 8396566 allocs/op
BenchmarkAgentGo-32    	      33	 643075392 ns/op	      8139 events/sec	163685404 B/op	 1766417 allocs/op
BenchmarkAgentNodeJS-32	      64	 340074864 ns/op	      5916 events/sec	85341083 B/op	 1058090 allocs/op
BenchmarkAgentPython-32	      18	1165133915 ns/op	      6032 events/sec	297537464 B/op	 3737762 allocs/op
BenchmarkAgentRuby-32  	      39	 563313710 ns/op	      6662 events/sec	145305742 B/op	 1743757 allocs/op
```

Issue #7540, opened 2022-03-15
Benchmark 2.0 production ready


simitt:
In https://github.com/elastic/apm-server/issues/7216 @marclop worked on a POC for a new benchmarking tool for APM data ingestion. The POC is leveraging Opbeans services to create apm events, which are recorded into ndjson files by the new Intake Receiver component. The `apmbench` loader reads from the ndjson files and the runner sends the data to an APM Server, while also collecting statistics which are indexed into a dedicated Elasticsearch. Find a more detailed description in https://github.com/elastic/apm-server/issues/7216#issue-1125938474.

After successfully building the POC, all the components should be tight together in a maintainable way.

## Goals
* New Benchmark tooling is wired up to run automated benchmarks on a fixed set of testdata on a regular schedule
* Can be integrated into the teams workflow when building new features or refactoring (ad-hoc and automated benchmarks)
* Can easily be used by agent developers to create a new testdata set for a new agent version for automated benchmarks, and ad-hoc benchmarks.
* Can be used by engineers and developers outside the apm-server team for throughput ananlysis with a targeted throughput per unit. 
* Update the [processing & performance guide](https://www.elastic.co/guide/en/apm/guide/current/processing-and-performance.html) with updated numbers

## Tasks
* [ ] benchmark limits for `events/unit` should be configurable (in addition to currently supported max. load)
* [ ] collect additional metrics for better insights 
* [ ] Implement alternative solution to committing the generated `ndjson` files in the APM Server repo. As one option, we discussed uploading generated files to s3 and providing a small uploader tool that is easy to use. Ideally the tool 
* [ ] Support feature specific use cases, at least, there should be enough data, or data (trace_ids, ..) being manipulated without losing semantic meaning, to test tail based sampling. 
* [ ] https://github.com/elastic/observability-dev/issues/2014
* [ ] build a meaningful dashboard 
* [ ] collaborate with the automation team for plumbing everything together and automatically run the benchmarks on a regular basis

PR #7987, opened 2022-04-28
apmbench: Warm-up with specified agent number

marclop:
## Motivation/summary

Changes the benchmark warm-up to use the same amount of specified agents
and fixes the broken glob expression, causing warm-up to be skipped.

## Checklist

## How to test these changes

1. Build `apmbench`: `cd systemtest/cmd/apmbench && go build .`
2. Run it with an `-agents` > 1: `./apmbench -agents=5 -run BenchmarkAgent`
3. Navigate to `localhost:8200/debug/expvar`, check that the total number of `acked` events is > `25000` (5000 warm up events by 5 agents).

## Related issues

Part of #7540


systemtest/benchtest/main.go
```diff
@@ -240,3 +236,33 @@ func Run(allBenchmarks ...BenchmarkFunc) error {
 	}
 	return nil
 }
+
+func warmup(agents int, events uint) error {
+	// Assume a base ingest rate of at least 5000 per second, and dynamically
+	// set the context timeout based on this ingest rate, or if lower, default
+	// to 15 seconds. The default 5000 / 5000 = 1, so the default 15 seconds
+	// will be used instead.
+	timeout := time.Duration(math.Max(float64(events/5000), 15)) * time.Second
```
lahsivjar, line 245: We should also consider `max-rate` here. A lower max-rate and corresponding config for `warump-events` can lead to cases were warmup always times out since the base assumption will not align. One option would be to consider `min(5000, maxEPM*60)` instead of hardcoded 5000. WDYT?

systemtest/benchtest/main.go
```diff
@@ -240,3 +236,33 @@ func Run(allBenchmarks ...BenchmarkFunc) error {
...
+
+	for i := 0; i < agents; i++ {
+		h, err := newEventHandler(`*.ndjson`, getNewLimiter())
+		if err != nil {
+			return fmt.Errorf("unable to create warm-up handler: %w", err)
+		}
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			h.WarmUpServer(ctx, events)
```
lahsivjar, line 258: We should handle errors from the `WarmUpServer` function (maybe by using a return channel)

lahsivjar:
LGTM

lahsivjar (approved):

PR #7998, opened 2022-04-28
ci: run the packaging only once

v1v:
## Motivation/summary

Use the packaging downstream pipeline to generate the packages automatically. Therefore, the URL in the packages will be not shown in the `main pipeline` but in the packaging pipeline instead.

Therefore:

- the `main pipeline` does not generate any packages anymore
- the `packaging pipeline` generates the packages regardless of the git event (branches, tags, prs)
- the `packaging pipeline` gets triggered always on a PR basis in the parallel stages in the `main pipeline`
- the `packaging pipeline` runs on branches at the very end of the `main pipeline`.
- the `packaging pipeline` can be triggered manually if needed.

I also added a filter to run the `apmpackage` in the `packaging pipeline` if no a PR (otherwise if triggered manually on a PR basis it will create the PR).

## How to test these changes

In the CI

## Test

This [build](https://apm-ci.elastic.co/blue/organizations/jenkins/apm-server%2Fapm-server-mbp/detail/PR-7998/9/pipeline) triggered this [packaging build](https://apm-ci.elastic.co/blue/organizations/jenkins/apm-server%2Fapm-server-package-mbp/detail/PR-7998/8/pipeline)

image


kuisathaverat (approved):

cachedout (approved):

axw (commented):
AFAICS, `env.BEATS_UPDATED` isn't used anymore. What's the rationale for running the packaging on all PRs? I think it's OK, just wondering.

v1v:
> AFAICS, env.BEATS_UPDATED isn't used anymore. What's the rationale for running the packaging on all PRs? I think it's OK, just wondering.

Good catch

Let me explain the reason for using a single packaging. When I enabled the `DRA`, Daily Releasable Artifacts, I knew I duplicated the package logic in a different pipeline, since the `main pipeline` had the `package` stage. Therefore I saw a potential refactor to use the `package pipeline` for everything (PRs, tags, branches), and enable/disable the `DRA` or `apmpackage` stages accordingly.

This is the proposal, to use the `package pipeline` in favour of removing duplicated code in the pipelines

It works as is so if there is no agreement this PR does nothing but refactoring (and also fix a condition in one stage, that probably I'll need to fix it in a different PR in case this one is closed)

axw (approved):
LGTM, thanks!

Issue #7986, opened 2022-04-28
Support benchmarking an unreleased locally built APM Server


marclop:
## Description

We should document or automate (if it proves easy enough) how to package and run a locally built version of APM Server with some changes in ESS / ECE.

marclop:
My approach for this issue has been to reuse and extend some of the code that we use to build the APM Server for the system tests and re-use the same packaging to produce a Docker Image that would be able to run in ESS or ECE.

For that purpose I introduced a new `buildapm` binary ([link](https://github.com/elastic/apm-server/blob/main/systemtest/cmd/buildapm/main.go)) that builds a docker image from the `fleet-server` service as specified in our `docker-compose.yml` (`docker.elastic.co/beats/elastic-agent`).

I also introduced a new `-cloud` flag in that binary which instead of using the upstream elastic-agent _standard_ image, uses the docker image that would be used in ESS (`docker.elastic.co/cloud-release/elastic-agent-cloud`).

As documented in the Observability-dev documentation PR, the packaging should work and it's a fairly easy experience to do build, package and build a docker image which is then uploaded to a repository which is allowed to run in ESS / ECE.

When using `buildapm` and running the image manually (or just using `runapm` which uses a very similar code path) to run the image locally, the injected `apm-server` runs without any problems and testing changes locally is pretty easy. The same is not applicable for the apm-server in the cloud-release image. It seems that something is missing or the packaging doesn't work exactly right. Although the APM Server binary is executed by the Elastic Agent in the built container, some libbeat code (before any of the APM Server code is run) returns with an error:

```
Exiting: error initializing publisher: missing field accessing 'output.elasticsearch.hosts.0'
```

The code that returns that error is: https://github.com/elastic/beats/blob/ff1041fc099ec69caea11c5c9c83d6ca59170ee7/libbeat/cmd/instance/beat.go#L387

Which is run before the APM Server's Creator func.

It's unclear to me how to debug this further and make headway since debugging is not straightforward.

axw:
I haven't gotten to the bottom of why, but commenting out https://github.com/elastic/apm-server/blob/e960ca3736acbe1d59d3b7ff7b6f35d3c08383ac/cmd/root.go#L57-L67 allows things to proceed. IIANM we should get rid of that override anyway, since we set the default for modelindexer in https://github.com/elastic/apm-server/blob/e960ca3736acbe1d59d3b7ff7b6f35d3c08383ac/elasticsearch/config.go#L76

[...]

PR #8053, opened 2022-05-05
testing: Introduce `buildapm` command

marclop:
## Motivation/summary

Introduces a new command under `systemtest/cmd/buildapm`, allowing devs
to build an Elastic Agent image without the need to have the docker
compose project up and running. The majority of times, it'll be useful
to use `systemtest/cmd/runapm`, but there are cases where some changes
may need to be made and packaged in a docker image to be run separately.

## Checklist

## How to test these changes

* Read `TESTING.md` and verify that the instructions work.

## Related issues

Part of #7986


TESTING.md
```diff
@@ -143,3 +145,29 @@ Usage of /var/folders/35/r4w8sbqj2md1sg944kpnzyth0000gn/T/go-build3644709196/b00
...
+
+It's possible to run `runapm` (as pictured above) and re-use the image that `runapm` builds to use in docker-compose
+files or run in ECE / ESS. However, it's also possible to only build a docker image without requiring the docker-compose
+project containers to be up and running with `systemtest/cmd/buildapm`.
+
+`buildapm` reads the `docker-compose.yml` at the root of the repository and uses that information to build an Elastic Agent
+docker image with an APM Server bundled that contains any local changes you might have made.
+
+By default, the same architecture (or platform in Docker lingo) as your local machine will be used. This may not be ideal
+if you have an architecture different than `amd64`, so you can specify the `-arch` flag.
```
stuartnelson3, line 159: would it be better to have the default be `amd64`, since that's probably going to be the correct arch 99% of the time? How is this handled with docker for mac? I remember it being a virtualized amd64 linux machine, but that could (and probably is) be incorrect.

marclop: I'm not it that since it's probably a safe assumption. I just defaulted to using whatever `runtime.GOARCH` is set to, but thinking about it now, `amd64` is way more widespread than `arm64`, so it sounds like a better default for now, until  (if) `arm64` takes over as the default architecture.

marclop: addressed

stuartnelson3 (approved):

stuartnelson3:
confirmed in https://github.com/elastic/apm-server/pull/8165

Issue #7520, opened 2022-03-14
Store well-defined agent-produced metrics in shared metrics data stream


axw:
Currently, builtin metrics sent by all agents like `system.memory.total` are stored in the data stream `metrics-apm.app.<service>-<namespace>`. This has the effect of creating many data streams, and hence shards, by default.

We should only create service-specific data streams where there is some risk of mapping conflicts, i.e. for custom application metrics. All others should be stored in shared data streams: either the existing `metrics-apm.internal-<namespace>`, or a new one.

PR #8064, opened 2022-05-06
modeldecoder/v2: add `metricset.internal`

axw:
## Motivation/summary

Introduce `internal` field to metricsets, enabling agents to indicate that a metricset contains only internal (non-application) metrics, such as runtime, system, and process metrics.

If `internal` is set to `true`, then the server will discard any metrics in the metricset that it does not know to be internal. This is necessary as the internal metrics data stream uses strict mapping.

If `internal` is is not set at all, then the server will check if all metrics are known to by internal; if any are not, then the entire set is sent to a service-specific application metrics data stream.

## Checklist

## How to test these changes

1. Run all opbeans services and ensure no service-specific data streams are created.
2. We should also modify one of the agents to set the `internal` flag for internal metrics, and check that known internal metrics are stored in the internal metrics data stream, and all others are discarded

## Related issues

https://github.com/elastic/apm-server/issues/7520


axw:
Let's come back to this later, it's not urgent.

PR #8080, opened 2022-05-10
mergify: Attempt auto-merge fix for beats-update

marclop:
## Motivation/summary

Attempt to fix auto-merges for beats-update PRs

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -137,7 +137,7 @@ pull_request_rules:
     conditions:
       - check-success=apm-ci/pr-merge
       - label=automation
-      - base~=^update-beats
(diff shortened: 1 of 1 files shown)


.mergify.yml
```diff
@@ -137,7 +137,7 @@ pull_request_rules:
     conditions:
       - check-success=apm-ci/pr-merge
       - label=automation
-      - base~=^update-beats
+      - head~=^update-beats
```
marclop, line 140: The condition was wrong, the actual check needs to be performed against the head (current branch) rather than base (target branch).

stuartnelson3 (approved):

PR #8081, opened 2022-05-10
processor: Avoid leaking labels between batches

marclop:
## Motivation/summary

Fixes a bug introduced in #6682 by performing a copy of the `baseEvent`
before each of the events is decoded. The event was being shallow copied
and it was reusing the same maps between batches.

The linked PR changed the behavior (although not in a very obviously)
and was instead merging the labels onto the passed `APMEvent`. This
patch reintroduces the previous behavior in a more obvious manner.

## Checklist

## How to test these changes

1. `docker-compose up -d` and install the APM Integration
2. `make && ./apm-server -E output.elasticsearch.username=admin -E output.elasticsearch.password=changeme -E apm-server.expvar.enabled=true -e`
3. `curl -H "Content-type: application/x-ndjson" --data-binary @testdata/intake-v2/events.ndjson http://localhost:8200/intake/v2/events`
4. Navigate to Kibana dev tools and issue the following query:
```
GET traces*/_search
{
  "_source": false, 
  "fields": [
    "labels.*",
    "numeric_labels.*"
  ], 
  "query": {
    "match": {
      "labels.organization_uuid": "9f0e9d64-c185-4d21-a6f4-4673ed561ec8"
    }
  }
}
```
5. Verify there's only 1 hit and it's: 
```javascript
{
    "_index": ".ds-traces-apm-default-2022.05.10-000001",
    "_id": "joj0rYABBkOJH5O9XGQe",
    "_score": 0.2876821,
    "fields": {
        "labels.group": [
            "experimental"
        ],
        "labels.ab_testing": [
            "true"
        ],
        "numeric_labels.segment": [
            5
        ],
        "labels.organization_uuid": [
            "9f0e9d64-c185-4d21-a6f4-4673ed561ec8"
        ]
    }
}
```

### DIFF
--- a/processor/stream/processor.go
+++ b/processor/stream/processor.go
@@ -161,7 +161,9 @@ func (p *Processor) readBatch(
 			// required for backwards compatibility - sending empty lines was permitted in previous versions
 			continue
 		}
-		input := modeldecoder.Input{Base: baseEvent}
+		// We copy the event for each iteration of the batch, as to avoid
+		// shallow copies of Labels and NumericLabels.
+		input := modeldecoder.Input{Base: copyEvent(baseEvent)}
 		switch eventType := p.identifyEventType(body); string(eventType) {
 		case errorEventType:
 			err = v2.DecodeNestedError(reader, &input, batch)
@@ -306,4 +308,17 @@ func (sr *streamReader) wrapError(err error) error {
 		}
 	}
 	return err
+}
+
+// copyEvent returns a shallow copy of the APMEvent with a deep copy of the
+// labels and numeric labels.
+func copyEvent(e model.APMEvent) model.APMEvent {
+	var out = e
+	if out.Labels != nil {
+		out.Labels = out.Labels.Clone()
+	}
+	if out.NumericLabels != nil {
+		out.NumericLabels = out.NumericLabels.Clone()
+	}
+	return out
 }
--- a/beater/test_approved_es_documents/TestPublishIntegrationEvents.approved.json
+++ b/beater/test_approved_es_documents/TestPublishIntegrationEvents.approved.json
@@ -214,13 +214,10 @@
             "labels": {
                 "ab_testing": "true",
                 "group": "experimental",
-                "organization_uuid": "9f0e9d64-c185-4d21-a6f4-4673ed561ec8",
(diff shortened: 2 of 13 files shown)


processor/stream/processor.go
```diff
@@ -307,3 +309,16 @@ func (sr *streamReader) wrapError(err error) error {
 	}
 	return err
 }
+
+// copyEvent returns a shallow copy of the APMEvent with a deep copy of the
+// labels and numeric labels.
+func copyEvent(e model.APMEvent) model.APMEvent {
```
stuartnelson3, line 315: so, it makes sense that we need to make a deep copy and not re-use the labels between decodings.

What I still can't answer is "how did this work before?" The bug surfaced when we moved from using `common.MapStr` to `map[string](NumericLabelValue|LabelValue)`([pr](https://github.com/elastic/apm-server/pull/6682)), so I don't know if the map is 100% at fault. Reverting the above PR resulted in the bug no longer surfacing.

stuartnelson3 (approved):
The explanation for why we need this makes sense, as does why this is only seen when metadata labels exist (the apmevent that is getting copied now has a shared map that doesn't get reset). I still don't know how it worked before, but I'll see if I can make sense of it.

marclop:
Verified with the reproduction steps in `8.2.1` BC2:

```
{
  "took" : 7,
  "timed_out" : false,
  "_shards" : {
    "total" : 1,
    "successful" : 1,
    "skipped" : 0,
    "failed" : 0
  },
  "hits" : {
    "total" : {
      "value" : 1,
      "relation" : "eq"
    },
    "max_score" : 0.2876821,
    "hits" : [
      {
        "_index" : ".ds-traces-apm-default-2022.05.24-000001",
        "_id" : "cVo19YAByBhr8ADC4Z26",
        "_score" : 0.2876821,
        "fields" : {
          "labels.group" : [
            "experimental"
          ],
          "labels.ab_testing" : [
            "true"
          ],
          "numeric_labels.segment" : [
            5.0
          ],
          "labels.organization_uuid" : [
            "9f0e9d64-c185-4d21-a6f4-4673ed561ec8"
          ]
        }
      }
    ]
  }
}
```

Issue #7872, opened 2022-04-14
Allow clients to request non-blocking ingest behaviour


axw:
When using libbeat outputs, we have an arbitrary one second timeout when enqueuing events to the libbeat output event queue: https://github.com/elastic/apm-server/blob/4e5755408350f6868a1cf78270e0a358a81b989a/publish/pub.go#L160

With the new Elasticsearch output (`modelindexer`), we no longer have this arbitrary timeout. Instead, it is the client's responsibility to set a request timeout; the server will block until the request times out/the client closes the connection.

Some clients, such as Lambda, would prefer to minimise latency as the expense of possibly losing events. For Lambda, time is literally money, as invocations are billed by the amount of wall-clock time they are active.

We should look at options for clients to request non-blocking ingest, meaning that if there is queue capacity then APM Server should enqueue an event, and otherwise it should be immediately rejected.

axw:
As an alternative, we should investigate just never blocking if all bulk indexers are flushing, i.e. today's equivalent of "queue is full". This would be similar to the old arbitrary 1 second timeout on enqueue, but now instead it would be immediate. When we investigate this option, we'll need to re-run benchmarks for backend agents.

marclop:
@axw my idea was to introduce a `fail_fast` query parameter to the `intake/v2/events` endpoint (intake/v2/events?fail_fast=true) and return immediately if the current semaphore cannot be acquired. 

Is that sufficient for how we want to treat "immediate" processing, or would we want go a step further and create a `context.WithTimeout()` with a maximum deadline (say 1s or 500ms) to try and ensure speedy processing for that fail_fast request?

With the current interfaces, there's no way to identify when all the bulk indexers are being flushed, if we wanted to expose that in the current decoder, we'd need to do some refactoring to expose that somehow.

axw:
> my idea was to introduce a fail_fast query parameter to the intake/v2/events endpoint (intake/v2/events?fail_fast=true) and return immediately if the current semaphore cannot be acquired. Alternatively, a header could also be used to signal that behavior, but I'm not sure it's "cleaner" or better. I think I'm leaning more towards the query parameter rather than the header since it's more visible and seems easier to use.

I think a query param is fine.

Would it be enough to just get past the semaphore though? IIANM, it is still possible for the final model processor (i.e. modelindexer) to block, say if Elasticsearch is inaccessible for a while.

[...]

PR #8102, opened 2022-05-16
processor/stream: Re-introduce queue is full

marclop:
## Motivation/summary

This patch re-introduces the `publish.QueueFull` error on the stream
processor layer. Allowing the APM Server to push back back when the
event output rate is significantly slower than the received event rate.

Since the semaphore is bound to 200 by default, this behavior should
only manifest when the concurrent number of in-flight requests exceeds
200 and Elasticsearch isn't able to process the APM Server's events as
fast as APM Server receives them.

The resulting behavior is slightly different than before since the back-
off will occur not on any bounded queue, but instead takes into account
the currently open requests that are being decoded.

This results in a very effective and relatively fast circuit breaker
when APM Server is overwhelmed and should prevent OOM issues when the
received event rate wildly exceeds the output event rate.

There are some open questions to this approach:

1. Is `1s` an acceptable "timeout" period?
  1.1. Should it be configurable?
2. Is `200` a good limit to set for the open and decoding requests?
  2.1. Would a lower number reduce the contention on `activeMu` lock and
       provide better performance?

## Checklist

- [ ] Benchmark change

## How to test these changes

TODO

## Related issues

Part of #7872


processor/stream/processor.go
```diff
@@ -272,6 +277,23 @@ func (p *Processor) getStreamReader(r io.Reader) *streamReader {
 	}
 }
 
+func (p *Processor) acquireLock(ctx context.Context, d time.Duration) (func(), error) {
+	select {
+	case p.sem <- struct{}{}:
+	default:
+		t := acquireTimer(d)
+		defer releaseTimer(t)
```
axw, line 285: Did you run some benchmarks that found the pooling to be worthwhile? I wonder if we should keep it simple and use `context.WithTimeout`, and then check for `context.DeadlineExceeded` and convert that to `publish.ErrFull`?

axw (commented):
This approach looks pretty reasonable to me. 1 second might still be too long for some clients (namely Lambda), but if nothing else we'll avoid backing up an indefinite number of requests in memory.

I don't love the arbitrary 1 second delay, but I don't have a strong opinion on making it configurable. It wasn't configurable before, and that was not terrible.

processor/stream/processor.go
marclop: that may be better indeed. I didn't benchmark the change and I think it may be too much to begin with, but since It's in the hot path, I wanted to reuse the timers, since `time.After` will add to the number of GC objects, but that may not be a problem since the `time.Timer` looks like a pretty small `struct`.

I can run some benchmarks and see how it compares.

If it turns out the overhead is pretty small and pooling isn't worth it, do you have a preference between using `context.WithTimeout` and `time.After`? The latter seems much more contained than the former, so it may be slightly less wasteful to use `time.After`?

axw: I suppose I would slightly prefer a timer, as the context could in theory already have a deadline which should not be converted to ErrFull.

marclop: did some benchmarks and unsurprisingly the difference isn't massive, but at first the difference between the pooled and the `time.After` approach was bigger than I expected:

```
$ benchstat pool.txt timeafter.txt
name                             old time/op    new time/op    delta
BackendProcessorParallelBlock-8    6.62ms ± 2%    6.84ms ± 1%  +3.25%  (p=0.000 n=10+9)

name                             old speed      new speed      delta
BackendProcessorParallelBlock-8  60.2MB/s ± 2%  58.3MB/s ± 1%  -3.15%  (p=0.000 n=10+9)

name                             old alloc/op   new alloc/op   delta
BackendProcessorParallelBlock-8    13.5MB ± 0%    13.6MB ± 0%  +0.01%  (p=0.010 n=10+9)

name                             old allocs/op  new allocs/op  delta
BackendProcessorParallelBlock-8     24.8k ± 0%     24.8k ± 0%  +0.03%  (p=0.000 n=10+10)
```

This is the benchmark I used:

```go
func BenchmarkBackendProcessorParallelBlock(b *testing.B) {
	cfg := config.DefaultConfig()
	cfg.MaxConcurrentDecoders = 1
	processor := BackendProcessor(cfg, make(chan struct{}, cfg.MaxConcurrentDecoders))
	file := "../../testdata/intake-v2/heavy.ndjson"
	const batchSize = 10
	batchProcessor := nopBatchProcessor{} // This is probably unrealistic since the processors will take time to process each event.
	data, err := ioutil.ReadFile(file)
	if err != nil {
		b.Error(err)
	}
	b.SetBytes(int64(len(data)))

	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
	defer cancel()
	b.ResetTimer()
	b.RunParallel(func(p *testing.PB) {
		r := bytes.NewReader(data)
		for p.Next() {
			var result Result
			processor.HandleStream(ctx, model.APMEvent{}, r, batchSize, batchProcessor, &result)
			r.Seek(0, io.SeekStart)
		}
	})
}
```

Then I decided to write a small mock processor that has 2 muteness similar to what the model indexer has:

```go
type mutexBatchProcessor struct {
	rwm sync.RWMutex
	mu  sync.Mutex
}

func (p *mutexBatchProcessor) ProcessBatch(_ context.Context, b *model.Batch) error {
	p.rwm.RLock()
	defer p.rwm.RUnlock()
	// Simulate modelindexer
	for range *b {
		p.mu.Lock()
		p.mu.Unlock()
	}
	return nil
}
```

Using that `mutexBatchProcessor` in the above benchmark resulted in a smaller difference in performance, but the pooled approach still looks like it's generally faster:

```
$ benchstat pool-mi-sim.txt timeafter-mi-sim.txt
name                             old time/op    new time/op    delta
BackendProcessorParallelBlock-8    6.62ms ± 2%    6.80ms ± 1%  +2.63%  (p=0.000 n=10+9)

name                             old speed      new speed      delta
BackendProcessorParallelBlock-8  60.2MB/s ± 2%  58.7MB/s ± 1%  -2.57%  (p=0.000 n=10+9)

name                             old alloc/op   new alloc/op   delta
BackendProcessorParallelBlock-8    13.5MB ± 0%    13.5MB ± 0%    ~     (p=0.739 n=10+10)

name                             old allocs/op  new allocs/op  delta
BackendProcessorParallelBlock-8     24.8k ± 0%     24.8k ± 0%    ~     (p=0.190 n=10+10)
```

Then I tried to increase the number of GC'd and re-used `time.Timer` by setting `-cpu=50`, the deltas were pretty similar, though:

```
benchstat pool-mi-sim-50-cpu.txt timeafter-mi-sim-50-cpu.txt
name                              old time/op    new time/op    delta
BackendProcessorParallelBlock-50    7.00ms ± 2%    7.16ms ± 1%  +2.34%  (p=0.000 n=9+10)

name                              old speed      new speed      delta
BackendProcessorParallelBlock-50  57.0MB/s ± 2%  55.7MB/s ± 1%  -2.28%  (p=0.000 n=9+10)

name                              old alloc/op   new alloc/op   delta
BackendProcessorParallelBlock-50    13.7MB ± 0%    13.7MB ± 0%    ~     (p=0.579 n=10+10)

name                              old allocs/op  new allocs/op  delta
BackendProcessorParallelBlock-50     24.8k ± 0%     24.8k ± 0%  +0.01%  (p=0.000 n=9+10)
```

----

I then increased the capacity of the semaphore to `10` and set the `-cpu=100` to increase concurrency and see how it fares then, the differences were much smaller, so small that I'm not sure if they'll have any effect in production:

```
name                               old time/op    new time/op    delta
BackendProcessorParallelBlock-100    2.74ms ± 1%    2.76ms ± 2%  +0.71%  (p=0.046 n=9+8)

name                               old speed      new speed      delta
BackendProcessorParallelBlock-100   146MB/s ± 1%   145MB/s ± 2%  -0.70%  (p=0.043 n=9+8)

name                               old alloc/op   new alloc/op   delta
BackendProcessorParallelBlock-100    13.7MB ± 0%    13.7MB ± 0%  -0.10%  (p=0.000 n=10+8)

name                               old allocs/op  new allocs/op  delta
BackendProcessorParallelBlock-100     24.8k ± 0%     24.8k ± 0%    ~     (p=0.983 n=10+8)
```

After seeing this maybe it's worth just using `time.After` and fix it in the future if it's broken.

marclop:
I ran the benchmarks again and the difference is really negligible, the variance I'd attribute to my own not extremely idle machine (tried to make it as idle as possible though):

```
name                                   old time/op    new time/op    delta
BackendProcessorParallelSemSize/1-100    6.82ms ± 1%    6.88ms ± 1%   ~     (p=0.400 n=3+3)
BackendProcessorParallelSemSize/2-100    4.18ms ± 0%    4.21ms ± 0%   ~     (p=0.100 n=3+3)
BackendProcessorParallelSemSize/4-100    3.11ms ± 4%    3.05ms ± 1%   ~     (p=0.700 n=3+3)
BackendProcessorParallelSemSize/8-100    2.85ms ± 0%    2.87ms ± 0%   ~     (p=0.100 n=3+3)

name                                   old speed      new speed      delta
BackendProcessorParallelSemSize/1-100  58.5MB/s ± 1%  58.0MB/s ± 1%   ~     (p=0.400 n=3+3)
BackendProcessorParallelSemSize/2-100  95.5MB/s ± 0%  94.7MB/s ± 0%   ~     (p=0.100 n=3+3)
BackendProcessorParallelSemSize/4-100   129MB/s ± 4%   131MB/s ± 1%   ~     (p=0.700 n=3+3)
BackendProcessorParallelSemSize/8-100   140MB/s ± 0%   139MB/s ± 0%   ~     (p=0.100 n=3+3)

name                                   old alloc/op   new alloc/op   delta
BackendProcessorParallelSemSize/1-100    13.8MB ± 0%    13.8MB ± 0%   ~     (p=0.100 n=3+3)
BackendProcessorParallelSemSize/2-100    13.8MB ± 0%    13.8MB ± 0%   ~     (p=0.100 n=3+3)
BackendProcessorParallelSemSize/4-100    13.8MB ± 0%    13.8MB ± 0%   ~     (p=0.100 n=3+3)
BackendProcessorParallelSemSize/8-100    13.8MB ± 0%    13.7MB ± 0%   ~     (p=0.100 n=3+3)

name                                   old allocs/op  new allocs/op  delta
BackendProcessorParallelSemSize/1-100     24.8k ± 0%     24.8k ± 0%   ~     (p=0.300 n=3+3)
BackendProcessorParallelSemSize/2-100     24.8k ± 0%     24.8k ± 0%   ~     (p=1.000 n=3+3)
BackendProcessorParallelSemSize/4-100     24.8k ± 0%     24.8k ± 0%   ~     (p=1.000 n=3+3)
BackendProcessorParallelSemSize/8-100     24.9k ± 0%     24.9k ± 0%   ~     (p=1.000 n=3+3)
```

processor/stream/processor.go
axw: Thanks, looks good. We could add small optimisation to use `time.NewTimer` & `defer timer.Stop()` to GC the timer before the timer fires (i.e. if the queue opens back up). Not a big deal though.

marclop:
I'm closing this PR off for now since it's uncertain if we want to go this route.

Issue #6844, opened 2021-12-07
Docs: Update documentation for managing API keys to authorize requests to the APM Server


gbamparop:
Agent API key management support has been added as part of https://github.com/elastic/kibana/issues/77966 and the work has been done in https://github.com/elastic/kibana/pull/119543 and https://github.com/elastic/kibana/pull/120373.

[API keys documentation ](https://www.elastic.co/guide/en/apm/server/current/api-key.html
) in the APM server could be updated to reflect these changes.

A how to guide could also be added in the Kibana [APM docs](https://www.elastic.co/guide/en/kibana/7.16/apm-how-to.html).

The management view is located in Kibana (APM -> Settings -> Agent Keys) and users can check the privileges they want to assign to the API key being created instead of pasting them in JSON into the role descriptors box like they do in the stack management.

https://user-images.githubusercontent.com/5831975/145042317-432602f7-89ae-4f14-9ed7-1c71047ee87e.mov

Please also find some screenshots below that you might find useful:

![image](https://user-images.githubusercontent.com/5831975/145041520-e8481135-1093-433c-940e-461a18bd157b.png)

![image](https://user-images.githubusercontent.com/5831975/145041852-b0f70cc5-f324-4874-b4ec-2df0e7750cdc.png)

gbamparop:
@bmorelli25 could you please have a look at this?

simitt:
Updated and now available in https://www.elastic.co/guide/en/apm/guide/current/api-key.html#create-an-api-key.

gbamparop:
> Updated and now available in https://www.elastic.co/guide/en/apm/guide/current/api-key.html#create-an-api-key.

@simitt the docs in this link seem to suggest to create an API key through `Stack Management > API keys` instead of `APM settings -> Agent Keys` mentioned in the description.

bmorelli25:
Repeating what I [posted](https://elastic.slack.com/archives/C95SB62AG/p1651870344637949) on Slack:

@elastic/apm-server I'm confused about the `sourcemap:write` privilege. Is this privilege still required for anything? As far as I can tell, the `/assets/v1/sourcemaps` endpoint was removed in https://github.com/elastic/apm-server/pull/6447.

I tried spinning up an 8.2 standalone APM Server and when I hit `http://127.0.0.1:8200/assets/v1/sourcemaps`, I get:

```
{
  "error": "404 page not found"
}
```

[...]

PR #8111, opened 2022-05-16
docs: add APM app API key instructions

bmorelli25:
### Summary

This PR updates our APM API key instructions. The new recommended way to create an APM agent API key is with the user-friendly **APM app** > **Settings** > **Agent Keys** page.

* Closes https://github.com/elastic/apm-server/issues/6844.


docs/secure-agent-communication.asciidoc
```diff
@@ -122,28 +122,70 @@ privileges.
...
+You will need this for the next step, and you will not be able to view it again.
+
+[role="screenshot"]
+image::images/apm-ui-api-key.png[APM app API key]
+
+[[create-an-api-key-alternate]]
+[float]
+=== Create an API key in {kib} Stack Management
+
+If you don't have access to the APM app, API keys can be created in {kib} Stack Management.
```
bmorelli25, line 184: What do we think of this? There used to be cases where a user wouldn't have access to the APM app. I don't think that's the case anymore, but I wasn't 100% sure.

docs/secure-agent-communication.asciidoc
```diff
@@ -122,28 +122,70 @@ privileges.
...
+Using a superuser account, or a user with the role created in the previous step,
+open {kib} and navigate to **Observability** > **APM** > **Settings** > **Agent keys**.
+Enter a name for your API key and select at least one privilege.
+
+For example, to create an API key that can be used to ingest APM events
+and read agent central configuration, select `config_agent:read` and `event:write`.
+
+IMPORTANT: The `sourcemap:write` privilege is outdated and will be removed in a future release.
+To learn more about the privileges required to upload a source map,
+see the {kibana-ref}/rum-sourcemap-api.html[RUM source map API].
```
bmorelli25, line 172: I'm assuming we're going to open an issue to remove this privilege from the APM app. In the meantime, I've added this disclaimer.

docs/secure-agent-communication.asciidoc
axw: As you say, since the licensing changes we can safely assume everyone has access to the APM app. I think we can drop this.

axw (approved):

docs/secure-agent-communication.asciidoc
gbamparop: I am not sure if it's worth adding a note for this, but only the API keys with `metadata.application': 'apm'` are displayed in the APM app. This means that when a key is created from Stack Management, it won't be displayed in the APM app, unless the above metadata is added.

bmorelli25: > I think we can drop this.

👍 

> This means that when a key is created from Stack Management, it won't be displayed in the APM app, unless the above metadata is added.

I think it's important to mention this as we likely have a lot of users that previously created API keys but aren't seeing them in this view. I'm not sure about adding the metadata req'd though as we'd then want to doc how to add that metadata.

PR #8119, opened 2022-05-17
apmbench: Support multiple apm servers in cloud

### LINKED ISSUE #7841: apmbench: Support multi-node APM Server deployment through cloud proxy
## Description

`apmbench` should support running the benchmark scenarios against a multi-node APM Server that is load balanced by the Elastic Cloud proxy. If apmbench is run against a deployment with more than 1 APM Server, the metrics reported by `apmbench` are incorrect since the _expvar_ requests are load balanced and  the calculations are incorrect when the before / after requests land in different APM Server instances.

## Possible solution

The cloud proxy returns additional metadata in the HTTP response header to let the requester know where their request landed, an example curl request to an APM Server deployment with header information enabled shows:

```console
$ curl -i -k -H "Authorization: Bearer ${ELASTIC_APM_SECRET_TOKEN}" ${ELASTIC_APM_SERVER_URL}/debug/pprof
HTTP/2 301
content-type: text/html; charset=utf-8
date: Mon, 11 Apr 2022 07:18:06 GMT
location: /debug/pprof/
x-cloud-request-id: LXsMdTemS1OEkqBqiJWobA
x-found-handling-cluster: 3560bfe7e0f54b73bb29deba1da411c8
x-found-handling-instance: instance-0000000000
content-length: 48
....
```

We should parse the `x-found-handling-instance` header and aggregate the results per instance, rather than globally. One challenge will be to ensure that we always calculate the metrics for all instances, perhaps apmbench could try to auto discover how many instances are available for the benchmark, and if no cloud proxy headers are detected, then assume a single APM Server.

marclop:
## Motivation/summary

Adds support for benchmarking multiple APM Servers running in ESS / ECE
behind the cloud proxy. It uses the cloud proxy response headers and
auto-discovers the number of APM Servers respond to the `/debug/vars`
requests, and aggregates the final responses.

## Checklist

## How to test these changes

1. `cd testing/benchmark`
2. `make terraform.tfvars`
3. Set the following settings in `terraform.tfvars`:
```tf
apm_server_size = "4g"
apm_server_zone_count = 2
elasticsearch_size = "8g"
worker_instance_type = "c6i.2xlarge"
```
4. `make`
5. `make run-benchmark BENCHMARK_AGENTS=1000 BENCHMARK_WARMUP=0 BENCHMARK_TIME=1m`
_Warm up timeouts seem to be broken atm_ that's why they're set to 0.

## Related issues

Closes #7841


systemtest/benchtest/expvar.go
```diff
@@ -72,12 +76,54 @@ func queryExpvar(out *expvar, srv string) error {
...
+		if err != nil {
+			return err
+		}
+		if _, ok := agg[id]; ok {
+			seen++
+		}
+		agg[id] = tmp
+		// We must ensure that we have made enough requests to the remote APM
+		// Server to guarantee with a degree of certainty that all the remote
+		// APM Servers metrics have been queried.
```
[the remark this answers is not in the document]
marclop: I suppose that may be slightly more deterministic when it comes to knowing which servers accepted batches. I suppose the best course would be to store these on a per-benchmark basis?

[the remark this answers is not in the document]
marclop: I was thinking about it but I don't think your suggestions will work quite well since before we run any benchmark, we need to aggregate the initial stats from the expvar endpoint, which at that point (first benchmark) we haven't sent any data to the APM Server. I'm not sure how we'd do what you suggest given this constraint.

Also, I checked with the cloud proxy team, there isn't a way (currently) to ask the the proxy to route the request to a specific instance, although it would be possible to add it should we require it.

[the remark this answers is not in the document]
axw: @marclop ah good point, thanks for humouring me anyway :)

We might eventually need an alternative way to measure throughput, like querying Elasticsearch. We did that in hey-apm, and I moved away from it to (a) simplify things and (b) enable benchmarking when not using the Elasticsearch output. If things are going to get more complicated by only relying on expvar, though, then perhaps we should revert.

axw (approved):

systemtest/benchtest/expvar.go
```diff
@@ -98,3 +144,57 @@ func WaitUntilServerInactive(ctx context.Context, server string) error {
...
+	to.NextGC += from.NextGC
+	to.LastGC += from.LastGC
+	to.PauseTotalNs += from.PauseTotalNs
+	to.NumGC += from.NumGC
+	to.NumForcedGC += from.NumForcedGC
+	to.GCCPUFraction += from.GCCPUFraction
+}
+
+func aggregateLibbeatStats(from LibbeatStats, to *LibbeatStats) {
+	to.ActiveEvents += from.ActiveEvents
```
lahsivjar, line 182: @marclop (Apologies for putting the message after the PR is reviewed and merged)

IIUC, some of these metrices are monotonic counters (strictly increasing for the lifetime of the apm-server process). In this case since we are aggregating different servers, it might result in a case where taking a diff between 2 timestamps (t0, t1 where t1 > t0) will result in an unexpected negative value if query over one expvar fails or one apm-server just dies. Is this correct? Do we need to take this into account?

marclop: > t might result in a case where taking a diff between 2 timestamps (t0, t1 where t1 > t0) will result in an unexpected negative value if query over one expvar fails or one apm-server just dies.

as you mention, the negative result would happen if an APM Server has been restarted (due to running out of memory). We could implement some protection against this by writing more logic that protects against this case. We can open an improvement issue to change the way this works to account for this case.

I think you mean to say where 2 timestamps (t0, t1 where t0 > t1), not the opposite, right?

lahsivjar: I meant `t1 - t0`

lahsivjar: Makes sense, we can add some protection to protect us from these cases. BTW, I think in the future we will have to look at a different way to get these metrices (we can discuss this later).

PR #8120, opened 2022-05-17
apmbench: Remove `b.Error()` from BenchmarkAgent

marclop:
## Motivation/summary

Removes the `b.Error()` calls in all `BenchmarkAgent*` benchmarks since
they are prone to make the whole benchmark fail when the APM Server
starts being overwhelmed. This defeats the purpose of `apmbench` since
we want to be able to push the APM Server as much as we want and we rely
on the `expvar` metrics for throughput calculations.

## Checklist

~- [ ] Documentation has been updated`~

## How to test these changes

N/A

## Related issues

Part of #7540


axw (approved):
LGTM. I guess I was overcautious; we calculated client-side throughput in hey-apm, ignoring transmission errors IIRC. This seems sensible: throughput should drop if there are errors.

marclop:
@axw I would've liked to keep the errors as some sort of logs or warnings using `b.Log` calls, but unfortunately, the messages won't be printed anywhere, thus my preference for removing it particularly since the throughput calculations are driven by metrics in the APM Server.

PR #8122, opened 2022-05-17
make: Rename `go.make` to `go.mk`

marclop:
## Motivation/summary

Renames to `go.mk` so the IDEs and editors can do syntax highlighting.


axw (approved):
Thanks :)
I instinctively looked for `go.mk` the other day, should have opened a PR then.

PR #8132, opened 2022-05-18
Use modelindexer instead of libbeat ES output in tests

axw:
## Motivation/summary

- Introduce package modelindexertest, which provides some test functions related to modelindexer and mocking Elasticsearch `_bulk` request handling
- Stop using libbeat/esleg/eslegclient in tests for encoding events
- In beater integration tests configure the elasticsearch output (which uses modelindexer) rather than using "chanClient" (now removed), exercising more production code. The instrumentation tests have changed a little as a result, as using the elasticsearch output means we now have additional trace events.

## Checklist

## How to test these changes

N/A

## Related issues

https://github.com/elastic/beats/pull/31553 broke our tests. We should be using modelindexer anyway.


beater/integration_test.go
```diff
@@ -32,39 +32,51 @@ import (
...
+	if err != nil {
+		panic(err)
+	}
+	if time.Since(timestamp) < 5*time.Minute {
+		var err error
+		doc, err = sjson.SetBytes(doc, "@timestamp", timestampOverride.Format(timestampFormat))
+		if err != nil {
+			panic(err)
+		}
+		if gjson.GetBytes(doc, "processor.name").String() != "metric" {
```
stuartnelson3, line 59: these seem like two very helpful packages

stuartnelson3 (approved):
looks good

model/modelindexer/modelindexertest/modelindexertest.go
```diff
@@ -0,0 +1,129 @@
...
+			panic(err)
+		}
+		var actionType string
+		for actionType = range action {
+		}
+		if !scanner.Scan() {
+			panic("expected source")
+		}
+
+		doc := append([]byte{}, scanner.Bytes()...)
```
marclop, line 88: cool way to copy the slice, I would've done something much longer like checking the length of the scanned bytes, creating a `[]byte` with that length and using `copy`.

model/modelindexer/modelindexertest/modelindexertest.go
```diff
@@ -0,0 +1,129 @@
...
+		if !json.Valid(doc) {
```
marclop, line 89: TIL

marclop (approved):
LGTM! `modelindexertest` is pretty neat.

PR #8138, opened 2022-05-18
apmbench: Adjust warmup timeout to account agents

marclop:
## Motivation/summary

Changes the calculation of the warmup timeout to add more weight when
a higher number of agents is used.

## Checklist

## How to test these changes

1. `cd testing/benchmark`
2. `make`
3. `make run-benchmark BENCHMARK_AGENTS=1000 BENCHMARK_WARMUP=1000 BENCHMARK_TIME=1m`


systemtest/benchtest/main.go
```diff
@@ -239,13 +239,15 @@ func Run(allBenchmarks ...BenchmarkFunc) error {
...
 	// set the context timeout based on this ingest rate, or if lower, default
-	// to 15 seconds. The default 5000 / 2000 ~= 2, so the default 15 seconds
+	// to 15 seconds. The default 5000 / 1000 ~= 5, so the default 15 seconds
 	// will be used instead. Additionally, the number of agents is also taken
 	// into account, since each of the agents will send the number of specified
 	// events to the APM Server, increasing its load.
-	timeout := warmupTimeout(2000, events, maxEPM, agents)
+	// Also account the GOMAXPROCS setting, since it will dictate the goroutines
+	// that are scheduled at any time.
+	timeout := warmupTimeout(1000, events, maxEPM, agents, runtime.GOMAXPROCS(0))
```
marclop, line 250: I'm starting to wonder if this logic is just too complicated and we may be better off limiting the maximum number of agents we warm up with?

marclop:
This PR fixes the flakiness seen in the tests by removing those tests since they don't seem to be deterministic when running in the CI machine, it may be good to merge the PR for now to unblock other PRs and restore system test and CI stability. However, we may want to revisit the warmup strategy and limit the number of agents and all these timeout calculations, I fear I've overcomplicated things unnecessarily.

axw (approved):
I don't fully grok all the changes, but if it unblocks CI :+1:

I don't understand why we need to take CPUs into account. Isn't the number of agents independent of that? I mean, you can have more agents (goroutines) running concurrently than there are CPUs. +1 to simplifying this logic if we can.

systemtest/benchtest/main.go
lahsivjar: My hypothesis here is that the warmup timeout calculation is overlooking the overhead of having a high number of agents and If the hypothesis is correct I don't think we need to take the GOMAXPROCS into account. Explanation:

`timeout = math.Max(events/rate*agents, 15) seconds`

Based on the above formula, timeout is proportional to the number of agents as well as number of events. But the math.Max introduces a lower bound of 15 seconds which would be applicable for smaller events and larger agent count. For example, at one extreme, having 1 event and 20000 agents will also give us 15 seconds timeout which will imply that apm-server has to process 20000 requests of 1 event each in 15 seconds.

We can test the hypothesis by fixing the agents to a high number and using lower and higher value events. I tested it with 80 agents and 1 event vs 80 agents and 1000 events. The first one failed with deadline exceeded (calculated timeout = 15 seconds) while the latter passed (calculated timeout = 40 seconds).

One way to address this (assuming the hypothesis is correct) would be to assign a higher weight to the agents.

Issue #7823, opened 2022-04-08
docs: update service name incorrect


graphaelli:
While answering https://discuss.elastic.co/t/how-to-delete-amp-service/301908/2 I noticed https://www.elastic.co/guide/en/apm/server/current/update-existing-data.html references `context.service.name` instead of `service.name`.

graphaelli:
Per that discussion, we need to update the docs to handle the data streams case as well.  I think `.ds-*-apm` would be reasonable.  `apm-agent-configuration,.apm-custom-link` could also use the same treatment though in 8.0+ those are system indices I'd recommend only adding those to the 7.x instructions.

PR #8145, opened 2022-05-18
[7.17] docs: how to delete other APM data

bmorelli25:
### Summary

This PR adds two new examples to the 7.17 **Delete data** page:

* Delete all custom link documents
* Delete all central config documents

For https://github.com/elastic/apm-server/issues/7823.


axw (commented):
I think this should be fine, but we'd better check with the UI team that there are no gotchas like cross-references in Kibana saved objects.

@sqren is it safe to delete docs from these indices like this?

sqren (approved):
Thank you!

docs/legacy/storage-management.asciidoc
```diff
@@ -248,6 +248,38 @@ POST /apm-*/_delete_by_query
...
+To delete all central agent configurations, run:
+
+[source,console]
+----
+POST /.apm-agent-configuration/_delete_by_query
+{
+  "query": {
+    "match_all": {}
+  }
+}
```
sqren, line 280: Agent configurations are sync'd with fleet so deleting from here might not be enough. I'll have to go check when and how the syncing mechanism works.

sqren: Upon creating/deleting an agent config a call is made to sync the state to fleet: 
https://github.com/elastic/kibana/blob/7591fb61556bc56aaae725d6db51bd80e958d91b/x-pack/plugins/apm/server/routes/settings/agent_configuration/route.ts#L115-L125

If the user simply deletes the agent config index the sync will not happen until they manually create a new agent config. This means that agents will still receive the old agent configs (via fleet), even though the agent config index was created. I can see how this can be confusing.

I'm not sure what the best solution would be. We could sync the agent configs periodically (eg. every hour), or when Kibana starts.

bmorelli25: Ahh. Interesting. I suppose we could document that the user needs to create a temporary agent config value to sync the index. I assume deleting the entire index is only useful for users with a large number of configuration settings, so deleting the index and creating a single config could be faster than clicking all of the `trash` icons?

bmorelli25: Thanks! Fixed in [`102ab5a`](https://github.com/elastic/apm-server/pull/8145/commits/102ab5aa2a3cd29482118c9e16f19cd3f720cf52). Appreciate the help; it's good to have you back.

sqren (approved):

Issue #8018, opened 2022-05-02
Traces in Kibana are not data streams, ILM policies don't apply, filling up disk space


slhck:
**APM Server version** apm-server version 8.1.3 (amd64), libbeat 8.1.3 [1c7bc9e0bb5a3ab54cb67a881c49b4fda7a62241 built 2022-04-19 07:04:09 +0000 UTC]

**Description of the problem including expected versus actual behavior**:

I have two ELK clusters that I have upgraded from 7.x to 8.x using the recommended migration guides. I have set up APM Server binary with a few agents sending data using Ruby and Python libraries. Both operate fine, except that the APM traces are now filling up the disk space.

Here is the index:

![image](https://user-images.githubusercontent.com/582444/166216201-7179be74-3cdd-4e09-a388-77a8212c86f7.png)

Here are the templates:

![image](https://user-images.githubusercontent.com/582444/166216267-88d9bbc8-c901-4f6a-b557-e4a9f6ddd205.png)

As you can see, the default policy for traces is not applied:

![image](https://user-images.githubusercontent.com/582444/166216336-6c19d909-1e15-4ba0-bd4a-3ee81070ad09.png)

I am unable to create my own ILM policy with rollup, since the traces-apm-default index is just an index, not an index pattern with different underlying indices; just one huge index that eats up disk space.

I have tried [migrating to data stream](https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-migrate-to-data-stream.html#indices-migrate-to-data-stream-request) but it fails with:

```
➜ curl -k -XPOST -u elastic:XXXXX https://localhost:9200/_data_stream/_migrate/traces-apm-default
{"error":{"root_cause":[{"type":"illegal_argument_exception","reason":"alias [traces-apm-default] does not exist"}],"type":"illegal_argument_exception","reason":"alias [traces-apm-default] does not exist"},"status":400}%                                                                                                                                                                
```

Also, using `traces-apm*` does not work:

```json
[...]

PR #8152, opened 2022-05-19
Prevent index failed precondition

stuartnelson3:
this patch closes the `publishReady` channel only
if all preconditions have been met. Previously,
any result from the waitReady check would result
in unblocking the roundtripper that was supposed
to prevent indexing until all preconditions had
been met.

when the server is shut down, the ctx in
s.waitReady is canceled. this caused the function
to return and execute its defer statement, closing
`publishReady`.  this resulted in any queued events
being sent to the elasticsearch, which is not the
desired behavior.

closes https://github.com/elastic/apm-server/issues/8018

# how to test

1. start elasticsearch/kibana
2. navigate to fleet ui / agent policies; remove apm-1
3. go to add integrations / apm / settings; remove the integration
4. start apm server; verify preconditions aren't met from log output
5. send events to apm-server
6. stop apm-server
7. verify no events were indexed


beater/waitready.go
```diff
@@ -79,10 +81,17 @@ func waitReady(
...
 }
 
 func (c *waitReadyRoundTripper) RoundTrip(r *http.Request) (*http.Response, error) {
 	select {
+	case <-c.drain:
+		return &http.Response{
+			Request:    r,
+			StatusCode: http.StatusPreconditionFailed,
+			Body:       io.NopCloser(new(bytes.Buffer)),
+		}, nil
```
stuartnelson3, line 94: choosing StatusPreconditionFailed was somewhat arbitrary; it stops the es client from retrying (I originally was returning an error, but then the client would retry X times) and has the word "precondition". Happy to change this to something else.

axw: I don't think it matters a great deal, as this is all internal to the server, but I'd prefer not to use `http.StatusPreconditionFailed` as it's a client error (4xx). Perhaps use `http.StatusInternalServerError`?

I think ideally we would return an error. It's possible to customise the retry-on-error behaviour of go-elasticsearch, but it's probably a bit of a rigmarole just for this?

beater/waitready.go
```diff
@@ -79,10 +81,17 @@ func waitReady(
...
+	drain <-chan struct{}
 }
 
 func (c *waitReadyRoundTripper) RoundTrip(r *http.Request) (*http.Response, error) {
 	select {
+	case <-c.drain:
+		return &http.Response{
+			Request:    r,
+			StatusCode: http.StatusPreconditionFailed,
+			Body:       io.NopCloser(new(bytes.Buffer)),
```
axw, line 93:
```suggestion
			Body:       io.NopCloser(strings.NewReader("server shutting down")),
```

might be a bit more understandable in logs?

axw (approved):
LGTM! Just a couple of minor things.

stuartnelson3:
@axw surprisingly little rigamarole, actually. let me know if this looks good to you!

axw (approved):
Nice :) LGTM, thanks!

axw:
Verified with 8.3.0 BC2. I ran the stack locally, and ran the apm-server image locally. Starting apm-server, sending some data, and stopping the server did not lead to any indices being created. OTOH, running with 8.2.0 did produce indices.

PR #8155, opened 2022-05-19
Allow indexing of empty service.{target, name}

lahsivjar:
## Motivation/summary

As per the spec for `service.target.{type, name}`, it is expected to index an empty value for the fields `service.target.type` and `service.target.name`.

Full spec list available at https://github.com/elastic/apm/pull/627

## Checklist

## Related issues

Fixes specs implemented in #7734


model/service_test.go
```diff
@@ -90,6 +90,55 @@ func TestServiceTransform(t *testing.T) {
...
+			},
+		},
+		{
+			Service: Service{
+				Name: serviceName,
+				Node: ServiceNode{Name: serviceNodeName},
+				Target: &ServiceTarget{
+					Name: "",
+					Type: "",
+				},
```
axw, line 135: Can this really happen? If it can't, maybe we should tighten up the JSON-Schema so constraints so if target is specified, either field must be present.

axw (commented):
Code looks good, I just have a question about the possibility of type and name both being empty. I don't _think_ it should be possible from looking at the spec.

model/service_test.go
lahsivjar: Hmm, you are correct. There will be 2 cases:

1. Older agents will not be sending us target information so target itself will be `nil` however it will be inferred by the resource field => as per the spec this is where the `target.type` can be nil but both `name` and `type` shouldn't be `nil`.
2.  For newer agents, as per the spec, `service.target.type` should always be present, if it is not present then even name is ignored.

Based on this I think we can add a required constraint on `target.type` if target is provided.

axw: That sounds good to me. I don't know if it needs to be captured in the spec or not, but can you please run it by the agent devs to make sure they're at least aware?

lahsivjar: Updated the code to validate empty target (without both `type` or `name` provided)

axw (approved):
LGTM, thanks!

trentm:
Perhaps curious that this resulted in "synchronize schema spec" PRs *twice* to each agent repo.

PR #8158, opened 2022-05-19
Fix Destination target fields legacy regex

SylvainJuge:
The regex previously defined (has now been fixed in https://github.com/elastic/apm/pull/627/commits/45c44ed6b01ca44a09f8481c175fa249fcdb8272) in the https://github.com/elastic/apm/pull/627 to infer the new `service.target.type` and `service.target.name` fields from `destination.service.resource` was not working as expected for `H2` database.

It also impacts other types of databases from the ones listed in https://github.com/elastic/apm/blob/main/tests/agents/json-specs/span_types.json  we currently have `db2`, `h2` and `sqlite3` for which this bug will be triggered.

Without this change, the spans sent by agents that DO NOT send the new `service.target.*` fields but with `destination.service.resource` set to `h2` or `h2/my-db` will have their service target fields incorrectly set
- `service.target.type` is absent, whereas `h2` value would be expected
- `service.target.name` should be either absent (when `resource` = `h2`) or equal to `my-db` (when `resource is `h2/my-db`.

## How to test these changes

- [ ] capture some DB spans with [opbeans-java](https://github.com/elastic/opbeans-java) application that uses `h2` as database by default.

## Related issues

https://github.com/elastic/apm/pull/627


model/modeldecoder/v2/decoder.go
```diff
@@ -72,7 +72,7 @@ var (
 	// reForServiceTargetExpr regex will capture service target type and name
 	// Service target type comprises of only lowercase alphabets
 	// Service target name comprises of all word characters
-	reForServiceTargetExpr = regexp.MustCompile(`^([a-z]+)(?:/(\w+))?$`)
+	reForServiceTargetExpr = regexp.MustCompile(`^([a-z][a-z0-9]*)(?:/(\w+))?$`)
```
marclop, line 75: I'm not a regex expert, but isn't this expression equivalent to `w+`?

lahsivjar: `\w` would also include uppercase letters.

lahsivjar: Do we have a specific set of middleware we are trying to capture by using this regex? I am concerned that there might be some other targets which might be an outlier, from the top of my mind I think there is a RDBMS named 4d which will not fit in the above regex.

Assuming we are converting the target type to lowercase somewhere in the pipeline `[a-z0-9]+` might work in most scenarios. WDYT?

lahsivjar: > Do we have a specific set of middleware we are trying to capture by using this regex?

Got answer to this one from the link in the description, still WDYT about having a more loose regex as `[a-z0-9]+` for capturing service type?

SylvainJuge: +1 on loosening it to `[a-z0-9]+`, that's simpler, I'll update the spec with that.

For reference, here is the quite exhaustive list of databases that are listed in the [OpenTelemetry semantic conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/database.md), while there is none that starts with a number like `4d` there might be in the future.

lahsivjar, line 75:
```suggestion
	reForServiceTargetExpr = regexp.MustCompile(`^([a-z0-9]+)(?:/(\w+))?$`)
```

lahsivjar (approved):

PR #8174, opened 2022-05-20
Collect aggregates from expvar metrices

### LINKED ISSUE #7817: benchmarking: Record more APM Server usage metrics
## Description

We currently record a few metrics related to the APM Server's event processing and golang memory allocations. We're still missing some more metrics to help us collect data points that are useful as it comes to sizing APM Server and understanding its memory usage.

We should aim to record (ideally in a histogram so we can see distributions):

* RSS memory consumption 
* Goroutines
* Available indexers

GC metrics ([Go docs](https://pkg.go.dev/runtime/metrics)): 

* GC cycles
* Heap information

lahsivjar:
## Motivation/summary

Collect aggregated metrics from the result of expvar. Currently collecting max, min and average.

## Checklist

## How to test these changes

1. Run the `apmtest` benchmark
2. Run apmbench test with `-detailed` flag
3. Observe the following fields in the output: `gc_cycles`, `max_rss`, `max_goroutines`, `max_heap_alloc`, `max_heap_objects` and `mean_available_indexers`

## Related issues

Closes #7817


axw (commented):
@lahsivjar @marclop what do you think about making these metrics opt-in?

I fear the output will become unwieldy if we add so many metrics. Do you anticipate using these regularly, or only (interactively) after diagnosing a drop in throughput?

lahsivjar:
@axw I agree with the output getting unwieldy. For automated benchmark, IMO, we should always index the results to ES but for other use cases around manual benchmarking I think optional metric opt-in flag for extra metrics will be really useful. @marclop WDYT?

marclop:
@axw I think that would be better and we should consider categorizing the metrics in essential and optional categories. This may improve the output and the mental strain of parsing all these number thrown at you. I should've added some of these ideas to the the original issue when I created it and start the discussion then.

There are some metrics that aren't really that useful, particularly when the number is 0, one that comes to mind is `error_responses/sec`.

Since we mimick the output of `go test -bench`, we have certain constraints on what we need to publish (`name`, `iterations` and `ns/op`), additionally we print `B/op` and `allocs/op` which would be the metrics reported when `-benchmem` is set in `go test -bench`.

Of our custom metrics, we don't really need all of them by default, if we're just doing some quick benchmarks, it may be better to limit the number of other metrics that are printed out. I believe that it would be more effective to print less metrics by default and omit those that are 0. the output of these could be as below (the final number of metrics to be printed is yet TBD)

```
$ ./apmbench
benchmarkName   100   1000 ns/op  6000 events/sec   1276948 B/op   2831 allocs/op
```

In case that there are some error responses, we could report those, but they needn't be included in all the results:

```
$ ./apmbench
benchmarkName   100   1000 ns/op   1 error_responses/sec  6000 events/sec   1276948 B/op   2831 allocs/op
```

We can introduce a new flag that when set, includes all the metrics that we're recording to be displayed `-detailed` or a similar flag name could work:

```
$ ./apmbench -detailed
benchmarkName   100   1000 ns/op   1 error_responses/sec  6000 events/sec   1000 errors/sec 1000 metrics/sec   1000 txs/sec   3000 spans/sec   10 gc/op   1276948 B/op   2831 allocs/op ...
```

I think that the usefulness of the broken down `event/s` (metrics, errors, txs, spans) metrics is not immediately obvious and we should most likely omit that to free some space.

| metric name | Required | Optional |
|----|----|----|
| Benchmark Name | x | - |
| Iterations | x | - |
| ns/op |  | - |
| B/op | x | - |
| allocs/op | x | - |
| events/sec | x | - |
| events/sec | x | - |
| events/sec | x | - |
| events/sec | x | - |

axw:
Thanks for the proposal @marclop, that all sounds good to me.

systemtest/benchtest/expvarmetrics/metrics.go
```diff
@@ -0,0 +1,253 @@
...
+//
+// Watchers are one-time hooks into collector to know about when
+// the a specific state of the metric is observed by the collector.
+type Collector struct {
+	l         sync.Mutex
+	serverURL string
+	period    time.Duration
+	metrics   map[Metric]*AggregateStats
+	watchers  map[Metric]*watchItem
+}
```
marclop, line 76: I think we can simplify this structure a little bit and only store what's necessary. `serverURL` and `period` are only used once when the collector is initialized, we can just pass these to the functions that require them.

Additionally, I see no reason to pollute the heap with AggregateStats and watchItems, since it's relatively easy to access or pass around the values to be updated within the collector.

Last, `watches` seems like a better name than `watchers`.

The `closed` boolean is used later on when the watches are closed and the loop function isn't running anymore (See comments below, towards the end of the review).

```suggestion
// Watches are one-time hooks into collector to know about when
// the a specific state of the metric is observed by the collector.
type Collector struct {
	l       sync.Mutex
	metrics map[Metric]AggregateStats
	watches map[Metric]watch
	closed  bool
}
```

systemtest/benchtest/expvarmetrics/metrics.go
```diff
@@ -0,0 +1,253 @@
...
+
+// Get returns the aggregated stat of a given metric.
+func (c *Collector) Get(m Metric) AggregateStats {
+	c.l.Lock()
+	defer c.l.Unlock()
+
+	stats := c.metrics[m]
+	statsCopy := *stats
+	return statsCopy
```
marclop, line 97:
```suggestion
	return c.metrics[m]
```

systemtest/benchtest/expvarmetrics/metrics.go
```diff
@@ -0,0 +1,253 @@
...
+// StartNewCollector creates a new collector and starts
+// querying expvar endpoint at the specified interval.
+func StartNewCollector(ctx context.Context, serverURL string, period time.Duration) (*Collector, error) {
+	c := &Collector{
+		serverURL: serverURL,
+		period:    period,
+		metrics:   make(map[Metric]*AggregateStats),
+		watchers:  make(map[Metric]*watchItem),
+	}
+	return c, c.start(ctx)
```
marclop, line 87:
```suggestion
	return c, c.start(ctx, serverURL, period)
```

systemtest/benchtest/expvarmetrics/metrics.go
```diff
@@ -0,0 +1,253 @@
...
+	stats.Last = value
+	stats.sampleCount += 1
+	stats.Mean += (float64(value) - stats.Mean) / float64(stats.sampleCount)
+	if stats.Min > value {
+		stats.Min = value
+	}
+	if stats.Max < value {
+		stats.Max = value
+	}
+}
```
marclop, line 190: What do you think about changing the function signature to return the updated stats aggregated. It seems like it would work slightly better if we don't store the pointer to the structure in the heap. Since we've got to hold the mutex anyway, we may as well then replace the entry in the map with the return value.

```suggestion
func (c *Collector) updateMetric(stats AggregateStats, value int64) AggregateStats {
	// Initialize for first value
	if stats.samples == 0 {
		stats.First, stats.Min, stats.Max = value, value, value
	}

	stats.Last = value
	stats.samples++
	stats.Mean += (float64(value) - stats.Mean) / float64(stats.samples)
	if stats.Min > value {
		stats.Min = value
	}
	if stats.Max < value {
		stats.Max = value
	}
	return stats
}
```

systemtest/benchtest/expvarmetrics/metrics.go
```diff
@@ -0,0 +1,253 @@
...
+	c.processMetric(MemBytes, int64(e.TotalAlloc))
+	c.processMetric(HeapAlloc, int64(e.HeapAlloc))
+	c.processMetric(HeapObjects, int64(e.HeapObjects))
+}
+
+func (c *Collector) processMetric(m Metric, val int64) {
+	if _, ok := c.metrics[m]; !ok {
+		c.metrics[m] = &AggregateStats{}
+	}
+	c.updateMetric(c.metrics[m], val)
```
marclop, line 165: What do you think about changing this to not use the pointer approach? Looks slightly more idiomatic and easier to parse:

```suggestion
	stats := c.metrics[m]
	c.metrics[m] = c.updateMetric(stats, val)
```

systemtest/benchtest/expvarmetrics/metrics.go
```diff
@@ -0,0 +1,253 @@
...
+			case <-errChan:
+				c.rejectWatchers()
+				return
+			case m := <-outChan:
+				c.addExpvar(m)
+			}
+		}
+	}()
+	return nil
+}
```
marclop, line 225: I think we can modify the signature to accept the `serverURL` and `period` to avoid storing them in the Collector structure, since those values aren't used elsewhere.

Also, the `rejectWatchers()` is called twice on error or context cancellation, we can defer calling that function inside the goroutine so the code doesn't have to be repeated.

```suggestion
func (c *Collector) start(ctx context.Context, serverURL string, period time.Duration) error {
	var first expvar
	if err := queryExpvar(ctx, &first, serverURL); err != nil {
		return err
	}
	c.addExpvar(first)

	outChan, errChan := run(ctx, serverURL, period)
	go func() {
		defer c.closeWatches()
		for {
			select {
			case <-ctx.Done():
				return
			case <-errChan:
				return
			case m := <-outChan:
				c.addExpvar(m)
			}
		}
	}()
	return nil
}
```

systemtest/benchtest/expvarmetrics/metrics.go
```diff
@@ -0,0 +1,253 @@
...
+
+func (c *Collector) rejectWatchers() {
+	c.l.Lock()
+	defer c.l.Unlock()
+
+	for _, watcher := range c.watchers {
+		watcher.notifyChan <- false
+	}
+}
```
marclop, line 199: Since this function is only expected to be called once (from what I can read in the current logic) and its purpose is signal to the channel receiver that the watch isn't going to be updated anymore, I think it would be better to close all the watches's channels to signal this. Otherwise, clients that might be receiving the channel in a loop will never exit (not that we do that now, but provides some safety).

Then, since we're closing the channel and is an operation which can only be performed once (at the risk of panic), we can add a `closed` boolean to the Collector structure to signal this.

```suggestion
func (c *Collector) closeWatches() {
	c.l.Lock()
	defer c.l.Unlock()

	if c.closed {
		return
	}
	for _, watcher := range c.watches {
		close(watcher.notifyChan)
	}
	c.closed = true
}
```

systemtest/benchtest/expvarmetrics/metrics.go
```diff
@@ -0,0 +1,253 @@
...
+				if err != nil {
+					errChan <- err
+					return
+				}
+				outChan <- e
+			}
+		}
+	}()
+	return outChan, errChan
+}
```
marclop, line 253: Couple of changes to always stop the ticker and close the channel in the defer call.

Also, the local `expvar` shouldn't be sent over to `outChan` when `err != nil`.

```suggestion
func run(ctx context.Context, serverURL string, period time.Duration) (<-chan expvar, <-chan error) {
	outChan := make(chan expvar)
	errChan := make(chan error)

	go func() {
		ticker := time.NewTicker(period)
		defer func() {
			ticker.Stop()
			close(outChan)
		}()
		query := func(e *expvar) error {
			ctx, cancel := context.WithTimeout(ctx, period)
			defer cancel()
			return queryExpvar(ctx, e, serverURL)
		}
		for {
			select {
			case <-ctx.Done():
				return
			case <-ticker.C:
				var e expvar
				if err := query(&e); err != nil {
					errChan <- err
					continue
				}
				outChan <- e
			}
		}
	}()
	return outChan, errChan
}
```

systemtest/benchtest/expvarmetrics/metrics_test.go
```diff
@@ -0,0 +1,85 @@
...
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestStart(t *testing.T) {
+	server := getTestServer(t)
+	defer server.Close()
+	ctx, cancel := context.WithCancel(context.Background())
```
marclop, line 34: I think it would be better to update the `context.WithCancel` calls in these tests to `context.WithTimeout` with a 1s timeout to avoid deadlocks in the tests

systemtest/benchtest/expvarmetrics/metrics_test.go
```diff
@@ -0,0 +1,85 @@
...
+	return httptest.NewServer(
+		http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			if r.URL.Path != "/debug/vars" {
+				t.Errorf("unexpcted path: %s", r.URL.Path)
+			}
+			w.WriteHeader(http.StatusOK)
+			w.Write([]byte(fmt.Sprintf(`{"beat.runtime.goroutines": %d}`, count)))
+			count++
+		}),
+	)
```
marclop, line 84:
```suggestion
	var count uint64
	return httptest.NewServer(
		http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
			if r.URL.Path != "/debug/vars" {
				t.Errorf("unexpected path: %s", r.URL.Path)
			}
			w.WriteHeader(http.StatusOK)
			w.Write([]byte(fmt.Sprintf(`{"beat.runtime.goroutines": %d}`, atomic.AddUint64(&count, 1))))
		}),
	)
```

systemtest/benchtest/expvarmetrics/metrics.go
```diff
@@ -0,0 +1,253 @@
...
+// lifecycle of a watcher.
+func (c *Collector) AddWatch(m Metric, expected int64) (<-chan bool, error) {
+	c.l.Lock()
+	defer c.l.Unlock()
+
+	if _, ok := c.watchers[m]; ok {
+		return nil, errors.New("watcher already exists for the given metric")
+	}
+
+	out := make(chan bool)
```
marclop, line 127: We should change the channel to be buffered, in case the condition is reached, but there aren't any receivers actively listening for the change, otherwise, when the condition is reached, it'll block forever on the `watcher.notifyChan <- true` statement.

```suggestion
	// Create a buffered channel in case there aren't any receivers when the
	// condition is reached.
	out := make(chan bool, 1)
```

systemtest/benchtest/expvarmetrics/metrics_test.go
```diff
@@ -0,0 +1,85 @@
...
+	assert.NoError(t, err)
+	select {
+	case w := <-watcher:
+		assert.True(t, w)
+		assert.GreaterOrEqual(t, c.Get(Goroutines).Last, int64(5))
+	case <-time.After(100 * time.Millisecond):
+		assert.Fail(t, "timed out while waiting for watcher")
+	}
+}
+
```
marclop, line 72: Can we add another test to verify that the watcher channel returns false when the context is cancelled:

```go
func TestWatchFalse(t *testing.T) {
	server := getTestServer(t)
	defer server.Close()
	ctx, cancel := context.WithTimeout(context.Background(), 20*time.Millisecond)
	defer cancel()
	c, _ := StartNewCollector(ctx, server.URL, 10*time.Millisecond)
	watcher, err := c.AddWatch(Goroutines, 20)
	require.NoError(t, err)
	select {
	case w := <-watcher:
		assert.False(t, w)
	case <-time.After(100 * time.Millisecond):
		assert.Fail(t, "timed out while waiting for watcher")
	}
}
```

We should also have a test that asserts that the entire watch processing isn't blocked when no receivers are set:

```go
func TestWatchNonBlocking(t *testing.T) {
	server := getTestServer(t)
	defer server.Close()
	timeout := 100 * time.Millisecond
	ctx, cancel := context.WithTimeout(context.Background(), timeout)
	defer cancel()
	c, _ := StartNewCollector(ctx, server.URL, 10*time.Millisecond)
	watcher, err := c.AddWatch(Goroutines, 5)
	require.NoError(t, err)

	// Wait for the context to be cancelled.
	time.Sleep(timeout)

	// Verify the values are updated even when there isn't an active receiver
	// on the watch.
	assert.GreaterOrEqual(t, c.Get(Goroutines).Max, int64(5))
	select {
	case w := <-watcher:
		assert.True(t, w)
	case <-time.After(100 * time.Millisecond):
		assert.Fail(t, "timed out while waiting for watcher")
	}
}
```

marclop (commented):
Overall approach looks good. I left a few comments on the overall watcher and map structure.

systemtest/benchtest/expvarmetrics/metrics.go
lahsivjar: ACK, looks cleaner

systemtest/benchtest/expvarmetrics/metrics.go
lahsivjar: I have updated the PR with code to delete watches, let me know what you think.

systemtest/benchtest/main.go
```diff
@@ -72,39 +78,47 @@ func runBenchmark(f BenchmarkFunc) (testing.BenchmarkResult, bool, error) {
...
-func addExpvarMetrics(result testing.BenchmarkResult, before, after expvar) {
-	result.MemAllocs = after.MemStats.Mallocs - before.MemStats.Mallocs
-	result.MemBytes = after.MemStats.TotalAlloc - before.MemStats.TotalAlloc
-	result.Bytes = after.UncompressedBytes - before.UncompressedBytes
-	result.Extra["events/sec"] = float64(after.TotalEvents-before.TotalEvents) / result.T.Seconds()
-	result.Extra["txs/sec"] = float64(after.TransactionsProcessed-before.TransactionsProcessed) / result.T.Seconds()
-	result.Extra["spans/sec"] = float64(after.SpansProcessed-before.SpansProcessed) / result.T.Seconds()
-	result.Extra["metrics/sec"] = float64(after.MetricsProcessed-before.MetricsProcessed) / result.T.Seconds()
-	result.Extra["errors/sec"] = float64(after.ErrorsProcessed-before.ErrorsProcessed) / result.T.Seconds()
+func addExpvarMetrics(result testing.BenchmarkResult, collector *expvar.Collector, detailed bool) {
```
marclop, line 97: This function is starting to contain ad decent amount of logic, could you write a few tests that cover the basic logic and the detailed flag?

systemtest/benchtest/main.go
```diff
@@ -89,12 +94,12 @@ func runBenchmark(f BenchmarkFunc) (testing.BenchmarkResult, bool, error) {
...
 	})
 	if result.Extra != nil {
-		addExpvarMetrics(result, collector, *detailed)
+		addExpvarMetrics(&result, collector, *detailed)
 	}
 	return result, ok, nil
 }
 
-func addExpvarMetrics(result testing.BenchmarkResult, collector *expvar.Collector, detailed bool) {
+func addExpvarMetrics(result *testing.BenchmarkResult, collector metricsCollector, detailed bool) {
```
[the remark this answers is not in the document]
lahsivjar: Hmm, writing tests with mocking HTTP calls is too brittle. It takes away the readability and adds a lot of "extra" dependency on the test making it hard to maintain long term. Additionally, I believe for unit testing this piece of code it doesn't need to explicitly define the whole logic for how expvar is collected. So comparing a simple interface vs the above, I chose the former. Also, golang interfaces are designed to be slimmer than other OOP so there's not that much of an overhead. WDYT?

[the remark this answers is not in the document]
marclop: I'd rather we introduced a more complete test for this particular piece of code rather than mock the entire surface. It will assert the correctness of the Calculations provided by `expvar.Collector`. Mocking the collector will only verify that the returned values are correct as per the mock.

I disagree that mocking using HTTP calls is too brittle and that it takes away readability. On the contrary, `httptest` provides all that's needed to mock the necessary expvar calls, and makes the collector interaction with the APM Server more obvious and easier to understand when the tests are read.

Looking at how the `expvar.Collector` works, it's shouldn't be difficult to write the necessary httptest.Server and assert its behavior in this case.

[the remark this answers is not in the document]
lahsivjar: > It will assert the correctness of the Calculations provided by expvar.Collector.

Ah, I thought we were unit testing the `addExpvarMetrics`, do we intend to test it end to end here?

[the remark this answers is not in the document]
marclop: We aren't overly strict at limiting the scope and surface of the unit tests. We tend not to over-abstract components just to be able to test things in isolation as long as they are reasonably easy to mock and test together. For example https://github.com/elastic/apm-server/blob/main/model/modelindexer/indexer_test.go tests slightly more than what you'd expect strictly by the definition of "unit" tests.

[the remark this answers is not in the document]
lahsivjar: I see, I agree having integration test have their own benefits but, honestly, I am still not convinced that adding integration tests to test calculations for collector is the best approach since it can be tested via unit more simply and thoroughly. However, I see the benefit of adding tests while considering each metric since there are metric specific logic in the code. Let me update the tests.

systemtest/benchtest/main_test.go
```diff
@@ -197,3 +199,147 @@ func Test_warmupTimeout(t *testing.T) {
...
+	var count int64
+	return httptest.NewServer(
+		http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			if r.URL.Path != "/debug/vars" {
+				t.Errorf("unexpcted path: %s", r.URL.Path)
+			}
+			w.WriteHeader(http.StatusOK)
+			response := "{}"
+			// 12 comes from the repeated calls made while querying expvar
+			if count > 12 {
```
lahsivjar, line 310: Couldn't figure out a better way to do this, any ideas to do it differently?

marclop (approved):
LGTM!

PR #8176, opened 2022-05-20
prefer integration config's data_stream.namespace

### LINKED ISSUE #8087: Changing the APM integration namespace breaks ingestion
**APM Server version** (`apm-server version`): 8.2.0

**Description of the problem including expected versus actual behavior**: Changing the APM integration namespace after setting up the integration breaks event ingestion. 

**Steps to reproduce**:

I tested both locally (8.3.0-snapshot) and on Elastic Cloud (8.2.0)

 1. Spin up the Stack and add the APM integration. Use the recommended `default` namespace.
 2. Send trace events to APM Server.
 3. Edit the APM integration and change the namespace to something like `prod`. Save the integration and wait for the Elastic Agent to update.
 4. Send trace events to APM Server. As far as I can tell, there aren't any agent or server logs to indicate ingestion isn't working; the APM Server logs `request accepted` (see screenshot below).
 5. View your data streams and indices and note that there are no new `*apm-prod*` data streams or indices and that events haven't been indexed.

Screen Shot 2022-05-11 at 8 34 11 AM

stuartnelson3:
The incoming data_stream.namespace on the
integration config was never being used to set the
namespace for ingestion.

If set, prefer this value to that set in the
apm-server config.

closes #8087

## How to test these changes

1. start the stack and install the apm integration
2. ingest some documents; verify datastream `traces-apm-default` is there
3. update the apm integration to use a different namespace
4. ingest some documents, verify datastream `traces-apm-$NAMESPACE` is there

### DIFF
--- a/beater/beater.go
+++ b/beater/beater.go
@@ -314,6 +314,7 @@ func (r *reloader) reload() error {
 		RawConfig:                r.rawConfig,
 		FleetConfig:              r.fleetConfig,
 		OutputConfig:             r.outputConfig,
+		Namespace:                r.namespace,
 	})
 	if err != nil {
 		return err
@@ -371,6 +372,7 @@ type serverRunnerParams struct {
 	RawConfig    *agentconfig.C
 	FleetConfig  *config.Fleet
 	OutputConfig agentconfig.Namespace
+	Namespace    string
 }
 
 type sharedServerRunnerParams struct {
@@ -395,6 +397,9 @@ func newServerRunner(ctx context.Context, args serverRunnerParams) (*serverRunne
 	}
 
 	runServerContext, cancel := context.WithCancel(ctx)
+	if args.Namespace == "" {
+		args.Namespace = cfg.DataStreams.Namespace
+	}
 	return &serverRunner{
 		backgroundContext:      ctx,
 		runServerContext:       runServerContext,
@@ -407,7 +412,7 @@ func newServerRunner(ctx context.Context, args serverRunnerParams) (*serverRunne
 		fleetConfig:               args.FleetConfig,
 		acker:                     args.Acker,
 		pipeline:                  args.Beat.Publisher,
-		namespace:                 cfg.DataStreams.Namespace,
+		namespace:                 args.Namespace,
 		beat:                      args.Beat,
 		logger:                    args.Logger,
 		tracer:                    args.Tracer,
--- a/beater/server_test.go
+++ b/beater/server_test.go
@@ -597,6 +597,7 @@ func TestServerOutputConfigReload(t *testing.T) {
 	}
 
 	inputConfig := agentconfig.MustNewConfigFrom(map[string]interface{}{
+		"data_stream.namespace": "custom",
 		"apm-server": map[string]interface{}{
 			"host": "localhost:0",
 			"sampling.tail": map[string]interface{}{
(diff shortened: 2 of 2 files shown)


beater/server_test.go
```diff
@@ -612,6 +613,7 @@ func TestServerOutputConfigReload(t *testing.T) {
 
 	runServerArgs := <-runServerCalls
 	assert.Equal(t, "", runServerArgs.Config.Sampling.Tail.ESConfig.Username)
+	assert.Equal(t, "custom", runServerArgs.Namespace)
```
stuartnelson3, line 616: At first I was trying to verify that this was "default" and then below it would change, but I was getting the following error trying to update that config:

```
=== RUN   TestServerOutputConfigReload
    server_test.go:625: 
        	Error Trace:	server_test.go:625
        	Error:      	Received unexpected error:
        	            	more than one namespace configured accessing config
        	Test:       	TestServerOutputConfigReload
```

I wasn't sure how to fix it. But, seeing as a user might use a custom namespace when first installing apm-integration, I figured this also tests the code path.

beater/beater.go
```diff
@@ -395,6 +397,9 @@ func newServerRunner(ctx context.Context, args serverRunnerParams) (*serverRunne
 	}
 
 	runServerContext, cancel := context.WithCancel(ctx)
+	if args.Namespace == "" {
+		args.Namespace = cfg.DataStreams.Namespace
+	}
```
axw, line 402: Do you think it would be slightly simpler if we removed this conditional, and instead updated `reloader.reload` to set the initial value of `namespace` to `cfg.DataStreams.Namespace`? Probably not too big of a difference, I was just slightly surprised to find the if/else in this function.

beater/server_test.go
axw: Yeah I think this is perfectly reasonable.

axw (approved):
Yikes. Thanks!

beater/beater.go
```diff
@@ -275,14 +275,18 @@ func (r *reloader) Reload(configs []*reload.ConfigWithMeta) error {
...
 	r.rawConfig = integrationConfig.APMServer
+	// Merge in datastream namespace passed in from apm integration
+	if integrationConfig.DataStream != nil && integrationConfig.DataStream.Namespace != "" {
+		c := agentconfig.MustNewConfigFrom(map[string]interface{}{
+			"data_streams.namespace": integrationConfig.DataStream.Namespace,
+		})
+		if r.rawConfig, err = agentconfig.MergeConfigs(r.rawConfig, c); err != nil {
+			return err
+		}
+	}
```
stuartnelson3, line 289: @axw I added the `integrationConfig.DataStream.Namespace != ""` check because, although the integration doesn't allow the user to enter a blank value, it seemed prudent to also check server-side. If you don't think this is necessary I'm happy to remove it.

axw: Agreed, seems prudent to me too.

axw (approved):

axw:
Tested with 8.3.0 BC2 in cloud QA, looks good. I had to edit the ingest pipeline to drop the processor which fails if the version is too old, since the bundled integration package is stale (should be fixed in BC3). Then I sent some data with sendotlp, updated the integration namespace to "nondefault", and sent more. APM Server produced `traces-apm-default` and `traces-apm-nondefault` for the two runs, respectively.

PR #8221, opened 2022-05-25
modelindexer: Fix compressed `BytesTotal` metric

### LINKED ISSUE #7625: modelindexer: BytesTotal is invalid when compression is enabled
https://github.com/elastic/apm-server/blob/36c6166c0e4fa2b6cc11ab896af78952d0cee022/model/modelindexer/indexer.go#L369-L370

The act of calling `Flush` will close the gzip writer, flushing to the buffer. Counting must therefore happen after closing the gzip writer.

Also, I believe our implementation of the metric differs substantially from the libbeat imlementation which counts bytes written at the transport level: https://github.com/elastic/beats/blob/36d7fffb19d497b71b6ac40e7741677aed076b94/libbeat/common/transport/stats.go#L59. I don't think this particularly matters in practice, but maybe we should at least document the difference (internally, for developers), as well as our definition of the metric:

https://github.com/elastic/apm-server/blob/36c6166c0e4fa2b6cc11ab896af78952d0cee022/model/modelindexer/indexer.go#L474-L475

marclop:
## Motivation/summary

Fixes the calculation on the `BytesTotal` reported by the modelindexer
to happen after the gzip writer has been flushed. Correcting the godoc
comment on the struct field.

## Checklist

## How to test these changes

N/A?

## Related issues

Closes #7625


model/modelindexer/bulk_indexer.go
```diff
@@ -135,21 +135,26 @@ func (b *bulkIndexer) writeMeta(item elasticsearch.BulkIndexerItem) {
 	b.jsonw.Reset()
 }
 
-// Flush executes a bulk request if there are any items buffered, and clears out the buffer.
-func (b *bulkIndexer) Flush(ctx context.Context) (elasticsearch.BulkIndexerResponse, error) {
+func (b *bulkIndexer) newBulkRequest() (esapi.BulkRequest, error) {
```
axw, line 138: We may want to extract `bulkIndexer` out of this package in the future, for reuse in Elastic Agent. To make that possible, we should avoid referring to unexported fields and methods.

What do you think about adding a `bytesFlushed` field to `bulkIndexer`, updating it in the `Flush` method if `req.Do` returns a `nil` error? Then we would expose it with a new `BytesFlushed` method.

marclop: that sounds like a better approach.

model/modelindexer/bulk_indexer.go
```diff
@@ -140,22 +146,24 @@ func (b *bulkIndexer) Flush(ctx context.Context) (elasticsearch.BulkIndexerRespo
 	if b.itemsAdded == 0 {
 		return elasticsearch.BulkIndexerResponse{}, nil
 	}
+	req := esapi.BulkRequest{Body: &b.buf, Header: esHeader}
```
axw, line 149: Is there some benefit to having this up here? No big deal, but I'd prefer to keep the request construction as it was.

model/modelindexer/indexer_test.go
```diff
@@ -127,21 +130,22 @@ func TestModelIndexerAvailableBulkIndexers(t *testing.T) {
 		require.NoError(t, err)
 	}
 	stats := indexer.Stats()
-	stats.BytesTotal = 0
 	// FlushBytes is set arbitrarily low, forcing a flush on each new
 	// event. There should be no available bulk indexers.
```
axw, line 134: IMO the assertion on BytesTotal distracts from the rest of this test. We're already checking it's working in TestModelIndexer, WDYT about dropping it from this test and go back to setting BytesTotal to zero before comparing?

systemtest/fleet_test.go
```diff
@@ -76,6 +76,14 @@ func TestFleetIntegrationMonitoring(t *testing.T) {
...
 	}
 	apmIntegration.getBeatsMonitoringStats(t, &metrics)
+	// We don't have a way to assert the writtenBytes at this layer
+	// use the value of `output.write.bytes`.
+	var writeBytes float64
+	if o := metrics.Libbeat["output"].(map[string]interface{}); len(o) > 0 {
+		if w := o["write"].(map[string]interface{}); len(w) > 0 {
+			writeBytes = w["bytes"].(float64)
+		}
+	}
```
axw, line 86: How about we extract that field, check that it's > 0, and then remove it from the map before the comparison below?

axw (commented):
LGTM overall, just a few quibbles

model/modelindexer/bulk_indexer.go
marclop: I suppose not, the compiler optimizations should take care of grouping things in the way that's best for it. I'll revert it.

beater/server_test.go
```diff
@@ -959,7 +959,7 @@ func TestServerElasticsearchOutput(t *testing.T) {
 			},
 			"type": "elasticsearch",
 			"write": map[string]interface{}{
-				"bytes": int64(10),
+				"bytes": int64(0),
```
axw, line 962:
```suggestion
				// _bulk requests haven't completed, so bytes flushed won't have been updated.
				"bytes": int64(0),
```

axw (approved):
LGTM! Thanks for the updates.

PR #8232, opened 2022-05-26
cmd: don't override compression_level

axw:
## Motivation/summary

We have been defaulting to level 5 gzip compression since https://github.com/elastic/apm-server/pull/6675. There is no need to set the config override for cloud.

## Checklist

## How to test these changes

Check that `_bulk` requests to Elasticsearch in Cloud are still gzipped, e.g. by looking at the proxy logs.

## Related issues

None


cmd/root.go
```diff
@@ -54,17 +52,6 @@ var libbeatConfigOverrides = func() []cfgfile.ConditionalOverride {
 				},
 			}),
 		},
-		{
-			Check: func(_ *agentconfig.C) bool {
-				return os.Getenv(cloudEnv) != ""
```
lahsivjar, line 59: Is this flag used to check if APM Server is running on cloud?

lahsivjar (approved):
LGTM, had one question regarding the flag used previously.

cmd/root.go
axw: Yes, it is set only on Cloud. Its value is the memory size of the APM instance. We used to use it for configuring queue sizes, and number of goroutines for the libbeat Elasticsearch output. When we switched to our new modelindexer implementation, that became unnecessary: https://github.com/elastic/apm-server/pull/6656/files#diff-ab967ab1a2f3a1b769106eeb7bfe892ef0e81d1d27811fa15be08e6749feee1fL123

PR #8240, opened 2022-05-26
refactor: remove deprecated ioutil references

kruskall:
## Motivation/summary

`io/ioutil` was deprecated in Go 1.16 .  
The [io/ioutil](https://tip.golang.org/pkg/io/ioutil/) package has turned out to be a poorly defined and hard to understand collection of things. All functionality provided by the package has been moved to other packages.  

See https://tip.golang.org/doc/go1.16#ioutil

## Checklist

### DIFF
--- a/agentcfg/fetch.go
+++ b/agentcfg/fetch.go
@@ -23,7 +23,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
-	"io/ioutil"
 	"net/http"
 	"strings"
 	"time"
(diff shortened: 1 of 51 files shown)


axw (approved):
Thanks @kruskall!

axw:
@kruskall the files under `internal/otel_collector` are vendored from opentelemetry-collector, so we shouldn't modify those ones. Can you please revert the changes to those?

kruskall:
> The files under internal/otel_collector are vendored from opentelemetry-collector, so we shouldn't modify those ones.

TIL. Thank you for this!

@axw Thank you for the feedback! I've pushed some changes, let me know if I should squash the commits

axw:
LGTM, thanks for the updates :)

PR #8253, opened 2022-05-30
feat: add multierrors to config validation

kruskall:
## Motivation/summary

Multiple errors are returned if multiple settings are invalid.
This should improve clarity and general productivity when debugging
config errors.  

As bonus effect, this replace old `pkg/errors` usage with Go 1.13+ errors

## Checklist

- [ ] ~~Update [package changelog.yml](https://github.com/elastic/apm-server/blob/main/apmpackage/apm/changelog.yml) (only if changes to `apmpackage` have been made)~~
- [ ] ~~Documentation has been updated~~

### DIFF
--- a/changelogs/head.asciidoc
+++ b/changelogs/head.asciidoc
@@ -28,6 +28,7 @@ https://github.com/elastic/apm-server/compare/8.2\...main[View commits]
 ==== Added
 - System, process, and well-defined runtime metrics are now sent to the shared `metrics-apm.internal-<namespace>` data stream {pull}7882[7882]
 - Number of parallel bulk requests are now configurable via `output.elasticsearch.max_requests` {pull}8055[8055]
+- Config validation returns more descriptive errors if multiple settings are invalid {pull}8253[8253]
 
 
 // Added but still being debugged


x-pack/apm-server/sampling/config.go
```diff
@@ -153,90 +155,115 @@ type PolicyCriteria struct {
...
+	ErrPoliciesMissing              = errors.New("Policies unspecified")
+	ErrDefaultPolicyMissing         = errors.New("Policies does not contain a default (empty criteria) policy")
+	ErrIngestRateDecayFactorInvalid = errors.New("IngestRateDecayFactor unspecified or out of range (0,1]")
+	ErrCompressionLevelInvalid      = errors.New("CompressionLevel out of range [-1,9]")
+	ErrElasticsearchMissing         = errors.New("Elasticsearch unspecified")
+	ErrDBMissing                    = errors.New("DB unspecified")
+	ErrStorageDirMissing            = errors.New("StorageDir unspecified")
+	ErrStorageGCIntervalInvalid     = errors.New("StorageGCInterval unspecified or negative")
+	ErrTTLInvalid                   = errors.New("TTL unspecified or negative")
+	ErrSampleRateInvalid            = errors.New("SampleRate unspecified or out of range [0,1]")
```
axw, line 172: I'm not too keen on exporting these errors. I don't think there's any good reason for a client of this package to check the specific error value.

x-pack/apm-server/sampling/config_test.go
```diff
@@ -14,73 +14,76 @@ import (
...
 	var config sampling.Config
-	assertInvalidConfigError := func(expectedError string) {
+	assertInvalidConfigError := func(expectedError error) {
 		t.Helper()
 		agg, err := sampling.NewProcessor(config)
-		require.Error(t, err)
 		require.Nil(t, agg)
-		assert.EqualError(t, err, "invalid tail-sampling config: "+expectedError)
+		require.ErrorIs(t, err, expectedError)
+		assert.ErrorContains(t, err, "invalid tail-sampling config: ")
```
axw, line 27: So I think this is actually a step backwards for the test. The intention of the test was to verify that the output is readable and useful in identifying the error. These assertions would still pass if you had a badly formatted error like: `BeatID unspecified invalid tail-sampling config: `.

Maybe that's a bit contrived, but my point is more that the test was fine as it was, and there's really no need for an `errors.Is` level assertion here. Could you please revert to `assert.EqualError`, and show what the complete error messages look like now?

x-pack/apm-server/sampling/pubsub/config.go
```diff
@@ -65,38 +66,52 @@ type DataStreamConfig struct {
...
 }
 
+var (
+	ErrClientMissing         = errors.New("Client unspecified")
+	ErrBeatIDMissing         = errors.New("BeatID unspecified")
+	ErrSearchIntervalInvalid = errors.New("SearchInterval unspecified or negative")
+	ErrFlushIntervalInvalid  = errors.New("FlushInterval unspecified or negative")
+	ErrTypeMissing           = errors.New("Type unspecified")
+	ErrDatasetMissing        = errors.New("Dataset unspecified")
+	ErrNamespaceMissing      = errors.New("Namespace unspecified")
```
axw, line 76: As above, let's please not export these.

axw (commented):
Thanks again @kruskall!

Sorry to say, but I'm not convinced this change is worthwhile. User configuration is verified higher up the stack, e.g. tail-based sampling config is verified according to the tags in https://github.com/elastic/apm-server/blob/main/beater/config/sampling.go.

The validation code you modified is for soundness within this package. We might transform user configuration on the way to this package, hence we validate again, at the entry point to the package. Realistically, these error messages should never be bubbling up to users, so I don't think it matters if we combine the errors or not.

kruskall:
> Thanks again @kruskall!
> 
> Sorry to say, but I'm not convinced this change is worthwhile. User configuration is verified higher up the stack, e.g. tail-based sampling config is verified according to the tags in https://github.com/elastic/apm-server/blob/main/beater/config/sampling.go.
> 
> The validation code you modified is for soundness within this package. We might transform user configuration on the way to this package, hence we validate again, at the entry point to the package. Realistically, these error messages should never be bubbling up to users, so I don't think it matters if we combine the errors or not.

@axw Oh I see, thank you for the explanation! :bow:   

I agree, we can close this!

Issue #8263, opened 2022-06-01
Hard-coded internal metric names are incomplete


axw:
Some (most) opbeans' internal metrics are still going to service-specific data streams. This is because the hard-coded list of metrics in APM Server is incomplete.

See https://github.com/elastic/apm-server/pull/7882#issuecomment-1142011880

PR #8264, opened 2022-06-01
Fix agent metrics

axw:
## Motivation/summary

Add omitted agent metrics, and fix names of agent metrics.

Also, if an agent sends "type" or "unit" for internal metrics, strip those from the metrics to ensure "_metric_descriptions" isn't set in the metricset document. This is set in the application metrics data stream for dynamic mapping; it is not used in internal metrics.

## Checklist

## How to test these changes

1. Run APM Server
2. Run all opbeans applications, pointed at APM Server (e.g. using apm-integration-testing with `--all-opbeans`)
3. Wait at least 1 minute to ensure all agents have sent some metrics
4. Ensure there are no service-specific application metrics data streams (`metrics-apm.app.*-*`)

Note that opbeans-python is instrumented with Prometheus, so it will generate `metrics-apm.app.opbeans_python-*` by default. Prometheus metrics can be disabled with the env var `ELASTIC_APM_PROMETHEUS_METRICS=False`, which should lead to the service-specific data stream no longer being populated.

## Related issues

Closes https://github.com/elastic/apm-server/issues/8263


marclop (approved):
Shouldn't we also add `backport-8.3` since we've already branched out?

axw:
Indeed, thanks for catching that.

lahsivjar (approved):

stuartnelson3:
confirmed with d920cdd3487563f235eb40c554cce316a15bfdde

PR #8287, opened 2022-06-02
Fix flaking test for TestAggregate

lahsivjar:
## Motivation/summary

Fix flaky test for expvar collector

## Checklist


systemtest/benchtest/expvar/metrics_test.go
```diff
@@ -137,16 +133,55 @@ func TestWatchMetricNonBlocking(t *testing.T) {
...
+			}
+		}
+		done <- true
+	}()
+
+	return server
+}
+
+func getTestServer(t *testing.T, resChan <-chan string, timeout time.Duration) *httptest.Server {
 	return httptest.NewServer(
```
axw, line 172: Should we be closing this server with a t.Cleanup?

axw (commented):
I'm having a hard time wrapping my head around the changes.

I feel like we could simplify the tests by separating construction of the collector from starting its query loop (in a background goroutine), and then synchronously unblocking server requests and responses in the test. So `TestAggregate` would look something like:

```go
func TestAggregate(t *testing.T) {
	collector, reqs := newTestCollector(t)

	for i := 0; i < 5; i++ {
		// unblock a request
		req := <-reqs
		close(req.done) // allow response
	}
	req <-reqs // wait for 6th request, but don't respond

	stats := c.Get(Goroutines)
	assert.Equal(t, 1, stats.Last)
	assert.Equal(t, 5, stats.Last)
	assert.Equal(t, 3.0, stats.Mean) // (1+2+3+4+5)/5
	assert.Equal(t, 5, stats.Max)
	assert.Equal(t, 1, stats.Min)
}
```

Maybe even have the test body specify the metric value to send back.

WDYT?

systemtest/benchtest/expvar/metrics_test.go
lahsivjar: Sounds good, will update the PR

lahsivjar:
@axw I kept the construction and start of query loop together to avoid sync protection for the `start` method. This could be handled by utilizing sync primitives easily, however, I think the complexity will still remain unless we remove the logic to do the [first query to expvar in sync](https://github.com/elastic/apm-server/blob/15a29254315748fa2f0056f46b266cd134a77719/systemtest/benchtest/expvar/metrics.go#L202-L208) before starting the query loop. Removing this will mean the start point for our metrics may not always be a clean slate w.r.t. requests to the APM server.

WDYT?

axw (approved):
@lahsivjar I've taken a second look with fresh eyes, and it looks fine. Thanks, LGTM.

PR #8323, opened 2022-06-07
systemtest/cmd/sendotlp: add http/protobuf support

axw:
## Motivation/summary

Update `sendotlp` with a `-protocol` flag, which can be used to specify the transport protocol for sending OTLP data to APM Server. The flag value may be either "grpc" or "http/protobuf".

For now the protocol defaults to "grpc", but in the future (when OTLP/HTTP has been available in APM Server for a while) we should consider changing it to "http/protobuf" to align with the OpenTelemetry SDK default.

## Checklist

## How to test these changes

1. Start an APM Server 8.3.0+ deployment
2. Run `sendotlp -endpoint <apm-server-url> -secret-token <secret-token>`
3. Run `sendotlp -endpoint <apm-server-url> -secret-token <secret-token> -protocol http/protobuf`

Check that events show up both times, and check (e.g. through proxy logs) that the appropriate protocol is used.

## Related issues

None


systemtest/cmd/sendotlp/main.go
```diff
@@ -214,9 +193,10 @@ func generateSpans(ctx context.Context, tracer trace.Tracer) error {
 }
 
 func generateMetrics(ctx context.Context, meter metric.Meter) error {
-	metric.Must(meter).NewFloat64Counter("float64_counter").Add(ctx, 1)
+	counter, _ := meter.SyncFloat64().Counter("float64_counter")
```
lahsivjar, line 196: Should we handle the error here? I think error may not happen while only using the command but the command may fail without a good explanation in case some error happens while developing something on top of this.

axw: I figured the program was small enough that it would be easy to debug, but it doesn't hurt to check - updated.

lahsivjar (approved):
LGTM

lahsivjar:
Confirmed on 8.4.0 BC3

PR #8334, opened 2022-06-08
Support service.target.{type, name} for OTLP events

### LINKED ISSUE #7980: Backend dependencies granularity: add support for otlp events
Support `span.service.target.*` fields also for otlp events. 

From https://github.com/elastic/apm/pull/627/files#diff-78ae87d6c58be8cc0ed40d4dcd78faa83a58e462ad0b836912194a5cf8aed336R220:

> APM server already infers the `span.destination.service.resource` value from OTel span attributes, this algorithm needs
to be updated in order to also infer the values of `span.context.service.target.*` fields.
> - `span.context.service.target.type` should be set from the inferred value of `span.subtype` with fallback to `span.type`
>   - For database spans: use value of `db.system` attribute
>   - For HTTP client spans: use `http`
>   - For messaging spans: use value of `messaging.system` attribute
>   - For RPC spans: use value of `rpc.system` attribute
> - `span.context.service.target.name` should be set from OTel attributes if they are present
>   - For database spans: use value of `db.name` attribute
>   - For HTTP client spans: create `<host>:<port>` string from `http.host`, `net.peer.port` attributes or equivalent
>   - For messaging spans: use value of `messaging.destination` attribute if `messaging.temp_destination` is `false` or absent to limit cardinality
>   - For RPC spans: use value of `rpc.service`

Related to https://github.com/elastic/apm-server/issues/7715

lahsivjar:
## Motivation/summary

Infer `service.target.{type, name}` from OTLP events.

## Checklist

## How to test these changes

Observe that each span type the `service.target.{type, name}` in the output doc are populated as per the corresponding tables with the appropriate inputs.

1. DB spans:
    | [in]peer.service | [in]db.system | [in]db.name | [out]target.type | [out]target.name |
    |---------------|---------------|-------------|------------------|------------------|
    | testsvc | postgresql    | testdb      | postgresql       | testdb           |
    | testsvc | postgresql    | <nil>      | postgresql       | testsvc           |
    | <nil> | <nil>         | testdb      | db               | testdb           |
    | <nil> | <nil>         | <nil>       | db               | ""               |
2. HTTP spans:
    | [in]peer.service | [in]http.url | [out]target.type | [out]target.name |
    |---|---|---|---|
    | testsvc | https://test-url:443/ | http | test-url:443 |

    | [in]peer.service | [in]http.scheme | [in]http.host | [in]http.target | [out]target.type | [out]target.name |
    |---|---|---|---|---|---|
    | <nil> | https | test-url:443 | / | http | test-url:443 |

    | [in]peer.service | [in]http.scheme | [in]net.peer.name | [in]net.peer.ip | [in]net.peer.port | [in]http.target | [out]target.type | [out]target.name |
    |---|---|---|---|---|---|---|---|
    | <nil> | https | test-url | ::1 | 443 | / | http | test-url:443 |

    | [in]peer.service | [in]http.scheme | [in]net.peer.ip | [in]net.peer.port | [in]http.target | [out]target.type | [out]target.name |
    |---|---|---|---|---|---|---|
    | <nil> | https | ::1 | 443 | / | http | [::1]:443 |
3. RPC spans:
    | [in]peer.service | [in]rpc.system | [in]rpc.service | [out]target.type | [out]target.name |
    |----------------|----------------|-----------------|------------------|------------------|
    | testsvc | grpc           | test            | grpc             | testsvc             |
    | testsvc | grpc           | <nil>            | grpc             | testsvc             |
    | <nil>          | test            | external         | test             |
4. Messaging spans:
    | [in]peer.service | [in]messaging.system | [in]messaging.destination | [out]target.type | [out]target.name |
    |----------------------|----------------------|---------------------------|------------------|------------------|
    | testsvc | kafka                | test-topic                | kafka            | test-topic       |
    | testsvc | kafka                | <nil>                | kafka            | testsvc       |
    | <nil> | <nil>                | test-topic                | messaging        | test-topic       |

## Related issues

Closes #7980

### DIFF
--- a/processor/otel/traces.go
+++ b/processor/otel/traces.go
@@ -497,8 +497,14 @@ func TranslateSpan(spanKind pdata.SpanKind, attributes pdata.AttributeMap, event
 	)
 
 	var (
-		messageSystem    string
-		messageOperation string
+		messageSystem          string
+		messageOperation       string
+		messageTempDestination bool
+	)
+
+	var (
+		rpcSystem  string
+		rpcService string
 	)
 
 	var http model.HTTP
@@ -507,8 +513,8 @@ func TranslateSpan(spanKind pdata.SpanKind, attributes pdata.AttributeMap, event
 	var message model.Message
 	var db model.DB
 	var destinationService model.DestinationService
+	var serviceTarget model.ServiceTarget
 	var foundSpanType int
-	var rpcSystem string
 	var samplerType, samplerParam pdata.AttributeValue
 	attributes.Range(func(kDots string, v pdata.AttributeValue) bool {
 		if isJaeger {
@@ -527,7 +533,13 @@ func TranslateSpan(spanKind pdata.SpanKind, attributes pdata.AttributeMap, event
 		case pdata.AttributeValueTypeArray:
 			setLabel(k, event, ifaceAttributeValueSlice(v.SliceVal()))
 		case pdata.AttributeValueTypeBool:
-			setLabel(k, event, strconv.FormatBool(v.BoolVal()))
+			switch kDots {
+			case semconv.AttributeMessagingTempDestination:
+				messageTempDestination = v.BoolVal()
+				fallthrough
+			default:
+				setLabel(k, event, strconv.FormatBool(v.BoolVal()))
+			}
 		case pdata.AttributeValueTypeDouble:
 			setLabel(k, event, v.DoubleVal())
 		case pdata.AttributeValueTypeInt:
@@ -625,6 +637,8 @@ func TranslateSpan(spanKind pdata.SpanKind, attributes pdata.AttributeMap, event
 				rpcSystem = stringval
 				foundSpanType = rpcSpan
 			case semconv.AttributeRPCService:
+				rpcService = stringval
+				foundSpanType = rpcSpan
 			case semconv.AttributeRPCMethod:
 
 			// miscellaneous
@@ -688,6 +702,7 @@ func TranslateSpan(spanKind pdata.SpanKind, attributes pdata.AttributeMap, event
 		}
 	}
 
+	serviceTarget.Name = peerService
 	destinationService.Name = peerService
 	destinationService.Resource = peerService
 	if peerAddress != "" {
@@ -706,7 +721,8 @@ func TranslateSpan(spanKind pdata.SpanKind, attributes pdata.AttributeMap, event
 		event.Span.Subtype = subtype
 		event.HTTP = http
 		event.URL.Original = httpURL
-		if destinationService.Name == "" && fullURL != nil {
+		serviceTarget.Type = event.Span.Subtype
+		if serviceTarget.Name == "" && fullURL != nil {
 			url := url.URL{Scheme: fullURL.Scheme, Host: fullURL.Host}
 			resource := url.Host
 			if destPort == schemeDefaultPort(url.Scheme) {
@@ -719,19 +735,25 @@ func TranslateSpan(spanKind pdata.SpanKind, attributes pdata.AttributeMap, event
 				}
 			}
 
+			serviceTarget.Name = resource
 			destinationService.Name = url.String()
 			destinationService.Resource = resource
 		}
 	case dbSpan:
 		event.Span.Type = "db"
-		if db.Type != "" {
-			event.Span.Subtype = db.Type
+		event.Span.Subtype = db.Type
+		serviceTarget.Type = event.Span.Type
+		if event.Span.Subtype != "" {
+			serviceTarget.Type = event.Span.Subtype
 			if destinationService.Name == "" {
 				// For database requests, we currently just identify
 				// the destination service by db.system.
 				destinationService.Name = event.Span.Subtype
 				destinationService.Resource = event.Span.Subtype
 			}
+		}
+		if serviceTarget.Name == "" {
+			serviceTarget.Name = db.Instance
 		}
 		event.Span.DB = &db
 	case messagingSpan:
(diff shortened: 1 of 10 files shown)


processor/otel/traces.go
```diff
@@ -688,6 +702,7 @@ func TranslateSpan(spanKind pdata.SpanKind, attributes pdata.AttributeMap, event
 		}
 	}
 
+	serviceTarget.Name = peerService
```
lahsivjar, line 705: @SylvainJuge @axw For `service.target.name` I have kept `peer.service` as highest priority unlike `destination.service.resource` which has `peer.address` as highest priority followed by `peer.service`.

The main motivation for this was that `peer.service` aligns well the definition of `service.target.name`. Additionally IIUC, `peer.address` is defined in the deprecated opentracing spec. The somewhat equivalent of `peer.address` in open-telemetry spec is `net.peer.*` which is handled in a different way.

Maybe it is time to remove handing `peer.address` entirely? What are your thoughts on this?

axw: I think those priorities make sense.

> Maybe it is time to remove handing peer.address entirely? What are your thoughts on this?

IMO we should keep it for now. I think we should have a broader discussion about deprecating both the Jaeger intake and OpenTracing bridges at some point.

SylvainJuge: After checking the usage of `peer.service` within the OpenTelemetry Java agent:
- when there is no value for `peer.service`, the value of `peer.service` is set t the same value as `peer.ip`
- the value of `peer.service` contains the host name for HTTP spans (or whatever is in URL).
- for some instrumentations like gRPC and Dubbo the service name is not set to a protocol-specific name (gRPC has a concept of service name which is not used here).

With that, I think the best option (for now) would be to :
- make the `peer.service` value only override the hostname part of the resource when applicable (HTTP spans and RPC spans)
- ignore the value of `peer.service` for the database spans as the values will contain the host/IP of the database, whereas we use it for the database name.

axw: @SylvainJuge thanks for the additional context. That's kind of annoying, but reading the semantic conventions again, it's only a "SHOULD", so I suppose you're right that we should give `peer.service` higher precedence in general.

> - make the peer.service value only override the hostname part of the resource when applicable (HTTP spans and RPC spans)
> - ignore the value of peer.service for the database spans as the values will contain the host/IP of the database, whereas we use it for the database name.

This sounds a little bit complicated. Can we simplify it, and only use `peer.service` for any span other than HTTP, RPC, database, and messaging? i.e. just as a fallback. WDYT?

SylvainJuge: Yes, that would definitely be a fair compromise here, let's just make `peer.service` have higher priority on the `service.target.name` field and call it a day. From the user perspective that would also be a bit better as the value set explicitly in the code would be reflected in the map and dependencies.

lahsivjar: @SylvainJuge @axw  Thanks for the review. IIUC, the conclusion here is to use `peer.service` to have highest priority when inferring `service.target.name` for **ALL** span types. Let me know if I am misunderstanding above discussion.

axw: Sorry, crossed wires between me and @SylvainJuge I think. I think that would be ideal _if_ the OTel SDK didn't default to using `peer.ip`.

@SylvainJuge what I meant was:

 - for HTTP, RPC, database, and messaging spans: follow the rules laid out in https://github.com/elastic/apm-server/issues/7980
 - for all other span types, use `peer.service` for `service.target.name`

SylvainJuge: Thanks for the clarification @axw, using `peer.service` only on "other" span types makes sense and allows to keep maximum compatibility here, it should even allow to prevent duplication of dependencies with a mixed set of OTel and our agents.

For `peer.ip` I don't have any strong opinion, thus I'll rely on you for this one.

lahsivjar: Thanks for the clarification. I have updated the PR and the tests in the description as per our conclusion.

axw (approved):
LGTM, thanks @lahsivjar !

SylvainJuge (approved):
For RPC spans, the agents (at least the Java one) currently use the `host:port` format for the `service.target.name` and the `destination.service.resource`, but this small difference will be handled once we start work on https://github.com/elastic/apm/issues/646 , as I don't want to further complicate nor delay this I'll approve and we can deal with that later.

axw:
Tested with 8.4.0-BC1, sending spans according to the description. Looks good except for one discrepancy, where the RPC spans' target name is set to `rpc.service`. I think this is expected according to the spec, but I forget where exactly we landed.

[The spec](https://github.com/elastic/apm/blob/main/specs/agents/tracing-spans-service-target.md#otel-and-bridge-compatibility) says `span.context.service.target.name` should be set to `rpc.service`. @lahsivjar Is there an error in the test cases for RPC in the description?

lahsivjar:
@axw After discussing with @SylvainJuge we decided that the definition of `peer.service` ([ref](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/span-general.md#general-remote-service-attributes)) aligns well with the expectation of `service.target.name`. I missed creating an issue to update the spec after the discussion, I will open a discussion/PR to correct it.

axw:
OK, I think we need to make a small fix for 8.4 then. Currently, `rpc.service` takes priority. Specifically, with the following set of attributes:

```go
semconv.PeerServiceKey.String("testsvc"),
semconv.RPCSystemKey.String("grpc"),
semconv.RPCServiceKey.String("test"),
```

`service.target.name` ends up as "test", rather than "testsvc".

axw:
Discussed with @lahsivjar on Slack, `rpc.service` should take precedence.

PR #8375, opened 2022-06-14
[tbs] log config err and continue

### LINKED ISSUE #8220: APM Server should log warning when Tail Based Sampling policy is invalid rather than crashing
**APM Server version** (`apm-server version`): >= 8.1

Whenever an invalid Tail Based Sampling policy is provided, the APM Server logs it and shuts down. It should instead log an Error but keep running, as it is not justified that a misconfiguration is completely shutting down the APM Server. 
We have seen several deployments being negatively impacted by this. 

**Steps to reproduce**:
Run an Elastic Agent managed APM Server and configure a malformed TBS policy. Observe that the APM Server shuts down (and EA tries to restart it over and over).

stuartnelson3:
## Motivation/summary

If the provided tail-based sampling config is
invalid, log that there was an error and continue
running with tbs disabled.

closes #8220

## Checklist

## How to test these changes

- run apm-server with an invalid tbs config, like
    ```yml
    apm-server:
      sampling:
        tail:
          enabled: true
    ```
- observe the logs for both error and info logs saying there was an issue with tbs, and the server is running with it disabled
- observe apm-server is running

### DIFF
--- a/beater/config/sampling.go
+++ b/beater/config/sampling.go
@@ -23,6 +23,7 @@ import (
 	"github.com/pkg/errors"
 
 	"github.com/elastic/apm-server/elasticsearch"
+	logs "github.com/elastic/apm-server/log"
 	"github.com/elastic/elastic-agent-libs/config"
 	"github.com/elastic/elastic-agent-libs/logp"
 )
@@ -71,15 +72,26 @@ type TailSamplingPolicy struct {
 }
 
 func (c *TailSamplingConfig) Unpack(in *config.C) error {
+	var err error
+	defer func() {
+		if err != nil {
+			logger := logp.NewLogger(logs.Config)
+			logger.Errorf("failed to setup tail sampling: %v", err)
+			logger.Info("continuing with tail sampling disabled")
+			*c = TailSamplingConfig(defaultTailSamplingConfig())
+		}
+	}()
 	type tailSamplingConfig TailSamplingConfig
 	cfg := tailSamplingConfig(defaultTailSamplingConfig())
-	if err := in.Unpack(&cfg); err != nil {
-		return errors.Wrap(err, "error unpacking tail sampling config")
+	if err = in.Unpack(&cfg); err != nil {
+		err = errors.Wrap(err, "error unpacking config")
+		return nil
 	}
 	cfg.Enabled = in.Enabled()
 	*c = TailSamplingConfig(cfg)
 	c.esConfigured = in.HasField("elasticsearch")
-	return errors.Wrap(c.Validate(), "invalid tail sampling config")
+	err = errors.Wrap(c.Validate(), "invalid config")
+	return nil
 }
 
 func (c *TailSamplingConfig) Validate() error {
--- a/beater/config/sampling_test.go
+++ b/beater/config/sampling_test.go
@@ -35,18 +35,20 @@ func TestSamplingPoliciesValidation(t *testing.T) {
 		assert.NoError(t, err)
 	})
 	t.Run("NoPolicies", func(t *testing.T) {
-		_, err := NewConfig(config.MustNewConfigFrom(map[string]interface{}{
+		c, err := NewConfig(config.MustNewConfigFrom(map[string]interface{}{
 			"sampling.tail.enabled": true,
 		}), nil)
-		assert.EqualError(t, err, "Error processing configuration: invalid tail sampling config: no policies specified accessing 'sampling.tail'")
+		assert.NoError(t, err)
+		assert.False(t, c.Sampling.Tail.Enabled)
 	})
 	t.Run("NoDefaultPolicies", func(t *testing.T) {
-		_, err := NewConfig(config.MustNewConfigFrom(map[string]interface{}{
+		c, err := NewConfig(config.MustNewConfigFrom(map[string]interface{}{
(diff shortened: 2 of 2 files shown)


beater/config/sampling.go
```diff
@@ -71,15 +72,26 @@ type TailSamplingPolicy struct {
 }
 
 func (c *TailSamplingConfig) Unpack(in *config.C) error {
+	var err error
+	defer func() {
+		if err != nil {
+			logger := logp.NewLogger(logs.Config)
+			logger.Errorf("failed to setup tail sampling: %v", err)
+			logger.Info("continuing with tail sampling disabled")
```
stuartnelson3, line 80: happy to change these messages to whatever the reviewer things is appropriate. also, i felt like there should be an additional log message about running with tbs disabled, but for some reason i thought it should be `info` level. happy to change or discuss this, as I don't have any well-formulated reason for choosing `info`.

marclop (approved):

beater/config/sampling.go
```diff
@@ -71,15 +72,26 @@ type TailSamplingPolicy struct {
...
 	type tailSamplingConfig TailSamplingConfig
 	cfg := tailSamplingConfig(defaultTailSamplingConfig())
-	if err := in.Unpack(&cfg); err != nil {
-		return errors.Wrap(err, "error unpacking tail sampling config")
+	if err = in.Unpack(&cfg); err != nil {
+		err = errors.Wrap(err, "error unpacking config")
+		return nil
 	}
 	cfg.Enabled = in.Enabled()
 	*c = TailSamplingConfig(cfg)
```
lahsivjar, line 91: Should we avoid setting an invalid config completely?

lahsivjar (commented):
LGTM

beater/config/sampling.go
```diff
@@ -71,15 +72,26 @@ type TailSamplingPolicy struct {
...
+	var err error
+	defer func() {
+		if err != nil {
+			logger := logp.NewLogger(logs.Config)
+			logger.Errorf("failed to setup tail sampling: %v", err)
+			logger.Info("continuing with tail sampling disabled")
+			*c = TailSamplingConfig(defaultTailSamplingConfig())
+		}
+	}()
 	type tailSamplingConfig TailSamplingConfig
```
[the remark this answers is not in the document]
stuartnelson3: Not sure, tbh

beater/config/sampling.go
stuartnelson3: you mean calling `Validate()` on the cfg, before setting `*c` to be equal to it? Since the code was already doing that and, should `Validate()` fail, `c` is set to the default config, I don't think it's a big deal, but I'm happy to clean it up if you think we should avoid this.

beater/config/sampling.go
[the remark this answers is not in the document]
simitt: @axw might have some details

beater/config/sampling.go
lahsivjar: I was concerned about the bad tbs config affecting the downstream code flow and causing unexpected cases. Since earlier we were terminating the server and thus such cases might not have been handled. If you think it is not a concern or protected by some other code flow then it is okay with me.

beater/config/sampling.go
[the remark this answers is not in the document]
axw: The type `tailSamplingConfig` drops the Unpack method from TailSamplingConfig, but retains all of its fields with their `config:"..."` tags. This allows us to unpack according to the field tags, and then run some custom code afterwards (i.e. the setting of esConfigured)

beater/config/sampling.go
stuartnelson3: In the event `Validate()` fails (returns a non-nil error), the defer func defined above should set `c` to be the default config. `Validate()` isn't available on `cfg` (as mentioned above by @axw, it drops the methods of the `TailSamplingConfig` type). Because the default config has enabled=false, it would actually be good to see if that effects downstream code, because then the code isn't respecting the fact that tbs is disabled

axw:
Verified with 8.4.0-BC1.

`docker run -p 8200:8200 --rm docker.elastic.co/staging/apm-server:8.4.0-91109cf0 apm-server --environment=container -E apm-server.sampling.tail.enabled=true -E output.elasticsearch.enabled=false -E output.console.enabled=true -E apm-server.data_streams.wait_for_integration=false`

Logs:

>`
{"log.level":"info","@timestamp":"2022-08-01T05:16:20.373Z","log.origin":{"file.name":"instance/beat.go","file.line":470},"message":"apm-server start running.","service.name":"apm-server","ecs.version":"1.6.0"}                                                                
{"log.level":"error","@timestamp":"2022-08-01T05:16:20.373Z","log.logger":"config","log.origin":{"file.name":"config/sampling.go","file.line":82},"message":"failed to setup tail sampling: invalid config: no policies specified","service.name":"apm-server","ecs.version":"1.6.
0"}                                                                                                                                                                                                                                                                               
{"log.level":"info","@timestamp":"2022-08-01T05:16:20.373Z","log.logger":"config","log.origin":{"file.name":"config/sampling.go","file.line":83},"message":"continuing with tail sampling disabled","service.name":"apm-server","ecs.version":"1.6.0"}
`

Then confirmed the server still runs and accepts events, and sends them to the output.

PR #8391, opened 2022-06-16
spanmetrics: aggregate on span.name

### LINKED ISSUE #8274: Add span.name to aggregation key for service destination metrics
Add the `span.name` to the aggregation key for service destination metrics. 

* exclude `span.name` for RUM events due to high cardinality values
* add `span.name` for any other event that has `span.destination.service` or the newly introduced `service.target.{type, name}` fields set. 

The aggregation already has a default limit of 10_000 time series per minute. When the limit is reached, metrics docs per event are written to ES. We may want to change behaviour when the limit is reached, e.g. stop aggregating by `span.name` to avoid breaking existing charts in Kibana, such as the dependencies list.

axw:
## Motivation/summary

Add `span.name` to the service_destination metrics aggregation key. To protect against high cardinality span names (which will eventually be fixed, but are currently a problem in some instrumentation) we will stop aggregating on `span.name` once 50% of the aggregation groups capacity is reached, and log a warning.

## Checklist

## How to test these changes

1. Instrument an application to send 10000 exit spans to APM Server with low cardinality span names
2. Check that APM Server produces `service_destination` metrics all with `span.name` set
3. Update the instrumentation to use high cardinality span names, and send another 10000 exit spans such that the 10000 limit is reached within 1 minute.
4. Check that APM Server produces 5000 metrics with `span.name` set, and logs a warning once that 50% capacity is reached. The remaining metrics are captured without `span.name` set. (Because we only send 10000 events, there should be no immediately published metricsets.)

## Related issues

Closes #8274


x-pack/apm-server/aggregation/spanmetrics/aggregator_test.go
```diff
@@ -484,10 +542,25 @@ func TestAggregatorOverflow(t *testing.T) {
...
 	require.NoError(t, err)
 	assert.Empty(t, batchMetricsets(t, batch))
+	assert.Equal(t, 1, observed.FilterMessage("service destination groups reached 50% capacity").Len())
+	assert.Len(t, observed.TakeAll(), 1)
+
+	// After hitting 50% capacity (two buckets), then subsequent new metrics will
+	// be aggregated without span.name.
+	span3 := makeSpan("service", "agent", "destination3", "trg_type_3", "trg_name_3", "success", 100*time.Millisecond, 1)
+	span4 := makeSpan("service", "agent", "destination4", "trg_type_4", "trg_name_4", "success", 100*time.Millisecond, 1)
+	span4.Span.Name = span3.Span.Name
```
marclop, line 552: Why are we setting the span4.name to the span3.name? It is not obvious why this is being done. From what I can read in the test:

1. Add 2 distinct span with different span names (`service:destination1`, `service:destination2`). Assert we have reached 50% capacity, no metric sets need to be published immediately (2/4).
2. Add 2 more spans with the same name?  (`service:destination3`, `service:destination3`). Assert we don't get any spans to publish immediately, but we're in fact aggregating the two spans above in the same key since we don't take the `Span.Name` into consideration.

Wouldn't it be better to remove this statement? `Span.Name` won't be considered, so why bother setting it to anything.

marclop (approved):
LGTM, just one question about the test code

x-pack/apm-server/aggregation/spanmetrics/aggregator_test.go
```diff
@@ -465,17 +516,24 @@ func TestAggregateTimestamp(t *testing.T) {
 	assert.Equal(t, t0.Add(30*time.Second), metricsets[1].Timestamp)
 }
 
-func TestAggregatorOverflow(t *testing.T) {
+func TestAggregatorMaxGroups(t *testing.T) {
+	core, observed := observer.New(zapcore.DebugLevel)
+	logger := logp.NewLogger("", zap.WrapCore(func(in zapcore.Core) zapcore.Core {
+		return zapcore.NewTee(in, core)
+	}))
+	_ = observed
```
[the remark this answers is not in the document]
axw: Indeed, thanks

x-pack/apm-server/aggregation/spanmetrics/aggregator_test.go
axw: I don't know what I was thinking now. I've removed that line, leaving `Span.Name` as the distinct generated values.

x-pack/apm-server/aggregation/spanmetrics/aggregator.go
```diff
@@ -264,13 +271,18 @@ func (a *Aggregator) processDroppedSpanStats(event *model.APMEvent, dss model.Dr
 		dss.DestinationServiceResource,
 		dss.ServiceTargetType,
 		dss.ServiceTargetName,
+
+		// BUG(axw) dropped span statistics do not contain span name.
+		// Capturing the service name requires changes to Elastic APM agents.
```
felixbarny, line 276: While aggregating metrics in APM agents would be feasible and more accurate, I'm not quite sure if it's worth the additional memory overhead in the agents. Exit spans that accidentally create high cardinality may also be more impactful in APM Agents.

x-pack/apm-server/aggregation/spanmetrics/aggregator.go
```diff
@@ -327,11 +365,13 @@ func makeAggregationKey(event *model.APMEvent, resource, targetType, targetName
 		serviceEnvironment: event.Service.Environment,
 		agentName:          event.Agent.Name,
 
+		spanName: spanName,
```
felixbarny, line 368: Did you consider adding a config option to let users opt-out of span.name aggregation?

axw: I did consider it, but would prefer to start out with minimal configuration. I see no reason to opt out except in in the case of high cardinality span names, which is handled automatically by stopping aggregation on `span.name` once the capacity reaches 50%.

x-pack/apm-server/aggregation/spanmetrics/aggregator.go
axw: Yeah fair point. Perhaps we should just document that the operation stats don't include dropped spans.

x-pack/apm-server/aggregation/spanmetrics/aggregator.go
```diff
@@ -264,13 +271,18 @@ func (a *Aggregator) processDroppedSpanStats(event *model.APMEvent, dss model.Dr
 		dss.DestinationServiceResource,
 		dss.ServiceTargetType,
 		dss.ServiceTargetName,
+
+		// BUG(axw) dropped span statistics do not contain span name.
+		// Capturing the service name requires changes to Elastic APM agents.
+		"",
```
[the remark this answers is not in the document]
axw: My thinking was: the UI needs to decide what to do with older `service_destination` metrics which lack the `span.name` field (using `missing`), so that should also apply in the case where `span.name` is unavailable for aggregation.

lahsivjar:
Confirmed on 8.4.0-BC1:

**Result with 10 fixed cardinality for `span.name` each with 1000 spans**

with_span_name_fixed_cardinality

**Result with 10_000 cardinality for `span.name`**

warning_log_svc_dest_group_50%_capacity

with_high_cardinality

PR #8400, opened 2022-06-17
Create lambda_metrics.yml

### LINKED ISSUE #8399: Add AWS Lambda metrics to the internal metrics data stream
As per https://github.com/elastic/apm-server/pull/8373, Lambda metrics are currently sent to service-specific data streams. As mentioned [here](https://github.com/elastic/apm-server/pull/8373#issuecomment-1157453127) and [here](https://github.com/elastic/apm-server/issues/7714#issue-1186015549), the Lambda metrics should be sent to the `metrics-apm.internal-*` data stream. 

This requires adding Lambda metrics to `apmpackage/apm/data_stream/internal_metrics/fields` (e.g. `lambda_metrics.yml`).

jlvoiseux:
## Motivation/summary
Resolves #8399.

### Discussion items
- In the current implementation, only Lambda metrics proper were added to `lambda_metrics.yml`. Do we need to add the `faas` fields as well?

## Checklist

## How to test these changes

The test plan is similar to https://github.com/elastic/apm-aws-lambda/pull/220, with the addition of verifying if the metrics are sent to the right data stream.


axw:
@jlvoiseux looking at the metric names again, they feel a bit off. For one thing, I don't think they need "metrics" in the name. For another, do they need to be Lambda-specific? Should they rather be `faas.duration`, `faas.billed_duration`, etc.?

apmpackage/apm/data_stream/internal_metrics/fields/lambda_metrics.yml
```diff
@@ -0,0 +1,20 @@
+- name: aws.lambda.metrics.billed_duration
```
jlvoiseux, line 1: I agree with a possible change to `faas.billed_duration` - however, we might want to keep a naming element allowing us to differentiate these metrics from the `faas` fields introduced in the model.

beater/test_approved_es_documents/TestPublishIntegrationMetricsets.approved.json
```diff
@@ -303,6 +303,69 @@
...
+            "data_stream.dataset": "apm.internal",
+            "data_stream.namespace": "default",
+            "data_stream.type": "metrics",
+            "ecs": {
+                "version": "1.12.0"
+            },
+            "faas": {
+                "coldstart": true,
+                "execution": "6f7f0961f83442118a7af6fe80b88d56",
+                "id": "arn:aws:lambda:us-east-2:123456789012:function:custom-runtime"
```
jlvoiseux, line 326: I had to double check, but it seems that the Extensions API [does not provide the Layer version](https://docs.aws.amazon.com/lambda/latest/dg/runtimes-extensions-api.html) - here that would be
```
arn:aws:lambda:us-east-2:123456789012:function:custom-runtime:1
```
That information would be useful to have, but there is not much we can do for now that improve that I think.

axw (approved):

axw:
With 8.4.0-BC3 and using the Lambda extension built from the main branch, I'm seeing Lambda metrics sent to `metrics-apm.internal-*` as expected:

image

PR #8407, opened 2022-06-17
sampling: Handle `ErrNOSPC` and `os.PathError`

### LINKED ISSUE #8314: APM Server can fill up disk with vlog files
**APM Server version** (`apm-server version`): 8.1.2

With tail-based sampling enabled, APM Server will happily continue writing vlog files until the disk (quota) is filled. We should not just consume all disk; we should set a reasonable default limit, make it configurable, and define/document what happens when the limit is exceeded.

**Steps to reproduce**:

(It's advisable to run APM Server in a container with disk quota when running these steps.)

 1. Enable tail-based sampling, set default sampling rate to 0.9999
 2. Send traces as fast as possible to APM Server
 3. Watch it run out of disk

> writeRequests: Unable to write to value log file: ".../tail_sampling/002192.vlog": write .../tail_sampling/002192.vlog: no space left on device

marclop:
## Motivation/summary

This patch handles how the tail sampling processor behaves when writes
to the BadgerDB fail with `ErrNOSPC` or `os.PathError`. By default, any
traces will be sampled when after BadgerDB has consumed all disk space.

Additionally, we're ignoring the two errors above when failing to write
the subscriber position to disk, since that would cause the server to
shut down.

## Checklist

## How to test these changes

It is not the most straightforward change to test in macOS, since using Docker for desktop doesn't allow you to change the VM filesystem, for this reason it can only be tested in Linux.

I used an AWS VM (Amazon Linux 2) to test these changes, but had to do a few changes to how the root filesystem is mounted to allow the XFS `pquota,prjquota` to take effect.

Screen Shot 2022-06-17 at 09 04 28

## Related issues

Closes #8314

### DIFF
--- a/dev_docs/tbs.md
+++ b/dev_docs/tbs.md
@@ -61,3 +61,13 @@ As mentioned above, APM Servers will buffer events locally until a sampling deci
 For this purpose, we use [BadgerDB](https://github.com/dgraph-io/badger). BadgerDB is designed for
 fast writes, which is important for our use case: we write all events here, and read out only the
 sampled ones. All non-sampled events will be removed from the store through TTL expiry.
+
+BadgerDB doesn't provide a setting to limit disk usage, which is not ideal, since APM Server will
+most likely be running with a limit on storage and allowing BadgerDB to consume all disk space
+will eventually cause the APM Server to malfunction. For this reason, we default to sampling all
+incoming events should writes fail when the disk is full. Eventually, this behavior should be
+configurable and allow users to either discard or sample incoming events when the disk is full.
+
+Furthermore, we should limit how much the BadgerDB size is allowed to grow to, to avoid consuming
+all disk space thus reducing the chances of causing the APM Server to malfunction due no disk space
+available.
--- a/docs/sampling.asciidoc
+++ b/docs/sampling.asciidoc
@@ -150,6 +150,10 @@ Requiring this default policy ensures that traces are only dropped intentionally
 If you enable tail-based sampling and send a transaction that does not match any of the policies,
 APM Server will reject the transaction with the error `no matching policy`.
 
+Please note that by default APM Server doesn't have a hard limit on how much disk space the
+local trace buffer is allowed to consume. For this reason, high throughput deployments may use all
+available disk space.
+
 ===== Example configuration
 
 This example defines three tail-based sampling polices:
--- a/x-pack/apm-server/sampling/eventstorage/storage.go
+++ b/x-pack/apm-server/sampling/eventstorage/storage.go
@@ -140,12 +140,18 @@ func (rw *ReadWriter) WriteTraceEvent(traceID string, id string, event *model.AP
 	if err != nil {
 		return err
 	}
-	return rw.writeEntry(badger.NewEntry(key[:], data).WithMeta(entryMetaTraceEvent).WithTTL(rw.s.ttl))
+	return rw.writeEntry(badger.NewEntry(key[:], data).
+		WithMeta(entryMetaTraceEvent).
+		WithTTL(rw.s.ttl),
+	)
 }
 
 func (rw *ReadWriter) writeEntry(e *badger.Entry) error {
 	rw.pendingWrites++
 	err := rw.txn.SetEntry(e)
+	// If the transaction is already too big to accommodate the new entry, flush
+	// the existing transaction and set the entry on a new one, otherwise,
+	// returns early.
 	if err != badger.ErrTxnTooBig {
 		return err
 	}
--- a/x-pack/apm-server/sampling/processor.go
+++ b/x-pack/apm-server/sampling/processor.go
@@ -11,6 +11,7 @@ import (
 	"path/filepath"
 	"sync"
 	"sync/atomic"
+	"syscall"
 	"time"
 
 	"github.com/dgraph-io/badger/v2"
@@ -144,24 +145,29 @@ func (p *Processor) ProcessBatch(ctx context.Context, batch *model.Batch) error 
 	for i := 0; i < len(events); i++ {
 		event := &events[i]
 		var report, stored bool
+		var err error
 		switch event.Processor {
 		case model.TransactionProcessor:
-			var err error
 			atomic.AddInt64(&p.eventMetrics.processed, 1)
 			report, stored, err = p.processTransaction(event)
-			if err != nil {
-				return err
-			}
 		case model.SpanProcessor:
-			var err error
 			atomic.AddInt64(&p.eventMetrics.processed, 1)
 			report, stored, err = p.processSpan(event)
-			if err != nil {
-				return err
-			}
 		default:
 			continue
 		}
+
+		// If processing the transaction or span returns with an error and
+		// it's caused by the OS returning a "no space left on device", we
+		// sample the transaction or span by default.
+		// NOTE(marclop) This behavior should be configurable so users who
+		// rely on tail sampling for cost cutting, can discard events once
+		// the disk is full.
+		if err != nil && ignoreENOSPCErr(err) == nil {
+			p.logger.Info("received disk is full error, sampling transaction by default")
+			report = true
+		}
+
 		if !report {
 			// We shouldn't report this event, so remove it from the slice.
 			n := len(events)
@@ -274,11 +280,10 @@ func (p *Processor) processSpan(event *model.APMEvent) (report, stored bool, _ e
 		return false, false, err
 	}
 	// Tail-sampling decision has been made, report or drop the event.
-	if !traceSampled {
-		return false, false, nil
+	if traceSampled {
+		atomic.AddInt64(&p.eventMetrics.sampled, 1)
 	}
-	atomic.AddInt64(&p.eventMetrics.sampled, 1)
-	return true, false, nil
+	return traceSampled, false, nil
 }
 
 // Stop stops the processor, flushing event storage. Note that the underlying
@@ -446,11 +451,11 @@ func (p *Processor) Run() error {
 				remoteDecision = true
 			case traceID = <-localSampledTraceIDs:
 			}
-			if err := p.storage.WriteTraceSampled(traceID, true); err != nil {
+			if err := ignoreENOSPCErr(p.storage.WriteTraceSampled(traceID, true)); err != nil {
 				return err
 			}
 			var events model.Batch
-			if err := p.storage.ReadTraceEvents(traceID, &events); err != nil {
+			if err := ignoreENOSPCErr(p.storage.ReadTraceEvents(traceID, &events)); err != nil {
 				return err
 			}
 			if n := len(events); n > 0 {
@@ -490,7 +495,7 @@ func (p *Processor) Run() error {
 			case <-ctx.Done():
 				return ctx.Err()
 			case pos := <-subscriberPositions:
-				if err := writeSubscriberPosition(p.config.StorageDir, pos); err != nil {
+				if err := ignoreENOSPCErr(writeSubscriberPosition(p.config.StorageDir, pos)); err != nil {
 					return err
 				}
 			}
@@ -530,4 +535,26 @@ func sendTraceIDs(ctx context.Context, out chan<- string, traceIDs []string) err
 		}
 	}
 	return nil
+}
+
+// ignoreENOSPCErr returns nil if the error is `syscall.ENOSPC` or os.PathError,
+// which is returned by calls to `os.Created`, if the error doesn't match any
+// of the two errors, the error is returned.
+func ignoreENOSPCErr(err error) error {
+	if err == nil {
+		return nil
+	}
+	// Since most times the "no space left on device" is triggered by writing
+	// to an existing vlog file, check if the error is syscall.ENOSPC first.
+	if errors.Is(err, syscall.ENOSPC) {
+		return nil
+	}
+
+	// If "no space left on device" occurs when a new value log file is created,
+	// then we return after checking that the error is `os.PathError`.
+	var pathErr *os.PathError
+	if errors.As(err, &pathErr) {
+		return nil
+	}
+	return err
 }
--- a/x-pack/apm-server/sampling/eventstorage/badger.go
+++ b/x-pack/apm-server/sampling/eventstorage/badger.go
@@ -27,6 +27,6 @@ func OpenBadger(storageDir string, valueLogFileSize int64) (*badger.DB, error) {
 	if valueLogFileSize > 0 {
 		badgerOpts.ValueLogFileSize = valueLogFileSize
 	}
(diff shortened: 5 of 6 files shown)


x-pack/apm-server/sampling/processor.go
```diff
@@ -144,24 +145,29 @@ func (p *Processor) ProcessBatch(ctx context.Context, batch *model.Batch) error
...
 		}
+
+		// If processing the transaction or span returns with an error and
+		// it's caused by the OS returning a "no space left on device", we
+		// sample the transaction or span by default.
+		// NOTE(marclop) This behavior should be configurable so users who
+		// rely on tail sampling for cost cutting, can discard events once
+		// the disk is full.
+		if err != nil && ignoreENOSPCErr(err) == nil {
+			p.logger.Info("received disk is full error, sampling trace by default")
```
axw, line 167: Might be a good idea to capture another metric for this (count of write failures), updating it in updateProcessorMetrics

x-pack/apm-server/sampling/processor.go
```diff
@@ -446,11 +451,11 @@ func (p *Processor) Run() error {
 				remoteDecision = true
 			case traceID = <-localSampledTraceIDs:
 			}
-			if err := p.storage.WriteTraceSampled(traceID, true); err != nil {
+			if err := ignoreENOSPCErr(p.storage.WriteTraceSampled(traceID, true)); err != nil {
```
axw, line 454: Perhaps instead of ignoring some errors, we should log a warning and continue? At least then we can index what we have.

x-pack/apm-server/sampling/processor.go
```diff
@@ -446,11 +451,11 @@ func (p *Processor) Run() error {
...
 				return err
 			}
 			var events model.Batch
-			if err := p.storage.ReadTraceEvents(traceID, &events); err != nil {
+			if err := ignoreENOSPCErr(p.storage.ReadTraceEvents(traceID, &events)); err != nil {
```
axw, line 458: Similarly, should we log the error rather than bail?

docs/sampling.asciidoc
```diff
@@ -150,6 +150,10 @@ Requiring this default policy ensures that traces are only dropped intentionally
 If you enable tail-based sampling and send a transaction that does not match any of the policies,
 APM Server will reject the transaction with the error `no matching policy`.
 
+Please note that by default APM Server doesn't have a hard limit on how much disk space the
```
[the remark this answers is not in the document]
marclop: do you mean open a public issue on the BadgerDB repo or on the APM Server?

beater/config/sampling.go
```diff
@@ -143,5 +145,7 @@ func defaultTailSamplingConfig() TailSamplingConfig {
 		IngestRateDecayFactor: 0.25,
 		StorageGCInterval:     5 * time.Minute,
 		TTL:                   30 * time.Minute,
+		StorageLimit:          3 * 1024 * 1024 * 1024, // 3GB default size limit?
```
marclop, line 148: I am unsure whether we should default this size to unbounded or to something like 3 or 4GB, which seems like a decent amount of storage to default to. This is kind of an arbitrary threshold.

marclop (commented):
I've updated the code to include a new Limiter component which should give us better guarantees than the current approach.

docs/sampling.asciidoc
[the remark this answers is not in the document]
axw: I meant in apm-server.

axw (commented):
Looks great overall @marclop.

Given that Badger already has its own loop for updating the size asynchronously, I wonder if we could simplify things by not having a separate `Limiter` at all, and directly calling `rw.s.db.Size()` inside `ReadWriter.Flush`? We would just need to pass the size limit into `eventstorage.New`.

x-pack/apm-server/sampling/processor.go
```diff
@@ -143,25 +165,39 @@ func (p *Processor) ProcessBatch(ctx context.Context, batch *model.Batch) error
...
+			var action string
+			switch p.writeFailStrategy {
+			case writeFailDiscardStrategy:
+				action = "discarding"
+				report = false
+			case writeFailSampleStrategy:
+				action = "sampling"
+				report = true
+			}
+			p.rateLimittedLogger.Infof("processing trace failed, %s trace by default", action)
```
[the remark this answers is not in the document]
marclop: thanks! I didn't realize that.

x-pack/apm-server/sampling/eventstorage/storage_limitter.go
```diff
@@ -0,0 +1,121 @@
...
+	"context"
+	"errors"
+	"sync/atomic"
+	"time"
+)
+
+const (
+	defaultStorageLimit = 3 * 1024 * 1024 * 1024 // 3GB storage limit.
+	defaultThreshold    = 0.90                   // Allow 90% of the quota to be used.
+	defaultPeriod       = 100 * time.Millisecond
```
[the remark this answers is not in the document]
marclop: I've removed all the polling and local caches and instead rely on the `db.Size()` calls.

x-pack/apm-server/sampling/eventstorage/storage.go
```diff
@@ -74,14 +107,14 @@ func (s *Storage) NewReadWriter() *ReadWriter {
...
-	pendingWrites int
+	s   *Storage
+	txn *badger.Txn
 
 	// readKeyBuf is a reusable buffer for keys used in read operations.
 	// This must not be used in write operations, as keys are expected to
 	// be unmodified until the end of a transaction.
-	readKeyBuf []byte
+	readKeyBuf    []byte
+	pendingWrites int
```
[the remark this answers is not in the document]
marclop: I moved it to save 8 bytes of padding, it's a very small amount of padding (we'll only save 8 bytes * the number of CPUs on the machine), so I'm happy to revert and leave it where it was too.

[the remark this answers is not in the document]
axw: Ah I see. I just wanted to understand why, happy to keep your change.

beater/config/sampling.go
```diff
@@ -49,6 +49,7 @@ type TailSamplingConfig struct {
 	IngestRateDecayFactor float64               `config:"ingest_rate_decay" validate:"min=0, max=1"`
 	StorageGCInterval     time.Duration         `config:"storage_gc_interval" validate:"min=1s"`
 	TTL                   time.Duration         `config:"ttl" validate:"min=1s"`
+	StorageLimit          uint64                `config:"ttl" validate:"min=1048576"` // 1MB
```
simitt, line 52: `config:"storage_limit" validate:"min=1048576"`

changelogs/head.asciidoc
```diff
@@ -14,6 +14,7 @@ https://github.com/elastic/apm-server/compare/8.4\...main[View commits]
 [float]
 ==== Bug fixes
 - Fix race for deducing destination service fields for OTel bridge {pull}8363[8363]
+- Limit the the local badger database space to 3GB by default for tail-based sampling {pull}8407[8407]
```
simitt, line 17:
```suggestion
When tail-based sampling is enabled, limit the local disk space for the feature to 3GB by default {pull}8407[8407]
```

dev_docs/tbs.md
```diff
@@ -61,3 +61,13 @@ As mentioned above, APM Servers will buffer events locally until a sampling deci
...
 fast writes, which is important for our use case: we write all events here, and read out only the
 sampled ones. All non-sampled events will be removed from the store through TTL expiry.
+
+BadgerDB doesn't provide a setting to limit disk usage, which is not ideal, since APM Server will
+most likely be running with a limit on storage and allowing BadgerDB to consume all disk space
+will eventually cause the APM Server to malfunction. For this reason, we have implemented a storage
+limit based on the size of the Badger database as reported by the SDK. This accounting has some lag
+of up to 1 minute and may allow the database to grow over the configured limit.
+
+Once the limit has been reached, or the writes to badger fail, the default default behavior is to
```
simitt, line 71:
```suggestion
Once the limit has been reached, or the writes to badger fail, the default behavior is to
```

stuartnelson3, line 71: there's an extra `default`

docs/sampling.asciidoc
```diff
@@ -150,6 +150,10 @@ Requiring this default policy ensures that traces are only dropped intentionally
 If you enable tail-based sampling and send a transaction that does not match any of the policies,
 APM Server will reject the transaction with the error `no matching policy`.
 
+IMPORTANT: Please note that from version `8.3.1` APM Server implements a storage limit of 3GB
```
stuartnelson3, line 153: default storage limit?

simitt (commented):
The new config option was not introduced to `apm-server.yml` and neither to the `apmpackage` config options as a default value. 
Is it intended that users can configure the size or not yet? (in which case, please remove the word _default_ from the changelog description).

dev_docs/tbs.md
```diff
@@ -61,3 +61,13 @@ As mentioned above, APM Servers will buffer events locally until a sampling deci
 For this purpose, we use [BadgerDB](https://github.com/dgraph-io/badger). BadgerDB is designed for
 fast writes, which is important for our use case: we write all events here, and read out only the
 sampled ones. All non-sampled events will be removed from the store through TTL expiry.
+
+BadgerDB doesn't provide a setting to limit disk usage, which is not ideal, since APM Server will
+most likely be running with a limit on storage and allowing BadgerDB to consume all disk space
+will eventually cause the APM Server to malfunction. For this reason, we have implemented a storage
+limit based on the size of the Badger database as reported by the SDK. This accounting has some lag
+of up to 1 minute and may allow the database to grow over the configured limit.
```
stuartnelson3, line 69: I would say the first half of this paragraph can be dropped. The fact that badger doesn't have a limit isn't (I am guessing) important to the user; they can now set a limit which we roughly enforce. Maybe something like:

>We have implemented a configurable storage
limit based on the size of the Badger database as reported by the SDK. This accounting has some lag
of up to 1 minute and may allow the database to grow over the configured limit.

stuartnelson3: Ah, and maybe mentioning that the limit can be disabled by setting its value to 0

marclop:
run elasticsearch-ci/docs

marclop:
I was testing the change locally and it isn't working as I was expecting, and the main reason seems to be because majority of the errors are coming from reading from badger, rather than writing.https://github.com/elastic/apm-server/blob/1278851cb903790c618b91b90d94c261ab4999b9/x-pack/apm-server/sampling/eventstorage/storage.go#L169-L181

Each time we read a local sampling decision, we flush the existing transaction where the writes are buffered if the pendingWrites are > 100.

https://github.com/elastic/apm-server/blob/1278851cb903790c618b91b90d94c261ab4999b9/x-pack/apm-server/sampling/processor.go#L433-L455

Then on the write path which `ProcessBatch` takes, the writes are flushed only when the transaction is too big. Which in general will be triggered much less often due to the `ReadTraceEvents` method attempting to flush the transaction when it has more than 100 written events to it.

https://github.com/elastic/apm-server/blob/1278851cb903790c618b91b90d94c261ab4999b9/x-pack/apm-server/sampling/eventstorage/storage.go#L146-L156

This results in unexpectedly dropping most if not all traces rather than sampling them. In the image below, we can see that no more traces are indexed into Elasticsearch after 58'.

Screen Shot 2022-06-21 at 16 09 08

The APM UI shows that the ingest rate is the same since it's powered by the metric documents, which aren't processed by the tail sampling processor.

Screen Shot 2022-06-21 at 16 09 48

I tested this locally with the updated instructions in the How to test this section in this PR.

----

We have a few options regarding before we merge this in, ordered by effort:

1. We don't significantly change how this is working and just update the docs to say that the traces are discarded by default when the storage is updated.
2. We change `ReadWriter.ReadTraceEvents` to not flush the txn on 100 stored events (@axw left a comment that we do it to speed up searching) and potentially use the number of pending writes where we flush the transaction (on `ReadWriter.writeEntry`). We may still lose some (but less) traces since `p.storage.WriteTraceSampled` is called outside of the ProcessBatch method.

### Option 1

It's by far the less involved option, but we would be introducing a change in behavior (although we're fixing a bug which currently will cause the APM Server to shut down) in `v8.3.1`. Traces will be lost and users may be unhappy about that. The metric sets that power the UI will continue to be indexed, but the product won't be vey useful for distributed tracing purposes.

### Option 2

Much more involved and may have some performance implications, but will most likely result in a more predictable outcome and won't "break" the product. Although it'll result in tracing nearly 100% of the received traces (minus those lost).

I've added one last commit to this PR (2324a8905). The results can be observed below. The event rate is mostly unchanged (towards the end the benchmark stopped, so the ingest rate decayed, I ran it with 20m of bench time). But the traces kept being indexed instead of all indexing being stopped.

Screen Shot 2022-06-21 at 17 09 36

Screen Shot 2022-06-21 at 17 09 39

beater/config/sampling.go
```diff
@@ -49,6 +50,8 @@ type TailSamplingConfig struct {
 	IngestRateDecayFactor float64               `config:"ingest_rate_decay" validate:"min=0, max=1"`
 	StorageGCInterval     time.Duration         `config:"storage_gc_interval" validate:"min=1s"`
 	TTL                   time.Duration         `config:"ttl" validate:"min=1s"`
+	StorageLimit          string                `config:"storage_limit"`
```
axw, line 53: It doesn't have to be in this PR, but it might be nice to make a `type Bytes uint64` with its own Unpack method which does the parsing.

x-pack/apm-server/sampling/eventstorage/storage.go
```diff
@@ -140,12 +183,24 @@ func (rw *ReadWriter) WriteTraceEvent(traceID string, id string, event *model.AP
...
+	return rw.writeEntry(badger.NewEntry(key[:], data).
+		WithMeta(entryMetaTraceEvent).
+		WithTTL(rw.s.ttl),
+	)
 }
 
 func (rw *ReadWriter) writeEntry(e *badger.Entry) error {
 	rw.pendingWrites++
 	err := rw.txn.SetEntry(e)
+	// Attempt to flush if there are 200 or more uncommitted writes.
```
axw, line 195:
```suggestion
	// Attempt to flush if there are 200 or more uncommitted writes.
	// This ensures calls to ReadTraceEvents are not slowed down;
	// ReadTraceEvents uses an iterator, which must sort all keys
	// of uncommitted writes.
```

axw (approved):
Thanks for the benchmarking, looks great.

axw:
Confirmed with 8.4.0-BC1. I created a deployment in cloud, connected an agent running locally on my machine, and modified the Fleet package policy to set the hidden var. Otherwise I followed the steps in the description.

Issue #8177, opened 2022-05-20
Document APM Server support for otlp http Intake via protobuf


simitt:
Document that APM Server also supports otlp protobuf over HTTP intake (once https://github.com/elastic/apm-server/issues/6128 is finished, [draft PR](https://github.com/elastic/apm-server/pull/8156) is up). 

* Describe the protobuf is now supported via `grpc` and `http` ([protobuf via http](https://github.com/open-telemetry/oteps/blob/main/text/0099-otlp-http.md))
* Update the [known limitations](https://www.elastic.co/guide/en/apm/guide/current/open-telemetry.html#open-telemetry-otlp-limitations) in the APM Server docs. The server only supports protobuf via http, not [`json` encoding](https://github.com/open-telemetry/oteps/blob/main/text/0122-otlp-http-json.md), mainly because the json encoding is not widely adopted yet. 
* Update the [architecture](https://github.com/elastic/apm-server/blob/main/ARCHITECTURE.md) to include `http`

PR #8429, opened 2022-06-21
Document APM Server support for OTLP/HTTP

bmorelli25:
### Summary

- [x] Describes that ProtoBuf is now supported via gRPC and HTTP.
- [x] Updates the known limitations to explain that JSON encoding is not yet supported.
- [x] Updates the architecture diagram to include HTTP from 8.3 onwards.

### Related

* Closes https://github.com/elastic/apm-server/issues/8177
* Closes https://github.com/elastic/observability-docs/issues/1930


bmorelli25:
For the architecture diagram, all I ended up doing was adding another blue (OTLP) box that says `HTTP` and updating the title. I'm not sure if there's more to do. Elastic employees can better view the changes [here](https://whimsical.com/ingest-flow-7rLfKj6HddmnJghU8RchD3).

docs/open-telemetry.asciidoc
```diff
@@ -26,10 +28,10 @@ business KPIs and technical components with the {stack}.
 [[open-telemetry-native]]
 ==== APM Server native support of OpenTelemetry protocol
 
-IMPORTANT: The https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/elasticexporter#legacy-opentelemetry-collector-exporter-for-elastic[OpenTelemetry Collector exporter for Elastic]
-was deprecated in 7.13 and replaced by the native support of the OpenTelemetry Line Protocol in
-Elastic {observability} (OTLP). To learn more, see
-https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/elasticexporter#migration[migration].
+****
+Starting in {stack} version 8.3,
```
axw, line 32: We've had native support for OTLP/gRPC since 7.13, it's only OTLP/HTTP that was added in 8.3. Am I misreading this?

axw (commented):
Looks good, just one comment about the version.

docs/open-telemetry.asciidoc
bmorelli25: If you're "misreading it," then others might too. I'll make it clearer that only OTLP/HTTP is new.

axw: Maybe I just misunderstood the purpose of this callout. Is it like a "what's new" type of thing? Slightly concerned that new users will take the statement to mean we support OTLP/HTTP (only) as of 8.3, and the non-mention of OTLP/gRPC may suggest that we don't support that.

bmorelli25: Yes—"what's new". That's how I interpreted the first bullet point of the referenced issue. I'm all for just removing this and letting the changelog and official "Observability what's new" do the talking.

axw: I think that might be best. My suggestion is:
 - in the 8.3.0 "Observability what's new", mention that we now support OTLP/HTTP as well as OTLP/gRPC
 - don't specifically highlight that we added OTLP/HTTP in 8.3 here
 - remove lack of OTLP/HTTP from limitations
 - remove the "IMPORTANT: ..." bit about the old opentelemetry-collector-contrib and introduction of native OTLP support in 7.13. 7.13 was quite a while ago, and the [opentelemetry-collector-contrib exporter docs](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/elasticexporter) mention that it is deprecated and point here. I don't think we need to continue drawing attention to the old deprecated thing.

@simitt @cyrille-leclerc any concerns?

simitt: Andrew's suggestions sound good to me; 
the old exporter is only deprecated and not removed, so I am on the fence on mentioning this in the docs, but I am ok with only referencing them directly from the opentelemetry-collector-contrib-exporter as suggested.

axw: OK, sounds good. Thanks @bmorelli25!

axw (approved):

docs/open-telemetry.asciidoc
bmorelli25: Thanks for your help!

PR #8434, opened 2022-06-22
Build apm-server binary for darwin/arm64

lahsivjar:
## Motivation/summary

Enable build for arm64 arch

## Checklist

## How to test these changes

Run `make` on arm64

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,7 +31,7 @@ APM_SERVER_BINARIES:= \
 	build/apm-server-darwin-amd64
 
 .PHONY: $(APM_SERVER_BINARIES)
-$(APM_SERVER_BINARIES): $(MAGE)
+$(APM_SERVER_BINARIES) build/apm-server-darwin-arm64: $(MAGE)
 	@$(MAGE) build
 
 build/apm-server-linux-%: export GOOS=linux


magefile.go
```diff
@@ -55,6 +55,7 @@ func init() {
 		"windows/386",
 		"windows/amd64",
 		"darwin/amd64",
+		"darwin/arm64",
```
lahsivjar, line 58: [For reviewers] Not sure about this since [v7 beats](https://github.com/elastic/beats/blob/v7.0.0/dev-tools/mage/platforms.go#L37) doesn't seem to support `Default` for `darwin/arm64`

axw (commented):
@lahsivjar what happens when you run `make` on darwin/arm64 without the changes?

APM_SERVER_BINARIES is meant to be the set of binaries that we package and release. By modifying that, `make release` will start building darwin/arm64. That's not desirable, since we don't currently want to support that platform.

I _think_ you just need to update `magefile.go`. Can you please add a comment next to it noting that it's just there for development purposes?

axw:
> The error seems to indicate that it cannot find any rule for the target build/apm-server-darwin-arm64 in makefile so we will have to add another target in makefile I think. Maybe something like:

Ah I see. Yes, your patch there looks good.

axw (approved):

lahsivjar:
@axw Updated the PR to include the agreed changes in Makefile. I have also reverted the changes in `magefile.go` because it doesn't seem to be affecting the build. IIUC, `Platforms` is used by [CrossBuild](https://github.com/elastic/beats/blob/v7.0.0/dev-tools/mage/crossbuild.go#L114) whereas we are using [Build](https://github.com/elastic/beats/blob/v7.0.0/dev-tools/mage/build.go).

axw:
> IIUC, Platforms is used by [CrossBuild](https://github.com/elastic/beats/blob/v7.0.0/dev-tools/mage/crossbuild.go#L114) whereas we are using [Build](https://github.com/elastic/beats/blob/v7.0.0/dev-tools/mage/build.go).

So it is - good catch. I'll create a PR to remove the setting of Platforms in our magefile.

PR #8438, opened 2022-06-22
Sourcemap upload documentation: align curl example with custom app example

jin-harmoney:
## Motivation/summary

The examples concerning the `bundle_filepath` are not in line. The goal of this PR is to align them.

The documentation mentions: `bundle_filepath` - The absolute path of the final bundle as used in the web application.

So I'm not sure if should now be `http://localhost/app.min.js` or just `app.min.js`? This change aligns both examples, but I'm still not 100% sure if this is fully correct. I do however know that using `http://localhost/app.min.js` worked when we used Elastic v7.

## Checklist


simitt:
Apologies for coming back to you that late. The `bundle_filepath` is expected to be working also with a relative path.

axw:
@jin-harmoney thanks for the PR, but as @simitt said above it's expected that a relative `bundle_filepath` should work. We had a bug with that which will be fixed in 8.5.0: https://github.com/elastic/apm-server/issues/8650

Issue #8383, opened 2022-06-15
"Output Events Rate" in stack monitoring is always zero


axw:
**APM Server version** (`apm-server version`): 8.3.0-BC4

**Description of the problem including expected versus actual behavior**:

"Output Events Rate" in stack monitoring is always zero.

**Steps to reproduce**:

 1. Start 8.3.0-BC4 with stack monitoring enabled.
 2. Send some events, check that they show up in the APM UI.
 3. Navigate to stack monitoring, observe the "Output Events Rate" chart is always reporting zero.

![image](https://user-images.githubusercontent.com/843579/173733244-3139f554-a2dd-47b0-b674-01945ba68ec3.png)

axw:
Hmm, I just reconfigured the integration with expvar enabled, and now it's working. Maybe there's race condition?

axw:
Happened again after upgrading from 8.2.3 to 8.3.0-BC4. Initially the output was zero, after reconfiguring the integration (this time changing the event rate limit), the output went non-zero.

axw:
This is apparently still an issue, at least in system tests, as seen here:

https://apm-ci.elastic.co/blue/organizations/jenkins/apm-server%2Fapm-server-mbp%2FPR-9014/detail/PR-9014/1/pipeline/

lahsivjar:
I haven't been able to reproduce this exact error. However, due to the way our instrumentation works it is possible that after a reload event the old modelindexer is still receiving data while the instrumentation has moved to the new modelindexer. This is due to the fact that we [wait for the old modelindexer to gracefully shutdown](https://github.com/elastic/apm-server/blob/653897827e1aaee5b9dd461b3be0fe5c7220f3f0/internal/beater/beater.go#L383-L388) however, we [switch the monitoring to new modelindexer](https://github.com/elastic/apm-server/blob/653897827e1aaee5b9dd461b3be0fe5c7220f3f0/internal/beater/beater.go#L702-L706) before the old one exits.

The above will result in the instrumentation data to report `0` until the old indexer shuts down.

simitt:
Moving this to backlog since we haven't spend more time recently to track this down.

tegenterter:
It appears that this bug lead up to an incident (https://github.com/elastic/cloud/issues/110723) and should be prioritized

simitt:
Moved it into the `8.7` milestone again to be picked up and verified if this is still a bug in current versions.

axw:
[...]

PR #8439, opened 2022-06-22
add config reload debouncer

stuartnelson3:
prevent multiple reloads in rapid succession.

to be tested in cloud, an attempt at seeing if this fixes the bug setting the output registry (https://github.com/elastic/apm-server/issues/8383)

## how to test

1. Start the stack in cloud
2. In the admin console, enable logs and metrics. Send them to this deployment.
3. Send events to the apm-server (grab the secret token from the apm integration, but do not update the integration!)
4. Navigate to stack monitoring; observe that metrics appear under "Output Events Rate"


axw (commented):
Did this fix the issues seen in Cloud? If so, why, and can we add a test to cover whatever the problem was? The reload stuff is already complicated, I'm wary of making it even more so.

beater/beater.go
```diff
@@ -264,6 +273,73 @@ type reloader struct {
...
+// will contain the return value of fn is returned to the caller, which they
+// are responsible for draining.
+func (d *debouncer) trigger(ctx context.Context) <-chan error {
+	res := make(chan error, 1)
+	select {
+	case d.active <- struct{}{}:
+		// Update the internal state to show we're currently debouncing
+		// fn execution and start the debounce method.
+		go func() {
+			d.debounce(ctx)
```
marclop, line 302: Don't we need to send the result of debounce back to res? Otherwise the error won't be propagated and line 308 will never happen either.

stuartnelson3: I believe that, when we send `res` on line 308, that will be added to the slice of channels that will have the error sent to them. I think this is being verified in the test.

However, maybe the code would be more readable if we send on this channel, and move `d.triggerc <- res` under `default`?

stuartnelson3:
After building apm-server locally and packaging that into an elastic-agent container which is then run on cloud, it appears that the issue isn't related to the actual code within the apm-server, but potentially something around the compilation in CI. The locally built apm-server binary did not have the same bug as is seen with the BC created by CI, so this fix is likely unnecessary. Potentially in the future it might be nice to reduce the number of reloads if both an update to the config and the output config appear in rapid succession.

Issue #8303, opened 2022-06-03
Run automated smoke tests on ESS


simitt:
## Problem

We want to own automated tests of the APM Server setup (with and without Elastic Agent) on Elastic ESS.

## Solution
The APM Server systemtests should contain an ESS testing package that runs the tests specified under _Validation Criteria_. Tests are mostly testing the overall setup, that all involved components on ESS work together as expected. 

There is no requirement to run the tests on a PR basis, but rather on a daily basis for now.  

The tests have to create a deployment on ESS on which a specified test set can then be run. 
For the deployment creation we will receive support from the eng productivity team (cc @cachedout) and try to leverage some of the work that has been done for benchmarking. 

### Validation Criteria

* Daily Jenkins job is set up to run a defined set of APM Server tests on the ESS infrastructure.

The tests suit runs tests for:
* standalone APM Server (created in `7.17`)
* managed APM Server (created in `8.0`)
* standalone APM Server migrated to managed APM Server (`7.17`)

We want to test upgrades for:
* `7.17.x-1` upgrading to `7.17.x`
* `8.x`: the last released version to the new minor version under development

We want to run following test cases:
  * data ingestion to APM Server for backend ~and RUM~ agents where events are successfully stored to the expected data streams. 
  * fetching index & component templates, ILM policies and index pipelines and verifying that they are set up in the expected version

## No-gos
In this first iteration we focus on ESS; ECE is out of scope.
We do not focus on triggering the tests on a PR basis for now.

cachedout:
Hi @simitt 

> For the deployment creation we will receive support from the eng productivity team (cc @cachedout) and try to leverage some of the work that has been done for benchmarking.

Regarding the ESS provisioning piece, there are a few options available:

1. You can [the ESS Terraform](https://github.com/elastic/terraform-provider-ec/) provider directly. This will provide you with maximum control over the deployment and the way that it is configured for testing.
2. You can use `oblt-cli` to create a cluster and then tear it back down again. Depending on your specific needs, we may need to create a specific recipe to deploy the cluster according to your needs.

Regarding the provisioning of RUM agents, could you please talk a bit more about what you envision here? Does it specifically need to be RUM data that's ingested or will any APM data be sufficient?

cc: @kuisathaverat for any additional thoughts.

simitt:
Update: I removed the work on RUM agents from the scope for `8.4` work.

marclop:
[...]

PR #8458, opened 2022-06-23
testing/smoke: Add initial smoke tests

marclop:
## Motivation/summary

This patch adds an initial set of smoke tests which use the existing
`ec_deployment` module for the underlying infrastructure and a bash
script to drive the scripted smoke test cases. The script takes in a
`MAJOR.MINOR` version combination.

The initial smoke tests are slightly different depending if the major
version is `7` or `8`, the former will be test upgrades `7.17.last-1` to
`7.17.last`, while the latter, will upgrade from the previous minor to
the current minor (`8.3.0`) will upgrade from `8.2.3` to `8.3.0`.

The initial set of assertions is kept to a minimum to enable iterating
over the approach rather than adding all the assertions in the first PR.

Initially, we're testing that:

- Create deployment succeeds
- Data is sent to APM Server
- Each of the events sent to APM Server can be found in Elasticsearch.
- Upgrade to the next version succeeds.
- Data is sent to the APM Server
- Each of the events sent to APM Server can be found in Elasticsearch.

As the last step and if any errors occur during the smoke test, the
underlying infrastructure will be destroyed.

I've chosen to use bash since it doesn't seem too bad with the current
set of assertions, we can always improve it and write it in Go, but it
seems more involved than what we need right now.

I've created a [gist with the output](https://gist.github.com/marclop/81d567ecafd23058c221e41f147ee2aa) of the tested execution.

## Checklist

## How to test these changes

1. `export your cloud credentials (export EC_API_KEY)`
2. `make smoketest SMOKETEST_VERSIONS=7.17,latest`
3. Verify that the exit status is 0 and that the smoke tests all passed (check for 4 instances of the log message `-> Smoke tests passed!`)

## Related issues

Part of #8303


lahsivjar (approved):
LGTM with 1 concern, since we are using local backend, in the event of SIGKILL the trap command will not destory terraform (cases like runners dying). This might lead to some cases where the infra will continue to exist like a zombie.

marclop:
@lahsivjar I was thinking about using a cleanup task in jenkins even after the script finishes that runs `terraform destroy -auto-approve` so that in the event the trap code doesn't fire, the infrastructure is destroyed.

This gives us the advantage of 1. we always try and clean up locally, 2. Jenkins will do it twice, and in the event the infrastructure has already been destroyed, the `terraform destroy -auto-approve` is a no-op.

PR #8471, opened 2022-06-24
ci: enable report benchmark diff on a PR basis

v1v:
## Motivation/summary

- Enable micro-benchmark on a PR basis too
- Run `benchstat` in a Jenkins shared library (so others can consume it) to generate diff report 
- Notify go benchmark report on a GitHub comment

## Issues

Consumes https://github.com/elastic/apm-pipeline-library/pull/1748
Requires https://github.com/elastic/apm-server/pull/8470

## Test

See https://github.com/elastic/apm-server/pull/8471#issuecomment-1165633680


Jenkinsfile
```diff
@@ -440,6 +434,7 @@ pipeline {
                   sh(label: 'Run benchmarks', script: './.ci/scripts/bench.sh')
                 }
                 sendBenchmarks(file: "bench.out", index: "benchmark-server")
+                generateGoBenchmarkDiff(current: 'bench.out', filter: 'exclude')
```
v1v, line 437:
```suggestion
                generateGoBenchmarkDiff(file: 'bench.out', filter: 'exclude')
```

Jenkinsfile
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env groovy
-@Library('apm@current') _
+@Library('apm@test/report-bench-diff') _
```
v1v, line 2: To be reverted

Jenkinsfile
```diff
@@ -30,15 +30,15 @@ pipeline {
...
     booleanParam(name: 'windows_ci', defaultValue: false, description: 'Enable Windows CI')
-    booleanParam(name: 'intake_ci', defaultValue: true, description: 'Enable test')
-    booleanParam(name: 'test_ci', defaultValue: true, description: 'Enable test')
-    booleanParam(name: 'test_sys_env_ci', defaultValue: true, description: 'Enable system and environment test')
+    booleanParam(name: 'intake_ci', defaultValue: false, description: 'Enable test')
+    booleanParam(name: 'test_ci', defaultValue: false, description: 'Enable test')
+    booleanParam(name: 'test_sys_env_ci', defaultValue: false, description: 'Enable system and environment test')
     booleanParam(name: 'bench_ci', defaultValue: true, description: 'Enable benchmarks')
-    booleanParam(name: 'release_ci', defaultValue: true, description: 'Enable build the release packages')
+    booleanParam(name: 'release_ci', defaultValue: false, description: 'Enable build the release packages')
```
v1v, line 41: To be reverted

Jenkinsfile
v1v:
```suggestion
@Library('apm@current') _
```

kuisathaverat (approved):

marclop:
@v1v looking at the benchmark report diff, there seems to be an unexpectedly large difference between the main branch result and this PR (5% up to 28% slower performance in the PR runs 🤔):

```
CompressedRequestReader/gzip_content_encoding-4                                                  9.55µs ± 8%   12.25µs ±11%  +28.27%  (p=0.008 n=5+5)
CompressedRequestReader/deflate_sniff-4                                                          10.0µs ± 7%    11.3µs ±18%  +13.36%  (p=0.032 n=5+5)
pkg:github.com/elastic/apm-server/model/modelindexer goos:linux goarch:amd64
ModelIndexer/BestSpeed-4                                                                         7.14µs ± 1%    7.71µs ± 7%   +8.10%  (p=0.008 n=5+5)
ModelIndexer/DefaultCompression-4                                                                16.1µs ± 3%    17.5µs ± 4%   +8.70%  (p=0.008 n=5+5)
ModelIndexer/BestCompression-4                                                                   22.2µs ± 2%    23.9µs ± 3%   +7.75%  (p=0.008 n=5+5)
```

Are we running the PR micro-benchmarks in the same "environment" or hardware that the main branch benchmarks are being run on? If we're not, we can't compare them as we are.

v1v:
> Are we running the PR micro-benchmarks in the same "environment" or hardware that the main branch benchmarks are being run on?

The existing benchmark stage in the CI runs on ephemeral workers (Cloud based one) that are always based on the same specs.

See https://github.com/elastic/apm-server/blob/c600c1bc3feb9db51c97d0cd37e760ff7fe8eedc/Jenkinsfile#L419

If those benchmarks should run on baremetal then we can enable it, though we might need to review the hardware specs

marclop:
@v1v given the performance delta when there aren't any significant changes, I think that would be best, since having such a large delta defeats the purpose of having the automated micro benchmark comparison run on PRs if we can't reliably trust it.

As long as the benchmarks are run on an idle bare metal machine we should be getting consistent results between runs.

v1v:
@marclop , by doing so, we will be limited to the existing number of static workers, only 3 workers -> https://apm-ci.elastic.co/label/linux/. And one of those workers is 32 Ram based versus the other two that are double. I don't know if that's an issue.

There are some other baremetals, but those are aimed to be used by the profiling team.

Question:

1) Should we use baremetals then? If so, I'll put a new PR only for using the baremetals.
2) Does the benchmark runs on the host or it uses any kind of containers?
3) I'll unblock the issue we had regarding deprecating the existing 32 Gb Ram worker to be similar to the other two.

marclop:
@v1v ideally we'd always run the benchmarks in exactly the same hardware to minimize differences, 32gb of ram should be plenty. It's OK to only run the micr-benchmarks in one machine, they shouldn't take a long time.

> Should we use baremetals then? If so, I'll put a new PR only for using the baremetals.
I think that would be best.

> Does the benchmark runs on the host or it uses any kind of containers?

Runs on the host via `go test -bench=./...` so we'd need to make sure Go is available and using the appropriate version.

> I'll unblock the issue we had regarding deprecating the existing 32 Gb Ram worker to be similar to the other two.

I'm fine deprecating the odd 32GB ram worker. For us, it shouldn't matter as long as the hardware is consistently the same.

v1v:
Thanks for all the details, https://github.com/elastic/apm-server/pull/8533 is the one driving the change to use the baremetals instead.

I won't merge this PR until the above one is merged so, we can compare results again.

OTOH, I'll proceed with deprecating the 32 Gb ram in favour of a new 64 Gb ram one, to avoid any deviation in the HW

v1v:
@marclop , the benchmarking is now executed in a subset of the baremetals. I just double checked that the `main` branch was triggered and used [worker-1095690](https://apm-ci.elastic.co/job/apm-server/job/apm-server-mbp/job/main/467/execution/node/199/log/) -> https://apm-ci.elastic.co/computer/worker-1095690/

While the last build for this PR used the [worker-1095690](https://apm-ci.elastic.co/job/apm-server/job/apm-server-mbp/job/PR-8471/19/execution/node/216/log/) -> https://apm-ci.elastic.co/computer/worker-1095690/

If I run:

```bash
wget https://apm-ci.elastic.co/job/apm-server/job/apm-server-mbp/job/main/467/artifact/bench.out -O main.out
wget https://apm-ci.elastic.co/job/apm-server/job/apm-server-mbp/job/main/467/artifact/bench.out -O pr.out
benchstat main.out pr.out | grep -v 'all equal' | grep -v '~'
```

Then I don't see any major diffs

But https://github.com/elastic/apm-server/pull/8471#issuecomment-1165633680 is based on the diff of [main@464](https://apm-ci.elastic.co/job/apm-server/job/apm-server-mbp/job/main/464/execution/node/199/log/) that used a different bare-metal

```bash
wget https://apm-ci.elastic.co/job/apm-server/job/apm-server-mbp/job/main/467/artifact/bench.out -O main.out
wget https://apm-ci.elastic.co/job/apm-server/job/apm-server-mbp/job/main/467/artifact/bench.out -O pr.out
benchstat main.out pr.out | grep -v 'all equal' | grep -v '~'
```

My proposal:
- Fix the bare-metal to https://apm-ci.elastic.co/computer/worker-1225339/ until we normalise all the existing baremetals (might take a couple of weeks or so)
  - Pros: we unblock this PR
  - Cons: if several concurrent builds then there is a chance they might take a bit long 

What are your thoughts?

PR #8527, opened 2022-07-04
Chore: retry benchmarks once when failed

pazone:
## Motivation/summary

Some low % of benchmark executions suffer of cloud connectivity issues.
The entire benchmark stage will be retried in event of any error
ESS and executor(EC2) instances will be recreated. 

TL;DR Added retry option to the benchmark job. If it's failed once - it will be retried. 

## How to test these changes
run it in jenkins

### DIFF
--- a/.ci/benchmarks.groovy
+++ b/.ci/benchmarks.groovy
@@ -36,7 +36,10 @@ pipeline {
     }
 
     stage('Benchmarks') {
-      options { skipDefaultCheckout() }
+      options {
+        skipDefaultCheckout()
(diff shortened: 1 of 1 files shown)


v1v (approved):
LGTM, thanks for the clarification. So we don't reuse the existing environment but try again from scratch to avoid any kind unexpected misbehaviours with the existing environment

pazone:
In case of 1 retry the benchmarks can take up to 3-4 hours.

marclop (approved):
LGTM

PR #8590, opened 2022-07-12
Implementing JVM discovery for Java Agent attacher integration

### LINKED ISSUE #8304: Support the java agent team in making changes to the java agent attacher
@felixbarny & the java agent team will be driving the java agent attacher improvement work, described in https://github.com/elastic/apm-agent-java/issues/2649.

We need to ensure to provide apm-server and golang specific knowledge for supporting them.

eyalkoren:
Closes #8304 
Context: https://github.com/elastic/apm-agent-java/issues/2649

So far, JVM discovery was done by a Java application (the attacher) that first and foremost requires Java that is accessible and executable by the APM Server user. 
The main motivation behind this PR is to discover the Java executable for each JVM, as well as the user that is used to run it, and invoke the attacher through it. I transported the JVM discovery rules logic as well, which has some benefits over trying to discover all data by the attacher, but we can decide later on whether we keep it or not.

Unit tests are definitely still missing, I will add them later. 
Windows is currently not fully supported and requires further investigation. 
Eventually it would be tested e2e as described in https://github.com/elastic/apm-agent-java/issues/2196.


beater/java_attacher/java_attacher.go
```diff
@@ -22,61 +22,389 @@ import (
...
-			cfg.JavaBin = bin
 		}
-	} else {
-		// Ensure we're using the correct separators for the system
-		// running apm-server
-		cfg.JavaBin = filepath.FromSlash(cfg.JavaBin)
+	}()
+}
+
+func (j *JavaAttacher) discoverJvmsForAttachment(ctx context.Context) (map[string]*JvmDetails, error) {
```
simitt, line 154: similar to comment below - when you expect that the return value should be used even when an error is returned, I'd rather log the error in this function, and change the signature to not return an error, and return `nil` or `jvms` accordingly

simitt (commented):
Generally looks good already, left a couple of comments, mostly related to error handling.

beater/java_attacher/discovery_rules.go
```diff
@@ -0,0 +1,61 @@
...
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package javaattacher
+
+import "regexp"
+
+type DiscoveryRule interface {
```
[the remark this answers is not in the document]
eyalkoren: 👍 
Yeah, I didn't at first, but coming from Java, it was too hard for me to not being able to easily see those are interface and implementations 🙂

beater/java_attacher/java_attacher.go
```diff
@@ -22,61 +22,389 @@ import (
...
+					"",
+				}
+			}
+		}
+	}
+	if err := scanner.Err(); err != nil {
+		j.logger.Errorf("error reading output for command '%v': %v", strings.Join(cmd.Args, " "), err)
+	}
+	err = cmd.Wait()
+	return jvms, err
```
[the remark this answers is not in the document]
eyalkoren: Very useful feedback, thanks!

beater/java_attacher/java_attacher.go
```diff
@@ -22,61 +22,389 @@ import (
...
+		defer close(c)
+		wg.Wait()
+	}()
+	select {
+	case <-ctx.Done():
+		return false
+	case <-c:
+		return false
+	case <-time.After(timeout):
+		return true // timed out
```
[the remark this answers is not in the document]
eyalkoren: The returns indicates whether it returned due to timeout or not. 
If `true`, I can handle accordingly (log that it timed out). I less need to distinguish between proper completion and cancellation due to context close. Do you think it should be reversed and logging about time out better be done here?

beater/java_attacher/java_attacher.go
```diff
@@ -22,61 +22,389 @@ import (
...
+
+func (j *JavaAttacher) executeForEachJvm(ctx context.Context, jvms map[string]*JvmDetails,
+	executable func(ctx context.Context, jvm *JvmDetails) error, removeOnError bool, timeout time.Duration) bool {
+	var wg sync.WaitGroup
+	for pid, jvm := range jvms {
+		wg.Add(1)
+		go func(jvm *JvmDetails, pid string) {
+			err := executable(ctx, jvm)
+			if err != nil {
+				j.logger.Debug(err)
```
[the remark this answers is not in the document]
eyalkoren: No, better `debug` here - either can't get accurate start time (we have a default), or can't get full cmd args (in which case some discovery rules won't work) or failed to run `java --version`, which is expected if the discovered process is not actually a Java process. In all cases we should proceed, but better be able to debug.

beater/java_attacher/java_attacher.go
```diff
@@ -22,61 +22,389 @@ import (
...
 	}
+	if err := cmd.Start(); err != nil {
+		return err
+	}
+	scanner := bufio.NewScanner(stdout)
+	if scanner.Scan() {
+		jvmDetails.startTime = scanner.Text()
+	}
+	if err := scanner.Err(); err != nil {
+		j.logger.Errorf("error obtaining accurate start time for process %v using 'ps -o lstart=': %v", err)
```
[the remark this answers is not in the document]
eyalkoren: no good reason apparently, I will change

beater/java_attacher/java_attacher.go
```diff
@@ -22,61 +22,389 @@ import (
...
+	if err := scanner.Err(); err != nil {
+		j.logger.Errorf("error reading output of command '%v' when running as user %v: %v", strings.Join(cmd.Args, " "), jvm.user, err)
+	}
+	return cmd.Wait()
+}
+
+func (j *JavaAttacher) setRunAsUser(jvm *JvmDetails, cmd *exec.Cmd) error {
+	currentUser, err := user.Current()
+	if err != nil {
+		j.logger.Infof("failed to get the current user: %v", err)
```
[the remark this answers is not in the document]
eyalkoren: I don't think so. In such case, the user will be the zero value, which means that we will still try to run as the JVM user.
AFAIK, we currently don't expect the current user to be anything other than `root`, in which case I want to try to run as the JVM user by default.

beater/java_attacher/java_attacher.go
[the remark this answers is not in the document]
eyalkoren: Since there are other errors that I want to return **instead of a list**, I will keep the signature, but make sure I either return non-nil error or non-nil list. I believe that follows the convention.

beater/java_attacher/java_attacher.go
```diff
@@ -22,61 +22,389 @@ import (
...
+	if err := cmd.Start(); err != nil {
+		return err
+	}
+	scanner := bufio.NewScanner(stdout)
+	if scanner.Scan() {
+		// only use first line of the output
+		jvm.version = scanner.Text()
+	}
+	if err := scanner.Err(); err != nil {
+		j.logger.Errorf("error reading output of command '%v' when running as user %v: %v", strings.Join(cmd.Args, " "), jvm.user, err)
```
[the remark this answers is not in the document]
eyalkoren: Isn't there a risk to not `Wait()` for the command to exit, thus accumulating unbounded number of forked processes?
I will use `cmd.Process.Kill()` whenever I return early, let me know if you think that is misuse

eyalkoren:
With the latest changes, the following e2e tests successfully pass on macOS (M1 chip) and Linux amd64 (AWS EC2):

System:
- OpenJDK Java 11 (JRE in Linux, JDK in macOS) with a non-root user as owner
- Tomcat 9 running as the same non-root user
- 8.4 snapshot deployment on ESS

Scenario:
1. Start Tomcat 9
2. Start Elastic agent with APM integration with Java agent attachment including a rule matching the `elastic.apm.attach` system property
3. Invoke some traffic on the webapps
4. Verify that the agent was not attached (as no discovery rule matched)
5. Restart Tomcat, this time with the system property `-Delastic.apm.attach=true`
6. Invoke some traffic
7. Verify that agent is attached and APM data is properly sent

beater/java_attacher/discovery_rules.go
```diff
@@ -0,0 +1,81 @@
...
+
+type discoveryRule interface {
+	include() bool
+	match(jvm *JvmDetails) bool
+	asString() string
+}
+
+type includeAllRule struct{}
+
+func (includeAllRule) match(jvm *JvmDetails) bool {
```
kruskall, line 33: **_suggestion_**: these should probably be pointer receivers to avoid copying the struct each time. See https://go.dev/tour/methods/4

(I'll only mention it here to avoid noise but it applies to all the methods in this file)

beater/java_attacher/java_attacher.go
```diff
@@ -25,50 +25,369 @@ import (
...
+			case "include-vmarg":
+				attacher.addCmdLineDiscoveryRule(value, true, "include-vmarg")
+			case "exclude-vmarg":
+				attacher.addCmdLineDiscoveryRule(value, false, "exclude-main")
+			default:
+				logger.Errorf("Unknown discovery rule - '%v'", name)
+			}
+		}
+	}
+	return attacher, nil
```
kruskall, line 104: **_question_**: does it make sense to return an error if there are no (known) discovery rules ?

beater/java_attacher/java_attacher.go
```diff
@@ -25,50 +25,369 @@ import (
...
+			time.Sleep(time.Second)
+		}
+	}
+}
+
+func (j *JavaAttacher) discoverJavaExecutable() error {
+	if j.javaBin == "" {
 		if jh := os.Getenv("JAVA_HOME"); jh != "" {
-			cfg.JavaBin = filepath.Join(jh, "/bin/java")
+			j.javaBin = filepath.Join(jh, "/bin/java")
```
kruskall, line 160: **_issue_**: avoid os specific path separator. 
```suggestion
			j.javaBin = filepath.Join(jh, "bin", "java")
```

beater/java_attacher/java_attacher.go
```diff
@@ -25,50 +25,369 @@ import (
...
 		} else {
 			bin, err := exec.LookPath("java")
 			if err != nil {
-				return JavaAttacher{}, fmt.Errorf("no java binary found: %v", err)
+				return fmt.Errorf("no java binary found: %v", err)
```
kruskall, line 164: **_issue_**: should use the specific modifier to support Go 1.13 error wrapping. See https://go.dev/blog/go1.13-errors#wrapping-errors-with-w

```suggestion
				return fmt.Errorf("no java binary found: %w", err)
```

beater/java_attacher/java_attacher.go
```diff
@@ -25,50 +25,369 @@ import (
...
+		}
+	}
+	if err := scanner.Err(); err != nil {
+		j.logger.Errorf("error reading output for command '%v': %v", strings.Join(cmd.Args, " "), err)
+	}
+	err = cmd.Wait()
+	if err != nil {
+		j.logger.Errorf("error executing command '%v': %v", strings.Join(cmd.Args, " "), err)
+	}
+	return jvms, nil
```
kruskall, line 269: **issue**: the last two errors are not being returned to the caller

beater/java_attacher/java_attacher.go
```diff
@@ -25,50 +25,369 @@ import (
...
+	for pid := range encounteredJvmCache {
+		_, found := jvms[pid]
+		if !found {
+			delete(encounteredJvmCache, pid)
+		}
+	}
+
+	// trying to improve start time accuracy - worth to consider optimizing as this runs every time the loop runs (1 second or so)
+	if err := j.executeForEachJvm(ctx, jvms, j.obtainAccurateStartTime, false, time.Second); err != nil {
+		j.logger.Infof("finding accurate start time for %v jvms did not finish successfully: %v", len(jvms), err)
```
kruskall, line 193: **issue**: because this is logging an error, it should use the error level (`Errorf`)

Note: it applies to other lines in this file.

beater/java_attacher/java_attacher.go
kruskall: I don't think the conversation has been resolved. This is still returning `error` but it seems to log an error and return `nil` in some cases.

I'll unresolve but feel free to solve it again if I am missing something :bow:

beater/java_attacher/java_attacher.go
```diff
@@ -77,6 +396,36 @@ func (j JavaAttacher) Run(ctx context.Context) error {
...
+func (j *JavaAttacher) attach(ctx context.Context, jvm *JvmDetails) error {
+	cmd := j.build(ctx, jvm)
+	err := j.setRunAsUser(jvm, cmd)
+	if err != nil {
+		j.logger.Errorf("failed to attach to JVM pid %v as user %v: %v. Trying to attach as current user,", jvm.pid, jvm.user, err)
+	}
+	j.logger.Infof("starting java attacher with command: %s", strings.Join(cmd.Args, " "))
+	stdout, err := cmd.StdoutPipe()
+	if err != nil {
+		return fmt.Errorf("cannot read from attacher standard output: %v", err)
```
kruskall, line 419: **issue**: same as above, use `%w`

beater/java_attacher/set_run_as_user.go
```diff
@@ -0,0 +1,52 @@
...
+	"os/exec"
+	"os/user"
+	"strconv"
+	"syscall"
+)
+
+func (j *JavaAttacher) setRunAsUser(jvm *JvmDetails, cmd *exec.Cmd) error {
+	currentUser, err := user.Current()
+	if err != nil {
+		j.logger.Infof("failed to get the current user: %v", err)
```
kruskall, line 34: **issue**: the error is logged but execution continues. The method should return or avoid using `currentUser` if an error was returned.

kruskall (commented):
Thanks for the contribution!

I left a few comments, mostly inconsistencies with logging an error vs returning an error

beater/java_attacher/discovery_rules.go
```diff
@@ -0,0 +1,81 @@
...
+
+import (
+	"fmt"
+	"regexp"
+)
+
+type discoveryRule interface {
+	include() bool
+	match(jvm *JvmDetails) bool
+	asString() string
```
axw, line 28:
```suggestion
```

There's a standard interface [`fmt.Stringer`](https://pkg.go.dev/fmt#Stringer). By renaming the `toString` methods on the structs to `String() string`, you'll implement that interface. String formatting functions know about that interface, so there's no need to call the method explicitly. Instead, you can just pass the value in and `%s` will call the `String()` method

beater/java_attacher/discovery_rules.go
```diff
@@ -0,0 +1,81 @@
...
+
+func (includeAllRule) match(jvm *JvmDetails) bool {
+	return true
+}
+
+func (includeAllRule) include() bool {
+	return true
+}
+
+func (includeAllRule) asString() string {
```
axw, line 41:
```suggestion
func (includeAllRule) String() string {
```

etc.

beater/java_attacher/java_attacher.go
```diff
@@ -25,50 +25,369 @@ import (
...
+	discoveryRules       []discoveryRule
+	rawDiscoveryRules    []map[string]string
+	agentConfigs         map[string]string
+	downloadAgentVersion string
+	javaBin              string
+}
+
+func (j *JavaAttacher) addDiscoveryRule(rule discoveryRule) {
+	j.discoveryRules = append(j.discoveryRules, rule)
+	j.logger.Debugf("added discovery rule: %v", rule.asString())
```
axw, line 66:
```suggestion
	j.logger.Debugf("added discovery rule: %s", rule)
```

beater/java_attacher/java_attacher.go
```diff
@@ -25,50 +25,369 @@ import (
...
+	defer cancelFunc()
+	var wg sync.WaitGroup
+	for pid, jvm := range jvms {
+		wg.Add(1)
+		go func(jvm *JvmDetails, pid string) {
+			err := executable(timeoutCtx, jvm)
+			if err != nil {
+				j.logger.Debug(err)
+				if removeOnError {
+					delete(jvms, pid)
```
axw, line 291: As this is in a goroutine, there may be concurrent map deletions. We need to synchronise these.

beater/java_attacher/java_attacher.go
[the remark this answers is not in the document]
axw: Doesn't appear to have been changed, unresolving.

beater/java_attacher/java_attacher.go
axw: Done here and elsewhere.

beater/java_attacher/set_run_as_user.go
axw: The error is now returned.

beater/java_attacher/java_attacher.go
axw: fixed

beater/java_attacher/java_attacher.go
axw: As above.

beater/java_attacher/set_run_as_user.go
simitt: @eyalkoren explained previously (don't remember if 1:1 or on the PR) that not finding the currentUser should only be logged but not be considered an error.

beater/java_attacher/java_attacher.go
eyalkoren: I suppose we can return `nil` in this case

beater/java_attacher/java_attacher.go
```diff
@@ -24,58 +24,419 @@ import (
...
+		return fmt.Errorf("java attacher is disabled")
+	}
+
+	if runtime.GOOS == "windows" {
+		// On Windows, we run the attacher by a JVM available to the
+		// current user in continuous mode.
+		if err := j.discoverJavaExecutable(); err != nil {
+			return err
+		}
+		return j.runAttacherContinuous(ctx)
```
eyalkoren, line 150: I think we don't have to do this now that we have support for JVM discovery on Windows. Whenever we can attach to a specific process with the actual Java that it uses, it should be a better option.

beater/java_attacher/java_attacher.go
```diff
@@ -24,58 +24,419 @@ import (
...
+			username = user.Username
+		}
+
+		jvms[pid] = &jvmDetails{
+			user:        username,
+			uid:         uid,
+			gid:         gid,
+			pid:         pid,
+			startTime:   info.StartTime,
+			command:     info.Exe,
```
eyalkoren, line 281: 🤯

eyalkoren, line 281: 🤯

beater/java_attacher/java_attacher.go
```diff
@@ -24,58 +24,419 @@ import (
...
+
+	return jvms, nil
 }
 
-// javaAttacher is bundled by the server
-var javaAttacher = filepath.FromSlash("./java-attacher.jar")
+// discoverAllRunningJavaProcesses returns a map of PIDs to running Java processes,
+// by looking for processes with "java" in the command.
+func (j *JavaAttacher) discoverAllRunningJavaProcesses(ctx context.Context) (map[int]*jvmDetails, error) {
+	processes, err := sysinfo.Processes()
```
eyalkoren, line 236: So much better!

beater/java_attacher/java_attacher.go
eyalkoren: Ohh man, that's right! Coming from Java, I looked whether it is even safe to remove from a map while iterating, but neglected this aspect when changed from sequential to concurrent execution. Thanks for catching this!

beater/java_attacher/java_attacher.go
eyalkoren: Clarifying - the attacher does not support running as a different user anyway, but even if it did, it's better to provide it with the user details and a PID rather than start it with the incorrect JVM in continuous mode and let it do all the heavy discovery

beater/java_attacher/java_attacher.go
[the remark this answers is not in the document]
eyalkoren: Irrelevant

beater/java_attacher/java_attacher.go
```diff
@@ -22,61 +22,389 @@ import (
...
 	if err := cmd.Start(); err != nil {
 		return err
 	}
+	scanner := bufio.NewScanner(stdout)
+	if scanner.Scan() {
+		jvmDetails.cmdLineArgs = scanner.Text()
+	}
+	if err := scanner.Err(); err != nil {
+		j.logger.Errorf("error obtaining command line args for process %v using 'ps -o command=': %v", err)
+	}
```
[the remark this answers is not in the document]
eyalkoren: Irrelevant

beater/java_attacher/java_attacher.go
eyalkoren: I did address this by not returning both error and a non-`nil` return value.
However, this and the function called by it (`discoverAllRunningJavaProcesses`) can have two types of errors - such that make them unable to provide a return value and such that don't. I want this function to return errors (with `nil` return value) of the first type and only log errors of the second type.

beater/java_attacher/java_attacher.go
```diff
@@ -24,58 +24,419 @@ import (
...
+				if removeOnError {
+					mu.Lock()
+					delete(jvms, pid)
+					mu.Unlock()
+				}
+			}
+			return err
+		})
+	}
+	return g.Wait()
```
eyalkoren, line 331: @axw One of the things that were lost with this better code is the active listening on the context, thus aborting quickly if it is closed. Here we wait up to the timeout, which may be 30 seconds for attachment. 
The thing is that `runAttacherCommand` can potentially block for long on `scanner.Scan()`, which may delay abort. Nevertheless, IIUC, the attach command itself would be killed once the context is closed, which means that `scanner.Scan()` should get `EOF` quite quickly. Is that correct?

axw: > Nevertheless, IIUC, the attach command itself would be killed once the context is closed, which means that scanner.Scan() should get EOF quite quickly. Is that correct?

That's right. The code now is dependant on the `f` in `f(ctx, jvm)` doing the right thing, and returning quickly once ctx is cancelled. That's either if the passed-in ctx was cancelled, or if the timeout triggers; whichever comes first. `CommandContext` will honour the ctx's cancellation.

beater/java_attacher/set_run_as_user.go
axw: It's returned here, but the caller logs it and continues to try running as the apm-server user.

eyalkoren: That's not optimal, it is only a failure to get the current user.
If we fail to get the current user, the default path should assume that the current user is `root` and the JVM user is not, thus trying to run as the target user. Trying to attach as `root` to a JVM ran by a non-root may fail.
Any reason why you think trying to run as target user is not preferable over running as APM Server user?

axw: Ah, now I understand. So if we can't get the current user, we should assume current user != target JVM user. I'll update it.

axw: Fixed and added a comment to clarify the intent.

axw (approved):

PR #8600, opened 2022-07-13
smoketest: Add APM Integration assertions

### LINKED ISSUE #8303: Run automated smoke tests on ESS

marclop:
## Motivation/summary

Adds the remaining assertions for the assets which are created by fleet
from the APM Integration (or APM package). It includes:

- Component templates and ILM Policies (these two are performed together
  since they are very closely tied).
- Ingest pipelines.

It also fixes the smoke tests which were previously broken since the
elastic artifacts API only returns the last few previous versions and
doesn't provide a comprehensive directory for all the released versions.
We now use a regex expression to match the previous minor latest patch
version (if 8.3.2 is the latest, then `8.2.latest` will be used).

## Checklist

## How to test these changes

1. Export credentials for ESS.
2. `make smoketest/all SMOKETEST_VERSIONS=7.17,latest`
3. Ensure the status code is 0.

## Related issues

Closes #8303

### DIFF
--- a/testing/smoke/lib.sh
+++ b/testing/smoke/lib.sh
@@ -246,17 +246,11 @@ data_stream_assert_pipelines() {
     # NOTE(marclop) we could assert that the pipelines have some sort of version suffix
     # in their name, however, the APM package version may not equal the deployment version.
     echo "-> Asserting ingest pipelines..."
-    local SUCCESS=true
     local RESPONSE=$(elasticsearch_curl '/_ingest/pipeline/*apm*')
     local HAS_APM_PIPELINES=$(echo ${RESPONSE} | jq -r '.|length>0')
     if [[ ${HAS_APM_PIPELINES} != true ]]; then
-        SUCCESS=false
         echo "-> Did not find any APM ingest pipelines"
         echo ${RESPONSE}
-    fi
-
-    if [[ ${SUCCESS} == false ]]; then
-        echo "-> Failed asserting ingest pipelines"
         return 31
     fi
 }
--- a/testing/smoke/lib.sh
+++ b/testing/smoke/lib.sh
@@ -4,7 +4,7 @@ terraform_apply() {
     echo "-> Creating / Upgrading deployment to version ${1}"
     echo stack_version=\"${1}\" > terraform.tfvars
     if [[ ! -z ${2} ]] && [[ ${2} ]]; then echo integrations_server=true >> terraform.tfvars; fi
-    terraform init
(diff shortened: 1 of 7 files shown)


testing/smoke/lib.sh
```diff
@@ -227,14 +225,74 @@ legacy_ingest_pipelines() {
...
+    local SUCCESS=true
+    local RESPONSE=$(elasticsearch_curl '/_ingest/pipeline/*apm*')
+    local HAS_APM_PIPELINES=$(echo ${RESPONSE} | jq -r '.|length>0')
+    if [[ ${HAS_APM_PIPELINES} != true ]]; then
+        SUCCESS=false
+        echo "-> Did not find any APM ingest pipelines"
+        echo ${RESPONSE}
+    fi
+
+    if [[ ${SUCCESS} == false ]]; then
```
lahsivjar, line 258: IIUC, we can merge this condition block with `if [[ ${HAS_APM_PIPELINES} != true ]]; then` since `SUCCESS = false` will only happen if the first `if` evaluates to true

lahsivjar (approved):
LGTM

Issue #5243, opened 2021-05-09
service metrics: add numeric outcome 


dgieselaar:
Add a numeric outcome to 
- transaction events (mapped as `byte`)
- transaction metrics (mapped as `byte` as it has an `event.outcome` dimension)
- service metrics (mapped as [`aggregate_metric_double`](https://www.elastic.co/guide/en/elasticsearch/reference/current/aggregate-metric-double.html), as it has no `event.outcome` dimension). 

All fields should be mapped as doc-value only fields.

The concrete field name needs to be defined. The working title is `event.outcome_numeric`.

This will help with reducing dimensions while keeping the structure of transaction metrics and service metrics similar. 

Mapping of `event.outcome` to `event.outcome_numeric`

| `event.outcome` | `event.outcome_numeric`|
|---|---|
|`unknown` | `null` |
| `failure` | 0 |
| `success` | 1 |

For the `aggregate_metric_double`, the following sub-fields should be populated:
- `sum`: the number of success events
- `value_count`: the number of success + error events

We can then simply do an average aggregation over the `event.outcome_numeric` to get the success rate. We need to validate that the avg aggregation works correctly on both the `byte` and `aggregate_metric_double` mapped fields.

---

Original request: 
Right now we use a terms aggregation or multiple filter aggregations on event.outcome to calculate transaction error rate. If we store failure as 0 and success in 1 in a single field (and not populate it for unknown), we can get the percentage by doing an avg aggregation on that single field. This would simplify our queries and the post processing that we are doing, will probably be a small performance improvement, and would make it easier for users to build custom dashboards.

axw:
[`event.outcome`](https://www.elastic.co/guide/en/ecs/current/ecs-event.html#field-event-outcome) is an ECS field with a prescribed set of allowed values, so I don't think we should be recording its value as a number. We could provide a separate field.

dgieselaar:
@axw yes, sorry, that is the idea. I don't really have a good suggestion for the name of the field though 😀

simitt:
@dgieselaar I updated the original description to what we have planned now for `8.7`.

axw:
@felixbarny @dgieselaar would either of you be able to help clarify the definition a bit please?

It might help if we knew how the fields would be aggregated, for each of transaction events, transaction metrics, and service metrics.

felixbarny:
done, let me know if something is missing or unclear

carsonip:
[...]

PR #8607, opened 2022-07-13
Service Metrics Aggregator

Mpdreamz:
Introduces a new derivative datastream that holds latency and throughput metrics pivoted by
 `service.name+service.environment+transaction.type`

 This yields a significant reduction in the documents we need to query on the Service Inventory Page and gives us a more natural place to query for Services (e.g Service Groups).

## Motivation/summary

This work stems from my most recent APM On Week: https://github.com/elastic/apm-dev/issues/768

Which itself is a continuation of the prior On Week effort to make synthtrace behave more closely to how apm-server behaves so we can prototype new aggregations more easily and realistic. https://github.com/elastic/kibana/pull/127257

In the context of APM UI performance investigations we used this to prototype a new service metrics datastream: https://github.com/elastic/kibana/pull/132889

Which we could use to query for service metrics at 100x performance increase by working over significantly less documents/values. In our 2 main test datasets the reduction was as followed.

```
500 million traces
50 million transaction metrics 
500k services timeseries 
```

A few other benefits:

* this new datastream maps really well to TSDB and will benefit further from downsampling
* We can be very explicit about what metadata we want to include and drive other features such as service groups and entity models. 
* The higher the load on APM the better the compression to service metrics works in our favor.

## Checklist

## How to test these changes

Terminal 1: 

```
apm-server $ docker compose up --force-recreate --build
```

Terminal 2: 

```
apm-server $ cd systemtest/cmd/runapm
run-apm $ go run main.go -arch amd64 -f -reinstall
```

I have been using this modified version of synthtrace which currently still only lives in https://github.com/elastic/kibana/pull/136530

```
node scripts/synthtrace packages/elastic-apm-synthtrace/src/scenarios/high_throughput.ts --local --maxDocs 1000000 --apm http://localhost:8200/ --username admin --skipPackageInstall --clean
```

This allows you to send `--maxDocs` traces directly to APM using various datageneration scenario's.

## Related issues

https://github.com/elastic/apm-server/issues/8756


apm-server.yml
```diff
@@ -326,8 +326,8 @@ output.elasticsearch:
 
   # Authentication credentials - either API key or username/password.
   #api_key: "id:api_key"
-  #username: "elastic"
-  #password: "changeme"
+  username: "admin"
```
Mpdreamz, line 329: This needs reverting

apmpackage/apm/data_stream/service_metrics/fields/ecs.yml
```diff
@@ -0,0 +1,137 @@
+- external: ecs
```
Mpdreamz, line 1: this field set is too liberal. Need to prune this down to only the fields we actually intend to use.

beater/interceptors/logging.go
```diff
@@ -62,7 +62,7 @@ func Logging(logger *logp.Logger) grpc.UnaryServerInterceptor {
 			logger.With("error.message", res.Message()).Error(logp.Error(err))
 			return nil, err
 		}
-		logger.Info("request accepted")
+		logger.Debugf("request accepted")
```
Mpdreamz, line 65: Controversial and might warrant its own PR but I found this to be very noisy just running with `-e`

x-pack/apm-server/main.go
```diff
@@ -84,6 +85,19 @@ func newProcessors(args beater.ServerParams) ([]namedProcessor, error) {
 		return nil, errors.Wrapf(err, "error creating %s", spanName)
 	}
 	processors = append(processors, namedProcessor{name: spanName, processor: spanAggregator})
+
+	const serviceName = "service metrics aggregation"
+	args.Logger.Infof("creating %s with config: %+v", spanName, args.Config.Aggregation.ServiceDestinations)
```
Mpdreamz, line 90: this still needs its own configuration section

simitt:
@Mpdreamz can you please link to an issue with some more context and justification for this change.

Mpdreamz:
@simitt Updated the motivation with more context and included links to all prior work that lead to this PR.

apmpackage/apm/data_stream/service_metrics/manifest.yml
```diff
@@ -0,0 +1,31 @@
...
+              type: object
+              properties:
+                aggregate:
+                  type: aggregate_metric_double
+                  metrics:
+                    - min
+                    - max
+                    - sum
+                    - value_count
+                  default_metric: max
```
axw, line 27: Are min and max needed? I don't see any min/max aggs in https://github.com/elastic/kibana/blob/main/x-pack/plugins/apm/dev_docs/apm_queries.md, and the service overview only uses avg AFAIK.

apmpackage/apm/data_stream/service_metrics/manifest.yml
```diff
@@ -0,0 +1,31 @@
...
+      # mapping this here because aggregate_metric_double is unsupported in field sets
+      # https://github.com/elastic/package-spec/issues/363
+      properties:
+        transaction:
+          type: object
+          properties:
+            duration:
+              type: object
+              properties:
+                aggregate:
```
axw, line 20: We have a concept of "summary" metrics. I would suggest calling this field `transaction.duration.summary`.

apmpackage/apm/data_stream/service_metrics/manifest.yml
```diff
@@ -0,0 +1,31 @@
+title: APM service metrics
+type: metrics
+dataset: apm.service
+ilm_policy: metrics-apm.service_metrics-default_policy
+elasticsearch:
+  index_template:
+    mappings:
+      # Internal metrics should have all fields strictly mapped;
+      # we are in full control of the field names.
+      dynamic: false
```
axw, line 10: Some copy pasta here, but also you've changed it from strict to dynamic. Any particular reason why we should allow unmapped fields?

beater/interceptors/logging.go
```diff
@@ -62,7 +62,7 @@ func Logging(logger *logp.Logger) grpc.UnaryServerInterceptor {
 			logger.With("error.message", res.Message()).Error(logp.Error(err))
 			return nil, err
 		}
-		logger.Info("request accepted")
+		logger.Infof("request accepted")
```
axw, line 65:
```suggestion
		logger.Info("request accepted")
```

model/metricset.go
```diff
@@ -132,6 +132,33 @@ func (s *SummaryMetric) fields() mapstr.M {
 	return mapstr.M{"value_count": s.Count, "sum": s.Sum}
 }
 
+// AggregateMetric holds extended summary metrics (value_count, sum, min, max).
+type AggregateMetric struct {
```
axw, line 136: There's an existing type [`model.SummaryMetric`](https://github.com/elastic/apm-server/blob/0420137bea7f9e8ad10ce6b5bbb751b0e0134baa/model/metricset.go#L122-L129), which is a subset of this. If Min and Max aren't needed, we should use that. If they are needed, I'd like to find a way to unify them.

model/transaction.go
```diff
@@ -56,6 +56,12 @@ type Transaction struct {
 	// duration metrics.
 	DurationHistogram Histogram
 
+	// DurationAggregate stores the value_count, min, max and sum of observed duration for the service metrics.
+	DurationAggregate AggregateMetric
+
+	// FailureCount is used by service metrics documents to indicate how many failed transactions were observed
+	FailureCount *int
```
axw, line 63: We have `event.outcome`, which we group by in transaction metrics to calculate the failed transaction rate. Can we do the same here?

axw (commented):
I've only take a quick look through this. Code looks good, and straightforward.

I'm not too keen on inventing a new way to count failed transactions for a service, as it feels a bit like overfitting to the current service overview page. Did you already try grouping by `event.outcome`, and comparing the resulting storage and query efficiency?

Conversely, although having min/max is nice and flexible, I _would_ fit to the current use of aggregations. The `aggregate_metric_double` type is rather inflexible, and you have to know up front which subfields you may populate. In this case we know, but in the case of dynamic summary metrics we do not. I'd prefer to be consistent, and just say we won't ever store min/max for summary metrics (unless the constraints of `aggregate_metric_double` change). If we do need them, perhaps we should be storing a latency histogram instead.

Mpdreamz:
> I'm not too keen on inventing a new way to count failed transactions for a service, as it feels a bit like overfitting to the current service overview page. 

I think 'overfitting' is not a negative as its actually intentional. 
`Service metrics` primary goal is to produce dedicated precalculated metric timeseries for the overview page. 

Secondary goal would be: 
- Service Groups: attach enough metadata (not dimensions) to create a curated query/grouping capabilities.  

> Did you already try grouping by event.outcome, and comparing the resulting storage and query efficiency?

I would caution more on the side of introducing more dimensions vs more metrics. We have no need to create additional timeseries per `event.outcome`. 

What's the main worry here? It's not necessary a different way but a new mechanism to deliver the same answer with one being pre-calculated. 

> Conversely, although having min/max is nice and flexible, I would fit to the current use of aggregations. The aggregate_metric_double type is rather inflexible, and you have to know up front which subfields you may populate. In this case we know, but in the case of dynamic summary metrics we do not.

I'm not invested in keeping `p0` and `p100` around. They are however reasonably cheap to keep create and keep around. I would love to hear from @elastic/machine-learning team if it has any benefits for any of their jobs. 

> If we do need them, perhaps we should be storing a latency histogram instead.

I would not be opposed to doing **both** but would not choose either histogram or metric_aggregate_double. Key for us reducing number of values we need to aggregate over and precalculating as much as possible.  


Today our service inventory page does not display p95/p99 I would love to add latency histograms to the emitted metrics to support that in a follow up PR though. The downside is that there is no [play for histograms in TSDB](https://github.com/elastic/elasticsearch-adrs/blob/master/analytics/tsdb/tsdb-rollups-design.md) and introducing these would potentially delay Service Metrics utilzing TSDB as we work with the team to create a play for histograms. 

Or we need to emit two metric documents per `service.name+service.environment+transaction.type`.

axw:
> I think 'overfitting' is not a negative as its actually intentional.
> Service metrics primary goal is to produce dedicated precalculated metric timeseries for the overview page.

What I meant is that it's overfitting to the _current_ service overview page. If it changes (say we want to add percentiles), then we need to adapt or introduce new metrics. Every addition bears a tech-debt cost, and data lives for a long time. So if we can make it a little more flexible then that would be nice.

Of course, there are always tradeoffs. If we do group by `event.outcome` then there will naturally be more docs, so it depends on the cost.

> What's the main worry here? It's not necessary a different way but a new mechanism to deliver the same answer with one being pre-calculated.

At the moment, the way the UI queries for [failed transaction rate](https://github.com/elastic/kibana/blob/main/x-pack/plugins/apm/dev_docs/apm_queries.md#failed-transaction-rate) is by filtering on `event.outcome: failed`. By introducing a new field, the UI (and any other consumer) now needs to deal with two ways of calculating failed transaction rate. I'm questioning whether the increase in complexity is worthwhile.

> Today our service inventory page does not display p95/p99 I would love to add latency histograms to the emitted metrics to support that in a follow up PR though. The downside is that there is no [play for histograms in TSDB](https://github.com/elastic/elasticsearch-adrs/blob/master/analytics/tsdb/tsdb-rollups-design.md) and introducing these would potentially delay Service Metrics utilzing TSDB as we work with the team to create a play for histograms.

Do you have any numbers on the size of service metrics is vs. transaction metrics? Is it worthwhile rolling up only service metrics in the short-to-mid term?

I realise we need to start somewhere, but it seems of limited value to users to roll up service metrics if we can't roll up the more expensive transaction metrics. So, I think on the whole I'd prefer to go with histogram if it's the eventually ideal data type, even if it means we don't get rollups in the short term.

Mpdreamz:
> At the moment, the way the UI queries for [failed transaction rate](https://github.com/elastic/kibana/blob/main/x-pack/plugins/apm/dev_docs/apm_queries.md?rgh-link-date=2022-07-25T02%3A48%3A46Z#failed-transaction-rate) is by filtering on event.outcome: failed. By introducing a new field, the UI (and any other consumer) now needs to deal with two ways of calculating failed transaction rate. I'm questioning whether the increase in complexity is worthwhile.

I think as we progress with more data transform capabilities in the stack the need to query either pre-aggregated or raw will only increase. The UI already has to discover how to query today even with a single common field. 

Dimensions are progressively more costly so would love to keep this to an absolute minimum unless our desired functionality  dictate otherwise.  

> Do you have any numbers on the size of service metrics is vs. transaction metrics? Is it worthwhile rolling up only service metrics in the short-to-mid term?

It's not in terms of number of documents (expectation) but it is also  worthwhile as a means to gain confidence in onboarding our metrics indices to TSDB. 

> I realise we need to start somewhere, but it seems of limited value to users to roll up service metrics if we can't roll up the more expensive transaction metrics. So, I think on the whole I'd prefer to go with histogram if it's the eventually ideal data type, even if it means we don't get rollups in the short term.

Yeah TSDB/Rollups are not a hard requirement at all but I would still keep the initial `Service Metrics` implementation as minimal and simple as possible to proof its performance claims in the real world. 

100% agree we need to include histograms in another iteration of this though.

> What I meant is that it's overfitting to the current service overview page. If it changes (say we want to add percentiles), then we need to adapt or introduce new metrics. 

I think this is less of an issue, we can include histograms later without being breaking.

axw:
Just coming back to the `failure_count` vs. `event.outcome` debate...

I'd like us to try and find a way to address https://github.com/elastic/apm-server/issues/5243 at the same time. While looking at that, it occurred to me that just storing `failure_count` is not good enough. The failed transaction rate is calculated from `failure(failure+success)` -- this excludes transactions with `event.outcome: unknown`. We would also need to store the number of successful transactions.

x-pack/apm-server/aggregation/servicemetrics/aggregator.go
```diff
@@ -0,0 +1,342 @@
...
+}
+
+// ProcessBatch aggregates all service latency metrics
+func (a *Aggregator) ProcessBatch(ctx context.Context, b *model.Batch) error {
+	a.mu.RLock()
+	defer a.mu.RUnlock()
+	for _, event := range *b {
+		tx := event.Transaction
+		if event.Processor == model.TransactionProcessor && tx != nil {
+			if msEvent := a.processTransaction(&event); msEvent.Metricset != nil {
```
Mpdreamz, line 182: I wonder if this can be extended to also observe `errors` and `metrics` and emits this as `errorCount` and `metricsCount` metrics that way querying for services can be even further simplified:

https://github.com/elastic/kibana/blob/2f7603de181a47cde19eccb6fd4b665954bd105b/x-pack/plugins/apm/server/routes/services/get_services/get_services_items.ts

Today merges two results client side to create a holistic picture.

x-pack/apm-server/aggregation/servicemetrics/aggregator.go
```diff
@@ -0,0 +1,342 @@
...
+		},
+		Agent: model.Agent{
+			Name: metrics.agentName,
+		},
+		Processor: model.MetricsetProcessor,
+		Metricset: &model.Metricset{
+			Name: metricsetName,
+		},
+		Transaction: &model.Transaction{
+			FailureCount: &metrics.sumFailures,
```
Mpdreamz, line 333: Missing transaction type in emitted doc (already part of aggregation key)

axw: Why do we need to aggregate on transaction type? If we're aggregating on transaction type, then they're starting to look a lot like transaction metrics, not service metrics.

axw:
Moving this into draft while I merge main and make some updates.

x-pack/apm-server/aggregation/servicemetrics/aggregator_test.go
```diff
@@ -0,0 +1,344 @@
...
+			err := agg.ProcessBatch(context.Background(), &batch)
+			require.NoError(t, err)
+			assert.Equal(t, model.Batch{transaction}, batch)
+		}(in)
+	}
+	wg.Wait()
+
+	// Start the aggregator after processing to ensure metrics are aggregated deterministically.
+	go agg.Run()
+	defer agg.Stop(context.Background())
```
marclop, line 108:
```suggestion
	t.Cleanup(func() { agg.Stop(context.Background()) })
```

marclop (approved):
LGTM. Should we add a system test as well?

sqren:
@Mpdreamz Are we setting a `metricset.name` for the service metrics, eg `metricset.name: "services"`? Having a unique identifier for the service metrics is important for the UI when querying for them and determining whether they are available or not.

Mpdreamz:
@sqren yep! https://github.com/elastic/apm-server/pull/8607/files#diff-4e538a1d3c8af2d271c66e8044baca93a083e0d85779d51cc31d4426a9ff0a95R300

It's currently `service` should it be plural?

sqren:
> @sqren yep! https://github.com/elastic/apm-server/pull/8607/files#diff-4e538a1d3c8af2d271c66e8044baca93a083e0d85779d51cc31d4426a9ff0a95R300
> 
> It's currently `service` should it be plural?

Perfect! Singular is fine.

x-pack/apm-server/aggregation/servicemetrics/aggregator_test.go
axw: What's the benefit? I tend to use Cleanup when creating resources outside of the main test function, which should be disposed of when the test ends.

marclop: I suppose there's no huge benefit in this case, but since we're writing "new" tests for this aggregator, I thought we may as well use `t.Cleanup`, but it doesn't really matter.

axw: OK. I'm going to leave it as it is, as there's no immediate benefit. Maybe we can take a look later and see if there would be some benefit in having a test helper function for creating the aggregator, and stopping it on cleanup.

PR #8440, opened 2022-06-22
Add command to generate continuous events for soak testing

lahsivjar:
## Checklist

~~- [ ] Update [CHANGELOG.asciidoc](https://github.com/elastic/apm-server/blob/main/CHANGELOG.asciidoc)~~
~~- [ ] Update [package changelog.yml](https://github.com/elastic/apm-server/blob/main/apmpackage/apm/changelog.yml) (only if changes to `apmpackage` have been made)~~
- [x] Documentation has been updated

## How to test these changes

Send continuous events to a running APM-Server using:

```
$ cd systemtest
$ go run cmd/apmsoak/main.go -max-rate 60eps
```

## Related issues

Closes #8406

PR #8609, opened 2022-07-14
apmbench: Fix broken `-benchtime`

marclop:
## Motivation/summary

Fixes the `-benchtime` flag which has no effect currently (benchmarks
are currently only running for 1s, regardless of what is set in the
benchtime flag).

## Checklist

## How to test these changes

Run apmbench with a -benchtime of 1m and verify it runs for at least 1m. (set -warmup-events to 0 so there's no warmup).

## Related issues

Fixes the regression introduced by #8440.


lahsivjar (approved):
Ah! I missed that. Thanks for catching it.

Issue #7868, opened 2022-04-13
benchmarking: Create meaningful dashboards for apmbench results


marclop:
## Description

The new benchmarking framework will also need new dashboards to be built. Since we're rebuilding them from scratch, we can look at what could be improved when comparing them to the existing observability-benchmarks dashboard.

### Dashboard consumers

While the main goal is to allow us (Go devs) to track the APM Server's performance over time and gauge the impact that code changes have in that ingest performance, but we should also think about how to make this data useful to allow sizing recommendations to happen as a result of our long running benchmarks.

I can think a couple of roles that may look into the dashboards (or docs that are updated from the dashboards):

- Support engineers looking into to a case where the APM Server (perhaps Elasticsearch as well) may not be appropriately sized for the number of events/s received.
- Solution Architects helping a customer plan their Elastic Stack deployment.

### Dashboard metrics

Each profile of user will be interested in different metrics, however, all parties are likely interested in an overview summary that shows higher-level view of the data, broken down by APM Server / Elasticsearch size:

- Aggregate throughput (events/s).
- Performance regressions.
- Resource usage.

#### Server team

As the main apm-server contributors, we'll be interested in tracking how the APM Server's resource usage evolves over time, these  are tracked in https://github.com/elastic/apm-server/issues/7817.

marclop:
This work has been mostly block by the nightly benchmarks failing due to #8426. There is a half-built dashboard available in the observability-benchmarks deployment: https://observability-benchmarks.elastic.dev/goto/de5d4aba9eb958c62a68731bec3a86bb.

marclop:
Benchmarks are running regularly again and I'm working on building the new dashboard

marclop:
Turns out that the `-benchtime` flag was broken and had no effect (Fixed in #8609). The benchmarks had been running with 1s duration (default) and it's quite possible that's why we saw so much variance in the daily benchmarks.

marclop:
The dashboard has been created and it's awaiting some feedback from the wider team.

marclop:
I've incorporated the feedback on the dashboard so far and this issue can be closed since all the scoped work has been done.

PR #8616, opened 2022-07-14
benchmark: Add `BENCHMARK_MAX_RATE` tune warmup

marclop:
## Motivation/summary

Adds a new `BENCHMARK_MAX_RATE` variable which defaults to `0eps` or
disabled max rate, tunes the `BENCHMARK_WARMUP` to `40,000` events to be
sent as the warmup to the APM Server (`10,000` was enough, but `40,000`
ensures that the CPU credts have been exhausted for a 1g deployment).

Also, modifies the warmup functon to not return an error on warmup
should the error be `context.DeadlineExceeded`, and adds more weight to
the number of agents specified.

## Checklist

## How to test these changes

0. (export ess credentials)
1. `cd testing/benchmark`
2. `make all run-benchmark BENCHMARK_WARMUP=50000 BENCHMARK_MAX_RATE=5000eps BENCHMARK_TIME=30s BENCHMARK_COUNT=1`

## Related issues

Improvements while working on #7868

### DIFF
--- a/systemtest/benchtest/main.go
+++ b/systemtest/benchtest/main.go
@@ -264,7 +264,7 @@ func warmup(agents int, events uint, url, token string) error {
 	close(errs)
 	var merr multierror.Errors
 	for err := range errs {
-		if err != nil {
+		if err != nil && !errors.Is(err, context.DeadlineExceeded) {
 			merr = append(merr, err)
 		}
 	}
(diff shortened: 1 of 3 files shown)


systemtest/benchtest/main.go
```diff
@@ -289,6 +289,6 @@ func warmupTimeout(ingestRate float64, events uint, epm, agents, cpus int) time.
 	// the apmbench runner can do.
 	factor := math.Max(1, float64(agents)/float64(cpus))
 	timeoutSeconds := math.Max(float64(events)/ingestRate, 1) *
-		float64(agents) * factor
+		float64(agents*2) * factor
```
lahsivjar, line 292: LGTM, just a thought for later, maybe it makes more sense to have a time dependent warmup rather than events dependent warmup, we can do a time dependent warmup with the event rate specified.

lahsivjar (approved):

systemtest/benchtest/main.go
marclop: I totally agree, we discussed this with Silvia yesterday and agreed that a time based warmup would make much more sense. I'll open an improvement issue.

marclop: Opened #8621

Issue #8651, opened 2022-07-20
Global label sanitisation may lead to concurrent map modification/access


axw:
**APM Server version** (`apm-server version`): 7.17.1

**Description of the problem including expected versus actual behavior**:

Server panics, e.g.

```
Jun 21 03:44:25 rn000043550 apm-server[71111]: panic: runtime error: invalid memory address or nil pointer dereference
Jun 21 03:44:25 rn000043550 apm-server[71111]: [signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x561e096719d1]
Jun 21 03:44:25 rn000043550 apm-server[71111]: goroutine 197 [running]:
Jun 21 03:44:25 rn000043550 apm-server[71111]: github.com/elastic/apm-server/model.sanitizeLabels(0xc00e831d50)
Jun 21 03:44:25 rn000043550 apm-server[71111]: #011/go/src/github.com/elastic/apm-server/model/labels.go:35 +0x118
Jun 21 03:44:25 rn000043550 apm-server[71111]: github.com/elastic/apm-server/model.(*APMEvent).BeatEvent(0xc00e8318a8, {0xc00afa6ec0, 0xc00e831828})
--
Jun 21 05:55:29 rn000043550 apm-server[82942]: panic: runtime error: invalid memory address or nil pointer dereference
Jun 21 05:55:29 rn000043550 apm-server[82942]: [signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x561df7a1f9d1]
Jun 21 05:55:29 rn000043550 apm-server[82942]: goroutine 224 [running]:
Jun 21 05:55:29 rn000043550 apm-server[82942]: github.com/elastic/apm-server/model.sanitizeLabels(0xc011b9dd50)
Jun 21 05:55:29 rn000043550 apm-server[82942]: #011/go/src/github.com/elastic/apm-server/model/labels.go:35 +0x118
Jun 21 05:55:29 rn000043550 apm-server[82942]: github.com/elastic/apm-server/model.(*APMEvent).BeatEvent(0xc011b9d8a8, {0xc01167ec40, 0xc011b9d828})
--
Jun 21 05:57:40 rn000043550 apm-server[89635]: panic: runtime error: invalid memory address or nil pointer dereference
Jun 21 05:57:40 rn000043550 apm-server[89635]: [signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x558e90b2a5c1]
[...]

PR #8652, opened 2022-07-20
model: clone labels before sanitizing

axw:
## Motivation/summary

If an agent sends global labels, then events without event-specific labels will share the same map of global labels. If those global labels need sanitisation, then the server may panic due to concurrently modifying/reading the map. To prevent this, clone the labels map if sanitisation is required.

## Checklist

~- [ ] Update [CHANGELOG.asciidoc](https://github.com/elastic/apm-server/blob/main/CHANGELOG.asciidoc)~ to be added at release time, to avoid updating live docs

## How to test these changes

Send a continuous stream of events with global labels that require sanitisation (e.g. a label key with a dot in it), and no event-specific labels. The server should not panic.

## Related issues

Closes https://github.com/elastic/apm-server/issues/8651

### DIFF
--- a/model/apmevent_test.go
+++ b/model/apmevent_test.go
@@ -20,6 +20,7 @@ package model
 import (
 	"context"
 	"net"
+	"sync"
 	"testing"
 	"time"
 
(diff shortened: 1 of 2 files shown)


lahsivjar (approved):
LGTM

Issue #6858, opened 2021-12-08
Investigate enabling golangci-lint in CI/make check


axw:
I ran golangci-lint locally, and there were some good findings which resulted in #6857. There was also some noise, e.g. related to not checking error return on `bytes.Buffer.Write` calls which will only error under pathological conditions.

We should investigate using golangci-lint, perhaps with some code excluded: tests, and certain code such as described above. We should consider retiring reviewdog at the same time.

axw:
We can alternatively expand our use of staticcheck, and still get rid of golint. _Maybe_ we also want golangci-lint, or perhaps we should just cherry-pick the specific linters/checks we want to use.

axw:
We ended up expanding our use of staticcheck: https://github.com/elastic/apm-server/pull/8680

PR #8680, opened 2022-07-21
Remove golint, just use staticcheck

axw:
## Motivation/summary

golint is deprecated and frozen (https://github.com/golang/go/issues/38968). Remove its use, along with reviewdog, and turn on all (but one) staticcheck checks to compensate. I've fixed a few small issues that staticcheck found as a result of this.

## Checklist

## How to test these changes

N/A

(If you really want you can introduce a lint issue and check that it gets picked up, e.g. change a type's comment such that it doesn't begin with the type name.)

## Related issues

https://github.com/elastic/apm-server/issues/6858


lahsivjar (approved):
LGTM. For ST1000, should we create an issue or track it somewhere?

axw:
> For ST1000, should we create an issue or track it somewhere?

Yep, good call, I'll open an issue.

PR #8681, opened 2022-07-21
Reduce magefile

axw:
## Motivation/summary

Simplify our build process, and tidy up the repository a little bit more.

- Remove `mage update` and `mage config`; remove _meta/beat.yml.
  keeping committed apm-server.yml and apm-server.docker.yml.
  Use sed to create apm-server.docker.yml from apm-server.yml.
- Remove `mage version`, use `make get-version` where needed.
- Remove `mage pythonEnv`, create venv directly.
- Remove `mage clean`, just remove the files using `rm -f`
- Remove `mage check`, replace with a script that checks if
  the git working tree is clean using `git status --porcelain`.
  The other mage checks are irrelevant: `go vet` is already
  run when running tests; we don't have Python tests; not sure
  why we would care if YAML files are executable, not worthwhile.

## Checklist

## How to test these changes

N/A

## Related issues

None


lahsivjar (approved):
LGTM, one question, why do we need to keep both `apm-server.docker.yml` and a way to generate it from `apm-server.yml`?

axw:
> LGTM, one question, why do we need to keep both apm-server.docker.yml and a way to generate it from apm-server.yml?

@lahsivjar I originally intended to remove it, but `apm-server.docker.yml` is referenced in the docs: https://www.elastic.co/guide/en/apm/guide/current/running-on-docker.html#_example_configuration_file

So to avoid breaking users, I've left it alone for now.

PR #8492, opened 2022-06-27
Add initial support of smoketests in CI

amannocci:
Signed-off-by: Adrien Mannocci <adrien.mannocci@elastic.co>

## Motivation/summary

* Implementation of smoke tests within the CI.

## Checklist

- [ ] Update [CHANGELOG.asciidoc](https://github.com/elastic/apm-server/blob/main/CHANGELOG.asciidoc)
- [ ] Update [package changelog.yml](https://github.com/elastic/apm-server/blob/main/apmpackage/apm/changelog.yml) (only if changes to `apmpackage` have been made)
- [ ] Documentation has been updated

For functional changes, consider:
- Is it observable through the addition of either **logging** or **metrics**?
- Is its use being published in **telemetry** to enable product improvement?
- Have system tests been added to avoid regression?

## Related issues

Ref #8303

PR #8682, opened 2022-07-21
Add failure report for smoketests

amannocci:
Signed-off-by: Adrien Mannocci <adrien.mannocci@elastic.co>

## Checklist

## How to test these changes

* Run the following [Jenkins job](https://apm-ci.elastic.co/job/apm-server/job/smoke-tests-mbp/view/change-requests/job/PR-8682/) with `6.8` as version.
* Validate the failure report in the slack channel.
* Run the following [Jenkins job](https://apm-ci.elastic.co/job/apm-server/job/smoke-tests-mbp/view/change-requests/job/PR-8682/) with `latest` as version.
* Validate that no reported failure is present in the slack channel.

## Related issues

Ref: https://github.com/elastic/apm-server/issues/8303

### DIFF
--- a/.ci/smoke-tests.groovy
+++ b/.ci/smoke-tests.groovy
@@ -76,8 +76,8 @@ pipeline {
     }
   }
   post {
-    failure {
-      notifyBuildResult(slackComment: true, slackNotify: true)
+    cleanup {
+      notifyBuildResult(slackComment: true)
     }
   }
 }
--- a/.ci/smoke-tests.groovy
+++ b/.ci/smoke-tests.groovy
@@ -10,6 +10,7 @@ pipeline {
     EC_KEY_SECRET = 'secret/observability-team/ci/elastic-cloud/observability-pro'
     TERRAFORM_VERSION = '1.2.3'
     CREATED_DATE = "${new Date().getTime()}"
+    SLACK_CHANNEL = "#apm-server"
   }
 
   options {
@@ -72,6 +73,11 @@ pipeline {
           }
         }
       }
+    }
+  }
+  post {
+    failure {
+      notifyBuildResult(slackComment: true, slackNotify: true)
     }
   }
 }


amannocci:
Screenshot 2022-07-21 at 13 20 50

.ci/smoke-tests.groovy
```diff
@@ -78,6 +79,11 @@ pipeline {
       }
     }
   }
+  post {
+    failure {
+      notifyBuildResult(slackComment: true, slackNotify: true)
+    }
```
v1v, line 85: I'd say to run this always similar pattern done in https://github.com/elastic/apm-pipeline-library/blob/467418b3dc74c766080235ca02f57a8bac7b22ab/resources/JenkinsfileTemplate.groovy#L392-L396

> * slackNotify: Whether to send or not the slack notifications, by default it sends notifications on Failed builds that are not pull request.


https://github.com/elastic/apm-pipeline-library/blob/467418b3dc74c766080235ca02f57a8bac7b22ab/vars/notifyBuildResult.groovy#L67



```suggestion
    cleanup {
      notifyBuildResult(slackComment: true)
    }
```

If needed you can add some headers to the slack comment , see https://github.com/elastic/apm-pipeline-library/blob/main/vars/notifyBuildResult.txt#L33

amannocci: Do we want to avoid running report failure on PR?

v1v: IIUC, the smoke-tests pipeline will only run in branches as per https://github.com/elastic/apm-server/pull/8492#discussion_r907419202 

Did it change that?

amannocci: We should be able to run smoketests on PRs, main branch, 7.17 branch, and 8.x branches.
Without the argument `slackNotify`, we will drop the report failure notification on PRs.

v1v: Can you confirm if that's the case for PRs? I cannot see any references in  https://github.com/elastic/apm-server/issues/8303

IMO, I'd say no notifications on PRs, the GitHub checks and GitHub build comments should be the ones driving those notifications on a PR basis, without any lack notification.

amannocci: We can run smoketests on PRs only on-demand AFAIK (https://github.com/elastic/apm-server/blob/main/.ci/jobs/apm-server-smoketests-mbp.yml#L16).
I don't think it's a requirement but more "a maybe nice to have".

v1v: For the time being, I'd say no notifications on a PR basis as we might need to revisit the user experience in terms of running the smoke tests. For some other cases, we provide some integrations with GitHub commands or GitHub tags, so anyone at Elastic, could trigger a particular stage/build without the need to use the Jenkins UI

simitt: In general the smoke tests should be run on a daily basis, not per PR. It would be nice (but can be added later if it doesn't fit this scope) to be able to trigger them on a PR basis with github commands, similar as to what we can use for other tests on PRs already (e.g. `/test` command).

v1v (approved):

PR #8685, opened 2022-07-21
ci: enable github actions for macos

v1v:
## Motivation/summary

Use GitHub actions for MacOS.

GitHub actions are SaaS so we don't need to worry about the MacOS workers but just use them

## Issue

A similar issue was done for beats in https://github.com/elastic/beats/pull/29032 and also disabled in the default CI controller in https://github.com/elastic/beats/pull/30591

## Follow ups

To export the telemetry data with -> https://github.com/elastic/beats/pull/32267 but to be done in the near future


axw:
@v1v we're moving away from mage, so rather than invest time in trying to get things working with it I've pushed a change to your branch to just run `go test -v ./...` directly. It does mean that there's no coverage files, but it's not really necessary anyway; we don't have any macOS-specific code, so coverage from other platforms is sufficient.

axw (approved):

PR #8689, opened 2022-07-22
Simplify CI

axw:
## Motivation/summary

- Stop calling mage in CI scripts, and replace
  "withMageEnv" with "withGoEnv" in Jenkinsfile.
- Remove "build" steps. `make apm-server` just builds the apm-server binary. If that would fail, `make test` would also fail.
- Just run Windows unit tests on Windows 2022, remove Windows 2019 stage.
- Post JUnit results only on linux/amd64; we don't have OS-specific code, so there's no need to repeat the results for each platform.
- Rename stages to use GOOS/GOARCH naming.
- Remove `df -h` debugging.
- Move gotestsum dependency to tools/go.mod.
- Remove GoTestUnit magefile target.
- Remove "/hey-apm" trigger and associated stage

## Checklist

## How to test these changes

N/A

## Related issues

None


v1v (approved):
LGTM

I've just triggered a new build with support on Windows (I added the `ci:windows` label) since it was not tested by default on this PR

PR #8727, opened 2022-07-27
fix: minor fix in make clean command

simitt:
## Motivation/summary

`make clean` is expected to remove a directory and files, the `-r` flag was missing from the command. 

## How to test these changes
run `make clean`


axw:
Oops, thanks!

Issue #8697, opened 2022-07-23
`make check-full` should check that BEATS_VERSION is correct


axw:
The apm-server main branch should have BEATS_VERSION=main, 8.3 should have 8.3, etc.
We should check this with `make check-full`, if possible, or in CI (on branch builds) otherwise.

PR #8728, opened 2022-07-27
feat: add BEATS_VERSION check make task

kruskall:
## Motivation/summary
add a new task to make sure the BEATS_VERSION variable
is synced with the current branch.
'check-beats-version' is also run as part of 'check-full'.

## Checklist

## Related issues

Closes https://github.com/elastic/apm-server/issues/8697


axw (changes requested):
This will fail on all PRs and local branches, where the apm-server branch name would be expected not to match.

I think we probably only want to run the target in CI, maybe as part of the daily `make update-beats` job? Otherwise we would need to figure out the expected branch from the version.

kruskall:
> This will fail on all PRs and local branches, where the apm-server branch name would be expected not to match.
> 
> I think we probably only want to run the target in CI, maybe as part of the daily `make update-beats` job? Otherwise we would need to figure out the expected branch from the version.

Good point! I've moved the check to be part of the `update-beats` job. Let me know what you think!

.ci/scripts/update-beats.sh
```diff
@@ -5,6 +5,8 @@
 #
 set -euo pipefail
 
+make check-beats-version
```
axw, line 8: I'm not sure that the branch name is necessarily as expected here either.

How about we introduce another script under `.ci/scripts` which accepts the branch name as an argument? Also, since the check is not generally going to work outside of CI, I would suggest we keep it contained within the script, rather than introducing a Make target. e.g.

```bash
#!/usr/bin/bash
set -xeu
exec grep "BEATS_VERSION?=$BRANCH_VERSION" Makefile
```

The update-beats pipeline (https://github.com/elastic/apm-server/blob/main/.ci/update-beats.groovy) already has a `$BRANCH_VERSION` environment variable we can use. So we would introduce another another `sh` step before the update which calls the script: https://github.com/elastic/apm-server/blob/ad545525e697164a454d4940c30dde12413ec016/.ci/update-beats.groovy#L58

kruskall: I like that! I've pushed a new commit 👍

.ci/update-beats.groovy
```diff
@@ -55,6 +55,7 @@ pipeline {
         dir("${BASE_DIR}"){
           withGoEnv(){
             setupAPMGitEmail(global: true)
+            sh(label: 'make check-beats-version', script: '.ci/scripts/check-beats-version.sh')
```
axw, line 58:
```suggestion
            sh(label: 'check BEATS_VERSION matches branch', script: '.ci/scripts/check-beats-version.sh')
```

axw (approved):
Thank you!

Issue #8644, opened 2022-07-19
Reproducible builds


axw:
We should ensure the apm-server binary is reproducibly built, meaning that given the same commit then `apm-server` binaries built are byte-for-byte identical, regardless of the machine on which they are built and the time at which they are built.

The build process currently injects the git commit and build timestamp via `-ldflags -X foo=bar`, which breaks reproducibility. Rather than injecting the build date, we should use the commit date. If the binary is reproducible, the time at which it was built is not important. Ideally we would extract the commit hash and date from [runtime/debug.BuildInfo](https://pkg.go.dev/runtime/debug#BuildInfo); this requires Go 1.18+. Alternatively, we can inject the values at build time, like we do today.

Doing this has several benefits, including auditability, and speeding up builds by caching.

simitt:
> Ideally we would extract the commit hash and date from [runtime/debug.BuildInfo](https://pkg.go.dev/runtime/debug#BuildInfo); this requires Go 1.18+.

https://github.com/elastic/apm-server/issues/8752 to track go 1.18 support

axw:
#8732 just injects the vars as before. We can consider using `runtime/debug` in the future.

PR #8732, opened 2022-07-28
Move building of binaries, Docker images, and packages to Makefile

axw:
## Motivation/summary

Move building of binaries, Docker images, and and packages out of Mage and into our Makefile. The goals are to simplify the build process, take greater ownership of the way in which we build and package APM Server, and better integrate with tools such as Skaffold (which can build a non-templated Dockerfile) for a better local development experience.

We introduce our own Dockerfile. This is based on the beats templated Dockerfile, with the following major changes:

 - We commit the `tini` binaries to the repo, to avoid having to download and verify during every image build step. The binaries are, as expected, tiny.
 - The Dockerfile is not templated, but uses build args where necessary. This would allow us to better integrate with tools like Skaffold, automatically building when source has changed.
 - We build the apm-server binary in Docker. This means the Docker image can be built without any prior steps.
 - We use BuildKit features to cache the Go module and Go build cache directories. On Linux, this requires
   setting DOCKER_BUILDKIT=1; on macOS and Windows this is the default with Docker Desktop.
 - We no longer explicitly create the "apm-server" user, instead we just set the UID to 1000. This simplifies the build, removing the need for installing packages. By avoiding the addition of any packages, we can switch the base image from Ubuntu to RedHat UBI8 without any other changes.

These are the major changes related to building binaries and packages:

 - Mage is no longer used at all, and has been removed as a dependency.
 - Static files added to packages are no longer templated.
 - We now use [nfpm](https://nfpm.goreleaser.com/), rather than [fpm](https://github.com/jordansissel/fpm). nfpm is a pure Go alternative to fpm, which means we have no dependency on Docker for building packages.
 - We now inject the Git commit timestamp into binaries for the "build time", rather than the system time at which the build happened. This means our Go binaries are now reproducible. We continue to use the system time for the Docker image builds: the base images may change from time to time (i.e. tags may be replaced), so they are not fully reproducible anyway.
- Docker image tarballs, and dependencies.csv, are now named according to the expectations of the unified release manager configuration.

Packaging targets have been renamed from "release-manager-release" and "release-manager-snapshot" to "package" and "package-snapshot" respectively. As we now use Make exclusively, building binaries and packages can run in parallel with `make -j<N> package`.

## Checklist

## How to test these changes

1. Install and smoke test each package and Docker image.
2. Confirm the commit hash and date are reported in the apm-server logs.
3. Compare IronBank Docker build context tarball to one built from the 8.4 branch, verify there are no discrepancies.
4. Build an apm-server binary multiple times, confirm that the build ID remains the same.
5. Confirm apm-server binaries are stripped
6. Confirm 64-bit apm-server are built with PIE, and 32-bit binaries are not

## Related issues

Closes https://github.com/elastic/apm-server/issues/8644


cachedout:
@axw This is a substantial change and I'll look at it but I'd also like @v1v to have a look as well. Since he's now out on leave, it will be about 2-3 weeks. Since you've targeted this for 8.5, I assume that's going to be fine but please let me know if you'd like to move faster than that.

axw:
@cachedout sorry, I should have stated earlier: I think the server team can review, I tagged observablt-ci as I changed some Jenkins bits. Those changes are not that significant though, so just a quick pass over should be fine, and I'd be happy even if it's post-commit. If there's any fallout after merging I'll fix it ASAP or revert.

.ci/packaging.groovy
```diff
@@ -263,9 +260,8 @@ def publishArtifactsDev() {
   uploadArtifacts(bucketLocation: "gs://${JOB_GCS_BUCKET}/${URI_SUFFIX}")
 
   dockerLogin(secret: env.DOCKER_SECRET, registry: env.DOCKER_REGISTRY)
-  dir("${BASE_DIR}"){
-    sh(label: 'Push', script: "./.ci/scripts/push-docker.sh ${env.GIT_BASE_COMMIT} ${env.DOCKER_IMAGE}")
-  }
+  sh(label: 'Tag Docker image', script: "docker tag ${dockerImage} ${env.DOCKER_IMAGE}:${env.GIT_BASE_COMMIT}")
+  sh(label: 'Push Docker image', script: "docker push ${env.DOCKER_IMAGE}:${env.GIT_BASE_COMMIT}")
```
cachedout, line 264: Nit: We often wrap this in a retry like `retryWithSleep(retries: 3, seconds: 5, backoff: true){`

cachedout (approved):

.ci/packaging.groovy
axw: Thanks! Added.

packaging/docker/docker-entrypoint
```diff
@@ -0,0 +1,33 @@
...
+  # If we _did_ get a subcommand, pass it to apm-server.
+  for subcommand in $subcommands; do
+      if [[ $1 == $subcommand ]]; then
+        exec $APM_SERVER "$@"
+      fi
+  done
+fi
+
+# If neither of those worked, then they have specified the binary they want, so
+# just do exactly as they say.
```
lahsivjar, line 32: Two questions, What does "they have specified the binary they want" mean? and Under what circumstances will this case be useful?

axw: FYI this is copied from beats. This means the user has run something like `docker run apm-server-image bash` (run `bash` inside the container). The alternatives are `docker run apm-server-image -e` (run `apm-server -e` inside the container) and `docker run apm-server-image test output` (run `apm-server test output` in the container).

TBH I think this entrypoint is overcomplicated, and it might be better to just require users to specify the apm-server command like `apm-server run apm-server-image apm-server -e`. Nevertheless, it remains this way for backwards compatibility.

lahsivjar: Got it, thanks for the explanation.

lahsivjar (approved):
LGTM

simitt:
@axw should `make package` still work? I ran into an issue with it.

axw:
@simitt yes it should, and it has been passing in CI. e.g. https://apm-ci.elastic.co/job/apm-server/job/apm-server-package-mbp/job/PR-8732/7/console

What was the issue? Was it perhaps related to permissions? If so, try removing the build directory first, I believe the old Docker-based packaging was creating files as root.

simitt (approved):
LGTM! The `make package` job worked smoothly after manually removing the old `build` folder.

packaging.mk
```diff
@@ -0,0 +1,216 @@
...
+	sed 's/$${APM_SERVER_VERSION_MAJORMINOR}/$(APM_SERVER_VERSION_MAJORMINOR)/' packaging/ironbank/README.md > $@/README.md
+$(DISTDIR)/%-docker-build-context.tar.gz: $(ARCHIVES_DIR)/%-docker-build-context
+	tar -C $< -czf $@ .
+
+##############################################################################
+# Unified release targets.
+##############################################################################
+
+PACKAGE_SUFFIXES := \
+	darwin-x86_64.tar.gz \
```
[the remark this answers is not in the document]
axw: We don't currently distribute a package for that combination. I added the binary target to make it possible for people to build and test locally, but intentionally kept the set of packages the same. We can follow up with additional packages later.

lahsivjar:
> Install and smoke test each package and Docker image.
> Confirm the commit hash and date are reported in the apm-server logs.

Confirmed `publish_ready` state, commit hash plus date in APM-Server logs, and successful event ingestion.
- [x] Linux AMD 64bit (tested using AWS EC2 instance)
- [x] Linux ARM 64bit (tested using AWS EC2 instance)
- [x] Windows AMD 64bit (tested using AWS EC2 instance)
<details>
<summary>Successful ingestion logs</summary>

</details>

> Compare IronBank Docker build context tarball to one built from the 8.4 branch, verify there are no discrepancies.

Verified.
<details>
<summary>Diff of docker build context tarball (8.4 vs 8.5)</summary>

```diff
$ diff apm-server-ironbank-8.4.3-docker-build-context apm-server-ironbank-8.5.0-docker-build-context                                                                                [14:50:18]
diff apm-server-ironbank-8.4.3-docker-build-context/Dockerfile apm-server-ironbank-8.5.0-docker-build-context/Dockerfile
11c11
< ARG ELASTIC_STACK=8.4.3
---
> ARG ELASTIC_STACK=8.5.0
diff apm-server-ironbank-8.4.3-docker-build-context/README.md apm-server-ironbank-8.5.0-docker-build-context/README.md
33c33
< License](https://github.com/elastic/apm-server/blob/8.4/licenses/ELASTIC-LICENSE.txt),
---
> License](https://github.com/elastic/apm-server/blob/8.5/licenses/ELASTIC-LICENSE.txt),
38c38
< [here](https://www.elastic.co/guide/en/apm/guide/current/release-notes-8.4.html).
---
> [here](https://www.elastic.co/guide/en/apm/guide/current/release-notes-8.5.html).
diff apm-server-ironbank-8.4.3-docker-build-context/hardening_manifest.yaml apm-server-ironbank-8.5.0-docker-build-context/hardening_manifest.yaml
11c11
< - "8.4.3"
---
> - "8.5.0"
18c18
<   ELASTIC_STACK: "8.4.3"
---
>   ELASTIC_STACK: "8.5.0"
32c32
<   org.opencontainers.image.version: 8.4.3"
---
>   org.opencontainers.image.version: "8.5.0"
42,43c42,43
<   - filename: "apm-server-8.4.3-linux-x86_64.tar.gz"
<     url: "<artifact_path>/apm-server-8.4.3-linux-x86_64.tar.gz"
---
>   - filename: "apm-server-8.5.0-linux-x86_64.tar.gz"
>     url: "<artifact_path>/apm-server-8.5.0-linux-x86_64.tar.gz"
```
</details>

> Build an apm-server binary multiple times, and confirm that the build ID remains the same.

Verified.

> Confirm apm-server binaries are stripped

Confirmed for `linux` and `windows` binaries. @axw For `darwin` binaries, I could not confirm this. Additionally, the darwin binaries don't show file reduction with and without the `-s` build flag. I also noticed that we don't strip the `DWARF` symbol table (by passing `-w` build flag). FWIW, the `-w` flag reduces the darwin build size though I am not sure what is the reason behind this.
<details>
<summary>Proof using `file` command</summary>

```bash
$ file build/apm-server-linux-386
build/apm-server-linux-386: ELF 32-bit LSB executable, Intel 80386, version 1 (SYSV), statically linked, Go BuildID=VZT495-qfGRzG1HJsOjC/cat_vQ1yxp3EFhJmy4Mv/VXm9k-st7GuxEdBUHKsm/YTVwaA75ysYM-9j0B2ju, stripped

$ file build/apm-server-linux-amd64
build/apm-server-linux-amd64: ELF 64-bit LSB pie executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, Go BuildID=seQlhj1ru04LvlaSd7rl/CJ-_aW7Vgj4JE2JoBlDa/B-LTGLFV9GQRbIp2ol1W/J7ZPcQCLVkAnvxVaHry3, stripped

$ file build/apm-server-linux-arm64
build/apm-server-linux-arm64: ELF 64-bit LSB pie executable, ARM aarch64, version 1 (SYSV), dynamically linked, interpreter /lib/ld-linux-aarch64.so.1, Go BuildID=TI4ukGiBplkO3htyRmAF/0i-gKYFwCA6XVEIP9Br_/b2YS814PuC_E2Uj3emRd/MHJ-5BvOwtSghY-p_83b, stripped

$ file build/apm-server-windows-386.exe
build/apm-server-windows-386.exe: PE32 executable (console) Intel 80386 (stripped to external PDB), for MS Windows

$ file build/apm-server-windows-amd64.exe
build/apm-server-windows-amd64.exe: PE32+ executable (console) x86-64 (stripped to external PDB), for MS Windows
```
</details>

> Confirm 64-bit apm-server are built with PIE, and 32-bit binaries are not

Verified for linux and darwin binaries.
<details>
<summary>Proof using `file` command for linux binaries and `otool` for darwin binary</summary>

```bash
$ file build/apm-server-linux*
build/apm-server-linux-386:   ELF 32-bit LSB **executable**, Intel 80386, version 1 (SYSV), statically linked, Go BuildID=VZT495-qfGRzG1HJsOjC/cat_vQ1yxp3EFhJmy4Mv/VXm9k-st7GuxEdBUHKsm/YTVwaA75ysYM-9j0B2ju, stripped
build/apm-server-linux-amd64: ELF 64-bit LSB **pie executable**, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, Go BuildID=seQlhj1ru04LvlaSd7rl/CJ-_aW7Vgj4JE2JoBlDa/B-LTGLFV9GQRbIp2ol1W/J7ZPcQCLVkAnvxVaHry3, stripped
build/apm-server-linux-arm64: ELF 64-bit LSB **pie executable**, ARM aarch64, version 1 (SYSV), dynamically linked, interpreter /lib/ld-linux-aarch64.so.1, Go BuildID=TI4ukGiBplkO3htyRmAF/0i-gKYFwCA6XVEIP9Br_/b2YS814PuC_E2Uj3emRd/MHJ-5BvOwtSghY-p_83b, stripped

$ otool -hv build/apm-server-darwin-amd64
build/apm-server-darwin-amd64:
Mach header
      magic  cputype cpusubtype  caps    filetype ncmds sizeofcmds      flags
MH_MAGIC_64   X86_64        ALL  0x00     EXECUTE    15       2760 DYLDLINK PIE
```
</details>

axw:
@lahsivjar thanks for the thorough testing and detailed writeup!

> Confirmed for linux and windows binaries. @axw For darwin binaries, I could not confirm this. Additionally, the darwin binaries don't show file reduction with and without the -s build flag.

I think it's good enough to have confirmed on Linux, since we use the same flags for all target platforms.

> I also noticed that we don't strip the DWARF symbol table (by passing -w build flag). FWIW, the -w flag reduces the darwin build size but not for other distros though I am not sure what is the reason behind this.

AFAIK the beats/mage code also doesn't pass `-w` to the Go linker -- only `-s`. If you think it would be worthwhile, please open an enhancement issue.

PR #8770, opened 2022-08-02
.ci: cache Go modules

axw:
## Motivation/summary

Speed up CI by caching Go module dependencies. Note that the build will still fetch missing modules if `go.mod` changes, and will still check cached modules match `go.sum`; this will not lead to stale dependencies being used.

Remove some magefile cruft.

## Checklist

## How to test these changes

N/A

## Related issues

N/A


.ci/packer_cache.sh
```diff
@@ -18,3 +19,8 @@ if [ -x "$(command -v docker)" ]; then
   (retry 2 docker pull "${image}") || echo "Error pulling ${image} Docker image, we continue"
   done
 fi
+
+# Download Go module dependencies.
+go mod download
```
v1v, line 24: If this is not required then `|| true` for each command otherwise the cache might fail and affect the cache image generations for some other projects.

v1v (commented):
As mentioned, if possible avoid failing if any of the commands failed

.ci/packer_cache.sh
axw: TIL, thanks. Good thing you pointed this out, since there was an error with one of the `go mod download` lines.

v1v (approved):

Issue #8042, opened 2022-05-04
Otel Java agent 1.13.0 renamed JVM GC metrics to follow Otel Semantic Conventions


cyrille-leclerc:
The [Otel Java agent 1.13.0](https://github.com/open-telemetry/opentelemetry-java-instrumentation/releases/tag/v1.13.0) renamed JVM GC metrics to follow the [OpenTelemetry Semantic Conventions for JVM metrics](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/runtime-environment-metrics.md#jvm-metrics)

Old metrics:
```
runtime.jvm.memory.area{type="used",area="heap"} 2000000
runtime.jvm.memory.area{type="committed",area="non_heap"} 200000
runtime.jvm.memory.area{type="used",pool="PS Eden Space"} 2000
```

New metrics:
```
 process.runtime.jvm.memory.init{type="heap",pool="G1 Eden Space"} 1000000
 process.runtime.jvm.memory.usage{type="heap",pool="G1 Eden Space"} 2500000
 process.runtime.jvm.memory.committed{type="heap",pool="G1 Eden Space"} 3000000
 process.runtime.jvm.memory.max{type="heap",pool="G1 Eden Space"} 4000000
 process.runtime.jvm.memory.init{type="non_heap",pool="Metaspace"} 200
 process.runtime.jvm.memory.usage{type="non_heap",pool="Metaspace"} 400
 process.runtime.jvm.memory.committed{type="non_heap",pool="Metaspace"} 500
```

See new code https://raw.githubusercontent.com/open-telemetry/opentelemetry-java-instrumentation/v1.13.1/instrumentation/runtime-metrics/library/src/main/java/io/opentelemetry/instrumentation/runtimemetrics/MemoryPools.java

cyrille-leclerc:
@AlexanderWert could you please help us map the new Otel metrics to the Elastic APM GUI metrics?

atahar-shapuri:
@kruskall  Can you please let us know ,Which version this is fixed ?

PR #8777, opened 2022-08-02
feat: add support for Otel semantic conventions for JVM metrics

kruskall:
## Motivation/summary

Handle new metrics properly:

- 'type' was previously called 'area'
- 'name' includes what was previously called 'type'

See https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/runtime-environment-metrics.md#jvm-metrics

## Checklist

## Related issues

Closes https://github.com/elastic/apm-server/issues/8042


internal/processor/otel/metrics.go
```diff
@@ -176,6 +176,34 @@ func (b *apmMetricsBuilder) accumulate(m pmetric.Metric) {
...
+					case "usage":
+						key.jvmType = "used"
+					case "committed":
+						key.jvmType = "committed"
+					}
+
+					if key.jvmType != "" {
+						b.jvmMemory[key] = apmMetricValue{dp.Timestamp().AsTime(), sample.Value}
+					}
 				case "runtime.jvm.memory.area":
```
axw, line 207: Could you please add a comment to mention that `runtime.jvm.*` metrics were renamed in the OTel Java SDK v1.13.0 (https://github.com/open-telemetry/opentelemetry-java-instrumentation/releases/tag/v1.13.0)? We should remove this code some time in the future.

axw (approved):
LGTM, thanks!

internal/processor/otel/metrics.go
kruskall: Good point! Added

PR #8779, opened 2022-08-02
Actually running attacher as another user

eyalkoren:
The `java --version` command was ran properly as the target user, however not the actual attacher command 🤦‍♂️ 
Probably dropped when this part was refactored.


simitt (approved):

simitt:
Eventually there should be a test covering that the cmd is triggered by the expected user.

eyalkoren:
> Eventually there should be a test covering that the cmd is triggered by the expected user.

Eventually... 
Unfortunately it competes with my other duties.

Issue #7256, opened 2022-02-11
Rename `process.ppid` to `process.parent.pid`


axw:
[ECS 8.0](https://github.com/elastic/ecs/releases/tag/v8.0.0) removed `process.ppid`, which has been replaced with `process.parent.pid`. We should update the server to do the same thing, and investigate whether an alias from `process.ppid` is needed.

We will also need to update our ingest pipelines to rename `process.ppid` sent by older versions of APM Server.

simitt:
@sqren is the APM UI using the `process.ppid` somewhere or is the information just shown? I'm leaning towards just renaming the field without adding an alias if it is not used anywhere.

sqren:
Feel free to rename. It's used in one place but we can easily change that:
https://github.com/elastic/kibana/blob/8b7395d594e87070e3b84ffdc91ceca1ca8c58cb/x-pack/plugins/apm/common/correlations/constants.ts#L27

simitt:
@sqren would that break anything when viewing older data that still have `process.ppid`?

sqren:
No, it wouldn't break anything. We only use it in an exclude list so as not to display that value for that field to the user. We can simply add the new field to that list.

sqren:
I created an issue to track this on the APM UI side: https://github.com/elastic/kibana/issues/127849

PR #8783, opened 2022-08-02
feat: rename process.ppid to process.parent.pid

kruskall:
## Motivation/summary

ECS 8.0 removed process.ppid, which has been
replaced with process.parent.pid

## Checklist

## Related issues

Closes https://github.com/elastic/apm-server/issues/7256


axw (changes requested):
@kruskall a few things:

- The intake API should remain unchanged, it's only the Elasticsearch document structure that should be updated. In other words, there should be no changes to `modeldecoder/v2/model.go`.
- The `process.ppid` field needs to be renamed in `apmpackage/data_stream/*/fields/ecs.yml`. I think we should probably also add an alias from `process.ppid` to `process.parent.pid`
- We will need to update our ingest pipelines to rename `process.ppid` sent by older versions of APM Server, for each of the data streams modified above.

For the last point, I suggest defining a new ingest pipeline in https://github.com/elastic/apm-server/blob/main/apmpackage/cmd/genpackage/pipelines.go, and then inserting a reference to it in each data stream's `ingest_pipeline/default.yml`.

kruskall:
> @kruskall a few things:
> 
>     * The intake API should remain unchanged, it's only the Elasticsearch document structure that should be updated. In other words, there should be no changes to `modeldecoder/v2/model.go`.
> 
>     * The `process.ppid` field needs to be renamed in `apmpackage/data_stream/*/fields/ecs.yml`. I think we should probably also add an alias from `process.ppid` to `process.parent.pid`
> 
>     * We will need to update our ingest pipelines to rename `process.ppid` sent by older versions of APM Server, for each of the data streams modified above.
> 
> 
> For the last point, I suggest defining a new ingest pipeline in https://github.com/elastic/apm-server/blob/main/apmpackage/cmd/genpackage/pipelines.go, and then inserting a reference to it in each data stream's `ingest_pipeline/default.yml`.

Thanks for the help and the explanation! 🙇 

I've updated the code!

axw (approved):
LGTM, thanks!

Issue #8755, opened 2022-07-29
Improve support for mobile agents - map additional fields


simitt:
Elastic mobile agents (Android and iOS) are built on OpenTelemetry, and therefore use the otel protocol, instead of the APM Server Intake API. https://github.com/elastic/apm/blob/main/specs/agents/mobile/README.md lists mobile agent specific fields and their desired mappings. Not all of these fields are already mapped by the APM Server. Unmapped fields are currently stored under `labels.*`. 

We want to improve the support for resource and span attributes, collected by these agents, to improve the experience when consuming these data. This issue is about mapping the remaining fields that are sent by the mobile agents: 

* `host.os.version`, `host.os.name`, `host.os.type` (`os.type` has been extended to contain `ios` and `android`, see https://github.com/elastic/ecs/pull/1999/files)
* `session.id`: see existing [RFC](https://github.com/elastic/ecs/blob/75d9daae17fdb3c3fef84cd5cbfd017ca156427a/rfcs/text/0004-session.md)

To be removed from ES documents:
*  `telemetry.sdk.elastic_export_timestamp` [is set to deal with time skews](https://github.com/elastic/apm-server/blob/f5f0272b1d18e42db75a14e3739deb6cc1d341e2/processor/otel/timestamps.go#L29). It is only relevant for APM Server internal usage, and should therefore be removed from the document that ends up in ES. 

## Out of Scope
* Mapping the `device.*` fields is handled in a separate issue https://github.com/elastic/apm-server/issues/4879.

PR #8784, opened 2022-08-02
feat: map additional fields from mobile agents

kruskall:
## Motivation/summary

Add mapping for 'host.os.version' and 'host.os.name'.
Android and ios are recognized and os type is set
accordingly.

## Checklist

## Related issues

Closes https://github.com/elastic/apm-server/issues/8755


internal/processor/otel/metadata.go
```diff
@@ -153,6 +157,13 @@ func translateResourceMetadata(resource pcommon.Resource, out *model.APMEvent) {
 		out.Host.OS.Type = "unix"
 	}
 
+	switch out.Host.Name {
```
axw, line 160:
```suggestion
	switch out.Host.OS.Name {
```

internal/processor/otel/metadata.go
```diff
@@ -153,6 +157,13 @@ func translateResourceMetadata(resource pcommon.Resource, out *model.APMEvent) {
...
+	case "Android":
+		out.Host.Type = "android"
+	case "iOS":
+		out.Host.Type = "ios"
```
axw, line 164:
```suggestion
		out.Host.OS.Type = "android"
	case "iOS":
		out.Host.OS.Type = "ios"
```

axw (commented):
Can you please add tests which show that `host.os.{name,type}` are set correctly for iOS and Android?

Also, these changes aren't quite enough to close https://github.com/elastic/apm-server/issues/8755. There's also `session.id` to handle, and we need to make sure `telemetry.sdk.elastic_export_timestamp` isn't recorded as a label. Assuming we'll do those as followups, please remove the "Closes" from the description.

internal/processor/otel/metadata.go
kruskall: Good catch! Changed 👍

kruskall:
> Can you please add tests which show that `host.os.{name,type}` are set correctly for iOS and Android?
Added 👍 

> Also, these changes aren't quite enough to close #8755. There's also `session.id` to handle, 

I see that the documentation doesn't mention a convention yet (see https://github.com/elastic/apm/blob/main/specs/agents/mobile/README.md#additional-span-attributes). Should we go with the RFC linked in the issue ?

> and we need to make sure `telemetry.sdk.elastic_export_timestamp` isn't recorded as a label.

I've pushed another commit to handle that and I've added assertions to verify that.

axw:
> I see that the documentation doesn't mention a convention yet (see https://github.com/elastic/apm/blob/main/specs/agents/mobile/README.md#additional-span-attributes). Should we go with the RFC linked in the issue ?

There's two parts to this that aren't yet specified:
 - the OTel side, there's no convention yet for encoding session IDs. However, the Elastic APM iOS agent is encoding this as a span attribute called `session.id`: https://github.com/elastic/apm-agent-ios/blob/93febce05fca0f407136b5066c9f87a79d44f86c/Sources/apm-agent-ios/Instrumentation/TapLogger.swift#L68
 - in ECS, there's only the RFC which you mentioned. Nevertheless, we've been recording `session.id` for RUM for a while already: https://github.com/elastic/apm-server/blob/b6ede69b407ee58ba3b9bbd4c8b63b970c28ab11/internal/model/apmevent.go#L50

So we should look for OTel `session.id` span attributes (for both transactions and spans I suppose), and record its value in the field `model.APMEvent.Session.ID`.

internal/processor/otel/metadata.go
```diff
@@ -106,6 +106,10 @@ func translateResourceMetadata(resource pcommon.Resource, out *model.APMEvent) {
 		case "host.arch":
 			out.Host.Architecture = truncate(v.StringVal())
 
+		// session.*
+		case "session.id":
+			out.Session.ID = truncate(v.StringVal())
```
axw, line 111: This one will come from [span attributes](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/README.md), not [resource attributes](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/semantic_conventions/README.md).

In case you're not already familiar with the distinction:
 - resource attributes describe the entity producing observability data, and the environment in which that entity exists. This comprises things like `service.name`, `process.pid`, `os.name`, etc.
 - span attributes describe an individual operation, e.g. `http.method` for an HTTP request, or `db.statement` for a database query

Session IDs are event-specific, so we'll need to handle them in `traces.go`: I think both in `TranslateTransaction` and `TranslateSpan`.

kruskall: Oh, I see. Thank you for the explanation! 🙇 

I've updated the code!

axw (approved):
LGTM, thanks :)
Please remember to update changelog(s).

PR #8809, opened 2022-08-08
Fix event loss during reload of TBS processor

### LINKED ISSUE #8339: Reconfiguring tail-based sampling leads to event loss
**APM Server version** (`apm-server version`): 8.3.0 (main)

When APM Server has tail-based sampling enabled, and its policies are modified or TBS is disabled, then the hot-reload of APM Server's configuration leads to events being dropped.

When the server is reconfigured, it will internally start a new processor pipeline and a listener which sends requests through to that pipeline. It should then close the old network listener, wait for existing HTTP requests to be processed, and then stop the existing processor pipeline.

During the reconfiguration, we're seeing the error `processor is stopped`, which is logged when events are processed by a stopped tail-based sampling processor. This suggests that the processor is being stopped before in-flight events have all been processed.

**Steps to reproduce**:

 1. Run apm-server with tail-based sampling enabled, with a single catch-all policy: `[{"sample_rate":0.5}]`
 2. Send a continuous stream of transactions
 3. Disable tail-based sampling
 4. Observe `processor is stopped` in the logs, and the rate of transactions docs drops

lahsivjar:
## Motivation/summary

TBS processor might lead to loss of events on reload. This PR fixes two edge cases for the same:

1. As per the reload logic for configurations, it is possible that multiple TBS processors are live for a small period of time during reloading. This introduces race between the two processors which utilize trace ID based sharded read writers. Additionally, badger uses a serializable snapshot isolation strategy which will cause a newly created read-write shard not to see data committed by the older shard for the same trace ID if it was created before the older shard flushed all its contents.
2. On shutdown of the processor, the TBS reservoir is not flushed which will lead to event loss during reloading. 

## Checklist

## How to test these changes

TBS reservoir flushing before shutdown can be tested by running `TestGracefulShutdown` testcase:

```
go test -v ./... -run TestGracefulShutdown
```

## Related issues

Closes #8339


x-pack/apm-server/sampling/processor.go
```diff
@@ -397,6 +367,7 @@ func (p *Processor) Run() error {
 		case <-ctx.Done():
 			return ctx.Err()
 		case <-p.stopping:
+			<-time.After(shutdownGracePeriod)
```
lahsivjar, line 370: **[For reviewers]** The benchmark reports a high delta for [BenchmarkProcess](https://github.com/elastic/apm-server/blob/cf8f0a616f94aadb00595c81ccdc2a61aa8ebd2d/x-pack/apm-server/sampling/processor_bench_test.go#L22). This is caused due to the introduction of the shutdown grace period.

I have run the same benchmarks by disabling the wait here with the following results (lines with `all equal` are removed):

```
name                                                            old time/op    new time/op    delta
pkg:github.com/elastic/apm-server/internal/agentcfg goos:darwin goarch:arm64
FetchAndAdd/FetchFromCache-10                                     62.1ns ± 2%    62.8ns ± 4%     ~     (p=0.548 n=5+5)
FetchAndAdd/FetchAndAddToCache-10                                  132ns ± 0%     134ns ± 4%   +1.91%  (p=0.032 n=4+5)
pkg:github.com/elastic/apm-server/internal/decoder goos:darwin goarch:arm64
CompressedRequestReader/uncompressed-10                            114ns ± 2%     111ns ± 4%     ~     (p=0.056 n=5+5)
CompressedRequestReader/gzip_content_encoding-10                  4.03µs ± 5%    3.51µs ±12%  -12.78%  (p=0.032 n=4+5)
CompressedRequestReader/gzip_sniff-10                             4.43µs ±65%    3.73µs ±15%     ~     (p=0.841 n=5+5)
CompressedRequestReader/deflate_content_encoding-10               6.49µs ±48%    3.51µs ±23%  -45.99%  (p=0.032 n=5+5)
CompressedRequestReader/deflate_sniff-10                          5.23µs ±43%    3.37µs ± 8%     ~     (p=0.056 n=5+5)
pkg:github.com/elastic/apm-server/internal/model/modelindexer goos:darwin goarch:arm64
ModelIndexer/NoCompression-10                                     1.65µs ±10%    1.54µs ± 6%     ~     (p=0.421 n=5+5)
ModelIndexer/BestSpeed-10                                         3.40µs ±11%    3.22µs ± 0%     ~     (p=0.151 n=5+5)
ModelIndexer/DefaultCompression-10                                5.28µs ± 2%    5.68µs ±21%     ~     (p=0.690 n=5+5)
ModelIndexer/BestCompression-10                                   8.44µs ± 7%    8.38µs ± 8%     ~     (p=0.841 n=5+5)
pkg:github.com/elastic/apm-server/internal/processor/stream goos:darwin goarch:arm64
BackendProcessor-10                                               0.00ns ±59%    0.00ns ±60%     ~     (p=0.286 n=5+5)
RUMV3Processor-10                                                 0.00ns ±48%    0.00ns ±24%     ~     (p=0.095 n=5+5)
BackendProcessorParallel/BenchmarkBackendProcessorParallel0-10    0.00ns ±67%    0.00ns ±63%     ~     (p=0.841 n=5+5)
BackendProcessorParallel/BenchmarkBackendProcessorParallel2-10    0.00ns ±21%    0.00ns ±24%  -21.65%  (p=0.032 n=5+5)
BackendProcessorParallel/BenchmarkBackendProcessorParallel4-10    0.00ns ±25%    0.00ns ±28%     ~     (p=0.111 n=5+5)
BackendProcessorParallel/BenchmarkBackendProcessorParallel8-10    0.00ns ±36%    0.00ns ±53%     ~     (p=0.222 n=5+5)
pkg:github.com/elastic/apm-server/internal/publish goos:darwin goarch:arm64
Publisher-10                                                       1.00s ± 0%     1.00s ± 0%   +0.16%  (p=0.032 n=5+5)
pkg:github.com/elastic/apm-server/x-pack/apm-server/aggregation/spanmetrics goos:darwin goarch:arm64
AggregateSpan-10                                                   955ns ± 4%    1020ns ±11%     ~     (p=0.381 n=5+5)
pkg:github.com/elastic/apm-server/x-pack/apm-server/aggregation/txmetrics goos:darwin goarch:arm64
AggregateTransaction-10                                            214ns ± 1%     203ns ± 7%     ~     (p=0.095 n=5+5)
pkg:github.com/elastic/apm-server/x-pack/apm-server/sampling goos:darwin goarch:arm64
TraceGroups-10                                                     309ns ± 6%     291ns ± 2%   -5.76%  (p=0.016 n=5+5)
Process-10                                                       51.0ms ±135%    76.9ms ±68%     ~     (p=0.548 n=5+5)
pkg:github.com/elastic/apm-server/x-pack/apm-server/sampling/eventstorage goos:darwin goarch:arm64
ShardedWriteTransactionUncontended-10                             1.12µs ±28%    1.12µs ±24%     ~     (p=0.730 n=5+5)
ShardedWriteTransactionContended-10                               6.48µs ± 1%    6.46µs ± 1%     ~     (p=0.841 n=5+5)
WriteTransaction/json_codec-10                                    5.63µs ± 2%    5.65µs ± 4%     ~     (p=0.841 n=5+5)
WriteTransaction/json_codec_big_tx-10                             6.79µs ± 3%    6.80µs ± 1%     ~     (p=1.000 n=5+5)
WriteTransaction/nop_codec-10                                      214ns ± 6%     212ns ± 2%     ~     (p=0.841 n=5+5)
WriteTransaction/nop_codec_big_tx-10                               211ns ± 2%     207ns ± 1%   -1.90%  (p=0.040 n=5+5)
ReadEvents/json_codec/0_events-10                                  323ns ± 5%     321ns ± 2%     ~     (p=0.841 n=5+5)
ReadEvents/json_codec/1_events-10                                 11.8µs ±11%    12.1µs ±13%     ~     (p=0.841 n=5+5)
ReadEvents/json_codec/10_events-10                                 103µs ± 2%     101µs ± 0%   -1.95%  (p=0.008 n=5+5)
ReadEvents/json_codec/100_events-10                                942µs ± 3%     927µs ± 2%     ~     (p=0.548 n=5+5)
ReadEvents/json_codec/199_events-10                               1.95ms ± 4%    1.95ms ± 4%     ~     (p=0.841 n=5+5)
ReadEvents/json_codec/399_events-10                               3.96ms ± 1%    3.93ms ± 1%   -0.89%  (p=0.032 n=5+5)
ReadEvents/json_codec/1000_events-10                              10.1ms ± 1%    10.0ms ± 2%     ~     (p=0.548 n=5+5)
ReadEvents/json_codec_big_tx/0_events-10                           327ns ± 2%     322ns ± 2%     ~     (p=0.222 n=5+5)
ReadEvents/json_codec_big_tx/1_events-10                          13.3µs ± 1%    13.9µs ± 4%   +4.29%  (p=0.032 n=5+5)
ReadEvents/json_codec_big_tx/10_events-10                          119µs ± 5%     115µs ± 2%     ~     (p=0.056 n=5+5)
ReadEvents/json_codec_big_tx/100_events-10                        1.12ms ± 3%    1.11ms ± 3%     ~     (p=0.222 n=5+5)
ReadEvents/json_codec_big_tx/199_events-10                        2.24ms ± 3%    2.21ms ± 4%     ~     (p=0.310 n=5+5)
ReadEvents/json_codec_big_tx/399_events-10                        4.60ms ± 4%    4.46ms ± 0%   -3.08%  (p=0.032 n=5+5)
ReadEvents/json_codec_big_tx/1000_events-10                       11.5ms ± 5%    11.4ms ± 2%     ~     (p=0.690 n=5+5)
ReadEvents/nop_codec/0_events-10                                   331ns ± 3%     324ns ± 1%     ~     (p=0.151 n=5+5)
ReadEvents/nop_codec/1_events-10                                  1.99µs ± 5%    1.99µs ± 4%     ~     (p=0.841 n=5+5)
ReadEvents/nop_codec/10_events-10                                 11.0µs ±15%    10.9µs ± 9%     ~     (p=0.841 n=5+5)
ReadEvents/nop_codec/100_events-10                                98.6µs ±17%    92.9µs ±10%     ~     (p=0.310 n=5+5)
ReadEvents/nop_codec/199_events-10                                 179µs ±11%     170µs ± 2%     ~     (p=0.690 n=5+5)
ReadEvents/nop_codec/399_events-10                                 332µs ±20%     326µs ± 4%     ~     (p=0.421 n=5+5)
ReadEvents/nop_codec/1000_events-10                                765µs ±12%     738µs ± 6%     ~     (p=0.690 n=5+5)
ReadEvents/nop_codec_big_tx/0_events-10                            331ns ± 1%     330ns ± 4%     ~     (p=0.254 n=5+5)
ReadEvents/nop_codec_big_tx/1_events-10                           1.99µs ± 7%    1.98µs ± 3%     ~     (p=0.889 n=5+5)
ReadEvents/nop_codec_big_tx/10_events-10                          10.4µs ± 2%    10.6µs ± 1%     ~     (p=0.063 n=5+4)
ReadEvents/nop_codec_big_tx/100_events-10                         94.9µs ±17%    92.6µs ±15%     ~     (p=1.000 n=5+5)
ReadEvents/nop_codec_big_tx/199_events-10                          178µs ±13%     170µs ± 5%     ~     (p=0.841 n=5+5)
ReadEvents/nop_codec_big_tx/399_events-10                          345µs ±20%     342µs ±16%     ~     (p=1.000 n=5+5)
ReadEvents/nop_codec_big_tx/1000_events-10                         826µs ±13%     729µs ±17%     ~     (p=0.095 n=5+5)
IsTraceSampled/sampled-10                                         93.9ns ± 5%    95.5ns ± 4%     ~     (p=0.421 n=5+5)
IsTraceSampled/unsampled-10                                       92.5ns ± 3%    94.6ns ± 4%     ~     (p=0.095 n=5+5)
IsTraceSampled/unknown-10                                          428ns ± 3%     429ns ± 2%     ~     (p=0.690 n=5+5)

name                                                            old alloc/op   new alloc/op   delta
pkg:github.com/elastic/apm-server/internal/agentcfg goos:darwin goarch:arm64
pkg:github.com/elastic/apm-server/internal/decoder goos:darwin goarch:arm64
CompressedRequestReader/gzip_content_encoding-10                  45.5kB ± 0%    45.5kB ± 0%     ~     (p=0.897 n=5+5)
CompressedRequestReader/gzip_sniff-10                             45.5kB ± 0%    45.5kB ± 0%     ~     (p=0.825 n=5+5)
CompressedRequestReader/deflate_content_encoding-10               44.9kB ± 0%    44.9kB ± 0%   -0.05%  (p=0.016 n=5+5)
CompressedRequestReader/deflate_sniff-10                          44.9kB ± 0%    44.9kB ± 0%     ~     (p=0.175 n=5+5)
pkg:github.com/elastic/apm-server/internal/model/modelindexer goos:darwin goarch:arm64
ModelIndexer/NoCompression-10                                     3.28kB ± 2%    3.30kB ± 1%     ~     (p=0.460 n=5+5)
ModelIndexer/BestSpeed-10                                         2.59kB ± 1%    2.58kB ± 0%     ~     (p=0.214 n=5+5)
ModelIndexer/DefaultCompression-10                                2.63kB ± 1%    2.63kB ± 1%     ~     (p=0.690 n=5+5)
ModelIndexer/BestCompression-10                                   2.69kB ± 2%    2.68kB ± 1%     ~     (p=0.643 n=5+5)
pkg:github.com/elastic/apm-server/internal/processor/stream goos:darwin goarch:arm64
pkg:github.com/elastic/apm-server/internal/publish goos:darwin goarch:arm64
Publisher-10                                                       293kB ±11%     279kB ±14%     ~     (p=0.421 n=5+5)
pkg:github.com/elastic/apm-server/x-pack/apm-server/aggregation/spanmetrics goos:darwin goarch:arm64
pkg:github.com/elastic/apm-server/x-pack/apm-server/aggregation/txmetrics goos:darwin goarch:arm64
AggregateTransaction-10                                            2.60B ±23%     2.00B ± 0%     ~     (p=0.095 n=5+4)
pkg:github.com/elastic/apm-server/x-pack/apm-server/sampling goos:darwin goarch:arm64
Process-10                                                       39.7MB ±124%    59.2MB ±50%     ~     (p=0.421 n=5+5)
pkg:github.com/elastic/apm-server/x-pack/apm-server/sampling/eventstorage goos:darwin goarch:arm64
ShardedWriteTransactionUncontended-10                             3.43kB ± 3%    3.54kB ± 4%     ~     (p=0.111 n=5+5)
ShardedWriteTransactionContended-10                               7.63kB ± 0%    7.62kB ± 1%     ~     (p=0.413 n=4+5)
WriteTransaction/json_codec_big_tx-10                             3.52kB ± 0%    3.52kB ± 0%     ~     (p=0.190 n=5+5)
ReadEvents/json_codec/1_events-10                                 8.23kB ± 0%    8.23kB ± 0%     ~     (p=0.714 n=5+5)
ReadEvents/json_codec/10_events-10                                75.3kB ± 0%    75.3kB ± 0%     ~     (p=0.341 n=5+5)
ReadEvents/json_codec/100_events-10                                750kB ± 0%     750kB ± 0%     ~     (p=0.690 n=5+5)
ReadEvents/json_codec/199_events-10                               1.02MB ± 0%    1.02MB ± 0%   -0.09%  (p=0.040 n=5+5)
ReadEvents/json_codec/399_events-10                               1.57MB ± 0%    1.56MB ± 0%     ~     (p=0.270 n=5+5)
ReadEvents/json_codec/1000_events-10                              3.36MB ± 0%    3.29MB ± 0%   -2.27%  (p=0.029 n=4+4)
ReadEvents/json_codec_big_tx/1_events-10                          10.1kB ± 0%    10.1kB ± 0%     ~     (p=0.643 n=5+5)
ReadEvents/json_codec_big_tx/10_events-10                         94.5kB ± 0%    94.5kB ± 0%     ~     (p=0.690 n=5+5)
ReadEvents/json_codec_big_tx/100_events-10                         942kB ± 0%     942kB ± 0%     ~     (p=0.135 n=5+5)
ReadEvents/json_codec_big_tx/199_events-10                        1.33MB ± 0%    1.33MB ± 0%     ~     (p=0.571 n=5+5)
ReadEvents/json_codec_big_tx/399_events-10                        2.11MB ± 0%    2.11MB ± 0%     ~     (p=0.524 n=5+4)
ReadEvents/json_codec_big_tx/1000_events-10                       4.62MB ± 0%    4.62MB ± 0%     ~     (p=0.571 n=5+4)
ReadEvents/nop_codec/1_events-10                                  3.25kB ± 0%    3.26kB ± 0%     ~     (p=0.381 n=5+5)
ReadEvents/nop_codec/10_events-10                                 25.6kB ± 0%    25.6kB ± 0%     ~     (p=1.000 n=5+5)
ReadEvents/nop_codec/100_events-10                                 249kB ± 0%     249kB ± 0%     ~     (p=0.310 n=5+5)
ReadEvents/nop_codec/199_events-10                                 467kB ± 0%     467kB ± 0%     ~     (p=0.310 n=5+5)
ReadEvents/nop_codec/399_events-10                                 890kB ± 0%     890kB ± 0%     ~     (p=1.000 n=5+5)
ReadEvents/nop_codec/1000_events-10                               2.15MB ± 0%    2.15MB ± 0%     ~     (p=0.230 n=5+5)
ReadEvents/nop_codec_big_tx/1_events-10                           3.25kB ± 0%    3.25kB ± 0%     ~     (p=1.000 n=5+4)
ReadEvents/nop_codec_big_tx/10_events-10                          25.6kB ± 0%    25.6kB ± 0%     ~     (p=0.087 n=5+5)
ReadEvents/nop_codec_big_tx/100_events-10                          249kB ± 0%     249kB ± 0%     ~     (p=0.690 n=5+5)
ReadEvents/nop_codec_big_tx/199_events-10                          467kB ± 0%     467kB ± 0%     ~     (p=0.302 n=5+4)
ReadEvents/nop_codec_big_tx/399_events-10                          890kB ± 0%     889kB ± 0%     ~     (p=0.516 n=5+5)
ReadEvents/nop_codec_big_tx/1000_events-10                        2.15MB ± 0%    2.14MB ± 0%   -0.33%  (p=0.008 n=5+5)
IsTraceSampled/unknown-10                                           334B ± 4%      334B ± 4%     ~     (p=1.000 n=5+5)

name                                                            old allocs/op  new allocs/op  delta
pkg:github.com/elastic/apm-server/internal/agentcfg goos:darwin goarch:arm64
pkg:github.com/elastic/apm-server/internal/decoder goos:darwin goarch:arm64
pkg:github.com/elastic/apm-server/internal/model/modelindexer goos:darwin goarch:arm64
pkg:github.com/elastic/apm-server/internal/processor/stream goos:darwin goarch:arm64
pkg:github.com/elastic/apm-server/internal/publish goos:darwin goarch:arm64
Publisher-10                                                       1.21k ±14%     1.21k ±14%     ~     (p=1.000 n=5+5)
pkg:github.com/elastic/apm-server/x-pack/apm-server/aggregation/spanmetrics goos:darwin goarch:arm64
pkg:github.com/elastic/apm-server/x-pack/apm-server/aggregation/txmetrics goos:darwin goarch:arm64
pkg:github.com/elastic/apm-server/x-pack/apm-server/sampling goos:darwin goarch:arm64
Process-10                                                          467 ±120%       681 ±50%     ~     (p=0.548 n=5+5)
pkg:github.com/elastic/apm-server/x-pack/apm-server/sampling/eventstorage goos:darwin goarch:arm64
ReadEvents/json_codec/1000_events-10                               12.5k ± 0%     12.5k ± 0%     ~     (p=0.333 n=4+5)
```

I think, having a fixed grace period for shutdown should be okay for our case since it only comes into the picture while stopping the processor, additionally, during reload, we first start a new processor and then stop the older one. Let me know if my assumptions are not correct here, I think we can invest some time in having hooks to properly detect clean stop of the sampling processor rather than just waiting for a fixed amount of time.

axw, line 370:
```suggestion
			select {
			case <-ctx.Done():
				return ctx.Err()
			case <-time.After(shutdownGracePeriod):
			}
```

This would allow the processor to exit faster than the grace period, if the final flush completes faster.

x-pack/apm-server/sampling/processor.go
```diff
@@ -33,10 +33,11 @@ const (
 	// loggerRateLimit is the maximum frequency at which "too many groups" and
 	// "write failure" log messages are logged.
 	loggerRateLimit = time.Minute
-)
 
-// ErrStopped is returned when calling ProcessBatch on a stopped Processor.
-var ErrStopped = errors.New("processor is stopped")
+	// shutdownGracePeriod is the time that the processor has to gracefully
+	// terminate after the stop method is called.
+	shutdownGracePeriod = 500 * time.Millisecond
```
axw, line 39: This seems very low. If we don't wait for the full grace period every time, how would you feel about increasing this to something higher, like 5 seconds?

x-pack/apm-server/main.go
```diff
@@ -162,6 +169,19 @@ func getBadgerDB(storageDir string) (*badger.DB, error) {
 	return badgerDB, nil
 }
 
+func getStorage(db *badger.DB, ttl time.Duration, limit int64) *eventstorage.ShardedReadWriter {
+	storageMu.Lock()
+	defer storageMu.Unlock()
+	if storage == nil {
+		eventCodec := eventstorage.JSONCodec{}
+		// TTL and storage limit are not dynamically configurable
```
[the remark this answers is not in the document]
lahsivjar: > Can you remind me what the rationale is behind reusing the exact same Storage is?

Storage creates a pool of read writers (badger transactions). There were 2 issues with keeping `Storage` object per `Processor`:
1. Logic behind the beater reload is that it starts a new processor before shutting down the older one. This might lead to 2 processors running simultaneously and thus introducing race conditions for finalizing traces.
2. Badger uses [serializable snapshot isolation](https://dgraph.io/blog/post/badger-txn/) and our storage is sharded by badger transactions (each read writer is basically a new badger transaction). This introduces race between when a transaction is created. If a transaction for the same `traceID` is created before the previous `traceID` is committed, the new transaction will not see the commit.

    <details>
    <summary>An example test case to showcase point 2</summary>
    
    ```golang
    func TestBadgerSSI(t *testing.T) {
	    tempdir, err := os.MkdirTemp("", "samplingtest")
	    require.NoError(t, err)
    
	    opts := badger.DefaultOptions(tempdir)
	    badgerDB, err := badger.Open(opts)
	    require.NoError(t, err)
    
	    // NewProcessor creates a new badger transaction for each ReadWriter
	    tx1 := badgerDB.NewTransaction(true)
	    require.NoError(t, tx1.Set([]byte("testkey"), []byte("testval")))
	    require.NoError(t, tx1.Set([]byte("testkey:1"), []byte("testval")))
    
	    // NewProcessor creates a new badger transaction for each ReadWriter
	    tx2 := badgerDB.NewTransaction(true)
	    // ReadWriter#Flush commits the transaction on Processor#Stop
	    require.NoError(t, tx1.Commit())
	    require.NoError(t, tx2.Set([]byte("testkey:2"), []byte("testval")))
	    require.NoError(t, tx2.Set([]byte("testkey:3"), []byte("testval")))
    
	    iterOpts := badger.DefaultIteratorOptions
	    iterOpts.Prefix = []byte("testkey")
    
	    // `testkey` and `testkey:1` will not be visible to `tx2`
	    keysSet := map[string]bool{
		    "testkey:2": true,
		    "testkey:3": true,
	    }
    
	    // ReadWriter#ReadTraceEvents reads the trace events for a traceID to push to ES
	    //
	    // If we were to get the keys explicitly by calling GET for keys
	    // `testkey` or `testkey:1` the GET will fail with transaction conflict
	    // since badger SSI will track reads but iterating over keys will miss entries
	    iter := tx2.NewIterator(iterOpts)
	    for iter.Rewind(); iter.Valid(); iter.Next() {
		    item := iter.Item()
		    if item.IsDeletedOrExpired() {
			    continue
		    }
		    key := string(item.Key())
		    _, ok := keysSet[key]
    
		    require.True(t, ok, "unexpected key")
		    delete(keysSet, key)
	    }
	    iter.Close()
    
	    assert.Equal(t, 0, len(keysSet))
	    require.NoError(t, tx2.Commit())
    }
    ```
    </details>

> The alternative being for each processor to hold a mutex when it starts, then acquire storage; then flush/close storage when it stops, and release the mutex.

IIUC, this approach can take care of point 1 but point 2 will not get resolved only by this. Maybe we can employ lazy creation of read writers but I think that might introduce more complexity than the current approach.

For handling TTL and storage limit another option would be to handle them in main lifecycle by detecting changes on reload.

x-pack/apm-server/sampling/processor.go
lahsivjar: Sounds good. Thanks :D

x-pack/apm-server/main.go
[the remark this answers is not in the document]
axw: > IIUC, this approach can take care of point 1 but point 2 will not get resolved only by this. Maybe we can employ lazy creation of read writers but I think that might introduce more complexity than the current approach.

I think maybe I was unclear. Essentially what I'm suggesting is to not start the 2nd processor until the 1st one has exited, though I described it in a rather round about way. If we only ever have one TBS processor running at any time (i.e. by using a mutex), then we can guarantee that there is only one SharedReadWriter active, hence only one Badger transaction active for the same shard, and hence there should not be any write conflicts.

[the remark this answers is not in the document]
lahsivjar: Sorry, I am still not able to see the mutex solution with the second case. The issue with the second point comes due to the creation of a new `ShardedReadWriter` before the old one is flushed. CMIIW, a mutex can solve the case of 2 simultaneously active `ShardedReadWriter` (by putting mutexes in `Processor#Run`) but, as per the current reload logic, we will still create a new processor and thus a new `ShardedReadWriter` before we stop, and subsequently flush, the old storage. Are you suggesting somehow safeguarding NewProcessor creation also with mutexes?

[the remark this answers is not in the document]
axw: For posterity: Vishal and I discussed this live, and I had mistakenly been thinking that we lazily created badger transactions. So, what we can do instead is lazily create the Storage & SharedReadWriter in Processor.Run, and block in ProcessBatch until it has been created. We would hold a global mutex/semaphore in Processor.Run to prevent two processors from accessing storage at any one time.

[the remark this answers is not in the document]
lahsivjar: @axw I have attempted another approach to avoid implementing a special case for reloading configuration OR explicitly synchronizing processors.

Since TTL and storage limit are not tied directly to the underlying badger transaction, I have exposed them as arguments to global `ShardedReadWriter` (as well as the underlying ReadWriter(s)). The global `ShardedReadWriter` is wrapped in a processor-lifecycle-scoped ReadWriter which uses the global ReadWriter with the configuration as arguments. This allows us to keep badger transactions globally while allowing for the processor lifecycle to take care of configuration updates.

Let me know your thoughts on the approach.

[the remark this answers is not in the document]
axw: Just scanned over the PR quickly - this looks way better than the ideas I had. I'll give it a proper review early next week.

x-pack/apm-server/main.go
```diff
@@ -230,14 +246,33 @@ func closeBadger() error {
 	return nil
 }
 
+func closeStorage() (err error) {
+	if storage != nil {
+		// Force flush without limit consideration
+		err = storage.Flush(0)
```
axw, line 252: Do we need to flush here? I would expect it to be the responsibility of the processor to flush when it exits.

x-pack/apm-server/sampling/processor_test.go
```diff
@@ -60,18 +60,31 @@ func TestProcessAlreadyTailSampled(t *testing.T) {
...
+	wOpts := eventstorage.WriterOpts{
+		TTL:                 time.Minute,
+		StorageLimitInBytes: 0,
+	}
+	assert.NoError(t, writer.WriteTraceSampled(trace1.ID, true, wOpts))
+	assert.NoError(t, writer.Flush(wOpts.StorageLimitInBytes))
 	writer.Close()
 
-	storage = eventstorage.New(config.DB, eventstorage.JSONCodec{}, -1, 0) // expire immediately
+	wOpts.TTL = -1
```
axw, line 73:
```suggestion
	wOpts.TTL = -1 // expire immediately
```

x-pack/apm-server/sampling/processor_test.go
```diff
@@ -60,18 +60,31 @@ func TestProcessAlreadyTailSampled(t *testing.T) {
...
+	assert.NoError(t, writer.Flush(wOpts.StorageLimitInBytes))
 	writer.Close()
 
+	// Overwrite global storage to allow the transactions created by
+	// sharded read writers to view the explicitly sampled traces so far
+	eventCodec := eventstorage.JSONCodec{}
+	config.Storage = eventstorage.
+		New(config.DB, eventCodec).
+		NewShardedReadWriter()
+	t.Cleanup(func() { config.Storage.Close() })
```
axw, line 86: IIUC, this is necessary because `config.Storage` is created before the transactions are written, so they are not visible to the storage's transaction due to SSI. Could you add a bit more detail to the comment to help others (and future me :smile:) understand the rationale?

Maybe instead of replacing the storage, we could just flush it? That will cause new transactions to be created.

axw (commented):
Basically LGTM, just a few small questions.

x-pack/apm-server/sampling/processor_test.go
lahsivjar: > Could you add a bit more detail to the comment to help others (and future me 😄) understand the rationale?

Thanks, will help my future self to save some confusion too 😄 

> Maybe instead of replacing the storage, we could just flush it?

Yeah, this will work too and will be cleaner—I will update.

x-pack/apm-server/main.go
lahsivjar: Good point, while writing the code I decided that flush should be handled globally but with the introduction of wrapped read writers residing in the Processor lifecycle, it does make more sense to flush the processor on `Stop`. Let me make the required changes.

x-pack/apm-server/sampling/processor_test.go
```diff
@@ -563,6 +579,9 @@ func TestStorageGC(t *testing.T) {
 	require.NoError(t, err)
 	t.Cleanup(func() { badgerDB.Close() })
 	config.DB = badgerDB
+	config.Storage = eventstorage.
```
axw, line 582: Should this be closed on test cleanup?

axw (approved):
Looks great, thank you!

x-pack/apm-server/sampling/processor_test.go
lahsivjar: Thanks, I mistakenly removed it in my earlier commit. I have also moved the processor stop to `Cleanup` function for `BenchmarkProcess` so that it doesn't affect benchmark results (thanks @marclop for the tip).

x-pack/apm-server/sampling/processor.go
lahsivjar: Since the grace period was added in `Processor#Stop`, moving `Processor#Stop` to `Cleanup` removes the added time from benchmark results ([relevant commit](https://github.com/elastic/apm-server/pull/8809/commits/4e092dc5036e4c70c2967b329ecbeb0f5de2b2db))

marclop:
should we look into back porting this for `8.4` and hold merging until `8.4.0` is released?

lahsivjar:
> should we look into back porting this for `8.4` and hold merging until `8.4.0` is released?

SGTM, @axw WDYT? Should we backport to `8.4` targeting `8.4.1`?

axw:
Yes, SGTM too - let's hold off for a bit and backport to 8.4.1.

PR #8815, opened 2022-08-08
.ci: fix check-beats-version

axw:
Should be BRANCH_NAME, not BRANCH_VERSION.

### DIFF
--- a/.ci/scripts/check-beats-version.sh
+++ b/.ci/scripts/check-beats-version.sh
@@ -2,4 +2,4 @@
 
 set -xeu
 
-exec grep "BEATS_VERSION?=$BRANCH_VERSION" Makefile
+exec grep "BEATS_VERSION?=$BRANCH_NAME" Makefile


.ci/scripts/check-beats-version.sh
```diff
@@ -2,4 +2,4 @@
 
 set -xeu
 
-exec grep "BEATS_VERSION?=$BRANCH_VERSION" Makefile
+exec grep "BEATS_VERSION?=$BRANCH_NAME" Makefile
```
kruskall, line 5: **_question_**: should we fail early if BRANCH_NAME is empty ?

kruskall (approved):
LGTM!

.ci/scripts/check-beats-version.sh
axw: I don't think it'll matter in practice, let's come back and improve that if it needed.

PR #8839, opened 2022-08-10
txmetrics: Add global labels to metric sets

### LINKED ISSUE #8139: Global labels in transaction metrics
**Feature request**
Global labels should be available on transaction metrics.

**Background**
For 8.4 the APM UI team is planning to compute which services that belong in a service group at runtime (https://github.com/elastic/kibana/issues/128623). 

When a user creates a service group they are asked to provide a kql string that matches the services they want to include.
This kql string is then used as a filter when retrieving the services for a selected service group. 
The fields in the kql string must be limited to dimensions available in transaction metrics, to avoid falling back to transaction events at runtime.

Customers have requested the ability to create service groups based on business-specific global labels like `department_name` and `department_owner_email`.

marclop:
## Motivation/summary

Adds global labels to the aggregated metric (as parsed in the `metadata`
event) dimensions that are published by the APM Server.

## Checklist

## How to test these changes

1. docker-compose up -d
2. cd systemtest && go run cmd/runapm/main.go -f
3. Run the [following program](https://gist.github.com/marclop/a976a2b78b1a90dd8d87b2c31faf67fe)
4. Ensure that the indexed metric documents contain the global labels
```
GET /metrics-apm.internal-default/_search
{
  "_source": false, 
  "fields": ["labels.*"], 
  "query": {"match_all": {}}
}
```
```
  "hits": {
    "total": {
      "value": 3,
      "relation": "eq"
    },
    "max_score": 1,
    "hits": [
      {
        "_index": ".ds-metrics-apm.internal-default-2022.08.10-000001",
        "_id": "4eDLhYIBxUgzjit0xKMG",
        "_score": 1,
        "_source": {},
        "fields": {
          "labels.department_name": [
            "apm"
          ],
          "labels.company": [
            "elastic"
          ],
          "labels.organization": [
            "observability"
          ]
        }
      },
      {
        "_index": ".ds-metrics-apm.internal-default-2022.08.10-000001",
        "_id": "4uDLhYIBxUgzjit0xKMG",
        "_score": 1,
        "_source": {},
        "fields": {
          "labels.department_name": [
            "fleet-server"
          ],
          "labels.company": [
            "elastic"
          ],
          "labels.organization": [
            "observability"
          ]
        }
      },
      {
        "_index": ".ds-metrics-apm.internal-default-2022.08.10-000001",
        "_id": "4-DLhYIBxUgzjit0xKMG",
        "_score": 1,
        "_source": {},
        "fields": {
          "labels.department_name": [
            "kibana"
          ],
          "labels.company": [
            "elastic"
          ],
          "labels.organization": [
            "platform"
          ]
        }
      }
    ]
  }
}
```
5. Enable Service groups in kibana ([link](https://www.elastic.co/guide/en/kibana/8.3/services.html#service-groups))
6. Create a 3 service groups:
   1. Observability: Matches `labels.organization: observability`
   2. APM: Matches `labels.department_name: apm`
   3. Platform: Matches `labels.organization: platform`

7. Verify that Observability has two services (`apm`, `fleet-server`)
8. Verify that APM has 1 service (`apm`)
9. Verify that Platform has 1 service (`kibana`)

## Related issues

Closes #8139


internal/model/labels.go
```diff
@@ -107,6 +156,53 @@ func (l NumericLabels) Clone() NumericLabels {
...
+	}
+
+	var strbld strings.Builder
+	for key, value := range l {
+		strbld.WriteString(key)
+		if value.Value > 0 {
+			strbld.WriteString(strconv.FormatFloat(value.Value, 'e', -1, 64))
+			continue
+		}
+		for _, v := range value.Values {
```
lahsivjar, line 199: Same as for `Labels`

internal/model/labels.go
```diff
@@ -59,6 +61,53 @@ func (l Labels) Clone() Labels {
...
+	}
+
+	var strbld strings.Builder
+	for key, value := range l {
+		strbld.WriteString(key)
+		if value.Value != "" {
+			strbld.WriteString(value.Value)
+			continue
+		}
+		for _, v := range value.Values {
```
lahsivjar, line 104: > in cases where both are set, the `Values` field will be used and `Value` will be ignored.

As per the struct docs, `Values` is preferred over `Value`, shouldn't `Flatten` keep the same priority?

internal/model/labels.go
```diff
@@ -59,6 +61,53 @@ func (l Labels) Clone() Labels {
...
 }
 
+// Equal returns true when the two maps of labels are equal.
+func (l Labels) Equal(labels Labels) bool {
+	for key, localV := range l {
+		v, ok := labels[key]
+		if !ok {
+			return false
+		}
+		if v.Value != localV.Value {
```
lahsivjar, line 71: > in cases where both are set, the Values field will be used and Value will be ignored.

The `Equal` will not follow this priority order. From the description it sounds like if `Values` is set then `Value` field doesn't matter.

internal/model/labels.go
```diff
@@ -107,6 +156,53 @@ func (l NumericLabels) Clone() NumericLabels {
...
 }
 
+// Equal returns true when the two maps of labels are equal.
+func (l NumericLabels) Equal(labels NumericLabels) bool {
+	for key, localV := range l {
+		v, ok := labels[key]
+		if !ok {
+			return false
+		}
+		if v.Value != localV.Value {
```
lahsivjar, line 166: Same as for `Labels`

internal/model/apmevent.go
```diff
@@ -72,6 +72,14 @@ type APMEvent struct {
 	// Supports slice values.
 	NumericLabels NumericLabels
 
+	// GlobalLabels are only populated when the batch metadata contains labels
+	// and the decoded event is a transaction event. This is done to be able to
+	// differentiate between event and global labels since we're generating
+	// transaction metrics which include global labels as dimensions to power
+	// certain parts of the APM UI. Not indexed.
+	GlobalLabels        Labels
+	GlobalNumericLabels NumericLabels
```
axw, line 81: I was thinking that instead of introducing `Global*` counterparts, we could add a `Global bool` field on the `model.LabelValue` and `model.NumericLabelValue` types. Then when merging global/non-global labels, we would set the Global flag to false if the event-specific label's value is not the same as the global label's value. WDYT?

internal/model/labels.go
```diff
@@ -59,6 +61,53 @@ func (l Labels) Clone() Labels {
 	return cp
 }
 
+// Equal returns true when the two maps of labels are equal.
+func (l Labels) Equal(labels Labels) bool {
+	for key, localV := range l {
```
axw, line 66: We should also check `len(l) == len(labels)`, otherwise one may be a superset of the other.

internal/model/labels.go
```diff
@@ -59,6 +61,53 @@ func (l Labels) Clone() Labels {
...
+			return false
+		}
+		if len(v.Values) != len(localV.Values) {
+			return false
+		}
+		for _, s := range [][]string{v.Values, localV.Values} {
+			if !sort.StringsAreSorted(s) {
+				sort.Strings(s)
+			}
+		}
```
axw, line 81: Seems dangerous and wrong to me for an `Equal` method to mutate the value. Do we need to sort them?

internal/model/labels.go
```diff
@@ -59,6 +61,53 @@ func (l Labels) Clone() Labels {
...
+			if v != localV.Values[i] {
+				return false
+			}
+		}
+	}
+	return true
+}
+
+// Flatten returns a string interpretation of the labels.
+func (l Labels) Flatten() string {
```
axw, line 92: Until we find another need for flattening labels, can we keep this in the txmetrics code?
Also, it would be better not to build a string only to write it to a hash, as it will create garbage: instead pass in an io.Writer.

internal/model/apmevent.go
marclop: that is much cleaner than what I did, thank you for the suggestion!

x-pack/apm-server/aggregation/txmetrics/aggregator.go
```diff
@@ -594,3 +644,35 @@ func transactionCount(tx *model.Transaction) float64 {
 	}
 	return 1
 }
+
+func writeLabels(w io.StringWriter, l model.Labels) {
+	if len(l) == 0 {
+		return
+	}
+	for key, value := range l {
```
axw, line 652: Labels is a map, so iteration will be non-deterministic. Should we iterate over the sorted keys?

Perhaps when copying the global labels in makeTransactionAggregationKey, create a slice and sort that?

x-pack/apm-server/aggregation/txmetrics/aggregator.go
```diff
@@ -594,3 +644,35 @@ func transactionCount(tx *model.Transaction) float64 {
...
+}
+
+func writeNumericLabels(w io.StringWriter, l model.NumericLabels) {
+	if len(l) == 0 {
+		return
+	}
+	for key, value := range l {
+		w.WriteString(key)
+		if value.Value != 0 {
+			w.WriteString(strconv.FormatFloat(value.Value, 'e', -1, 64))
```
axw, line 671: Can we use encoding/binary.LittleEndian.PutUint64 & math.Float64bits? Should be faster and create less garbage I think?

internal/model/labels.go
```diff
@@ -59,6 +61,39 @@ func (l Labels) Clone() Labels {
...
+	if len(l) != len(labels) {
+		return false
+	}
+	for key, localV := range l {
+		v, ok := labels[key]
+		if !ok {
+			return false
+		}
+		// If the slice value is set, ignore the Value field.
+		if len(v.Values) == 0 && v.Value != localV.Value {
```
lahsivjar, line 75: Should we also check for `v.Global != localV.Global`? From the API, it seems possible for such a case to occur.

internal/model/labels.go
```diff
@@ -59,6 +61,39 @@ func (l Labels) Clone() Labels {
 	return cp
 }
 
+// Equal returns true when the two maps of labels are equal.
```
axw, line 64: I wonder if this is premature generalisation too. If we were to copy labels to a sorted slice when constructing the aggregation key, then we could do a more efficient comparison (i.e. by stepping through the slices in order). Also, by having the Equal method here, I would expect the Global field to be compared too - it's not relevant for the aggregation as we only compare global labels.

internal/model/modeldecoder/modeldecoderutil/labels.go
```diff
@@ -26,11 +26,19 @@ import (
 	"github.com/elastic/apm-server/internal/model"
 )
 
-// LabelsFrom populates the Labels and NumericLabels.
+// LabelsFrom populates the Labels and NumericLabels for the metadata object.
 func LabelsFrom(from mapstr.M, to *model.APMEvent) {
```
axw, line 30:
```suggestion
// GlobalLabelsFrom populates the Labels and NumericLabels from global labels in the metadata object.
func GlobalLabelsFrom(from mapstr.M, to *model.APMEvent) {
```

internal/model/labels.go
```diff
@@ -83,6 +118,8 @@ type NumericLabelValue struct {
 	Values []float64
 	// Value holds the label `float64` value.
 	Value float64
+	// Global is `true` when the label has been decoded from the metadata.
```
axw, line 121:
```suggestion
	// Global is `true` when the label is defined at the agent level, rather than being event-specific.
	// Global labels may be included as dimensions in aggregations.
```

internal/model/labels.go
```diff
@@ -83,6 +118,8 @@ type NumericLabelValue struct {
...
+	Global bool
 }
 
 func (l NumericLabels) Set(k string, v float64) {
```
axw, line 125:
```suggestion
// Set sets the label k to value v. If there existed a label in l with the same key,
// it will be replaced and its Global field will be set to false.
func (l NumericLabels) Set(k string, v float64) {
```

"Metadata" is kind of specific to the Elastic APM protocol, so I'd like to avoid using that term in the model, which may come from OpenTelemetry.

axw (commented):
Should we also consider labels set from OpenTelemetry resource attributes to be global?

internal/model/labels.go
marclop: I've moved the equality comparison to a private function inside the transaction aggregator where all the compared labels are global.

internal/model/labels.go
```diff
@@ -83,12 +90,18 @@ type NumericLabelValue struct {
 	Values []float64
 	// Value holds the label `float64` value.
 	Value float64
+	// Global is `true` when the label has been decoded from the metadata.
```
axw, line 93:
```suggestion
	// Global is `true` when the label is defined at the agent level, rather
	// than being event-specific.
```

x-pack/apm-server/aggregation/txmetrics/aggregator.go
```diff
@@ -594,3 +649,103 @@ func transactionCount(tx *model.Transaction) float64 {
 	}
 	return 1
 }
+
+func writeLabels(w io.Writer, aggKey *transactionAggregationKey) {
+	if len(aggKey.labelKeys) == 0 && len(aggKey.numericLabels) == 0 {
+		return
+	}
```
axw, line 656:
```suggestion
```

x-pack/apm-server/aggregation/txmetrics/aggregator.go
```diff
@@ -594,3 +649,103 @@ func transactionCount(tx *model.Transaction) float64 {
...
+			for _, localValue := range localV.Values {
+				if value == localValue {
+					found = true
+					break
+				}
+			}
+			if !found {
+				return false
+			}
+		}
```
axw, line 714:
```suggestion
		for i, value := range v.Values {
			if localV.Values[i] != value {
				return false
			}
		}
```

Do we need to treat array values as sets for comparison? I think we should consider them to be equal if they are exactly the same order.

x-pack/apm-server/aggregation/txmetrics/aggregator.go
```diff
@@ -594,3 +649,103 @@ func transactionCount(tx *model.Transaction) float64 {
...
+			return false
+		}
+		// If the slice value is set, ignore the Value field.
+		if len(v.Values) == 0 && v.Value != localV.Value {
+			return false
+		}
+		if len(v.Values) != len(localV.Values) {
+			return false
+		}
+		for _, value := range v.Values {
```
axw, line 737: same as above

axw (commented):
Mostly LGTM, just a couple of small things.

x-pack/apm-server/aggregation/txmetrics/aggregator.go
marclop: In that case slices that contain the exact same items but are out of order won't be equal:

```go
{"a": {Values: []string{"b", "a"}}}
```
```go
{"a": {Values: []string{"a", "b"}}}
```

Is that what we want? That seems counter intuitive to me.

axw: I haven't seen this explicitly specified anywhere, but as far as I've seen OTel will always maintain ordering of array-type attribute values. I don't think we should start treating them as sets, as this will lead to surprises if some OTel instrumentation is relying on the ordering being maintained. In practice I doubt it will matter.

The other reason is to make the code simpler and faster :)

marclop: I see, thanks for the extra information. I'll change the code se we handle the order as intended.

axw (approved):
Looks great, thank you :)

axw:
Verified with 8.5.0-BC1.

PR #8840, opened 2022-08-10
Tidy up building and installing integration package

axw:
## Motivation/summary

Tidy up the Makefile targets for building the APM integration package:

- Separate building integration packages from linting
- Introduce separate build directory for release/snapshot integration packages, so they don't clobber each other.
- Use the git commit timestamp in the snapshot package build tag, for reproducibility.
- Because building the snapshot/preview integration package no longer clobbers the release package, I have updated the CI job to only sign/publish preview package zips.

We now install the integration package in system tests by uploading directly to Kibana, rather than injecting into package-registry, which is undergoing changes. This approach also ensures the system tests have an up-to-date package, rather than requiring developers to first run `make build-package` and restart the package-registry service.

To keep a simple method of installing the local APM integration package in Kibana outside of the system tests, one can use `systemtest/cmd/runapm` with the new `-init` flag. This will just initialise Fleet, install the integration package, and exit.

## Checklist

## How to test these changes

N/A

## Related issues

None


docker-compose.yml
```diff
@@ -88,18 +77,6 @@ services:
       - "./testing/docker/fleet-server/certificate.pem:/etc/pki/tls/certs/fleet-server.pem"
       - "./testing/docker/fleet-server/key.pem:/etc/pki/tls/private/fleet-server-key.pem"
 
-  package-registry:
```
marclop, line 91: 🎉

marclop (approved):
Great improvement! I'm not totally familiar with how the packages are published right now on CI, but the changes in that file look sensible to me.

Issue #8775, opened 2022-08-02
Run benchmarks on scaling APM Sever on ESS beyond 8GB


simitt:
Verify that the APM Server performance does not degrade when increasing the maximum allowed container size on ESS beyond the current limits of 8GB per instance. 

We should leverage the recently reworked apm benchmark tooling for conducting such tests efficiently. 
@mieciu has offered to provide a dedicated testing envionment on ESS for this and can provide more details via PM.

axw:
We should also compare scaling up vs. scaling out. We have a default limit of 200 concurrent request decoders, which could be a problem if we were to start scaling out at a higher memory threshold.

marclop:
We've concluded the scheduled work on this topic for the `8.5` milestone. We'll look at introducing some changes to the  model indexer in `8.6`, so we can maximize resource usage on bigger instances.

PR #8855, opened 2022-08-12
apmbench: Replace -warmup-events for -warmup-time

marclop:
## Motivation/summary

Removes the `-warmup-events` flag and the complicated logic required to
handle timeouts it in favor of a new flag `-warmup-time` which will send
events for the specified duration. This strategy falls more in line with
the intended usage for a warmup phase, where a user wants to warm up the
target server for an amount of time using the specified agents and does
not care much about how many events are sent for that warmup phase.

Also re-formats some unformatted tf files, and adds a new make benchmark
target which auto-tunes the number of agents from the APM Server size.

Last, it changes apmbench to not return an error when the warm-up phase
returns an error. Warm-ups are best-effort and shouldn't impede the
benchmarks from running.

## Checklist

## How to test these changes

Run `apmbench` with `-warmup-time` and see if it warms up the APM Server for the set time.

## Related issues

Part of #8775


systemtest/benchtest/main_test.go
```diff
@@ -37,166 +37,56 @@ import (
...
-					}
-
-					if !strings.HasPrefix(r.URL.Path, "/intake") {
+		t.Run(fmt.Sprintf("%d_agent_%s_events", c.agents, c.duration.String()), func(t *testing.T) {
+			var received uint64
+			srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				if r.URL.Path == "/debug/vars" {
+					// Report idle APM Server.
+					w.Write([]byte(`{"libbeat.output.events.active":0}`))
+				}
```
kruskall, line 54: **issue**: we should return early here. if `w.Write` is called the server has already sent a response to the client

systemtest/benchtest/main.go
```diff
@@ -265,6 +256,9 @@ func warmup(agents int, events uint, url, token string) error {
 	var merr multierror.Errors
 	for err := range errs {
 		if err != nil && !errors.Is(err, context.DeadlineExceeded) {
+			if e, ok := err.(timeoutError); ok && e.Timeout() {
```
kruskall, line 259: **suggestion:** I think we can use `errors.As` and `net.Error` to check if `err` is a network error and then call `Timeout()` on it.
We should also add a TODO to remove this once we update to Go 1.19 as the workaround is no longer necessary: https://go.dev/doc/go1.19#net

systemtest/benchtest/main_test.go
```diff
@@ -37,166 +37,56 @@ import (
...
-		t.Run(tt.name, func(t *testing.T) {
-			got := warmupTimeout(
-				tt.args.ingestRate,
-				tt.args.events,
-				tt.args.epm,
-				tt.args.agents,
-				tt.args.cpus,
-			)
-			assert.Equal(t, tt.expected, got)
+				}
```
kruskall, line 82: **issue:** `scanner.Err()` should be checked before returning status accepted

kruskall (changes requested):

marclop:
@kruskall thanks for the suggestions. Addressed them in the last commit.

kruskall (approved):
LGTM!

PR #8859, opened 2022-08-14
Remove all uses of Python apart from Rally

axw:
## Motivation/summary

Remove all uses of Python apart from Rally, to simplify and speed up our build, and reduce the amount of custom code we are maintaining. Because only Rally requires Python now, we remove `script/requirements.txt`, and the `autopep8` and `check-autopep8` Makefile targets. CI was relying on docker-compose being upgraded by the version specified in `script/requirements.txt`; this was a CI-only thing, so I've moved this to the relevant CI script.

`script/copy_package.py` is no longer needed, as snapshot packages are automatically published to storage. Non-snapshot packages do not need any modifications, and can be copied verbatim.

`script/copy-docs.py` is replaced with a simple `rsync` command. `make update-beats-docs` now copies from the beats module specified in go.mod.

`script/generate_notice.py` is replaced with a simpler "go list" command that is directly piped to `go-licence-detector`. I have copied the beats NOTICE.txt/dependencies.csv templates, rules, and overrides, with our modifications made directly in the templates. The order of dependencies in NOTICE.txt has changed, hence the file has been regenerated.

`script/latest_snapshot_version.py` is replaced with a simple shell script using `curl` and `jq`.

## Checklist

## How to test these changes

N/A

## Related issues

None


script/latest_snapshot_version.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
+#
+# Find the latest snapshot version for the specified branch.
+#
+# Example usage: ./latest_snapshot_version.py 7.x
+
+BRANCH=$1
```
kruskall, line 7: **suggestion**: should we guard against `BRANCH` being empty and return an error ?

script/latest_snapshot_version.sh
```diff
@@ -0,0 +1,14 @@
...
+
+# The snapshots API has not yet been updated to use "main".
+if [ "$BRANCH" = "main" ]; then
+  BRANCH=master
+fi
+
+curl --silent "https://snapshots.elastic.co/latest/$BRANCH.json" | jq -r .version
```
kruskall, line 14: **issue**: curl is suppressing errors. I think we should only ignore stdout and let stderr print the errors (`jq` will fail anyway but at least we would have the error message)

Something like `curl --silent --show-error` should be okay

kruskall, line 14: **issue**: `jq` doesn't return a proper exit code. We should use `-e` to set the exit code and try to handle empty input or missing key.

To reproduce:

```
~ % echo "" | jq -r .version
~ % echo $?
0
```

Fix attempt using `-e -n`:

```
~ % echo "" | jq -e -n -r 'input.version'
jq: error (at <stdin>:1): break
~ % echo $?
1
```

kruskall (changes requested):
Great change! Just a minor issue in the shell script

axw:
Thanks @kruskall, good catches! Updated with your suggestions.

kruskall (approved):
Thanks, LGTM!

PR #8867, opened 2022-08-15
testing/benchmark: Allow setting apm index shards

marclop:
## Motivation/summary

Introduces a new variable `apm_shards` which allows tunning the APM
component templates `number_of_shards` setting, with no default.

This allows us to benchmark the APM Server with a more suitable config
as the APM Server size increases, and also test different permutations
of config options.

Also defaults the benchmarking to always use the `integrations_server`
resource and tunes the 16GB APM Server profile.

## Checklist

## How to test these changes

Set the shard setting to something different than 1 and check that the component templates are updated.

## Related issues

Part of #8775


testing/benchmark/README.md
```diff
@@ -90,3 +90,9 @@ to have use the committed tags in the `docker-compose.yml` file: `eval $(make do
 It is also possible to override the docker image to one that is allowed to run in ESS. For more information
 on which repositories can be used, please refer to our internal docs. To override the docker image, you'll need
 to specify the full object of images that is defined in `variables.tf`: `docker_image_override`.
+
+### Set APM index shards
+
+By default, the APM indices shiop with `number_of_shards` set to `1`. To override this behavior, you can modify the
```
kruskall, line 96:
```suggestion
By default, the APM indices ship with `number_of_shards` set to `1`. To override this behavior, you can modify the
```

kruskall (changes requested):
LGTM! Justa minor typo

testing/benchmark/variables.tf
```diff
@@ -68,6 +68,19 @@ variable "docker_image_override" {
...
+    "traces-apm@custom" : 1,
+    "logs-apm.app@custom" : 1,
+    "logs-apm.error@custom" : 1,
+    "metrics-apm.app@custom" : 1,
+    "metrics-apm.internal@custom" : 1,
+    "traces-apm.rum@custom" : 1,
+    "traces-apm@custom" : 1,
+  }
+  type = map(string)
+}
```
axw, line 82: Could we not set anything at all by default? I would prefer if the default came from the integration package, so we don't inadvertently test with something other than the actual defaults. i.e. if we were to start setting number_of_shards in the package, then our benchmarks should pick that up.

marclop: yep, that sounds better. I'll remove the default.

axw (approved):
LGTM, thanks for the update!

Issue #5682, opened 2021-07-12
Metrics without any explicit mapping can lead to document rejection


lucabelluccini:
Some APM clients support metrics which are unknown by APM Server (e.g. `micrometer` for the APM Java Agent).
APM doesn't know the field names nor their type.

It can happen some of the fields, due to the fact there's no default mapping, might be mapped as `float` instead of `long` (or the opposite), leading to document rejections depending on the first document to be indexed after a rollover or index creation.

Some examples:
```
hikaricp_connections_acquire.count
jvm_memory_max
jvm_buffer_total_capacity
jvm.gc.time
```

It is possible to observe this behavior especially when Logstash is in the middle as we have a 400 errors in the logs.

The problem is not easy to tackle, except defaulting to `float` instead of `long` for `integer` values (e.g. using a [dynamic template](https://www.elastic.co/guide/en/elasticsearch/reference/current/dynamic-field-mapping.html))?

simitt:
@lucabelluccini with the new data streams scheme, service specific metrics are sent to service specific data streams, and only common metrics share a data stream. I believe this issue can be closed now as a sideeffect of the new indexing strategy. @lucabelluccini do you agree?

lucabelluccini:
Hello @simitt
The Data Streams for sure help, avoiding mapping conflicts if the same metric was used by different services and they used different types.

Still, the problem could occur even when the metric is reported by the same agent/service.

The most common issue can happen for representation of `float` numbers in Golang JSON serialization.
See this golang playground: https://play.golang.com/p/tPYQZFGLajT

_Example_

If `jvm.gc.time` is `1.0`, it will be serialized as `1` by APM-Server.
Elasticsearch will default to `long`.
If the next `jvm.gc.time` is `0.9`, it will be rejected.

axw:
I think we should probably map all simple numeric metrics as `double`, to avoid this issue.

PR #8868, opened 2022-08-15
fix: prevent floats without decimal values from being encoded as int

kruskall:
## Motivation/summary

Floats are encoded as ints and trailing zeroes are being stripped,
causing document rejection due to mismatched types.
APM server is relying on fastjson, which is using strconv, to encode
metrics.
The precision is set to -1 so that the encoded string uses the
smallest number of digits. To avoid losing type information, truncate
the value by casting to int and compare the result to the original
value: if they are equal conversion is carried out with precision=1
so that the trailing zero is not stripped.

## Checklist

## Related issues

Closes https://github.com/elastic/apm-server/issues/5682


internal/model/modelindexer/indexer.go
```diff
@@ -308,6 +309,17 @@ func encodeAny(v interface{}, out *fastjson.Writer) error {
...
 	case map[string]interface{}:
 		return encodeMap(v, out)
+	case float64:
+		// truncate the value with int conversion to
+		// check for decimals.
+		if float64(v) == float64(int64(v)) {
+			// we cannot rely on fastjson.Marshal as it would strip
+			// the leading zero and encode the value as an "int".
+			// So we call AppendFloat directly with precision=1.
+			out.RawBytes(strconv.AppendFloat(nil, float64(v), 'f', 1, 64))
```
marclop, line 319: nit: There's no need to cast `v` to `float64` since the switch/case clause already does that for you.
```suggestion
		if v == float64(int64(v)) {
			// we cannot rely on fastjson.Marshal as it would strip
			// the leading zero and encode the value as an "int".
			// So we call AppendFloat directly with precision=1.
			out.RawBytes(strconv.AppendFloat(nil, v, 'f', 1, 64))
```

marclop (approved):
LGTM. It looks like the processing performance has improved as well (>-50% improvement)?

May be worth back porting this to the `8.4.` branch and hold merging until `8.4.0` is out, so we can include the fix in `8.4.1`.

axw (commented):
I don't think this is quite enough to close https://github.com/elastic/apm-server/issues/5682.

Although it's much less likely, there's still the possibility that a double value can trigger a mapping exception when relying on dynamic mapping, as double values are dynamically mapped to the "float" field type: https://www.elastic.co/guide/en/elasticsearch/reference/current/dynamic-field-mapping.html

Consider this:

```
PUT /metrics
{
  "mappings": {
    "properties": {
      "float": {"type": "float"},
      "double": {"type": "double"}
    }
  }
}

# Fails
POST /metrics/_doc
{
  "float": 3.40282346638528859811704183484516925440e+39
}

# Fails
POST /metrics/_doc
{
  "dynamic": 3.40282346638528859811704183484516925440e+39
}

# Succeeds
POST /metrics/_doc
{
  "double": 3.40282346638528859811704183484516925440e+39
}
```

This is why I think we should explicitly map simple metric values to `double`. To do that, we would add another named dynamic template to the `app_metrics` data stream.

See https://github.com/elastic/apm-server/blob/1efe50c7c098b7aae754b2d4922b8aff03141a31/apmpackage/apm/data_stream/app_metrics/manifest.yml#L15-L25 and https://github.com/elastic/apm-server/blob/1efe50c7c098b7aae754b2d4922b8aff03141a31/apmpackage/apm/data_stream/app_metrics/elasticsearch/ingest_pipeline/default.yml#L22-L26

internal/model/modelindexer/indexer.go
kruskall: Good catch!

kruskall:
> I don't think this is quite enough to close https://github.com/elastic/apm-server/issues/5682.
> 
> Although it's much less likely, there's still the possibility that a double value can trigger a mapping exception when relying on dynamic mapping, as double values are dynamically mapped to the "float" field type: https://www.elastic.co/guide/en/elasticsearch/reference/current/dynamic-field-mapping.html

Ah, I see. Good point!

I've updated the code and added the dynamic template

apmpackage/apm/data_stream/app_metrics/manifest.yml
```diff
@@ -23,6 +23,10 @@ elasticsearch:
                 - sum
                 - value_count
               default_metric: value_count
+        - doubles:
+            match_mapping_type: float
```
axw, line 27:
```suggestion
```

We should always explicitly specify the dynamic template by name, so it shouldn't be necessary to match on type.

apmpackage/apm/data_stream/app_metrics/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -23,6 +23,8 @@ processors:
             dynamic_templates[name] = "histogram";
           } else if (metric_type == "summary") {
             dynamic_templates[name] = "summary";
+          } else if (metric_type == "counter" || metric_type == "gauge") {
```
axw, line 26:
```suggestion
          } else {
```

apmpackage/apm/data_stream/app_metrics/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -23,6 +23,8 @@ processors:
...
+            dynamic_templates[name] = "doubles";
```
axw, line 27:
```suggestion
            dynamic_templates[name] = "double";
```

Singular, for consistency with "summary" and "histogram".

apmpackage/apm/data_stream/app_metrics/manifest.yml
```diff
@@ -23,6 +23,10 @@ elasticsearch:
                 - sum
                 - value_count
               default_metric: value_count
+        - doubles:
```
axw, line 26:
```suggestion
        - double:
```

internal/model/modelindexer/indexer.go
```diff
@@ -308,6 +309,17 @@ func encodeAny(v interface{}, out *fastjson.Writer) error {
...
+		// truncate the value with int conversion to
+		// check for decimals.
+		if v == float64(int64(v)) {
+			// we cannot rely on fastjson.Marshal as it would strip
+			// the leading zero and encode the value as an "int".
+			// So we call AppendFloat directly with precision=1.
+			out.RawBytes(strconv.AppendFloat(nil, v, 'f', 1, 64))
+			return nil
+		}
+		return fastjson.Marshal(out, v)
```
axw, line 322:
```suggestion
```

This should not be necessary. We'll map it as a double even if it's encoded as an integer, by specifying the dynamic template by name.

systemtest/metrics_test.go
```diff
@@ -170,7 +170,7 @@ func TestApplicationMetrics(t *testing.T) {
 			"full_name": "a.b.c",
 			"mapping": map[string]interface{}{
 				"c": map[string]interface{}{
-					"type": "long",
+					"type": "float",
```
axw, line 173: This should be `double` now, right?

axw (approved):
LGTM, thanks :)

I think this one deserves a changelog entry.

internal/model/modelindexer/indexer.go
kruskall: The system test is failing without this line. Am I missing something ?

axw: I overlooked something: I was thinking that all metrics would have an entry in the "_metric_descriptions" map which is used in the ingest processor for specifying dynamic templates. This isn't the case: we only add entries for metrics with either Type or Unit set. So we'll need some more changes:

- Always send a metric description, even if the type and unit are empty. This would enable us to identify metric fields in the ingest pipeline, and assign them the correct dynamic template. The changes would need to be made in https://github.com/elastic/apm-server/blob/a1d1cfb3e07dbf5e6ab95e5f1b7a95faddfbd374/internal/model/metricset.go#L163-L172
- Update the `internal_metrics` data stream's ingest pipeline to discard metric descriptions, since in that data stream we do not use dynamic mapping at all.

I'm open to alternative ideas.

kruskall: Sorry for the long wait 🙇 

I learned quite a lot while working on this. I've pushed some changes!

apmpackage/apm/data_stream/internal_metrics/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -9,3 +9,7 @@ processors:
       name: process_ppid
   - pipeline:
       name: client_geoip
+  - script:
+      if: ctx._metric_descriptions != null
+      source: |
+        ctx.remove("_metric_descriptions");
```
axw, line 15:
```suggestion
  - remove:
      field: _metric_descriptions
      ignore_missing: true
```

Should be a bit more efficient.

internal/model/metricset.go
```diff
@@ -165,9 +165,9 @@ func (me *Metricset) setFields(fields *mapStr) {
 		sample.set(name, fields)
 
 		var md mapStr
-		md.maybeSetString("type", string(sample.Type))
-		md.maybeSetString("unit", sample.Unit)
-		metricDescriptions.maybeSetMapStr(name, mapstr.M(md))
+		md.set("type", string(sample.Type))
+		md.set("unit", sample.Unit)
```
axw, line 169:
```suggestion
		md.maybeSetString("type", string(sample.Type))
		md.maybeSetString("unit", sample.Unit)
```

Can we always send the map, but only send type or unit if non-empty? This would cut down the payload sizes a little.

kruskall: Good point! Changed!

axw (approved):
Looks great, thanks :)

Issue #8754, opened 2022-07-29
[MVP] Tooling for benchmarking Elasticsearch ingest changes


simitt:
## Problem
We would like to increase our visibility into the effects of changing Elasticsearch index mappings, and ingest pipelines. For example, what would be the change in indexing time and Elasticsearch storage size if we were to [start mapping SQL statements with the `text` field type](https://github.com/elastic/apm-server/issues/6586)?

By introducing tooling to analyse these kinds of changes, we can make changes more quickly, and with confidence.

## Expected Outcome
Build an MVP so that developers can run benchmarks on-demand, from their laptop, to analyse the impact of integration packages on Elasticsearch performance and storage cost. The output should be a Rally [summary report](https://esrally.readthedocs.io/en/stable/summary_report.html).

* [x] Ability to inject a custom apm integration package into a Kibana image
* [x] Spin up ESS deployment of a specific version, allowing to override the Kibana image to use the one with the custom apm integration package.
* [x] Generate the ES corpus from APM agent data sent to an APM Server of a specific version (locally or directly on worker VM). 
* [x] Run `rally` on local to benchmark ES cluster using the ES corpus from agent data.
* [ ] Run `rally` on the worker VM to [benchmark the Elastic Cloud cluster](https://esrally.readthedocs.io/en/stable/recipes.html#benchmarking-an-elastic-cloud-cluster) using the ES corpus from agent data.

## Outside of scope
- benchmarking query performance
- indexing results in Elasticsearch
- no CI integration for running benchmarks regularly

## Relates to
https://github.com/elastic/apm-server/issues/6115

simitt:
> Ability to inject a custom apm integration package into a Kibana image

https://github.com/elastic/kibana/issues/137759 seems relevant for this

lahsivjar:
#8856 Introduces support for injecting a custom APM Package into kibana image as well as ability to spin-up ESS deployment with a custom integrations package.

PR #8878, opened 2022-08-16
Add gencorpora cmd to generate ES corpora for rally

lahsivjar:
## Motivation/summary

Add command for generating ES corpus based on APM agent data sent to a test APM-Server.

## Checklist

## How to test these changes

```
$ cd systemtest/cmd/gencorpora
$ go run main.go
```
Examine stdout for ES corpus.

## Related issues

Part of #8754


marclop (commented):
couple of small comments.

systemtest/cmd/gencorpora/main.go
```diff
@@ -0,0 +1,110 @@
...
+		if err != nil {
+			log.Fatal(err)
+		}
+	}()
+	if err := apmServer.Start(); err != nil {
+		log.Fatal(err)
+	}
+
+	// Wait for APM Server to be ready to accept requests
+	<-time.After(2 * time.Second)
```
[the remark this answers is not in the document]
lahsivjar: Thanks for the tip, I didn't know we had something like this. BTW, still working on a few things before marking it ready for review.

systemtest/gencorpora/apmserver.go
```diff
@@ -0,0 +1,164 @@
...
+	"sync"
+	"time"
+
+	"github.com/elastic/apm-server/systemtest/apmservertest"
+)
+
+// APMServer represents a wrapped exec CMD for APMServer process
+type APMServer struct {
+	apmHost string
+	cmd     *apmservertest.ServerCmd
```
axw, line 37: Is there a reason why you're not using apmservertest.Server?

axw (commented):
Only scanned over so far, but what I've seen looks good - I will be happy to see cat_bulk.py replaced. Main question so far is whether we can use apmservertest.Server.

systemtest/gencorpora/apmserver.go
axw: I read a little further and probably answered my own question: apmservertest.Server requires a `testing.TB`. Maybe we can get rid of that requirement?

lahsivjar: That was my first approach but I decided to abandon it because the `apmservertest.Server` is tightly coupled with `testing.TB` utilizing `Cleanup` function to prevent unnecessary handling to leak into the test cases. I was thinking to extend the `APMServer` command to meet the requirements of test cases in a generic way and then replace `apmservertest#Server` to use it with `testing.TB`. But, this proved to be a lot more work than I anticipated and I thought we can address it at a later point. Let me know if you think it should be addressed now.

lahsivjar:
>  I will be happy to see cat_bulk.py replaced

@axw Yup, I am working on this in my next PR.

systemtest/gencorpora/apmserver.go
axw: Yes, that's fair enough.

I spiked on making the `testing.TB` part optional - I might create a followup on your PR if you don't mind.

systemtest/gencorpora/catbulk.go
```diff
@@ -0,0 +1,147 @@
...
+		server: &http.Server{
+			Addr:    addr,
+			Handler: handleReq(writer),
+		},
+		writer: writer,
+	}, nil
+}
+
+// Start starts the fake ES server on a listener.
+func (s *CatBulkServer) Start() error {
```
axw, line 74: Nit: maybe call this Serve? When I see "Start" I think of non-blocking methods like `os/exec.Command.Start`

systemtest/gencorpora/catbulk.go
```diff
@@ -0,0 +1,147 @@
...
+	server   *http.Server
+	Addr     string
+
+	writer io.WriteCloser
+}
+
+// GetCatBulkServer returns a HTTP Server which can serve as a
+// fake ES server writing the response of the bulk request to the
+// provided writer. Writes to the provided writer must be thread safe.
+func GetCatBulkServer() (*CatBulkServer, error) {
```
axw, line 50:
```suggestion
func NewCatBulkServer() (*CatBulkServer, error) {
```

axw (approved):
LGTM, just a couple of nitpicks

systemtest/gencorpora/catbulk.go
lahsivjar: Resolved, also updated `GetAPMServer` -> `NewAPMServer`

PR #8880, opened 2022-08-16
feat: migrate to the new net/netip package

kruskall:
## Motivation/summary

The new net/netip package defines a new IP address type, Addr.
Compared to the existing net.IP type, the netip.Addr type takes
less memory, is immutable, and is comparable so it supports ==
and can be used as a map key.

See https://go.dev/doc/go1.18#netip

## Checklist


internal/processor/otel/traces.go
```diff
@@ -345,13 +346,13 @@ func TranslateTransaction(
 				foundSpanType = httpSpan
 				httpServerName = stringval
 			case semconv.AttributeHTTPClientIP:
-				event.Client.IP = net.ParseIP(stringval)
+				event.Client.IP = netip.MustParseAddr(stringval)
```
marclop, line 349: We shouldn't be using the `Must` function here since it'll cause APM Server to panic. We don't seem to have a panic / recover interceptor for gRPC, so it's likely to cause APM Server to crash. Instead, we should use [`netip.ParseAddr`](https://pkg.go.dev/net/netip@go1.18.5#ParseAddr) an only assign the value on `err == nil`
```suggestion
				if ip, err := netip.ParseAddr(stringval); err == nil {
					event.Client.IP = ip
				}
```

internal/processor/otel/traces.go
```diff
@@ -345,13 +346,13 @@ func TranslateTransaction(
...
 			case semconv.AttributeHTTPUserAgent:
 				event.UserAgent.Original = stringval
 
 			// net.*
 			case semconv.AttributeNetPeerIP:
-				event.Source.IP = net.ParseIP(stringval)
+				event.Source.IP = netip.MustParseAddr(stringval)
```
marclop, line 355: Same here:
```suggestion
				if ip, err := netip.ParseAddr(stringval); err == nil {
					event.Source.IP = ip
				}
```

marclop (changes requested):
LGTM, minus the usage of `netip.MustParseAddr` in non test code with user-supplied values.

internal/processor/otel/traces.go
kruskall: Thanks for spotting this!

marclop (approved):
Looks great!! Thanks

PR #8898, opened 2022-08-18
systemtest/apmservertest: make use of testing.TB optional

axw:
Rename `NewServer` to `NewServerTB`, and `NewUnstartedServer` to `NewUnstartedServerTB`. Introduce a new version of `NewUnstartedServer` which does not take a `testing.TB`.

Existing uses of `TB.Fatal` have been replaced with panics, and cleanup has been refactored. The `New*TB` functions will call `TB.Cleanup` to automatically close the server when the text exits, and set up a log file. Callers of the non-TB function are expected to call `Server.Close` themselves, and there is no log file set up.


lahsivjar (approved):
Thanks, this looks much better than my plan to refactor.

PR #8900, opened 2022-08-18
Fix race condition in custom libbeat instrumentation

### LINKED ISSUE #8383: "Output Events Rate" in stack monitoring is always zero

lahsivjar:
## Motivation/summary

The PR fixes race condition in libbeat instrumentation that surfaces during reload (initial and on subsequent configuration change). The race condition is in setting up custom libbeat instrumentation by using beats `monitoring#DefaultRegistry`.  The race arises from two reload events captured by APM-Server for every reload operation including the initial config load. The two reloads are used by APM-Server to load `inputs` using the beats reload registry and `output` using beats `OutputConfigReloader`.

Specifically, the race condition is in `serverRunner#newFinalBatchProcessor` (called from `serverRunner#run`) due to updates to a global `monitoring#Default` registry and the two reloads explained above. Even though the access to the `monitoring#Default` is protected by mutex in `serverRunner#newFinalBatchProcessor`; it is still possible that `serverRunner#newFinalBatchProcessor` call for a temporary reload happens later for a reload event that happens earlier. This will cause `monitoring#Default` to be in an incorrect state:
1. For initial load, the incorrect state would be no custom instrumentation
2. For subsequent reloads, the incorrect state would be instrumentation with an incorrect model indexer

## Checklist

## How to test these changes

1. Run `TestFleetIntegrationMonitoring` multiple times and observe they are not flaky
2. APM-Server metrics for elasticsearch output events should always be present in stack monitoring

## Related issues
Closes #8383


internal/beater/beater.go
```diff
@@ -312,14 +312,23 @@ func (r *reloader) reloadOutput(config *reload.ConfigWithMeta) error {
...
-	r.mu.Unlock()
 	return r.reload()
 }
 
-func (r *reloader) reload() error {
+func (r *reloader) reloadOnce() error {
 	r.mu.Lock()
 	defer r.mu.Unlock()
+
+	return r.reload()
```
axw, line 324: Seeing as this is only called in one place, should we inline it there?

axw (approved):
Nice one! Thanks for tracking this down.

Maybe in the future we can look at splitting `serverRunner.run` into two methods: one which initialises things, called with a mutex; and one which actively runs, which may be concurrent.

internal/beater/beater.go
marclop: I think it's also called in `reloadOutput`, on line 317, right?

internal/beater/beater.go
```diff
@@ -345,6 +354,17 @@ func (r *reloader) reload() error {
...
 	}()
+
+	// Wait for the new runner to start; this avoids the race condition in updating
+	// the monitoring#Deafult global registry inside the runner due to two reloads,
+	// one for the inputs and the other for the elasticsearch output
+	select {
+	case <-runner.done:
+		return errors.New("runner exited unexpectedly")
+	case <-runner.started:
+		// runner has started
```
marclop, line 365: Nice catch and great solution!

marclop (approved):
LGTM! Should this PR also be back ported to `8.4`? Seems like a good candidate to include with the TBS reload fix.

internal/beater/beater.go
axw: `reload` is called directly there, because `reloadOutput` already holds the mutex

lahsivjar: Removed `reloadOnce` and inlined the mutex acquisition for standalone mode in `run`.

PR #8917, opened 2022-08-22
apmbench: Rewrite handler to be concurrently safe

marclop:
## Motivation/summary

Changes the architecture of the `eventhandler.Handler` to make it safe
for concurrent use. Drops the previous usage of `bytes.Reader` in the
batches and instead uses the `[]byte` representation. A new `sync.Pool`
is used internally and re-uses `bytes.Reader` before sending the batch.

The main issue with creating new eventhandlers concurrently in all the
BenchmarkParallel goroutines is that the exact same data is compressed
over and over, resulting in a lot of unnecessary CPU cycles burnt and
also results part of the benchmark time spent initializing the event
handlers rather than sending events to the APM Server.

The benchmark results appear to be much more stable than before and the
throughput numbers also appear to be much closer between benchmark runs:

```
name              old events/sec  new events/sec  delta
AgentAll-2048     9.46k ±44%      18.42k ± 2%     +94.73% (p=0.10 n=3+3)
AgentGo-2048      12.4k ±49%      20.7k ± 3%      +67.36% (p=0.10 n=3+3)
AgentNodeJS-2048  11.6k ±33%      16.4k ± 4%      +41.24% (p=0.10 n=3+3)
AgentPython-2048  15.4k ± 7%      18.5k ± 2%      +19.71% (p=0.10 n=3+3)
AgentRuby-2048    17.8k ± 3%      17.9k ± 7%      ~     (p=0.70 n=3+3)
```

The above result was obtained while benchmarking a 60g APM Server backed
with an Elasticsearch cluster of 116GB of RAM (58GB instance x 4) and
12 shards for each of the APM data streams.

## Checklist

## How to test these changes

Run the apmbench benchmarks using `testing/benchmark`

## Related issues

Part of #8775


systemtest/cmd/apmbench/main.go
```diff
@@ -94,11 +94,19 @@ func BenchmarkAgentRuby(b *testing.B, l *rate.Limiter) {
...
+			if err != nil {
+				b.Fatal(err)
+			}
+			return h
+		})
 		for pb.Next() {
 			h.SendBatches(context.Background())
 		}
+		benchtest.ReleaseEventHandler(expr, h)
 	})
```
axw, line 110: This seems awkward, since it's adding to a global pool (i.e. updating the pool map) - but the handler is parameterised with a rate limit and that's not part of the map key. Is the map necessary?

Would it be possible to simplify to something like this?

```go
pool := sync.Pool{New: func() any { return benchtest.NewEventHandler(b, expr, l) }}
b.RunParallel(func(pb *testing.PB) {
    h := pool.Get().(*eventhandler.Handler)
    defer pool.Put(h)
    defer h.Reset()
    // use h
})
```

systemtest/loadgen/eventhandler/handler.go
```diff
@@ -183,6 +185,39 @@ func (h *Handler) WarmUpServer(ctx context.Context) error {
 	}
 }
 
+// Clone creates a new instance of the Handler, copying the batches contents
+// to a new slice.
+func (h *Handler) Clone() (*Handler, error) {
```
axw, line 190: Perhaps a stupid question: is it needed to clone, or can we reset the batches in the Handler.Reset method? I think resetting to the initial state after using the handler would fit better with the pool pattern.

systemtest/benchtest/main.go
```diff
@@ -232,22 +232,51 @@ func Run(allBenchmarks ...BenchmarkFunc) error {
...
+			// we use log.Fatal in case there's an error.
+			if err != nil {
+				log.Fatal(err)
+			}
+			return h
+		})
+		// NOTE(marclop) h.Reset() is unlikely to do much when the handler
+		// is used in non-testing code, since the URL and token are set at
+		// a global level and remain unchanged for the duration of the program.
+		h.Reset(url, token)
```
axw, line 260: When do the URL or token ever change?

marclop: not during the normal execution of the code. I was thinking bout removing this Reset function and changing the test code to initialize a single http test server that is re-used for the warm-up tests. That would allow us to remove this.

systemtest/cmd/apmbench/main.go
marclop: I was trying to not have to change the signature of the current `BenchmarkFunc`. I decided to use the map, since currently `benchtest.NewEventHandler` filters the files using a the provided glob expression. If we're indiscriminately storing the event handlers into a single pool, then the wrong handlers will be re-used.

If we implement as you suggest, then the pool won't survive after the benchmark function has exited, since it's local to that function, so the pool won't be very helpful.

Another option would be to have `NewEventHandler` scan all the files and store the compressed batches in a `map[string][]batch`. That would allow us to use a single pool and re-initialize the event handler and reset it just before putting it back into pool.

systemtest/loadgen/eventhandler/handler.go
marclop: I think I answered this question above.

systemtest/cmd/apmbench/main.go
axw: > If we implement as you suggest, then the pool won't survive after the benchmark function has exited, since it's local to that function, so the pool won't be very helpful.

Ah - I was thinking that they were already local to the function, but I see now.

> Another option would be to have NewEventHandler scan all the files and store the compressed batches in a map[string][]batch. That would allow us to use a single pool and re-initialize the event handler and reset it just before putting it back into pool.

This sounds sensible to me.

systemtest/benchtest/pool.go
```diff
@@ -0,0 +1,51 @@
...
+var handlerPoolMapMu sync.RWMutex
+
+// NewPooledEventHandler creates or returns a previously used eventhandler.Handler.
+func NewPooledEventHandler(p string, newFunc func() any) *eventhandler.Handler {
+	handlerPoolMapMu.RLock()
+	pool, ok := handlerPoolMap[p]
+	handlerPoolMapMu.RUnlock()
+	if ok {
+		return pool.Get().(*eventhandler.Handler)
+	}
```
simitt, line 36: In theory this could lead to returning a handler from an existing pool with a different `newFunc`.

marclop:
I've dropped the `sync.Pool` to re-use `eventhandler.Handler` and made it safe for concurrent use. Now different goroutines can use the same compressed batch representation since it's immutable and create / use `bytes.Reader` before the batch is sent. The `bytes.Reader` are now reused and should provide a more fitting design with `sync.Pool` within the event handler.

simitt (approved):

systemtest/benchtest/main.go
```diff
@@ -232,22 +232,35 @@ func Run(allBenchmarks ...BenchmarkFunc) error {
...
-			return fmt.Errorf("unable to create warm-up handler: %w", err)
-		}
 		wg.Add(1)
 		go func() {
 			defer wg.Done()
+			// Create the context within the first goroutine that calls once.
+			// This ensures that the timeout starts counting from the moment
+			// the warm-up traffic starts.
+			// This delay starts becoming more noticeable once the number of
+			// agents increases.
```
axw, line 253: Did you make this change before you made the EventHandler safe for concurrent use? I'm surprised that creating the timeout context here vs. just before the loop makes a difference, seeing as there's practically nothing done at the beginning of the loop anymore.

axw (approved):
LGTM. Thanks, this is much easier for me to follow. Just one question about the sync.Once, but it's not blocking.

systemtest/benchtest/main.go
marclop: oh I forgot to remove this code. Thanks for catching this.

PR #8918, opened 2022-08-22
tf/ec_deployment: Use script for shard settings

marclop:
## Motivation/summary

Removes the `elasticstack` provider in favor of using a shell script to
set the APM shard settings in the component templates. The reason being
that replacing and removing the deployment is awkward since when destroy
is called, the component templates are managed by Terraform and deleting
them is not an option. With a shell script we just ensure the shards are
set to the number we desire, without worrying about clean-up.

## Checklist

## How to test these changes

Change the shard settings in `testing/benchmark`.

## Related issues

Part of #8775


axw (approved):
> since when destroy is called, the component templates are managed by Terraform and deleting them is not an option.

Not an option because they're referenced by the index templates, right? Seems like something that could be improved.

With the upcoming support for `@custom` ingest pipelines, they're just ignored if they don't exist. If we had something like that for component templates, then we would be able to delete them right?

marclop:
@axw yes, that's part of the issue, another issue (terraform it seems) that any time you mark the ec_deployment as tainted or it needs to be replaced, the reconciliation of the component template resource fails, as it tries to refresh the state, but fails to do so.

PR #8920, opened 2022-08-22
testing/benchmark: Add make cleanup-elasticsearch

marclop:
## Motivation/summary

Adds a new Makefile target `cleanup-elasticsearch` which can be used to
wipe all data created by the `apmbench` benchmarks.

## Checklist

## How to test these changes

1. Run the benchmarks (`testing/benchmark`)
2. Run: `make cleanup-elasticsearch`

## Related issues

Part of #8775


axw (approved):

testing/benchmark/Makefile
```diff
@@ -117,3 +117,12 @@ ssh:
...
 	@echo 'export TF_VAR_docker_image_tag_override={"elasticsearch":"$(TAG)","kibana":"$(TAG)","apm":"$(TAG)"}'
+
+.PHONY: cleanup-elasticsearch
+cleanup-elasticsearch:
+	$(eval ELASTICSEARCH_URL = $(shell terraform output elasticsearch_url))
+	$(eval ELASTICSEARCH_USER = $(shell terraform output -raw elasticsearch_username))
+	$(eval ELASTICSEARCH_PASS = $(shell terraform output -raw elasticsearch_password))
+	$(eval APM_DATA_STREAMS = traces-apm*,metrics-apm*,logs-apm*)
+	@ echo "-> Deleting APM Server data streams..."
+	@ curl -u $(ELASTICSEARCH_USER):$(ELASTICSEARCH_PASS) -XDELETE $(ELASTICSEARCH_URL)/_data_stream/$(APM_DATA_STREAMS)
```
Mpdreamz, line 128: Drive by nitpick: Could this be a default target dependency of `run-benchmark`, that way `run-benchmark` itself can just simply be repeated.

marclop: I thought about it but ended up not adding it as a dependency to the `run-benchmark` target, since we don't necessarily want to wipe data every time we run the benchmarks right now. We can always add it to that target as a dependency or call it explicitly when / if we want to re-run benchmarks against the same deployment

PR #8926, opened 2022-08-22
systemtest: Fix apm-server log getter from test container

lahsivjar:
I stumbled across this issue where APM-Server logs were not shown on an error of some systemtest even when the code expects them to be printed. After debugging, the root cause was that the docker container is using `UTC` timezone whereas the code which fetches the logs uses the machine's current local time.

I was also thinking maybe we can make APM-Server logs to be visible in `docker-logs` (maybe by configuring APM-Server run by elastic-agent to log to `stderr`?). Let me know what other @elastic/apm-server folks think about this. I can create an issue in the backlog if required.

### DIFF
--- a/systemtest/containers.go
+++ b/systemtest/containers.go
@@ -566,7 +566,7 @@ func (c *ElasticAgentContainer) APMServerLog() (io.ReadCloser, error) {
 	return c.container.CopyFileFromContainer(
 		context.Background(), fmt.Sprintf(
 			"/usr/share/elastic-agent/state/data/logs/default/apm-server-%s-1.ndjson",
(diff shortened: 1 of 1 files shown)


axw (approved):
> I was also thinking maybe we can make APM-Server logs to be visible in docker-logs (maybe by configuring APM-Server run by elastic-agent to log to stderr?). Let me know what other https://github.com/orgs/elastic/teams/apm-server folks think about this. I can create an issue in the backlog if required.

I think it would be nice if the logs were in stderr, but I worry about doing anything too different to how APM Server will run in the wild.

Issue #8928, opened 2022-08-23
Refactor "beater" package


axw:
We currently have a [big ball of mud](https://en.wikipedia.org/wiki/Big_ball_of_mud) in the codebase: the "beater" package and its sub-packages. These comprise:
 - our interface to libbeat (including Elastic Agent/Fleet management)
 - setup of HTTP & gRPC servers
 - definition of HTTP middleware & gRPC interceptors
 - route registration (intake, agent config, ...)
 - model processor creation
 - configuration
 - Java agent attachment
 - telemetry
 - authentication/authorization framework
 - self-instrumentation

This is a growing source of complexity and tech debt, manifesting in issues such as https://github.com/elastic/apm-server/issues/8383. We should look at refactoring "beater" into several distinct modules, with clearly defined responsibilities.

If nothing else, we must define a clear line between libbeat and the rest of apm-server code. This will enable us to better test things like reloading configuration and its effect on monitoring (see issue linked above) without running a complete apm-server. It would also enable us to explore alternative methods of deployment, e.g. splitting APM Server into multiple services to fit a stream processing architecture.

PR #8929, opened 2022-08-23
Remove dependencies from non-beater code on beater/config

axw:
## Motivation/summary

Stop depending directly on `beater/config` from all the packages. Instead, limit use of `beater/config` to `beater` itself, and translate configuration into just the bits needed by each downstream package. This will enable us to make changes to configuration independently of updating downstream users.

This does lead to a bit of duplication, but as Rob says, "a little copying is better than a little dependency".

## Checklist

## How to test these changes

N/A

## Related issues

https://github.com/elastic/apm-server/issues/8928


internal/beater/api/mux_config_agent_test.go
```diff
@@ -38,18 +37,22 @@ func TestConfigAgentHandler_AuthorizationMiddleware(t *testing.T) {
 		rec, err := requestToMuxerWithPattern(cfg, AgentConfigPath)
 		require.NoError(t, err)
 		require.Equal(t, http.StatusUnauthorized, rec.Code)
-		approvaltest.ApproveJSON(t, approvalPathConfigAgent(t.Name()), rec.Body.Bytes())
+		assert.JSONEq(t,
```
marclop, line 40: TIL

marclop (approved):
Looks good! I too prefer de-coupling configs, I think that in practice it'll make little or no difference to be copying around the configuration since most of the copying will happen just once.

lahsivjar (approved):
Love the beater refactoring ❤️

PR #8937, opened 2022-08-23
systemtest/gencorpora: generate corpus metadata

lahsivjar:
## Motivation/summary

A `corpora` is required for defining the rally track. The [corpora schema](https://esrally.readthedocs.io/en/stable/track.html#corpora) requires a list of document files and each document has the following requirements:
1. `document-count` and `source-file` are mandatory
2. `compressed-bytes` and `uncompressed-bytes` is recommended

This PR adds support to generate the document file with the above fields (other than `compressed-bytes`).

## Checklist

## How to test these changes

N/A

## Related issues

Part of: #8754


systemtest/gencorpora/catbulk.go
```diff
@@ -113,18 +171,31 @@ func handleReq(writer io.Writer) http.HandlerFunc {
...
+				n, err := fmt.Fprintln(writer, scanner.Text())
+				stat.bytes += n
+
 				// create mock bulk response considering all request items
 				// exist with action_and_metadata followed by source line
 				if count%2 == 0 {
+					respStatus := http.StatusOK
+					if err != nil {
+						respStatus = http.StatusInternalServerError
+						stat.count++
```
axw, line 185: Is it intentional that we are counting failed docs twice? (stat.count++ here and below)
If so, why?

systemtest/gencorpora/catbulk.go
```diff
@@ -113,18 +171,31 @@ func handleReq(writer io.Writer) http.HandlerFunc {
 
 			mockResp := esutil.BulkIndexerResponse{}
 			scanner := bufio.NewScanner(reader)
+			var stat docsStat
 			for count := 0; scanner.Scan(); count++ {
-				fmt.Fprintln(writer, scanner.Text())
+				n, err := fmt.Fprintln(writer, scanner.Text())
```
axw, line 176: Maybe panic if there's an error? We wouldn't expect an error except in case of something exceptional like disk full, right?

axw (commented):
A few early comments. Looks good overall.

systemtest/gencorpora/catbulk.go
lahsivjar: Apologies, I was in between commits and pushed a mess 😓 . I have fixed this now.

systemtest/gencorpora/catbulk.go
lahsivjar: I have reworked the PR and now I am logging the errors in this case and discarding the whole request. The error log should be visible in APM-Server logs also along with the `gencorpora` logs, let me know if it makes sense.

systemtest/gencorpora/catbulk.go
```diff
@@ -139,9 +210,9 @@ func handleReq(writer io.Writer) http.HandlerFunc {
 	})
 }
 
-func getWriter(writePath string) (io.WriteCloser, error) {
-	if writePath == "" {
+func getWriter(path string) (io.WriteCloser, error) {
+	if path == "" {
 		return os.Stdout, nil
```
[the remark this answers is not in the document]
lahsivjar: Ah, good point. Not sure if `NopCloser` for a writer is a good idea. What do you think about dropping `stdout` output and making the `directory` a required property?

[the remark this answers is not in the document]
axw: That sounds fine to me, it probably doesn't help to write the two different files to stdout anyway.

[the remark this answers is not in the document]
lahsivjar: Agreed. I have pushed a commit to remove `stdout` and now the write defaults to the current directory (didn't implement required on `write-dir` since it is not supported by default for the `flag` package and introduces additional logic for explicit validation in the code path).

systemtest/gencorpora/catbulk.go
```diff
@@ -53,44 +62,93 @@ func NewCatBulkServer() (*CatBulkServer, error) {
...
 
 // Serve starts the fake ES server on a listener.
 func (s *CatBulkServer) Serve() error {
+	go func() {
+		if err := s.metaWriter(); err != nil {
+			log.Println("failed to write metadata", err)
+		}
+	}()
+
 	if err := s.server.Serve(s.listener); err != http.ErrServerClosed {
```
kruskall, line 93:
```suggestion
	if err := s.server.Serve(s.listener); err != nil && err != http.ErrServerClosed {
```

**nit**: Non blocking, but I think this should check for `nil` before returning. It doesn't change the behaviour of the code but it helps with understanding the flow

systemtest/gencorpora/catbulk.go
```diff
@@ -113,18 +171,40 @@ func handleReq(writer io.Writer) http.HandlerFunc {
 
 			mockResp := esutil.BulkIndexerResponse{}
 			scanner := bufio.NewScanner(reader)
+
+			var stat docsStat
+			var buf bytes.Buffer
 			for count := 0; scanner.Scan(); count++ {
-				fmt.Fprintln(writer, scanner.Text())
+				if _, err := fmt.Fprintln(&buf, scanner.Text()); err != nil {
```
kruskall, line 178:
```suggestion
				if _, err := buf.Write(scanner.Bytes()); err != nil {
```

**suggestion**: I think we can avoid a byte->string conversion and use the buffer directly. WDYT ?

systemtest/gencorpora/catbulk.go
lahsivjar: Good catch, thanks!

systemtest/gencorpora/catbulk.go
lahsivjar: This will not work since the bulk API requires newline delimited JSON and scanner strips EOL markers. However, it is possible to optimize performance here by customizing the scanner -- I was being lazy here since it is not in the critical code path. Let me update this.

axw (approved):
LGTM!

systemtest/gencorpora/catbulk.go
axw: Looks good :)

lahsivjar:
I have updated the code to split input expecting NDJSON doc with metadata followed by source. This should cleanup the code a bit and have less memory footprint.

axw (approved):
Nice. LGTM.

PR #8950, opened 2022-08-24
Remove top-level "images" directory

axw:
These files seem to be a mistake, as they are duplicates of images in `docs/legacy/guide/images`, and are not used in any docs references. They were added in https://github.com/elastic/apm-server/pull/7218


bmorelli25 (approved):
Thank you!

PR #8952, opened 2022-08-24
Allow gencorpora to replay events

lahsivjar:
## Motivation/summary

By playing the recorded events without replay the ES corpus has **18817** documents accounting for **~56MBs** of data. This PR allows replaying events in order to increase the corpus size for testing with rally. APM-Server might aggregate some events thus the generated corpus will not always be a multiple of the replay count.

## Checklist

## How to test these changes

N/A

## Related issues

Part of: #8754


systemtest/gencorpora/gencorpora.go
```diff
@@ -0,0 +1,100 @@
...
+func generateLoad(ctx context.Context, serverURL string, replayCount int) error {
+	inf := loadgen.GetNewLimiter(0)
+	handler, err := loadgen.NewEventHandler(`*.ndjson`, serverURL, "", inf)
+	if err != nil {
+		return err
+	}
+
+	for i := 0; i < replayCount; i++ {
+		if _, err = handler.SendBatches(ctx); err != nil {
+			return err
```
marclop, line 96: Should we include a bit more information on this error? Since it's possible we the `handler.SendBatches` fails on any iteration, it may be interesting to send which iteration the error was found in. What do you think?

```suggestion
			return fmt.Errorf("failed sending batches on iteration %d: %w ", i+1, err)
```

systemtest/gencorpora/gencorpora.go
```diff
@@ -0,0 +1,100 @@
...
+	"golang.org/x/sync/errgroup"
+
+	"github.com/elastic/apm-server/systemtest/loadgen"
+)
+
+// RunBlocking runs APM-Server and CatBulk server followed by
+// sending load to APM-Server and writing the generated corpus
+// to a file. RunBlocking exits on error or on successful
+// corpora generation.
+func RunBlocking(rootCtx context.Context) error {
```
marclop, line 36: `RunBlocking` seems to imply that there is also a non-blocking function available, but I don't think that's the case.
Also, I would expect that `RunBlocking` blocks forever or until the context is cancelled, but that isn't the case here either.

 Would it make sense to just have a `Run()` method? The godoc comment above is informative enough where callers can make good decisions by just reading that.

systemtest/gencorpora/gencorpora.go
lahsivjar: Sounds good, I have updated the PR.

marclop (approved):
LGTM!

systemtest/gencorpora/gencorpora.go
lahsivjar: Agree with the logic, I would expect `RunBlocking` to run until I cancel the context. I have updated the function name to `Run`. Thanks for catching this.

PR #8955, opened 2022-08-24
test: fix processor benchmarks testdata path

kruskall:
## Motivation/summary

BenchmarkBackendProcessor and BenchmarkRUMV3Processor were a no-op due to the
glob filter not matching anything.
Update filter to use the correct path.

## Checklist


marclop (approved):
good find!

PR #8967, opened 2022-08-25
Introduce interfaces for gRPC auth & metrics

axw:
## Motivation/summary

Rather than injecting callbacks into the auth and metrics interceptors at construction time, we remove the direct dependency from the interceptor to service implementations by introducing interfaces that gRPC services may implement.

We also have a global callback registry for registering monitoring maps, which is necessary (at least for now) as we cannot extend the OTLP gRPC service implementations to implement the metrics interface. This isn't needed for auth, as the OTLP services rely on the default auth ("authorization" metadata) implementation.

This will enable us to implement gRPC services under x-pack. This will be used in a followup, where the gRPC server will be exposed via ServerParams.

## Checklist

## How to test these changes

N/A

## Related issues

None


lahsivjar (approved):
LGTM!

simitt:
Too late for review, but huge +1 on this change!

PR #8971, opened 2022-08-25
Fix commit compare links in changelogs

lahsivjar:
Commit compare links are not correct for multiple past versions.


kruskall (approved):
Good catch!

Non-blocking question: should this be backported to other branches ? It seems the doc is using the broken links: https://www.elastic.co/guide/en/apm/guide/current/release-notes-8.4.html

lahsivjar:
Sounds good to me, atleast we can backport to `8.4`. @axw @simitt WDYT?

axw:
Backporting to 8.4 sounds like a good idea. Probably not worth going to the trouble of backporting to the other branches, seeing as it'll have to be fixed up for each other release branch.

PR #8717, opened 2022-07-26
CI: Publish apm package to Package Storage v2

mtojek:
Related: https://github.com/elastic/apm-server/issues/8491

This PR follows the [v2 migration guide](https://github.com/elastic/elastic-package/blob/main/docs/howto/use_package_storage_v2.md) to onboard the apm package with Package Storage v2.

The `apmpackage` stage will call the signing pipeline and the publishing pipeline. It appends a timestamp to the semver prerelease.

Testing:
https://apm-ci.elastic.co/job/apm-server/job/apm-server-package-mbp/job/PR-8717/11/ (dry-run)
.zip + .zip.sig in [build artifacts](https://apm-ci.elastic.co/job/apm-server/job/apm-server-package-mbp/job/PR-8717/11/artifact/build/packages/)

PR #8972, opened 2022-08-25
fix update-apm package-storage

v1v:
### What

Set `PACKAGE_STORAGE_BRANCH` value only once.

### Why

Solves the issue with the automation for the package-store.

```
Creating pull request for update-apm-20220825154830 into snapshot in elastic/package-storage

pull request create failed: GraphQL: Head sha can't be blank, Base sha can't be blank, No commits between snapshot and update-apm-20220825154830, Head ref must be a branch (createPullRequest)
```

https://github.com/elastic/package-storage/pull/5414 was automatically created.

### DIFF
--- a/.ci/scripts/Makefile
+++ b/.ci/scripts/Makefile
@@ -5,7 +5,7 @@
 BUILD=../../build
 PACKAGE_STORAGE=.package-storage
 PACKAGE_STORAGE_PATH=$(BUILD)/$(PACKAGE_STORAGE)
-PACKAGE_STORAGE_BRANCH = $(eval PACKAGE_STORAGE_BRANCH := $$(shell date "+%Y%m%d%H%M%S"))$(PACKAGE_STORAGE_BRANCH)
+PACKAGE_STORAGE_BRANCH:=update-apm-$(shell date "+%Y%m%d%H%M%S")
 APM_SERVER_VERSION=$(shell make --no-print-directory -C ../../ get-version)
 REPO_PROJECT=elastic/package-storage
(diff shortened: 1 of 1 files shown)


v1v:
@elastic/apm-server  and @mtojek , do we need to support the legacy system? or v2?

I've just seen https://github.com/elastic/apm-server/pull/8717, so maybe this change is not needed anymore but removing the support for v1?

mtojek:
We may need it in case of emergency rollback to v1. Once we're good with https://github.com/elastic/kibana/issues/137759 and there are no more blockers, we will switch to v2.

.ci/scripts/Makefile
```diff
@@ -5,8 +5,10 @@
 BUILD=../../build
 PACKAGE_STORAGE=.package-storage
 PACKAGE_STORAGE_PATH=$(BUILD)/$(PACKAGE_STORAGE)
-PACKAGE_STORAGE_BRANCH=update-apm-$(shell date "+%Y%m%d%H%M%S")
+PACKAGE_STORAGE_BRANCH = $(eval PACKAGE_STORAGE_BRANCH := $$(shell date "+%Y%m%d%H%M%S"))$(PACKAGE_STORAGE_BRANCH)
```
axw, line 8:
```suggestion
PACKAGE_STORAGE_BRANCH:=update-apm-$(shell date "+%Y%m%d%H%M%S")
```

I think we'll always expand that variable? So may as well just evaluate it immediately, and keep the Makefile a bit simpler.

axw (approved):

PR #8973, opened 2022-08-25
test: try to make the benchmarks more stable

kruskall:
## Motivation/summary

Separate initialization from benchmarked code with subtests and avoid
calling StopTimer/StartTimer in the bench loop as the overhead might
affect the bench time.
Replace reader.Reset with r.Seek as we're reading for the same byte
slice and there is no need to allocate a new reader each time.

## Checklist


marclop (approved):
LGTM!

PR #8979, opened 2022-08-26
intake: Support non-blocking event processing

### LINKED ISSUE #7872: Allow clients to request non-blocking ingest behaviour

marclop:
## Motivation/summary

Introduces a new bool `async` query parameter for `intake/v2/events`.
This allows clients to ask the APM Server to process the contents of the
request in an async manner. This will only happen if the APM Server has
enough capacity to service the request, otherwise it will return the
infamous `publish.ErrFull` error.

## Checklist

## How to test these changes

1. `docker-compose up -d`
2. `cd systemtest && go run cmd/runapm/main.go`
3. `export RUNAPM_PORT=REPLACE_WITH_RUNAPM_PORT`
3. `curl -H 'Content-type: application/x-ndjson' --data-binary @testdata/intake-v2/errors_transaction_id.ndjson "http://localhost:${RUNAPM_PORT}/intake/v2/events?async=true"`
4. `curl -H 'Content-type: application/x-ndjson' --data-binary @testdata/intake-v2/invalid-event-type.ndjson "http://localhost:${RUNAPM_PORT}/intake/v2/events?async=true"`, see error:
```
{
  "accepted": 0,
  "errors": [
    {
      "message": "tennis-court: did not recognize object type",
      "document": "{\"tennis-court\": {\"name\": \"Centre Court, Wimbledon\"}}"
    }
  ]
}
```
5. Check that `logs-apm.error-default/_search` contains 1 document.
6. `for i in {1..5}; do curl -H 'Content-type: application/x-ndjson' --data-binary @testdata/intake-v2/heavy.ndjson "http://localhost:${RUNAPM_PORT}/intake/v2/events?async=true"; done`. Observe at least 2 "queue is full" responses.
```
{
  "accepted": 0,
  "errors": [
    {
      "message": "queue is full"
    }
  ]
}
```

## Related issues

Closes #7872


internal/beater/api/intake/handler.go
```diff
@@ -218,3 +247,61 @@ type jsonError struct {
...
+
+func (b *pooledBodyReader) Read(p []byte) (n int, err error) {
+	n, err = b.Buffer.Read(p)
+	if err != nil && !b.closed {
+		// Close the pooledBodyReader so it can be returned to the pool after
+		// io.EOF is returned (all the contents have been read).
+		if errors.Is(err, io.EOF) {
+			b.Close()
+		}
+	}
```
marclop, line 291: Alternatively, we could not do this in the `Read()` wrapping method and instead wrap the compressed reader with:

```go
// dualCloserReader wraps a two io.ReadCloser using
type dualCloserReader struct {
	// ReadCloser will be used for Reads
	io.ReadCloser
	closer io.Closer
}

//
func newDualCloserReader(reader, closer io.ReadCloser) *dualCloserReader {
	return &dualCloserReader{ReadCloser: reader, closer: closer}
}

// Close closes the two io.ReadCloser
func (d *dualCloserReader) Close() error {
	d.ReadCloser.Close()
	return d.closer.Close()
}
```

Then:
```go
 		reader, err := decoder.CompressedRequestReader(c.Request)
 		if err != nil {
 			writeError(c, compressedRequestReaderError{err})
			return
		}
		// In order to re-use the pooledBodyReaders, they muyst be returned to
		// the pool.
		if c.Async {
			reader = newDualCloserReader(reader, c.Request.Body)
		}
```

internal/processor/stream/processor.go
```diff
@@ -212,31 +218,70 @@ func (p *Processor) readBatch(
...
-	select {
-	case p.sem <- struct{}{}:
-	case <-ctx.Done():
-		return ctx.Err()
+	// (determined by batchSize) from the batch.
+	// The ceiling may also reduce the contention on the modelindexer.activeMu.
+	// Clients can set a async to true which makes the processor handle the
+	// events in the background. Returns with an error `publish.ErrFull` if the
+	// semaphore is full.
+	if err := p.acquireLock(ctx, async); err != nil {
```
lahsivjar, line 235: (I might be lacking context here) The use of semaphore seems to limit the concurrency which, in effect, also controls the load APM-Server is processing at any moment. An async operation will defeat this purpose (if it is intended).

*brain dump*: I think we can eliminate semaphore entirely and add a middleware that implements [token based rate limiter](https://pkg.go.dev/golang.org/x/time/rate) for achieving a similar end result as what is expected from the semaphore. We can additionally expose the configuration of a rate limiter so that users can configure it based on their specific workload. WDYT?

marclop: > (I might be lacking context here) The use of semaphore seems to limit the concurrency which, in effect, also controls the load APM-Server is processing at any moment. An async operation will defeat this purpose (if it is intended).

The semaphore won't be released back until the goroutine has finished work when `async=true`. So the semaphore will still have the desired effect. The only difference is in case the server is under load, then the semaphore won't be acquired and `publish.ErrFull` is returned.

> brain dump: I think we can eliminate semaphore entirely and add a middleware(?) that implements [token based rate limiter](https://pkg.go.dev/golang.org/x/time/rate) for achieving a similar end result as what is expected from the semaphore. We can additionally expose the configuration of a rate limiter so that users can configure it based on their specific workload. WDYT?

I had some thoughts on moving the semaphore to a middleware, but after this feature I'm not so sure. It seems to add a fair bit of complexity.

I'm not sure the rate limiter fits entirely with the limitation that we're trying to achieve here. The rate limiter would limit on a period of time (which?), while the semaphore is agnostic to time and always limits the concurrent number of decodes in the server. 

Since the rate limiter would be limiting based on a time period, which we don't want, it wouldn't matter to it if the events are happening serially (very fast) or concurrently.

internal/processor/stream/processor.go
```diff
@@ -212,31 +218,70 @@ func (p *Processor) readBatch(
...
+
+	// By default, streams are handled synchronously, however, clients can
+	// set the request to be processed asynchronously, in which case, a
+	// the processing will happen in the background within an new goroutine,
+	// and since error cannot be sent back to the client, they are logged.
+	if !async {
+		defer releaseResources()
+		return p.handleStream(ctx, baseEvent, sr, batchSize, processor, result)
+	}
+	go func() {
```
axw, line 255: Instead of doing _everything_ async, I was thinking of just doing the processing (ProcessBatch) async.

So we would decode the body in the request handler into batches as before, and send them to a queue/channel. If the queue is full, we would return with a "queue is full" error. Otherwise, it goes on the queue to be picked up by another short lived goroutine to send to ProcessBatch.

Then the question becomes how to size the channel of course :)

internal/beater/api/intake/handler.go
```diff
@@ -77,6 +81,30 @@ func Handler(handler StreamHandler, requestMetadataFunc RequestMetadataFunc, bat
...
+		//    work, because the HTTP server keeps a reference to the body and
+		//    closes that.
+		// This will result in an async request to perform synchronous work.
+		// The amount of bytes copied is minimized by copying the body contents
+		// before we attempt to create the de-compressors. It's still possible
+		// that a request is uncompressed, in which case, more copying will
+		// take place.
+		ctx := c.Request.Context()
+		if c.Async {
+			ctx = auth.CopyAuthorizer(ctx, context.Background())
```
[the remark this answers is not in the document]
marclop: I would also prefer that, but how would we access the private context key (struct) that we're using in the auth package? https://github.com/elastic/apm-server/blob/5dfaba7ca9e9840b590e6f6d4cfb3eef04847c0b/internal/beater/auth/context.go#L28-L33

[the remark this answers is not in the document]
marclop: That sounds good. I think we can export `AuthorizerFromContext` first, and revert that once the processing chains are split. Does that sounds good to you?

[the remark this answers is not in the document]
axw: SGTM

internal/beater/api/intake/handler.go
marclop: this isn't applicable anymore.

internal/processor/stream/processor.go
```diff
@@ -260,13 +283,30 @@ func (p *Processor) HandleStream(
...
+			return err
+		}
+	}
+}
+
+func (p *Processor) asyncProcessor(
+	ctx context.Context,
+	processor model.BatchProcessor,
+	batchChan chan *model.Batch,
+	result *Result,
```
axw, line 299: `result` isn't updated - should it even be passed in? I think it should only be updated by the synchronous code path, since it'll be returned to the client.

internal/processor/stream/processor.go
```diff
@@ -212,43 +218,60 @@ func (p *Processor) readBatch(
...
 	defer sp.End()
 
+	// Create a new processor var to avoid replacing the passed reference.
+	var processor model.BatchProcessor
+	if async {
+		// Allow 5 async batches to be processing per stream. The lambda
+		// extension doesn't (yet) stream to the server.
+		// If a single batch with more than the batchSize is sent (10),
+		// then the request can process 50 events asynchronously.
+		batchChan := make(chan *model.Batch, 5)
```
axw, line 265: Hmm, now that I see this in code: it means we could have higher memory usage for async=true, right? i.e. if you had 200 concurrent requests. Also this channel size feels a bit arbitrary. (I realise this is what I described in my other comment, I just hadn't quite thought it through.)

Would it make sense to insead allow a single request to process up to 200 (i.e. semaphore size) batches asynchronously? Something like this, in the loop below:
1. acquire the semaphore at the top of the loop, returning 503 if there's none available
2. call readBatch
3. start a new goroutine to process the batch; the goroutine will release semaphore when processing is done

internal/beater/api/intake/handler.go
```diff
@@ -77,6 +79,26 @@ func Handler(handler StreamHandler, requestMetadataFunc RequestMetadataFunc, bat
...
+		// errors while processing the batch be communicated back to the client.
+		// Instead, errors are logged by the APM Server.
+		async := asyncRequest(c.Request)
+
+		// When asynchronous processing is requested, the context authorizer is
+		// copied into a new context, since the current context is cancelled
+		// after the server handles the request.
+		ctx := c.Request.Context()
+		if async {
+			if authorizer, ok := auth.AuthorizationFromContext(ctx); ok {
```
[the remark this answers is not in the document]
marclop: great! I wasn't aware of that function. I've updated the code with this change and what you proposed.

internal/processor/stream/processor.go
```diff
@@ -212,61 +217,89 @@ func (p *Processor) readBatch(
...
 	defer sp.End()
 
+	var acquireLock bool
 	for {
+		// Async requests will re-acquire the lock after the first batch is
+		// scheduled to be processed. If the semaphore is full, then we return
+		// with `publish.ErrFull`.
+		if async {
+			if acquireLock {
+				if err := p.acquireLock(ctx, async); err != nil {
```
axw, line 265: I'm pretty sure there are some error cases where we're acquiring the lock, but never releasing it. e.g. if `readBatch` returns `io.EOF` but there are no events to decode, because it's the end of the stream.

internal/processor/stream/processor.go
```diff
@@ -212,61 +217,89 @@ func (p *Processor) readBatch(
...
+		sr.release()
+		p.releaseLock()
 		return err
 	}
 
+	defer func() {
+		sr.release()
+		if !async { // Release lock for sync processing.
+			p.releaseLock()
+		}
```
axw, line 252: We did acquire the lock above even if it's async - what has responsibility for releasing the lock now?

axw (commented):
Looking good!

The locking is a bit concerning - it looks like it may deadlock. Can you please check over that?

_Generally_ I find it tends to be best to follow the pattern that the code that acquires a lock releases the lock, i.e. always defer release after acquire. Since we need to reacquire the lock, this would mean ending the loop with the lock acquired but not released on the last loop.

Perhaps your initial thought of doubling up on the semaphore acquisition on first batch is going to simplify things?

internal/processor/stream/processor.go
```diff
@@ -213,29 +218,33 @@ func (p *Processor) readBatch(
...
 	sr := p.getStreamReader(reader)
+
+	// Release the stream reader after all the batches have been read. Synchronous
+	// requests also release the semaphore since batch processing is blocking.
 	defer func() {
 		sr.release()
-		<-p.sem
+		if !async {
+			p.semRelease()
+		}
```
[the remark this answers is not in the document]
marclop: good catch! I'll change the code to do what you propose, I think it looks pretty clean as well.

axw (approved):
LGTM :tada:

axw:
Confirmed with 8.5.0-BC1. To verify the "queue is full" behaviour, I ran a bunch of parallel `curl ... heavy.ndjson` requests, with request routing to Elasticsearch disabled.

PR #8983, opened 2022-08-26
beater: expose gRPC server in ServerParams

### LINKED ISSUE #8988: `host.hostname` is inconsistent between transaction events and aggregated transaction metrics
**APM Server version** (`apm-server version`): 8.4.0 (but also older versions)

`host.hostname` is inconsistent between transaction events and transaction metrics aggregated by APM Server.

**Steps to reproduce**:

 1. Send some transaction events to APM Server with `kubernetes.node.name` set, e.g. by setting the environment variable `KUBERNETES_NODE_NAME=...` and sending events using an Elastic APM agent.
 2. Wait for transaction events and transaction metrics to be recorded.
 3. Compare the `host.hostname` field of each document. They should be the same, but there is a bug with the ordering of batch processors which means that aggregation is occurring before preprocessing -- in particular, `modelprocessor.SetHostHostname`.

axw:
## Motivation/summary

Refactor beater so that the gRPC server can be exposed in ServerParams, enabling x-pack code to register additional gRPC services. This requires that we initialise various things outside of runServer, such as the authenticator and rate-limiter cache.

The creation of the batch processing chain, and injection of behaviour through wrapping the RunServerFunc (and now ServerParams) has been refactored to hopefully make things a bit more linear and easier to follow. It turns out that the existing method of injecting BatchProcessors in x-pack was buggy, and led to the aggregation code to run before preprocessors. This has been fixed.

## Checklist

## How to test these changes

N/A

## Related issues

Closes #8988


marclop (approved):
Looks good and thanks for all the additional GoDoc comments, helps clarify the parameters and the layers of wrapping a bit in the beater.

PR #8986, opened 2022-08-26
Add terraform module to benchmark Elasticsearch on Elastic Cloud with locally built APM-Server and Integration package

### LINKED ISSUE #8754: [MVP] Tooling for benchmarking Elasticsearch ingest changes

lahsivjar:
## Motivation/summary

Allows running rally to benchmark a locally built Elastic Cloud cluster. This allows developers to benchmark changes introduced in ES schema used by APM-Server.

## Checklist

## How to test these changes

1. Navigate to `testing/rally-cloud` and run `make init` if you are running for the first
time
2. Get `EC_API_KEY` by following [this guide](https://www.elastic.co/guide/en/cloud-enterprise/current/ece-restful-api-authentication.html#ece-api-keys).
3. Checkout the `testing/rally-cloud/variables.tf` to customize your benchmark.
4. Run `EC_API_KEY=<ec_api_key> make apply` to create the required infra and run rally.

The rally benchmark will execute everytime `make apply` is called however, the infrastucture
will be created only once. After testing is done make sure to destroy all infrastructure using
`EC_API_KEY=<ec_api_key> make destroy`

## Related issues

Closes #8754


axw (commented):
Looks great!

Regarding the need to run `make` in `testing/cloud`: perhaps instead of building a custom Kibana image, we would be better off using the default image, and uploading a custom integration package to Kibana after it starts like we do in the system tests: https://github.com/elastic/apm-server/blob/fd81083076545be7a35194013859d75a3cf6309c/systemtest/kibana.go#L137

testing/rally-cloud/Makefile
```diff
@@ -0,0 +1,42 @@
+GENCORPORA_PATH ?= ../../systemtest/cmd/gencorpora
+GENCORPORA_GOOS ?= linux
+GENCORPORA_GOARCH ?= amd64
```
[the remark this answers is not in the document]
lahsivjar: Thanks for catching this. They should be removed, they are remnants from when I was building and running the gencorpora on the remote rally worker machine.

testing/rally-cloud/variables.tf
```diff
@@ -0,0 +1,95 @@
...
+  type        = string
+  description = "GCP region"
+  default     = "us-west2"
+}
+
+variable "gcp_zone" {
+  type        = string
+  description = "GCP zone"
+  default     = "us-west2-b"
+}
```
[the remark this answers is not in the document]
lahsivjar: You are correct, the reason of zone variable was to minimise network overhead between rally workers and coordinator. 

The `google_compute_zones` looks good, I was not aware of this. The implementation also looks straightforward, we can just choose a random zone from all available zones. I will give this a try in the current PR itself.

testing/infra/terraform/modules/ec_deployment/variables.tf
```diff
@@ -123,3 +123,17 @@ variable "apm_index_shards" {
...
   type        = number
 }
+
+# Install custom APM integration package
+
+variable "install_custom_apm_integration_pkg" {
+  type = bool
+  description = "Whether or not to install a custom APM integration package"
+  default = false
+}
```
axw, line 133: Can we infer this from `custom_apm_integration_pkg_path` being non-empty? Is there any reason to specify one without the other?

axw (commented):
Nice - uploading the custom package is much more sensible than bundling in a custom Kibana. We should probably update testing/cloud later to make use of this too.

It would be great if we could combine the two new vars into one.

testing/infra/terraform/modules/ec_deployment/variables.tf
lahsivjar: I wanted to be more explicit here since not all users of `ec_deployment` module will want to install custom integration package. I have fixed this to use only 1 variable and updated the description of the variable to notify users that empty string will not install a custom integration.

axw: Got it. If it creates confusion then we can revisit, but IMO the name and description are (now) sufficiently self-explanatory.

lahsivjar: Makes sense, I have updated it to use a single variable.

axw (approved):
Looks great :)

axw:
I tested this earlier. Not with a BC, since it's independent of the stack release.

PR #9005, opened 2022-08-30
Add experimental profiling collection agent code

axw:
## Motivation/summary

Integrate the profiling collection agent code into APM Server, as a technical preview / experiment.

This feature is disabled by default, and can be enabled by setting `apm-server.profiling.enabled=true`. The profiling code will use its own go-elasticsearch client to write to `profiling-*` indices and data streams. Management of these indices / data streams is handled externally to APM Server.

When running under Elastic Agent & Fleet, APM Server's privileges are insufficient to writing to `profiling-*`, so it is necessary to provide an additional API Key for the profiling code to use. This can be injected (e.g. [using fleetctl](https://github.com/elastic/apm-server/tree/main/systemtest/cmd/fleetctl#updating-arbitrary-config)) as `apm-server.profiling.elasticsearch.api_key`.

## Checklist

## How to test these changes

1. Run `apm-server -E apm-server.profiling.enabled=true ...`
2. Run `pf-host-agent -disable-tls -collection-agent localhost:8200 -project-id 123 -secret-token abc`
3. Confirm data shows up in Kibana

Also check that apm-server's secret token auth is honoured. It is not currently possibly to configure the host agent to use API Key auth.

## Related issues

https://github.com/elastic/apm-server/issues/9051


inge4pres:
note on `systemtests/cmd/runapm`: it's missing the `extra_config` flag setup.
Do we want to amend that binary for local dev purposes?

axw:
@inge4pres can you please have a look primarily at the x-pack/apm-server/profiling code, which is _mostly_ a copy of pf-elastic-agent, but with some adaptations to fit into this codebase.

@elastic/apm-server can you please have a look primarily at the bits other than x-pack/apm-server/profiling, i.e. the glue code. There are some auth & monitoring bits in x-pack/apm-server/profiling which are relevant too: auth.go and monitoring.go

internal/beater/config/profiling.go
```diff
@@ -0,0 +1,89 @@
...
+func (c *ProfilingConfig) setup(log *logp.Logger, outputESCfg *config.C) error {
+	if !c.Enabled {
+		return nil
+	}
+	// NOTE(axw) this behaviour below is a bit different to what we have
+	// in other code that has Elasticsearch config. Here we always merge
+	// the provided Elasticsearch config over the top of the Elasticsearch
+	// output config. This allows one to provide only an API Key for
+	// profiling (-E apm-server.profiling.elasticsearch.api_key=...), but
+	// still use the same Elasticsearch output host, etc.
```
simitt, line 70: nit: are there use cases for configuring something other than the API Key at this point? If not, should we only extract the API Key for now to avoid weird edge cases where other ES configs (URLs, credentials, etc) get weirdly merged with each other?

internal/beater/config/profiling.go
```diff
@@ -0,0 +1,89 @@
...
+	if outputESCfg != nil {
+		if err := outputESCfg.Unpack(&c.ESConfig); err != nil {
+			return errors.Wrap(err, "error unpacking output.elasticsearch config for profiling collection")
+		}
+	}
+	if c.es != nil {
+		if err := c.es.Unpack(&c.ESConfig); err != nil {
+			return errors.Wrap(err, "error unpacking elasticsearch config for profiling collection")
```
simitt, line 78: the two error messages are ident, might be helpful to have an indicator which one is failing (generic ES output or profiling one)

simitt (approved):
Only minor comments from my side (focusing on the glue work); changes look pretty contained - thanks for keeping it that clean.

internal/beater/config/profiling.go
axw: They're not quite identical, the one above says `unpacking output.elasticsearch`. Anyway, to make it more obvious I'll change this one to say `error unpacking apm-server.profiling.elasticsearch`.

internal/beater/config/profiling.go
axw: I'm inclined to leave it, to allow some flexibility for 8.5. It'll get ripped out soon enough.

inge4pres (commented):
Looking good! 👍🏼 
Thanks for pulling this off, it's great that we are able to re-use the existing gRPC utilities for Profiling 👏🏼 

I left a couple of questions around validation of config and metrics.
I will test this once more in Cloud before final approval.

Should we include the new ILM code to manage non-datastream indices, are we going to raise a new PR?

inge4pres:
On the metrics topic: I noticed we already expose some metrics in the group `apm-server.profile`, can we re-use them to publish Profiling metrics?

axw:
> On the metrics topic: I noticed we already expose some metrics in the group apm-server.profile, can we re-use them to publish Profiling metrics?

Yes, we could do.

internal/beater/config/profiling.go
```diff
@@ -0,0 +1,122 @@
...
+			c.metricsES = c.es
+		}
+	}
+
+	return errors.Wrap(c.Validate(), "invalid config")
+}
+
+func (c *ProfilingConfig) Validate() error {
+	return nil
+}
```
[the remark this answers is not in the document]
axw: Fair enough, done

x-pack/apm-server/profiling/collector.go
```diff
@@ -0,0 +1,632 @@
...
+)
+
+var (
+	// metrics
+	//indexedDocs = metrics.NewLabelCounter("elastic_collector",
+	//	"indexed_document_count",
+	//	"grpc_method")
+	//indexedErrDocs = metrics.NewLabelCounter("elastic_collector",
+	//	"indexed_document_failed_count",
+	//	"grpc_method", "reason")
```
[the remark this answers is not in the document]
axw: > Would love to replace these 2 with 2 *monitoring.Ints with the same purpose.

We can, but bear in mind that `elastic-agent-libs/monitoring` does not support dimensional metrics. So we can either have 2 metrics for tracking all docs indexed/failed, or we can have a pair of metrics for each endpoint or perhaps each document type.

> What would be the best way to make this happen?

How about we get this PR merged first, and then iterate?

> My understanding is that these metrics would exposed via the Go expvar package, is that correct?

Yes. Bear in mind that expvar is disabled by default. It can be enabled, but there's no auth, and we haven't done an audit on exposure of things like command line args.

image

I think ideally we would be sending these metrics to telemetry.

axw:
> Should we include the new ILM code to manage non-datastream indices, are we going to raise a new PR?

Yes, I think we should merge this sooner than later, and then iterate.

x-pack/apm-server/profiling/collector.go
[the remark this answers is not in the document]
inge4pres: > How about we get this PR merged first, and then iterate?

Agreed 😄 👍🏼

> I think ideally we would be sending these metrics to the cloud monitoring cluster.

Was chatting about this yesterday with Silvia, I think we have a way of making that happen

inge4pres (approved):
👍🏼 

Just finished re-testing the latest commit in Cloud, all looks good with the addition config `metrics.elasticsearch` that _has to have_ the `hosts` field in it.

PR #9006, opened 2022-08-30
Fix: trigger signing if there is an unsigned revision

mtojek:
This PR fixes a bug in the `packaging.groovy`.

Unpublished package revisions should be signed before publishing.


axw (approved):
Thank you!

PR #9007, opened 2022-08-30
bump version to 8.4.2

simitt:
## Motivation/summary

Bump version to 8.4.2

### DIFF
--- a/apmpackage/apm/changelog.yml
+++ b/apmpackage/apm/changelog.yml
@@ -1,7 +1,7 @@
 # newer versions go on top
 #
 # change type can be one of: enhancement, bugfix, breaking-change
-- version: "8.4.1"
+- version: "8.4.2"
   changes:
     - description: Placeholder
       type: bugfix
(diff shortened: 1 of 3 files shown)


apmpackage/apm/changelog.yml
```diff
@@ -1,7 +1,7 @@
 # newer versions go on top
 #
 # change type can be one of: enhancement, bugfix, breaking-change
-- version: "8.4.1"
+- version: "8.4.2"
```
kruskall, line 4: **question**: shouldn't this go on top of `8.4.1` ?

kruskall (approved):
LGTM! Just a minor question

apmpackage/apm/changelog.yml
simitt: We haven't actually had any changes for the `8.4.1` package, so we are not releasing it and only bump the version to `8.4.2`.

kruskall: Ah, that entry is a placeholder 🤦

PR #9017, opened 2022-08-31
Set the runtime for mobile agents using the OS information

AlexanderWert:
For mobile agents the `runtime.*` fields can be directly derived from the `os.*` fields (unless it's explicitly overwritten by the agents). This is common for both, iOS and Android. 

## Motivation/summary
With this PR, the APM server sets the `service.runtime.name` and `service.runtime.version` to the same values as `host.os.name` and `host.os.version` in case that the `process.runtime.name` field is not explicitly provided (which is the default for the iOS and Android agents).

## Checklist

### DIFF
--- a/internal/processor/otel/metadata.go
+++ b/internal/processor/otel/metadata.go
@@ -169,6 +169,11 @@ func translateResourceMetadata(resource pcommon.Resource, out *model.APMEvent) {
 		out.Host.OS.Type = "ios"
 	}
 
+	if out.Service.Runtime.Name == nil && (out.Host.OS.Name == "Android" || out.Host.OS.Name == "iOS") {
+		out.Service.Runtime.Name = out.Host.OS.Name
+		out.Service.Runtime.Version = out.Host.OS.Version
+	}
+
 	if strings.HasPrefix(exporterVersion, "Jaeger") {
 		// version is of format `Jaeger-<agentlanguage>-<version>`, e.g. `Jaeger-Go-2.20.0`
 		const nVersionParts = 3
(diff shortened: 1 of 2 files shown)


axw:
Can the agents send `process.runtime.name` and `process.runtime.version` fields, per https://opentelemetry.io/docs/reference/specification/resource/semantic_conventions/process/#process-runtimes?

Is "Android" the correct runtime name? OTel says that for Java runtimes:

> process.runtime.name - Fill in the value of java.runtime.name as is
> process.runtime.version - Fill in the value of java.runtime.version as is

I'm not sure what those values will be for ART - the OTel docs give "Android Runtime" an an example.

AlexanderWert:
> Can the agents send `process.runtime.name` and `process.runtime.version` fields, per https://opentelemetry.io/docs/reference/specification/resource/semantic_conventions/process/#process-runtimes?
> 
> Is "Android" the correct runtime name? OTel says that for Java runtimes:
> 
> > process.runtime.name - Fill in the value of java.runtime.name as is
> > process.runtime.version - Fill in the value of java.runtime.version as is
> 
> I'm not sure what those values will be for ART - the OTel docs give "Android Runtime" an an example.

* For iOS there is no meaningful "runtime" it's just iOS with its version. 
* For Android we have the OS (Android), the ART and the Android API version. Though, there is a 1:1 mapping between all of them. Cesar commented that usually Android devs are interested in the Android version and the rest (ART, JVM, Android API) are coupled to that. 

We can also set the `process.runtime.*` fields in the agents, just thought if we would set it anyways to the same values as the `host.os.fields` APM server might be a better place for it. But don't have a strong opinion, fine with both ways. WDYT?

AlexanderWert:
agreed to set this info directly in the mobile agents

PR #9020, opened 2022-08-31
perf: replace samples map with a slice

kruskall:
## Motivation/summary

Reduce allocations by replacing the samples map with a slice
with perfect size.

Benchmark on heavy.ndjson:

```
name                       old time/op    new time/op    delta
ReadBatch/heavy.ndjson-16    11.2ms ± 9%    10.5ms ± 8%   -5.85%  (p=0.029 n=10+10)

name                       old speed      new speed      delta
ReadBatch/heavy.ndjson-16  35.7MB/s ± 9%  37.9MB/s ± 8%   +6.20%  (p=0.029 n=10+10)

name                       old alloc/op   new alloc/op   delta
ReadBatch/heavy.ndjson-16    1.28MB ± 0%    0.83MB ± 0%  -35.09%  (p=0.000 n=10+10)

name                       old allocs/op  new allocs/op  delta
ReadBatch/heavy.ndjson-16     23.1k ± 0%     22.6k ± 0%   -2.09%  (p=0.000 n=10+9)
```

## Checklist


internal/model/metricset.go
```diff
@@ -70,6 +70,9 @@ type MetricsetSample struct {
 	// If Type is unspecified or invalid, it will be ignored.
 	Type MetricType
 
+	// Name holds the name of the metric type.
```
axw, line 73:
```suggestion
	// Name holds the metric name.
	//
	// Name is required.
```

axw (commented):
Looks great! There's a couple of unrelated test failures (I'm looking into the service metrics one now), but there's one that seems likely to be related: `processor/otel.TestConsumeMetrics_JVM`

kruskall:
Flaky tests should be fixed 🏓

Issue #8957, opened 2022-08-24
Process and map dedicated device.* fields from mobile agents


simitt:
Elastic mobile agents (Android and iOS) are built on OpenTelemetry, and therefore use the otel protocol, instead of the APM Server Intake API. https://github.com/elastic/apm/blob/main/specs/agents/mobile/README.md lists mobile agent specific fields and their desired mappings. A majority of the missing fields have been added to the APM Server processing in https://github.com/elastic/apm-server/issues/8755. 

The remaining part is mapping the  `device.*` fields. Following device related fields should be accepted and mapped by the APM Server, according to [ECS spec](https://github.com/elastic/ecs/blob/main/schemas/device.yml): 

* `device.id`
* `device.model.identifier`
* `device.manufacturer` (if provided directly by agents)

## Out of Scope
* Mapping of `device.model.name` (and potentially `device.manufacturer` where it cannot directly be retrieved by the agents) via leveraging an ES built in pipeline (https://github.com/elastic/apm-server/issues/4879).

PR #9021, opened 2022-08-31
feat: process and map dedicated device.* fields

kruskall:
## Checklist

## Related issues

Closes https://github.com/elastic/apm-server/issues/8957


internal/model/device.go
```diff
@@ -0,0 +1,57 @@
...
+	ID string
+
+	// Architecture holds the host machine architecture.
+	Model Model
+
+	// Manufacturer the vendor name of the device manufacturer.
+	Manufacturer string
+}
+
+type Model struct {
```
axw, line 35:
```suggestion
type DeviceModel struct {
```

Calling this DeviceModel will make the name a bit more meaningful when looking at it out of the context of this file, e.g. by looking at godoc.

internal/model/device.go
```diff
@@ -0,0 +1,57 @@
...
+
+import (
+	"github.com/elastic/elastic-agent-libs/mapstr"
+)
+
+type Device struct {
+	// ID holds the unique identifier of a device.
+	ID string
+
+	// Architecture holds the host machine architecture.
```
axw, line 28:
```suggestion
	// Model holds information about the device model.
```

internal/model/device.go
```diff
@@ -0,0 +1,57 @@
...
+	Name string
+
+	// Identifier holds the machine readable identifier of the device model.
+	Identifier string
+}
+
+func (d *Device) fields() mapstr.M {
+	if d == nil {
+		return nil
+	}
```
axw, line 46:
```suggestion
	if d == nil {
		return nil
	}
```

Do we need it? APMEvent.Device is not a pointer.

axw (changes requested):
Looks great! We should also add these fields to `ecs.yml` in `app_logs`, `app_metrics`, `error_logs`, and `traces` data streams.

I forgot about that for the previous PR, so can you please also add the fields added in https://github.com/elastic/apm-server/pull/8784/files while you're at it?

internal/model/device.go
kruskall: Good point! Updated!

internal/model/device.go
kruskall: Thank you for spotting this! 🙇

internal/model/device.go
kruskall: I couldn't see a consistent usage of nil checks in other `fields` methods. I guess it's fine to remove it.

kruskall:
> ooks great! We should also add these fields to `ecs.yml` in `app_logs`, `app_metrics`, `error_logs`, and `traces` data streams.
> 
> I forgot about that for the previous PR, so can you please also add the fields added in https://github.com/elastic/apm-server/pull/8784/files while you're at it?

@axw sory, I'm a bit lost here 😓 

`build/apmpackage` is failing but the definition in `ecs.yml` seems to be correct. I see the apm-server is using an older version of https://github.com/elastic/elastic-package/ but updating yields the same result.

axw:
@kruskall I think the issue is that `device.*` fields aren't in any released version of ECS. We have a couple of options:
 - update https://github.com/elastic/apm-server/blob/bd0969db37e6f6200b496cf2fa10b3d3239323ff/apmpackage/apm/_dev/build/build.yml#L4 to use `git@main`
 - write out those the fields by hand

I think it's probably safe to use the main branch. Maybe try that, and diff the build package before and after, to check that none of the fields types have changed incompatibly.

kruskall:
> I think the issue is that `device.*` fields aren't in any released version of ECS. We have a couple of options:
> 
>     * update https://github.com/elastic/apm-server/blob/bd0969db37e6f6200b496cf2fa10b3d3239323ff/apmpackage/apm/_dev/build/build.yml#L4

Thank you for the help and guidance! 🙇 Seems to be fixed!

axw (approved):
LGTM!

PR #9035, opened 2022-09-04
perf: replace strings.Split with strings.Cut in span model mapping

kruskall:
## Motivation/summary

memprofile shows a few allocations when splitting the string
with strings.Split.
Replace the method with strings.Cut.

## Checklist


axw (approved):
Nice, looks good. This splitting is also done in modeldecoder/rumv3, can you please update that too?

axw (approved):

PR #9068, opened 2022-09-08
Add logs support to events intake API

### LINKED ISSUE #8757: Add Logs Intake API (POC)
The AWS lambda extension is already forwarding metrics and trace signals to the APM Server. Now the extension is adding support for collecting logs (https://github.com/elastic/apm-aws-lambda/issues/256). The APM server should be extended to also support proxying these logs to Elasticsearch. 

This issue can be used as meta issue and be further split up, if necessary. 

The POC should include:
* Extend the http server to support receiving logs events (provide a `logs` endpoint or add logs event support to existing event intake API).
* Add JSON spec for the logs format
* Parse & process log events
* Add a dedicated data stream, including mappings, ILM policy and ingest pipeline to the `apmpackage`. 
* Ship logs into one common `logs-apm.*` ES data stream for this POC.

## Expected outcome
* have a working prototype
* identify further work items and potential bottlenecks or issues.

## Out of Scope
* Define final data stream strategy. 
* Full end-to-end test coverage

lahsivjar:
## Motivation/summary

Adds support for `log` event type to the intake API.

## Checklist

## How to test these changes

1. Create a request body for the `log` event type:
    ```json
    {"metadata": {"user": {"domain": "ldap://abc", "id": "123", "email": "s@test.com", "username": "john"}, "process": {"ppid": 6789, "pid": 1234,"argv": ["node", "server.js"], "title": "node"}, "system": {"platform": "darwin", "hostname": "prod1.example.com", "architecture": "x64", "container": {"id": "container-id"}, "kubernetes": {"namespace": "namespace1", "pod": {"uid": "pod-uid", "name": "pod-name"}, "node": {"name": "node-name"}}}, "labels": {"tag1": "label1"}, "service": {"name": "testsvc", "language": {"version": "8", "name": "ecmascript"}, "agent": {"version": "3.14.0", "name": "elastic-node"}, "environment": "staging", "framework": {"version": "1.2.3", "name": "Express"}, "version": "5.1.3", "runtime": {"version": "8.0.0", "name": "node"}},"cloud":{"account":{"id":"account_id","name":"account_name"},"availability_zone":"cloud_availability_zone","instance":{"id":"instance_id","name":"instance_name"},"machine":{"type":"machine_type"},"project":{"id":"project_id","name":"project_name"},"provider":"cloud_provider","region":"cloud_region","service":{"name":"lambda"}}}}
    {"log": {"message": "test message"}}
    {"log": {"message": "test message 2", "timestamp": 1662616049000000}}
    {"log": {"message": "test log message with faas", "@timestamp": 1662616049000000,"faas":{"coldstart":true,"execution":"6f7f0961f83442118a7af6fe80b88d56","id":"arn:aws:lambda:us-east-2:123456789012:function:custom-runtime"}}}
    ```
2. Send a curl request with the above request body:
    ```bash
    curl -i -H "Content-type: application/x-ndjson" --data-binary @<request_body_path> http://localhost:8200/intake/v2/events
    ```

## Related issues

Closes #8757


docs/spec/v2/log.json
```diff
@@ -0,0 +1,839 @@
...
+      "type": "string"
+    },
+    "severity": {
+      "description": "Severity represents the numreic severity of the logs as advertised by the source. Meaning behind severity is upto the implementer.",
+      "type": [
+        "null",
+        "integer"
+      ]
+    },
+    "span_id": {
```
felixbarny, line 812: Note that unlike traces and spans, logs are expected to be already in ECS format, as we'll use our ECS logging libraries to convert log events to JSON. So it shouldn't be `span_id` but `span.id`. This also applies to `timestamp` -> `@timestamp`, `trace_id` -> `trace.id` and others.

I think it will end up being much easier to not even have a dedicated spec for `log`. Just forward everything agents send to Elasticserach.

felixbarny: Having said that, the global metadata ofc needs to be converted to ECS.

What I had in mind is that APM Server would just send a doc like
```
{
  ... global metadata
  "message": "<whatever the agent has sent>"
}
```

And then use an ingest node pipeline to do a JSON decoding of the `message` field that merges the fields into the root of the doc.

Dynamic fields become an issue then. If we want to disallow dynamic fields in the first iteration, we can add `dynamic: false` to the mappings. However, we could also just add `dynamic: runtime`. This makes the fields searchable but avoids  (some) mapping issues related to dynamic fields and can save space and index time for fields that are not commonly used in searches.
Other mechanisms to avoid issues with dynamic fields are to enable `ignore_malformed` for the whole data stream and to enable `subobjects: true` on the root level once https://github.com/elastic/elasticsearch/issues/88934 has been resolved.

lahsivjar: > maybe something like raw.

Or we can put everything else under `labels` as suggested by Andrew in the slack conversation.

axw: IIANM, for Lambda all we get is a timestamp and a string for function logs. Is that right? If so then I agree with the suggestions above to just send `@timestamp` and `message`.

For the first cut, I would suggest we just store `message` as it is received. Then we can follow up with what Felix suggests, and attempt to decode the message as JSON into the root, and use `dynamic: runtime`.

internal/model/modeldecoder/v2/model.go
```diff
@@ -951,6 +956,31 @@ type transaction struct {
...
+	// Message logged as part of the log. In case a parameterized message is
+	// captured, Message should contain the same information, but with any placeholders
+	// being replaced.
+	Message nullable.String `json:"message"`
+	// Severity represents the numeric severity of the logs as advertised by the
+	// source. Meaning behind severity is upto the implementer.
+	Severity nullable.Int `json:"severity"`
+	// FAAS holds fields related to Function as a Service events.
+	FAAS faas `json:"faas"`
+}
```
axw, line 982: Should we trim this back to just `@timestamp`, `message`, and and `faas`? IIANM that's all we need for now?

lahsivjar: > Should we trim this back to just @timestamp, message, and and faas?

Yes, this sounds good. Do we also need `Context`? I am assuming that global metadata will have context fields that will require to be converted to ECS format.

axw: I don't think we need Context for logs, at least not right now. Global metadata should be sufficient, and does indeed hold data that will be translated to ECS, such as `host.name`.

lahsivjar: Updated the code to have only the required fields.

internal/model/log.go
```diff
@@ -34,10 +34,13 @@ var (
 type Log struct {
 	// Level holds the log level of the log event.
 	Level string
+	// LoggerName holds the name of the logger for the log event.
+	LoggerName string
```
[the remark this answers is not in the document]
lahsivjar: Removed `LoggerName` since it is not required for our current usecase.

axw (commented):
Looks great! Should we add a small system test?

lahsivjar:
> Looks great! Should we add a small system test?

This is done.

axw (approved):
Nice work!

changelogs/head.asciidoc
```diff
@@ -23,6 +23,7 @@ https://github.com/elastic/apm-server/compare/8.5\...main[View commits]
   When set to `true`, the batch will be processed asynchronously if the APM Server can
   service the incoming request immediately, otherwise a "503 queue is full" error will
   be returned {pull}8979[8979]
+- [EXPERIMENTAL] Add logs support to v2 intake API {pull}9068[9068]
```
simitt, line 26: Better to follow the convention: `- experimental:[] Add logs support to v2 intake API {pull}9068[9068]`

It will let it appear as `preview` automatically (see [8.0 release notes](https://www.elastic.co/guide/en/apm/guide/current/release-notes-8.0.html))

Screenshot 2022-09-09 at 10 26 46

simitt (approved):

lahsivjar:
@axw I missed adding `faas` fields to the logs datastream. I have added it now, please give it a quick look.

changelogs/head.asciidoc
lahsivjar: Thanks, TIL. I have updated it.

apmpackage/apm/changelog.yml
```diff
@@ -1,11 +1,11 @@
 # newer versions go on top
 #
 # change type can be one of: enhancement, bugfix, breaking-change
-- version: "generated"
+- version: "8.5.0"
```
axw, line 4: Not sure where I commented, commenting again here: this one should be left as "generated" - it will be replaced by a script. Just update the placeholder entry below.

axw (approved):
Good catch! Just one thing about the package version, otherwise LGTM.

apmpackage/apm/changelog.yml
lahsivjar: Updated, thanks.

felixbarny:
What do you think about also adding the fields from the ECS logging spec? https://github.com/elastic/ecs-logging/blob/main/spec/spec.json This way, we could use this as the foundation for https://github.com/elastic/apm-dev/issues/833 and merge https://github.com/elastic/apm-agent-java/pull/2694 as an experimental feature.

Dynamic mappings can come later.

lahsivjar:
@felixbarny Thanks for the message. I have created #9100 to handle this, feel free to add any more information to it if required. I will try to push the required changes for 8.5.

marclop:
Verified with 8.5.0 BC1:

```console
$ curl -i -H "Authorization: Bearer 123" -H "Content-type: application/x-ndjson" --data-binary @testdata/intake-v2/logs.ndjson https://5fa0c22c644145439c4c4f9daa176b05.apm.us-west2.gcp.elastic-cloud.com:443/intake/v2/events
HTTP/2 202
date: Tue, 27 Sep 2022 07:21:56 GMT
x-cloud-request-id: AprFqdBaR7akPHcDaGQT-Q
x-content-type-options: nosniff
x-found-handling-cluster: 5fa0c22c644145439c4c4f9daa176b05
x-found-handling-instance: instance-0000000000
content-length: 0
```

image

I couldn't verify the log message with `faas` fields since that change is in the unreleased package which isn't being used for ESS atm. I will retest as soon as that's available 

image

marclop:
Verified with the latest `8.5.0` BC:

image

PR #9096, opened 2022-09-12
sampling: call RunValueLogGC in a loop

axw:
## Motivation/summary

`RunValueLogGC` returns `ErrNoRewrite` when there are no changes to be made for GC, and should be called in a loop until it returns an error. Doing this should lead to value logs being garbage collected more quickly, as otherwise we'll wait for another tick before repeating.

See https://github.com/dgraph-io/dgraph/blob/832ebb9766f1732186eaecbf3792aff9e3766c6c/x/x.go#L1089-L1127 for how dgraph itself uses the method.

:tipping_hand_man: :tophat: @marclop 

## Checklist

## How to test these changes

N/A

## Related issues

None

### DIFF
--- a/x-pack/apm-server/sampling/processor.go
+++ b/x-pack/apm-server/sampling/processor.go
@@ -387,7 +387,13 @@ func (p *Processor) Run() error {
 				return ctx.Err()
 			case <-ticker.C:
 				const discardRatio = 0.5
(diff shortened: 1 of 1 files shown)


marclop (approved):
Thanks for making the change! I had forgotten about this after I worked on the soft limit 😳 .

Issue #7846, opened 2022-04-11
Benchmarking 2.0 infrastructure and automation


marclop:
## Description

The new benchmarking framework will take a slightly different route than our current approach, namely it will look to leverage either ESS (preferred) or an on-demand ECE environment (ecetest) to run nightly periodic benchmarks that allow us to track the APM Server performance.

The lowest effort would be to use a region in ESS where we can benchmark the APM Server's throughput, collect the results and index them to a long-lived remote Elasticsearch cluster with [`gobench`](https://github.com/elastic/gobench). If there are unforeseen limitations when using ESS to run the benchmarks, we can then look into spinning up an on-demand ECE environment, but that is much costlier, in terms of time, resources and monetary cost.

## Considerations

- Keep the benchmarks run by hey-apm for a while and only decommission those after we're happy with the new approach.
- Run `apmbench` in a machine that is appropriate for the work and is as close as possible to the workload (same CSP region).
- The hardware profile must be able to handle a high level of concurrency since we'll be looking to run `apmbench` with a medium to high number of agents, and decent network performance.

### Docker image tag

Since the Elastic Stack and APM Server will be running in ESS, the software must be packaged in docker images, it is out of scope for this issue to build these images, but to reduce the risk of running the benchmarks against an upstream version that doesn't completely work, we should have some guarantees in place and a vetting process for the "latest" version.

Since we already have a workflow that updates each of the APM Server's active branches docker images, we could rely on the docker image tags that are used in our `docker-compose.yml` file and specify the current image's tag as the docker image to use in `<elasticsearch|kibana|apm>.config.docker_image` when creating the ESS deployment. [See the Terraform provider acceptance test](https://github.com/elastic/terraform-provider-ec/blob/master/ec/acc/testdata/deployment_docker_image_override.tf#L37) that uses `docker_image`.

### Deployment lifecycle

The most cost effective and efficient approach is to create a new deployment in ESS and a new VM in the same region with the desired hardware profile for the apmbench runner and upload the credentials for apmbench to connect to the deployment. After the benchmarks have been run, and the results uploaded to a persistent deployment where we'd store them, the deployment and apmbench vm should be town down, to cut down costs.

The terraform configuration for the benchmark deployment could live in the APM Server repo and it could also be used for APM Server developers when there are changes in the APM Server that wish to be benchmarked, a limitation, however, would be that a cloud docker image would need to be built and uploaded to allow the testing to take place.

## Automation work

[...]

PR #9099, opened 2022-09-12
chore: CI: Benchmark slack notifications

pazone:
## Motivation/summary

Send slack messages to apm-server channel about benchmark results

https://github.com/elastic/apm-server/issues/7846

### DIFF
--- a/.ci/benchmarks.groovy
+++ b/.ci/benchmarks.groovy
@@ -9,11 +9,14 @@ pipeline {
     AWS_ACCOUNT_SECRET = 'secret/observability-team/ci/elastic-observability-aws-account-auth'
     EC_KEY_SECRET = 'secret/observability-team/ci/elastic-cloud/observability-pro'
     BENCHMARK_ES_SECRET = 'secret/apm-team/ci/benchmark-cloud'
+    BENCHMARK_KIBANA_SECRET = 'secret/observability-team/ci/apm-benchmark-kibana'
+    PNG_REPORT_FILE = 'out.png'
     TERRAFORM_VERSION = '1.1.9'
 
     CREATED_DATE = "${new Date().getTime()}"
     JOB_GCS_BUCKET_STASH = 'apm-ci-temp'
     JOB_GCS_CREDENTIALS = 'apm-ci-gcs-plugin'
+    SLACK_CHANNEL = "#apm-server"
   }
 
   options {
@@ -51,7 +54,6 @@ pipeline {
         TF_VAR_ENVIRONMENT= 'ci'
         TF_VAR_BRANCH = "${env.BRANCH_NAME.toLowerCase().replaceAll('[^a-z0-9-]', '-')}"
         TF_VAR_REPO = "${REPO}"
-
         GOBENCH_TAGS = "branch=${BRANCH_NAME},commit=${GIT_BASE_COMMIT},pr=${CHANGE_ID},target_branch=${CHANGE_TARGET}"
       }
       steps {
@@ -67,6 +69,7 @@ pipeline {
               }
             }
           }
+          downloadPNGReport()
         }
       }
       post {
@@ -86,6 +89,14 @@ pipeline {
             }
           }
         }
+        success {
+          dir("${BASE_DIR}") {            
+            sendSlackReportSuccessMessage()
+          }
+        }
+        failure {
+          notifyBuildResult(slackNotify: true, slackComment: true)
+        }  
       }
     }
   }
@@ -108,5 +119,53 @@ def withESBenchmarkEnv(Closure body) {
             'url': 'GOBENCH_HOST', 
             'password': 'GOBENCH_PASSWORD'] ) {
     body()
+  }
+}
+
+def downloadPNGReport() {
+  withSecretVault(
+      secret: "${BENCHMARK_KIBANA_SECRET}", 
+      data: ['user': 'KIBANA_USER', 
+            'password': 'KIBANA_PASSWORD', 
+            'kibana_url': 'KIBANA_ENDPOINT'] ) {
+    sh(label: 'Run unit tests', script: '.ci/scripts/download-png-from-kibana.sh $KIBANA_ENDPOINT $KIBANA_USER $KIBANA_PASSWORD $PNG_REPORT_FILE')
+    archiveArtifacts(
+      allowEmptyArchive: false,
+      artifacts: "${env.PNG_REPORT_FILE}",
+      defaultExcludes: false
+    )
+  }
+}
+
+def sendSlackReportSuccessMessage() {
+  withSecretVault(
+      secret: "${BENCHMARK_KIBANA_SECRET}", 
+      data: ['kibana_dashboard_url': 'KIBANA_DASHBOARD_URL'] ) {
+    slackMessageBlocks = [
+      [
+        "type": "section",
+        "text": [
+          "type": "mrkdwn",
+          "text": "Nightly benchmarks succesfully executed\n\n <${env.BUILD_URL}|Jenkins Build ${env.BUILD_DISPLAY_NAME}>"
+        ]
+      ],
+      [
+        "type": "divider"
+      ],
+      [
+        "type": "image",
+        "image_url": "${env.BUILD_URL}artifact/${env.PNG_REPORT_FILE}",
+        "alt_text": "image is not downloaded"
+      ],
+      [
+        "type": "section",
+        "text": [
+          "type": "mrkdwn",
+          "text": "<${env.KIBANA_DASHBOARD_URL}|Kibana Dashboard>"
+        ]        
+      ]
+    ]
+
+    slackSend(channel: "${env.SLACK_CHANNEL}", blocks: slackMessageBlocks)  
   }
 }
--- a/.ci/scripts/download-png-from-kibana.sh
+++ b/.ci/scripts/download-png-from-kibana.sh
@@ -1,0 +1,30 @@
+#!/usr/bin/env bash
+
+kibana_host=$1
+kibana_user=$2
+kibana_pwd=$3
+png_output_file=${4:-'out.png'}
+table_width_px=${5:-850}
+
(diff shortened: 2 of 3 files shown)


.ci/benchmarks.groovy
```diff
@@ -9,11 +9,16 @@ pipeline {
     AWS_ACCOUNT_SECRET = 'secret/observability-team/ci/elastic-observability-aws-account-auth'
     EC_KEY_SECRET = 'secret/observability-team/ci/elastic-cloud/observability-pro'
     BENCHMARK_ES_SECRET = 'secret/apm-team/ci/benchmark-cloud'
+    KIBANA_SECRET = 'secret/k8s/elastic-apps/oauth-proxy/prod/domains/observability-benchmarks.elastic.dev'
+    PNG_REPORT_FILE = 'out.png'
+    KIBANA_DASHBOARD_URL = 'https://observability-benchmarks.elastic.dev/app/r/s/afdFh'
```
v1v, line 14: I'd not expose the Kibana URL but use the Vault secret in favour of more dynamism

.ci/benchmarks.groovy
```diff
@@ -85,6 +93,14 @@ pipeline {
             }
           }
         }
+        success {
+          dir("${BASE_DIR}") {            
+            sendSlackReportSuccessMessage()
+          }
+        }
+        failure {
+          notifyBuildResult(slackComment: true, slackChannel: ${env.SLACK_CHANNEL})
```
v1v, line 102: typo, use double quotes for the interpolation

```suggestion
          notifyBuildResult(slackComment: true, slackChannel: "${env.SLACK_CHANNEL})"
```

or 

```suggestion
          notifyBuildResult(slackComment: true, slackChannel: env.SLACK_CHANNEL)
```

Although, SLACK_CHANNEL is the default behaviour, see https://github.com/elastic/apm-pipeline-library/blob/0b1b3269e6a25bf50e5d78ed6102894c7f712fce/vars/notifyBuildResult.groovy#L66


```suggestion
          notifyBuildResult(slackComment: true)
```

The idea in this section is to send a slack message with the build failure only, correct?

.ci/benchmarks.groovy
```diff
@@ -109,3 +125,47 @@ def withESBenchmarkEnv(Closure body) {
...
   }
 }
+
+def downloadPNGReport() {
+  withSecretVault(
+      secret: "${KIBANA_SECRET}", 
+      data: ['username': 'KIBANA_USER', 
+            'password': 'KIBANA_PASSWORD', 
+            'endpoint': 'KIBANA_ENDPOINT'] ) {
+    sh(label: 'Run unit tests', script: ".ci/scripts/download-png-from-kibana.sh $KIBANA_ENDPOINT $KIBANA_USER $KIBANA_PASSWORD ${env.PNG_REPORT_FILE}")
```
v1v, line 135: To play safe, I'd say those env variables are implicitly used within the script, to avoid any weirdness from the security point of view in the deprecated Blue Ocean view. See https://issues.jenkins.io/browse/JENKINS-59214, or use the single quotes as proposed in https://www.jenkins.io/doc/book/pipeline/jenkinsfile/#interpolation-of-sensitive-environment-variables

```suggestion
    sh(label: 'Run unit tests', script: '.ci/scripts/download-png-from-kibana.sh $KIBANA_ENDPOINT $KIBANA_USER $KIBANA_PASSWORD $PNG_REPORT_FILE')
```

or


```suggestion
    sh(label: 'Run unit tests', script: '.ci/scripts/download-png-from-kibana.sh')
```

and change the `.ci/scripts/download-png-from-kibana.sh` to use those env variables

.ci/benchmarks.groovy
```diff
@@ -109,3 +125,47 @@ def withESBenchmarkEnv(Closure body) {
...
+      "type": "section",
+      "text": [
+        "type": "mrkdwn",
+        "text": "Nightly benchmarks succesfully executed\n\n [Jenkins Build ${env.BUILD_DISPLAY_NAME}](${env.BUILD_URL})"
+      ]
+    ],
+    [
+		  "type": "divider"
+	  ],
+    [
```
v1v, line 156: NIT: ident

.ci/benchmarks.groovy
```diff
@@ -109,3 +125,47 @@ def withESBenchmarkEnv(Closure body) {
...
+      ],
+      "accessory": [
+        "type": "image",
+        "image_url": "${env.BUILD_URL}artifact/${env.PNG_REPORT_FILE}",
+        "alt_text": "No image for now"
+      ]
+    ]
+  ]
+
+  slackSend(channel: ${env.SLACK_CHANNEL}, blocks: slackMessageBlocks)
```
v1v, line 170: interpolation is required when using `${env.}` or use directly the env map;

```suggestion
  slackSend(channel: "${env.SLACK_CHANNEL}", blocks: slackMessageBlocks)  
```

or 

```suggestion
  slackSend(channel: env.SLACK_CHANNEL, blocks: slackMessageBlocks)  
```

.ci/benchmarks.groovy
```diff
@@ -9,11 +9,16 @@ pipeline {
     AWS_ACCOUNT_SECRET = 'secret/observability-team/ci/elastic-observability-aws-account-auth'
     EC_KEY_SECRET = 'secret/observability-team/ci/elastic-cloud/observability-pro'
     BENCHMARK_ES_SECRET = 'secret/apm-team/ci/benchmark-cloud'
+    KIBANA_SECRET = 'secret/k8s/elastic-apps/oauth-proxy/prod/domains/observability-benchmarks.elastic.dev'
```
v1v, line 12: What do you think if this secret is not exposed but a new one is created in `secret/observability-team/ci`?
Maybe `secret/observability-team/ci/benchmark-cloud-reporting`, in addition a new elasticsearh user is created with the right permissions so we don't expose the superuser within the CI?

v1v (commented):
I added some questions for the Vault secrets and securise the Elastic Deployment with a new  user with the right permissions to access to the required entrypoints.

By doing so, we can avoid any security concerns, we manage the vault secrets and we don't expose the URLs

.ci/benchmarks.groovy
```diff
@@ -85,6 +96,14 @@ pipeline {
             }
           }
         }
+        success {
+          dir("${BASE_DIR}") {            
+            sendSlackReportSuccessMessage()
+          }
+        }
+        failure {
+          notifyBuildResult(slackComment: true)
```
v1v, line 105: default behaviour won't notify slack messages for PRs, https://github.com/elastic/apm-pipeline-library/blob/0b1b3269e6a25bf50e5d78ed6102894c7f712fce/vars/notifyBuildResult.txt#L36

```suggestion
          notifyBuildResult(slackComment: true, slackNotify: true)
```

.ci/benchmarks.groovy
```diff
@@ -8,10 +8,8 @@ pipeline {
     BASE_DIR = "src/github.com/elastic/${env.REPO}"
     AWS_ACCOUNT_SECRET = 'secret/observability-team/ci/elastic-observability-aws-account-auth'
     EC_KEY_SECRET = 'secret/observability-team/ci/elastic-cloud/observability-pro'
-    BENCHMARK_ES_SECRET = 'secret/apm-team/ci/benchmark-cloud'
-    KIBANA_SECRET = 'secret/k8s/elastic-apps/oauth-proxy/prod/domains/observability-benchmarks.elastic.dev'
+    BENCHMARK_ES_SECRET = 'secret/apm-team/ci/benchmark-cloud'
```
v1v, line 11: I'd say to use `secret/observability-team/ci/benchmark-cloud` instead, 

`secret/apm-team/ci/benchmark-cloud` is deprecated and all the references in our docs point to the above one. 


```
status                  DEPRECATED, use secret/observability-team/ci/benchmark-cloud instead
```

I'm still concerned with the existing user/pass, is it possible to create a new user with only the right permissions to query the `kibana_dashboard_url` and `png` generation, in favour of reducing the chances of any security disclosure?

.ci/benchmarks.groovy
```diff
@@ -109,3 +120,51 @@ def withESBenchmarkEnv(Closure body) {
...
+          "text": "Nightly benchmarks succesfully executed\n\n <${env.BUILD_URL}|Jenkins Build ${env.BUILD_DISPLAY_NAME}>"
+        ]
+      ],
+      [
+        "type": "divider"
+      ],
+      [
+        "type": "image",
+        "image_url": "${env.BUILD_URL}artifact/${env.PNG_REPORT_FILE}",
+        "alt_text": "image is not downloaded"
```
v1v, line 157: NIT: what's the reason for this message?

.ci/scripts/download-png-from-kibana.sh
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env bash
+
+kibana_host=$1
+kibana_user=$2
+kibana_pwd=$3
+png_output_file=${4:-'out.png'}
+table_width_px=${5:-850}
+
+png_url_path=$(curl -XPOST --silent -L -u "$kibana_user:$kibana_pwd" -H 'kbn-xsrf: true' "$kibana_host/api/reporting/generate/pngV2?jobParams=%28browserTimezone%3AEurope%2FBerlin%2Clayout%3A%28dimensions%3A%28height%3A0%2Cwidth%3A$table_width_px%29%2Cid%3Apreserve_layout%29%2ClocatorParams%3A%28id%3ADASHBOARD_APP_LOCATOR%2Cparams%3A%28dashboardId%3Aa5bc8390-2f8e-11ed-a369-052d8245fa04%2CpreserveSavedFilters%3A%21t%2CtimeRange%3A%28from%3Anow-30d%2Cto%3Anow%29%2CuseHash%3A%21f%2CviewMode%3Aview%29%2Cversion%3A%278.3.2%27%29%2CobjectType%3Adashboard%2Ctitle%3Aapp_bench_diff_shifts_slack%2Cversion%3A%278.3.2%27%29" \
```
marclop, line 9: It looks like the previous dashboard has been modified.

Where we had markdown objects with detailed explanations on what to look for before, they are now gone. What happened? Would you mind bringing those back?

marclop (commented):
It currently looks like we're just doing a time shift and comparing a day results against a week ago (time shifted by `1w`). This is exactly what we were doing before and has tended to not be a reliable way to compare, since a single day outlier will result in a significant diff.

I think a better approach would be to aggregate the last 7 days and compare that aggregation with the previous 7 days. Do folks in @elastic/apm-server have opinions on what we should be comparing?

.ci/scripts/download-png-from-kibana.sh
pazone: The idea is to grab the data from the existing table. It allows us to add/modify any metrics to comparison without changing any code. 
However it turned our that CSV reporting is now broken in Kibana. As soon as it fixed I going to use csv to create detailed comparison with markdown. 
Temporarily use the png image to share the results. It is less representative but the report can be adjusted on the fly.

testing/benchmark/README.md
```diff
@@ -101,3 +101,9 @@ do that in `terraform.tfvars.example`.
 
 `make cleanup-elasticsearch` will delete all the APM data streams. This may be useful in case you'd like to re-run
 the benchmarks without destroying the deployment.
+
+### Slack reporting
+Reporting data is taken from the [app_bench_diff_shifts_slack](https://observability-benchmarks.elastic.dev/app/dashboards#/view/a5bc8390-2f8e-11ed-a369-052d8245fa04?_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:now-30d,to:now)))
```
v1v, line 106: @pazone  , can we avoid exposing those details publicly?

PR #9105, opened 2022-09-13
benchmark: Use GCP Belgium (CPU optimized profile)

marclop:
## Motivation/summary

Updates the default region and deployment template so that we make use of the Belgium CFT region which does have the CPU optimized hardware profile. This should give us a better benchmarking environment where the APM Server isn't being throttled all the time.

## Checklist

## How to test these changes

N/A

## Related issues

Use the right hardware profile for benchmarks.

### DIFF
--- a/testing/benchmark/variables.tf
+++ b/testing/benchmark/variables.tf
@@ -8,13 +8,13 @@ variable "user_name" {
 ## Deployment configuration
 
 variable "ess_region" {
-  default     = "gcp-us-west2"
-  description = "Optional ESS region where the deployment will be created. Defaults to gcp-us-west2"
+  default     = "gcp-europe-west1"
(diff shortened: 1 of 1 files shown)


testing/benchmark/variables.tf
```diff
@@ -83,8 +83,8 @@ variable "apm_shards" {
 ## Worker configuraiton
 
 variable "worker_region" {
-  default     = "us-west-2"
-  description = "Optional ESS region where the deployment will be created. Defaults to us-west-2 (AWS)"
+  default     = "eu-west-3"
```
marclop, line 86: Paris is the closest AWS region from GCP Belgium. The network latency should be negligible

axw (approved):

PR #9127, opened 2022-09-15
Profiling: improve storing interpreter frame filenames

inge4pres:
## Motivation/summary

To save network bandwidth in the profiling agent, filenames for interpreted frames are only sent on the first frame encounter. Thereafter the filenames will be empty (or refreshed eventually).

The collector so far stores each frame whenever it is reported: this replaces an existing record and thus often leads to records with empty source filenames. In turn, Kibana often fails to fetch and display the filenames of interpreted frames.

To prevent this issue, the PR adds an LRU cache for source filenames (key is the source FileID) to enrich frame metadata when it's source filename is empty.

The cache is local to every apm-server replica and has been sized to 131072 entries: each entry is made of a key of 2 uint64 and a value that is a variable length `string`. On Linux a filename has max 255 chars, therefore the theoretical maximum size for each entry in the cache is 259 bytes, resulting in a theoretical maximum total cache size of ~32 MiB.

## Checklist

- [ ] Update [CHANGELOG.asciidoc](https://github.com/elastic/apm-server/blob/main/CHANGELOG.asciidoc) N/A

- Is it observable through the addition of either **logging** or **metrics**? 🚫 
- Is its use being published in **telemetry** to enable product improvement? 🚫 

## How to test these changes

Run the local stack as usual, and fleetctl to configure profiling intake.
Running with `runapm` you'll be able to see frames are inserted but not overwritten (check the counters in the expvar metrics)

## Related issues

#9005


rockdaboot (approved):
LGTM

athre0z (approved):

axw (approved):

PR #9177, opened 2022-09-23
Added android for client IP mapping

AlexanderWert:
Adds the android agent to the logic of setting client IP in client metadata.


axw (approved):
LGTM! Thanks :)

PR #9207, opened 2022-09-27
profiling: fix off-by-one error in run-length encoder

inge4pres:
Signed-off-by: inge4pres <francesco.gualazzi@elastic.co>

## Motivation/summary

The Profiling collector compresses an array of frame types using run-length encoding and then writes the result to Elasticsearch. The compressed result is later read by Kibana and decoded.

Given an uint8 array with 256 elements of the same value, we should see a byte array that encodes two runs:

* a 255-length encoding for that value
* a 1-length encoding for that value

Two runs are expected since we use a uint8 to encode the run length so 256 is not valid.

In the current implementation, the compressed result for 256 elements of the same value is instead a 0-length encoding for that value. This is due to integer overflow wrapping.
This means that we are missing frame types in Elasticsearch for stacktraces longer than 255.

## Checklist

- Is it observable through the addition of either **logging** or **metrics**? N/A
- Is its use being published in **telemetry** to enable product improvement? N/A

## How to test these changes

There is an additional unit test that validates the unforeseen case


axw (approved):
@inge4pres presumably you'll want this in 8.5.0? If so, please apply the label `backport-8.5` to backport to the 8.5 branch after it's merged.

inge4pres:
> @inge4pres presumably you'll want this in 8.5.0? If so, please apply the label `backport-8.5` to backport to the 8.5 branch after it's merged.

Thanks! This is indeed a bug fix for 8.5 🙏🏼

rockdaboot (approved):
:+1:

PR #9208, opened 2022-09-27
Make ecs.version a constant_keyword field

### LINKED ISSUE #9043: Define `ecs.version` as a `constant_keyword` field
We set `ecs.version` using a model processor, in the apm-server code, based on the version of the `github.com/elastic/ecs/code/go/ecs` module in use:

https://github.com/elastic/apm-server/blob/0c6c95ff8d6fb81a32ebf38420016b7d6ba765c2/internal/beater/processors.go#L64-L72

An older version of APM Server may send documents to the data streams, and the ingest pipeline is responsible for upgrading/transforming documents to the appropriate format. Therefore I believe it makes sense for the ECS version to be set by the installed ingest pipeline and index template. We can do this by making `ecs.version` a [`constant_keyword`](https://www.elastic.co/guide/en/elasticsearch/reference/current/keyword.html#constant-keyword-field-type) field.

This will also allow the code to be simplified, and documents to be made (slightly) smaller.

axw:
## Motivation/summary

Instead of adding `ecs.version` as a field in every document written by APM Server, make it a constant_keyword field and set its version in the data stream mappings. This means that `ecs.version` will not be available in `_source`, and will only be available through the Elasticsearch fields API. Nothing uses `ecs.version` yet, so it doesn't matter in practice.

The data stream's ingest pipeline is responsible for upgrading documents to the expected format, so it follows that the integration package (ingest pipeline, mappings) should define the ECS version. Given that nothing currently relies on `ecs.version`, the pipelines now simply drop any `ecs.version` field defined in the incoming document, in favour of the `constant_keyword` value.

We're currently pinning to a recent commit of ECS due to our reliance on the unreleased `device.*` fields. When they are released, we should update to the released version.

## Checklist

## How to test these changes

1. Install the integration, send some data to APM Server
2. Ensure the documents have `ecs.version` set

## Related issues

Closes #9043


simitt (commented):
Generally LGTM! 

@axw checking to see if I fully grasp the intended upgrade flow here: The apm package sets the ecs version, whenever the package gets upgraded, the templates get created and data streams are rolled over; therefore the ecs version is set on package upgrade. If there are still older apm-server instances sending to ES, there are no conflicts expected, as ecs related changes are expected to be backwards compatible and follow the stack versioning. Correct?

axw:
> If there are still older apm-server instances sending to ES, there are no conflicts expected, as ecs related changes are expected to be backwards compatible and follow the stack versioning. Correct?

Right. Hypothetically if there were a breaking change in ECS, then we would update our ingest pipelines to "upgrade" the docs coming from older APM Servers so they conform to the `ecs.version` set in the mapping. We could use the `observer.version` field from the docs to do that.

simitt (approved):

simitt:
Tested with BC5 that the index templates have the `ecs.version` set accordingly. 
Screenshot 2022-12-05 at 17 25 37

PR #9213, opened 2022-09-27
test: wait for requests to finish when testing async handling

kruskall:
## Motivation/summary

TestConcurrentAsync has a race condition when writing/reading the 'processed' field of the batch processor due to the async nature of the processor. The waitgroup is already handling the concurrent part but because 'handleStream' returns without waiting for the processor to finish, multiple goroutines can access the field, leading to flaky tests.

Add a for loop that blocks trying to acquire/wait for the unreleased locks held by the unfinished processor goroutines.

## Checklist

## How to test these changes

go test -v -race -run=TestConcurrentAsync ./...


internal/processor/stream/processor_test.go
```diff
@@ -392,6 +392,13 @@ func TestConcurrentAsync(t *testing.T) {
...
 		}
 		wg.Wait()
+		if !tc.fullSem {
+			// Try to acquire the lock to make sure all the requests have been handled
+			// and the locks have been released.
+			for i := 0; i < tc.sem; i++ {
+				p.semAcquire(context.Background(), false)
+			}
+		}
 		pResult.Accepted += int(batchProcessor.processed)
```
marclop, line 402: We should also access the field with `atomic.LoadUint64` too.

kruskall: Good catch! Updated!

marclop (approved):
LGTM. I can't think of another way to fix the flakiness without making the test even more complicated

PR #9224, opened 2022-09-28
ci: run smoke tests in parallel even for different versions

v1v:
## Motivation/summary

Some smoke tests run sequentially, let's run them in parallel 

I also took the lead to refactor a bit the pipeline
- use GO_VERSION env variable
- remove extra stage since the parallel creates stages for every entry in the map
- use smoke test filename rather than the full path to help with the UI visualisation in Blue Ocean
- set the env variable with `SMOKETEST_VERSIONS` with the params value, since if no build exists then default values won't be gathered, so this should help this particular corner case.
- run more things in parallel in the CI.

## Test

Marc, :), triggered the [build](https://apm-ci.elastic.co/job/apm-server/job/smoke-tests-mbp/job/PR-9224/15/) manually, let's see how it goes

image


testing/smoke/legacy-managed/test.sh
```diff
@@ -2,8 +2,8 @@
 
 set -eo pipefail
 
-if [[ ${1} != 7.17 ]]; then
-    echo "-> Skipping smoke test..."
+if [[ "${1}" != "7.17" ]]; then
```
v1v, line 5: This should be the fix

testing/smoke/legacy-standalone-major-managed/test.sh
```diff
@@ -2,8 +2,8 @@
 
 set -eo pipefail
 
-if [[ ${1} != 7.17 ]]; then
-    echo "-> Skipping smoke test..."
+if [[ "${1}" != "7.17" ]]; then
```
v1v, line 5: This should be the fix

marclop:
I've opened #9233, which removes the usage of artifacts API in favor of using the ESS API directly.

marclop (commented):
Thanks for the quick fix @v1v! Looked at the execution of the tests themselves and they now seem to be running as they should.

If we can revert the changes in the `lib.sh` and various `test.sh` and use the changes in #9233 instead, I think this PR is in a good state to merge.

I can't comment on the specific groovy changes since I'm unsure what the best practice are for those, but they look good to me.

marclop (approved):
🚀

PR #9225, opened 2022-09-28
smoke-test: check if artifacts api is available

v1v:
### What

Artifacts older than 30 days are not available in the artifacts-api.

This will provide further details when that happens otherwise the information provided by the `test.sh` is not useful in the CI

### Follow up

To define what to do in the future, so @elastic/apm-server I'll let you decide what to do

### DIFF
--- a/testing/smoke/basic-upgrade/test.sh
+++ b/testing/smoke/basic-upgrade/test.sh
@@ -22,8 +22,8 @@ if [[ ${MAJOR_VERSION} -eq 7 ]]; then
     ASSERT_EVENTS_FUNC=legacy_assertions
 
     # Check if the version is available.
-    if ! curl --fail ${ARTIFACTS_API}/versions/${MAJOR_VERSION}.${MINOR_VERSION} ; then
-        echo "-> Warning there are no new artifacts to be downloaded in artifacts-api.elastic.co ..."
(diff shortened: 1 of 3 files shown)


testing/smoke/basic-upgrade/test.sh
```diff
@@ -18,7 +20,14 @@ MINOR_VERSION=$(echo ${VERSION} | cut -d '.' -f2 )
...
     ASSERT_EVENTS_FUNC=legacy_assertions
-    LATEST_VERSION=$(curl -s --fail https://artifacts-api.elastic.co/v1/versions/${MAJOR_VERSION}.${MINOR_VERSION} | jq -r '.version.builds[0].version')
+
+    # Check if the version is available.
+    if ! curl -s --fail ${ARTIFACTS_API}/versions/${MAJOR_VERSION}.${MINOR_VERSION} ; then
+        echo "-> Skipping there are no new artifacts to be downloaded in artifacts-api.elastic.co ..."
+        exit 0
+    fi
+
+    LATEST_VERSION=$(curl -s --fail ${ARTIFACTS_API}/versions/${MAJOR_VERSION}.${MINOR_VERSION} | jq -r '.version.builds[0].version')
```
amannocci, line 30: It should be possible to check if the `LATEST_VERSION` is empty or not to trigger a skip.
This will avoid a double call to the artifacts-api.

v1v:
Superseded by https://github.com/elastic/apm-server/pull/9233

PR #9233, opened 2022-09-29
smoke: Remove usage of artifacts API

marclop:
## Motivation/summary

Removes the usage of the Elastic artifacts API, in favor of using the Elastic Cloud API to obtain all the stack pack versions available in the configured ESS region.

This approach is better than using the artifacts API for two reasons:

    1. Artifacts API and ESS regions may have different stack packs
    2. Unused artifacts are removed after 30 days.

By using the ESS API, we no longer will see failures due to a missing BC or old versions removed (or the `7.17.x` older versions).

Last, it also adds the `SKIP_DESTROY` conditional to all the `test.sh` scripts.

## Checklist

## How to test these changes

Run smoke tests for version `7.17`:
`make smoketest/all SMOKETEST_VERSIONS=7.17`

## Related issues

Supersedes ##9225

### DIFF
--- a/testing/smoke/basic-upgrade/test.sh
+++ b/testing/smoke/basic-upgrade/test.sh
@@ -5,8 +5,8 @@ set -eo pipefail
 # Load common lib
 . $(git rev-parse --show-toplevel)/testing/smoke/lib.sh
 
-# Load the latest versions except SNAPSHOTS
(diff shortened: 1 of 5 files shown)


axw (approved):
Nice! LGTM.

PR #9234, opened 2022-09-29
tf/ec_deployment: Match ESS defaults

### LINKED ISSUE #9123: testing/infra/ec_deployment: set behavioural variable defaults to match Elastic Cloud defaults
Our `ec_deployment` Terraform module has several variables which influence the behaviour of APM Server:
- integrations_server (defaults to false)
- apm_server_expvar (defaults to true)
- apm_server_pprof (defaults to true)

In a vanilla 8.x Elastic Cloud deployment created through the cloud console, we would get:
- Integrations Server
- expvar disabled by default
- pprof disabled by default

I expect all uses of `ec_deployment` apart from benchmarking and soak testing will not care about expvar or pprof, so I suggest we default to disabling them, and explicitly enable them where needed. Also, since APM Server may behave a bit differently when running under Elastic Agent, I would suggest that we default to using Integrations Server.

marclop:
## Motivation/summary

Updates the `ec_deployment` terraform module variables defaults to match the default variables in an ESS deployment.

- `integrations_server`: `false` => `true`
- `apm_server_expvar`: `true` => `false`
- `apm_server_pprof`: `true` => `false`

## Checklist

## How to test these changes

N/A

## Related issues

Closes #9123

### DIFF
--- a/testing/benchmark/main.tf
+++ b/testing/benchmark/main.tf
@@ -44,13 +44,14 @@ module "ec_deployment" {
   deployment_template    = var.deployment_template
   deployment_name_prefix = local.name_prefix
 
-  integrations_server   = true
   apm_server_size       = var.apm_server_size
   apm_server_zone_count = var.apm_server_zone_count
   apm_index_shards      = var.apm_shards
+  apm_server_expvar     = true
+  apm_server_pprof      = true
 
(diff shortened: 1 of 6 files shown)


axw (approved):
:heart:

Issue #9173, opened 2022-09-23
systemtest: TestIntakeAsync/semaphore_size_default is flaky


axw:
Run the test in a loop with `-count=100 -failfast`, and it'll soon fail:

```
intake.go:46:                                                                                                                        
                Error Trace:    /home/andrew/projects/apm-server/systemtest/intake.go:71                                                 
                                                        /home/andrew/projects/apm-server/systemtest/intake.go:46                         
                                                        /home/andrew/projects/apm-server/systemtest/intake_async_test.go:75              
                                                        /home/andrew/projects/apm-server/systemtest/asm_amd64.s:1594                     
                Error:          Not equal:                                                                                               
                                expected: 202                                                                                            
                                actual  : 503                                                                                            
                Test:           TestIntakeAsync/semaphore_size_default                                                                   
                Messages:       {"accepted":0,"errors":[{"message":"queue is full"}]}
```

simitt:
This has happened a bunch of times now on the `8.5` branch (e.g. [8.5-CI](https://apm-ci.elastic.co/blue/organizations/jenkins/apm-server%2Fapm-server-mbp/detail/8.5/12/pipeline)); currently blocking to merge https://github.com/elastic/apm-server/pull/9244.

PR #9268, opened 2022-10-04
test: fix TestIntakeAsync/semaphore_size_default flaky test

kruskall:
## Motivation/summary

The test is using the default semaphore size (10) and sending a request with a lot of events followed by 9 concurrent requests. The first request (heavy.ndjson) is acquiring and releasing the semaphore 193 times and is also processed asynchronously on the server. Because of the async nature of the requests, the heavy request can hold more than a semaphore at the same time leading to a failure (queue is full) while processing the 9 concurrent requests.
The race condition can be avoided by processing the heavy request synchronously followed by 10 concurrent request to make sure the async processing on the intake handler is working.

## Checklist

## Related issues

Closes https://github.com/elastic/apm-server/issues/9173


systemtest/intake_async_test.go
```diff
@@ -56,17 +56,17 @@ func TestIntakeAsync(t *testing.T) {
 		systemtest.CleanupElasticsearch(t)
 		srv := apmservertest.NewServerTB(t)
 
-		systemtest.SendBackendEventsAsyncPayload(t, srv.URL, `../testdata/intake-v2/errors.ndjson`)
+		systemtest.SendBackendEventsPayload(t, srv.URL, `../testdata/intake-v2/errors.ndjson`)
```
marclop, line 59: Let's revert this change, there's no need to process these synchronously since the line below asserts that the documents have been indexed to Elasticsearch, at which point, all the requests have been handled.
```suggestion
		systemtest.SendBackendEventsAsyncPayload(t, srv.URL, `../testdata/intake-v2/errors.ndjson`)
```

marclop (commented):
Mostly LGTM. Just one small comment

systemtest/intake_async_test.go
```diff
@@ -56,17 +56,17 @@ func TestIntakeAsync(t *testing.T) {
...
 		// Ensure the 5 errors are ingested.
 		systemtest.Elasticsearch.ExpectMinDocs(t, 5, "logs-apm.error-*", nil)
 
 		// Send a request with a lot of events (~1920) and expect it to be processed
 		// without any errors.
-		systemtest.SendBackendEventsAsyncPayload(t, srv.URL, `../testdata/intake-v2/heavy.ndjson`)
+		systemtest.SendBackendEventsPayload(t, srv.URL, `../testdata/intake-v2/heavy.ndjson`)
```
simitt, line 65: When this is the root cause, shouldn't https://github.com/elastic/apm-server/pull/9268/files#diff-74763070977bc9b95333184e98c4a318a3dcf2f6f6a62ba88de993bc65c726c1R41 also need to be changed then; to avoid that the `503` is issued due to the async requests using up the available resources?

marclop: `systemtest.SendBackendEventsPayload` will perform the Intake HTTP request synchronously, while `systemtest.SendBackendEventsAsyncPayload` will do that asynchronously.

The Queue is full errors are only thrown for async requests that can't be processed

systemtest/intake_async_test.go
kruskall: Good catch!

marclop (approved):
LGTM! Thanks for fixing the flakiness.

Issue #9243, opened 2022-09-30
apmbench: disable OTLP benchmarks


axw:
Our benchmark results are currently all over the place because the OTLP benchmark is driving APM Server to OOM. We should disable the OTLP benchmarks until we've addressed https://github.com/elastic/apm-server/issues/9242

PR #9270, opened 2022-10-04
apmbench: disable OTLP benchmarks

kruskall:
## Motivation/summary

The OTLP benchmark is driving APM Server to OOM.

## Checklist

## Related issues

Closes https://github.com/elastic/apm-server/issues/9243


systemtest/cmd/apmbench/main.go
```diff
@@ -49,6 +49,8 @@ func Benchmark1000Transactions(b *testing.B, l *rate.Limiter) {
 }
 
 func BenchmarkOTLPTraces(b *testing.B, l *rate.Limiter) {
+	b.Skipf("Disabled until we've addressed https://github.com/elastic/apm-server/issues/9242")
```
marclop, line 52: Rather than disabling it right here disallowing us from running it on demmand, would it be best to update the `BENCHMARK_RUN` expression on the benchmarks so that the test isn't run? WDYT?

kruskall: Good point!

I think we should be able to run it on demand but I'm not sure about changing the benchmark statement. I went with a different approach and used the same code we are using for integration tests. Let me know what you think!

marclop (approved):

systemtest/cmd/apmbench/main.go
```diff
@@ -49,6 +51,12 @@ func Benchmark1000Transactions(b *testing.B, l *rate.Limiter) {
 }
 
 func BenchmarkOTLPTraces(b *testing.B, l *rate.Limiter) {
+	switch strings.ToLower(os.Getenv("OTLPTRACES_TEST")) {
```
marclop, line 54: Seems a bit more suitable than suffixing `_TEST`?
```suggestion
	switch strings.ToLower(os.Getenv("RUN_OTLPTRACES")) {
```

systemtest/cmd/apmbench/main.go
```diff
@@ -49,6 +51,12 @@ func Benchmark1000Transactions(b *testing.B, l *rate.Limiter) {
...
+	case "1", "true":
+	default:
+		b.Skip("Disabled until we've addressed https://github.com/elastic/apm-server/issues/9242, export OTLPTRACES_TEST=1 to run")
```
marclop, line 57:
```suggestion
		b.Skip("Disabled until we've addressed https://github.com/elastic/apm-server/issues/9242, export RUN_OTLPTRACES=1 to run")
```

PR #9308, opened 2022-10-06
fix: recognize OpenTelemetry grpc spans generated by the Javascript SDK

### LINKED ISSUE #9299: Unknown transaction.type with OTLP and GRPC Spans Created by Node.js Instrumentation
**APM Server version** (`apm-server version`): 8.4

OpenTelemetry GRPC spans sent to the APM Server OTLP endpoint create a derived transaction with a `transaction.type` of `unknown` when the spans originate from the Node.js OpenTelemetry GRPC instrumentation.  (GRPC spans in other OpenTelemetry language API/SDKs appear to have the correct transaction.type of `request` sent)

**Steps to reproduce**:

 1. Install and run the Elastic Forked, OTel Instrumented Online Boutique microservices-demo, pointing it at a stock Elastic Cloud instance: https://github.com/julianocosta89/opentelemetry-
 2. Observe the `APM -> Services` section

Expected Behavior: `emailservice` (a python GRPC service) and `currencyservice` (a Node.js GRPC service) should both have a transaction type of `request`

Actual Behavior: Only `emailservice` has a transaction type of request, `currencyservice` has a transaction type of `unknown`

file

Not logs, but using the Console Span exporter the Node.js GRPC spans look like the following

```
{
  traceId: '76e1db1e74ed86b0e529d7ce76d683ce',
  parentId: undefined,
  name: 'grpc.hipstershop.CurrencyService/Convert',
  id: 'a3a33ad55a7cc96f',
  kind: 1,
  timestamp: 1665035192673723,
  duration: 1820,
  attributes: { 'grpc.kind': 1, 'rpc.grpc.status_code': '1' },
  status: { code: 0 },
  events: []
}
{
  traceId: '9ac218a5377d9ff2569ebbf603a160c4',
  parentId: undefined,
  name: 'grpc.hipstershop.CurrencyService/GetSupportedCurrencies',
  id: '9c9ca41e54fd117d',
  kind: 1,
  timestamp: 1665035192680413,
  duration: 715,
  attributes: { 'grpc.kind': 1, 'rpc.grpc.status_code': '1' },
  status: { code: 0 },
  events: []
}
```

My best guess is that [the code in this package fails](https://github.com/elastic/apm-server/blob/626e906f2043174152214ef77572effe0eff2a2f/internal/processor/otel/traces.go#L410-L418) to identify these spans and defaults to a type of unknown.

astorm:
## Motivation/summary

Closes #9299.

The [Node.js OpenTelemetry GRPC instrumentation](https://github.com/open-telemetry/opentelemetry-js/tree/main/experimental/packages/opentelemetry-instrumentation-grpc) does not set the `semconv.AttributeRPCSystem` attribute.  This results in GRPC services listing their transaction.type as `unknown` which is misaligned with the other languages (both Agents and OTel instrumentations)  

This PR ensures that the GRPC spans sent by the current GRPC instrumentation are correctly identified and labeled.

## Checklist

- [ ] ~Update [package changelog.yml](https://github.com/elastic/apm-server/blob/main/apmpackage/apm/changelog.yml) (only if changes to `apmpackage` have been made)~
- [ ] ~Documentation has been updated~

## How to test these changes

PR includes new `TestGRPCTransactionFromNodejsSDK` tests.

If a tester was feeling extra bold, they could spin up [the "@julianocosta89 forked and instrumented with OTel" GCP microservices demo](https://github.com/julianocosta89/opentelemetry-microservices-demo)

## Related issues

https://github.com/elastic/apm-server/issues/9299


internal/processor/otel/traces.go
```diff
@@ -304,6 +304,8 @@ func TranslateTransaction(
 				netHostPort = int(v.IntVal())
 			case "rpc.grpc.status_code":
 				event.Transaction.Result = codes.Code(v.IntVal()).String()
+			case "grcp.kind":
+				event.Transaction.Type = "request"
```
axw, line 308:
```suggestion
```

I would prefer to minimise non-standard attributes like this, and just stick to those defined by the `rpc` semantic conventions: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/rpc.md#semantic-conventions-for-rpc-spans

changelogs/head.asciidoc
```diff
@@ -13,6 +13,7 @@ https://github.com/elastic/apm-server/compare/8.5\...main[View commits]
 
 [float]
 ==== Bug fixes
+- OpenTelemetry GRPC Spans from the Javascript API/SDK/Instrumentations are now correctly transformed into transactions with type=`request`
```
axw, line 16:
```suggestion
- OpenTelemetry GRPC Spans from the Javascript API/SDK/Instrumentations are now correctly transformed into transactions with type=`request`
```

I'm being a little bit pedantic, but this isn't a bug fix, it's an enhancement. The Node.js instrumentation is incorrect: `rpc.system` is a required attribute, per https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/rpc.md#common-attributes

axw (commented):
I'm not _strongly_ opposed to inferring the type from `rpc.grpc.status_code`, but I would prefer not to make this change. I think we should instead send a fix to opentelemetry-js. The gRPC instrumentation appears to have experimental status anyway.

changelogs/head.asciidoc
astorm: That's a fair point of view.  I've moved this down to the `Added` section.

internal/processor/otel/traces.go
astorm: Again, fair.  This isn't strictly necessary and more a covering our bases sort of thing (and cargo culting other non-standard attributes I saw in this file.  Removing this.

astorm:
> but I would prefer not to make this change. I think we should instead send a fix to opentelemetry-js. The gRPC instrumentation appears to have experimental status anyway.

@axw fair, (and ultimately not my repo so it's not my call :)), but a couple things on that.

First -- experimental or not -- folks _are_ starting to use OpenTelemetry and we're trying to support users/customers as best we can.  The experimental label is _not_ going to stop everyone from using OTel (ex. this is not an arbitrary PR -- it's coming from some real world product folks trying to use OTel and running into this problem).  So I'd push back lightly that an experimental label means we shouldn't try to to support this. (Definitely a strict in what you send, liberal in what you accept point of view.  `¯\_(ツ)_/¯`)

Second -- if we do go the "fixing this in `opentelementy-js` route" -- what would that mean to you? My naive read would be that we'd need to get the OpenTelemetry folks to finalize semantic conventions around GRPC attributes, and then get a PR into the GRPC instrumentation `contrib` repo, and then get the OpenTelemetry JS folks to do a release of the contrib repo.  That's all theoretically reasonable -- but also something that's going to be very bureaucratically time consuming and probably wouldn't happen before out next stack release.  Whereas this small change would fix the problem before that. APM Server is one of the reasons Elastic is a superior platform for OpenTelemetry (or data-ingest from any tracing/APM platform) -- we can make this sort of small adjustments here without needing to get the rest of the world to get their act together.

astorm:
> Second -- if we do go the "fixing this in `opentelemetry-js` route" -- what would that mean to you? 

Just saw: https://github.com/elastic/apm-server/issues/9299#issuecomment-1272666550 (the joys of reading email backwards).

> The bug is with the instrumentation, and should be fixed there. https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/rpc.md#common-attributes states that rpc.system is required.

Presuming the above would be the "fixing this in `opentelemetry-js`"

internal/processor/otel/traces.go
```diff
@@ -566,6 +568,8 @@ func TranslateSpan(spanKind ptrace.SpanKind, attributes pcommon.Map, event *mode
 
 			switch kDots {
 			// http.*
+			case "rpc.grpc.status_code":
```
axw, line 571:
```suggestion
			case semconv.AttributeRPCGRPCStatusCode:
```

Also can you please move this down to the "// rpc.*" section below?

internal/processor/otel/traces_test.go
```diff
@@ -1341,6 +1341,25 @@ func TestServiceTarget(t *testing.T) {
...
+	}
+
+	test(t, map[string]interface{}{
+		"rpc.grpc.status_code": codes.Unavailable,
+	})
+
+	test(t, map[string]interface{}{
+		"rpc.grpc.status_code": codes.Unavailable,
+		"grcp.kind":            1,
+	})
```
axw, line 1360:
```suggestion
```

axw (commented):
@astorm you've convinced me that it's small enough to be worthwhile.

I'm not the biggest fan of Postel's Law in general as it tends to entrench bugs, but I think it's reasonable to say if there are any `rpc.*` attributes then it's an RPC transaction/span, and not really care about the presence of `rpc.system` specifically.

internal/processor/otel/traces.go
astorm: moved in 2cc84eb8f8ac7d6f80585604ef2453d333a26d53

internal/processor/otel/traces.go
```diff
@@ -649,6 +651,8 @@ func TranslateSpan(spanKind ptrace.SpanKind, attributes pcommon.Map, event *mode
 			case semconv.AttributeRPCService:
 				rpcService = stringval
 				foundSpanType = rpcSpan
+			case semconv.AttributeRPCGRPCStatusCode:
+				event.Transaction.Type = "request"
```
axw, line 655: I just noticed that this code is meant for translating spans, so we shouldn't be setting transaction type.

```suggestion
				rpcSystem = "grpc"
				foundSpanType = rpcSpan
```

internal/processor/otel/traces_test.go
```diff
@@ -1341,6 +1341,20 @@ func TestServiceTarget(t *testing.T) {
 	})
 }
 
+func TestGRPCTransactionFromNodejsSDK(t *testing.T) {
```
axw, line 1344: Can you please also add a test (maybe a subtest of this one) that creates a gRPC client span, and verify that the `span.type` and `span.subtype` are "external" and "grpc" respectively?

axw (changes requested):

internal/processor/otel/traces_test.go
astorm: 👍 Makes sense, tackling this today.

astorm: @axw Change made, thanks for catching this one.

axw (approved):
LGTM, thanks @astorm!

axw:
Removing 8.4 backport, since 8.5 is imminent.

Issue #9183, opened 2022-09-23
Enable doc-value-only mappings for APM data streams


simitt:
We want to improve indexing speed and reduce storage size for indexed events where possible. To achieve this, we want to look into utilising the [doc-value-only feature introduced in Elasticsearch 8.1](https://www.elastic.co/blog/whats-new-elasticsearch-kibana-cloud-8-1-0). For fields that are rarely queried, using `doc-value-only` fields can be a good choice. 

Goal of this effort is to identify a list of fields that should be doc-value-only and provide benchmarks, proofing that the APM UI performance is not negatively impacted, while storage size is reduced and ingestion speed increased. 

Recommended steps to achieve this
* [x] Identify a list of fields that should be `doc-value-only`. Fields with field types `keyword`, `date`, `ip`, `boolean`, `geo_point`, and numerical fields are general candidates. For any such field where searches are expected to never or only rarely make sense, a  reduced storage size would be a better default. Collaborate with someone from the APM UI team to identify which fields are used in the APM UI. 
* [x] Implement a POC with the field type changes. For this the integration package needs to be updated
* [ ] Perform benchmarks on storage size, leveraging the recently introduced Rally tooling (see https://github.com/elastic/apm-dev/issues/772). 
* [ ] ~Evaluate indexing speed improvements, leveraging the recently introduced Rally tooling.~ 
* [x] Consult with the the UI team after they conducted benchmarks for verfiying that changes do not negatively effect the APM UI performance (see https://github.com/elastic/kibana/issues/142337).
* [ ] Verify that users can manually opt into indexing on a field-by-field basis

kruskall:
Suggestions for doc-value-only fields:

- `cloud.origin.provider`
- `faas.version`
- `network.carrier.icc`
- `network.carrier.mcc`
- `network.carrier.mnc`
- `network.carrier.name`
- `network.connection.type`
- `network.connection.subtype`
- `observer.ephemereal_id`
- `observer.id`
- `service.framework.version`
- `service.origin.version`
- `service.runtime.version`
- `span.db.rows_affected`

kruskall:
List of fields not using the reverse index, ordered by size:

```
Field                                                   Size
-----                                                   ----
labels.request_id                                       1mb
labels.worker                                           900.6kb
labels.thread_name                                      592.7kb
labels.process_runtime_description                      384.6kb
labels.telemetry_auto_version                           383.7kb
labels.b                                                138.2kb
[...]

PR #9316, opened 2022-10-10
poc: enable doc-value-only mappings for APM data streams

kruskall:
## Motivation/summary

Mark `timestamp.us` as doc-value-only field for a quick poc.

Add test to verify the field is indexed as expected.

This is a PoC and `timestamp.us` is used as an example. The actual list of fields should be decided after coordinating with the UI team.

## Checklist

## Related issues

Related to https://github.com/elastic/apm-server/issues/9183


apmpackage/apm/data_stream/traces/manifest.yml
```diff
@@ -16,3 +16,9 @@ elasticsearch:
             mapping:
               type: scaled_float
               scaling_factor: 1000000
+      properties:
+        timestamp:
+          properties:
+            us:
+              type: long
+              index: false
```
axw, line 24: I think we should instead update `fields/fields.yml`:

https://github.com/elastic/apm-server/blob/ff5e1792298a52fffef1e5079917942b6b364c94/apmpackage/apm/data_stream/traces/fields/fields.yml#L248-L251

IIANM you can just add `index: false` to the field definition there.

axw (changes requested):
I'm not convinced that the test is necessary, since it's second guessing that Fleet is working as expected. WDYT?

apmpackage/apm/data_stream/traces/manifest.yml
kruskall: Thanks for the tip! :bow: 

I've updated the poc to use the fields mentioned in the related issue

kruskall:
> I'm not convinced that the test is necessary, since it's second guessing that Fleet is working as expected. WDYT?

Agree, I've removed the test. I've updated the PR to use the new fields.

apmpackage/apm/data_stream/traces/manifest.yml
axw: Thanks, looks good :)

axw (approved):
Looks good!

Later on we should revisit _how_ we define the fields that should or should not be indexed. For example we might update the package generator (apmpackage/cmd/genpackage) to inject the property across all the data streams for certain fields.

That would save us from having to manually keep them in sync. Also it would enable us to define some policies, like "all metric fields should be doc-value-only". We might also want to opt into indexing, rather than opting out, depending on how many fields are actually used in searches.

Issue #9182, opened 2022-09-23
Ensure APM Server makes full usage of available CPU resources


simitt:
Previous benchmarks indicate that the APM Server performs better when scaling horizontally than vertically. This is a meta issue collecting steps necessary to ensure APM Server's throughput scales well when scaling the server's available CPU resources up. 

Previous investigations identified the modelindexer to be a bottle neck, and two action items were identified: 

- [x] https://github.com/elastic/apm-server/issues/9180
- [x] https://github.com/elastic/apm-server/issues/9181

The improvements need to be benchmarked; the goal is to provide full resource usage and (almost) linear scaling for APM Server's up to `32GB` of RAM (with aliquot CPU time) on ESS.

kruskall:
I've reviewed the code in the linked PRs and looked at the trace files for specific benchmarks like `BenchmarkModelIndexer`.

I could not find any issues but I noticed the flush goroutine for the active indexers (introduced in https://github.com/elastic/apm-server/pull/9318) being always busy.
A goroutine being active might slow down GC and increase memory usage. I don't think that's going to be a big issue here due to the flush timer channel but I noticed some improvements and opened a PR to reduce the cpu time spent in the flush gorutine: https://github.com/elastic/apm-server/pull/9760

PR #9318, opened 2022-10-10
modelindexer: Run active in dedicated goroutine

### LINKED ISSUE #9180: Move event compression outside of the modelindexer activeMu lock
From @marclop 's findings:

> As the modelindexer processes the batches it receives, it compresses each event by default before they are stored in the bulk indexer cache. This is done with the activeMu lock held and caused the entire processing pipeline to stop until a [single event is compressed](https://github.com/elastic/apm-server/blob/57d94db8d8cc8a56003018921814a45b40fada46/internal/model/modelindexer/indexer.go#L270-L274). If we can defer the compression for later and even better remove the activeMu lock to eliminate its contention, the processing can progress much faster.

> Since the modelindexer activeMu lock seems to be the offending component causing a significant bottleneck and reduced throughput, we benchmarked a modified version of APM Server where the modelindexer activeMu lock is eliminated and a new channel inside the modelindexer is introduced to decouple the cache writing from the HTTP request lifecycle. Also, since we have a queue where the BulkRequestItems are sent before they are compressed, we can scale the compression and flushing of the bulk indexers to increase the number of "active" consumers from the queue. See for more details on the design.

Implement a solution where the event compression is not causing lock contention so that the APM Server can better make use of available CPU resources.

marclop:
## Motivation/summary

Remove the `i.activeMu` from the modelindexer and shared sync primitives
from the indexer structure, instead, run the active indexer in its own
goroutine, minimizing the amount of shared locking that needs to happen
when `ProcessBatch` is called. To allow (potentially) multiple active
indexers from running in parallel, a new internal `bulkItems` channel is
introduced, with a variable buffer depending on how powerful the machine
where APM Server is run. The bigger the machine, the bigger the channel
buffer, and vice versa.

Additionally, the default `FlushBytes` reduced from `5MB` to `2MB`, and
to keep the same total "bufferable" size, the default `MaxRequests` is
increased to 25 (25 * 2mb = 50mb). The motivation behind this change is
is twofold; Allowing the modelindexer to have more background flushes
if enough events are being processed by the APM Server, and, creating
slightly smaller bulk requests to Elasticsearch, since, with the default
settings (5MB FlushBytes), the number of bulk documents is ~`18-24K`.

Initial micro-benchmarks indicate that the performance improvement is
~42% with default settings, and macro-benchmarks ~27%:

```
$ benchstat old.txt single-pr.txt
name                               old time/op    new time/op    delta
ModelIndexer/NoCompression-8         1.11µs ± 1%    1.14µs ± 2%   +3.53%  (p=0.008 n=5+5)
ModelIndexer/BestSpeed-8             2.03µs ± 3%    1.70µs ± 1%  -15.90%  (p=0.008 n=5+5)
ModelIndexer/DefaultCompression-8    4.10µs ± 2%    2.36µs ± 1%  -42.53%  (p=0.008 n=5+5)
ModelIndexer/BestCompression-8       6.95µs ± 1%    4.26µs ± 1%  -38.71%  (p=0.008 n=5+5)

name                               old alloc/op   new alloc/op   delta
ModelIndexer/NoCompression-8         2.07kB ± 0%    2.14kB ± 0%   +3.54%  (p=0.016 n=4+5)
ModelIndexer/BestSpeed-8             2.57kB ± 0%    2.56kB ± 0%   -0.33%  (p=0.008 n=5+5)
ModelIndexer/DefaultCompression-8    2.55kB ± 0%    2.53kB ± 0%   -0.79%  (p=0.008 n=5+5)
ModelIndexer/BestCompression-8       2.58kB ± 0%    2.55kB ± 0%   -0.96%  (p=0.008 n=5+5)

name                               old allocs/op  new allocs/op  delta
ModelIndexer/NoCompression-8           20.0 ± 0%      20.0 ± 0%     ~     (all equal)
ModelIndexer/BestSpeed-8               24.0 ± 0%      24.0 ± 0%     ~     (all equal)
ModelIndexer/DefaultCompression-8      24.0 ± 0%      24.0 ± 0%     ~     (all equal)
ModelIndexer/BestCompression-8         24.0 ± 0%      24.0 ± 0%     ~     (all equal)
```

```
name          old events/sec  new events/sec  delta
AgentAll-512  26.8k ± 0%      34.1k ±11%      +27.11%  (p=0.100 n=3+3)
```

## Checklist

## How to test these changes

TODO

## Related issues

Closes #9180


marclop:
I've included two commits of the implementation, the first one uses a local `sync.Mutex` in the active goroutine, not dissimilar to how the model indexer used locking previously (451e7f2).

```console
$ benchstat old.txt single.txt
name                               old time/op    new time/op    delta
ModelIndexer/NoCompression-8         1.11µs ± 1%    1.19µs ±15%   +7.61%  (p=0.008 n=5+5)
ModelIndexer/BestSpeed-8             2.03µs ± 3%    1.89µs ± 2%   -6.91%  (p=0.008 n=5+5)
ModelIndexer/DefaultCompression-8    4.10µs ± 2%    2.34µs ± 1%  -42.98%  (p=0.008 n=5+5)
ModelIndexer/BestCompression-8       6.95µs ± 1%    4.20µs ± 1%  -39.58%  (p=0.008 n=5+5)

name                               old alloc/op   new alloc/op   delta
ModelIndexer/NoCompression-8         2.07kB ± 0%    2.14kB ± 0%   +3.30%  (p=0.016 n=4+5)
ModelIndexer/BestSpeed-8             2.57kB ± 0%    2.56kB ± 0%   -0.35%  (p=0.008 n=5+5)
ModelIndexer/DefaultCompression-8    2.55kB ± 0%    2.52kB ± 0%   -0.83%  (p=0.008 n=5+5)
ModelIndexer/BestCompression-8       2.58kB ± 0%    2.56kB ± 0%   -0.62%  (p=0.008 n=5+5)

name                               old allocs/op  new allocs/op  delta
ModelIndexer/NoCompression-8           20.0 ± 0%      20.0 ± 0%     ~     (all equal)
ModelIndexer/BestSpeed-8               24.0 ± 0%      24.0 ± 0%     ~     (all equal)
ModelIndexer/DefaultCompression-8      24.0 ± 0%      24.0 ± 0%     ~     (all equal)
ModelIndexer/BestCompression-8         24.0 ± 0%      24.0 ± 0%     ~     (all equal)
```

The second one implements the same "algorithm" but without using channels instead of a mutex. It seems to have comparable performance: (ef72d77):

```console
$ benchstat old.txt single-lockless.txt
name                               old time/op    new time/op    delta
ModelIndexer/NoCompression-8         1.11µs ± 1%    1.15µs ± 4%   +4.41%  (p=0.008 n=5+5)
ModelIndexer/BestSpeed-8             2.03µs ± 3%    1.88µs ± 1%   -7.14%  (p=0.008 n=5+5)
ModelIndexer/DefaultCompression-8    4.10µs ± 2%    2.36µs ± 2%  -42.32%  (p=0.008 n=5+5)
ModelIndexer/BestCompression-8       6.95µs ± 1%    4.21µs ± 1%  -39.48%  (p=0.008 n=5+5)

name                               old alloc/op   new alloc/op   delta
ModelIndexer/NoCompression-8         2.07kB ± 0%    2.14kB ± 0%   +3.44%  (p=0.016 n=4+5)
ModelIndexer/BestSpeed-8             2.57kB ± 0%    2.56kB ± 0%   -0.31%  (p=0.008 n=5+5)
ModelIndexer/DefaultCompression-8    2.55kB ± 0%    2.52kB ± 0%   -0.87%  (p=0.008 n=5+5)
ModelIndexer/BestCompression-8       2.58kB ± 0%    2.57kB ± 0%   -0.52%  (p=0.008 n=5+5)

name                               old allocs/op  new allocs/op  delta
ModelIndexer/NoCompression-8           20.0 ± 0%      20.0 ± 0%     ~     (all equal)
ModelIndexer/BestSpeed-8               24.0 ± 0%      24.0 ± 0%     ~     (all equal)
ModelIndexer/DefaultCompression-8      24.0 ± 0%      24.0 ± 0%     ~     (all equal)
ModelIndexer/BestCompression-8         24.0 ± 0%      24.0 ± 0%     ~     (all equal)
```

```console
$ benchstat single.txt single-lockless.txt
name                               old time/op    new time/op    delta
ModelIndexer/NoCompression-8         1.19µs ±15%    1.15µs ± 4%   ~     (p=0.841 n=5+5)
ModelIndexer/BestSpeed-8             1.89µs ± 2%    1.88µs ± 1%   ~     (p=0.889 n=5+5)
ModelIndexer/DefaultCompression-8    2.34µs ± 1%    2.36µs ± 2%   ~     (p=0.548 n=5+5)
ModelIndexer/BestCompression-8       4.20µs ± 1%    4.21µs ± 1%   ~     (p=0.690 n=5+5)

name                               old alloc/op   new alloc/op   delta
ModelIndexer/NoCompression-8         2.14kB ± 0%    2.14kB ± 0%   ~     (p=0.579 n=5+5)
ModelIndexer/BestSpeed-8             2.56kB ± 0%    2.56kB ± 0%   ~     (p=0.667 n=5+5)
ModelIndexer/DefaultCompression-8    2.52kB ± 0%    2.52kB ± 0%   ~     (p=0.373 n=5+5)
ModelIndexer/BestCompression-8       2.56kB ± 0%    2.57kB ± 0%   ~     (p=0.905 n=5+5)

name                               old allocs/op  new allocs/op  delta
ModelIndexer/NoCompression-8           20.0 ± 0%      20.0 ± 0%   ~     (all equal)
ModelIndexer/BestSpeed-8               24.0 ± 0%      24.0 ± 0%   ~     (all equal)
ModelIndexer/DefaultCompression-8      24.0 ± 0%      24.0 ± 0%   ~     (all equal)
ModelIndexer/BestCompression-8         24.0 ± 0%      24.0 ± 0%   ~     (all equal)
```

marclop:
I think the channel lockless approach has a race condition which causes the active indexer to never be flushed. I haven't managed to track down where/how it happens, so it may be best to leave that approach as a future optimization and update this PR to use the local mutex, which seems to be equal performance wise.

internal/model/modelindexer/indexer.go
```diff
@@ -434,6 +397,72 @@ func (i *Indexer) flush(ctx context.Context, bulkIndexer *bulkIndexer) error {
...
+		var mu sync.Mutex
+		// `active` and `timer` must be accessed with `mu` locked.
+		var active *bulkIndexer
+		var timer *time.Timer
+		flush := make(chan *bulkIndexer)
+		defer func() {
+			atomic.AddInt64(&i.activeBulkRequests, -1)
+			close(flush)
+		}()
+		for item := range i.bulkItems {
```
axw, line 417: I wonder if this could be simplified, and remove some locking, by selecting on the `flush` event in the foreground? Something like this:

```go
var closed bool
var active *bulkIndexer
flushTimer := time.NewTimer(0)
if !flushTimer.Stop() {
    <-flushTimer.C
}
for !closed {
    var shouldFlush bool
    select {
    case <-flushTimer.C:
        shouldFlush = true
    default:
        select {
        case <-flushTimer.C:
            shouldFlush = true
        case event, ok := <-i.bulkItems:
            if !ok {
                closed = true
                break
            }
            if active == nil {
                active = <-i.available
                atomic.AddInt64(&i.availableBulkRequests, -1)
                flushTimer.Reset(i.config.FlushInterval)
            }
            if err := active.Add(item); err != nil {
                i.logger.Errorf("failed adding event to bulk indexer: %v", err)
            }
            // Flush the active bulk indexer when it's at or exceeds the
            // configured FlushBytes threshold.
            if active.Len() >= i.config.FlushBytes {
                shouldFlush = true
                if !flushTimer.Stop() {
                    <-flushTimer.C
                }
            }
        }
    }

    if shouldFlush {
        indexer := active
        active = nil
        if !flushTimer.Stop() {
            <-flushTimer.C
        }
        i.g.Go(func() error {return i.flushIndexer(context.Background(), indexer)})
    }
}
```

internal/model/modelindexer/indexer.go
```diff
@@ -434,6 +397,72 @@ func (i *Indexer) flush(ctx context.Context, bulkIndexer *bulkIndexer) error {
...
+			mu.Lock()
+			if active == nil {
+				active = <-i.available
```
axw, line 420: Should this be in a select? Is it possible for available to be depleted for some non-negligible amount of time?

axw (commented):
Looking good! This is starting to look a bit like the APM Go Agent's tracer loop :)

internal/model/modelindexer/indexer.go
```diff
@@ -234,54 +235,21 @@ func (i *Indexer) processEvent(ctx context.Context, event *model.APMEvent) error
...
+	// Send the BulkIndexerItem to the internal channel, allowing individual
+	// events to be processed by an active bulk indexer in a dedicate goroutine,
+	// which in turn speeds up event processing.
+	case i.bulkItems <- elasticsearch.BulkIndexerItem{
+		Index:  r.indexBuilder.String(),
 		Action: "create",
 		Body:   r,
-	}); err != nil {
-		return err
+	}:
```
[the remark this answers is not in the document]
marclop: That makes sense, however, for better or worse we'd be performing that processing by the consumer which was a more limited number of concurrency (1 at the moment of this PR), rather than the producer (up to 200 concurrent json encodes). This could slow down the pipeline, but if we implement multiple active indexers, then I think that makes a lot of sense since we'd have more than a single consumer

internal/model/modelindexer/indexer.go
```diff
@@ -434,6 +397,72 @@ func (i *Indexer) flush(ctx context.Context, bulkIndexer *bulkIndexer) error {
 	return nil
 }
 
+// runActiveIndexer starts a new active indexer which pulls items from the
+// bulkItems channel. The more active indexers there are, the faster events
+// will be pulled out of the queue, but also the more likely it is that the
+// outgoing Elasticsearch bulk requests are flushed due to the idle timer,
+// rather than due to being full.
+func (i *Indexer) runActiveIndexer() {
+	i.g.Go(func() error {
```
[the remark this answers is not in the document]
marclop: that's fair, I'll move it there 👍🏼

internal/model/modelindexer/indexer.go
```diff
@@ -434,6 +397,72 @@ func (i *Indexer) flush(ctx context.Context, bulkIndexer *bulkIndexer) error {
...
+		atomic.AddInt64(&i.activeBulkRequests, 1)
+		var mu sync.Mutex
+		// `active` and `timer` must be accessed with `mu` locked.
+		var active *bulkIndexer
+		var timer *time.Timer
+		flush := make(chan *bulkIndexer)
```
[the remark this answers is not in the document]
marclop: I initially gave it a buffer of 1 but removed it since it didn't seem to make much of a difference, I might re-introduce it and micro-benchmark and see if there's any meaningful difference

internal/model/modelindexer/indexer.go
marclop: I've implemented a very similar approach in [5e5b24d](https://github.com/elastic/apm-server/pull/9318/commits/5e5b24da3f6c3f69ed83667cac3798a43f6d14c0). Let me know what you think.

internal/model/modelindexer/indexer.go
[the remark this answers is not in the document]
axw: I think we should still try and measure it later. It may be that removing the intermediate buffer & copying offsets the loss of concurrency. Anyway, not for now.

internal/model/modelindexer/indexer.go
```diff
@@ -434,6 +393,60 @@ func (i *Indexer) flush(ctx context.Context, bulkIndexer *bulkIndexer) error {
...
+	for !closed {
+		select {
+		case <-flushTimer.C:
+		default:
+			select {
+			case <-flushTimer.C:
+			case event, ok := <-i.bulkItems:
+				if !ok {
+					closed = true
+					break
```
axw, line 419:
```suggestion
					break // we'll flush one last time below, if there's an active indexer
```

internal/model/modelindexer/indexer.go
```diff
@@ -434,6 +393,60 @@ func (i *Indexer) flush(ctx context.Context, bulkIndexer *bulkIndexer) error {
...
+				}
+				if active == nil {
+					active = <-i.available
```
axw, line 422: Should this be a select with `i.closed`, to cause the loop to end immediately when the context passed to Close is cancelled?

internal/model/modelindexer/indexer.go
axw: Looks great!

internal/model/modelindexer/indexer.go
marclop: Isn't that redundant? Given that we close the `i.bulkItems` when the indexer is closing? Maybe I'm missing an obvious condition

axw: What I'm wondering is if it's possible for `<-available` to block if the indexer is closed after we've received from bulkItems. I suppose not, since flushIndexer will put the indexers back into available when `i.closed` is signalled. So never mind :)

axw (approved):
LGTM! Might be worth running the benchmarks one more time, after the changes?

marclop:
@axw benchmarks look good however I do notice an increase in overall RSS which could be problematic (and is) for 1GB APM Servers, and cause them to run out of memory.

Bear in mind that I have done this testing with a slightly smaller ES deployment (6 nodes instead of 12, totaling `116g`, rather than `232g`). Even then, we've got a ~20% performance increase in all sizes except `1gb` and  2GB` which is impressive.

## 2GB

```console
$ name          old time/op                  new time/op                  delta
AgentAll-128                   2.18s ± 0%                   2.13s ± 0%    -2.20%  (p=0.100 n=3+3)

AgentAll-128                   8.28k ± 0%                   8.46k ± 0%    +2.21%  (p=0.100 n=3+3)

AgentAll-128                     146 ± 8%                     198 ±21%   +35.70%  (p=0.100 n=3+3)

AgentAll-128                    896M ± 1%                   1059M ± 6%   +18.24%  (p=0.100 n=3+3)

AgentAll-128                   8.24M ± 1%                   9.49M ± 4%   +15.17%  (p=0.100 n=3+3)

AgentAll-128                    965M ± 1%                   1167M ± 5%   +20.98%  (p=0.100 n=3+3)

AgentAll-128                   558MB ± 0%                   560MB ± 0%    +0.37%  (p=0.100 n=3+3)

AgentAll-128                   7.93M ± 0%                   7.94M ± 0%    +0.10%  (p=0.100 n=3+3)
```

## 4GB

```console
$ name          old time/op                  new time/op                  delta
AgentAll-256                   1.34s ± 1%                   1.12s ± 0%   -16.57%  (p=0.100 n=3+3)

AgentAll-256                   13.4k ± 1%                   16.1k ± 0%   +19.87%  (p=0.100 n=3+3)

AgentAll-256                     176 ± 8%                     250 ± 4%   +41.97%  (p=0.100 n=3+3)

AgentAll-256                    914M ± 2%                   1096M ± 0%   +19.95%  (p=0.100 n=3+3)

AgentAll-256                   8.14M ± 1%                   9.81M ± 1%   +20.46%  (p=0.100 n=3+3)

AgentAll-256                    987M ± 1%                   1195M ± 1%   +21.10%  (p=0.100 n=3+3)

AgentAll-256                    5.78 ± 1%                   19.82 ± 0%  +242.65%  (p=0.100 n=3+3)

AgentAll-256                   566MB ± 0%                   560MB ± 0%    -1.06%  (p=0.100 n=3+3)

AgentAll-256                   8.02M ± 0%                   7.94M ± 0%    -0.91%  (p=0.100 n=3+3)
```

## 8GB

```console
$ name          old time/op                  new time/op                  delta
AgentAll-512                   673ms ± 0%                   566ms ± 8%   -15.94%  (p=0.100 n=3+3)

AgentAll-512                   26.8k ± 0%                   32.0k ± 8%   +19.26%  (p=0.100 n=3+3)

AgentAll-512                     284 ± 1%                     434 ± 6%   +52.93%  (p=0.100 n=3+3)

AgentAll-512                    936M ± 1%                   1190M ± 2%   +27.12%  (p=0.100 n=3+3)

AgentAll-512                   7.86M ± 2%                  10.50M ± 2%   +33.57%  (p=0.100 n=3+3)

AgentAll-512                   1.04G ± 2%                   1.31G ± 2%   +25.60%  (p=0.100 n=3+3)

AgentAll-512                    2.40 ± 1%                    9.59 ±14%  +299.17%  (p=0.100 n=3+3)

AgentAll-512                   562MB ± 0%                   574MB ± 0%    +2.25%  (p=0.100 n=3+3)

AgentAll-512                   7.98M ± 0%                   8.06M ± 0%    +0.94%  (p=0.100 n=3+3)
```

In terms of CPU usage, the old indexer only used about 50% CPU on an 8GB APM Server, while the new one goes as high as 67%.

I think we may need to revisit the semaphore size, particularly for smaller sizes and perhaps size it according to the amount of memory available to APM Server.

simitt (commented):
Changes generally look good. The higher memory consumption is concerning, will you address this in more detail in a follow up @marclop ?

marclop:
I suspect that the culprit may be the internal model indexer channel size: `bulkItems: make(chan elasticsearch.BulkIndexerItem, 1000),`. In smaller instances that is quite a lot of events to buffer, it would probably be better to dynamically size the channel based on memory rather than always having a buffered channel with capacity for 1000 events as they now may take up to `1000 * model.APMEvent`.

I think it may be better to reduce that channel size to something much smaller for now and look into dynamic sizing based on memory.

marclop:
After lowering the channel size to `100`, the total memory increased is more contained: ~8% for 1G instances, and ~16% for the rest. 

I think we should merge this PR, and I'll work on a PR to resize both the channel and the semaphore size based on `cgroup` (or in its absence, machine) memory limit.

axw:
This should be tested as part of https://github.com/elastic/apm-server/issues/9182

Issue #9178, opened 2022-09-23
Set `error.id` for OpenTelemetry exception span events


axw:
The APM UI expects error events to have an `error.id` field, which is used for fetching error metadata: https://github.com/elastic/kibana/issues/141590

We currently do not set this for error events derived from OpenTelemetry exception span events. We should generate a UUID.

axw:
Verified with BC2. I used `systemtest/cmd/sendotlp` to send an exception span event, confirmed the resulting error document has an `error.id` field, and confirmed that the UI shows error metadata when looking at the error sample.

PR #9327, opened 2022-10-10
fix: set error.id for OpenTelemetry exception span events

kruskall:
## Motivation/summary

The APM UI expects error events to have an error.id field, which is used for fetching error metadata.
Generate a random UUID for otlp errors.

## Checklist

## Related issues

Closes https://github.com/elastic/apm-server/issues/9178


marclop:
Also, this is a pretty small change, I think we should also make it available as a bug fix in `8.5.1`.

Issue #3878, opened 2020-06-11
model: align error with ECS


axw:
The ECS [Error](https://www.elastic.co/guide/en/ecs/current/ecs-error.html) definition only defines a minimal set of values. There is no concept for chained exceptions. There is no description how to relate error logs. 

APM uses an extensive fieldset in the error namespace, with the main sub-namespaces being `error.exception`, `error.log` and `error.grouping_key`. We should propose changes (or adapt our model) to ECS.

We should take into account https://github.com/elastic/apm-server/issues/3877 while implementing this.

PR #9349, opened 2022-10-12
Add ECS log files to the intake log endpoint

### LINKED ISSUE #9100: Add fields from ECS logging spec to log support in v2 intake API
Extend the log support in v2 intake API to handle fields in [ECS logging spec](https://github.com/elastic/ecs-logging/blob/main/spec/spec.json).

lahsivjar:
## Motivation/summary

Add ECS logging fields to intake API's log endpoint.

## Checklist

## How to test these changes

1. Create a request body for the `log` event type:
```json
{"metadata":{"process":{"pid":1234,"title":"/usr/lib/jvm/java-10-openjdk-amd64/bin/java","ppid":1,"argv":["-v"]},"system":{"architecture":"amd64","detected_hostname":"8ec7ceb99074","configured_hostname":"host1","platform":"Linux","container":{"id":"8ec7ceb990749e79b37f6dc6cd3628633618d6ce412553a552a0fa6b69419ad4"},"kubernetes":{"namespace":"default","pod":{"uid":"b17f231da0ad128dc6c6c0b2e82f6f303d3893e3","name":"instrumented-java-service"},"node":{"name":"node-name"}}},"service":{"name":"1234_service-12a3","version":"4.3.0","node":{"configured_name":"8ec7ceb990749e79b37f6dc6cd3628633618d6ce412553a552a0fa6b69419ad4"},"environment":"production","language":{"name":"Java","version":"10.0.2"},"agent":{"version":"1.10.0","name":"java","ephemeral_id":"e71be9ac-93b0-44b9-a997-5638f6ccfc36"},"framework":{"name":"spring","version":"5.0.0"},"runtime":{"name":"Java","version":"10.0.2"}},"labels":{"group":"experimental","ab_testing":true,"segment":5}}}
{"log": {"message": "test log message without timestamp"}}
{"log": {"message": "test log message with timestamp", "@timestamp": 1662616049000000}}
{"log": {"message": "test log message with string timestamp", "@timestamp": "2022-09-08T13:47:29.111-0800"}}
{"log": {"message": "test log message with faas", "@timestamp": 1662616049000000,"faas":{"coldstart":true,"execution":"6f7f0961f83442118a7af6fe80b88d56","id":"arn:aws:lambda:us-east-2:123456789012:function:custom-runtime"}}}
{"log": {"trace.id": "trace-id", "transaction.id": "txn-id", "log.level": "warn", "log.logger": "testLogger", "log.origin.file.name": "testFile", "log.origin.file.line": 10, "log.origin.function": "testFunc", "service.name": "testSvc", "service.version": "v1.0.0", "service.environment": "prod", "process.thread.name": "testThread", "dataset": "accesslog", "labels": {"str": "str", "int": 1, "float": 1.1, "bool": true}, "message": "test log message with faas and ecs fields", "@timestamp": 1662616049000000,"faas":{"coldstart":true,"execution":"6f7f0961f83442118a7af6fe80b88d56","id":"arn:aws:lambda:us-east-2:123456789012:function:custom-runtime"}}}
```

2. Send a curl request with the above request body:
```bash
curl -i -H "Content-type: application/x-ndjson" -H "Authorization: Bearer {APM_SERVER_SECRET_TOKEN}" --data-binary @<request_body_path> {APM_SERVER_URL}:8200/intake/v2/events
```

## Related issues

Closes #9100


internal/model/modeldecoder/v2/model.go
```diff
@@ -960,12 +960,54 @@ type log struct {
...
+	ServiceEnvironment nullable.String `json:"service.environment" validate:"maxLength=1024"`
+	// ServiceNodeName represents a unique node name per host for the service which
+	// originated the log line.
+	ServiceNodeName nullable.String `json:"service.node.name" validate:"maxLength=1024"`
+	// ProcessThreadName represents the name of the thread.
+	ProcessThreadName nullable.String `json:"process.thread.name" validate:"maxLength=1024"`
+	// Dataset identifies the source which originated the log line.
+	Dataset nullable.String `json:"dataset" validate:"maxLength=1024"`
+	// Labels are a flat mapping of user-defined key-value pairs.
+	Labels mapstr.M `json:"labels" validate:"inputTypesVals=string;bool;number,maxLengthVals=1024"`
```
lahsivjar, line 1010: [**For reviewers**] I have kept the input in flat JSON format, I am not sure if that will work with all ECS logging libraries or if some of the structs need to support both formats (flat and nested)

CC: @felixbarny

lahsivjar: ([Relevant comment](https://github.com/elastic/apm-server/issues/9100#issuecomment-1245207345))

felixbarny: Ideally, all properties should be supported both in a flat and a nested notation.

internal/model/error.go
```diff
@@ -42,6 +42,12 @@ type Error struct {
 	// This may be set when a stack trace cannot be parsed.
 	StackTrace string
 
+	// LogMessage holds ECS compatible error message for log event type.
+	LogMessage string
```
axw, line 46:
```suggestion
	// Message holds an error message.
	//
	// Message is the ECS field equivalent of the APM field `error.log.message`.
	Message string
```

Probably as a followup: we should think about setting `error.message` wherever we currently set `error.log.message`, and add an alias from `error.log.message` to `error.message`. See also https://github.com/elastic/apm-server/issues/3878

internal/model/event.go
```diff
@@ -38,14 +38,20 @@ type Event struct {
 	// Severity holds the numeric severity of the event for log events.
 	Severity int64
 
-	// Severity holds the action captured by the event for log events.
+	// Action holds the action captured by the event for log events.
 	Action string
+
+	// Dataset holds the the dataset which produces the events. If an event
+	// source publishes more than one type of log or events (e.g. access log,
+	// error log), the dataset is used to specify which one the event comes from
```
axw, line 46:
```suggestion
	// source publishes more than one type of log or events (e.g. access log,
	// error log), the dataset is used to specify which one the event comes from.
```

internal/model/log.go
```diff
@@ -34,10 +34,39 @@ var (
 type Log struct {
 	// Level holds the log level of the log event.
 	Level string
+	// Logger holds the name of the logger instance.
+	Logger string
+	Origin *LogOrigin
```
axw, line 39: Should these be non-pointer types, i.e. embedded structs? Using pointer types will likely force heap allocations, and it looks like we can rely on the zero values for omission.

internal/model/process.go
```diff
@@ -28,6 +28,7 @@ type Process struct {
 	Argv        []string
 	CommandLine string
 	Executable  string
+	Thread      *ProcessThread
```
axw, line 31: As commented elsewhere, should we use a non-pointer type and take the zero value to mean it should be omitted?

axw (commented):
Looks good, just left a couple of suggestions.

Handling maybe-flattened-maybe-not JSON will be interesting. I haven't looked into using it myself before, but json-iterator has an extension API which might be useful? https://pkg.go.dev/github.com/json-iterator/go#Extension

docs/spec/v2/log.json
```diff
@@ -76,12 +105,122 @@
...
+      ],
+      "maxLength": 1024
+    },
+    "transaction.id": {
+      "description": "TransactionID holds the hex encoded 64 random bits ID of the correlated transaction.",
+      "type": [
+        "null",
+        "string"
+      ],
+      "maxLength": 1024
```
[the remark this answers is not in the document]
lahsivjar: Initially, I added it but the [ECS spec](https://github.com/elastic/ecs-logging/blob/main/spec/spec.json) doesn't mention `span.id`. Not sure if any ECS library will use it.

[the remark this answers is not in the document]
axw: Right, I see. [`apmzap`](https://github.com/elastic/apm-agent-go/blob/main/module/apmzap/fields.go) does set it. @felixbarny do you have any concern about allowing `span.id` even if it's not in the ECS logging spec?

internal/model/log.go
lahsivjar: ACK on the non-pointer type. I thought about this and went with the current implementations for optional values thinking that the size of the struct won't make it escape to the heap.

>  i.e. embedded structs

Do you mean a move from nested to embedded structs? I personally find embedded structs introduce some unwanted complexity and I am not seeing a lot of benefits here (not aware of any performance difference either).

axw: > Do you mean a move from nested to embedded structs? I personally find embedded structs introduce some unwanted complexity and I am not seeing a lot of benefits here (not aware of any performance difference either).

I'm not sure why I said embedded now. Just a plain old struct field is what I actually had in mind.

lahsivjar:
> Handling maybe-flattened-maybe-not JSON will be interesting. I haven't looked into using it myself before, but json-iterator has an extension API which might be useful? https://pkg.go.dev/github.com/json-iterator/go#Extension

@axw Thanks for this. I was not aware of this and my searches also came up empty, it sounds way better than what I am doing right now which is introducing a map to decode the nested structure into and finally merging them with the flat decoded values via generated code. Something like this:

```
type ecsLogService struct {
	NestedStruct map[string]interface{} `json:"service" flattenSource:"true"`
	ServiceName nullable.String `json:"service.name" validate:"maxLength=1024"`
	ServiceVersion nullable.String `json:"service.version" validate:"maxLength=1024"`
	ServiceEnvironment nullable.String `json:"service.environment" validate:"maxLength=1024"`
	ServiceNodeName nullable.String `json:"service.node.name" validate:"maxLength=1024"`
}
```

Finally, the generator will scan the `json` tags and output a function that can be called to merge the `nested` decoded values into the flat values.

docs/spec/v2/log.json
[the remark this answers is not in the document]
felixbarny: I know that some agents do add a `span.id` so I think we should add it and consider adding it to the ECS spec. The thing is that currently the `span.id` in logs is not used in the product but that might change in the future. But either way, there's no harm in adding it. Especially since we do want to support dynamic fields going forward.

internal/model/modeldecoder/v2/model.go
lahsivjar: @felixbarny Is it correct to assume that a nested notation will have ALL fields as nested. For example:

1. Case 1:
    ```json
    {
        "service": {
            "name": "testsvc",
            "version": "v1.2.0",
            "node.name": "testnode"
        }
    }
    ```
2. Case 2:
    ```json
    {
        "service": {
            "name": "testsvc",
            "version": "v1.2.0",
            "node": {
                "name": "testnode"
            }
        }
    }
    ```
In the above 2 cases, `Case 1` will be incorrect since it has `node.name` in flat notation inside `service` which is in nested notation (the incorrect field will be silently dropped during encoding).

felixbarny: Yes, that is correct to assume - a given object path will either be fully nested or fully dotted. However, `log.level` may be dotted while there is also a nested `log` object in the same message. For example:

```json
{
  "log.level": "INFO",
  "log": { "logger": "foo" }
}
```

lahsivjar: I have updated the code to allow for both nested and flat JSON schema

felixbarny:
Another edge case that's worth discussing is how to deal with duplicate keys. As ECS loggers allow users to add arbitrary keys at the top level, the resulting JSON may contain duplicate keys, such as `log.level` appearing twice. Ideally, the JSON processing should not fail but just use the last value. But that's definitely something we can look at in a follow-up.

SylvainJuge:
When testing this PR with Java agent draft PR, I get a parsing error on the `@timestamp` field which seems to be parsed a uint64 field and not as an usual `datetime` field, is that intentional ?

It's quite easy to reproduce with the following payload:
```
{"metadata":{"process":{"pid":1234,"title":"/usr/lib/jvm/java-10-openjdk-amd64/bin/java","ppid":1,"argv":["-v"]},"system":{"architecture":"amd64","detected_hostname":"8ec7ceb99074","configured_hostname":"host1","platform":"Linux","container":{"id":"8ec7ceb990749e79b37f6dc6cd3628633618d6ce412553a552a0fa6b69419ad4"},"kubernetes":{"namespace":"default","pod":{"uid":"b17f231da0ad128dc6c6c0b2e82f6f303d3893e3","name":"instrumented-java-service"},"node":{"name":"node-name"}}},"service":{"name":"1234_service-12a3","version":"4.3.0","node":{"configured_name":"8ec7ceb990749e79b37f6dc6cd3628633618d6ce412553a552a0fa6b69419ad4"},"environment":"production","language":{"name":"Java","version":"10.0.2"},"agent":{"version":"1.10.0","name":"java","ephemeral_id":"e71be9ac-93b0-44b9-a997-5638f6ccfc36"},"framework":{"name":"spring","version":"5.0.0"},"runtime":{"name":"Java","version":"10.0.2"}},"labels":{"group":"experimental","ab_testing":true,"segment":5}}}
{"log": {"message": "test log message with timestamp", "@timestamp": "2022-10-14T14:11:54.312+0200"}}

```

felixbarny:
Great point. Yeah, the `@timestamp` field should support being provided as a ISO date or as a number. Unless the server needs to process the timestamp in some way (like when aggregating metrics on top of logs), it would be reasonable to just forward the timestamp to Elasticsearch which can handle the timestamp in both number and ISO date string format. The default date format supported by ES is `strict_date_optional_time||epoch_millis`. See also https://www.elastic.co/guide/en/elasticsearch/reference/current/date.html.

Thanks for testing this @SylvainJuge 👏

internal/model/modeldecoder/v2/model.go
sabrican3532: s***********8@gmail.com

https://user-images.githubusercontent.com/115670295/196031034-ea0fbe58-ab63-46bf-bf06-e98a43d7ec76.mp4

lahsivjar:
Thanks, @SylvainJuge, and @felixbarny for the proactive feedback and testing.

Currently, APM-Server only supports `epoch_millis` for `timestamp`. Since APM-Server batches data before sending I think sending `timestamp` string directly to Elasticsearch might not be suitable as it would lead to failures in the ES layer even though APM-Server accepted the request. I am thinking of supporting this at the APM-Server level, do you think supporting a stricter format of `strict_date_time || epoch_millis` would be enough, or do we need to support the full `strice_date_optional_time || epoch_millis`?

felixbarny:
Yes, I think `strict_date_time || epoch_millis` would be good enough.

lahsivjar:
@felixbarny Thanks for the clarification. Based on [this doc](https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-date-format.html#built-in-date-formats), IIUC, `strict_date_time` accepts only the format `yyyy-MM-dd'T'HH:mm:ss.SSSZ`. Is this correct?

felixbarny:
Yes, that's correct.

docs/spec/v2/log.json
```diff
@@ -9,6 +9,169 @@
...
     },
+    "dataset": {
+      "description": "Dataset identifies the source which originated the log line.",
+      "type": [
+        "null",
+        "string"
+      ],
+      "maxLength": 1024
+    },
+    "ecslogerrorfields": {
```
axw, line 20: I wonder if we should just not produce JSON Schema for logs for now? IIRC that's what @felixbarny suggested earlier. Otherwise we should probably ensure the embedded structs don't affect the schema.

axw (approved):
LGTM! Just a couple of minor things.

docs/spec/v2/log.json
lahsivjar: > I wonder if we should just not produce JSON Schema for logs for now?

How are the generated specs utilized? FWIW, I was planning on fixing the embedded struct handling in the schema generation code but if we don't have a use case for the schema or if it is not required for the logs then removing it sounds good to me.

axw: At least some of the agents use the schema for payload validation in tests. We used to use it in APM Server, before we generated the validation code. Let's drop it for now and come back to it if agents want to start validate logs payloads.

lahsivjar: Thanks @axw, I have removed the JSON spec generation for now.

lahsivjar:
@SylvainJuge Thanks for the earlier testing. Whenever you have time can you test again with the JAVA agent as all the required PRs are now merged.

SylvainJuge:
> @SylvainJuge Thanks for the earlier testing. Whenever you have time can you test again with the JAVA agent as all the required PRs are now merged.

I just did a quick test and it seems that not all "string-based" timestamps formats are supported, see https://github.com/elastic/apm-server/pull/9376#issuecomment-1285177363 for an example.

Issue #9181, opened 2022-09-23
Autoscale number of modelindexers to increase throughput and ensure full resource usage


simitt:
From @marclop 's findings:

> We should come up with a design that allows high throughput and to communicate back payload problems to the producing agents. Currently, we would still respond with an error if a bulk indexer failed to compress an agent's event, yet it is highly unlikely that the agent or customer is at fault for it. A better strategy would be to log those errors and decouple the time intensive operations from agent requests, since not doing so slows down the entire pipeline. A PoC with autoscaling of active indexers can be found in: https://github.com/marclop/apm-server/tree/vertical-scaling.

Autoscale the number of modelindexers up and down depending on ES and apm agent load.

marclop:
After merging #9318, it seems that we could be processing more events if we had more active indexers pulling from the internal model indexer queue. Based CPU utilization metrics for a 12 hot node 58gb RAM Elasticsearch, and APM Server indices configured to 12 shards, we aren't pushing the underlying Elasticsearch hard enough:

<img width="1436" alt="image" src="https://user-images.githubusercontent.com/7286993/195516467-fe948553-2612-43df-9b13-da4b0f418db8.png">

<img width="1436" alt="image" src="https://user-images.githubusercontent.com/7286993/195516567-091edba9-5f19-456a-929e-65a3e6c3aeab.png">

<img width="1436" alt="image" src="https://user-images.githubusercontent.com/7286993/195516725-eb19a8f6-0632-4500-b9c5-5b18097535c7.png">

Looking at the APM Server CPU usage metrics, it also looks like while we use more CPU when available (after the change to dedicated goroutine active indexer), we still aren't taking advantage of bigger instances with more CPUs

<img width="1436" alt="image" src="https://user-images.githubusercontent.com/7286993/195516884-a2a708da-98a0-4686-bd08-b9bc332bd70d.png">

Looking at this metrics, it may be that scaling up the active indexers up to `GOMAXPROCS / 3` can increase our event processing rate. Autoscaling could be performed using a mix of these metrics:

- When certain number of consecutive full flushes occurs (respecting scale up cooldowns).
- When a timed flush takes place, we need to downscale an active indexer (respecting scale down cooldown)

I think it is a good place to start for autoscaling and keep it simple.

Afterwards we could use other metrics to fine tune how autoscaling behaves: 

- How full the model indexer channel is ( `len(chan) / cap(chan) = decimal utilization` ) and also look into using that as a pressure indicator.
[...]

PR #9358, opened 2022-10-13
beater: Set semaphore, modelindexer chan from mem

### LINKED ISSUE #9182: Ensure APM Server makes full usage of available CPU resources

### LINKED ISSUE #9341: 1GB APM Server may still run out of memory
**APM Server version** (`apm-server version`): 8.2+

When Elasticsearch ingestion is significantly lower than the APM processing rate, and given enough CPU processing power for APM Server to decode `200` concurrent intake requests, the APM Server may run out of memory in some cases.

**Steps to reproduce**:

 1. Create a CPU Optimized ESS deployment with a small backing Elasticsearch (16gb) and a 1GB APM Server.
 2. Run `apmbench` against it
 3. Watch it run out of memory.

marclop:
## Motivation/summary

Update the MaxConcurrentDecoders, the modelindexer internal channel and the MaxRequests based on the total amount of memory that the APM server can access. It looks at cgroups if available, otherwise, falls back to the total system memory.

Updates the default semaphore size to `128` from `200`, the default modelindexer internal queue size to `1024` from `100` and the modelindexer.MaxRequests to `10` down from `50`.

This fixes a couple of issues; 1GB instances could still OOM if enough concurrency was used (> 80 concurrent agents sending data as fast as possible), and another issue where APM Server doesn't use the total available memory to its advantage.

## Checklist

## How to test these changes

Benchmark with different sizes and observe more resource utilization when available.

## Related issues

Closes #9182 
Closes #9341


internal/beater/beater.go
```diff
@@ -874,6 +936,13 @@ func (s *serverRunner) newFinalBatchProcessor(
 	return indexer, indexer.Close, nil
 }
 
+func eventBufferSize(memLimit float64) int {
+	if memLimit > 1 {
+		return int(256 * memLimit)
```
marclop, line 941: Memory usage is till pretty low, we could likely increase from `256` to `512` or even further.

marclop:
# Initial benchmarks

I ran a benchmark matrix with different APM Server instances (1, 2, 8, 15 and 30 gigabytes of RAM) with a constant Elasticsearch deployment of 15 nodes (3 dedicated masters, 12 hot nodes) and 12 shards for all the APM data streams. Peak `events/sec` was reached at and above `8g` and bigger instances didn't seem to improve the performance much.

## CPU

image

## Memory

image

## Network

image

## `benchstat` comparison

WIP.

marclop:
After merging #9393, all instances had 50 available indexers. Because of how we cycle through them, 1GB and 2GB APM instances were running out of memory. I've modified the PR to also size the number of available indexers based on RAM, so that we can avoid unpleasant OOMs in smaller instances.

Taking profiles gave me the clue that I needed, after sizing the 1GB instance to 10 available indexers

```
heap profile: 377: 74804032 [681456: 883525128] @ heap/1048576
10: 41943040 [10: 41943040] @ 0xaaaab7611238 0xaaaab7610b8c 0xaaaab7611018 0xaaaab83fa898 0xaaaab83fcb04 0xaaaab83fdb68 0xaaaab7956c00 0xaaaab7578d64
#	0xaaaab7611237	bytes.makeSlice+0x57											bytes/buffer.go:229
#	0xaaaab7610b8b	bytes.(*Buffer).grow+0x10b										bytes/buffer.go:142
#	0xaaaab7611017	bytes.(*Buffer).ReadFrom+0x47										bytes/buffer.go:202
#	0xaaaab83fa897	github.com/elastic/apm-server/internal/model/modelindexer.(*bulkIndexer).Flush+0x327			github.com/elastic/apm-server/internal/model/modelindexer/bulk_indexer.go:178
#	0xaaaab83fcb03	github.com/elastic/apm-server/internal/model/modelindexer.(*Indexer).flush+0x253			github.com/elastic/apm-server/internal/model/modelindexer/indexer.go:408
#	0xaaaab83fdb67	github.com/elastic/apm-server/internal/model/modelindexer.(*Indexer).runActiveIndexer.func2+0x37	github.com/elastic/apm-server/internal/model/modelindexer/indexer.go:535
#	0xaaaab7956bff	golang.org/x/sync/errgroup.(*Group).Go.func1+0x5f							golang.org/x/sync@v0.0.0-20220819030929-7fc1605a5dde/errgroup/errgroup.go:75
```

It turns out that reading the Elasticsearch bulk responses can take quite a bit of memory, which makes sense since we're sending pretty big bulk requests (10K+ events). https://github.com/elastic/apm-server/blob/8e27af7e9abb93e1b2dd2145404441610397c251/internal/model/modelindexer/bulk_indexer.go#L178-L180

internal/beater/memlimit_system.go
```diff
@@ -0,0 +1,38 @@
...
+
+//go:build !darwin
+// +build !darwin
+
+package beater
+
+import (
+	"fmt"
+
+	"github.com/elastic/gosigar"
```
axw, line 26: Should we be using go-sysinfo? With Joseph's changes, it provides non-cgo implementations that will work on all our supported platforms.

systemtest/intake_async_test.go
```diff
@@ -26,19 +26,19 @@ import (
...
+		srv := apmservertest.NewServerTB(t, "-E", "apm-server.max_concurrent_decoders=5")
 
 		systemtest.SendBackendEventsAsyncPayload(t, srv.URL, `../testdata/intake-v2/errors.ndjson`)
 		// Ensure the 5 errors are ingested.
 		systemtest.Elasticsearch.ExpectMinDocs(t, 5, "logs-apm.error-*", nil)
 
 		// Send a request with a lot of events (~1920) and expect it to be processed
 		// without any errors.
-		systemtest.SendBackendEventsAsyncPayload(t, srv.URL, `../testdata/intake-v2/heavy.ndjson`)
+		systemtest.SendBackendEventsAsyncPayloadError(t, srv.URL, `../testdata/intake-v2/heavy.ndjson`)
```
axw, line 41: Why is this failing now? The comment above doesn't match.

internal/beater/memlimit_system.go
```diff
@@ -0,0 +1,38 @@
...
+)
+
+// systemMemoryLimit returns 50% of the total system memory in GigaBytes.
```
axw, line 29: It's surprising to me that this function is halving the value - that tightly couples it to how it's used. Should we do that bit in the caller?

internal/beater/beater.go
```diff
@@ -177,6 +177,33 @@ func (s *Runner) Run(ctx context.Context) error {
...
 	}
 
+	// Obtain the memory limit for the APM Server process. Certain config
+	// values will be sized according to the maximum memory set for the server.
+	var memLimit float64
+	if cgroupReader := newCgroupReader(); cgroupReader != nil {
+		if limit, err := cgroupMemoryLimit(cgroupReader); err != nil {
+			s.logger.Warn(err)
+		} else {
+			memLimit = float64(limit) / 1024 / 1024 / 1024
```
axw, line 187: Did you say you were going to limit the use of cgroup memory, to allow for other processes in the container?

internal/beater/beater.go
```diff
@@ -637,6 +685,37 @@ func (s *Runner) newFinalBatchProcessor(
...
+	if opts.MaxRequests > 0 {
+		return opts
+	}
+	// This formula yields the following max requests for APM Server sized:
+	// 1	2 	4	8	15	30
+	// 10	13	16	22	32	55
+	maxRequests := int(float64(10) + memLimit*1.5)
+	if maxRequests > 60 {
+		maxRequests = 60
+	}
```
axw, line 711: How did you arrive at these formulae?

If we assumed the max event size (in bytes on the wire) could fit exactly into as many bytes in memory, then isn't a channel size of 2048 for 1GB memory is too much? The default max event size is 300KB. `2048 * 300 * 1024 == 629145600`, which is (substantially) greater than `1024 * 1024 * 1024 == 1073741824`.

What do you think about defining the defaults in terms of:
 - some percentage of available memory
 - maximum event size * (in-memory model objects overhead factor)

?

marclop: I think that is a big higher than I intended the 1GB instance to hold, however, it would still be within the limit I think?

1,073,741,824 limit in bytes
0,629,145,600 current buffer max usage in bytes (based on 300kb event size). (600MB)

I think it would be safer to to reduce the base queue by 1024 so we end up with up to 300MB of cached events in the queue (And allows some overhead for the in-memory representation).

> How did you arrive at these formulae?

I did some calculations on paper and then adjusted them observing the behavior of benchmarks. The queue didn't have a big impact in throughput to be honest, but It should allow a bit of data to be cached (in practice, it would be less than 1s at maximal throughput, but that's not what the queue is there for).

Modelindexer bulk indexer size = ~10MB (based on some profiles taken that indicated that with 50 available bulk requests, the buffers grew to ~300MB / 50 = ~6MB. per available indexer).

So for a 1GB APM Instance
- Max requests of 128 * 10 * 300kb = 375MB
- Available bulk indexers 11 * 6 = 66MB
- Modelindexer internal queue = 1024 * 300kb = 300MB
- +25% error for other aspects of the server and not consuming all the cgroup memory = 816MB

systemtest/intake_async_test.go
marclop: It's because the semaphore wasn't being set to what I intended, so it ended up with the default semaphore size, which is not that test's aim

internal/beater/beater.go
marclop: I capped it to 80% of the cgroup memory limit

marclop:
## Benchstat

Overall there's a clear improvement in throughput with this PR, I believe as a product of the increased semaphore + autoscaling + tuned buffers. All instances showed improvement, and I believe that 30g shows less of an improvement due to the Elasticsearch deployment size? (although the deployment is 580g x 3z).

### 1g

Main ran out of memory, while PR diff didn't.

## 2g

```console
$ benchstat -alpha 1 main/2g-30s-30n.txt pr/2g-30s-30n.txt
name          old time/op                  new time/op                  delta
AgentAll-256                   1.91s ±24%                   1.61s ±27%  -15.64%  (p=0.400 n=3+3)
AgentAll-256                   9.77k ±28%                  11.64k ±25%  +19.20%  (p=0.400 n=3+3)
name          old gc_cycles                new gc_cycles                delta
AgentAll-256                    224 ±118%                     438 ±55%  +95.68%  (p=0.200 n=3+3)
AgentAll-256                     278 ±81%                     349 ±54%     ~     (p=1.000 n=3+3)
AgentAll-256                   1.08G ±14%                   0.77G ±27%  -28.09%  (p=0.100 n=3+3)
AgentAll-256                   9.86M ±14%                   5.54M ±25%  -43.85%  (p=0.100 n=3+3)
AgentAll-256                   1.18G ±13%                   0.89G ±24%  -24.53%  (p=0.100 n=3+3)
AgentAll-256                    41.4 ±26%                     6.2 ±59%  -84.93%  (p=0.100 n=3+3)
AgentAll-256                   559MB ± 1%                   570MB ± 1%   +1.98%  (p=0.100 n=3+3)
AgentAll-256                   7.90M ± 0%                   7.97M ± 1%   +0.92%  (p=0.200 n=3+3)
```

### 4g

```console
$ benchstat -alpha 1 main/4g-30s-30n.txt pr/4g-30s-30n.txt
name          old time/op                  new time/op                  delta
AgentAll-512                   1.25s ± 0%                   1.07s ± 1%  -14.12%  (p=0.100 n=3+3)
AgentAll-512                   14.4k ± 0%                   16.8k ± 1%  +16.44%  (p=0.100 n=3+3)
name          old gc_cycles                new gc_cycles                delta
AgentAll-512                     159 ± 1%                     199 ± 1%  +25.10%  (p=0.100 n=3+3)
AgentAll-512                     211 ±10%                     223 ± 7%   +6.01%  (p=0.400 n=3+3)
AgentAll-512                   1.05G ± 1%                   0.98G ± 0%   -6.33%  (p=0.100 n=3+3)
AgentAll-512                   9.41M ± 1%                   7.01M ± 1%  -25.50%  (p=0.100 n=3+3)
AgentAll-512                   1.16G ± 2%                   1.14G ± 0%   -1.76%  (p=0.400 n=3+3)
AgentAll-512                    45.4 ± 0%                    10.8 ± 1%  -76.32%  (p=0.100 n=3+3)
name          old metrics/sec              new metrics/sec              delta
AgentAll-512                   2.21k ± 0%                   2.57k ± 1%  +16.40%  (p=0.100 n=3+3)
AgentAll-512                   579MB ± 0%                   566MB ± 0%   -2.28%  (p=0.100 n=3+3)
AgentAll-512                   8.07M ± 0%                   7.95M ± 0%   -1.50%  (p=0.100 n=3+3)
```

### 8g

```console
$ benchstat -alpha 1 main/8g-30s-30n.txt pr/8g-30s-30n.txt
name           old time/op                  new time/op                  delta
AgentAll-1024                   610ms ±16%                   350ms ± 1%  -42.67%  (p=0.100 n=3+3)
name           old errors/sec               new errors/sec               delta
AgentAll-1024                     599 ±18%                    1029 ± 1%  +71.83%  (p=0.100 n=3+3)
name           old events/sec               new events/sec               delta
AgentAll-1024                   30.0k ±18%                   51.5k ± 1%  +71.83%  (p=0.100 n=3+3)
name           old gc_cycles                new gc_cycles                delta
AgentAll-1024                     286 ±14%                     278 ±11%   -2.68%  (p=0.700 n=3+3)
name           old max_goroutines           new max_goroutines           delta
AgentAll-1024                     410 ±28%                     561 ± 4%  +36.86%  (p=0.100 n=3+3)
name           old max_heap_alloc           new max_heap_alloc           delta
AgentAll-1024                   1.19G ± 5%                   2.07G ± 5%  +73.42%  (p=0.100 n=3+3)
name           old max_heap_objects         new max_heap_objects         delta
AgentAll-1024                   10.5M ± 8%                   14.0M ± 5%  +33.12%  (p=0.100 n=3+3)
name           old max_rss                  new max_rss                  delta
AgentAll-1024                   1.32G ± 6%                   2.31G ± 4%  +74.57%  (p=0.100 n=3+3)
name           old mean_available_indexers  new mean_available_indexers  delta
AgentAll-1024                    39.2 ±11%                     1.4 ±10%  -96.43%  (p=0.100 n=3+3)
name           old alloc/op                 new alloc/op                 delta
AgentAll-1024                   580MB ± 1%                   573MB ± 0%   -1.20%  (p=0.100 n=3+3)
name           old allocs/op                new allocs/op                delta
AgentAll-1024                   8.09M ± 0%                   8.01M ± 0%   -0.90%  (p=0.100 n=3+3)
```

### 15g

```console
$ benchstat -alpha 1 main/15g-30s-30n.txt pr/15g-30s-30n.txt
name           old time/op                  new time/op                  delta
AgentAll-1920                   371ms ± 0%                   323ms ± 0%   -12.82%  (p=0.100 n=3+3)
name           old events/sec               new events/sec               delta
AgentAll-1920                   48.6k ± 0%                   55.7k ± 0%   +14.59%  (p=0.100 n=3+3)
name           old gc_cycles                new gc_cycles                delta
AgentAll-1920                     405 ± 1%                     148 ± 9%   -63.34%  (p=0.100 n=3+3)
name           old max_goroutines           new max_goroutines           delta
AgentAll-1920                     548 ± 3%                     607 ± 5%   +10.83%  (p=0.100 n=3+3)
name           old max_heap_alloc           new max_heap_alloc           delta
AgentAll-1920                   1.21G ± 1%                   3.84G ± 4%  +216.61%  (p=0.100 n=3+3)
name           old max_heap_objects         new max_heap_objects         delta
AgentAll-1920                   10.8M ± 2%                   26.9M ± 3%  +149.36%  (p=0.100 n=3+3)
name           old max_rss                  new max_rss                  delta
AgentAll-1920                   1.36G ± 0%                   4.29G ± 1%  +214.90%  (p=0.100 n=3+3)
name           old mean_available_indexers  new mean_available_indexers  delta
AgentAll-1920                    31.2 ± 2%                     7.6 ± 4%   -75.80%  (p=0.100 n=3+3)
name           old alloc/op                 new alloc/op                 delta
AgentAll-1920                   567MB ± 0%                   568MB ± 0%    +0.23%  (p=0.100 n=3+3)
name           old allocs/op                new allocs/op                delta
AgentAll-1920                   7.98M ± 0%                   7.99M ± 0%    +0.15%  (p=0.700 n=3+3)
```

### 30g

```console
$ benchstat -alpha 1 main/30g-30s-30n.txt pr/30g-30s-30n-2048d.txt
name           old time/op                  new time/op                  delta
AgentAll-3840                   181ms ± 1%                   173ms ± 3%    -4.27%  (p=0.008 n=5+5)
name           old events/sec               new events/sec               delta
AgentAll-3840                   99.3k ± 0%                  103.6k ± 3%    +4.34%  (p=0.008 n=5+5)
name           old gc_cycles                new gc_cycles                delta
AgentAll-3840                     725 ± 0%                     313 ±28%   -56.76%  (p=0.004 n=5+6)
name           old max_goroutines           new max_goroutines           delta
AgentAll-3840                   1.02k ± 1%                   1.05k ±10%    +2.75%  (p=0.667 n=5+5)
name           old max_heap_alloc           new max_heap_alloc           delta
AgentAll-3840                   1.50G ± 2%                   4.09G ±25%  +171.76%  (p=0.004 n=5+6)
name           old max_heap_objects         new max_heap_objects         delta
AgentAll-3840                   13.6M ± 2%                   30.6M ±27%  +124.10%  (p=0.004 n=5+6)
name           old max_rss                  new max_rss                  delta
AgentAll-3840                   1.73G ± 1%                   4.64G ±24%  +168.33%  (p=0.004 n=5+6)
name           old mean_available_indexers  new mean_available_indexers  delta
AgentAll-3840                    15.9 ± 1%                    15.3 ±38%    -3.65%  (p=0.758 n=5+6)
name           old alloc/op                 new alloc/op                 delta
AgentAll-3840                   558MB ± 0%                   561MB ± 1%    +0.46%  (p=0.121 n=5+6)
name           old allocs/op                new allocs/op                delta
AgentAll-3840                   7.89M ± 0%                   7.88M ± 1%    -0.14%  (p=0.909 n=5+6)
```

internal/beater/beater.go
axw: > I think that is a big higher than I intended the 1GB instance to hold, however, it would still be within the limit I think?
>
> 1,073,741,824 limit in bytes
> 0,629,145,600 current buffer max usage in bytes (based on 300kb event size). (600MB)

:man_facepalming: indeed, I was just off by an order of magnitude.

internal/beater/beater.go
```diff
@@ -433,6 +462,25 @@ func (s *Runner) Run(ctx context.Context) error {
...
+	if memLimit > 1 {
+		// Limit the number of concurrent decodes to 2048
+		max := uint(128 * memLimit)
+		if max > 2048 {
+			return 2048
+		}
+		return max
+	}
+	return defaultDecoders
+}
```
axw, line 482: A couple of things:
 - I think if a user sets a value explicitly, we should always use that. In the case where they explicitly set it to the default value, it shouldn't be changed based on memory.
 - I think we should assume memLimit(GB) is always valid (set to something conservative, in case of errors)

With those changes, we can get rid of `defaultDecoders` and simplify this:

```suggestion
func maxConcurrentDecoders(memLimitGB float64, decoders uint) uint {
	// If the decoders has been explicitly set, just return that value.
	if decoders > 0 {
		return decoders
	}
	// Allow 128 concurrent decoders for each 1GB memory, limited to at most 2048.
	const max = 2048
	decoders = uint(128 * memLimit)
	if decoders > max {
		decoders = max
	}
	return decoders
}
```

internal/beater/beater.go
```diff
@@ -177,6 +177,36 @@ func (s *Runner) Run(ctx context.Context) error {
...
+		s.logger.Info("no cgroups detected, falling back to total system memory")
+		if limit, err := systemMemoryLimit(); err != nil {
+			s.logger.Warn(err)
+		} else {
+			// If no cgroup limit is set, only return 50% of the total memory.
+			// to have a margin of safety for other processes.
+			memLimit = float64(limit) / 1024 / 1024 / 1024 * 0.5
+		}
+	}
+	if memLimit > 0 {
```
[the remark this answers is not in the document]
axw: Also, perhaps only call maxConcurrentDecoders (probably renamed to include "memory" in the name somewhere) in case MaxConcurrentDecoders hasn't been explicitly specified. The log message below is a bit misleading otherwise, as it's not based on the memory.

internal/beater/config/config.go
```diff
@@ -159,6 +161,6 @@ func DefaultConfig() *Config {
 		AgentAuth:             defaultAgentAuth(),
 		JavaAttacherConfig:    defaultJavaAttacherConfig(),
 		WaitReadyInterval:     5 * time.Second,
-		MaxConcurrentDecoders: 200,
+		MaxConcurrentDecoders: 128,
```
axw, line 164: Should we set this to zero by default, with zero meaning it is set based on memory?

axw (changes requested):
The magic numbers (in modelIndexerConfig in particular) make me a bit uncomfortable, but this is a clear improvement so I'm good with merging it and iterating as needed.

Just a few more minor things, otherwise looks good.

internal/beater/tracing.go
```diff
@@ -30,8 +30,7 @@ import (
 	"github.com/elastic/apm-server/internal/model"
 )
 
-func newTracerServer(listener net.Listener, logger *logp.Logger, batchProcessor model.BatchProcessor) (*http.Server, error) {
-	cfg := config.DefaultConfig()
+func newTracerServer(cfg *config.Config, listener net.Listener, logger *logp.Logger, batchProcessor model.BatchProcessor) (*http.Server, error) {
```
axw, line 33: This was a bit surprising to me, since the modelindexer is passed in. I suppose this is needed for the MaxConcurrentDecoders though, is that right?

This seems like a bit of a problem: it means when self-instrumentation is enabled, there could be 2x the concurrent decoders (1x from external clients, 1x from the server itself). I think this is another reason why the modelindexer should control the rate of input.

(No action required at the moment.)

axw (approved):
LGTM, thanks! Could you please open an issue to follow up on modelIndexerConfig and maxConcurrentDecoders, to take the max event size and flush size into account?

internal/beater/tracing.go
marclop: Absolutely right. I thought about using a smaller arbitrary value instead, but like you say I think we can address the problem by encapsulating the semaphore in a different manner and have the model indexer act as the controller.

axw:
I think we can test this as part of https://github.com/elastic/apm-server/issues/9181 and https://github.com/elastic/apm-server/issues/9341

PR #9359, opened 2022-10-13
Fix wrong punctuation for arguments in curl commands

LHB6540:
## Motivation/summary
Removed the redundant symbol ' " ' carried in the curl parameter in the example. 
In the original example, this will cause the id of the sourcemap managed by kibana to have more escape characters when uploading by referring to the example, which will cause the sourcemap to map to the error stack and cannot work. For example The correct id should be "apm-foo-1.0-xxx", the original example would result in an id of "apm:\"demo\"-\"1.0\-xxx"".

## Checklist
- [ ] Update [source-map-how-to.asciidoc](https://github.com/elastic/apm-server/blob/main/docs/source-map-how-to.asciidoc)

## Related issues
 This issue is mentioned in [sourcemap to error stack doesn't work](https://github.com/elastic/apm-agent-rum-js/issues/1289)


docs/source-map-how-to.asciidoc
```diff
@@ -124,10 +124,10 @@ curl -X POST "http://localhost:5601/api/apm/sourcemaps" \
 -H 'Content-Type: multipart/form-data' \
 -H 'kbn-xsrf: true' \
 -H 'Authorization: ApiKey ${YOUR_API_KEY}' \ <2>
--F 'service_name="foo"' \
--F 'service_version="$SERVICEVERSION"' \
--F 'bundle_filepath="/test/e2e/general-usecase/bundle.js.map"' \
--F 'sourcemap="@./dist/app.min.js.map"'
+-F 'service_name=foo' \
+-F 'service_version=$SERVICEVERSION' \
+-F 'bundle_filepath=/test/e2e/general-usecase/bundle.js.map' \
```
andriijas, line 129: I dont think the bundle_path should be pointing the .map but the actual .js file.

The docs is confusing because it shows .js in one example and .map in one example.

While we are at this I think someone need to clarify this

devcorpio: you are right, it should point to the .js file 👍

simitt: @LHB6540 while you're at it, would you mind also removing the `map` part from `bundle_filepath`?

LHB6540: > @LHB6540 while you're at it, would you mind also removing the `map` part from `bundle_filepath`?
When I actually use it, bundle_filepath actually points to the js file instead of the js.map file. I agree with the above, it should be a bug in the documentation. Should I revoke this pr and submit it after the revision or open another pr for this error?

simitt: It is a pure doc issue, so changing to `-F 'bundle_filepath=/test/e2e/general-usecase/bundle.js' \` with this PR would be ok.

LHB6540: > It is a pure doc issue, so changing to `-F 'bundle_filepath=/test/e2e/general-usecase/bundle.js' \` with this PR would be ok.
done, commit: https://github.com/elastic/apm-server/pull/9359/commits/e0277492ac9c1bf7e30e03f10cd7c795c1eff74a

simitt (approved):

simitt (approved):

LHB6540:
@simitt I don't understand what this means, is there anything I need to fix?
elasticsearch-ci/docs Expected — Waiting for status to be reported
image

simitt:
@bmorelli25 could you take a look at the pending `elasticsearch-ci/docs` job?

axw:
run elasticsearch-ci/docs

axw:
The `elasticsearch-ci/docs` task only runs for PRs created by Elastic employees, or when one of us triggers it with the magic words "run elasticsearch-ci/docs" :)

Thanks for your contribution @LHB6540!

PR #9363, opened 2022-10-14
apmbench: Increase `waitInactiveTimeout` to `60s`

marclop:
## Motivation/summary

Double the `waitInactiveTimeout` since we are sometimes exceeding the `30s`. In particular when warming up bigger instances.

## Checklist

## How to test these changes

N/A

## Related issues

Part of #9182


lahsivjar (approved):
LGTM. A non-blocking question:

> Double the waitInactiveTimeout since we are sometimes exceeding the 30s. In particular when warming up bigger instances.

Do we see timeout exceeded due to the size of warmup events for bigger instances or does it have something to do with the instance size itself?

marclop:
@lahsivjar It's mostly to do with bigger instances. Since bigger instances can process and send more background flushes to Elasticsearch, I've found that the warm up failed a few times in particular when benchmarking `30g` of RAM.

PR #9382, opened 2022-10-17
modelindexer: don't allow Close to be blocked

axw:
## Motivation/summary

Remove all mutexes from `ProcessBatch`, making it impossible for `ProcessBatch` to block `Close`. Now if `Close` is called while `ProcessBatch` is blocked waiting to enqueue an item, `ProcessBatch` will return `ErrClosed`.

## Checklist

## How to test these changes

N/A

## Related issues

None

### DIFF
--- a/internal/model/modelindexer/indexer.go
+++ b/internal/model/modelindexer/indexer.go
@@ -82,7 +82,7 @@ type Indexer struct {
 	logger                *logp.Logger
 	available             chan *bulkIndexer
 	bulkItems             chan elasticsearch.BulkIndexerItem
-	errgroup              *errgroup.Group
+	errgroup              errgroup.Group
 	errgroupContext       context.Context
 	cancelErrgroupContext context.CancelFunc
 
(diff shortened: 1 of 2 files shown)


internal/model/modelindexer/indexer.go
```diff
@@ -167,9 +168,10 @@ func New(client elasticsearch.Client, cfg Config) (*Indexer, error) {
...
 	defer i.mu.Unlock()
-	if !i.closing {
-		i.closing = true
-		close(i.bulkItems)
+	select {
+	case <-i.closing:
+	default:
+		close(i.closing)
 		// Close i.closed when ctx is cancelled,
 		// unblocking any ongoing flush attempts.
```
lahsivjar, line 176: Instead of using `i.closed` will it be more readable to use errgroup context and propagate it? (In this case `i.closed` can be signaled by the errgroup context being canceled) and in `Close` we can just have our goroutine run as:

```go
i.g.Go(func() error {
    select {
    case <-done:
    case <-ctx.Done():
        return context.Canceled
    }
})
```

axw: That sounded like a great idea until I realised that we need to wait for errgroup.Wait to return in Close. The act of adding that Go call would cause Close to always block until the context passed to Close is cancelled.

axw: We could create a separate cancellable context for all flush calls though, which is cancelled by Close when the Close context is cancelled. That might be neater than what we have now.

lahsivjar: > until I realised that we need to wait for errgroup.Wait to return in Close

Ah, yes.

> We could create a separate cancellable context for all flush calls though, which is cancelled by Close when the Close context is cancelled. That might be neater than what we have now.

This looks good, I can follow it easily now. Thanks 😄

lahsivjar (approved):
LGTM

PR #9383, opened 2022-10-17
Enable Service Metrics by default

simitt:
## Motivation/summary

In `8.5` we introduce service metrics aggregation as an experimental feature, disabled by default.
We now want to enable it by default to make it easier to use and gather feedback. 

## Checklist

## How to test these changes

0. Enable APM Server and ingest some events. 
1. Check that service metrics events are created (`metricset.name: service`).


axw:
I realise this is in draft, so maybe you were already going to do this: I would expect that we would have some data on the feature (cost & benefit) before enabling it by default. Users can already opt in, and IIANM it is (or should be) enabled in the edge cluster.

simitt:
I had not planned that, but you raise a good point, and I can look into the storage increase caused by service metrics.

simitt:
Closing this until we have further data for justifying this feature.

PR #9388, opened 2022-10-18
testing: use "self" for observability by default

axw:
## Motivation/summary

Upgrade terraform ec provider to 0.5.0.

Replace the `ec_deployment` module's "monitor_deployment" var with "observability_deployment", which is either the ID of an existing deployment, "self" to send logs and metrics to the same deployment, "dedicated" to create a dedicated deployment for sending logs and metrics, or "none" to disable observability.

For `ec_deployment`, "none" is the default. For `testing/cloud`, "self" is the default.

## Checklist

## How to test these changes

`cd testing/cloud && EC_API_KEY=... make apply`

Check that stack monitoring data is available.

## Related issues

None


marclop:
In general I'm in favor of this change, but I am a bit worried on the impact that it may have for Benchmarking. Since we're self-monitoring the deployment it may divert some CPU time from the benchmarking itself into the monitoring functions.

Did you see if there was a meaningful difference in performance with the monitoring turned on?

axw:
@marclop good call, I didn't think about that. I don't think we should enable this in the benchmarking case. I'll mark this as draft while I think about how to handle that. Maybe opt-out of self monitoring.

testing/infra/terraform/modules/ec_deployment/variables.tf
```diff
@@ -24,10 +24,10 @@ variable "stack_version" {
...
 }
 
-variable "monitor_deployment" {
-  default     = false
-  type        = bool
-  description = "Optionally monitor the deployment in a separate deployment"
+variable "observability_deployment" {
+  default     = "none"
+  type        = string
+  description = "Optional deployment ID, for platform observability. Use 'self' to send data to the same deployment, 'dedicated' to create a dedicated observability deployment, or 'none' (the default) to disable observability. Any other value is assumed to be the ID of an existing, externally managed Elasticsearch deployment."
```
marclop, line 30: Can we add a validation rule for this variable? Since the values are fixed (or deployment ID which is of length 32), and are expected to be one of a list, it fits the `validation` use case quite nicely: https://developer.hashicorp.com/terraform/language/values/variables#custom-validation-rules.

```suggestion
  description = "Optional deployment ID, for platform observability. Use 'self' to send data to the same deployment, 'dedicated' to create a dedicated observability deployment, or 'none' (the default) to disable observability. Any other value is assumed to be the ID of an existing, externally managed Elasticsearch deployment."
  validation {
    condition     = can(regex("^self$|^dedicated$|^none$", var.observability_deployment)) || length(var.observability_deployment) == 32
    error_message = "invalid observability_deployment value. Must be one of 'self','dedicated','none', '<deployment id>'"
  }
```

testing/cloud/variables.tf
```diff
@@ -42,6 +42,12 @@ variable "elasticsearch_zone_count" {
   description = "Optional Elasticsearch zone count"
 }
 
+variable "observability_deployment" {
+  default     = "self"
+  type        = string
+  description = "Optional deployment ID, for platform observability. Use 'self' (the default) to send data to the same deployment, 'dedicated' to create a dedicated observability deployment, or 'none' to disable observability. Any other value is assumed to be the ID of an existing, externally managed Elasticsearch deployment."
```
marclop, line 48: Can we add a validation rule for this variable? Since the values are fixed (or deployment ID which is of length 32), and are expected to be one of a list, it fits the `validation` use case quite nicely: https://developer.hashicorp.com/terraform/language/values/variables#custom-validation-rules.

```suggestion
  description = "Optional deployment ID, for platform observability. Use 'self' (the default) to send data to the same deployment, 'dedicated' to create a dedicated observability deployment, or 'none' to disable observability. Any other value is assumed to be the ID of an existing, externally managed Elasticsearch deployment."
  validation {
    condition     = can(regex("^self$|^dedicated$|^none$", var.observability_deployment)) || length(var.observability_deployment) == 32
    error_message = "invalid observability_deployment value. Must be one of 'self','dedicated','none', '<deployment id>'"
  }
```

marclop (approved):
Looks good! I think it'd be better to add input validation to the `observability_deployment` variable since the input is a bit complex for that one.

testing/infra/terraform/modules/ec_deployment/variables.tf
axw: Thanks for the tip, will do.

PR #9393, opened 2022-10-18
modelindexer: Scale active indexers based on load

### LINKED ISSUE #9181: Autoscale number of modelindexers to increase throughput and ensure full resource usage

marclop:
## Motivation/summary

Modifies the `modelindexer` to scale up and down the number of "active" indexers (goroutines pulling out of the internal modelindexer queue), based on consecutive number of flushes up to 25% of the `GOMAXPROCS`.

When N number of full flushes (Reached or exceeded `FlushBytes`) occur, a new active indexer will be created by the scale action, as long as the scaling cooldown has elapsed.

Equally, when N number of timed flushes (due to `FlushInterval`) occur, an active indexer will be scaled down, since not enough load is going through the server to warrant the current number of active indexers.

Active indexer downscaling can also be triggered due to a change in the `GOMAXPROCS`. This is particularly important for containerized or cgroup environments where CPU quotas may be updated "live". In this case, the downscale cooldown is ignored and active indexers are scaled down until until `active <= math.RoundToEven(GOMAXPROCS / 4)`.

When traffic isn't going through an active indexer, a new timer has been introduced to allow completely idle indexers to be scaled back. The idle check interval can be configured via `IdleInterval`.

Scaling is enabled by default, but can be disabled via the configuration option `output.elasticsearch.scaling.enabled: false`.

Last, modifies the default settings for `max_requests` and `flush_bytes` to be `50` and `1MB` respectively. This allows smaller payloads to be sent to Elasticsearch, more available indexers can be used by instances with more processing power, and the indexers are cycled faster, which results in better usage and performance.

## Checklist

## How to test these changes

Run benchmarks on instances `8g` and bigger in ESS and observe CPU utilization reaching ~100%.

## Related issues

Closes #9181


marclop:
## Benchstat diff

### 1 to 4g APM Server

All had similar performance to main, since the active indexers were only scaled up briefly when CPU credits were available which to significant bursts in throughput when that occurred.

### 8g APM Server

This one has less of a noticeable increase, but the CPU was utilized more fully. Bursts had way higher throughput.

```console
$ benchstat -alpha 0.11 sizes/main-active/8g-12s-12n.txt sizes/scaling/8g-24s-24n1mb-50av-gomaxprocs-by-4.txt
name          old time/op                  new time/op                  delta
AgentAll-512                   606ms ± 1%                   593ms ±25%      ~     (p=0.700 n=3+3)

name          old error_responses/sec      new error_responses/sec      delta
AgentAll-512                    0.03 ±58%                    0.00       -100.00%  (p=0.100 n=3+3)

AgentAll-512                   29.5k ± 1%                   31.5k ±28%      ~     (p=0.700 n=3+3)

name          old gc_cycles                new gc_cycles                delta
AgentAll-512                     318 ± 1%                     312 ±26%      ~     (p=0.700 n=3+3)

AgentAll-512                     382 ± 2%                     398 ±29%      ~     (p=0.700 n=3+3)

AgentAll-512                   1.11G ± 2%                   1.15G ± 5%      ~     (p=0.200 n=3+3)

AgentAll-512                   10.0M ± 2%                   10.1M ± 6%      ~     (p=1.000 n=3+3)

AgentAll-512                   1.22G ± 1%                   1.26G ± 5%      ~     (p=0.400 n=3+3)

AgentAll-512                    16.1 ± 1%                    40.7 ± 8%  +153.35%  (p=0.100 n=3+3)

AgentAll-512                   570MB ± 1%                   581MB ± 1%    +2.03%  (p=0.100 n=3+3)

AgentAll-512                   7.99M ± 1%                   8.15M ± 1%    +2.09%  (p=0.100 n=3+3)
```

### 15g APM Server

From this size onwards is where the autoscaling really shines. CPU utilization was increased and so was throughput.

```console
$ benchstat -alpha 0.11 sizes/main-active/15g-12s-12n.txt sizes/scaling/15g-24s-24n1mb-50av-gomaxprocs-by-4.txt
name          old time/op                  new time/op                  delta
AgentAll-960                   561ms ±10%                   336ms ± 0%   -40.08%  (p=0.100 n=3+3)

name          old error_responses/sec      new error_responses/sec      delta
AgentAll-960                    0.06 ±85%                   0.00 ±200%   -95.79%  (p=0.100 n=3+3)

AgentAll-960                   31.8k ±10%                   53.6k ± 0%   +68.34%  (p=0.100 n=3+3)

name          old gc_cycles                new gc_cycles                delta
AgentAll-960                     317 ± 3%                     448 ± 3%   +41.28%  (p=0.100 n=3+3)

AgentAll-960                     386 ± 3%                     562 ± 2%   +45.64%  (p=0.100 n=3+3)

AgentAll-960                   1.11G ± 0%                   1.19G ± 1%    +7.72%  (p=0.100 n=3+3)

AgentAll-960                   9.89M ± 1%                  10.67M ± 0%    +7.91%  (p=0.100 n=3+3)

AgentAll-960                   1.23G ± 2%                   1.33G ± 1%    +7.84%  (p=0.100 n=3+3)

AgentAll-960                    15.5 ± 7%                    33.4 ± 0%  +115.79%  (p=0.100 n=3+3)

AgentAll-960                   564MB ± 1%                   571MB ± 0%    +1.20%  (p=0.100 n=3+3)

AgentAll-960                   7.92M ± 1%                   8.07M ± 0%    +1.82%  (p=0.100 n=3+3)
```

### 30g APM Server

```console
$ benchstat -alpha 0.11 sizes/main-active/30g-12s-12n.txt sizes/scaling/30g-24s-24n1mb-50av-gomaxprocs-by-4-pr.txt
name           old time/op                  new time/op                  delta
AgentAll-1920                   585ms ± 5%                   183ms ± 0%   -68.71%  (p=0.100 n=3+3)

name           old error_responses/sec      new error_responses/sec      delta
AgentAll-1920                    0.04 ±18%                    0.01 ±50%   -76.61%  (p=0.100 n=3+3)

name           old events/sec               new events/sec               delta
AgentAll-1920                   30.7k ± 5%                   98.1k ± 0%  +219.84%  (p=0.100 n=3+3)

name           old gc_cycles                new gc_cycles                delta
AgentAll-1920                     335 ± 5%                     772 ± 1%  +130.55%  (p=0.100 n=3+3)

name           old max_goroutines           new max_goroutines           delta
AgentAll-1920                     390 ± 2%                    1026 ± 1%  +163.22%  (p=0.100 n=3+3)

name           old max_heap_alloc           new max_heap_alloc           delta
AgentAll-1920                   1.09G ± 2%                   1.41G ± 1%   +29.59%  (p=0.100 n=3+3)

name           old max_heap_objects         new max_heap_objects         delta
AgentAll-1920                   9.18M ± 3%                  13.15M ± 1%   +43.31%  (p=0.100 n=3+3)

name           old max_rss                  new max_rss                  delta
AgentAll-1920                   1.23G ± 2%                   1.58G ± 0%   +28.20%  (p=0.100 n=3+3)

name           old mean_available_indexers  new mean_available_indexers  delta
AgentAll-1920                    15.8 ± 3%                    12.2 ± 7%   -22.64%  (p=0.100 n=3+3)

name           old alloc/op                 new alloc/op                 delta
AgentAll-1920                   573MB ± 0%                   560MB ± 0%    -2.19%  (p=0.100 n=3+3)

name           old allocs/op                new allocs/op                delta
AgentAll-1920                   8.00M ± 0%                   7.96M ± 0%    -0.59%  (p=0.100 n=3+3)
```

marclop:
## Metric Screenshots

The last 3 distributions are all for the same 30g APM Server instance. I ran the benchmarks 3 times with that size to gather some more data.

## APM Server

Screen Shot 2022-10-18 at 15 57 11

## Elasticsearch

464GB of RAM! Cluster over 3 zones. With a total of 24 hot nodes.

Screen Shot 2022-10-18 at 16 06 53

Screen Shot 2022-10-18 at 16 07 23

Screen Shot 2022-10-18 at 16 08 24

internal/model/modelindexer/indexer.go
```diff
@@ -384,15 +442,33 @@ func (i *Indexer) flush(ctx context.Context, bulkIndexer *bulkIndexer) error {
...
+		timedFlush++
+		fullFlush = 0
+	}
+	countFullFlush := func() {
+		if i.config.Scaling.Disabled {
+			return
+		}
+		fullFlush++
+		timedFlush = 0
+	}
```
axw, line 470: It's probably faster to just increment/clear the counters inline. This code might get inlined, but even if it is there's some branching in there (i.e. the conditional on Disabled). I don't think it's harmful to update the counters even if scaling is disabled, is it?

internal/model/modelindexer/indexer.go
```diff
@@ -401,12 +477,23 @@ func (i *Indexer) runActiveIndexer() {
...
 	for !closed {
 		select {
 		case <-flushTimer.C:
+			countTimedFlush()
+		case <-idleTimer.C:
+			countTimedFlush()
 		default:
+			// When the queue utilization is below 5%, reset the idleTimer. When
+			// traffic to the APM Server is interrupted or stopped, it allows excess
+			// active indexers that have been idle for a the IdleInterval to be
```
axw, line 490:
```suggestion
			// active indexers that have been idle for the IdleInterval to be
```

internal/model/modelindexer/indexer.go
```diff
@@ -401,12 +477,23 @@ func (i *Indexer) runActiveIndexer() {
...
+			// scaled down.
+			activeIndexers := atomic.LoadInt64(&i.activeBulkRequests)
+			lowChanCapacity := float64(len(i.bulkItems))/float64(cap(i.bulkItems)) <= 0.05
+			if lowChanCapacity && activeIndexers > 1 {
+				idleTimer.Reset(i.config.Scaling.IdleInterval)
+			}
```
axw, line 496: Doing this on every iteration of the loop concerns me a little bit. Would it make sense to require `FlushInterval >= IdleInterval`, and only reset the idle timer before the first iteration of the loop, and then after flushing? i.e. whenever the flush timer is inactive.

Also, I think the idle timer should only be started when autoscaling is enabled?

internal/model/modelindexer/indexer_test.go
```diff
@@ -683,10 +685,200 @@ func TestModelIndexerCloseBusyIndexer(t *testing.T) {
...
+		})
+		indexer, err := modelindexer.New(client, cfg)
+		require.NoError(t, err)
+		t.Cleanup(func() { indexer.Close(context.Background()) })
+		return indexer
+	}
+	sendEvents := func(t *testing.T, indexer *modelindexer.Indexer, events int) {
+		for i := 0; i < events; i++ {
+			// The compressed size for this event is  is roughly ~190KB with
+			// default compression.
```
axw, line 707: 190KB for this almost empty event?

internal/model/modelindexer/indexer_test.go
```diff
@@ -683,10 +685,200 @@ func TestModelIndexerCloseBusyIndexer(t *testing.T) {
...
+				Dataset:   "apm_server",
+				Namespace: "testing",
+			}}}
+			assert.NoError(t, indexer.ProcessBatch(context.Background(), &batch))
+		}
+	}
+	waitForScaleUp := func(t *testing.T, indexer *modelindexer.Indexer, n int64) {
+		if n == 0 {
+			n = 2
+		}
```
axw, line 719: What's this about? We appear to ever pass in 0, so is it needed?

internal/model/modelindexer/indexer_test.go
```diff
@@ -683,10 +685,200 @@ func TestModelIndexerCloseBusyIndexer(t *testing.T) {
...
+			}
+		}
+		stats = indexer.Stats()
+		assert.Greater(t, stats.UpScales, int64(0), "No upscales took place: %+v", stats)
+		assert.Less(t, stats.ActiveBulkRequests, int64(runtime.GOMAXPROCS(0)))
+	}
+	waitForScaleDown := func(t *testing.T, indexer *modelindexer.Indexer, n int64) {
+		if n == 0 {
+			n = 1
+		}
```
axw, line 738: As above

internal/model/modelindexer/indexer.go
```diff
@@ -101,25 +112,52 @@ type Config struct {
...
+	UpScale ScaleActionConfig
+
+	// IdleInterval defines how long an active indexer performs an inactivity
+	// check based on the DownScale settings.
+	IdleInterval time.Duration
+}
+
+// ScaleActionConfig holds the configuration for a scaling action
+type ScaleActionConfig struct {
+	// Threshold is the number on which the scaling action will be triggered.
```
axw, line 155: The number of what? Could you add some comments to the ScalingConfig.DownScale and ScalingConfig.UpScale fields to describe what the thresholds are in terms of? Alternatively, maybe we should fold these fields directly into ScalingConfig and come up with less abstract names. I couldn't think of any off the top of my head though.

axw (commented):
I have a bunch of comments and suggestions, but this is very nice :)

internal/model/modelindexer/indexer.go
```diff
@@ -444,7 +538,111 @@ func (i *Indexer) runActiveIndexer() {
...
+		return scaleDown()
+	}
+	if *timedFlush < i.config.Scaling.DownScale.Threshold {
+		return false
+	}
+	// Reset timedFlush after it has exceeded the threshold
+	// it avoids unnecessary precociousness to scale down.
+	*timedFlush = 0
+	lastScaleT := time.Unix(atomic.LoadInt64(&i.lastScaleAction), 0)
+	if !lastScaleT.Add(i.config.Scaling.DownScale.CoolDown).Before(now) {
```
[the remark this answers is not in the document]
marclop: I wasn't aware of that. I think it's better to store the `time.Time` as an `atomic.Value` then.

internal/model/modelindexer/indexer.go
```diff
@@ -444,7 +538,111 @@ func (i *Indexer) runActiveIndexer() {
...
+	// cooldown, since doing so would result in using much more CPU for longer.
+	if activeIndexers > activeLimit() {
+		i.logger.Infof("active indexers > active limit, scaling down: %d",
+			atomic.LoadInt64(&i.activeBulkRequests),
+		)
+		return scaleDown()
+	}
+	if *timedFlush < i.config.Scaling.DownScale.Threshold {
+		return false
+	}
```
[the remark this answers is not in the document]
marclop: we could, but then it wouldn't scale down indexers when the CFS quota has been updated, which would trigger excessive throttling in the cgroup.

internal/model/modelindexer/indexer_test.go
marclop: I'll drop the `K`. I meant B.

internal/model/modelindexer/indexer.go
```diff
@@ -401,12 +477,23 @@ func (i *Indexer) runActiveIndexer() {
...
+			countTimedFlush()
+		case <-idleTimer.C:
+			countTimedFlush()
 		default:
+			// When the queue utilization is below 5%, reset the idleTimer. When
+			// traffic to the APM Server is interrupted or stopped, it allows excess
+			// active indexers that have been idle for a the IdleInterval to be
+			// scaled down.
+			activeIndexers := atomic.LoadInt64(&i.activeBulkRequests)
+			lowChanCapacity := float64(len(i.bulkItems))/float64(cap(i.bulkItems)) <= 0.05
```
simitt, line 493: Wouldn't that rather be `lowChanUsage`?

simitt (commented):
Excited about these changes! 
Left a couple of mostly nitpicks around naming and comments. 

Have you tested this against an undersized ES? Given that the number of active indexers is based on events processed in the APM Server and not also on ES response and pushback, I wonder if a situation might get worse for setups where APM Server is processing more events than ES can handle, by increasing the pressure towards ES.

dev_docs/ARCHITECTURE.md
```diff
@@ -1,9 +1,86 @@
...
+High level overview over incoming data and their flow through the APM Server until being passed on
+to the output publisher pipeline.
+
+![ingest-flow](../docs/images/ingest-flow.png)
+
+## Modelindexer Architecture
+
+When APM Server runs under managed mode or using the Elasticsearch output in standalone mode, it uses
+a custom Elasticsearch output called `modelindexer`.  This custom output differs from the previous
+libbeat output in multiple ways, but its main difference is that it fills a local cache until it is full,
```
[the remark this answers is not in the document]
marclop: Isn't it still useful for us devs to come back to this since `7.17` is LTS? I expect maybe some team members may want to look into where we came from and how that looked compared to what we have

internal/model/modelindexer/indexer.go
```diff
@@ -444,7 +539,109 @@ func (i *Indexer) runActiveIndexer() {
 				return err
 			})
 		}
+		if i.config.Scaling.Disabled {
+			continue
+		}
+		now := time.Now().Add(time.Duration(id))
```
axw, line 545: Why is `id` being added to the time?

internal/model/modelindexer/indexer.go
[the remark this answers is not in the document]
axw: I see, thanks - I missed that there are two paths that lead to scaling down.

dev_docs/ARCHITECTURE.md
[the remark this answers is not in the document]
simitt: My main concern is that we do not have any libbeat related dev-docs that would explain how libbeat output was working, so I didn't consider this comparison very helpful. I don't feel strongly about this, if you believe it adds value for new team members, then please keep it.

internal/model/modelindexer/indexer.go
marclop: Yep, you're right. I've implemented what you suggest in the latest commit.

internal/beater/beater.go
```diff
@@ -622,6 +630,10 @@ func (s *Runner) newFinalBatchProcessor(
 		v.OnInt(stats.ActiveBulkRequests)
 		v.OnKey("completed")
 		v.OnInt(stats.BulkRequests)
+		v.OnKey("active_created")
+		v.OnInt(stats.IndexersDestroyed)
+		v.OnKey("active_destroyed")
+		v.OnInt(stats.IndexersCreated)
```
axw, line 636: These names still read a bit awkwardly to me. Maybe rename to `output.elasticsearch.indexers.{created,destroyed}`?

internal/model/modelindexer/indexer.go
axw: Thanks, looks good.

axw (commented):
Thanks for the updates, it's looking pretty good now. Only one more comment on the metric names from me - I think the main remaining question is @simitt's one about behaviour when ES is underpowered.

marclop:
@simitt @axw Thanks for the reviews. I have tested the performance of an 8GB APM Server (4vCPUs with 8 as the burstable max) and it does seem like the pressure is a bit worse since the autoscaling does take place after 60 consecutive flushes. The backing Elasticsearch cluster is a 3 zone 8GB cluster.

Comparing the current PR with main (main has 25 available indexers of 2.5MB of size):

```console
$ benchstat -alpha 0.11 sizes/scaling/undersized/8g-2s-2n-main.txt sizes/scaling/undersized/8g-2s-2n-pr.txt
name          old time/op                  new time/op                  delta
AgentAll-512                   1.91s ± 4%                   2.01s ± 3%      ~     (p=0.200 n=3+3)

name          old error_responses/sec      new error_responses/sec      delta
AgentAll-512                    0.00                         0.00           ~     (all equal)

AgentAll-512                   9.46k ± 4%                   8.98k ± 3%      ~     (p=0.200 n=3+3)

name          old gc_cycles                new gc_cycles                delta
AgentAll-512                     130 ± 8%                     120 ± 2%    -7.93%  (p=0.100 n=3+3)

AgentAll-512                     232 ±10%                     294 ± 4%   +26.69%  (p=0.100 n=3+3)

AgentAll-512                    989M ± 3%                   1015M ± 2%      ~     (p=0.400 n=3+3)

AgentAll-512                   8.82M ± 5%                   9.00M ± 5%      ~     (p=0.700 n=3+3)

AgentAll-512                   1.10G ± 2%                   1.13G ± 1%    +2.65%  (p=0.100 n=3+3)

AgentAll-512                    2.65 ±11%                    0.00       -100.00%  (p=0.100 n=3+3)

AgentAll-512                   638MB ± 0%                   644MB ± 1%      ~     (p=0.200 n=3+3)

AgentAll-512                   8.59M ± 0%                   8.63M ± 0%      ~     (p=0.200 n=3+3)
```

The number of 429 was significant (up to 23% of all requests).

I made some small changes (not pushed to this PR) which disallowed scaling when 1% or more of the total indexed documents results in a 429, that seemed to give a very similar performance to main:

```console
$ benchstat -alpha 0.11 sizes/scaling/undersized/8g-2s-2n-main.txt sizes/scaling/undersized/8g-2s-2npr-block-as.txt
name          old time/op                  new time/op                  delta
AgentAll-512                   1.91s ± 4%                   1.92s ± 1%      ~     (p=1.000 n=3+3)

name          old error_responses/sec      new error_responses/sec      delta
AgentAll-512                    0.00                         0.00           ~     (all equal)

AgentAll-512                   9.46k ± 4%                   9.39k ± 1%      ~     (p=1.000 n=3+3)

name          old gc_cycles                new gc_cycles                delta
AgentAll-512                     130 ± 8%                     123 ± 5%      ~     (p=0.500 n=3+3)

AgentAll-512                     232 ±10%                     304 ± 6%   +30.85%  (p=0.100 n=3+3)

AgentAll-512                    989M ± 3%                    974M ± 0%      ~     (p=0.700 n=3+3)

AgentAll-512                   8.82M ± 5%                   8.21M ± 3%    -6.96%  (p=0.100 n=3+3)

AgentAll-512                   1.10G ± 2%                   1.11G ± 0%      ~     (p=0.700 n=3+3)

AgentAll-512                    2.65 ±11%                    0.00       -100.00%  (p=0.100 n=3+3)

AgentAll-512                   638MB ± 0%                   642MB ± 0%    +0.60%  (p=0.100 n=3+3)

AgentAll-512                   8.59M ± 0%                   8.62M ± 0%      ~     (p=0.200 n=3+3)
```

patch: 

```diff
diff --git a/internal/model/modelindexer/indexer.go b/internal/model/modelindexer/indexer.go
index de50be0a1..e50844c68 100644
--- a/internal/model/modelindexer/indexer.go
+++ b/internal/model/modelindexer/indexer.go
@@ -594,6 +594,14 @@ func (i *Indexer) maybeScaleDown(now time.Time, info scalingInfo, timedFlush *ui
 		}
 		info = i.scalingInformation() // refresh scaling info if CAS failed.
 	}
+	// If more than 1% of the requests result in 429, scale down.
+	if i.indexFailureRate() >= 0.01 {
+		if new := info.ScaleDown(now); i.scalingInfo.CompareAndSwap(info, new) {
+			i.logger.Infof("Elasticsearch 429 too many rate exceeded 1%, scaling down to: %d", new)
+			return true
+		}
+		return false
+	}
 	if *timedFlush < i.config.Scaling.ScaleDown.Threshold {
 		return false
 	}
@@ -624,6 +632,10 @@ func (i *Indexer) maybeScaleUp(now time.Time, info scalingInfo, fullFlush *uint)
 	// Reset fullFlush after it has exceeded the threshold
 	// it avoids unnecessary precociousness to scale up.
 	*fullFlush = 0
+	// If more than 1% of the requests result in 429, do not scale up.
+	if i.indexFailureRate() >= 0.01 {
+		return false
+	}
 	if info.withinCoolDown(i.config.Scaling.ScaleUp.CoolDown, now) {
 		return false
 	}
@@ -642,6 +654,11 @@ func (i *Indexer) scalingInformation() scalingInfo {
 	return i.scalingInfo.Load().(scalingInfo)
 }
 
+func (i *Indexer) indexFailureRate() float64 {
+	return float64(atomic.LoadInt64(&i.tooManyRequests)) /
+		float64(atomic.LoadInt64(&i.eventsAdded))
+}
+
 // activeLimit returns the value of GOMAXPROCS / 4. Which should limit the
 // maximum number of active indexers to 25% of GOMAXPROCS.
 // NOTE: There is also a sweet spot between Config.MaxRequests and the number

```

If we're ok merging this PR and doing some more investigation on backstop of pressure identifiers as a follow up then we can properly test all these changes and perhaps think about a different way to look for overwhelmed Elasticsearch.

dev_docs/ARCHITECTURE.md
```diff
@@ -1,9 +1,84 @@
...
-![](./docs/images/ingest-flow.png)
\ No newline at end of file
+High level overview over incoming data and their flow through the APM Server until being passed on
+to the output publisher pipeline.
+
+![ingest-flow](../docs/images/ingest-flow.png)
+
+## Modelindexer Architecture
+
+When APM Server uses a custom Elasticsearch output called `modelindexer`. It fills a local cache until it
```
simitt, line 17:
```suggestion
When APM Server uses a custom Elasticsearch output called `modelindexer`, it fills a local cache until it
```

simitt (approved):
Apart from a minor [comment](https://github.com/elastic/apm-server/pull/9393/files#r999340608) that hasn't been addressed yet, this LGTM. 
Let's wait for @axw 's approval also, and then merge this in and address underscaled ES behavior in a follow up.

internal/beater/beater.go
```diff
@@ -612,15 +620,19 @@ func (s *Runner) newFinalBatchProcessor(
 		v.OnInt(indexer.Stats().Added)
 	})
 	monitoring.Default.Remove("output")
-	monitoring.NewFunc(monitoring.Default, "output.elasticsearch.bulk_requests", func(_ monitoring.Mode, v monitoring.Visitor) {
+	monitoring.NewFunc(monitoring.Default, "output.elasticsearch.indexers", func(_ monitoring.Mode, v monitoring.Visitor) {
```
axw, line 623: Not quite what I had in mind. This will produce:

- `output.elasticsearch.indexers.active` (currently active number of indexers)
- `output.elasticsearch.indexers.available` (sounds like available indexers, is actually available bulk requests)
- `output.elasticsearch.indexers.created` (counter of indexers that have been created)
- `output.elasticsearch.indexers.destroyed` (counter of indexers that have been destroyed)
- `output.elasticsearch.indexers.requests_completed` (counter of bulk requests that have been completed)

So I see a couple of issues:
 - the existing "active bulk requests" metric (`output.elasticsearch.bulk_requests.active`) has been removed but probably shouldn't have been
 - `output.elasticsearch.indexers.available` doesn't describe what it really is

Active bulk requests counts the `*bulkIndexers` that have been acquired from, but not yet released back to, the "available" channel.

I suggest reinstating all the existing metrics, and just introducing new `output.elasticsearch.indexers.*` metrics. Here's what I would expect to see at the end:

- `output.elasticsearch.indexers.active` (currently active number of indexers)
- `output.elasticsearch.indexers.created` (counter of indexers that have been created)
- `output.elasticsearch.indexers.destroyed` (counter of indexers that have been destroyed)
- `output.elasticsearch.bulk_requests.active` (currently active number of bulk requests)
- `output.elasticsearch.bulk_requests.available` (currently available number of bulk requests)
- `output.elasticsearch.bulk_requests.completed` (counter of bulk requests that have been completed)

axw (commented):
> If we're ok merging this PR and doing some more investigation on backstop of pressure identifiers as a follow up then we can properly test all these changes and perhaps think about a different way to look for overwhelmed Elasticsearch.

SGTM.

The changes all look good apart from a couple of issues with the metrics.

internal/beater/beater.go
marclop: Sounds much better, I may not have thought the metric name change through.

internal/model/modelindexer/indexer.go
```diff
@@ -203,7 +282,9 @@ func (i *Indexer) Stats() Stats {
 		TooManyRequests:       atomic.LoadInt64(&i.tooManyRequests),
 		BytesTotal:            atomic.LoadInt64(&i.bytesTotal),
 		AvailableBulkRequests: atomic.LoadInt64(&i.availableBulkRequests),
-		ActiveBulkRequests:    atomic.LoadInt64(&i.activeBulkRequests),
+		ActiveBulkRequests:    i.scalingInformation().activeIndexers,
```
axw, line 285: We should revert to how we were tracking the `ActiveBulkRequests` metric before, and maybe expose `activeIndexers` as a new metric.

axw (changes requested):
One last thing, I think.

internal/model/modelindexer/indexer.go
```diff
@@ -101,25 +112,52 @@ type Config struct {
...
+	// If unset, scaling is enabled by default.
+	Scaling ScalingConfig
+}
+
+// ScalingConfig holds the modelindexer scaling configuration.
+type ScalingConfig struct {
+	// Disabled toggles active indexer scaling on.
+	//
+	// It is enabled by default.
+	Disabled bool
```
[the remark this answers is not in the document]
axw: FWIW I think this is okay for the moment, maybe we can come back to it if it's a pain. The zero value for ScalingConfig currently means to use the default config; it should be enabled by default. So we would need to make it a `*bool` to preserve that while also making it "enabled".

axw (approved):
Thank you!

axw:
This should be tested as part of https://github.com/elastic/apm-server/issues/9182

PR #9395, opened 2022-10-19
Simplify Kibana client

axw:
## Motivation/summary

Remove the asynchronous connection loop from the Kibana client, and remove version checking. We only support Kibana >= APM Server, and all recent versions of APM Server support the central config API (the only place we care about the version.)

Also, remove allowance for (and unit testing of) an impossible code path related to having a nil *kibana.Client from the Kibana-based agent config fetcher. We will now panic if a nil client is passed into NewKibanaFetcher, enforcing that a non-nil client is required in the contract.

Extraction of VCS info is moved to package version, so it can be used for defining the Kibana client's User-Agent.

## Checklist

## How to test these changes

N/A. Non-functional change.

## Related issues

None


internal/beater/api/config/agent/handler_test.go
```diff
@@ -233,9 +208,9 @@ func TestAgentConfigHandlerAnonymousAccess(t *testing.T) {
 }
 
 func TestAgentConfigHandlerAuthorizedForService(t *testing.T) {
-	cfg := config.KibanaAgentConfig{Cache: config.Cache{Expiration: time.Nanosecond}}
-	f := agentcfg.NewKibanaFetcher(nil, cfg.Cache.Expiration)
-	h := NewHandler(f, cfg, "", nil)
+	// TODO
```
marclop, line 211: Is there anything left todo in this test?

internal/kibana/client_test.go
```diff
@@ -0,0 +1,128 @@
...
+	assert.Equal(t, []string{"abc", "123"}, headers.Values("Combined"))
+}
+
+func TestClientSend(t *testing.T) {
+	var requestURL *url.URL
+	var requestBody []byte
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		requestURL = r.URL
+		requestBody, _ = io.ReadAll(r.Body)
+		w.WriteHeader(http.StatusTeapot)
```
marclop, line 103: TIL 😂

marclop (approved):
Thanks for the simplification 🙂 the code looks great!

internal/beater/api/config/agent/handler_test.go
axw: Oops, thanks. I intended to replace the `&kibana.Client{}` with an actual client. Fixed.

PR #9411, opened 2022-10-20
Fix time parsing format for timezone

lahsivjar:
## Motivation/summary

Fixes timezone format for parsing timestamp field.

## Checklist

## How to test these changes

See #9100 

## Related issues

Fixes issue with #9100


axw (approved):

axw (approved):

lahsivjar:
@felixbarny @SylvainJuge I am confused between `2006-01-02T15:04:05Z0700` and `2006-01-02T15:04:05Z07:00` formats. I discussed a bit with Sylvain earlier and he was suggesting `2006-01-02T15:04:05Z07:00` however, since then, I have found that [Go ECS libraries](https://github.com/elastic/ecs-logging-go-logrus/blob/main/formatter.go#L133) seem to use `2006-01-02T15:04:05Z0700` format.

felixbarny:
Which of these timestamps can be parsed by Elasticsearch's default date format?

lahsivjar:
@felixbarny I just tested out with an ES deployment which seems to support full ISO8601 i.e. `hh:mm`, `hhmm` and `hh`. Let me update the PR to support all three formats.

internal/model/modeldecoder/nullable/nullable.go
```diff
@@ -27,6 +27,12 @@ import (
 	jsoniter "github.com/json-iterator/go"
 )
 
+var supportedTSFormats = []string{
+	time.RFC3339,
```
axw, line 31:
```suggestion
	"2006-01-02T15:04:05Z07:00", // time.RFC3339,
```

Might be worth writing out the layout here to highlight the difference to the ones below?

internal/model/modeldecoder/nullable/nullable.go
```diff
@@ -27,6 +27,12 @@ import (
 	jsoniter "github.com/json-iterator/go"
 )
 
+var supportedTSFormats = []string{
```
axw, line 30: Could you please add a comment explaining that the below are variations of time.RFC3339, supporting more formats for the timezone offset?

axw (approved):
LGTM. _Maybe_ in the future we should split on "Z" and parse the timezone separately, but it's probably fine to just do this.

lahsivjar:
> LGTM. Maybe in the future we should split on "Z" and parse the timezone separately, but it's probably fine to just do this.

Yeah, we can deduce the format by checking the length of the string after `Z`. I chose to do it like this because I thought the readability is better this way and not that many performance gains.

SylvainJuge:
I just did a test locally, and I can confirm that now the Java APM agent can send logs without errors :tada: .

Issue #3874, opened 2020-06-11
[ECS] align with ECS observer fieldset


axw:
APM Server and (potentially other beats acting as an observer) sends several `observer` fields that are not part of the ECS fieldset definition:
 - `observer.ephemeral_id`
 - `observer.id`

We should propose a change to the ECS observer fieldset, to include these fields.

~In the case of `observer.version_major`, we would probably want to wait for a resolution on https://github.com/elastic/elasticsearch/issues/48878; rather than continuing to store the major version, we would instead store a semver tuple. See also https://github.com/elastic/ecs/issues/842.~ version_major was removed in https://github.com/elastic/apm-server/pull/7399

PR #9412, opened 2022-10-20
Drop `observer.id` and `observer.ephemeral_id`

axw:
## Motivation/summary

We don't use these fields in the UI anywhere, and for debugging we can rely on the hostname to distinguish one APM Server instance from another. The `observer.id` and `observer.ephemeral_id` fields suggest that we should care about specific instances (and think of them as long-lived), and their continued presence contradicts our intentions for a more modern, cloud native, approach, where APM Server instances may come and go.

For now we need to continue persisting a UUID to disk for stack monitoring, as this is used to identify instances of APM Server. In the future we expect to revisit how stack monitoring works, and at that point we should redesign the documents to not enshrine a "pets" mentality. We may still want to identify individual instances (e.g. for calculating time-series IDs), but statefulness should not be required.

For tail-based sampling we now generate a UUID (for identifying its documents) at process start. The field has been renamed from `observer.id` to `agent.ephemeral_id`, the latter of which is an ECS field (`observer.id` is not).

We no longer pass `beat.Info` from `beatcmd` into package `beater`. Instead, `beater` hard-codes `observer.type: apm-server`, uses package `version` for `observer.version`, and calls `os.Hostname()` for `observer.hostname`. We need to recreate `beat.Info` for the libbeat outputs. If at some point we drop support for libbeat outputs, then we would have no more dependencies on `beat.Info` outside of package `beatcmd`.

## Checklist

## How to test these changes

1. Ensure tail-based sampling still works (with multiple servers)
2. Ensure stack monitoring still works (with multiple servers)

## Related issues

Closes https://github.com/elastic/apm-server/issues/3874


apmpackage/apm/data_stream/sampled_traces/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,7 @@
+---
+description: Pipeline for ingesting APM tail-sampled trace IDs.
+processors:
+  - rename:
+      field: observer.id
+      target_field: agent.ephemeral_id
```
simitt, line 6: ECS says:
>This id normally changes across restarts, but `agent.id` does not.

I believe semantically `observer.id` -> `agent.id` would be more correct.

axw: The ID changes on restart now - it's generated at process init.

simitt (approved):

PR #9421, opened 2022-10-24
Introduce Kustomize and Tilt

axw:
## Motivation/summary

Introduce Kustomize manifests in `testing/infra/k8s`, using ECK for running Elasticsearch, Kibana, Fleet Server, and an Elastic Agent with the APM integration injected.

We use the Kustomize manifests from Tiltfile, injecting a live-reloaded custom image into the apm-server Agent pod spec.

## Checklist

## How to test these changes

1. Install Tilt: https://docs.tilt.dev/install.html. On macOS you can use Docker Desktop, on Linux you should use Kind (optionally using https://github.com/tilt-dev/ctlptl for managing the cluster.)
2. Run `tilt up`
3. Send some APM data, check that it shows up in Kibana

## Related issues

None


simitt (approved):
Works great 🚀 . The one thing missing is to add a small documentation section. It might be worth adding something directly to the README. 

Minor note: maybe also mention `tilt down` for how to stopp the deployment again.

axw:
@simitt I've just updated `dev_docs/TESTING.md` for now. I think it might be worth updating README.md too, but would rather take my time on that and don't want to delay merging this.

Tiltfile
```diff
@@ -0,0 +1,47 @@
...
+#
+# Without this, ECK refuses to recreate the trial license. That's expected
+# for actual trial use, but we rely an Enterprise feature for local development:
+# installing the integration package by uploading to Fleet.
+if config.tilt_subcommand == "down":
+  print(local("kubectl delete --ignore-not-found namespace/elastic-system"))
+
+# Add a button for sending trace events and metrics to APM Server.
+load('ext://uibutton', 'cmd_button')
+cmd_button(
```
marclop, line 41: This is pretty cool

testing/infra/k8s/base/kustomization.yaml
```diff
@@ -0,0 +1,7 @@
+resources:
+  - https://download.elastic.co/downloads/eck/2.4.0/crds.yaml
+  - https://download.elastic.co/downloads/eck/2.4.0/operator.yaml
```
marclop, line 3: I wonder how we'll keep these up to date.

marclop (commented):
excited to try this out!

testing/infra/k8s/base/kustomization.yaml
axw: Like end users, I don't think it's too much of a problem if we're not always on the latest version. Only if there's some security issue or critical bug, or a new feature that we want to take advantage of.

PR #9429, opened 2022-10-25
Map RUM PerformanceResourceTiming fields

axw:
## Motivation/summary

Add field mappings for:

 - http.response.transfer_size
 - http.response.encoded_body_size
 - http.response.decoded_body_size

The fields are mapped with `index: false`, to minimise storage overhead. They can still be used in searches, just a bit slower.

The intake schema for these fields has been changed from float64 to int, as these are originally integer values.
See https://w3c.github.io/resource-timing/#sec-performanceresourcetiming

testdata, and hence approvals, have been updated to change the mock values from floats to integers.

## Checklist

## How to test these changes

1. Send some RUM page-load transactions and HTTP spans to APM Server
2. Ensure the fields can be searched and aggregated

## Related issues

None


vigneshshanmugam (approved):
LGTM from the RUM agent side.

simitt (approved):

SylvainJuge:
Regarding applying this change to other non-RUM agents, do we have (or plan to) do anything beyond updating the schema for now ? For example do we have to also attempt to capture similar attributes on HTTP spans ?

axw:
@SylvainJuge no, there's no plans for that at the moment

carsonip:
Confirmed that we can search and aggregate on `http.response.transfer_size`, `http.response.encoded_body_size` and `http.response.decoded_body_size` and they are all integers. One thing I noticed is that decoded_body_size does not have `unit: byte`. Not sure if this is intended / affects any functionality.

In mapping:
```
          "response": {
            "properties": {
              "decoded_body_size": {
                "type": "long",
                "index": false
              },
              "encoded_body_size": {
                "type": "long",
                "index": false,
                "meta": {
                  "unit": "byte"
                }
              },
              "finished": {
                "type": "boolean"
              },
              "headers": {
                "type": "object"
              },
              "status_code": {
                "type": "long"
              },
              "transfer_size": {
                "type": "long",
                "index": false,
                "meta": {
                  "unit": "byte"
                }
              }
            }
          }
```
In code:
https://github.com/elastic/apm-server/pull/9429/files#diff-3d9ce867f164a3002e6829a4b35a8d9fafd3c3cd84f137c98fb1cedc97bcbe25R74

axw:
Good catch @carsonip. AFAIK it's not currently used, so it doesn't matter at the moment.

Issue #9184, opened 2022-09-23
Set up automated smoke tests for supported Operating Systems (linux, windows)


simitt:
We want to run automated smoke tests for officially supported operation systems. Currently this effort is limited to Linux and Windows Server. The officially supported distributions and versions can be found at https://www.elastic.co/support/matrix#matrix_compatibility. 

The required steps are:
 - [x] Implement minimum smoke tests (start apm-server, ingesting sample events and querying them from ES)
 - [x] Use existing terraform automation for spinning up an ES and Kibana in the ESS CFT region on AWS.
 - [x] Build automation with terraform to provision the according linux ~or windows server~ on EC2 AWS in the same region as the ESS stack deployment (ES+Kibana). 
 - [x] Install APM Server (once in standalone mode, ~once with Elastic Agent + APM integration~) on EC2 
 - [x] Send test events to APM Server
 - [x] Query expected events from ES and run assertions
 - [x] Tear everything down. 

The work should leverage previous work around setup automation with terraform and existing smoke tests in the apm-server where possible. 

After the smoke tests are defined and the setup is automated, the productivity engineers will contribute Jenkins automation and slack notification integration in case of failures.

PR #9440, opened 2022-10-26
feat: set up automated smoke tests for supported Operating Systems

kruskall:
## Motivation/summary

Add smoke testing for the following os (from the support matrix):

- ubuntu 18.04
- ubuntu 20.04
- ubuntu 22.04
- debian 10
- debian 11
- Amazon Linux 2
- RHEL 7
- RHEL 8

The terraform script provisions an ec2 instance and downloads the
apm server from the artifacts api.
The apm-server is installed and linked to the es cluster.

## Checklist

## Related issues

Related to https://github.com/elastic/apm-server/issues/9184


testing/smoke/supported-os/test.sh
```diff
@@ -0,0 +1,65 @@
...
+if [[ -z ${VERSION} ]] || [[ "${VERSION}" == "latest" ]]; then
+    VERSION=$(echo ${VERSIONS} | jq -r 'last')
+    echo "-> unspecified version, using $(echo ${VERSION} | cut -d '.' -f1-2)"
+fi
+MAJOR_VERSION=$(echo ${VERSION} | cut -d '.' -f1 )
+MINOR_VERSION=$(echo ${VERSION} | cut -d '.' -f2 )
+
+OBSERVER_VERSION=$(echo ${VERSION} | cut -d '-' -f1 )
+
+if [[ ${MAJOR_VERSION} -eq 7 ]]; then
```
axw, line 25: Are we testing old versions of APM Server or the stack? Is this code path relevant?

testing/smoke/supported-os/test.sh
```diff
@@ -0,0 +1,65 @@
...
+    "RHEL-8"
+)
+
+for os in "${os_names[@]}"
+do
+    terraform_apply ${LATEST_VERSION} ${INTEGRATIONS_SERVER} ${KEY_NAME} $os
+    healthcheck 1
+    send_events
+    ${ASSERT_EVENTS_FUNC} ${OBSERVER_VERSION}
+    terraform_destroy
```
axw, line 64: Applying and destroying in a loop means we'll create the whole ESS deployment from fresh each time, right? That sounds like it'll be a bit slow.

Would it work to apply each time, delete data streams at the end of each OS's tests, and destroy right at the end? Subsequent applies should update changed and delete unused resources, right?

axw (commented):
Looks nice overall. I think mixing the ESS and OS smoke test Terraform together might be a recipe for complexity, though.

None of the AWS instance stuff is relevant to the ESS smoke tests. Some of the ESS things (secret token, server URL, default stack version) are not relevant to the OS smoke tests. So, would we be better of creating a dedicated Terraform module for the OS smoke tests? That way we can avoid conditionals, and minimise the amount of context required to understand them.

If we had a Terraform module just for the OS smoke tests we would create an ESS deployment with Elasticsearch, Kibana, and Integrations Server for its Fleet Server component. The ESS deployment would always be 8.x deployment, as we won't be doing automated OS smoke tests for 7.x.

testing/smoke/supported-os/test.sh
kruskall: I don't think we are. Good catch! :+1:

testing/smoke/supported-os/test.sh
kruskall: Good point! I've update the code to reuse the cluster :+1:

kruskall:
> Looks nice overall. I think mixing the ESS and OS smoke test Terraform together might be a recipe for complexity, though.
> 
> None of the AWS instance stuff is relevant to the ESS smoke tests. Some of the ESS things (secret token, server URL, default stack version) are not relevant to the OS smoke tests. So, would we be better of creating a dedicated Terraform module for the OS smoke tests? That way we can avoid conditionals, and minimise the amount of context required to understand them.

Does it need a module ? All the unrelated stuff are in the `testing/smoke/main.tf` script. I think we can achieve the same result by using a separate tf script without the conditionals.

axw:
> Does it need a module ? All the unrelated stuff are in the testing/smoke/main.tf script. I think we can achieve the same result by using a separate tf script without the conditionals.

@kruskall I _think_ we're saying the same thing. What I meant was to create a separate `.tf` file, and not reuse `testing/smoke/main.tf` - since that one is more appropriate for testing APM Server running in ESS.

kruskall:
> @kruskall I _think_ we're saying the same thing. What I meant was to create a separate `.tf` file, and not reuse `testing/smoke/main.tf` - since that one is more appropriate for testing APM Server running in ESS.

Yep, that's right, I thought you wanted to create a new `module`, similar to `ec_deployment`.
I've updated to code to use a separate file, removed the `count` field with the indexes and the conditionals.

testing/smoke/supported-os/test.sh
```diff
@@ -0,0 +1,62 @@
...
+
+for os in "${os_names[@]}"
+do
+    append_tfvar "aws_provisioner_key_name" ${KEY_NAME} 1
+    append_tfvar "aws_os" $os
+    terraform_apply
+    healthcheck 1
+    send_events
+    ${ASSERT_EVENTS_FUNC} ${OBSERVER_VERSION}
+    delete_all
```
axw, line 61: Should we move this `delete_all` directly after the `terraform_apply`?

IIANM, it is possbile that APM Server is still running (and indexing events/metrics/etc.) here, so we may end up with pollution between test cases. If we do it directly after `terraform_apply`, the previous test case's APM Server should have been stopped.

(Probably warrants a `# comment` if you do make that change.)

testing/smoke/supported-os/main.tf
```diff
@@ -0,0 +1,218 @@
...
+    host        = self.public_ip
+    private_key = file("${var.aws_provisioner_key_name}")
+  }
+
+  provisioner "remote-exec" {
+    inline = [
+      "curl ${data.external.getlatestapmserver.result.deb} -o apm-server.deb && curl ${data.external.getlatestapmserver.result.rpm} -o apm-server.rpm",
+      "sudo dpkg -i apm-server.deb || sudo yum -y install apm-server.rpm",
+      "sudo rm /etc/apm-server/apm-server.yml",
+      "echo \"apm-server:\n  host: \\\"0.0.0.0:${local.apm_port}\\\"\noutput:\n  elasticsearch:\n    hosts: [\\\"${module.ec_deployment.elasticsearch_url}\\\"]\n    username: \\\"${module.ec_deployment.elasticsearch_username}\\\"\n    password: \\\"${module.ec_deployment.elasticsearch_password}\\\"\" | sudo tee -a /etc/apm-server/apm-server.yml",
```
axw, line 135: Should we be injecting the secret token here too?

testing/smoke/lib.sh
```diff
@@ -49,11 +72,12 @@ terraform_destroy() {
     if [[ ${exit_code} -gt 0 ]]; then
         echo "-> Smoke tests FAILED!!"
         echo "-> Printing terraform logs:"
-        cat tf.log
+	#cat tf.log
```
axw, line 75: Did you mean to leave this commented out?

axw (approved):
Thanks, this looks much better now. Just a couple of more small issues, then I think this is good to go.

We'll still need to figure out Windows and macOS, but this is obviously much better than what we had ;)  Ideally we would also test aarch64.

testing/smoke/supported-os/test.sh
kruskall: Oh, good catch! Updated :+1:

testing/smoke/lib.sh
kruskall: Yes, I find the previous behaviour really confusing.

Any error during the test assertions is getting suppressed by the tf.log printed after the test failed, so you end up having to scroll to find the error message.

I restored the previous behaviour and we can discuss this in a followup since it's unrelated to the changes.

testing/smoke/supported-os/main.tf
kruskall: We don't need to. The APM server is not using the secret token afaik but the output variable is needed because the bash script sends the token in a http header.

axw: OK. I don't think it hurts either way, we can leave it for now.

testing/smoke/supported-os/test.sh
```diff
@@ -0,0 +1,66 @@
...
+
+for os in "${os_names[@]}"
+do
+    append_tfvar "aws_provisioner_key_name" ${KEY_NAME} 1
+    append_tfvar "aws_os" $os
+    terraform_apply
+    #  it is possbile that APM Server is still running (and indexing events/metrics/etc.),
+    #  so we may end up with pollution between test cases. 
+    #  If we do it directly after terraform_apply, the previous test case's APM Server 
+    #  should have been stopped.
```
axw, line 61:
```suggestion
    # The previous test case's APM Server should have been stopped by now,
    # so there should be no new documents indexed. Delete all existing data.
```

axw (approved):

simitt:
@kruskall could you fix the linting issues so we can get this merged in?

kruskall:
> @kruskall could you fix the linting issues so we can get this merged in?

The linting issue is actually the docker image not being updated: `docker-compose.yml: image docker.elastic.co/elasticsearch/elasticsearch:8.6.0-c9d1a5e9-SNAPSHOT is out of date (latest 'main' snapshot version is 8.7.0-SNAPSHOT)`

It should be fixed by https://github.com/elastic/apm-server/pull/9601

Issue #9340, opened 2022-10-11
Record sample rate in trace event documents


axw:
It may be useful to know the sampling rate that was configured when an event was captured. For example, the APM storage explorer could use this to compare configured vs. effective sampling rate, and highlight changes to the sampling rate configuration in the timeline.

We already propagate this to APM Server, and use it for calculating transaction and service-destination metrics; we just don't record it in the event documents.

sqren:
+1 for this. As mentioned in the issue description, this is useful in features like Storage Explorer. For the recent work on Mobile this is even more important, since features like Maps is purely based on sampled transactions. Without the sampling rate we won't be able to calculate metrics like throughput for a given region (country/city).

axw:
It occurs to me now that this won't be enough for mobile, we'll also need them to send the sample rate to APM Server. Currently that's only done for agents that use intake-v2. I think that should be doable though.

PR #9458, opened 2022-10-28
Add {span,transaction}.representative_count fields

axw:
## Motivation/summary

If RepresentativeCount is set (>0), store it as `{span,transaction}.representative_count` in the span/transaction documents. This may be used in queries for calculating the effective sample rate over a time range, or for calculating weighted averages from trace events.

## Checklist

## How to test these changes

1. Configure an agent with sample rate 0.016
2. Trigger requests that will cause transactions and spans to be sent to APM Server
3. Check that each transaction and span document has a `transaction.representative_count` or `span.representative_count` field with value 62.5.

## Related issues

Closes https://github.com/elastic/apm-server/issues/9340


internal/processor/stream/test_approved_es_documents/testIntakeIntegrationSpanLinks.approved.json
```diff
@@ -97,6 +97,7 @@
             },
             "transaction": {
                 "id": "abcdef1478523690",
+                "representative_count": 1,
```
sqren, line 100: Is "representative_count" equivalent of "sampling rate"? If so, why not call it the later?

axw: It's effectively the inverse sampling rate, except for old agents which send all transactions regardless of whether they're sampled: in this case we don't know the sample rate, so we set representative_count to 1 for every event we record.

Less important (since it could also be done with a runtime field): representative_count will be easier to deal with in https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-metrics-weight-avg-aggregation.html

simitt (commented):
Ack that this is currently in draft; I don't see any reason why not to move forward with this.

lahsivjar (approved):
LGTM

simitt:
Tested with BC5 on ESS: set sampling rate for go agent instrumented app and verified that `*.representative_count` fields were set accordingly.

PR #9463, opened 2022-10-28
modelindexer: disable scale up when 429 > 1%

marclop:
## Motivation/summary

Disables the scale up actions when the 429 response rate exceeds 1% of the total response rate. Additionally, scale down respecting the scale down parameters when the rate is breached.

## Checklist

## How to test these changes

Run benchmarks with >= APM Server 8GB against a small ES (8gb) for example.

## Related issues

Part of #9181


axw (approved):
LGTM!

Did you consider scaling down based on any failures, rather than just 429s?

marclop:
@axw I did, however, "other" failures could be due to malformed documents (bad mappings, values, etc), so I decided not to for the time being.

The other status codes we may consider is looking for 499 (client timeouts?), 502 and 503? That however, we can do on a follow up PR, perhaps since I haven't tested it, and we're not collecting those already.

axw:
Verified with 8.6.0-BC5, running on a GCP VM. I pointed it at an ESS cluster's Elasticsearch; scaled up ES and waited for APM Server to scale up the indexers; then scaled down ES an observed APM Server scale down too.

Issue #9485, opened 2022-11-01
Disable indexing for custom application metrics


axw:
In https://github.com/elastic/apm-server/pull/9316, we have added `index: false` to well-defined metrics, such as system metrics (e.g. CPU usage), and language/runtime-specific metrics (e.g. JVM garbage collector metrics). The rationale behind this is that it is unusual to search on metric values -- they are typically only aggregated -- so we can disable indexing to save some storage space.

We should also do this for custom, application-defined metrics. These are dynamically mapped, so we would need to update the dynamic templates used for mapping them: https://github.com/elastic/apm-server/blob/d87c885d3cedb15c45adb7ca3c007157aa60a9a6/apmpackage/apm/data_stream/app_metrics/manifest.yml#L20-L32

lahsivjar:
Tested with the latest 8.6.0 BC: `8.6.0-8cf9e954`. Validated that indexing is `false` for custom app metrics.

PR #9504, opened 2022-11-03
feat: mark custom metrics as doc-value-only

kruskall:
## Checklist

## Related issues

Closes https://github.com/elastic/apm-server/issues/9485


apmpackage/apm/data_stream/app_metrics/manifest.yml
```diff
@@ -30,8 +30,10 @@ elasticsearch:
         - double:
             mapping:
               type: double
+              index: false
         - numeric_labels:
             path_match: numeric_labels.*
             mapping:
               type: scaled_float
+              index: false
```
axw, line 38: I don't think we want this. Labels may be used for searching; they're not necessarily metrics.

axw (changes requested):

apmpackage/apm/data_stream/app_metrics/manifest.yml
kruskall: Oh, good point! Updated :+1:

axw (approved):

Issue #8958, opened 2022-08-24
Adapt to Elastic Agent v2 spec and libbeat changes


simitt:
Elastic Agent switches to the v2 spec in `8.5` and most likely removes support for v1. Libbeat is updated accordingly (https://github.com/elastic/beats/pull/32673). 

APM Server needs to align with these changes to ensure it keeps working with the Elastic Agent in `8.5`. See https://github.com/elastic/beats/pull/32673 for more detailed information on any required changes. 

This is time critical once the changes are merged into libbeat and Elastic Agent, to not break development images and automated test setups. 

@cmacknz @blakerouse please notify here when changes need to be applied.

cmacknz:
The agent [feature-arch-v2](https://github.com/elastic/elastic-agent/tree/feature-arch-v2) branch is ready for testing with APM server now. 

You will need likely the libbeat version from https://github.com/elastic/beats/pull/32673 to get the new protocol client. It has been tested with Filebeat, Metricbeat, and Osquerybeat so far.

The changes on your side should hopefully be small. Let us know if you encounter any unexpected issues.

simitt:
The switch to Elastic Agent v2 architecture is delayed. Changing the labels on this accordingly.

axw:
Closed by https://github.com/elastic/apm-server/pull/9544

PR #9544, opened 2022-11-11
Adapt to Elastic Agent v2 protocol changes

axw:
## Motivation/summary

Update libbeat, and adapt to the Elastic Agent v2 protocol changes.

Synthetic source is disabled while we attempt to isolate a bug in Elasticsearch. `TestIngestPipelineDataStreamMigration` was failing, and could be made to pass after either removing the `_doc_count` field from the source document, or disabling synthetic source.

## Checklist

## How to test these changes

N/A -- we can rely on system tests and ESS smoke tests

## Related issues

https://github.com/elastic/apm-server/issues/8958


axw:
Fleet Server is failing to start with the latest 8.6.0-SNAPSHOT, so we're blocked at the moment.

axw:
Latest snapshot appears to be working :crossed_fingers:

axw:
Or not. Various system tests, such as TestProfiling, are failing. Unclear why at the moment.

axw:
Ruled out other stack component changes. Something to do with Elastic Agent and/or Fleet Server. TestProfiling passes if I add a sleep just after creating the integration, so may be related to unexpected server restarts.

axw:
> so may be related to unexpected server restarts.

Indeed, Elastic Agent (or libbeat) is sending the same input config twice. We can suppress multiple reloads of the same config by comparing the revision to the most recently applied config.

@cmacknz is this behaviour expected?

axw:
OK, stale config changes are now suppressed, and that gets some system tests passing.

The TestRUMErrorSourcemapping system test is failing in Fleet mode. I haven't had time to dig into that one yet, but this morning I saw a comment that Endpoint is having issues fetching Fleet artifacts: https://github.com/elastic/kibana/issues/145202. (We store and deliver source maps to APM Server as Fleet Artifacts.)

axw:
@cmacknz I don't see `fleet` in the input config with the v2 changes. That means APM Server doesn't know how to connect to Fleet Server for fetching artifacts. Is there a new way to get that now?

cmacknz:
> Indeed, Elastic Agent (or libbeat) is sending the same input config twice. We can suppress multiple reloads of the same config by comparing the revision to the most recently applied config.

No this is not intentional, we'll file an issue for this to follow up on after we address the higher priority issues.

>  I don't see fleet in the input config with the v2 changes. That means APM Server doesn't know how to connect to Fleet Server for fetching artifacts. Is there a new way to get that now?

The fleet configuration isn't provided by default anymore. We will have to modify the agent to send it to you. We already do this for endpoint security so hopefully we can change this quickly.

cmacknz:
Issues to track the issues discovered here so far:

1. https://github.com/elastic/elastic-agent/issues/1737
2. https://github.com/elastic/elastic-agent/issues/1738

@axw is there an individual test that reproduces the multiple configuration issue? Or did you find this by testing manually?

axw:
> is there an individual test that reproduces the multiple configuration issue? Or did you find this by testing manually?

I noticed it while running one of our system tests, TestProfiling, but it only became obvious by looking through logs and debugging.

FWIW: the reason it fails is because the test is telling APM Server to listen on an ephemeral port, parses the logs to find which port it actually opened, and then connects to it to run some tests. The extraneous reload causes APM Server to internally restart/hot reload, and open a new ephemeral port.

I see the issue can be reproduced in Shipper (https://github.com/elastic/elastic-agent/issues/1738#issuecomment-1317751632), so assuming it can be fixed I'll revert my changes to suppress multiple reloads after that lands.

> The fleet configuration isn't provided by default anymore. We will have to modify the agent to send it to you. We already do this for endpoint security so hopefully we can change this quickly.

Thanks.

axw:
Next issue: apm-server appears not to be creating a unix socket for serving /stats or /state.

Looks like there's an error in the elastic-agent code that enriches command line args: https://github.com/elastic/elastic-agent/blob/969c18a6cec550840f2b488a2d3e10fe870e0af6/internal/pkg/agent/application/monitoring/v1_monitor.go#L157

APM Server does understand the `http.*` and `logging.*` flags.

axw:
Opened https://github.com/elastic/elastic-agent/pull/1748 to fix the monitoring issue

axw:
With that fix, all system tests are passing for me locally.

axw:
> With that fix, all system tests are passing for me locally.

Spoke too soon, there's one failing test: TestIngestPipelineDataStreamMigration. That's not likely related to Elastic Agent or Fleet though. Looks like something is up with Elasticsearch. I'm getting this error when searching `metrics-apm.internal-*` after running that test:

```json
{
  "error": {
    "root_cause": [
      {
        "type": "array_index_out_of_bounds_exception",
        "reason": "Index 129 out of bounds for length 129"
      }
    ],
    "type": "search_phase_execution_exception",
    "reason": "all shards failed",
    "phase": "query",
    "grouped": true,
    "failed_shards": [
      {
        "shard": 0,
        "index": ".ds-metrics-apm.internal-migrated-2022.11.18-000001",
        "node": "mZGHctCjQZirjOrrnDnuRQ",
        "reason": {
          "type": "array_index_out_of_bounds_exception",
          "reason": "Index 129 out of bounds for length 129"
        }
      }
    ],
    "caused_by": {
      "type": "array_index_out_of_bounds_exception",
      "reason": "Index 129 out of bounds for length 129",
      "caused_by": {
        "type": "array_index_out_of_bounds_exception",
        "reason": "Index 129 out of bounds for length 129"
      }
    }
  },
  "status": 500
}
```

axw:
Removing `_doc_count` from the transaction metric doc that we're reindexing allows the test to pass. I suspect there's a bug in Elasticsearch with synthetic source, still digging.

axw:
Confirmed, disabling synthetic source enables the test to pass.

internal/beatcmd/reloader.go
```diff
@@ -126,10 +148,15 @@ func (r *Reloader) reloadInputs(configs []*reload.ConfigWithMeta) error {
 func (r *Reloader) reloadOutput(cfg *reload.ConfigWithMeta) error {
 	r.mu.Lock()
 	defer r.mu.Unlock()
+	if r.outputConfig != nil && configEqual(cfg.Config, r.outputConfig) {
```
simitt, line 151: Is the `configEqual` check strictly necessary for switching to Elastic Agent v2? In the past we consciously didn't introduce `reflect.*` in production code. In which cases would the config be reloaded without any change, do we expect performance improvement from this?

simitt (approved):
LGTM apart from one question/concern.

systemtest/approvals/TestAgentConfig.approved.json
```diff
@@ -3,24 +3,17 @@
...
-                "dataset": "apm.internal",
-                "namespace": "default",
-                "type": "metrics"
-            },
-            "ecs": {
-                "version": "dynamic"
-            },
+            "data_stream.dataset": "apm.internal",
+            "data_stream.namespace": "default",
+            "data_stream.type": "metrics",
```
simitt, line 8: Why did the keys change here? This does not look related to EA v2 changes.

axw: This is related to disabling synthetic source. Synthetic source always produces hierarchical objects, but the original docs use flattened fields in some cases.

internal/beatcmd/reloader.go
axw: This is necessary for working around https://github.com/elastic/elastic-agent/issues/1738

This is not a hot code path, so it won't matter. Nevertheless I intend to back this change out after the issue is fixed.

PR #9553, opened 2022-11-14
ci: Populate `USER` variable with build id

marclop:
## Motivation/summary

Populates the empty variable `USER` with `bench-ci-<build_id>` to avoid jobs from running into collisions should any of the previous jobs fail to clean up the resources properly.

## Checklist

## How to test these changes

N/A

## Related issues

Failing nightly benchmarks

### DIFF
--- a/.ci/benchmarks.groovy
+++ b/.ci/benchmarks.groovy
@@ -46,6 +46,7 @@ pipeline {
         TF_VAR_public_key = "./id_rsa_terraform.pub"
         BENCHMARK_RESULT = "benchmark-result.txt"
         TFVARS_SOURCE = "system-profiles/8GBx1zone.tfvars" // Default to use an 8gb profile
+        USER = "benchci-${env.BUILD_ID}" // use build as prefix
         // cloud tags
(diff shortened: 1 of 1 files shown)


.ci/benchmarks.groovy
```diff
@@ -46,6 +46,7 @@ pipeline {
         TF_VAR_public_key = "./id_rsa_terraform.pub"
         BENCHMARK_RESULT = "benchmark-result.txt"
         TFVARS_SOURCE = "system-profiles/8GBx1zone.tfvars" // Default to use an 8gb profile
+        USER = "benchci-${env.BUILD_ID}" // use build as prefix
```
v1v, line 49: there are corner-cases if the same Build ID is used between executions, though IIUC, this won't be the case by default since it won't run on PRs but on branches. is that right?

marclop: I hadn't considered that, but you're right, we cover 99% of the runs with this prefix. In practice, I don't think we're running the benchmarks in PRs right now.

v1v (approved):

v1v (approved):

amannocci (approved):

PR #9558, opened 2022-11-14
Universal Profiling: implement custom ILM strategy for K/V data

inge4pres:
## Motivation/summary

In Universal Profiling beta we use both data streams and raw indices.
Elasticsearch does not provide a built-in ILM strategy to handle K/V data (discussing in https://github.com/elastic/elasticsearch/issues/85273), and we ought give users the ability to limit the storage used for profiling data.

We were able to provide ILM policies for data stream, but for raw indices we yet don't have nothing.
This PR provides basic ILM capabilities to rollover raw indices using a sliding window data lifecyle: the index age gates the rollover.

## Checklist

- Is it observable through the addition of either **logging** or **metrics**? ✅ 
- Is its use being published in **telemetry** to enable product improvement? 🔴 

## How to test these changes

Locally with runapm + fleetctl:
* start the docker compose stack and run the Elastic Agent container with `runapm`
* create an API key with the `admin` role
* fetch the package policy ID from Fleet for the APM input
* use `fleetctl` to configure profiling intake and the new ILM runners
```bash
go run main.go set-policy-config <PACKAGE_POLICY_ID>  apm-server.profiling.enabled=true apm-server.profiling.elasticsearch.api_key="<API_KEY>" apm-server.profiling.metrics.elasticsearch.hosts='["http://localhost:9200"]' apm-server.profiling.metrics.elasticsearch.api_key="<API_KEY>" apm-server.profiling.keyvalue_retention.execution_interval="10s" server.profiling.keyvalue_retention.age="10m"
```
* ingest data with a profiling host-agent connected to the apm-server intake
* observe the metrics `"beat.stats.apm_server.profiling.ilm.custom_ilm.execution.count"` and `"beat.stats.apm_server.profiling.ilm.custom_ilm.skipped_because_time_constraints.count"`, representing the counters of when indices rollover is taking place or skipped, respectively


testing/docker/kibana/kibana.yml
```diff
@@ -19,3 +19,5 @@ xpack.fleet.agentPolicies:
         id: default-fleet-server
         package:
           name: fleet_server
+
+xpack.profiling.enabled: true
```
inge4pres, line 23: Intentionally added to speed up development: if you prefer to remove it let me know

x-pack/apm-server/profiling/ilm.go
```diff
@@ -0,0 +1,474 @@
...
+
+// ILMRunner describes the behavior of a custom ILM strategy executor
+type ILMRunner interface {
+	configure() error
+	preFlight(policy ilmPolicy) error
+	execute(policy ilmPolicy) error
+}
+
+type optimisticAliasSwap struct {
+	client *es.Client
```
inge4pres, line 130: We can't rely on the client defined in other parts of apm-server, at least this is what I could grasp.
Is there a concern in adding the v8 client here?

internal/beater/config/profiling.go
```diff
@@ -112,13 +127,31 @@ func (c *ProfilingConfig) setup(log *logp.Logger, outputESCfg *config.C) error {
...
 		Enabled:         false,
 		ESConfig:        elasticsearch.DefaultConfig(),
 		MetricsESConfig: elasticsearch.DefaultConfig(),
+		ILMConfig: &ILMConfig{
+			// 2 months
+			Age: 1440 * time.Hour,
+			// 100 GiB
+			SizeInBytes: 107374182400,
+			Interval:    12 * time.Hour,
+		},
```
inge4pres, line 155: Defaults defined in case a user config is not provided

internal/beater/config/profiling.go
```diff
@@ -112,13 +127,35 @@ func (c *ProfilingConfig) setup(log *logp.Logger, outputESCfg *config.C) error {
...
 	}
+	if c.ilm != nil {
+		if err := c.ilm.Unpack(&c.ILMConfig); err != nil {
+			return errors.Wrap(err, "error unpacking apm-server.profiling.keyvalue_retention config for profiling K/V data retention")
+		}
+	}
 	return nil
 }
 
+type ILMConfig struct {
```
axw, line 138:
```suggestion
type ProfilingILMConfig struct {
```

testing/docker/kibana/kibana.yml
axw: I think this is fine, thanks.

x-pack/apm-server/main.go
```diff
@@ -250,6 +250,9 @@ func runServerWithProcessors(ctx context.Context, runServer beater.RunServerFunc
 func newProfilingCollector(args beater.ServerParams) (*profiling.ElasticCollector, func(context.Context) error, error) {
 	logger := args.Logger.Named("profiling")
 
+	ctx, stopILM := context.WithCancel(context.Background())
+	profiling.ScheduleILMExecution(ctx, logger.Named("ilm"), args.Config.Profiling)
```
axw, line 254: Should this be moved down close to the NewCollector call so the goroutine won't be left running if this function returns with an error?

x-pack/apm-server/main.go
```diff
@@ -287,6 +290,7 @@ func newProfilingCollector(args beater.ServerParams) (*profiling.ElasticCollecto
 		logger,
 	)
 	cleanup := func(ctx context.Context) error {
+		stopILM()
```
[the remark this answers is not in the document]
axw: Cleanup is called when server shutdown begins. This is fine.

x-pack/apm-server/profiling/ilm.go
axw: It's fine. I think you could use the internal/elasticsearch.Client interface by using the esapi request struct types, but it doesn't matter.

x-pack/apm-server/profiling/ilm.go
```diff
@@ -0,0 +1,474 @@
...
+			o.logger.Warnf("Failed to index document in lock index: %v", err)
+		}
+	}()
+	rolloverCounter.Inc()
+
+	// Create a new index: setting the right name will allow inheriting
+	// the configuration from the template
+	newIdx := fmt.Sprintf("%s-%s", o.currentAlias,
+		time.Now().UTC().Format("2006.02.01-15.04.05"))
+	if resp, err := o.client.Indices.Create(newIdx); checkESAPIError(http.StatusOK,
```
axw, line 188: I think you should always be closing resp.Body as long as err == nil. I didn't check through ES API calls, but all the ones I saw were using this pattern, and it only closes the body when the status code indicates an error.

axw (approved):
LGTM. I've only scanned through the main business logic for things which might impact regular apm-server behaviour.

x-pack/apm-server/profiling/ilm.go
inge4pres: Are you referring to `checkESAPIError` in general or this particular call site?
What you mentioned is true for `http.Response.Body`, is it the same for `esapi.Response`?

x-pack/apm-server/main.go
inge4pres: I actually left it outside to signal that it's a different route than the collector, and because it uses parts of the configuration from `args` that i'd prefer not to propagate to the collector (the ILM strategy here is temporary™️).

I'll move it below after all err handling, just before the `cleanup` closure.

x-pack/apm-server/profiling/ilm.go
axw: Yes, you are expected to close the response body for `esapi.Response`s too.

The `esapi.Response.String()` doc comment says it is intended for debugging. I guess it's fine, just pointing that out in case you missed it.

inge4pres: NVM, just read the comment on top of it...

inge4pres: So yeah, after re-reading `elastictransport.Client.Perform` implementation, I _guess_ it's safer to always close the body.
I'll apply these changes.

Issue #9018, opened 2022-08-31
txmetrics: remove `timeseries.instance` field


axw:
The `timeseries.instance` field is not used anywhere. Eventually we should move these metrics to using TSDB, which generates a timeseries ID from the dimensions.

simitt:
Wasn't the original intention to use this e.g. for telemetry to get a better understanding of usage and potential issues with the metrics aggregation?

axw:
@simitt I don't recall tbqh. If it was, we could also do that with TSDB using the [`_tsid`](https://www.elastic.co/guide/en/elasticsearch/reference/current/tsds.html#tsid) field. We could alternatively compare transaction event count to transaction metric count for a rough approximation.

PR #9565, opened 2022-11-15
Remove `timeseries.instance`

axw:
## Motivation/summary

Remove this field as we are not using it, and we can use `_tsid` instead when we migrate to TSDS.

## Checklist

## How to test these changes

1. Ensure transaction metrics are recorded
2. Check that `timeseries.instance` is not indexed

## Related issues

Closes https://github.com/elastic/apm-server/issues/9018


simitt (approved):

apmpackage/apm/data_stream/internal_metrics/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -16,3 +16,7 @@ processors:
   - remove:
       field: _metric_descriptions
       ignore_missing: true
+  - remove:
+      # Removed in 8.6.0
+      field: timeseries.instance
```
marclop, line 21: When would this stage of the ingest pipeline be triggered?

From what I understand, we're not emitting the filed from the APM Server anymore and everything is version aligned.

axw: This is to handle data coming from an old APM Server, that hasn't yet been upgraded. First we upgrade Kibana, which upgrades the integration package; then we upgrade Elastic Agent and therefore APM Server.

axw: So to be a little clearer:

> everything is version aligned

Not exactly. What we guarantee is that APM Server's version is no greater than Kibana/the integration package's version.

PR #9575, opened 2022-11-16
ci: Disable failing benchmark report

marclop:
## Motivation/summary

Disables the Kibana PNG download and reporting to Slack. Main reason is that it causes the build to fail unnecessarily.


amannocci (approved):
LGTM

PR #9580, opened 2022-11-16
testing/k8s: Set memory limit, deployment strategy

marclop:
## Motivation/summary

Updates the resource "limits" and "requests" for fleet-server, the ECK operator and the apm-server so memory is allocated a bit better in machines with limited RAM.

## How to test these changes

Non functional changes

## Related issues

None

### DIFF
--- a/testing/infra/k8s/overlays/local/kustomization.yaml
+++ b/testing/infra/k8s/overlays/local/kustomization.yaml
@@ -4,6 +4,24 @@ resources:
 
 patches:
 - patch: |-
+    apiVersion: apps/v1
+    kind: StatefulSet
+    metadata:
+      name: elastic-operator
+      namespace: elastic-system
+    spec:
+      selector:
+        control-plane: elastic-operator
+      template:
+        spec:
+          containers:
+            - name: manager
+              resources:
+                limits:
+                  memory: "512Mi"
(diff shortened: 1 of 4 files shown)


testing/infra/k8s/overlays/local/kustomization.yaml
```diff
@@ -3,6 +3,24 @@ resources:
...
+        control-plane: elastic-operator
+      template:
+        spec:
+          containers:
+            - name: manager
+              resources:
+                limits:
+                  memory: "512Mi"
+                requests:
+                  memory: "256Mi"
```
axw, line 23: We could potentially be even more conservative here. I just took a look at the eck-operator Helm chart, and it requests 150Mi by default: https://github.com/elastic/cloud-on-k8s/blob/b79bb0f644d568f0b082438f0222bf8f2b621ff7/deploy/eck-operator/values.yaml#L35

axw (approved):
Nice!

testing/infra/k8s/overlays/local/kustomization.yaml
marclop: I see, I haven't seen the container go higher than 100MB to be honest. I'll change it

PR #9581, opened 2022-11-16
universal profiling: reduce tombstones on stacktraces index

inge4pres:
## Motivation/summary

In Universal Profiling we currently store a `@timestamp` field in `profiling-stacktraces` index.
This field is not used in the UI, and its presence generates a ~25% storage overhead due to tombstones when documents are replaced (host-agent periodically re-sends them).
Since we introduced ILM for K/V indices with #9558, we can remove this field.

## Checklist

- [X] Update [CHANGELOG.asciidoc](https://github.com/elastic/apm-server/blob/main/CHANGELOG.asciidoc)

- Is it observable through the addition of either **logging** or **metrics**? N/A
- Is its use being published in **telemetry** to enable product improvement? N/A

## How to test these changes

Steps are the same as #9558, but instead of reading metrics on ILM execution, you can should:

* add a data view for the `profiling-stacktraces-*` index pattern
* query the data view, see that no documents contain the `@timestamp` field


x-pack/apm-server/profiling/timestamp.go
```diff
@@ -8,14 +8,6 @@ import (
...
 
-// GetStartOfWeek returns the seconds since the epoch for time 00:00:00 of the current week.
-// Updating an existing index entry is expensive as the old entry is marked 'deleted'
-// and a new entry is created. But this only happens if at least one value changed.
-// Since a per-week resolution is enough for us, we remove the second-of-the-day.
-func GetStartOfWeek() uint32 {
-	return GetStartOfWeekFromTime(time.Now())
-}
-
 // GetStartOfWeekFromTime is the same as GetStartOfWeek, except that it doesn't take the
```
rockdaboot, line 11: Can you also copy the amended comment from the prodfiler PR here ?

inge4pres:
I noticed the failing test: it's due to the removal of `@timestamp` in this PR.

axw (approved):
@inge4pres you can fix the system test by running `make check-approvals` in the top of the repo, and committing the modified `.approved` file.

rockdaboot (approved):
:+1:

inge4pres:
> @inge4pres you can fix the system test by running `make check-approvals` in the top of the repo, and committing the modified `.approved` file.

I think there's more than that unfortunately, apparently the removal of the timestamp field messed up the retrieved events.
I'll try your suggestion as well, thanks for the pointer Andrew! 🙏🏼

inge4pres:
I refactored the test to be more similar to what actually happens now in ES.
See e71af7663

Issue #9596, opened 2022-11-17
APM server 8.5.1 won't start on Kubernetes 1.24


MeijerM1:
**APM Server version** (`apm-server version`): 
8.5.1

**Description of the problem including expected versus actual behavior**:
APM server 8.5.1 is unable to start with the following error:
```
error initializing beat: error loading config file: open apm-server.yml: permission denied
```

**Steps to reproduce**:
Use ECK 2.4.0 to rollout APM server 8.5.1 on Kubernetes 1.24.

APM Definition:
```
apiVersion: apm.k8s.elastic.co/v1
kind: ApmServer
metadata:
  labels:
    app: apm-server
  name: apm
  namespace: my-namespace
spec:
  version: 8.5.1
  count: 2
  kibanaRef:
    name: kibana
  elasticsearchRef:
    name: elasticsearch
  secureSettings:
    - secretName: apm-secret-settings
  http:
    tls:
      selfSignedCertificate:
        disabled: true
  podTemplate:
    metadata:
      labels:
        app: apm-server
    spec:
      containers: 
      - name: apm-server
      automountServiceAccountToken: true
      topologySpreadConstraints:
        - maxSkew: 1
          topologyKey: topology.kubernetes.io/zone
          whenUnsatisfiable: DoNotSchedule
          labelSelector:
            matchLabels:
              common.k8s.elastic.co/type: apm-server
      securityContext:
        runAsUser: 1000
        runAsGroup: 2000
        fsGroup: 2000
  config:
    setup.template.overwrite: true
    apm-server:
      ilm:
        enabled: "auto"
        setup:
          overwrite: true
          require_policy: false
    output:
      elasticsearch:
        hosts: ["some-url"]
        username: myUser
        password: myPassword 
```
**Provide logs (if relevant)**:
The `elasticsearch-internal-init-keystore` init container logs the following:
```
Initializing keystore.
+ keystore_initialized_flag=/usr/share/apm-server/data/elastic-internal-init-keystore.ok
+ [[ -f /usr/share/apm-server/data/elastic-internal-init-keystore.ok ]]
+ echo 'Initializing keystore.'
+ /usr/share/apm-server/apm-server keystore create --force
error initializing beat: error loading config file: open apm-server.yml: permission denied
```

SansDK:
You should use the Elastic agent.

simitt:
[...]

PR #9604, opened 2022-11-18
docker: fix file access for apm-server user

axw:
## Motivation/summary

Ensure the apm-server.yml is readable by the apm-server user (UID 1000), and the `data/` directory is writeable by the same user.

This enables users to run the image with a non-root GID. This was possible in older versions, as the apm-server user was created with the root supplementary group.

## Checklist

## How to test these changes

1. Create a local Kubernetes cluster using Kind
2. Install ECK 2.4.0
3. Create Elasticsearch and Kibana 8.5.1 resources
4. Create an ApmServer 8.5.1 resource with non-root user and group; this should fail to come up, per the linked issue
5. Build the image off this branch (or take a snapshot/BC image), and replace the ApmServer image (`spec.image: ...`); this should work.

## Related issues

Closes https://github.com/elastic/apm-server/issues/9596


changelogs/head.asciidoc
```diff
@@ -18,6 +18,7 @@ https://github.com/elastic/apm-server/compare/8.5\...main[View commits]
 [float]
 ==== Bug fixes
 - Null APM agent configuration entries are now treated as empty, and will not cause the server to exit with an error {pull}9546[9546]
+- Change ownership of apm-server.yml and data/ to apm-server user in Docker image {pull}TODO[TODO]
```
marclop, line 21:
```suggestion
- Change ownership of apm-server.yml and data/ to apm-server user in Docker image {pull} 9604[9604]
```

marclop (approved):

packaging/docker/Dockerfile
```diff
@@ -77,6 +77,7 @@ RUN sed -i 's/localhost:8200/0.0.0.0:8200/' apm-server.yml
 RUN sed -i 's/localhost:9200/elasticsearch:9200/' apm-server.yml
 RUN mkdir --mode=0775 data
 RUN echo 'apm-server:*:1000:0::/usr/share/apm-server:/bin/false' >> /etc/passwd
+RUN chown apm-server apm-server.yml data # ensure apm-server user can read/write
```
kruskall, line 80: **suggestion**: I think we can use `COPY --chown` (https://docs.docker.com/engine/reference/builder/#copy) to avoid calling chown a creating a layer

changelogs/head.asciidoc
axw: Oops, thanks :)

packaging/docker/Dockerfile
axw: We can do that for apm-server.yml, but the data directory is created with mkdir above, so we still need to chown.

axw: Updated

PR #9625, opened 2022-11-22
compliance: Tag CSP resources according to policy

marclop:
## Motivation/summary

Introduces a new `tags` in `testing/infra/terraform/modules/tags` which defines the default CSP tags and accepts a `project` variable, to define the purpose of each importing module.

Also re-formats a few files, and updates the default values for a couple of unknown variables if unset.

## How to test these changes

N/A

## Related issues

CSP tagging policy.

### DIFF
--- a/testing/infra/terraform/modules/tags/README.md
+++ b/testing/infra/terraform/modules/tags/README.md
@@ -1,5 +1,5 @@
 <!-- BEGIN_TF_DOCS -->
-## Rally worker module
+## Default tags module
 
 This modules exports the default tags / labels to use on Cloud Service Providers based on the Elastic tagging policy.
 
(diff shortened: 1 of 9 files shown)


testing/infra/terraform/modules/tags/README.md
```diff
@@ -0,0 +1,18 @@
+<!-- BEGIN_TF_DOCS -->
+## Rally worker module
```
lahsivjar, line 2: Default tags module maybe?

testing/infra/terraform/modules/tags/header.md
```diff
@@ -0,0 +1,3 @@
+## Rally worker module
```
lahsivjar, line 1: Same as above

lahsivjar (approved):
LGTM, minor fix required.

simitt:
Thanks for picking this up that quickly!

Issue #9643, opened 2022-11-22
Fetch sourcemaps from Elasticsearch


simitt:
## Problem Statement

At the moment the APM integration requires Fleet for some features. Because of this, Elastic Agent standalone mode is not supported for APM Server. Other integrations do not have this requirement. It means that users need to use Fleet as soon as they want to add the APM integration to an Elastic Agent policy. We want to remove this dependency, while still keeping support for Fleet.

One of the features that needs to be changed is how the APM Server retrieves sourcemaps. The standalone APM Server and the Fleet managed APM Server have implemented two different logics. The Fleet managed APM Server queries from Fleet, to which the APM UI syncs the sourcemaps. The standalone APM Server either fetches from Kibana or from Elasticsearch. 

## Solution Outline
We want to implement one common way to fetch the sourcemaps from Elasticsearch, replacing all other current implementations. 

In contrast to the current Elasticsearch implementation, the server will have a regularly running background job that fetches sourcemap metadata from Elasticsearch and caches them. Whenever processing a stack trace, the server then queries the cache with the metadata to identify if a certain sourcemap exists and if so, it fetches it from another cache or if not yet cached from Elasticsearch. 

## Implementation Details

The currently existing logic needs to be kept as a fallback. This helps to decouple the APM UI changes from the server changes. All the systemtests are expected to keep working (first with, then without the fallback). Once the APM UI team has implemented the required changes, the fallback in the server should be removed.

For the new logic, the APM Server starts a background routine on startup, that fetches the sourcemap metadata from Elasticsearch every 30 seconds. An internal cache is updated with the result (old sourcemaps that have not been returned anymore by Elasticsearch need to be removed from cache). The sourcemap metadata contain the `service.name` , `service.version`, `bundle_filepath` and a `content-hash` over the sourcemap body. The `service.name` , `service.version`, `bundle_filepath` create the unique identifier for a sourcemap. 

The sourcemap bodies are cached in another cache. This should be a LRU cache with a configurable size. Once a sourcemap is cached in the LRU cache it only gets refreshed if the `content-hash` for the sourcemap changes. 

Whenever the server is processing a stacktrace, it checks the metadata cache if a certain sourcemap exists, if not, no further sourcemap handling is required. If a sourcemap metadata entry exists, it fetches the sourcemap content from the LRU cache, if not yet cached, it fetches from Elasticsearch and adds the sourcemap body to the LRU cache. 

[...]

PR #8193, opened 2022-05-23
agentcfg: implement Elasticsearch fetcher

axw:
## Motivation/summary

Implement a new `agentcfg.Fetcher` which queries agent config from Elasticsearch directly, searching the `.apm-agent-configuration` index created by Kibana.

This enables running APM Server standalone, directly or with Elastic Agent, without requiring APM Server to communicate with Kibana to enable APM agent configuration.

## Checklist

- [ ] Update [CHANGELOG.asciidoc](https://github.com/elastic/apm-server/blob/main/CHANGELOG.asciidoc)
~- [ ] Update [package changelog.yml](https://github.com/elastic/apm-server/blob/main/apmpackage/apm/changelog.yml) (only if changes to `apmpackage` have been made)~
- [ ] Documentation has been updated

## How to test these changes

1. `docker-compose up -d`
2. `apm-server -E output.elasticsearch.username=admin -E output.elasticsearch.password=changeme -E instrumentation.enabled=true`
3. create agent configuration matching service name "apm-server"
4. `curl http://localhost:8200/config/v1/agents?service.name=apm-server`

The response should include the matching agent configuration.

## Related issues

None

PR #9722, opened 2022-12-02
feat: fetch sourcemaps from Elasticsearch

kruskall:
## Motivation/summary

Add a new caching metadata fetcher that fetch sourcemap metadata from ES and forward the request to the backend fetcher if it exists. The backend fetcher is an ES fetcher wrapped in a LRU cache to cache the sourcemap body.
Sourcemap metadata are periodically synced with ES by the caching metadata fetcher.
Keep fleet and kibana fetcher until the UI team has updated the sourcemap uploading flow.
Update ES sourcemap response to include sourcemap metadata.

## Checklist

## Related issues

Closes https://github.com/elastic/apm-server/issues/9643

### DIFF
--- a/internal/beater/beater.go
+++ b/internal/beater/beater.go
@@ -320,13 +320,7 @@ func (s *Runner) Run(ctx context.Context) error {
 		if err != nil {
 			return err
 		}
-		cachingFetcher, err := sourcemap.NewCachingFetcher(
-			fetcher, s.config.RumConfig.SourceMapping.Cache.Expiration,
-		)
-		if err != nil {
-			return err
-		}
-		sourcemapFetcher = cachingFetcher
+		sourcemapFetcher = fetcher
 	}
 
 	// Create the runServer function. We start with newBaseRunServer, and then
@@ -827,18 +821,33 @@ func newSourcemapFetcher(
 			}
 		}
 
-		return sourcemap.NewFleetFetcher(
+		ff, err := sourcemap.NewFleetFetcher(
 			&client,
 			fleetCfg.AccessAPIKey,
 			fleetServerURLs,
 			artifactRefs,
 		)
+		if err != nil {
+			return nil, err
+		}
+
+		cachingFetcher, err := sourcemap.NewCachingFetcher(ff, cfg.Cache.Expiration)
+		if err != nil {
+			return nil, err
+		}
+
+		return cachingFetcher, nil
 	}
 
 	// For standalone, we query both Kibana and Elasticsearch for backwards compatibility.
 	var chained sourcemap.ChainedFetcher
 	if kibanaClient != nil {
-		chained = append(chained, sourcemap.NewKibanaFetcher(kibanaClient))
+		cachingFetcher, err := sourcemap.NewCachingFetcher(sourcemap.NewKibanaFetcher(kibanaClient), cfg.Cache.Expiration)
+		if err != nil {
+			return nil, err
+		}
+
+		chained = append(chained, cachingFetcher)
 	}
 	esClient, err := newElasticsearchClient(cfg.ESConfig)
 	if err != nil {
@@ -846,7 +855,15 @@ func newSourcemapFetcher(
 	}
 	index := strings.ReplaceAll(cfg.IndexPattern, "%{[observer.version]}", version.Version)
 	esFetcher := sourcemap.NewElasticsearchFetcher(esClient, index)
-	chained = append(chained, esFetcher)
+	cachingFetcher, err := sourcemap.NewCachingFetcher(esFetcher, cfg.Cache.Expiration)
+	if err != nil {
+		return nil, err
+	}
+	metadataCachingFetcher := sourcemap.NewMetadataCachingFetcher(esClient, cachingFetcher, index)
+	metadataCachingFetcher.StartBackgroundSync()
+
+	chained = append(chained, metadataCachingFetcher)
+
 	return chained, nil
 }
 
--- a/internal/sourcemap/elasticsearch.go
+++ b/internal/sourcemap/elasticsearch.go
@@ -59,7 +59,12 @@ type esSourcemapResponse struct {
 		Hits []struct {
 			Source struct {
 				Sourcemap struct {
-					Sourcemap string
+					Service struct {
+						Name    string
+						Version string
+					}
+					BundleFilepath string `json:"bundle_filepath"`
+					Sourcemap      string
 				}
 			} `json:"_source"`
 		}
--- a/internal/sourcemap/metadata.go
+++ b/internal/sourcemap/metadata.go
@@ -1,0 +1,206 @@
+// Licensed to Elasticsearch B.V. under one or more contributor
+// license agreements. See the NOTICE file distributed with
+// this work for additional information regarding copyright
+// ownership. Elasticsearch B.V. licenses this file to you under
+// the Apache License, Version 2.0 (the "License"); you may
+// not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package sourcemap
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"sync"
+	"time"
+
+	"github.com/go-sourcemap/sourcemap"
+	"github.com/pkg/errors"
+
+	"github.com/elastic/apm-server/internal/elasticsearch"
+	"github.com/elastic/apm-server/internal/logs"
+	"github.com/elastic/elastic-agent-libs/logp"
+	"github.com/elastic/go-elasticsearch/v8/esapi"
+)
+
+type Identifier struct {
+	name    string
+	version string
+	path    string
+}
+
+type Metadata struct {
+	id          Identifier
+	contentHash string
+}
+
+type MetadataCachingFetcher struct {
+	esClient elasticsearch.Client
+	set      map[Identifier]bool
+	mu       sync.RWMutex
+	backend  Fetcher
+	logger   *logp.Logger
+	once     sync.Once
+	index    string
+}
+
+func NewMetadataCachingFetcher(
+	c elasticsearch.Client,
+	backend Fetcher,
+	index string,
+) *MetadataCachingFetcher {
+	return &MetadataCachingFetcher{
+		esClient: c,
+		index:    index,
+		set:      make(map[Identifier]bool),
+		backend:  backend,
+		logger:   logp.NewLogger(logs.Sourcemap),
+	}
+}
+
+func (s *MetadataCachingFetcher) Fetch(ctx context.Context, name, version, path string) (*sourcemap.Consumer, error) {
+	if len(s.set) == 0 {
+		// First run, populate cache
+		s.once.Do(func() {
+			s.sync(ctx)
+		})
+	}
+
+	key := Identifier{
+		name:    name,
+		version: version,
+		path:    path,
+	}
+
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+
+	if _, found := s.set[key]; found {
+		// Only fetch from ES if the sourcemap id exists
+		return s.backend.Fetch(ctx, name, version, path)
+	}
+
+	return nil, nil
+}
+
+func (s *MetadataCachingFetcher) Update(ids []Identifier) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	for k := range s.set {
+		delete(s.set, k)
+	}
+
+	for _, k := range ids {
+		s.set[k] = true
+	}
+}
+
+func (s *MetadataCachingFetcher) StartBackgroundSync() {
+	go func() {
+		// TODO make this a config option ?
+		t := time.NewTicker(30 * time.Second)
+		defer t.Stop()
+
+		for {
+			select {
+			case <-t.C:
+				ctx, cleanup := context.WithTimeout(context.Background(), 10*time.Second)
+				s.sync(ctx)
+				cleanup()
+			}
+		}
+	}()
+}
+
+func (s *MetadataCachingFetcher) sync(ctx context.Context) error {
+	resp, err := s.runSearchQuery(ctx)
+	if err != nil {
+		return errors.Wrap(err, errMsgESFailure)
+	}
+	defer resp.Body.Close()
+
+	// handle error response
+	if resp.StatusCode >= http.StatusMultipleChoices {
+		if resp.StatusCode == http.StatusNotFound {
+			return nil
+		}
+		b, err := io.ReadAll(resp.Body)
+		if err != nil {
+			return errors.Wrap(err, errMsgParseSourcemap)
+		}
+		return errors.New(fmt.Sprintf("%s %s", errMsgParseSourcemap, b))
+	}
+
+	// parse response
+	body, err := parseResponse(resp.Body, s.logger)
+	if err != nil {
+		return err
+	}
+
+	var ids []Identifier
+	for _, v := range body.Hits.Hits {
+		id := Identifier{
+			name:    v.Source.Sourcemap.Service.Name,
+			version: v.Source.Sourcemap.Service.Version,
+			path:    v.Source.Sourcemap.BundleFilepath,
+		}
+
+		ids = append(ids, id)
+	}
+
+	// Update cache
+	s.Update(ids)
+	return nil
+}
+
+func (s *MetadataCachingFetcher) runSearchQuery(ctx context.Context) (*esapi.Response, error) {
+	var buf bytes.Buffer
+	if err := json.NewEncoder(&buf).Encode(queryMetadata()); err != nil {
+		return nil, err
+	}
+
+	req := esapi.SearchRequest{
+		Index:          []string{s.index},
+		Body:           &buf,
+		TrackTotalHits: true,
+	}
+	return req.Do(ctx, s.esClient)
+}
+
+func queryMetadata() map[string]interface{} {
+	return map[string]interface{}{
+		"_source": []string{"sourcemap.service.*", "sourcemap.bundle_filepath"},
+	}
+}
+
+func parseResponse(body io.ReadCloser, logger *logp.Logger) (esSourcemapResponse, error) {
+	b, err := io.ReadAll(body)
+	if err != nil {
+		return esSourcemapResponse{}, err
+	}
+
+	var esSourcemapResponse esSourcemapResponse
+	if err := json.Unmarshal(b, &esSourcemapResponse); err != nil {
+		return esSourcemapResponse, err
+	}
+	hits := esSourcemapResponse.Hits.Total.Value
+	if hits == 0 || len(esSourcemapResponse.Hits.Hits) == 0 {
+		return esSourcemapResponse, nil
+	}
+
+	return esSourcemapResponse, nil
+}
--- a/systemtest/sourcemap_test.go
+++ b/systemtest/sourcemap_test.go
@@ -18,6 +18,10 @@
 package systemtest_test
 
 import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"io"
 	"os"
 	"testing"
 
@@ -27,6 +31,7 @@ import (
 	"github.com/elastic/apm-server/systemtest"
 	"github.com/elastic/apm-server/systemtest/apmservertest"
 	"github.com/elastic/apm-server/systemtest/estest"
+	"github.com/elastic/go-elasticsearch/v8/esapi"
 )
 
 func TestRUMErrorSourcemapping(t *testing.T) {
@@ -129,6 +134,77 @@ func TestNoMatchingSourcemap(t *testing.T) {
 		// RUM events have the source port recorded, and in the tests it will be dynamic
 		"source.port",
 	)
+}
+
+func TestSourcemapElasticsearch(t *testing.T) {
+	t.Skip("DISABLED FOR NOW")
+
+	systemtest.CleanupElasticsearch(t)
+	srv := apmservertest.NewUnstartedServerTB(t)
+	srv.Config.RUM = &apmservertest.RUMConfig{Enabled: true}
+	err := srv.Start()
+	require.NoError(t, err)
+
+	sourcemap, err := os.ReadFile("../testdata/sourcemap/bundle.js.map")
+	require.NoError(t, err)
+
+	type Source struct {
+		Timestamp string `json:"@timestamp"`
+		Processor struct {
+			Name string `json:"name"`
+		} `json:"processor"`
+		Sourcemap struct {
+			Service struct {
+				Name    string `json:"name"`
+				Version string `json:"version"`
+			} `json:"service"`
+			BundleFilepath string `json:"bundle_filepath"`
+			Sourcemap      string `json:"sourcemap"`
+		} `json:"sourcemap"`
+	}
+
+	s := Source{
+		Timestamp: "1970-01-01T00:00:01.000Z",
+		Processor: struct {
+			Name string `json:"name"`
+		}{
+			Name: "sourcemap",
+		},
+		Sourcemap: struct {
+			Service struct {
+				Name    string `json:"name"`
+				Version string `json:"version"`
+			} `json:"service"`
+			BundleFilepath string `json:"bundle_filepath"`
+			Sourcemap      string `json:"sourcemap"`
+		}{
+			Service: struct {
+				Name    string `json:"name"`
+				Version string `json:"version"`
+			}{
+				Name:    "apm-agent-js",
+				Version: "1.0.1",
+			},
+			BundleFilepath: "http://localhost:8000/test/e2e/general-usecase/bundle.js.map",
+			Sourcemap:      string(sourcemap),
+		},
+	}
+
+	b, err := json.Marshal(s)
+	require.NoError(t, err)
+
+	_, err = systemtest.Elasticsearch.Do(context.Background(), &esapi.IndexRequest{
+		Index: "apm-test-sourcemap",
+		Body:  bytes.NewReader(b),
+	}, nil)
+	require.NoError(t, err)
+
+	systemtest.Elasticsearch.ExpectMinDocs(t, 1, "apm-*-sourcemap", nil)
+
+	// Index an error, applying source mapping and caching the source map in the process.
+	systemtest.SendRUMEventsPayload(t, srv.URL, "../testdata/intake-v2/errors_rum.ndjson")
+	result := systemtest.Elasticsearch.ExpectDocs(t, "logs-apm.error-*", nil)
+	assertSourcemapUpdated(t, result, true)
 }
 
 func TestSourcemapCaching(t *testing.T) {


internal/sourcemap/elasticsearch.go
```diff
@@ -59,7 +59,12 @@ type esSourcemapResponse struct {
 		Hits []struct {
 			Source struct {
 				Sourcemap struct {
```
kruskall, line 61: **TODO**: 

> Once a sourcemap is cached in the LRU cache it only gets refreshed if the content-hash for the sourcemap changes.

kruskall, line 61: > Align with the UI team on the document structure and whether the current sourcemap index and structure can be reused, and only the field for the content hash would be added.

**issue**: I think we should reach out now that we have laid down the impl on our side.

internal/beater/beater.go
```diff
@@ -827,26 +821,49 @@ func newSourcemapFetcher(
...
 		return nil, err
 	}
 	index := strings.ReplaceAll(cfg.IndexPattern, "%{[observer.version]}", version.Version)
 	esFetcher := sourcemap.NewElasticsearchFetcher(esClient, index)
-	chained = append(chained, esFetcher)
+	cachingFetcher, err := sourcemap.NewCachingFetcher(esFetcher, cfg.Cache.Expiration)
+	if err != nil {
+		return nil, err
+	}
+	metadataCachingFetcher := sourcemap.NewMetadataCachingFetcher(esClient, cachingFetcher, index)
```
kruskall, line 862: **question**: 

> A configuration section (similar to the [config for agent config](https://github.com/elastic/apm-server/pull/8193/files#diff-fbadce52f216b92bda4dab62b076f334502b170af5e51bf3e442e2033b4a4e9dR201-R223)) needs to be added for fetching from Elasticsearch. Querying these ES indices will require additional privileges.

I'm not sure I get this part. The ES fetcher is using the configured es client and reusing the old indices. The metadata fetcher is sharing the same client and querying the same index.
Why does it need additional privileges ? 
If it needs additional privileges does that mean that the current ES fetcher is not working ?

internal/sourcemap/metadata.go
```diff
@@ -0,0 +1,206 @@
...
+	return nil
+}
+
+func (s *MetadataCachingFetcher) runSearchQuery(ctx context.Context) (*esapi.Response, error) {
+	var buf bytes.Buffer
+	if err := json.NewEncoder(&buf).Encode(queryMetadata()); err != nil {
+		return nil, err
+	}
+
+	req := esapi.SearchRequest{
```
carsonip, line 176: Is 1 search request enough? Do we need scroll search instead?

internal/sourcemap/elasticsearch.go
kruskall: Fixed!

kruskall: See https://github.com/elastic/kibana/pull/147208

kruskall:
How to test:

- checkout https://github.com/elastic/kibana/pull/147208 and build a kibana docker image with it
- Replace the current kibana docker image with the custom one at https://github.com/elastic/apm-server/blob/main/docker-compose.yml#L34
- Bump elasticsearch docker image at https://github.com/elastic/apm-server/blob/main/docker-compose.yml#L4 so that it includes https://github.com/elastic/elasticsearch/pull/92160
- Change the ES user to admin at https://github.com/elastic/apm-server/blob/main/systemtest/apmservertest/config.go#L34

internal/sourcemap/elasticsearch.go
```diff
@@ -97,12 +100,33 @@ func (s *esFetcher) Fetch(ctx context.Context, name, version, path string) (*sou
...
+		return nil, fmt.Errorf("failed to base64 decode string: %w", err)
+	}
+
+	r, err := zlib.NewReader(bytes.NewReader(d))
+	if err != nil {
+		return nil, fmt.Errorf("failed to create zlib reader: %w", err)
+	}
+	defer r.Close()
+
+	bbb, err := io.ReadAll(r)
```
simitt, line 119: What is `bbb`? ;)

internal/sourcemap/metadata.go
```diff
@@ -0,0 +1,254 @@
...
+		logger:           logp.NewLogger(logs.Sourcemap),
+		invalidationChan: invalidationChan,
+	}
+}
+
+func (s *MetadataCachingFetcher) Fetch(ctx context.Context, name, version, path string) (*sourcemap.Consumer, error) {
+	if len(s.set) == 0 {
+		// If cache is empty and we're trying to fetch a sourcemap
+		// make sure the initial cache population has ended.
+		s.once.Do(func() {
```
simitt, line 71: What happens if ES is not yet reachable at this point but later?

internal/sourcemap/metadata.go
```diff
@@ -0,0 +1,254 @@
...
+		// invalidate the sourcemap cache.
+		s.invalidationChan <- k
+
+		// remove from metadata cache
+		delete(s.set, k)
+	}
+
+	for _, m := range ms {
+		s.set[m.id] = m.contentHash
+	}
```
simitt, line 141: This logic seems a bit inefficient. 

Two ideas: 
(1) In the `sync` method, make `ms` a `map` with the key being `id`. Then iterate over `ms` once, check if the `id` is in the `ms` map, and act accordingly. Remove the items from the map an add the remaining map items afterwards. You save an iteration and a lot of writes to the cache. 

(2) You could use a cache with a time based eviction, that is set to something between `metadataFetcherInterval < cacheTimeout < 2*metadataFetcherInterval`. On eviction, the sourcemap key is added to the `s.invalidationChan`. On every update, the time is updated for the valid sourcemap keys.

simitt (commented):
Generally looks good, left some comment related to updating the cache. Please also update the description as the fallback logic is now removed. 

We should also add metrics for better insights into the sourcemap handling: at least track the number of requests to ES for sourcemaps, track sucess of metadata requests.

internal/sourcemap/metadata.go
simitt: I believe option 1 is simpler and good enough.

internal/beater/beater.go
kruskall: I've added a section to the agent config to inject the api key based on the open kibana PR

internal/sourcemap/elasticsearch.go
kruskall: ~A result of the grammar { S -> b, S -> Sb} used to name variables~Thanks spotting this! I've updated the variables in this PR so use more meaningful names :+1:

internal/sourcemap/metadata.go
kruskall: Looking back at this, I don't think it's a good idea, this will basically block until the cache is populated since we cannot retrieve sourcemaps until then with the current implementation.

I'm not sure what's the best way to proceed here. Returning nothing might be misleading. Should we bypass the cache and send the request to ES if the cache is not populated yet ?

internal/sourcemap/metadata.go
kruskall: Thanks for spotting this! I've pushed some changes :bow:

internal/sourcemap/metadata.go
simitt: This would be if the metadata cache is not ready. If a sourcemap is part of the metadata cache but the body is not yet  cached, it will be fetched from ES on request. 
So let's take a look at the metadata cache. What are the scenarios for when the cache might not be ready?
- ES is not reachable, due to a restart, misconfiguration, etc.: in this case, trying to fetch directly from ES would not succeed either. 
- Fetching metadata from ES is ongoing. The fetching and cache update for metadata can be expected to be fairly quick, even when many sourcemaps are stored. Bypassing the metadata cache would work here, but seems like a hack. I'd rather either block the request processing (and collect logs & metrics so we get an understanding how long this blocking period is) or treat it as _no sourcemap is available_ (and again, collect logs & metrics for the scenario). The third option would be to block the APM Server listeners until everything is readily set up, but this would definitely lead to apm event loss, which is definitely worse.

apm-server.docker.yml
```diff
@@ -198,9 +198,13 @@ apm-server:
...
-  # When using APM agent configuration, information fetched from Kibana will be cached in memory for some time.
-  # Specify cache key expiration via this setting. Default is 30 seconds.
-  #agent.config.cache.expiration: 30s
+  #agent:
+    # When using APM agent configuration, information fetched from Kibana will be cached in memory for some time.
+    # Specify cache key expiration via this setting. Default is 30 seconds.
+    #config.cache.expiration: 30s
+
+    #elasticsearch:
+      #api_key: "id:api_key"
```
simitt, line 207: Why is code from the `agentConfig` bleeding into the sourcemapping logic? There is already a `source_mapping.elasticsearch` section - shouldn't that be used?

internal/sourcemap/caching.go
```diff
@@ -61,13 +61,13 @@ func NewCachingFetcher(
 	}
 
 	go func() {
-		for i := range invalidationChan {
-			c.Remove(i)
+		for identifier := range invalidationChan {
+			lruCache.Remove(identifier)
```
simitt, line 65: I believe the variable name changes were triggered by my comment about the `bbb` variable. My mistake, I didn't intend to make you change all variable names to something meaningful. It is perfectly fine to use something like `i` in a loop. It is also fine to use `c` for a `cache`, if the scope of the variable is a couple of lines within the same method. 

Using `b`, `d` and `bbb` for `body`, `decodedBody` and `uncompressedBody` was already a bit harder to follow. IMO there it makes sense to be a bit more descriptive.

internal/sourcemap/metadata.go
```diff
@@ -110,34 +110,24 @@ func maybeParseURLPath(s string) string {
...
 
-		// remove from metadata cache
-		delete(s.set, k)
-	}
-
-	for _, m := range ms {
-		s.set[m.id] = m.contentHash
+			// remove from metadata cache
+			delete(s.set, id)
+		}
```
simitt, line 119: I am not following where the values are actually updated in the set now.

It might need to be something like this:
```
	for id, contentHash := range s.set {
		if metadata, ok := ms[id]; ok {
			// content hash changed, invalidate the sourcemap cache
			if contentHash != metadata.contentHash {
				s.invalidationChan <- id
				// remove from metadata cache
				delete(s.set, id)

			} else {
				// remove entry from map if already part of the cache set and the content is unchanged
				delete(ms, id)
			}
		} else {
			// the sourcemap no longer exists in ES.
			// invalidate the sourcemap cache.
			s.invalidationChan <- id

			// remove from metadata cache
			delete(s.set, id)
		}
	}
	for id, m := range ms {
		s.set[id] = m.contentHash
	}
```

internal/sourcemap/metadata.go
```diff
@@ -196,7 +186,7 @@ func (s *MetadataCachingFetcher) sync(ctx context.Context) error {
 		return err
 	}
 
-	var ms []Metadata
+	ms := make(map[Identifier]Metadata, len(body.Hits.Hits))
```
simitt, line 176: What about making it `map[Identifier]string`?

simitt (commented):
Left a couple more comments, but the PR is already very close.

apm-server.docker.yml
kruskall: This is what kibana is using: https://github.com/elastic/kibana/pull/147650/files#diff-a94aa3619e34f92c6719ca26e5efa03eb62eb2a077bb98a87cf0c7c9e751e7f8R17

Unless I misunderstood that PR, we need to match the path they are injecting to.

internal/sourcemap/caching.go
kruskall: No worries, I changed `i` because it is usually used for int/index so it wasn't obvious to me it was referring to identifiers.
For cache, I was worried `c` might be confused with the `CachingFetcher` so I made it explicit.

I don't have a strong opinion about it, let me know if I should revert the change :bow:

internal/sourcemap/metadata.go
kruskall: My mistake, Thanks for spotting this! I've pushed some changes!

internal/sourcemap/metadata.go
kruskall: Good point! Updated :+1:

internal/sourcemap/metadata.go
kruskall: > ES is not reachable, due to a restart, misconfiguration, etc.: in this case, trying to fetch directly from ES would not succeed either.

I don't think the metadata cache should be concerned If ES is not ready/reachable/available. A failure in that scenario would be "working as intended" imo because the metadata cache is proxying what it got from the ES fetcher (error, timeout, etc.).

> Fetching metadata from ES is ongoing. The fetching and cache update for metadata can be expected to be fairly quick, even when many sourcemaps are stored. Bypassing the metadata cache would work here, but seems like a hack. I'd rather either block the request processing (and collect logs & metrics so we get an understanding how long this blocking period is) or treat it as no sourcemap is available (and again, collect logs & metrics for the scenario). The third option would be to block the APM Server listeners until everything is readily set up, but this would definitely lead to apm event loss, which is definitely worse.

I was thinking of "if init is not completed, forward the request to the backend fetcher". In theory the first request for each sourcemap would be forwarded to ES anyway so this wouldn't create additional work unless we get a request for a missing sourcemap on ES during init. The reasoning behind this was that when the metadata cache is not initialized:
- if the sourcemap exists on ES: the metadata fetcher will send the request to the es fetcher anyway. No additional time/work/blocking: this would be the same as the metadata cache receiving a request for the first time. 
- if the sourcemap does not exist on ES: the metadata will send the request to the ES fetcher, creating additional work. The additional processing time is offset by the fact we would have had to wait for the init phase to finish if we had blocked so my assumption was that there wasn't gonna be a time difference between the two in the end.

apm-server.docker.yml
simitt: Ideally they would inject it at the agentcfg and the sourcemap level. Please check with them how much effort it is to change that.

internal/sourcemap/caching.go
simitt: no need to revert anything, thanks for the changes.

apm-server.docker.yml
kruskall: Left a comment on the kibana PR and pinged you for visibility :+1:

internal/sourcemap/metadata.go
```diff
@@ -0,0 +1,235 @@
...
+
+func (s *MetadataCachingFetcher) sync(ctx context.Context) error {
+	resp, err := s.runSearchQuery(ctx)
+	if err != nil {
+		return errors.Wrap(err, errMsgESFailure)
+	}
+	defer resp.Body.Close()
+
+	// handle error response
+	if resp.StatusCode >= http.StatusMultipleChoices {
```
carsonip, line 166: Why is it >=300 instead of >=400?

internal/beater/beater.go
```diff
@@ -808,78 +800,32 @@ func (s *Runner) newLibbeatFinalBatchProcessor(
...
+	cachingFetcher, err := sourcemap.NewCachingFetcher(esFetcher, invalidationChan, size)
 	if err != nil {
 		return nil, err
 	}
-	index := strings.ReplaceAll(cfg.IndexPattern, "%{[observer.version]}", version.Version)
-	esFetcher := sourcemap.NewElasticsearchFetcher(esClient, index)
-	chained = append(chained, esFetcher)
-	return chained, nil
+	metadataCachingFetcher := sourcemap.NewMetadataCachingFetcher(esClient, cachingFetcher, apmSourcemapIndex, invalidationChan)
+	metadataCachingFetcher.StartBackgroundSync()
```
carsonip, line 826: Tests call `newSourcemapFetcher` and `newSourcemapFetcher` starts goroutines, meaning that the tests will leak the background sync goroutine even after they finish. Is that how it works? In my PR I return a function for the caller (beater) to call instead.

internal/sourcemap/metadata.go
simitt: So this is really used for triggering the first fetch, in case it hasn't yet been initiated?

simitt:
@kruskall I fear we cannot remove the Kibana fetcher, as we would introduce a breaking change for scenarios where the standalone APM Server is configured to output to something else than Elasticsearch. In such cases, when Kibana is configured, sourcemaps are fetched from Kibana, and user interaction would be required on upgrade for the sourcemap logic to keep working. 
For standalone server that was using ES configuration for fetching sourcemaps, the previous configuration should still succeed in fetching sourcemaps, since they are still stored in the same index and therefore do not require additional privileges. 

It is fine to remove the Fleet fetcher logic, as this is replaced with the new Elasticsearch fetcher logic, and the APM Integration does only support ES output. 

So, in summary, please re-add the Kibana fetcher and ping the team when this is ready for a final review.

apm-server.docker.yml
kruskall: Changed!

internal/sourcemap/metadata.go
```diff
@@ -0,0 +1,292 @@
...
+}
+
+func (s *MetadataCachingFetcher) StartBackgroundSync() {
+	go func() {
+		// First run, populate cache
+		ctx, cleanup := context.WithTimeout(context.Background(), 10*time.Second)
+		defer cleanup()
+
+		if err := s.sync(ctx); err != nil {
+			s.logger.Error("failed to fetch sourcemaps metadata: %v", err)
```
carsonip, line 191:
```suggestion
			s.logger.Errorf("failed to fetch sourcemaps metadata: %v", err)
```

internal/sourcemap/metadata.go
```diff
@@ -0,0 +1,292 @@
...
+	go func() {
+		// TODO make this a config option ?
+		t := time.NewTicker(30 * time.Second)
+		defer t.Stop()
+
+		for range t.C {
+			ctx, cleanup := context.WithTimeout(context.Background(), 10*time.Second)
+
+			if err := s.sync(ctx); err != nil {
+				s.logger.Error("failed to sync sourcemaps metadata: %v", err)
```
carsonip, line 207:
```suggestion
				s.logger.Errorf("failed to sync sourcemaps metadata: %v", err)
```

internal/sourcemap/metadata.go
kruskall: The logic has been reworked and now we forward the request to ES if the metadata cache is not ready. 

This does not create additional costs since the response is cached in the lru cache.

internal/sourcemap/metadata.go
kruskall: That's what the current elasticsearch fetcher is using: https://github.com/elastic/apm-server/blob/da18d54ecdb3f442d973007ebe2ebdd33692a143/internal/sourcemap/elasticsearch.go#L84

internal/beater/beater.go
kruskall: Good catch! Updated :+1:

internal/sourcemap/metadata.go
kruskall: It is not, thanks for spotting this! Updated :+1:

kruskall:
> @kruskall I fear we cannot remove the Kibana fetcher, as we would introduce a breaking change for scenarios where the standalone APM Server is configured to output to something else than Elasticsearch. In such cases, when Kibana is configured, sourcemaps are fetched from Kibana, and user interaction would be required on upgrade for the sourcemap logic to keep working.
> For standalone server that was using ES configuration for fetching sourcemaps, the previous configuration should still succeed in fetching sourcemaps, since they are still stored in the same index and therefore do not require additional privileges.
> 
> It is fine to remove the Fleet fetcher logic, as this is replaced with the new Elasticsearch fetcher logic, and the APM Integration does only support ES output.
> 
> So, in summary, please re-add the Kibana fetcher and ping the team when this is ready for a final review.

Ah I see, thanks for mentioning this! I've added back the kibana fetcher as fallback and a systemtest to make sure it is working as intended.

internal/sourcemap/elasticsearch.go
```diff
@@ -58,11 +59,17 @@ type esSourcemapResponse struct {
 		}
```
carsonip, line 59: Any struct tags needed for Total and Value?

internal/beater/config/rum.go
```diff
@@ -72,20 +74,35 @@ func (c *RumConfig) setup(log *logp.Logger, outputESCfg *config.C) error {
 		return errors.Wrapf(err, "Invalid regex for `exclude_from_grouping`: ")
 	}
 
-	// No need to unpack the ESConfig if SourceMapMetadata exist
 	if len(c.SourceMapping.Metadata) > 0 {
-		return nil
+		// We don't have the fleet fetcher anymore.
+		// Ignore metadata and setup the elasticsearch config.
+		log.Warn("Ignoring sourcemap metadata")
```
simitt, line 80: I'd completely remove `SourceMapping.Metadata`. Afaics this is not used anywhere anymore and was only used with the FleetFetcher, so it should never reach this code.

internal/sourcemap/metadata.go
```diff
@@ -0,0 +1,409 @@
...
+		set:              make(map[Identifier]string),
+		alias:            make(map[Identifier]*Identifier),
+		backend:          backend,
+		logger:           logp.NewLogger(logs.Sourcemap),
+		init:             make(chan struct{}),
+		invalidationChan: invalidationChan,
+	}
+}
+
+func (s *MetadataCachingFetcher) Fetch(ctx context.Context, name, version, path string) (*sourcemap.Consumer, error) {
```
simitt, line 70: The `MetadataCachingFetcher` should always only run as a background job, fetching metadata and populating a cache. Whereas the `Cache` and `Elasticsearch` Fetcher are responsible for synchr. fetching soucemap bodies from either the cache or ES. The `Fetch` method is called sync when processing incoming requests. Why exposing this `Fetch` from the `MetadataCachingFetcher`? This causes quite a strong coupling between the fetchers, especially the `MetadataCachingFetcher` doesn't need to know anything about the sourcemap body fetcher. It only needs to expose a way to access the cache with the metadata. 
I guess I had rather expected it the other way around, that the `MetadataCachingFetcher` would expose a public method `GetID` and the fetcher would be passed as a parameter to the `ElasticsearchFetcher`. 


Nit: How about also renaming the `ElasticsearchFetcher` to something indicating what it fetches from ES, as the `MetadataCachingFetcher` also fetches from ES.

internal/sourcemap/metadata.go
```diff
@@ -0,0 +1,409 @@
...
+		if cleanPath != path {
+			s.logger.Debugf("original filepath %s converted to %s", path, cleanPath)
+			return s.Fetch(ctx, name, version, cleanPath)
+		}
+	}
+
+	return nil, nil
+}
+
+func (s *MetadataCachingFetcher) hasID(key Identifier) bool {
```
simitt, line 136: How about abstracting the concept of `aliases` to the caller and changing this to `func (s *MetadataCachingFetcher) getID(key Identifier) string`. If the key is not presented, return an empty string, otherwise return the ID, even if the `key` is an alias. This way the caller would always only have to make one call and not be concerned with these internas.

internal/sourcemap/caching_test.go
```diff
@@ -164,176 +159,9 @@ func TestStore_Fetch(t *testing.T) {
...
-	mapper, err := store.Fetch(context.Background(), name, version, path)
-	require.NoError(t, err)
-	assert.Equal(t, &sourcemap.Consumer{}, mapper)
-
-	time.Sleep(25 * time.Millisecond)
-	// cache is cleared, sourcemap is fetched from ES leading to an error
-	mapper, err = store.Fetch(context.Background(), name, version, path)
-	require.Error(t, err)
-	assert.Nil(t, mapper)
-}
```
simitt, line 313: Why did you remove this test?

systemtest/approvals/TestRUMRoutingIntegration.approved.json
```diff
@@ -447,9 +447,6 @@
                         "line": {
                             "column": 3,
                             "number": 7666
-                        },
-                        "sourcemap": {
-                            "error": "unable to find sourcemap.url for service.name=apm-a-rum-test-e2e-general-usecase service.version=0.0.1 bundle.path=http://localhost:8000/test/e2e/general-usecase/app.e2e-bundle.min.js?token=secret"
```
simitt, line 452: Why did this change?

internal/sourcemap/metadata.go
```diff
@@ -0,0 +1,409 @@
...
+	"github.com/go-sourcemap/sourcemap"
+	"github.com/pkg/errors"
+
+	"github.com/elastic/apm-server/internal/elasticsearch"
+	"github.com/elastic/apm-server/internal/logs"
+	"github.com/elastic/elastic-agent-libs/logp"
+	"github.com/elastic/go-elasticsearch/v8/esapi"
+)
+
+type MetadataCachingFetcher struct {
```
simitt, line 40: Are there any tests for the `MetadataCachingFetcher` logic? I was expecting a fair amount of unit tests covering the cache update logic etc. 
I might be missing something here, but with the escape hatch to fetch the body directly from ES, if the metadata cache is not ready, the system tests might still be running as expected even if the metadata cache is not populated.

simitt (changes requested):

internal/beater/config/rum.go
```diff
@@ -34,7 +35,7 @@ const (
 	defaultExcludeFromGrouping      = "^/webpack"
 	defaultLibraryPattern           = "node_modules|bower_components|~"
 	defaultSourcemapCacheExpiration = 5 * time.Minute
-	defaultSourcemapIndexPattern    = "apm-*-sourcemap*"
+	defaultSourcemapIndexPattern    = ".apm-source-map"
```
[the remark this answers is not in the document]
kruskall: That's correct. Both the index and the way sourcemaps are stored in ES changed. This happened in the kibana PR (https://github.com/elastic/kibana/pull/147208/files#diff-a968c157df4cd16bb68294e0741431359aba6574abbed2020b7ca10e4f592159R28), it is necessary and we are just defaulting to what they use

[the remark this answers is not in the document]
simitt: Totally agree that as long as the Kibana side changes this, we have to adapt, but I'd like to question whether the changes in the Kibana UI code are necessary. @kruskall and I discussed that he will follow up with the UI folks.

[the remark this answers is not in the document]
sqren: I was of the impression that `apm-*-sourcemap*` was no longer used, and the ability to configure the index was removed to simplify things:
https://github.com/elastic/kibana/pull/147208/commits/4050578be14da202662bd863b867ef0b33b7f366

My concern was that if the end user could configure the source map index, the internal kibana user might not have permissions to query it. With a hardcoded index (`.apm-source-map`) I was sure this was not the case.

I can easily change this back to `apm-{version}-sourcemap`. It won't be user configurable but it'll match the old pattern.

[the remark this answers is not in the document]
sqren: Btw. I suppose we still need to create the index for recent versions of the stack. Or does APM Server still create the index at startup?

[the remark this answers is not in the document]
simitt: Yes Kibana will still need to create the index as it was previously done; at least for the upgrade order (Kibana before APM Server).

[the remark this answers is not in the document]
simitt: >My concern was that if the end user could configure the source map index, the internal kibana user might not have permissions to query it. With a hardcoded index (.apm-source-map) I was sure this was not the case.

Understood, and this is a valid concern.@kruskall could you check how the sourcemap index is currently setup (before your changes) when the server runs in standalone mode and Kibana is not enabled?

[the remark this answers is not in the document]
sqren: Would this solve the problem on your end?
https://github.com/elastic/kibana/pull/149403

Btw I changed the index to `apm-agent-sourcemap` so it still matches `apm-*-sourcemap*`

internal/sourcemap/metadata.go
```diff
@@ -0,0 +1,409 @@
...
+func (s *MetadataCachingFetcher) runScrollSearchQuery(ctx context.Context, id string) (*esapi.Response, error) {
+	var buf bytes.Buffer
+	if err := json.NewEncoder(&buf).Encode(scrollQueryMetadata(id)); err != nil {
+		return nil, err
+	}
+
+	req := esapi.ScrollRequest{
+		ScrollID: id,
+		Scroll:   time.Minute,
+		Body:     &buf,
```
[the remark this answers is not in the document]
kruskall: Good catch! Updated :+1:

internal/sourcemap/metadata.go
```diff
@@ -0,0 +1,409 @@
...
+		// requests, it doesn’t always change — in any case, only the most recently
+		// received _scroll_id should be used.
+		if id != "" {
+			scrollID = id
+		}
+
+		// Stop if there are no new updates
+		if before == len(updates) {
+			break
+		}
```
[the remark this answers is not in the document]
kruskall: I like that! Changed it!

internal/beater/beater.go
```diff
@@ -823,73 +813,37 @@ func newSourcemapFetcher(
...
-				return nil, err
-			}
-			u, err := url.Parse(urlString)
-			if err != nil {
-				return nil, err
-			}
-			fleetServerURLs[i] = u
-		}
+	size := 128
+	invalidationChan := make(chan sourcemap.Identifier, size)
```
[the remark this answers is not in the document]
kruskall: Good catch! Fixed :+1:

internal/sourcemap/elasticsearch.go
kruskall: Added :+1:

internal/beater/beater.go
```diff
@@ -818,78 +806,31 @@ func (s *Runner) newLibbeatFinalBatchProcessor(
 	return publisher, stop, nil
 }
 
+const apmSourcemapIndex = ".apm-source-map"
+
 func newSourcemapFetcher(
 	cfg config.SourceMapping,
 	fleetCfg *config.Fleet,
```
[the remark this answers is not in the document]
kruskall: Good catch!

internal/beater/beater.go
```diff
@@ -330,20 +325,15 @@ func (s *Runner) Run(ctx context.Context) error {
 
 	var sourcemapFetcher sourcemap.Fetcher
 	if s.config.RumConfig.Enabled && s.config.RumConfig.SourceMapping.Enabled {
-		fetcher, err := newSourcemapFetcher(
+		fetcher, cleanup, err := newSourcemapFetcher(
```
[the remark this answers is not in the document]
kruskall: Done :+1:

internal/beater/config/rum.go
kruskall: It is not used in the APM Server but I think kibana (?) might be injecting this option. 

For example, in  `TestRUMErrorSourcemapping` this code path is triggered and we're logging a warning.

I agree we can remove it here :+1:

systemtest/approvals/TestRUMRoutingIntegration.approved.json
kruskall: Should be fixed!

internal/sourcemap/caching_test.go
kruskall: The BodyCachingFetcher is using a LRU cache now so time-based expiration no longer applies. I'll add other tests to replace the old ones and make sure the new logic is working.

internal/sourcemap/metadata.go
kruskall: I've moved the logic for syncing and sending updates to a separate file and decoupled the fetchers.

Only the sync worker and the Elasticsearch fetchers are coupled/knows about ES. The fetchers have been renamed for clarity and they are decoupled from each other.
I'm not sure I fully understand where you are heading, do you want to merge Metadata and Body Caching fetchers ?

internal/sourcemap/metadata.go
kruskall: Done! :+1:

kruskall:
Updated with the changes discusses in the previous sync

axw (commented):
Just a small drive-by comment, I haven't reviewed much of the code - just noticed that we're still using the index pattern. IIANM that index is not configurable on the Kibana side, so I think we should just remove the config from apm-server.yml & hard-code the index name in the code.

internal/sourcemap/sourcemap_fetcher.go
```diff
@@ -0,0 +1,125 @@
...
+			return s.fetch(ctx, i)
+		}
+	default:
+		s.logger.Debugf("Metadata cache not populated. Falling back to backend fetcher for id: %s, %s, %s", name, version, path)
+		// init is in progress, ignore the metadata cache and fetch the sourcemap directly
+		// return if we get a valid sourcemap or an error
+		if c, err := s.backend.Fetch(ctx, original.name, original.version, original.path); c != nil || err != nil {
+			return c, err
+		}
+
```
simitt, line 65: I am missing why the following part is necessary. If the metadata fetcher is not done, `s.backend.Fetch` is triggered. If `c == nil && err == nil` it means that the sourcemap doesn't exist. What is the reason for blocking and waiting until metadata is ready etc?  

Especially if ES is just not reachable, this might lead to blocking any request.

internal/sourcemap/metadata_fetcher.go
```diff
@@ -0,0 +1,320 @@
...
+		if len(result.Hits.Hits) == 0 {
+			break
+		}
+	}
+
+	s.update(ctx, sourcemaps)
+	return nil
+}
+
+func (s *MetadataESFetcher) update(ctx context.Context, sopurcemaps map[Identifier]string) {
```
simitt, line 171:
```suggestion
func (s *MetadataESFetcher) update(ctx context.Context, sourcemaps map[Identifier]string) {
```

internal/sourcemap/metadata_fetcher.go
```diff
@@ -0,0 +1,320 @@
...
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	var invalidation []Identifier
+
+	for id, contentHash := range s.set {
+		if updatedHash, ok := sopurcemaps[id]; ok {
+			// already in the cache, remove from the updates.
+			delete(sopurcemaps, id)
```
simitt, line 180: Since the ID is deleted from `sourcemaps` here, if `contenthash!=updatedHash`, the `s.set[id]=updatedHash` is missing.

internal/sourcemap/metadata_fetcher.go
```diff
@@ -0,0 +1,320 @@
...
+	// handle error response
+	if resp.StatusCode >= http.StatusMultipleChoices {
+		b, err := io.ReadAll(resp.Body)
+		if err != nil {
+			return nil, fmt.Errorf("failed to read response body: %w", err)
+		}
+		if resp.StatusCode == http.StatusNotFound || resp.StatusCode == http.StatusForbidden {
+			return nil, fmt.Errorf("%w: %s: %s", ErrFetcherUnvailable, resp.Status(), string(b))
+		}
+		return nil, fmt.Errorf("ES returned unknown status code: %s", resp.Status())
```
simitt, line 268: We should also explicitly handle if ES is not yet reachable.

internal/sourcemap/sourcemap_fetcher.go
```diff
@@ -0,0 +1,125 @@
...
+			// Only fetch from ES if the sourcemap id exists
+			return s.fetch(ctx, i)
+		}
+	default:
+		s.logger.Debugf("Metadata cache not populated. Falling back to backend fetcher for id: %s, %s, %s", name, version, path)
+		// init is in progress, ignore the metadata cache and fetch the sourcemap directly
+		// return if we get a valid sourcemap or an error
+		if c, err := s.backend.Fetch(ctx, original.name, original.version, original.path); c != nil || err != nil {
+			return c, err
+		}
```
simitt, line 64: I'm not 100% convinced about this fallback. Running your code, if an unreachable Elasticsearch is configured, this is what is logged:
```
{"log.level":"debug","@timestamp":"2023-02-02T08:35:02.675+0100","log.logger":"sourcemap","log.origin":{"file.name":"sourcemap/sourcemap_fetcher.go","file.line":59},"message":"Metadata cache not populated. Falling back to backend fetcher for id: apm-agent-js, 1.0.1, http://localhost:8000/test/e2e/general-usecase/bundle.js.map","service.name":"apm-server","ecs.version":"1.6.0"}
{"log.level":"error","@timestamp":"2023-02-02T08:35:03.229+0100","log.logger":"beater","log.origin":{"file.name":"beater/waitready.go","file.line":64},"message":"precondition failed: dial tcp [::1]:9200: connect: connection refused","service.name":"apm-server","ecs.version":"1.6.0"}
{"log.level":"debug","@timestamp":"2023-02-02T08:35:07.676+0100","log.logger":"stacktrace","log.origin":{"file.name":"sourcemap/processor.go","file.line":124},"message":"failed to fetch sourcemap with path (http://localhost:8000/test/e2e/general-usecase/bundle.js.map): failure querying ES: context deadline exceeded","service.name":"apm-server","ecs.version":"1.6.0"}
{"log.level":"debug","@timestamp":"2023-02-02T08:35:07.676+0100","log.logger":"sourcemap","log.origin":{"file.name":"sourcemap/sourcemap_fetcher.go","file.line":59},"message":"Metadata cache not populated. Falling back to backend fetcher for id: apm-agent-js, 1.0.1, http://localhost:8000/test/e2e/general-usecase/bundle.js.map","service.name":"apm-server","ecs.version":"1.6.0"}
{"log.level":"debug","@timestamp":"2023-02-02T08:35:07.676+0100","log.logger":"stacktrace","log.origin":{"file.name":"sourcemap/processor.go","file.line":124},"message":"failed to fetch sourcemap with path (http://localhost:8000/test/e2e/general-usecase/bundle.js.map): failure querying ES: context deadline exceeded","service.name":"apm-server","ecs.version":"1.6.0"}
{"log.level":"debug","@timestamp":"2023-02-02T08:35:07.676+0100","log.logger":"sourcemap","log.origin":{"file.name":"sourcemap/sourcemap_fetcher.go","file.line":59},"message":"Metadata cache not populated. Falling back to backend fetcher for id: apm-agent-js, 1.0.1, http://localhost:8000/test/e2e/general-usecase/bundle.js.map","service.name":"apm-server","ecs.version":"1.6.0"}
{"log.level":"debug","@timestamp":"2023-02-02T08:35:07.676+0100","log.logger":"stacktrace","log.origin":{"file.name":"sourcemap/processor.go","file.line":124},"message":"failed to fetch sourcemap with path (http://localhost:8000/test/e2e/general-usecase/bundle.js.map): failure querying ES: context deadline exceeded","service.name":"apm-server","ecs.version":"1.6.0"}
{"log.level":"debug","@timestamp":"2023-02-02T08:35:07.676+0100","log.logger":"sourcemap","log.origin":{"file.name":"sourcemap/sourcemap_fetcher.go","file.line":59},"message":"Metadata cache not populated. Falling back to backend fetcher for id: apm-agent-js, 1.0.1, http://localhost:8000/test/e2e/general-usecase/bundle.js.map","service.name":"apm-server","ecs.version":"1.6.0"}
{"log.level":"debug","@timestamp":"2023-02-02T08:35:07.676+0100","log.logger":"stacktrace","log.origin":{"file.name":"sourcemap/processor.go","file.line":124},"message":"failed to fetch sourcemap with path (http://localhost:8000/test/e2e/general-usecase/bundle.js.map): failure querying ES: context deadline exceeded","service.name":"apm-server","ecs.version":"1.6.0"}
{"log.level":"debug","@timestamp":"2023-02-02T08:35:07.676+0100","log.logger":"sourcemap","log.origin":{"file.name":"sourcemap/sourcemap_fetcher.go","file.line":59},"message":"Metadata cache not populated. Falling back to backend fetcher for id: apm-agent-js, 1.0.1, http://localhost:8000/test/e2e/general-usecase/bundle.js.map","service.name":"apm-server","ecs.version":"1.6.0"}
{"log.level":"debug","@timestamp":"2023-02-02T08:35:07.676+0100","log.logger":"stacktrace","log.origin":{"file.name":"sourcemap/processor.go","file.line":124},"message":"failed to fetch sourcemap with path (http://localhost:8000/test/e2e/general-usecase/bundle.js.map): failure querying ES: context deadline exceeded","service.name":"apm-server","ecs.version":"1.6.0"}
{"log.level":"debug","@timestamp":"2023-02-02T08:35:07.676+0100","log.logger":"sourcemap","log.origin":{"file.name":"sourcemap/sourcemap_fetcher.go","file.line":59},"message":"Metadata cache not populated. Falling back to backend fetcher for id: apm-agent-js, 1.0.1, http://localhost:8000/test/e2e/general-usecase/bundle.js.map","service.name":"apm-server","ecs.version":"1.6.0"}
{"log.level":"debug","@timestamp":"2023-02-02T08:35:07.676+0100","log.logger":"stacktrace","log.origin":{"file.name":"sourcemap/processor.go","file.line":124},"message":"failed to fetch sourcemap with path (http://localhost:8000/test/e2e/general-usecase/bundle.js.map): failure querying ES: context deadline exceeded","service.name":"apm-server","ecs.version":"1.6.0"}
{"log.level":"info","@timestamp":"2023-02-02T08:35:07.678+0100","log.logger":"request","log.origin":{"file.name":"middleware/log_middleware.go","file.line":61},"message":"request accepted","service.name":"apm-server","url.original":"/intake/v2/rum/events","http.request.method":"POST","user_agent.original":"curl/7.79.1","source.address":"127.0.0.1","http.request.id":"9dab8633-386a-47c6-b431-171ba122b9b9","event.duration":5003973201,"http.response.status_code":202,"ecs.version":"1.6.0"}
{"log.level":"error","@timestamp":"2023-02-02T08:35:13.223+0100","log.logger":"agentcfg","log.origin":{"file.name":"agentcfg/elasticsearch.go","file.line":134},"message":"refresh cache error: context deadline exceeded","service.name":"apm-server","ecs.version":"1.6.0"}
{"log.level":"error","@timestamp":"2023-02-02T08:35:18.234+0100","log.logger":"beater","log.origin":{"file.name":"beater/waitready.go","file.line":64},"message":"precondition failed: dial tcp [::1]:9200: connect: connection refused","service.name":"apm-server","ecs.version":"1.6.0"}
{"log.level":"error","@timestamp":"2023-02-02T08:35:23.221+0100","log.logger":"sourcemap","log.origin":{"file.name":"sourcemap/metadata_fetcher.go","file.line":117},"message":"failed to sync sourcemaps metadata: failed to run initial search query: context deadline exceeded","service.name":"apm-server","ecs.version":"1.6.0"}
```

The fallback only makes sense if ES is reachable & accessible, but the background sync job is not yet finished.

internal/sourcemap/metadata_fetcher.go
```diff
@@ -0,0 +1,320 @@
...
+
+		if err := s.sync(ctx); err != nil {
+			s.logger.Errorf("failed to fetch sourcemaps metadata: %v", err)
+		} else {
+			// only close the init chan and mark the fetcher as ready if
+			// sync succeeded
+			close(s.init)
+		}
+
+		s.logger.Info("init routine completed")
```
simitt, line 104: This should either be moved into the `else` close, or rephrased. Right now it leads to
```
{"log.level":"error","@timestamp":"2023-02-02T08:34:43.220+0100","log.logger":"sourcemap","log.origin":{"file.name":"sourcemap/metadata_fetcher.go","file.line":97},"message":"failed to fetch sourcemaps metadata: failed to run initial search query: context deadline exceeded","service.name":"apm-server","ecs.version":"1.6.0"}
{"log.level":"info","@timestamp":"2023-02-02T08:34:43.220+0100","log.logger":"sourcemap","log.origin":{"file.name":"sourcemap/metadata_fetcher.go","file.line":104},"message":"init routine completed","service.name":"apm-server","ecs.version":"1.6.0"}
```

internal/sourcemap/chained.go
```diff
@@ -35,10 +36,15 @@ func (c ChainedFetcher) Fetch(ctx context.Context, name, version, path string) (
 	var lastErr error
 	for _, f := range c {
 		consumer, err := f.Fetch(ctx, name, version, path)
+		if !errors.Is(err, ErrFetcherUnvailable) {
+			return consumer, err
+		}
```
simitt, line 41: Shouldn't this be 
```

if err != nil {
  if errors.Is(err, ErrFetcherUnvailable){
    return nil, err
  }
  lastErr = err
  continue
}

if consumer != nil{
  return consumer, nil
}



```

internal/sourcemap/metadata_fetcher.go
```diff
@@ -0,0 +1,320 @@
...
+
+	"github.com/elastic/apm-server/internal/elasticsearch"
+	"github.com/elastic/apm-server/internal/logs"
+	"github.com/elastic/elastic-agent-libs/logp"
+	"github.com/elastic/go-elasticsearch/v8/esapi"
+)
+
+const (
+	syncTimeout = 10 * time.Second
+)
```
simitt, line 38: Some unit tests would be good for this and the `SourcemapFetcher`.

simitt (changes requested):
Overall looks good, but please take a look at the logs and behavior when Elasticsearch is not reachable, e.g. still starting up.

internal/sourcemap/metadata_fetcher.go
kruskall: Updatd :+1:

internal/sourcemap/metadata_fetcher.go
kruskall: Thanks :facepalm:

internal/beater/beater.go
```diff
@@ -820,76 +810,39 @@ func (s *Runner) newLibbeatFinalBatchProcessor(
...
-			if err != nil {
-				return nil, err
-			}
-			u, err := url.Parse(urlString)
-			if err != nil {
-				return nil, err
-			}
-			fleetServerURLs[i] = u
-		}
+	index := strings.ReplaceAll(cfg.IndexPattern, "%{[observer.version]}", version.Version)
```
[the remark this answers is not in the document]
kruskall: Removed :+1:

internal/sourcemap/fetcher.go
```diff
@@ -0,0 +1,114 @@
...
+	"github.com/go-sourcemap/sourcemap"
+)
+
+var (
+	ErrFetcherUnvailable  = errors.New("fetcher unavailable")
+	ErrMalformedSourcemap = errors.New("sourcemap malformed")
+)
+
+// Fetcher is an interface for fetching a source map with a given service name, service version,
+// and bundle filepath.
```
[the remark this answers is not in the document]
kruskall: Updated :+1:

internal/sourcemap/sourcemap_fetcher.go
kruskall: > I am missing why the following part is necessary. If the metadata fetcher is not done, s.backend.Fetch is triggered. If c == nil && err == nil it means that the sourcemap doesn't exist. What is the reason for blocking and waiting until metadata is ready etc?

If `c== nil && err == nil` it means there's no sourcemap associated with the provided name/version/path but the sourcemap could be an alias, in which case we block to populate the aliases: https://github.com/elastic/apm-server/pull/9722/files#diff-2941f7040f3ffdd6900280715c51cf111d7e6c28b0e7ba8449fdcd185a1dcaadR68
We're also checking the request context to avoid blocking requests indefinitely: https://github.com/elastic/apm-server/pull/9722/files#diff-2941f7040f3ffdd6900280715c51cf111d7e6c28b0e7ba8449fdcd185a1dcaadR71

internal/sourcemap/metadata_fetcher.go
kruskall: Good catch! Fixed :+1:

internal/sourcemap/metadata_fetcher.go
kruskall: I've pushed some changes, it should correctly handle ES not being available now

internal/sourcemap/sourcemap_fetcher.go
kruskall: Good catch! I've added a test for ES unreachable.

The current issue was the ES fetcher not returning errFetcherUnavailable when ES is not available.

The problems is that by the time the sourcemap fetcher realizes ES is not reachable/available the context is expired as you can see from your log message. That means the next fetcher will fail regardless since the deadline is exceeded.
Should we create a new context with a 1s (?) timeout when trying the next chained fetcher ?

internal/sourcemap/chained.go
kruskall: The idea is to always return what we get when the fetcher worked.

worked= it returns either a sourcemap or an error except `ErrFetcherUnvailable`

If we get `ErrFetcherUnvailable` then the fetcher is not available (e.g. ES not reachable) and we try the next chained fetcher.
I've pushed a new change to simplify and document the logic a bit better.

internal/sourcemap/body_caching.go
```diff
@@ -0,0 +1,111 @@
...
+		}
+		logger.Debugf("Removed id %v", key)
+	})
+
+	if err != nil {
+		return nil, fmt.Errorf("failed to create lru cache for caching fetcher: %w", err)
+	}
+
+	go func() {
+		logger.Info("listening for invalidation...")
```
simitt, line 59: nit: I'd also make this `logger.Debug` for consistency

internal/sourcemap/sourcemap_fetcher.go
simitt: I started the APM Server with pointing to a non-existent ES instance, and still see the above logs. The same behavior happens when configuring another output than ES. In such cases, or if the configured credentials do not have the right privileges, the Kibana fallback fetcher should be used. The extra query to ES for every error stacktrace frame before using the Kibana fetcher is not efficient.

simitt (changes requested):

internal/sourcemap/sourcemap_fetcher.go
kruskall: That's strange, I'll investigate a bit more!

The wrong credential and ES unreachable should be covered by the systemtests: https://github.com/elastic/apm-server/pull/9722/files#diff-d98c3b8e30f1c6e2ae6dfb07a2b2f35fd940f74fd52149b5b25bf648aa622b9cR182-R237

internal/sourcemap/body_caching.go
kruskall: Changed :+1:

internal/sourcemap/body_caching_test.go
```diff
@@ -0,0 +1,167 @@
...
+func Test_NewCachingFetcher(t *testing.T) {
+	_, err := NewBodyCachingFetcher(nil, -1, nil)
+	require.Error(t, err)
+
+	f, err := NewBodyCachingFetcher(nil, 100, nil)
+	require.NoError(t, err)
+	assert.NotNil(t, f.cache)
+}
+
+func TestStore_Fetch(t *testing.T) {
```
axw, line 51: I'm struggling to follow this test. Why does it test anything to do with Elasticsearch? AFAICS the BodyCachingFetcher is generic, so could we use a mock backend Fetcher which exercises the caching logic? Rather than updating the cache directly, make a call that would be uncached, have it hit the backend (mock) fetcher, and return a specified nil/non-nil consumer. Then you can check that the caching fetcher does the right thing.

EDIT: I see this is pre-existing code that's been moved around :)
I suppose it's fine for now, but something we should come back and improve.

axw (commented):
I haven't been following closely, so apologies if any of my questions are dumb ones.

We need a lot more unit testing in this package, particularly for a change of this size. `metadata_fetcher.go` and `sourcemap_fetcher.go` have zero coverage.

internal/sourcemap/sourcemap_fetcher.go
axw: Alternatively, I think it would be acceptable if the server doesn't apply source maps during startup while the cache is empty. That would be simplest.

simitt: +1 - I also think it is acceptable, as the initial sourcemap metadata fetching should be triggered immediately and be fairly quick.

internal/sourcemap/metadata_fetcher.go
```diff
@@ -0,0 +1,321 @@
...
+	// Try to retrieve the sourcemap from the alias map
+	i, ok := s.alias[key]
+	return i, ok
+}
+
+func (s *MetadataESFetcher) ready() <-chan struct{} {
+	return s.init
+}
+
+func (s *MetadataESFetcher) startBackgrounSync(parent context.Context) {
```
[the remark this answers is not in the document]
kruskall: Good catch!

internal/sourcemap/elasticsearch.go
```diff
@@ -76,34 +75,63 @@ func NewElasticsearchFetcher(c *elasticsearch.Client, index string) Fetcher {
 func (s *esFetcher) Fetch(ctx context.Context, name, version, path string) (*sourcemap.Consumer, error) {
 	resp, err := s.runSearchQuery(ctx, name, version, path)
 	if err != nil {
-		return nil, errors.Wrap(err, errMsgESFailure)
+		var networkErr net.Error
+		if errors.As(err, &networkErr) {
+			return nil, fmt.Errorf("failed to reach elasticsearch: %w, %v: ", errFetcherUnvailable, err)
```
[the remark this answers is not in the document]
kruskall: Ah, that's a mistake. Fixed :+1:

internal/sourcemap/metadata_fetcher.go
```diff
@@ -0,0 +1,321 @@
...
+		if err := s.sync(ctx); err != nil {
+			s.logger.Errorf("failed to fetch sourcemaps metadata: %v", err)
+		} else {
+			// only close the init chan and mark the fetcher as ready if
+			// sync succeeded
+			close(s.init)
+			s.logger.Info("init routine completed")
+		}
+
+		go func() {
```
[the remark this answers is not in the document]
kruskall: Good catch! I wrapped the initial method in a goroutine and forgot to remove this!

internal/sourcemap/body_caching.go
```diff
@@ -0,0 +1,111 @@
...
+	backend Fetcher,
+	cacheSize int,
+	invalidationChan <-chan []identifier,
+) (*BodyCachingFetcher, error) {
+	logger := logp.NewLogger(logs.Sourcemap)
+
+	lruCache, err := lru.NewWithEvict(cacheSize, func(key, value interface{}) {
+		if !logger.IsDebug() {
+			return
+		}
```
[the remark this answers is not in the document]
kruskall: This was carried over from the old code. Updated!

internal/sourcemap/elasticsearch.go
```diff
@@ -116,97 +144,34 @@ func (s *esFetcher) runSearchQuery(ctx context.Context, name, version, path stri
...
+func requestBody(name, version, path string) map[string]interface{} {
+	id := name + "-" + version + "-" + path
+
+	return search(
+		size(1),
+		source("content"),
+		query(
+			boolean(
+				must(
+					term("_id", id),
```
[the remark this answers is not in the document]
kruskall: Oh, that's a great idea! Changed :+1:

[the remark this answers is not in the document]
simitt: I wonder if the logic at https://github.com/elastic/apm-server/pull/9722/files#diff-2941f7040f3ffdd6900280715c51cf111d7e6c28b0e7ba8449fdcd185a1dcaadR84-R110 could be removed if the sourcemap fetching query still considered the absolute and relative path within one query. Is there a specific reason why the previously used [`boostedTerm`](https://github.com/elastic/apm-server/blob/main/internal/sourcemap/elasticsearch.go#L147-L153) would not work anymore?

internal/sourcemap/metadata_fetcher.go
```diff
@@ -0,0 +1,321 @@
...
+
+func (s *MetadataESFetcher) runSearchQuery(ctx context.Context) (*esapi.Response, error) {
+	var buf bytes.Buffer
+	if err := json.NewEncoder(&buf).Encode(queryMetadata()); err != nil {
+		return nil, fmt.Errorf("failed to encode metadata query: %w", err)
+	}
+
+	req := esapi.SearchRequest{
+		Index:          []string{s.index},
+		Body:           &buf,
```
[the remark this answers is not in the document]
kruskall: TIL! Thanks :+1:

internal/sourcemap/sourcemap_fetcher.go
kruskall: Updated to block until the metadata cache is populated

internal/sourcemap/metadata_fetcher.go
```diff
@@ -0,0 +1,321 @@
...
+		// First run, populate cache
+		ctx, cancel := context.WithTimeout(parent, syncTimeout)
+		defer cancel()
+
+		if err := s.sync(ctx); err != nil {
+			s.logger.Errorf("failed to fetch sourcemaps metadata: %v", err)
+		} else {
+			// only close the init chan and mark the fetcher as ready if
+			// sync succeeded
+			close(s.init)
```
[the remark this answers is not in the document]
kruskall: the logic changed (we are blocking until init is completed) so the channel is always closed now

internal/sourcemap/metadata_fetcher.go
```diff
@@ -0,0 +1,321 @@
...
+	var invalidation []identifier
+
+	for id, contentHash := range s.set {
+		if updatedHash, ok := sourcemaps[id]; ok {
+			// already in the cache, remove from the updates.
+			delete(sourcemaps, id)
+
+			// content hash changed, invalidate the sourcemap cache and update hash
+			if contentHash != updatedHash {
+				sourcemaps[id] = updatedHash
```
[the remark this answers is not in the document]
kruskall: Good catch! Updated :+1:

internal/sourcemap/metadata_fetcher.go
```diff
@@ -0,0 +1,321 @@
...
+			delete(s.set, id)
+
+			// remove aliases
+			for _, k := range getAliases(id.name, id.version, id.path) {
+				delete(s.alias, k)
+			}
+		}
+	}
+
+	s.invalidationChan <- invalidation
```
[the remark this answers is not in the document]
kruskall: Good catch!

systemtest/sourcemap_test.go
```diff
@@ -157,6 +159,83 @@ func TestSourcemapCaching(t *testing.T) {
...
+	err = srv.Start()
+	require.NoError(t, err)
+
+	// Index an error, applying source mapping and caching the source map in the process.
+	systemtest.SendRUMEventsPayload(t, srv.URL, "../testdata/intake-v2/errors_rum.ndjson")
+	result := systemtest.Elasticsearch.ExpectDocs(t, "logs-apm.error-*", nil)
+	assertSourcemapUpdated(t, result, true)
+}
+
+func TestSourcemapKibana(t *testing.T) {
```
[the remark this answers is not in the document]
kruskall: Updated :+1:

systemtest/sourcemap_test.go
```diff
@@ -157,6 +159,83 @@ func TestSourcemapCaching(t *testing.T) {
...
+			ESConfig: &apmservertest.ElasticsearchOutputConfig{
+				// Use an unreachable address
+				Hosts: []string{"127.0.0.1:12345"},
+			},
+		},
+	}
+	err = srv.Start()
+	require.NoError(t, err)
+
+	// Index an error, applying source mapping and caching the source map in the process.
```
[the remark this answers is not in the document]
kruskall: I've moved the test to named subtests, it should better explain their purpose

internal/sourcemap/metadata_fetcher.go
```diff
@@ -0,0 +1,368 @@
...
+	}()
+}
+
+func (s *MetadataESFetcher) ping(ctx context.Context) error {
+	// we cannot use PingRequest because the library is
+	// building a broken url and the request is timing out.
+	req := esapi.IndicesGetRequest{
+		Index: []string{s.index},
+	}
+	_, err := req.Do(ctx, s.esClient)
```
axw, line 162: We should be closing the response body when err == nil.

internal/sourcemap/sourcemap_fetcher.go
```diff
@@ -0,0 +1,107 @@
...
+	case <-s.metadata.ready():
+		if err := s.metadata.err(); err != nil {
+			return nil, err
+		}
+	case <-ctx.Done():
+		return nil, fmt.Errorf("error waiting for metadata fetcher to be ready: %w", ctx.Err())
+	}
+
+	// the mutex is shared by the update goroutine, we need to release it
+	// as soon as possible to avoid blocking updates.
```
axw, line 60: Is this a stale comment? I don't see any mutexes

axw (changes requested):
Thanks for the updates @kruskall, I think this is looking a lot better now.

IMO the blocking behaviour is fine since it's bounded to a ping + search request, which is roughly the same as the old ES implementation. @simitt do you still have any concerns about that?

I'd like to see some unit tests for MetadataFetcher and SourcemapFetcher before this is merged.

internal/sourcemap/metadata_fetcher.go
kruskall: :facepalm:

internal/sourcemap/sourcemap_fetcher.go
kruskall: It is, the mutex was inside `getID` but that has been moved to another struct

simitt:
>IMO the blocking behaviour is fine since it's bounded to a ping + search request, which is roughly the same as the old ES implementation. @simitt do you still have any concerns about that?

The existing behavior pre `8.7` was already blocking RUM ingest requests until sourcemaps were fetched, so while I don't think it's ideal, I have no concerns on moving forward with blocking.

simitt (approved):
Changes look good now!

The `sourcemap_fetcher_test`s only test the happy path, some checks for error cases would also be good, but test coverage looks good enough to be merged now. 

Thanks for all the effort that went into this!

internal/sourcemap/metadata_fetcher.go
```diff
@@ -0,0 +1,373 @@
...
+			} else {
+				// only close the init chan and mark the fetcher as ready if
+				// sync succeeded
+				s.logger.Info("init routine completed")
+			}
+		}
+
+		close(s.init)
+
+		// TODO make this a config option ?
```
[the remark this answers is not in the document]
kruskall: Removed :+1:

internal/beater/config/config_test.go
```diff
@@ -270,9 +269,8 @@ func TestUnpackConfig(t *testing.T) {
 					AllowOrigins: []string{"example*"},
 					AllowHeaders: []string{"Authorization"},
 					SourceMapping: SourceMapping{
-						Enabled:      true,
-						Cache:        Cache{Expiration: 8 * time.Minute},
-						IndexPattern: "apm-test*",
+						Enabled: true,
+						Cache:   Cache{Expiration: 8 * time.Minute},
```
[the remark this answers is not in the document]
kruskall: Good catch! Removed :+1:

axw:
First test with Fleet-managed APM Server is successful:
 1. Created an Elastic Cloud deployment
 2. Send a RUM error, confirmed that its stack trace was not source mapped (naturally, there are no source maps uploaded yet)
 3. Uploaded a source map
 4. Deleted existing error documents, send another RUM error; confirmed that it was source mapped

I will run some more tests running without Fleet.

axw:
I tried running Elastic Agent locally in Docker, communicating with Elasticsearch in an Elastic Cloud us-west2 deployment. Errors were indexed with: 

`error.exception.stacktrace.sourcemap.error: failed to ping es cluster: fetcher unavailable: context deadline exceeded`

Looking at the logs, this is coming from the metadata fetcher, which has a 1 second ping timeout.

Opened https://github.com/elastic/apm-server/issues/10338

PR #9750, opened 2022-12-06
Universal Profiling: avoid K/V document delete on duplicates

inge4pres:
## Motivation/summary

The Universal Profiling intake currently generates tombstones when the same stacktrace/stackframe document is ingested from 2 different agents.
We initially understood that an `op_type=index` is a no-op if the document _id is the same and the document has is not changed, but this is not true.

## Checklist

- [X] Update [CHANGELOG.asciidoc](https://github.com/elastic/apm-server/blob/main/CHANGELOG.asciidoc)

- Is it observable through the addition of either **logging** or **metrics**? ✅ 
- Is its use being published in **telemetry** to enable product improvement? ⛔ 

## How to test these changes

Enable the profiling intake and send some data through it via an Universal Profiling agent: there should be no deleted docs in the `profiling-stacktraces` and `profiling-stackframes` reading their status via the `_cat` API.


x-pack/apm-server/profiling/collector.go
```diff
@@ -403,24 +403,24 @@ func (e *ElasticCollector) SetFramesForTraces(ctx context.Context,
...
+				_ error,
 			) {
-				counterStacktracesFailure.Inc()
-				e.logger.With(
-					logp.Error(err),
-					logp.String("error_type", resp.Error.Type),
-				).Errorf("failed to index stacktrace metadata: %s", resp.Error.Reason)
+				// Error is expected here, as we tried to "create" an existing document.
+				// We increment the metric to understand the origin-to-duplicate ratio.
+				counterStacktracesDuplicate.Inc()
```
axw, line 416: How accurate should this metric be? There might be other errors - should you be parsing the error reason, and only incrementing the counter for specific errors?

rockdaboot:
```
const docIDAlreadyExists = "version_conflict_engine_exception"
...
if resp.Error.Type == docIDAlreadyExists {
    counterStacktracesDuplicate.Inc()
    return
}
// log other errors here
```

inge4pres: You are right, thanks 🙏🏼 
@rockdaboot shared a suggestion that I'm going to implement here.

inge4pres:
Btw this PR has been labeled with `port-into-apm` because a corresponding change should be applied into apm-server

rockdaboot (approved):
LGTM

axw (approved):

PR #9775, opened 2022-12-08
Universal Profiling: Reduce go-elasticsearch per-worker buffer size

christos68k:
From 16MB to 4MB. This not only reduces memory usage spikes (from >1GB to <350MB in my most heavy testing push-symbols workload) but results in faster overall processing and smoother operation.

This configuration change is synergistic with go-elasticsearch memory leak fixes that will be part of the `8.6.0` release.


axw (approved):
> This configuration change is synergistic with go-elasticsearch memory leak fixes that will be part of the 8.6.0 release.

If this should be in 8.6.x, please apply the `backport-8.6` label.

rockdaboot (approved):
:+1:

PR #9786, opened 2022-12-12
Replace modelindexer with go-docappender

axw:
## Motivation/summary

The headline change here is to extract modelindexer out of apm-server as go-docappender. This implies several other changes:

 - go-docappender uses zap for logging, so we now pass in a zap logger using the same core as used by logp.
 - go-docappender is decoupled from apm-server and its model, so the model.BatchProcessor API has been removed, and now exists as a processor in package beater. This includes the encoding of events to JSON.
 - go-docappender uses go-elasticsearch directly, whereas modelindexer used internal/elasticsearch. We remove much of the use of internal/elasticsearch from apm-server, and use go-elasticsearch directly.

Eventually we might want to remove internal/elasticsearch altogether, but that is out of scope for now.

## Checklist

## How to test these changes

N/A -- non-functional change.

## Related issues

None


internal/beater/processors.go
```diff
@@ -74,3 +80,52 @@ func newObserverBatchProcessor() model.ProcessBatchFunc {
...
+}
+
+type pooledReader struct {
+	pool         *sync.Pool
+	jsonw        fastjson.Writer
+	indexBuilder strings.Builder
+}
+
+func (r *pooledReader) Read(p []byte) (int, error) {
+	panic("should've called WriterTo")
```
lahsivjar, line 118:
```suggestion
	panic("should've called WriteTo")
```

internal/beater/processors.go
```diff
@@ -74,3 +80,52 @@ func newObserverBatchProcessor() model.ProcessBatchFunc {
...
+				return err
+			}
+			r.indexBuilder.WriteString(event.DataStream.Type)
+			r.indexBuilder.WriteByte('-')
+			r.indexBuilder.WriteString(event.DataStream.Dataset)
+			r.indexBuilder.WriteByte('-')
+			r.indexBuilder.WriteString(event.DataStream.Namespace)
+			index := r.indexBuilder.String()
+			if err := a.Add(ctx, index, r); err != nil {
+				r.reset()
```
lahsivjar, line 103: The reader is not calling `reset` on success and thus not releasing `indexBuilder` and `jsonw` back to the pool. Is that expected?

axw: Yes, in the success case the reader is in the docappender's queue. When the document body is consumed, the reader's WriteTo method will be called, and that will call reset.

lahsivjar (approved):

PR #9790, opened 2022-12-12
Stop vendoring opentelemetry-collector

axw:
## Motivation/summary

By creating our own HTTP handlers and gRPC services, which are fairly trivial to implement now, we can remove the opentelemetry-collector vendoring and mixin script. This should make it more straightforward to keep the opentelemetry-collector dependency up to date, e.g. by using Renovate or Dependabot.

## Checklist

## How to test these changes

N/A, non-functional change.

## Related issues

None


internal/beater/otlp/http.go
```diff
@@ -44,28 +50,121 @@ func init() {
...
+	UnmarshalProto([]byte) error
+}
+
+type protoMarshaler interface {
+	MarshalProto() ([]byte, error)
+}
 
-	tracesHandler, err := otlpreceiver.TracesHTTPHandler(context.Background(), consumer)
+func (h HTTPHandlers) readRequest(req *http.Request, out protoUnmarshaler) error {
+	body, err := io.ReadAll(req.Body)
```
lahsivjar, line 130: I think we should close the `req.Body` here.

axw: Why? The net/http server will close the request body once the handler returns.

lahsivjar: Ah yes, you are right.

lahsivjar (approved):

PR #9828, opened 2022-12-15
Universal Profiling: Upsert and noop for AddExecutableMetadata

christos68k:
### Summary

Rewrote AddExecutableMetadata to use a "scripted_upsert" that should reduce the number of deleted documents (a document write will be a noop if the document exists and has the same timestamp).

Refactored `ecsVersionString` into the 'common' package.


axw (approved):

rockdaboot (approved):
:+1:

simitt:
Afaik a new BC for `8.6.0` is planned for today, if this is targeting the version it'd be good to get the changes merged.

Issue #9659, opened 2022-11-23
mobile: translate otel log events to error events 


simitt:
When receiving otel log events that represent errors, translate them to APM error events. If they are coming from mobile crashes, the app crash information needs to be mapped to the APM error exception event. 

For translating a log event to an error event, use the existence of `exception.*` attributes to signal that the event is an error: https://opentelemetry.io/docs/reference/specification/logs/semantic_conventions/exceptions/

For the additional information that the error represents an app crash, one additional label will be set. Hopefully this label can be added to the semantiv conventions long term, but for now we start with a label `event.type: mobile-crash`. The information that the error event was created by a mobile app crash needs to be available via an Elasticsearch `error.exception.*` field.

Build on existing code for converting [resource logs](https://github.com/elastic/apm-server/blob/main/internal/processor/otel/logs.go) and [translating exceptions and parsing stacktraces](https://github.com/elastic/apm-server/blob/main/internal/processor/otel/exceptions.go) whereever possible. 

--- 
updated to reflect below comments (28/11/2022)

felixbarny:
We were discussing to add a dedicated attribute to signal that the log represents an app crash. The mobile agent team would be responsible to propose a name and to contribute it to the OTel semantic conventions.

axw:
Aside from the attribute mentioned above, we could consider logs with `exception.*` attributes to be errors: https://opentelemetry.io/docs/reference/specification/logs/semantic_conventions/exceptions/

If they have those, create an error. Otherwise, create an application log. That's essentially what we do for span events today.

AlexanderWert:
We discussed in the last mobile sync that, actually, there might be more than just crashes in the future. 
For example, there are ANR (Application Not Responding) events, that are very similar to errors as well, but need to be marked as a special kind of events. So, an additional field that would describe the type of error / event (e.g. crash vs. error vs. ANR) would be beneficial.

AlexanderWert:
@simitt @kruskall 

Update: We discussed this topic in the mobile sync last week. One update on this is that we will use the `event.name` = `mobile-crash` field (instead of the `event.type` as specified above).

AlexanderWert:
Also, we need to define which field we will use on the ES / ECS side to denote a mobile crash. Any preferences / ideas? Maybe the ECS field `error.type`? Right now we are using `error.exception.type` for the exception class name.

@felixbarny WDYI?

kruskall:
[...]

PR #9842, opened 2022-12-19
feat: translate otel log events to error events

kruskall:
## Motivation/summary

Translate otel logs events that represent an error to error events. Based on the otel spec for recording exceptions, at least one of the following sets of attributes is required:

exception.type
exception.message

Additionally we want to map an app crash to an error event by checking the event.name.

Add test to validate the expected APMEvent output.

## Checklist

## Related issues

Closes https://github.com/elastic/apm-server/issues/9659

### DIFF
--- a/internal/processor/otel/logs.go
+++ b/internal/processor/otel/logs.go
@@ -40,6 +40,7 @@ import (
 
 	"go.opentelemetry.io/collector/pdata/pcommon"
 	"go.opentelemetry.io/collector/pdata/plog"
+	semconv "go.opentelemetry.io/collector/semconv/v1.5.0"
 
 	"github.com/elastic/apm-data/model"
 	apmserverlogs "github.com/elastic/apm-server/internal/logs"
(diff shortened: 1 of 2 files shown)


internal/processor/otel/logs.go
```diff
@@ -120,12 +128,56 @@ func (c *Consumer) convertLogRecord(
...
 	}
 	return event
 }
 
+func updateErrorType(m pcommon.Map, event *model.APMEvent) {
+	if event.Error == nil {
+		event.Error = &model.Error{}
+	}
+
+	event.Error.Type = "crash"
```
kruskall, line 147: Suggested by @AlexanderWert in https://github.com/elastic/apm-server/issues/9659#issuecomment-1337134933

I like error.type. +1 for that!

AlexanderWert, line 147: Setting the `event.Error.Type = "crash"` is mobile-specific, but this logic here is generic for all OTel errors / exceptions. 

Instead, we should set the `event.Error.Type` to the value of the `event.name` OTel attribute.

internal/processor/otel/logs.go
```diff
@@ -120,12 +128,56 @@ func (c *Consumer) convertLogRecord(
 		}
 		event.Span.ID = spanID.HexString()
 	}
-	if attrs := record.Attributes(); attrs.Len() > 0 {
+	attrs := record.Attributes()
```
AlexanderWert, line 131: I think we should in addition set the event categorization fields (for all of the events / logs):
* `event.kind` = `event`
* `event.category` = value of the `event.domain` OTel attribute
* `event.type` = `info` (in case of logs / normal events) and `error` in case of errors / crashes, etc

kruskall:
Superseded by https://github.com/elastic/apm-data/pull/5

internal/processor/otel/logs.go
kruskall: Added! :+1:

internal/processor/otel/logs.go
kruskall: Good catch!

Issue #9703, opened 2022-11-30
Add additional aggregation intervals for metrics


simitt:
## Problem Statement

Today, if a large volume of apm data is ingested, the APM UI might take some time to be able to load data. The APM Server pre-aggregates some metrics and therefore decreases the amount of documents that need to be taken into account by the APM UI queries. The current interval of `1m` is not efficient enough though for achieving the desired UI performance. 

## Solution
APM Server will additionally roll up service, transaction and span metrics into larger intervals. It should aggregate multiple time ranges (`1m`, `10m` and `60m`) concurrently, and flush them to Elasticsearch periodically. Because this is expected to result in some storage increase, the metrics will be written to dedicated data streams, enabling interval-specific ILM policies. This enables users to adapt how long they want to keep high resolution data. 
When shutting down, the APM Server will flush the metrics so that buffered data are not lost. 

## Changes

- [x] Add additional `10m` and `60m` to transaction metrics (https://github.com/elastic/apm-server/pull/9846).
- [x] Add additional `10m` and `60m` to service destination metrics (https://github.com/elastic/apm-server/pull/9926).
- [ ] Add additional `10m` and `60m` to service metrics (#9960).

marclop:
## How to test the changes in the linked PRs



Run the following sequence


```console
$ tilt up
# Wait for APM Server to be healthy.
$ cd systemtest && go run cmd/fleetctl/main.go set-policy-config apm-1 apm-server.aggregation.service.enabled=true
$ Run the ./test.sh script from this link https://gist.github.com/marclop/2b8d1e7278d84e0e787f2b2e35578f62.
# Restart the APM Server
$ go run cmd/fleetctl/main.go set-policy-config apm-1 apm-server.aggregation.service.enabled=false
```

Issue the following query in Kibana dev tools and ensure that there are 27 metrics documents:

```
GET /metrics-apm*/_search
{
  "_source": [false]
}
```

- 9 for service metrics (3 services * 3 metricset.interval (1m, 10m, 60m))
- 9 for service destination metrics (1 dropped span per service * 3 metricset.interval (1m, 10m, 60m))
- 9 for transaction metrics (3 transactions * 3 metricset.interval (1m, 10m, 60m))

Ensure that 3 APM Services are displayed:

<img width="1792" alt="image" src="https://user-images.githubusercontent.com/7286993/210351499-3c65ffa9-36ef-4dec-9df6-4d41804f97dc.png">


And that the service map also works

[...]

PR #9846, opened 2022-12-19
txmetrics: Add `10m`, `1h` aggregation intervals

marclop:
## Motivation/summary

Modifies the existing `txmetrics` aggregator to support a list of extra
aggregation intervals by introducing a new `RollUpIntervals` (internal
only) config option. The configuration option is not user facing, but
rather allow us to configure smaller intervals for testing purposes.
Also removes the `aggregation.transactions.interval` config option.

Two new intervals are added and published by default, 10 and 60m (1h).
and these are flushed when the APM Server is stopped or restarted.

Introduces a new `metrics-apm.transaction.${interval}m` datastream
which contains the mapping definitions for transaction metrics with a
series of changes to the apmpackage generator to allow interval
templating and individual ILM policies to be tied to each interval.
A new `Metricset.Interval` field is added to the ingested documents, to
allow them to contain the aggregated interval they represent. The field
format is `${interval}m` or `${interval}s` if < 60s.

Older APM Servers sending data to an Elasticsearch cluster with updated
APM Package, will continue indexing metrics to `metrics-apm.internal`.

## Checklist

## How to test these changes

1. `tilt up`
2. Run the `test.sh` script in https://gist.github.com/marclop/01a8f4270cc442573fd8319ca140015d.
3. Toggle expvar in the APM integration so metrics are published immediately on hot restart.
4. Verify that 3 metric sets per application and data stream `metrics-apm.transaction.1m`, `metrics-apm.transaction.10m`, `metrics-apm.transaction.60m`, and that they contain `event.duration` in nanoseconds matching the data stream interval (`constant_keyword`).
```
GET /metrics-apm*/_search
{
  "_source": [false],
  "fields": [
    "metricset.name",
    "event.duration",
    "service.name",
    "metricset.interval"
  ]
}
```
5. Verify that each of the different data streams has a dedicated ILM policy.

## Related issues

Part of #9703
Depends on https://github.com/elastic/apm-data/pull/7


apmpackage/apm/data_stream/internal_metrics_interval/manifest.yml
```diff
@@ -0,0 +1,36 @@
+title: APM internal metrics
+type: metrics
+dataset: apm.internal-interval
```
axw, line 3: [`data_stream.dataset`](https://www.elastic.co/guide/en/ecs/current/ecs-data_stream.html#field-data-stream-dataset) is not allowed to contain "-", as that's the separator used for identifying the data stream type, dataset, and namespace.

x-pack/apm-server/aggregation/txmetrics/aggregator.go
```diff
@@ -55,7 +55,9 @@ type Aggregator struct {
 	tooManyGroupsLogger *logp.Logger
 
 	mu               sync.RWMutex
-	active, inactive *metrics
+	active, inactive map[time.Duration]*metrics
```
axw, line 58: Given that we iterate over the intervals in order, would it be worth making these `[]*metrics`?

x-pack/apm-server/aggregation/txmetrics/aggregator.go
```diff
@@ -141,16 +160,28 @@ func (a *Aggregator) Run() error {
...
-				"publishing transaction metrics failed: %s", err,
-			)
+		// Publish the metricsets for all configured intervals.
+		for _, interval := range a.intervals {
+			// Publish $interval MetricSets when:
+			//  - ticks * MetricsInterval % $interval == 0.
+			//  - Aggregator is stopped.
+			if !stop && (ticks*uint64(a.config.MetricsInterval))%uint64(interval) != 0 {
+				continue
+			}
```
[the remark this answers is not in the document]
marclop: I did have that check initially and removed it since we aren't planning  to allow users to tweak that setting it seemed superfluous, I'm happy adding it back.

x-pack/apm-server/aggregation/txmetrics/aggregator.go
marclop: I suppose we could, would you prefer that?

axw: I don't have a very strong opinion on this - slight preference because it'll be slightly faster. Not sure if it's worth it. If it makes things harder to read then don't worry about it.

apmpackage/apm/data_stream/internal_metrics_interval/manifest.yml
felixbarny: > I would strongly prefer if we introduced transaction and service metric data streams.

I’m good with separating out data streams for metrics that we’re downsampling - transaction, service, and service destination metrics.

marclop: So we'd have 1 internal metric data stream (`apm-metrics.internal`), and 3 "aggregation" metrics data streams (1 per aggregated interval) then.

apmpackage/apm/data_stream/internal_metrics_transaction/fields/faas_metrics.yml
```diff
@@ -0,0 +1,20 @@
...
+- name: faas.duration
+  type: float
+  metric_type: counter
+  unit: ms
+  description: The total number of milliseconds elapsed during the execution of the FaaS invocation.
+- name: faas.timeout
+  type: long
+  metric_type: counter
+  unit: ms
+  description: The FaaS invocation timeout, expressed in milliseconds.
```
axw, line 20: Are these relevant to transaction metrics? Doesn't look like it, looking at txmetrics code.

apmpackage/apm/data_stream/internal_metrics_transaction/fields/fields.yml
```diff
@@ -0,0 +1,85 @@
...
+  description: |
+    Request ID of the function invocation.
+- name: faas.trigger.request_id
+  type: keyword
+  description: |
+    The ID of the origin trigger request.
+- name: faas.trigger.type
+  type: keyword
+  description: |
+    The trigger type.
```
axw, line 20: I don't think these are relevant either

apmpackage/apm/data_stream/internal_metrics_transaction/fields/fields.yml
```diff
@@ -0,0 +1,85 @@
...
+- name: faas.version
+  type: keyword
+  description: |
+    The lambda function version.
+- name: kubernetes.pod.name
+  type: keyword
+  description: |
+    Kubernetes pod name
+- name: metricset.name
+  type: keyword
```
axw, line 34: Can we make this `constant_keyword`, and set the value to "transaction" here?

axw (commented):
Looking good! As discussed on Slack, we'll need to handle transaction metrics from older servers going to `metrics-apm.internal-*`, and either upgrade and re-route, or drop them.

apmpackage/apm/data_stream/internal_metrics_transaction/fields/faas_metrics.yml
marclop: absolutely right, I'll delete these.

apmpackage/apm/data_stream/internal_metrics_transaction/fields/fields.yml
marclop: `faas.trigger.type` is used, but the other two aren't, I'll remove them.

apmpackage/apm/data_stream/internal_metrics/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -1,6 +1,8 @@
 ---
 description: Pipeline for ingesting APM internal metrics.
 processors:
+  - pipeline: # drop transaction metrics coming from older APM Server versions.
+      name: drop_tx_metrics
```
axw, line 5: Should we inline the pipeline here? It's not used anywhere else.

apmpackage/apm/data_stream/internal_metrics_transaction/manifest.yml
```diff
@@ -0,0 +1,19 @@
+title: APM transaction metrics
+type: metrics
+dataset: apm.transaction
+dataset_is_prefix: true
+ilm_policy: metrics-apm.internal_metrics_transaction-default_policy
```
axw, line 5: Should we change the directory name to "transaction_metrics", to be more in line with the data stream name? Then the ILM policy name would be more consistent too.

apmpackage/apm/data_stream/internal_metrics_transaction/manifest.yml
```diff
@@ -0,0 +1,19 @@
...
+elasticsearch:
+  index_template:
+    mappings:
+      # Internal metrics should have all fields strictly mapped;
+      # we are in full control of the field names.
```
axw, line 10:
```suggestion
      # Transaction metrics should have all fields strictly mapped;
      # we are in full control of the field names.
```

axw (commented):
LGTM from a high level. Just a few minor comments. I haven't pored over the txmetrics/aggregator.go code changes.

x-pack/apm-server/aggregation/txmetrics/aggregator.go
marclop: I looked into making the change but in my opinion it made the code slightly less readable and cryptic. I'm happy to revisit if you feel it would be better to move it to a slice.

axw: Totally fine, thanks for looking into it.

felixbarny (approved):
LGTM. Note to self and others: `event.duration` is mapped as a constant_keyword and describes the aggregation interval.
One question I have is how the UI is supposed to know which intervals to target as the aggregation intervals are configurable. Should it do a terms_enum on the `event.duration`, query for `metrics-apm.transaction-*` data streams, or hard-code the intervals?

marclop:
@felixbarny I was wondering about that myself too. I had a couple of thoughts on what to do, my understanding is that the UI will attempt to use `10m` buckets on longer time ranges, I don't know how they'll _discover_ which time ranges are available, and how to fill them, Aggregation intervals are available as:

- Templated on the index patterns (`metrics-apm.transaction.{1m,10m,60m}`)
- In each individual document as `event.duration` as `constant_keyword` instead of `long` type.

felixbarny:
> Templated on the index patterns (`metrics-apm.transaction.{1m,10m,60m}`)

Can the UI rely on these granularities or can users modify that, so that there are these data streams `metrics-apm.transaction.{30s,5m,30m}`, for example?

marclop:
@felixbarny we discussed this in the stand-up and have an agreement that we should disallow users to change any of the aggregation intervals going forward. This will be a breaking change but will guarantee the UI behavior going forward.

Another open point is that the current data stream definitions don't allow dedicated ILM policies, I'll investigate how we can define an ILM policy per interval without making the APM package definitions look too bloated.

felixbarny:
> I'll investigate how we can define an ILM policy per interval without making the APM package definitions look too bloated.

To avoid duplicating the field definitions, can we use symlinks in package definitions?

marclop:
I've updated the code to:

- Create 3 different data streams from 3 intervals using templates. This allows uses to configure the retention policy for the different periods. By default I've configured some rollover and delete settings in the different intervals, please let me know if we should change the retention windows. 
- Disallow configuring the aggregation intervals for transaction metrics.

felixbarny:
> Create 3 different data streams from 3 intervals using templates. 

I see that the ILM policies are so that we keep 30d@1m, 90d@10m, and 180d@60m. Previously, we were storing 90d@1m. What's the relative overall storage overhead or storage saving that results from this change?

marclop:
@felixbarny, I haven't run the numbers. I had assumed higher resolution for most cases for 30d, then lower resolution for 90d, and even more up to 180. I'm happy to keep it 90d, 180d and 360d if you feel that'd be more appropriate.

dgieselaar:
@marclop we need a constant keyword in the index that denotes the aggregation interval to select the right data stream. I've used `metricset.interval` in my tests but I've just made that up, maybe there's an ECS-compatible alternative?

marclop:
@dgieselaar. I've "re-purposed" `event.duration` for all the `metrics-apm.transaction.{1m,10m,60m}` data streams. The raw value is stored as a `constant_keyword` in nanoseconds.

https://github.com/elastic/apm-server/blob/242f4d8bf6c80eb362b9ad9c69b8e96938afc075/systemtest/approvals/TestTransactionAggregation.approved.json#L19

dgieselaar:
@marclop it would be easier for us if it was a term, like `1m` or `10m`. Also for custom dashboards, e.g. consider the difference between `event.duration: 3600000000000` and `metricset.interval:60m`. I realise `event.duration` should be nanoseconds but maybe we can use something else?

felixbarny:
IINM, we should be saving around 53% with these settings:

```
30/1m + 90/10m + 180/60m=42 day's worth of 1m buckets
1-(42@1m/90@1m)=53.3%
```

felixbarny:
If we wanted, we could even add an additional layer of 10s granularity stored for 7d and still use less storage than before.

marclop:
@dgieselaar Regarding the format of this new `metricset.interval` field, should it always be formatted in minutes with the `m` unit? Or are we planning to use different units based on the value?

I'm asking because I think we should define which unit is the minimum and maximum unit for that field. I've created a small snipped so you can run it and see different outputs: https://go.dev/play/p/V31MJ_SQNHr.

`60 minutes` could be formatted either `1h0m0s`, `1h` or `60m`. If we introduce intervals lower than 1m, then the problem expands.

I think we should store either the value in seconds (`50s`, `10s`, etc) or in minutes if `value >= minute` (`1m`, `60m`, `720m` (`12h`)).

marclop:
I've opened https://github.com/elastic/apm-data/pull/7 to update the model to include the new field.

dgieselaar:
@marclop re:
> I think we should store either the value in seconds (50s, 10s, etc) or in minutes if value >= minute (1m, 60m, 720m (12h)).

This sounds good to me. I'm also fine with 1h, 12h etc beyond a 1h interval.

simitt:
>Looking good! As discussed on Slack, we'll need to handle transaction metrics from older servers going to metrics-apm.internal-*, and either upgrade and re-route, or drop them.

Apologies for being late to this discussion, but dropping them would basically mean that as soon as one apm-server is upgraded, all transaction metrics from not (yet) upgraded APM Servers would be dropped. The APM UI needs to be able to handle the case that tx metrics are available in different data streams anyways, why do we drop transaction metrics that are going to the _old_ datastreams, rather than making it a UI problem how to deal with them? As soon as all APM Servers are upgraded, the _old_ tx metrics won't be there anyways, up until then, why not just ignoring them but preserving the data? Dropping seems radical, especially in a minor.

apmpackage/apm/data_stream/transaction_interval_metrics/elasticsearch/ilm/default_policy.60m.json
```diff
@@ -0,0 +1,23 @@
+{
+    "policy": {
+        "phases": {
+            "hot": {
+                "actions": {
+                    "rollover": {
+                        "max_age": "30d",
+                        "max_size": "20gb"
```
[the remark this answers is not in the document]
marclop: I've changed the `max_size` to `50gb` as before. I can't remember why I changed the max size, but I do remember thinking that 50gb worth of metrics (especially only `transaction metrics`), seemed like quite a large index.

I did lower the `max_age` for the smaller intervals:

- `1m` => `7d`
- `10m` => `14d`
- `60m` => `30d` (default)

marclop:
> Apologies for being late to this discussion, but dropping them would basically mean that as soon as one apm-server is upgraded, all transaction metrics from not (yet) upgraded APM Servers would be dropped. The APM UI needs to be able to handle the case that tx metrics are available in different data streams anyways, why do we drop transaction metrics that are going to the old datastreams, rather than making it a UI problem how to deal with them? As soon as all APM Servers are upgraded, the old tx metrics won't be there anyways, up until then, why not just ignoring them but preserving the data? Dropping seems radical, especially in a minor.

@simitt Since we're removing the transaction, service and service destination metrics from the `metrics-apm.internal`, my understanding is that we'll remove the fields that are specific to those metrics (tx, service and service destination) from `metrics-apm.internal` as well, because of that, each PR that moves those metrics to its own dedicated interval data stream, should include an ingest pipeline that drops those metric sets.

We reasoned with @axw that only the last `1m` interval metrics would be lost (time between the new integration package being installed and the APM Servers being upgraded) and as such, it's a very small amount of data "lost".

simitt:
> We reasoned with @axw that only the last 1m interval metrics would be lost (time between the new integration package being installed and the APM Servers being upgraded) and as such, it's a very small amount of data "lost".

This makes sense for the upgrade scenario, which will cover a majority of usecases. I wonder if the setup of multiple APM Servers of different versions writing to the same Elasticsearch instance is a real scenario or a theoretical problem. 
How much work would it be to add a pipeline that sends the 1m interval metrics from older APM Servers to the new data stream? The _old_ and the _new_ 1m interval datastreams would be compatible afaics.

marclop:
We discussed the drop pipeline with @simitt and came to the conclusion that the best course of action is leaving the `metrics-apm.internal` as is without a drop pipeline, let users upgrade and introduce a drop pipeline in an upcoming version providing a clear upgrade path so metrics aren't lost.

marclop:
@simitt I believe I've updated all the docs for metrics. Let me know if I've missed anything

simitt (approved):
Changes LGTM! 

Please also update the [dev-docs on transaction metrics](https://github.com/elastic/apm-server/blob/main/dev_docs/trace_metrics.md#transaction-metrics). You can do that in a follow up, when working on the other aggregation metrics.

marclop:
systemtest is flaky, merging anyway

Issue #9656, opened 2022-11-23
service metrics: add histograms


simitt:
Add histograms to service metrics. This lets us use service metrics for the service overview page in the future, without having to do more pre-flight checks on whether the service metrics have histograms or not.

Issue #9655, opened 2022-11-23
service metrics: add aggregate_metric_double


simitt:
Add a `aggregate_metric_double` field for the duration on transaction metrics and service metrics (for example `transaction.duration.summary`) to more efficiently aggregate the average duration, without having to iterate over the histogram buckets.

carsonip:
Service metrics is already using DurationSummary and DurationHistogram in https://github.com/elastic/apm-server/blob/04ca7b537fb8dd83da47a484833541913f8168c4/x-pack/apm-server/aggregation/servicemetrics/aggregator.go#L442 while transaction metrics is only using DurationHistogram but not DurationSummary in https://github.com/elastic/apm-server/blob/04ca7b537fb8dd83da47a484833541913f8168c4/x-pack/apm-server/aggregation/txmetrics/aggregator.go#L555

Is there anything extra needed for service metrics in this case? Or is this task only about adding DurationSummary to transaction metrics?

simitt:
IMO adding `transaction.duration.summary` to transaction metrics would fulfill this ask, and also bring service and transaction metrics in line again. @felixbarny @dgieselaar were you expecting something additionally?

dgieselaar:
@simitt no, don't think so.

PR #9864, opened 2022-12-21
feat: add histogram support to service metrics

kruskall:
## Motivation/summary

Add histograms to service metrics.
Add histogram significant digit setting. (default 5)
Update tests.

## Checklist

## Related issues

Closes https://github.com/elastic/apm-server/issues/9656

### DIFF
--- a/x-pack/apm-server/aggregation/servicemetrics/aggregator.go
+++ b/x-pack/apm-server/aggregation/servicemetrics/aggregator.go
@@ -291,18 +291,19 @@ func (mb *metricsBuffer) storeOrUpdate(key aggregationKey, metrics serviceMetric
 	entry := &mb.space[mb.entries]
 	entry.aggregationKey = key
 
+	entry.serviceMetrics = serviceMetrics{
+		transactionDuration: metrics.transactionDuration,
+		transactionCount:    metrics.transactionCount,
+		failureCount:        metrics.failureCount,
+		successCount:        metrics.successCount,
+	}
+
 	if entry.serviceMetrics.histogram == nil {
-		entry.serviceMetrics = serviceMetrics{
-			transactionDuration: metrics.transactionDuration,
-			transactionCount:    metrics.transactionCount,
-			failureCount:        metrics.failureCount,
-			successCount:        metrics.successCount,
-			histogram: hdrhistogram.New(
-				minDuration.Microseconds(),
-				maxDuration.Microseconds(),
-				mb.significantFigures,
-			),
-		}
+		entry.histogram = hdrhistogram.New(
+			minDuration.Microseconds(),
+			maxDuration.Microseconds(),
+			mb.significantFigures,
+		)
 	} else {
 		entry.serviceMetrics.histogram.Reset()
 	}
--- a/x-pack/apm-server/aggregation/servicemetrics/aggregator.go
+++ b/x-pack/apm-server/aggregation/servicemetrics/aggregator.go
@@ -15,6 +15,7 @@ import (
 	"github.com/pkg/errors"
 
 	"github.com/elastic/elastic-agent-libs/logp"
+	"github.com/elastic/go-hdrhistogram"
 
 	"github.com/elastic/apm-data/model"
 	"github.com/elastic/apm-server/internal/logs"
@@ -23,6 +24,11 @@ import (
 
 const (
 	metricsetName = "service"
+
+	minDuration time.Duration = 0
+	maxDuration time.Duration = time.Hour
+
+	histogramCountScale = 1000
 )
 
 // AggregatorConfig holds configuration for creating an Aggregator.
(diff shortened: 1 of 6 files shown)


x-pack/apm-server/aggregation/servicemetrics/aggregator.go
```diff
@@ -268,13 +291,23 @@ func (mb *metricsBuffer) storeOrUpdate(key aggregationKey, metrics serviceMetric
...
-		transactionDuration: metrics.transactionDuration,
-		transactionCount:    metrics.transactionCount,
-		failureCount:        metrics.failureCount,
-		successCount:        metrics.successCount,
+	if entry.serviceMetrics.histogram == nil {
+		entry.serviceMetrics = serviceMetrics{
+			transactionDuration: metrics.transactionDuration,
+			transactionCount:    metrics.transactionCount,
+			failureCount:        metrics.failureCount,
+			successCount:        metrics.successCount,
```
simitt, line 299: Why would you set `duration`, `count`, `failureCount` and `successCount` dependent of `histogram==nil`? Isn't that independent of the histogram initialization?

simitt (commented):
Generally looks good; afaics the same defaults are used as in the transaction metrics aggregation. Left one comment.

x-pack/apm-server/aggregation/servicemetrics/aggregator.go
kruskall: Good catch! Updated! +1

x-pack/apm-server/aggregation/servicemetrics/aggregator.go
```diff
@@ -381,6 +442,17 @@ func makeMetricset(key aggregationKey, metrics serviceMetrics) model.APMEvent {
 				Count: metricCount,
 				Sum:   float64(time.Duration(math.Round(metrics.transactionDuration)).Microseconds()),
 			},
+			DurationHistogram: model.Histogram{
+				Counts: counts,
+				Values: values,
+			},
```
lahsivjar, line 448: Now that we have histograms, do we need to keep the `SummaryMetric`?

kruskall: Good question! I think so but my assumption is based on the output document @axw posted in the kibana issue: https://github.com/elastic/kibana/issues/146734#issuecomment-1342467414

axw: Yes we still need them, in fact we'll be adding them to transaction metrics too: https://github.com/elastic/apm-server/issues/9655

The reason for having both is that calculating sum, count, and average over histograms is considerably more expensive due to the way they're stored. Elasticsearch ends up having to run through all the buckets of the histogram, vs. looking up one or two sub-fields of an `aggregate_metric_double`.

In the future we should be able to store a histogram along with sum, count, and optionally min & max in a single field: https://github.com/elastic/elasticsearch/issues/81031

lahsivjar: Ah, that makes sense. Thanks for the clarification, Andrew.

lahsivjar (approved):
LGTM!

simitt (approved):

marclop:
Tested with 8.7.0 BC2 on ESS:

```go
package main

import (
	"fmt"
	"time"

	"go.elastic.co/apm/v2"
)

func main() {
	tracer := apm.DefaultTracer()
	for i := 0; i < 3_000; i++ {
		for _, duration := range []int{1, 100} {
			tx := tracer.StartTransaction(fmt.Sprintf("tx-%d", i), fmt.Sprintf("type-%d", i))
			// duration to report is 2ms.
			span := tx.StartSpanOptions(fmt.Sprintf("span-%d", i), fmt.Sprintf("type-%d", i), apm.SpanOptions{
				ExitSpan: true,
			})
			span.End()
			tx.End()
		}
		if i%500 == 0 {
			tracer.Flush(nil)
		}
	}
	tracer.Flush(nil)
}
```

```
GET /metrics-apm*/_search?filter_path=hits.hits
{
  "size": 2, 
  "_source": [false], 
  "fields": [
    "metricset.name",
    "transaction.duration.histogram"
  ], 
  "query": {
    "match": {
      "metricset.name": "service_transaction"
    }
  }
}
```

```
{
  "hits": {
    "hits": [
      {
        "_index": ".ds-metrics-apm.service_transaction.1m-default-2023.02.20-000001",
        "_id": "w6vpbIYBZ-vqtropJh-E",
        "_score": 1,
        "_source": {},
        "fields": {
          "metricset.name": [
            "service_transaction"
          ],
          "transaction.duration.histogram": [
            {
              "values": [
                1003,
                100351
              ],
              "counts": [
                1,
                1
              ]
            }
          ]
        }
      },
      {
        "_index": ".ds-metrics-apm.service_transaction.1m-default-2023.02.20-000001",
        "_id": "xKvpbIYBZ-vqtropJh-E",
        "_score": 1,
        "_source": {},
        "fields": {
          "metricset.name": [
            "service_transaction"
          ],
          "transaction.duration.histogram": [
            {
              "values": [
                1003,
                100351
              ],
              "counts": [
                1,
                1
              ]
            }
          ]
        }
      }
    ]
  }
}
```

PR #9881, opened 2022-12-23
fix: fall back to system memory if cgroup exceeds the available memory

kruskall:
## Motivation/summary
I couldn't find any issue in the cgroup reading logic that would explain the bug in https://github.com/elastic/apm-server/pull/9868

On the other hand, it looks like `9223372036854771712` is the default value for cgroup v1 for unlimited memory (no limit).

I was able to reproduce the content of the limit_in_bytes file by creating a fedora 30 instance in aws and running a docker container inside of it. Then reading the memory limit inside the container:

```
# cat /sys/fs/cgroup/memory/memory.limit_in_bytes 
9223372036854771712
```

Note: fedora 30 is important as later versions are using cgroup v2

The PR is adding a fall back to system memory if cgroup returns a value that exceeds the total memory.

## Checklist

### DIFF
--- a/internal/beater/beater.go
+++ b/internal/beater/beater.go
@@ -199,8 +199,8 @@ func (s *Runner) Run(ctx context.Context) error {
 			memLimit = float64(limit) / 1024 / 1024 / 1024 * 0.5
 		}
 		if memLimit > float64(limit) {
-			s.logger.Info("cgroup memory limit exceed available memory, falling back to total system memory")
-			memLimit = float64(limit)
+			s.logger.Info("cgroup memory limit exceed available memory, falling back to the total system memory")
+			memLimit = float64(limit) / 1024 / 1024 / 1024 * 0.5
 		}
 	}
 	if memLimit <= 0 {
--- a/internal/beater/beater.go
+++ b/internal/beater/beater.go
@@ -189,14 +189,18 @@ func (s *Runner) Run(ctx context.Context) error {
 			memLimit = float64(limit) / 1024 / 1024 / 1024 * 0.8
 		}
 	}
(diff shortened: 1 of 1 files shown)


internal/beater/beater.go
```diff
@@ -189,15 +189,19 @@ func (s *Runner) Run(ctx context.Context) error {
...
+	} else {
+		if memLimit <= 0 {
+			s.logger.Info("no cgroups detected, falling back to total system memory")
 			// If no cgroup limit is set, only return 50% of the total memory.
 			// to have a margin of safety for other processes.
 			memLimit = float64(limit) / 1024 / 1024 / 1024 * 0.5
 		}
+		if memLimit > float64(limit) {
+			s.logger.Info("cgroup memory limit exceed available memory, falling back to total system memory")
+			memLimit = float64(limit)
```
lahsivjar, line 203: I think the case for keeping the memory limit to be a fraction of total memory should hold for this case also.

lahsivjar (commented):
LGTM other than the final value of the limit.

internal/beater/beater.go
kruskall: Agreed! Updated :+1:

lahsivjar (approved):
LGTM!

lahsivjar:
@simitt Should we backport this to `8.6`?

simitt (approved):
Great work @kruskall in reproducing and finding the root cause 🚀

simitt:
>@simitt Should we backport this to 8.6?

yes, it would introduce a known bug otherwise, and I consider it low risk to backport. @kruskall can you take care of it please?

PR #9911, opened 2022-12-28
Use process time as event time for overflow metric

### LINKED ISSUE #9648: Stop sending single-event metrics if max groups limit is exceeded
# Context

To prevent OOMs, we place grouping cardinality limits on metrics aggregation. By default we allow at most 10000 groups of `transaction`, `service`, and `service_destination` metrics per reporting interval (1 minute). After each report, the groups are reset. Once the limits are reached, we start emitting single-event metrics.

Producing these single-event metrics means that, in the face of misconfigured or misbehaving agents that produce high-cardinality transaction names (or other dimensions), the ingest performance, query performance, and storage cost are unbounded. The impact on ingest performance will be exacerbated as we move to [TSDS](https://www.elastic.co/guide/en/elasticsearch/reference/current/tsds.html), which improves query performance but at the cost of ingest performance.

# Proposed change

We will stop emitting single-event metrics to set some bounds on performance and storage cost. At the same time, we must ensure that users who do have high cardinality transactions/etc. do not have a worse experience than today. To that end we will ensure that users can observe that limits have been reached, and that the UI can inform users of how they might address this (e.g. fix instrumentation/agent configuration, or scale up APM Server).

Initially we will address transaction and service metrics.

We will have the following limits:
 1. max total transaction groups ([`apm-server.aggregation.transaction.max_groups`](https://www.elastic.co/guide/en/apm/guide/master/transaction-metrics.html#transactions-max_groups))
 2. max transaction groups per service to avoid a single misbehaving service consuming all the buckets
 3. max services

## Transaction groups per service limit

Some services (e.g. RUM) may produce high cardinality transaction names, while others are well behaved. To avoid one misbehaving service to eat up the limit for all other services, we'll impose a per-service transaction group limit.

Once the limit has been reached, we will aggregate transactions in a dedicated "other" transaction group per service, e.g. `transaction.name: "other"`. This special metricset has an additional metric that tracks the unique count (aka cardinality) of all transaction names that were grouped in this bucket.

## Total transaction groups limit

When the [apm-server.aggregation.transaction.max_groups](https://www.elastic.co/guide/en/apm/guide/master/transaction-metrics.html#transactions-max_groups) limit is reached, we will increment the "other" transaction name bucket of the corresponding services. We'll ensure that even if the total transaction groups limit is reached, we can collect the "other" transaction name until the service metrics limit is reached.

This ensures that even if a small number of services use up all the aggregation buckets (10 services with 1000 transaction groups can exhaust the default limit on a small instance), we can still track the metrics that are needed for the service overview page.

## Service metrics limit

We'll introduce a dedicated limit for the number of services for which we're collecting service metrics. Once the service limit has been reached, we will aggregate service metrics in an "other" service name. The service metrics for this special "other" service name has an additional metric that tracks the unique count (aka cardinality) of all service names that were grouped in this bucket.

## Configuration

The transaction group limit is currently configurable in standalone APM Server, and defaults to 10000; it is not configurable for the integration. Similarly, the service limit defaults to 10000. We will maintain these going forward, but will change the defaults to be based on available memory.

The transaction groups per service limit will be 10% of the transaction group limit.

| Memory | tx groups | tx groups per service | services |
|--|--|--|--|
| 1GB | 10000 | 1000 | 10000 |
| 8GB | 80000 | 8000 | 80000 |
| 32GB | 320000 | 32000 | 320000 |

lahsivjar:
## Motivation/summary

In PR #9856 we introduced dedicated overflow buckets for recording transaction and service metrics after limits get breached. The overflow metrics for these buckets were published along with the aggregated metrics document using the timestamp of the first event that overflowed. Due to this behavior, it was possible that an old timestamped event metric might skew the overflowed event's timestamp. In this PR, we use the processing time as the event timestamp for the overflow buckets.

## Checklist

## How to test these changes

NA; to be tested with #9856 

## Related issues

Closes #9648 
Related to #9856


marclop (approved):
I think this approach makes more sense as you describe. I found the test a bit hard to follow, though.

lahsivjar:
@marclop Thanks for the review, I have taken a stab at simplifying the test. Let me know if it is easier to follow now.

marclop (approved):
Looks better now, thanks for the change!

simitt (approved):
+1 on this change

PR #9918, opened 2022-12-29
test: reduce retries and avoid blocking on failed request in TestQueryElasticsearchClusterName

kruskall:
## Motivation/summary

Copy the default es config and reduce the number of retries to 1. 
Disable backoff to avoid blocking by changing the backoff init to 0.

Test time went from 10s to 0s

## Checklist


simitt (approved):
Great comments! Very helpful when coming across this code.

PR #9928, opened 2022-12-29
ci: Move systemtests to GitHub actions

marclop:
## Motivation/summary

Removes the system tests from Jenkins over to GitHub Actions, since they take a long time to run in Jenkins.

## Related issues

#9567


marclop:
All the CI jobs were run in ~15m, vs ~30m before.

reakaleek (approved):
LGTM

However, I see you are omitting the junit test report. 

We have established a pattern to be able to post junit reports as github checks.

- https://github.com/elastic/gmux/blob/main/.github/workflows/test.yml
- https://github.com/elastic/gmux/blob/main/.github/workflows/test-reporter.yml
- [Example report](https://github.com/elastic/gmux/pull/14/checks?check_run_id=10180803086)

carsonip (approved):
It appears that we are removing junit reporting. Not sure how things will look when tests fail but the performance improvement over Jenkins really is good.

marclop:
@elastic/apm-server happy to spend some time looking at the Junit reports, in the past, I haven't found these that useful, and opening the `go test` output seems to provide the same value and none of the overhead and complication of JUnit. Let me know what you folks think.

.ci/Jenkinsfile
```diff
@@ -122,48 +122,6 @@ pipeline {
...
-              expression { return env.ONLY_DOCS == "false" }
-            }
-          }
-          steps {
-            withGithubNotify(context: 'System Tests', tab: 'tests') {
-              deleteDir()
-              unstash 'source'
-              dir("${BASE_DIR}"){
-                withGoEnv(){
-                  sh(label: 'Run Linux tests', script: './.ci/scripts/linux-test.sh')
```
kruskall, line 146: **suggestion**: can we delete `linux-test.sh` ? Is it used anywhere ?

kruskall (changes requested):

kruskall (approved):
LGTM!

marclop:
@reakaleek I've added the job as instructed but the report generator doesn't seem to be running. Is there any step I've missed or anything else I need to do?

marclop:
I just read that new workflows need to be in main before they run, so I'll merge this PR and check on a draft PR the results.

reakaleek:
> I just read that new workflows need to be in main before they run, so I'll merge this PR and check on a draft PR the results.

Yes, your are right! Sorry, I should have mentioned that before you started wondering!

kruskall:
@Mergifyio backport 8.7

PR #9931, opened 2022-12-29
Add _doc_count to service destination metrics

### LINKED ISSUE #8340: Set _doc_count for service destination metrics
When transaction metrics are sent to Elasticsearch, [`_doc_count`](https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-doc-count-field.html) is set, which means that bucketing aggregations will use this field to calculate the doc count for a bucket, which frees us from having to do a `value_count` sub-aggregation, which can be expensive.

Similarly, for service destination metrics, we can send up `_doc_count` as well, removing the need to do a `sum` aggregation on `span.destination.service.response_time.count`. This is especially useful when switching between span events and service destination metrics, similarly to how we deal with transaction metrics, because the aggregations needed to calculate statistics are the same, just on different fields.

carsonip:
## Motivation/summary

TODO

## Checklist

## How to test these changes

TODO

## Related issues

Closes #8340


simitt (approved):
Thanks for picking this up, changes look great!

marclop:
Tested with 8.7.0 BC2 on ESS:

```go
package main

import (
	"fmt"
	"time"

	"go.elastic.co/apm/v2"
)

func main() {
	tracer := apm.DefaultTracer()
	for i := 0; i < 3_000; i++ {
		for _, duration := range []int{1, 100} {
			tx := tracer.StartTransaction(fmt.Sprintf("tx-%d", i), fmt.Sprintf("type-%d", i))
			// duration to report is 2ms.
			span := tx.StartSpanOptions(fmt.Sprintf("span-%d", i), fmt.Sprintf("type-%d", i), apm.SpanOptions{
				ExitSpan: true,
			})
			span.End()
			tx.End()
		}
		if i%500 == 0 {
			tracer.Flush(nil)
		}
	}
	tracer.Flush(nil)
}
```

```
GET /metrics-apm*/_search?filter_path=hits.hits
{
  "size": 2, 
  "_source": [false], 
  "fields": [
    "metricset.name",
    "_doc_count"
  ], 
  "query": {
    "match": {
      "metricset.name": "service_destination"
    }
  }
}
```

```
{
  "hits": {
    "hits": [
      {
        "_index": ".ds-metrics-apm.service_destination.1m-default-2023.02.20-000001",
        "_id": "X6vpbIYBZ-vqtropJkKM",
        "_score": 1,
        "_source": {},
        "fields": {
          "metricset.name": [
            "service_destination"
          ],
          "_doc_count": [
            7919
          ]
        }
      }
    ]
  }
}
```

Issue #9699, opened 2022-11-30
Set labels for resources created with terraform


simitt:
Similar to what @marclop did in https://github.com/elastic/apm-server/pull/9625, we need to set tags for all other terraform modules with the default CSP tags and accept a project variable, to define the purpose of each importing module.

kruskall:
Both PRs are merged. Closing this.

PR #9947, opened 2023-01-01
refactor: update labels in benchmark terraform script

kruskall:
## Motivation/summary

Use the module/tags module for concistency across resources. 
Set created_date label.
Fix coalesce being unused.

## Checklist

## Related issues

Related https://github.com/elastic/apm-server/issues/9699


testing/benchmark/main.tf
```diff
@@ -9,23 +9,22 @@ terraform {
...
+    }
   }
 }
 
+resource "time_static" "created_date" {}
+
 locals {
   ci_tags = {
-    environment  = coalesce(var.ENVIRONMENT, "dev")
-    repo         = coalesce(var.REPO, "apm-server")
```
kruskall, line 18: `coalesce` wasn't working because environment and repo are never null/empty since they have a default value.

testing/benchmark/main.tf
```diff
@@ -9,23 +9,22 @@ terraform {
...
+    environment  = var.ENVIRONMENT
+    repo         = var.REPO
```
kruskall, line 24: repo is basically a duplicate of the team tag in the case of apm-server. Is it worth adding it ?

testing/benchmark/main.tf
```diff
@@ -38,9 +37,6 @@ provider "ec" {}
```
kruskall, line 37:
```terraform
module "tags" {
  source  = "../infra/terraform/modules/tags"
  project = "benchmarks"
}
```

I'm not sure about this. We're setting `project = benchmarks`, using the convention for `a team/shared owned resource` but benchmarks are actually always individually owned. Am I missing something ?

marclop: That makes sense when the benchmark is launched locally, but if it's launched as a CI run, then that doesn't hold anymore since some resources may be left dangling, and we own it as a team since the CI scheduled run can't own that.

testing/infra/terraform/modules/benchmark_executor/instance.tf
```diff
@@ -30,18 +42,17 @@ module "vpc" {
...
 
-  tags = {
-    Owner       = var.user_name
-    Environment = "dev"
-  }
+  tags = merge(local.ci_tags, var.tags)
+
   vpc_tags = {
-    Name = "vpc-${var.user_name}-worker"
+    name = "vpc-${var.user_name}-worker"
```
marclop, line 48: `Name` is a special tag that is used for resource naming, it should have the Title format.

```suggestion
    Name = "vpc-${var.user_name}-worker"
```

marclop (approved):

testing/benchmark/Makefile
```diff
@@ -54,7 +54,7 @@ init-aws-profile: ~/.aws/credentials
 	@touch $@
 
 terraform.tfvars:
-	@sed "s/USER/$(USER)/" $(TFVARS_SOURCE) > terraform.tfvars
+	@cp $(TFVARS_SOURCE) terraform.tfvars
```
marclop, line 57: This will break `make terraform.tfvars` locally, since the current user won't be used for prefixes, I'd rather keep it as it was.

testing/benchmark/main.tf
kruskall: I've pushed some changes to account for CI! Let me know what you think!

testing/benchmark/Makefile
kruskall: Updated!

testing/infra/terraform/modules/benchmark_executor/variables.tf
```diff
@@ -33,6 +33,12 @@ variable "private_key" {
   type    = string
 }
 
+variable "tags" {
+  type        = map(string)
+  default     = {}
+  description = "Optional set of tags to use for all deployments"
```
marclop, line 39:
```suggestion
  description = "Optional set of tags to use for all resources"
```

testing/infra/terraform/modules/benchmark_executor/instance.tf
```diff
@@ -42,6 +52,7 @@ module "vpc" {
 resource "aws_key_pair" "worker" {
   key_name   = "${var.user_name}_worker_key"
   public_key = file(var.public_key)
+  tags       = var.tags
```
marclop, line 55: I think all AWS resources should have the tags set in `ec2_tags`. Given that, it would be easier to define the `provider "aws"` block within this module, so all created resources have the same tags consistently. This will already happen in `testing/benchmark` since the module is consumed by that folder and we define a provider block, however, for other users, we can enforce the tag standard within the module consistently.

marclop (commented):
We should set `default_tags` in `testing/benchmark`, and the `benchmark_executor` module, so other importers of this module also have the tags set for them.

testing/infra/terraform/modules/benchmark_executor/instance.tf
kruskall: I've added `ec2_tags` to all resources

testing/benchmark/main.tf
```diff
@@ -9,38 +9,36 @@ terraform {
...
+  project = startswith(var.user_name, "benchci") ? "benchmarks" : var.user_name
 }
 
 provider "ec" {}
 
 provider "aws" {
   region = var.worker_region
-  default_tags {
-    tags = merge(local.ci_tags, module.tags.tags)
-  }
```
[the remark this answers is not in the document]
kruskall: I tired to use `default_tags` and it's only applying to the `aws_key_pair`: https://github.com/elastic/apm-server/blob/main/testing/infra/terraform/modules/benchmark_executor/instance.tf#L42

It seems the other resources are using a different provider (https://registry.terraform.io/modules/terraform-aws-modules/vpc/aws/latest for example) so we need to set tags explicitly.

Let me know what you think!

testing/benchmark/main.tf
```diff
@@ -9,40 +9,34 @@ terraform {
...
 }
 
 provider "ec" {}
 
-provider "aws" {
-  region = var.worker_region
-  default_tags {
-    tags = merge(local.ci_tags, module.tags.tags)
-  }
-}
```
marclop, line 44: Did you mean to remove the entire `provider` block? AFAIK we still need it so we can specify the worker region. Before we merge, can you make sure trigger the benchmarks and that they pass? It should also be a good way to see the exact tags that are applied.

kruskall: Good catch! I've added back the block.

I've also confirmed that the terraform script are working and the benchmarks run :+1:

marclop:
I triggered the benchmarks on the CI and they're currently failing, can you please take a look?

kruskall:
> I triggered the benchmarks on the CI and they're currently failing, can you please take a look?

@marclop Thanks for that! It seems ci was using an old version of terraform. I've updated the version

marclop:
Please don't merge the PR until the `apm-server-benckmarks/pr-merge` check is passing.

kruskall:
> Please don't merge the PR until the apm-server-benckmarks/pr-merge check is passing. If we do, we'll break the daily benchmarks. https://github.com/orgs/elastic/teams/observablt-ci Could check why the updated Terraform version is not being used in the CI runner?

@marclop Thanks for mentioning this! I've pinged them on slack, see the related channel for the root cause. I think the issue is solved now :+1:

kruskall:
`apm-server-benckmarks/pr-merge` is green!

Thank you @v1v for the super fast response! :bow:

PR #9960, opened 2023-01-02
servicemetrics: Add `10m`, `60m` aggregation intervals

### LINKED ISSUE #9703: Add additional aggregation intervals for metrics

marclop:
## Motivation/summary

Modifies the `servicemetrics` aggregator to support extra aggregation
intervals by introducing a new `RollUpIntervals` (internal only) config
option. The configuration option is not user facing, but rather allow us
to configure smaller intervals for testing purposes. Also removing the
`aggregation.service.interval` config option.

Two new intervals are added and published by default, 10 and 60m (1h).
and these are flushed when the APM Server is stopped or restarted.

Introduces 3 `metrics-apm.service.${interval}m` datastreams
which contain the mapping definitions for transaction metrics with a
series of changes to the apmpackage generator to allow interval
templating and individual ILM policies to be tied to each interval.
A new `Metricset.Interval` field is added to the ingested documents, to
allow them to contain the aggregated interval they represent. The field
format is `${interval}m` or `${interval}s` if < 60s.

Older APM Servers sending data to an Elasticsearch cluster with updated
APM Package, will continue indexing metrics to `metrics-apm.internal`.

## Checklist

## How to test these changes

See related issue.

## Related issues

Closes #9703


apmpackage/apm/data_stream/service_interval_metrics/manifest.yml
```diff
@@ -0,0 +1,37 @@
+title: APM service metrics {{ .Interval }}
```
lahsivjar, line 1: Can this be processed by the `apmpackage/genpackage` command? I don't see those changes in the PR.

x-pack/apm-server/aggregation/servicemetrics/aggregator.go
```diff
@@ -184,21 +200,23 @@ func (a *Aggregator) processTransaction(event *model.APMEvent) {
 	if event.Transaction == nil || event.Transaction.RepresentativeCount <= 0 {
 		return
 	}
-	key := makeAggregationKey(event, a.config.Interval)
-	metrics := makeServiceMetrics(event)
-	a.active.storeOrUpdate(key, metrics, a.config.Interval, a.config.Logger)
+	for _, interval := range a.Intervals {
```
lahsivjar, line 203: I wonder if there is any point in doing this concurrently. Do we have some benchmarking data on this?

marclop: I haven't done any benchmarking on it, but given the locking strategy and how this function is called, I fear we'd have goroutine explosion, since this is called on each batch that is processed, and we acquire the full `metricsBuffer` lock for each event we process.

apmpackage/apm/data_stream/service_interval_metrics/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,22 @@
...
+      name: process_ppid
+  - pipeline:
+      name: client_geoip
+  - remove:
+      field: _metric_descriptions
+      ignore_missing: true
+  - remove:
+      # Removed in 8.6.0
+      field: timeseries.instance
+      ignore_missing: true
```
simitt, line 22: Do we really need all these pipeline processors? E.g. why the `process_ppid`, `user_agent` and `client_geoip`, or the two remove processors?

simitt (approved):
Apart from the pipeline definition, LGTM. 

Afaik, this has been the last in a row. Looking forward to some benchmarking results under high load. Might also be worth to pull in the UI team for collaborating on benchmarks.

apmpackage/apm/data_stream/service_interval_metrics/elasticsearch/ingest_pipeline/default.yml
marclop: good catch, we don't need them. I'll remove the ones that are not needed.

Issue #9644, opened 2022-11-22
PoC: de-obfuscate Android stack traces


simitt:
For support of the mobile APM android agent, it is crucial to support de-cyphering obfuscated stack traces. The Elastic APM solution already offers an [API for uploading sourcemaps](https://www.elastic.co/guide/en/kibana/master/rum-sourcemap-api.html) that should be reused for stack traces sent via the Android agent. 
The APM Server fetches the sourcemaps and applies them to incoming event stack traces. The internal sourcemap handling is [subject to change](https://github.com/elastic/apm-server/issues/9643), but changes won't be user facing and they are independent of this task. 

This issue focuses on implementing the logic to de-obfuscate the error stack traces by applying a sourcemap. The [specification for the ProGuard/ReTrace mapping file format](https://www.guardsquare.com/manual/tools/retrace) should be used for implementing the mapping logic. 

@LikeTheSalad will provide an example sourcemap and error event for defining the spec and testing. 

--- 

updated to provide more context and details (28/11/2022)

axw:
There is a specification for the ProGuard/ReTrace mapping file format at https://www.guardsquare.com/manual/tools/retrace (thanks @LikeTheSalad for the link)

gbamparop:
Are the mapping files expected to be stored to ES like the sourcemaps (which will be moved there for 8.7)? If so, will we need another API in Kibana?

LikeTheSalad:
Hi! I just wanted to mention here that we found the source code for Google's tool (R8 Retrace) here: https://r8.googlesource.com/r8/+/refs/heads/main/src/main/java/com/android/tools/r8/retrace/Retrace.java - In case it can help.

Also, this is the folder with the stacktrace examples: https://drive.google.com/drive/folders/1oA66ADF3kMP4JvjHTIyPbFcuXbT0FKlz?usp=sharing

simitt:
>Are the mapping files expected to be stored to ES like the sourcemaps (which will be moved there for 8.7)? If so, will we need another API in Kibana?

We are planning on reusing the existing API endpoint and using the same ES index. The one part that is unclear from my perspective at this point is whether any of the currently used fields for the sourcemap key need to change/be extended, especially the `bundle_filepath`. @LikeTheSalad @AlexanderWert have probably more knowledge here.

AlexanderWert:
@simitt From our understanding, for mobile we only need the fields `service.name` and `service.version`. We don't see a use case for `bundle_filepath`.

[...]

PR #9980, opened 2023-01-05
feat: add package to de-obfuscate Android stack traces

kruskall:
## Motivation/summary

Add package to parse and map android stack traces
Add tests

## Checklist

## Related issues

Closes https://github.com/elastic/apm-server/issues/9644


internal/retrace/parse.go
```diff
@@ -0,0 +1,213 @@
...
+
+type parser struct {
+	// used to keep track of the current class
+	// when descending into the method names.
+	currentClass string
+}
+
+// Mapping holds the information stored in the mapping file.
+type Mapping struct {
+	// obfuscated class name -> original class metdata
```
carsonip, line 35:
```suggestion
	// obfuscated class name -> original class metadata
```

internal/retrace/parse.go
```diff
@@ -0,0 +1,213 @@
...
+		if err != nil {
+			return fmt.Errorf("failed to parse obfuscatedStartLine: %w", err)
+		}
+	}
+
+	originalStartLine := 0
+	// check if we have the original line numbers
+	originalMethodSignature, originalLineNumbers, ok := bytes.Cut(originalMethodInfo, []byte(":"))
+	if ok {
+		orgiginalStartLineStr, _, _ := bytes.Cut(bytes.TrimPrefix(originalLineNumbers, []byte(":")), []byte(":"))
```
carsonip, line 186:
```suggestion
		originalStartLineStr, _, _ := bytes.Cut(bytes.TrimPrefix(originalLineNumbers, []byte(":")), []byte(":"))
```

internal/retrace/parse.go
```diff
@@ -0,0 +1,213 @@
...
+
+		var err error
+		originalStartLine, err = strconv.Atoi(string(orgiginalStartLineStr))
```
carsonip, line 189:
```suggestion
		originalStartLine, err = strconv.Atoi(string(originalStartLineStr))
```

carsonip (commented):
still reading the PR but want to point out some minor typos first

simitt (commented):
PR looks good. 

The testdata usecases are generally great, but I was wondering if the whole mapping file is required, or whether it could be reduced in size without losing value?
Ideally we could also use the mapping file to more easily verify that every part of the spec is implemented & tested. 
If that's not easily achievable, you could also look at adding a dedicated parser test, using a hand crafted, small mapping file that shows that the spec is fully implemented.

internal/retrace/map.go
```diff
@@ -0,0 +1,168 @@
...
+	sb.WriteRune(' ')
+	sb.WriteString(originalClassName)
+	sb.WriteRune('.')
+	sb.WriteString(originalMethodNames[0].name)
+	sb.WriteRune('(')
+	sb.WriteString(sourceFile)
+	if hasLineNumber {
+		sb.WriteRune(':')
+		if originalMethodNames[0].line == 0 {
+			sb.WriteString(strconv.Itoa(obfuscatedLineNumber))
```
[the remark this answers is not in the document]
kruskall: Done!

internal/retrace/map.go
```diff
@@ -0,0 +1,168 @@
...
+
+	// trim spaces to avoid issues with editors
+	sourceFileInfo = strings.TrimSpace(sourceFileInfo)
+
+	// validate that the source file info is not malformed
+	if !strings.HasSuffix(sourceFileInfo, ")") {
+		return errors.New("missing ')' token suffix in source file info")
+	}
+
+	obfuscatedLineNumber := 0
```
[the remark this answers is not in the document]
kruskall: The line number is part of the Stacktrace, that's a java object which is not part of the retrace specification. The equivalent document would be the JLS (Java Language Specification) or the jvm specification. AFAIK line numbers start from 1.

The following (keep it on the same/first line):

```java
public class TestLineNumber {public static void main(String args[]) {System.out.println("Line Number: "+Thread.currentThread().getStackTrace()[1].getLineNumber());}}
```

Should produce `Line Number: 1`

kruskall:
> The testdata usecases are generally great, but I was wondering if the whole mapping file is required, or whether it could be reduced in size without losing value?

Good point! I've compressed them. Size went from 7Mb to 600K.

internal/retrace/map.go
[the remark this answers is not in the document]
kruskall: > The line number is part of the Stacktrace, that's a java object which is not part of the retrace specification. The equivalent document would be the JLS (Java Language Specification) or the jvm specification. AFAIK line numbers start at 1.
> 
> The following (keep it on the same/first line):
> 
> ```java
> public class TestLineNumber {public static void main(String args[]) {System.out.println("Line Number: "+Thread.currentThread().getStackTrace()[1].getLineNumber());}}
> ```
> 
> Should produce `Line Number: 1`

It surely does. Unfortunately retrace does not care and uses zero line numbers just fine. :sob: 

I've update the code to use -1 (if that's a valid line number whoever designed the language is messing with us).

internal/retrace/parse.go
```diff
@@ -199,6 +202,9 @@ func (p *parser) parseMethodline(m *Mapping, line []byte) error {
 	originalMethodName := originalMethodSignature[:argStart]
 
 	k := methodID{
+		// Always use startline in the key.
+		// If the line numbers are not present (either in the mapping or the stacktrace)
+		// a methoid with a line 0 is going to be used.
```
simitt, line 210:
```suggestion
		// a methodid with a line 0 is going to be used.
```

internal/retrace/parse.go
```diff
@@ -193,6 +193,9 @@ func (p *parser) parseMethodline(m *Mapping, line []byte) error {
 		if err != nil {
 			return fmt.Errorf("failed to parse originalStartLine: %w", err)
 		}
+	} else {
+		// no inlined method, the leading line number are the original line numbers.
```
simitt, line 197:
```suggestion
		// no inlined method, the leading line numbers are the original line numbers.
```

simitt (approved):
Looks great, thanks! Only minor cosmetical suggestions.

PR #10002, opened 2023-01-10
ci: Allow system-tests check to be set as required

marclop:
## Motivation/summary

Follows GitHub's guidance on how to create a step with the same name as the required action so that a status check may be set as required in a protected branch.

see: https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/troubleshooting-required-status-checks#handling-skipped-but-required-checks

### DIFF
--- a/.github/workflows/ci-docs.yml
+++ b/.github/workflows/ci-docs.yml
@@ -6,6 +6,7 @@ on:
     # this workflow when code changes also contain docs changes:
     # https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-excluding-paths
     paths-ignore: # This expression needs to match the paths ignored on ci.yml.
+      - '**' # Ignore all file changes, except for the below.
       - '!**.md'
       - '!**.asciidoc'
       - '!**.png'
--- a/.github/workflows/ci-docs.yml
+++ b/.github/workflows/ci-docs.yml
@@ -1,0 +1,19 @@
+name: ci
+
+on:
+  pull_request:
+    paths: # This expression needs to match the paths ignored on ci.yml.
+      - '**.md'
+      - '**.asciidoc'
+      - '**.png'
+      - '**.svg'
+
+# limit the access of the generated GITHUB_TOKEN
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-docs.yml
```diff
@@ -0,0 +1,19 @@
+name: ci
+
+on:
+  pull_request:
+    paths: # This expression needs to match the paths ignored on ci.yml.
```
axw, line 5: Should we be excluding any PR that has files that don't match the below?

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore says

> If at least one path matches a pattern in the paths filter, the workflow runs.

So if we had a PR with a `.go` file as well as a `.md` file, then this would run, right?

Maybe we should use `paths-ignore` with negated patterns?

.github/workflows/ci-docs.yml
```diff
@@ -0,0 +1,19 @@
+name: ci
```
axw, line 1:
```suggestion
name: ci-docs
```

marclop: unfortunately, the docs explicitly state that the names must be the same for the checks to match:

> Make sure that the name key and required job name in both the workflow files are the same. For more information, see "[Workflow syntax for GitHub Actions](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions)".

axw: Ah sorry, I did not RTFM

.github/workflows/ci-docs.yml
marclop: that's true, I will change to `paths-ignore`. Thanks for catching this.

marclop: I will open a PR with changes in a few places after we merge this one so we can ensure things work as we expect them to work.

axw (approved):

PR #9818, opened 2022-12-15
packaging: disable --strict.perms for Docker

axw:
## Motivation/summary

Disable strict permission checks in Docker and make the config file world-readable and data directory world-writeable. This permits running the image with arbitrary UIDs and GIDs.

## Checklist

~- [ ] Update [CHANGELOG.asciidoc](https://github.com/elastic/apm-server/blob/main/CHANGELOG.asciidoc)~
~- [ ] Update [package changelog.yml](https://github.com/elastic/apm-server/blob/main/apmpackage/apm/changelog.yml) (only if changes to `apmpackage` have been made)~
~- [ ] Documentation has been updated~

## How to test these changes

```bash
make package-docker
docker run --rm --user 123:456 $(cat build/docker/apm-server-ubi8-8.7.0.txt) /usr/share/apm-server/apm-server keystore create
```

## Related issues

https://github.com/elastic/cloud-on-k8s/issues/6227

PR #10011, opened 2023-01-10
fix: change docker group to not use root group

BlackDark:
## Motivation/summary

Because currently the images require the root group `0` maybe this could be an easy fix to be able to run on the group `1000` group without problems?

## Checklist


axw:
@BlackDark sorry for the silence. https://github.com/elastic/apm-server/pull/9818 went into 8.6.0, which was released a couple of days after you opened this PR. Could you please try that version and let us know if you still have issues?

BlackDark:
As best practice it is always best to not use any root related groups or users.
Even if the change works you will have to manually adjust each manifest in kubernetes etc to be able to run the image in hardened environments where you restrict root access.
My change changes the default group of the user to 1000 which is no root and as such should work out of the box.

axw:
@BlackDark fair enough. I will take another look at this next week. We had some assumptions about the group being root in the past. I believe those have all been fixed, but would like to verify first.

axw (approved):
Thanks @BlackDark! LGTM.

BlackDark:
Is also compatible as backport-7.x

axw:
> Is also compatible as backport-7.x
> If I can help with that just ping me.

For 7.17.x we rely on the Dockerfile template provided by beats: https://github.com/elastic/beats/blob/main/dev-tools/packaging/templates/docker/Dockerfile.tmpl. I don't think it would be straightforward to backport, but if you have a need for this in 7.x then you're welcome to send a PR and somebody will review. Otherwise I think we'll leave it.

PR #10013, opened 2023-01-10
Fix a panic caused by accepting new grpc connections during shutdown

carsonip:
## Motivation/summary

There is a possible panic related to gmux. It is due to the order we shutdown httpServer and grpcServer. Occasionally when a new grpc connection comes in after grpcServer shutdown and before httpServer shutdown, the program panics.
See https://github.com/elastic/gmux/issues/13

## Checklist


simitt (approved):
Great finding!

Issue #9654, opened 2022-11-23
service metrics: extend to non-transaction events and add zero counts


simitt:
Service metrics should also be produced when application metrics, errors & logs for a service are observed. When during an aggregation period no transaction events are seen for this service, then a zero count for transaction throughput needs to be stored with the service metric.

simitt:
Any non trace event should trigger the creation of a metricset, but with zero values for the `Metricset` and `Transaction` counts. The aggregation logic needs to be changed to not only process transaction events.
What's problematic is that the aggregation key contains the `transaction.type`. The type is not known if no transaction event is available. @felixbarny does it make sense to create the service metrics with `transaction.type: "unknown"` for the UI queries?

felixbarny:
Good question, Silvia. I don't have a clear answer but a few thoughts.

> The aggregation logic needs to be changed to not only process transaction events.

This is a bit at odds with https://github.com/elastic/apm-server/issues/5243#issuecomment-1345504925

> I would strongly prefer that service metrics remain a subset of transaction metrics so that we could later derive them in Elasticsearch by just dropping dimensions.

Maybe we should have another metricset (or metricsets) for different signals that summarize, for example, the rate of metrics, logs, errors, and maybe the number of profiling frames for a service. These metricsets don't have a `transaction.type` dimension. 

Therefore, the closest thing to an answer to your question

> does it make sense to create the service metrics with transaction.type: "unknown" for the UI queries?

is that there shouldn't be a `transaction.type` dimension for non-trace-related service metrics.

Another issue is that not all signals are necessarily passing through APM Server (for example logs that are shipped by Filebeat). So if we only rely on service metrics in the service inventory UI (which is the premise of this issue), we might miss some services. I think it's important to show services that only have logs, for example. We could query the logs index in addition to the service metrics but the risk is that it's going to be very slow.

cc @dgieselaar

felixbarny:
To make a concrete suggestion, APM server could emit a service metric document that only has a `service.name` and a `service.environment` dimension and nothing else if it receives any signal from a service. 

The document may look like this (note the lack of any numerical metric value):

```json
{
  "@timestamp": "2006-01-02T15:04:05.000Z",
  "metricset.name": "service",
  "service": {
    "environment": "prod",
    "name": "systemtest"
  }
}
```

This would be in addition to the service metrics that have a `transaction.duration` metric and a `transaction.type` dimension. Alternatively, we could emit these metrics only if there are no service-transaction metrics.

[...]

PR #10029, opened 2023-01-12
Add service summary metrics aggregator

carsonip:
## Motivation/summary

Add a service summary metrics aggregator which aggregates `agent.name`, `service.environment`, `service.name`, `labels` and `numeric_labels` over configured intervals from all events except spans. The aggregation writes to a new data stream `metrics-apm.service_summary.${interval}` with intervals `1m`, `10m` and `60m`.

## Checklist

## How to test these changes

- Ensure service_summary data stream has the aggregated document on receiving any events

## Related issues

Part of #9654


x-pack/apm-server/aggregation/servicesummarymetrics/aggregator.go
```diff
@@ -0,0 +1,264 @@
...
+	return a.config.BatchProcessor.ProcessBatch(ctx, &batch)
+}
+
+// ProcessBatch aggregates all service summary metrics.
+func (a *Aggregator) ProcessBatch(ctx context.Context, b *model.Batch) error {
+	a.mu.RLock()
+	defer a.mu.RUnlock()
+	for _, event := range *b {
+		// Not filtering on Processor.
+		a.processEvent(&event, b)
```
carsonip, line 149: Are there any events that we need to exclude? There is absolutely 0 filtering now

x-pack/apm-server/main.go
```diff
@@ -150,6 +151,19 @@ func newProcessors(args beater.ServerParams) ([]namedProcessor, error) {
...
 		processors = append(processors, namedProcessor{name: serviceName, processor: serviceAggregator})
+
+		const serviceSummaryName = "service summary aggregation"
+		args.Logger.Infof("creating %s with config: %+v", serviceSummaryName, args.Config.Aggregation.Service)
+		serviceSummaryAggregator, err := servicesummarymetrics.NewAggregator(servicesummarymetrics.AggregatorConfig{
+			BatchProcessor:  args.BatchProcessor,
+			Interval:        metricsInterval,
+			RollUpIntervals: rollUpMetricsIntervals,
+			MaxGroups:       args.Config.Aggregation.Service.MaxGroups,
+		})
```
carsonip, line 162: Reusing service metrics config `enabled` and `max_groups` here. Is this fine or should we introduce a new set of config for service summary?

x-pack/apm-server/aggregation/servicesummarymetrics/aggregator.go
```diff
@@ -0,0 +1,264 @@
...
+		if !a.active[period].storeOrUpdate(key, a.config.Logger) {
+			intervalStr := interval.FormatDuration(period)
+			*b = append(*b, makeMetricset(key, intervalStr))
+		}
+	}
+}
+
+type metricsBuffer struct {
+	mu sync.RWMutex
+	m  map[uint64]aggregationKey
```
carsonip, line 166: pending discussion on hash collision and whether to use a chained hash table like service metrics

x-pack/apm-server/aggregation/servicesummarymetrics/aggregator.go
simitt: strictly thinking we could ignore spans, they are only adding noise and no value, but they do not increase cardinality, so also not a big issue to not have a filter at all

x-pack/apm-server/main.go
```diff
@@ -104,7 +105,7 @@ type processor interface {
 // newProcessors returns a list of processors which will process
 // events in sequential order, prior to the events being published.
 func newProcessors(args beater.ServerParams) ([]namedProcessor, error) {
-	processors := make([]namedProcessor, 0, 3)
+	processors := make([]namedProcessor, 0, 5)
```
simitt, line 108: nice catch

x-pack/apm-server/main.go
simitt: clarified on the issue - fine to reuse for now

simitt (commented):
PR generally looks good to me. There is no overflow handling if the max groups limit is exceeded, @dgieselaar @felixbarny would you expect this to be aligned with the newly introduced overflowbucket `_other` for transactions and service transaction metrics? (would that even be useful for the UI)?

simitt:
@carsonip can you add the logic for the overflow bucket to this PR?

x-pack/apm-server/aggregation/servicesummarymetrics/aggregator.go
axw: We definitely need to handle collisions. How that's implemented is less important, but we should probably be consistent and use separate chaining.

simitt:
@carsonip having an indicator without an actual aggregation is fine for now, as the next step will be to add the counters per event type.

carsonip:
> @carsonip having an indicator without an actual aggregation is fine for now, as the next step will be to add the counters per event type. I'd also be fine to do this in a dedicated PR though, as you are right that it makes much more sense once the counters are added.

In that case, let's do it in a dedicated PR. Currently service summary will add the event to the batch without aggregating on overflow, similar to what's in service destination metrics.

x-pack/apm-server/aggregation/servicesummarymetrics/aggregator.go
carsonip: Done, thanks.

simitt (approved):
Changes look good. 

Please note on the related issue which parts are implemented with this PR and which parts are left open.

PR #10034, opened 2023-01-13
Universal Profiling: Fix for GRPC compression not being enabled

christos68k:
If this import does not exist, the backend will drop compressed RPCs as `Unimplemented`.

Was previously removed in:
https://github.com/elastic/apm-server/commit/06b1a97767942ce52e65cd07a80980679c1198ba

```
internal/otel_collector/config/configgrpc/gzip.go:19:	_ "google.golang.org/grpc/encoding/gzip"
internal/otel_collector/config/configgrpc/configgrpc.go:34:	"google.golang.org/grpc/encoding/gzip"
```

**TODO:**

- ~Looking into updating tests~ DONE

### DIFF
--- a/systemtest/profiling_test.go
+++ b/systemtest/profiling_test.go
@@ -27,12 +27,18 @@ import (
 	"github.com/stretchr/testify/require"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials/insecure"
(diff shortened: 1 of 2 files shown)


axw (approved):
Thank you :pray:

PR #10049, opened 2023-01-16
action: export traces with the OTEL GH action

v1v:
## Motivation/summary

Back in the days we built the Jenkins OTEL integration, so let's use the same approach for the GH actions, so every workflow and step will be tracked as traces/spans.

## How to test these changes

See traces in the Elastic deployment

This has been already tested for quite sometime in some other projects

image

https://github.com/elastic/apm-pipeline-library/blob/main/.github/workflows/opentelemetry.yml

### DIFF
--- a/.github/workflows/ci-docs.yml
+++ b/.github/workflows/ci-docs.yml
@@ -1,4 +1,4 @@
-name: ci-docs
+name: ci
 
 on:
   pull_request:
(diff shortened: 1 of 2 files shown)


reakaleek (approved):

kruskall (approved):

.github/workflows/ci-docs.yml
```diff
@@ -1,4 +1,4 @@
-name: ci
+name: ci-docs
```
axw, line 1: The name needs to be "ci" -- see https://github.com/elastic/apm-server/pull/10002#discussion_r1065297736

axw (changes requested):

PR #10061, opened 2023-01-17
Add overflow bucket to service summary

carsonip:
## Motivation/summary

Add overflow bucket to service summary to align with transaction metrics and service metrics.

## Checklist

~~- [] Documentation has been updated~~

## How to test these changes

- On overflow, check for document that has field `service_summary.aggregation.overflow_count`

## Related issues

Part of #9654


x-pack/apm-server/aggregation/servicesummarymetrics/aggregator.go
```diff
@@ -177,44 +200,61 @@ func newMetricsBuffer(maxSize int) *metricsBuffer {
...
+			mb.other = &mb.space[len(mb.space)-1]
+			mb.otherCardinalityEstimator = hyperloglog.New14()
 		}
-		mb.m[hash] = append(mb.m[hash], &key)
+		mb.otherCardinalityEstimator.InsertHash(hash)
+		return
+	} else if mb.entries == mb.maxSize/2-1 {
+		logger.Warn("service summary groups reached 50% capacity")
+	} else if mb.entries == mb.maxSize-1 {
+		logger.Warn("service summary groups reached 100% capacity")
```
simitt, line 253: We are not issuing these warnings anymore for the service metrics or the transaction metrics. Can you align the overflow logic where possible with these aggregations?

simitt (changes requested):
Changes largely look good, just a bit more alignment with how this is handled in the other aggregations.

carsonip:
@simitt @axw I really like how the `mb.space` backing array was implemented in service metrics to enable sequential memory access and reduce gc pressure. Given that the cardinality may be lower in service summary metrics, are we worried about reserving `maxGroups * size(aggregationKey)` in memory while most of it may be unused?

simitt:
@carsonip could you provide micro-benchmarks on both options? 
What's the additionally reserved memory for the default values for a 1GB instance?

carsonip:
> @carsonip could you provide micro-benchmarks on both options? What's the additionally reserved memory for the default values for a 1GB instance?

sizeof(aggregationKey{}) is 136B, maxGroups is configurable. e.g. maxGroups = 1000, the size of the backing array will be ~1.36KB. Given maxGroups is auto-sized using `maxGroupsForAggregation`, for an APM server with 1GB mem limit, maxGroups will be 10000 and the backing array will therefore be ~13.6KB. With 64GB or more mem limit, maxGroups will be autosized to 640000, which is 870KB. This is already the existing behavior for service metrics.

Side note: Since map entry contains labels, which contains slices, we may retain references to more objects on the heap and keeping more memory in use than just 136B.

With backing array mb.space, benchmarking `processEvent`:
```
goos: linux
goarch: amd64
pkg: github.com/elastic/apm-server/x-pack/apm-server/aggregation/servicesummarymetrics
cpu: 11th Gen Intel(R) Core(TM) i7-11800H @ 2.30GHz
BenchmarkProcessEvent
BenchmarkProcessEvent-16    	 2826516	       437.3 ns/op	     473 B/op	       5 allocs/op
PASS
```

Without backing array and just keeping addresses of structs on the heap:
```
goos: linux
goarch: amd64
pkg: github.com/elastic/apm-server/x-pack/apm-server/aggregation/servicesummarymetrics
cpu: 11th Gen Intel(R) Core(TM) i7-11800H @ 2.30GHz
BenchmarkProcessEvent
BenchmarkProcessEvent-16    	 2260258	       531.4 ns/op	     905 B/op	       8 allocs/op
PASS
```

IMO, if it doesn't seem wise to reserve memory for this purpose, we can implement the one without backing array first, and optimize it later if it becomes an issue.

carsonip:
@simitt I've removed the mb.space backing array optimization for now to unblock it first. We can add optimizations later on.

axw:
Reserving 13.6KB on a 1GB machine, and 870KB on a 64GB machine, is not even remotely concerning to me. Generating a lot of garbage is. Nevertheless, I'm fine with keeping it simple to start with and optimising later - it's fairly insignificant in the grand scheme of things.

simitt:
I fully agree with @axw these numbers are very small and I'd prefer movign forward with the backing array.

simitt (approved):

carsonip:
Updated my math on memory usage because it was very wrong. But let's merge the changes in first and optimize the performance later. There are other aggregators that can use this optimization as well.

marclop:
@carsonip I tried to test this with 8.7.0 BC2 on ESS but it doesn't seem to be working.

I noticed that there are some error messages around the time when I was triggering the overflow that make me think that we forgot to remove the `_metric_descriptions` field. https://github.com/elastic/apm-server/blob/7c365e460b3cf229cb1c9eeeb077cc6e88f6ef29/apmpackage/apm/data_stream/service_transaction_interval_metrics/elasticsearch/ingest_pipeline/default.yml#L10-L12

vs

https://github.com/elastic/apm-server/blob/7c365e460b3cf229cb1c9eeeb077cc6e88f6ef29/apmpackage/apm/data_stream/service_summary_interval_metrics/elasticsearch/ingest_pipeline/default.yml#L1-L9

```
failed to index document (strict_dynamic_mapping_exception): mapping set to strict, dynamic introduction of [_metric_descriptions] within [_doc] is not allowed
```

I used the following program, as `main.go`, and also the script below as `test.sh`.

```go
package main

import (
	"bytes"
	"encoding/binary"
	"fmt"
	"io"
	"log"
	"math/rand"
	"net/http"
	"os"
	"runtime"
	"strconv"
	"sync"
	"time"

	"go.elastic.co/apm/v2/model"
	"go.elastic.co/fastjson"
)

var host = os.Getenv("ELASTIC_APM_SERVER_URL")
var auth = os.Getenv("ELASTIC_APM_SECRET_TOKEN")

var (
	language = model.Language{Name: "go", Version: runtime.Version()}
	rtime    = model.Runtime{Name: runtime.Compiler, Version: runtime.Version()}
	agent    = model.Agent{Name: "go", Version: "2.2.0"}
)

func main() {
	txGroups := 10
	if groups := os.Getenv("TX_GROUPS"); groups != "" {
		if v, _ := strconv.ParseInt(groups, 10, 64); v > 0 {
			txGroups = int(v)
		}
	}
	services := 1
	if groups := os.Getenv("SERVICES"); groups != "" {
		if v, _ := strconv.ParseInt(groups, 10, 64); v > 0 {
			services = int(v)
		}
	}
	var frequency time.Duration = 0 // <= 0 means send once and exit
	if str := os.Getenv("FREQUENCY"); str != "" {
		if v, err := time.ParseDuration(str); err == nil {
			frequency = v
		}
	}

	var wg sync.WaitGroup
	log.Printf("starting load generator with frequency %s", frequency)

	for i := 0; i < services; i++ {
		wg.Add(1)
		go func(id int) {
			defer wg.Done()
			var ticker *time.Ticker
			if frequency > 0 {
				ticker = time.NewTicker(frequency)
			}
			for {
				if ticker != nil {
					<-ticker.C
				}

				var w fastjson.Writer
				var batchSize int
				for batchSize < 10 {
					batchSize = rand.Intn(100)
				}
				defer flushService(&w, auth)

				for i := 0; i < txGroups; i++ {
					if i%batchSize == 0 {
						if i > 0 {
							flushService(&w, auth)
						}
						writeMetadata(&w, id)
					}
					txID := i
					writeTransaction(&w, txID, id)
				}

				if ticker == nil {
					break
				}
			}
		}(i)
	}
	wg.Wait()
	log.Println("done")
}

func writeMetadata(json *fastjson.Writer, id int) error {
	json.RawString(`{"metadata":`)
	json.RawString(`{"service":`)
	svc := model.Service{
		Name:        fmt.Sprint("service-", id),
		Language:    &language,
		Runtime:     &rtime,
		Agent:       &agent,
		Environment: "dev",
		Node: &model.ServiceNode{
			ConfiguredName: "service-node",
		},
	}
	if err := svc.MarshalFastJSON(json); err != nil {
		return err
	}
	json.RawByte('}')
	json.RawString("}\n")
	return nil
}

func writeTransaction(w *fastjson.Writer, txID, id int) {
	t := true
	sr := 1.0
	tx := model.Transaction{
		Timestamp:  model.Time(time.Now()),
		Name:       fmt.Sprintf("tx-%d", txID),
		Type:       "txtype",
		Duration:   10 + float64(id),
		Outcome:    "success",
		Result:     "HTTP 2xx",
		Sampled:    &t,
		SampleRate: &sr,
	}
	binary.LittleEndian.PutUint64(tx.TraceID[:8], rand.Uint64())
	binary.LittleEndian.PutUint64(tx.TraceID[8:], rand.Uint64())
	binary.LittleEndian.PutUint64(tx.ID[:], rand.Uint64())
	w.RawString(`{"transaction":`)
	tx.MarshalFastJSON(w)
	w.RawString("}\n")
}

func flushService(w *fastjson.Writer, auth string) {
	buf := make([]byte, w.Size())
	copy(buf, w.Bytes())
	w.Reset()
	req, err := http.NewRequest(http.MethodPost,
		host+"/intake/v2/events",
		bytes.NewReader(buf),
	)
	if err != nil {
		panic(err)
	}
	req.Header.Add("Authorization", fmt.Sprintf("Bearer %s", auth))
	req.Header.Add("Content-type", "application/x-ndjson")
	res, err := http.DefaultClient.Do(req)
	if err != nil {
		panic(err)
	}
	defer res.Body.Close()
	if res.StatusCode >= 400 {
		io.Copy(os.Stdout, res.Body)
	}
}
```

```bash
#!/bin/bash

export ELASTIC_APM_SERVER_URL=${1:-"http://localhost:8200"}

export TX_GROUPS=1
export SERVICES=3000

go run main.go
```

carsonip:
@marclop thanks for testing this. I'll add a system test and make sure this is fixed.

carsonip:
Since service groups are autosized, unless we duplicate the autosizing in logic in system test, we can't rely on a const and be sure that it will overflow. The best approach imo will be to wait for #10207 for metrics on overflow, so that it is observable from the outside whether it has overflowed. @lahsivjar what do you think?

lahsivjar:
@carsonip I am not sure if I understand the issue correctly.

> Since service groups are autosized, unless we duplicate the autosizing in logic in system test, we can't rely on a const and be sure that it will overflow

Based on what I understand, in the systemtest, we can disable autoscaling by setting the specific configurations values to be `0`.

> so that it is observable from the outside whether it has overflowed

We can assert this reliably by checking for documents with `transaction.name: _other`

carsonip:
> Based on what I understand, in the systemtest, we can disable autoscaling by setting the specific configurations values to be a positive number.
> We can assert this reliably by checking for documents with `transaction.name: _other`

Ah, you're right. I'll proceed with setting the config and checking for `_other`. Thanks.

carsonip:
Fixing in #10302

marclop:
@carsonip I'm re-testing this and still seeing the same issue:

```
GET metrics-apm.service_summary*/_search?filter_path=hits.hits
{
  "query": {
    "match": {
      "service.name": "_other"
    }
  }
}
```
```
{
  "hits": {
    "hits": []
  }
}
```

The logs still show that error: `failed to index document (strict_dynamic_mapping_exception): mapping set to strict, dynamic introduction of [_metric_descriptions] within [_doc] is not allowed`

marclop:
There's a stale PR blocking the 8.7 branch updates https://github.com/elastic/kibana/pull/150647, that would be a good explanation as to why this isn't working with the latest BC (BC4).

marclop:
Verified with BC5 on ESS:

```
GET metrics-apm.service_summary.1m-default/_search?filter_path=hits.hits
{
  "query": {
    "match": {
      "service.name": "_other"
    }
  }
}
```
```json
{
  "hits": {
    "hits": [
      {
        "_index": ".ds-metrics-apm.service_summary.1m-default-2023.03.10-000001",
        "_id": "Cm_8yoYBePWpGPOHo5_2",
        "_score": 7.379257,
        "_source": {
          "@timestamp": "2023-03-10T10:05:00.000Z",
          "data_stream": {
            "dataset": "apm.service_summary.1m",
            "namespace": "default",
            "type": "metrics"
          },
          "ecs": {
            "version": "8.6.0-dev"
          },
          "event": {
            "agent_id_status": "missing",
            "ingested": "2023-03-10T10:06:40Z"
          },
          "metricset": {
            "interval": "1m",
            "name": "service_summary"
          },
          "observer": {
            "hostname": "c9e482615886",
            "type": "apm-server",
            "version": "8.7.0"
          },
          "processor": {
            "event": "metric",
            "name": "metric"
          },
          "service": {
            "name": "_other"
          },
          "service_summary": {
            "aggregation": {
              "overflow_count": 500
            }
          }
        }
      },
      {
        "_index": ".ds-metrics-apm.service_summary.1m-default-2023.03.10-000001",
        "_id": "c2_7yoYBePWpGPOHuY2N",
        "_score": 7.379257,
        "_source": {
          "@timestamp": "2023-03-10T10:04:00.000Z",
          "data_stream": {
            "dataset": "apm.service_summary.1m",
            "namespace": "default",
            "type": "metrics"
          },
          "ecs": {
            "version": "8.6.0-dev"
          },
          "event": {
            "agent_id_status": "missing",
            "ingested": "2023-03-10T10:05:40Z"
          },
          "metricset": {
            "interval": "1m",
            "name": "service_summary"
          },
          "observer": {
            "hostname": "c9e482615886",
            "type": "apm-server",
            "version": "8.7.0"
          },
          "processor": {
            "event": "metric",
            "name": "metric"
          },
          "service": {
            "name": "_other"
          },
          "service_summary": {
            "aggregation": {
              "overflow_count": 100
            }
          }
        }
      }
    ]
  }
}
```

PR #10066, opened 2023-01-18
feat: bump apm data for otel error mapping

kruskall:
## Checklist

## Related issues

Closes https://github.com/elastic/apm-server/issues/9659


axw (approved):
LGTM, please add changelog entries.

axw (commented):
Looks good. Something just occurred to me: did you already update apmpackage with new fields? I don't think we've defined `event.category`, for example.

kruskall:
> did you already update apmpackage with new fields? I don't think we've defined event.category, for example.

Good catch! Fixed :+1:

apmpackage/apm/data_stream/app_logs/fields/ecs.yml
```diff
@@ -82,6 +82,12 @@
   name: event.severity
 - external: ecs
   name: event.action
+- external: ecs
+  name: event.type
+- external: ecs
+  name: event.category
+- external: ecs
+  name: event.kind
```
axw, line 90: Field changes should be accompanied by an enhancement entry in apmpackage/apm/changelog.yml

axw (approved):
LGTM - please add an apmpackage changelog entry too. It's helpful for reviewers when you either tick or strike out the checklist boxes in the PR template - so we know if you intentionally skipped a step or not.

apmpackage/apm/data_stream/app_logs/fields/ecs.yml
kruskall: Good catch! Updated :+1:

kruskall:
> please add an apmpackage changelog entry too. It's helpful for reviewers when you either tick or strike out the checklist boxes in the PR template - so we know if you intentionally skipped a step or not.

Thanks for mentioning this, I'll keep that in mind :bow:

axw (approved):
LGTM, thanks for the updates :)

PR #10070, opened 2023-01-18
Hide non-default rollup interval data stream

### LINKED ISSUE #10047: Use hidden datastreams for aggregated metrics with a non-default rollup interval
Update Jan 23rd 2023:

Instead of adding a suffix to the `metricset.name`, we are moving forward with leveraging hidden datastreams for metrics with a non-default aggregation interval (other than 1minute). 
This should be applied to transaction, service and the new service summary metrics. 

Outdated: 
~According to https://github.com/elastic/kibana/issues/148507#issuecomment-1380830848, add a suffix `.internal` to the `metricset.name` for any rolled up metrics with a interval different than the `1m` default interval.~

~`.internal` is used instead of `-internal` to align with the [data stream naming scheme](https://www.elastic.co/blog/an-introduction-to-the-elastic-data-stream-naming-scheme) because data stream name is derived from metricset.name currently~

carsonip:
## Motivation/summary

Set data streams with non-default rollup intervals as hidden. This will avoid changing metricset.name to distinguish between default (1m) and non-default rollup intervals

## Checklist

## How to test these changes

- Check that 1m rollup is not hidden while 10m, 60m rollup interval data streams are hidden

## Related issues

Closes #10047


carsonip:
Blocked by https://github.com/elastic/package-spec/pull/465 and corresponding fleet changes at the moment

Issue #10097, opened 2023-01-24
APM Server panics on cgroup memory limit check


simitt:
**APM Server version** (`apm-server version`): 8.6.0

**Description of the problem including expected versus actual behavior**:

The APM Server panics if `stats.Memory.Mem.Limit.Bytes` is not available, so no cgroup limit is set. 

https://github.com/elastic/apm-server/blob/main/internal/beater/memlimit_cgroup.go#L58 should check if the limit exists, similar to how this is done for [v2](https://github.com/elastic/apm-server/blob/main/internal/beater/memlimit_cgroup.go#L64). 

**Steps to reproduce**:
 1. download, extract and start `apm-server-8.6.0-linux-x86_64.tar.gz`
 2. start server and observe panic

@kruskall could you verify that this panic on linux machines using cgroup `v1` please?

PR #10098, opened 2023-01-24
fix: make sure the memory subsystem is available before accessing the memory limit

kruskall:
## Checklist

## Related issues

Closes https://github.com/elastic/apm-server/issues/10097

### DIFF
--- a/internal/beater/memlimit_cgroup.go
+++ b/internal/beater/memlimit_cgroup.go
@@ -55,6 +55,9 @@ func cgroupMemoryLimit(rdr *cgroup.Reader) (uint64, error) {
 		if err != nil {
 			return 0, fmt.Errorf("unable to read cgroup limits: %w", err)
 		}
+		if stats.Memory == nil {
+			return 0, fmt.Errorf("cgroup memory subsystem unavailable")
+		}
 		return stats.Memory.Mem.Limit.Bytes, nil
 	case cgroup.CgroupsV2:
 		stats, err := rdr.GetV2StatsForProcess(pid)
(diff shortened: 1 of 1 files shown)


internal/beater/memlimit_cgroup.go
```diff
@@ -55,6 +55,9 @@ func cgroupMemoryLimit(rdr *cgroup.Reader) (uint64, error) {
 		if err != nil {
 			return 0, fmt.Errorf("unable to read cgroup limits: %w", err)
 		}
+		if stats.Memory == nil {
+			return 0, fmt.Errorf("cgroup memory subsystem unavailable")
+		}
```
simitt, line 60: What if `stats.Memory` is available, but `stats.Memory.Mem.Max.Bytes` isn't? Would aligning the [behavior with the cgroupsV2 logic](https://github.com/elastic/apm-server/pull/10098/files#diff-9465d4addb3bf8eac10da4a31232f2e5911ed5840bcf9afb3f024f81730448b2R67) make sense?

kruskall: > What if stats.Memory is available, but stats.Memory.Mem.Max.Bytes isn't?

`stats.Memory.Mem.Limit.Bytes` is an `uint64` so the value would be 0

> Would aligning the [behavior with the cgroupsV2 logic](https://github.com/elastic/apm-server/pull/10098/files#diff-9465d4addb3bf8eac10da4a31232f2e5911ed5840bcf9afb3f024f81730448b2R67) make sense?

`stats.Memory.Mem.Max.Bytes` is an `opt.Uint` so the type is different. I don't think we can use the same method but the behaviour should be the same.

kruskall: To verify: 

```go
s := &cgroup.StatsV1{}
s.Memory = &cgv1.MemorySubsystem{}
fmt.Println(s.Memory.Mem.Limit.Bytes) # prints 0
```

simitt (approved):

simitt (approved):

PR #10106, opened 2023-01-25
refactor: add more logging to sourcemap processing failure

kruskall:
## Motivation/summary

Debugging sourcemap issues has proven to be quite difficult during SDHs.

Add more log messages and make the message meaningful (e.g. add the requested path).
Remove logger ratelimit: the ratelimit is limiting a message to once every 60s, losing potentially important messages when debugging sourcemap issues.

## Checklist


simitt (approved):
The additional logs look good. I'm a bit concerned with removing the rate limiter, as sourcemap logging used to be quite verbose. But given the amount of time we spent trouble shooting on this topic, I'm ok with moving forward.

PR #10108, opened 2023-01-25
Add meaningful assertion message to TestAPMServerInstrumentation

carsonip:
Add meaningful assertion message to TestAPMServerInstrumentation so that we can understand the reason the next time it fails on GH actions

### DIFF
--- a/systemtest/instrumentation_test.go
+++ b/systemtest/instrumentation_test.go
@@ -89,8 +89,10 @@ func TestAPMServerInstrumentation(t *testing.T) {
 		if !ok {
 			continue
 		}
-		assert.Equal(t, transactionDoc.Trace.ID, traceID)
-		assert.Equal(t, transactionDoc.Transaction.ID, entry.Fields["transaction.id"])
+		assert.Equal(t, transactionDoc.Trace.ID, traceID,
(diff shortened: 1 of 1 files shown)


systemtest/instrumentation_test.go
```diff
@@ -89,8 +89,10 @@ func TestAPMServerInstrumentation(t *testing.T) {
 		if !ok {
 			continue
 		}
-		assert.Equal(t, transactionDoc.Trace.ID, traceID)
-		assert.Equal(t, transactionDoc.Transaction.ID, entry.Fields["transaction.id"])
+		assert.Equal(t, transactionDoc.Trace.ID, traceID,
+			"expecting log with trace id %s; got trace id %s and message \"%s\" instead", transactionDoc.Trace.ID, traceID, entry.Message)
+		assert.Equal(t, transactionDoc.Transaction.ID, entry.Fields["transaction.id"],
```
kruskall, line 94: **suggestion**: I think we can use `assert#Equalf` for clarity. WDYT ?

kruskall (approved):
LGTM! Just a minor suggestion!

systemtest/instrumentation_test.go
carsonip: Agreed for clarity. `Equalf` uses `Equal` anyways.

kruskall (approved):

PR #10127, opened 2023-01-27
Lower default service tx aggregation histogram sig fig

carsonip:
Current default of service transaction aggregation HDR histogram significant figures of 5 will allocate a [2097152]int64 for `histogram.counts`, meaning that each histogram is at least 16MB.

Given service tx aggregation max_groups is autosized to 10k per 1GB memory limit, it will take up a dangerous amount of memory. Lower the default histogram significant figures to 2 so that a [3328]int64 is allocated instead.


axw (approved):
thanks @carsonip!

kruskall (approved):

simitt (approved):

carsonip:
Tested locally with a modified system test. 
For:
- 100 service transactions, memory usage dropped from ~17GB to 4280844K ~= 4GB.
- 200 service transactions, memory usage dropped from ~26GB to 4281356K ~= 4GB.

PR #10128, opened 2023-01-27
Added field mappings for mobile use cases to error_logs schema

AlexanderWert:
Some of the fields for mobile logs and errors were missing in the field mappings:

- `session.id`
- `event.type`
- `event.kind`
- `event.category`
- `error.type`

### DIFF
--- a/apmpackage/apm/data_stream/app_logs/fields/ecs.yml
+++ b/apmpackage/apm/data_stream/app_logs/fields/ecs.yml
@@ -75,6 +75,8 @@
 - external: ecs
   name: error.stack_trace
 - external: ecs
+  name: error.type
+- external: ecs
   name: event.dataset
 - external: ecs
   name: event.outcome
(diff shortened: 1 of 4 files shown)


kruskall (approved):
LGTM! Thanks for the fix!

apmpackage/apm/data_stream/app_logs/fields/ecs.yml
```diff
@@ -74,6 +74,8 @@
   name: ecs.version
 - external: ecs
   name: error.stack_trace
+- external: ecs
+  name: error.type
```
simitt, line 78: Is the `error.type` really set for the `logs-apm.app_logs-*` datastream and not only for `logs-apm.error_logs-*`?

simitt:
@AlexanderWert coud you also update the dependency to `apm-data`?

PR #10129, opened 2023-01-27
systemtest: Fix ES query and cleanup on hidden data stream

carsonip:
## Motivation/summary

Fix failing system test due to recently merged hidden non-default rollup interval data stream #10070 and related kibana changes

## Checklist

N/A

## How to test these changes

N/A


systemtest/elasticsearch.go
```diff
@@ -93,11 +93,14 @@ func CleanupElasticsearch(t testing.TB) {
...
-		"metrics-apm*",
-		"logs-apm*",
-	}}, nil)
+	_, err := Elasticsearch.Do(context.Background(), &esapi.IndicesDeleteDataStreamRequest{
+		Name: []string{
+			"traces-apm*",
+			"metrics-apm*",
+			"logs-apm*",
+		},
+		ExpandWildcards: "all,hidden",
```
simitt, line 102: Is `hidden` necessary when setting `all`? See https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-update-settings.html

carsonip: For delete data stream, the docs https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-delete-data-stream.html indicate that the valid values are `all,hidden`, `open,closed` and `none`, while the default is `open`, which isn't even a valid value. Not sure if my brain parses the docs right, because it confuses me a little bit. I haven't tried `open,hidden` for this yet. Will try and see if it works the same.

carsonip: When I look at the docs again, I think it meant `all` == `hidden`. Updated PR to use `all` only.

kruskall (approved):

PR #10130, opened 2023-01-27
Fix aggregations make batch size

carsonip:
## Motivation/summary

Allocate the right size for batch, on overflow & hash collision.

## Checklist

N/A

## How to test these changes

N/A


simitt:
@carsonip is this still relevant?

carsonip:
> @carsonip is this still relevant?

This is still relevant. The current batch make does not account for overflow bucket which means when there's an overflow, the backing array of batch will need to be extended. Not sure if this counts as a 8.7 bug.

simitt:
Then please make this ready for review; borderline to consider it a bugfix, but if we bring it in before the next BC, I'd be ok with it.

marclop (approved):

PR #10133, opened 2023-01-30
test: move TestProfiling to standalone APM Server

kruskall:
## Motivation/summary

Reduce fleet dependency in test and improve test time.
TestProfiling time went from 32s to 10s

## Checklist

### DIFF
--- a/systemtest/apmservertest/config.go
+++ b/systemtest/apmservertest/config.go
@@ -71,7 +71,7 @@ type Config struct {
 	Output OutputConfig `json:"output"`
 
 	// ProfilingConfig holds configuration related to profiling.
-	Profiling *ProfilingConfig `json:"apm-server.profiling"`
+	Profiling *ProfilingConfig `json:"apm-server.profiling,omitempty"`
 }
 
 // Args formats cfg as a list of arguments to pass to apm-server,
--- a/systemtest/apmservertest/config.go
+++ b/systemtest/apmservertest/config.go
@@ -69,6 +69,9 @@ type Config struct {
 
 	// Output holds configuration for the libbeat output.
 	Output OutputConfig `json:"output"`
+
+	// ProfilingConfig holds configuration related to profiling.
+	Profiling *ProfilingConfig `json:"apm-server.profiling"`
 }
 
 // Args formats cfg as a list of arguments to pass to apm-server,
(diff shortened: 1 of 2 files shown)


systemtest/profiling_test.go
```diff
@@ -82,31 +83,20 @@ func TestProfiling(t *testing.T) {
...
-							"hosts":   []string{"elasticsearch:9200"},
-							"api_key": apiKey.ID + ":" + apiKey.APIKey,
-						},
-					},
-				},
-			},
-		})
-
-	apmServerURL, _ := url.Parse(apmIntegration.URL)
+	srv := apmservertest.NewUnstartedServerTB(t)
```
axw, line 86: With this change, the comment at line 58 doesn't make sense, and the test doesn't quite cover what it was doing. We could simulate what Elastic Agent does by creating a more limited API Key for the Elasticsearch output.

kruskall: Good catch! Wouldn't the behaviour be the same with standalone ? In this case: the standalone APM Server is using the default elasticsearch credentials (with narrow privileges) and we are creating and passing an api_key to write to the profiling indices.

axw: Ah, true. I was thinking the standalone apm-server run by system tests used superuser privileges, but they don't. So, I think we just need to update the comment then.

kruskall: Updated! :+1:

systemtest/apmservertest/config.go
```diff
@@ -69,6 +69,9 @@ type Config struct {
 
 	// Output holds configuration for the libbeat output.
 	Output OutputConfig `json:"output"`
+
+	// ProfilingConfig holds configuration related to profiling.
+	Profiling *ProfilingConfig `json:"apm-server.profiling"`
```
axw, line 74:
```suggestion
	Profiling *ProfilingConfig `json:"apm-server.profiling,omitempty"`
```

systemtest/apmservertest/config.go
```diff
@@ -169,6 +172,28 @@ type TailSamplingPolicy struct {
...
+
+	// ILMConfig
+	ILMConfig *ProfilingILMConfig `json:"keyvalue_retention"`
+}
+
+type ProfilingILMConfig struct {
+	Age         time.Duration `json:"age"`
+	SizeInBytes uint64        `json:"size_bytes"`
+	Interval    time.Duration `json:"execution_interval"`
+}
```
axw, line 195: I think we should leave ILMConfig out, and remove the ProfilingILMConfig. It's not used AFAICS. If it were, I think we'd need to add a custom MarshalJSON method like in other structs, for marshalling the time.Duration as expected.

axw (approved):
LGTM with one minor change.

systemtest/apmservertest/config.go
kruskall: Good catch! Updated!

PR #10140, opened 2023-01-30
docs: Remove txmetrics config

carsonip:
Removing transaction metrics config as they are not officially supported.


docs/legacy/configuring.asciidoc
```diff
@@ -73,8 +72,6 @@ include::{libbeat-dir}/shared-ssl-config.asciidoc[]
 include::ssl-input-settings.asciidoc[]
 // END SSL SECTION --------------------------------------------
 
-include::./transaction-metrics.asciidoc[]
```
kruskall, line 76: **issue**: this might require coordinating with the kibana team to remove the companion page on their documentation (should be a config option): https://github.com/elastic/kibana/pull/82379

docs/legacy/configuring.asciidoc
```diff
@@ -21,7 +21,6 @@ include::{libbeat-dir}/shared/configuring-intro.asciidoc[]
 * <<configuration-path>>
 * <<configuration-rum>>
 * <<configuration-ssl-landing>>
-* <<transaction-metrics>>
```
kruskall, line 24: **issue**: this needs to be backported to 7.x as well

kruskall (changes requested):
cc @bmorelli25 for visibility since he was the owner of the original PR

docs/legacy/configuring.asciidoc
bmorelli25: Yup. This will need to be removed and backported from the Kibana docs before docs ci will pass on this PR (as that content links to this content). I'll open a PR in the Kibana repo take care of that.

docs/legacy/configuring.asciidoc
bmorelli25: Yup. I updated the backport labels. We don't usually backport doc changes to previous versions of 8.x, but I can if y'all think that's necessary in this case.

docs/legacy/configuring.asciidoc
bmorelli25: Opened https://github.com/elastic/kibana/pull/149896

bmorelli25: Merged and backported in the Kibana repo. Rerunning docs ci now...

bmorelli25:
@elasticmachine, run elasticsearch-ci/docs

bmorelli25 (approved):

carsonip:
@bmorelli25 sorry for the unclear comms. We want to stop documenting these apm-server transaction aggregation configs because they are not officially supported. e.g. `interval` option has been removed, and a poor `hdrhistogram_significant_figures` value may cause the system to run out of memory.

I thought the change in kibana docs will be to only remove the pointer but not the paragraph. I'm not sure if we need to remove the paragraph as it is not exactly about apm-server but more about the UI, and they may still be valid and officially supported.

cc @simitt

carsonip:
@simitt Do you know what "additional configuration" it is trying to imply in the kibana docs?

simitt:
@bmorelli25, I discussed with @sqren that removing the paragraph from Kibana UI is fine, and even overdue. 
@carsonip we had exposed the time interval, significant numbers etc in the config, the line refered to these options.

carsonip:
@bmorelli25 let me know if this PR is ready to merge or simply takeover from here. Thanks!

sqren:
> I thought the change in kibana docs will be to only remove the pointer but not the paragraph. I'm not sure if we need to remove the paragraph as it is not exactly about apm-server but more about the UI, and they may still be valid and officially supported.

I'm in favour of removing the setting on the UI side as well. In the past we allowed the user to disable transaction metrics but I don't see any good reason for doing that anymore. When the setting is removed it'll keep the same default behaviour (`auto`) meaning the UI will attempt to use metrics first, and only fallback to transaction events if they are not available.

simitt (approved):

PR #10156, opened 2023-02-01
perf: reduce aggregation initial memory usage

kruskall:
## Motivation/summary

we are currently allocating a map with a size of maxGroups/maxServices/maxSize. This is consuming a lot of memory, especially for people that do not come close to the maximum value.
Use (max/16)+1 as an hint to reduce memory usage. The new value will make sure we have at most 4 reallocation (table doubling = 2->4->8->16) and that the final value is at least equal to the maximum.

## Checklist


simitt:
@kruskall are you intending on making this production ready and providing numbers for the improvement?

kruskall:
> @kruskall are you intending on making this production ready and providing numbers for the improvement?

Yes, I think there's an opportunity for improvement here. Unfortunately this is a bit more complicated than it seems because we're using chaining as a collision resolution strategy so we need the big array.

kruskall:
Closing this. I'll open an issue with the details and then we can shedule time for it

PR #10185, opened 2023-02-07
Fix incorrect `transaction.duration.summary.sum` calculation

### LINKED ISSUE #10177: Incorrect `transaction.duration.summary.sum` calculation
Existing `transaction.duration.summary.sum` calculation is incorrect. It should be a sum of products of count*val instead of sum of just the counts from histogram. Also, there is a precision loss in the values from the histogram. Ideally the count & sum should be tracked separately in transactionMetrics.

carsonip:
## Motivation/summary

`transaction.duration.summary.sum` calculation was incorrect. Fix it by introducing a separate count and sum per transactionMetric. This will also provide more accurate summaries.

## Checklist

## How to test these changes

- Check that `transaction.duration.summary.sum` is correct on txmetrics aggregation

## Related issues

Closes #10177


carsonip:
cc @dgieselaar fyi there was a bug with wrong `transaction.duration.summary.sum` and we are fixing it now. Also, it is now possible to have a different number in `transaction.duration.summary.sum` vs values in histogram because there is a precision loss associated in histogram. It is most noticeable when count=1.

axw (approved):
LGTM, thanks.

Should we have `makeMetricset` call `transactionMetrics.histogramBuckets`, rather than requiring the caller to pass the results in? Also, should we stop returning a total count from `histogramBuckets` and use the `transactionMetrics.count` field for setting `APMEvent.Metricset.DocCount` (the only place we're using totalCount)?

carsonip:
> Should we have `makeMetricset` call `transactionMetrics.histogramBuckets`, rather than requiring the caller to pass the results in? 

I did it because we did the same in service transaction metrics, which looked a bit weird imo. This is definitely an area for improvement.

> Also, should we stop returning a total count from `histogramBuckets` and use the `transactionMetrics.count` field for setting `APMEvent.Metricset.DocCount` (the only place we're using totalCount)?

Good catch, I shouldn't have missed that.

carsonip:
> Good catch, I shouldn't have missed that.

@axw We are also using histogram's total count in service tx metrics. Total count from histograms should be *almost* as accurate as our separate count. There might be a small precision loss when we multiply the count (float64) by 1000 before placing it into the histogram and divide it by 1000 when taking it out. That's why I'm not sure if this should be counted as a bug, or just code clean up. I'll go forward to make this improvement but should this be backported to 8.7?

marclop:
Tested with 8.7.0 BC2

Using the program below the durations in nanoseconds are: `1000 + 100000 + 1000000 = 1,101,000`. So `transaction.duraiton.summary.sum` must be `1,101,000`, and `transaction.duraiton.summary.count`, 3.

```go
package main

import (
	"fmt"
	"time"

	"go.elastic.co/apm/v2"
)

func main() {
	tracer := apm.DefaultTracer()
	for _, duration := range []int{1, 100, 1000} {
		tx := tracer.StartTransaction("transaction", "type")
		if duration == 1 {
			// duration to report is 2ms.
			span := tx.StartSpanOptions("span", "type", apm.SpanOptions{
				ExitSpan: true,
			})
			span.End()
		}
		tx.End()
	}
	tracer.Flush(nil)
}
```

And run it with 3 different service names: `apm`, `kibana`, `fleet-server`.

```
GET /metrics-apm.transaction*/_search?filter_path=hits.hits
{
  "_source": false,
  "fields": [
    "transaction.duration.summary",
    "service.name"
  ]
}
```

```
{
  "hits": {
    "hits": [
      {
        "_index": ".ds-metrics-apm.transaction.1m-default-2023.02.20-000001",
        "_id": "QqerbIYBZ-vqtropTfNb",
        "_score": 1,
        "fields": {
          "service.name": [
            "fleet-server"
          ],
          "transaction.duration.summary": [
            {
              "sum": 1101000,
              "value_count": 3
            }
          ]
        }
      },
      {
        "_index": ".ds-metrics-apm.transaction.1m-default-2023.02.20-000001",
        "_id": "Q6erbIYBZ-vqtropTfNb",
        "_score": 1,
        "fields": {
          "service.name": [
            "kibana"
          ],
          "transaction.duration.summary": [
            {
              "sum": 1101000,
              "value_count": 3
            }
          ]
        }
      },
      {
        "_index": ".ds-metrics-apm.transaction.1m-default-2023.02.20-000001",
        "_id": "RKerbIYBZ-vqtropTfNb",
        "_score": 1,
        "fields": {
          "service.name": [
            "apm"
          ],
          "transaction.duration.summary": [
            {
              "sum": 1101000,
              "value_count": 3
            }
          ]
        }
      }
    ]
  }
}
```

PR #10217, opened 2023-02-09
fix: improve smoke testing flexibility and resolve deployment issues

kruskall:
## Motivation/summary

The current smoke testing scripts have quite a few bugs that were uncovered by the supported-os test.

- fix delete_all function: delete_all is using the delete index api but the request had no effect and returned no error.
According to the documentation, allow_no_indices defaults to true and changing it to false shows the following error:

"type":"index_not_found_exception","reason":"no such index [[_all]]"

Setting expand_wildcards=all also fails with the following:

"type":"illegal_argument_exception","reason":"index [.ds-logs-elastic_agent.filebeat-default-2023.02.09-000001] is the write index for data stream [logs-elastic_agent.filebeat-default] and cannot be deleted"

The function has been migrated to the delete_by_query API.

- fix supported-os test always using the latest version the supported-os test is not writing the stack-version tfvar which defaults to the ec provider using the latest available. This resulted in a broken deployment because the script was picking up 8.7.0 (non snapshot) as the latest version instead of the one retrieved from the cloud API.
The script has been updated to always set the stack-version consistently to what is being print to stdout.

- fix terraform_apply overwriting variables if called with an explicit version the terraform_apply function will overwrite the tfvar file, losing any previous variable provided with append_tfvar.
Removing the tfvar file was moved to a separate function which is called explicitly.

- fix terraform_apply not refreshing stack variables the terraform_apply will only retrieve the deployment endpoints the first time it is called.
This was breaking the supported-os test as the apm server url is changing between runs when testing different oses.

## Checklist


testing/smoke/lib.sh
```diff
@@ -43,28 +43,25 @@ terraform_init() {
...
+}
+
 append_tfvar() {
-    if [[ -f terraform.tfvars && ! -z ${3} && ${3} -gt 0 ]]; then rm terraform.tfvars; fi
+    echo "-> Adding tfvar ${1}=\"${2}\""
     echo ${1}=\"${2}\" >> terraform.tfvars
 }
 
 terraform_apply() {
-    echo "-> Creating / Upgrading deployment to version ${1}"
```
simitt, line 52: Some output would still be helpful when checking the logs.

simitt (approved):
Great catches!

testing/smoke/lib.sh
kruskall: Added!

kruskall:
@Mergifyio backport 8.7

PR #10224, opened 2023-02-10
test: move sourcemap systemtests to standalone apm server

kruskall:
## Motivation/summary

More changes to remove the dependency from fleet from system tests: migrate sourcemap tests to standalone apm server.
A new system test in `fleet_test` is added to make sure the sourcemap feature is working with the APM Integration.

This also saves a bit less than a minute in test time.

## Checklist


systemtest/fleet_test.go
```diff
@@ -153,6 +153,26 @@ func TestFleetIntegrationAnonymousAuth(t *testing.T) {
 	test("denied_service", "allowed_agent", http.StatusForbidden)
 }
 
+func TestFleetIntegrationSourcemap(t *testing.T) {
```
kruskall, line 156: Adding a systemtest to make sure sourcemapping is working with fleet. Does it make sense to have this ?

axw: I think this is a good idea for now, but maybe later we can make it a smoke test.

axw (approved):

Issue #9969, opened 2023-01-04
docs: clarify when auth is enabled/disabled in the anonymous auth guide


simitt:
By default, no auth handling between APM agents and APM Server is enabled. To enforce auth, either the `apm-server.auth.api-key.enabled: true` or `apm-server.auth.secret_token: xxx` needs to be set. Anonymous auth handling is coupled with these config settings; if no auth handling for backend agents is configured, then the auth handling for RUM agents are also not taken effect. If any of the two options are set, anonymous auth handling is also enforced with its default settings. 

The docs should be updated to better reflect that, by default, RUM related auth handling is disabled, but enabled as soon as auth handling is generally enabled.

PR #10227, opened 2023-02-10
docs: Clarify anonymous authentication details

bmorelli25:
### Summary

> The docs should be updated to better reflect that, by default, RUM related auth handling is disabled, but enabled as soon as auth handling is generally enabled.

The goal of this PR is to better clarify when and why anonymous requests to the APM Server are accepted or rejected. Here's the thought process—and please correct me if I'm wrong:

We have two anonymous authentication pages in the docs. One is for the [APM integration](https://www.elastic.co/guide/en/apm/guide/master/anonymous-auth.html) and one is for [standalone APM Server](https://www.elastic.co/guide/en/apm/guide/master/secure-communication-unauthenticated.html). This PR updates these pages to clarify that the APM Server’s default response to anonymous requests depends on whether an API key or Secret Token has been configured. If and only if an API key or Secret token has been configured, you can enable anonymous authentication in the APM Server to allow the ingestion of unauthenticated client-side APM data while still requiring authentication for server-side services.

In addition to the above pages, the standalone APM Server documentation has a [configuration reference](https://www.elastic.co/guide/en/apm/guide/master/configuration-anonymous.html) for anonymous authentication. Previously, this page stated:

> Anonymous authentication must be enabled to collect RUM data. For this reason, anonymous auth will be enabled automatically if [apm-server.rum.enabled](https://www.elastic.co/guide/en/apm/guide/master/configuration-rum.html#rum-enable) is set to true, and [apm-server.auth.anonymous.enabled](https://www.elastic.co/guide/en/apm/guide/master/configuration-anonymous.html#config-auth-anon-enabled) is not explicitly defined.

IIUC, this is only true if a secret token or API key has been configured. I've reworked the text on this page to better reflect this.

I think those are the important bits. Have a good weekend 👋 

### Related

* Follows up on @lucabelluccini's PR (thanks!) https://github.com/elastic/apm-server/pull/10076
* Closes https://github.com/elastic/apm-server/issues/9969

### DIFF
--- a/docs/empty.yml
+++ b/docs/empty.yml
@@ -1,0 +1,1 @@
+# delete me
--- a/docs/empty.yml
+++ b/docs/empty.yml
@@ -1,1 +1,1 @@
-# delete me
+#delete me
--- a/docs/empty.yml
+++ b/docs/empty.yml
@@ -1,1 +1,1 @@
-#delete me
+# delete me
--- a/docs/legacy/configuration-anonymous.asciidoc
+++ b/docs/legacy/configuration-anonymous.asciidoc
@@ -9,8 +9,12 @@ IMPORTANT: {deprecation-notice-config}
 If you're using {fleet} and the Elastic APM integration, please see <<input-apm>> instead.
 
 Elastic APM agents can send unauthenticated (anonymous) events to the APM Server.
-This is useful for agents that run on clients, like the Real User Monitoring (RUM) agent running in a browser,
-or the iOS/Swift agent running in a user application.
+An event is considered to be anonymous if no authentication token can be extracted from the incoming request.
+This is useful for agents that run on clients, like the Real User Monitoring (RUM)
+agent running in a browser, or the iOS/Swift agent running in a user application.
+
+Enable anonymous authentication in the APM Server to allow the
+ingestion of unauthenticated client-side APM data while still requiring authentication for server-side services.
 
 Example configuration:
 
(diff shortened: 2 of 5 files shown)


lucabelluccini (approved):
I think the only point which is not clear now is explain what happens if APM Server/Integration allows both api key & secret token.

At least from https://github.com/elastic/apm/issues/183 and from some tests we did: the api key is sent out by the agent if both are configured and so the event should be accepted/allowed

simitt:
>what happens if APM Server/Integration allows both api key & secret token.

The server behaves the same if only the api key or secret token or both are enabled. The differentiator is whether or not any auth is required by enabling at least one of them. 

> At least from https://github.com/elastic/apm/issues/183 and from some tests we did: the api key is sent out by the agent if both are configured and so the event should be accepted/allowed

I'm not sure I follow; the apm agent will use whatever is configured for the agent (not in the apm-server) - api key, secret token or nothing, and this can be different between apm agents.

bmorelli25:
To add to Silvia's response, we mention in the [docs](https://www.elastic.co/guide/en/apm/guide/master/secure-agent-communication.html) that

>  If both API keys and a secret token are enabled, APM agents can choose whichever mechanism they support.

bmorelli25:
@Mergifyio backport 8.7

bmorelli25:
@Mergifyio backport 8.7

PR #10249, opened 2023-02-14
test: fix most goroutine leaks in unit tests

kruskall:
## Motivation/summary

More goroutine leaks found using https://github.com/uber-go/goleak:

- http servers not being closed
- pipelines not being closed
- response bodies not being closed
- blocking forever by reading from nil channel

## Checklist


internal/sourcemap/body_caching.go
```diff
@@ -55,6 +55,13 @@ func NewBodyCachingFetcher(
 	go func() {
 		logger.Debug("listening for invalidation...")
 
+		// prevent goroutine leak
+		// this can only happens in tests, otherwise
+		// invalidationChan is never null
+		if invalidationChan == nil {
+			return
+		}
+
```
marclop, line 64: Can't we make the channel non nil in the tests? We could pass in a closed channel which would cause the coroutine to shut down immediately.

kruskall: Oh, that works too. Changed :+1:

marclop (approved):

PR #10250, opened 2023-02-14
fix: accept --environment macos_service and udpate testcase to avoid race condition

kruskall:
## Motivation/summary

The env variable is a loop variable and is captured by the func which is executed in parallel.
As a result the test only runs with `windows_service` and the other values are never tested.
Ironically one of the untested values is currently broken: https://github.com/elastic/elastic-agent-libs/pull/101

## Checklist


kruskall:
PR on elastic-agent-libs has been merged. I've released a new version and bumped the library here.

Do we want to backport this ?

carsonip (approved):
good catch! is the go mod bump needed?

kruskall:
> good catch! is the go mod bump needed?

Yes, the bug is in the library: https://github.com/elastic/elastic-agent-libs/pull/101

axw:
Oof! Good catch, thank you!

PR #10258, opened 2023-02-14
test: fix io.pipe locking the goroutine forever while writing 

kruskall:
## Motivation/summary

The following is a sample output from using https://github.com/uber-go/goleak on a systemtest starting an apm server instance:

```
 Goroutine 94 in state select, with io.(*pipe).write on top of the stack:
goroutine 94 [select]:
io.(*pipe).write(0xc00068ade0, {0xc000402000, 0x1, 0x600})
        /usr/lib/go/src/io/pipe.go:86 +0x20b
io.(*PipeWriter).Write(0x40db7d?, {0xc000402000?, 0x135d2a0?, 0xc000347e01?})
        /usr/lib/go/src/io/pipe.go:165 +0x25
bytes.(*Reader).WriteTo(0xc00091e660, {0x15a6180?, 0xc000015418?})
        /usr/lib/go/src/bytes/reader.go:143 +0x87
io.copyBuffer({0x15a6180, 0xc000015418}, {0x15a3b80, 0xc00091e660}, {0xc00054e000, 0x8000, 0x8000})
        /usr/lib/go/src/io/io.go:409 +0x16e
io.(*multiReader).writeToWithBuffer(0xc000494e58, {0x15a6180, 0xc000015418}, {0xc00054e000, 0x8000, 0x8000})
        /usr/lib/go/src/io/multi.go:54 +0xe5
io.(*multiReader).WriteTo(0x18?, {0x15a6180, 0xc000015418})
        /usr/lib/go/src/io/multi.go:45 +0x56
io.copyBuffer({0x15a6180, 0xc000015418}, {0x15a61a0, 0xc000494e58}, {0x0, 0x0, 0x0})
        /usr/lib/go/src/io/io.go:409 +0x16e
io.Copy(...)
        /usr/lib/go/src/io/io.go:386
github.com/elastic/apm-server/systemtest/apmservertest.(*Server).consumeStderr(0xc000597520, {0x15a61e0?, 0xc0003b4ba0})
        /apm-server/systemtest/apmservertest/server.go:421 +0x61a
created by github.com/elastic/apm-server/systemtest/apmservertest.(*Server).start
        /apm-server/systemtest/apmservertest/server.go:236 +0x94c
```

Writing to the stderrPipeWriter blocks forever.
Reads and Writes on the pipe are matched one to one except when multiple Reads are needed to consume a single Write. The data is copied directly from the Write to the corresponding Read; there is no internal buffering.
As a result the consumeStderr goroutine leaks and run forever. Because the data has already been written to the log at this point, we can just write directly to stderr.
Don't change the exit error stderr message to avoid duplicate messages.

## Checklist


systemtest/apmservertest/server.go
```diff
@@ -418,8 +412,8 @@ func (s *Server) consumeStderr(procStderr io.Reader) {
 
 	// Send the remaining stderr to s.Stderr.
 	procStderr = io.MultiReader(decoder.Buffered(), procStderr)
-	_, err := io.Copy(stderrPipeWriter, procStderr)
-	stderrPipeWriter.CloseWithError(err)
+	_, err := io.Copy(os.Stderr, procStderr)
+	return err
```
axw, line 416: `apmservertest.Server` is intended to have the sort of behaviour you might expect from `os/exec.Cmd`. i.e. it's not expected to write to `os.Stderr` unless told to. It it possible to fix the issue without changing the behaviour?

kruskall: Updated to avoid changing the current behaviour! :+1:

systemtest/apmservertest/server.go
```diff
@@ -449,6 +450,9 @@ func (s *Server) Close() error {
 	if err := interruptProcess(s.cmd.Process); err != nil {
 		s.cmd.Process.Kill()
 	}
+	// drain stderr even if there was no err so that the
+	// consumeStderr goroutine exits
+	io.ReadAll(s.Stderr)
```
axw, line 455: Should we close `s.Stderr`, rather than reading it? If the read end of the pipe is closed, then writes will fail. That should cause `consumeStderr` to exit more quickly, IIANM.

axw (approved):
Thanks, looks good. Just one suggestion.

systemtest/apmservertest/server.go
kruskall: That works too!

PR #10271, opened 2023-02-15
test: fix system test panic on compile error and avoid double close

kruskall:
## Motivation/summary

Trying to run system tests if there is a compile error will result in the error being printed to stderr followed by a nil pointer dereference:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x4b50f9]

goroutine 98 [running]:
os.(*Process).signal(0xc0004aab60?, {0x15a7998?, 0x1f0e850?})
        /usr/lib/go/src/os/exec_unix.go:63 +0x39
os.(*Process).Signal(...)
        /usr/lib/go/src/os/exec.go:138
github.com/elastic/apm-server/systemtest/apmservertest.interruptProcess(...)
        /apm-server/systemtest/apmservertest/server_unix.go:31
github.com/elastic/apm-server/systemtest/apmservertest.(*Server).Close(0xc0004aab60)
        /apm-server/systemtest/apmservertest/server.go:434 +0x5a
github.com/elastic/apm-server/systemtest/apmservertest.NewUnstartedServerTB.func1.1()
        /apm-server/systemtest/apmservertest/server.go:138 +0x26
created by github.com/elastic/apm-server/systemtest/apmservertest.NewUnstartedServerTB.func1
        /apm-server/systemtest/apmservertest/server.go:138 +0x1a5
```

Additionally: closing apmservertest twice results in a silent error.

## Checklist

### DIFF
--- a/systemtest/aggregation_test.go
+++ b/systemtest/aggregation_test.go
@@ -38,7 +38,7 @@ import (
 
 func TestTransactionAggregation(t *testing.T) {
 	systemtest.CleanupElasticsearch(t)
-	srv := apmservertest.NewUnstartedServerTB(t)
+	srv := apmservertest.NewUnstartedServer()
 	srv.Config.Monitoring = &apmservertest.MonitoringConfig{
 		Enabled:       true,
 		MetricsPeriod: 100 * time.Millisecond,
@@ -125,7 +125,8 @@ func TestTransactionAggregation(t *testing.T) {
 
 func TestTransactionAggregationShutdown(t *testing.T) {
 	systemtest.CleanupElasticsearch(t)
-	srv := apmservertest.NewServerTB(t)
+	srv := apmservertest.NewUnstartedServer()
+	require.NoError(t, srv.Start())
 
 	// Send a transaction to the server to be aggregated.
 	tracer := srv.Tracer()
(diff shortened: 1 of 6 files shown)


systemtest/aggregation_test.go
```diff
@@ -38,7 +38,7 @@ import (
 
 func TestTransactionAggregation(t *testing.T) {
 	systemtest.CleanupElasticsearch(t)
-	srv := apmservertest.NewUnstartedServerTB(t)
+	srv := apmservertest.NewUnstartedServer()
```
axw, line 41: If the test fails part way through, the server will be left running won't it? The idea behind NewUnstartedServerTB is that it should always close the server that it starts, when the test ends.

Changing NewUnstartedServerTB to NewUnstartedServer means transferring the responsibility of closing the server back to each test - that seems a bit error prone to me.

Could we change `Server.Close` to be safe to call multiple times, including concurrently?

axw (changes requested):

systemtest/aggregation_test.go
kruskall: Good catch! Updated!

axw (approved):

PR #10272, opened 2023-02-15
[poc] agent telemetry

SylvainJuge:
- add new agent telemetry endpoint

## Checklist
- [ ] get and send the cluster ID
- [ ] make it configurable: boolean toggle (disabled by default), telemetry URL (dev URL by default)
- [ ] modify default configuration for production deployment in cloud: feature should be enabled and use production URL for telemetry.

## Checklist

### DIFF
--- a/internal/beater/api/telemetry/handler.go
+++ b/internal/beater/api/telemetry/handler.go
@@ -1,0 +1,77 @@
+// Licensed to Elasticsearch B.V. under one or more contributor
+// license agreements. See the NOTICE file distributed with
+// this work for additional information regarding copyright
+// ownership. Elasticsearch B.V. licenses this file to you under
+// the Apache License, Version 2.0 (the "License"); you may
+// not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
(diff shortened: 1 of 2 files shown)


internal/beater/api/telemetry/handler.go
```diff
@@ -0,0 +1,77 @@
...
+	ClusterId string
+
+	// Version contains the APM Server version
+	Version string
+}
+
+func Handler(cfg HandlerConfig) request.Handler {
+
+	return func(c *request.Context) {
+		err := sendTelemetry(cfg.TelemetryUrl, c.Request.Header.Get("content-type"), c.Request.Body, cfg.ClusterId, cfg.Version)
```
SylvainJuge, line 47: `/!\` the current issue is around here, we can properly get the `content-length` header with the expected value, but the original request body appears to be empty :-(

basepi: After banging my head against this much longer than I'd like to admit (why is `ContentLength` -1, even when there's a valid body??? I *still* don't know) -- I think the problem was that you were trying to pass the incoming request's `io.Reader` to the outbound request without repackaging it. Not certain why that would be a problem, since you weren't calling `ReadAll` on it, but with my contributions where I read it out to a `[]byte` and then create a new buffer wrapped in an `io.Reader`, it's working for me. Perhaps it gets closed out of band or something? I am not sure.

SylvainJuge: Thanks for taking care of this, I also tried doing a copy, but without any success, so congrats for finding a work-around !

SylvainJuge:
Not yet, for now it was mostly for experimentation and it's definitely not production grade.
There are a few aspects that we need to solve first before maybe moving forward with this.

simitt:
sounds good, closing the PR then for now

PR #10056, opened 2023-01-17
Add global labels to service_destination metrics

carsonip:
## Motivation/summary

Add global labels to service_destination metrics for filtering APM data in APM UI

## Checklist

- [x] Update [CHANGELOG.asciidoc](https://github.com/elastic/apm-server/blob/main/CHANGELOG.asciidoc)
~~- [ ] Update [package changelog.yml](https://github.com/elastic/apm-server/blob/main/apmpackage/apm/changelog.yml) (only if changes to `apmpackage` have been made)~~
~~- [ ] Documentation has been updated~~

## How to test these changes

- Confirm that global labels are present in service_destination metrics

## Related issues

Closes #10045

Issue #10045, opened 2023-01-16
Add global labels to service_destination metrics


simitt:
We want to allow using global labels for filtering APM data everywhere in the APM UI. For this to work, global labels also need to be added to the `service_destination` (span) metrics. 

The groundwork for adding labels is already done via https://github.com/elastic/apm-server/tree/main/x-pack/apm-server/aggregation/labels.

PR #10297, opened 2023-02-20
Fix service destination labels regression

carsonip:
## Motivation/summary

Fix regression introduced by #10056 where service destination labels are not stored successfully.

## Checklist

## How to test these changes

- Ensure that service destination labels are stored

## Related issues

Related to #10045

### DIFF
--- a/systemtest/aggregation_test.go
+++ b/systemtest/aggregation_test.go
@@ -151,6 +151,7 @@ func TestTransactionAggregationShutdown(t *testing.T) {
 }
 
 func TestServiceDestinationAggregation(t *testing.T) {
+	t.Setenv("ELASTIC_APM_GLOBAL_LABELS", "department_name=apm,organization=observability,company=elastic")
 	systemtest.CleanupElasticsearch(t)
 	srv := apmservertest.NewServerTB(t)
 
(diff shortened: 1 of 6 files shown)


systemtest/aggregation_test.go
```diff
@@ -151,13 +151,17 @@ func TestTransactionAggregationShutdown(t *testing.T) {
 }
 
 func TestServiceDestinationAggregation(t *testing.T) {
+	t.Setenv("ELASTIC_APM_GLOBAL_LABELS", "department_name=apm,organization=observability,company=elastic")
```
marclop, line 154: Can we add a numeric label so we can verify that both mappings are applied?

carsonip: Do you have an easy way to add numeric labels with the Go agent? I'm looking at Go agent tracer code and in `parseGlobalLabels` it parses labels from env into kv of string and string.

marclop: Yeah, you're right. Looking at the implementation we only ever added support for k/v string label parsing. It'd be ideal if we could test that both are parsed in APM Server, but that'd require sending a literal payload and creating another test. I think it's worth the trouble. What do you think?

carsonip: If we do it I would love to do it for all aggregations, because this test is missing for all of them.

marclop: let's open a follow up issue then to add numeric labels tests to all aggregations.

marclop (approved):

systemtest/aggregation_test.go
carsonip: Created #10324

marclop:
I re-verified this on the linked issue with BC5.

PR #10329, opened 2023-02-22
fix: invert args order for prefix check in deprecation notice

kruskall:
## Motivation/summary

_Recalling the exact syntax for the built-in stab() function, you make a sane assumption and call shoot(GUN, FOOT); The foot shoots your gun_

## Checklist


axw (approved):
oops :) sorry, I clearly didn't do a good review either!

PR #10330, opened 2023-02-22
Collect monitoring for the overflow metrics for all aggregations

### LINKED ISSUE #10207: Add monitoring data for metrics aggregation
From `8.7` onwards the APM Server creates service summary, service transaction, transaction and span destination metrics aggregations and also collects overflow information in an _other_ bucket. The only aggregator that currently reports monitoring data about the aggregation is the [tx metrics aggregator](https://github.com/elastic/apm-server/blob/main/x-pack/apm-server/aggregation/txmetrics/aggregator.go#L187-L188). 

- All aggregators need to report metrics on overflow buckets and current number of active groups.
- The tx metrics aggregator reporting needs to be updated to include monitoring information about which group has overflowed (service, tx, per-service-tx).

This was an oversight to not be added during development and the fix should be backported to `8.7`.

lahsivjar:
## Motivation/summary

Collect monitoring for the overflow metrics for all aggregations. Some major changes in the PR are:
1. The PR moves the metric updates to `publish` method, this would mean that the metrics are updated periodically based on the metric interval (set to `1m`).
2. The overflow metrics collected represent an estimate of the total number of unique transaction groups that overflowed rather than total number of overflow.
3. Changes the metric name from `txmetrics.overflowed` to `txmetrics.overflowed.total`.
4. Introduces 3 more metrics: `txmetrics.overflowed.services`, `txmetrics.overflowed.per_service_txn_groups` and `txmetrics.overflowed.txn_groups` to represent overflow reason.

## Checklist

## How to test these changes

Create overflow as per [these steps](https://github.com/elastic/apm-server/issues/9648#issuecomment-1423607724) and check if the metrics are published.

## Related issues

Fixes #10207


x-pack/apm-server/aggregation/txmetrics/aggregator.go
```diff
@@ -218,12 +227,23 @@ func (a *Aggregator) publish(ctx context.Context, period time.Duration) error {
...
+			if isMetricsPeriod {
+				activeGroups = int64(current.entries)
+				if svc == overflowBucketName {
+					servicesOverflow += overflowCount
+				} else if svcEntry.entries >= a.config.MaxTransactionGroupsPerService {
+					perSvcTxnGroupsOverflow += overflowCount
+				} else {
+					txnGroupsOverflow += overflowCount
+				}
+			}
```
lahsivjar, line 240: [**For reviewers**] Collection of metrics in `publish` method was driven by the following factors:

1. Collect total number of unique transaction groups that overflowed due to _max services limit_, _per service max transaction limit_, and _max transaction limit_. To do this, we would require the estimate from the hyperloglog sketch. While, the value would still be an estimate due to hash-collisions and hyperloglog usage but the metrics will be better aligned with the published metricset.
2. Hyperloglog++ sketch with sparse representation requires a write lock even for estimating cardinality.
3. IIUC, `CollectMonitoring` is part of a pull-based monitoring architecture, and doing a full lock on the whole struct didn't sound like a good idea.

x-pack/apm-server/aggregation/txmetrics/aggregator_test.go
```diff
@@ -93,49 +93,59 @@ func TestTxnAggregatorProcessBatch(t *testing.T) {
...
+			expectedOverflowReasonPerSvcTxnGrps: 16,
+			expectedOverflowReasonTxnGrps:       0,
+			expectedOverflowReasonSvc:           0,
 		},
 		{
-			name:                             "record_into_other_txn_if_txn_grps_limit_breached",
-			uniqueTxnCount:                   60,
-			uniqueServices:                   20,
-			expectedActiveGroups:             40,
-			expectedPerSvcTxnLimitOverflow:   2,
```
simitt, line 118: why is this 0 now? (same for the above example)

lahsivjar: I changed the meaning behind the variables. Previously the value `expectedPerSvcTxnLimitOverflow` represented the total number of events that were recorded in the overflow bucket. These included the overflow due to max transaction limit as well as due to per service transaction limit (since if the max transaction limit is breached we will still record the value in the overflow bucket of the specific service). In this PR, I updated the variables to represent the overflow due to a specific limit breach. The older and newer values can be translated as:

`expectedPerSvcTxnLimitOverflow = expectedOverflowReasonPerSvcTxnGrps + expectedOverflowReasonTxnGrps`
`expectedOtherSvcTxnLimitOverflow = expectedOverflowReasonSvc`

PR #10353, opened 2023-02-24
test: update testcontainer library and remove reaper code

kruskall:
## Motivation/summary

`containerRepaer.connect` creates a goroutine that never completes because we are not closing the channel returned by the function: https://github.com/elastic/apm-server/blob/main/systemtest/containers.go#L91
While trying to fix that goroutine leak I notice we don't really need the reaper since we're closing the agent in a cleanup hook: https://github.com/elastic/apm-server/blob/main/systemtest/fleet_test.go#L205.
The agent container leak was fixed in https://github.com/elastic/apm-server/pull/10334.
The only other place using the reaper is runapm but that's settig the `Reap` field explicitly so this is not changing behaviour.

## Checklist


axw (commented):
> While trying to fix that goroutine leak I notice we don't really need the reaper since we're closing the agent in a cleanup hook: https://github.com/elastic/apm-server/blob/main/systemtest/fleet_test.go#L205.

The point of the reaper is to remove containers when the tests exit uncleanly. For example, say the process gets wedged for whatever reason and you interrupt it (Ctrl+C or whatever) - the container will be left running. The reaper ensures that container will eventually be destroyed.

kruskall:
Thanks for the feedback! :bow: 

I'll try to revisit this and find another way to prevent the goroutine from leaking

axw:
@kruskall can we close this in the mean time?

(FWIW I don't think it matters if we have some orphaned goroutines in the system tests, as long as they're not accumulating for each test. Happy to improve things if it doesn't significantly complicate anything though.)

kruskall:
@axw I've updated the PR to close the reaper goroutine once we close the agent container (we don't need the reaper anymore at that point since it's after the container has been killed).
The code was changed to create a goroutine/connection for each test but that's fine since `testcontainers` is reusing the reaper so there's at most one ryuk container running.
A simple test is to remove the `agent.Close` call (purposely leaking docker containers) and run `go test -run=TestFleetIntegration -v .` (matches more than one test) in the systemtest folder.
By watching the `docker ps` output:
- one ryuk container is created despite having a few `elastic-agent-systemtest` container (one for each test).
- once the timeout expires the reaper will terminate all agent containers.

axw (approved):

Issue #10338, opened 2023-02-23
sourcemap: failing to ping ES within 1s in metadata fetcher renders fetcher permanently broken


axw:
If the new Elasticsearch sourcemap fetcher fails to initialise, then it sets an error that is never cleared, and all future attempts fetch will fail with that error. Temporary issues, such as Elasticsearch being unavailable during startup, will render the fetcher permanently broken.

Moreover, initialisation involves making a successful request to Elasticsearch within 1 second, which may not be a sufficient amount of time when crossing a WAN:

https://github.com/elastic/apm-server/blob/c30c3ec56005d6041b758c5a3f8563bda315d8f3/internal/sourcemap/metadata_fetcher.go#L104-L113

Do we need to check the initErr in the SourcemapFetcher? I'm thinking we should do this instead:
- In `MetadataESFetcher.startBackgroundSync`: remove the `ping`, and change the code to `close(s.init)` after the first successful `s.sync`. It should not be closed immediately after the first _attempt_. Maybe remove the `syncTimeout`.
- In `SourcemapFetcher.Fetch`, add an overall timeout (5s?) which applies to both waiting for the metadata fetcher to be initialised, and subsequently calling the backend fetcher's Fetch method.

That way the metadata fetcher will continue trying to sync in the background, and `SourcemapFetcher.Fetch` can avoid blocking without affecting the synchronisation.

kruskall:
> If the new Elasticsearch sourcemap fetcher fails to initialise, then it sets an error that is never cleared, and all future attempts fetch will fail with that error. Temporary issues, such as Elasticsearch being unavailable during startup, will render the fetcher permanently broken.

I think I missed resetting the channel here: https://github.com/elastic/apm-server/blob/27f0419d3ffc368155437e3093355b833ffecfce/internal/sourcemap/metadata_fetcher.go#L140

An error during init is not stopping the goroutine and the metadata fetcher is still fetching ES in the background periodically. The errored state is not updated, we should reset `initErr` to nil.

> Moreover, initialisation involves making a successful request to Elasticsearch within 1 second, which may not be a sufficient amount of time when crossing a WAN:

Good point!
I did something similar in the chaned fetcher to add a 500ms context to fetch from the next fetcher: https://github.com/elastic/apm-server/blob/27f0419d3ffc368155437e3093355b833ffecfce/internal/sourcemap/chained.go#L50

[...]

PR #10360, opened 2023-02-27
fix: reset metadata fetcher init error after successful sync

kruskall:
## Motivation/summary

From the related issue:

> If the new Elasticsearch sourcemap fetcher fails to initialise, then it sets an error that is never cleared, and all future attempts fetch will fail with that error. Temporary issues, such as Elasticsearch being unavailable during startup, will render the fetcher permanently broken.

## Checklist

## Related issues

Related to https://github.com/elastic/apm-server/issues/10338


axw (commented):
This is an improvement, but I still think we can do better.

Can you please respond to my comment here? https://github.com/elastic/apm-server/issues/10338#issuecomment-1442805292

What is the benefit of the initial ping? Why do we block in `SourcemapFetcher.Fetch`? Would it be a problem if we just returned `ErrUnavailable` until the metadata cache is populated? As it is, there's a window where `SourcemapFetcher.Fetch` may behave as if the cache is empty, after the initial ping succeeds but before the `sync` completes.

axw (approved):
LGTM. I think we should probably remove `initErr` in https://github.com/elastic/apm-server/pull/10367 though, and only mark the metadata fetcher as ready when it has _successfully_ synced. Any errors would be logged, but not returned to the SourcemapFetcher.

kruskall:
@mergifyio backport 8.7

PR #10362, opened 2023-02-27
Fixes issues with transaction metrics aggregation.

### LINKED ISSUE #10349: Per svc max transaction group limit is not reset on publish for transaction metrics aggregation
**APM Server version** (`apm-server version`): 8.7, main

**Description of the problem including expected versus actual behavior**: The max per service transaction group limit doesn't reset the number of transaction group seen during an interval at publish. This causes the limit to eventually breach leading to the creation of overflow bucket.

**Steps to reproduce**:

 1. Create a 1 GB deployment of APM-Server.
 2. Generate load such that `x` transactions are sent per second. Keep `x` to be lower than `500/60` to ensure that the limit of per service transaction group should not be breached.
 3. Assert that the limits are breached at around `(500/x)*time.Minute`.

N/A

lahsivjar:
## Motivation/summary

The PR fixes the following issues:
* Resets limits to 0 after each publish of the transction metrics.
* Uses correct interval for calculating timestamp of overflow metrics.
* Fixes the monitoring and logging to correctly report the reason for overflow.

Along with the above fixes, the PR also refactors the transaction metrics aggregator to be more readable and maintainable while keeping the amount of changes minimum for backport. Would be happy to hear other opinions and suggestions.

## Checklist

## How to test these changes

The fix has a unit test. Can also be tested by following the below steps:

1. Create a 1 GB deployment of APM-Server.
2. Generate load such that x transactions are sent per second. Keep x to be lower than 500/60 to ensure that the limit of per service transaction group should not be breached.
3. Assert that the limits are NOT breached even after (500/x)*time.Minute.

## Related issues

Fixes #10349


x-pack/apm-server/aggregation/txmetrics/aggregator.go
```diff
@@ -303,147 +310,137 @@ func (a *Aggregator) ProcessBatch(ctx context.Context, b *model.Batch) error {
...
 	if count <= 0 {
-		return
+		return nil
 	}
 	for _, interval := range a.Intervals {
-		key := a.makeTransactionAggregationKey(event, interval)
-		a.updateTransactionMetrics(key, count, event.Event.Duration, interval)
+		key := makeTransactionAggregationKey(event, interval)
+		if err := a.updateTransactionMetrics(key, count, event.Event.Duration, interval); err != nil {
+			return err
```
axw, line 321: IIANM, the error would only ever occur if there's a bug in our code, and not because of something wrong with the data. Right? In that case, should we just log the error rather than returning it?

axw: Actually we're not even doing anything with the error, it's being ignored in ProcessBatch. Should we log it there?

x-pack/apm-server/aggregation/txmetrics/aggregator.go
```diff
@@ -248,22 +258,19 @@ func (a *Aggregator) publish(ctx context.Context, period time.Duration) error {
 			entry.histogram.Reset()
 			a.histogramPool.Put(entry.histogram)
 			entry.histogram = nil
-			svcEntry.other = nil
-			svcEntry.otherCardinalityEstimator = nil
-			svcEntry.entries = 0
 		}
+		svcEntry.reset()
 		delete(current.m, svc)
```
axw, line 263: Should we clear the map in the reset method too?

x-pack/apm-server/aggregation/txmetrics/aggregator.go
```diff
@@ -248,22 +258,19 @@ func (a *Aggregator) publish(ctx context.Context, period time.Duration) error {
 			entry.histogram.Reset()
 			a.histogramPool.Put(entry.histogram)
 			entry.histogram = nil
```
axw, line 260: Should we reset/recycle the histogram in the reset method?

axw (approved):
Looks good. I think it would help clarity if we made the new/reset methods symmetrical, so creating an entry is always undone by calling reset, and not spread across reset & publish.

x-pack/apm-server/aggregation/txmetrics/aggregator.go
lahsivjar: > Actually we're not even doing anything with the error, it's being ignored in ProcessBatch. Should we log it there?

Ah, this was a miss, thanks for catching it. I planned to add a log entry in `ProcessBatch` and also return the error from there.

> IIANM, the error would only ever occur if there's a bug in our code, and not because of something wrong with the data. Right?

Yes, this is true.

> In that case, should we just log the error rather than returning it?

I was thinking that if there is a bug the process batch should indicate this to the client rather than only logging it. This would give the error more visibility. I don't have a strong preference either way, what do you prefer?

x-pack/apm-server/aggregation/txmetrics/aggregator.go
lahsivjar: I didn't do this because the `histogramPool` is defined in the aggregator and resetting + recycling the histogram would mean that a `reset` either is a method on aggregator struct or has a reference to the pool passed to it. Both approaches didn't feel clean. Do you have any suggestions on how to handle this?

axw: Hm yeah I see. I think I'd be inclined to pass the pool into `reset`, just like you've done with `newMetricsEntry`. Maybe not the cleanest, but it's at least keeps the create/destroy code similar.

axw: Hah, just saw your commit - you beat me to it :)

lahsivjar: > or has a reference to the pool passed to it

Added a [commit for this](https://github.com/elastic/apm-server/pull/10362/commits/78fdd93390a1efe1a58cf847531e30c3a66ae15c). It doesn't look that bad now 😅, WDYT?

axw: Yup, LGTM - thanks.

x-pack/apm-server/aggregation/txmetrics/aggregator.go
```diff
@@ -611,28 +608,152 @@ func makeMetricset(key transactionAggregationKey, metrics transactionMetrics, in
...
+func (m *metrics) newServiceEntry(svcName string, overflow bool) (*svcMetricsMapEntry, error) {
+	var err error
+	if overflow {
+		// put the overflow bucket as a service map to avoid handling it separately
+		// while publishing the metrics.
+		svc, ok := m.m[overflowBucketName]
+		if !ok {
+			svc, err = m.svcSpace.Next()
+			if err != nil {
+				return svc, err
```
marclop, line 668: same here

x-pack/apm-server/aggregation/txmetrics/aggregator.go
```diff
@@ -611,28 +608,152 @@ func makeMetricset(key transactionAggregationKey, metrics transactionMetrics, in
...
+			}
+			m.m[overflowBucketName] = svc
+		}
+		return svc, nil
+	}
+
+	svc, err := m.svcSpace.Next()
+	if err != nil {
+		return svc, err
```
marclop, line 677: same here

marclop (commented):
Thanks for the refactor, the application flow is much easier to read and reason about now.

x-pack/apm-server/aggregation/txmetrics/aggregator.go
```diff
@@ -303,147 +310,137 @@ func (a *Aggregator) ProcessBatch(ctx context.Context, b *model.Batch) error {
...
-		} else {
-			svcEntry = &m.svcSpace[m.services]
-			m.services++
-			m.m[key.serviceName] = svcEntry
-		}
-		if svcEntry.m == nil {
-			svcEntry.m = make(map[uint64][]*metricsMapEntry)
-		}
+	svc, entry, _ = m.searchMetricsEntry(hash, key, offset)
+	if entry != nil {
```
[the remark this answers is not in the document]
lahsivjar: Actually, in both cases just checking if entry is not nil is enough. I was being lazy in defining a `var` without the lint complaining 😅. I will fix it, thanks.

marclop (approved):
Thanks!

axw:
Verified with BC5, following steps in the description -- I sent 500 unique transaction names every minute, for 10 minutes.

For good measure I also tested that sending 500 unique transaction every second still produced an `_other` metric.

PR #10367, opened 2023-02-28
fix: remove hardcoded timeout and don't block when fetching sourcemaps

kruskall:
## Motivation/summary

A couple of changes:

- revert the delete hook in systemtests: `TestNoMatchingSourcemaps` was failing because the resulting document had a sourcemap applied. It looks like there was a sourcemap in ES despite the `cleanupElasticsearch` call. Reverting the delete hook seems to make things more deterministic.
- remove hardcoded timeouts: both the 1s and the 500ms timeout are gone.
- the sourcemap_fetcher is not blocking anymore as a side effect sourcemap remapping won't work before init has completed

 

## Checklist

## Related issues

Related to https://github.com/elastic/apm-server/issues/10338


internal/beater/beater_test.go
```diff
@@ -76,6 +77,9 @@ func TestStoreUsesRUMElasticsearchConfig(t *testing.T) {
 	)
 	require.NoError(t, err)
 	defer cancel()
+
+	time.Sleep(1 * time.Second)
```
axw, line 81: is this leftover from debugging?

systemtest/kibana.go
```diff
@@ -314,5 +314,27 @@ func CreateSourceMap(t testing.TB, sourcemap []byte, serviceName, serviceVersion
...
 	})
 
+	t.Cleanup(func() {
+		DeleteSourceMap(t, result.ID)
+	})
+
 	return result.ID
 }
+
+// DeleteSourceMap deletes a source map with the given ID.
```
axw, line 324: FWIW I think this is cleaner anyway. We upload source maps through Kibana, so we should delete them through Kibana too; otherwise our system tests may be affected by internal Kibana changes.

internal/sourcemap/metadata_fetcher.go
```diff
@@ -101,9 +101,9 @@ func (s *MetadataESFetcher) startBackgroundSync(parent context.Context) {
 	go func() {
 		s.logger.Debug("populating metadata cache")
 
-		ctx, cancel := context.WithTimeout(parent, 1*time.Second)
-		err := s.ping(ctx)
-		cancel()
+		// the parent ctx does not have a deadline so we
+		// rely on the es client timeout (default 5s).
+		err := s.ping(parent)
```
axw, line 106: Do we still need the ping? Can we get rid of the the first run code, and just keep the loop  below and close `s.init` on the first successful call to `s.sync`?

Like this:

```go
func (s *MetadataESFetcher) ready() <-chan struct{} {                        
        return s.readyChan                                                   
}                                                                            

func (s *MetadataESFetcher) startBackgroundSync(ctx context.Context) {
        go func() {
                defer close(s.invalidationChan)
                defer s.logger.Info("sourcemap metadata update routine done")
                s.logger.Debug("populating sourcemap metadata cache")
                                                                             
                ticker := time.NewTicker(30 * time.Second)
                defer ticker.Stop()
                first := make(chan struct{})
                close(first)

                var setAvailable bool
                for {
                        select {
                        case <-ctx.Done():
                                return
                        case <-first:
                                first = nil
                        case <-ticker.C:
                        }
                        if err := s.sync(ctx); err != nil {
                                s.logger.With(
                                        logp.Error(err),
                                ).Error("failed to sync sourcemaps metadata")
                                continue
                        }
                        if !setAvailable {
                                setAvailable = true
                                close(s.readyChan)
                        }
                }
        }()
}
```

axw (changes requested):
Thanks, this is looking better.

We should pretty much never have a `time.Sleep(...)` in a test -- that will lead to flakiness. So if that wasn't just for debugging, let's find a deterministic alternative please.

internal/beater/beater_test.go
kruskall: No, this was to ensure the metadata fetcher init was completed.

I removed the time.Sleep and used a more deterministic approach

internal/sourcemap/metadata_fetcher.go
kruskall: I got rid of the ping and the arbitrary 10s timeout. Sync was update to return errFetcherUnavailable on err.

internal/sourcemap/fetcher.go
```diff
@@ -38,6 +38,8 @@ type Fetcher interface {
 	//
 	// If there is no such source map available, Fetch returns a nil Consumer.
 	Fetch(ctx context.Context, name string, version string, bundleFilepath string) (*sourcemap.Consumer, error)
+
+	Wait()
```
axw, line 42: Please don't add a Wait method to the interface. This is irrelevant to all but one implementation, which is a good indicator that the method belongs only on the one implementation.

internal/beater/beater_test.go
```diff
@@ -76,6 +76,9 @@ func TestStoreUsesRUMElasticsearchConfig(t *testing.T) {
 	)
 	require.NoError(t, err)
 	defer cancel()
+
+	fetcher.Wait()
```
axw, line 80: As mentioned elsewhere, we should not be adding methods to the interface just for tests. This complicates all of the implementations.

What I suggest we do instead is just create the fetcher, but don't invoke it. It should poll the test server in the background. Then in the test server, send to a channel rather than setting `called = true`, and wait for that channel to be signalled here.

I don't think we need to invoke the fetcher here. The basic functionality is tested in internal/sourcemap.

axw (changes requested):
Thanks, almost there. Just needs a change in the tests, otherwise looks good.

internal/sourcemap/fetcher.go
kruskall: Updated :+1:

internal/beater/beater_test.go
kruskall: Updated :+1:

axw (approved):
LGTM, thanks

axw:
Repeated the tests in https://github.com/elastic/apm-server/pull/9722 with BC5, all good now. Thanks @kruskall!

PR #10369, opened 2023-02-28
.ci: fix update-beats branch condition

axw:
Set `disablesourceinput: true` on the update-beats branch condition to stop the commit hash being added to the command line.


reakaleek (approved):
Oh, thank you for noticing this silently failing check and fixing it!

PR #10378, opened 2023-02-28
.ci: fix update-beats branch condition, take 3

axw:
Remove the condition altogether. The Github Action already checks which branches are active releases. If we should still have this check, I think it should be a separate Github Action step, and not an updatecli condition.


marclop (approved):
🙏🏼 🤞🏼

PR #10394, opened 2023-03-01
mage: set safe.directory for crossbuild

axw:
## Motivation/summary

Like https://github.com/elastic/beats/pull/34614, but for the apm-server directory. Works around strict directory ownership checks that fail in the Docker crossbuild image. Only relevant to the 7.17 branch.

## Checklist

## How to test these changes

`SNAPSHOT=true PLATFORMS=darwin/amd64 TYPES=docker mage package`

## Related issues

None

### DIFF
--- a/magefile.go
+++ b/magefile.go
@@ -72,6 +72,15 @@ func Build() error {
 // GolangCrossBuild build the Beat binary inside of the golang-builder.
 // Do not use directly, use crossBuild instead.
 func GolangCrossBuild() error {
+	// beats/dev-tools sets the *beats* directory as a safe directory,
+	// we need to do the same for apm-server since we're out of tree.
+	repo, err := mage.GetProjectRepoInfo()
+	if err != nil {
+		panic(err)
+	}
+	if err := sh.Run("git", "config", "--global", "--add", "safe.directory", repo.RootDir); err != nil {
+		return err
+	}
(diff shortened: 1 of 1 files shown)


marclop (approved):

magefile.go
```diff
@@ -72,6 +72,15 @@ func Build() error {
 // GolangCrossBuild build the Beat binary inside of the golang-builder.
 // Do not use directly, use crossBuild instead.
 func GolangCrossBuild() error {
+	// beats/dev-tools sets the *beats* directory as a safe directory,
+	// we need to do the same for apm-server since we're out of tree.
+	repo, err := mage.GetProjectRepoInfo()
+	if err != nil {
+		panic(err)
+	}
+	if err := sh.Run("git", "config", "--global", "--add", "safe.directory", repo.RootDir); err != nil {
```
kruskall, line 81: **issue**: this is changing the global config. I think we should either reset the setting once we are done or try to use something less invasive (https://git-scm.com/docs/git-config/#Documentation/git-config.txt-coresharedRepository ?)

kruskall: Looks like I was too late :facepalm:

axw: @kruskall it would probably be better, but on the other hand this is only run within Docker containers, and only in 7.17. I think we should just do what beats is doing and not spend a lot of time on it.

PR #10424, opened 2023-03-06
Rename organisation label to obs as per guidelines

lahsivjar:
Rename organisation label to use `obs` instead of `observability`


marclop (approved):

kruskall (approved):

lahsivjar:
Confirmed that aws-lambda uses the tags module ([smoke test](https://github.com/elastic/apm-aws-lambda/blob/b654e78e741d08140f7eaf6f108a2af5ca2d52f7/tf/main.tf#L5-L8), [benchmarking](https://github.com/elastic/apm-aws-lambda/blob/b654e78e741d08140f7eaf6f108a2af5ca2d52f7/testing/benchmarking/main.tf#L39-L42))

PR #10456, opened 2023-03-09
feat: stores app logs in service-specific data stream

### LINKED ISSUE #10291: Service-specific log data streams
## Problem Description

The APM Server currently sends all `log` events to one shared datastream `logs-apm.app-default`. This behavior is the same for OTLP and intake v2, but is not aligned with the best practices for the data stream naming convention. It also makes it difficult to have per-service settings for ILM, permissions, processing etc. It also means that the common field limit is exceeded more quickly.

## Solution
The behavior of the APM Server should be changed to automatically send logs to `logs-apm.app.<service>-<namespace>` instead of `logs-apm.app-default`

The server relies on Fleet for the index & component template setup, which does not (yet) support automatic creation of new templates for dynamic datastreams. The topic of creating service specific ILM policies won't be solved with this feature, but it sets up the general data structure for this.

kyungeunni:
## Motivation/summary
As described in #10291, we want to store app logs per service to follow best practices for the data stream naming convention.

## Checklist

## How to test these changes
1. Spin up tilt
2. Fire some events to intake endpoints
example:
```
curl --request POST \
  --url http://localhost:8200/intake/v2/events \
  --header 'content-type: application/x-ndjson' \
  --data '{"metadata":{"service":{"name":"test-service-1","version":"4.3.0","environment":"production", "language":{"name":"Java","version":"10.0.2"},"agent":{"version":"1.10.0","name":"java"}}, "labels":{"group":"experimental","ab_testing":true,"segment":5}}}
{"log": {"message": "test log message without timestamp"}}
{"log": {"trace.id": "trace-id", "transaction.id": "txn-id", "log.level": "warn", "log":{"logger":"testLogger","origin":{"file":{"name":"testFile","line":10},"function":"testFunc"}},"service":{"name":"test-service-2","version":"v1.0.0","environment":"prod"},"message": "test log message with nested ecs fields", "@timestamp": 1662616049000000}}'
```
3. Go to Kibana, check the datastreams created with the service names by querying:
```
GET logs-apm.app*/_search 
{
  "_source": "false"
}

# datastrteam `logs-apm.app.test_service_1-default` and `logs-apm.app.test_service_2-default` are created
```
4. Also check the correct index pattern is created.
image

## Related issues
closes #10291 
depends on https://github.com/elastic/apm-data/pull/23

### DIFF
--- a/NOTICE.txt
+++ b/NOTICE.txt
@@ -288,11 +288,11 @@ SOFTWARE.
 
 --------------------------------------------------------------------------------
 Dependency : github.com/elastic/apm-data
-Version: v0.1.1-0.20230223083125-b7f31c1a6768
+Version: v0.1.1-0.20230306102740-534660e4a151
 Licence type (autodetected): Apache-2.0
 --------------------------------------------------------------------------------
 
-Contents of probable licence file $GOMODCACHE/github.com/elastic/apm-data@v0.1.1-0.20230223083125-b7f31c1a6768/LICENSE:
+Contents of probable licence file $GOMODCACHE/github.com/elastic/apm-data@v0.1.1-0.20230306102740-534660e4a151/LICENSE:
 
                                  Apache License
                            Version 2.0, January 2004
--- a/NOTICE.txt
+++ b/NOTICE.txt
@@ -288,11 +288,11 @@ SOFTWARE.
 
 --------------------------------------------------------------------------------
 Dependency : github.com/elastic/apm-data
-Version: v0.1.1-0.20230306102740-534660e4a151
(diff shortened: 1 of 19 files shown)


marclop (commented):
Looks good! But I think the regular changelog should also have an entry, since that's where most of our users will look at.

kyungeunni:
Thanks @marclop, updated!

marclop (approved):
LGTM!

axw:
Tested with 8.8.0 BC1 in ESS. I sent the payload from the description, and confirmed the data streams were created as expected.

PR #10462, opened 2023-03-09
feat: add APIKey for apmsoak

### LINKED ISSUE #10443: apmsoak: add API Key auth support
We should extend `apmsoak` with support for API Key auth, to enable soak testing of APM Servers without secret token auth support.

kyungeunni:
## Motivation/summary
Support API key support for apmsoak. Deliberately didn't update `benchtest` as it's out of scope even though it uses the same `loadgen` module.

## How to test these changes
run apmsoak with apikey instead of secret token

## Related issues
closes #10443


systemtest/loadgen/eventhandler.go
```diff
@@ -35,13 +35,13 @@ var events embed.FS
 
 // NewEventHandler creates a eventhandler which loads the files matching the
 // passed regex.
-func NewEventHandler(p, url, token string, l *rate.Limiter) (*eventhandler.Handler, error) {
+func NewEventHandler(p, url, token string, l *rate.Limiter, apiKey string) (*eventhandler.Handler, error) {
```
kyungeunni, line 38: I wanted to put `apiKey` close to `token` param but it could break the existing usage so I put it as the last argument just to be safe, but I'm happy to reorganise the implementations with the new order after putting the `apiKey` next to `token`. 

In my previous experience, I prefer to pass params as a struct (object to be precise as it was javascript :p) when there are many arguments for a function so that the order of the params doesn't matter.

testing/infra/terraform/modules/soaktest_workers/README.md
```diff
@@ -35,6 +35,7 @@ This module sets up worker with load generation binary for soaktest configured a
 | <a name="input_apm_loadgen_agents_replicas"></a> [apm\_loadgen\_agents\_replicas](#input\_apm\_loadgen\_agents\_replicas) | Number of agents replicas to use, each replica launches 4 agents, one for each type | `string` | n/a | yes |
 | <a name="input_apm_loadgen_max_rate"></a> [apm\_loadgen\_max\_rate](#input\_apm\_loadgen\_max\_rate) | Max load generation rate | `string` | n/a | yes |
 | <a name="input_apm_secret_token"></a> [apm\_secret\_token](#input\_apm\_secret\_token) | Secret token for auth against the given server URL | `string` | n/a | yes |
+| <a name="input_apm_api_key"></a> [apm\_api\_key](#input\_apm\_api\_key) | API Key for auth against the given server URL. `apm_api_key` takes precedence over `api_secret_token` | `string` | n/a | yes |
```
axw, line 38: Did you hand-write this? Or did you miss adding the change to `variables.tf`? The docs are/should be generated from the variable definitions.

systemtest/loadgen/eventhandler.go
axw: > In my previous experience, I prefer to pass params as a struct (object to be precise as it was javascript :p) when there are many arguments for a function so that the order of the params doesn't matter.

I would be happy to introduce a config/params struct for this function.

axw (commented):
Looks good! I think there's a missing change to `variables.tf` though?

testing/infra/terraform/modules/soaktest_workers/README.md
kyungeunni: Yes I did hand-write this :D thanks, I didn't know this is auto-generated file. Will add it to `variables.tf` 👍

systemtest/loadgen/eventhandler.go
```diff
@@ -33,15 +33,23 @@ import (
 //go:embed events/*.ndjson
 var events embed.FS
 
+type EventHandlerParams struct {
+	Path    string
+	Url     string
```
axw, line 38:
```suggestion
	URL     string
```

Nit: let's please follow https://github.com/golang/go/wiki/CodeReviewComments#initialisms

axw (approved):
LGTM!

systemtest/loadgen/eventhandler.go
kyungeunni: Oh thansk, will update 👍

PR #10468, opened 2023-03-10
feat: add `burst` option for apmsoak

### LINKED ISSUE #10444: apmsoak: add support for generating traffic bursts
`apmsoak` currently supports specifying a maximum event rate (eps), so we can verify that APM Server is able to handle a sustained workload. This also sets the burst size to `max(1000, 2*eps)`.

We should expand on this with support for configuring the burst size, to verify the server's ability to handle sustained bursty workloads. For example, it should be possible to configure `apmsoak` to send a burst of 10000 events every 5 seconds.

kyungeunni:
## Motivation/summary
Support for configuring the burst size, to verify the server's ability to handle sustained bursty workloads.

## Related issues
closes #10444


axw (changes requested):
@kyungeunni the code all looks great, but I don't think it quite does what I had in mind. That's my mistake, I had misremembered what apmsoak already did.

At the moment, apmsoak sends batches in the same size as they are recorded, but with a configured rate limit. Before sending a batch, rate limiter tokens are acquired for the batch's full size. So if the batch is N events, and we configure the rate limit to be 1 event/sec, apmsoak will do nothing for N seconds and then send all the events at once.

So it's already kind of bursty, but we don't have a good way to control the rate of bursts. We also don't have a way to make apmsoak send events with a steady rate. We'll need to address both of those things. At least for the latter, I think we'll need to split the batches according to the configured event rate.

kyungeunni:
closing as it needs to take a different approach

PR #10492, opened 2023-03-20
Adding R8 stacktrace deobfuscator

LikeTheSalad:
## Motivation/summary

Android app crashes' stacktraces are often obfuscated in production for security reasons. We need to allow our users to get their stacktraces deobfuscated, otherwise app crash analysis would be very difficult if not impossible for them.

## Checklist

## Related issues

https://github.com/elastic/apm-agent-android/issues/32


internal/r8/deobfuscator.go
```diff
@@ -0,0 +1,222 @@
...
+	var currentType *MappedType
+	var currentMappedMethodCall *MappedMethodCall
+
+	for scanner.Scan() {
+		line := scanner.Text()
+		typeMatch := typePattern.FindStringSubmatch(line)
+		if typeMatch != nil {
+			currentMappedMethodCall = nil
+			if currentType != nil {
+				ensureAllMethodsProvided(&res, currentType)
```
carsonip, line 128: This line is only called on every new type but not called at the end of loop (end of file), which is an edge case

testdata/r8/deobfuscator/1/de-obfuscated-crash.txt
```diff
@@ -0,0 +1,22 @@
+E/AndroidRuntime: FATAL EXCEPTION: main
```
carsonip, line 1: I wonder if it will pass windows CI if we remove `.txt` extension from de-obfuscated-crash.txt and obfuscated-crash.txt. I believe it will stop git cloning the txt with crlf and fix the tests.

internal/r8/deobfuscator.go
```diff
@@ -0,0 +1,222 @@
...
+			return v
+		}
+	}
+	return nil
+}
+
+func findMappingFor(symbols []StacktraceType, mapFilePath string) map[string]string {
+	mapFile, err := os.Open(mapFilePath)
+	if err != nil {
+		log.Fatal(err)
```
carsonip, line 110: log.Fatal will terminate the program with non-0 exit code. I suggest that we propagate the error to caller instead

internal/r8/deobfuscator.go
```diff
@@ -0,0 +1,222 @@
...
+		} else if currentType != nil {
+			methodMatch := methodPattern.FindStringSubmatch(line)
+			if methodMatch != nil {
+				currentMappedMethodCall = handleMappedMethodCall(&res, methodMatch, currentType, currentMappedMethodCall)
+			}
+		}
+	}
+
+	if err := scanner.Err(); err != nil {
+		log.Fatal(err)
```
carsonip, line 146: Propagate the error

carsonip (changes requested):
Thanks for the PR and welcome to Go! PR works well and most of the comments are around coding style in Go. A lot of the utility functions can be removed once `map` is used instead of slices.

testdata/r8/deobfuscator/1/de-obfuscated-crash.txt
LikeTheSalad: I've added those changes, let's wait to see if that works!

changelogs/8.7.asciidoc
```diff
@@ -50,3 +50,4 @@ https://github.com/elastic/apm-server/compare/8.6\...8.7[View commits]
 - Use Elasticsearch fetcher for agent remote configuration where possible {pull}9720[9720] {pull}10088[10088]
 - Fetch sourcemaps from Elasticsearch {pull}9722[9722]
 - Add overflow metrics for transaction, span metrics, service transaction and service summary metrics {pull}10330[10330]
+- R8 stacktrace deobfuscator {pull}10492[10492]
```
carsonip, line 53: The changelog line should live in head.asciidoc, instead of 8.7.

testdata/r8/deobfuscator/1/de-obfuscated-crash.txt
carsonip: I managed to fix it by forcing LF line endings on all platforms. See https://github.com/elastic/apm-server/pull/10545/commits/9516803085a614a4be95608d74333a138b84fd3d

internal/r8/deobfuscator.go
```diff
@@ -0,0 +1,177 @@
...
+	return res, nil
+}
+
+func handleMappedMethodCall(res map[string]string, methodMatch MethodMatch, currentType *MappedType, currentMappedMethodCall *MappedMethodCall) *MappedMethodCall {
+	methodNameReference := methodMatch.methodObfuscatedName
+	if methodMatch.sourceFileStart != "" {
+		if methodMatch.sourceFileStart != methodMatch.sourceFileEnd {
+			// This is probably due an edge-case where the mapping line starts with different numbers (e.g 1:2). We don't
+			// have that case in our tests, therefore we are ignoring it.
+			return currentMappedMethodCall
```
carsonip, line 149: Thanks, the comment is helpful. I'm wondering if we want to log a warning for this case where we haven't encountered but something that might happen in live environment. We can discuss this later.

internal/r8/deobfuscator.go
```diff
@@ -0,0 +1,205 @@
...
+
+// Sometimes the map file contains a type but not a method name belonging to said type, this might happen because the type name was obfuscated
+// but its method wasn't, in this case we use this function to make sure we add the leftover de-obfuscated type names from the map file,
+// along with the method names as found in the stacktrace (which should be de-obfuscated already).
+func mapLeftoverUnmappedMethods(res map[string]string, currentType *MappedType) {
+	for methodName, callSite := range currentType.obfuscated.methods {
+		key := getKey(currentType.obfuscated.name, methodName, callSite)
+		res[key] = getKey(currentType.realName, methodName, callSite)
+	}
+}
```
carsonip, line 201: Sorry I missed this `res` earlier
```suggestion
func mapLeftoverUnmappedMethods(mapping map[string]string, currentType *MappedType) {
	for methodName, callSite := range currentType.obfuscated.methods {
		key := getKey(currentType.obfuscated.name, methodName, callSite)
		mapping[key] = getKey(currentType.realName, methodName, callSite)
	}
}
```

internal/r8/deobfuscator.go
```diff
@@ -0,0 +1,205 @@
...
+	methods     map[string]string // Maps method references to their stacktrace call site
+}
+
+type MappedType struct {
+	obfuscated StacktraceType
+	realName   string
+}
+
+type MappedMethodCall struct {
+	reference string
```
carsonip, line 40: a comment to describe `reference` would be good. It is `obfuscatedTypeName:obfuscatedMethodName(:sourcefileStart)` if i understand correctly

internal/r8/deobfuscator.go
```diff
@@ -0,0 +1,205 @@
...
+		methodNameReference = fmt.Sprintf("%s:%s", methodMatch.methodObfuscatedName, methodMatch.sourceFileStart)
+	}
+	mapReference := currentType.obfuscated.name + ":" + methodNameReference
+	methodCallSite, ok := currentType.obfuscated.methods[methodNameReference]
+	if ok {
+		// Found this method in the list of methods parsed from the stacktrace for the currentType.
+		delete(currentType.obfuscated.methods, methodNameReference)
+		key := getKey(currentType.obfuscated.name, methodMatch.methodObfuscatedName, methodCallSite)
+		mapping[key] = getKey(currentType.realName, methodMatch.methodRealName, methodCallSite)
+		return &MappedMethodCall{mapReference, key}
```
carsonip, line 184: nit: minor readability improvement
```suggestion
		return &MappedMethodCall{reference: mapReference, key: key}
```

internal/r8/deobfuscator_test.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+}
+
+func readFile(path string) string {
+	bytes, err := os.ReadFile(path)
+	if err != nil {
+		log.Fatal(err)
+	}
+	return string(bytes)
+}
```
carsonip, line 64: Fail the test instead of logging in test. `log.Fatal` will exit with 1 instead of just failing the test
```suggestion
func readFile(t *testing.T, path string) string {
	bytes, err := os.ReadFile(path)
	require.Nil(t, err)
	return string(bytes)
}
```

internal/r8/deobfuscator_test.go
```diff
@@ -0,0 +1,64 @@
...
+		t.Run(fmt.Sprintf("(%s)->(%s)", inputPath, expectedOutputPath), func(t *testing.T) {
+			obfuscated := readFile(inputPath)
+			expected := readFile(expectedOutputPath)
+			deobfuscated, _ := Deobfuscate(obfuscated, reader)
+			assert.Equal(t, expected, deobfuscated)
+			err := reader.Close()
+			if err != nil {
+				log.Fatal(err)
+			}
+		})
```
carsonip, line 54: Fail the test instead of logging in test. Isolate the failure to the test case only.
```suggestion

		t.Run(fmt.Sprintf("(%s)->(%s)", inputPath, expectedOutputPath), func(t *testing.T) {
			t.Parallel()
			reader, err := os.Open(mapFilePath)
			require.Nil(t, err)
			defer reader.Close()
			obfuscated := readFile(t, inputPath)
			expected := readFile(t, expectedOutputPath)
			deobfuscated, _ := Deobfuscate(obfuscated, reader)
			assert.Equal(t, expected, deobfuscated)
		})
```

carsonip (approved):

PR #10508, opened 2023-03-22
benchmarks: run benchmarks 6 times, for 95% CI

axw:
## Motivation/summary

With the recent changes to `benchstat` output, it became obvious that we're not capturing enough samples. We should run the benchmarks at least 6 times each for a 95% confidence interval.

## Checklist

## How to test these changes

N/A

## Related issues

None


testing/benchmark/Makefile
```diff
@@ -6,7 +6,7 @@ TFVARS_SOURCE ?= terraform.tfvars.example
 
 BENCHMARK_WARMUP_TIME ?= 5m
 BENCHMARK_AGENTS ?= 64
-BENCHMARK_COUNT ?= 3
+BENCHMARK_COUNT ?= 6
```
marclop, line 9: We should also double the benchmark timeout to 8h here: https://github.com/elastic/apm-server/blob/30f2301fd67082fefd0f59aedf5055df72b7f691/.ci/benchmarks.groovy#L22

axw: 8h :flushed:

OK

marclop: yeah... That would be the worst case scenario. In practice it's unlikely that the benchmarks run for more than 3 hours, unless the first run fails and the retry runs for the full length.

marclop (approved):
Looks good!

PR #10563, opened 2023-03-30
feat: steady rate load generation for loadgen

### LINKED ISSUE #10444: apmsoak: add support for generating traffic bursts

kyungeunni:
## Motivation/summary
Support for configuring the burst size and interval to verify the server's ability to handle sustained bursty workloads. 

## Checklist

## Related issues
closes #10444

### DIFF
--- a/systemtest/loadgen/config/config.go
+++ b/systemtest/loadgen/config/config.go
@@ -24,6 +24,7 @@ import (
 	"os"
 	"strconv"
 	"strings"
+	"time"
 )
 
 var Config struct {
@@ -31,10 +32,46 @@ var Config struct {
 	SecretToken       string
 	APIKey            string
 	Secure            bool
-	MaxEPM            int
+	EventRate         RateFlag
 	RewriteIDs        bool
 	RewriteTimestamps bool
 	Headers           map[string]string
+}
+
+type RateFlag struct {
+	Burst    int
+	Interval time.Duration
+}
+
+func (f *RateFlag) String() string {
+	return fmt.Sprintf("%d/%s", f.Burst, f.Interval)
+}
+
+func (f *RateFlag) Set(s string) error {
+	before, after, ok := strings.Cut(s, "/")
+	if !ok || before == "" || after == "" {
+		return fmt.Errorf("invalid rate %q, expected format burst/duration", s)
+	}
+
+	burst, err := strconv.Atoi(before)
+	if err != nil {
+		return fmt.Errorf("invalid burst %s in event rate: %w", before, err)
+	}
+
+	if !(after[0] >= '0' && after[0] <= '9') {
+		after = "1" + after
+	}
+	interval, err := time.ParseDuration(after)
+	if err != nil {
+		return fmt.Errorf("invalid interval %q in event rate: %w", after, err)
+	}
+	if interval <= 0 {
+		return fmt.Errorf("invalid interval %q, must be positive", after)
+	}
+
+	f.Burst = burst
+	f.Interval = interval
+	return nil
 }
 
 func init() {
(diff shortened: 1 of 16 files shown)


systemtest/loadgen/config/config.go
```diff
@@ -52,29 +53,42 @@ func init() {
...
+				Config.Interval = t
 			case "m":
-				Config.MaxEPM = rateVal
+				Config.Interval = t * 60
 			default:
 				return fmt.Errorf(errStr, rate)
 			}
+			Config.EPS = burst / Config.Interval
 			return nil
 		})
```
axw, line 91: IMO this code has become sufficiently complicated that we should create a custom flag:

```go
type RateFlag struct {
        burst    int
        interval time.Duration
}

func (f *RateFlag) String() string {
        return fmt.Sprintf("%d/%s", f.burst, f.interval)
}

func (f *RateFlag) Set(s string) error {
        before, after, ok := strings.Cut(s, "/")
        if !ok || before == "" || after == "" {
                return fmt.Errorf("invalid rate %q, expected format burst/duration", s)
        }

        burst, err := strconv.Atoi(before)
        if err != nil {
                return fmt.Errorf("invalid burst %s in event rate: %w", before, err)
        }

        if !(after[0] >= '0' && after[0] <= '9') {
                after = "1" + after
        }
        interval, err := time.ParseDuration(after)
        if err != nil {
                return fmt.Errorf("invalid interval %q in event rate: %w", after, err)
        }
        if interval <= 0 {
                return fmt.Errorf("invalid interval %q, must be positive", after)
        }

        f.burst = burst
        f.interval = interval
        return nil
}
```

(May require some changes, I haven't fully tested that.)

axw (commented):
Looks good! My comments are mostly about readability than anything. I think the functionality looks sound.

systemtest/loadgen/eventhandler/handler.go
```diff
@@ -266,10 +266,45 @@ func (h *Handler) sendBatch(
...
+		err := h.sendHTTPRequest(ctx, b, baseTimestamp, randomBits)
+		if err != nil {
+			return 0, err
+		}
+		return len(b.events), nil
 	}
 
+	events := make([]event, len(b.events))
+	copy(events, b.events)
+	eventsLen := len(b.events)
```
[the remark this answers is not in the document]
kyungeunni: `events` slice is reassigned later on ([line 295](https://github.com/elastic/apm-server/pull/10563/files#diff-a9132018cf11cb7b898a19b33d048de3f436b455881eca781ab0a98ae3d969ffR295)) 
Instead, I can just keep the last index and avoid copying all the events. 

Regarding `eventsLen`, yes we can do that. `len(b.events)` was used everywhere and I kind of preferred to have it as a named variable, but no strong opinion. Will update!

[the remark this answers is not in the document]
axw: `events = events[n:]` doesn't modify the memory of the original `events` slice, it creates a new slice, offset by `n` elements from the beginning of the original slice. There's no need to copy just for that.

systemtest/loadgen/limiter.go
```diff
@@ -18,24 +18,12 @@
...
 
 import (
-	"math"
-
 	"golang.org/x/time/rate"
 )
 
-func GetNewLimiter(epm int) *rate.Limiter {
-	if epm <= 0 {
+func GetNewLimiter(eps int, interval int) *rate.Limiter {
```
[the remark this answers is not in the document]
kyungeunni: Agreed to all 👍

systemtest/loadgen/eventhandler/handler.go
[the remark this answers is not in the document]
kyungeunni: Oh, I see. I must have misread about the underlying array can be affected when doing so. TIL, thanks!

systemtest/loadgen/config/config.go
kyungeunni: It works like a charm 👍

systemtest/loadgen/limiter.go
```diff
@@ -18,24 +18,21 @@
...
 import (
-	"math"
+	"time"
 
 	"golang.org/x/time/rate"
 )
 
-func GetNewLimiter(epm int) *rate.Limiter {
-	if epm <= 0 {
+// Return a rate limiter that is configured to rate limit burst per interval.
```
axw, line 26:
```suggestion
// GetNewLimiter returns a rate limiter that is configured to rate limit burst per interval.
```

It's conventional to start doc comments with the name of the method or type or whatever else you're documenting.

systemtest/loadgen/limiter.go
```diff
@@ -18,24 +18,21 @@
...
-}
-
-func getBurstSize(eps int) int {
-	burst := eps * 2
-	// Allow for a batch to have 1000 events minimum
-	if burst < 1000 {
-		burst = 1000
-	}
-	return burst
+	eps := float64(burst) / float64(interval/time.Second)
```
axw, line 36:
```suggestion
	eps := float64(burst) / interval.Seconds()
```

testing/apmsoak/terraform.tfvars.example
```diff
@@ -20,9 +20,9 @@ gcp_zone = "us-west2-b"
 # API Key for auth against the given server URL
 # apm_api_key = "<api_key>"
 
-# Max rate at which events will be sent to the APM server
-# Defaults to `4000eps` (i.e. 4000 events per second)
-# apm_loadgen_max_rate = "8000eps"
+# Event rate at which events will be sent to the APM server
+# Defaults to `4000/s` (i.e. 4000 events per second)
+# apm_loadgen_event_rate = "8000/s"
```
axw, line 25:
```suggestion
# apm_loadgen_event_rate = "4000/s"
```

The supposed default is out of sync with `variables.tf` and the comment above, let's fix it while we're here.

axw (approved):
LGTM! Just a few minor things

PR #10573, opened 2023-03-31
Universal Profiling: Versioned RPC protocol

christos68k:
Added checks for RPC protocol version mismatches. 

I tested with HA v2.4.1 and v2.5.2. HAs older than v2.4.1 (AFAIK we never made any such older version available to beta customers) that do not send RPC protocol version will error on startup. 

This should be part of the upcoming 8.7.1 release.


rockdaboot (approved):
LGTM

PR #10574, opened 2023-03-31
systemtest/loadgen: add `-rewrite-*-names` flags

### LINKED ISSUE #10446: apmsoak: support event field randomisation
`apmsoak` replays canned event payloads to APM Server in a loop. This means that data may be unrealistic, e.g. due to trace IDs being non-unique/repeated across distributed traces. This in turn affects features such as tail-based sampling, and may bias our indexing performance. Furthermore, it means we cannot use `apmsoak` to verify the server's ability to handle dynamic data, e.g. high cardinality transaction names.

We should provide a means of randomising event data -- ideally arbitrary fields. This must be done in a consistent manner per iteration, meaning that if two events have the same field value (e.g. same `trace.id`) then the field values will be adjusted in the same way.

As mentioned above, randomisation may also be useful for verifying the server's ability to handle high cardinality metric dimensions, such as `transaction.name` and `service.name`. `service.name` supports only a limited character set, so we will either need to hardcode this for the `service.name` field in the randomiser, or else we'll need to make the randomisation configurable; the former is probably sufficient for the foreseeable future, and would be simpler to use.

We will need to ensure low overhead, so `apmsoak` can continue to sustain high throughput.

axw:
## Motivation/summary

Add flags for apmsoak to rewrite name fields:

```
-rewrite-service-names service.name
      replace service.name in events
-rewrite-service-node-names service.node.name
      replace service.node.name in events
-rewrite-service-target-names service.target.name
      replace service.target.name in events
-rewrite-span-names span.name
      replace span.name in events
-rewrite-transaction-names transaction.name
      replace transaction.name in events
```

If any of these flags are specified, then the corresponding fields will be rewritten by replacing ASCII letters and digits with a random rune of the same category.

These flags are intended for testing the impacts of high cardinality.

## Checklist

## How to test these changes

Run `apmsoak` with each flag in turn, verifying the corresponding event fields are randomised.

## Related issues

Closes #10446

### DIFF
--- a/systemtest/loadgen/eventhandler/handler.go
+++ b/systemtest/loadgen/eventhandler/handler.go
@@ -33,6 +33,7 @@ import (
 
 	"github.com/klauspost/compress/zlib"
 	"github.com/tidwall/gjson"
+	"github.com/tidwall/sjson"
 	"go.elastic.co/fastjson"
 	"golang.org/x/time/rate"
 )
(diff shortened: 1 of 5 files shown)


kyungeunni (approved):
LGTM!!

systemtest/loadgen/eventhandler/handler.go
```diff
@@ -412,6 +508,46 @@ func randomizeTraceID(out *bytes.Buffer, in string, randomBits uint64) bool {
...
+	data, err := sjson.SetBytes(data, path, scratch.String())
+	if err != nil {
+		return nil, fmt.Errorf("failed to rewrite %q: %w", path, err)
+	}
+	return data, nil
+}
+
+// randomizeASCII replaces ASCII letter and digit runes in the input
+// with random ASCII runes in the same category.
+func randomizeASCII(out *bytes.Buffer, in string, randomBits uint64) {
```
marclop, line 528: This is pretty cool

PR #10591, opened 2023-04-03
feat: wait for interval and sends events at once

### LINKED ISSUE #10575: apmsoak: make event rate more regular
We can instruct the apmsoak to send `n` events to be sent per `t` duration since #10563 is introduced.

However, depending on the length of the source data, it is possible that the timing of sending the data can be shifted a bit while the rate is still valid.

For example, we have an event rate of 100/5s. The rate limiter will wait for 5 seconds when the source provided 100 events.
However, when 80 events are provided from the source, the rate limiter will wait only 4 seconds and send the available 80 events. The next batch job will start at 4 seconds mark so thus the expectation _sending 100 events every 5 seconds_ is not met.

We need to improve this and make the event rate is more regular and predictable.

kyungeunni:
## Motivation/summary
if event rate is `n/t`,
Send a burst of `n` data every `t` rather than spread the load over duration `t`

## Checklist

## How to test these changes
TBC

## Related issues
closes #10575

### DIFF
--- a/systemtest/loadgen/eventhandler/handler.go
+++ b/systemtest/loadgen/eventhandler/handler.go
@@ -259,6 +259,16 @@ func New(config Config) (*Handler, error) {
 	return &h, nil
 }
 
+// WarmUpServer will "warm up" the remote APM Server by sending events until
+// the context is cancelled.
+func (h *Handler) WarmUpServer(ctx context.Context) error {
+	for {
+		if _, err := h.SendBatches(ctx); err != nil {
+			return err
+		}
+	}
+}
+
 // SendBatches sends the loaded trace data to the configured transport,
 // returning the total number of documents sent and any transport errors.
 func (h *Handler) SendBatches(ctx context.Context) (int, error) {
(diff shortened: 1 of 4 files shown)


systemtest/loadgen/eventhandler/handler.go
```diff
@@ -71,8 +71,10 @@ type Handler struct {
 	config     Config
 	writerPool sync.Pool
 
-	batches      []batch
-	minTimestamp time.Time // across all batches
+	batches       []batch
+	queuedBatches []batch
+	queuedEvents  int
```
axw, line 76: Sorry I didn't notice this earlier: Handler may be used by multiple concurrent callers, so we can't store queuedBatches and queuedEvents on the struct. I think they would have to be stored on the stack in SendBatches.

systemtest/loadgen/eventhandler/handler.go
```diff
@@ -259,73 +259,107 @@ func New(config Config) (*Handler, error) {
...
+			return ctx.Err()
+		default:
+			if _, err := h.sendBatches(ctx, &s); err != nil {
+				return err
+			}
+			// reset after batches so it doesn't exceed th limit
+			// but keep the remainder so the next batches know where to start
+			if s.burst > 0 {
+				s.sent = s.sent % s.burst
+			}
```
axw, line 286: Does it matter if s.sent > s.burst? It was intentional that s.sent be the total events sent, not just the number of events sent for the current iteration.

axw (approved):
Just one question, non-blocking. LGTM, thank you :)

systemtest/loadgen/eventhandler/handler.go
kyungeunni: Oh, it's just a safeguard not to exceed the `math.MaxInt`

kyungeunni: Will revisit it when it becomes problematic.

systemtest/loadgen/eventhandler/handler.go
```diff
@@ -259,73 +259,107 @@ func New(config Config) (*Handler, error) {
...
+
+	for {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		default:
+			if _, err := h.sendBatches(ctx, &s); err != nil {
+				return err
+			}
+			// reset after batches so it doesn't exceed th limit
```
kyungeunni, line 282:
```suggestion
			// safeguard `s.sent` so that it doesn't exceed math.MaxInt
```

systemtest/loadgen/eventhandler/handler.go
kyungeunni: Updated the comment to explain what it does more clearly

PR #10622, opened 2023-04-06
Limit docker compose log max size

carsonip:
## Motivation/summary

Limit docker compose log max size to avoid taking up all the disk space when the containers run for a long time. This only affects local development and testing.

## Checklist

## How to test these changes

N/A


simitt (approved):

kruskall (commented):
**suggestion**: can we use extensions and fragments to reuse the configuration and avoid duplicate code ? (See https://docs.docker.com/compose/compose-file/11-extension/#using-extensions-as-fragments)

PR #10628, opened 2023-04-07
Fix recursive statement in docs

felixbarny:


pquentin (approved):
Thanks! That was confusing. :)

PR #10632, opened 2023-04-09
Use local Kibana in Tiltfile if available

dgieselaar:
Closes https://github.com/elastic/apm-dev/issues/937.

In short:
- if `--local-kibana` is set, checks port 5601 to see if there is a service running
- if it is, see if it is a Docker container or Kibana running locally in development mode
- if it's Kibana running locally, spawn a wait script that checks whether ES is available. this allows the user to start Kibana locally, without a working ES connection, and wait with installing the package until Kibana is available
- if there is no local Kibana, start the Kibana container
- Wait until either the local Kibana or the Kibana container is available when installing the package

I've added `KIBANA_BASE_PATH` to apm-server because in development mode, Kibana is often running on a different path rather than just the root.

### DIFF
--- a/Tiltfile
+++ b/Tiltfile
@@ -1,3 +1,44 @@
+def printup(*args):
+  if config.tilt_subcommand == "up":
+    print(*args)
+
+config.define_bool('local-kibana')
+
+parsed_config = config.parse()
+
+script_dir = os.path.join(config.main_dir, 'script')
+
+default_kibana_port = 5601
+kibana_host = "localhost"
+kibana_base_path = ""
+
+# checks whether (dev mode) kibana is already running at given port,
+# and returns the URL it is redirected to
+def discover_kibana():
+
+    default_kibana_host = "localhost"
+
+    count = int(
+      str(
+        local("lsof -n -i :%s | grep LISTEN | wc -l | tr -d ' '" % default_kibana_port, quiet=True, echo_off=True)
+      ).strip()
+    )
+
+    if count <= 0:
+      fail("No service found on port " + str(default_kibana_port))
+      return None
+
+
+    kibana_redirect_url = str(
+      local(
+        "curl --fail -s -I -o /dev/null -w '%{redirect_url}' " + 'http://{}:{}'.format(default_kibana_host, default_kibana_port) + ' || [ $? -eq 7 ]',
+        quiet=True,
+        echo_off=True
+      )
+    ).strip().replace("http://", "")
+
+    return kibana_redirect_url
+
 custom_build(
   'apm-server',
   'DOCKER_BUILDKIT=1 docker build -t $EXPECTED_REF -f packaging/docker/Dockerfile .',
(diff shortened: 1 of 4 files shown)


Tiltfile
```diff
@@ -1,31 +1,109 @@
...
+    cmd = 'ifconfig | grep "inet " | grep -v 127.0.0.1 | cut -d\\  -f2'
+    return str(local(cmd, quiet=True, echo_off=True)).strip()
+
+def printup(*args):
+  if config.tilt_subcommand == "up":
+    print(*args)
+
+script_dir = os.path.join(config.main_dir, 'script')
+
+default_kibana_port = 5601
```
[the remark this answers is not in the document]
dgieselaar: I think that's fine for now, I think it's fairly uncommon for folks to change this in development mode.

Tiltfile
```diff
@@ -1,31 +1,109 @@
...
+
+# checks whether (dev mode) kibana is already running at given port,
+# and returns the URL it is redirected to
+def discover_kibana():
+
+    default_kibana_host = "localhost"
+
+    count = int(
+      str(
+        local("lsof -i :%s | grep LISTEN | wc -l | tr -d ' '" % default_kibana_port, quiet=True, echo_off=True)
```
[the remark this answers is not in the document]
dgieselaar: Thanks! I'll make the change.

dgieselaar:
@marclop:
> Mostly looks good. I'm curious as to how the "external" Kibana will be running. Do you have a Mac? Or Linux-based system?
> 
> I'm asking because of Docker on Mac and how the networking works in K8s. I haven't tested this locally, but if you were to have a local process (not in K8s or docker). I'm not sure how the APM Server will reach the Kibana socket.

I was under the assumption that this was necessary so I started out by getting the host IP and then passing that on to APM Server, but because the APM package install process runs on the host, it's no longer necessary. We can add it later if we want APM Server to talk to Kibana.

dgieselaar:
I wanted to demo this but I ran into a discrepancy between a local Kibana instance and the docker image. In this case there's a breaking change in the mappings without a migration script (which happens between snapshot builds). I need to think about a way to deal with this because this ES instance ships with `action.destructive_requires_name` set to true, which makes it harder to resolve.

simitt:
drive-by comment: @axw might have additional input on this when returning from PTO.

SylvainJuge:
This sounds to be a good alternative to using a "patch" like https://github.com/elastic/apm-server/issues/10662 (and thus would also allow to close this issue).

Tiltfile
```diff
@@ -1,31 +1,105 @@
...
+
+script_dir = os.path.join(config.main_dir, 'script')
+
+default_kibana_port = 5601
+kibana_host = "localhost"
+kibana_path = ""
+
+# checks whether (dev mode) kibana is already running at given port,
+# and returns the URL it is redirected to
+def discover_kibana():
```
kruskall, line 13: **question**: Is there a reason people would run kibana outside of docker ? 

It feels like this discovery process is adding a lot of "responsibility"/"feature creep" as opposed to a custom path+overlay+docker

dgieselaar: Yes, during development, where it is not practical to wait for a docker rebuild of Kibana.

systemtest/apmservertest/config.go
```diff
@@ -38,6 +38,7 @@ const (
 	defaultKibanaPort = "5601"
 	defaultKibanaUser = "apm_user_ro"
 	defaultKibanaPass = "changeme"
+	defaultKibanaPath = ""
```
axw, line 41: Base path maybe? To me, "path" sounds like it's referring to a filesystem path to where Kibana is located.

Tiltfile
```diff
@@ -1,31 +1,105 @@
...
 )
 
+k8s_yaml(kustomize('testing/infra/k8s/overlays/local'))
+
+k8s_kind('ApmServer', image_json_path='{.spec.image}')
+k8s_kind('Elasticsearch')
+
+
+# Conditionally set up Kibana resource
+discovered_kibana_url = discover_kibana()
```
axw, line 64: I don't really love this as a default behaviour. It feels a bit too magical, and may lead to confusion if someone has Kibana running and didn't intend for it to be used by the Tilt setup.

What do you think about making it an explicit opt-in, via [Tilt config](https://docs.tilt.dev/tiltfile_config.html)?

Tiltfile
```diff
@@ -1,31 +1,105 @@
...
 # Build and install the APM integration package whenever source under
 # "apmpackage" changes.
-script_dir = os.path.join(config.main_dir, 'script')
 run_with_go_ver = os.path.join(script_dir, 'run_with_go_ver')
+
+local_resource(
+  "init_kibana_system_user",
+  cmd = """curl -s -X POST -H "Content-Type: application/json" -d '{"username":"kibana_system_user", "password":"changeme", "roles": [ "kibana_system"]}' http://admin:changeme@localhost:9200/_security/user/kibana_system_user""",
+  resource_deps=["elasticsearch"]
+)
```
axw, line 90: Is this relevant to the containerised Kibana, or only for the external one? I suppose the latter, seeing as we didn't do this before. Maybe it should be encapsulated in wait_for_kibana.sh (perhaps rename to init_kibana.sh?)?

Tiltfile
```diff
@@ -1,31 +1,105 @@
...
+
 local_resource(
   'apmpackage',
   cmd = [os.path.join(script_dir, 'run_with_go_ver'), 'go', 'run', './cmd/runapm -init'],
   dir = 'systemtest',
   deps = ['apmpackage'],
-  resource_deps=['kibana'],
+  resource_deps=['init_kibana_system_user'],
```
axw, line 97: Should this be depending on "kibana" still? The kibana local_resource depends on init_kibana_system_user.

axw (commented):
Looks good overall, but I'd much it were an explicit opt-in via config. I imagine that would be acceptable for UI developers, but WDYT?

dgieselaar:
@axw agreed on using config rather than inferring, I'll make that change next week (had to pick up some other urgent work this week).

systemtest/apmservertest/config.go
dgieselaar: changed to `defaultKibanaBasePath`

Tiltfile
dgieselaar: Made it opt in:
`tilt up -- --local-kibana`

Tiltfile
dgieselaar: Yes, I've changed this back to `kibana` because it was no longer necessary anyway. I can't remember why I added it here anyway :)

sqren (approved):
Took this for a spin. Works as expected, although data ingestion is a little tricky.

Tiltfile
```diff
@@ -1,31 +1,115 @@
...
+
+    count = int(
+      str(
+        local("lsof -n -i :%s | grep LISTEN | wc -l | tr -d ' '" % default_kibana_port, quiet=True, echo_off=True)
+      ).strip()
+    )
+
+    if count <= 0:
+      fail("No service found on port " + str(default_kibana_port))
+      return None
```
axw, line 29: Is this discovery logic still needed, now that we're explicitly saying to use a local kibana? I'd rather like to drop it if it's not necessary.

Tiltfile
```diff
@@ -1,31 +1,115 @@
...
+    is_docker_kibana = str(
+      local(
+        """curl --fail -s -w '%header{kbn-name}'""" + " http://{}:{}".format(default_kibana_host, default_kibana_port),
+        quiet=True,
+        echo_off=True
+      )
+    ).strip() == "kibana"
+
+    if (is_docker_kibana):
+      return None
```
axw, line 40: Same as above -- can we just assume that it's running on the host if `--local-kibana` has been passed?

axw (commented):
Basically LGTM, but can we trim down the Tiltfile a bit more?

Tiltfile
dgieselaar: This is needed, still, to get the redirect URL (which includes the kibana base path).

axw: The curl is needed, but the lsof bit above? It's second guessing the flag. Could we just have the curl request fail if kibana isn't running?

dgieselaar: IMHO I think it's a good thing, the flag doesn't control the service, an explicit failure is still helpful here, especially because curl doesn't really give user-friendly error messages. If you feel strongly about it I can remove it.

axw: OK, let's leave it for now. Just trying to keep this thing minimal and maintainable :)
Should be fine.

axw (approved):

axw:
@dgieselaar can you please squash to a signed commit and force-push? We recently added a requirement for commit signing.

Issue #10619, opened 2023-04-06
logging options ignored in containers or systemd environment


kruskall:
**APM Server version** (`apm-server version`): https://github.com/elastic/apm-server/commit/fa98d98a3fae65ac7e3cb91b7246b8636250030c

**Description of the problem including expected versus actual behavior**:

While working on https://github.com/elastic/apm-server/issues/10467 I was having issues trying to setup a provisioner that would copy the apm-server logs to the local environment before destroying the ec2 instance.
Further investigation revealed that the logging options were being ignored.
I assume that's because the process will always log to stderr in a container or systemd environment:

https://github.com/elastic/apm-server/blob/fa98d98a3fae65ac7e3cb91b7246b8636250030c/internal/beatcmd/logging.go#L56

I don't think we should ignore the logging option if logging to file has been explicitly set. 

The library we are using (elastic-agent-libs) already has something similar, falling back to stderr for specific environments if an explicit option has not been configured: https://github.com/elastic/elastic-agent-libs/blob/main/logp/core.go#L134

**Steps to reproduce**:

Please include a *minimal* but *complete* recreation of the problem,
including server configuration, agent(s) used, etc.  The easier you make it
for us to reproduce it, the more likely that somebody will take the time to
look at it.

 1. start apm-server in a container or systemd environment with the following options:
```
logging.to_files: true
logging.files:
  path: /var/log/apm-server
  name: apm-server
```

**Provide logs (if relevant)**:

Issue #10467, opened 2023-03-09
test: apm server managed by EA should not require Fleet anymore


simitt:
In `8.7`, the sourcemap and agent config logic were reworked to not depend on Fleet anymore. After this, there shouldn't be any technical requirements left why Fleet would be a required component when managed by Elastic Agent. Before we can officially support standalone Elastic Agent, we need to do some manual testing.
Setting up a [standalone Elastic Agent](https://www.elastic.co/guide/en/fleet/8.6/install-standalone-elastic-agent.html) and configure the apm server as integration, and test that
- [x] configuration and confg changes are picked up correctly by the apm-server
- [x] core features are working (data ingestion)
- [x] sourcemap works
- [x] apm agent configuration is working
- [ ] test whether output to logstash is working (not a hard requirement, but good to be aware if there are issues)
- [ ] test whether output to kafka is working (not a hard requirement, but good to be aware if there are issues)

If time permits, also add an automated smoke test for standalone Elastic agent with apm.

simitt:
@kruskall afair from our sync, this is done - could you tick the boxes and close the issue, except if there is anything else left to do.

kruskall:
Ok, I can confirm everything is working and we now have a systemtest for EA managed APM Server.

I haven't tested the kafka/logstash output

kruskall:
> Ok, I can confirm everything is working and we now have a systemtest for EA managed APM Server.

FWIW I thought this was an issue with the agentcfg default credentials but there seems to be some strange issue with the go agent.
To expand on this: I used the go agent and the capture headers option to quickly send requests with curl to a small go program to make sure remote agent config was being applied.

After some investigation, debug logging and changing agent option I confirmed the feature was working but here's the debug log from the APM Go agent:

```
{"level":"debug","message":"central config update: updated capture_body to all"}
{"level":"warning","message":"central config failure: unsupported config: capture_headers"}
{"level":"debug","message":"central config update: updated log_level to debug"}
```

The capture headers doesn't seem to be recognized so the settings in the APM tab has no effect. I don't think this is an issue with the Elastic agent managed APM Server but I don't think this is intended.

simitt:
@kruskall can you bring the go-agent observations to some other format (slack, dedicated issue, etc)?

Thanks for the update, closing this issue then.

kruskall:
> @kruskall can you bring the go-agent observations to some other format (slack, dedicated issue, etc)?

[...]

PR #10636, opened 2023-04-11
test: retrieve apm-server log during smoke tests and move config to template

kruskall:
## Motivation/summary

Improve the current smoke test for supported os:
- switch to terraform `templatefile` and move the template to a separate file
- retrieve apm-server logs with a destroy time provisioner

Log retrieval is currently broken because of https://github.com/elastic/apm-server/issues/10619

This PR is the result of backporting the work on https://github.com/elastic/apm-server/issues/10467 to the current supported-os smoke test

## Checklist

## How to test these changes

1. `EC_API_KEY=<api-key> TEST_DIR=testing/smoke/supported-os/ make smoketest/run`
1. smoke tests are passing


testing/smoke/supported-os/main.tf
```diff
@@ -164,6 +175,19 @@ resource "aws_instance" "apm" {
...
 
+resource "null_resource" "getapmserverlogs" {
+  triggers = {
+    user        = local.image_ssh_users[var.aws_os]
+    host        = aws_instance.apm.public_ip
+    private_key = aws_key_pair.provisioner_key.key_name
+  }
+
+  provisioner "local-exec" {
+    when    = destroy
```
marclop, line 186: Nice!

marclop (commented):
Looks good, should we also update the CI pipeline to store that log file as an artifact?

kruskall:
> should we also update the CI pipeline to store that log file as an artifact?

That's a great idea!

smoke tests are being moved to github actions so I think we should do it in https://github.com/elastic/apm-server/pull/10654. WDYT ?

testing/smoke/supported-os/main.tf
```diff
@@ -164,6 +175,19 @@ resource "aws_instance" "apm" {
   vpc_security_group_ids = [aws_security_group.main.id]
 }
 
+resource "null_resource" "getapmserverlogs" {
```
marclop, line 178: I haven't tested this, will it always run before the `aws_instance` is destroyed?

marclop:
```suggestion
resource "null_resource" "apm_server_log" {
```

marclop (approved):
@kruskall could you create an issue about doing that or make sure that gets done, so we don't forget about it?

PR #10654, opened 2023-04-13
ci: migrate smoke tests to github actions

amannocci:
## Description

* Remove Jenkins Smoke Tests ESS pipeline.
* Remove Jenkins Smoke Tests OS pipeline.
* Migrate Smoke Tests OS to GitHub Actions.
* Migrate Smoke Tests ESS to GitHub Actions.

## Motivation & Context

* Jenkins will be decommissioned and we need to move away from it.


dmathieu:
Do you have a link to a run of that new action?

.github/workflows/smoke-tests.yml
```diff
@@ -0,0 +1,90 @@
...
+          vault-url: ${{ secrets.VAULT_ADDR }}
+          vault-role-id: ${{ secrets.VAULT_ROLE_ID }}
+          vault-secret-id: ${{ secrets.VAULT_SECRET_ID }}
+      - name: Run smoke tests ${{ matrix.test }} for ${{ matrix.version }}
+        run: make smoketest/run-version TEST_DIR=${{ matrix.test }} SMOKETEST_VERSION=${{ matrix.version }}
+      - if: always()
+        name: Teardown smoke test infra
+        run: make smoketest/cleanup TEST_DIR=${{ matrix.test }}
+
+  smoke-tests-os:
```
simitt, line 68: These tests are currently run for `'main', '8.<minor>', '8.<next-patch>'` (see 
[def](https://github.com/elastic/apm-server/pull/10654/files#diff-74ea6b734b6243d4e4e4dddde4234041d59d005d03fbc87f7265d32065a73df2L24)); where is this defined in the github worfklow?

simitt (changes requested):
Thanks for making these changes. Haven't done a completely deep review, but left some comments already. 

The [Smoke test status in the README](https://raw.githubusercontent.com/elastic/apm-server/main/README.md) and the [dev-docs section on smoke tests](https://github.com/elastic/apm-server/blob/main/dev_docs/TESTING.md#smoke-testing) also need to get updated to point to the changed test runs.

.github/workflows/smoke-tests.yml
amannocci: I checked the current [behavior](https://apm-ci.elastic.co/job/apm-server/job/smoke-tests-os-mbp/ ) and currently only the `main` branch is scheduled.
It isn't scheduling `8.x` due to the missing `.ci/smoke-tests-os.groovy` Jenkins script.
Does the list above explicit? Do we need to run `Smoke Tests OS` on `main`, `8.7` and `8.7.1` branches?

simitt: oh interesting; yes it would be good to run the smoke tests on these branches; 
`'8.<minor>', '8.<next-patch>'` make more sense when we are in FF phase, e.g. once we branch out `8.8` but before releasing it we would run the smoke tests on `main` (`8.9.0`), `8.8.0` (unreleased), `8.7.1` (next patch version to be released).

.ci/scripts/smoke-tests-matrix.sh
reakaleek, line 1: The output of this script looks similar to the output of the [elastic-stack-snapshot-branches](https://github.com/elastic/apm-pipeline-library/blob/main/.github/actions/elastic-stack-snapshot-branches/README.md) action.

I understand that the output of this script is slightly different.

Nevertheless, WDYT? Can we maybe add this behaviour to the action with an extra attribute? Would that make sense?

.github/workflows/generate-smoke-tests-list/action.yml
```diff
@@ -0,0 +1,38 @@
+---
+
+name: generate-smoke-tests-list
+description: Generate smoke tests list
+
+inputs:
+  vault-url:
```
reakaleek, line 7: I know kebab-case is actually the GH convention for naming inputs. But in the apm-pipeline-library actions, we mainly use camelCase. 
Would it make sense to align this?

reakaleek (approved):
LGTM. Just some minor notes.

.ci/scripts/smoke-tests-matrix.sh
amannocci: It makes sense to use it for the normal phase. In that case, we need an extra argument to exclude some branches.
However, I'm not sure it will work as expected for the FF phase since we have a new branch.
This new branch will only become active after the release.

v1v: > The output of this script looks similar to the output of the [elastic-stack-snapshot-branches](https://github.com/elastic/apm-pipeline-library/blob/main/.github/actions/elastic-stack-snapshot-branches/README.md) action.

It reads [branches.json](https://storage.googleapis.com/artifacts-api/snapshots/branches.json) and that file is generated automatically as part of the `source of truth` automation done in [branches.json](https://github.com/elastic/apm-pipeline-library/blob/main/.ci/generate-snapshots.sh) (definitely to be improved in the future but at the moment the contract is the files generated in the [artifacts-api](https://storage.googleapis.com/artifacts-api) Google bucket)

The [getBranchesFromAliases](https://github.com/elastic/apm-pipeline-library/blob/main/vars/getBranchesFromAliases.groovy) function reads the file [releases.properties](https://github.com/elastic/apm-pipeline-library/blob/main/resources/versions/releases.properties) and does some manipulation.

I think `elastic-stack-snapshot-branches` and `getBranchesFromAliases` are doing the same for different CIs.

On the other hand, when a FF phase, then `elastic-stack-snapshot-branches` will list all the minor active branches, so if a new branch is created then `main`, `8.X`, `8.X-1` and `7.17`

if FF phase is regarding a new path then `elastic-tack-snapshot-branches` will list `main`, `8.X`, `7.17`

where `X` represents the minor number and `X-1` represents the previous minor number, i.e: `8.8` and `8.7`

Will it be possible to use `elastic-stack-snapshot-branches` in favour of reusing the existing implementation as it used by the CI itself?

.github/workflows/smoke-tests.yml
amannocci: I replace the script with a GitHub Actions doing the same.

.github/workflows/generate-smoke-tests-list/action.yml
v1v: I just raised an internal GH issue so we can agree, as a team, what to do with these conventions

.github/workflows/setup-cluster-env/action.yml
```diff
@@ -0,0 +1,42 @@
...
+      with:
+        url: ${{ inputs.vault-url }}
+        roleId: ${{ inputs.vault-role-id }}
+        secretId: ${{ inputs.vault-secret-id }}
+        method: approle
+        secrets: |
+          ${{ env.EC_KEY_SECRET }} apiKey | EC_API_KEY ;
+          ${{ env.AWS_ACCOUNT_SECRET }} user | AWS_PROFILE ;
+          ${{ env.AWS_ACCOUNT_SECRET }} csv | AWS_ACCOUNT_IMPORT
+    - run: |
```
v1v, line 37: Nit: step name so it can help with knowing what it does when running a multiline script

Makefile
```diff
@@ -332,7 +332,7 @@ testing/rally/corpora:
 SMOKETEST_VERSIONS ?= latest
 # supported-os tests are exclude and hence they are not running as part of this process
 # since they are required to run against different versions in a different CI pipeline.
-SMOKETEST_DIRS = $$(find $(CURRENT_DIR)/testing/smoke -mindepth 1 -maxdepth 1 -type d | grep -v supported-os)
+SMOKETEST_DIRS = $$(find $(CURRENT_DIR)/testing/smoke -mindepth 1 -maxdepth 1 -type d | grep -v supported-os | jq -cnR '[inputs | select(length > 0)]')
```
v1v, line 335: What's the reason for using the pipe with the jq?

README.md
```diff
@@ -1,5 +1,5 @@
 [![Build Status](https://apm-ci.elastic.co/buildStatus/icon?job=apm-server/apm-server-mbp/main)](https://apm-ci.elastic.co/job/apm-server/job/apm-server-mbp/view/change-requests/job/main/)
-[![Smoke Tests on ESS](https://apm-ci.elastic.co/buildStatus/icon?job=apm-server/smoke-tests-ess-mbp/main&subject=smoke%20tests)](https://apm-ci.elastic.co/job/apm-server/job/smoke-tests-ess-mbp/job/main/)
+[![Smoke Tests](https://github.com/elastic/apm-server/actions/workflows/smoke-tests-schedule.yml/badge.svg)](https://github.com/elastic/apm-server/actions/workflows/smoke-tests-schedule.yml)
```
v1v, line 2: Maybe a build badge per smoke test type?

dev_docs/TESTING.md
```diff
@@ -145,8 +145,8 @@ be configured to use an ECE installation) with some light bash scripting which i
 and upgrades, but ensures there aren't any major problems with APM Server accepting and indexing events where
 it should.
 
-The smoke tests can be found under [`testing/smoke`](./../testing/smoke) and the latest CI runs can be found in
-the [APM CI dashboard](https://ela.st/apm-server-smoke-tests).
+The smoke tests can be found under [`testing/smoke`](./../testing/smoke) and the latest CI runs can be found on 
+[Github Actions](https://github.com/elastic/apm-server/actions/workflows/smoke-tests-schedule.yml).
```
v1v, line 149: Ditto, maybe one per smoke test types?

v1v (commented):
Overall LGTM, just a few minor comments

Makefile
amannocci: I allow us to return a JSON array instead of a list of paths separated by `\n`.

README.md
amannocci: Since we are using `workflow_call`, they aren't defined as separate workflows. Hence badges per smoke test type aren't defined.
Would you like to generate a badge per smoke test type? If so, we can add a `workflow_dispatch` trigger on both.

.github/workflows/setup-cluster-env/action.yml
```diff
@@ -0,0 +1,43 @@
...
+    - uses: hashicorp/vault-action@v2.5.0
+      with:
+        url: ${{ inputs.vault-url }}
+        roleId: ${{ inputs.vault-role-id }}
+        secretId: ${{ inputs.vault-secret-id }}
+        method: approle
+        secrets: |
+          ${{ env.EC_KEY_SECRET }} apiKey | EC_API_KEY ;
+          ${{ env.AWS_ACCOUNT_SECRET }} user | AWS_PROFILE ;
+          ${{ env.AWS_ACCOUNT_SECRET }} csv | AWS_ACCOUNT_IMPORT
```
v1v, line 36: These env variables are defined in the consumers of this reusable action, since they are the same, will it make sense to set those env variables within the context of this step?

Makefile
v1v: This breaks the existing implementation in the Makefile:
https://github.com/elastic/apm-server/blob/ae7acf930a2ef7b6074ce1de847025a0fec76c0d/Makefile#L360-L367

if `smoketest/discover` is the one used in this particular changeset, and no other consumers maybe the `jq` pipe can be added to the `smoketest/discover` make goal instead?

v1v (changes requested):

PR #10661, opened 2023-04-18
apmsoak: update configs with new flags

kyungeunni:
Update apmsoak configs to use recently introduced flags


lahsivjar (approved):
LGTM. Will also need to update tf docs and format files by running `make tf-docs`

PR #10668, opened 2023-04-19
Fix systemd.tpl file to use all arguments

lahsivjar:
PR https://github.com/elastic/apm-server/pull/10661 updated the apmsoak to use rewrite flags, however, there was a typo due to which systemd was ignoring the new arguments.


marclop (approved):

kruskall (approved):

kyungeunni:
Yikes, thanks for fixing this @lahsivjar

PR #10675, opened 2023-04-19
build: bump beats library

kruskall:
## Motivation/summary

bump beats library and update code to fix compile issues

Most of the changes are implementing new methods to satisfy the interface requirement.

The only invasive change is in the libbeat publisher test: the pipeline creation code changed and I've updated it to use the new arguments. 
The benchmark hangs so I've disabled it. I don't know if it's still relevant given that libbeat processors are not supported.

## Checklist

## Related issues

Supersedes https://github.com/elastic/apm-server/pull/10645
Supersedes https://github.com/elastic/apm-server/pull/10679


internal/publish/pub_test.go
```diff
@@ -106,6 +104,8 @@ func TestPublisherStopShutdownInactive(t *testing.T) {
 }
 
 func BenchmarkPublisher(b *testing.B) {
```
axw, line 106: Do you know why the benchmark is hanging? I think we need to know why before disabling it, in case there's some real issue being masked.

internal/publish/acker.go
```diff
@@ -70,6 +70,8 @@ func (w *WaitPublishedAcker) Close() {
 	w.decref(1)
 }
 
+func (w *WaitPublishedAcker) ClientClosed() {}
```
axw, line 73: Should this be decrementing the reference count? Maybe this is why the benchmark is hanging...?

internal/beater/beater.go
```diff
@@ -924,12 +925,23 @@ func queryClusterUUID(ctx context.Context, esClient *elasticsearch.Client) error
...
+func (p *nopProcessingSupporter) Processors() []string {
+	procList := []string{}
+	for _, proc := range p.processors {
+		procList = append(procList, proc.String())
+	}
+	return procList
+}
+
+func (p *nopProcessingSupporter) Create(cfg beat.ProcessingConfig, _ bool) (beat.Processor, error) {
+	p.processors = append(p.processors, cfg.Processor)
```
axw, line 945: AFAICS the libbeat default implementation of Create does not store `cfg.Processor` on the supporter. I don't think we should either - that might be a source of concurrency issues. Can we remove this append, and have `Processors()` just return an empty slice? AFAICS `Processors()` isn't relevant to apm-server anyway; it's only used by `cmd/instance/beat.go` in elastic/beats, and we don't use that code.

kruskall: Thank you!, good point! :+1:

internal/publish/acker.go
kruskall: Good catch! That's the issue!

internal/beater/beater.go
```diff
@@ -828,6 +828,7 @@ func (s *Runner) newLibbeatFinalBatchProcessor(
 			// No output defined, so the acker will never be closed.
 			return nil
 		}
+		defer acker.ClientClosed()
```
axw, line 831: Should we be calling this directly? Looking at libbeat code, it appears to be called by when the pipeline.Client is closed.

internal/beater/beater.go
```diff
@@ -807,7 +807,7 @@ func (s *Runner) newLibbeatFinalBatchProcessor(
 		output, err := outputs.Load(indexSupporter, beatInfo, stats, outputName, s.outputConfig.Config())
 		return outputName, output, err
 	}
-	pipeline, err := pipeline.Load(beatInfo, monitors, pipeline.Config{}, nopProcessingSupporter{}, outputFactory)
+	pipeline, err := pipeline.Load(beatInfo, monitors, pipeline.Config{}, &nopProcessingSupporter{}, outputFactory)
```
axw, line 810:
```suggestion
	pipeline, err := pipeline.Load(beatInfo, monitors, pipeline.Config{}, nopProcessingSupporter{}, outputFactory)
```

Can we keep the diff minimal, and revert to a non-pointer type?

axw (commented):
Code looks reasonable apart from maybe calling ClientClosed directly. There's a failing test on macOS which looks suspiciously related, so I guess there's a timing issue.

internal/beater/beater.go
kruskall: Good catch! Fixed!

internal/beater/beater.go
kruskall: Updated!

axw:
Still failing. Maybe try running the tests in a loop with `-race` to reproduce locally.

axw:
Seems the behaviour of libbeat has changed when no libbeat output is defined. I've got another PR in the works that should fix this.

axw (approved):

axw:
@Mergify backport 8.8

PR #10680, opened 2023-04-21
Make machine type configurable for soaktest workers

lahsivjar:
Soak test cluster has been upgraded to use the latest version of `apmsoak`. Due to the changes in `apmsoak` to rewrite ids, timestamps, etc the resource requirement for the worker has increased. This PR allows the machine type to be used as a variable while configuring the soak test workers.


kruskall (approved):
LGTM!

PR #10694, opened 2023-04-24
apmsoak: Allow rewriting global labels

carsonip:
## Motivation/summary

Add an option to apmsoak such that global labels can be replaced with a constant JSON string specified in the `-rewrite-global-labels` argument. It will be directly placed into `metadata.labels`. This will allow us to simulate events with a large number of labels. 

However, this is not designed to simulate different labels between batches as there complexity around randomizing global labels, e.g. number of labels, length of key, length of value, string vs numeric values. It functions a little different from the existing `rewrite*` options as they are flags and have randomization while this accepts a constant string argument.

Example usage:
```
apmsoak -rewrite-global-labels '{"foo":"bar","num":1.1,"bar":"baz"}'
```

## Checklist

### DIFF
--- a/systemtest/loadgen/eventhandler/handler.go
+++ b/systemtest/loadgen/eventhandler/handler.go
@@ -155,6 +155,9 @@ type Config struct {
 	// RewriteTransactionTypes controls the rewriting of `transaction.type`
 	// in events.
 	RewriteTransactionTypes bool
+
+	// ReplaceGlobalLabels replaces metadata.labels with the specified string.
+	ReplaceGlobalLabels string
 }
 
 // New creates a new Handler with config.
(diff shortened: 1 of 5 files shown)


systemtest/loadgen/eventhandler/handler.go
```diff
@@ -155,6 +155,9 @@ type Config struct {
 	// RewriteTransactionTypes controls the rewriting of `transaction.type`
 	// in events.
 	RewriteTransactionTypes bool
+
+	// ReplaceGlobalLabels replaces metadata.labels with the specified string.
+	ReplaceGlobalLabels string
```
lahsivjar, line 160: Can we use `RewriteGlobalLabels`? I know that it will be slightly different than other flags in the way that it will rewrite with the provided input but it still performs a similar functionality.

carsonip: Renamed

axw:
@carsonip when will we use this flag? Would it exercise some overflow that isn't otherwise exercised by another flag?

carsonip:
> when will we use this flag?

For stress testing.

> Would it exercise some overflow that isn't otherwise exercised by another flag?

It would not. It sets the global labels for all the metadata emitted to the same provided labels. This is to measure the overhead of processing labels (especially when there is a long list of labels).

axw:
> Would it exercise some overflow that isn't otherwise exercised by another flag?
>
> It would not. It sets the global labels for all the metadata emitted to the same provided labels. This is to measure the overhead of processing labels (especially when there is a long list of labels).

OK. Seems like it would be better tested with a microbenchmark, with varying label sizes/amounts. Is there something to be gained by doing it as a macrobenchmark? The reason we have the other flags is more to trigger grouping overflow.

carsonip:
> OK. Seems like it would be better tested with a microbenchmark, with varying label sizes/amounts. 

Agreed.

> Is there something to be gained by doing it as a macrobenchmark? The reason we have the other flags is more to trigger grouping overflow.

Nothing at this point other than facilitating e2e stress testing. This is also why I was a little hesitant with this change at first.

carsonip:
Closing this PR as this may not be useful outside the specific e2e stress testing context.

PR #10697, opened 2023-04-25
Add support for mobile UI filtering

bryce-b:
## Motivation/summary
The mobile ui wants to filter data based on host.os.version & service.version. Http spans are a key part of the mobile ui, and when filters are added for either of these attributes all http related charts will be empty.

## Checklist

## Related issues

https://github.com/elastic/apm-agent-ios/issues/145

### DIFF
--- a/changelogs/head.asciidoc
+++ b/changelogs/head.asciidoc
@@ -17,4 +17,4 @@ https://github.com/elastic/apm-server/compare/8.8\...main[View commits]
 
 [float]
 ==== Added
-- Span events now carry `host` and `service.version` attributes.
+- Span events now carry `host` and `service.version` attributes. {pull}10697[10697]
--- a/changelogs/head.asciidoc
+++ b/changelogs/head.asciidoc
@@ -18,3 +18,4 @@ https://github.com/elastic/apm-server/compare/8.7\...main[View commits]
 
 [float]
 ==== Added
+- Span events now carry `host` and `service.version` attributes.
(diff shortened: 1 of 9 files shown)


axw (commented):
@felixbarny do you think this is worth benchmarking, or would the design win anyway?

I wonder if we should drop that `remove` processor altogether. This has been around since forever, when we used to store a transaction document even for non-sampled transactions, and when we didn't have a default ILM delete phase . I feel like we have better options for storage management these days.

felixbarny:
If we already had synthetic _source, it would probably be a no-brainer to drop the `remove` processor altogether. Instinctively, I don't expect this to have a huge impact even now but I'd prefer to do some testing for that. If that testing would be very difficult to do (which I hope it isn't, that would kinda be another issue in itself) I suppose we could skip it if we preserve just `host` and `service.name`.

axw (approved):

changelogs/head.asciidoc
```diff
@@ -17,3 +17,4 @@ https://github.com/elastic/apm-server/compare/8.8\...main[View commits]
 
 [float]
 ==== Added
+- Span events now carry `host` and `service.version` attributes.
```
kruskall, line 20:
```suggestion
- Span events now carry `host` and `service.version` attributes. {pull}10697[10697]
```

kruskall (changes requested):

axw:
Verified with BC2. Send testdata/intake-v2/spans.ndjson to the hosted APM Server, confirmed that `host.*` and `service.version` fields are present in span documents.

PR #10708, opened 2023-04-27
Bump up version for 8.7.2

lahsivjar:
To be merged after tag for 8.7.1 is cut

### DIFF
--- a/apmpackage/apm/changelog.yml
+++ b/apmpackage/apm/changelog.yml
@@ -1,5 +1,10 @@
 - version: "generated"
   changes:
+    - description: Placeholder
+      type: enhancement
+      link: https://github.com/elastic/apm-server/pull/123
+- version: "8.7.1"
+  changes:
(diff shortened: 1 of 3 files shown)


simitt (approved):

carsonip (commented):
What about updating changelog in `apmpackage/apm/changelog.yml`? I assume do should do it here instead of #10707 right?

lahsivjar:
> What about updating changelog in apmpackage/apm/changelog.yml? 

I was expecting our automation to create this PR (similar to https://github.com/elastic/apm-server/pull/9529/files) but it seems like that automation is broken and doesn't even seem to be triggering anymore.

I have updated the PR to add changelog for apmpackage. Thanks for noticing this.

simitt:
@lahsivjar I believe automation might only kick in after the `8.7.1` release? But would be good to follow up with the prod.eng team.

PR #10717, opened 2023-04-28
Fix incorrect labels definition for systemtest

lahsivjar:
`labels` should be inside `metadata` rather than at the same level as `metadata`:

```
{
    "metadata": {
        "service": {...},
        "labels": {...},
        ...
    }
}
```


carsonip (approved):

kruskall (approved):

lahsivjar:
[**For reviewers**]: We were earlier using the backend endpoint for the rum events, I have updated it to use rum endpoint and updated the ES query to search for global labels for the `traces` making the test fail with timeout without this fix.

axw (approved):

PR #10637, opened 2023-04-11
test: add smoke testing for EA managed apm-server

kruskall:
## Motivation/summary

Add smoke test for elastic-agent managed APM server.

This is similar to the current test for standalone apm-server binary but we are using elastic agent and passing an elastic agent config.
Terraform `templatefile` is used and the apm-server logs are retrieved before the deployment is destroyed.

Make sure core features, config and data ingestion is working.

## Checklist

~- [ ] Update [CHANGELOG.asciidoc](https://github.com/elastic/apm-server/blob/main/CHANGELOG.asciidoc)~
~- [ ] Update [package changelog.yml](https://github.com/elastic/apm-server/blob/main/apmpackage/apm/changelog.yml) (only if changes to `apmpackage` have been made)~
~- [ ] Documentation has been updated~

For functional changes, consider:
- Is it observable through the addition of either **logging** or **metrics**?
- Is its use being published in **telemetry** to enable product improvement?
- Have system tests been added to avoid regression?

## How to test these changes

`EC_API_KEY=<api-key> TEST_DIR=testing/smoke/managed/ make smoketest/run`

## Related issues

Related to https://github.com/elastic/apm-server/issues/10467

PR #10719, opened 2023-04-28
Terraform fmt

dmathieu:
## Motivation/summary

Following this comment: https://github.com/elastic/apm-server/pull/10637#discussion_r1180021907, this adds a GH Action check which runs `terraform fmt` on every terraform file, if any changed. And expects the diff to be clean (if any file is changed by the command, it means they have formatting issues).

## Checklist

This is a CI change, there are no functional changes.

## How to test these changes

The CI run should pass.

## Related issues

N/A

### DIFF
--- a/.github/workflows/terraform-fmt.yml
+++ b/.github/workflows/terraform-fmt.yml
@@ -14,4 +14,4 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - uses: hashicorp/setup-terraform@v2
-      - run: find . -name '*.tf' -exec terraform fmt {} \;; git diff --exit-code
+      - run: terraform fmt -recursive -check -diff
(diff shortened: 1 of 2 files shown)


.github/workflows/terraform-fmt.yml
```diff
@@ -0,0 +1,17 @@
...
+    paths:
+      - '**.tf'
+
+jobs:
+  terraform-fmt:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: hashicorp/setup-terraform@v2
+      - run: find . -name '*.tf' -exec terraform fmt {} \;; git diff --exit-code
```
kruskall, line 17: **suggstion**: I think we can avoid git diff by using `terraform fmt -check -diff`. We can probably avoid `find` too by using the `-recursive` flag

kruskall (changes requested):

.github/workflows/terraform-fmt.yml
dmathieu: Ah, good find.

kruskall (approved):

PR #10730, opened 2023-05-01
added app_events data stream for otel events

bryce-b:
## Motivation/summary
The mobile agents & mobile UI are starting to depend on `events`, a special class of log in the Otel environment.  We think it would be valuable to have them ingested in their own data stream to differentiate them from logs, and avoid having them lost in the noise of the logs data stream. 

## Checklist

## Related issues
https://github.com/elastic/apm-data/pull/28


bryce-b:
I'm not entirely sure this is all the points that need to get added for another data-stream. See also the PR to elastic/apm-data#28

axw:
The code change looks fine. What is the definition of an "app event"? How does this differ from any other kind of log? What are they used for, and why does it help to store them a separate data stream?

bryce-b:
@axw an event is a special log that contains `event.name` and `event.domain`. In open telemetry, a log provider cannot create events unless the provider is created with the `event.domain` set. I used the `app` prefix because I thought `event` on it's own was too broad; maybe more thought should go into the name? 

From the [Otel event-api spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/event-api.md) : 
> While the logging space has a diverse legacy with many existing logging libraries in different languages, there is not ubiquitous alignment with OpenTelemetry events. In some logging libraries, producing records shaped as OpenTelemetry events is clunky or error-prone.

The reason we want to put them in a separate data stream is because they represent more critical data points for developers, such as crashes, or [lifecycle events
](https://github.com/elastic/apm/blob/main/specs/agents/mobile/events.md#application-lifecycle-events)

bryce-b:
@kruskall could you point me towards the tests for `apmpackage/apm/data_stream`? I'm having difficulty finding them

kruskall:
> @kruskall could you point me towards the tests for `apmpackage/apm/data_stream`? I'm having difficulty finding them

Sorry, wrong PR, I meant in apm-data: https://github.com/elastic/apm-data/pull/28

We could probably have some system tests here too but there is a more important discussion going on so I'll defer to other reviewers

AlexanderWert:
> * Do we want to store one document per received event or do we want to pre-aggregate an not store them?

One use case is to see what happened within individual sessions. So a timeline of events. For that we need to store the individual events (not aggregated).

> * Do we want users to see their events when looking at the logs of the application?

I would say: NO. (And that was the primary reason to think about a separate data-stream.) Reasoning: we do not want to "pollute" the logs view with events that have a very specific meaning. 

> * Where and how are we planning to visualize events?

For example in a session timeline. 

> * What are the downsides of querying the logs data stream and adding an exists query on `event.name` and `event.domain`?

The main reason was to logically separate different types of data (also semantically and logically). And when showing regular logs of a service / app to not include the special events (like breadcrumbs, lifecycle events, etc.).

felixbarny:
> we do not want to "pollute" the logs view with events that have a very specific meaning.

How do we plan to filter out the app events from the service logs, then? The events will also match the `logs-*` index pattern and have a `service.name` entry.

felixbarny:
> Is that due to the dataset being set to logs?

Yes.

> Would it work to change the dataset to something like events? I'm not sure what the constraints of that field are.

That would work but there are constraints, yes. However, these constrains aren't set in stone:

https://github.com/elastic/package-spec/blob/13493cded215333659c3c6070956182c5acc6370/spec/integration/data_stream/manifest.spec.yml#L203-L211

bryce-b:
We chatted about this PR in the mobile agent weekly meeting, and decided it's not necessary to split events out of logs for mobile.

Issue #10718, opened 2023-04-28
Document TBS config options for standalone apm server


simitt:
Following up on https://github.com/elastic/apm-server/issues/10293, the config options available via Fleet configuration should also be added to the `apm-server.yml` and be publicly documented for the standalone apm-server.

PR #10740, opened 2023-05-02
docs: document TBS config options for standalone apm server

kruskall:
## Motivation/summary

Update apm-server config files with new tbs settings and list them in the config option page.

## Checklist

## Related issues

Closes https://github.com/elastic/apm-server/issues/10718

### DIFF
--- a/apm-server.docker.yml
+++ b/apm-server.docker.yml
@@ -279,6 +279,18 @@ apm-server:
     # Configure curve types for ECDHE based cipher suites.
     #ssl.curve_types: []
 
+  #---------------------------- APM Server - tail-based sampling ----------------------------
+
+  #sampling.tail:
+    #Set to `true` to enable tail based sampling. Disabled by default.
+    #enabled: false
+
+    #Synchronization interval for multiple APM Servers. Should be in the order of tens of seconds or low minutes.
+    #interval: 1m
+
+    #Criteria used to match a root transaction to a sample rate.
+    #policies: []
+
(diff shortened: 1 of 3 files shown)


apm-server.docker.yml
```diff
@@ -279,6 +279,18 @@ apm-server:
...
+
+  #sampling.tail:
+    #Set to `true` to enable tail based sampling. Disabled by default.
+    #enabled: false
+
+    #Synchronization interval for multiple APM Servers. Should be in the order of tens of seconds or low minutes.
+    #interval: 1m
+
+    #Criteria used to match a root transaction to a sample rate.
+    #policies: []
```
marclop, line 292: Maybe we can add a commented policy? So users have an easier time writing these?

apm-server.docker.yml
```diff
@@ -279,6 +279,18 @@ apm-server:
     # Configure curve types for ECDHE based cipher suites.
     #ssl.curve_types: []
 
+  #---------------------------- APM Server - tail-based sampling ----------------------------
+
+  #sampling.tail:
+    #Set to `true` to enable tail based sampling. Disabled by default.
```
dmathieu, line 285: Nit: other doc comments in the file have a space right after the `#`.

apm-server.docker.yml
kruskall: I think the example is currently reflecting the default values. 

I think examples should go in the doc which already has an example configuration for TBS. WDYT ?

marclop: Sure, let's leave it like that then

PR #10744, opened 2023-05-03
R8 map fetching function

LikeTheSalad:
## Motivation/summary

We need to be able to fetch Android map files, similar to what we do with JS source maps right now. There's a `Fetcher` interface that is used for source map file fetching for which some of its implementations might be reusable for fetching Android map files too. I'd like to be able to adapt this functionality for Android R8 map files in this PR.

I'll create it as a draft first because I have several questions on what would be the best approach for adapting this process to Android which I'll be adding to each file below.

## Checklist


internal/sourcemap/body_caching.go
LikeTheSalad, line 1: It seems like we can store the sourcemap `Consumer` object in the cache, though I'm wondering if the same should be done for Android's map file? Since it's considerably bigger than a source map one. For our sample project, this map file size is > 7mb, so it could easily get bigger for bigger projects. What would be the best way to "cache" this file?

internal/sourcemap/fetcher.go
```diff
@@ -38,6 +39,8 @@ type Fetcher interface {
 	//
 	// If there is no such source map available, Fetch returns a nil Consumer.
 	Fetch(ctx context.Context, name string, version string, bundleFilepath string) (*sourcemap.Consumer, error)
+
+	FetchAndroidMap(ctx context.Context, name string, versionName string, versionCode string) (*io.ReadCloser, error)
```
LikeTheSalad, line 43: I'm not sure if adding a new function to this interface is the best approach.

internal/sourcemap/kibana.go
```diff
@@ -86,6 +86,10 @@ func (s *kibanaFetcher) Fetch(ctx context.Context, name, version, path string) (
 	return nil, nil
 }
 
+func (s *kibanaFetcher) FetchAndroidMap(ctx context.Context, name string, versionName string, versionCode string) (*io.ReadCloser, error) {
+	return nil, nil
```
LikeTheSalad, line 90: It seems like we'd need to do an HTTP GET to kibana's endpoint, I didn't copy/paste what was done for the sourcemap because probably it'd need to be slightly different due to Android's map file sizes... Not sure if that request might have to be done in multiple parts.

LikeTheSalad:
Hi @kruskall @carsonip @axw - This part of the work needed for adding R8 deobfuscation support seems to be one of the trickiest ones due to the different implementations needed to be addressed for the current `sourcemap.Fetcher` interface. I've left a couple of questions above on several files, I've mostly only done the `esFetcher` implementation, although I'm not sure if it would be the ideal approach, I'm also not quite sure if the same way we're caching sourcemaps could be done with Android maps due to the difference in size. I've also left some comments about it. Please have a look when you get a chance, I'm open to any ideas.

internal/sourcemap/kibana.go
LikeTheSalad: I've also tried to upload an android map file using [Kibana's API](https://www.elastic.co/guide/en/apm/guide/8.7/source-map-how-to.html#source-map-rum-upload) and I noticed that it's not possible for two reasons:

- There's a file size limit that is lower than Android's map file size.
- Source maps are JSON files and this Kibana endpoint is aware of their format, so it only allows JSON content that also contains the key/values expected in a sourcemap file.

I'm thinking a new endpoint would be needed to upload Android map files.

kruskall:
Sorry I missed the ping. We briefly discussed it in one of our sync, I agree with @carsonip proposal. We can create a new interface and refactor later. A separate package is also fine, if there's a need to abstract common code we can do it as needed.
Again, sorry for the late response :bow:

LikeTheSalad:
Thank you for your inputs, @carsonip @kruskall - I'm going to close this PR and leave it as a reference for [this new draft](https://github.com/elastic/apm-server/pull/11005) with a new approach, based on what was mentioned here, which would allow us to make this functionality work for now without having to modify the existing interface.

PR #10751, opened 2023-05-04
New APM documentation structure

bmorelli25:
### Summary

This PR sets us up for success with future APM Guide improvements. This is a tough one to review because the diff is so large; future PRs will be much smaller. It does the following:

- [x] Removes the `Legacy APM Overview` ([this](https://www.elastic.co/guide/en/apm/guide/current/legacy-apm-overview.html)).
- [x] Deletes the duplicate `API` section in the APM Server Reference ([this](https://www.elastic.co/guide/en/apm/guide/current/intake-api.html))
- [x] Deletes the duplicate `How to` section in the APM Server Reference ([this](https://www.elastic.co/guide/en/apm/guide/current/howto-guides.html))
- [x] Removes all deprecation notices from the `Legacy APM Overview` and the `Legacy APM Server Reference`.
- [x] Up-levels the remaining pages in the `Legacy APM Server Reference`. This will make it easier to combine the content into one APM Guide. You can see this change in the [TOC in the doc preview](https://apm-server_10751.docs-preview.app.elstc.co/guide/en/apm/guide/master/getting-started-apm-server.html).

### Notes for reviewers

I think the biggest thing to look for in this PR is that I didn't accidentally remove content we want to keep. I'm following [this doc](https://docs.google.com/document/d/1x_eKhOgOZwD4AzPOcTK2BIfpSR85rYupYmGRofmpwOs/edit#) (sorry, internal only) but it'd be good to get a second set of eyes.


changelogs/6.1.asciidoc
simitt, line 1: Why are the changelogs removed?

docs/legacy/copied-from-beats/docs/keystore.asciidoc
```diff
@@ -12,7 +12,7 @@
 [[keystore]]
 === Secrets keystore for secure settings
 
-IMPORTANT: {deprecation-notice-installation}
+IMPORTANT: -installation}
```
simitt, line 15: Should that be completely removed?

simitt (commented):
We talked about this yesterday, but just to be sure - everything that is removed here stays removed (as it is either not needed anymore, or already available in the Fleet managed guide) and everything that is brought to top level will be combined with the Fleet managed guide. This will happen in a dedicated follow up PR to keep the review scope smaller. Is this correct?

changelogs/6.1.asciidoc
bmorelli25: I removed the changelogs because we only publish `8.x` and `7.last` changelogs to the 8.x docs.

If you think there's utility in keeping the source files around still, I'm fine with that.

docs/legacy/copied-from-beats/docs/keystore.asciidoc
bmorelli25: Good catch :)

bmorelli25:
> everything that is removed here stays removed (as it is either not needed anymore, or already available in the Fleet managed guide) and everything that is brought to top level will be combined with the Fleet managed guide. This will happen in a dedicated follow up PR to keep the review scope smaller. Is this correct?

That's correct. The only exception to this is the exported fields documentation, which I've removed in a separate PR to allow for discussion.

changelogs/6.1.asciidoc
simitt: got it; I think it's fine to remove them - we can go back to older branches when looking for the changes.

PR #10772, opened 2023-05-08
docs: remove exported fields

bmorelli25:
### Summary

This PR removes the exported fields documentation.

### For the reviewer

Is this what we want to do? Does it still make sense to document these fields when any changes are made in conjunction with changes to the APM app?


simitt (approved):
IMO there is no value in documenting the exported fields. They are available via the index templates and data stream mappings, and I don't see why this would be relevant upfront and thus need to be in the docs. 

Curious to hear if @axw has any concerns..

PR #10793, opened 2023-05-11
apmpackage: prepare for metrics output schema change

axw:
## Motivation/summary

Metrics produced by apm-server (apm-data)
will soon change. Update the ingest pipelines
to cater for this change. There is no change
to how metrics are indexed; the ingest pipeline
takes care of restructuring things.

## Checklist

## How to test these changes

N/A non-functional change until we update apm-data

## Related issues

### DIFF
--- a/apmpackage/apm/data_stream/internal_metrics/elasticsearch/ingest_pipeline/default.yml
+++ b/apmpackage/apm/data_stream/internal_metrics/elasticsearch/ingest_pipeline/default.yml
@@ -14,9 +14,11 @@ processors:
   - pipeline:
       name: client_geoip
   - remove:
-      field: _metric_descriptions
-      ignore_missing: true
-  - remove:
       # Removed in 8.6.0
       field: timeseries.instance
+      ignore_missing: true
+  - pipeline:
+      name: set_metrics
+  - remove:
+      field: _dynamic_templates
       ignore_missing: true
--- a/apmpackage/apm/data_stream/service_destination_interval_metrics/elasticsearch/ingest_pipeline/default.yml
+++ b/apmpackage/apm/data_stream/service_destination_interval_metrics/elasticsearch/ingest_pipeline/default.yml
@@ -7,6 +7,8 @@ processors:
       name: observer_ids
   - pipeline:
       name: ecs_version
+  - pipeline:
+      name: set_metrics
   - remove:
-      field: _metric_descriptions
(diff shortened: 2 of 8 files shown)


apmpackage/apm/data_stream/internal_metrics/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -13,10 +13,12 @@ processors:
...
-      ignore_missing: true
   - remove:
       # Removed in 8.6.0
       field: timeseries.instance
       ignore_missing: true
+  - pipeline:
+      name: set_metrics
+  - remove:
+      field: _dynamic_templates
+      ignore_missing: true
```
simitt, line 24: Why is the remove pipeline necessary here? Previously, the `set_metrics` script wasn't called here, but now that it is, it removes the `_metric_descriptions` and `samples` anyways. 
I also generally don't follow why `_dynamic_templates` should now be removed?

apmpackage/apm/data_stream/service_destination_interval_metrics/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -7,6 +7,8 @@ processors:
       name: observer_ids
   - pipeline:
       name: ecs_version
+  - pipeline:
+      name: set_metrics
   - remove:
-      field: _metric_descriptions
+      field: _dynamic_templates
```
simitt, line 13: same comment as above

apmpackage/apm/data_stream/internal_metrics/elasticsearch/ingest_pipeline/default.yml
axw: Note that we're removing `_dynamic_templates` here, not `_metric_descriptions. `_dynamic_templates` is related to dynamic mapping -- it's set by the `set_metrics` pipeline to tell Elasticsearch how to dynamically map a field.

We only rely on dynamic mapping for the app_metrics data stream. We were previously relying on metrics being set at the top level of the document, and we removed `_metric_descriptions`, which was sent by APM Server. Now we're sending `metricset.samples` instead, which maps metrics to the top level and sets `_dynamic_templates`. We want the first bit, but not the second.

lahsivjar (approved):

PR #10814, opened 2023-05-16
docs: Update exploring data in {es}

bmorelli25:
### Summary

This PR moves the "Exploring data in Elasticsearch" section to the "Manage storage" bucket. It also condenses some of the content into a single page within expandable boxes.

Screenshot 2023-05-15 at 10 27 17 PM

"Application metrics" being a sub-nested heading was the best solution I could think of. Open to other ideas.


docs/legacy/metricset-indices.asciidoc
```diff
@@ -1,9 +1,5 @@
 [[metricset-indices]]
-== Metrics documents
-
-++++
-<titleabbrev>Metrics documents</titleabbrev>
-++++
+== Application metrics
```
simitt, line 2: IMO we'll need to rework the metrics docs for some reasons:
- the main question that we see raised is which metrics end up in service specific datastreams and which in the generic one
- irritation around the aggregated metrics

We'll need to better cover these topics, but without over-documenting it, as the concrete (aggregated) metrics data streams are considered an internal detail, that we might change, as long as the UI changes the queries in a compatible way. 
Should we have an in-depth conversation for this topic on this PR, or do you prefer a follow up issue to not slow down the structural changes?

bmorelli25: 👍 Makes sense. I'd prefer to move the conversation to a separate issue.

PR #10841, opened 2023-05-17
docs: Secure communication

bmorelli25:
### Summary

//soon


docs/secure-comms.asciidoc
```diff
@@ -0,0 +1,25 @@
...
+include::secure-agent-communication.asciidoc[]
+
+// APM privileges
+include::{docdir}/legacy/feature-roles.asciidoc[]
+
+// APM API keys
+include::{docdir}/legacy/api-keys.asciidoc[]
+
+// Linux Seccomp
+include::{docdir}/legacy/copied-from-beats/docs/security/linux-seccomp.asciidoc[]
```
axw, line 24: seccomp configuration is unrelated to communication with the stack

I think we should just remove the seccomp section altogether. The only time we fork/exec is for the java attacher, and disallowing those syscalls will break that feature. Otherwise if you want to disallow syscalls it's better to apply seccomp rules in the calling environment, e.g. using systemd sandboxing or Docker seccomp config.

axw: @bmorelli25 did you see this comment?

bmorelli25: Whoops. Thanks for the bump on this! Removed.

axw: :+1: 

@simitt please shout if you think we should still document seccomp. You may recall we disabled it by default a while back. I think it's still possible to configure, but as mentioned above I think we should guide users to external configuration for this, if needed at all.

axw (approved):

PR #10881, opened 2023-05-23
Add service.language.name to service destination mapping

carsonip:
## Motivation/summary

Add `service.language.name` to service destination mapping to align with other aggregations. There will be no change to actual aggregation implementation yet.

## Checklist


axw:
This did not make it into 8.9.0-BC1, the bundled package hasn't been updated since May 17: https://github.com/elastic/kibana/pull/160272

axw:
Confirmed that the field is in the component template mappings in BC2.

Issue #10531, opened 2023-03-27
Service target is not considered in service destination aggregation


lahsivjar:
**APM Server version** (`apm-server version`): 8.x

**Description of the problem including expected versus actual behavior**: `service.target.*` fields are not considered for service_destination aggregation. This is because service destination aggregation skips all spans with destination service set as `nil` ([ref](https://github.com/elastic/apm-server/blob/f701332a56aabec12f6454a649b97afa48f3cae0/x-pack/apm-server/aggregation/spanmetrics/aggregator.go#L230-L232)) and we don't infer destination service fields if only service target fields are set ([ref](https://github.com/elastic/apm-data/blob/3ad1a5caedc9bf5b36ebd62cdc8193e7cedc7d05/input/elasticapm/internal/modeldecoder/v2/decoder.go#L1032-L1035)).

**Steps to reproduce**:
1. Send exit spans with `service.target.*` set but missing `destination.service.*` fields.
2. Observe that `service_destination` aggregations are not produced.

**Provide logs (if relevant)**: NA

dmathieu:
I'm having a look at this, and I wonder what the expected behavior should be.
The aggregation [requires a destination](https://github.com/elastic/apm-server/blob/f701332a56aabec12f6454a649b97afa48f3cae0/x-pack/apm-server/aggregation/spanmetrics/aggregator.go#L262), so it seems we should detect one from the target? What should that be?

lahsivjar:
When decoding the model, as per the spec, we do infer target from destination resource but not vice versa ([ref](https://github.com/elastic/apm-data/blob/main/input/elasticapm/internal/modeldecoder/v2/decoder.go)). Full spec available [here](https://github.com/elastic/apm/blob/main/specs/agents/tracing-spans-service-target.md).

> The aggregation [requires a destination](https://github.com/elastic/apm-server/blob/f701332a56aabec12f6454a649b97afa48f3cae0/x-pack/apm-server/aggregation/spanmetrics/aggregator.go#L262), so it seems we should detect one from the target?

IIANM, the aggregation should work without destination resource if `service.target.*` are specified.

PR #10922, opened 2023-06-01
Keep making aggregation keys if there is no destination

dmathieu:
## Motivation/summary

If no service destination has been set, we can still keep creating aggregation keys, with a blank resource.

## Checklist

- [x] **N/A** Update [package changelog.yml](https://github.com/elastic/apm-server/blob/main/apmpackage/apm/changelog.yml) (only if changes to `apmpackage` have been made)
- [x] **Should be N/A** Documentation has been updated

## How to test these changes

This is unit-tested

## Related issues

Closes https://github.com/elastic/apm-server/issues/10531

### DIFF
--- a/x-pack/apm-server/aggregation/spanmetrics/aggregator.go
+++ b/x-pack/apm-server/aggregation/spanmetrics/aggregator.go
@@ -227,9 +227,6 @@ func (a *Aggregator) ProcessBatch(ctx context.Context, b *model.Batch) error {
 }
 
 func (a *Aggregator) processSpan(event *model.APMEvent) {
-	if event.Span.DestinationService == nil || event.Span.DestinationService.Resource == "" {
-		return
-	}
 	if event.Span.RepresentativeCount <= 0 {
 		// RepresentativeCount is zero when the sample rate is unknown.
 		// We cannot calculate accurate span metrics without the sample
@@ -256,11 +253,17 @@ func (a *Aggregator) processSpan(event *model.APMEvent) {
 		count: float64(count) * event.Span.RepresentativeCount,
 		sum:   float64(duration) * event.Span.RepresentativeCount,
 	}
+
+	var resource string
+	if event.Span.DestinationService != nil && event.Span.DestinationService.Resource != "" {
+		resource = event.Span.DestinationService.Resource
+	}
+
 	for _, interval := range a.Intervals {
 		key := makeAggregationKey(
 			event,
 			event.Event.Outcome,
-			event.Span.DestinationService.Resource,
+			resource,
 			serviceTargetType,
 			serviceTargetName,
 			event.Span.Name,
--- a/x-pack/apm-server/aggregation/spanmetrics/aggregator_test.go
+++ b/x-pack/apm-server/aggregation/spanmetrics/aggregator_test.go
@@ -262,7 +262,40 @@ func TestAggregatorRun(t *testing.T) {
 				{serviceName: "service-A", agentName: "java", outcome: "success", count: 1},
 			},
 			getExpectedEvents: func(now time.Time, interval time.Duration) []model.APMEvent {
-				return []model.APMEvent{}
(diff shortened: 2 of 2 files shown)


x-pack/apm-server/aggregation/spanmetrics/aggregator.go
```diff
@@ -227,9 +227,6 @@ func (a *Aggregator) ProcessBatch(ctx context.Context, b *model.Batch) error {
 }
 
 func (a *Aggregator) processSpan(event *model.APMEvent) {
-	if event.Span.DestinationService == nil || event.Span.DestinationService.Resource == "" {
-		return
-	}
```
lahsivjar, line 232:
```go
	if event.Service.Target == nil {
		return
	}
```

I think this should be updated to NOT process if service target is not set, provided there are no other edge cases. IIUC, there will be three cases when we should produce the service destination aggregation:

1. `service.target.*` is not set and `destination_service.resource` is set
2. `service.target.*` is set and `destination_service.resource` is not set
3. both are set

For `1`, when decoding we will infer `service.target.*` and set it whereas for `2` and `3` we will not infer anything ([ref](https://github.com/elastic/apm-data/blob/c0b8c4f469fdc878aec0ea0e488cb084baa66605/input/elasticapm/internal/modeldecoder/v2/decoder.go#L1032-L1035)).

x-pack/apm-server/aggregation/spanmetrics/aggregator_test.go
```diff
@@ -262,7 +262,40 @@ func TestAggregatorRun(t *testing.T) {
 				{serviceName: "service-A", agentName: "java", outcome: "success", count: 1},
 			},
 			getExpectedEvents: func(now time.Time, interval time.Duration) []model.APMEvent {
-				return []model.APMEvent{}
```
lahsivjar, line 265: In my understanding, this should be empty.

x-pack/apm-server/aggregation/spanmetrics/aggregator.go
dmathieu: It can. But since today, there can be aggregations emitted if there is no service target (but there is a destination), that seemed like a strange change.

lahsivjar: > But since today, there can be aggregations emitted if there is no service target (but there is a destination), that seemed like a strange change

Is this possible, I think when we decode an event with no service target but destination, we would infer the service target: https://github.com/elastic/apm-data/blob/c0b8c4f469fdc878aec0ea0e488cb084baa66605/input/elasticapm/internal/modeldecoder/v2/decoder.go#L1032-L1035

dmathieu: That makes sense, thanks. I've updated the PR.

lahsivjar (approved):

changelogs/head.asciidoc
```diff
@@ -20,3 +20,4 @@ https://github.com/elastic/apm-server/compare/8.8\...main[View commits]
 ==== Added
 - Span events now carry `host` and `service.version` attributes. {pull}10697[10697]
 - Accept more options in agentcfg ES config; Reliably use `agent.config.elasticsearch` and `rum.source_mapping.elasticsearch` credentials when merging with `output.elasticsearch` credentials {pull}10783[10783]
+- Make aggregation keys even if there is no service destination. {pull}10922[10922]
```
lahsivjar, line 23:
```suggestion
- Emit service destination metrics when only `service.target.*` fields are set. {pull}10922[10922]
```
Or something along similar lines maybe?

axw:
Verified with 8.9.0-BC1.

I modified https://github.com/elastic/apm-server/blob/main/testdata/intake-v2/spans.ndjson to send just the event with "target", but with the "destination" attributes removed. I then confirmed that after a minute the service destination metric was produced.

PR #10968, opened 2023-06-12
Use "dynamic: runtime" instead of strict mapping for internal, TBS, and aggregation indices

### LINKED ISSUE #10808: Revisit decision to use strict mappings
In our well-defined metrics data streams, we currently specify strict mappings. e.g. `metrics-apm.internal-*`:

https://github.com/elastic/apm-server/blob/f61db13498b1ab6af88d5993d4b2c04f6501d65f/apmpackage/apm/data_stream/internal_metrics/manifest.yml#L8-L10

The reason for doing so is that we are (ostensibly) in full control of the mappings, and there should never be any additional fields. We occasionally introduce bugs that invalidate this and lead to users having a bad time, e.g. https://github.com/elastic/apm-server/issues/10773. Other times we find issues with ingest pipelines that unexpectedly and unintentionally introduce fields, e.g. https://github.com/elastic/elasticsearch/issues/87345. The latter kind of issue is what strict mappings are supposed to surface.

Blocking ingestion is an extreme response to unexpected fields. On the other hand we should not allow dynamic mapping, as that may also lead to documents being rejected, e.g. due to mapping conflicts. We should consider setting `dynamic: false` or `dynamic: runtime`.

carsonip:
## Motivation/summary

Use "dynamic: runtime" instead of strict mapping for internal, TBS, and aggregation indices. Avoid blocking ingestion on unexpected fields.

## Checklist

## How to test these changes

- Create an unexpected field (by dropping a field in the mapping / emitting a new field in apm-server) and make sure it doesn't block ingestion.

## Related issues

Closes #10808


axw (approved):

axw:
Verified with BC2. I listed all component templates matching `*apm*@package`, and confirmed that apart from `metrics-apm.app@package` which is expected to have `dynamic: true`, all others use either `dynamic: false` or `dynamic: runtime`

PR #11005, opened 2023-06-15
Adding R8 deobfuscation to otlp batch processors

LikeTheSalad:
## Motivation/summary

This change will allow doing an Android exception stacktrace deobfuscation when receiving an OTLP mobile exception log. In order to do so, part of this PR is related to fetching an Android deobfuscation map file from Elasticsearch, which is currently done in a similar way to how it's done for JS RUM Source maps. In [the first draft](https://github.com/elastic/apm-server/pull/10744), the idea was to reuse the same `Fetcher` interface as the one used for JS Source maps, however, it was [later mentioned](https://github.com/elastic/apm-server/pull/10744#issuecomment-1561735340) that said interface is too tied to the specific JS RUM use case, making it complicated to be reused for other purposes. Because of that reason, and because the overall Android mapping process hasn't been fully defined yet, the way the Android map file is fetched as part of this PR doesn't involve reusing the `Fetcher` interface, but picking only parts of the `Fetcher` implementation related to the Elasticsearch query that would help fetching the Android map file, without having to touch the existing JS RUM interface nor its implementations for now.

Once the first iteration is working, I believe we would get a better overview of the similarities across both types of fetching mechanisms, so that we can pinpoint what parts of them can be shared and do the refactoring accordingly.

## Goals

* Add a processor that deobfuscates Android stacktraces for the OTLP data ingestion.
* Fetch android maps from ES.
* Store maps in cache and reuse them.

## Checklist


LikeTheSalad:
I'm planning to wait for the new Kibana endpoint (to upload new Android map files to Elasticsearch) to be available before I can fully finish this PR and mark it as ready for review.

LikeTheSalad:
In this Kibana PR we're adding the android map uploading endpoint which will end up adding map files into ES: https://github.com/elastic/kibana/pull/161252

internal/beater/otlp/stacktrace_processor.go
```diff
@@ -0,0 +1,64 @@
...
+func NewStacktraceProcessor(fetcher MapFetcher) *StacktraceProcessor {
+	return &StacktraceProcessor{fetcher, crashDeobfuscatorFunc(r8.Deobfuscate)}
+}
+
+func (p StacktraceProcessor) ProcessBatch(ctx context.Context, batch *modelpb.Batch) error {
+	for _, event := range *batch {
+		if isMobileCrash(event) {
+			err := p.deobfuscate(ctx, event)
+			if err != nil {
+				return err
```
LikeTheSalad, line 35: I'm not sure if raising an error in case one event's deobfuscation fails is the recommended way to address this case, given that any unchecked events (if any) will not get deobfuscated because of it. I'm open to suggestions!

internal/beater/server.go
```diff
@@ -191,13 +192,19 @@ func newServer(args ServerParams, listener net.Listener) (server, error) {
...
 
-	otlpBatchProcessor := args.BatchProcessor
+	var otlpBatchProcessor modelprocessor.Chained
+
+	esClient, err := args.NewElasticsearchClient(elasticsearch.DefaultConfig())
+	if err != nil {
+		return server{}, err
+	}
+	mapFetcher := r8.NewMapFetcher(esClient)
+	size := 128
```
LikeTheSalad, line 202: I'm not sure what's the format of this value, though I'd like to mention that an Android map could easily get to over 10MB in size, so I wanted to mention this here in case we should provide a higher number in here.

internal/beater/server.go
```diff
@@ -191,13 +192,19 @@ func newServer(args ServerParams, listener net.Listener) (server, error) {
 		return server{}, err
 	}
 
-	otlpBatchProcessor := args.BatchProcessor
+	var otlpBatchProcessor modelprocessor.Chained
+
+	esClient, err := args.NewElasticsearchClient(elasticsearch.DefaultConfig())
```
LikeTheSalad, line 197: I couldn't find how the esClient instance is initialized for RUM sourcemaps. I found about this `DefaultConfig` and it allowed me to compile so that's why I used it, although I'm not sure if this is the proper way to initialize it, or if there's another way for us to get an instance of esClient without having to initialize it in here. I'm open to any advice on this!

LikeTheSalad:
Hi @carsonip, @kruskall - I've just added all the changes I was planning to add to provide support for Android stacktrace deobfuscation. In the end, my goal is to add the deobfuscation processing to the OTLP event ingestion where we'd get the android map files from ES and will also cache them to reuse those later. The only thing that is present in the sourcemap RUM processor that I didn't add into this Android one was a listener to invalidate the cache, and the reason is that I couldn't wrap my head around how that works 😔 I'm not sure if that's crucial when it comes to storing things in cache, if so, I'm going to need some help to figure it out.

I've also added some concerns of mine across this PR, so please have a look at those as well, I'm open to any suggestions!

internal/r8/map_fetcher.go
```diff
@@ -0,0 +1,100 @@
...
+		var networkErr net.Error
+		if errors.As(err, &networkErr) {
+			return nil, fmt.Errorf("failed to reach elasticsearch: %w: %v ", errFetcherUnvailable, err)
+		}
+		return nil, fmt.Errorf("failure querying ES: %w", err)
+	}
+
+	// handle error response
+	if resp.StatusCode >= http.StatusMultipleChoices {
+		defer resp.Body.Close()
```
kruskall, line 49: **issue**: we should move this before the if, right after the error check

kruskall (changes requested):

internal/beater/server.go
kruskall: It happens here: https://github.com/elastic/apm-server/blob/main/internal/beater/beater.go#L340-L353

We're passing a callback to the method which creates the fetcher.

kruskall:
@LikeTheSalad This is great! Thanks for working on this!

I think we should add a system test at this point.

internal/beater/server.go
LikeTheSalad: Thanks for pointing me in the right direction. I just tried to replicate the same setup for the esClient instance as it's done in `beater.go`, however, I noticed that there are a couple of intricate setup parameters and processes that require creating tracers and callbacks and launching goroutines that seem too many things that I can do wrongly, so I then noticed that this code in `server.go` that adds the OTLP processors is called from `beater.go` from the same place that you mentioned, so I decided to pass the esClient initialization function and config from `beater.go` instead and that way not having to set it all up again in `server.go`. I'm unsure if I'm breaking any rules by doing so, so please have another look and let me know if there's something I need to change with that approach.

internal/beater/server.go
```diff
@@ -191,13 +192,19 @@ func newServer(args ServerParams, listener net.Listener) (server, error) {
...
+	size := 128
+	cachingFetcher, err := otlp.NewMapCachingFetcher(mapFetcher, size)
+	otlpBatchProcessor = append(otlpBatchProcessor, otlp.NewStacktraceProcessor(cachingFetcher))
 	if args.Config.AugmentEnabled {
 		// Add a model processor that sets `client.ip` for events from end-user devices.
-		otlpBatchProcessor = modelprocessor.Chained{
-			modelpb.ProcessBatchFunc(otlp.SetClientMetadata),
-			otlpBatchProcessor,
-		}
+		otlpBatchProcessor = append(otlpBatchProcessor, modelpb.ProcessBatchFunc(otlp.SetClientMetadata))
```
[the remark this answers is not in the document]
LikeTheSalad: No reason in particular, I've just made the changes to add the stacktrace deobfuscator processor after the client metadata one.

internal/r8/map_fetcher.go
LikeTheSalad: Thank you, I've just applied the changes.

LikeTheSalad:
Thank you @kruskall - I've just addressed your comments and I hope it's now better. I'm not familiarized with what you mentioned about a "system test" but if you can give me some hints about how to create it or where to find some examples, I can give it a try.

internal/beater/server.go
```diff
@@ -191,14 +199,20 @@ func newServer(args ServerParams, listener net.Listener) (server, error) {
 		return server{}, err
 	}
 
-	otlpBatchProcessor := args.BatchProcessor
+	var otlpBatchProcessor modelprocessor.Chained
```
carsonip, line 202: `args.BatchProcessor` is now ignored, is that intended? I think this will cause everything to send to nowhere.

internal/beater/server.go
carsonip: I believe this is `size` as in number of entries, not number of bytes of the data cached.

internal/r8/map_fetcher.go
```diff
@@ -0,0 +1,100 @@
...
+
+	"github.com/elastic/apm-server/internal/elasticsearch"
+	"github.com/elastic/apm-server/internal/logs"
+	"github.com/elastic/elastic-agent-libs/logp"
+	"github.com/elastic/go-elasticsearch/v8/esapi"
+)
+
+const mapIndex = ".apm-android-map"
+
+var errFetcherUnvailable = errors.New("fetcher unavailable")
```
carsonip, line 23:
```suggestion
var errFetcherUnavailable = errors.New("fetcher unavailable")
```

internal/r8/map_fetcher.go
```diff
@@ -0,0 +1,100 @@
...
+
+type MapFetcher struct {
+	client *elasticsearch.Client
+	logger *logp.Logger
+}
+
+// NewMapFetcher returns a MapFetcher.
+func NewMapFetcher(c *elasticsearch.Client) *MapFetcher {
+	logger := logp.NewLogger(logs.Sourcemap)
```
carsonip, line 32: Create a new `logs.R8` logger name

internal/r8/map_fetcher.go
```diff
@@ -0,0 +1,100 @@
...
+	body, err := io.ReadAll(resp.Body)
+
+	if err != nil {
+		return nil, err
+	}
+
+	decodedBody, err := base64.StdEncoding.DecodeString(string(body))
+	if err != nil {
+		return nil, fmt.Errorf("failed to base64 decode string: %w", err)
+	}
```
carsonip, line 73: I'm thinking about creating a streaming base64 decoder and zlib reader from resp.Body so that we avoid having multiple copies of large payload in memory, something like `zlib.NewReader(base64.NewDecoder(base64.StdEncoding, resp.Body))`

internal/beater/otlp/map_caching.go
```diff
@@ -0,0 +1,71 @@
...
+	return &MapCachingFetcher{
+		cache:   lruCache,
+		backend: backend,
+		logger:  logger,
+	}, nil
+}
+
+// Fetch fetches an android source map from the cache or wrapped backend.
+func (f *MapCachingFetcher) Fetch(ctx context.Context, name, version string) ([]byte, error) {
+	key := sourcemap.Identifier{
```
carsonip, line 44: I'd argue that we don't need to reuse anything from sourcemap package to avoid an unnecessary dependency on a mostly unrelated package. This package can have its own struct representing an identifier.

internal/beater/otlp/map_caching.go
```diff
@@ -0,0 +1,71 @@
+package otlp
```
carsonip, line 1: Should this file go into r8 package instead? Since this is more about r8 instead of about otlp.

internal/beater/otlp/map_caching.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+// NewMapCachingFetcher returns a MapCachingFetcher that wraps backend, caching results for the configured cacheExpiration.
+func NewMapCachingFetcher(
+	backend MapFetcher,
+	cacheSize int,
+) (*MapCachingFetcher, error) {
+	logger := logp.NewLogger(logs.Sourcemap)
+
+	lruCache, err := lru.NewWithEvict(cacheSize, func(key, value interface{}) {
```
carsonip, line 27: a little concerned about using an in-memory lru cache if the mapping files are going to be >10MB. We'll need to think about how much memory usage this will add. Otherwise, we'll have to cache some on local disk instead.

internal/beater/otlp/stacktrace_processor.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+type StacktraceProcessor struct {
+	fetcher      MapFetcher
+	deobfuscator CrashDeobfuscator
+}
+
+// NewStacktraceProcessor returns a StacktraceProcessor.
+func NewStacktraceProcessor(fetcher MapFetcher) *StacktraceProcessor {
+	return &StacktraceProcessor{fetcher, crashDeobfuscatorFunc(r8.Deobfuscate)}
```
carsonip, line 27: May be good to not bake in r8 here and have the caller compose it instead. Removes package dependency to r8, at least in this file.
```suggestion
func NewStacktraceProcessor(fetcher MapFetcher, deobfuscator CrashDeobfuscator) *StacktraceProcessor {
	return &StacktraceProcessor{fetcher, deobfuscator}
```

carsonip (commented):
First pass. Looking good. We'll also need some system tests on this as I suspect the current deobfuscated events are sent to nowhere.

internal/beater/server.go
LikeTheSalad: This is a great point, I didn't notice it 😨 - I'll add it back.

internal/r8/map_fetcher.go
LikeTheSalad: Than you, I'll add the changes locally to get the IDE to replace the name where this is referenced.

internal/beater/server.go
LikeTheSalad: I've just re-added it

internal/r8/map_fetcher.go
LikeTheSalad: Thanks, it's there now.

internal/r8/map_fetcher.go
LikeTheSalad: Thank you, I've applied this change and it looks way cleaner now.

internal/beater/otlp/map_caching.go
LikeTheSalad: Thanks, I've created an ide only for android maps caching.

internal/beater/otlp/map_caching.go
LikeTheSalad: Makes sense, I've moved it to R8.

internal/beater/otlp/map_caching.go
LikeTheSalad: Sure, what do you suggest we could do to verify whether this will use too much memory or not? Or should we just go ahead and cache it on the local disk right away? I'm not sure how to proceed here.

carsonip: I don't have the background knowledge here to make a decision on this. How large can a mapping file be, in normal cases and extreme cases? We'll have to ensure that `lru_size * mapping_file_size` is within reasonable memory usage while also reserving enough memory for normal APM server operation (e.g. aggregations). I'm highlighting this to make sure that we don't OOM just because we are caching large mapping files, especially for small APM servers (e.g. 1-2GB)

LikeTheSalad: Got it. Our opbeans app's map file size is about 7 MB, so considering that it's a dummy app, I'd say that it's quite easy for an app's map file size to get bigger than 10 MB. However, I don't think it's safe to say that it'd reach sizes bigger than 50 MB, so probably we can say that an extreme case could be around 50 MB (giving it some margin of error just to be safe). On top of that, it is expected that a single application can have multiple versions running at once, and each version must have its own map file. Therefore, I think it's very likely that a lot of map files can be in-memory at a time for a single app in case there are several versions of it being used (which tends to be a common scenario unless the app has some sort of force-upgrade mechanism built-in). I hope it helps to get a better idea of how much memory this feature can take.

carsonip: @simitt what do you think about this matter? Personally at least for smaller instances, the idea of caching a few 50MB mapping files sounds scary to me since it takes away a lot of memory headroom for apm-server. My feeling is that a local disk cache would make sense in this case.

simitt (changes requested):
Thank you @LikeTheSalad for all your efforts on this. After reviewing the logic in more detail, I unfortunately have several concerns with this direction.

My biggest concern is on an architectural level. Caching potentially large files completely in memory for an ever growing number of different versions (assuming that we would need to support many old versions for mobile) does not seem to be the best path forward. This quickly grows to a substantial RAM requirement, potentially requiring switching to more optimised hardware. Before moving forward in such a direction, we need to discuss alternatives. I just synced with @AlexanderWert on this topic and he'll arrange for you to discuss in further details. But one idea that Alex brought up was to look into de-obfuscating at query time. 

I also have some concerns more specific to implementation, that I would want to be tackled before merging this. But given the doubts on an architectural level, I'll safe that for now and come back to it if we decide to move forward with the de-obfuscation in the apm-server eventually.

simitt:
Closing this PR for now; we can re-open if relevant, when the final direction is clarified.

PR #11021, opened 2023-06-19
docs: describe all aggregated metrics

### LINKED ISSUE #10977: [docs] Improve and update the metrics data model docs
The [Data Model - Metrics](https://www.elastic.co/guide/en/apm/guide/current/data-model-metrics.html) documentation is somewhat half baked. The docs should be updated to explain the aggregated metrics documents and on a high level what they are used for and which fields the aggregation keys are made of. 

Before the tech writer team can pick this up, we need to provide the technical details.

carsonip:
## Motivation/summary

- Update docs on all aggregated metrics
- Add section on limits and overflows

## Checklist

## Related issues

Closes #10977


carsonip:
Should this be backported to 8.8?

docs/data-model.asciidoc
```diff
@@ -378,6 +427,40 @@ include::./data-streams.asciidoc[tag=metrics-data-streams]
...
+==== Aggregated metrics: limits and overflows
+
+For all aggregated metrics, namely transaction, service-transaction, service-destination, and service-summary metrics,
+there are limits on the number of unique groups tracked at any given time.
+
+[float]
+===== Limits
+
+For transaction metrics, the limits are 1000 services per GB of APM Server, 5000 transaction
+groups per GB of APM Server, and 500 transaction groups per service per GB of APM Server.
```
felixbarny, line 440: Maybe clarify a bit further the difference between the global limit of 5000 transaction groups and the per-service limit of 500 transaction groups.

```suggestion
For transaction metrics, the limits are 1000 services per GB of APM Server, and 5000 transaction
groups per GB of APM Server. Additionally, each service may only consume up to 10% of the transaction groups, which is 500 transaction groups per service per GB of APM Server.
```

docs/data-model.asciidoc
```diff
@@ -378,6 +427,40 @@ include::./data-streams.asciidoc[tag=metrics-data-streams]
...
+
+[float]
+===== Overflows
+
+When a dimension has a high cardinality and exceeds the limit, the metrics will be aggregated
+under a dedicated overflow bucket.
+
+For example, when `transaction.name` has a lot of unique values and reaches the limit
+of unique transaction groups tracked, any transactions with new `transaction.name` will all be aggregated under
+`transaction.name`: `_other`.
```
felixbarny, line 458: Please add another example where the transaction.name cardinality is modest but the service has a lot of instances.

felixbarny:
> Should this be backported to 8.8?

Yes, please do.

docs/data-model.asciidoc
```diff
@@ -457,6 +457,10 @@ For example, when `transaction.name` has a lot of unique values and reaches the
 of unique transaction groups tracked, any transactions with new `transaction.name` will be aggregated under
 `transaction.name`: `_other`.
 
+Another example is that `transaction.name` contains just a few unique values and stays within the transaction groups limit,
+but `service.name` has high cardinality and reaches the max services limit. Any transactions with new `service.name` will
+be aggregated under `service.name`: `_other` and `transaction.name`: `_other`.
```
felixbarny, line 462: I was going for something different: the transaction name and the service name have low cardinality but there a service has many different instances (is deployed on a lot of hosts).


```suggestion
Another example of how the transaction group limit may be reached is if `transaction.name` contains just a few unique values,
but the service is deployed on a lot of different hosts. As `host.name` is part of the aggregation key for transaction metrics, the max transaction group limit is reached for a service that has 100 instances and 40 different transaction names when connected to an APM Server with 8GB of RAM . Once this limit is reached, any new combinations of transaction.name and host.name for that service will be aggregated under `transaction.name`: `_other`.
```

If you want to apply the suggestion verbatim, you'll probably need to adjust the line lengths.

docs/data-model.asciidoc
```diff
@@ -337,6 +337,34 @@ You can filter and group by these dimensions (some of which are optional, for ex
...
+[float]
+===== Service-transaction metrics
+
+To power {kibana-ref}/xpack-apm.html[{apm-app}] visualizations,
+APM Server aggregates transaction events into service-transaction metrics.
+
+*`transaction.duration.summary`* and *`transaction.duration.histogram`*::
++
+--
+These metrics measure the aggregated latency and latency distribution of service transaction groups,
```
felixbarny, line 349:
```suggestion
These metrics represent the latency summary and latency distribution of service transaction groups,
```

felixbarny (approved):

docs/data-model.asciidoc
```diff
@@ -337,6 +337,34 @@ You can filter and group by these dimensions (some of which are optional, for ex
...
+used to power visualizations and analytics in Elastic APM.
```
simitt, line 350: Do we expect users to be able to know the difference between transaction groups and service transaction groups?

docs/data-model.asciidoc
```diff
@@ -378,6 +427,47 @@ include::./data-streams.asciidoc[tag=metrics-data-streams]
...
+groups per GB of APM Server. Additionally, each service may only consume up to 10% of the transaction groups,
+which is 500 transaction groups per service per GB of APM Server.
+
+For service-transaction metrics, the limit is 1000 service transaction groups per GB of APM Server.
+
+For service-destination metrics, the limit is a constant of 10000 service destination groups.
+
+For service-summary metrics, the limit is 1000 service summary groups per GB of APM Server.
+
+Note that all of the above limits may change in the future with further improvements.
```
simitt, line 449: Can this be highlighted a bit more?

docs/data-model.asciidoc
```diff
@@ -378,6 +427,47 @@ include::./data-streams.asciidoc[tag=metrics-data-streams]
...
+the max transaction group limit is reached for a service that has 100 instances and 40 different transaction names when
+connected to an APM Server with 8GB of RAM. Once this limit is reached, any new combinations of `transaction.name` and
+`host.name` for that service will be aggregated under `transaction.name`: `_other`.
+
+This issue can be resolved by increasing memory available to APM Server, or by ensuring that the dimensions do not use values
+that are based on parameters that can change. For example, user ids, product ids, order numbers, query parameters, etc.,
+should be stripped away from the dimensions.
+
 [float]
 ==== Example metric document
```
simitt, line 472: The example documents should show an example for every metric type.

docs/data-model.asciidoc
simitt, line 1: I would call it out at the beginning of the _Calculated metrics_ section that calculated metrics are treated as an implementation detail to power the UI and while we aim for stability also for these data models, we might change the dimensions and concrete limits for aggregations within minor version updates. 
WDYT?

simitt (changes requested):
Left some suggestions.

docs/data-model.asciidoc
```diff
@@ -378,6 +427,47 @@ include::./data-streams.asciidoc[tag=metrics-data-streams]
 
 See <<apm-data-streams>> to learn more.
 
+[float]
+==== Aggregated metrics: limits and overflows
```
felixbarny, line 431: When this is merged, we should also link https://www.elastic.co/guide/en/kibana/8.8/troubleshooting.html#troubleshooting-too-many-transactions to this section.

Longer term, we could think about how to combine the UI-centric and the Server-centric documentation with something more wholistic.

docs/data-model.asciidoc
carsonip: Actually how do we explain the service transaction grouping?

docs/metricset-indices.asciidoc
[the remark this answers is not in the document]
carsonip: Yes I'm wondering the value of this duplicated page. I think it was for "legacy" APM Server. I'll refactor a bit.

docs/data-model.asciidoc
simitt: How about we highlight that they are generally producing the same kind of metrics, but that the service transaction metrics usually have a much lower cardinality as they have significantly less dimensions. The UI uses them where less details of the transactions are needed - so basically service transaction metrics are used for service oriented visualisations, whereas transaction metrics are powerin transaction oriented pages, where more details and slicing of the transaction groups is helpful.

docs/data-model.asciidoc
```diff
@@ -354,50 +452,118 @@ These metric documents can be identified by searching for `metricset.name: servi
...
+under a dedicated overflow bucket.
+
+For example, when `transaction.name` has a lot of unique values and reaches the limit
+of unique transaction groups tracked, any transactions with new `transaction.name` will be aggregated under
+`transaction.name`: `_other`.
+
+Another example of how the transaction group limit may be reached is if `transaction.name` contains just a few unique values,
+but the service is deployed on a lot of different hosts. As `host.name` is part of the aggregation key for transaction metrics,
+the max transaction group limit is reached for a service that has 100 instances and 40 different transaction names when
+connected to an APM Server with 8GB of RAM. Once this limit is reached, any new combinations of `transaction.name` and
```
felixbarny, line 560: That example doesn't account for the transaction result also being dimension?

```suggestion
the max transaction group limit is reached for a service that has 100 instances, 10 different transaction names, and 4 transaction results per transaction name when
connected to an APM Server with 8GB of RAM. Once this limit is reached, any new combinations of `transaction.name` and
```

cc @dgieselaar

carsonip: hmm, do we want to update the next sentence as well to avoid confusion?
>  any new combinations of `transaction.name` and...

simitt (approved):
Thanks, LGTM!

docs/data-model.asciidoc
carsonip: took the liberty to update the following sentence to include `transaction.result` as well.

felixbarny: thanks!

Issue #10716, opened 2023-04-28
Stop dropping metadata fields in span documents


axw:
From https://github.com/elastic/apm-server/pull/10697#pullrequestreview-1400922831:

> I wonder if we should drop that `remove` processor altogether. This has been around since forever, when we used to store a transaction document even for non-sampled transactions, and when we didn't have a default ILM delete phase . I feel like we have better options for storage management these days.

We should investigate how much storage these fields actually take, using Rally, and consider dropping that pipeline altogether so spans and transactions have the same metadata fields.

dmathieu:
I have been running the rally tests against the main branch, and with the following diff, dropping the remove processor for span documents:

```diff
diff --git a/apmpackage/apm/data_stream/traces/elasticsearch/ingest_pipeline/default.yml b/apmpackage/apm/data_stream/traces/elasticsearch/ingest_pipeline/default.yml
index fba9ac0fa..870baac68 100644
--- a/apmpackage/apm/data_stream/traces/elasticsearch/ingest_pipeline/default.yml
+++ b/apmpackage/apm/data_stream/traces/elasticsearch/ingest_pipeline/default.yml
@@ -26,20 +26,3 @@ processors:
       field: event.success_count
       value: 0
       ignore_failure: true
-  # end of event.success_count logic
-  - remove:
-      # Remove some metadata from spans that is available in the parent transaction, to cut down on storage costs.
-      if: ctx.processor?.event == 'span'
-      field:
-        - process
-        - user
-        - user_agent
-        - container
-        - kubernetes
-        - service.node
-        - service.language
-        - service.runtime
-        - service.framework
-      ignore_missing: true
-      ignore_failure: true
```

I ran the tests multiple times. All of them had roughly similar results.

This diff is a run with the rally config from main:
```diff
diff --git a/original.txt b/original.txt
index 7fd24d6..309bd54 100644
--- a/original.txt
+++ b/original.txt
@@ -1,37 +1,37 @@
 |                                                         Metric |   Task |            Value |   Unit |
 |---------------------------------------------------------------:|-------:|-----------------:|-------:|
-|                     Cumulative indexing time of primary shards |        |      0.533967    |    min |
+|                     Cumulative indexing time of primary shards |        |      0.650567    |    min |
 |             Min cumulative indexing time across primary shards |        |      0           |    min |
 |          Median cumulative indexing time across primary shards |        |      8.33333e-06 |    min |
[...]

PR #11073, opened 2023-06-26
Set max-num-segments to 1 in rally config

dmathieu:
## Motivation/summary

When running force-merge after rally tests, this change ensures we merge the data into a single segment, to provide more stable test results.

See https://esrally.readthedocs.io/en/stable/track.html#force-merge

## Checklist

- [x] **NA** Update [CHANGELOG.asciidoc](https://github.com/elastic/apm-server/blob/main/CHANGELOG.asciidoc)
- [x] **NA** Update [package changelog.yml](https://github.com/elastic/apm-server/blob/main/apmpackage/apm/changelog.yml) (only if changes to `apmpackage` have been made)
- [x] **NA** Documentation has been updated

## How to test these changes

Run the rally-cloud test suite.

## Related issues

* #10716

### DIFF
--- a/testing/rally/track.json
+++ b/testing/rally/track.json
@@ -38,7 +38,8 @@
     },
     {
       "operation": {
-        "operation-type": "force-merge"
+        "operation-type": "force-merge",
+        "max-num-segments": 1
       }
     }
   ]


axw (approved):
thank you!

PR #11089, opened 2023-06-28
Stop dropping metadata fields in span documents

### LINKED ISSUE #10716: Stop dropping metadata fields in span documents

dmathieu:
## Motivation/summary

We drop these fields for spans, but not for transactions. So not dropping them makes both events more aligned.

## Checklist

- [x] **N/A** Update [package changelog.yml](https://github.com/elastic/apm-server/blob/main/apmpackage/apm/changelog.yml) (only if changes to `apmpackage` have been made)
- [x] **N/A** Documentation has been updated

## How to test these changes

Run the suite of automated changes.

## Related issues

Closes #10716.


apmpackage/apm/data_stream/traces/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -26,20 +26,3 @@ processors:
       field: event.success_count
       value: 0
       ignore_failure: true
-  # end of event.success_count logic
```
kruskall, line 29: **question**: why was this removed ? I think we want to keep this comment

changelogs/head.asciidoc
```diff
@@ -17,3 +17,5 @@ https://github.com/elastic/apm-server/compare/8.9\...main[View commits]
 
 [float]
 ==== Added
+
```
carsonip, line 20:
```suggestion
```

carsonip (approved):

axw (approved):
LGTM!

@AlexanderWert @felixbarny FYI. There's a small storage size increase (see https://github.com/elastic/apm-server/issues/10716), but this is necessary for us to converge transactions and spans.

felixbarny (approved):

axw:
Verified with 8.10.0-BC1. I sent some transactions and spans and confirmed that the spans had `process.*` fields in tact.

PR #11116, opened 2023-06-30
Add sourcemap periodic refresh tracing

dmathieu:
## Motivation/summary

Sourcemap periodic refreshes are making external calls to ES. Those calls produce spans, but they can't generate any data, as there is no parent transaction.
This creates a transaction for each sync run, so we can observe them.

## Checklist

## How to test these changes

Outside of unit tests, I have been testing these changes by running apm-server with tilt and rum/self instrumentation enabled.

Screenshot 2023-06-30 at 09 42 52

I'm not 100% sure about the `initialSearch` and `scrollSearch` spans, since they do nothing but wrap the ES call. They also help in figuring out which method makes those calls though.

## Related issues

See #1014.


dmathieu:
System tests failed once du to a flaky test. They passed on a second run, but somehow, the `System Tests Results` is still marked as failed. And rerunning it doesn't change anything.

changelogs/head.asciidoc
```diff
@@ -17,3 +17,4 @@ https://github.com/elastic/apm-server/compare/8.9\...main[View commits]
 
 [float]
 ==== Added
+- Add a transaction to the source map periodic refresh action. {pull}11116[11116]
```
carsonip, line 20: Would be great to mention this is for instrumentation only.

carsonip (approved):

changelogs/head.asciidoc
dmathieu: How about now?

carsonip (approved):

axw:
Verified with 8.10.0-BC1. Running apm-server under Tilt, I configured `instrumentation.enabled: true` and `apm-server.rum.enabled: true` and observed the "MetadataESFetcher.sync" traces being captured:

image

I think the transaction name could mention sourcemaps somewhere, but this is good enough for now.

PR #11134, opened 2023-07-03
test: ignore tail based sampling policy for 7.x smoke tests

kruskall:
## Checklist

## How to test these changes

`make smoketest/run-version TEST_DIR=testing/smoke/legacy-managed SMOKETEST_VERSION=7.17`

### DIFF
--- a/testing/smoke/lib.sh
+++ b/testing/smoke/lib.sh
@@ -344,19 +344,25 @@ data_stream_assert_pipelines() {
 data_stream_assert_templates_ilm() {
     echo "-> Asserting component templates and ILM policies..."
     local SUCCESS=true
+    local VERSION=${1}
 
     local COMPOSABLE_TEMPLATES=($(elasticsearch_curl "/_component_template" | jq -c -r '.component_templates|to_entries[]|select(.value.component_template._meta.package.name == "apm")|.value'))
     for ct in "${COMPOSABLE_TEMPLATES[@]}"; do
         local CT_NAME=$(echo "${ct}" | jq -r '.name')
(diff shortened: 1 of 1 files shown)


carsonip (commented):
Could you fill in the "Motivation/summary" of the PR? 7.17 TBS ILM policy name is broken and this test is catching it. Is this fix just temporary?

kruskall:
Thank you for the reminder! :bow: 

I've updated the description!

testing/smoke/lib.sh
```diff
@@ -344,19 +344,25 @@ data_stream_assert_pipelines() {
...
         if [[ ! -z "${ILM_POLICY_NAME}" ]]; then
-            # Component template has ILM policy attached, check if ILM policy actually exists.
-            local RES
-            local RC=0
-            RES=$(elasticsearch_curl "/_ilm/policy/${ILM_POLICY_NAME}") || RC=$?
-            if [ $RC -ne 0 ]; then
-                echo "-> ILM policy ${ILM_POLICY_NAME} error: ${RES}"
-                SUCCESS=false
+            # ignore tail based sampling ILM policy for 7.x
+            if [ "${ILM_POLICY_NAME}" = "traces-apm.sampled-default_policy" ] && [ "${VERSION}" != "${VERSION#7.}" ]; then
```
simitt, line 355: Am I reading this wrong - or should it be `"${VERSION}" == "${VERSION#7.}"`? We want to skip the check only for `7.x`.

kruskall: I don't think that would work.

The idea is that if we trim the prefix `7.` and the version changes then we are on version 7 and we should skip the test.

On 7.17.11 this check would become: `7.17.11 != 17.11`
On 8.9: `8.9 != 8.9`

simitt (approved):

PR #11137, opened 2023-07-04
sampling: use Marshal/UnmarshalVT

axw:
## Motivation/summary

Minor improvement to use generated marshal/unmarshal methods for tail-based sampling. Remove the JSON codec, it's not used anymore.

## Checklist

## How to test these changes

N/A, non-functional change.

## Related issues

None


lahsivjar (approved):

marclop (approved):
Looking forward to this! Should a change log entry be added as well?

axw:
> Should a change log entry be added as well?

@marclop I don't think so: there's no functional change, so users shouldn't care.

marclop:
I was thinking that's an improvement where APM Server will be able to store more traces in badger vs before, but perhaps that's not important to point out.

axw:
Ah I see. This was just a switch from reflection-based protobuf encoding/decoding to vtprotobuf-generated code. Maybe we want to talk about the overall switch to using protobuf encoding.

Issue #10991, opened 2023-06-14
Enable routing data to different data streams


felixbarny:
We'd like to allow users to use a [`reroute`](https://www.elastic.co/guide/en/elasticsearch/reference/current/reroute-processor.html) processor to send APM data to arbitrary namespaces. Currently, when using a Fleet-managed APM Server, sending data to a different data stream using the reroute process isn't possible. That's because the API Key permissions are very narrow and only allow sending data to an exact data stream.

Similarly to what we've done for some other integrations where we want to allow routing (see https://github.com/elastic/integrations/pull/6340), we just need to add the `elasticsearch.dynamic_dataset: true` and `elasticsearch.dynamic_namespace: true` flags, accordingly.

I'm proposing to add `elasticsearch.dynamic_namespace: true`  to all data streams and to add `elasticsearch.dynamic_dataset: true` only to the [app_metrics](https://github.com/elastic/apm-server/tree/main/apmpackage/apm/data_stream/app_metrics) and [`app_logs`](https://github.com/elastic/apm-server/tree/main/apmpackage/apm/data_stream/app_logs) data streams. That's because changing the data set for other data stream, such as [`internal_metrics`](https://github.com/elastic/apm-server/tree/main/apmpackage/apm/data_stream/internal_metrics) could confuse the APM UI. If required, we can always add `elasticsearch.dynamic_dataset: true` to other data streams, such as the [`traces`](https://github.com/elastic/apm-server/tree/main/apmpackage/apm/data_stream/traces) data stream.

felixbarny:
One caveat for allowing routing to different datasets for custom metrics is that the target data streams need to have to have the dynamic templates and ingest processors that are used in the `metrics-apm.app_metrics-*` data stream.

PR #11168, opened 2023-07-10
Allow dynamic datasets and namespaces

dmathieu:
## Motivation/summary

This change grants the permissions for agents running the apmpackage integration to send data to other data streams, and other namespaces, making the use of the `reroute` processor possible.

## Checklist

## How to test these changes

I ran this integration locally with tilt.

## Related issues

Closes https://github.com/elastic/apm-server/issues/10991


apmpackage/apm/changelog.yml
```diff
@@ -3,6 +3,9 @@
     - description: Placeholder
```
dmathieu, line 3: Should the placeholder be removed manually, or do we remove it when the release happens?

simitt: Afair, it's removed when we branch out.

simitt (commented):
From the issue description:
> to add elasticsearch.dynamic_dataset: true only to the [app_metrics](https://github.com/elastic/apm-server/tree/main/apmpackage/apm/data_stream/app_metrics) and [app_logs](https://github.com/elastic/apm-server/tree/main/apmpackage/apm/data_stream/app_logs) data streams. That's because changing the data set for other data stream, such as [internal_metrics](https://github.com/elastic/apm-server/tree/main/apmpackage/apm/data_stream/internal_metrics) could confuse the APM UI.

Was this overlooked or were there any follow up conversations/considerations that led to adding `dynamic_dataset: true` everywhere?

dmathieu:
> Was this overlooked or were there any follow up conversations/considerations that led to adding dynamic_dataset: true everywhere?

Ah yes, my bad. I overlooked that. I've kept it only for app_metrics and app_logs.
I do wonder if we should would be interesting to have it for traces as well?

simitt:
I was wondering the same tbh. @felixbarny was there a special reason for keeping `traces-*` out of this? Any concerns for the APM UI?

felixbarny:
I don't have a strong opinion on whether to allow dynamic_dataset for traces. I suppose there's no harm in allowing that.

However, we should add `dynamic_namespace` to all APM data streams. This enables users to send all APM data to a dedicated namespace so that they can use custom retention and security policies.

axw (approved):
LGTM. If we go ahead with https://github.com/elastic/elasticsearch/pull/97546, then I expect we'll end up giving APM append-only privileges for `traces-*`, `metrics-*`, and `logs-*`. This seems pretty well aligned.

axw:
Verified with 8.10.0-BC1.

I created `@custom` ingest pipelines with the `reroute` processor, setting dataset to a constant value ("apm.app.all") for app logs/metrics; and setting namespace to a template (some "{{agent.name}}", some "{{service.name}}")

Sending test data with apm-tool, I verified this with `logs-apm.app.*`, `metrics-apm.app.*`, `traces-apm-*`, `logs-apm.error-*`, and `metrics-apm.service_destination.1m-*`.

PR #11169, opened 2023-07-10
fix incorrect "same-kind" strategy default

jackshirazi:
## Motivation/summary

The same-kind span compression strategy default is 0ms in all agents, not 5ms


marclop (approved):

AlexanderWert (approved):

docs/span-compression.asciidoc
```diff
@@ -52,7 +52,7 @@ The agent selects this strategy if two adjacent spans have the same:
 === Settings
 
 The agent has configuration settings to define upper thresholds in terms of span duration for both strategies.
-For the "Same-Kind" strategy, the limit is 5 milliseconds. For the "Exact-Match" strategy, the limit is 50 milliseconds.
+For the "Same-Kind" strategy, the limit is 0 milliseconds. For the "Exact-Match" strategy, the limit is 50 milliseconds.
```
AlexanderWert, line 55:
```suggestion
For the "Same-Kind" strategy, the default limit is 0 milliseconds, which means the "Same-Kind" strategy is disabled by default. For the "Exact-Match" strategy, the default limit is 50 milliseconds.
```

PR #11166, opened 2023-07-10
systemtest: Use `elastic/apm-tools`

marclop:
## Motivation/summary

Uses the `elastic/apm-tools` module and removes the approvals from the systemtest module.

## Checklist

N/A

## How to test these changes

N/A

## Related issues

N/A

PR #11190, opened 2023-07-12
systemtest: Remove loadgen, use elastic/apm-perf

marclop:
## Motivation/summary

Removes `systemtest/loadgen` since a more recent version exists in the `elastic/apm-perf` repository.

Also caused a systemtest output to be updated since the systemtest were using the APM Go Agent v1.

## Checklist

N/A

## How to test these changes

N/A

## Related issues

unblocks #11166


axw (approved):
:heart:

PR #11225, opened 2023-07-17
Fix mergify 7.17 status checks

reakaleek:
## Details

https://github.com/elastic/apm-server/pull/11194 is not mergable because it's waiting for the status check `system-test`. However, this check does not exist on `7.17`. 

This PR fixes it by creating a job named `system-test` that depends on the outcome of `go-system-tests` and `python-system-tests`.


axw (approved):
nice :)

kuisathaverat (approved):

PR #11344, opened 2023-08-07
Feat/badgerv4

kruskall:
## Motivation/summary

badger v2 is severely out of date and newer versions have important performance improvements

## Checklist


x-pack/apm-server/sampling/processor_test.go
```diff
@@ -573,87 +573,6 @@ func TestStorageMonitoring(t *testing.T) {
 	assert.NotZero(t, metrics.Ints, "sampling.storage.value_log_size")
 }
 
-func TestStorageGC(t *testing.T) {
```
kruskall, line 576: For reviewers: this will fail because the meaning of vlog files changed, see https://github.com/dgraph-io/badger/pull/1555

axw (commented):
Looks good!

IIUC this would be a breaking change as the data format has changed, so old files wouldn't be able to be loaded. Is that correct? If so then we should list it in the changelog. Also, have you confirmed the behaviour when upgrading from an older apm-server version?

kruskall:
This is what happened when trying to load a directory on the previous version:

 
> panic: manifest has unsupported version: 7 (we support 8).
> Please see https://dgraph.io/docs/badger/faq/#i-see-manifest-has-unsupported-version-x-we-support-y-error on how to fix this. 

I don't think the migration procedure is feasible. Could we change the directory path ? e.g. `tail_sampling` -> `tail_sampling_v4` ? WDYT ?

axw:
> I don't think the migration procedure is feasible. Could we change the directory path ? e.g. tail_sampling -> tail_sampling_v4 ? WDYT ?

That sounds good. The alternative would be to automatically delete the old data, but I think it would be better to leave it there and give the user an option to manually backup/restore.

OTOH that would mean the old data will be left behind forever. Maybe we could check the modification time of files in the `tail_sampling` directory at startup (if it exists), and remove it if all files are older than the TTL. WDYT?

x-pack/apm-server/main.go
```diff
@@ -117,6 +119,23 @@ func newTailSamplingProcessor(args beater.ServerParams) (*sampling.Processor, er
...
+		var newest time.Time
+		for _, de := range entries {
+			if i, err := de.Info(); err == nil {
+				if t := i.ModTime(); t.After(newest) {
+					newest = t
+				}
+			}
+		}
+
+		if time.Since(newest) > tailSamplingConfig.TTL {
```
axw, line 132:
```suggestion
		if newest.IsZero() || time.Since(newest) > tailSamplingConfig.TTL {
```

If there are no entries, we should still delete the directory.

changelogs/head.asciidoc
```diff
@@ -5,6 +5,7 @@ https://github.com/elastic/apm-server/compare/8.10\...main[View commits]
 
 [float]
 ==== Breaking Changes
+- Update badger to v4 {pull}11344[11344]
```
axw, line 8: We should mention why it's breaking - what the implications are for users.

x-pack/apm-server/sampling/processor_test.go
axw: Thinking about this again, I'm not sure we should just remove the test. GC is still needed, it's just that how it works has changed. Can we instead update the test?

axw (commented):
Thanks for the updates @kruskall. Looks good overall, my main concern now is the removed test.

PR #11374, opened 2023-08-10
action: support for DRA

v1v:
## Motivation/summary

Use GitHub actions and Buildkite to package and run the DRA.

This will run on push events or manually.

## How to test these changes

In the CI

## Related issues

Requires https://github.com/elastic/apm-server/pull/11373 and https://github.com/elastic/apm-server/pull/11399

## Tasks

- [ ] exclude if main and staging.
- [ ] exclude if branch is not supported anymore isBranchUnifiedReleaseAvailable()
- [ ] upload/download artifacts so they can be consumed when running the DRA.
- [ ] Add pipeline in the catalog entry

### DIFF
--- a/.buildkite/hooks/pre-command
+++ b/.buildkite/hooks/pre-command
@@ -1,0 +1,23 @@
+#!/usr/bin/env bash
+##
+##  This script prepares the Vault context and required tooling
+##  for the release and snapshot pipelines.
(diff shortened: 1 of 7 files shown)


v1v:
I'll close this particular PR so I can use a feature branch instead

PR #11072, opened 2023-06-26
Universal Profiling: Remove all UP-related code and tests

christos68k:
### Summary

UP now runs outside of APM server starting from 8.9.

**NOTE:** This should NOT be merged unless the profiling team has signed off on merging, specifically. In the meantime it's useful to have reviews done so that this PR becomes mergeable.

There are standalone profiling tests remaining to be performed in production and we'd like to keep UP-in-APM-server available as a backup in case we run into major problems. The expectation being that no major issues arise and we merge this as part of 8.9.

PR #11434, opened 2023-08-20
fix: add back gzip support for grpc otlp endpoint

kruskall:
## Motivation/summary

gzip support was removed in https://github.com/elastic/apm-server/pull/11072 along with UP code.
`_ "google.golang.org/grpc/encoding/gzip"` was imported in `x-pack/apm-server/profiling/collector.go` in the `profiling` package which was imported into the main package, enabling gzip support.

This PR adds back the import which runs the `init` function enabling gzip support.
The grpc otlp tests have been updated to send gzip request to validate that the feature is working.
The fix is backported to 8.10 and 8.9 and it's related to SDH `#1053`

## Checklist

### DIFF
--- a/changelogs/head.asciidoc
+++ b/changelogs/head.asciidoc
@@ -11,6 +11,7 @@ https://github.com/elastic/apm-server/compare/8.10\...main[View commits]
 
 [float]
 ==== Bug fixes
+- Add back gzip support for grpc otlp endpoint {pull}11434[11434]
 
 [float]
 ==== Intake API Changes


axw (approved):
:anguished:

LGTM, please add a changelog entry.

systemtest/otlp_test.go
```diff
@@ -206,7 +206,7 @@ func TestOTLPGRPCMetrics(t *testing.T) {
 
 	// opentelemetry-go does not support sending Summary metrics,
 	// so we send them using the lower level OTLP/gRPC client.
-	conn, err := grpc.Dial(serverAddr(srv), grpc.WithInsecure(), grpc.WithBlock())
+	conn, err := grpc.Dial(serverAddr(srv), grpc.WithInsecure(), grpc.WithBlock(), grpc.WithDefaultCallOptions(grpc.UseCompressor("gzip")))
```
dmathieu, line 209: Could there be a future regression where it's requests without gzip that don't work? Should we also test without gzip enabled?

axw:
Verified with 8.10.0-BC1 using https://github.com/elastic/apm-tools/pull/22

PR #11437, opened 2023-08-21
apmpackage: remove `processor.name` field

axw:
## Motivation/summary

This field is not used by the UI, and is redundant; its value can always be inferred from `data_stream.dataset`.

See https://github.com/elastic/kibana/issues/161718#issuecomment-1638294156

## Checklist

## How to test these changes

Verify that apm-server 8.9.0 can still index into an 8.11.0 deployment

## Related issues

https://github.com/elastic/kibana/issues/161718

### DIFF
--- a/apmpackage/apm/changelog.yml
+++ b/apmpackage/apm/changelog.yml
@@ -1,8 +1,8 @@
 - version: generated
   changes:
-    - description: Placeholder
+    - description: Remove unused `processor.name` field
       type: enhancement
-      link: https://github.com/elastic/apm-server/pull/123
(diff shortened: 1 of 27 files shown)


carsonip (approved):
lgtm. There shouldn't be any backcompat issues as we do not use strict mappings now.

lahsivjar:
Verified the following points with `8.11.0 BC3`:
- `8.11.0` APM Server can index events without setting `processor.name` field
- `8.9.0` APM Server can index fields without error (but with `processor.name` field)

Screenshot 2023-10-24 at 7 04 10 PM

Issue #10623, opened 2023-04-06
Create a PoC for collecting apm-server metrics for built-in observability


simitt:
Create a PoC for metrics collection for improving the apm-server observability. 

Use the OpenTelemetry library and setup custom telemetry where needed. The otel SDK can be configured to have two exporters. A custom exporter should be used to make data available to the customers. The custom exporter should reduce the amount of data that is sent to customers cluster by default, to avoid overhead and storage costs for the customer.

dmathieu:
I've been thinking about this, and it seems to me an unknown remains.
We already have [self instrumentation](https://github.com/elastic/apm-server/blob/fc5774e409c8849a27b166ebe85d0eea2685bf9f/apm-server.yml#L736), which uses apm-go-agent to emit traces and (all, though only internal) metrics.

We could either reuse the self instrumentation configuration, or introduce something new which would emit metrics.
The problem with the introduction of something new is that it duplicates very similar configuration and could be quite confusing to know which one someone wants to use.
The problem with reusing self instrumentation is that it sends all of the telemetry, not only a subset of the metrics, as we want to do here.

If we want to reuse self instrumentation, we could start very easily right now by enabling the [otel metrics exporter](https://github.com/elastic/apm-agent-go/blob/main/docs/opentelemetry.asciidoc#initializing-the-metrics-exporter), and start emitting metrics with OpenTelemetry.
If we introduce a new configuration, we probably need a bigger RFC to converge on the proper new configuration.

simitt:
I don't expect this to be configurable by users in general. We want to always collect a minimal, pre-defined set of metrics which should be indexed into the users Elasticsearch. Whereas the self-instrumentation could be configured to index into a different ES endpoint.

dmathieu:
With https://github.com/elastic/apm-server/pull/11582, we are now exposing the counters from HTTP and gRPC requests.
With https://github.com/elastic/apm-server/pull/11613, we will also be exposing the request duration histogram, which will give us all the metrics described for this PoC, and should let us close this issue.

dmathieu:
Closing per the comment above.

PR #11470, opened 2023-08-23
Otel exporter for internal telemetry

dmathieu:
## Motivation/summary

This sets up an OpenTelemetry exporter to collect internal metrics, export them as internal APM events and send them to the processor.

It is the boilerplate for #10623, yet no metrics are emitted yet. So this code isn't used at the moment.

## Checklist

## How to test these changes

This PR is unit-tested.

## Related issues

* #10623

### DIFF
--- a/internal/telemetry/metric_exporter.go
+++ b/internal/telemetry/metric_exporter.go
@@ -135,7 +135,14 @@ func addMetric(sm metricdata.Metrics, ms metricsets) error {
 	case metricdata.Histogram[float64]:
 		// Float Histogram
 	case metricdata.Sum[int64]:
-		// Int Sum
+		for _, dp := range m.DataPoints {
+			sample := modelpb.MetricsetSample{
+				Name:  sm.Name,
+				Type:  modelpb.MetricType_METRIC_TYPE_COUNTER,
+				Value: float64(dp.Value),
+			}
+			ms.upsert(dp.Time, dp.Attributes, &sample)
+		}
 	case metricdata.Sum[float64]:
 		for _, dp := range m.DataPoints {
 			sample := modelpb.MetricsetSample{
@@ -146,9 +153,23 @@ func addMetric(sm metricdata.Metrics, ms metricsets) error {
 			ms.upsert(dp.Time, dp.Attributes, &sample)
 		}
 	case metricdata.Gauge[int64]:
-		// Int Gauge
+		for _, dp := range m.DataPoints {
+			sample := modelpb.MetricsetSample{
+				Name:  sm.Name,
+				Type:  modelpb.MetricType_METRIC_TYPE_GAUGE,
+				Value: float64(dp.Value),
+			}
+			ms.upsert(dp.Time, dp.Attributes, &sample)
+		}
 	case metricdata.Gauge[float64]:
-		// Float Gauge
+		for _, dp := range m.DataPoints {
+			sample := modelpb.MetricsetSample{
+				Name:  sm.Name,
+				Type:  modelpb.MetricType_METRIC_TYPE_GAUGE,
+				Value: dp.Value,
+			}
+			ms.upsert(dp.Time, dp.Attributes, &sample)
+		}
 	default:
 		return fmt.Errorf("unknown metric type %q", m)
 	}
--- a/internal/telemetry/metric_exporter.go
+++ b/internal/telemetry/metric_exporter.go
@@ -76,10 +76,6 @@ func (e *MetricExporter) Export(ctx context.Context, rm *metricdata.ResourceMetr
 			Name:     "apm-server",
 			Language: &modelpb.Language{Name: "go"},
 		},
-		Agent: &modelpb.Agent{
-			Name:    "internal",
-			Version: "unknown",
-		},
 		Event: &modelpb.Event{
 			Received: modelpb.FromTime(now),
 		},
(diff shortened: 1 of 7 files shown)


internal/telemetry/metric_exporter.go
```diff
@@ -0,0 +1,341 @@
...
+			event := baseEvent.CloneVT()
+			event.Timestamp = modelpb.FromTime(key.timestamp)
+			metrs := make([]*modelpb.MetricsetSample, 0, len(ms.samples))
+			for _, s := range ms.samples {
+				metrs = append(metrs, s)
+			}
+			event.Metricset = &modelpb.Metricset{Samples: metrs, Name: "app"}
+			if ms.attributes.Len() > 0 {
+				event.Labels = modelpb.Labels(event.Labels).Clone()
+				event.NumericLabels = modelpb.NumericLabels(event.NumericLabels).Clone()
```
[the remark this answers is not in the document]
dmathieu: We may not need to clone them, I did it that way here because that's what apm-data does.
https://github.com/elastic/apm-data/blob/main/input/otlp/metadata.go#L322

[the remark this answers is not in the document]
dmathieu: Actually, it seems the `Clone()` is used to initialize. Without it, I'm getting `panic: assignment to entry in nil map` panics.

internal/telemetry/metric_exporter_config.go
```diff
@@ -0,0 +1,109 @@
...
+// WithMetricFilter configured the metrics filter. Any metric filtered here
+// will be the only ones to be exported. All other metrics will be ignored.
+//
+// Defaults to not exporting anything.
+func WithMetricFilter(f []string) ConfigOption {
+	return func(cfg Config) Config {
+		cfg.MetricFilter = f
+		return cfg
+	}
+}
```
axw, line 89: Should we be using Views? e.g. https://pkg.go.dev/go.opentelemetry.io/otel/sdk/metric#example-NewView-Drop

internal/telemetry/metric_exporter.go
```diff
@@ -0,0 +1,341 @@
...
+
+	baseEvent := modelpb.APMEvent{
+		Service: &modelpb.Service{
+			Name:     "apm-server",
+			Language: &modelpb.Language{Name: "go"},
+		},
+		Agent: &modelpb.Agent{
+			Name:    "internal",
+			Version: "unknown",
+		},
```
axw, line 80: Is this needed?

internal/telemetry/metric_exporter.go
```diff
@@ -0,0 +1,341 @@
...
+func (e *MetricExporter) isMetricFiltered(n string) bool {
+	for _, v := range e.metricFilter {
+		if v == n {
+			return false
+		}
+	}
+	return true
+}
+
+func addMetric(sm metricdata.Metrics, ms metricsets) error {
```
axw, line 142: Could you please add a reference to https://github.com/elastic/apm-data/blob/main/input/otlp/metrics.go? And maybe open an issue/add a TODO to find a way to reuse the apm-data code?

axw (commented):
I didn't look too closely at the modelpb usage, just the overall structure. Looks good, just a few comments.

internal/telemetry/metric_exporter_config.go
dmathieu: A view is applied to every instrumentation which matches it. So if we used it, it would filter the metrics for every exporter, not just this one.
We don't want to filter metrics that way for apmotel.

internal/telemetry/metric_exporter.go
dmathieu: apm-data sets it, I figured kibana may uses it somehow?

internal/telemetry/metric_exporter.go
dmathieu: I have added a reference. However, I don't think we can/want to reuse the apm-data code. While the logic is extremely similar, the structs are different (apm-data uses the collector ones, while this one must use the SDK ones).

axw: Yeah I don't mean directly, I mean indirectly by converting to the collector types. As we discussed live, it would add some overhead and may be complex - so remains to be seen whether it's worthwhile. OTOH it would avoid duplicating the logic for non-trivial metric conversions (grouping by dimensions/attributes, histogram conversion).

internal/telemetry/metric_exporter.go
axw: Yes, under the assumption that all metrics are coming from an external agent. I'm not sure it makes sense here. These metrics won't show up in the APM app, so I think we can remove these fields.

internal/telemetry/metric_exporter.go
dmathieu: See https://github.com/elastic/apm-data/issues/144

internal/telemetry/metric_exporter_config.go
axw: @dmathieu explained on slack that we may want to use a single MeterProvider with two readers, in which case that wouldn't work.

internal/telemetry/metric_exporter.go
dmathieu: I have removed the agent. I've kept the service, as we'd still want to know where the metric comes from.

axw (approved):

internal/telemetry/metric_exporter.go
axw: :+1: thanks

dmathieu:
I have added the new exporter to the global meter provider (it won't do anything yet, since we don't have any metric being emitted).

Issue #7842, opened 2022-04-11
Collect APM Server throughput numbers


simitt:
The goal is to replace the ([current numbers](https://www.elastic.co/guide/en/apm/guide/current/processing-and-performance.html)) around processing and performance for APM Server. 

For different APM Server sizes (at least 1GB, 4GB, 8GB) define
* throughput baseline for medium sized events 
* expected amount of concurrent apm agents that can be handled
* expected size of ES

Ideally we also provide numbers for costly use cases, such as sourcemapping. 

This is the ground work for an updated sizing guide (https://github.com/elastic/apm-server/issues/7840).

simitt:
Updated the description to contain a concrete setup for the benchmarks.

dmathieu:
We are going to run those benchmarks in Jenkins, which allows us to trigger them in parallel and without the need to keep a laptop from sleeping.
The process, and results details are described in [this document](https://docs.google.com/document/d/1Vb-KzXZBJ9I01LOQAONru69SDSH_LDQjOenxyJ3FcXk/edit#heading=h.9muqipp9lkgq).

simitt:
Labeled this as blocked until after `8.9.0` has been released. The benchmarks should include the protobuf changes as they have a relevant impact on performance. Tests should be carried out in a stable environment, therefore prefering to run them on production. CFT regions do not exist for `aws` and `azure`, thus we'll need to wait until the new version is publicly released, to be able to test against all three cloud providers.

dmathieu:
This has been run and documented in the document linked in my previous comment above.
The default system profiles have been updated to match good values here: https://github.com/elastic/apm-server/pull/11600

Once that PR is merged, I believe this issue can be closed.

PR #11486, opened 2023-08-25
Customization for benchmarks manual runs

dmathieu:
## Motivation/summary

Since benchmarks moves to GH Actions, we lost some customization possibilities. This allows setting some attributes (the ones I need for #7842), while keeping the same defaults for the cronjobs.

## Checklist

- [x] **NA** Update [CHANGELOG.asciidoc](https://github.com/elastic/apm-server/blob/main/CHANGELOG.asciidoc)
- [x] **NA** Update [package changelog.yml](https://github.com/elastic/apm-server/blob/main/apmpackage/apm/changelog.yml) (only if changes to `apmpackage` have been made)
- [x] **NA** Documentation has been updated

## How to test these changes

This can be tested by manually running the workflow.
A test run: https://github.com/elastic/apm-server/actions/runs/5973256723/job/16205170670

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -7,6 +7,11 @@ on:
         description: 'The system profile used to run the benchmarks'
         required: false
         type: string
+      runOnStable:
+        description: 'Run the benchmarks on the latest stable version'
+        required: false
+        type: boolean
+        default: false
   schedule:
     - cron: '0 17 * * *'
 
@@ -85,10 +90,14 @@ jobs:
       - name: Build apmbench
         run: make apmbench $SSH_KEY terraform.tfvars
 
+      - name: Override docker committed version
+        if: ${{ !github.event.inputs.runOnStable }}
+        run: make docker-override-committed-version
+
       - name: Spin up benchmark environment
         id: deploy
         run: |
-          make docker-override-committed-version init apply
+          make init apply
           admin_console_url=$(terraform output -raw admin_console_url)
           echo "admin_console_url=$admin_console_url" >> "$GITHUB_OUTPUT"
           echo "-> infra setup done"
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -39,8 +39,6 @@ jobs:
       TF_VAR_ENVIRONMENT: ci
       TF_VAR_REPO: ${{ github.repository }}
       GOBENCH_TAGS: branch=${{ github.head_ref || github.ref }},commit=${{ github.sha }},target_branch=${{ github.base_ref }}
-
-      BENCHMARK_AGENTS: ${{ inputs.benchmarkAgents || 64 }}
     steps:
       - uses: actions/checkout@v3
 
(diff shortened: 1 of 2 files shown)


.github/workflows/benchmarks.yml
```diff
@@ -20,11 +34,13 @@ jobs:
...
       TF_VAR_private_key: ./id_rsa_terraform
       TF_VAR_public_key: ./id_rsa_terraform.pub
-      TFVARS_SOURCE: 'system-profiles/8GBx1zone.tfvars' # // Default to use an 8gb profile
+      TFVARS_SOURCE: ${{ inputs.profile || 'system-profiles/8GBx1zone.tfvars' }} # // Default to use an 8gb profile
       TF_VAR_BUILD_ID: ${{ github.run_id }}
       TF_VAR_ENVIRONMENT: ci
       TF_VAR_REPO: ${{ github.repository }}
       GOBENCH_TAGS: branch=${{ github.head_ref || github.ref }},commit=${{ github.sha }},target_branch=${{ github.base_ref }}
+
+      BENCHMARK_AGENTS: ${{ inputs.benchmarkAgents || 64 }}
```
carsonip, line 43: I think having a default here will interfere with the autotuning in `make run-benchmark-autotuned`

dmathieu: The benchmarks readme mentions 64 as being the default, so I assumed it would be fine. I've switched it to only be set if the value isn't empty.

carsonip: `BENCHMARK_AGENTS` defaults to 64 for `run-benchmark`, but we are using `run-benchmark-autotuned` which is essentially a wrapper over `run-benchmark` with a calculated `BENCHMARK_AGENTS`. See Makefile. Therefore `inputs.benchmarkAgents` will still be ignored in the current state of PR .

imo the right thing to do here to achieve what you want is that we `run-benchmark` instead of `run-benchmark-autotuned` when `inputs.benchmarkAgents` is set. Also good to have README updated to mention `run-benchmark-autotuned`

dmathieu: I've changed the action, and documented the command.

.github/workflows/benchmarks.yml
```diff
@@ -80,17 +98,26 @@ jobs:
...
           echo "admin_console_url=$admin_console_url" >> "$GITHUB_OUTPUT"
           echo "-> infra setup done"
 
-      - name: Run benchmarks
+      - name: Run benchmarks autotuned
+        if: ${{ !github.event.inputs.benchmarkAgents == '' }}
         run: make run-benchmark-autotuned index-benchmark-results
 
+      - name: Run benchmarks self tuned
+        if: ${{ !github.event.inputs.benchmarkAgents != '' }}
```
carsonip, line 118: same as above

.github/workflows/benchmarks.yml
```diff
@@ -80,17 +98,26 @@ jobs:
...
         run: |
-          make docker-override-committed-version init apply
+          make init apply
           admin_console_url=$(terraform output -raw admin_console_url)
           echo "admin_console_url=$admin_console_url" >> "$GITHUB_OUTPUT"
           echo "-> infra setup done"
 
-      - name: Run benchmarks
+      - name: Run benchmarks autotuned
+        if: ${{ !github.event.inputs.benchmarkAgents == '' }}
```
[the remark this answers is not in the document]
dmathieu: Good catch. It's a mistake on my end. I have removed it.

carsonip (approved):
lgtm! I believe this should work but would be great to test it with the new params after merge

dmathieu:
We can test it without a merge, running a workflow can pick any branch.
I have several runs triggered with various options: https://github.com/elastic/apm-server/actions/workflows/benchmarks.yml?query=branch%3Acustomize-benchmarks

Issue #11533, opened 2023-08-31
`queue.*` config is ignored for libbeat outputs


axw:
**APM Server version** (`apm-server version`): 8.6.0+

**Description of the problem including expected versus actual behavior**:

When using the Logstash output, configuring `queue.mem.events`, `queue.mem.flush.min_events`, and `output.logstash.bulk_max_size` has no effect. The default event batch size (2048), flush size (2048) and queue size (4096) are used.

`queue.*` configuration is ignored for libbeat outputs -- logstash, kafka, etc. This is not relevant for the Elasticsearch output, as the queue configuration is not applicable there.

This regression was introduced in 8.6.0 while we were refactoring beater. The `pipeline.Config{}` value on this line is the culprit (it should be a populated config): https://github.com/elastic/apm-server/blob/c4a31fac93f7c0aa247ccfbf5a7dbe891175e73d/internal/beater/beater.go#L847

**Steps to reproduce**:

 1. Run https://github.com/elastic/go-lumber/tree/main/cmd/tst-lj
 2. Run `apm-server  --strict.perms=false -E apm-server.data_streams.wait_for_integration=false -E queue.mem.events=350000 -E queue.mem.flush.min_events=10000 -E output.elasticsearch.enabled=false -E output.logstash.enabled=true -E output.logstash.bulk_max_size=10000 -E output.logstash.hosts=localhost:5044`
 2. Send at least 10000 events in 1s
 3. Observe that only 2048 events are sent at a time

PR #11534, opened 2023-08-31
Handle queue.* config for libbeat outputs

axw:
## Motivation/summary

Unpack and pass `queue.*` config to libbeat. We were ignoring `queue.*` config when using libbeat outputs, such as logstash and kafka.

## Checklist

## How to test these changes

See reproduction steps in https://github.com/elastic/apm-server/issues/11533

## Related issues

Closes https://github.com/elastic/apm-server/issues/11533

### DIFF
--- a/internal/beater/beater.go
+++ b/internal/beater/beater.go
@@ -844,7 +844,11 @@ func (s *Runner) newLibbeatFinalBatchProcessor(
 		output, err := outputs.Load(indexSupporter, beatInfo, stats, outputName, s.outputConfig.Config())
 		return outputName, output, err
 	}
-	pipeline, err := pipeline.Load(beatInfo, monitors, pipeline.Config{}, nopProcessingSupporter{}, outputFactory)
+	var pipelineConfig pipeline.Config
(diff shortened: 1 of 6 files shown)


lahsivjar (approved):

internal/beater/beater.go
```diff
@@ -844,7 +844,11 @@ func (s *Runner) newLibbeatFinalBatchProcessor(
 		output, err := outputs.Load(indexSupporter, beatInfo, stats, outputName, s.outputConfig.Config())
 		return outputName, output, err
 	}
-	pipeline, err := pipeline.Load(beatInfo, monitors, pipeline.Config{}, nopProcessingSupporter{}, outputFactory)
+	var pipelineConfig pipeline.Config
```
lahsivjar, line 847: As a follow-up probably, I wonder if it would be possible to have a test to assert that the full pipeline config is loaded correctly.

axw: Should be possible, but I didn't have the patience to figure it out right now - the system test should cover us for the moment. I've opened https://github.com/elastic/apm-server/issues/11535

PR #11558, opened 2023-08-31
fix: clone each event in the libbeat publisher processor

kruskall:
## Motivation/summary

This was okay when `model.Batch` was a slice of concrete types and the models had concrete types but now we have pointers everywhere.

## Checklist

## Related issues

Closes https://github.com/elastic/apm-server/issues/11563


carsonip:
Is this something that needs to be backported to 8.10?

simitt:
@kruskall there is no issue for this. Could you please always ensure to create a bug issue, and describe the consequences of the bug (in this case, just add that to the PR please).

carsonip (approved):

axw (approved):

PR #11560, opened 2023-08-31
[8.9] docs: Removes // from Node.js link

KOTungseth:
Changes `{apm-node-ref-v}//configuration.html#capture-body` to `{apm-node-ref-v}/configuration.html#capture-body`


bmorelli25 (approved):
I can approve, but unfortunately I don't have permission to merge.

PR #11589, opened 2023-09-05
Add 8.9.2 release notes

bmorelli25:
I see https://github.com/elastic/apm-server/pull/11534 was added to the 8.10 release notes, but it looks like this was backported to 8.9? Should this note be included in the 8.9.2 rns?


axw:
@bmorelli25 yes, we should add it here too.

Issue #11138, opened 2023-07-04
Adding `event.received` to APM events


endorama:
Recently we added support for `event.received` field in apm-data: https://github.com/elastic/apm-data/pull/85

As part of data ingestion one interesting element we want to keep track of is when a certain event has been received by the ingestion backend.

This timestamp is not user-controlled like @timestamp and can guarantee us a certain timestamp for event reception, the step before ingestion (which is registered as event.ingested).

We also added setting the value for this field to OTLP ingested events, but we are lacking the same for APM events.

We also need to update APM server integration package data streams to provide proper mapping for the new field.

PR #11602, opened 2023-09-07
Prevent indexing of `event.received` field

endorama:
## Motivation/summary

We added support for `event.received` field in https://github.com/elastic/apm-data/pull/85.

This field is useful to compute delay between event reception and ingestion, but is not currently used by APM UI.

To prevent ingesting unnecessary data, this PR adds a `remove` processor to ingest pipelines to prevent indexing `event.received`.

Data streams updated:
- `app_logs`
- `app_metrics`
- `error_logs`
- `internal_metrics`
- `rum_traces`
- `traces`

**Note for reviewers**: I'm not sure if this PR needs to be backported.

## Checklist

## Related issues

Related to https://github.com/elastic/apm-server/issues/11138

### DIFF
--- a/apmpackage/apm/changelog.yml
+++ b/apmpackage/apm/changelog.yml
@@ -12,6 +12,9 @@
     - description: Remove unused `processor.event` field from logs data streams
       type: enhancement
       link: https://github.com/elastic/apm-server/pull/11494
+    - description: Remove `event.received` fields from logs, metrics and traces data streams
+      type: bugfix
+      link: 
 - version: "8.10.0"
   changes:
     - description: Add permissions to reroute to dedicated datasets for logs, metrics and traces
--- a/apmpackage/apm/changelog.yml
+++ b/apmpackage/apm/changelog.yml
@@ -14,7 +14,7 @@
       link: https://github.com/elastic/apm-server/pull/11494
     - description: Remove `event.received` fields from logs, metrics and traces data streams
       type: bugfix
-      link: 
+      link: https://github.com/elastic/apm-server/pull/11602
 - version: "8.10.0"
   changes:
     - description: Add permissions to reroute to dedicated datasets for logs, metrics and traces
--- a/apmpackage/apm/changelog.yml
+++ b/apmpackage/apm/changelog.yml
@@ -13,7 +13,7 @@
       type: enhancement
       link: https://github.com/elastic/apm-server/pull/11494
     - description: Remove `event.received` fields from logs, metrics and traces data streams
-      type: bugfix
+      type: enhancement
       link: https://github.com/elastic/apm-server/pull/11602
 - version: "8.10.0"
   changes:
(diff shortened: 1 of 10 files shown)


apmpackage/apm/data_stream/app_logs/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -9,3 +9,8 @@ processors:
       name: observer_ids
   - pipeline:
       name: ecs_version
+  - remove:
```
carsonip, line 12: Even better, define a pipeline like the one in https://github.com/elastic/apm-server/blob/main/apmpackage/cmd/genpackage/pipelines.go#L155 so that it can be reused in multiple ingest pipelines.

carsonip (commented):
1. Could you also check whether we need this for aggregated metrics data streams?
2. Remove https://github.com/elastic/apm-server/blob/8.9/internal/beater/processors.go#L89 and its usage so that we can confirm the ingest pipeline works.
3. System test to ensure the field is removed if possible

endorama:
> Could you also check whether we need this for aggregated metrics data streams?

Checked and I don't think we need to. 

> Remove [8.9/internal/beater/processors.go#L89](https://github.com/elastic/apm-server/blob/8.9/internal/beater/processors.go?rgh-link-date=2023-09-07T12%3A57%3A15Z#L89) and its usage so that we can confirm the ingest pipeline works.

Done in [8ac04bd](https://github.com/elastic/apm-server/pull/11602/commits/8ac04bdc9f8da3d61810e34e2dcfcd240e41ff6a), thanks for pointing that out!

> System test to ensure the field is removed if possible

I reviewed the system tests and this is already tested. When removing the `removeEventReceivedBatchProcessor` processor a series of tests fail:
- `TestJaeger`
- `TestJaeger/batch_0`
- `TestJaeger/batch_1`
- `TestOTLPGRPCTraces`
- `TestOTLPGRPCMetrics`
- `TestOTLPGRPCLogs`

The same tests pass when adding the ingest pipeline. Also is my understanding that as https://github.com/elastic/apm-server/issues/11138 has not been completed APM events don't have `event.received`.

apmpackage/apm/changelog.yml
```diff
@@ -12,6 +12,9 @@
     - description: Remove unused `processor.event` field from logs data streams
       type: enhancement
       link: https://github.com/elastic/apm-server/pull/11494
+    - description: Remove `event.received` fields from logs, metrics and traces data streams
+      type: bugfix
```
carsonip, line 16:
```suggestion
      type: enhancement
```

There's no bug in apmpackage (mapping and ingest pipeline) itself

endorama: thanks, followed up in [cee297b](https://github.com/elastic/apm-server/pull/11602/commits/cee297b83c3485d41e51e86f24c714f89af522fd)

endorama:
After internal discussion we prefer keeping the current approach instead of moving to an ingest pipeline. I'm closing this PR as is no longer relevant.

Issue #11266, opened 2023-07-24
Change instance specific dimensions in aggregated metrics


simitt:
Currently, instance specific metrics are part of the transaction metrics dimensions, and can lead to time series explosion.

To reduce the number of transaction metric timeseries, we remove the instance-specific dimensions and introduce a new metric set that has instance specific dimensions but not `transaction.name`.

Required changes:
- review dimensions and propose which ones classify as instance-specific
- remove the instance dimensions from transaction metrics
- create a new metric set with instance dimensions
- update documentation
- update apmpackage

axw:
I think all the `faas.*` fields should be on transaction metrics only. @AlexanderWert would you agree with that?

I believe we still need `transaction.type` on the instance metrics, as the transaction type dropdown on the service overview page of the UI affects the transaction metrics of the Instances table:

![image](https://github.com/elastic/apm-server/assets/843579/c77726c5-02ab-426e-972b-d14970192f5f)

Given this, I'd say the name ought to be `service_instance_transaction` (`service_node_transaction`?) metrics; they're essentially the same as `service_transaction`, but with instance/node dimensions.

> EventOutcome exists as a dimension in both keys. Does it sound right?

If we consider the new metrics as being the same as `service_transaction` but with additional instance dimensions, then I think it would make sense to instead set `event.success_count` and leave `event.outcome` out of the dimensions. We would continue to have `event.outcome` as a dimension for transaction metrics.

AlexanderWert:
> I think all the `faas.*` fields should be on transaction metrics only. @AlexanderWert would you agree with that?

Yes, that's true.

dgieselaar:
I remember there were some oddities with FAAS data where some of it was handled differently than regular transactions. I don't immediately remember what it was, but maybe @AlexanderWert knows? The reason why I'm bringing that up is _if_ we want to align FAAS data with regular transaction data, maybe this is an opportunity to do so?

simitt:
Shouldn't the `Service*` fields be a dimension on both metrics?

AlexanderWert:
> I remember there were some oddities with FAAS data where some of it was handled differently than regular transactions. I don't immediately remember what it was, but maybe @AlexanderWert knows? The reason why I'm bringing that up is if we want to align FAAS data with regular transaction data, maybe this is an opportunity to do so?

[...]

PR #11614, opened 2023-09-08
feat: add service instance transaction

kruskall:
## Motivation/summary

Bump apm-aggregation and add service instance transaction metrics.

The fields were already decided in previous PRs so there's nothing new here.

The `RemoveEventReceived` processor was moved later down the middleware chain to actually remove the field: the new version of apm-aggregation is adding an `event.received` field and we need to propagate the field to the aggregation processor.
This means we should remove the field after the aggregation is complete but before sending to ES.

An option was added to the aggregation config in apm-server with the same default option based on available memory as the other aggregations.

A system-test has been added to ensure the feature is working as intended.

## Checklist

## Related issues

Closes https://github.com/elastic/apm-server/issues/11266

### DIFF
--- a/NOTICE.txt
+++ b/NOTICE.txt
@@ -573,11 +573,11 @@ Contents of probable licence file $GOMODCACHE/github.com/elastic/apm-data@v0.1.1
 
 --------------------------------------------------------------------------------
 Dependency : github.com/elastic/beats/v7
-Version: v7.0.0-alpha2.0.20230912134242-8e7e2065c73b
+Version: v7.0.0-alpha2.0.20230913143010-b7d1f806f6ec
 Licence type (autodetected): Elastic
 --------------------------------------------------------------------------------
 
-Contents of probable licence file $GOMODCACHE/github.com/elastic/beats/v7@v7.0.0-alpha2.0.20230912134242-8e7e2065c73b/LICENSE.txt:
+Contents of probable licence file $GOMODCACHE/github.com/elastic/beats/v7@v7.0.0-alpha2.0.20230913143010-b7d1f806f6ec/LICENSE.txt:
 
 Source code in this repository is variously licensed under the Apache License
 Version 2.0, an Apache compatible license, or the Elastic License. Outside of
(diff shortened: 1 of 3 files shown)


simitt (commented):
Please also adapt the docs 
- https://www.elastic.co/guide/en/apm/guide/current/apm-data-streams.html#apm-data-streams-list
- https://www.elastic.co/guide/en/apm/guide/current/data-model-metrics.html#_calculated_metrics

systemtest/approvals/TestServiceDestinationAggregation.approved.json
```diff
@@ -17,7 +17,8 @@
             "event": {
                 "agent_id_status": "missing",
                 "ingested": "dynamic",
-                "outcome": "success"
+                "outcome": "success",
+                "received": "dynamic"
```
carsonip, line 21: Why is event.received still present? Isn't apm-data updated to remove the field?

carsonip (commented):
Also, could you please update the PR description? I'll review the PR again when the docs are updated.

internal/beater/beater.go
```diff
@@ -447,6 +454,11 @@ func (s *Runner) Run(ctx context.Context) error {
 		modelprocessor.NewDropUnsampled(false /* don't drop RUM unsampled transactions*/, func(i int64) {
 			transactionsDroppedCounter.Add(i)
 		}),
+
+		// Add a model processor that removes `event.received`, which is added by
+		// apm-data, but which we don't yet map.
```
carsonip, line 459: Outdated comment. Instead of "we don't yet map", we intentionally do not map it now.

What about also mentioning that it has to be after aggregation because aggregation adds it as well?

carsonip (commented):
missing changelog for apm-server and apmpackage; otherwise looks good

internal/beater/beater.go
```diff
@@ -447,6 +447,11 @@ func (s *Runner) Run(ctx context.Context) error {
...
+		modelprocessor.RemoveEventReceived{},
```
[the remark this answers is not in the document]
carsonip: Before this PR, apm-aggregation was on a version that does not add event.received. Now we are ensuring event.received added by apm-aggregation is removed as well.

simitt:
Unfortunately this cannot be merged into `8.11` development, see https://github.com/elastic/apm-server/issues/11266#issuecomment-1717847803

PR #11621, opened 2023-09-11
Create new dev/test certificate for fleet-server

axw:
The old one expired, create a new one with a 10 year expiry. Generated with:

```
openssl req -x509 -newkey rsa:4096 -keyout key.pem -out certificate.pem
-sha256 -days 3650 -nodes -subj "/CN=fleet-server" -addext
"subjectAltName=DNS:fleet-server,DNS:localhost"
```

I've removed the separate `ca.pem`, the new certificate is self-signed.


axw:
> The output for this was a bit obscure in the CI results. How about a check at the beginning of systemtests to fail quickly if the certificate is expired? (I guess that can come in another PR)

The new expiry time is 10 years, so I think it may be unnecessary, but if a similar issue comes up then adding some pre-flight checks sounds reasonable. We could maybe also improve the logging of fleet-server.

PR #11632, opened 2023-09-12
Remove `ecs.version` field from mappings

axw:
## Motivation/summary

Remove the `ecs.version` field mapping from APM data streams. We don't use this field at all in APM UI, and its presence works against converging on OTel SemConv, and decoupling the index templates from apm-server.

## Checklist

## How to test these changes

1. Ingest some APM data (e.g. with `apmtool`)
2. Verify the `ecs.version` field is not present in any APM docs
3. Verify basic APM UI functionality still works

## Related issues

https://github.com/elastic/elasticsearch/pull/97546


apmpackage/cmd/genpackage/pipelines.go
```diff
@@ -103,7 +103,7 @@ var observerIDsPipeline = []map[string]interface{}{{
 	},
 }}
 
-var ecsVersionPipeline = []map[string]interface{}{{
+var removeECSVersionPipeline = []map[string]interface{}{{
```
lahsivjar, line 106: Out of curiosity, even the previous `ecsVersionPipeline` was removing the field `ecs.version`. Does this mean we were never indexing `ecs.version` so it was not used in the mappings anyway?

lahsivjar (approved):

apmpackage/cmd/genpackage/pipelines.go
axw: The ingest pipeline was there to remove the field from documents written by older versions of apm-server. The field would get added back into documents as a `constant_keyword` field.

I've left the removal there for now, but I don't think we'll keep that in https://github.com/elastic/elasticsearch/pull/97546. We'll just dynamically map the field for older versions of apm-server - no big deal.

lahsivjar:
Tested successfully that `ecs.version` field is not present in any APM docs.

Screenshot 2023-10-24 at 1 12 17 PM
^ All docs with `ecs.version` not present.

Screenshot 2023-10-24 at 1 12 27 PM
^ All docs with `ecs.version` present.

PR #11638, opened 2023-09-13
Minor changes to support running system tests with x-pack-apm

axw:
## Motivation/summary

These are some of the changes needed for the system tests to pass when using the x-pack-apm ES plugin templates and pipelines (https://github.com/elastic/elasticsearch/pull/97546), rather than the Fleet integration package ones. There will be more changes needed, which we'll add later when we remove the data streams from the integration package.

## Checklist

## How to test these changes

N/A

## Related issues

https://github.com/elastic/elasticsearch/pull/97546

### DIFF
--- a/systemtest/errorgrouping_test.go
+++ b/systemtest/errorgrouping_test.go
@@ -22,7 +22,7 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
-	"github.com/tidwall/gjson"
+	"github.com/stretchr/testify/require"
 	"go.elastic.co/apm/v2"
 
 	"github.com/elastic/apm-server/systemtest"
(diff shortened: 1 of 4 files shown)


systemtest/errorgrouping_test.go
```diff
@@ -48,7 +48,9 @@ func TestErrorGroupingName(t *testing.T) {
 
 	var names []string
 	for _, hit := range result.Hits.Hits {
-		names = append(names, gjson.GetBytes(hit.RawSource, "error.grouping_name").String())
+		values := hit.Fields["error.grouping_name"]
```
axw, line 51: Note for reviewers: I made this change because in the Elasticserach plugin, we are using a script field which is evaluated at index time. It ends up in `fields`, but not in `_source`.

systemtest/ingest_test.go
```diff
@@ -123,10 +123,11 @@ func TestIngestPipelineEventDuration(t *testing.T) {
 	}
 
 	tests := []test{{
-		// No transaction.* field, no update.
-		source: `{"@timestamp": "2022-02-15", "observer": {"version": "8.2.0"}, "processor": {"event": "transaction"}}`,
+		// No transaction.* field, no event.duration: set transaction.duration.us to zero.
+		source:                        `{"@timestamp": "2022-02-15", "observer": {"version": "8.2.0"}, "processor": {"event": "transaction"}}`,
+		expectedTransactionDurationUS: 0.0,
```
axw, line 128: Note for reviewers: there's a small functional change here, which won't actually ever manifest in reality - transactions and spans always have a `transaction` or `span` object defined.

dmathieu (approved):

PR #11642, opened 2023-09-13
Fix tail-based sampling discarding low throughput and low sample rate traces

### LINKED ISSUE #11372: Tail-based sampling discards all traces when sampling rate and throughput are low
**APM Server version** (`apm-server version`): 8.9.0

Let's say we have a service that produces 4 transactions per minute, and we have configured tail-based sampling with a rate of 0.1. Roughly speaking, we calculate the desired number of traces as `sample_rate * ingest_rate`, rounding half away from zero. In this case, that would be 4*0.1=0.4 -- so the desired total number of traces is calculated as zero.

Even at low throughput we should be using random sampling. In the above scenario there should be a 10% chance of capturing any one trace. That may still lead to infrequently captured traces -- users can then address this by adding more specific sampling policies with higher sampling rates for their low-throughput services.

**Steps to reproduce**:

 1. Send 4 transactions every minute to apm-server
 2. Configure TBS with a single policy with a sample rate of 0.1
 3. Observe that no trace events are even indexed

Sample event generator:

```go
package main

import (
        "time"

        "go.elastic.co/apm/v2"
)

func main() {
        tracer := apm.DefaultTracer()
        defer tracer.Flush(nil)

        ticker := time.NewTicker(time.Minute)
        for range ticker.C {
                for i := 0; i < 4; i++ {
                        tx.Duration = time.Duration(i+1) * time.Second
                        tx.End()
                }
        }
}
```

carsonip:
## Motivation/summary

In tail-based sampling, use math.ceil instead of math.Round when computing desired total such that on low throughput and low sample rate traces e.g. throughput=1, sample rate = 0.4, desired total = math.Ceil(1 * 0.4) = 1 as opposed to 0 when using math.Round. This will fix cases where these traces were missing before the fix.

## Checklist

## How to test these changes

Enable TBS, set sample rate to 0.4, send 1 trace. Check if the trace shows up.

## Related issues

Closes #11372

### DIFF
--- a/x-pack/apm-server/sampling/groups.go
+++ b/x-pack/apm-server/sampling/groups.go
@@ -224,7 +224,7 @@ func (g *traceGroup) finalizeSampledTraces(traceIDs []string, ingestRateDecayFac
 
 	// Resize the reservoir, so that it can hold the desired fraction of
 	// the observed ingest rate.
-	newReservoirSize := int(math.Round(g.samplingFraction * g.ingestRate))
+	newReservoirSize := int(math.Ceil(g.samplingFraction * g.ingestRate))
 	if newReservoirSize < minReservoirSize {
 		newReservoirSize = minReservoirSize
 	}
--- a/x-pack/apm-server/sampling/groups.go
+++ b/x-pack/apm-server/sampling/groups.go
@@ -211,7 +211,7 @@ func (g *traceGroup) finalizeSampledTraces(traceIDs []string, ingestRateDecayFac
 		g.ingestRate *= 1 - ingestRateDecayFactor
 		g.ingestRate += ingestRateDecayFactor * float64(g.total)
 	}
-	desiredTotal := int(math.Round(g.samplingFraction * float64(g.total)))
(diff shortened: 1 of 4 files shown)


x-pack/apm-server/sampling/groups.go
```diff
@@ -224,7 +224,7 @@ func (g *traceGroup) finalizeSampledTraces(traceIDs []string, ingestRateDecayFac
 
 	// Resize the reservoir, so that it can hold the desired fraction of
 	// the observed ingest rate.
-	newReservoirSize := int(math.Round(g.samplingFraction * g.ingestRate))
+	newReservoirSize := int(math.Ceil(g.samplingFraction * g.ingestRate))
```
carsonip, line 227: [For reviewer] I don't think this is needed for the fix to work, and that we have a min size. But might be good to be consistent and use math.Ceil in all samplingFraction multiplication.

axw (approved):
LGTM, thank you

lahsivjar:
Tested successfully by enabling TBS with a low sample rate and sending a single trace, the trace was indexed successfully.

PR #11660, opened 2023-09-18
Bump to 8.10.2

marclop:


apmpackage/apm/changelog.yml
```diff
@@ -3,6 +3,11 @@
     - description: Placeholder
       type: enhancement
       link: https://github.com/elastic/apm-server/pull/123
+- version: "8.10.1"
```
carsonip, line 6: I think we can just skip the entry if there's nothing interesting.

marclop: Fixed

carsonip (approved):

PR #11690, opened 2023-09-21
Bump up version to prepare for 8.10.3

lahsivjar:


marclop (commented):
you need to also update the apmpackage changelog to contain `8.10.3`.

lahsivjar:
> you need to also update the apmpackage changelog to contain 8.10.3.

@marclop There are no changes to the apmpackage I think. I was under the impression that in such cases we don't make any changes.

marclop:
I agree with you, but the `make check-full` target will fail and tell you due to the version bump.

marclop:
oh never mind... Seems I was mistaken, ignore my comment

dmathieu (approved):

PR #11699, opened 2023-09-22
Adding geolocation attrs to app_logs

LikeTheSalad:
## Motivation/summary

We need geolocation information on mobile events in order to properly display location-related charts in Kibana's mobile dashboard.

These changes aim to add the client.geo* attrs to logs that contain the attr `client.ip`, which at the moment is only set to mobile-related logs.

## Checklist


apmpackage/apm/data_stream/app_logs/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -9,3 +9,14 @@ processors:
...
       name: remove_ecs_version
+  - geoip:
+      if: ctx.event?.category == 'device'
+      field: client.ip
+      target_field: client.geo
+      ignore_missing: true
+      database_file: GeoLite2-City.mmdb
+      on_failure:
+        - remove:
+            field: client.ip
```
carsonip, line 20: Do we have to remove the field on failure?

carsonip (commented):
- A larger problem now that I've checked apm-data is that we are never parsing/populating client.ip for logs, but only for errors. This will need to be solved before this PR works.
- It will also be helpful to have a system test so that this doesn't break in the future.

apmpackage/apm/data_stream/app_logs/elasticsearch/ingest_pipeline/default.yml
LikeTheSalad: Thanks for taking a look Carson!

Regarding the first item, I'm a bit confused because I'm taking a look at [this log](https://edge-lite-oblt.kb.us-west2.gcp.elastic-cloud.com/app/discover#/doc/apm_static_index_pattern_id/.ds-logs-apm.app.opbeans_android-default-2023.09.24-000012?id=nRUey4oBpy6-VGJ8ASZU) in `edge-lite` that's not an error, which has the `client.ip` attribute. That's the kind of log I'm trying to append geo attrs with these changes so it seems like we might be able to work with those?

> It will also be helpful to have a system test so that this doesn't break in the future.

I'm up for adding those kinds of tests, although I'm struggling a bit with finding similar tests (that check log's enhancements) in the [current tests](https://github.com/elastic/apm-server/blob/main/systemtest/ingest_test.go) for ingest processes, and without some similar tests to take as a base for mine, I'm not sure how to set things up on my own in order to do the logs validations. Is there any other test file I'm missing where I could find some ingest tests that verify attrs appended to logs? Otherwise, I'm also up to jump on a call to discuss what steps I can follow to set up these kind of tests for logs.

LikeTheSalad:
Thanks for taking a look @carsonip !

> A larger problem now that I've checked apm-data is that we are never parsing/populating client.ip for logs, but only for errors. This will need to be solved before this PR works.

I'm a bit confused about this because I'm taking a look at [this log](https://edge-lite-oblt.kb.us-west2.gcp.elastic-cloud.com/app/discover#/doc/apm_static_index_pattern_id/.ds-logs-apm.app.opbeans_android-default-2023.09.24-000012?id=nRUey4oBpy6-VGJ8ASZU) in edge-lite that's not an error, which has the client.ip attribute. That's the kind of log I'm trying to append geo attrs with these changes so it seems like we might be able to work with those?

> It will also be helpful to have a system test so that this doesn't break in the future.

I'm up for adding those kinds of tests, although I'm struggling a bit with finding similar tests (that check log's enhancements) in the [current tests](https://github.com/elastic/apm-server/blob/main/systemtest/ingest_test.go) for ingest processes, and without some similar tests to take as a base for mine, I'm not sure how to set things up on my own in order to do the logs validations. Is there any other test file I'm missing where I could find some ingest tests that verify attrs appended to logs? Otherwise, I'm also up to jump on a call to discuss what steps I can follow to set up these kind of tests for logs.

carsonip:
> That's the kind of log I'm trying to append geo attrs with these changes so it seems like we might be able to work with those?

You're right if the apm-server is appropriately configured. When the `capture_personal_data` is set in apm-server, Client.IP from OTLP will be parsed and recorded. Here's the relevant line of code: https://github.com/elastic/apm-server/blob/main/internal/beater/server.go#L198

I believe this is also why you're having difficulties finding such a test, because it is 1. requires a config, 2. uses otlp and only works on swift and android, 3. needs to be a system test. I did a quick search and [otlp_test.go](https://github.com/elastic/apm-server/blob/main/systemtest/otlp_test.go) would be a good starting point. Think of a mix of TestOTLPAnonymous (the part where we override agent name), TestOTLPGRPCLogs (how we generate logs and assert them).

LikeTheSalad:
> You're right if the apm-server is appropriately configured. When the capture_personal_data is set in apm-server, Client.IP from OTLP will be parsed and recorded. Here's the relevant line of code: https://github.com/elastic/apm-server/blob/main/internal/beater/server.go#L198
> 
> I believe this is also why you're having difficulties finding such a test, because it is 1. requires a config, 2. uses otlp and only works on swift and android, 3. needs to be a system test. I did a quick search and [otlp_test.go](https://github.com/elastic/apm-server/blob/main/systemtest/otlp_test.go) would be a good starting point. Think of a mix of TestOTLPAnonymous (the part where we override agent name), TestOTLPGRPCLogs (how we generate logs and assert them).

Cool! Thanks for taking the time 👍 I'll take a look at those tests to try and create some for this use case then

carsonip:
Also remember to set `capture_personal_data` in config before `srv.Start()` like what we did in TestOTLPAnonymous. We never use this setting in system tests so there's no existing field in `systemtest/apmservertest/config.go`. Under Config please add something like 
```
AugmentEnabled                       *bool      `json:"apm-server.capture_personal_data,omitempty"`
```

apmpackage/apm/data_stream/app_logs/elasticsearch/ingest_pipeline/default.yml
LikeTheSalad: I don't see why we should tbh, although it seems like the same is done in other places when using the `geoip` processor which made me think that there might be a reason for it, so I just left it to keep consistency with all other usages, but I'm ok to remove this part if it's not needed.

LikeTheSalad: I'm fine with using that predefined pipeline, however, doing so would append geo attrs to **all** logs, instead of only the ones with the attribute `event.category` set to `device`. I'm not sure is that could cause problems to the apm server.

AlexanderWert: @LikeTheSalad you can execute the pipeline conditionally as well:

```
- pipeline: 
      if: ...
      name: client_geoip
```

carsonip: I've just discussed this with Silvia. We think that unless there's a specific reason, we'd prefer running the pipeline without the `if` check. Since we are already running client_geoip ingest pipeline for most of the other data streams, the fact that it is missing in this app_logs data stream appears to be only an oversight. Also, client.ip is only populated for swift / android otel agents AND `capture_personal_data: true`, so it should be fine to run client_geoip whenever client.ip exists.

LikeTheSalad: Cool! That makes things simpler, cheers!

apmpackage/apm/changelog.yml
```diff
@@ -1,5 +1,8 @@
 - version: generated
   changes:
+    - description: Adding client.geo* attrs to apm logs
```
carsonip, line 3:
```suggestion
    - description: Add geoip processing to add client.geo.* fields for app_logs
```

systemtest/ingest_test.go
```diff
@@ -300,3 +307,60 @@ func TestIngestPipelineBackwardCompatibility(t *testing.T) {
 		require.NoError(t, err)
 	}
 }
+
+func TestMobileLogsWithGeoLocation(t *testing.T) {
+	systemtest.CleanupElasticsearch(t)
+	srv := apmservertest.NewUnstartedServerTB(t)
+	err := srv.Start()
```
carsonip, line 314: hmm why does this work without the capture_personal_data config?

apmpackage/apm/data_stream/app_logs/elasticsearch/ingest_pipeline/default.yml
LikeTheSalad: I finished creating the system test for this use case and I also applied the changes to use the existing `client_geoip` pipeline without conditions.

I found out a couple of things while I was working on the test:

* The `capture_personal_data` seems to be enabled by default (at least for the system tests), because when I was adding a new config inside `systemtest/apmservertest/config.go`, I printed some logs [in here](https://github.com/elastic/apm-server/blob/main/internal/beater/server.go#L195) where we're reading the config and realized that the `if` body was always getting executed no matter if I added the new config or not.
* There were 2 reasons why I wasn't able to see geo attributes locally: The first one is that the `agent.name` has to be either `android/java` or `iOS/swift` for the `client.ip` to be populated (as mentioned by @carsonip), and the second one was that the client.ip value received by the server during the tests was `127.0.0.1` which the server doesn't know how to translate into a location (which makes sense), so I had to override the local IP during testing by setting a real one using the `X-Forwarded-For` header so that the server could find a location from the client.ip attr.

systemtest/ingest_test.go
LikeTheSalad: It seems like the capture_personal_data flag is enabled by default (at least for tests). I've added more info on it in [this other comment](https://github.com/elastic/apm-server/pull/11699#discussion_r1336089308).

apmpackage/apm/data_stream/app_logs/elasticsearch/ingest_pipeline/default.yml
carsonip: - You're right. capture_personal_data is default to true for both apm-server standalone and integration.
- Overriding X-Forwarded-For sounds good.

systemtest/ingest_test.go
```diff
@@ -300,3 +307,60 @@ func TestIngestPipelineBackwardCompatibility(t *testing.T) {
 		require.NoError(t, err)
 	}
 }
+
+func TestMobileLogsWithGeoLocation(t *testing.T) {
```
carsonip, line 311: nit: imo the test fits marginally better in otlp_test.go although this specific PR is related to ingest pipeline. This is just a special case of TestOTLPGRPCLogs. Maybe call it `TestOTLPGRPCLogsClientIP`.

carsonip (commented):
One nit, otherwise looks good!

systemtest/ingest_test.go
LikeTheSalad: You mean to rename it and also move it to `otlp_test.go` or to just rename it?

carsonip: move + rename

carsonip (approved):
Looks perfect! thanks for adding the missing test

carsonip (approved):

dmathieu (approved):

axw:
Worked for me. I added `android/java` to allowed anonymous agents, and used this code:

```go
package main

import (
        "context"
        "crypto/tls"
        "log"
        "time"

        "go.opentelemetry.io/collector/pdata/pcommon"
        "go.opentelemetry.io/collector/pdata/plog"
        "go.opentelemetry.io/collector/pdata/plog/plogotlp"
        semconv "go.opentelemetry.io/collector/semconv/v1.5.0"
        "google.golang.org/grpc"
        "google.golang.org/grpc/credentials"
)

func main() {
        conn, err := grpc.Dial(
                "cc6895f8c08348a4ab785f00f15e971b.apm.us-west2.gcp.elastic-cloud.com:443",
                grpc.WithBlock(),
                grpc.WithTransportCredentials(credentials.NewTLS(&tls.Config{})),
        )
        if err != nil {
                log.Fatal(err)
        }
        defer conn.Close()

        logsClient := plogotlp.NewGRPCClient(conn)

        logs := plog.NewLogs()
        resourceLogs := logs.ResourceLogs().AppendEmpty()
        resourceAttrs := logs.ResourceLogs().At(0).Resource().Attributes()
        resourceAttrs.PutStr(semconv.AttributeTelemetrySDKLanguage, "java")
        resourceAttrs.PutStr(semconv.AttributeTelemetrySDKName, "android")

        scopeLogs := resourceLogs.ScopeLogs().AppendEmpty()
        otelLog := scopeLogs.LogRecords().AppendEmpty()
        otelLog.SetTraceID(pcommon.TraceID{1})
        otelLog.SetSpanID(pcommon.SpanID{2})
        otelLog.SetSeverityNumber(plog.SeverityNumberInfo)
        otelLog.SetSeverityText("Info")
        otelLog.SetTimestamp(pcommon.NewTimestampFromTime(time.Unix(1, 0)))
        otelLog.Attributes().PutStr("key", "value")
        otelLog.Attributes().PutDouble("numeric_key", 1234)
        otelLog.Body().SetStr("a log message")

        _, err = logsClient.Export(context.Background(), plogotlp.NewExportRequestFromLogs(logs))
        if err != nil {
                log.Fatal(err)
        }
}
```

axw:
@kruskall in discover, make sure you're looking back far enough. The test program creates a log in 1970.

image

PR #11739, opened 2023-09-28
Scale service destination based on available memory

### LINKED ISSUE #11721: Scale max service destination and max per service service destination aggregations based on available memory
Currently, we scale transactions and service transaction groups based on available memory ([ref](https://github.com/elastic/apm-server/blob/c220ed40b6bed913b7d92ed4be10e416aa449e6e/internal/beater/beater.go#L232-L251)). However, we don't do the same for service destination aggregations. The max service destination aggregations default to a constant size of 10_000 ([ref](https://github.com/elastic/apm-server/blob/c220ed40b6bed913b7d92ed4be10e416aa449e6e/internal/beater/config/aggregation.go#L20-L22)).

lahsivjar:
Closes #11721

### DIFF
--- a/internal/beater/beater.go
+++ b/internal/beater/beater.go
@@ -230,23 +230,34 @@ func (s *Runner) Run(ctx context.Context) error {
 	}
 
 	if s.config.Aggregation.MaxServices <= 0 {
-		s.config.Aggregation.MaxServices = maxGroupsForAggregation(memLimitGB)
+		// scale based on available memory considering 1K groups for 1GB
+		s.config.Aggregation.MaxServices = linearScaledValue(1_000, memLimitGB)
 		s.logger.Infof("Aggregation.MaxServices set to %d based on %0.1fgb of memory",
 			s.config.Aggregation.MaxServices, memLimitGB,
 		)
 	}
 
+	if s.config.Aggregation.ServiceTransactions.MaxGroups <= 0 {
+		// scale based on available memory considering 1K groups for 1GB
+		s.config.Aggregation.ServiceTransactions.MaxGroups = linearScaledValue(1_000, memLimitGB)
+		s.logger.Infof("Aggregation.ServiceTransactions.MaxGroups for service aggregation set to %d based on %0.1fgb of memory",
+			s.config.Aggregation.ServiceTransactions.MaxGroups, memLimitGB,
+		)
+	}
+
 	if s.config.Aggregation.Transactions.MaxGroups <= 0 {
-		s.config.Aggregation.Transactions.MaxGroups = maxTxGroupsForAggregation(memLimitGB)
+		// scale based on available memory considering 5K groups for 1GB
+		s.config.Aggregation.Transactions.MaxGroups = linearScaledValue(5_000, memLimitGB)
 		s.logger.Infof("Aggregation.Transactions.MaxGroups set to %d based on %0.1fgb of memory",
 			s.config.Aggregation.Transactions.MaxGroups, memLimitGB,
 		)
 	}
 
-	if s.config.Aggregation.ServiceTransactions.MaxGroups <= 0 {
-		s.config.Aggregation.ServiceTransactions.MaxGroups = maxGroupsForAggregation(memLimitGB)
-		s.logger.Infof("Aggregation.ServiceTransactions.MaxGroups for service aggregation set to %d based on %0.1fgb of memory",
-			s.config.Aggregation.ServiceTransactions.MaxGroups, memLimitGB,
+	if s.config.Aggregation.ServiceDestinations.MaxGroups <= 0 {
+		// scale based on available memory considering 5K groups for 1GB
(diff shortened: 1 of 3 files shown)


internal/beater/beater.go
```diff
@@ -230,23 +230,34 @@ func (s *Runner) Run(ctx context.Context) error {
...
 		)
 	}
 
-	if s.config.Aggregation.ServiceTransactions.MaxGroups <= 0 {
-		s.config.Aggregation.ServiceTransactions.MaxGroups = maxGroupsForAggregation(memLimitGB)
-		s.logger.Infof("Aggregation.ServiceTransactions.MaxGroups for service aggregation set to %d based on %0.1fgb of memory",
-			s.config.Aggregation.ServiceTransactions.MaxGroups, memLimitGB,
+	if s.config.Aggregation.ServiceDestinations.MaxGroups <= 0 {
+		// scale based on available memory considering 5K groups for 1GB
+		s.config.Aggregation.ServiceDestinations.MaxGroups = linearScaledValue(5_000, memLimitGB)
```
lahsivjar, line 258: [**For reviewers**] Not sure what is a good value here, previous default was 10K for all, now we will have much greater values. However, service destination is not very costly so I kept it 5k/GB. Let me know if others have any concerns.

carsonip: question: if it was a constant 10k for all, do we consider this a breaking change for 1GB users? do we accept the risk?

simitt: Without indicators that current limits were too high, I'd keep the 10k for 1GB, and then afterwards start 5k steps per GB.

carsonip (commented):
- this change may be breaking for 1GB
- please also update `docs/data-model.asciidoc`

internal/beater/beater.go
carsonip: service destination doesn't use histograms and only uses 2 float64, which makes it very light on memory. +1 on keeping 10k for 1GB.

lahsivjar: Starting with 10K for 1GB sounds good to me. One potential drawback can be that if we were to introduce summary or histograms in the future in the service destination metrics then we might have to introduce a breaking change to reduce the limits but probably better to do a breaking change then than now.

> afterwards start 5k steps per GB.

Should we make it 10K per GB instead or will that be too high for bigger APM servers?

carsonip: > Should we make it 10K per GB instead or will that be too high for bigger APM servers?

Wouldn't be a problem now as there are no histograms. But as you've said, it we were to introduce histograms, it will be 2x the other limits and require us to make a breaking change to reduce the limits, but we will then have a pressing reason to do so. So I'm fine with 10k per GB.

lahsivjar: Updated code to use 10k for 1GB and then 5k per GB for svc destination limit

internal/beater/beater.go
```diff
@@ -568,26 +575,14 @@ func maxConcurrentDecoders(memLimitGB float64) uint {
...
-	return int(memLimitGB * 1_000)
-}
-
-// maxTxGroupsForAggregation calculates the maximum transaction groups that a
-// particular memory limit can have. This will be scaled linearly for bigger
-// instances.
-func maxTxGroupsForAggregation(memLimitGB float64) int {
+// linearScaledValue calculates linearly scaled value based on memory limit using
+// the formula y = (perGBIncrement * memLimitGB) + c
+func linearScaledValue(perGBIncrement, memLimitGB, c float64) int {
```
carsonip, line 580:
```suggestion
// linearScaledValue calculates linearly scaled value based on memory limit using
// the formula y = (perGBIncrement * memLimitGB) + constant
func linearScaledValue(perGBIncrement, memLimitGB, constant float64) int {
```

a more descriptive param name in the signature would be useful!

internal/beater/beater.go
```diff
@@ -568,26 +575,14 @@ func maxConcurrentDecoders(memLimitGB float64) uint {
...
 	const maxMemGB = 64
 	if memLimitGB > maxMemGB {
 		memLimitGB = maxMemGB
 	}
-	return int(memLimitGB * 5_000)
+	return int(memLimitGB*perGBIncrement) + c
```
carsonip, line 585:
```suggestion
	return int(memLimitGB*perGBIncrement + constant)
```

carsonip (approved):

carsonip (approved):

carsonip:
Testing notes:

:x: test-plan-fail

- No `"Aggregation.ServiceDestinations.MaxGroups set to %d based on %0.1fgb of memory"` found in logs. Probably due to mixing a default & `s.config.Aggregation.ServiceDestinations.MaxGroups <= 0` check.

lahsivjar:
Tested via the regression fix PR: https://github.com/elastic/apm-server/pull/11905#issuecomment-1782554173. Marking as `test-plan-ok` now.

Issue #10057, opened 2023-01-17
docs: Update standalone apm server API key required privileges


carsonip:
The documentation around required APM server API keys privileges is outdated. https://www.elastic.co/guide/en/apm/guide/current/feature-roles.html

> The privileges that the fleet-server has are listed directly in ES https://github.com/elastic/elasticsearch/blob/main/x-pack/docs/en/rest-api/security/get-service-accounts.asciidoc?plain=1#L64C28-L85. These privileges need to contain all the privileges that apm needs, as it is the basis for creating API Keys when using the Elastic Agent and Fleet are created from them.

simitt:
@bmorelli25 could you take a look at this please, we had some users running into issues on upgrade as we haven't updated the required security privileges.

bmorelli25:
Taking a look at this now, and I'm slightly confused. The page you link to in the docs is for standalone APM Server, but those privileges are for Fleet. Also, what role are changes required for? All of them?

bmorelli25:
Okay so...

ES docs indicate the following `cluster` level privileges for Fleet Server:

```
  "cluster":[
    "manage_own_api_key",
    "monitor"
  ],
```

ES docs indicate privileges are required on the following indices:

```
  "indices":[
    {
      "names":[
        "logs-*",
        "metrics-*",
        "traces-*",
        ".logs-endpoint.diagnostic.collection-*",
        ".logs-endpoint.action.responses-*"
      ],
```

ES docs indicate that the following index-level privileges are needed:

```
          "privileges": [
            "write",
            "create_index",
            "auto_configure"
          ],
```

> These privileges need to contain all the privileges that apm needs, as it is the basis for creating API Keys when using the Elastic Agent and Fleet are created from them.

So our API key creation needs to include all of the above, plus the following application-level privileges:

```
  "applications":[
    {
      "application":"apm",
      "privileges":[
        "event:write"
      ],
      "resources":[
        "*"
      ]
    }
  ]
``` 

Which results in this. Does this sound right?

```
{
  "cluster":[
    "manage_own_api_key",
    "monitor"
  ],
  "indices":[
    {
      "names":[
        "logs-*",
        "metrics-*",
        "traces-*",
        ".logs-endpoint.diagnostic.collection-*",
        ".logs-endpoint.action.responses-*"
      ],
      "privileges":[
        "write",
        "create_index",
        "auto_configure"
      ],
      "allow_restricted_indices":false
    }
  ],
  "applications":[
    {
[...]

PR #11741, opened 2023-09-29
docs: Document privileges to read RUM source maps; Update examples

### LINKED ISSUE #11722: Outdated docs on privileges to use RUM sourcemapping
[Docs on privileges to perform RUM sourcemapping](https://www.elastic.co/guide/en/apm/guide/8.9/privileges-to-publish-events.html) is outdated. Since the introduction of ES sourcemap fetcher in #9722 , to use ES sourcemap fetcher, it requires `read` on `.apm-source-map`.

Also, the right place for the docs should be [here](https://www.elastic.co/guide/en/apm/guide/8.10/secure-comms-stack.html#feature-roles).

Ideally backport the doc update to 8.7+ since the required privilege is new from 8.7.

carsonip:
## Motivation/summary

- Document required privileges to read RUM source map from ES since 8.7.
- Update API key creation examples, add agentcfg role to it.

## Checklist

## Related issues

Closes #11722


docs/feature-roles.asciidoc
```diff
@@ -340,6 +337,21 @@ assign the user the following privileges:
...
+
+To grant an APM Server user with the required privileges for reading RUM source maps from {es} directly without {kib},
+assign the user the following privileges:
+
+[options="header"]
+|====
+|Type | Privilege | Purpose
+
+|Index
+|`read` on `.apm-source-map` index
```
carsonip, line 351: TODO: double check if `allow_restricted_indices: true` is required

docs/feature-roles.asciidoc
```diff
@@ -340,6 +337,29 @@ assign the user the following privileges:
...
+|Allow {beatname_uc} to read RUM source maps from {es}
+|====
+
+The above privileges should be sufficient for RUM source mapping to work properly
+as long as {beatname_uc} communicates with {es} successfully.
+If it fails, it may fallback to read source maps via {kib} if configured,
```
carsonip, line 359: Keeping it slightly ambiguous to avoid describing the entire decision tree for fallback logic. This paragraph exists to maintain the pointer to the relevant kibana docs.

docs/feature-roles.asciidoc
carsonip: Tested with `allow_restricted_indices: false` and it still works.

docs/feature-roles.asciidoc
```diff
@@ -91,10 +92,6 @@ that has the following privileges:
 |Write events into {es}
 |====
 
-. If <<configuration-rum,real user monitoring>> is enabled, additional privileges are required to read source maps.
-See {kibana-ref}/rum-sourcemap-api.html[RUM source map API] for more details.
-Assign these extra privileges to the *general writer role*.
-
```
lahsivjar, line 97: Maybe keep this section too and refer it to the new `privileges-rum-source-mapping`? While not directly related, reading source maps is related to writing RUM events as expected.

lahsivjar (approved):
A small suggestion

docs/feature-roles.asciidoc
```diff
@@ -340,6 +339,29 @@ assign the user the following privileges:
 TIP: Looking for privileges and roles needed use central configuration from the {apm-app} or {apm-app} API?
 See {kibana-ref}/apm-app-central-config-user.html[{apm-app} central configuration user].
 
+[[privileges-rum-source-mapping]]
+==== APM Server RUM source mapping
```
bmorelli25, line 343: This will move your changes off of the central config page

```suggestion
[[privileges-agent-central-config]]
=== Grant privileges and roles needed for reading source maps

++++
<titleabbrev>Create a _source map_ user</titleabbrev>
++++

[[privileges-rum-source-mapping]]
==== APM Server RUM source mapping
```

carsonip:
The example here may need to be updated as well: https://www.elastic.co/guide/en/apm/guide/current/beats-api-keys.html#beats-api-key-es

bmorelli25 (dismissed):

docs/access-api-keys.asciidoc
```diff
@@ -123,11 +123,15 @@ POST /_security/api_key
         {
           "names": ["{beat_default_index_prefix}-*"],
           "privileges": ["create_index", "create_doc"]
-        },
+        }
+      ]
+    },
+    "{beat_default_index_prefix}_sourcemap": {
```
carsonip, line 129: tbh I'm not sure what's the point of this example, since the permissions are good enough to run apm-server with basic functionalities + rum but it is still lacking agentcfg role and monitoring role etc. I've updated this section so that it follows what's recommended in the updated RUM role. I'm wondering if we should expand this example to include all roles that allow apm-server to do everything it'll do but potentially making the doc hard to maintain, or stick with what we have now. In any case, this shouldn't block this PR.

carsonip: This example is about "Create an API key for writing events" and RUM source mapping kind of falls under it but at the same time not exactly.

bmorelli25: This is a good point. I think adding agent config makes sense. We should add it to both of the examples on this page (like you did with source maps). Monitoring I think we're okay to leave separate. It already has a separate section on this page.

bmorelli25 (dismissed):

simitt (commented):
The changes look good; 
overall the examples are outdated, as the indices do not start with `apm*` anymore, but with `logs-*, metrics-*, traces-*`.  There is another issue to tackle this https://github.com/elastic/apm-server/issues/10057#issuecomment-1770179922, so if prefered, this can be tackled extra.

carsonip:
> but I don't think the index name is right .{beat_default_index_prefix}-source-map - shouldn't it be .apm-source-map

Effectively the compiled docs still show `.apm-source-map`, but I agree that it is better to hardcode it.

docs/access-api-keys.asciidoc
carsonip: Effectively the doc compiles to `.apm-source-map`, but I agree that it is better to hardcode it.

carsonip: added agentcfg role to example.

simitt (approved):

PR #11780, opened 2023-10-05
ironbank: bump version to 8.8

v1v:
## Motivation/summary

> The Container Hardening Team (CHT) has identified a few of the Elastic images with UBI 8.7 as a base image.  This image is now EOL and 8.8 is the latest but will be EOL in the first quarter of 2024, so the CHT is recommending moving to UBI 9.x, which is being updated more often by Red Hat with security fixes.

## Checklist

## How to test these changes

In the CI and with the unified release automation

## Further detail 

I'd propose to bump the version to the next minor and then work around supporting 9.x in a follow up

### DIFF
--- a/packaging/ironbank/Dockerfile
+++ b/packaging/ironbank/Dockerfile
@@ -4,7 +4,7 @@
 ################################################################################
 ARG BASE_REGISTRY=registry1.dsop.io
 ARG BASE_IMAGE=ironbank/redhat/ubi/ubi8
-ARG BASE_TAG=8.7
+ARG BASE_TAG=8.8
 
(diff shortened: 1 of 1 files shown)


v1v:
> Is there a reason why we can't do that ?

It's major version change, and we might need to verify things work nicely, even though I don't think there is nothing to stop us so we could go all the way to `9.x` if needed. 
OTOH,  no clue what's the latest 9.x supported version. We might need to also review if things are not breaking though, for such, there is a private repo explaining how to run some smoke tests.

While investigating this, maybe this could be merged?

reakaleek (approved):

PR #11781, opened 2023-10-05
fix: invert assertion check in panic middleware to actually use the error

kruskall:
## Motivation/summary

We're ignoring panic messages. This is because the error wrapping logic will be peformed if the assertion check fails which means the error is always nil.

## Checklist


axw (approved):
good catch! maybe worth backporting this one?

carsonip:
I doubt this can be easily tested. This change is ok judging from the diff.

test-plan-skip

PR #11808, opened 2023-10-09
build: mark apm-server.yml as config|noreplace in nfpm config

kruskall:
## Motivation/summary

Upgrade apm-server rpm package silently replace the old config regardless of whether it was modified.

Set the config file as `config|noreplace` in nfpm to make sure it is not overwritten during an upgrade.

## Checklist


axw (dismissed):
LGTM, please add changelog and backport to 8.10 and 8.11.

axw (approved):

PR #11881, opened 2023-10-17
docs: clarify the usage of the server information endpoint

rodrigc:
Discussed in [case 01501264](https://support.elastic.co/cases/5008X00002UnA9UQAV)

Signed-off-by: Craig Rodrigues <rodrigc@crodrigues.org>

## Motivation/summary

I found the existing documentation for the server information endpoint to be a bit confusing.
I didn't understand that GET and POST behave differently.

## Checklist

### DIFF
--- a/docs/api-info.asciidoc
+++ b/docs/api-info.asciidoc
@@ -8,22 +8,56 @@ This lightweight endpoint is useful as a server up/down health check.
 [[api-info-endpoint]]
 === Server Information endpoint
 
-Send an `HTTP GET` request to the server information endpoint:
+This is the server information endpoint:
 
 [source,bash]
 ------------------------------------------------------------
 http(s)://{hostname}:{port}/
 ------------------------------------------------------------
 
-This endpoint always returns an HTTP 200.
+Sending an `HTTP GET` or `HTTP POST` request to the server information endpoint
+will return an HTTP 200, indicating that the server is up.
 
-If <<api-key>> or a <<secret-token>> is configured, requests to this endpoint must be authenticated.
+To configure authenticated access to the APM server,
+the instructions at <<api-key,APM API key>> or <<secret-token,APM Secret Token>>,
+must be followed to configure the correct permissions for APM access.
+
+If an <<api-key>> or a <<secret-token>> is passed along with
+the `HTTP GET` or `HTTP POST` request, in addition to an HTTP 200,
+the response payload will include some information about the APM server.
 
 [float]
-[[api-info-examples]]
-==== Example
+[[api-info-example-get-without-credentials]]
+==== Example: GET, without credentials
 
-Example APM Server information request:
+Example APM Server status request with GET, without credentials:
+
+["source","sh",subs="attributes"]
+---------------------------------------------------------------------------
+curl --verbose -X GET http://127.0.0.1:8200
+
+*   Trying 127.0.0.1:8200...
+* TCP_NODELAY set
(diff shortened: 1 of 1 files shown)


docs/api-info.asciidoc
```diff
@@ -8,22 +8,55 @@ This lightweight endpoint is useful as a server up/down health check.
...
 ------------------------------------------------------------
 
-This endpoint always returns an HTTP 200.
+Sending an `HTTP GET` request to the server information endpoint
+will return an HTTP 200, indicating that the server is up,
+with an empty response body.
+
+Sending an `HTTP POST` request to the server information endpoint
+will return an HTTP 200, indicating that the server is up,
+with the body containing basic server information.
```
axw, line 24: It's not GET vs. POST that makes the difference, it's whether the request has a valid auth token.

axw (commented):
Thanks for sending the PR @rodrigc! I can't see your support case, but I guess there was some miscommunication: the endpoint's response body shouldn't differ based on whether it's a GET or POST method, what matters is whether the client is authorized. This is to minimise leaking information about the service to unauthorized (and potentially malicious) clients.

docs/api-info.asciidoc
rodrigc: Oh OK, the support agent communicated that GET would only return an HTTP 200 response,
so looks like the support agent was confused by these docs also.

In my support ticket, I communicated that I am using Elastic APM 8.10.3, and that I sent an authenticated request using

```
curl --verbose --request POST --header 'Authorization: ApiKey XXX' http://localhot:8200
```

and that just returned an HTTP 200 response, and did not return a payload.

I guess that led to confusion on both my part and the part of the support agent in [case 01501264](https://support.elastic.co/cases/5008X00002UnA9UQAV)

rodrigc: @axw OK, I revised this patch, and clarified the text around the examples

axw: Looks great, thank you :)

axw (approved):

axw:
@rodrigc sorry, could you please sign your commit and force push again?

rodrigc:
Done

axw (approved):

rodrigc:
@axw can you trigger the docs CI?  I'm not sure, but I think something like:

@elasticsearchmachine run elasticsearch-ci/docs

is required

rodrigc:
@axw Thanks for the review and merge of this PR.

I have one question.  Would it be possible to include a default role in the apm-server code which is similar to
the one mentioned here:
https://www.elastic.co/guide/en/apm/guide/current/api-key.html#create-api-key-user

That way, the end user would not need to worry so much about creating a role
to use an API key which works with APM

axw:
@rodrigc we could predefine a role in Elasticsearch (not in apm-server), but you would still need to create/assign a user that role, so it doesn't seem entirely ideal either.

I think the long-term vision is that this should be built into a standard, but broader, role like "editor". The idea being that you have users that are admins, viewers, or editors; admins and editors should be able to create API Keys for agents, whereas viewers should not. If you wanted a more fine-grained role, then you would still have the option to create one.

Does that sound sensible to you?

rodrigc:
That seems like a reasonable approach.  I agree that more thought needs to be put into this.

axw:
@rodrigc I've opened https://github.com/elastic/apm/issues/831 to track this

PR #11988, opened 2023-11-01
Removed auth from secret-token configuration path

mshwf:
The apm-server.yml (v 8.10.4) doesn't have the section 'apm-server.auth.secret-token'  but 'apm-server.secret-token'

### DIFF
--- a/docs/tab-widgets/secret-token.asciidoc
+++ b/docs/tab-widgets/secret-token.asciidoc
@@ -12,6 +12,6 @@ Set the secret token in `apm-server.yaml`:
 
 [source,yaml]
 ----
-apm-server.auth.secret_token: <secret-token>
+apm-server.secret_token: <secret-token>
 ----
 // end::binary[]


axw (commented):
> The apm-server.yml (v 8.10.4) doesn't have the section 'apm-server.auth.secret-token' but 'apm-server.secret-token'

This is not true - what made you think that? Here's the relevant apm-server.yml: https://github.com/elastic/apm-server/blob/b8562dfee49574f43b636bf95f1839a8ba63b4f2/apm-server.yml#L5-L20. Note that `secret_token` is nested under `auth`, which is nested under `apm-server`.

It used to be `apm-server.secret_token` in older versions.

mshwf:
@axw My bad sorry, had another look now and it's there indeed. What made me think that, is that I uncommented the line that has the `secret_token:` and became under `apm-server:`, and the integration with APM Server worked (my first run with APM Server after many failed trials), but now I think the integration worked not because the secret_token was correct, but because it was ignored.
Thanks, without this PR I would take this misunderstanding as a fact :D
Rejecting the PR...

axw:
No worries, thank you for the PR all the same :)

PR #11992, opened 2023-11-02
chore: enable dependabot version updates of github-actions yaml files

v1v:
## What does this PR do?

* Update dependabot to run on [Sundays](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#groups) and use [groups](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#groups).
* Add dependabot for github actions.
* Add dependabot for github composite actions.

Unfortunately it's not possible to pass different directories yet, see https://github.com/dependabot/dependabot-core/issues/2178

## Why is it important?

Reduce the noise during the week and create one PR with all the changes per package ecosystem

## Checklist

## How to test these changes

In the CI


v1v:
> question: do we need to enable this on multiple branches since we support multiple branches ?

As far as I see dependabot does not support multiple branches. I guess it could be done with mergify, so by tagging those PRs from dependabot with their relevant `backport-8.x`

amannocci (dismissed):

PR #12014, opened 2023-11-07
Update shared-docker.asciidoc

jkakavas:
Update base image reference in docs from CentOS to Ubuntu. CentOS is unsupported/unmaintained and this reference might discourage potential users.

### DIFF
--- a/docs/shared-docker.asciidoc
+++ b/docs/shared-docker.asciidoc
@@ -2,7 +2,7 @@
 ==== Run {beatname_uc} on Docker
 
 Docker images for {beatname_uc} are available from the Elastic Docker
-registry. The base image is https://hub.docker.com/_/centos/[centos:7].
+registry. The base image is https://hub.docker.com/_/ubuntu[ubuntu:20.04].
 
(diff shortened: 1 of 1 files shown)


docs/shared-docker.asciidoc
```diff
@@ -2,7 +2,7 @@
 ==== Run {beatname_uc} on Docker
 
 Docker images for {beatname_uc} are available from the Elastic Docker
-registry. The base image is https://hub.docker.com/_/centos/[centos:7].
+registry. The base image is https://hub.docker.com/_/ubuntu[ubuntu:20.04].
```
carsonip, line 5: I think we switched to 22.04 in Oct 2023: https://github.com/elastic/apm-server/blob/main/packaging/docker/Dockerfile#L1

carsonip: Thanks for catching this. May need 2 separate PRs then, one for main to show 22.04 and one up to 8.11 to show 20.04.

carsonip (approved):

PR #12035, opened 2023-11-10
perf: set max conns idle per host as docappender max request 

kruskall:
## Motivation/summary

By default the http transport will use 2 as the max idle connections per host.
The docappender client will send request a single host and will have at most
MaxRequest appenders sharing the same client. To improve performance and
connection reuse we set the number of max idle conns per host to MaxRequest

## Checklist


internal/beater/beater.go
```diff
@@ -700,6 +700,7 @@ func (s *Runner) newFinalBatchProcessor(
 	}
 	esConfig.FlushInterval = time.Second
 	esConfig.Config = elasticsearch.DefaultConfig()
+	esConfig.MaxIdleConnsPerHost = 10
```
simitt, line 703: Why setting it to `10` here and then only later to `esConfig.MaxRequests`?

simitt: Other default values are set in the `elasticsearch/config.go`; IMO the default value should also be set there.

simitt: So in this case, shouldn't this default value then always be set via `esConfig.MaxRequests` in line `724`? Why setting it to a hardcoded default, that easily gets outdated (as the docappender default value comes from a different source)?

simitt (approved):
LGTM; would be good to have some actual numbers on this improvement.

PR #12066, opened 2023-11-17
Depend on apm-data plugin

axw:
## Motivation/summary

Updates docker-compose.yml to enable the apm-data Elasticsearch plugin, which installs the built-in Elasticsearch index templates and ingest pipelines. Update the system tests to stop installing the integration package.

`apm-server.data_streams.wait_for_integration` configuration has been removed. This PR should therefore not be merged until the apm-data Elasticsearch plugin is enabled by default.

## Checklist

- [x] Update [CHANGELOG.asciidoc](https://github.com/elastic/apm-server/blob/main/CHANGELOG.asciidoc)
~- [ ] Update [package changelog.yml](https://github.com/elastic/apm-server/blob/main/apmpackage/apm/changelog.yml) (only if changes to `apmpackage` have been made)~
~- [ ] Documentation has been updated~ docs will be updated separately, they're not in this repo anymore

## How to test these changes

Run the system tests.

## Related issues

https://github.com/elastic/apm-server/issues/11528
https://github.com/elastic/apm-server/issues/11529

PR #12102, opened 2023-11-24
apmpackage: prepare for move to apm-data plugin

axw:
## Motivation/summary

This is a precursor to https://github.com/elastic/apm-server/pull/12066.

The undocumented, best-effort 7.x migration data stream migration pipeline has been removed. To migrate, users can install an older version and perform a stepped upgrade.

We add mappings for various fields to the integration package that were previously unmapped.

The following fields are all newly mapped as `flattened`:
 - `error.custom`
 - `error.exception.attributes`
 - `error.exception.stacktrace`
 - `error.log.stacktrace`
 - `http.request.body` (always send as a JSON object with a single field, `original`)
 - `http.request.env`
 - `http.request.cookies`
 - `span.message.headers`
 - `span.stacktrace`
 - `transaction.custom`
 - `transaction.message.headers`

`http.{request,response}.headers` have been changed from `object` to `flattened`. Even though these were mapped as `object` type, they were inheriting `dynamic: false` and thus the headers never got mapped. The result of this change is headers will be returned in fields now, as `http.request.headers: {Content-Length: ...}`. There's no change to the `_source` for these fields.

`http.response.headers_sent` is mapped as a boolean.
`transaction.span_count.started` is mapped as a long for consistency with `transaction.span_count.dropped`.

The following fields are mapped as `keyword`, in line with the dynamic mapping rules of the apm-data plugin's templates:
- `span.db.instance`
- `span.db.type`
- `span.db.user.name`
- `span.db.statement`
- `span.message.body`
- `span.message.routing_key`
- `transaction.message.body`
- `transaction.message.routing_key`

## Checklist

## How to test these changes

- Check that error/exception stack traces show up correctly in APM UI
- Check that HTTP request/response headers show up in APM UI
- Check that SQL statements show up in APM UI

## Related issues

https://github.com/elastic/apm-server/pull/12066


lahsivjar (approved):

simitt:
Verified with BC-2 that
- error stacktraces show up
- http headers show up
- sql statements show up as expected

PR #12120, opened 2023-11-29
feat: compute tbs entry size before writing and flush to avoid limit error

### LINKED ISSUE #11857: Potential missing spans when TBS storage limit is reached in a multi APM server setup
**APM Server version** (`apm-server version`): reproduced in 8.10 but should affect more versions

In a multi APM server setup,

- Expected: When TBS is enabled and storage limit is reached, since everything is indexed by default, traces should be complete.
- Actual: When TBS is enabled and storage limit is reached, we may not see spans that are expected in the trace.

**Steps to reproduce**:

look at it.

 1. Setup 2 APM servers with TBS enabled and a low storage limit
 2. Configure 2 APM agents sending to different APM servers
 3. Send distributed traces
 4. Traces may look incomplete with missing spans in the waterfall

kruskall:
## Motivation/summary

Keep track of pending writes in readwrites and use an atomic counter for the overall size of pending writes in the backing storage.
Catch storage issues earlier and avoid losing events. 
Estimate the size of an entry by hardcoding values from badger internal code.

## Checklist

## How to test these changes

    Setup 1 APM server with TBS enabled and a low storage limit
    Configure 1 APM agent to send a transaction containing spans in high throughput.
    Traces may look incomplete with missing spans in the waterfall

## Related issues

Closes #11857

### DIFF
--- a/changelogs/head.asciidoc
+++ b/changelogs/head.asciidoc
@@ -20,3 +20,4 @@ https://github.com/elastic/apm-server/compare/8.12\...main[View commits]
 ==== Added
 - map OTel's `span.Status: Unset` to `event.outcome: success` instead of `event.outcome: unknown`
 - Add support for Otel code.stacktrace {pull}12096[12096]
+- `http.request.body.bytes` now reports the bytes read from request body even if Content-Length is -1 {pull}12451[12451]
--- a/internal/beater/middleware/log_middleware.go
+++ b/internal/beater/middleware/log_middleware.go
@@ -46,6 +46,7 @@ func LogMiddleware() Middleware {
 			}
 			h(c)
 			c.Logger = c.Logger.With("event.duration", time.Since(c.Timestamp))
+			c.Logger = c.Logger.With("http.request.body.bytes", c.RequestBodyBytes())
 			if c.MultipleWriteAttempts() {
 				c.Logger.Warn("multiple write attempts")
 			}
@@ -74,9 +75,6 @@ func loggerWithRequestContext(c *request.Context) *logp.Logger {
 	}
 	if c.ClientIP.IsValid() && c.ClientIP != c.SourceIP {
 		logger = logger.With("client.ip", c.ClientIP.String())
-	}
-	if c.Request.ContentLength != -1 {
-		logger = logger.With("http.request.body.bytes", c.Request.ContentLength)
 	}
 	return logger
 }
--- a/internal/beater/middleware/log_middleware_test.go
+++ b/internal/beater/middleware/log_middleware_test.go
@@ -19,6 +19,8 @@ package middleware
 
 import (
 	"net/http"
+	"net/http/httptest"
+	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
@@ -135,4 +137,21 @@ func TestLogMiddleware(t *testing.T) {
 			}
 		})
 	}
+}
+
+func TestLogMiddlewareRequestBodyBytes(t *testing.T) {
+	configure.Logging("APM Server test", agentconfig.MustNewConfigFrom(`{"ecs":true}`))
+	require.NoError(t, logp.DevelopmentSetup(logp.ToObserverOutput()))
+
+	c := request.NewContext()
+	c.Reset(
+		httptest.NewRecorder(),
+		httptest.NewRequest(http.MethodGet, "/", strings.NewReader("content")),
+	)
+	Apply(LogMiddleware(), HandlerIdle)(c)
+
+	entries := logp.ObserverLogs().TakeAll()
+	require.Equal(t, 1, len(entries))
+	field := entries[0].ContextMap()["http.request.body.bytes"]
+	assert.Equal(t, int64(len("content")), field)
 }
--- a/internal/beater/request/context.go
+++ b/internal/beater/request/context.go
@@ -54,6 +54,7 @@ type Context struct {
 	// compressedRequestReadCloser will be initialised for requests
 	// with a non-zero ContentLength and empty Content-Encoding.
 	compressedRequestReadCloser compressedRequestReadCloser
+	countingReadCloser          countingReadCloser
 	gzipReader                  *gzip.Reader
 	zlibReader                  zlibReadCloseResetter
 
@@ -155,6 +156,13 @@ func (c *Context) setRequest(r *http.Request) {
 		c.SourceNATIP = c.ClientIP
 		c.SourceIP, c.ClientIP = ip, ip
 		c.SourcePort, c.ClientPort = int(port), int(port)
+	}
+
+	if c.Request.ContentLength != -1 {
+		c.countingReadCloser.n = c.Request.ContentLength
+	} else {
+		c.countingReadCloser.ReadCloser = c.Request.Body
+		c.Request.Body = &c.countingReadCloser
 	}
 
 	if err := c.decodeRequestBody(); err != nil {
@@ -316,6 +324,29 @@ func (c *Context) resetGzip(r io.Reader) (io.ReadCloser, error) {
 		err = c.gzipReader.Reset(r)
 	}
 	return c.gzipReader, err
+}
+
+// RequestBodyBytes returns the original c.Request.ContentLength if it
+// was not -1, otherwise it returns the number of bytes read from the
+// request body.
+//
+// RequestBodyBytes must not be called concurrently with
+// c.Request.Body.Read().
+func (c *Context) RequestBodyBytes() int64 {
+	return c.countingReadCloser.n
+}
+
+type countingReadCloser struct {
+	io.ReadCloser
+	n int64
+}
+
+func (r *countingReadCloser) Read(p []byte) (int, error) {
+	n, err := r.ReadCloser.Read(p)
+	if n > 0 {
+		r.n += int64(n)
+	}
+	return n, err
 }
 
 type compressedRequestReadCloser struct {
--- a/internal/beater/request/context_test.go
+++ b/internal/beater/request/context_test.go
@@ -123,6 +123,8 @@ func TestContext_Reset(t *testing.T) {
 			assert.False(t, timestamp.Before(before))
 		case "compressedRequestReadCloser":
 			assert.Zero(t, c.compressedRequestReadCloser)
+		case "countingReadCloser":
+			assert.Zero(t, c.countingReadCloser)
 		case "zlibReader":
 			assert.Nil(t, c.zlibReader)
 		case "gzipReader":
@@ -249,6 +251,35 @@ func TestContextResetContentEncoding(t *testing.T) {
 	test("gzip_sniff", "", bytes.NewReader(gzipCompressed), "contents", "gzip")
 	test("deflate", "deflate", bytes.NewReader(deflateCompressed), "contents", "deflate")
 	test("deflate_sniff", "", bytes.NewReader(deflateCompressed), "contents", "deflate")
+}
+
+func TestContextRequestBodyBytes(t *testing.T) {
+	t.Run("no_body", func(t *testing.T) {
+		w := httptest.NewRecorder()
+		r := httptest.NewRequest(http.MethodGet, "/", nil)
+		c := Context{Request: r, ResponseWriter: w}
+		c.Reset(w, r)
+		assert.Equal(t, int64(0), c.RequestBodyBytes())
+	})
+
+	t.Run("client_specified_content_length", func(t *testing.T) {
+		w := httptest.NewRecorder()
+		r := httptest.NewRequest(http.MethodGet, "/", strings.NewReader("contents"))
+		c := Context{Request: r, ResponseWriter: w}
+		c.Reset(w, r)
+		assert.Equal(t, int64(len("contents")), c.RequestBodyBytes())
+	})
+
+	t.Run("unspecified_content_length", func(t *testing.T) {
+		w := httptest.NewRecorder()
+		compressed := gzipCompressString("contents")
+		r := httptest.NewRequest(http.MethodGet, "/", bytes.NewReader(compressed))
+		r.ContentLength = -1
+		c := Context{Request: r, ResponseWriter: w}
+		c.Reset(w, r)
+		assertReaderContents(t, "contents", c.Request.Body)
+		assert.Equal(t, int64(len(compressed)), c.RequestBodyBytes())
+	})
 }
 
 func BenchmarkContextResetContentEncoding(b *testing.B) {


x-pack/apm-server/sampling/eventstorage/storage.go
```diff
@@ -206,9 +209,14 @@ func (rw *ReadWriter) writeEntry(e *badger.Entry, opts WriterOpts) error {
 	if err := rw.Flush(opts.StorageLimitInBytes); err != nil {
 		return err
 	}
+	rw.pendingSize += entrySize
```
axw, line 212: This doesn't seem right. I think we'll need to track the amount of bytes buffered across all writers. Maybe this should be an `atomic.Uint64` on the `Storage` type?

x-pack/apm-server/sampling/eventstorage/storage.go
```diff
@@ -206,9 +209,14 @@ func (rw *ReadWriter) writeEntry(e *badger.Entry, opts WriterOpts) error {
...
 	return rw.txn.SetEntry(e)
 }
 
+func estimateSize(e *badger.Entry) int {
+	return len(e.Key) + len(e.Value) + 12 + 2
```
axw, line 217: Can you please add some comments explaining what the magic numbers are?

x-pack/apm-server/sampling/eventstorage/storage.go
```diff
@@ -20,6 +20,8 @@ const (
 	entryMetaTraceSampled   = 's'
 	entryMetaTraceUnsampled = 'u'
 	entryMetaTraceEvent     = 'e'
+
+	defaultPendingSize = 1024
```
axw, line 24: Where does this number come from?

x-pack/apm-server/sampling/eventstorage/storage.go
```diff
@@ -185,18 +180,26 @@ func (rw *ReadWriter) WriteTraceEvent(traceID string, id string, event *modelpb.
...
 	rw.pendingWrites++
-	err := rw.txn.SetEntry(e.WithTTL(opts.TTL))
-	// Attempt to flush if there are 200 or more uncommitted writes.
-	// This ensures calls to ReadTraceEvents are not slowed down;
-	// ReadTraceEvents uses an iterator, which must sort all keys
-	// of uncommitted writes.
-	// The 200 value yielded a good balance between read and write speed:
-	// https://github.com/elastic/apm-server/pull/8407#issuecomment-1162994643
-	if rw.pendingWrites >= 200 {
+	entrySize := int64(estimateSize(e)) + 10
```
axw, line 183: Why are you adding 10 to it? Should that be part of `estimateSize`?

x-pack/apm-server/sampling/eventstorage/storage.go
```diff
@@ -126,19 +130,10 @@ func (rw *ReadWriter) Close() {
 // files exceeds the configured threshold.
 func (rw *ReadWriter) Flush(limit int64) error {
```
axw, line 131: We should remove the limit param, right? Also the doc comment needs updating since ErrLimitReached won't be returned any more.

x-pack/apm-server/sampling/eventstorage/storage.go
```diff
@@ -126,19 +130,10 @@ func (rw *ReadWriter) Close() {
...
-		rw.pendingWrites = 0
-		return fmt.Errorf(
-			flushErrFmt+" (current: %d, limit: %d)",
-			ErrLimitReached, current, limit,
-		)
-	}
 	err := rw.txn.Commit()
 	rw.txn = rw.s.db.NewTransaction(true)
 	rw.pendingWrites = 0
+	rw.pendingSize = defaultPendingSize
```
axw, line 136: Like in `ReadWriter.writeEntry`, this should be coordinated across all ReadWriters.

x-pack/apm-server/sampling/eventstorage/storage.go
```diff
@@ -65,8 +67,9 @@ func (s *Storage) NewShardedReadWriter() *ShardedReadWriter {
 // The returned ReadWriter must be closed when it is no longer needed.
 func (s *Storage) NewReadWriter() *ReadWriter {
 	return &ReadWriter{
-		s:   s,
-		txn: s.db.NewTransaction(true),
+		s:           s,
+		txn:         s.db.NewTransaction(true),
+		pendingSize: defaultPendingSize,
```
axw, line 72: Instead of defaulting to 1024, should we perhaps use `s.db.Size()`? And rather than calculating `pendingSize`, maybe we could track the remaining capacity? i.e. `StorageLimitInBytes - Size()`

x-pack/apm-server/sampling/eventstorage/storage.go
```diff
@@ -185,30 +166,66 @@ func (rw *ReadWriter) WriteTraceEvent(traceID string, id string, event *modelpb.
...
+	entrySize := estimateSize(e)
+	lsm, vlog := rw.s.db.Size()
+
+	// there are multiple ReadWriters writing to the same storage so add
+	// the entry size and consider the new value to avoid TOCTOU issues.
+	pendingSize := rw.s.pendingSize.Add(entrySize)
+	rw.pendingSize += entrySize
+
+	if pendingSize+entrySize+lsm+vlog >= opts.StorageLimitInBytes {
+		if err := rw.Flush(); err != nil {
```
lahsivjar, line 178: I am a bit confused here, how will flushing avoid reaching the limit? IIUC, flushing will also contribute towards `StorageLimitInBytes`.

simitt:
@kruskall I moved this to `ms-8.13` - can you please pick this up again, the PR looks stale.

x-pack/apm-server/sampling/eventstorage/storage.go
```diff
@@ -185,30 +172,72 @@ func (rw *ReadWriter) WriteTraceEvent(traceID string, id string, event *modelpb.
...
+	lsm, vlog := rw.s.db.Size()
+
+	// there are multiple ReadWriters writing to the same storage so add
+	// the entry size and consider the new value to avoid TOCTOU issues.
+	pendingSize := rw.s.pendingSize.Add(entrySize)
+	rw.pendingSize += entrySize
+
+	if current := pendingSize + lsm + vlog; opts.StorageLimitInBytes != 0 && current >= opts.StorageLimitInBytes {
+		// Discard the txn and re-create it if the soft limit has been reached.
+		rw.txn.Discard()
```
carsonip, line 185: IIUC this is still discarding tx which are accepted (and returned as ok to the caller). I think we should flush the rw immediately instead. Also I'm thinking if we need to add a mutex somewhere to avoid race and overshooting the storage limit

x-pack/apm-server/sampling/eventstorage/storage.go
carsonip: `flush` was called with an overestimated rw.s.pendingSize due to added entrySize. Probably doesn't make any difference, but would be better to call flush with the right pendingSize, i.e. reverting it before calling.

x-pack/apm-server/sampling/eventstorage/storage.go
```diff
@@ -181,14 +181,13 @@ func (rw *ReadWriter) writeEntry(e *badger.Entry, opts WriterOpts) error {
...
 		// Discard the txn and re-create it if the soft limit has been reached.
 		rw.txn.Discard()
 		rw.txn = rw.s.db.NewTransaction(true)
-		// reset pending size
-		rw.s.pendingSize.Add(-rw.pendingSize)
-		rw.pendingWrites = 0
-		rw.pendingSize = baseTransactionSize
-		rw.s.pendingSize.Add(baseTransactionSize)
 		return fmt.Errorf("%w (current: %d, limit: %d)", ErrLimitReached, current, opts.StorageLimitInBytes)
 	} else if rw.pendingWrites >= 200 {
```
carsonip, line 192: line 199 rw.Flush is called with overestimated rw.s.pendingSize

carsonip (commented):
The code becomes harder to reason. Maybe refactor it a bit?

x-pack/apm-server/sampling/eventstorage/storage.go
```diff
@@ -185,30 +172,71 @@ func (rw *ReadWriter) WriteTraceEvent(traceID string, id string, event *modelpb.
...
+		// The 200 value yielded a good balance between read and write speed:
+		// https://github.com/elastic/apm-server/pull/8407#issuecomment-1162994643
+		if err := rw.Flush(); err != nil {
 			return err
 		}
 	}
+	err := rw.txn.SetEntry(e.WithTTL(opts.TTL))
+	// the current ReadWriter flushed the transaction and reset the pendingSize so add
+	// the entrySize again.
+	if rw.pendingSize == 0 {
```
carsonip, line 206: rw.pendingSize will not be 0. Even after flush it will be baseTransactionSize. Put this block inside rw.pendingWrites >= 200 because it will only be flushed without aborting inside

x-pack/apm-server/sampling/eventstorage/storage.go
carsonip: Flush will always recreate tx. Should be no need to discard and create new tx. Update the comment to say "Discard the entry"

x-pack/apm-server/sampling/processor_test.go
```diff
@@ -704,8 +704,10 @@ func TestStorageLimit(t *testing.T) {
 	config := newTempdirConfig(t)
 	// Write 5K span events and close the DB to persist to disk the storage
 	// size and assert that none are reported immediately.
-	writeBatch(5000, config, func(b modelpb.Batch) { assert.Empty(t, b) })
-	assert.NoError(t, config.Storage.Flush(0))
+	writeBatch(5000, config, func(b modelpb.Batch) {
+		assert.Empty(t, b, fmt.Sprintf("expected empty but size if %d", len(b)))
```
carsonip, line 708:
```suggestion
		assert.Empty(t, b, fmt.Sprintf("expected empty but size of %d", len(b)))
```

x-pack/apm-server/sampling/eventstorage/storage.go
```diff
@@ -185,30 +172,70 @@ func (rw *ReadWriter) WriteTraceEvent(traceID string, id string, event *modelpb.
...
+		// The 200 value yielded a good balance between read and write speed:
+		// https://github.com/elastic/apm-server/pull/8407#issuecomment-1162994643
+		if err := rw.Flush(); err != nil {
 			return err
 		}
+
+		// the current ReadWriter flushed the transaction and reset the pendingSize so add
+		// the entrySize again.
+		rw.pendingSize += entrySize
+		rw.s.pendingSize.Add(entrySize)
```
carsonip, line 205: this is possible to race & overshoot storage limit. Imagine this:

normal scenario:
rw1: line 198 flush, rw1.s.pendingSize.Add(-rw1.pendingSize)
rw1: line 204-205 pendingSize increase
rw2: line 175-183, failed size check, flush immediately and reject current entry

race scenario that leads to overshooting the limit:
rw1: line 198 flush, rw1.s.pendingSize.Add(-rw1.pendingSize)
rw2: line 175-183, passes size check
rw2: line 208-214, setEntry successfully
rw1: line 204-205 pendingSize increase
rw1: line 208-214, setEntry successfully

x-pack/apm-server/sampling/eventstorage/storage.go
```diff
@@ -185,30 +172,70 @@ func (rw *ReadWriter) WriteTraceEvent(traceID string, id string, event *modelpb.
...
+	err := rw.txn.SetEntry(e.WithTTL(opts.TTL))
+
 	// If the transaction is already too big to accommodate the new entry, flush
 	// the existing transaction and set the entry on a new one, otherwise,
 	// returns early.
 	if err != badger.ErrTxnTooBig {
 		return err
 	}
-	if err := rw.Flush(opts.StorageLimitInBytes); err != nil {
+	if err := rw.Flush(); err != nil {
```
carsonip, line 216: ditto

x-pack/apm-server/sampling/eventstorage/storage.go
```diff
@@ -185,30 +172,70 @@ func (rw *ReadWriter) WriteTraceEvent(traceID string, id string, event *modelpb.
...
-	// ReadTraceEvents uses an iterator, which must sort all keys
-	// of uncommitted writes.
-	// The 200 value yielded a good balance between read and write speed:
-	// https://github.com/elastic/apm-server/pull/8407#issuecomment-1162994643
+	entrySize := estimateSize(e)
+	lsm, vlog := rw.s.db.Size()
+
+	// there are multiple ReadWriters writing to the same storage so add
+	// the entry size and consider the new value to avoid TOCTOU issues.
+	pendingSize := rw.s.pendingSize.Add(entrySize)
```
carsonip, line 180: Given that 176 and 180 are not protected by a mutex, will it obtain an inconsistent view on db size and pendingSize? i.e. 1 is updated while another isn't, by another thread

x-pack/apm-server/sampling/eventstorage/storage.go
```diff
@@ -185,30 +172,70 @@ func (rw *ReadWriter) WriteTraceEvent(traceID string, id string, event *modelpb.
...
 	rw.pendingWrites++
-	err := rw.txn.SetEntry(e.WithTTL(opts.TTL))
-	// Attempt to flush if there are 200 or more uncommitted writes.
-	// This ensures calls to ReadTraceEvents are not slowed down;
-	// ReadTraceEvents uses an iterator, which must sort all keys
-	// of uncommitted writes.
-	// The 200 value yielded a good balance between read and write speed:
-	// https://github.com/elastic/apm-server/pull/8407#issuecomment-1162994643
+	entrySize := estimateSize(e)
+	lsm, vlog := rw.s.db.Size()
```
carsonip, line 176: Add a comment here to mention Size() is updated in the background every minute

x-pack/apm-server/sampling/eventstorage/storage_test.go
```diff
@@ -300,11 +298,6 @@ func TestStorageLimit(t *testing.T) {
 		TTL:                 time.Minute,
 		StorageLimitInBytes: 1, // ignored in the write, because there's no implicit flush
```
carsonip, line 299:
```suggestion
		StorageLimitInBytes: 1,
```

stale comment

carsonip (commented):
I've written a test for the error scenario:
```go

func TestStorageLimitWritesPersisted(t *testing.T) {
	// Ensure that if WriteTraceEvent does not return an error,
	// the event is eventually written to storage.
	const storageLimitInBytes = int64(10000)
	tempdir := t.TempDir()
	opts := func() badger.Options {
		opts := badgerOptions()
		opts = opts.WithInMemory(false)
		opts = opts.WithDir(tempdir).WithValueDir(tempdir)
		return opts
	}

	// Open and close the database to create a non-empty value log file,
	// which will cause writes below to fail due to the storage limit being
	// exceeded. We would otherwise have to rely on Badger's one minute
	// timer to refresh the size.
	db := newBadgerDB(t, opts)
	db.Close()
	db = newBadgerDB(t, opts)

	store := eventstorage.New(db, eventstorage.ProtobufCodec{})

	rw1 := store.NewReadWriter()
	defer rw1.Close()

	rw2 := store.NewReadWriter()
	defer rw2.Close()

	write := func(rw *eventstorage.ReadWriter) error {
		traceID := uuid.Must(uuid.NewV4()).String()
		transactionID := uuid.Must(uuid.NewV4()).String()
		transaction := modelpb.APMEvent{Transaction: &modelpb.Transaction{Id: transactionID}}
		err := rw.WriteTraceEvent(traceID, transactionID, &transaction, eventstorage.WriterOpts{
			TTL:                 time.Minute,
			StorageLimitInBytes: storageLimitInBytes,
		})
		return err
	}

	var (
		rw1Written, rw2Written int
		wg                     sync.WaitGroup
	)
	wg.Add(2)
	go func() {
		for {
			err := write(rw1)
			if err != nil {
				assert.ErrorIs(t, err, eventstorage.ErrLimitReached)
				break
			}
			rw1Written++
		}
		wg.Done()
	}()

	go func() {
		for {
			err := write(rw2)
			if err != nil {
				assert.ErrorIs(t, err, eventstorage.ErrLimitReached)
				break
			}
			rw2Written++
		}
		wg.Done()
	}()
	wg.Wait()

	getEventCount := func() (count int) {
		opts := badger.DefaultIteratorOptions
		txn := db.NewTransaction(true)
		iter := txn.NewIterator(opts)
		defer iter.Close()
		for iter.Rewind(); iter.Valid(); iter.Next() {
			item := iter.Item()
			if item.IsDeletedOrExpired() {
				continue
			}
			switch item.UserMeta() {
			case 'e':
				count++
			default:
				// Unknown entry meta: ignore.
				continue
			}
		}
		return
	}

	assert.Equal(t, rw1Written+rw2Written, getEventCount())
}
```

this will fail before the PR and will pass with the fix:
```
=== RUN   TestStorageLimitWritesPersisted
    storage_test.go:405: 
        	Error Trace:	/home/carson/projects/apm-server/x-pack/apm-server/sampling/eventstorage/storage_test.go:405
        	Error:      	Not equal: 
        	            	expected: 5256598
        	            	actual  : 4925758
        	Test:       	TestStorageLimitWritesPersisted
--- FAIL: TestStorageLimitWritesPersisted (68.69s)

Expected :5256598
Actual   :4925758
```

```
=== RUN   TestStorageLimitWritesPersisted
--- PASS: TestStorageLimitWritesPersisted (0.04s)
PASS
```

x-pack/apm-server/sampling/eventstorage/storage.go
```diff
@@ -185,30 +172,70 @@ func (rw *ReadWriter) WriteTraceEvent(traceID string, id string, event *modelpb.
...
 	if err != badger.ErrTxnTooBig {
 		return err
 	}
-	if err := rw.Flush(opts.StorageLimitInBytes); err != nil {
+	if err := rw.Flush(); err != nil {
 		return err
 	}
+	rw.pendingSize += entrySize
+	rw.s.pendingSize.Add(entrySize)
 	return rw.txn.SetEntry(e)
```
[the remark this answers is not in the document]
carsonip: can you create an issue to determine whether this is a bug?

x-pack/apm-server/sampling/eventstorage/storage.go
carsonip: we had a comment like that in https://github.com/elastic/apm-server/blob/v8.11.4/x-pack/apm-server/sampling/eventstorage/storage.go#L77 and it helps readers understand why we don't really care about races in storage limit checking in this file.

carsonip:
> I think this is an acceptable tradeoff because of two reasons:
> 
> we are already being careful when estimating size (see the first flush which flushes before writing the entry because it would overflow the storage)

The main reason is the badger size reconciliation happens per-minute. So that will end up be the greatest source of error (as in inaccuracy) in our storage limit check.

> the previous behaviour was way worse and actually incorrect: the buffered writes weren't counted, leading to a huge overuse of storage

Agreed. We are definitely not making things worse here. A critical bug is being fixed. Initially I did not think about the overuse of storage but when I last tested it out last week, I realize it was dangerous.

carsonip (commented):
looks good. If you could update the PR description and respond to the comments, I think it should be good to go. @axw do you want to glance over it?

carsonip (approved):

carsonip:
@kruskall this is missing changelog. Sorry I missed it during the review.

carsonip:
## Testing notes

:heavy_check_mark:  test-plan-ok

Tested with sampling rate 0.1 with 1KB TBS size limit under 2 ESS versions: 8.12.0 (affected by bug) and 8.12.1 (with bugfix)

### Code snippet

<details><summary>Go agent code</summary>
<p>

```
package main

import (
	"context"
	"fmt"
	"go.elastic.co/apm/module/apmhttp/v2"
	"go.elastic.co/apm/v2"
	"golang.org/x/net/context/ctxhttp"
	"io/ioutil"
	"net"
	"net/http"
	"time"
)

var tracingClient = apmhttp.WrapClient(http.DefaultClient)

func serverHandler(w http.ResponseWriter, req *http.Request) {
	// Propagate the transaction context contained in req.Context().
	span, _ := apm.StartSpan(req.Context(), "serverHandler", "spantype")
	defer span.End()
	sleep(req.Context())
	resp, err := ctxhttp.Get(req.Context(), tracingClient, "https://httpbin.org/status/200")
	if err != nil {
		apm.CaptureError(req.Context(), err).Send()
		http.Error(w, "failed to query backend", 500)
		return
	}
	ioutil.ReadAll(resp.Body)
	sleep(req.Context())
	return
}

func sleep(ctx context.Context) {
	span, _ := apm.StartSpan(ctx, "sleepspan", "sleep")
	defer span.End()
	time.Sleep(100 * time.Millisecond)
}

const CONCURRENCY = 10

func main() {
	listener, err := net.Listen("tcp", "127.0.0.1:0")
	if err != nil {
		panic(err)
	}
	for i := 0; i < CONCURRENCY; i++ {
		go func() {
			for {
				//select {
				//case <-time.Tick(3 * time.Second):
				//}
				if _, err := http.Get(fmt.Sprintf("http://%s/", listener.Addr())); err != nil {
					panic(err)
				}
			}
		}()
	}
	if err := http.Serve(listener, apmhttp.Wrap(http.HandlerFunc(serverHandler))); err != http.ErrServerClosed {
		panic(err)
	}
}
```

</p>
</details> 

### Expected outcome

Traces should be complete like this:

Screenshot 2024-02-01 at 12-50-42 GET _ - Transactions - ___1high_throughput_8_12_0_ESS - Services - APM - Observability - Elastic

### 8.12.0 affected by bug

Traces are sometimes incomplete under high throughput.


Screenshot 2024-02-01 at 12-54-25 GET _ - Transactions - ___1high_throughput_8_12_0_ESS - Services - APM - Observability - Elastic
Screenshot 2024-02-01 at 12-54-42 GET _ - Transactions - ___1high_throughput_8_12_0_ESS - Services - APM - Observability - Elastic

### 8.12.1 with TBS bug fix

Did not find any incomplete traces after scrolling through 50+ traces.

Screenshot 2024-02-01 at 12-59-39 GET _ - Transactions - ___high_throughput_8_12_1_ESS_with_TBS_fix - Services - APM - Observability - Elastic

PR #12206, opened 2023-12-12
ci: remove check-docker-compose task

kruskall:
## Motivation/summary

docker compose is only used for system tests and we are automation to bump the version. This means there is nothing actionable on our end and it might cause CI to fail if the ES/Kibana docker image is not available.

This is currently the case for the 8.13.0 SNAPSHOT version.

## Checklist


lahsivjar (approved):
LGTM!

v1v:
run docs-build

PR #6714, opened 2021-11-24
package: Update `labels|data_stream` fields as ecs

marclop:
## Motivation/summary

Updates the `elastic-package` binary to the latest version which allows
us to benefit from the latest changes. In particular, references to ecs
external fields may override some of the field properties.

This patch updates the `data_stream.*`, `@timestamp` and `labels` fields
to reference the external ecs definition.

## Checklist

~- [ ] Update [CHANGELOG.asciidoc](https://github.com/elastic/apm-server/blob/master/CHANGELOG.asciidoc)~
- [x] Update [package changelog.yml](https://github.com/elastic/apm-server/blob/master/apmpackage/apm/changelog.yml) (only if changes to `apmpackage` have been made)
~- [ ] Documentation has been updated~

## How to test these changes

1. `make build-package`
2. `docker-compose up -d`
3. Install the APM Integration in Kibana.
4. Verify all the modified fields appear in the index template mappings with their defined ecs types.
5. Verify the "labels" field has `dynamic: true` set in all the index template mappings

## Related issues

Follow up from #6633

PR #12219, opened 2023-12-13
Define value for all constant_keyword fields in apmpackage

### LINKED ISSUE #12159: Define value for all constant-keyword fields in the apm package
The apm package defines fields of type `constant_keyword` in several places. Some of these fields have a fixed value set already in the template definition; others do not. This is generally supported by ES, but causes problems in combination with how Fleet handles mapping updates and rollovers, see details in https://github.com/elastic/elasticsearch/issues/102829. Solving this in Fleet would add undesired complexity, therefore it was suggested to instead require more restrictive behaviour via the package spec, by requiring a value to be defined for constant keyword types (https://github.com/elastic/package-spec/issues/677).

We need to update the apm-package to define values for all of these fields. This includes changing the package generation logic for the interval based templates (https://github.com/elastic/apm-server/blob/main/apmpackage/cmd/genpackage/main.go#L62).

carsonip:
## Motivation/summary

Define value for all `constant_keyword` fields in apmpackage to workaround Fleet upgrade bug. Under certain conditions, it is possible that apmpackage upgrade creates a lot of indices if constant_keyword does not have a value.

Targeting 8.12 branch because apmpackage has been moved to https://github.com/elastic/integrations/tree/main/packages/apm/data_stream on main.

## Checklist

## How to test these changes

- Ingest APM data, upgrade package to a version that contains the fix, and verify that DS rollover is successful and mapping contains a value for constant keyword fields.

## Related issues

Closes #12159

### DIFF
--- a/apmpackage/apm/data_stream/sampled_traces/fields/base-fields.yml
+++ b/apmpackage/apm/data_stream/sampled_traces/fields/base-fields.yml
@@ -1,12 +1,8 @@
 - name: '@timestamp'
-  type: date
-  description: Event timestamp.
+  external: ecs
 - name: data_stream.type
-  type: constant_keyword
-  description: Data stream type.
+  external: ecs
 - name: data_stream.dataset
-  type: constant_keyword
-  description: Data stream dataset.
+  external: ecs
 - name: data_stream.namespace
-  type: constant_keyword
-  description: Data stream namespace.
+  external: ecs
--- a/apmpackage/apm/changelog.yml
+++ b/apmpackage/apm/changelog.yml
@@ -1,5 +1,10 @@
 - version: generated
   changes:
+    - description: Define value for all constant_keyword fields; Use ecs for sampled_traces
+      type: bugfix
+      link: https://github.com/elastic/apm-server/pull/12219
+- version: 8.12.0
+  changes:
     - description: Add missing mappings for various fields
       type: enhancement
       link: https://github.com/elastic/apm-server/pull/12102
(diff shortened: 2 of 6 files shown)


apmpackage/apm/data_stream/sampled_traces/fields/base-fields.yml
```diff
@@ -1,12 +1,8 @@
 - name: '@timestamp'
-  type: date
-  description: Event timestamp.
+  external: ecs
```
carsonip, line 2: [For reviewer] This is similar to #6714

carsonip:
@simitt now that apmpackage no longer lives in apm-server repo, should I split this PR into 2 parts, 1 targeting apm-server 8.12 branch, another changing https://github.com/elastic/integrations/tree/main/packages/apm/data_stream ?

simitt:
Yeah I fear we need to put this into both places now unfortunately.

carsonip (commented):
Once this is merged I'll port the change to https://github.com/elastic/integrations/tree/main/packages/apm

bmorelli25:
run docs-build

apmpackage/apm/data_stream/sampled_traces/fields/base-fields.yml
carsonip: the rendered index template is identical. So there is nothing to worry about.

carsonip:
This change did not go into 8.12.1. Now the plan is to land this fix in 8.12.2.

carsonip:
## testing notes

Upgraded from a 8.12.1 deployment with existing data in metrics-apm.service_destination.1m-default etc to 8.12.2. Upgrade went well and data streams rolled over successfully.

PR #12222, opened 2023-12-13
chore: rename apm-mutating-webhook to apm-k8s-attacher

amannocci:
## What is the change being made?

* Rename `apm-mutating-webhook` to `apm-k8s-attacher`

## Why is the change being made?

* The repository `apm-mutating-webhook` has been renamed to `apm-k8s-attacher`.

### DIFF
--- a/docs/features.asciidoc
+++ b/docs/features.asciidoc
@@ -13,7 +13,7 @@
 * <<cross-cluster-search>>
 * <<span-compression>>
 * <<monitoring-aws-lambda>>
-* <<apm-mutating-admission-webhook>>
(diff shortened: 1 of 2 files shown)


bmorelli25 (changes requested):
Hi. Unfortunately changing a heading is easier said than done. With this change, you'll need to update all links that point to `apm-mutating-admission-webhook`. In addition, we'll want to set up a redirect to catch any external links. The obs-docs team can help with the redirects. I recommend doing a ctrl+f for `apm-mutating-admission-webhook` and updating this PR accordingly.

bmorelli25 (approved):

PR #11883, opened 2023-10-18
Make OTLP input return partial success response

carsonip:
## Motivation/summary

Make OTLP input return partial success response.

## Checklist

- [x] Update [CHANGELOG.asciidoc](https://github.com/elastic/apm-server/blob/main/CHANGELOG.asciidoc)
- [ ] Update [package changelog.yml](https://github.com/elastic/apm-server/blob/main/apmpackage/apm/changelog.yml) (only if changes to `apmpackage` have been made)
- [ ] Documentation has been updated

For functional changes, consider:
- Is it observable through the addition of either **logging** or **metrics**?
- Is its use being published in **telemetry** to enable product improvement?
- Have system tests been added to avoid regression?

PR #12276, opened 2023-12-22
add test for OTLP metrics partial success response

endorama:
## Motivation/summary

We recently introduced OTLP partial success response support in #11883. This PR adds a test case for OTLP metrics to ensure we are responding using a PartialSuccess.

Note that this test rely on triggering a partial response from the server by sending malformed Exponential Histogram data.

## Checklist

## How to test these changes

Run systemtests, verify that `TestOTLPGRPCMetrics_partialSuccess` passes.

## Related issues

Ref: https://github.com/elastic/apm-server/issues/12226


carsonip (approved):
thanks!

PR #12221, opened 2023-12-13
fix: always try to remap both span and error stacktraces

kruskall:
## Motivation/summary

When an event contains both a span and an error, the error is ignored by the sourcemap processor leading to stacktraces not being remapped.

## Checklist

- [ ] Update [CHANGELOG.asciidoc](https://github.com/elastic/apm-server/blob/main/CHANGELOG.asciidoc)
- [ ] Update [package changelog.yml](https://github.com/elastic/apm-server/blob/main/apmpackage/apm/changelog.yml) (only if changes to `apmpackage` have been made)
- [ ] Documentation has been updated

For functional changes, consider:
- Is it observable through the addition of either **logging** or **metrics**?
- Is its use being published in **telemetry** to enable product improvement?
- Have system tests been added to avoid regression?

## Related issues

Closes https://github.com/elastic/apm-server/issues/12132

PR #12390, opened 2024-01-10
Add 8.11.4 release notes

bmorelli25:
Add missing changelog for #12221


carsonip (commented):
Sorry we missed a changelog for a 8.12 PR which was backported to 8.11: https://github.com/elastic/apm-server/pull/12392

carsonip (commented):
I noticed that we have a "View commits" link in each section in 8.11 changelog, which does not exist in other changelogs. Do we want to align it with other versions?

bmorelli25:
> I noticed that we have a "View commits" link in each section in 8.11 changelog, which does not exist in other changelogs. Do we want to align it with other versions?

Oh. It looks like it was missed in the 8.10 file. It exists in all other versions. I can fix that in a different PR. This release already went out so I'd like to keep these changes separate.

carsonip (approved):

PR #12448, opened 2024-01-17
test: cleanup systemtest init

kruskall:
## Motivation/summary

we don't need to upload apmpackage so we can cleanup the init code. Avoid cleaning es during init, it should be cleaned in specific tests to support the -count flag.

## Checklist

### DIFF
--- a/NOTICE.txt
+++ b/NOTICE.txt
@@ -573,11 +573,11 @@ Contents of probable licence file $GOMODCACHE/github.com/elastic/apm-data@v1.3.0
 
 --------------------------------------------------------------------------------
 Dependency : github.com/elastic/beats/v7
-Version: v7.0.0-alpha2.0.20240612140227-60ba7bb46d03
+Version: v7.0.0-alpha2.0.20240613132816-f9fec1e4ff85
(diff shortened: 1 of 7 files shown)


v1v:
run docs-build

carsonip (commented):
> Avoid cleaning es during init, it should be cleaned in specific tests to support the -count flag.

which -count flag are you referring to?

axw:
> @axw do we still need to install the APM package for systemtests ? Do you think this makes sense now that we have the apm data plugin ?

I don't think we need it any more. The tests pass... :)
We're still using Fleet in ESS, and we have smoke tests there, so I think we're covered.

> Avoid cleaning es during init, it should be cleaned in specific tests to support the -count flag.

I don't follow this. If the individual tests still clean up ES, how will `-count` work?

axw:
> Individual tests cleaning up ES is the correct behaviour. If we rely on TestMain then it will be cleaned once when the first test run and every other iteration specified in -count won't clean ES.

Gotcha. Makes sense.

axw (approved):

PR #12451, opened 2024-01-18
Log number of bytes read from HTTP request body

axw:
## Motivation/summary

If Content-Length is not -1, report that in http.request.body.bytes, otherwise report the number of bytes read from the body. This can be useful for monitoring the rate of bytes ingested, even when agents use chunked encoding to stream data to the server.

## Checklist

## How to test these changes

1. Using `curl` send some requests to `/intake/v2/events`, with and without `Tranfer-Encoding: chunked`
2. Verify that the HTTP request logs report the request body size is reported correctly

## Related issues

None


marclop (approved):
Looks great!

simitt:
Tested with BC 5 on ESS:
sent a request via `curl` without -H "Content-type: application/x-ndjson" and one request with `-H "Content-type: application/x-ndjson"` to the apm-server. In both cases the number of bytes were logged at `http.request.body.bytes`.

PR #12453, opened 2024-01-18
TBS: Fix missing TTL after badger.ErrTxnTooBig

### LINKED ISSUE #12449: tbs code not setting entry TTL

TBS code is setting the entry without the TTL: https://github.com/elastic/apm-server/blob/03ec4426b2afffb789f03dd3abaa3846928d801c/x-pack/apm-server/sampling/eventstorage/storage.go#L209

This was spotted during the bugfix in https://github.com/elastic/apm-server/pull/12120#discussion_r1450628513 but was kept as is. We should investigate whether this is a bug.

**Steps to reproduce**:

look at it.

 1.
 2.
 3.

carsonip:
## Motivation/summary

Tail-based sampling: Fix a case where entry TTL was not set after encountering badger.TxnTooBig.

## Checklist

## How to test these changes

This is an edge case which is hard to test. The change should be easy enough to reason.

## Related issues

Closes #12449

### DIFF
--- a/x-pack/apm-server/sampling/eventstorage/storage.go
+++ b/x-pack/apm-server/sampling/eventstorage/storage.go
@@ -222,7 +222,7 @@ func (rw *ReadWriter) writeEntry(e *badger.Entry, opts WriterOpts) error {
 	}
 	rw.pendingSize += entrySize
 	rw.s.pendingSize.Add(entrySize)
-	return rw.txn.SetEntry(e)
+	return rw.txn.SetEntry(e.WithTTL(opts.TTL))
 }
 
 func estimateSize(e *badger.Entry) int64 {


simitt (approved):

carsonip (commented):
I intentionally did not add unit tests to this. I wrote one but it involves arbitrary numbers in badger configuration to specifically trigger ErrTxnTooBig on the n-th entry, flush the txn with the [0..n-1] entries, then add the n-th entry in a new txn. It is ugly.

carsonip:
Here's an incomplete test that will walk the concerned code path:
```
func TestErrTxnTooBig(t *testing.T) {
	tempdir := t.TempDir()
	opts := func() badger.Options {
		opts := badgerOptions()
		opts = opts.WithInMemory(false)
		opts = opts.WithDir(tempdir).WithValueDir(tempdir)
		opts = opts.WithMaxTableSize(10000)
		opts = opts.WithValueThreshold(1000)
		return opts
	}

	// Open and close the database to create a non-empty value log file,
	// which will cause writes below to fail due to the storage limit being
	// exceeded. We would otherwise have to rely on Badger's one minute
	// timer to refresh the size.
	db := newBadgerDB(t, opts)
	db.Close()
	db = newBadgerDB(t, opts)
	//lsm, vlog := db.Size()

	store := eventstorage.New(db, eventstorage.ProtobufCodec{})
	readWriter := store.NewReadWriter()
	defer readWriter.Close()

	traceID := uuid.Must(uuid.NewV4()).String()
	transactionID := uuid.Must(uuid.NewV4()).String()
	transaction := modelpb.APMEvent{Transaction: &modelpb.Transaction{Id: transactionID}}
	for i := 0; i < 100; i++ {
		err := readWriter.WriteTraceEvent(traceID, transactionID, &transaction, eventstorage.WriterOpts{
			TTL:                 time.Second,
			StorageLimitInBytes: 0,
		})
		assert.NoError(t, err)
	}
}
```

PR #12471, opened 2024-01-20
test: set memqueue min_events to 0 when using null output

kruskall:
## Motivation/summary

A null output is using a memqueue under the hood, buffering events and leading to delayed ack. On shutdown/close the Acker is hanging/waiting for the events to be acknowledged which happens after the timeout is hit.
This happens on multiple tests, particularly on TestServerCORS, causing a severe slowdown.

Set queue.mem.flush.min_events to 0 to flush immediately.

TestServerCORS time went from 30s to 0.07s

## Checklist

## How to test these changes

Run `go test -v ./internal/beater/. -run=TestServerCORS` before and after.

Observe test time going from 30s to ~0.05s


axw (approved):
Nice speedup, thank you!

PR #12508, opened 2024-01-29
feat: consume APM configuration over the elastic-agent control protocol

kruskall:
## Motivation/summary

Consume APM configuration over the elastic-agent control protocol. Pin elastic-agent-client version to 7.6.0 until beats is updated.

Create a new instrumentation package to retrieve the tracer dynamically. It might need some changes in go-docappender and other packages to allow updating the tracer.

## Checklist

## How to test these changes

Testing is a nightmare and can't be easily performed. We can't run elastic-agent locally and the ea-managed apm-server system test is broken.
Current approach is to start the ea-managed smoke test, manually install elastic-agent, upload apm-server binary to replace the apm-server component and restart elastic-agent.

### DIFF
--- a/internal/instrumentation/provider.go
+++ b/internal/instrumentation/provider.go
@@ -1,0 +1,161 @@
+// Licensed to Elasticsearch B.V. under one or more contributor
+// license agreements. See the NOTICE file distributed with
+// this work for additional information regarding copyright
+// ownership. Elasticsearch B.V. licenses this file to you under
+// the Apache License, Version 2.0 (the "License"); you may
+// not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package instrumentation
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"net"
+	"os"
+	"sync/atomic"
+
+	"go.elastic.co/apm/module/apmotel/v2"
+	"go.elastic.co/apm/v2"
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/sdk/metric"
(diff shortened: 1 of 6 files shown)


cmacknz:
> Testing is a nightmare and can't be easily performed. We can't run elastic-agent locally

Best solution to this is a VM, I like https://multipass.run/ especially if you are a Mac. Makes dealing with Ubuntu VMs as easy as containers.

internal/instrumentation/provider.go
```diff
@@ -0,0 +1,161 @@
...
+
+	ver := client.VersionInfo{
+		Name:    "apm-server",
+		Version: version.Version,
+		//BuildHash: version.CommitHash(),
+	}
+	c, _, err := client.NewV2FromReader(os.Stdin, ver)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create new v2 client: %w", err)
+	}
```
cmacknz, line 72: In libbeat we do this in https://github.com/elastic/beats/blob/main/x-pack/libbeat/management/managerV2.go#L176-L183
```go
		// Normal Elastic-Agent-Client initialisation
		agentClient, _, err = client.NewV2FromReader(os.Stdin, client.VersionInfo{
			Name:    "beat-v2-client",
			Version: version.GetDefaultVersion(),
			Meta: map[string]string{
				"commit":     version.Commit(),
				"build_time": version.BuildTime().String(),
			},
```

This is called during initialization from https://github.com/elastic/beats/blob/c5e79a25d05d5bdfa9da4d187fe89523faa42afc/libbeat/cmd/instance/beat.go#L837-L841
```go
	// initialize config manager
	b.Manager, err = management.NewManager(b.Config.Management, reload.RegisterV2)
	if err != nil {
		return err
	}
```

In APM Server I see this is already called in https://github.com/elastic/apm-server/blob/062c286636369a57967e9a75441963d3b5764beb/internal/beatcmd/beat.go#L152-L157

So what I suspect is happening is that you are reading an empty proto message from stdin because the one the agent wrote is already consumed. This is causing the Services field to be `nil` failing the check in https://github.com/elastic/elastic-agent-client/blob/17d2b257ff148cd86de88252359bef0f28df9ab8/pkg/client/reader.go#L86

The `io.ReadAll` must just be reading `EOF` if you aren't getting an error back, which seems weird but possible. 

Regardless of the actual mechanism causing this to fail you are trying to initialize the client a second time which is likely the source of your problems.

v1v:
run docs-build

Issue #12053, opened 2023-11-15
TBS: processor fails and stops when span is too large


simitt:
**APM Server version** (`apm-server version`):
`8.10` (and previous)

**Description of the problem including expected versus actual behavior**:
Under specific circumstances, the apm-server [`tail sampler` process](https://github.com/elastic/apm-server/blob/main/x-pack/apm-server/main.go#L101:L108) returns and stops without shutting down the apm-server. This seems to happen in relation to tail based sampling when a span cannot be deleted from the local storage due to its size (https://github.com/elastic/apm-server/blob/main/x-pack/apm-server/sampling/processor.go#L536). 
This manifests in an error message such as `failed to delete span from local storage: Txn is too big to fit into one request`. 
Right now the only way to mitigate is to manually restart the apm-server. 

To fix this we need to 
- already reject events that exceed a max size when they are received, as well as ensuring that events of allowed size can be processed by the system
- avoid situations where the processor stops running without stopping the apm-server.

PR #12509, opened 2024-01-29
fix: do not stop sampling processor when failing to delete trace events

kruskall:
## Motivation/summary

The sampling processor should never stop when apm-server is running. Instead log an error on Warn level and skip the current event.

Handle `ErrTxnTooBig` error when deleting trace events: because deleting adds a transaction it can increase the size above the limit. 
Flush the events before deleting.

## Checklist

## Related issues

Closes https://github.com/elastic/apm-server/issues/12053


simitt:
From the issue description:
>To fix this we need to

>already reject events that exceed a max size when they are received, as well as ensuring that events of allowed size can be processed by the system

This PR addresses the part of the github issue to not stop the TBS processor, but it does not address the problem that events exceeding the max size fill up storage without ever getting deleted. Events that are too large to be deleted, should be rejected already when received.

carsonip:
> already reject events that exceed a max size when they are received

There is the same `ErrTxnTooBig` check in `writeEntry`. Did the deletion txn get too large because writes are batched? Should we handle the ErrTxnTooBig thrown by flushing and retrying?

simitt:
>There is the same ErrTxnTooBig check in writeEntry, so we should already be doing it.

The [code](https://github.com/elastic/apm-server/blob/main/x-pack/apm-server/sampling/eventstorage/storage.go#L217-L225) ensures that an event is added to a new transaction if the max size of the write transaction would otherwise be exceeded. I overlooked that the event cannot be stored in the first place if the event itself is exceeding the max transaction size. 

>Did the deletion txn get too large because writes are batched? Should we handle the ErrTxnTooBig thrown by flushing and retrying?

Yes, that sounds like the only reason why this might fail. +1 on the suggested solution.

carsonip:
> I'm not sure what this means, can you elaborate ?

`DeleteTraceEvent` and `WriteTraceEvent` share the same `rw`. Imagine processor calls `WriteTraceEvent` multiple times right before reaching `ErrTxnTooBig`, then calls `DeleteTraceEvent`, which takes the txn right over the txn size limit. Since there is no `ErrTxnTooBig` handling inside `DeleteTraceEvent` unlike `writeEntry`, it will throw an error. I suspect this is the root cause.

To fix it, handle `ErrTxnTooBig` gracefully in `DeleteTraceEvent` like what we have in `writeEntry`

carsonip (approved):
lgtm

- Can you update the PR description?
- Do we want to backport this to 8.12, i.e. 8.12.2?

carsonip:
## Testing notes

Tested via https://github.com/elastic/apm-server/pull/12688 as it is almost impossible to manually test this behavior. Test in #12688 fails without the handling in this PR.

PR #12543, opened 2024-02-01
Fix dependabot PRs

reakaleek:
## Motivation/summary

Dependabot Pull Requests are not updating the NOTICE.txt file, which causes the lint job to fail.
This workflow will checkout the dependabot PR, update the NOTICE.txt file, and push the changes back to the PR.

## Checklist

### DIFF
--- a/.github/workflows/update-dependabot-pr.yml
+++ b/.github/workflows/update-dependabot-pr.yml
@@ -1,0 +1,41 @@
+---
+# Dependabot Pull Requests are not updating the NOTICE.txt file, which causes the lint job to fail.
+# This workflow will checkout the dependabot PR, update the NOTICE.txt file, and push the changes back to the PR.
+name: update-dependabot-pr
+
+on:
+  workflow_dispatch: ~ # This gives us the ability to run it for existing PRs where dependabot has already created a PR
+  push:
+    branches:
+      - dependabot/go_modules/**
+
+permissions:
+  contents: read
+
+jobs:
+  update-dependbot-pr:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: elastic/apm-pipeline-library/.github/actions/github-token@current
+        with:
+          url: ${{ secrets.VAULT_ADDR }}
+          roleId: ${{ secrets.VAULT_ROLE_ID }}
+          secretId: ${{ secrets.VAULT_SECRET_ID }}
+      - uses: elastic/apm-pipeline-library/.github/actions/setup-git@current
+        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/update-dependabot-pr.yml
```diff
@@ -0,0 +1,41 @@
...
+      - dependabot/go_modules/**
+
+permissions:
+  contents: read
+
+jobs:
+  update-dependbot-pr:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: elastic/apm-pipeline-library/.github/actions/github-token@current
```
reakaleek, line 19: @v1v 

I was also thinking.. maybe we could use the github actions token instead with

```
permissions:
  contents: write
```

But then, I guess, the tests won't run.

WDYT?

reakaleek: Oh wait. Maybe we actually want to use the github actions token.

pushing to the branch will cause the workflow to be triggered again and again.

v1v: In addition, using the `GitHub bot` will prevent the build from being rerun.

However, I wonder whether a changeset validation should be in place, so if changes other than `NOTICE`, then run the next steps.

What do you think?

reakaleek: Aha, 

we could do 

```
    paths-ignore:
      - NOTICE.txt
```

Then this check will only run if the `NOTICE.txt` file was not changed. Hence, this workflow won't run twice.

.github/workflows/update-dependabot-pr.yml
```diff
@@ -0,0 +1,43 @@
...
+          roleId: ${{ secrets.VAULT_ROLE_ID }}
+          secretId: ${{ secrets.VAULT_SECRET_ID }}
+      - uses: elastic/apm-pipeline-library/.github/actions/setup-git@current
+        with:
+          username: ${{ env.GIT_USER }}
+          email: ${{ env.GIT_EMAIL }}
+          token: ${{ env.GITHUB_TOKEN }}
+      - uses: actions/checkout@v4
+        with:
+          ref: ${{ inputs.branch }}
```
v1v, line 33: why?

reakaleek: woops.

https://github.com/elastic/apm-server/pull/12543/commits/6d12399494822c931d6773388522bf39d0a4b8d8

v1v (approved):

PR #12559, opened 2024-02-05
ci: ignore beats from dependabots update

kruskall:
## Motivation/summary

beats is using some special versioning and dependency bump is handled by updatecli.

Ignore beats from dependabotes update

## Checklist

## Related issues

See https://github.com/elastic/apm-server/pull/12539


axw (approved):
Ideally we'd just use Dependabot to update to the latest commit on the relevant branch, but it seems you can't do that: https://github.com/dependabot/dependabot-core/issues/2028

PR #12605, opened 2024-02-07
ci: add more dependabot groups

kruskall:
## Motivation/summary

spf13/cobra depends on spf13/pflag so group them together. otel jaeger translator depends on jaeger so group them together. otel consumer depends on otel pdata so group them together. golang.org/x/ dependencies tend to be bumped at the same time so group them.

This should reduce noise and duplicate dependabot PRs

## Checklist

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,7 +10,17 @@ updates:
     groups:
       otel:
         patterns:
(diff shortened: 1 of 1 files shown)


endorama (approved):
Looks good to me,

PR #12735, opened 2024-02-27
fix: set 8.12.2 version instead of 8.12.3

kyungeunni:
set version to 8.12.2 instead of 8.12.3 to fix smoke test


carsonip (commented):
we had to bump to 8.12.3 since we released 8.12.2. What's the problem with smoketests?

kyungeunni:
It wasn't able to find 8.12.3 snapshot till yesterday. It seems it's available now. closing this

PR #12862, opened 2024-03-25
[DOCS] Add openAPI specification for APM server information API

lcawl:
This PR creates an openAPI specification for documentation purposes to replace https://www.elastic.co/guide/en/observability/current/api-info.html

### DIFF
--- a/docs/spec/openapi/apm-openapi.yaml
+++ b/docs/spec/openapi/apm-openapi.yaml
@@ -1,0 +1,80 @@
+openapi: 3.1.0
+info:
+  title: APM Server
+  description: OpenAPI schema for APM Server APIs
+  version: 8.13.0
+  license:
+    name: Elastic License 2.0
+    url: https://www.elastic.co/licensing/elastic-license
+  contact:
+    name: APM Team
+servers:
+  - url: /
+tags:
+  - name: server info
+    description: APIs that query general server information.
+    x-displayName: APM server information
+paths:
+  /:
+    get:
+      summary: Retrieve general server information.
+      description: |
+        This lightweight endpoint is useful as a server up/down health check.
+        To configure authenticated access to the APM server, the instructions at [APM API key](https://www.elastic.co/guide/en/observability/current/api-key.html) or [APM Secret Token](https://www.elastic.co/guide/en/observability/current/secret-token.html) must be followed to configure the correct permissions for APM access.
+      operationId: getServerHealth
+      tags:
+        - server info
+      responses:
+        '200':
+          $ref: '#/components/responses/200_server_info'
+    post:
+      summary: Retrieve general server information.
+      operationId: postServerHealth
+      tags:
+        - server info
+      requestBody:
+        content:
+          application/json:
+            schema:
+              type: object
+              properties:
+                build_date:
+                  type: string
+                  format: time-date
+                  example: '2021-12-18T19:59:06Z'
+                build_sha:
+                  type: string
+                  example: 24fe620eeff5a19e2133c940c7e5ce1ceddb1445
+                publish_ready:
+                  type: boolean
+                  example: true
+                version:
+                  type: string
+                  example: 8.12.1
+      responses:
+        '200':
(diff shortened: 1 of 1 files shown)


docs/spec/openapi/apm-openapi.yaml
```diff
@@ -0,0 +1,80 @@
...
+                  example: '2021-12-18T19:59:06Z'
+                build_sha:
+                  type: string
+                  example: 24fe620eeff5a19e2133c940c7e5ce1ceddb1445
+                publish_ready:
+                  type: boolean
+                  example: true
+                version:
+                  type: string
+                  example: 8.12.1
```
axw, line 53: This information is in the response body, not the request body. This endpoint does not parse anything from the request body.

docs/spec/openapi/apm-openapi.yaml
```diff
@@ -0,0 +1,80 @@
...
+      type: http
+      scheme: bearer
+      bearerFormat: Secret token
+  responses:
+    200_server_info:
+      description: |
+        Indicates that the server is up.
+        If an API key or a secret token is passed along with the request, the response payload will include some information about the APM server.
+      content:
+        application/json:
```
axw, line 75: Is it possible with OpenAPI to convey that a response can have no content when the request is unauthenticated? I mean other than in the description.

lcawl: I've added a "oneOf" so that it's a bit clearer that either of the two responses are valid

axw: Only problem is, the empty content response does not respond with `Content-Type` - so I don't think the "application/json" bit above is correct in this case.

Not sure how much it matters - probably not worth worrying about?

axw (approved):

docs/spec/openapi/apm-openapi.yaml
lcawl: Is the example shown in the docs (without credentials) returned as a string? If so, I've drafted an another alternative in 445adc9d7044fbc1bb7c50fb4a424053c0ca95ea

axw: Nope, there's no content-type at all:

Unauthenticated (note no "Content-Type" header in the response):
```
$ curl -v http://localhost:8200
*   Trying 127.0.0.1:8200...
* Connected to localhost (127.0.0.1) port 8200 (#0)
> GET / HTTP/1.1
> Host: localhost:8200
> User-Agent: curl/7.81.0
> Accept: */*
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< X-Content-Type-Options: nosniff
< Date: Wed, 27 Mar 2024 02:01:29 GMT
< Content-Length: 0
<
```

Authenticated:
```
$ curl -H "Authorization: Bearer foo" -v http://localhost:8200
*   Trying 127.0.0.1:8200...
* Connected to localhost (127.0.0.1) port 8200 (#0)
> GET / HTTP/1.1
> Host: localhost:8200
> User-Agent: curl/7.81.0
> Accept: */*
> Authorization: Bearer foo
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< X-Content-Type-Options: nosniff
< Date: Wed, 27 Mar 2024 02:02:42 GMT
< Content-Length: 151
< 
{
  "build_date": "2024-03-20T19:10:39Z",
  "build_sha": "474ef34e6dc44f90b810ff7f31f9a20af172c46f",
  "publish_ready": false,
  "version": "8.14.0"
}
```

lcawl: Alas, then it doesn't seem like there's a supported option for capturing that.  There's a long thread on this sort of topic https://github.com/OAI/OpenAPI-Specification/issues/270 but no resolution that seems to work for our case. For now, I've reverted to just putting the info in the description.

axw: No worries, thanks for looking into it!

axw (approved):

Issue #12904, opened 2024-04-03
Fix automation, bring back openAPI spec


simitt:
The recently introduced openAPI spec (https://github.com/elastic/apm-server/pull/12862) by @lcawl has been removed by this automated [PR#12892](https://github.com/elastic/apm-server/pull/12892/) for updating beats. 

Identify why the automation deleted this file, fix the automation and bring back the openAPI spec.

Issue #12903, opened 2024-04-02
[DOCS] Create openAPI specifications for public APM APIs 


lcawl:
We need openAPI documents that are equivalent to these manually-maintained pages:

- [x] [APM server information API](https://www.elastic.co/guide/en/observability/current/apm-api-info.html): https://github.com/elastic/apm-server/pull/12862
- [x] [APM events intake API](https://www.elastic.co/guide/en/observability/current/apm-api-events.html) https://github.com/elastic/apm-server/pull/12982
- [x] [APM agent configuration API](https://www.elastic.co/guide/en/observability/current/apm-api-config.html) https://github.com/elastic/apm-server/pull/12948
- [x] [openTelemetry intake API](https://www.elastic.co/guide/en/observability/current/apm-api-otlp.html) https://github.com/elastic/apm-server/pull/13056
- [x] [jaeger event intake API](https://www.elastic.co/guide/en/observability/current/apm-api-jaeger.html) https://github.com/elastic/apm-server/pull/13056
- [x] Serverless-specific subset of APIs: https://github.com/elastic/apm-managed-service/pull/788

Progress is currently blocked until https://github.com/elastic/apm-server/issues/12904 is fixed.

Long-term, they'd ideally be generated automatically. In the near-term, they can be manually generated and maintained.

lcawl:
FYI I will wait for https://github.com/elastic/apm-server/issues/12904 to be fixed and the openAPI document to re-appear before starting on the next set.

PR #12948, opened 2024-04-09
[DOCS] Add openAPI specification for APM agent configuration endpoints

lcawl:
Relates to https://github.com/elastic/apm-server/issues/12903

This PR updates the OpenAPI specification to include details from https://www.elastic.co/guide/en/observability/current/apm-api-config.html

Related to the outstanding issue https://github.com/elastic/apm-server/issues/12904, this PR re-adds the original file from https://github.com/elastic/apm-server/pull/12862


docs/spec/openapi/apm-openapi.yaml
```diff
@@ -0,0 +1,162 @@
...
+        To configure authenticated access to the APM server, the instructions at [APM API key](https://www.elastic.co/guide/en/observability/current/api-key.html) or [APM Secret Token](https://www.elastic.co/guide/en/observability/current/secret-token.html) must be followed to configure the correct permissions for APM access.
+      operationId: getAgentConfig
+      tags:
+        - agent config
+      parameters:
+        - in: query
+          name: service.name
+          required: true
+          schema:
+            type: string
```
carsonip, line 54: According to https://github.com/elastic/apm/blob/main/specs/agents/configuration.md and https://github.com/elastic/apm-server/blob/main/internal/beater/api/config/agent/handler.go#L146, this may be missing `service.environment` param

docs/spec/openapi/apm-openapi.yaml
```diff
@@ -0,0 +1,162 @@
...
+        - agent config
+      requestBody:
+        content:
+          application/json:
+            schema:
+              type: object
+              required:
+                - service
+              properties:
+                CAPTURE_BODY:
```
carsonip, line 77: Q: what is `CAPTURE_BODY`?

docs/spec/openapi/apm-openapi.yaml
```diff
@@ -0,0 +1,162 @@
...
+      tags:
+        - agent config
+      requestBody:
+        content:
+          application/json:
+            schema:
+              type: object
+              required:
+                - service
+              properties:
```
carsonip, line 76: I think this is missing a few fields in https://github.com/elastic/apm-server/blob/main/internal/agentcfg/model.go#L55

docs/spec/openapi/apm-openapi.yaml
```diff
@@ -0,0 +1,162 @@
...
+                properties:
+                  capture_body:
+                   type: string
+                   example: off
+                  transaction_max_spans:
+                    type: integer
+                    example: 500
+                  transaction_sample_rate:
+                    type: number
+                    example: 0.3
```
carsonip, line 107: Is this supposed to be an exhaustive list? There are a lot more config options in https://github.com/elastic/kibana/blob/main/x-pack/plugins/observability_solution/apm/common/agent_configuration/setting_definitions/general_settings.ts

docs/spec/openapi/apm-openapi.yaml
```diff
@@ -0,0 +1,162 @@
...
+        '403':
+          description: APM Server is configured to fetch agent configuration from Elasticsearch but the configuration is invalid.
+        '503':
+          description: APM Server is starting up or Elasticsearch is unreachable.
+  /config/v1/rum/agents:
```
carsonip, line 112: I thought rum agentcfg endpoint is almost identical to the regular one, with the exception of auth check and the actual endpoint

carsonip:
And thanks for translating it from docs to openAPI spec. It really shows how much is missing from the existing docs.

docs/spec/openapi/apm-openapi.yaml
lcawl: It's only parity with what appears in the current docs, so definitely not an exhaustive list

docs/spec/openapi/apm-openapi.yaml
lcawl: Good question!  I omitted descriptions from the OpenAPI document for now since I didn't know the answer to questions like that, but we'd definitely want to add descriptions to make it more useable than our existing docs.

docs/spec/openapi/apm-openapi.yaml
lcawl: I've added `additionalProperties: true` to indicate there are more options than this currently covers. I've also pulled some descriptions from that file you linked, thanks!

docs/spec/openapi/apm-openapi.yaml
lcawl: I've added `additionalProperties: true` for now so that it indicates this isn't an exhaustive list.

docs/spec/openapi/apm-openapi.yaml
lcawl: For now, I've added a description based on what appears in https://github.com/elastic/kibana/blob/main/x-pack/plugins/observability_solution/apm/common/agent_configuration/setting_definitions/general_settings.ts

docs/spec/openapi/apm-openapi.yaml
lcawl: Thanks! Fixed in b78f0960d61d3c39b759b89f58452629395fb55d

docs/spec/openapi/apm-openapi.yaml
```diff
@@ -0,0 +1,162 @@
+openapi: 3.1.0
+info:
+  title: APM Server
+  description: OpenAPI schema for APM Server APIs
+  version: 8.13.2
```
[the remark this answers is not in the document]
lcawl: This property actually has nothing to do with the API or product version and just reflects the document version, so I've changed it to `0.1` for now (via a7d421c2c7f9563278647cb8690b46ee4dd02917).  We have an outstanding question about when/if we ought to increment that value, but from a docs point of view it seems to have no impact.

lcawl:
I think this PR can't be merged until https://github.com/elastic/apm-server/issues/12904 is fixed, since running "make fmt update" as recommended by the linter causes the file to be removed.

carsonip (approved):
let's merge on parity and revisit these suggestions later.

PR #12969, opened 2024-04-15
dra: fix explicit version

v1v:
## Motivation/summary

Fixes DRA packaging, I only changed the `.go-version`, but it required some further changes in the `docker file`. I'll work on honouring the `.go-version` in the `Dockerfile` in a follow-up

## Checklist

- [ ] Update [CHANGELOG.asciidoc](https://github.com/elastic/apm-server/blob/main/CHANGELOG.asciidoc)
- [ ] Documentation has been updated

For functional changes, consider:
- Is it observable through the addition of either **logging** or **metrics**?
- Is its use being published in **telemetry** to enable product improvement?
- Have system tests been added to avoid regression?

PR #12970, opened 2024-04-15
package: honour `.go-version` content when creating docker images

v1v:
## Motivation/summary

Avoid an explicit hardcoded value in the `Dockerfile` but honour the `.go-version` content. For such, the `GOLANG_VERSION` is created on the fly when running `make package*`

This helps reduce the steps when bumping the Golang version and the chances that something goes side-ways like it happened to me recently, and I needed to create a new follow-up:
- https://github.com/elastic/apm-server/pull/12969

I also enabled `make package-snapshot` as part of the CI to help detect breaking changes before merging. Otherwise, we only detect them when the DRA runs in Buildkite.

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -84,6 +84,6 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-go@v5
         with:
-          go-version-file: systemtest/go.mod
+          go-version-file: go.mod
           cache: true
       - run: make package-snapshot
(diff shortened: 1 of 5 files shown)


reakaleek (dismissed):

.github/workflows/ci.yml
```diff
@@ -77,3 +77,13 @@ jobs:
...
           name: test-results
           path: 'build/*.xml'
+
+  test-package:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-go@v5
+        with:
+          go-version-file: systemtest/go.mod
```
v1v, line 87:
```suggestion
          go-version-file: go.mod
```

PR #12987, opened 2024-04-17
fix: add rsync protect filter to preserve openapi spec

kruskall:
## Motivation/summary

we're syncing docs spec from apm-data which causes rsync to delete the openapi spec because it's not in the target dir (apm-data). This PR adds a protect filter to prevent the openapi folder from being deleted.

## Checklist

## How to test these changes

run `make docs/spec` 
openapi spec should still be there

## Related issues

Closes https://github.com/elastic/apm-server/issues/12904

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -332,3 +332,17 @@ pull_request_rules:
         labels:
           - "backport"
         title: "[{{ destination_branch }}] {{ title }} (backport #{{ number }})"
+  - name: backport patches to 8.14 branch
+    conditions:
+      - merged
+      - base=main
+      - label=backport-8.14
+    actions:
+      backport:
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -189,11 +189,6 @@ update-beats: update-beats-module update
 update-beats-module:
 	$(GO) get -d $(BEATS_MODULE)@$(BEATS_VERSION) && $(GO) mod tidy
 
-.PHONY: update-beats-docs
-update-beats-docs:
```
endorama, line 193: question: why this change? Does not seem related to the other change or with the linked issue to me, but I may lack the context. Could you clarify it?

endorama (approved):

PR #12999, opened 2024-04-18
buildkite: skip branch validation if UI build

v1v:
## Motivation/summary

Utilise the [BUILDKITE_SOURCE](https://buildkite.com/docs/pipelines/environment-variables#BUILDKITE_SOURCE) environment variable when running the DRA. If the Buildkite pipeline is triggered manually, this will skip the branch validation.

image

This is handy when the Unified Release process has not produced any artifacts for `main` yet, and `main` still points to the newly created minor branch. Normally it takes a few days to have new artifacts for `main` pointing to a new minor version.

https://github.com/elastic/apm-pipeline-library/blob/main/.ci/generate-snapshots.sh is the crazy script that generates the `branches.json` file.

## Checklist

### DIFF
--- a/.buildkite/scripts/dra.sh
+++ b/.buildkite/scripts/dra.sh
@@ -28,15 +28,21 @@ if [[ "${BUILDKITE_PULL_REQUEST:-false}" == "true" ]]; then
   exit 0
 fi
 
-curl -s https://storage.googleapis.com/artifacts-api/snapshots/branches.json > active-branches.json
-if ! grep -q "\"$BUILDKITE_BRANCH\"" active-branches.json ; then
-  echo "--- :arrow_right: Release Manager only supports the current active branches, skipping"
-  echo "BUILDKITE_BRANCH=$BUILDKITE_BRANCH"
-  echo "BUILDKITE_COMMIT=$BUILDKITE_COMMIT"
-  echo "VERSION=$VERSION"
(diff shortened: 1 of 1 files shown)


endorama:
Thank you for proposing this, but I would prefer not to merge it as it creates a UI only edge case that make the process harder to debug in case of issues. The behavior that led do this PR was unexpected but correct: no DRA was executed because we needed to wait upstream (technically we may not need to, but being a unified release I think it make sense to do so). 

We could address it with better documentation (on which I'm working) and having the script visibly fail/warn (not sure if this is possible) in Buildkite instead of skipping the step but looking like a success. What do you think?

v1v:
> We could address it with better documentation (on which I'm working) and having the script visibly fail/warn (not sure if this is possible) in Buildkite instead of skipping the step but looking like a success. What do you think?

I cannot see an easy way to skip a Buildkite build for a branch since it uses the `catalog-info.yml` file and does not allow runtime conditionals. For doing so, it's during the build time at the step level, see https://buildkite.com/docs/pipelines/conditionals

https://buildkite.com/docs/pipelines/skipping explains about the different options.

It will be possible to fail the build... but it's that what you want to? You will receive a Slack message saying it failed for every commit its merged. Unless we do something fancy with the slack notifications and its conditions.

reakaleek (approved):

Issue #12989, opened 2024-04-17
ci: minor-release automation is not aligned with expectations


endorama:
Release automation for minor releases does not:
- add `apm-` prefix to Asciidoc links and targets
- use tags instead of branches in compare links

A manual fix we applied for latest release was [`a0d4d2d` (#12984)](https://github.com/elastic/apm-server/pull/12984/commits/a0d4d2d60f292cf98d10c717cb903ec257c1e580), but we want to fix the automation so this is not a manual step.

simitt:
@cachedout could the eng.prod team help out with this and add it to your planning?

cachedout:
@simitt done (cc: @v1v )

PR #13065, opened 2024-05-02
ci(release-minor): add apm- prefix to Asciidoc links and targets

v1v:
## Motivation/summary

Release automation for minor releases does not:

* add apm- prefix to Asciidoc links and targets

That's caused by https://github.com/elastic/apm-server/pull/12820

## Checklist

## Related issues

Part of https://github.com/elastic/apm-server/issues/12989

### DIFF
--- a/release.mk
+++ b/release.mk
@@ -64,7 +64,7 @@ define CHANGELOG_TMPL
 [[release-notes-head]]
 == APM version HEAD
 
-https://github.com/elastic/apm-server/compare/$(RELEASE_BRANCH)\...main[View commits]
+https://github.com/elastic/apm-server/compare/$(RELEASE_VERSION)\...main[View commits]
 
 [float]
 ==== Breaking Changes
@@ -140,8 +140,8 @@ rename-changelog:
 	$(MAKE) common-changelog
 	@echo ">> rename-changelog"
 	echo "$$CHANGELOG_TMPL" > changelogs/head.asciidoc
-	@if ! grep -q 'release-notes-$(VERSION)' CHANGELOG.asciidoc ; then \
-		awk "NR==2{print \"* <<release-notes-$(VERSION)>>\"}1" CHANGELOG.asciidoc > CHANGELOG.asciidoc.new; \
+	@if ! grep -q 'apm-release-notes-$(VERSION)' CHANGELOG.asciidoc ; then \
+		awk "NR==2{print \"* <<apm-release-notes-$(VERSION)>>\"}1" CHANGELOG.asciidoc > CHANGELOG.asciidoc.new; \
 		mv CHANGELOG.asciidoc.new CHANGELOG.asciidoc ; \
 	fi
 	@if ! grep -q '$(VERSION).asciidoc' CHANGELOG.asciidoc ; then \
--- a/release.mk
+++ b/release.mk
@@ -64,7 +64,7 @@ define CHANGELOG_TMPL
 [[release-notes-head]]
 == APM version HEAD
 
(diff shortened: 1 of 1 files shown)


release.mk
```diff
@@ -165,7 +165,9 @@ common-changelog:
 	mv changelogs/head.asciidoc changelogs/$(VERSION).asciidoc
 	$(SED) 's#head#$(VERSION)#gI' changelogs/$(VERSION).asciidoc
 	$(SED) -E -e 's#(\...)main#\1$(VERSION)#g' changelogs/$(VERSION).asciidoc
-	awk "NR==5{print \"\n* <<release-notes-$(VERSION).0>>\n\n[float]\n[[release-notes-$(VERSION).0]]\n=== APM version $(VERSION).0\"}1" changelogs/$(VERSION).asciidoc > changelogs/$(VERSION).asciidoc.new
+	$(SED) 's#https://github.com/elastic/apm-server/compare/.*##gI' changelogs/$(VERSION).asciidoc
```
v1v, line 168: This sed won't remove the two empty lines. But if for anything is automated then changes should also be reflected in the `awk` since it uses a fixed line. (number 5)

release.mk
```diff
@@ -39,7 +39,7 @@ PROJECT_PATCH_VERSION ?= $(shell echo $(RELEASE_VERSION) | cut -f3 -d.)
 PROJECT_OWNER ?= elastic
 RELEASE_TYPE ?= minor
 
-CURRENT_RELEASE ?= $(shell gh api repos/elastic/apm-server/releases/latest | jq -r '.tag_name|sub("v"; ""; "")')
+CURRENT_RELEASE ?= $(shell gh release list --exclude-drafts --exclude-pre-releases --limit 10 --json tagName --jq '.[].tagName|select(. | startswith("v$(PROJECT_MAJOR_VERSION)"))' | sed 's|v||g' | sort -r | head -n 1)
```
v1v, line 42: We cannot use `latest` as long as latest can point to `7.17`. Already reported to the release team. 

image


I decided to use a different approach and filter by the major version and do some manipulation

release.mk
v1v: This is not the case anymore... I used a different approach, see https://github.com/elastic/apm-server/pull/13065/commits/55d1fbf24b2fb495850e2a169a748cc46fb9c551

therefore no more sed/awk craziness but templates

release.mk
```diff
@@ -157,16 +171,13 @@ update-changelog:
...
 common-changelog: VERSION=$${VERSION}
 common-changelog:
 	@echo ">> common-changelog"
-	mv changelogs/head.asciidoc changelogs/$(VERSION).asciidoc
-	$(SED) 's#head#$(VERSION)#gI' changelogs/$(VERSION).asciidoc
-	$(SED) -E -e 's#(\...)main#\1$(VERSION)#g' changelogs/$(VERSION).asciidoc
-	awk "NR==5{print \"\n* <<release-notes-$(VERSION).0>>\n\n[float]\n[[release-notes-$(VERSION).0]]\n=== APM version $(VERSION).0\"}1" changelogs/$(VERSION).asciidoc > changelogs/$(VERSION).asciidoc.new
-	mv changelogs/$(VERSION).asciidoc.new changelogs/$(VERSION).asciidoc
+	echo "$$CHANGELOG_MINOR_TMPL" > changelogs/$(VERSION).asciidoc
+	tail -n +6 changelogs/head.asciidoc >> changelogs/$(VERSION).asciidoc
```
v1v, line 180: as long as `head.asciidoc` uses the current format, we can use this approach, so less craziness commands to manipulate the file but using templates

reakaleek (approved):

endorama (approved):

PR #13103, opened 2024-05-07
updatecli: verify commits

v1v:
## Motivation/summary

Leverage https://github.com/updatecli/updatecli/issues/1914 and support [commit signature verifications](https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification)

## Checklist

- [ ] Update [CHANGELOG.asciidoc](https://github.com/elastic/apm-server/blob/main/CHANGELOG.asciidoc)
- [ ] Documentation has been updated

For functional changes, consider:
- Is it observable through the addition of either **logging** or **metrics**?
- Is its use being published in **telemetry** to enable product improvement?
- Have system tests been added to avoid regression?

## How to test these changes

I tested it in a playground:
- https://github.com/v1v/updatecli-demo/pull/43/commits

And also in my fork, for such I used this set of changes and tweaked a [bit](https://github.com/v1v/apm-server/pull/42) my main branch,

Then https://github.com/v1v/apm-server/pull/43/commits was created

## Related issues

Then if it works as expected in the CI, changes will be applied to the remaining updatecli manifests

PR #13107, opened 2024-05-08
updatecli: verify commits for golang and beats

v1v:
## Motivation/summary

Leverage https://github.com/updatecli/updatecli/issues/1914 and support [commit signature verifications](https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification)

## Checklist

## Related issues

Similar to https://github.com/elastic/apm-server/pull/13103

### DIFF
--- a/.github/workflows/bump-elastic-stack.yml
+++ b/.github/workflows/bump-elastic-stack.yml
@@ -39,4 +39,3 @@ jobs:
           command: '--experimental apply'
         env:
           BRANCH: ${{ matrix.branch }}
-          GITHUB_ACTOR: ${{ github.actor }}
--- a/.github/workflows/bump-golang.yml
+++ b/.github/workflows/bump-golang.yml
@@ -22,5 +22,3 @@ jobs:
           vaultSecretId: ${{ secrets.VAULT_SECRET_ID }}
           pipeline: ./.ci/bump-golang.yml
           command: '--experimental apply'
-        env:
-          GITHUB_ACTOR: ${{ github.actor }}
--- a/.github/workflows/bump-golang.yml
+++ b/.github/workflows/bump-golang.yml
@@ -21,3 +21,6 @@ jobs:
           vaultRoleId: ${{ secrets.VAULT_ROLE_ID }}
           vaultSecretId: ${{ secrets.VAULT_SECRET_ID }}
           pipeline: ./.ci/bump-golang.yml
+          command: '--experimental apply'
+        env:
+          GITHUB_ACTOR: ${{ github.actor }}
--- a/.github/workflows/update-beats.yml
+++ b/.github/workflows/update-beats.yml
@@ -34,11 +34,13 @@ jobs:
       - uses: elastic/apm-pipeline-library/.github/actions/updatecli@current
         env:
           BRANCH_NAME: ${{ matrix.branch }}
+          GITHUB_ACTOR: ${{ github.actor }}
         with:
           vaultUrl: ${{ secrets.VAULT_ADDR }}
           vaultRoleId: ${{ secrets.VAULT_ROLE_ID }}
           vaultSecretId: ${{ secrets.VAULT_SECRET_ID }}
           pipeline: ./.ci/update-beats.yml
+          command: '--experimental apply'
       - if: failure()
         uses: elastic/apm-pipeline-library/.github/actions/notify-build-status@current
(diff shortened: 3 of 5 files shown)


reakaleek (dismissed):

.github/workflows/update-beats.yml
```diff
@@ -34,11 +34,13 @@ jobs:
       - uses: elastic/apm-pipeline-library/.github/actions/updatecli@current
         env:
           BRANCH_NAME: ${{ matrix.branch }}
+          GITHUB_ACTOR: ${{ github.actor }}
         with:
           vaultUrl: ${{ secrets.VAULT_ADDR }}
           vaultRoleId: ${{ secrets.VAULT_ROLE_ID }}
           vaultSecretId: ${{ secrets.VAULT_SECRET_ID }}
           pipeline: ./.ci/update-beats.yml
+          command: '--experimental apply'
```
reakaleek, line 43: I'm wondering if it makes sense to set the `--experimental` flag in the action itself? Would it be backwards compatible?

.github/workflows/bump-golang.yml
```diff
@@ -21,3 +21,6 @@ jobs:
           vaultRoleId: ${{ secrets.VAULT_ROLE_ID }}
           vaultSecretId: ${{ secrets.VAULT_SECRET_ID }}
           pipeline: ./.ci/bump-golang.yml
+          command: '--experimental apply'
+        env:
+          GITHUB_ACTOR: ${{ github.actor }}
```
reakaleek, line 26: According to https://docs.github.com/en/actions/learn-github-actions/variables the `GITHUB_ACTOR` env is already set by default

.github/workflows/update-beats.yml
v1v: I was thinking to support a flag, but since command is a way to pass the different arguments to the updatecli binary, I think it's fine.

`--experimental` is a flag supported by the updatecli for quite some time, so I think it's backwards compatible. Or  what do you mean?

.github/workflows/bump-golang.yml
v1v: Good catch, I'm gonna try this in my playground and if it works then I'll remove it

.github/workflows/update-beats.yml
reakaleek: My thought was, if we set it by default in in the action, we don't have to set in every action usage.

reakaleek: Also, it might be easier to migrate in the future if it becomes "non-experimental"

v1v: No, `--experimental` feature by default it might produce some side-effects, see this [comment](https://github.com/updatecli/updatecli/issues/1914#issuecomment-2081151292); therefore I prefer to be explicit when to use it for the shake of reliability

reakaleek: I see. Thanks for the explanation.

.github/workflows/bump-golang.yml
v1v: Given https://github.com/v1v/apm-server/commit/6b2f1cd6ee49309614c10c0b990644736a7fde18
When I run the workflow
Then it uses the implicit env variable GITHUB_ACTOR

See https://github.com/v1v/apm-server/pull/44

v1v, line 26:
```suggestion
```

.github/workflows/update-beats.yml
```diff
@@ -34,11 +34,13 @@ jobs:
       - uses: elastic/apm-pipeline-library/.github/actions/updatecli@current
         env:
           BRANCH_NAME: ${{ matrix.branch }}
+          GITHUB_ACTOR: ${{ github.actor }}
```
v1v, line 37:
```suggestion
```

.github/workflows/bump-elastic-stack.yml
```diff
@@ -39,4 +39,3 @@ jobs:
           command: '--experimental apply'
         env:
           BRANCH: ${{ matrix.branch }}
-          GITHUB_ACTOR: ${{ github.actor }}
```
v1v, line 42: Remove unrequired env variable

reakaleek (approved):

endorama (approved):

PR #13113, opened 2024-05-08
fix: use Dialer DialContext to propagate context correctly

kruskall:
## Motivation/summary

When the elastic-agent-libs was updated the `Dialer` type added a `DialContext` method to comply with the stdlib interface.
However the code in apm-server is still using the old method.

## Checklist


endorama (approved):

endorama:
@kruskall may you clarify what is the benefit of using `DialContext` instead of `Dial` in the scope of the apm-server codebase?

PR #13137, opened 2024-05-10
build and push Docker image based on Chainguard base image

v1v:
## Motivation/summary

Release two flavours of Docker images:
- the one we always do
- the one based on [chainguard](https://www.chainguard.dev/unchained/introducing-wolfi-the-first-linux-un-distro) from [Chainguard](https://edu.chainguard.dev/chainguard/chainguard-images/overview/)

Please note that we are going to preserve the current `Dockerfile`, so that users will still be able to build their own custom images based on `ubuntu`: this is needed because `docker.elastic.co/wolfi/*` is not a public base image, so docker build would fail for unauthenticated users.

## Implementation details

I changed the Makefiles to support different Dockerfiles when running the docker build generation.
I added a tag for helping testing this in GitHub actions and then be able to push the docker images to an internal docker registry.
I fixed some warning regarding missing files when running the go build in the docker.

## Checklist

- [ ] Update [CHANGELOG.asciidoc](https://github.com/elastic/apm-server/blob/main/CHANGELOG.asciidoc)
- [ ] Documentation has been updated

For functional changes, consider:
- Is it observable through the addition of either **logging** or **metrics**?
- Is its use being published in **telemetry** to enable product improvement?
- Have system tests been added to avoid regression?

### CI

Will create the docker images automatically when working on a feature branch or main/releases

```shell
docker images docker.elastic.co/observability-ci/apm-server-internal
REPOSITORY                                               TAG                     IMAGE ID       CREATED          SIZE
docker.elastic.co/observability-ci/apm-server-internal   8.15.0-SNAPSHOT-wolfi   dce4529d1383   11 minutes ago   126MB
docker.elastic.co/observability-ci/apm-server-internal   8.15.0-SNAPSHOT         d2e84aceb911   12 minutes ago   209MB
```

### Locally

```shell
$ make package-snapshot
```

TBC what kind of tests should be done and how.

## Related issues

https://github.com/elastic/apm-server/pull/13139 will help with testing these changes in Buildkite without the need to be merged.
https://github.com/elastic/apm-server/pull/12671 is also another in-progress task

PR #13138, opened 2024-05-10
test conditionals

v1v:
test package without an env variable


v1v:
This was only for testing purposes. 

` test-package-and-push:` is now in place see, https://github.com/elastic/apm-server/pull/13137

PR #13139, opened 2024-05-10
buildkite(dra): support list for feature branches

v1v:
## Motivation/summary

Support feature branches to list DRA artifacts, since https://github.com/elastic/apm-server/pull/13001 introduced a regression when working on a feature branch.

## Checklist


.buildkite/scripts/dra.sh
```diff
@@ -37,30 +41,51 @@ if ! grep -q "\"$BUILDKITE_BRANCH\"" active-branches.json ; then
...
+    # to know if the branch was branched out from main or the release branches.
+    MAJOR_MINOR=${VERSION%.*}
+    if curl -s "https://storage.googleapis.com/artifacts-api/snapshots/main.json" | grep -q "$VERSION" ; then
+      DRA_BRANCH=main
+    else
+      if curl -s "https://storage.googleapis.com/artifacts-api/snapshots/$MAJOR_MINOR.json" | grep -q "$VERSION" ; then
+        DRA_BRANCH="$MAJOR_MINOR"
+      else
+        echo "It was not possible to know what's the original base branch. Let's stop here"
+        exit 0
```
endorama, line 59: question: My understanding is that exiting 0 here will report a success in buildkite, disregard this if it's not the case. As this looks like a fatal error, why is it exiting with `0` here?

v1v: A bit of context:

The DRA buildkite pipeline listens for events regarding different branches:
- https://github.com/elastic/apm-server/blob/8163c53d965b186f2b4d00615c3b144ad36da58d/catalog-info.yaml#L44

We enabled DRA for feature branches to test things before merging them into `main` or `[0-9].[0-9]`.

In this particular case, I want to support feature branches but without publishing any artifacts but listing them.

Unfortunately, feature branches do not match the name of the support DRA configurations (`main`, `7.17`, `8.14`); that's why I do some string manipulation to identify the base branch using the `VERSION` and some curls.

Regarding your question, since it's a feature branch and the name of the base branch could not be found, I decided to use `exit 0` rather than failing it. This is certainly an edge case, and I don't envision it going through that condition unless the Google bucket with the `<major>.<minor>.json` files is not accessible. As I said, all this logic is about supporting feature branches for DRA without failing.

reakaleek: Maybe an annotation or warning could be beneficial? Just an idea.

reakaleek (approved):

.buildkite/scripts/dra.sh
```diff
@@ -37,30 +41,51 @@ if ! grep -q "\"$BUILDKITE_BRANCH\"" active-branches.json ; then
...
+    # for such we will use the VERSION and https://storage.googleapis.com/artifacts-api/snapshots/<major.minor>.json
+    # to know if the branch was branched out from main or the release branches.
+    MAJOR_MINOR=${VERSION%.*}
+    if curl -s "https://storage.googleapis.com/artifacts-api/snapshots/main.json" | grep -q "$VERSION" ; then
+      DRA_BRANCH=main
+    else
+      if curl -s "https://storage.googleapis.com/artifacts-api/snapshots/$MAJOR_MINOR.json" | grep -q "$VERSION" ; then
+        DRA_BRANCH="$MAJOR_MINOR"
+      else
+        echo "It was not possible to know what's the original base branch. Let's stop here"
```
v1v, line 58:
```suggestion
        buildkite-agent annotate "It was not possible to know the original base branch for ${BUILDKITE_BRANCH}. This won't fail - this is a feature branch." --style 'info' --context 'ctx-info-feature-branch'
```

.buildkite/scripts/dra.sh
```diff
@@ -37,30 +41,51 @@ if ! grep -q "\"$BUILDKITE_BRANCH\"" active-branches.json ; then
   echo "VERSION=$VERSION"
   echo "Supported branches:"
   cat active-branches.json
-  buildkite-agent annotate "${BUILDKITE_BRANCH} is not supported yet. Look for the supported branches in ${BRANCHES_URL}" --style 'warning' --context 'ctx-warn'
-  exit 1
+  if [[ $BUILDKITE_BRANCH =~ "feature/" ]]; then
+    buildkite-agent annotate "${BUILDKITE_BRANCH} will list DRA artifacts. Feature branches are not supported. Look for the supported branches in ${BRANCHES_URL}" --style 'warning' --context 'ctx-warn'
```
v1v, line 45:
```suggestion
    buildkite-agent annotate "${BUILDKITE_BRANCH} will list DRA artifacts. Feature branches are not supported. Look for the supported branches in ${BRANCHES_URL}" --style 'info' --context 'ctx-info'
```

.buildkite/scripts/dra.sh
v1v: See https://github.com/elastic/apm-server/pull/13139/commits/4a326f57195433acd0173b0fc4fa67bffea331be

v1v:
This will help while working on https://github.com/elastic/apm-server/pull/13137 and the different follow-ups.

For instance, I detected at the very last moment that my changes would produce a failure when merged since the DRA in Buildktie will require a docker login. Unfortunately, I could not detect it until the very end when I found that I didn't change the Buildkite pipeline to run some docker login.

Therefore, if possible, I'd ask for some reviews 🙏

PR #13275, opened 2024-05-30
test: set flush_interval to 1ms for es output config

kruskall:
## Motivation/summary

`beatertest` is using a `flush_bytes` setting of 1 to avoid delay when flushing data but the setting has no effect because it's too small.
As a result `TestServerTracingEnabled` and `TestWrapServer` are waiting for the flush interval to kick in which defaults to 1s.
To fix the issue, set the `flush_interval` to 1ms to flush immediately and avoid delay. 
`TestWrapServer` test time went from 1.2s to 0.01
`TestServerTracingEnabled` test time went from 1.2s to 0.01

## Checklist

## How to test these changes

`make test`


axw:
> beatertest is using a flush_bytes setting of 1 to avoid delay when flushing data but the setting has no effect because it's too small.

Why is it too small? Shouldn't it trigger a flush as soon as 1 byte has been buffered?

marclop:
I haven't looked into it in detail, but it could be that compression is enabled by default and is not flushed to the underlying `bytes.Buffer`?

marclop:
I haven't looked into it in detail, but it could be that compression is enabled by default and is not flushed to the underlying `bytes.Buffer`?

axw (approved):

PR #13147, opened 2024-05-10
Test SNAPSHOT in smoketests

carsonip:
## Motivation/summary

Support "latest" in lagacy-managed and legacy-standalone-major-managed by testing snapshots.

## Checklist

- [ ] Update [CHANGELOG.asciidoc](https://github.com/elastic/apm-server/blob/main/CHANGELOG.asciidoc)
- [ ] Documentation has been updated

For functional changes, consider:
- Is it observable through the addition of either **logging** or **metrics**?
- Is its use being published in **telemetry** to enable product improvement?
- Have system tests been added to avoid regression?

## How to test these changes

```shell
(cd testing/smoke/legacy-standalone-major-managed && EC_API_KEY=redacted ./test.sh latest)
(cd testing/smoke/legacy-managed && EC_API_KEY=redacted ./test.sh latest)
```

## Related issues

Part of #13051

PR #13315, opened 2024-06-05
fix: use 7.17.22-SNAPSHOT instead of 7.17.22 temporarilly

kyungeunni:
## Motivation/summary

Temporarily use `7.17.22-SNAPSHOT` for smoke tests when the desired version is 7.17.22.

Previously the `7.17.22-SNAPSHOT` image was [broken](https://github.com/elastic/apm-server/pull/13147#issuecomment-2133906637) and now it's fixed. Now the `7.17.22` is broken for the same reason, and more recent build `7.17.22-SNAPSHOT` is now working. Until `7.17.22` is properly fixed, we should use 7.17.22-SNAPSHOT.

## How to test these changes

```
(cd testing/smoke/legacy-standalone-major-managed && EC_API_KEY=redacted ./test.sh 7.17)
(cd testing/smoke/legacy-managed && EC_API_KEY=redacted ./test.sh 7.17)
```

## Related issues
https://github.com/elastic/apm-server/pull/13147

### DIFF
--- a/testing/smoke/basic-upgrade/test.sh
+++ b/testing/smoke/basic-upgrade/test.sh
@@ -45,6 +45,18 @@ fi
 
 echo "-> Running basic upgrade smoke test for version ${VERSION}"
 
+# temporary fix to use SNAPSHOT image as 7.17.22 is broken
+if [ $LATEST_VERSION == "7.17.22" ]; then 
+    echo "using 7.17.22-SNAPSHOT instead of 7.17.22 temporarily"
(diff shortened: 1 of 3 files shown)


marclop:
I think it's not possible to upgrade a non-snapshot version deployment to a snapshot version. Did you check if this would work?

kyungeunni:
@marclop thanks for pointing this out, it doesn't indeed. 

Also, this temporary fix isn't necessary anymore given that there are new builds available. I will close this.

PR #13327, opened 2024-06-05
ubi: use ubi:9.4

v1v:
## Motivation/summary

Use the latest available ubi image based on 9.4 and available since [early May](https://repo1.dso.mil/dsop/redhat/ubi/9.x/ubi9-minimal/-/commit/dcb239d46b063c04579e3cd21477e505a0575492)

## Checklist


v1v:
> @v1v can we just publish the ubi image or do we need to append the version ?

We agreed to move away for `ubi-<major-version>` and use `ubi`.  https://github.com/elastic/apm-server/pull/11977 is the original one

reakaleek (approved):

PR #13359, opened 2024-06-07
smoketest: fix healthcheck to wait for the resource to be ready

kyungeunni:
## Motivation/summary
Healthcheck for smoke test parses API responses body to confirm that the APM Server is up and running before sending events. Currently, the condition is always true even when the `.publish_ready` isn't set to true, which makes the health check to be successful even the resource is not ready.

example of the case:
```
-> Waiting for 70 seconds for the APM Server to become available...
{"ok":false,"message":"The requested resource is currently unavailable."}
-> APM Server ready!
-> Sending events to APM Server...
{"ok":false,"message":"The requested resource is currently unavailable."}
```

The response didn't contain `.publish_ready` but the healthcheck passed.

This PR fixes the check to wait for the readiness correctly.

## Checklist

## How to test these changes
The successful response look like this:
```
{
  "build_date": "2024-06-06T00:00:00Z",
  "build_sha": "redacted",
  "publish_ready": true,
  "version": "8.13.0"
}
```

run this snippet script(copied from the lib.sh) locally with different RES:
```
healthcheck() {
    local PUBLISH_READY=$(echo ${RES}| jq '.publish_ready')
    if [[ ${PUBLISH_READY} != true ]]; then
        local MAX_RETRIES=10
        if [[ ${1} -gt 0 ]] && [[ ${1} -lt ${MAX_RETRIES} ]]; then
            echo "-> APM Server isn't ready to receive events, retrying (${1}/${MAX_RETRIES})..."
            # sleep $((1 * ${1}))
            # healthcheck $((1 + ${1}))
            return
        else
            echo "-> APM Server isn't ready to receive events, maximum retries exceeded"
            exit 1
        fi
    else
        echo "-> APM Server ready!"
    fi
}

healthcheck
```

It should print followings
❯ RES="{ }" ./test-healthcheck.sh
-> APM Server isn't ready to receive events, maximum retries exceeded

❯ RES="{\"publish_ready\": true }" ./test-healthcheck.sh
-> APM Server ready!

❯ RES="{\"publish_ready\": false }" ./test-healthcheck.sh
-> APM Server isn't ready to receive events, maximum retries exceeded


testing/smoke/lib.sh
```diff
@@ -181,7 +181,7 @@ healthcheck() {
     RES=$(curl_fail -H "${APM_AUTH_HEADER}" ${APM_SERVER_URL}) || RC=$?
     if [ $RC -ne 0 ]; then echo "${RES}"; fi
     local PUBLISH_READY=$(echo ${RES}| jq '.publish_ready')
-    if [[ ! ${PUBLISH_READY} ]]; then
+    if [[ ${PUBLISH_READY} != true ]]; then
```
endorama, line 184: I would suggest to explicitly use string comparison here:

```suggestion
    if [[ "${PUBLISH_READY}" != "true" ]]; then
```

In practice this behavior should not be affected, but it leaves the door open to nasty bugs (there is an interesting answer on StackOverflow detailing [some](https://stackoverflow.com/a/21210966)) especially if the server response changes at some point.

carsonip (approved):
nice catch, thanks!

testing/smoke/lib.sh
carsonip: even better!

kyungeunni: Thanks!

axw (approved):

PR #13398, opened 2024-06-11
feat: add require datastreams to bulk index requests

kruskall:
## Motivation/summary

Set require data stream for bulk index requests

## Checklist


axw (approved):
Can you please add a changelog entry? Could be useful in case there are any support issues related to this.

PR #13415, opened 2024-06-13
refactor: remove gimme and run_with_go_ver script

kruskall:
## Motivation/summary

Now that we have a toolchain directive explicitly defined in go.mod we can drop gimme and the shell script to ensure we are running the correct go version

## Checklist

### DIFF
--- a/NOTICE.txt
+++ b/NOTICE.txt
@@ -573,11 +573,11 @@ Contents of probable licence file $GOMODCACHE/github.com/elastic/apm-data@v1.3.0
 
 --------------------------------------------------------------------------------
 Dependency : github.com/elastic/beats/v7
-Version: v7.0.0-alpha2.0.20240613132816-f9fec1e4ff85
+Version: v7.0.0-alpha2.0.20240613222619-823943486f13
 Licence type (autodetected): Elastic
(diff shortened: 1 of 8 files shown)


go.mk
```diff
@@ -5,10 +5,9 @@ GITCOMMITTIMESTAMPUNIX ?= $(shell git log -1 --pretty=%ct)
...
 GITROOT ?= $(shell git rev-parse --show-toplevel)
 
-# Ensure the Go version in .go-version is installed and used.
-GOROOT?=$(shell $(GITROOT)/script/run_with_go_ver go env GOROOT)
+GOROOT?=$(go env GOROOT)
 GOLANG_VERSION=$(shell cat $(GITROOT)/.go-version)
-GO:=$(GOROOT)/bin/go
+GO:=go
 GOARCH:=$(shell $(GO) env GOARCH)
 export PATH:=$(GOROOT)/bin:$(PATH)
```
axw, line 12: I think most of this is unnecessary now? PATH must already contain the go binary, and not sure if we need GOROOT otherwise.

axw (approved):
:heart: thanks for the cleanup. I think we can get rid of a few more Makefile variables, otherwise LGTM

PR #8778, opened 2022-08-02
Introduce systemtest/cmd/fleetctl

axw:
## Motivation/summary

Add a basic CLI for listing Fleet package policies, and updating their config vars, for use in development and testing. This may be useful for manipulating config vars that are not shown in the APM integration policy editor.

Listing policies: `fleetctl -u <kibana_url> list-policies`

Updating vars: `fleetctl -u <kibana_url> set-policy-var <ID> tail_sampling_storage_limit=30MB`

## Checklist

~- [ ] Update [CHANGELOG.asciidoc](https://github.com/elastic/apm-server/blob/main/CHANGELOG.asciidoc)~
~- [ ] Update [package changelog.yml](https://github.com/elastic/apm-server/blob/main/apmpackage/apm/changelog.yml) (only if changes to `apmpackage` have been made)~
~- [ ] Documentation has been updated~

## How to test these changes

N/A

## Related issues

None

PR #13440, opened 2024-06-19
refactor: remove fleetctl cmd

kruskall:
## Motivation/summary

We are no longer using fleet for systemtests or local dev so the utility is unused.

## Checklist


systemtest/cmd/fleetctl/main.go
```diff
@@ -1,187 +0,0 @@
...
-	"fmt"
-	"log"
-	"net/url"
-	"reflect"
-	"strings"
-
-	"github.com/spf13/cobra"
-	"gopkg.in/yaml.v3"
-
-	"github.com/elastic/apm-server/systemtest/fleettest"
```
endorama, line 31: is `fleettest` still used? Should we look into removing it too? I noticed that part of it has been introduced with `cmd/fleetctl` in https://github.com/elastic/apm-server/pull/8778

(Maybe we should try with [deadcode](https://pkg.go.dev/golang.org/x/tools/internal/cmd/deadcode)?

endorama (approved):
left a question but can be addressed in a follow-up. Looks good to me

PR #13464, opened 2024-06-20
TBS: Optimize for ReadTraceEvents miss

### LINKED ISSUE #13463: High memory usage and disk IO when tail-based smapling is enabled
**APM Server version** (`apm-server version`): confirmed on 8.13.2 but affects all versions including the latest 8.14.1

When tail-based sampling (TBS) is enabled, the memory usage will go as high as the local TBS database storage size. When viewing `/proc/meminfo`, most of the memory usage shows up as "Mapped". This is particularly noticeable in setups which consist of multiple apm-servers and receive high load.

**Steps to reproduce**:

look at it.

 1. start 2 separate apm-servers (calling them A and B) and send load to them independently using apmsoak. Wait until their local TBS database size grows to >1GB.
 2. stop A, wait for 10mins, then restart A without sending load to it.
 3. observe that the A's memory usage increases to approximately to its local TBS database size

carsonip:
## Motivation/summary

When a sampling decision is received from a remote apm-server, if the trace is not stored locally, prefetching values in the iterator is causing mmap vlog to be read for no value, resulting in high memory usage and disk read IO.

Add an initial pass with PrefetchValues=false to optimize for a miss in ReadTraceEvents.

e.g. if there are 10 apm-servers all running TBS and synced via ES, and all transactions/spans from a trace are always sent to in 1 apm-server (i.e., a single trace never stored across multiple apm-servers), this PR should cut disk read IO related to sampling decision handling to 1/10 of before.

### Benchmark results

Since the benchmark runs badger DB with in-memory mode, I expect in reality, `hit=false after this fix` will stay the same, while all the other numbers to be 1 order of magnitude slower. i.e. the performance gain from this PR will be even greater.

```
goos: linux
goarch: amd64
pkg: github.com/elastic/apm-server/x-pack/apm-server/sampling/eventstorage
cpu: 11th Gen Intel(R) Core(TM) i7-11800H @ 2.30GHz
                                       │ bench-read-events-hit.before │     bench-read-events-hit.after     │
                                       │            sec/op            │    sec/op     vs base               │
ReadEventsHit/bigTX=true/hit=false-16                   167.06µ ± 14%   14.31µ ± 32%  -91.43% (p=0.002 n=6)
ReadEventsHit/bigTX=true/hit=true-16                     174.4µ ±  4%   186.3µ ±  4%   +6.81% (p=0.004 n=6)
ReadEventsHit/bigTX=false/hit=false-16                  129.49µ ±  3%   16.93µ ± 21%  -86.92% (p=0.002 n=6)
ReadEventsHit/bigTX=false/hit=true-16                    136.5µ ±  3%   139.8µ ±  3%   +2.40% (p=0.026 n=6)
geomean                                                  150.7µ         50.13µ        -66.73%
```

## Checklist

~~- [ ] Update [CHANGELOG.asciidoc](https://github.com/elastic/apm-server/blob/main/CHANGELOG.asciidoc)~~ Require backport release changelog

## How to test these changes

See benchmark

## Related issues

Closes #13463


x-pack/apm-server/sampling/eventstorage/storage.go
```diff
@@ -263,7 +263,23 @@ func (rw *ReadWriter) ReadTraceEvents(traceID string, out *modelpb.Batch) error
 	rw.readKeyBuf = append(append(rw.readKeyBuf[:0], traceID...), ':')
 	opts.Prefix = rw.readKeyBuf
 
+	// 1st pass: check whether there exist keys matching the prefix.
+	// Do not prefetch values so that the check is done in-memory.
+	// This is to optimize for cases when it is a miss.
```
vikmik, line 268: What is prefetched in case of a miss? Are you implying that badger prefetches stuff that is not relevant to the iterator?

(adding this for future reference)
https://github.com/dgraph-io/badger/blob/f5a7e0a1c83b7db39381e6e528ae23f21a4dd04f/iterator.go#L709

axw (dismissed):

x-pack/apm-server/sampling/eventstorage/storage.go
carsonip: > Are you implying that badger prefetches values that are not relevant to the iterator?

Exactly. The line you highlighted, `i.Valid`, `i` is `it.ittr`. `it.ittr.Valid` does not perform prefix checking like `it.Valid`.

carsonip: > What is prefetched in case of a miss?

100 values of irrelevant keys that don't match the supplied key prefix. I _think_ it should be considered a bug in badger v2. Still in process of checking if it is already fixed.

axw (approved):

x-pack/apm-server/sampling/eventstorage/storage.go
carsonip: While I don't think v4 fixed it, the memory / disk profile changed and I didn't dig into it. https://github.com/carsonip/tbs-badger-playground/tree/badgerv4/prefetch

At the very least, this PR should work on both v2 and v4. It will bring a significant improvement to cases where there's a miss, with little to no impact to happy path (hit).

PR #13523, opened 2024-06-28
fix: only retry when receiving a 429 status code

kruskall:
## Motivation/summary

The only safe status code to retry is 429 because it guarantee that the document has not been indexed. Everything else might create duplicate documents

## Checklist


carsonip (commented):
this likely warrants a changelog

1pkg (approved):

kyungeunni (approved):

PR #13555, opened 2024-07-03
8.14.2: add missing changelog entry

endorama:
Intially only merged on 8.14 branch, should have been merged on
main and backported.

Also add a missing changelog entry


carsonip (changes requested):
should only be merged when the version is released

carsonip (approved):
got confirmation from docs team: it is good to merge on release day or D-1.

PR #13571, opened 2024-07-03
updatecli: add backport labels

v1v:
## Motivation/summary

Add GitHub labels for the backports in branches other than `7.17`.
This will help with automating the backport label when creating PRs targeting the main for bumping the `golang` version.

We only support the latest minor version; if there are two active minor versions, then only the latest.

## Checklist


.ci/bump-golang.yml
```diff
@@ -1,6 +1,6 @@
 ---
 name: Bump golang-version to latest version
-pipelineid: 'updatecli-bump-golang'
+pipelineid: 'updatecli-bump-golang-{{ requiredEnv "GITHUB_BRANCH" }}'
```
v1v, line 3: To avoid reusing the same git  branch

PR #12604, opened 2024-02-07
feat: bump apm-data and go-docappender to stable version 

kruskall:
## Motivation/summary

Future updates should be picked up by dependabot

## Checklist

- [ ] Update [CHANGELOG.asciidoc](https://github.com/elastic/apm-server/blob/main/CHANGELOG.asciidoc)
- [ ] Documentation has been updated

For functional changes, consider:
- Is it observable through the addition of either **logging** or **metrics**?
- Is its use being published in **telemetry** to enable product improvement?
- Have system tests been added to avoid regression?

## Related issues

Closes https://github.com/elastic/apm-server/issues/12268

PR #13576, opened 2024-07-04
fix: Update `FlushBytes` parsing/defaults

### LINKED ISSUE #13024: Perf regression, too many small bulk requests
8.13.2

I believe there is a regression in the number of bulk requests/s emitted by APM Server towards Elastic. I don't have all data for it, but I believe the issue started in 8.12.x and then got worse with the upgrade to 8.13.x.

This is from the Elastic cloud console: On otherwise identical clients/agents, we see that with the upgrade from 8.12 to 8.13 the number of indexing requests/s jumped noticeably.

Screenshot_2024-04-16_at_23 59 50

The number of documents/s that get ingested didn't really change much, but our hot nodes started to run under higher load. Fortunately, we observed that setting `index.translog.durability=async` on the all APM indices drops IOPS per hot node back to acceptable levels:  

Screenshot_2024-04-21_at_21 35 36

On a decently tuned `bulk_max_size` setting I'd not expect such a big drop in IOPS/s. I am thus wondering if the APM server is misconfigured in newer versions. 

It almost feels like the server uses the agent defaults instead of its custom bulk size option?  Was it maybe forgotten to set `preset: custom` into the server config once https://github.com/elastic/elastic-agent/issues/3797 got merged into the agent?

marclop:
## Motivation/summary

Updates the `FlushBytes` setting to default to 1 mib and only override to 24kb if the user has explicitly set it to 24kb.

## Checklist

## How to test these changes

1. Spin up an APM Server and send data to it (without this fix).
2. Look at the request length for Elasticsearch requests and verify they're smaller than 50kb and very close to 24kb (26kb)
3. Now start apm sever with the code in this PR and send data to it
4. Verify that the flush size is significantly higher than before and if you sent enough data it should be ~1MB.

## Related issues

Fixes #13024

### DIFF
--- a/changelogs/head.asciidoc
+++ b/changelogs/head.asciidoc
@@ -9,6 +9,7 @@ https://github.com/elastic/apm-server/compare/8.14\...main[View commits]
 - Avoid data race due to reuse of `bytes.Buffer` in ES bulk requests {pull}13155[13155]
 - APM Server now relies on the Elasticsearch apm-data plugin's index templates, which reverts some unsafe uses of `flattened` field types {pull}12066[12066]
 - Add `error.id` to jaeger errors {pull}13196[13196]
+- Fix a performance regression for apm servers which don't specify `output.elasticsearch.flush_bytes` {pull}13576[13576]
 
 [float]
 ==== Breaking Changes
--- a/internal/beater/beater.go
+++ b/internal/beater/beater.go
@@ -633,7 +633,6 @@ func (s *Runner) newFinalBatchProcessor(
 	newElasticsearchClient func(cfg *elasticsearch.Config) (*elasticsearch.Client, error),
 	memLimit float64,
 ) (modelpb.BatchProcessor, func(context.Context) error, error) {
-
 	monitoring.Default.Remove("libbeat")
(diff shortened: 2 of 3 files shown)


marclop:
I committed the code without setting the default value for FlushBytes and the tests fail (so we can verify this bug fix) https://github.com/elastic/apm-server/actions/runs/9789313060/job/27028818438?pr=13576#step:4:173. I'll commit the fix in the next commit.

StephanErb:
Thank you! In case there is another 8.14 release, would it be possible to get this backportet?

axw (approved):
:scream: good find @marclop!

marclop:
We're still missing the `backport-8.15` label, I'll re-tag the PR once it's created.

inge4pres:
great catch @marclop 👏🏼

carsonip:
https://github.com/elastic/apm-server/pull/12604 was the PR that introduced the regression

v1v:
> We're still missing the backport-8.15 label, I'll re-tag the PR once it's created.


Let me solve this

marclop:
> Thank you! In case there is another 8.14 release, would it be possible to get this backportet?

@StephanErb We're looking at releasing a patch version `8.14.3` by the end of next week.

PR #13620, opened 2024-07-08
output: Retry document-level `429`s by default

### LINKED ISSUE #13619: Enable document-level retries by default
## Description

APM Server should retry 429s when Elasticsearch returns 429 status code in _bulk requests, rather than just logging the 429s, they should be retried.

https://github.com/elastic/go-docappender/blob/e702455c6fff497bf332d55347fb8f9e78dd729d/config.go#L71-L72

marclop:
## Motivation/summary

Updates the APM Server to automatically retry document-level `429`s from Elasticsearch to avoid dropping data. It can be configured/overwritten by `output.elasticsearch.max_retries`, and defaults to `3`.

It uses the default backoff configuration, which could wait up to 1m if enough retries are configured, but can be overwritten as well.

This reduces data loss whenever Elasticsearch is overhelmed and reduces the chances of data loss due to ES push-back.

## Checklist

## How to test these changes

This may need a bit of setup. It's thoroughly tested in the `go-docappender`, but we may want to verify that retries happen when document-level 429s happen.

I'd suggest writing a small "proxy" that returns 429s randomly and test that things are indexed (as duplicates).

## Related issues

Closes #13619

### DIFF
--- a/changelogs/head.asciidoc
+++ b/changelogs/head.asciidoc
@@ -18,4 +18,4 @@ https://github.com/elastic/apm-server/compare/8.15\...main[View commits]
 [float]
 ==== Added
 
-- APM Server now automatically retries document-level 429s from Elasticsearch to avoid dropping data. This is controlled by `output.elasticsearch.max_retries`, and defaults to `3`. {pull}13620[13620]
+- APM Server now automatically retries document-level 429s from Elasticsearch to avoid dropping data. `output.elasticsearch.max_retries` now controls both request-level and document-level retries, and defaults to `3`. {pull}13620[13620]
--- a/changelogs/head.asciidoc
+++ b/changelogs/head.asciidoc
@@ -17,3 +17,5 @@ https://github.com/elastic/apm-server/compare/8.15\...main[View commits]
 
 [float]
 ==== Added
+
+- APM Server now automatically retries document-level 429s from Elasticsearch to avoid dropping data. This is controlled by `output.elasticsearch.max_retries`, and defaults to `3`. {pull}13620[13620]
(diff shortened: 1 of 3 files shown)


carsonip (commented):
Retrying on document level 429 is the default is RetryOnDcoumentStatus is empty: https://github.com/elastic/go-docappender/blob/main/config.go#L76

marclop:
@carsonip yes, I'm aware. I wanted to make it explicit to avoid any potential breaking changes going forward. It's also clearer when reading the code what will happen.

carsonip (commented):
I thought this PR is a no-op, but I was wrong. `MaxDocumentRetries` was 0 and it is now non-zero. Changes lgtm but it will need a changelog.

changelogs/head.asciidoc
```diff
@@ -16,4 +16,6 @@ https://github.com/elastic/apm-server/compare/8.15\...main[View commits]
 ==== Intake API Changes
 
 [float]
-==== Added
\ No newline at end of file
+==== Added
+
+- APM Server now automatically retries document-level 429s from Elasticsearch to avoid dropping data. This is controlled by `output.elasticsearch.max_retries`, and defaults to `3`. {pull}13620[13620]
```
carsonip, line 21: nit: make it clear that `output.elasticsearch.max_retries` controls both request-level and document-level retries.
```suggestion
- APM Server now automatically retries document-level 429s from Elasticsearch to avoid dropping data. `output.elasticsearch.max_retries` now controls both request-level and document-level retries, and defaults to `3`. {pull}13620[13620]
```

carsonip (approved):

PR #13652, opened 2024-07-09
Rename APM Server APIs

bmorelli25:
### Summary

Renames `APM Server` API to `Observability Intake APIs`.

This PR is for https://github.com/elastic/docs-projects/issues/222.

I propose we merge these changes so we can go live with our new API docs (shoutout to @lcawl!) with a name that makes more sense publicly than `APM Server`. We can easily change this name in the future if product wants to call this set of APIs something else.


1pkg (approved):

lcawl (approved):
Thanks for the update!

PR #13666, opened 2024-07-11
Bump up version to 8.14.4

inge4pres:
Release of 8.14.3 today requires a new patch version to be prepared.


marclop (approved):

inge4pres:
Re-ran CI because only windows tests were failing, maybe a flaky one (🤞🏼 )

PR #13515, opened 2024-06-28
8.14.3: update versions

obltmachine:
Merge on request.

PR #13848, opened 2024-08-08
Bump up version to 8.15.1

inge4pres:
Release of 8.15.0 today requires bumping the versions to 8.15.1


carsonip (approved):
change lgtm but we have an automation to do this like https://github.com/elastic/apm-server/pull/13515

PR #13851, opened 2024-08-08
docs: clarify automation in release workflow

inge4pres:
## Motivation/summary

During the 8.15 release cycle we noticed some differences between what documented as needed and the available tooling for release.
This Pr is an attempt to clarify what's documented.


dev_docs/RELEASES.md
```diff
@@ -72,7 +72,9 @@ For patch releases, only the version on the existing major and minor version bra
 
 * A new [tag](https://github.com/elastic/apm-server/releases) will automatically be created on GitHub.
 
-* Bump the version in anticipation of the next release, e.g. [after 8.13.3 release](https://github.com/elastic/apm-server/pull/13066) bump to 8.13.4. **Prepare this PR ahead of time** but only merge it once pinged by the Release Manager on release date.
+* Merge the PRs created by the GitHub workflow ran as part of the steps in the ["Day after feature freeze"](#day-after-feature-freeze) section. 
+ > [!IMPORTANT]
+ > Only merge the PRs once pinged n Slack by the Release Manager on release date in the #mission-control channel
```
carsonip, line 77:
```suggestion
 > Only merge the PRs once pinged on Slack by the Release Manager on release date in the #mission-control channel
```

dev_docs/RELEASES.md
```diff
@@ -72,7 +72,9 @@ For patch releases, only the version on the existing major and minor version bra
 
 * A new [tag](https://github.com/elastic/apm-server/releases) will automatically be created on GitHub.
 
-* Bump the version in anticipation of the next release, e.g. [after 8.13.3 release](https://github.com/elastic/apm-server/pull/13066) bump to 8.13.4. **Prepare this PR ahead of time** but only merge it once pinged by the Release Manager on release date.
+* Merge the PRs created by the GitHub workflow ran as part of the steps in the ["Day after feature freeze"](#day-after-feature-freeze) section.
```
carsonip, line 75: This only applies to patch release, not minor release, right?

endorama: I agree and this needs to be clarified as there are a lot of PRs involved in this process and as we discovered recently other PRs (docs + version bumps) must not be merged before release date.

inge4pres: What do we need to do to clarify? Reach out to the robots?

> This only applies to patch release, not minor release, right?

Good point, thanks for the callout. Should we make a distinction like we do above for minor/patch release?
E.g. say: "for minor releases, merge this and that; for patch releases, merge this other and that other PR".

carsonip:
@inge4pres Do the changes still apply? Do you want to hand this PR over to @1pkg ?

dev_docs/RELEASES.md
```diff
@@ -17,15 +17,15 @@ For patch releases, only the version on the existing major and minor version bra
 
 * Trigger release workflow manually
   * For **patch releases**: run the [`run-patch-release`](https://github.com/elastic/apm-server/actions/workflows/run-patch-release.yml) workflow (In "Use workflow from", select `main` branch. Then in "The version", specify the **upcoming** patch release version - es: on `8.14.2` feature freeze you will use `8.14.2`).
-    This workflow will: create the release branch; update version across codebase; commit and create PR targeting the release branch.
-    Release notes for patch releases **must be manually added** (PR should target `main` branch and backported to the release branch):
+    This workflow will: create the `update-<VERSION>` branch, update version constants across the codebase and create a PR targeting the release branch. 
+    Release notes for patch releases **must be manually added** (PR should target `main` branch and should be backported to the release branch):
```
1pkg, line 21: I would add here **must be manually added a day before release**

1pkg (approved):

dev_docs/RELEASES.md
1pkg: I think the old bullet point was correct, no? Basically on the release day the next patch version bump has to happen, everything else is supposed to already be done by the day after feature freeze or the day before release.

dev_docs/RELEASES.md
carsonip: Teams do it on either D-1 or release day. I think it is fine to specify it as D-1 here for clarity.

dev_docs/RELEASES.md
inge4pres: It depends how we structure the flow, I think.
The old doc assumed the version bump PR is created on FF, and merged on release day.
I don't see why a time distance between the 2 events should exist, and I see more risk in creating and merging a PR for version bump _before_ we are requested to do so (as we would produce a DRA version that is too early).
That's why I added the "IMPORTANT" section below, but I also agree that the previous phrasing was more correct.
I'll amend that

inge4pres:
> how the preview renders the important block

works with latest commit

1pkg (dismissed):

carsonip (dismissed):

PR #13880, opened 2024-08-14
feat: bump go to 1.23.0

kruskall:
## Motivation/summary

use go get to update go versions in go.mod
drop go requirement from readme

the go requirement in the readme is less relevant since the introduction of go toolchains.
It's also not accurate becaure we're requiring a patch version.

## Checklist


README.md
```diff
@@ -17,7 +17,7 @@ To get started with APM, see our [Quick start guide](https://www.elastic.co/guid
 
 ### Requirements
 
-* [Go][golang-download] 1.22.x
+* [Go][golang-download]
```
endorama, line 20: What do you think about mentioning the `.go-version` file here without an explicit reference to the version?

```suggestion
* [Go][golang-download] (prefer installing the version specified in `.go-version`)
```

endorama (approved):
A minor nitpick, otherwise :+1:

1pkg (approved):

Issue #13859, opened 2024-08-09
Introduce profile-guided optimization to APM Server


1pkg:
# Context

Go compiler added support for profile-guided optimization (PGO) since Go 1.20, thus allowing to build further optimized Go binaries using CPU profiles. Depending on the workload, optimizations and the profile quality one can expect to extract somewhere from 2-14% of performance gains according to [go.dev/blog](https://go.dev/doc/pgo) which is quite meaningful. Further more for CPU bound services, in general, the expected results should be somewhere in the upper bound of the expected performance gains.

# Requested Changes

We should capture this performance improvement in APM Server.

We should consider how we will collect the profiles, initially we could leverage existing [benchmarks](https://github.com/elastic/apm-server/actions/workflows/benchmarks.yml) workflow to collect CPU profiles which will be included into the future PGO builds.

We should consider how we will store profiles and include them into the builds, the simplest and most convenient approach to commit them directly in the source code  for easy distribution and repeatable builds.

# Additional Links

[go.dev/doc/pgo](https://go.dev/doc/pgo)
[go.dev/blog/pgo](https://go.dev/blog/pgo)

1pkg:
The existing benchmarks in APM Server are highly coupled to the ElasticSearch stack, making them over indexing ES performance into the benchmark results. This could lead to a large distribution between individual benchmark results, up to 20%, and make them hard to measure and reason about. This is a particular problem for PGO, since PGO only adds an incremental gain to the performance. In my synthetical local test when benchmarking an isolated instance of APM Server the result was an average 5% increase in the throughput. These results are really hard to observe when the existing benchmark hide them between ES performance.

The solution to this problem would be to use a separate second set of benchmark workflow that targets a http API stub instead of real ES instance. This way we should be able to sufficiently isolate APM Server performance from underlying ES performance.

PR #13884, opened 2024-08-15
PGO: update the existing benchmarks workflow to enable PGO builds

1pkg:
## Motivation/summary

This PR implements changes outlined in https://github.com/elastic/apm-server/issues/13859. It updates the existing benchmarks workflow to run standalone APM Server instance that produces a relevant CPU profile for PGO, then it copies, uploads and injects the obtained CPU profile into a PR, see [example](https://github.com/elastic/apm-server/pull/14084).

## Benchmarks

The existing benchmarks results turned to be too unreliable to base PGO on. Because of the underlying dependency on ElasticSearch the difference in the throughput results could go above 10% from a workflow to workflow. The table below provides a view with the existing benchmarks results sample.

image

This all renders incremental PGO performance gains hard to observe and measure. Therefore, in this PR a new benchmark mode is introduced, which swaps ElasticSearch with a stubbed API http server (Moxy). Thus allowing us to better isolate and elevate APM Server performance component inside the benchmarks. The table below provides a view with the new isolated benchmarks results sample.

image

Using the benchmarks result sample data we can clearly observe that the results deviation for the new benchmark mode is in an order of magnitude lower in comparison to the existing ES based benchmarks. And now PGO performance improvements could be reliably observed.

The standalone APM Server benchmarks mode consists of running 3 separate EC2 instances in a VPC for apmbench, apm-server and moxy. Existing `benchmark_executor` and `standalone_apm_server` terraform modules are reused and a similar new terraform module `moxy` is created.

## Results

PGO enabled builds show 5% performance gain on average across the standalone APM Server benchmarks workflow.

## Checklist

## How to test these changes

To observe and validate the changes please refer to the indexed PGO benchmarks results.

## Related issues

#13859


inge4pres:
the only doubt I have is o the CI part, not sure how to test it

1pkg:
> the only doubt I have is on the CI part, not sure how to test it

Github Actions are problematic to test. AFAIK, the simplest way to test them is to just run them in Github directly, which is really manual and less than ideal. I know about few projects that can help reduce the "pain":
- https://github.com/nektos/act that allows running Github Actions locally; it falls short for a complex actions like https://github.com/elastic/apm-server/actions/workflows/benchmarks.yml.
- https://github.com/sethvargo/go-githubactions that allows creating Github Actions in go; I never used it in the pass, but it looks promising. The main problem with using go-githubactions, is that it's sufficiently different from all our existing Github Actions.

inge4pres (dismissed):
🚀 ship it!
looking forward to see it in action

marclop (commented):
I didn't review in detail, just posting a half review with the top files reviewed.

.github/workflows/benchmarks.yml
```diff
@@ -118,13 +132,24 @@ jobs:
...
       - name: Run benchmarks self tuned
         if: ${{ inputs.benchmarkAgents != '' }}
-        run: make run-benchmark index-benchmark-results
+        run: make run-benchmark
+
+      - name: Cat standalone server logs
+        if: ${{ inputs.runStandalone && failure() }}
+        run: make cat-apm-server-logs
+
+      # Results are only indexed and uploaded if the run happens on the main branch.
```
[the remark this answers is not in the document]
1pkg: Agree it's useful sometimes to benchmark non "main" branch like I did in this very PR. We also can easily filter the results in the dashboard directly, so there is no concern about intermixing the results.

testing/benchmark/main.tf
```diff
@@ -45,7 +45,46 @@ locals {
...
 }
 
+module "vpc" {
+  source  = "terraform-aws-modules/vpc/aws"
+  version = "3.14.0"
+
+  name = "${var.user_name}-worker"
+  cidr = var.vpc_cidr
+
+  azs                = [for letter in ["a", "b", "c"] : "${var.worker_region}${letter}"]
```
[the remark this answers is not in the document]
1pkg: Good catch, I agree 1 AZ should be good enough here, updated.

.github/workflows/benchmarks.yml
```diff
@@ -38,12 +44,14 @@ jobs:
       run:
         working-directory: ${{ env.WORKING_DIRECTORY }}
     permissions:
-      contents: read
+      contents: write
       id-token: write
+      pull-requests: write
```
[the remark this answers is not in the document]
1pkg: Yes, we need `pull-requests: write` since my current idea is to raise a PR for every benchmark PGO update so we can quickly eyeball it and confirm that the source benchmarks that created it was not a failure. After we get some confidence with this approach I'd move forward with auto merging PGO.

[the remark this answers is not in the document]
1pkg: See the current workflow action [here](https://github.com/elastic/apm-server/pull/13884/files#diff-22eef62c9d7b36d02f5fd5aaada92702e22d6795c464239cb7b8fe0f26ea1e1cR195) that requires the access.

testing/benchmark/Makefile
```diff
@@ -110,13 +127,25 @@ index-benchmark-results: _default-gobench-vars
 
 .PHONY: _default-gobench-vars
 _default-gobench-vars:
+ifeq ($(RUN_STANDALONE),true)
+	$(eval GOBENCH_DEFAULT_TAGS = $(GOBENCH_DEFAULT_TAGS),apm_server_size=$(shell echo var.standalone_apm_server_instance_size | terraform console | tr -d '"'))
+	$(eval GOBENCH_DEFAULT_TAGS = $(GOBENCH_DEFAULT_TAGS),moxy_size=$(shell echo var.standalone_moxy_instance_size | terraform console | tr -d '"'))
+	$(eval GOBENCH_DEFAULT_TAGS = $(GOBENCH_DEFAULT_TAGS),build_sha=$(shell git rev-parse HEAD))
+	$(eval GOBENCH_DEFAULT_TAGS = $(GOBENCH_DEFAULT_TAGS),bench_mode=standalone)
```
marclop, line 134: How come we aren't populating?
- `elasticsearch_size`
- `apm_server_zone_count`
- `elasticsearch_zone_count`

testing/benchmark/system-profiles/16GBx2zone.tfvars
```diff
@@ -10,5 +16,8 @@ elasticsearch_zone_count = 2
 apm_server_size = "16g"
 # Number of shards for the ES indices
 apm_shards = 4
-# Benchmarks executor size executor
-worker_instance_type = "c6i.2xlarge"
+
+# Standalone
+
+standalone_apm_server_instance_size = "c6i.4xlarge"
+standalone_moxy_instance_size       = "c6i.8xlarge"
```
marclop, line 23: Wow these are pretty big and beefy. Maybe we could add a comment here. Also looking at the size, `c6i.4xlarge` seems to be a 32GB instance, not 16GB. Is that what you intended?

testing/benchmark/system-profiles/32GBx2zone.tfvars
```diff
@@ -12,5 +18,8 @@ elasticsearch_dedicated_masters = true
 apm_server_size = "32g"
 # Number of shards for the ES indices
 apm_shards = 4
-# Benchmarks executor size executor
-worker_instance_type = "c6i.2xlarge"
+
+# Standalone
+
+standalone_apm_server_instance_size = "c6i.8xlarge"
```
marclop, line 24: Same here, 8xlarge has 64GB of RAM.

testing/infra/terraform/modules/standalone_apm_server/outputs.tf
```diff
@@ -5,6 +5,11 @@ output "apm_secret_token" {
 }
 
 output "apm_server_url" {
-  value       = "${aws_instance.apm.public_ip}:${local.apm_port}"
+  value       = "http://${aws_instance.apm.public_ip}:${local.apm_port}"
```
marclop, line 8: I wonder if it's important at all to test with HTTPS here.

marclop (commented):
mostly everything LGTM. Just a few questions we can merge soon after confirming things are as expected.

testing/benchmark/system-profiles/16GBx2zone.tfvars
1pkg: good catch, the sizing is fixed now

testing/benchmark/system-profiles/32GBx2zone.tfvars
1pkg: good catch, the sizing is fixed now

testing/benchmark/Makefile
1pkg: For Moxy based benchmarks there is no ES to speak of, we can add fake value for zone count=1 to keep the consistent tagging, if we think it's a good idea.

testing/infra/terraform/modules/standalone_apm_server/outputs.tf
1pkg: It's debatable but I think https shouldn't really matter here, since it shouldn't qualitatively change the resulting profile.

testing/benchmark/Makefile
marclop: oh of course there's no Elasticsearch. Ignore my comment then.

marclop (dismissed):
Thanks Kostya! Looks good now :)

1pkg:
Had to make some final changes to address some regression in shared modules for smoke-tests.
Now everything works as expected and aligned with the smoke tests results from the main branch.

See the workflow runs:

https://github.com/elastic/apm-server/actions/runs/11076905822
https://github.com/elastic/apm-server/actions/runs/11076854721
https://github.com/elastic/apm-server/actions/runs/11078611143

With this, this PR should be good to merge finally! Waiting for the final approvals + review from the robots team.

v1v:
Is this automation creating cloud resources but not deleting them? I see some leftovers in 

image


that's the AWS account used for the CI - we separated the Cloud accounts from dev/ci, that's the reason you cannot access to that one

1pkg:
> Is this automation creating cloud resources but not deleting them? I see some leftovers in
> 
> image
> that's the AWS account used for the CI - we separated the Cloud accounts from dev/ci, that's the reason you cannot access to that one

@v1v initially there was an unrelated authentication problem in the benchmarks pipeline that caused the issue, see for example https://github.com/elastic/apm-server/actions/runs/10567550442/job/29276727544. Because I was experimenting a lot at the same time, I accidentally created multiple VPC resources that were not properly cleaned. This long got fixed since then. And now this pipeline works as expected.

Please take a closer look at the actual changes so we can merge them to the upstream, thank you.

.github/workflows/benchmarks.yml
```diff
@@ -101,9 +110,23 @@ jobs:
           terraform_version: 1.3.7
           terraform_wrapper: false
 
+      - name: Init terraform module
+        run: |
+          AWS_REGION=${{ env.AWS_REGION }}
+          echo "TF_VAR_worker_region=$AWS_REGION" >> "$GITHUB_ENV"
+          echo "TF_VAR_run_standalone=$RUN_STANDALONE" >> "$GITHUB_ENV"
```
v1v, line 117: If these env variable are only used within the context of this step, you can use the `env`


```yaml
      - name: Init terraform module
        run: make init
        env:
          TF_VAR_worker_region: ${{ env.AWS_REGION }}
          TF_VAR_run_standalone: ${{ env.RUN_STANDALONE }}
```

Otherwise, do you think you can set them in the top level env? `L54`

.github/workflows/benchmarks.yml
```diff
@@ -101,9 +110,23 @@ jobs:
...
+
       - name: Build apmbench
         run: make apmbench $SSH_KEY terraform.tfvars
 
+      - name: Build APM Server and Moxy
+        run: |
+          if [ "$RUN_STANDALONE" = true ]; then
+            make apm-server
+            make moxy
+          fi
```
v1v, line 128: The condition might help to know if the step has been triggered or no easily:

```suggestion
        if: env.RUN_STANDALONE == 'true'
        run: |
          make apm-server
          make moxy
```

.github/workflows/benchmarks.yml
```diff
@@ -118,11 +141,21 @@ jobs:
...
         if: ${{ inputs.benchmarkAgents != '' }}
-        run: make run-benchmark index-benchmark-results
+        run: make run-benchmark
+
+      - name: Cat standalone server logs
+        if: ${{ failure() }}
+        run: |
+          if [ "$RUN_STANDALONE" = true ]; then
+            make cat-apm-server-logs
+          fi
```
v1v, line 155: step native condition

```suggestion
        if: ${{ failure() && env.RUN_STANDALONE == 'true' }}
        run: make cat-apm-server-logs
```

.github/workflows/benchmarks.yml
```diff
@@ -150,15 +183,50 @@ jobs:
 
       - name: Upload benchmark result
         uses: actions/upload-artifact@v4
-        if: always()
```
v1v, line 153: What's the reason it's not always?

.github/workflows/benchmarks.yml
```diff
@@ -150,15 +183,50 @@ jobs:
...
+        uses: actions/upload-artifact@v4
+        with:
+          name: cpu-profile
+          path: ${{ env.WORKING_DIRECTORY }}/${{ env.BENCHMARK_CPU_OUT }}
+          if-no-files-found: error
+
+      - name: Open PGO PR
+        if: ${{ github.ref == 'refs/heads/main' }}
+        run: |
+          if [ "$RUN_STANDALONE" = true ]; then
```
v1v, line 208: ditto.

OTOH, do you think this snippet could be added to the Makefile?

.github/workflows/benchmarks.yml
```diff
@@ -150,15 +183,50 @@ jobs:
...
+            git checkout -b "$BRANCH"
+            mv "$PROFILE_PATH" x-pack/apm-server/default.pgo
+            git add x-pack/apm-server/default.pgo
+            git commit -m "PGO: Update default.pgo from benchmarks $WORKFLOW."
+            git push -u origin "$BRANCH"
+            gh pr create -B main -H "$BRANCH" -t "PGO: Update default.pgo" -b "Update default.pgo CPU profile from the benchmarks [workflow]($WORKFLOW)." -R elastic/apm-server
+          fi
+        env:
+          PROFILE_PATH: ${{ env.WORKING_DIRECTORY }}/${{ env.BENCHMARK_CPU_OUT }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
v1v, line 224: This will not trigger a build when the PR is created. Unfortunately, GitHub does not trigger builds if the actor is a `github-actions bot`. For such, there are two alternatives:

1) Use PAT or finer-grained tokens
2) Or a GitHub app to create ephemeral tokens.


`1)` is not anymore what we use, but `2)`, you can use the GitHub app as done in https://github.com/elastic/apm-server/pull/14217/files#diff-68bec17d560c218f68b50a1e84ac73026a82e64927e706cffff69501cd151970R60-R83

testing/infra/terraform/modules/standalone_apm_server/variables.tf
```diff
@@ -33,50 +43,20 @@ variable "stack_version" {
...
+  description = "Optionally use the apm-server binary from the specified path instead"
+}
+
 variable "tags" {
   type        = map(string)
   default     = {}
   description = "Optional set of tags to use for all deployments"
 }
-
-# CI variables
```
v1v, line 60: What's the reason for deleting these labels?

v1v (commented):
I reviewed the GitHub workflow and TF variables, the other bits and pieces are not something I can say much though. I left a few comments

.github/workflows/benchmarks.yml
1pkg: This step sets variables for terraform, I'm not 100% sure if `terraform init` is the step than binds the ENV to variables or it happens during `terraform apply` let me double check this and update it if it works in `terraform init`.

.github/workflows/benchmarks.yml
1pkg: Unfortunately, Github Actions do no support condition on env variables natively. There is a workaround to pass it as a step output, let me try to check if it works better.

.github/workflows/benchmarks.yml
1pkg: Good point, will move this to Makefile

testing/infra/terraform/modules/standalone_apm_server/variables.tf
1pkg: This module didn't use those variables, so I decided to replace it unified `tags` variable instead.

.github/workflows/benchmarks.yml
1pkg: Updated via steps output now. Also apparently TF binds ENVs to variables at `terraform apply`.

.github/workflows/benchmarks.yml
1pkg: Updated via steps outputs now.

.github/workflows/benchmarks.yml
1pkg: updated

.github/workflows/benchmarks.yml
1pkg: @v1v thank you for pointing to this, I updated the workflow and it uses the same plumbing to setup all necessary pieces to prepare the github config and the access token. Now the runs work as expected see [link](https://github.com/elastic/apm-server/actions/runs/11134770091/job/30943589134) and the raised [PR](https://github.com/elastic/apm-server/pull/14229) has properly signed commits.

1pkg:
Final results after feedback from @v1v to properly set github access token.

The PGO standalone benchmark workflow run [link](https://github.com/elastic/apm-server/actions/runs/11134770091/job/30943589134) -> resulted in the next PGO update PR [link](https://github.com/elastic/apm-server/pull/14229).

While the old benchmark against ES cloud works as expected without regression [link](https://github.com/elastic/apm-server/actions/runs/11134115634/job/30941670041).

v1v (approved):

testing/benchmark/main.tf
```diff
@@ -33,6 +33,7 @@ module "tags" {
   # use the convention for team/shared owned resources if we are running in CI.
   # assume this is an individually owned resource otherwise.
   project = startswith(var.user_name, "benchci") ? "benchmarks" : var.user_name
+  build   = var.BUILD_ID
```
axw, line 36: This is also in ci_tags above, should we remove it from there?

.github/workflows/benchmarks.yml
```diff
@@ -150,15 +178,65 @@ jobs:
...
+        id: get_token
+        uses: tibdex/github-app-token@3beb63f4bd073e61482598c45c71c1019b59b73a # v2.1.0
+        with:
+          app_id: ${{ secrets.OBS_AUTOMATION_APP_ID }}
+          private_key: ${{ secrets.OBS_AUTOMATION_APP_PEM }}
+          permissions: >-
+            {
+              "contents": "write",
+              "pull_requests": "write"
+            }
```
axw, line 210: Why do we need this? Can't we add `pull_requests: write` to the workflow's permissions?

Is it because we are creating a PR that triggers another workflow?

.github/workflows/benchmarks.yml
```diff
@@ -150,15 +178,65 @@ jobs:
...
+          git_commit_gpgsign: true
+
+      - name: Open PGO PR
+        if: ${{ env.RUN_STANDALONE == 'true' && github.ref == 'refs/heads/main' }}
+        run: make push-pgo-pr
+        env:
+          WORKSPACE_PATH: ${{ github.workspace }}
+          PROFILE_PATH: ${{ env.WORKING_DIRECTORY }}/${{ env.BENCHMARK_CPU_OUT }}
+          GITHUB_TOKEN: ${{ steps.get_token.outputs.token }}
+          WORKFLOW: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}/attempts/${{ github.run_attempt }}
```
axw, line 235: I wonder if instead of creating a new PR on every benchmark run, could we just push a commit to the branch?

axw (dismissed):
Nice work - thank you for all the cleanups along the way!

.github/workflows/benchmarks.yml
v1v: > Is it because we are creating a PR that triggers another workflow?

Correct, as per the [documentation](https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#using-the-github_token-in-a-workflow), 'When you use the repository's GITHUB_TOKEN to perform tasks, events triggered by the GITHUB_TOKEN, with the exception of workflow_dispatch and repository_dispatch, will not create a new workflow run.'

https://github.com/elastic/apm-server/pull/13884#discussion_r1783310922 asked to change it

testing/benchmark/Makefile
```diff
@@ -149,6 +178,20 @@ cleanup-elasticsearch:
...
+push-pgo-pr:
+	@cd $(WORKSPACE_PATH); \
+	git fetch origin main; \
+	git checkout main; \
+	git checkout -b $(PGO_BRANCH); \
+	mv $(PROFILE_PATH) x-pack/apm-server/default.pgo; \
+	git add x-pack/apm-server/default.pgo; \
+	git commit -m "PGO: Update default.pgo from benchmarks $(WORKFLOW)."; \
+	git push -u origin $(PGO_BRANCH); \
+	gh pr create -B main -H $(PGO_BRANCH) -t "PGO: Update default.pgo" -b "Update default.pgo CPU profile from the benchmarks [workflow]($(WORKFLOW))." -R elastic/apm-server
```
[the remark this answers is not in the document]
v1v: That's me who I asked for moving the logic to a Makefile in https://github.com/elastic/apm-server/pull/13884#discussion_r1783309017

testing/benchmark/outputs.tf
```diff
@@ -4,38 +4,46 @@ output "public_ip" {
...
-  value       = module.ec_deployment.apm_url
+  value       = var.run_standalone ? module.standalone_apm_server[0].apm_server_url : module.ec_deployment[0].apm_url
   description = "The APM Server URL"
+  sensitive   = true
+}
+
+output "apm_server_ip" {
+  value       = var.run_standalone ? module.standalone_apm_server[0].apm_server_ip : ""
+  description = "The APM Server EC2 IP address"
+  sensitive   = true
```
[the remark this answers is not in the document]
1pkg: good catch, marked by a mistake, updated

testing/benchmark/Makefile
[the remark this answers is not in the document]
1pkg: Moved to .ci/scripts/ now

testing/benchmark/main.tf
1pkg: yes, it's a leftover, updated

.github/workflows/benchmarks.yml
1pkg: I think it's slightly risky to enable auto pushes to main branch right away, I'd prefer to start with more controlled PR based approach so we develop the confidence that this pipeline works well. Afterwards we can simplify and enable the direct merge.

1pkg: I also added a small update to the push-pgo-pr script so it enables auto merging too for PRs. This way we will only need to give it 1 approval and the pipeline tests need to pass.

axw: Fair enough.

axw (approved):
LGTM, thank you!

Issue #13957, opened 2024-08-30
APM Agents configuration cache refresh appears to be broken


up2neck:
**APM Server version** (`apm-server version`): 8.14.3 (linux/amd64), standalone

**Description of the problem including expected versus actual behavior**:

APM Server makes invalid search/scroll requests to ".apm-agent-configuration" Elasticsearch index:

![Image](https://github.com/user-attachments/assets/6ab9c6a7-50a9-4b88-b38a-453bc8abc6e4)

![Image](https://github.com/user-attachments/assets/e3ebaafd-bb14-4db4-9a1d-176fd1b7e69f)

![Image](https://github.com/user-attachments/assets/1b695a92-c1d5-4e05-a62f-f20f8cb49a4a)

<details>
  <summary>Invalid transaction JSON</summary>

```
{
  "@timestamp": [
    "2024-08-30T09:11:00.105Z"
  ],
  "agent.name": [
    "go"
  ],
  "agent.version": [
    "2.6.0"
  ],
  "cloud.availability_zone": [
    ""
  ],
  "cloud.instance.id": [
    ""
  ],
  "cloud.instance.name": [
    ""
  ],
  "cloud.project.id": [
    ""
  ],
  "cloud.provider": [
    ""
  ],
  "cloud.region": [
    ""
  ],
  "data_stream.dataset": [
    "apm"
  ],
  "data_stream.namespace": [
    "default"
  ],
  "data_stream.type": [
    "traces"
  ],
  "destination.address": [
    "elasticsearch-v1-es-http.apm-sandbox.svc"
  ],
  "destination.port": [
    9200
  ],
  "event.agent_id_status": [
    "missing"
  ],
  "event.ingested": [
    "2024-08-30T09:11:01.000Z"
  ],
  "event.outcome": [
    "failure"
  ],
  "event.success_count": [
    0
  ],
  "host.architecture": [
    "amd64"
  ],
  "host.hostname": [
    "apm-server-v1-apm-server-68576fc56-6pkms"
  ],
  "host.name": [
    "apm-server-v1-apm-server-68576fc56-6pkms"
  ],
  "host.os.platform": [
    "linux"
  ],
  "http.response.status_code": [
    400
  ],
  "labels.project": [
    "epm-paas"
  ],
  "observer.hostname": [
    "apm-server-v1-apm-server-68576fc56-6pkms"
  ],
  "observer.type": [
    "apm-server"
  ],
  "observer.version": [
    "8.14.3"
  ],
[...]

PR #13958, opened 2024-08-30
Fix ScrollID used for retrieval of APM Agents configuration

up2neck:
- Fixed broken APM Agents configuration cache

## Motivation/summary

APM Agents configuration cache appears to be broken, resulting frequent invalid requests made by APM Server to Elasticsearch cluster.

## Checklist

## How to test these changes

1. Deploy Elastic Stack: Elasticsearch, Kibana, APM Server.
2. Configure self-instrumentation for APM Server
3. Open Kibana/APM
4. Check APM Server has no failed operations of type "Elasticsearch: POST .apm-agent-configuration/_search" for Elasticsearch dependency

## Related issues

closes [#13957](https://github.com/elastic/apm-server/issues/13957)

### DIFF
--- a/internal/agentcfg/elasticsearch.go
+++ b/internal/agentcfg/elasticsearch.go
@@ -251,40 +251,38 @@ func (f *ElasticsearchFetcher) refreshCache(ctx context.Context) (err error) {
 
 func (f *ElasticsearchFetcher) singlePageRefresh(ctx context.Context, scrollID string) (cacheResult, error) {
 	var result cacheResult
+	var err error
+	var resp *esapi.Response
 
-	if scrollID == "" {
-		resp, err := esapi.SearchRequest{
+	switch scrollID {
+	case "":
+		resp, err = esapi.SearchRequest{
 			Index:  []string{ElasticsearchIndexName},
 			Size:   &f.searchSize,
 			Scroll: f.cacheDuration,
 		}.Do(ctx, f.client)
-		if err != nil {
-			return result, err
-		}
-		defer resp.Body.Close()
-
-		if resp.StatusCode >= http.StatusBadRequest {
-			// Elasticsearch returns 401 on unauthorized requests and 403 on insufficient permission
-			if resp.StatusCode == http.StatusUnauthorized || resp.StatusCode == http.StatusForbidden {
-				f.invalidESCfg.Store(true)
-			}
-			bodyBytes, err := io.ReadAll(resp.Body)
-			if err == nil {
-				f.logger.Debugf("refresh cache elasticsearch returned status %d: %s", resp.StatusCode, string(bodyBytes))
-			}
-			return result, fmt.Errorf("refresh cache elasticsearch returned status %d", resp.StatusCode)
-		}
-		return result, json.NewDecoder(resp.Body).Decode(&result)
+	default:
+		resp, err = esapi.ScrollRequest{
+			ScrollID: scrollID,
+			Scroll:   f.cacheDuration,
+		}.Do(ctx, f.client)
 	}
-
-	resp, err := esapi.ScrollRequest{
-		ScrollID: result.ScrollID,
-		Scroll:   f.cacheDuration,
-	}.Do(ctx, f.client)
 	if err != nil {
 		return result, err
 	}
 	defer resp.Body.Close()
+
+	if resp.StatusCode >= http.StatusBadRequest {
+		// Elasticsearch returns 401 on unauthorized requests and 403 on insufficient permission
+		if resp.StatusCode == http.StatusUnauthorized || resp.StatusCode == http.StatusForbidden {
+			f.invalidESCfg.Store(true)
+		}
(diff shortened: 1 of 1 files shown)


changelogs/head.asciidoc
```diff
@@ -7,6 +7,7 @@ https://github.com/elastic/apm-server/compare/8.15\...main[View commits]
 ==== Bug fixes
 
 - Track all bulk request response status codes {pull}13574[13574]
+- Fix APM Agents configuration cache renewal {pull}13958[13958]
```
carsonip, line 10: It is fine to remove this line, as this bugfix will likely be backported to 8.15.
```suggestion
```

carsonip (commented):
Thanks for the PR. Nice catch!

Ideally this PR also updates the test to assert scroll id payload in `elasticsearch_test.go`, but it may not be trivial. I can do it in a follow up PR.

up2neck:
@carsonip 
Alongside this, what do you think about adding check of "scroll_id" is not empty after cache renewal?

carsonip:
> Alongside this, what do you think about adding check of "scroll_id" is not empty after cache renewal?

It doesn't hurt, but personally I do not think there is a _need_ in handler to validate anything returned by Elasticsearch. We rely on Elasticsearch for a correct scroll ID, as well as correct search results. What's needed here is a better test, which I'm working on.

Either way the error will be surfaced, just more explicitly if an additional check is in place, or a little more obscurely in a form of HTTP 400 like this case.

up2neck:
> > Alongside this, what do you think about adding check of "scroll_id" is not empty after cache renewal?
> 
> It doesn't hurt, but personally I do not think there is a _need_ in handler to validate anything returned by Elasticsearch. We rely on Elasticsearch for a correct scroll ID, as well as correct search results. What's needed here is a better test, which I'm working on.
> 
> Either way the error will be surfaced, just more explicitly if an additional check is in place, or a little more obscurely in a form of HTTP 400 like this case.

But original code contains HTTP validation, but only for the first search request: 
https://github.com/elastic/apm-server/blob/a5aaf076514d6bd849eaf2bd8baed7cc5df5e845/internal/agentcfg/elasticsearch.go#L266

Probably we can modify it to handle both of the requests:
```
func (f *ElasticsearchFetcher) singlePageRefresh(ctx context.Context, scrollID string) (cacheResult, error) {
	var result cacheResult
	var err error
	var resp *esapi.Response

	if scrollID == "" {
		resp, err = esapi.SearchRequest{
			Index:  []string{ElasticsearchIndexName},
			Size:   &f.searchSize,
			Scroll: f.cacheDuration,
		}.Do(ctx, f.client)
	} else {
		resp, err = esapi.ScrollRequest{
			ScrollID: scrollID,
			Scroll:   f.cacheDuration,
		}.Do(ctx, f.client)
	}
	if err != nil {
		return result, err
	}
	defer resp.Body.Close()

	if resp.StatusCode >= http.StatusBadRequest {
		// Elasticsearch returns 401 on unauthorized requests and 403 on insufficient permission
		if resp.StatusCode == http.StatusUnauthorized || resp.StatusCode == http.StatusForbidden {
			f.invalidESCfg.Store(true)
		}
		bodyBytes, err := io.ReadAll(resp.Body)
		if err == nil {
			f.logger.Debugf("refresh cache elasticsearch returned status %d: %s", resp.StatusCode, string(bodyBytes))
		}
		return result, fmt.Errorf("refresh cache elasticsearch returned status %d", resp.StatusCode)
	}
	return result, json.NewDecoder(resp.Body).Decode(&result)
}
```

carsonip:
> But original code contains HTTP validation, but only for the first search request

Correct. There is no handling for bad HTTP response codes in scroll requests. In retrospect, there should be. Thanks for adding it to the PR.

However, that increases the scope of this PR a little bit, and it will need to be tested on our end to avoid introducing regression (I don't expect any).

carsonip (dismissed):
lgtm, thanks for the PR. I can confirm this fixes the scrolling bug by my follow-up PR to add tests: https://github.com/elastic/apm-server/pull/13959 . The refactoring to handle http errors should be fine, as expired scroll requests returns 404.

carsonip:
run docs-build

up2neck:
> > But original code contains HTTP validation, but only for the first search request
> 
> Correct. There is no handling for bad HTTP response codes in scroll requests. In retrospect, there should be. Thanks for adding it to the PR.
> 
> However, that increases the scope of this PR a little bit, and it will need to be tested on our end to avoid introducing regression (I don't expect any).

Made it

carsonip (approved):

carsonip:
run docs-build

up2neck:
@carsonip unfortunately, Github lost my commit signature after rebase in web UI. Have restored it.

carsonip (approved):

carsonip:
run docs-build

carsonip:
run docs-build

PR #13841, opened 2024-08-07
test: set owner v1v for testing the automation in the forked repo

v1v:
## Motivation/summary

I'd like to honour the CI details so we can tag any cloud resources with the Build ID or so, this will help to identify if the resources were created when running the terraform locally or in the CI.

## Checklist

- [ ] Update [CHANGELOG.asciidoc](https://github.com/elastic/apm-server/blob/main/CHANGELOG.asciidoc)
- [ ] Documentation has been updated

For functional changes, consider:
- Is it observable through the addition of either **logging** or **metrics**?
- Is its use being published in **telemetry** to enable product improvement?
- Have system tests been added to avoid regression?

PR #14233, opened 2024-10-02
8.15: fix smoke tests

endorama:
## Motivation/summary

tags module was using the build input var, but it was not defined.
This was producing failures like:
```
Error: Reference to undeclared input variable

  on ../../infra/terraform/modules/tags/output.tf line 7, in locals:
   7:     "build" : var.build

An input variable with the name "build" has not been declared. This variable
can be declared with a variable "build" {} block.
```

## How to test these changes

Manually trigger Smoke Test workflow on this branch.

https://github.com/elastic/apm-server/actions/workflows/smoke-tests-schedule.yml

### DIFF
--- a/testing/infra/terraform/modules/tags/vars.tf
+++ b/testing/infra/terraform/modules/tags/vars.tf
@@ -1,3 +1,8 @@
 variable "project" {
   description = "The value to use for the project tag/label"
 }
+
+variable "build" {
+  description = "The value to use for the build tag/label"
+  default     = "unknown"
+}


endorama:
Triggered smoke test run https://github.com/elastic/apm-server/actions/runs/11140665972

marclop (approved):

marclop:
I think we need to merge this to main, and backport to `8.x`, not open against 8.15.

endorama:
@marclop On `main` this is correct (see [here](https://github.com/elastic/apm-server/blob/afe87f9cce085bcc5f4881b177df60153956629e/testing/infra/terraform/modules/tags/vars.tf#L5)). The original PR wasn't backported, which is the source for these failures https://github.com/elastic/apm-server/pull/13841

endorama:
The previous run failed with the same error message, due to the way I triggered it. I triggered another run the should verify if this change works in https://github.com/elastic/apm-server/actions/runs/11147080061

(but it used the workflow file from `main` so I'm not sure it tests what I expect). Anyway if it don't pass I think it's safe to merge after @marclop confirm I address his concerns.

endorama:
@kruskall no because `8.15` already has it. Anyway, as the tests are broken, I opened a backport PR, which makes clearer where this fix come from.

Closing as superseded by https://github.com/elastic/apm-server/pull/14240 (the only difference is the docs)

PR #14257, opened 2024-10-04
benchmarks: add SetMemoryLimit to apmbench so it doesn't go OOM

1pkg:
## Motivation/summary

In benchmarks workflow apm-perf occasionally goes OOM, this change sets memory limit to 90% of host memory to prevent OOM happening again. 

## Checklist

### DIFF
--- a/systemtest/cmd/apmbench/main.go
+++ b/systemtest/cmd/apmbench/main.go
@@ -22,9 +22,11 @@ import (
 	"flag"
 	"fmt"
 	"log"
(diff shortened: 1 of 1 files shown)


systemtest/cmd/apmbench/main.go
marclop, line 1: We should really be using `apmbench` from `apm-perf`, and remove this binary from the systemtest folder.

marclop (approved):
LGTM to unblock OOM issues

PR #14334, opened 2024-10-10
ci: fix benchmarks cleanup

reakaleek:
## Motivation/summary

The benchmarks workflow does not clean up all the resources it's creating.

In this [workflow run](https://github.com/elastic/apm-server/actions/runs/11183981002) you can see that the [Spin up benchmark environment](https://github.com/elastic/apm-server/actions/runs/11183981002/job/31093763569#step:15:729) step creates 22 resources.

But the [Teardown benchmark environment](https://github.com/elastic/apm-server/actions/runs/11183981002/job/31093763569#step:30:381) only deletes 9 resources.

The reason for it is the differently passed `TF_VAR_*` environment variables in both steps.

This change fixes it and aligns the passed tf vars to both apply and destroy.

## Checklist

## How to test these changes

I was able to test this change in https://github.com/elastic/apm-server/actions/runs/11277362803/job/31363323001, where you can find 22 resources being destroyed.

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -134,102 +134,102 @@ jobs:
           admin_console_url=$(terraform output -raw admin_console_url)
           echo "admin_console_url=$admin_console_url" >> "$GITHUB_OUTPUT"
           echo "-> infra setup done"
-#      - name: Run benchmarks autotuned
-#        if: ${{ inputs.benchmarkAgents == '' }}
-#        run: make run-benchmark-autotuned
-#
-#      - name: Run benchmarks self tuned
-#        if: ${{ inputs.benchmarkAgents != '' }}
-#        run: make run-benchmark
-#
-#      - name: Cat standalone server logs
-#        if: ${{ env.RUN_STANDALONE == 'true' && failure() }}
-#        run: make cat-apm-server-logs
-#
-#      - name: Index benchmarks result
-#        run: make index-benchmark-results
-#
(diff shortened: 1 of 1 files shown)


.github/workflows/benchmarks.yml
```diff
@@ -133,10 +134,6 @@ jobs:
           admin_console_url=$(terraform output -raw admin_console_url)
           echo "admin_console_url=$admin_console_url" >> "$GITHUB_OUTPUT"
           echo "-> infra setup done"
-        env:
-          TF_VAR_worker_region: ${{ env.AWS_REGION }}
```
reakaleek, line 137: `env.AWS_REGION` doesn't exist. Hence I removed it.

1pkg (approved):

.github/workflows/benchmarks.yml
v1v: can you add the link to the reference that included it? In case there are some rationale?

v1v (approved):

.github/workflows/benchmarks.yml
reakaleek: https://github.com/elastic/apm-server/pull/13884/files#diff-22eef62c9d7b36d02f5fd5aaada92702e22d6795c464239cb7b8fe0f26ea1e1cR137

reakaleek: is this what you meant?

Issue #14337, opened 2024-10-10
Long delay between EA received input and loaded


carsonip:
**APM Server version** (`apm-server version`): 8.14.3

**Description of the problem including expected versus actual behavior**:

In EA managed apm-server, there are observations that there can be delay between log "received input from elastic-agent" and "loaded input config" in the order of days. It implies 2 servers are actually running in the apm-server process during this long period when the old server is stopping. The fact that 2 TBS gc goroutines may cause #14305 , actually makes #14305 also an evidence that 2 servers are running at the same time.

carsonip:
I'm fairly convinced that it is due to slow httpServer and grpcServer shutdown, as I managed to reproduce something similar with an artificial delay:
```
diff --git a/internal/beatcmd/beat.go b/internal/beatcmd/beat.go
index 8a62e1c89..6d19f67fb 100644
--- a/internal/beatcmd/beat.go
+++ b/internal/beatcmd/beat.go
@@ -365,7 +365,7 @@ func (b *Beat) Run(ctx context.Context) error {
 		return err
 	}
 
-	if b.Manager.Enabled() {
+	if b.Manager.Enabled() || true {
 		reloader, err := NewReloader(b.Info, b.newRunner)
 		if err != nil {
 			return err
@@ -377,6 +377,37 @@ func (b *Beat) Run(ctx context.Context) error {
 			return fmt.Errorf("failed to start manager: %w", err)
 		}
 		defer b.Manager.Stop()
+
+		g.Go(func() error {
+			for {
+				in := config.MustNewConfigFrom(map[string]interface{}{
+					"apm-server": map[string]interface{}{
+						"rum.enabled": true,
+						"host":        "0.0.0.0:8200",
+						"sampling.tail": map[string]interface{}{
+							"enabled": true,
+							"policies": []map[string]interface{}{
+								{"sampling_rate": 0.1},
+							},
+							"storage_gc_interval": "2s",
+						},
+					},
+				})
+				out := config.MustNewConfigFrom(map[string]interface{}{
+					"elasticsearch": map[string]interface{}{
+						"host":     []string{"localhost:9200"},
+						"username": "admin",
+						"password": "changeme",
+					},
+				})
+				if err := reloader.reload(in, out, nil); err != nil {
+					logp.Err("reload error")
+				}
+				time.Sleep(time.Minute)
+			}
+
+			return nil
+		})
 	} else {
 		if !b.Config.Output.IsSet() {
[...]

PR #14339, opened 2024-10-11
Apply shutdown timeout to http server to limit reload delay

carsonip:
## Motivation/summary

httpServer.stop may block indefinitely, potentially due to misbehaving connections.

Apply shutdown timeout to http server so that in a hot reload, the old and new server overlapping time is bounded.

## Checklist

## How to test these changes

May need to change the code to hang the http server to observe the shutdown timeout. Alternatively there may be a way to mock a misbehaving http client without code change.

## Related issues

Part of #14337


internal/beater/server.go
```diff
@@ -222,11 +222,10 @@ func (s server) run(ctx context.Context) error {
 	})
 	g.Go(func() error {
 		<-ctx.Done()
-		// httpServer should stop before grpcServer to avoid a panic caused by placing a new connection into
-		// a closed grpc connection channel during shutdown.
-		// See https://github.com/elastic/gmux/issues/13
-		s.httpServer.stop()
 		s.grpcServer.GracefulStop()
```
carsonip, line 225: [to reviewer] Moving GracefulStop first, as this was how the code was originally written before https://github.com/elastic/gmux/issues/13 was discovered. Now that it is fixed, we can revert it, so that it sends GOAWAY to grpc clients.

internal/beater/server.go
```diff
@@ -222,11 +222,10 @@ func (s server) run(ctx context.Context) error {
...
+		stopctx, cancel := context.WithTimeout(context.Background(), s.cfg.ShutdownTimeout)
+		defer cancel()
+		s.httpServer.stop(stopctx)
```
1pkg, line 228: why not just propagate `ctx` that already exists here, it should be already canceled

1pkg (approved):

internal/beater/server.go
carsonip: That's the ctx that get canceled on signal receive. We are supposed to wait for a while to gracefully shutdown and drain connections. see https://github.com/elastic/apm-server/blob/6ad5821fcf930ce5ca8f3aafc02609752c3b24e1/internal/beater/beater.go#L165

1pkg: Oh I didn't notice that we are using `ShutdownTimeout` here, though we just propagate an independent cancel context. Looks good then!

PR #14340, opened 2024-10-11
Limit badger gc concurrency to 1 to avoid panic

### LINKED ISSUE #14305: TBS: "Value log GC request rejected" error
Errors observed in the wild: message "tail sampler aborted" with error.message "Value log GC request rejected".

Due to #14337 , it is possible to have 2 TBS GC goroutine running concurrently, which will cause one of them to crash. In a hot reload under EA, if the new server's GC crashed, and the old server's GC later stopped as part of the reload, this will leave the server with no GC and the TBS storage size will keep growing.

**APM Server version** (`apm-server version`): observed on 8.14.3

carsonip:
## Motivation/summary

Badger GC will panic when run concurrently. 2 TBS processors may run concurrently during a hot reload. Make TBS processor concurrent-safe by protecting badger gc using a mutex.

## Checklist

## How to test these changes

See unit test

Alternatively, somehow (by modifying code locally?) trigger a slow EA reload, and observe that despite 2 TBS processors are running concurrently, gc never runs concurrently.

## Related issues

Fixes #14305


carsonip (commented):
appears that pubsub e.g. `readSubscriberPosition` and `writeSubscriberPosition` needs to change too

x-pack/apm-server/sampling/processor.go
```diff
@@ -558,7 +575,13 @@ func (p *Processor) Run() error {
 	return nil
 }
 
+// subscriberPositionFileMutex protects the subscriber file from concurrent RW, in case of hot reload.
+var subscriberPositionFileMutex sync.Mutex
```
carsonip, line 579: [to reviewer] This ended up as a mutex over the subscriber file only, but not the subscriber goroutine. Although this means possibly duplicate work (e.g. searching in ES) during the overlap, any position written to subscriber file is a position that is processed. Running 2 subscriber goroutines concurrently does not present any correctness issues.

x-pack/apm-server/sampling/processor.go
```diff
@@ -386,6 +391,16 @@ func (p *Processor) Run() error {
 		}
 	})
 	g.Go(func() error {
+		// Protect this goroutine from running concurrently when 2 TBS processors are active
+		// as badger GC is not concurrent safe.
+		select {
+		case <-p.stopping:
+			return nil
+		case gcCh <- struct{}{}:
```
carsonip, line 399: [to reviewer] Used a channel here instead of a sync.Mutex, just to avoid blocking the goroutine shutdown in case Stop is called. I cannot imagine a case where mu.Lock() will block the shutdown, but just to err on the safe side.

rubvs (approved):

1pkg (dismissed):

x-pack/apm-server/sampling/processor_test.go
```diff
@@ -668,6 +669,31 @@ func TestStorageGC(t *testing.T) {
 	t.Fatal("timed out waiting for value log garbage collection")
 }
 
+func TestStorageGCConcurrency(t *testing.T) {
```
carsonip, line 672: [to reviewer] added this test to reproduce the issue. I don't know how to better reproduce it in a simpler way other than setting a short GC interval and sleeping for a second.

1pkg (approved):

PR #14345, opened 2024-10-11
benchmark: Use static URL value for admin console URL for slack automation

1pkg:
## Motivation/summary

This is a small fix to make existing slack automation in benchmark workflow to work as expected with standalone benchmark mode.

## Checklist


testing/benchmark/outputs.tf
```diff
@@ -43,6 +43,6 @@ output "apm_server_ip" {
 }
 
 output "admin_console_url" {
-  value       = var.run_standalone ? "" : module.ec_deployment[0].admin_console_url
+  value       = var.run_standalone ? "https://cloud.elastic.co/deployments" : module.ec_deployment[0].admin_console_url
```
marclop, line 46: Shouldn't this be `admin.found.no`? If you're running it yourself, then you'll be able to see the deployment in the "normal" (user console) UI. But if it's running in CI, then you won't.

1pkg: I don't think there is a big difference, since it only uses this URL when run against the Moxy when no ESS deployment exists.

PR #14551, opened 2024-11-06
fix: clear scroll after completing scroll requests

kruskall:
## Motivation/summary

clear the search context using a clear scroll request for sourcemap and agentcfg search scroll requests

## Checklist

## How to test these changes

- update a lot of sourcemaps to ES
- run apm-server
- check open scroll context
- verify there are no open scroll context

## Related issues
Related to https://github.com/elastic/apm-server/issues/14619


internal/agentcfg/elasticsearch_test.go
```diff
@@ -79,6 +80,9 @@ func newElasticsearchFetcher(
 	i := 0
 
 	fetcher := NewElasticsearchFetcher(newMockElasticsearchClient(t, func(w http.ResponseWriter, r *http.Request) {
+		if r.Method == http.MethodDelete && strings.HasPrefix(r.URL.Path, "/_search/scroll") {
+			return
```
carsonip, line 84: nit: Can you check that the scroll id is `respTmpl["_scroll_id"]`? We had a bug with scrolling in the past

internal/sourcemap/metadata_fetcher.go
```diff
@@ -179,10 +179,24 @@ func (s *MetadataESFetcher) sync(ctx context.Context) error {
 		}
 	}
 
+	s.clearScroll(ctx, scrollID)
```
carsonip, line 182: do we have tests for sourcemap ES fetcher?

axw:
> Motivation/summary
> clear the search context using a clear scroll request for sourcemap and agentcfg search scroll requests

What's the motivation?

internal/agentcfg/elasticsearch.go
```diff
@@ -240,6 +240,8 @@ func (f *ElasticsearchFetcher) refreshCache(ctx context.Context) (err error) {
 		}
 	}
 
+	f.clearScroll(ctx, scrollID)
```
lahsivjar, line 243: I might be missing some context here but if the single page refresh reguarly errors then we won't be clearing the scroll which might lead to too many scrolls. I think a better way to handle this would be to create a method (something like `doScroll`) with the `for` loop and a defer call to clear scroll.

internal/sourcemap/metadata_fetcher.go
lahsivjar, line 182: Same concern as for elasticsearch fetcher.

rubvs (approved):

internal/sourcemap/metadata_fetcher.go
```diff
@@ -179,10 +180,28 @@ func (s *MetadataESFetcher) sync(ctx context.Context) error {
...
+	if err != nil {
+		s.logger.Warnf("failed to clear scroll: %v", err)
+		return
+	}
+
+	if resp.IsError() {
+		s.logger.Warn("clearscroll request returned error: %s", resp.Status())
+	}
+
+	defer resp.Body.Close()
```
1pkg, line 202: nit: no need to defer this call

1pkg (dismissed):

1pkg (approved):

PR #14561, opened 2024-11-06
changelog for 8.15.4

inge4pres:
Changelog entry for upcoming 8.15.4

### DIFF
--- a/changelogs/8.15.asciidoc
+++ b/changelogs/8.15.asciidoc
@@ -16,7 +16,7 @@ https://github.com/elastic/apm-server/compare/v8.15.3\...v8.15.4[View commits]
 [float]
 ==== Bug fixes
 
-- Rollover for all APM data streams is now executing correctly in Elasticsearch 8.15.4, fixing a https://www.elastic.co/guide/en/observability/current/apm-known-issues.html#_upgrading_to_v8_15_x_may_cause_ingestion_to_fail[known issue] present in all patch versions of this minor release.
+- Rollover for all APM data streams is now executing correctly in Elasticsearch 8.15.4, fixing a https://www.elastic.co/guide/en/observability/current/apm-known-issues.html#_upgrading_to_v8_15_x_may_cause_ingestion_to_fail[known issue] present in 8.15.0, 8.15.1, 8.15.2 and 8.15.3.
 
 [float]
 [[apm-release-notes-8.15.3]]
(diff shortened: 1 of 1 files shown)


changelogs/8.15.asciidoc
```diff
@@ -1,11 +1,23 @@
...
+[float]
+[[apm-release-notes-8.15.4]]
+=== APM version 8.15.4
+
+https://github.com/elastic/apm-server/compare/v8.15.3\...v8.15.4[View commits]
+
+[float]
+==== Bug fixes
+
+- Rollover for all APM data streams is now executing correctly in Elasticsearch 8.15.4, fixing a https://www.elastic.co/guide/en/observability/current/apm-known-issues.html#_upgrading_to_v8_15_x_may_cause_ingestion_to_fail[known issue] present in all patch versions of this minor release.
```
carsonip, line 19: Q: I don't see any changes in apm-server repo related to this. Does the fix sit in ES apm-data plugin? If so, is it possible to attach the PR here?

1pkg (dismissed):

changelogs/8.15.asciidoc
inge4pres: @lahsivjar told me the fix is entirely in ES, and I also had a doubt if the apm-server changelog should have this entry.
WDYT?

simitt: IMO it would make sense to have it in the apm-server changelog. The users might not really care which exact component, but rather that it was a problem in apm-server data ingestion that was fixed.

lahsivjar: +1 on adding PR link. Also, should we be more explicit for the "all patch versions of this minor release" but? We can say "for 8.15.0 to 8.15.3 versions"?

lahsivjar (commented):
LGTM! A minor comment

changelogs/8.15.asciidoc
inge4pres: Ok I'll add.
What link should be used, the ES PR?

lahsivjar: Yes, I think https://github.com/elastic/elasticsearch/pull/116219 should be good enough.

lahsivjar (approved):
LGTM!

simitt (approved):

Issue #13872, opened 2024-08-13
Make ubi the default image and drop the ubuntu-based image


kruskall:
We've released chainguard image in https://github.com/elastic/apm-server/pull/13137

Since 8.15.0 we're publishing both `apm-server` and ~`apm-server-wolfi`~ apm-server-ubi

In future relase we should drop the ubuntu based image and make ~wolfi~ ubi the default image, thus only maintaining a `apm-server` image based on ~wolfi~ ubi.

cachedout:
@simitt Per my discussion with @mgreau today, it seems like this work will need to be done by approximately the end of October if not sooner. (@mgreau please correct me if that's not correct in your view.)

cachedout:
@simitt Did this question get adequately answered via the Slack discussion in the or do you want more clarity in this issue?

simitt:
We have everything we need, thank you.

PR #14786, opened 2024-12-02
feat: migrate default docker image to ubi-micro

kruskall:
## Motivation/summary

In line with other products and with the direction in https://github.com/elastic/apm-server/issues/13872 we want to drop ubuntu and use ubi base images for the default image.

Similar to what we did with the chainguard image, the image is pinned to a digest for security/reproducibility and tini is dropped since it's not needed in modern docker versions.

## Checklist

## How to test these changes

`docker build . -f packaging/docker/Dockerfile -t ubi-micro-apm-server --build-arg GOLANG_VERSION=1.23`
`docker run ubi-micro-apm-server` (note if you don't pass an elasticsearch url it will use `elasticsearch` by default).


rubvs (dismissed):

axw (commented):
Can we remove the tini & docker-entrypoint files?

v1v:
Out of curiosity, I've seen `tini` is not needed when using `ub9`, if that's correct, does it mean we can remove https://github.com/elastic/apm-server/blob/767abab33d13e46652fdfde2963c317503e104a0/packaging/ironbank/hardening_manifest.yaml#L42-L46 in a follow-up?

1pkg (approved):

axw (approved):
LGTM! Thanks @kruskall!

Issue #14539, opened 2024-11-05
Replace md5 with FIPS compliant algorithm


maxgio92:
Hello,
it would be nice to replace [this call](https://github.com/elastic/apm-server/blob/f4b8e24ec4828569d47ef590282506d43367885c/internal/beater/config/agentconfig.go#L128) to `md5` with a call to a FIPS 140 allowed algorithm like sha256 to calculate the etag for the fleet agent configuration.

What do you think?

Thank you

inge4pres:
hello @maxgio92 👋🏼 

i don't know how the eTag is used, but it doesn't seem like sensitive information that needs strong hashing.

Why do you think we should update it?
What benefit would we gain compared to the current usage of md5?

maxgio92:
Hi @inge4pres, I see and understand the point.
I think anyway that the advantage is to be able to meet compliance with FIPS-140 that doesn't allow usage of md5 in general, at a pretty low cost.
Thank you

xnox:
@inge4pres it is not security sensitive, but many FIPS go toolchains block crypto/md5 usage such that it fails at runtime. This triggered go coverage to switch away from md5 to a non-security hash FNV-1a instead (which is available in go standard library, and faster than md5) see https://go-review.googlesource.com/c/go/+/617360

Another option, which webpack chose is to switch to xxhash (which is even faster than above two) for eTag like purposes.

md5 often raises suspicion, because whilst today it almost almost used as a non-security relevant hash; it used to be classed as a secure one, and today it is a very slow choice for non-security one.

Does this make sense?

axw:
+1 on switching away from MD5, but we'll need to be mindful of whether this is a breaking change. Similar issue: https://github.com/elastic/apm-data/issues/146

It's _might_ be possible that we can remove the linked code altogether, as we have a new mechanism for agent configuration these days. I think we kept that around for migration, but that's nearly 2 years old now.

xnox:
> +1 on switching away from MD5, but we'll need to be mindful of whether this is a breaking change. Similar issue: [elastic/apm-data#146](https://github.com/elastic/apm-data/issues/146)
> 
> It's _might_ be possible that we can remove the linked code altogether, as we have a new mechanism for agent configuration these days. I think we kept that around for migration, but that's nearly 2 years old now.

[...]

PR #14921, opened 2024-12-12
feat: drop fleet agentcfg and direct fetcher

kruskall:
## Motivation/summary

use ES fetcher and kibana fallback when available
replace nil directfetcher with empty fetcher
remove fleet agentcfg config options

## Checklist

## Related issues

Closes https://github.com/elastic/apm-server/issues/14539


internal/agentcfg/fetch.go
```diff
@@ -17,38 +17,46 @@
...
+	// agents matching the above constraints.
+	Config map[string]string
 }
 
-// NewDirectFetcher returns a new DirectFetcher that serves agent configuration
-// requests using cfgs.
-func NewDirectFetcher(cfgs []AgentConfig) *DirectFetcher {
-	return &DirectFetcher{cfgs}
+func NewEmptyFetcher() Fetcher {
+	return &emptyFetcher{}
```
simitt, line 53: Exposing this Fetcher here although it should only be used for testing is not ideal; IMO there should at least be a comment. 
Alternatively you could export `zeroResult()` and then add the emptyFetcher directly in the test functions where it is needed; also not great though.

simitt: Ah right, ignore my comment then.

simitt (approved):

Issue #11346, opened 2023-08-07
Benchmark and document tail-based sampling performance


axw:
We have a good benchmarking setup for general apm-server ingest performance, but tail-based sampling is a bit of a blind spot. We have done this manually in the past, but we don't have a framework for repeatable testing of TBS.

Once we have a baseline performance established, we should then add to the public documentation. This should include details about disks used, and what kinds of disks are recommended; and expectations about disk and memory usage in relation to ingest rate and sampling rate. Documentation on TBS performance should probably follow on from https://github.com/elastic/apm-server/issues/7842.

We will need https://github.com/elastic/apm-server/issues/7845. Assuming we use `apmbench`, we will need to enable `-rewrite-ids` to ensure `trace.id` and per-trace events are not repeated, which would affect TBS.

Note to whoever works on this:
- We should look at how both disk reads and writes grow with both event ingest rate and sampling rate. Rate of writes is generally proportional to ingest rate, but rate of reads is expected to be proportional to the ingest * sampling rate.
- We should compare Badger v2 (in use at the time of writing this) vs. v4 performance (proposed)

carsonip:
Adding links for prosperity.

> We should look at how both disk reads and writes grow with both event ingest rate and sampling rate. Rate of writes is generally proportional to ingest rate, but rate of reads is expected to be proportional to the ingest * sampling rate.

To be more exact, in a multi-apm-server setup, the expectation is that "Rate of writes is generally proportional to **local** ingest rate", ingest rate local to the apm-server under observation.

On the read side, the expectation is that "rate of reads is expected to be proportional to **local ingest** * sampling rate". However, before fix #13464, apm-server suffers from rate of reads = **global ingest** * sampling rate, which means unscalable disk IO and memory usage.

> We should compare Badger v2 (in use at the time of writing this) vs. v4 performance (proposed)

Related to #11546

raultorrecilla:
moving it from it-107 to it-108

carsonip:
9.0 8gb ram, gp2, TBS: https://github.com/elastic/apm-server/actions/runs/13698906719
9.0 8gb ram, nvme, TBS: https://github.com/elastic/apm-server/actions/runs/13698916450
9.0 32gb ram, gp2, TBS: https://github.com/elastic/apm-server/actions/runs/13698922879
9.0 32gb ram, nvme, TBS: https://github.com/elastic/apm-server/actions/runs/13698930437
[...]

PR #14985, opened 2024-12-17
Add option to enable TBS in benchmark terraform

carsonip:
## Motivation/summary

Update testing/benchmark terraform and github workflow to add an option to enable TBS.

## Checklist

## How to test these changes

Run testing/benchmark terraform locally

## Related issues

Related to #11346 . This should enable reproducible benchmarks on cloud.

### DIFF
--- a/testing/infra/terraform/modules/ec_deployment/scripts/enable_features.tftpl
+++ b/testing/infra/terraform/modules/ec_deployment/scripts/enable_features.tftpl
@@ -26,6 +26,7 @@ echo $${POLICY} | jq '.item' | \
     jq 'del(.created_at)' | jq 'del(.created_by)' | jq 'del(.updated_at)' | jq 'del(.updated_by)' |\
     jq 'select(.inputs[].policy_template == "apmserver").inputs[].vars.expvar_enabled = {type: "bool", value: ${enable_expvar}}' |\
     jq 'select(.inputs[].policy_template == "apmserver").inputs[].vars.pprof_enabled = {type: "bool", value: ${enable_pprof}}' |\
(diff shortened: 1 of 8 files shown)


testing/infra/terraform/modules/ec_deployment/scripts/enable_features.tftpl
```diff
@@ -14,19 +14,19 @@ if [[ "$${POLICY}" == *"$${NOT_FOUND_MSG}"* ]]; then
     exit 0
 fi
 
-echo $${POLICY} | grep '"expvar_enabled":{"type":"bool","value":true}' > /dev/null 2>&1 && EXPVAR_ENABLED=true
-echo $${POLICY} | grep '"pprof_enabled":{"type":"bool","value":true}' > /dev/null 2>&1 && PPROF_ENABLED=true
-if [[ $${EXPVAR_ENABLED} || $${PPROF_ENABLED} ]] ; then
-    echo "expvar or pprof already enabled"
-    exit 0
-fi
+# Abort early if features are already set as desired
```
carsonip, line 17: [to reviewer] unless there's a need to avoid making a no-op agent policy change, I don't see why we need this error-prone check. Before this PR, the check was useless anyway as the grep content `{"type":"bool","value":true}` had out of order JSON keys and it always failed.

carsonip: I'm inclined to remove it

marclop (commented):
Do you think it's worth adding the storage limit setting too?

carsonip:
> Do you think it's worth adding the storage limit setting too?

I don't want to further complicate the github workflow UI, so keeping it fixed for github workflow use case. But it will be configurable when running terraform locally.

marclop (dismissed):
LGTM. We may also need to add a GitHub action variable for the storage value, but not blocking.

reakaleek (approved):

PR #15003, opened 2024-12-19
Fix TF_VAR_apm_server_tail_sampling for scheduled gh actions

carsonip:
## Motivation/summary

A regression from #14985 causing nightly benchmark to fail.

The root cause of regression was that scheduled GH actions ignore `workflow_dispatch` defaults.

## Checklist


marclop (approved):

marclop (approved):
Thanks for the fix, and 💯 CI benchmarking karma points for fixing it

v1v (approved):

PR #15081, opened 2024-12-31
TBS: flatten badger when compaction is not running

### LINKED ISSUE #14923: TBS: apm-server never recovers from storage limit exceeded in rare cases
Tail based sampling: There are observations where after storage limit is exceeded, lsm size remains greatly higher than vlog size.

The assumption is that lsm size should usually be smaller than vlog. It is unclear whether compactions are done in the background to reclaim expired keys in LSM tree. If vlog << lsm, any vlog gc would not be effective in reclaiming storage, and apm-server may be indefinitely stuck in a state where storage exceeds limit.

There are also cases where vlog files are days old in the file system. My hypothesis is that vlog gc thread is still running, but because it relies on stats from compactions but compactions are not run, vlog files are not cleaned up.

carsonip:
## Motivation/summary

WIP

## Checklist

## Related issues

Fixes #14923

### DIFF
--- a/x-pack/apm-server/sampling/processor.go
+++ b/x-pack/apm-server/sampling/processor.go
@@ -43,6 +43,8 @@ const (
 var (
 	// gcCh works like a global mutex to protect gc from running concurrently when 2 TBS processors are active during a hot reload
 	gcCh = make(chan struct{}, 1)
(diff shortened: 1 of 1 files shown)


carsonip:
Superseded by #15106

Issue #14488, opened 2024-10-31
Investigate replacing libbeat monitoring usage with OpenTelemetry instrumentation


axw:
We have had issues related to libbeat monitoring plaguing us for years, notably https://github.com/elastic/apm-server/issues/8383.

We should consider removing our dependency on libbeat for monitoring apm-server, and instrumenting APM Server with the OpenTelemetry API instead. We would need to arrange for the metrics to be written to the monitoring Elasticsearch cluster with the same schema as produced by libbeat; we would also need to expose these metrics via the stats API, and reported in logs.

axw:
For context below, APM Server monitoring builds on libbeat monitoring, and relies on the `.monitoring-beats` index template at https://github.com/elastic/elasticsearch/blob/main/x-pack/plugin/core/template-resources/src/main/resources/monitoring-beats.json. There's tight coupling between this index template and any metrics we add to apm-server. Moreover, this existing template does not have any way of expressing metric dimensions, nor does it support complex metrics such as histograms.

There's a few approaches we could take to migrating to the OTel API, which have varying trade-offs.

1. Use the OTel API only in apm-server code bases, and adapt that to the existing libbeat monitoring framework. libbeat would continue to produce metrics directly to the libbeat monitoring framework
2. Use the OTel API for all metrics, export metrics to Elasticsearch in the same format as libbeat does with a new exporter
4. Use the OTel API for all metrics, export metrics to Elasticsearch in the new OTel-native format and update the stack monitoring & stack telemetry code in Kibana to match

(1) is probably the least amount of work. It will not remove any of the limitations that come with the existing libbeat monitoring template. Moreover, since the OTel API enables users to provide dimensions (attributes) and complex metrics, developers may be surprised when these are dropped or don't behave as they should.
(2) is an incremental improvement over (1) in that it would remove any reliance on the libbeat/monitoring API, but comes with extra work in the form of implementing a new exporter.
(3) is I think the ideal approach technically, but comes with the most amount of work: a new exporter and changes to Kibana.

axw:
I would say the main risks of not implementing this are:
 - It remains difficult to add/change metrics in apm-server code bases, i.e. tight coupling between code & ES templates, lack of support for dimensional metrics, lack of support for histogram metrics.
 - We may not have a solution to https://github.com/elastic/apm-server/issues/8383

[...]

PR #15094, opened 2025-01-01
 Translate otel metrics to libbeat monitoring 

kruskall:
## Motivation/summary

use otel api to record metrics and export them to beats monitoring

## Checklist

## How to test these changes

- run apm-server
- send data
- go to index management and validate the monitoring index is there and monitoring data is inside it

## Related issues

Related to https://github.com/elastic/apm-server/issues/14488


internal/beater/server_test.go
```diff
@@ -653,16 +641,34 @@ func TestWrapServerAPMInstrumentationTimeout(t *testing.T) {
...
-	equal, result := monitoringtest.CompareMonitoringInt(map[request.ResultID]int{
-		request.IDRequestCount:          2,
-		request.IDResponseCount:         2,
-		request.IDResponseErrorsCount:   1,
-		request.IDResponseValidCount:    1,
-		request.IDResponseErrorsTimeout: 1, // test data POST /intake/v2/events
-		request.IDResponseValidAccepted: 1, // self-instrumentation
-	}, intake.MonitoringMap)
-	assert.True(t, equal, result)
+	// TODO broken one request missing
```
axw, line 659: I think it's missing because the MeterProvider passed to newTracerServer is not being used, and a noop provider is used instead.

I'm wondering if we should really be counting instrumentation requests in the metrics

x-pack/apm-server/main_test.go
```diff
@@ -45,25 +48,137 @@ func TestMonitoring(t *testing.T) {
...
-			tailSamplingMonitoringSnapshot = monitoring.CollectFlatSnapshot(samplingMonitoringRegistry, monitoring.Full, false)
 			return runServerError
 		})
 		require.NoError(t, err)
 
 		err = runServer(context.Background(), serverParams)
 		assert.Equal(t, runServerError, err)
-		assert.NotEqual(t, monitoring.MakeFlatSnapshot(), tailSamplingMonitoringSnapshot)
+		// TODO broken: should not be empty
+		monitoringtest.ExpectOtelMetrics(t, reader, map[string]any{})
```
axw, line 77: The test is failing because ExpectOtelMetrics doesn't look for gauge metrics, and the TBS metrics are gauges.

x-pack/apm-server/main_test.go
```diff
@@ -45,25 +48,137 @@ func TestMonitoring(t *testing.T) {
...
+			},
+		}}
+		err := processor.ProcessBatch(context.Background(), &batch)
+		require.NoError(t, err)
+		assert.Empty(t, batch)
 	}
+
+	// Stop the processor and create a new one, which will reopen storage
+	// and calculate the storage size. Otherwise we must wait for a minute
+	// (hard-coded in badger) for storage metrics to be updated.
```
axw, line 108: I'm not sure how the old test is working (is it?), but I don't think this is correct in the current implementation. The storage is not reopened when the processor starts, it gets passed in via config. So I think what we'll need to do is:

- stop the processor, which will flush all writes
- close the DB (probably with the `closeBadger` function)
- reopen the DB with the existing storage dir & MeterProvider (probably with the `getBadgerDB` function)

If you do use `closeBadger` above, you might want to modify it to set `badgerDB = nil` before closing, otherwise `getBadgerDB` is a no-op.

internal/model/modelprocessor/eventcounter.go
```diff
@@ -33,46 +32,34 @@ import (
...
+func NewEventCounter(mp metric.MeterProvider) *EventCounter {
+	meter := mp.Meter("github.com/elastic/apm-server/internal/model/modelprocessor")
+	c := &EventCounter{}
+	for i := range c.eventCounters {
+		eventType := modelpb.APMEventType(i)
+		counter, err := meter.Int64Counter(
+			fmt.Sprintf("apm-server.processor.%s.transformations", eventType),
+		)
+		if err != nil {
+			// TODO(axw) return err
```
axw, line 51: Needs fixing

internal/beater/otlp/grpc.go
```diff
@@ -79,7 +54,17 @@ func RegisterGRPCServices(
 		Semaphore:        semaphore,
 		RemapOTelMetrics: true,
 	})
-	gRPCMonitoredConsumer.set(consumer)
+
+	// FIXME we should add an otel counter metric directly in the
```
axw, line 58: I think this needs fixing before we merge anything?

internal/beater/beater.go
```diff
@@ -281,15 +286,8 @@ func (s *Runner) Run(ctx context.Context) error {
...
-	if err != nil {
-		return err
-	}
-	meterProvider := metric.NewMeterProvider(
-		metric.WithReader(exporter),
-	)
-	otel.SetMeterProvider(meterProvider)
-	tracer.RegisterMetricsGatherer(exporter)
+	// TODO(axw) register apmotel gatherer when we pass it in from beatcmd.
+	//tracer.RegisterMetricsGatherer(exporter)
```
axw, line 290: Needs to be TODONE?

internal/beater/beater.go
```diff
@@ -102,6 +100,10 @@ type RunnerParams struct {
 	// Logger holds a logger to use for logging throughout the APM Server.
 	Logger *logp.Logger
 
+	MetricReader *sdkmetric.ManualReader
+
+	MeterProvider metric.MeterProvider
```
axw, line 105: Please add doc comments

internal/beatcmd/reloader.go
```diff
@@ -50,6 +52,19 @@ type RunnerParams struct {
 
 	// Logger holds a logger to use for logging throughout the APM Server.
 	Logger *logp.Logger
+
+	// MeterProvider holds a metric.MeterProvider that can be used for
+	// creating metrics. The same MeterProvider is execpted to be used
```
axw, line 57:
```suggestion
	// creating metrics. The same MeterProvider is expected to be used
```

internal/beatcmd/beat.go
```diff
@@ -457,6 +484,176 @@ func (b *Beat) registerMetrics() {
...
+			}
+		case "elasticsearch.indexers.created":
+			if value, ok := getScalarInt64(m.Data); ok {
+				monitoring.ReportInt(v, "indexers.created", value)
+			}
+		case "elasticsearch.indexers.destroyed":
+			if value, ok := getScalarInt64(m.Data); ok {
+				monitoring.ReportInt(v, "indexers.destroyed", value)
+			}
+			// TODO output.elasticsearch.indexers.active (created - destroyed?)
```
axw, line 652: TODO

internal/beatcmd/beat.go
```diff
@@ -457,6 +484,176 @@ func (b *Beat) registerMetrics() {
 	monitoring.NewBool(managementRegistry, "enabled").Set(b.Manager.Enabled())
 }
 
+func (b *Beat) registerStatsMetrics() {
+	// TODO: we should ensure all metrics are produced in the expected JSON
```
axw, line 488: Is this done?

internal/beatcmd/beat.go
```diff
@@ -109,6 +116,15 @@ func NewBeat(args BeatParams) (*Beat, error) {
 		beatName = hostname
 	}
 
+	metricReader := sdkmetric.NewManualReader()
+	meterProvider := sdkmetric.NewMeterProvider(
+		// TODO(axw) register an apmotel gatherer,
+		// pass it into the runner so it can be
+		// registered with tracers.
```
axw, line 123: TODO?

axw (commented):
There are still many TODOs in the code, so it's unclear if this is ready for final review. At any rate, I should probably not be the final reviewer since I was involved in the initial implementation. It would be great if we could do this in stages, since this PR is pretty massive.

simitt (changes requested):
Reviewed on a high level in an initial round and agree with the general direction. 
However, this PR is huge which makes it very hard to spot subtle bugs or potential issues. Therefore +1 on @axw 's suggestions to split this PR up into smaller chunks.

internal/beatcmd/beat.go
axw: The TODO was to make sure the metrics have the expected field names. I've fixed it in the other PR

x-pack/apm-server/main.go
```diff
@@ -254,8 +261,13 @@ func wrapServer(args beater.ServerParams, runServer beater.RunServerFunc) (beate
 // called concurrently with opening badger.DB/accessing the badgerDB global,
 // so it does not need to hold badgerMu.
 func closeBadger() error {
+	if badgerDBMetricRegistration != nil {
+		badgerDBMetricRegistration.Unregister()
```
axw, line 265:
```suggestion
		badgerDBMetricRegistration.Unregister()
		badgerDBMetricRegistration = nil
```

internal/beater/server_test.go
```diff
@@ -630,15 +628,12 @@ func TestWrapServerAPMInstrumentationTimeout(t *testing.T) {
...
-		request.IDResponseErrorsTimeout: 1, // test data POST /intake/v2/events
-		request.IDResponseValidAccepted: 1, // self-instrumentation
-	}, intake.MonitoringMap)
-	assert.True(t, equal, result)
+	monitoringtest.ExpectContainOtelMetrics(t, reader, map[string]any{
+		"http.server." + string(request.IDRequestCount):          1,
+		"http.server." + string(request.IDResponseCount):         1,
+		"http.server." + string(request.IDResponseErrorsCount):   1,
+		"http.server." + string(request.IDResponseErrorsTimeout): 1, // test data POST /intake/v2/events
+	})
```
axw, line 636: Probably worth adding a comment that the self-instrumentation requests are not counted in metrics.

This is a change in behaviour. I think it's probably what users would want, but would like to hear @simitt's opinion on this.

internal/beater/otlp/grpc_test.go
```diff
@@ -171,28 +179,21 @@ func TestConsumeLogsGRPC(t *testing.T) {
...
 
-	actual := map[string]interface{}{}
-	monitoring.GetRegistry("apm-server.otlp.grpc.logs").Do(monitoring.Full, func(key string, value interface{}) {
-		actual[key] = value
+	monitoringtest.ExpectContainOtelMetrics(t, reader, map[string]any{
+		"grpc.server.request.count":         2,
+		"grpc.server.response.valid.count":  1,
+		"grpc.server.response.count":        2,
+		"grpc.server.response.errors.count": 1,
 	})
```
axw, line 187: It appears that here and in other places the metric names have changed, i.e. in this case from `apm-server.otlp.grpc.logs.*` to `apm-server.grpc.server.*`. Is that intentional?

internal/beater/otlp/grpc.go
axw: > this shouldn't be needed now that we propagate the meterprovider instead of reusing a global meter.

I don't follow. The issue here is that:
 - the MeterProvider & Meters will survive across config reloads
 - on each config reload we will register a new callback
 - we never unregister those callbacks, so memory & CPU usage will increase with every reload

> I also think adding metrics to apm-data is not as simple since we need to keep other systems in mind (that's a library).

What's difficult about that? We can add a MeterProvider to apm-data/input/otlp.ConsumerConfig, and if it's not set then we can either use the global or a noop provider.

The alternative to adding the metric to apm-data would be to find a way to unregister the callback when the server is stopped.

axw (commented):
Thanks for extracting the other PR - this is much easier to review now.

internal/beatcmd/beat.go
```diff
@@ -539,6 +539,22 @@ func (b *Beat) registerStatsMetrics() {
...
 	})
+	monitoring.NewFunc(monitoring.Default, "apm-server", func(_ monitoring.Mode, v monitoring.Visitor) {
+		var rm metricdata.ResourceMetrics
+		if err := b.metricReader.Collect(context.Background(), &rm); err != nil {
+			return
+		}
+		v.OnRegistryStart()
+		defer v.OnRegistryFinished()
+		for _, sm := range rm.ScopeMetrics {
+			switch {
```
1pkg, line 550: nit: why to use switch case here? do we expect to handle more cases in the future?

internal/beater/api/intake/handler.go
```diff
@@ -82,9 +79,9 @@ func Handler(handler elasticapm.StreamHandler, requestMetadataFunc RequestMetada
 			batchProcessor,
 			&result,
 		)
-		eventsAccepted.Add(int64(result.Accepted))
-		eventsInvalid.Add(int64(result.Invalid))
-		eventsTooLarge.Add(int64(result.TooLarge))
+		eventsAccepted.Add(context.Background(), int64(result.Accepted))
```
1pkg, line 82: why not using `c.Request.Context()` here?

internal/model/modelprocessor/eventcounter.go
```diff
@@ -33,46 +32,30 @@ import (
...
+		c.eventCounters[i] = counter
+	}
+	return c
 }
 
 // ProcessBatch counts events in b, grouping by APMEvent.Processor.Event.
 func (c *EventCounter) ProcessBatch(ctx context.Context, b *modelpb.Batch) error {
 	for _, event := range *b {
-		eventType := event.Type()
-		if eventType == modelpb.UndefinedEventType {
```
1pkg, line 52: Not sure if the change of logic here is intentional, before we would ignore `UndefinedEventType` if any (which should never occur) now we are allow it, reserve a metric and increment it.

internal/beater/otlp/grpc.go
axw: > We cannot add apm-server specific metrics in apm-data. That's library and it could be used by other applications.
> If we add a metric to apm-data it should be agnostic but at that point we can't use the same name as the old beats monitoring metric.

Right, I see. I had in mind we'd do something similar to go-docappender: apm-data native metrics that we rename/transform to the legacy apm-server equivalents.

> Moreover if we add metrics to otlp consumer in apm-data we should consistently add all the consumer metrics, not just unsupported_dropped.

I agree we should have that eventually, but we can add them incrementally.

Anyway, since if we can unregister the callback it's fine if we do this later if at all.

axw (commented):
Thanks @kruskall. Changes LGTM, please wait for sign off from @simitt though

internal/beatcmd/beat.go
simitt: Aligning all of these monitoring functions to use `if` statements also sounds fine, I was mostly addressing the inconsistency for the same kind of checks.

internal/beater/api/mux.go
```diff
@@ -116,18 +117,18 @@ func NewMux(
...
 		{IntakeRUMPath, rumIntakeHandler},
 		{IntakeRUMV3Path, rumIntakeHandler},
-		{IntakePath, builder.backendIntakeHandler},
-		{OTLPTracesIntakePath, builder.otlpHandler(otlpHandlers.HandleTraces, otlp.HTTPTracesMonitoringMap)},
-		{OTLPMetricsIntakePath, builder.otlpHandler(otlpHandlers.HandleMetrics, otlp.HTTPMetricsMonitoringMap)},
-		{OTLPLogsIntakePath, builder.otlpHandler(otlpHandlers.HandleLogs, otlp.HTTPLogsMonitoringMap)},
+		{IntakePath, builder.backendIntakeHandler(meterProvider)},
+		{OTLPTracesIntakePath, builder.otlpHandler(otlpHandlers.HandleTraces, "apm-server.otlp.http.traces.", meterProvider)},
+		{OTLPMetricsIntakePath, builder.otlpHandler(otlpHandlers.HandleMetrics, "apm-server.otlp.http.metrics.", meterProvider)},
+		{OTLPLogsIntakePath, builder.otlpHandler(otlpHandlers.HandleLogs, "apm-server.otlp.http.logs.", meterProvider)},
```
simitt, line 131: Why are you passing the `metricsPrefix` as an argument to the otel routes vs. hardcode them inside the handler functions for non-otel?

internal/beater/server_test.go
simitt: I agree that we don't want to mingle the counters for user requests and self-instrumentation requests together, especially when not being able to encode a differentiator in the metadata. Seems like a bugfix to me.

internal/model/modelprocessor/eventcounter.go
simitt: I don't see how not specifically excluding `UndefinedEventType` would do any harm

simitt (approved):
Only one minor question; 
PR looks great and improves the readability of the metrics collection

x-pack/apm-server/sampling/processor_test.go
```diff
@@ -600,58 +593,21 @@ func TestGroupsMonitoring(t *testing.T) {
...
-	expectedMonitoring.Ints["sampling.events.processed"] = int64(config.MaxDynamicServices) + 2
-	expectedMonitoring.Ints["sampling.events.stored"] = int64(config.MaxDynamicServices)
-	expectedMonitoring.Ints["sampling.events.dropped"] = 1 // final event dropped, after service limit reached
-	expectedMonitoring.Ints["sampling.events.sampled"] = 0
-	expectedMonitoring.Ints["sampling.events.head_unsampled"] = 1
-	expectedMonitoring.Ints["sampling.events.failed_writes"] = 0
-	assertMonitoring(t, processor, expectedMonitoring, `sampling.events.*`, `sampling.dynamic_service_groups`)
-}
-
-func TestStorageMonitoring(t *testing.T) {
```
carsonip, line 614: @kruskall is it intended to remove this test?

PR #15106, opened 2025-01-02
TBS: drop and recreate badger db after exceeding storage limit for TTL time

carsonip:
## Motivation/summary

Workaround for cases where apm-server is stuck at storage limit exceeded state indefinitely because badger DB compaction conditions are not satisfied. This PR implements a goroutine that detects this state, and if the state persists for at least TTL time, as the entries in badger DB would have been expired, just drop and recreate the DB to get out of this state.

## Checklist

~~- [ ] Update [CHANGELOG.asciidoc](https://github.com/elastic/apm-server/blob/main/CHANGELOG.asciidoc)~~ change will be backported. Changelog should be added on docs release.
- [ ] Documentation has been updated

For functional changes, consider:
- Is it observable through the addition of either **logging** or **metrics**?
- Is its use being published in **telemetry** to enable product improvement?
- Have system tests been added to avoid regression?

## How to test these changes

2 ways to test:
1. Start apm-server with TBS, use apmsoak to send many small events <1KB, confirm it is affected by #14923, wait for TTL and ensure that DB is deleted and recreated, and writes to badger are resumed.
2. Manually bloat the badger DB with either APM events or other irrelevant data, start apm-server, ensure DB is deleted and recreated, and writes to badger are resumed.

## Related issues

Alternative to #15081 

Fixes #14923

PR #15112, opened 2025-01-03
TBS: refactor to encapsulate badger DB

carsonip:
## Motivation/summary

Introduce StorageManager to encapsulate badger DB access.

There is a minor difference in how gc mutex and subscriber position file mutex are no longer global variables, but they are per StorageManager, but there should be no apm-server behavior change, only possibly subtle difference in concurrent testing where there are 2 storage managers. apm-server has only 1 global storage manager created by x-pack main.

## Checklist

## How to test these changes

No behavior change. No test needed.

## Related issues

Facilitates changes in #15106


x-pack/apm-server/sampling/eventstorage/storage_manager.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+}
+
+// ManagedReadWriter is a read writer that is transparent to badger DB changes done by StorageManager.
+// It is a wrapper of the ShardedReadWriter under StorageManager.
+type ManagedReadWriter struct {
+	sm *StorageManager
+}
+
+func (s *ManagedReadWriter) ReadTraceEvents(traceID string, out *modelpb.Batch) error {
```
1pkg, line 144: Do we need pass `ManagedReadWriter` by pointer and having pointer receiver methods?

1pkg (approved):

x-pack/apm-server/sampling/eventstorage/storage_manager.go
carsonip: There's a minor difference in xpack main.go as we use nil to more easily distinguish between an initialized and uninitialize `storage   *eventstorage.ManagedReadWriter`, favoring the use of pointer slightly. Otherwise I agree value is better.

x-pack/apm-server/sampling/eventstorage/storage_manager.go
```diff
@@ -0,0 +1,173 @@
...
+}
+
+func (s *ManagedReadWriter) Flush() error {
+	return s.sm.rw.Flush()
+}
+
+// NewBypassReadWriter returns a ReadWriter directly reading and writing to the database,
+// bypassing any wrapper e.g. ShardedReadWriter.
+// This should be used for testing only, useful to check if data is actually persisted to the DB.
+func (s *StorageManager) NewBypassReadWriter() *ReadWriter {
```
carsonip, line 171: @1pkg I thought about it again. Merging rw interface into StorageManager makes it slightly more confusing especially now I introduce a BypassReadWriter for testing purposes. It is clearer that you can create a default shared read writer using NewReadWriter (which wraps ShardedReadWriter -> lockedReadWriter -> ReadWriter), and alternatively a NewBypassReadWriter.

x-pack/apm-server/sampling/eventstorage/storage_manager.go
```diff
@@ -0,0 +1,173 @@
...
+// It is a wrapper of the ShardedReadWriter under StorageManager.
+type ManagedReadWriter struct {
+	sm *StorageManager
+}
+
+func (s *ManagedReadWriter) ReadTraceEvents(traceID string, out *modelpb.Batch) error {
+	return s.sm.rw.ReadTraceEvents(traceID, out)
+}
+
+func (s *ManagedReadWriter) WriteTraceEvent(traceID, id string, event *modelpb.APMEvent, opts WriterOpts) error {
```
carsonip, line 148: [to reviewer] I intentionally did not add any RLocks here yet as this is a refactoring PR.

1pkg (approved):

PR #15117, opened 2025-01-03
Fix incorrect timeout status code in trace produced by self instrumentation

carsonip:
## Motivation/summary

WIP

TODO: implement the actual fix. Although trace transaction result is HTTP 5xx, it is not the 5xx returned by timeout middleware, as timeout middleware should return 503. It is actually the one from `writeStreamResult` `processStreamError` default HTTP 500: https://github.com/elastic/apm-server/blob/52849ff30f6211afacdbecc838a645eac93cb6da/internal/beater/api/intake/handler.go#L148

## Checklist

- [ ] Update [CHANGELOG.asciidoc](https://github.com/elastic/apm-server/blob/main/CHANGELOG.asciidoc)
- [ ] Documentation has been updated

For functional changes, consider:
- Is it observable through the addition of either **logging** or **metrics**?
- Is its use being published in **telemetry** to enable product improvement?
- Have system tests been added to avoid regression?

## Related issues

Fixes #14232

Issue #14232, opened 2024-10-02
Intake v2 request timeout causes inconsistent log and self-instrumented trace


up2neck:
APM Server log events and correlating trace (with self-instrumentation enabled) contains 2 different HTTP errors:

Whether log event contains 503 error with "request timed out" error:

<details>

<summary>Log entry</summary>

```json
{
  "message": "request timed out",
  "url.original": "/intake/v2/events",
  "http.request.body.bytes": 0,
  "log.level": "error",
  "transaction.id": "d91bb495a6ab25bb",
  "http.request.method": "POST",
  "ecs.version": "1.6.0",
  "http.response.status_code": 503,
  "error.message": "request timed out",
  "log.logger": "request",
  "@timestamp": "2024-10-02T07:22:04.439Z",
  "service.name": "apm-server",
  "user_agent.original": "apm-agent-java/1.45.0 (autocode-app 2.0.0)",
  "log.origin": {
    "function": "github.com/elastic/apm-server/internal/beater/api.apmMiddleware.LogMiddleware.func1.1",
    "file.name": "middleware/log_middleware.go",
    "file.line": 59
  },
  "event.duration": 3499790781,
  "http.request.id": "d91bb495a6ab25bb",
  "source.address": "000.000.000.000",
  "trace.id": "d91bb495a6ab25bbdede8859c7cf8c8a"
}
```

</details>

APM interface shown actual error was:

![Image](https://github.com/user-attachments/assets/a24ba459-8dbe-49af-8353-8df8bfb41ff1)

<details>

<summary>Trace body</summary>

```json
{
  "_index": ".ds-traces-apm-dummy_sub2-2024.09.20-000009",
  "_id": "urkbTJIBO4jYgZRmiL6k",
  "_version": 1,
  "_score": 0,
  "_source": {
    "agent": {
      "name": "go",
      "version": "2.6.0"
    },
    "process": {
      "args": [
        "apm-server",
        "run",
        "-e",
        "-c",
        "config/config-secret/apm-server.yml"
      ],
      "pid": 1,
      "title": "apm-server"
    },
    "source": {
      "nat": {
        "ip": "00.000.00.000"
      },
      "ip": "000.000.000.000"
    },
    "processor": {
      "event": "transaction"
    },
    "url": {
      "path": "/intake/v2/events",
      "scheme": "https",
      "domain": "mydomain",
      "full": "https://mydomain/intake/v2/events"
    },
    "labels": {
      "project": "dummy-sub2"
[...]

PR #15122, opened 2025-01-03
Instrumentation: fix log trace inconsistent status code with error result chaining

1pkg:
## Motivation/summary

This fix is branched from https://github.com/elastic/apm-server/pull/15117. 

This PR introduces additional error `request.Result` chaining for secondary errors that can happen after the response has already been written e.g. timeout or panic. The rationale behind this change to create an ability to pass multiple errors in the result object so they could be eventually logged while maintaining original response.

As a result of this changes self instrumentation transaction will always contain original error, while logging middleware will log both the original error and the secondary timeout error. 

This is alternative to errors chaining PR https://github.com/elastic/apm-server/pull/15123 which will use single error depending on when timeout happened.

## Checklist

## Related issues

https://github.com/elastic/apm-server/pull/15123
https://github.com/elastic/apm-server/pull/15117
https://github.com/elastic/apm-server/issues/14232


carsonip (commented):
I agree that this looks like an overkill which further complicates a part that is already failing due to its complexity. Let's work on #15123

1pkg:
> I agree that this looks like an overkill which further complicates a part that is already failing due to its complexity. Let's work on #15123

I agree, that this approach introduces too much complexity for a little benefit.

1pkg:
Closing this in favor of simple timeout check version from https://github.com/elastic/apm-server/pull/15123

PR #15159, opened 2025-01-07
TBS: Add `sampling.tail.discard_on_write_failure` config

### LINKED ISSUE #11127: Configurable option to handle events failed to be processed by TBS
Tail-based sampling currently doesn't expose a configuration option to select the behavior when a batch of events fail to be processed. There are 2 options to choose from for failure cases:

1. Index on failure
2. Drop on failure

Index on failure is the default behavior of APM-Server. The code does have the option to choose drop on failure however, it is not exposed as a config option ([ref](https://github.com/elastic/apm-server/blob/ccc7161a8239a10a9ce6ef7f8404a1311743af1b/x-pack/apm-server/sampling/processor.go#L84-L88))

Also update the docs created in https://github.com/elastic/apm-server/issues/11126

carsonip:
## Motivation/summary

Add config `sampling.tail.discard_on_write_failure` (default=false) for users to opt in to data loss when TBS storage limit is reached, as an escape hatch when increased ES indexing load is unacceptable.

## Checklist

~~- [ ] Update [CHANGELOG.asciidoc](https://github.com/elastic/apm-server/blob/main/CHANGELOG.asciidoc)~~ backported
~~- [ ] Documentation has been updated~~ intentionally undocumented

## How to test these changes

Enable TBS, trigger storage limit reached, check if behavior aligns with config = {true, false}

## Related issues

Fixes #11127


internal/beater/config/sampling.go
```diff
@@ -52,6 +52,7 @@ type TailSamplingConfig struct {
 	TTL                   time.Duration         `config:"ttl" validate:"min=1s"`
 	StorageLimit          string                `config:"storage_limit"`
 	StorageLimitParsed    uint64
+	DiscardOnWriteFailure bool `config:"discard_on_write_failure"`
```
carsonip, line 55: [to reviewer] Using `discard_on_write_failure` instead of `index_on_write_failure` to have a default of false, and give a stronger sense of opt-in to data loss. Lmk if there's a better name, e.g. `drop_on_write_failure`.

lahsivjar (approved):
LGTM!

1pkg (dismissed):
It would be really nice to have a test for DiscardOnWriteFailure case.

1pkg (approved):

Issue #14933, opened 2024-12-12
Revisit default TBS storage size limit `sampling.tail.storage_limit` and storage limit handling


carsonip:
`sampling.tail.storage_limit` has a default of 3GB which is usually insufficient for high throughput use cases. A storage limit too low will cause TBS to be bypassed and other downstream effects. Reconsider the default in 9.0 release.

### Tasks
- [x] https://github.com/elastic/apm-server/pull/15467
- [x] https://github.com/elastic/integrations/pull/12543
- [x] cloud: https://elasticco.atlassian.net/browse/CP-10146
- [x] https://github.com/elastic/observability-docs/pull/4814
- [x] https://github.com/elastic/docs-content/pull/581
- [x] https://github.com/elastic/docs-content/pull/582
- [x] https://github.com/elastic/apm-server/pull/15524

carsonip:
I have 2 ideas for the new default.
- to allow infinite disk usage
- to detect available disk space (on startup / periodically) and set the limit dynamically

The argument for either of them is that e.g. if you are running a mysql server, you expect it to use storage space but you don't expect yourself updating a "storage limit" every time it gets hit.

The argument for a detection is to avoid causing a node to be unhealthy in any environments.

simitt:
@carsonip the solution to detect available disk space would be more elegant. We would need to ensure that it works on all supported OS systems + docker and have a sensible fallback if disk size cannot be detected. 
Do you have capacity to look into this?

carsonip:
> Do you have capacity to look into this?

Yes, it is in scope of this task. I'll write down some possible solutions soon.

simitt:
>3. keep storage limit handling but default to unlimited (and maybe always set it to a number on ESS)
>4. removing storage limit handling altogether. This will simplify https://github.com/elastic/apm-server/pull/15235.

These options do not sound like customer friendly options to me. 
Option 1 and 2 both sound reasonable. If option 2 is risky or very complex, I wouldn't see a big problem with starting with option 1 and then potentially switching to option 2 later.

carsonip:
Option 1 is the baseline. It should be trivial.
I'm also optimistic about option 2 as we move to pebble in #15246 , because there seems to be a cross-platform disk usage stats (available, used, total) in pebble https://github.com/cockroachdb/pebble/blob/master/vfs/disk_usage_linux.go

carsonip:
[...]

PR #15235, opened 2025-01-14
TBS: Replace badger with pebble

### LINKED ISSUE #15246: TBS: Explore replacing badger with pebble
Look into whether it is feasible to replace badger with pebble to consolidate on KV store dependency. Ensure that there is no major performance regression, support existing apm-server features, and identify blockers to adopt pebble for TBS.

carsonip:
## Motivation/summary

This PR replaces badger with pebble as the database for tail-based sampling.

```
// The database of choice is Pebble, which does not have TTL handling built-in,
// and we implement our own TTL handling on top of the database:
// - TTL is divided up into N parts, where N is partitionsPerTTL.
// - A database holds N + 1 + 1 partitions.
// - Every TTL/N we will discard the oldest partition, so we keep a rolling window of N+1 partitions.
// - Writes will go to the most recent partition, and we'll read across N+1 partitions
```

### Benchmarks

TLDR: +3000% in indexed events/s, +76% intakev2 event rate, while on -75% memory usage and -44% disk usage

See [comment](https://github.com/elastic/apm-server/pull/15235#issuecomment-2620093781) for details.

### Major design changes

- As pebble does not support TTL, this PR introduces a partitioning method to enforce TTL. TTL is handled by assigning a rotating partitioning key (not timestamp based) as database key prefix. There's also a background thread to run a "TTL GC Loop": for every TTL, the prefix of keys are rotated, and the expired prefix will be deleted and compacted. This explicit TTL handling enables precise deletion and the lifetime of an entry in the database is strictly bounded to `2*TTL`. In fact, there's a knob `partitionsPerTTL` to adjust the available prefixes to trade between storage overhead and read amplification. e.g. partitionsPerTTL=1 keeps `2*TTL` entries with 2 partition reads per key read, while partitionsPerTTL=2 keeps `1.5*TTL` entries with 3 partition reads per key read.
- Not using timestamp based prefix enables TTL adjustments without data loss on EA hot reload / apm-server restart, as prefixes are fixed but TTL-truncated timestamps aren't.
- Sampling decision and events are stored separately in different pebble databases, as they have vastly different characteristics, which means the optimal pebble option would be different. Compression is enabled for events but disabled for decisions.

image

### Other implied changes

- As TTL is enforced by deleting entries and compacting actively in the TTL GC Loop, there is no TTL validation on stored entries at read time, i.e. it is possible to run apm-server TBS, stop it, wait for 1 year, then restart apm-server, and old entries will be respected. In contrast, badger checks for expiry at read time. As trace IDs are supposed to be unique, the downside of not doing read-time TTL validation should be minimal.
- Pebble does not use a vlog like badger. In addition to the event and decision DB separation, this also requires changes in monitoring metrics structure. In the draft PR, the sum of table size, WAL size across the 2 DBs are summed and reported in `sampling.tail.storage.lsm_size`, while `sampling.tail.storage.vlog_size` is always 0. The change is not decided yet.

TODO:
- [x] Initial benchmark numbers
- [x] Memory usage compared to main
- [x] TTL handling
- [x] Storage limit handling
- [x] Storage usage & numbers to facilitate https://github.com/elastic/apm-server/issues/14933
- [x] Measure sampling decision miss perf hit
- [x] check if possible to upgrade to pebble v2.0.0: https://github.com/cockroachdb/pebble/issues/4271
- [ ] ~~Check for memory leak by running it for a long time, as pebble does manual memory management~~
- [ ] ~~event loss on upgrade:~~ see #15423 

Useful but not necessary, out of scope of this PR:
- [x] support tbs in `testing/infra/terraform/modules/standalone_apm_server` for reproducible benchmark with moxy and without using ESS: https://github.com/elastic/apm-server/pull/15337

## Checklist

## How to test these changes

Enable TBS, try various sampling policies, send events, keep it running for over 2 * TTL, ensure that disk usage is bounded, and memory usage is expected.

## Related issues

Fixes #15246


x-pack/apm-server/sampling/eventstorage/storage_manager.go
```diff
@@ -67,274 +117,135 @@ func NewStorageManager(storageDir string) (*StorageManager, error) {
...
-	// Delete every file except subscriber position file
-	var (
-		rootVisited         bool
-		sstFiles, vlogFiles int
-		otherFilenames      []string
+// RotatePartitions rotates the partitions to clean up TTL-expired entries.
+func (sm *StorageManager) RotatePartitions() error {
+	sm.partitioner.Rotate()
+	// FIXME: potential race, wait for a bit before deleting?
+	pidToDelete := sm.partitioner.Inactive().ID()
```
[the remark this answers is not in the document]
carsonip: Given that StorageManager is the only writer to partitioner, there will be no competing writes. I don't see a need for a CAS. The use of atomics here is to avoid locking in the hot path (any PartitionedReadWriter actions)

x-pack/apm-server/sampling/eventstorage/partitioner.go
```diff
@@ -0,0 +1,105 @@
...
+// Rotate rotates partitions to the right by 1 position.
+//
+// Example for total=4:
+// (A: active, I: inactive, ^ points at the current active entry)
+// A-I-A-A
+// ^......
+//
+// After Rotate:
+// A-A-I-A
+// ..^....
```
carsonip, line 42: [to reviewer] this describes how the partitioned ring works

x-pack/apm-server/sampling/eventstorage/rw.go
```diff
@@ -0,0 +1,94 @@
...
+
+func (s SplitReadWriter) Close() error {
+	return nil
+}
+
+type storageLimitChecker interface {
+	StorageLimitReached() bool
+}
+
+type StorageLimitReadWriter struct {
```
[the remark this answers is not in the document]
carsonip: I've added quite a few RWs here, but I think it reflects the separation of concerns well. This seems to be an improvement over the old WriterOpts.

x-pack/apm-server/sampling/processor.go
```diff
@@ -102,7 +107,7 @@ func (p *Processor) CollectMonitoring(_ monitoring.Mode, V monitoring.Visitor) {
 	monitoring.ReportNamespace(V, "storage", func() {
 		lsmSize, valueLogSize := p.config.DB.Size()
 		monitoring.ReportInt(V, "lsm_size", int64(lsmSize))
-		monitoring.ReportInt(V, "value_log_size", int64(valueLogSize))
+		monitoring.ReportInt(V, "value_log_size", int64(valueLogSize)) // FIXME: either report everything under lsm_size, or report with new fine-grained metrics, e.g. wal_size
```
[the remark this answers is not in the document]
carsonip: [to reviewer] I haven't decided on how the metrics should be reported. We now have a decision DB and an events DB, each with table and WAL and misc disk usages. Updating monitoring metrics structure will affect tooling, kibana visualizations, ES and filebeat mappings.

I'm slightly inclined to mark it as a TODO instead of a FIXME and leave it as a follow up task (maybe consider it as a regression to tackle post 9.0 FF?), and just report a sum of everything under lsm_size for compatibility now. But I'm interested to hear what folks think about this, on timing (can a monitoring metrics structure change catch 9.0?) and on implementation (should we report a set of metrics for event and decision DB separately?)

x-pack/apm-server/sampling/eventstorage/storage_manager.go
```diff
@@ -26,25 +25,46 @@ const (
...
-	errDropAndRecreateInProgress = errors.New("db drop and recreate in progress")
+	// partitionsPerTTL holds the number of partitions that events in 1 TTL should be stored over.
+	// Increasing partitionsPerTTL increases read amplification, but decreases storage overhead,
+	// as TTL GC can be performed sooner.
+	//
+	// For example, partitionPerTTL=1 means we need to keep 2 partitions active,
+	// such that the last entry in the previous partition is also kept for a full TTL.
+	// This means storage requirement is 2 * TTL, and it needs to read 2 keys per trace ID read.
+	// If partitionPerTTL=2, storage requirement is 1.5 * TTL at the expense of 3 reads per trace ID read.
+	partitionsPerTTL = 1
```
carsonip, line 37: [to reviewer] e.g. pertitionsPerTTL=3 creates 4 active partitions, and 1 can be cleared out per rotation, i.e. 33% of entries can be cleared at `TTL/3` interval. This yields the most stable disk usage in @kruskall 's experiments, but I'm slightly worried about the read amplification, which is not well reflected in the experiment as it wasn't tested with remote sampling decisions. While it is possible to microbenchmark this, but I doubt how much it reflects the reality because of pebble block cache.

x-pack/apm-server/sampling/eventstorage/pebble.go
```diff
@@ -0,0 +1,60 @@
...
+		Levels: []pebble.LevelOptions{
+			{
+				BlockSize:    2 << 10,
+				Compression:  func() pebble.Compression { return pebble.NoCompression },
+				FilterPolicy: bloom.FilterPolicy(10),
+				FilterType:   pebble.TableFilter,
+			},
+		},
+	})
+}
```
carsonip, line 60: [to reviewer] would love feedback from folks with RocksDB configuration expertise. I obtained these config options by trial and error and isn't super scientific. TBH I'm rather new to this.

internal/beater/config/config_test.go
```diff
@@ -362,7 +362,6 @@ func TestUnpackConfig(t *testing.T) {
 						ESConfig:              elasticsearch.DefaultConfig(),
 						Interval:              1 * time.Minute,
 						IngestRateDecayFactor: 0.25,
-						StorageGCInterval:     5 * time.Minute,
```
carsonip, line 365: [to reviewer] I tested with apm-server.yml that an extra options are silently ignored, so a config removal should be fine

x-pack/apm-server/sampling/eventstorage/storage.go
```diff
@@ -5,341 +5,80 @@
...
-	s.pendingSize.Add(baseTransactionSize)
-	return &ReadWriter{
-		s:           s,
-		txn:         nil, // lazy init to avoid deadlock in storage manager
-		pendingSize: baseTransactionSize,
-	}
+	Get(key []byte) ([]byte, io.Closer, error)
+	Set(key, value []byte, opts *pebble.WriteOptions) error
+	Delete(key []byte, opts *pebble.WriteOptions) error
+	NewIter(o *pebble.IterOptions) (*pebble.Iterator, error)
```
carsonip, line 19: [to reviewer] I've got rid of ShardedReadWriter, as I'm not using any transactions ("batch" in pebble). The removal of transaction actually makes a lot of things much simpler, and reduced memory usage. There's no more locking, and storage limit handling is much easier without sharded and in-flight transactions. I haven't noticed any downsides to it yet, but let me know if I missed anything

x-pack/apm-server/sampling/eventstorage/storage_manager.go
```diff
@@ -53,288 +73,190 @@ type StorageManager struct {
...
+		eventRW:    sm.eventStorage.NewReadWriter(),
+		decisionRW: sm.decisionStorage.NewReadWriter(),
+	})
 }
 
-func (s *ManagedReadWriter) WriteTraceSampled(traceID string, sampled bool, opts WriterOpts) error {
-	ok := s.sm.mu.TryRLock()
-	if !ok {
-		return errDropAndRecreateInProgress
+// NewBypassReadWriter returns a SplitReadWriter directly reading and writing to the database,
```
axw, line 254: This doesn't seem to really be needed. I just replaced all uses with NewReadWriter locally, and removed the Close calls (which are no-ops on SplitReadWriter anyway), and all the tests still pass.

x-pack/apm-server/sampling/eventstorage/partition_rw.go
```diff
@@ -0,0 +1,81 @@
...
+)
+
+// PartitionReadWriter reads from and writes to storage across partitions.
+type PartitionReadWriter struct {
+	s *Storage
+}
+
+// Flush waits for preceding writes to be committed to storage.
+func (rw *PartitionReadWriter) Flush() error {
+	return nil
```
axw, line 20: Why is this a no-op? Perhaps we should have another interface separate from RW that has Flush? Maybe:
 - Reader
 - Writer
 - Flusher
 - ReadWriter
 - ReadWriteFlusher

(Perhaps a _little_ over the top, but avoids having methods just because we're implementing an overload wide interface)

x-pack/apm-server/sampling/eventstorage/rw.go
```diff
@@ -0,0 +1,116 @@
...
+	return s.nextRW.IsTraceSampled(traceID)
+}
+
+func (s StorageLimitReadWriter) DeleteTraceEvent(traceID, id string) error {
+	// Technically DeleteTraceEvent writes, but it should have a net effect of reducing disk usage
+	return s.nextRW.DeleteTraceEvent(traceID, id)
+}
+
+func (s StorageLimitReadWriter) Flush() error {
+	return s.nextRW.Flush()
```
axw, line 115: This won't flush deletions on non-current partitions. Is that intentional, and is it OK? IIANM that means (non rotation-related) deletes may not be flushed.

x-pack/apm-server/sampling/eventstorage/prefix.go
```diff
@@ -0,0 +1,151 @@
...
+	var lb bytes.Buffer
+	lb.Grow(1 + len(traceID) + 1)
+	lb.WriteByte(rw.prefix)
+	lb.WriteString(traceID)
+	lb.WriteByte(':')
+
+	var ub bytes.Buffer
+	ub.Grow(lb.Len())
+	ub.Write(lb.Bytes()[:lb.Len()-1])
+	ub.WriteByte(';') // This is a hack to stop before next ID
```
axw, line 50: Maybe use a single buffer, and slice it for LowerBound and UpperBound?

x-pack/apm-server/sampling/eventstorage/prefix.go
```diff
@@ -0,0 +1,151 @@
...
+	key := b.Bytes()
+	return rw.writeEntry(key, data)
+}
+
+func (rw PrefixReadWriter) writeEntry(key, data []byte) error {
+	if err := rw.db.Set(key, data, pebble.NoSync); err != nil {
+		return err
+	}
+	return nil
+}
```
axw, line 100:
```suggestion
	return rw.db.Set(key, data, pebble.NoSync)
}
```

x-pack/apm-server/sampling/eventstorage/prefix.go
axw: Actually, I don't think you need to specify the bounds at all if you're using SeekPrefixGE:

>> // SeekPrefixGE moves the iterator to the first key/value pair whose key is
>> // greater than or equal to the given key and which has the same "prefix" as
>> // the given key. The prefix for a key is determined by the user-defined
>> // Comparer.Split function. The iterator will not observe keys not matching the
>> // "prefix" of the search key.

x-pack/apm-server/sampling/eventstorage/partitioner.go
```diff
@@ -0,0 +1,105 @@
...
+	return PartitionIterator{
+		id:        int(p.current.Load()),
+		remaining: 0,
+		total:     p.total,
+	}
+}
+
+// PartitionIterator is for iterating on partition results.
+// In theory Partitioner could have returned a slice of partition IDs,
+// but returning an iterator should avoid allocs.
```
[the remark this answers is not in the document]
carsonip: only if i knew it exists :face_holding_back_tears:

x-pack/apm-server/sampling/eventstorage/partition_rw.go
carsonip: > Why is this a no-op?

All `Flush` are no-op at the moment, because we stopped using transaction/"batch". In the underlying implementation, all writes are committed immediately using a temporary batch. This has a side effect that if a read follows very closely to a concurrent write, and the read returned sooner than the write, I _think_ they may race and the read may return NotFound. We'll need to add back sharding and enforce strict ordering for trace ID concurrent RW if we need to fix this edge case.

> Perhaps we should have another interface separate from RW that has Flush?

I should have cleaned this up. RW Flush (i.e. transaction commit) is not used, while storage manager flush (i.e. DB flush to disk) is used on apm-server exit.

x-pack/apm-server/sampling/eventstorage/rw.go
carsonip: ditto, I'm more inclined to remove Flush from RW.

x-pack/apm-server/sampling/eventstorage/storage_manager.go
```diff
@@ -53,288 +73,190 @@ type StorageManager struct {
...
-}
-
-func (s *StorageManager) deleteBadgerFiles() error {
-	// Although removing the files in place can be slower, it is less error-prone than rename-and-delete.
-	// Delete every file except subscriber position file
-	var (
-		rootVisited         bool
-		sstFiles, vlogFiles int
-		otherFilenames      []string
+	// FIXME: potential race, wait for a bit before deleting?
```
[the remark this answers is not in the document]
carsonip: > Reads/deletes on the oldest partition that was active before the rotation?

Yes, or any write on the oldest partition. A write on the expiring partition returned later than the DeleteRange, causing unexpected residue in the expiring partition. This shouldn't be a worry though, as there is a buffer of `partitionsPerTTL` partitions between the current partition and the expired partition, so unless the write on the partition stalled for TTL, the race won't happen.

I'll look into using RWMutex.

x-pack/apm-server/sampling/eventstorage/partition_rw.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+// WriteTraceSampled records the tail-sampling decision for the given trace ID.
+func (rw *PartitionReadWriter) WriteTraceSampled(traceID string, sampled bool) error {
+	pid := rw.s.db.WritePartition().ID()
+	return NewPrefixReadWriter(rw.s.db, byte(pid), rw.s.codec).WriteTraceSampled(traceID, sampled)
```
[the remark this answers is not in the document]
carsonip: On top of that, since we have a special key `partitionerMetaKey = "~partitioner"`, all valid partition ids need to be smaller than 176. This arbitrary limitation is not my favorite, but I guess we can live with it. It has no disadvantage in practice.

x-pack/apm-server/sampling/eventstorage/prefix.go
carsonip: nice catch! Removed lb and ub from iter. @kruskall has some ideas on adding a fast path to not use a buffer for 16B trace ID, but that's out of scope of this PR.

x-pack/apm-server/sampling/eventstorage/storage_manager.go
```diff
@@ -53,288 +75,188 @@ type StorageManager struct {
...
 }
 
-// runGCLoop runs a loop that calls badger DB RunValueLogGC every gcInterval.
-func (s *StorageManager) runGCLoop(stopping <-chan struct{}, gcInterval time.Duration) error {
-	// This goroutine is responsible for periodically garbage
-	// collecting the Badger value log, using the recommended
-	// discard ratio of 0.5.
-	ticker := time.NewTicker(gcInterval)
+func (sm *StorageManager) runTTLGCLoop(stopping <-chan struct{}, ttl time.Duration) error {
+	ttlGCInterval := ttl / partitionsPerTTL
```
simitt, line 198: `partitionsPerTTL` gets increased by 1 active and 1 inactive partition; e.g. if `partitionsPerTTL==1`, overall there are 2 active and 1 inactive partition. Wouldn't it make sense to divide by number of active partitions here as there is 1 partition buffer already from the inactive one?

simitt (commented):
For the amount of complexity that this PR holds, it was fairly smooth to navigate through it. I haven't found anything major of concern (what hasn't been discussed or commented already). Comments are mostly nits except for 2 points. 

Given that pebble works with manual memory management, have you had a long running test already ensuring that there is no evident memory leak?

x-pack/apm-server/sampling/eventstorage/partitioner.go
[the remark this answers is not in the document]
carsonip: When I was writing the iterator, it made me miss Python [generators](https://wiki.python.org/moin/Generators). Glad that we have that in Go 1.23. Updated PR.

simitt:
>I'm finding the eventstorage interfaces a bit hard to follow, but I don't have any suggestions yet.

Initially I had a similar experience, but once I got the hang of them I found them fairly smooth. The `ID` part initially confused me as it is basically an `Index` position within a ring. However, I refrained from suggesting to change it, as currently the fact that it is an index is an implementation detail which doesn't need to be exposed.

x-pack/apm-server/sampling/eventstorage/partitioner.go
```diff
@@ -0,0 +1,116 @@
...
+	return PartitionIterator{
+		id:        int(p.current.Load()),
+		remaining: p.total - 1 - 1,
+		total:     p.total,
+	}
+}
+
+// Inactive returns a PartitionIterator pointing to the inactive partition.
+// It contains only 1 partition.
+func (p *Partitioner) Inactive() PartitionIterator {
```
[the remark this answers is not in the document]
carsonip: Updated to an int and replaced PartitionIterator with iter.Seq as suggested by Andrew.

x-pack/apm-server/sampling/eventstorage/prefix.go
```diff
@@ -0,0 +1,144 @@
...
+	b.Grow(1 + len(traceID))
+	b.WriteByte(rw.prefix)
+	b.WriteString(traceID)
+
+	item, closer, err := rw.db.Get(b.Bytes())
+	if err == pebble.ErrNotFound {
+		return false, ErrNotFound
+	}
+	defer closer.Close()
+	return item[0] == entryMetaTraceSampled, nil
```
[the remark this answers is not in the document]
carsonip: This is actually missing an `err != nil` check. I've updated to add it back but keeping this line as nil to be explicit.

x-pack/apm-server/sampling/eventstorage/pebble.go
```diff
@@ -0,0 +1,60 @@
...
+
+func OpenEventPebble(storageDir string) (*pebble.DB, error) {
+	opts := &pebble.Options{
+		FormatMajorVersion: pebble.FormatColumnarBlocks,
+		Logger:             logp.NewLogger(logs.Sampling),
+		MemTableSize:       16 << 20,
+		Levels: []pebble.LevelOptions{
+			{
+				BlockSize:    16 << 10,
```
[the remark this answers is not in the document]
carsonip: Not exactly science. They were some numbers from initial benchmarking before committing to replacing badger with pebble. They will need to be revalidated, especially now that we have built in TTL handling. But I can try to reason about the changes at a high level for now.

event DB:
- 16MB memtable size
  - In hopes for faster writes, especially events can be large. But recently I've reduced it from the initial 32MB to 16MB and noticed much better memory usage. I think we can further reduce it and check for perf impact
- 16KB block size
  - Increased the size as events can be large (I'm always assuming 1KB per event), and I assume bigger blocks mean more effective compression, but this is not backed by numbers. I should look at the metrics and decide on a better block size
  - There's also advice that with compression, block sizes should be smaller to reduce decompression overhead. Another reason to revisit this value.
- snappy compression
  - it is the default and it makes sense to compress events

decision DB:
- 2MB memtable size
  - In microbenchmarks I observe that having bigger memtables actually make things slower, and I assume it is due to pebble not supporting bloom filter on memtables. Checking a cached bloom filter on a table is actually faster.
  - Added benefit of reducing memory usage
- 2KB block size
  - entries in decision DB are independent and caching the blocks is unlikely to benefit from locality (except from the locality in partition ID). Reading a decision on trace AAA doesn't imply trace AAB's decision would be read soon, unlike in events db that transactions and spans of the same trace IDs are stored next to each other.
- no compression
  - I don't see why we need to compress a 1 byte value. Keys are already benefiting from delta encoding anyway.

x-pack/apm-server/sampling/eventstorage/storage_manager.go
carsonip: I've attached the image of how it currently works in the PR description, but I'll attach a copy here too. 
image

Let's imagine we adopt your suggestion that ttlGCInterval is `ttl / (partitionsPerTTL+1)` instead of `ttl / partitionsPerTTL`, then each partition covers 10m instead of 15m, then the 3 active partitions only cover 30m backwards from the ending time of the current partition. That also means if we are only 5mins into this current partition, we're only keeping 25mins of events while we SHOULD be keeping TTL=30m of events. It is not correct.

simitt: You are right; with your approach we are erroring on the side of keeping the data up to one ttl slice longer, which makes sense!

x-pack/apm-server/sampling/eventstorage/pebble.go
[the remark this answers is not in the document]
carsonip: @kruskall is now conducting experiments to determine the optimal values.

x-pack/apm-server/sampling/eventstorage/storage_manager.go
carsonip: This was originally intended to view a lower level rw for minimal perf overhead and an accurate & independent view of db, but in a lot of usages (e.g. in processor_test.go), they really shouldn't care about the implementation. I've removed it to minimize the confusion. If we find the need to do some benchmarking closer to db, we can add it back later.

carsonip:
> Given that pebble works with manual memory management, have you had a long running test already ensuring that there is no evident memory leak?

@kruskall mentioned that with cgo disabled, pebble doesn't use manual memory management. https://github.com/cockroachdb/pebble/blob/master/internal/manual/manual_nocgo.go

x-pack/apm-server/sampling/processor.go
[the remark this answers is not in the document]
carsonip: Deferring this as it shouldn't be a blocker.

x-pack/apm-server/sampling/eventstorage/storage_manager.go
```diff
@@ -67,274 +117,135 @@ func NewStorageManager(storageDir string) (*StorageManager, error) {
...
-// Close closes StorageManager's underlying ShardedReadWriter and badger DB
-func (s *StorageManager) Close() error {
-	s.mu.RLock()
-	defer s.mu.RUnlock()
-	s.rw.Close()
-	return s.db.Close()
+func (sm *StorageManager) Size() (lsm, vlog int64) { // FIXME: stop calling it vlog
+	// FIXME: separate WAL usage?
+	return int64(sm.eventDB.Metrics().DiskSpaceUsage() + sm.decisionDB.Metrics().DiskSpaceUsage()), 0
+}
```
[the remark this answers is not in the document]
carsonip: Ditto, deferring this.

x-pack/apm-server/sampling/eventstorage/pebble.go
carsonip: As mentioned we are now conducting experiments.

x-pack/apm-server/sampling/eventstorage/storage_manager.go
```diff
@@ -26,315 +25,301 @@ const (
...
+	})
 }
 
-// NewBypassReadWriter returns a ReadWriter directly reading and writing to the database,
-// bypassing any wrapper e.g. ShardedReadWriter.
-// This should be used for testing only, useful to check if data is actually persisted to the DB.
-func (s *StorageManager) NewBypassReadWriter() *ReadWriter {
-	return s.storage.NewReadWriter()
+// wrapNonNilErr only wraps an error with format if the error is not nil.
+func wrapNonNilErr(format string, err error) error {
```
carsonip, line 320: [to reviewer] I don't love this, but it seems to help keep the main code clean.

x-pack/apm-server/sampling/eventstorage/storage_manager.go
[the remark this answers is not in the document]
carsonip: Updated to use RWMutex. I still think it is not impossible to use atomics, but let's strive for correctness first.

x-pack/apm-server/sampling/eventstorage/storage_manager.go
```diff
@@ -26,315 +25,301 @@ const (
...
 	}
-}
 
-// ManagedReadWriter is a read writer that is transparent to badger DB changes done by StorageManager.
-// It is a wrapper of the ShardedReadWriter under StorageManager.
-type ManagedReadWriter struct {
-	sm *StorageManager
-}
+	// No lock is needed here as the only writer to sm.partitioner is exactly this function.
+	pidToDelete := sm.partitioner.Inactive()
```
axw, line 286: Perhaps Rotate should also return the inactive partition ID?

x-pack/apm-server/sampling/eventstorage/partitioner.go
```diff
@@ -0,0 +1,88 @@
...
+// Callers should obtain p.mu.RLock when using the returned PID.
+func (p *Partitioner) Inactive() int {
+	return (p.current + 1) % p.total
+}
+
+// Current returns the ID of the current partition (rightmost active).
+// Callers should obtain p.mu.RLock when using the returned PID.
+func (p *Partitioner) Current() int {
+	return p.current
+}
```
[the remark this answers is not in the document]
axw: Alternatively, can the methods be made to take the read lock? Seems like it, looking at partition_rw.go, particularly now that Actives implements iter.Seq

axw (approved):
Great job! LGTM, just minor comments

x-pack/apm-server/sampling/eventstorage/partitioner.go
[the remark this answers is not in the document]
carsonip: I was thinking about making methods to take the RLock, but we can't do it on Inactive and Current, unless they return a function that provides inactive and current as func arguments, or return a seq as well. I'm also aware that adding RWMutex to partitioner makes it leak the abstraction a bit.

x-pack/apm-server/sampling/eventstorage/partitioner_test.go
```diff
@@ -0,0 +1,58 @@
...
+
+	"github.com/elastic/apm-server/x-pack/apm-server/sampling/eventstorage"
+)
+
+func iterToSlice[T any](it iter.Seq[T]) (s []T) {
+	for i := range it {
+		s = append(s, i)
+	}
+	return
+}
```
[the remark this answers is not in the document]
carsonip: my google-fu failed me

x-pack/apm-server/sampling/eventstorage/partitioner.go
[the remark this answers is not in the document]
carsonip: > `currentID` return a function that provides current as func arguments

this looks really strange and not ergonomic for the caller

> `currentID` return a seq

this could work but can be confusing and forces the caller to use a range.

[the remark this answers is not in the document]
carsonip: pushed the changes, let me know if you like it: https://github.com/elastic/apm-server/pull/15235/commits/34213e172da5039793487e5e49134afcfbb85125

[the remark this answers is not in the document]
simitt: This LGTM, but let's wait what on Andrew's feedback

simitt (approved):
Changes look great! 
This is a pretty big and _invasive_ PR, so we need to ensure to do enough manual testing before releasing the changes.

carsonip:
## Benchmarks

Comparison is done on main using badger (plus #15434 , see [run](https://github.com/elastic/apm-server/actions/runs/13015242494/job/36302670809)) and this PR (plus #15434, see [run](https://github.com/elastic/apm-server/actions/runs/13015306337/job/36302890307)) on standalone apm server sending to moxy, with 60GB gp3 disk and 60GB TBS storage limit. The benchmark are called `Traces*` as they are introduced in #15434 to remove noise from TBS benchmarks.

<details>

<summary>full benchstat</summary>

```
                      │ main-run-1217.txt │          pebble-run-1218.txt          │
                      │      sec/op       │    sec/op     vs base                 │
TracesAgentAll-512         83.69m ±   ∞ ¹   74.37m ± 32%        ~ (p=0.177 n=5+6)
TracesAgentGo-512          88.43m ± 19%     75.63m ±  9%        ~ (p=0.394 n=6)
TracesAgentNodeJS-512      74.37m ± 35%     65.38m ±  7%  -12.08% (p=0.026 n=6)
TracesAgentPython-512      84.77m ± 28%     81.04m ±  2%        ~ (p=0.818 n=6)
TracesAgentRuby-512        72.55m ± 12%     70.84m ±  9%        ~ (p=0.065 n=6)
geomean                    80.52m           73.27m         -9.01%
¹ need >= 6 samples for confidence interval at level 0.95

                      │  main-run-1217.txt  │             pebble-run-1218.txt              │
                      │ error_responses/sec │ error_responses/sec  vs base                 │
TracesAgentAll-512            11.95 ±   ∞ ¹           13.40 ± 27%        ~ (p=0.126 n=5+6)
TracesAgentGo-512             11.31 ± 24%             12.80 ±  6%        ~ (p=0.394 n=6)
TracesAgentNodeJS-512         13.14 ± 31%             14.77 ± 15%  +12.32% (p=0.026 n=6)
TracesAgentPython-512         11.82 ± 22%             12.29 ±  4%        ~ (p=0.937 n=6)
TracesAgentRuby-512           13.93 ± 12%             14.05 ±  9%        ~ (p=0.485 n=6)
geomean                       12.39                   13.43         +8.38%
¹ need >= 6 samples for confidence interval at level 0.95

                      │ main-run-1217.txt │         pebble-run-1218.txt          │
                      │    errors/sec     │ errors/sec  vs base                  │
TracesAgentAll-512           0.000 ±  ∞ ¹   0.000 ± 0%       ~ (p=1.000 n=5+6) ²
TracesAgentGo-512            0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=6)   ²
TracesAgentNodeJS-512        0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=6)   ²
TracesAgentPython-512        0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=6)   ²
TracesAgentRuby-512          0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=6)   ²
geomean                                 ³               +0.00%                 ³
¹ need >= 6 samples for confidence interval at level 0.95
² all samples are equal
³ summaries must be >0 to compute geomean

                      │ main-run-1217.txt │            pebble-run-1218.txt            │
                      │    events/sec     │  events/sec    vs base                    │
TracesAgentAll-512         218.1 ±    ∞ ¹    3334.0 ± 26%   +1428.66% (p=0.004 n=5+6)
TracesAgentGo-512          93.47 ±  92%     3627.00 ± 19%   +3780.18% (p=0.002 n=6)
TracesAgentNodeJS-512      69.37 ± 761%     1142.50 ± 15%   +1547.08% (p=0.002 n=6)
TracesAgentPython-512      11.25 ± 154%     1881.00 ± 43%  +16616.29% (p=0.002 n=6)
TracesAgentRuby-512        96.41 ± 407%     1793.50 ± 46%   +1760.38% (p=0.002 n=6)
geomean                    68.73             2.156k         +3037.07%
¹ need >= 6 samples for confidence interval at level 0.95

                      │ main-run-1217.txt │          pebble-run-1218.txt          │
                      │     gc_cycles     │  gc_cycles   vs base                  │
TracesAgentAll-512          173.0 ±   ∞ ¹   270.0 ± 30%   +56.07% (p=0.004 n=5+6)
TracesAgentGo-512           115.5 ± 18%     293.0 ± 23%  +153.68% (p=0.002 n=6)
TracesAgentNodeJS-512       134.5 ± 34%     217.5 ± 10%   +61.71% (p=0.002 n=6)
TracesAgentPython-512       108.0 ± 21%     190.0 ± 18%   +75.93% (p=0.002 n=6)
TracesAgentRuby-512         109.0 ± 10%     262.0 ± 22%  +140.37% (p=0.002 n=6)
geomean                     125.9           243.5         +93.42%
¹ need >= 6 samples for confidence interval at level 0.95

                      │     main-run-1217.txt      │                 pebble-run-1218.txt                  │
                      │ intake_events_accepted/sec │ intake_events_accepted/sec  vs base                  │
TracesAgentAll-512                  11.96k ±   ∞ ¹                 25.31k ± 20%  +111.58% (p=0.004 n=5+6)
TracesAgentGo-512                   9.732k ± 37%                  20.361k ± 23%  +109.21% (p=0.002 n=6)
TracesAgentNodeJS-512               12.33k ± 46%                   14.20k ± 31%   +15.19% (p=0.041 n=6)
TracesAgentPython-512               6.827k ± 44%                  14.128k ± 11%  +106.96% (p=0.002 n=6)
TracesAgentRuby-512                 11.29k ± 50%                   18.41k ± 33%   +63.17% (p=0.004 n=6)
geomean                             10.20k                         18.03k         +76.68%
¹ need >= 6 samples for confidence interval at level 0.95

                      │    main-run-1217.txt     │                pebble-run-1218.txt                 │
                      │ intake_events_errors/sec │ intake_events_errors/sec  vs base                  │
TracesAgentAll-512                  0.000 ±  ∞ ¹                 0.000 ± 0%       ~ (p=1.000 n=5+6) ²
TracesAgentGo-512                   0.000 ± 0%                   0.000 ± 0%       ~ (p=1.000 n=6)   ²
TracesAgentNodeJS-512               0.000 ± 0%                   0.000 ± 0%       ~ (p=1.000 n=6)   ²
TracesAgentPython-512               0.000 ± 0%                   0.000 ± 0%       ~ (p=1.000 n=6)   ²
TracesAgentRuby-512                 0.000 ± 0%                   0.000 ± 0%       ~ (p=1.000 n=6)   ²
geomean                                        ³                             +0.00%                 ³
¹ need >= 6 samples for confidence interval at level 0.95
² all samples are equal
³ summaries must be >0 to compute geomean

                      │ main-run-1217.txt │           pebble-run-1218.txt           │
                      │  max_goroutines   │ max_goroutines  vs base                 │
TracesAgentAll-512          726.0 ±   ∞ ¹       634.5 ± 6%  -12.60% (p=0.004 n=5+6)
TracesAgentGo-512           726.0 ±  6%         621.5 ± 2%  -14.39% (p=0.002 n=6)
TracesAgentNodeJS-512       760.0 ±  5%         649.0 ± 9%  -14.61% (p=0.002 n=6)
TracesAgentPython-512       682.0 ±  7%         627.5 ± 2%   -7.99% (p=0.002 n=6)
TracesAgentRuby-512         729.0 ± 12%         629.0 ± 2%  -13.72% (p=0.002 n=6)
geomean                     724.2               632.2       -12.70%
¹ need >= 6 samples for confidence interval at level 0.95

                      │ main-run-1217.txt │           pebble-run-1218.txt           │
                      │  max_heap_alloc   │ max_heap_alloc  vs base                 │
TracesAgentAll-512         1.792G ±   ∞ ¹     1.057G ±  5%  -41.01% (p=0.004 n=5+6)
TracesAgentGo-512          1.943G ±  6%       1.161G ±  6%  -40.23% (p=0.002 n=6)
TracesAgentNodeJS-512      1.900G ± 19%       1.100G ±  4%  -42.10% (p=0.002 n=6)
TracesAgentPython-512      2.100G ±  5%       1.357G ± 12%  -35.41% (p=0.002 n=6)
TracesAgentRuby-512        1.963G ±  8%       1.145G ± 16%  -41.66% (p=0.002 n=6)
geomean                    1.937G             1.160G        -40.13%
¹ need >= 6 samples for confidence interval at level 0.95

                      │ main-run-1217.txt │            pebble-run-1218.txt            │
                      │ max_heap_objects  │ max_heap_objects  vs base                 │
TracesAgentAll-512         8.205M ±   ∞ ¹       4.272M ± 11%  -47.93% (p=0.004 n=5+6)
TracesAgentGo-512         10.218M ± 16%         3.920M ± 14%  -61.64% (p=0.002 n=6)
TracesAgentNodeJS-512     11.142M ±  4%         4.349M ± 21%  -60.97% (p=0.002 n=6)
TracesAgentPython-512      9.699M ± 17%         5.453M ±  5%  -43.77% (p=0.002 n=6)
TracesAgentRuby-512       12.012M ±  9%         4.836M ± 30%  -59.74% (p=0.002 n=6)
geomean                    10.17M               4.536M        -55.40%
¹ need >= 6 samples for confidence interval at level 0.95

                      │ main-run-1217.txt │          pebble-run-1218.txt          │
                      │      max_rss      │   max_rss     vs base                 │
TracesAgentAll-512         5.532G ±   ∞ ¹   1.199G ±  6%  -78.33% (p=0.004 n=5+6)
TracesAgentGo-512          5.524G ± 22%     1.264G ±  9%  -77.12% (p=0.002 n=6)
TracesAgentNodeJS-512      5.728G ± 41%     1.219G ±  5%  -78.71% (p=0.002 n=6)
TracesAgentPython-512      4.637G ± 11%     1.486G ±  9%  -67.95% (p=0.002 n=6)
TracesAgentRuby-512        5.176G ± 14%     1.323G ± 10%  -74.43% (p=0.002 n=6)
geomean                    5.305G           1.294G        -75.60%
¹ need >= 6 samples for confidence interval at level 0.95

                      │    main-run-1217.txt    │               pebble-run-1218.txt               │
                      │ mean_available_indexers │ mean_available_indexers  vs base                │
TracesAgentAll-512                 26.09 ±  ∞ ¹                26.54 ± 1%  +1.71% (p=0.004 n=5+6)
TracesAgentGo-512                  26.07 ± 0%                  26.34 ± 0%  +1.04% (p=0.002 n=6)
TracesAgentNodeJS-512              26.06 ± 4%                  26.52 ± 0%       ~ (p=0.065 n=6)
TracesAgentPython-512              26.51 ± 2%                  26.64 ± 0%       ~ (p=1.000 n=6)
TracesAgentRuby-512                26.05 ± 4%                  26.41 ± 1%       ~ (p=0.065 n=6)
geomean                            26.16                       26.49       +1.28%
¹ need >= 6 samples for confidence interval at level 0.95

                      │ main-run-1217.txt │          pebble-run-1218.txt          │
                      │    metrics/sec    │ metrics/sec   vs base                 │
TracesAgentAll-512         5.293 ±    ∞ ¹   6.172 ± 249%        ~ (p=0.429 n=5+6)
TracesAgentGo-512          2.539 ± 105%     4.482 ±  22%  +76.49% (p=0.041 n=6)
TracesAgentNodeJS-512      4.232 ±  27%     3.893 ± 100%        ~ (p=0.485 n=6)
TracesAgentPython-512      4.173 ±  65%     4.186 ±  16%        ~ (p=0.818 n=6)
TracesAgentRuby-512        2.268 ±  71%     3.630 ±  18%        ~ (p=0.065 n=6)
geomean                    3.517            4.393         +24.89%
¹ need >= 6 samples for confidence interval at level 0.95

                      │ main-run-1217.txt │            pebble-run-1218.txt             │
                      │     spans/sec     │   spans/sec     vs base                    │
TracesAgentAll-512         176.7 ±    ∞ ¹     2337.5 ± 34%   +1222.86% (p=0.004 n=5+6)
TracesAgentGo-512          72.52 ±  94%      2524.50 ± 18%   +3381.11% (p=0.002 n=6)
TracesAgentNodeJS-512      52.07 ± 100%       607.25 ± 15%   +1066.22% (p=0.002 n=6)
TracesAgentPython-512      5.293 ±    ?     1596.000 ± 49%  +30055.88% (p=0.002 n=6)
TracesAgentRuby-512        80.95 ±    ?       867.60 ± 49%    +971.71% (p=0.002 n=6)
geomean                    49.12              1.378k         +2704.74%
¹ need >= 6 samples for confidence interval at level 0.95

                      │ main-run-1217.txt │           pebble-run-1218.txt           │
                      │   tbs_lsm_size    │ tbs_lsm_size   vs base                  │
TracesAgentAll-512          5.568G ±  ∞ ¹    7.822G ± 69%         ~ (p=0.329 n=5+6)
TracesAgentGo-512           7.367G ± 6%     14.722G ± 34%   +99.85% (p=0.002 n=6)
TracesAgentNodeJS-512       6.745G ± 6%     16.940G ± 24%  +151.17% (p=0.002 n=6)
TracesAgentPython-512       7.404G ± 9%     19.090G ± 22%  +157.84% (p=0.002 n=6)
TracesAgentRuby-512         7.653G ± 7%     15.877G ± 20%  +107.48% (p=0.002 n=6)
geomean                     6.903G           14.27G        +106.69%
¹ need >= 6 samples for confidence interval at level 0.95

                      │ main-run-1217.txt │            pebble-run-1218.txt            │
                      │   tbs_vlog_size   │ tbs_vlog_size  vs base                    │
TracesAgentAll-512         12.72G ±   ∞ ¹      0.00G ± 0%  -100.00% (n=5+6)
TracesAgentGo-512          12.31G ±  8%        0.00G ± 0%  -100.00% (p=0.002 n=6)
TracesAgentNodeJS-512      19.33G ± 51%        0.00G ± 0%  -100.00% (p=0.002 n=6)
TracesAgentPython-512      30.43G ± 11%        0.00G ± 0%  -100.00% (p=0.002 n=6)
TracesAgentRuby-512        23.72G ± 39%        0.00G ± 0%  -100.00% (p=0.002 n=6)
geomean                    18.53G                          ?                      ² ³
¹ need >= 6 samples for confidence interval at level 0.95
² summaries must be >0 to compute geomean
³ ratios must be >0 to compute geomean

                      │ main-run-1217.txt │           pebble-run-1218.txt            │
                      │      txs/sec      │    txs/sec     vs base                   │
TracesAgentAll-512         36.02 ±    ∞ ¹    594.10 ± 24%  +1549.36% (p=0.004 n=5+6)
TracesAgentGo-512          19.20 ±  85%      647.80 ± 19%  +3274.84% (p=0.002 n=6)
TracesAgentNodeJS-512      12.61 ± 100%      293.40 ± 12%  +2226.72% (p=0.002 n=6)
TracesAgentPython-512      2.667 ±    ?     151.400 ± 20%  +5577.85% (p=0.002 n=6)
TracesAgentRuby-512        11.43 ±    ?      550.65 ± 44%  +4717.59% (p=0.002 n=6)
geomean                    12.16              393.3        +3134.93%
¹ need >= 6 samples for confidence interval at level 0.95

                      │ main-run-1217.txt │          pebble-run-1218.txt           │
                      │       B/op        │     B/op       vs base                 │
TracesAgentAll-512        51.58Mi ±   ∞ ¹   58.49Mi ± 25%        ~ (p=0.662 n=5+6)
TracesAgentGo-512         43.71Mi ± 20%     58.38Mi ± 10%  +33.56% (p=0.002 n=6)
TracesAgentNodeJS-512     45.90Mi ± 15%     46.65Mi ±  8%        ~ (p=0.699 n=6)
TracesAgentPython-512     50.60Mi ±  7%     51.34Mi ±  7%        ~ (p=0.589 n=6)
TracesAgentRuby-512       41.50Mi ± 16%     50.19Mi ±  9%  +20.94% (p=0.009 n=6)
geomean                   46.50Mi           52.80Mi        +13.56%
¹ need >= 6 samples for confidence interval at level 0.95

                      │ main-run-1217.txt │          pebble-run-1218.txt          │
                      │     allocs/op     │  allocs/op    vs base                 │
TracesAgentAll-512         257.2k ±   ∞ ¹   443.0k ± 40%        ~ (p=0.126 n=5+6)
TracesAgentGo-512          226.6k ± 30%     396.5k ±  7%  +74.96% (p=0.002 n=6)
TracesAgentNodeJS-512      351.7k ± 13%     358.9k ± 20%        ~ (p=0.589 n=6)
TracesAgentPython-512      275.2k ± 20%     445.1k ± 11%  +61.77% (p=0.002 n=6)
TracesAgentRuby-512        276.2k ± 30%     412.0k ± 14%  +49.14% (p=0.002 n=6)
geomean                    274.5k           409.8k        +49.29%
¹ need >= 6 samples for confidence interval at level 0.95
```

</details>

### Highlights

#### Elasticsearch indexing rate

Bear in mind that aggregated documents are also included here.

```
                      │ main-run-1217.txt │            pebble-run-1218.txt            │
                      │    events/sec     │  events/sec    vs base                    │
TracesAgentAll-512         218.1 ±    ∞ ¹    3334.0 ± 26%   +1428.66% (p=0.004 n=5+6)
TracesAgentGo-512          93.47 ±  92%     3627.00 ± 19%   +3780.18% (p=0.002 n=6)
TracesAgentNodeJS-512      69.37 ± 761%     1142.50 ± 15%   +1547.08% (p=0.002 n=6)
TracesAgentPython-512      11.25 ± 154%     1881.00 ± 43%  +16616.29% (p=0.002 n=6)
TracesAgentRuby-512        96.41 ± 407%     1793.50 ± 46%   +1760.38% (p=0.002 n=6)
geomean                    68.73             2.156k         +3037.07%
¹ need >= 6 samples for confidence interval at level 0.95
```

#### Intakev2 event rate

```
                      │     main-run-1217.txt      │                 pebble-run-1218.txt                  │
                      │ intake_events_accepted/sec │ intake_events_accepted/sec  vs base                  │
TracesAgentAll-512                  11.96k ±   ∞ ¹                 25.31k ± 20%  +111.58% (p=0.004 n=5+6)
TracesAgentGo-512                   9.732k ± 37%                  20.361k ± 23%  +109.21% (p=0.002 n=6)
TracesAgentNodeJS-512               12.33k ± 46%                   14.20k ± 31%   +15.19% (p=0.041 n=6)
TracesAgentPython-512               6.827k ± 44%                  14.128k ± 11%  +106.96% (p=0.002 n=6)
TracesAgentRuby-512                 11.29k ± 50%                   18.41k ± 33%   +63.17% (p=0.004 n=6)
geomean                             10.20k                         18.03k         +76.68%
¹ need >= 6 samples for confidence interval at level 0.95
```

#### Memory usage (max_rss)

```
                      │ main-run-1217.txt │          pebble-run-1218.txt          │
                      │      max_rss      │   max_rss     vs base                 │
TracesAgentAll-512         5.532G ±   ∞ ¹   1.199G ±  6%  -78.33% (p=0.004 n=5+6)
TracesAgentGo-512          5.524G ± 22%     1.264G ±  9%  -77.12% (p=0.002 n=6)
TracesAgentNodeJS-512      5.728G ± 41%     1.219G ±  5%  -78.71% (p=0.002 n=6)
TracesAgentPython-512      4.637G ± 11%     1.486G ±  9%  -67.95% (p=0.002 n=6)
TracesAgentRuby-512        5.176G ± 14%     1.323G ± 10%  -74.43% (p=0.002 n=6)
geomean                    5.305G           1.294G        -75.60%
¹ need >= 6 samples for confidence interval at level 0.95
```

#### Disk usage

This is slightly trickier to understand, as pebble doesn't use vlog. Therefore, the 2 metrics should be read together (i.e. compare tbs_lsm_size + tbs_vlog_size). Comparison on geomean will then be main 6.903+18.53G vs pebble 14.27G.

```
                      │ main-run-1217.txt │           pebble-run-1218.txt           │
                      │   tbs_lsm_size    │ tbs_lsm_size   vs base                  │
TracesAgentAll-512          5.568G ±  ∞ ¹    7.822G ± 69%         ~ (p=0.329 n=5+6)
TracesAgentGo-512           7.367G ± 6%     14.722G ± 34%   +99.85% (p=0.002 n=6)
TracesAgentNodeJS-512       6.745G ± 6%     16.940G ± 24%  +151.17% (p=0.002 n=6)
TracesAgentPython-512       7.404G ± 9%     19.090G ± 22%  +157.84% (p=0.002 n=6)
TracesAgentRuby-512         7.653G ± 7%     15.877G ± 20%  +107.48% (p=0.002 n=6)
geomean                     6.903G           14.27G        +106.69%
¹ need >= 6 samples for confidence interval at level 0.95

                      │ main-run-1217.txt │            pebble-run-1218.txt            │
                      │   tbs_vlog_size   │ tbs_vlog_size  vs base                    │
TracesAgentAll-512         12.72G ±   ∞ ¹      0.00G ± 0%  -100.00% (n=5+6)
TracesAgentGo-512          12.31G ±  8%        0.00G ± 0%  -100.00% (p=0.002 n=6)
TracesAgentNodeJS-512      19.33G ± 51%        0.00G ± 0%  -100.00% (p=0.002 n=6)
TracesAgentPython-512      30.43G ± 11%        0.00G ± 0%  -100.00% (p=0.002 n=6)
TracesAgentRuby-512        23.72G ± 39%        0.00G ± 0%  -100.00% (p=0.002 n=6)
geomean                    18.53G                          ?                      ² ³
¹ need >= 6 samples for confidence interval at level 0.95
² summaries must be >0 to compute geomean
³ ratios must be >0 to compute geomean
```

## Summary

Seems like a win here as there are net gains on all ends. This is also just a starting point as we haven't spent much time tuning the Pebble options and microbenchmarks for allocs.

x-pack/apm-server/sampling/eventstorage/partitioner.go
```diff
@@ -0,0 +1,91 @@
...
+//
+// As ActiveIDs holds RLock internally,
+// - Rotate should never be called within ActiveIDs
+// - the returned partition IDs may be outdated if used outside `range`
+func (p *Partitioner) ActiveIDs() iter.Seq[int] {
+	return func(yield func(int) bool) {
+		p.mu.RLock()
+		defer p.mu.RUnlock()
+		for i := 0; i < p.total-1; i++ {
+			if !yield((p.current + p.total - i) % p.total) {
```
1pkg, line 75: I'm not sure if want to consider adding some shuffling when active partitions are selected here. Right now we will always perform all operations against active partitions in descending order i.e. A, A-1, A-2, ... It's likely that it could be the most efficient order for an average case, however it'll definitely amplify reads against first active partitions significantly.

Maybe we should update PartitionReadWriter to collect some stats on how often ErrNotFound happens to have some data on this in the future.

x-pack/apm-server/sampling/eventstorage/partitioner_test.go
```diff
@@ -0,0 +1,60 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
```
1pkg, line 1: Could we also add some tests to validate that `ActiveIDs` works as expected. Plus the logic for `ActiveIDs`, `CurrentIDFunc` with a concurrent rotate.

x-pack/apm-server/sampling/eventstorage/prefix.go
```diff
@@ -0,0 +1,132 @@
...
+
+	"github.com/cockroachdb/pebble/v2"
+
+	"github.com/elastic/apm-data/model/modelpb"
+)
+
+const (
+	// NOTE(axw) these values (and their meanings) must remain stable
+	// over time, to avoid misinterpreting historical data.
+	entryMetaTraceSampled   byte = 's'
```
1pkg, line 20: Any reasons to not percolate these markers as slices in global variables directly?

x-pack/apm-server/sampling/eventstorage/storage_manager.go
```diff
@@ -26,315 +25,300 @@ const (
...
+// such that partitioner resumes from where it left off before an apm-server restart.
+func (sm *StorageManager) loadPartitionID() (int, error) {
+	item, closer, err := sm.decisionDB.Get([]byte(partitionerMetaKey))
+	if errors.Is(err, pebble.ErrNotFound) {
+		return 0, nil
+	} else if err != nil {
+		return 0, err
+	}
+	defer closer.Close()
+	var pid struct {
```
1pkg, line 150: nit: why not just use the same structure for both marshal and unmarshal in `loadPartitionID` and `savePartitionID` to avoid any implementation drifts in the future?

x-pack/apm-server/sampling/eventstorage/partitioner.go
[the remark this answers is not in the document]
axw: Ah, I previously missed the fact that we needed to stop the rotation to prevent current being invalidated. Change looks good though. Don't love the name "CurrentIDFunc", but also don't have a compelling suggestion :)

axw (dismissed):

1pkg (dismissed):
LGTM!

One request from me, could we please add a description to all struct that implement `RW` interface. There are couple of them and it takes time to understand how they interact, it'll save time in the future if we have some light description to explain their purpose.

x-pack/apm-server/sampling/eventstorage/partitioner.go
carsonip: i think the current setup is ok, given default ttl=30m and partitionsPerTTL=1. This means on every reads, it will first search the most recent TTL=30mins (obviously after after a rotation, as the partition is new, it can be less than 30mins). Even if ttl is lower, we still expect most traces to come in within minutes, if not seconds or milliseconds, and searching the most recent partition is the most logical decision.

x-pack/apm-server/sampling/eventstorage/storage_manager.go
```diff
@@ -26,315 +27,335 @@ const (
...
 	}
-	err := sm.reset()
-	if err != nil {
-		return nil, err
+	for _, opt := range opts {
+		opt(sm)
+	}
+
+	if sm.meterProvider != nil {
+		meter := sm.meterProvider.Meter("github.com/elastic/apm-server/sampling/eventstorage")
```
simitt, line 111: Shouldn't it be `meter := sm.meterProvider.Meter("github.com/elastic/apm-server/x-pack/apm-server/sampling/eventstorage")`?

carsonip: Nice catch. Shouldn't code at night.

x-pack/apm-server/sampling/config.go
```diff
@@ -95,24 +95,16 @@ type DataStreamConfig struct {
...
-	// is closed
-	Storage rw
-
-	// StorageDir holds the directory in which event storage will be maintained.
-	StorageDir string
-
-	// StorageGCInterval holds the amount of time between storage garbage collections.
-	StorageGCInterval time.Duration
+	// Storage overrides the default DB storage RW.
+	// For testing only.
```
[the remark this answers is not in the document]
carsonip: Fixed. This is not only for testing now.

simitt (approved):
Amazing work!

x-pack/apm-server/sampling/eventstorage/partitioner.go
1pkg: Agree with `partitionsPerTTL=1` this doesn't matter much, but if we ever consider increasing `partitionsPerTTL` value we should revisit this.

x-pack/apm-server/sampling/config.go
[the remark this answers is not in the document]
axw: Thanks!

PR #15245, opened 2025-01-14
build: move notice make tasks to go

kruskall:
## Motivation/summary

autopep8 is causing the lint task to fail complaining about lib2to3 being missing

update the notice generation go to use the go command (similar to what we do in 8 and main)
skip linting python files during check task
bump indirect dependency to pass license check

## Checklist


go.mod
```diff
@@ -138,7 +138,7 @@ require (
 	github.com/gorilla/mux v1.8.0 // indirect
 	github.com/gorilla/websocket v1.5.0 // indirect
 	github.com/h2non/filetype v1.1.1 // indirect
-	github.com/hashicorp/cronexpr v1.1.0 // indirect
+	github.com/hashicorp/cronexpr v1.1.2 // indirect
```
carsonip, line 141: Q: will the fix work without dep bump?

carsonip: It is auto-detected as Apache-2.0 although the file is not called LICENSE. Is the file name the problem?

rubvs (approved):

PR #15278, opened 2025-01-17
ci(dra): version should not include the qualifier

v1v:
## Motivation/summary

When introducing the `qualifier`, the version changed. This should help keep the expected behaviour and use a new target goal to avoid the qualifier in the name.

## Checklist

## How to test these changes

In the CI:

1) `dry-run mode` without `qualifier` -> https://buildkite.com/elastic/apm-server-package/builds/3707
2) `dry-run mode` with `qualifier` -> https://buildkite.com/elastic/apm-server-package/builds/3708

### DIFF
--- a/.buildkite/scripts/dra.sh
+++ b/.buildkite/scripts/dra.sh
@@ -13,8 +13,8 @@ set -eo pipefail
 # Either staging or snapshot
 TYPE="$1"
 
-## Read current version.
-VERSION=$(make get-version)
+## Read current version without the qualifier
+VERSION=$(make get-version-only)
 
 echo "--- Restoring Artifacts"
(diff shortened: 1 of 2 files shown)


.buildkite/scripts/dra.sh
```diff
@@ -13,8 +13,8 @@ set -eo pipefail
 # Either staging or snapshot
 TYPE="$1"
 
-## Read current version.
-VERSION=$(make get-version)
+## Read current version without the qualifier
+VERSION=$(make get-version-only)
```
v1v, line 17: DRA requires a version with the format `major.minor.patch(-SNAPSHOT)` but without the `qualifier`. 

`make get-version` changed to support the qualifier, but it also changed the contract with the `DRA`.

PR #15303, opened 2025-01-20
terraform: add tags module

v1v:
## Motivation/summary

The current CI integration with the Terraform `ec_deployment` uses OIDC; hence, it uses the robots automation, but unfortunately, if `ephemeral:true` is not set and the `tear-down` steps fail in the CI, those resources are not deleted automatically but manually.

We want to enforce using `ephemeral:true` in all the Terraform resources using `ec`, the `ephemeral:true` was added as part of https://github.com/elastic/apm-server/pull/14179

But I didn't see the `tags` to be set for the `cloud` and `functional tests` that run in the CI or can run.

## Checklist

### DIFF
--- a/testing/cloud/main.tf
+++ b/testing/cloud/main.tf
@@ -8,6 +8,11 @@ terraform {
   }
 }
 
+module "tags" {
+  source = "../infra/terraform/modules/tags"
+  project = "cloud"
+}
+
 provider "ec" {}
 
 locals {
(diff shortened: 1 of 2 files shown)


testing/cloud/main.tf
```diff
@@ -8,6 +8,11 @@ terraform {
   }
 }
 
+module "tags" {
+  source = "../infra/terraform/modules/tags"
```
v1v, line 12:
```suggestion
  source  = "../infra/terraform/modules/tags"
```

endorama (approved):

Issue #14100, opened 2024-09-17
Add functional tests for ILM & Data Stream Lifecycle


axw:
We should implement functional tests that verify ILM or Data Stream Lefecycle Management is used as expected according to user configuration, and depending on whether a cluster is created new or upgraded. See https://github.com/elastic/apm-server/issues/13898#issuecomment-2326277518

endorama:
I've been thinking how to implement functional tests. I explored 3 possible approaches:
1. as part of smoke tests
2. as part of system tests
3. with a new approach independent of the 2 above

TL;DR: my conclusion is that we may want to proceed with option 3.

Option 1 heavy leverages ESS, creating infrastructure using terraform, on top of a custom built Bash framework. We do not need leveraging ESS for functional testing and the framework looks built with that purpose. Moreover Bash flexibility comes at a great readability costs. Orchestration is also not clear to me, and documentation is lacking.

Option 2 uses Go tests, but is strictly focused on testing specific APM server behaviors. As we now rely on `apm-data` plugin we may need to test interactions with different Elasticsearch versions and more complex interactions, which do not fit the current model (that starts a single stack against which runs all tests).

A third option would be to build a framework that looks similar to Option 2, with guarantees provided by Go, but with a scope more similar to Option 1. I created a simple stub in https://github.com/endorama/apm-server/blob/af3ca3744e4746d4c6d7f65a162927d6c9e19331/functionaltests/main_test.go#L31 using `testcontainers` library and starting a specific Elasticsearch version in a container. The idea here would be to build the freedom to leverage single stack components, allowing us to express complex upgrade and assertion scenarios.
I have some concerns on this approach:
- we would duplicate packages already existing in `systemtests`, but that looks like are lacking the flexibility we would need to express complex cases
- we would duplicate some logic of smoke tests, but relying on docker would allow faster testing and using Go easier maintenance in the long term

The third option looks the best to me if I think at future use cases (es adding further cases to this logic https://github.com/elastic/apm-server/pull/13678 is more difficult in Bash than in Go), and I think a test framework must be ergonomic enough to encourage use.

inge4pres:
> depending on whether a cluster is created new or upgraded.

The upgrade part is going to be especially tricky IMO, because IIANM Elasticsearch will never allow an upgrade between a released version (e.g. 8.15.3) and an un-released one (SNAPSHOT).
If this is correct, I have no idea on how we could run the test for an upgrade _before_ a release is created.

[...]

PR #15317, opened 2025-01-21
functionaltests: wait for aggregation data steams

endorama:
## Motivation/summary

[Today](https://github.com/elastic/apm-server/actions/runs/12879293419/job/35906566924#step:7:66)'s `functional-tests` run failed due to executing the first data stream check before all data stream we expect to be present were there.
This may happen due to the 1m aggregation data streams being not yet created if the data stream check happens too fast.
Ideally this problem was fixed with the initial implementation, but it only reduced its occurrence.
This PR introduces an additional check in `RunBlocking` to verify if other dta strea'ms are present, ignoring the document count (which is problematic for aggregation data streams).

## How to test these changes

https://github.com/elastic/apm-server/actions/runs/12892900589

## Related issues

https://github.com/elastic/apm-server/issues/14100

### DIFF
--- a/functionaltests/internal/gen/generator.go
+++ b/functionaltests/internal/gen/generator.go
@@ -71,6 +71,8 @@ func (g *Generator) RunBlocking(ctx context.Context) error {
 // document count retrieved through ecc matches expected or timeout. It supports
 // specifying previous document count to offset the expectation based on a previous state.
 // expected and previous must be maps of <datastream name, document count>.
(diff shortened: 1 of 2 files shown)


endorama:
We decided to take a different approach than what is in this PR. This approach has been implemented in #15356.

I'm closing this PR as superseded by the other one. Thanks @kruskall for the review :rocket:

PR #15324, opened 2025-01-22
Refactor global labels removal in metrics aggregation process

up2neck:
## Motivation/summary

Code for aggregation in x-pack seems a bit improperly structured. Improved it for better readability/understanding.

## Checklist


x-pack/apm-server/aggregation/lsm.go
```diff
@@ -108,17 +113,17 @@ func wrapNextProcessor(processor modelpb.BatchProcessor) aggregators.Processor {
...
-	default:
-		return
+		return true
 	}
+	return false
+}
 
-	// Setting the labels to non-global so that they are ignored by the aggregator.
+// Setting the labels to non-global so that they are ignored by the aggregator.
+func removeGlobalLabels(event *modelpb.APMEvent) {
```
simitt, line 126: The structure you are proposing reads nicely. However, we are not intending, and it would be harmful, to remove global labels for anything non-rum in the future. Therefore I prefer the current code which is encapsulating the logic into one combined function. It prevents our future selves from just reusing the `removeGlobalLabels` function without the `isRumAgent` safeguard. 

Having this said, both implementations generally look fine to me, but I don't see an improvement by your suggestion over what is currently implemented.

simitt (changes requested):
Thank you for your contribution. However, I don't see this as an improvement for the reasons explained.

PR #15339, opened 2025-01-23
Update benchmark warmup time

### LINKED ISSUE #14948: Increase benchmark warmup time
Having rerun benchmarks for 8.16.0 we had discovered some high variances for AWS and Azure in the results due to CPU credits not being fully burnt through for small instance (<8gb), see [APM Dev](https://github.com/elastic/apm-dev/tree/main/docs/benchmarks/throughtput) for a detailed breakdown. As a result, we should increase our Nightly Benchmark warmup time from 5mins to 10mins.

ericywl:
## Motivation/summary

Fixes #14948

## Checklist

- Trim some words from benchmark workflow inputs to be more succinct
- Add benchmark warmup time to workflow inputs
- Increase benchmark warmup time for nightly non-standalone to 10m
- Reduce benchmark warmup time for scheduled standalone to 3m

## How to test these changes

Run [throughput benchmark](https://github.com/elastic/apm-dev/tree/main/docs/benchmarks/throughtput)

## Appendix

Old workflow inputs:
Screenshot 2025-02-06 at 4 14 34 PM

New workflow inputs:
Screenshot 2025-02-06 at 7 24 57 PM

[Test benchmark run on upstream](https://github.com/elastic/apm-server/actions/runs/13177698152)


ericywl:
@rubvs From what I can gather, the benchmark warmup time is configured by `BENCHMARK_WARMUP_TIME`. Let me know if I missed anything. I will try to run the benchmarks and verify.

rubvs:
@ericywl Correct. This should be the only change to make. You move this out-of-draft and let's get feedback from the rest of the team.

With this issue there isn't much "testing" you can do. I've already concluded this through running many benchmarks. However, this is a great first-issue:

- It'll force you to read and hopefully improve the docs, see [Throughput](https://github.com/elastic/apm-dev/tree/main/docs/benchmarks/throughtput).
- Trigger a benchmark in GitHub Actions that run on ESS.
- Trigger a benchmark from your local machine that runs in ESS.

carsonip (commented):
This is good for ESS, but probably not that helpful for standalone apm server (moxy) setup. Can we keep the 5m default for standalone apm setup?

1pkg:
> This is good for ESS, but probably not that helpful for standalone apm server (moxy) setup. Can we keep the 5m default for standalone apm setup?

We probably even should reduce it for standalone to not waste resources.

ericywl:
> This is good for ESS, but probably not that helpful for standalone apm server (moxy) setup. Can we keep the 5m default for standalone apm setup?

May I know what is ESS? Also I'm not sure if there's a way to separate between the two.

carsonip:
> I'm not sure if there's a way to separate between the two. If not I will simply close the PR.

The new default can be set in terraform files e.g. `testing/benchmark/main.tf`, just not in Makefile.

ericywl:
QQ: AFAIK you cannot configure benchmark warmup time or benchmark time etc. through GH actions currently. Would that be a good to have?

.github/workflows/benchmarks.yml
```diff
@@ -71,7 +76,7 @@ jobs:
       TF_VAR_apm_server_tail_sampling_storage_limit: ${{ inputs.tailSamplingStorageLimit || '10GB' }} # set the default again otherwise schedules won't work
       RUN_STANDALONE: ${{ inputs.runStandalone || github.event.schedule=='0 0 1 * *' }}
       PGO_EXPORT: ${{ inputs.pgoExport || github.event.schedule=='0 0 1 * *' }}
-      TFVARS_SOURCE: ${{ inputs.profile || 'system-profiles/8GBx1zone.tfvars' }} # // Default to use an 8gb profile
```
carsonip, line 74: scheduled workflow runs do not respect inputs defaults. Can you revert this change and add a comment like the one I added above `# set the default again otherwise schedules won't work`? Fair bit of time wasted on this

.github/workflows/benchmarks.yml
```diff
@@ -4,39 +4,44 @@ on:
...
+        description: 'Storage size limit of tail-based sampling on APM server'
         required: false
         type: string
         default: "10GB"
       profile:
-        description: 'The system profile used to run the benchmarks'
+        description: 'System profile for benchmark'
         required: false
         type: string
+        default: "system-profiles/8GBx1zone.tfvars"
```
carsonip, line 25: as explained below

.github/workflows/benchmarks.yml
```diff
@@ -4,39 +4,44 @@ on:
...
+        description: 'Number of benchmark agents sending data to APM Server'
         required: false
         type: string
       benchmarkRun:
-        description: 'Set the expression that matches the benchmark scenarios to run'
+        description: 'Expression for benchmark scenarios to run'
+        required: false
+        type: string
+      warmupTime:
+        description: 'Benchmark warmup time for APM Server'
```
carsonip, line 40:
```suggestion
        description: 'Benchmark warmup time for APM Server, e.g. 5m'
```
On the GH Actions UI, it is unclear what's the format for the warmup time input. People with a Go background _might_ assume that it takes a time unit, while Python people _might_ assume it is in seconds.

.github/workflows/benchmarks.yml
```diff
@@ -4,39 +4,44 @@ on:
...
         type: boolean
         default: false
       benchmarkAgents:
-        description: 'Set the number of agents to send data to the APM Server'
+        description: 'Number of benchmark agents sending data to APM Server'
         required: false
         type: string
       benchmarkRun:
-        description: 'Set the expression that matches the benchmark scenarios to run'
+        description: 'Expression for benchmark scenarios to run'
```
carsonip, line 36: This description is a little too brief. The original idea is the regex that matches benchmark scenarios. Now it is even less clear. See also `go help testflag`
```
        -run regexp
            Run only those tests, examples, and fuzz tests matching the regular
            expression. For tests, the regular expression is split by unbracketed
            slash (/) characters into a sequence of regular expressions, and each
            part of a test's identifier must match the corresponding element in
            the sequence, if any. Note that possible parents of matches are
            run too, so that -run=X/Y matches and runs and reports the result
            of all tests matching X, even those without sub-tests matching Y,
            because it must run them to look for those sub-tests.
            See also -skip.
```

carsonip (changes requested):
Can you also test the PR by pushing to a branch to upstream apm-server repo and triggering workflow runs manually in the GH Actions UI?

.github/workflows/benchmarks.yml
ericywl: Ok got it

.github/workflows/benchmarks.yml
ericywl: Got it, I will add mention of regex as well

.github/workflows/benchmarks.yml
```diff
@@ -4,39 +4,43 @@ on:
...
         type: boolean
         default: false
       benchmarkAgents:
-        description: 'Set the number of agents to send data to the APM Server'
+        description: 'Number of benchmark agents sending data to APM Server'
         required: false
         type: string
       benchmarkRun:
-        description: 'Set the expression that matches the benchmark scenarios to run'
+        description: 'Run only benchmark scenarios that match regex, i.e. "go test -run"'
```
carsonip, line 35:
```suggestion
        description: 'Run only benchmark scenarios that match regex, e.g. BenchmarkAgentAll
```

sorry I didn't describe it well. It isn't actually `go test -run`, but we mimic `go test -run` in https://github.com/elastic/apm-server/blob/main/systemtest/benchtest/config.go#L48 . But in this case I think it is slightly more helpful to provide an example.

testing/benchmark/README.md
```diff
@@ -65,7 +65,8 @@ The main commands are:
   - `BENCHMARK_TIME`: Set the amount of time to run each benchmark scenario for. Defaults to `2m`.
```
carsonip, line 65: update the new benchmark warmup time default

testing/benchmark/README.md
```diff
@@ -65,7 +65,8 @@ The main commands are:
...
   - `BENCHMARK_RUN`: Set the expression that matches the benchmark scenarios to run. Defaults to `Benchmark` (all).
   - `BENCHMARK_RESULT`: Set the output file where the results of the benchmark will be written. Defaults to `benchmark-result.txt`
-  - `BENCHMARK_DETAILED`: Sets the `-detailed` when running `apmbench`, displaying extra metrics for each benchmark. Defaults to `false`.
+  - `BENCHMARK_DETAILED`: Set the `-detailed` when running `apmbench`, displaying extra metrics for each benchmark. Defaults to `false`.
+  - `BENCHMARK_EVENT_RATE`: Set the rate of events sent to the APM Server.
```
carsonip, line 69: describe the default for consistency although I'm not a big fan of describing defaults in a readme as they get outdated over time, but that's a separate discussion.

testing/benchmark/README.md
carsonip: oh, forget about this. The testing/benchmark default didn't change. What changed was the github workflow setup only

carsonip (dismissed):
Other than the nits above, lgtm. Great work!

testing/benchmark/README.md
ericywl: Ok, let me go find out the defaults

testing/benchmark/README.md
```diff
@@ -65,7 +65,8 @@ The main commands are:
   - `BENCHMARK_TIME`: Set the amount of time to run each benchmark scenario for. Defaults to `2m`.
   - `BENCHMARK_RUN`: Set the expression that matches the benchmark scenarios to run. Defaults to `Benchmark` (all).
   - `BENCHMARK_RESULT`: Set the output file where the results of the benchmark will be written. Defaults to `benchmark-result.txt`
-  - `BENCHMARK_DETAILED`: Sets the `-detailed` when running `apmbench`, displaying extra metrics for each benchmark. Defaults to `false`.
+  - `BENCHMARK_DETAILED`: Set the `-detailed` when running `apmbench`, displaying extra metrics for each benchmark. Defaults to `false`.
+  - `BENCHMARK_EVENT_RATE`: Set the rate of events sent to the APM Server. Defaults to `0/s` (set in `apm-perf`), which means unlimited.
```
carsonip, line 69:
```suggestion
  - `BENCHMARK_EVENT_RATE`: Set the rate of events sent to the APM Server. Defaults to `0/s`, which means unlimited.
```
To be exact, the default is set in `testing/benchmark/Makefile`

carsonip (approved):

.github/workflows/benchmarks.yml
```diff
@@ -4,39 +4,43 @@ on:
...
         type: boolean
         default: false
       benchmarkAgents:
-        description: 'Set the number of agents to send data to the APM Server'
+        description: 'Number of benchmark agents sending data to APM Server'
         required: false
         type: string
       benchmarkRun:
-        description: 'Set the expression that matches the benchmark scenarios to run'
+        description: 'Run only benchmark scenarios that match regex, e.g. BenchmarkAgentAll'
```
1pkg, line 35: Should it be something like this to be consistent with other description changes?

```suggestion
        description: 'Benchmark scenarios that only match regex, e.g. BenchmarkAgentAll'
```

1pkg (approved):

testing/benchmark/README.md
ericywl: Right I forgot that its set to 0 there

1pkg (approved):

carsonip (approved):

PR #15355, opened 2025-01-23
functionaltests: handle terraform output errors

endorama:
## Motivation/summary

Code in functionaltests relying on `terraform.Runner.Output()` was not checking errors. If for any reason the output
value is wrong and the function returns an error this was ignored and the test would proceed, failing at a later
stage with ambiguous errors.

Add a clarifying error if the requested output name is not available from the Terraform outputs.

## How to test these changes

Run the functionaltests. I tested it locally.

### DIFF
--- a/functionaltests/8_15_test.go
+++ b/functionaltests/8_15_test.go
@@ -40,7 +40,7 @@ func TestUpgrade_8_15_4_to_8_16_0(t *testing.T) {
 	start := time.Now()
 	ctx := context.Background()
 
-	t.Log("creating deploment with terraform")
+	t.Log("creating deployment with terraform")
 	tf, err := terraform.New(t, t.Name())
 	require.NoError(t, err)
 	ecTarget := terraform.Var("ec_target", *target)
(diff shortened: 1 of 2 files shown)


functionaltests/8_15_test.go
```diff
@@ -60,11 +60,16 @@ func TestUpgrade_8_15_4_to_8_16_0(t *testing.T) {
 	})
 
 	var escfg esclient.Config
-	tf.Output("apm_url", &escfg.APMServerURL)
-	tf.Output("es_url", &escfg.ElasticsearchURL)
-	tf.Output("username", &escfg.Username)
-	tf.Output("password", &escfg.Password)
-	tf.Output("kb_url", &escfg.KibanaURL)
+	err = tf.Output("apm_url", &escfg.APMServerURL)
+	require.NoError(t, err)
```
marclop, line 64: Minor nit to reduce the number of lines, I think it's still quite readable:
```suggestion
	require.NoError(t, tf.Output("apm_url", &escfg.APMServerURL))
```

endorama: Agreed and updated

PR #14935, opened 2024-12-12
functionaltests: add first test case

endorama:
## Motivation/summary

Implement first test verifying upgrade from 8.15.4 to 8.16.0 for a fresh Elastic Cloud deployment.

Provides a new folder structure and helpers to quickly implement functional tests for APM Server on Elastic Cloud.

## Checklist

None

## How to test these changes

Run `cd functionaltests && go test -v ./`

## Related issues

Part of https://github.com/elastic/apm-server/issues/14100

PR #15356, opened 2025-01-23
functionaltests: restart Integrations Server after ingestion

endorama:
## Motivation/summary

This is an alternative https://github.com/elastic/apm-server/pull/15317.

Instead of expanding the logic to wait for ingestion completion by monitoring data streams and document count, rely on a side effect of shutting down the APM Server. On shutdown APM Server flushes all the data, included the aggregation for the data already ingested.

We use this behaviour triggering a restart of the Integration Server after the test data ingestion is done to ensure ingestion in completed before moving to checking document counts and data streams details.

I run 5 tests and in most cases the restart took ~1 minutes. In 1 case it took 6 and in 1 case it took more than 10 minutes (but the code has a timeout at 10 minutes).

In 1 case the test failed due to race condition when testing that the lazy rollover happened: it did not at the moment of testing but I confirmed manually that it indeed happened. This behaviour suggests lazy rollover may happen later than expected (at ingestion after an upgrade) and is not impacted by this change, but further investigation is required to prevent flakyness.

## How to test these changes

Functional tests pipeline for this branch should be green: https://github.com/elastic/apm-server/actions/workflows/functional-tests.yml?query=branch%3Afunctionaltests-restartapmserver

## Related issues

https://github.com/elastic/apm-server/issues/14100


functionaltests/main_test.go
```diff
@@ -116,3 +116,14 @@ func regionFrom(target string) string {
 		panic("target value is not accepted")
 	}
 }
+
+func endpointFrom(target string) string {
+	switch target {
+	case "qa":
+		return "https://public-api.qa.cld.elstc.co"
+	case "production":
+		return "https://api.elastic-cloud.com"
```
marclop, line 125: Won't this panic? https://github.com/elastic/apm-server/blob/4997d07805bcc55c64a2d58cffec75e73c80ce54/.github/workflows/functional-tests.yml#L28

endorama: yes. I updated it in [e528ec5](https://github.com/elastic/apm-server/pull/15356/commits/e528ec5ff28115f7fbcc00a1a33390f4481615a4) and also moved to using constants to reduce the chance of this occurring. In parallel I'm discussing how to make this clearer in https://github.com/elastic/apm-server/pull/14935#discussion_r1906393191

Issue #13738, opened 2024-07-22
Investigate nightly benchmarks 0 events/s issue


carsonip:
Nightly benchmarks occasionally report 0 events/s. Investigate the root cause of it.

lahsivjar:
## Status update

The first thing I looked at was what was getting reported by the benchmark failures. Here are 2 links to the benchmark run:

1. [Run with events/sec metric populated](https://github.com/elastic/apm-server/actions/runs/9978413243/job/27575260880) - [Link to APM-Server logs](https://platform-logging.kb.us-west2.gcp.elastic-cloud.com/app/r/s/HRaiU) - [Link to deployment](https://admin.found.no/deployments/ec3dfaed7456f6e248c674e1d758e85e/integrations_server)
2. [Run without events/sec metric populated](https://github.com/elastic/apm-server/actions/runs/9995659257/job/27628309676) - [Link to APM-Server logs](https://platform-logging.kb.us-west2.gcp.elastic-cloud.com/app/r/s/e7vG3) - [Link to deployment](https://admin.found.no/deployments/bad195b56131647756f8eda18c3ac26d/integrations_server)

Both of these show `500 internal error`, however, the logs for `0` events/sec additionally show `data validation errors` due to unexpected EOF. This could be an issue with our sender, however, the most intriguing thing is why only a subset of delta metrics are reported as `0`. For example: in the above link, the txn/sec and metrics/sec are reported correctly whereas other delta metrics are reported as zero.

I have tried reproducing the errors locally but haven't succeeded (note that the expvar metrics collection is designed for benchtimes in minutes so if testing locally make sure that you have a good enough benchtime to give expvar metrics to work correctly). I did see some special handling in the expvar metric collection but nothing explains this bug.

I have also created a PR to [log errors in expvar endpoint](https://github.com/elastic/apm-server/pull/13778) which was not done before. I am not sure how helpful it will be though.

simitt:
Is this still happening?

rubvs:
@simitt I had this happen to me in a run on GH Actions last week, see Slack Thread: https://elastic.slack.com/archives/C95SB62AG/p1729263104854879

raultorrecilla:
moving this out of iteration, to backlog. If it happens again more frequently we can reprioritize.
Out of curiosity, how many times this happened between end of October and now? cc @rubvs

rubvs:
[...]

PR #15360, opened 2025-01-24
 feat: start hiding beats monitoring behind otel abstraction 

kruskall:
## Motivation/summary

Split from https://github.com/elastic/apm-server/pull/15094/ 

part 1

## Checklist


internal/beater/server_test.go
```diff
@@ -487,44 +496,11 @@ func TestServerElasticsearchOutput(t *testing.T) {
...
-				"completed": int64(0),
-			},
-			"indexers": map[string]interface{}{
-				"active":    int64(1),
-				"destroyed": int64(0),
-				"created":   int64(0),
-			},
-		},
-	}, snapshot)
+	monitoringtest.ExpectOtelMetrics(t, reader, map[string]any{
```
axw, line 499: It seems we're missing a bunch of non-zero metrics that we had before, is that expected?

 - `elasticsearch.indexers.active`
 - `output.events.active`
 - `output.events.total`
 - `pipeline.events.total`

internal/beater/beater.go
```diff
@@ -90,7 +91,10 @@ type Runner struct {
 	outputConfig              agentconfig.Namespace
 	elasticsearchOutputConfig *agentconfig.C
 
-	listener net.Listener
+	metricReader   *sdkmetric.ManualReader
```
axw, line 94: This doesn't seem to be used, can we remove it and RunnerParams.MetricReader?

internal/beatcmd/reloader.go
```diff
@@ -86,6 +108,10 @@ type Reloader struct {
 	logger    *logp.Logger
 	newRunner NewRunnerFunc
 
+	meterProvider  metric.MeterProvider
+	metricReader   *sdkmetric.ManualReader
```
axw, line 112: metricReader doesn't seem to be needed in the reloader either?

internal/beatcmd/beat.go
```diff
@@ -457,6 +491,156 @@ func (b *Beat) registerMetrics() {
...
+			}
+		case "elasticsearch.indexers.created":
+			if value, ok := getScalarInt64(m.Data); ok {
+				monitoring.ReportInt(v, "indexers.created", value)
+			}
+		case "elasticsearch.indexers.destroyed":
+			if value, ok := getScalarInt64(m.Data); ok {
+				monitoring.ReportInt(v, "indexers.destroyed", value)
+			}
+			// TODO output.elasticsearch.indexers.active (created - destroyed?)
```
axw, line 639: Continuing from https://github.com/elastic/apm-server/pull/15094#discussion_r1927668560

> this is not produced by the go-docappender, it's one of the legacy metrics. If we think it's useful we should add it there. WDYT ?

It might be a good idea to add to go-docappender, but I'm not sure that we can wait for that. Can we calculate it here as created-destroyed, and add a TODO to either add it to go-docappender or remove the metric?

axw (commented):
Thanks for splitting this out, it's a lot more consumable!

Mostly looks good, just a few comments

internal/beater/server_test.go
axw: Ah, that would be expected here since this code is only looking at OTel metrics. We're missing a test in beatcmd for the mapping.

axw: Added a unit test in beatcmd

internal/beatcmd/reloader.go
axw: removed

internal/beatcmd/beat.go
axw: I've added it back (should be created-destroyed+1), and we can think about whether we want to remove it or add it to go-docappender later.

axw (dismissed):
LGTM, we'll need to make sure stack monitoring & self-instrumentation are manually tested I think.

internal/beatcmd/beat.go
```diff
@@ -457,6 +490,188 @@ func (b *Beat) registerMetrics() {
...
+	v.OnKey("events")
+	for _, m := range sm.Metrics {
+		switch m.Name {
+		case "elasticsearch.events.processed":
+			var acked, toomany, failed int64
+			data, _ := m.Data.(metricdata.Sum[int64])
+			for _, dp := range data.DataPoints {
+				status, ok := dp.Attributes.Value(attribute.Key("status"))
+				if !ok {
+					continue
```
simitt, line 578: Should we log a warning here? Afaics this wouldn't be expected.

simitt (approved):
Code looks good to me; only left one minor comment. 
Did review the metrics mapping in detail and nothing stood out. 

@kruskall before merging, please leave a note on the PR how you tested this manually (see @axw 's comment above).

carsonip:
for the record this introduced a regression where apmbench events/s is 0 (via expvar). @kruskall is working on a fix.

carsonip:
As explained in https://github.com/elastic/apm-server/issues/13738#issuecomment-2619397528 , I'm confident that the regression is caused by this PR. @kruskall can you study how this PR interacts with EA and instrumentation config on ESS? They are suspect.

Issue #15424, opened 2025-01-28
remove backward compatible flag support


kruskall:
**APM Server version** (`apm-server version`): 9.0

**Description of the problem including expected versus actual behavior**:

beats removed backward compatible flag support in https://github.com/elastic/beats/issues/42117 but we're adding it back at https://github.com/elastic/apm-server/blob/main/internal/beatcmd/init.go#L57-L61

We should remove the hack as well

**Steps to reproduce**:

Please include a *minimal* but *complete* recreation of the problem,
including server configuration, agent(s) used, etc.  The easier you make it
for us to reproduce it, the more likely that somebody will take the time to
look at it.

 1.
 2.
 3.

**Provide logs (if relevant)**:

PR #15425, opened 2025-01-28
feat: drop backward compatible flag

kruskall:
## Motivation/summary

beats removed backward compatible flag support in [elastic/beats#42117](https://github.com/elastic/beats/issues/42117) but we're adding it back at https://github.com/elastic/apm-server/blob/main/internal/beatcmd/init.go#L57-L61We should remove the hack as well

Closes https://github.com/elastic/apm-server/issues/15424

## Checklist


internal/beatcmd/init.go
```diff
@@ -52,11 +52,5 @@ func initRand() {
 }
 
 func initFlags() {
```
1pkg, line 54: nit: there is no more point to have dedicated `initFlags` function just move `cfgfile.Initialize()` directly to `init`

1pkg (approved):

PR #15434, opened 2025-01-28
[apmbench] Add trace only benchmarks

carsonip:
## Motivation/summary

Existing benchmarks are a mix of logs, metrics and traces (sampled and unsampled transactions). Using these benchmarks will introduce noise in benchmark results to assess TBS performance. It will create strange results where the slower the disk, the higher the throughput, because traces ingestion slows and ingestion of other events speeds up to take up the CPU time spared by slow disk, making benchmark results very hard to interpret.

Create a special benchmark with only sampled traces to accurately benchmark traces.

## Checklist


endorama (approved):

carsonip:
@Mergifyio backport 9.0 8.x 8.18

PR #15439, opened 2025-01-28
fix: always register beats registries

kruskall:
## Motivation/summary

beats monitoring registries are not being registered in ESS because the output name check is not passing.

also never remove the libbeat registry since it's only registered once on startup.

this is fine for non libbeat output because:
- old output registry metrics are cleared
- the mapping code only maps docappender metrics which are empty if non-es output is used

## Checklist

## Related issues

Related to https://github.com/elastic/apm-server/issues/13738

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -35,9 +35,14 @@ on:
         description: 'Set the expression that matches the benchmark scenarios to run'
         required: false
         type: string
+      pgoExport:
+        description: 'Run the benchmarks with PGO export'
+        required: false
+        type: boolean
+        default: false
   schedule:
     - cron: '0 17 * * *' # Scheduled regular benchmarks.
-    - cron: '0 5 */5 * *' # Scheduled PGO benchmarks.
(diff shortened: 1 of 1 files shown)


internal/beater/beater.go
```diff
@@ -685,8 +685,7 @@ func (s *Runner) newFinalBatchProcessor(
 	memLimit float64,
 ) (modelpb.BatchProcessor, func(context.Context) error, error) {
 	if s.elasticsearchOutputConfig == nil {
-		monitoring.Default.Remove("libbeat")
-		libbeatMonitoringRegistry := monitoring.Default.NewRegistry("libbeat")
+		libbeatMonitoringRegistry := monitoring.Default.GetRegistry("libbeat")
```
carsonip, line 688: Q: Is this related to the fix? I was under the impression that we specifically remove and create a new libbeat repo to workaround EA hot reload where 2 final processors run in parallel

1pkg (dismissed):

1pkg (approved):

PR #15467, opened 2025-01-30
[WIP] TBS: disk threshold

carsonip:
## Motivation/summary

Enforce disk usage threshold on top of existing db storage limit. Remove default storage limit.

## Checklist

### DIFF
--- a/x-pack/apm-server/sampling/eventstorage/storage_manager.go
+++ b/x-pack/apm-server/sampling/eventstorage/storage_manager.go
@@ -347,10 +347,13 @@ func (sm *StorageManager) WriteSubscriberPosition(data []byte) error {
 }
 
 func (sm *StorageManager) NewReadWriter() StorageLimitReadWriter {
-	return NewStorageLimitReadWriter(sm, SplitReadWriter{
+	splitRW := SplitReadWriter{
 		eventRW:    sm.eventStorage.NewReadWriter(),
 		decisionRW: sm.decisionStorage.NewReadWriter(),
-	})
+	}
+	storageLimitRW := NewStorageLimitReadWriter(sm, splitRW)
+	diskThresholdRW := NewStorageLimitReadWriter(NewDiskThresholdChecker(sm.storageDir, 0.9), storageLimitRW)
+	return diskThresholdRW
 }
 
 // wrapNonNilErr only wraps an error with format if the error is not nil.
--- a/x-pack/apm-server/sampling/eventstorage/disk.go
+++ b/x-pack/apm-server/sampling/eventstorage/disk.go
@@ -1,0 +1,35 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License 2.0;
+// you may not use this file except in compliance with the Elastic License 2.0.
+
+package eventstorage
+
+import "github.com/cockroachdb/pebble/v2/vfs"
+
+type DiskThresholdChecker struct {
+	path      string
+	threshold float64
+}
+
+func NewDiskThresholdChecker(path string, threshold float64) DiskThresholdChecker {
+	return DiskThresholdChecker{
+		path:      path,
+		threshold: threshold,
+	}
+}
+
+func (d DiskThresholdChecker) DiskUsage() uint64 {
+	usage, err := vfs.Default.GetDiskUsage(d.path) // FIXME: cache results
+	if err != nil {
+		panic(err) // FIXME: handle err
+	}
+	return usage.UsedBytes
+}
+
+func (d DiskThresholdChecker) StorageLimit() uint64 {
+	usage, err := vfs.Default.GetDiskUsage(d.path) // FIXME: cache results
+	if err != nil {
+		panic(err) // FIXME: handle err
+	}
+	return uint64(float64(usage.TotalBytes) * d.threshold)
+}


x-pack/apm-server/sampling/eventstorage/storage_manager.go
```diff
@@ -202,17 +219,42 @@ func (sm *StorageManager) Size() (lsm, vlog int64) {
...
+	return sm.storageLimit.Load()
 }
 
 func (sm *StorageManager) updateDiskUsage() {
-	sm.cachedDiskUsage.Store(sm.eventDB.Metrics().DiskSpaceUsage() + sm.decisionDB.Metrics().DiskSpaceUsage())
+	sm.cachedDBSize.Store(sm.eventDB.Metrics().DiskSpaceUsage() + sm.decisionDB.Metrics().DiskSpaceUsage())
+
+	usage, err := vfs.Default.GetDiskUsage(sm.storageDir)
+	if err != nil {
+		sm.rateLimitedLogger.With(logp.Error(err)).Warn("failed to get disk usage")
```
simitt, line 241: what is the fallback in this case?

carsonip: In the current implementation it effectively becomes unlimited. It is by design because of how they are 2 separate checkers and RW. But I _think_ it is possible to overwrite `dbStorageLimit` to a fallback value if we get an error, so that storage_limit checker takes over.

x-pack/apm-server/sampling/eventstorage/storage_manager.go
```diff
@@ -279,7 +329,10 @@ func (sm *StorageManager) Run(stopping <-chan struct{}, ttl time.Duration, stora
 		<-sm.runCh
 	}()
 
-	sm.storageLimit.Store(storageLimit)
+	if storageLimit != 0 {
```
1pkg, line 332: Is it a good idea to handle the `storageLimit` config overriding logic here in `Run` dynamically? Is there any reason we have to do it this way vs handle it once in `NewStorageManager` initialization?

carsonip: sigh, this is exactly the problem. StorageManager has the lifecycle of the global main function where config is not yet available, but configs (including storage limit) has the lifecycle of a processor. i.e. in a hot reload, the storage limit may change.

Overriding the storage limit in Run dynamically still opens to a race when events come in between processor is created and processor.Run is called. Doing `sm.storageLimit.Store(storageLimit)` in the #15235 was cutting corners and it comes back to bite us.

I'm also playing with a new PR where storage limit is not bound to StorageManager, but a ReadWriter, which will be the most correct implementation, but the change is going to be larger and will be slightly harder to review. I'll let you know when it is ready. This will be similar to what we have in badger where we had a wrappedRW per processor (see [code](https://github.com/elastic/apm-server/blob/8.17/x-pack/apm-server/sampling/processor.go#L589)) which knows the storage limit. It was done for a good reason.

carsonip: Not to mention that if pebble is backported to 8.x, we need to apply a similar fix in 8.x without porting the new "disk_threshold" handling, as this is a regression.

carsonip: Created PR https://github.com/elastic/apm-server/pull/15488

x-pack/apm-server/sampling/eventstorage/storage_manager.go
carsonip: Added fallback handling. A bit complex, but should cover all cases. All of it is written with the assumption that if GetDiskUsage ever returns an error, it will always return an error.

internal/beater/config/sampling.go
```diff
@@ -49,8 +49,23 @@ type TailSamplingConfig struct {
...
+	// DiskThresholdRatio controls the proportion of the disk to be filled at max, irrespective of db size.
+	// e.g. 0.9 means the last 10% of disk should not be written to.
+	//
+	// This is additional to StorageLimit which controls db size,
+	// i.e. if both StorageLimit and DiskThresholdRatio are set, both are enforced.
+	// If StorageLimit is 0 and any error occurs when getting filesystem stats,
+	// a fallback StorageLimit will be applied.
+	// To disable the fallback to allow unlimited db size and disk threshold under all circumstances,
+	// set both StorageLimit and DiskThresholdRatio to 0.
+	DiskThresholdRatio float64 `config:"disk_threshold_ratio"  validate:"min=0, max=1"`
```
carsonip, line 67: [to reviewer] open to a better name, maybe `disk_usage_ratio_threshold`?

internal/beater/config/sampling.go
```diff
@@ -49,8 +49,23 @@ type TailSamplingConfig struct {
...
 	TTL                   time.Duration         `config:"ttl" validate:"min=1s"`
-	StorageLimit          string                `config:"storage_limit"`
-	StorageLimitParsed    uint64
+
+	// StorageLimit is the user-configured tail-sampling database size limit.
+	// 0 means unlimited storage.
+	StorageLimit       string `config:"storage_limit"`
+	StorageLimitParsed uint64
+
+	// DiskThresholdRatio controls the proportion of the disk to be filled at max, irrespective of db size.
```
carsonip, line 58: [to reviewer] Q: @lahsivjar suggested that it doesn't need to be configurable. I've made it configurable but undocumented atm. As StorageLimit default is changed, we'll have to update observability docs anyway. I'm thinking we still have to either document this config, or describe this new default behavior as if it is not configurable. I'm slightly inclined to document it, but it also means we'll have to support it.

x-pack/apm-server/sampling/eventstorage/storage_manager.go
```diff
@@ -62,6 +67,19 @@ func WithMeterProvider(mp metric.MeterProvider) StorageManagerOptions {
 	}
 }
 
+// WithGetDiskUsage configures getDiskUsage function used by StorageManager.
+// For testing only.
+func WithGetDiskUsage(getDiskUsage func() (DiskUsage, error)) StorageManagerOptions {
```
1pkg, line 72: what is the purpose to expose this option publicly? do we expect any different implementation at all? AFAICT it's only useful for test purposes

carsonip: storage_manager_test is in `eventstorage_test` package, that is why the option has to be public. It is currently only useful for test purposes, but it _could_ serve a legit purpose to replace the GetDiskUsage implementation. But I don't have a strong opinion here. Either change the package of the test file, or keep this option public.

x-pack/apm-server/sampling/eventstorage/storage_manager.go
```diff
@@ -211,6 +243,28 @@ func (sm *StorageManager) dbSize() uint64 {
...
+
+	if sm.diskStatFailed.Load() {
+		// Skip GetDiskUsage under the assumption that
+		// it will always get the same error if GetDiskUsage ever returns one,
+		// such that it does not keep logging GetDiskUsage errors.
+		return
+	}
+	usage, err := sm.getDiskUsage()
+	if err != nil {
+		sm.logger.With(logp.Error(err)).Warn("failed to get disk usage")
```
1pkg, line 255: should we update warn message with a note that unlimited threshold be used.

carsonip: This line is for logging the error. Using unlimited threshold is implied in https://github.com/elastic/apm-server/pull/15467/files#diff-3ab172f1b1c0d77526ba1b808468554518bff5b2cffb34b539bf3d5384f430ecR436 . Welcome any suggestions for clearer log message!

1pkg (dismissed):

x-pack/apm-server/sampling/eventstorage/storage_manager.go
1pkg: It's fine let's keep it public, there is no serious harm in that.

carsonip:
Tested with #15492 

startup message
```json
{"log.level":"info","@timestamp":"2025-01-30T23:26:50.535Z","log.logger":"sampling","log.origin":{"function":"github.com/elastic/apm-server/x-pack/apm-server/sampling/eventstorage.(*StorageManager).NewReadWriter","file.name":"eventstorage/storage_manager.go","file.line":446},"message":"setting disk threshold ratio to 0.01 of total disk space of 2.0gb","service.name":"apm-server","ecs.version":"1.6.0"}
```

error message
```json
{"log.level":"warn","@timestamp":"2025-01-30T23:38:51.103Z","log.logger":"sampling","log.origin":{"function":"github.com/elastic/apm-server/x-pack/apm-server/sampling.(*Processor).ProcessBatch","file.name":"sampling/processor.go","file.line":127},"message":"processing trace failed, indexing by default","service.name":"apm-server","error":{"message":"disk usage ratio exceeding threshold of 0.01: configured storage limit reached (current: 38973440, limit 21474836)"},"ecs.version":"1.6.0"}
```

internal/beater/config/sampling.go
lahsivjar: Having this as a config option is okay for now. If we want to expose this (either as a config or documentation) we should discuss a bit on how we want both the options to interact and the user experience it entails. To start with, I would recommend documenting this (using up to 90% of the available disk) as default TBS behaviour with a mention on how to tweak the max using `storage_limit` option.

internal/beater/config/sampling.go
lahsivjar: Maybe `disk_usage_threshold`, not sure how much the `ratio` is adding... no objection with what is in the PR too.

x-pack/apm-server/sampling/eventstorage/storage_manager.go
```diff
@@ -46,6 +47,12 @@ const (
 
 	// diskUsageFetchInterval is how often disk usage is fetched which is equivalent to how long disk usage is cached.
 	diskUsageFetchInterval = 1 * time.Second
+
+	// dbStorageLimitFallback is the default fallback storage limit in bytes
+	// that applies when disk threshold cannot be enforced due to an error.
+	dbStorageLimitFallback = 5 << 30
```
lahsivjar, line 53: How did we come up with 5GB? Along the same lines, what was the reasoning for choosing 3GB default previously? Should we keep the fallback to 3GB? - If the previous value was also chosen randomly then no objection with 5GB.

x-pack/apm-server/sampling/eventstorage/storage_manager.go
```diff
@@ -341,30 +397,64 @@ func (sm *StorageManager) WriteSubscriberPosition(data []byte) error {
 // NewUnlimitedReadWriter returns a read writer with no storage limit.
 // For testing only.
```
lahsivjar, line 398: nit: Not a fan of these testing-only methods, can we move at least this one to the test file?

x-pack/apm-server/sampling/eventstorage/storage_manager.go
```diff
@@ -341,30 +397,64 @@ func (sm *StorageManager) WriteSubscriberPosition(data []byte) error {
...
+			return uint64(float64(sm.diskStat.total.Load()) * diskThresholdRatio)
+		}
+		// the total disk space could change in runtime, but it is still useful to print it out in logs.
+		sm.logger.Infof("setting disk threshold ratio to %.2f of total disk space of %0.1fgb", diskThresholdRatio, float64(sm.diskStat.total.Load())/gb)
+	}
+
+	// To limit actual disk usage percentage to diskThresholdRatio
+	diskThresholdChecker := NewStorageLimitCheckerFunc(sm.diskUsed, diskThreshold)
+	diskThresholdRW := NewStorageLimitReadWriter(
+		fmt.Sprintf("disk usage ratio exceeding threshold of %.2f", diskThresholdRatio),
```
lahsivjar, line 452: How is this used? The docs says that this is a `name` parameter

x-pack/apm-server/sampling/eventstorage/storage_manager.go
```diff
@@ -341,30 +397,64 @@ func (sm *StorageManager) WriteSubscriberPosition(data []byte) error {
...
 
-	dbStorageLimit := func() uint64 {
-		return storageLimit
-	}
-	if storageLimit == 0 {
-		sm.logger.Infof("setting database storage limit to unlimited")
+	// dbStorageLimit returns max size of db in bytes.
+	// If size of db exceeds dbStorageLimit, writes should be rejected.
+	var dbStorageLimit func() uint64
+	if sm.diskStatFailed.Load() && storageLimit == 0 && diskThresholdRatio > 0 {
```
lahsivjar, line 413: This section feels a bit confusing, what if the disk usage hasn't ran yet and fails after `NewReadWriter` is called?

How about we add the disk usage fallback in the `getDiskUsage` itself and remove the error return? - this way the method has to add fallback by definitione

x-pack/apm-server/sampling/eventstorage/storage_manager.go
```diff
@@ -341,30 +397,64 @@ func (sm *StorageManager) WriteSubscriberPosition(data []byte) error {
...
+		sm.logger.Infof("setting disk threshold ratio to %.2f of total disk space of %0.1fgb", diskThresholdRatio, float64(sm.diskStat.total.Load())/gb)
+	}
+
+	// To limit actual disk usage percentage to diskThresholdRatio
+	diskThresholdChecker := NewStorageLimitCheckerFunc(sm.diskUsed, diskThreshold)
+	diskThresholdRW := NewStorageLimitReadWriter(
+		fmt.Sprintf("disk usage ratio exceeding threshold of %.2f", diskThresholdRatio),
+		diskThresholdChecker,
+		dbStorageLimitRW,
+	)
```
simitt, line 455: This gets fairly complex now. I think it is fine to move forward given the time constraints, but would appreciate a bit of refactoring soon after. The responsibilities of ensuring disk space is available is now spread amongst several parts.

x-pack/apm-server/sampling/eventstorage/storage_manager.go
simitt: I wonder if we should keep falling back to `3GB`? This would yield in the same result for upgraded instances where the newly introduced storage limit cannot be fetched.

internal/beater/config/sampling.go
simitt: I like the suggestion of `disk_usage_threshold`.

internal/beater/config/sampling.go
simitt: >Having this as a config option is okay for now.

Once we expose it, we cannot simply revert this decision. 

With the current implementation, one needs to be aware of both config options and understand how they play together. While also checking for not exceeding a certain threshold is a guardrail, it also adds complexity for the user.  I am not a huge fan of this complexity. 
If the `DiskThresholdRatio` would only be checked if no `StorageLimit` is set, I think it would be easier to grasp. Either the APM Server always tries to write up to X GB or it writes up to x% of the overall disk size, but only if available.

x-pack/apm-server/sampling/eventstorage/storage_manager.go
carsonip: No problem with keeping it at 3GB for fallback

x-pack/apm-server/sampling/eventstorage/storage_manager.go
carsonip: The test file is in another package though, moving this means making the struct field public which is even worse imo. The one with least impact to prod code is to make the test file use the same package, which is probably the best option

x-pack/apm-server/sampling/eventstorage/storage_manager.go
carsonip: getDiskUsage is called on StorageManager initialization, the first time a TBS processor is created, and before any processor is Run. I can add a comment to make it clear.

x-pack/apm-server/sampling/eventstorage/storage_manager.go
lahsivjar: Hmm, isn't `NewReadWriter` a public method? I was thinking of moving the `newUnlimitedReadWriter` to the test package.

x-pack/apm-server/sampling/eventstorage/storage_manager.go
carsonip: It first started as a name of the StorageLimitReadWriter, but now I'm using it more like a description. It is prepended to the error. While changing it to "disk usage threshold" makes it become a name again, having the value of it seems to help understand, because `current: X, limit: Y` in the error message are absolute values and is not very easy to understand: `disk usage ratio exceeding threshold of 0.01: configured storage limit reached (current: 38973440, limit 21474836)`

x-pack/apm-server/sampling/eventstorage/storage_manager.go
```diff
@@ -83,6 +103,13 @@ type StorageManager struct {
 	// cachedDBSize is a cached result of db size.
 	cachedDBSize atomic.Uint64
 
+	getDiskUsage func() (DiskUsage, error)
```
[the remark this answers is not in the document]
carsonip: should have called it cachedDiskStat and changing diskStatFailed to getDiskUsageFailed. Keeping the call as getDiskUsage as the pebble function is called GetDiskUsage.

internal/beater/config/sampling.go
lahsivjar: > Once we expose it, we cannot simply revert this decision.

Hmm, I was thinking if it is exposed as undocumented configuration (i.e. only documenting the behaviour) we could revert the decision but if this is not true then I would be reluctant to release this config.

> If the DiskThresholdRatio would only be checked if no StorageLimit is set, I think it would be easier to grasp. Either the APM Server always tries to write up to X GB or it writes up to x% of the overall disk size, but only if available.

This sounds good to me, however, I would advocate for starting with the x% of overall disk size as a fixed default behaviour for now because I am not sure if this config is the best way forward. So, the behavior would be, if `storage_limit` is set then that is obeyed, otherwise, 90% of the overall disk size would be used for TBS. Does this sound reasonable?

x-pack/apm-server/sampling/eventstorage/storage_manager.go
carsonip: oh, sorry i thought you were talking about `WithGetDiskUsage`.

Yes, UnlimitedReadWriter can be moved, but quite a few test files use it. I can sort it out.

internal/beater/config/sampling.go
carsonip: I am fine with keeping it an undocumented config, and disable disk usage threshold check whenever StorageLimit is set. The edge case we'll have to handle is when user sets it to 0.

- What if the user really wants unlimited storage limit, without disk usage threshold check? 
- For the above case, what if GetDiskUsage returns an error?

i.e. are we happy with
- no more truly unlimited storage limit, either set storage_limit to non-0 and TBS will use that absolute value, or set it to 0 and disk usage threshold will be limited to 90%

Alternatives:
- make StorageLimit not an uint64 but an int64, where <0 means truly unlimited and disables disk usage threshold?

I'm happy with no more truly unlimited storage limit but just checking if we want an escape hatch for it. I think it is ok to tell users who want a truly unlimited to set the StorageLimit as maxUInt64.

x-pack/apm-server/sampling/eventstorage/storage_manager.go
lahsivjar: Hmm, tbh, it feels a bit of a smell. Maybe we can refactor it later to abstract the error outside `StorageLimitReadWriter` and remove the `name`.

x-pack/apm-server/sampling/eventstorage/storage_manager.go
carsonip: ++ on refactoring in a follow up PR when my head is clearer.

getDiskUsage is designed to mock pebble GetDiskUsage without modifying internal state. `cachedDiskStat`, `getDiskUsageFailed`, `cachedDBSize` are all refreshed regularly throughout the lifetime of StorageManager and should be seen as a (cached) source of truth. Therefore, the use of getDiskUsageFailed here is almost like calling getDiskUsage within NewReadWriter and handling its error.

x-pack/apm-server/sampling/eventstorage/storage_manager.go
lahsivjar: No need to do this now - we can do this in a future PR.

internal/beater/config/sampling.go
carsonip: Updated PR to implement what @simitt suggested. Any non-0 storage limit disables disk usage threshold checks. On non-0 storage limit, disk usage threshold checks using the undocumented default `disk_usage_threshold=0.9`, unless getDiskUsage errored, which would then fallback to 3GB db storage limit.

x-pack/apm-server/sampling/eventstorage/storage_manager.go
carsonip: updated it slightly to make the passed in argument more like a name. Abstracting the error outside StorageLimitReadWriter to create a DiskThresholdReadWriter is fine, but they would look too similar apart from the error / name.

x-pack/apm-server/sampling/eventstorage/storage_manager.go
```diff
@@ -210,7 +259,29 @@ func (sm *StorageManager) dbSize() uint64 {
...
+		// Skip GetDiskUsage under the assumption that
+		// it will always get the same error if GetDiskUsage ever returns one,
+		// such that it does not keep logging GetDiskUsage errors.
+		return
+	}
+	usage, err := sm.getDiskUsage()
+	if err != nil {
+		sm.logger.With(logp.Error(err)).Warn("failed to get disk usage")
+		sm.getDiskUsageFailed.Store(true)
+		sm.cachedDiskStat.total.Store(0) // setting total to 0 to disable any running disk usage threshold checks
```
simitt, line 274: Should `m.cachedDiskStat.used.Store` also be updated here? It might hold a value from previous calls.

simitt (approved):

x-pack/apm-server/sampling/eventstorage/storage_manager.go
```diff
@@ -210,7 +259,29 @@ func (sm *StorageManager) dbSize() uint64 {
...
+	if sm.getDiskUsageFailed.Load() {
+		// Skip GetDiskUsage under the assumption that
+		// it will always get the same error if GetDiskUsage ever returns one,
+		// such that it does not keep logging GetDiskUsage errors.
+		return
+	}
+	usage, err := sm.getDiskUsage()
+	if err != nil {
+		sm.logger.With(logp.Error(err)).Warn("failed to get disk usage")
+		sm.getDiskUsageFailed.Store(true)
```
lahsivjar, line 273: nit: While I can't think of what could lead to transient failures, I am not sure about _always_ failing on error bit - something to think about in a future PR.

lahsivjar (dismissed):
LGTM!

x-pack/apm-server/sampling/eventstorage/storage_manager.go
carsonip: This gives me a headache as well. What should existing disk usage threshold checks perform when getDiskUsage has transient failures, should it use a stale number, or should it become unlimited? With this assumption of _always_ failing, it simplifies the implementation.

lahsivjar (approved):

x-pack/apm-server/sampling/eventstorage/storage_manager.go
carsonip: fixed

simitt (approved):

PR #15517, opened 2025-01-31
fix: do not overwrite registries when publishing otel metrics to beats

kruskall:
## Motivation/summary

update apm-server registry mapping logic to avoid overwriting onkey handlers

discovered while backporting the otel changes. Already added in 8.x

if metrics have similar paths the onkey handler is overwritten. To avoid this process all the metrics path first to build a tree and produce them in order.

## Checklist


simitt (commented):
Can you please add a unit test for this logic.

1pkg (approved):

simitt (approved):
Thank you!

PR #15545, opened 2025-02-04
feat(fips): remove keystore subcommand and config handling

kruskall:
## Motivation/summary

the keystore is providing obfuscation of data on disk by using an empty password by default.
This fails in fips only mode with the following error:

`crypto/hmac: use of keys shorter than 112 bits is not allowed in FIPS 140-only mode`

Disable the keystore subcommand and do not try to create one when loading the config.

## Checklist

## How to test these changes

- download gotip/go1.24
- build apm-server with the stdlib in fips only mode
- run apm-server

## Related issues

closes https://github.com/elastic/apm-server/issues/15839


internal/beatcmd/config_nofips_test.go
```diff
@@ -0,0 +1,62 @@
...
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/elastic/elastic-agent-libs/keystore"
+)
+
+func TestLoadConfigKeystore(t *testing.T) {
```
1pkg, line 29: Will this test fail if build with `requirefips`?

internal/beatcmd/config_nofips.go
```diff
@@ -0,0 +1,34 @@
...
+
+import (
+	"github.com/elastic/beats/v7/libbeat/common"
+	"github.com/elastic/elastic-agent-libs/config"
+	libkeystore "github.com/elastic/elastic-agent-libs/keystore"
+	"github.com/elastic/elastic-agent-libs/paths"
+)
+
+// loadKeystore returns the appropriate keystore based on the configuration.
+func loadKeystore(cfg *config.C) (libkeystore.Keystore, error) {
```
1pkg, line 30: nit: why not to add this function directly to `internal/beatcmd/keystore_nofips.go` file?

internal/beatcmd/config_fips.go
```diff
@@ -0,0 +1,30 @@
...
+
+package beatcmd
+
+import (
+	"github.com/elastic/elastic-agent-libs/config"
+	libkeystore "github.com/elastic/elastic-agent-libs/keystore"
+)
+
+// loadKeystore returns the appropriate keystore based on the configuration.
+func loadKeystore(cfg *config.C) (libkeystore.Keystore, error) {
```
1pkg, line 28: nit: why not to add this function directly to `internal/beatcmd/keystore_fips.go` file?

1pkg (dismissed):

1pkg (approved):

PR #15550, opened 2025-02-04
mergify: support backport-active-8, backport-active-9 and backport-active-all

v1v:
## Motivation/summary

> Maybe this is solved already, but is there a way we can get an automatically kept up to date “backport-active-minors” label that auto backports to the set of minor releases that are still supported? I’d want this in Beats, Elastic Agent, and Fleet Server if it already exists. I’ve noticed an increase in forgotten backports because there are more releases to keep track of. If we can abstract this it’d make life easier, I think Kibana does something like this

## How does this PR solve the problem?

Enable `backport-active-all`, `backport-active-8` and `backport-active-9` GitHub labels in Mergify to backport to a set of active branches in addition to the existing automation with the `backport-<major>.<minor>` labels. This can help with the current 6 active branches, which we have today.

See https://docs.mergify.com/workflow/actions/backport/#parameters

image

I can find `regexes` too:

image

But I'm not sure whether that can solve our current release model easily

## Checklist


.mergify.yml
```diff
@@ -434,3 +436,44 @@ pull_request_rules:
...
+      backport:
+        assignees:
+          - "{{ author }}"
+        # NOTE: this list needs to be changed when a new minor branch is created
+        #       or an existing minor branch reached EOL.
+        branches:
+          - "9.0"
+          - "8.18"
+          - "8.17"
+          - "8.16"
```
simitt, line 455: I find the naming a big confusing, e.g. `9.0` and `8.x` branches are treated within `backport-active-minors` but they are indeed 2 different majors.

Could we do something like
- `backport-active-8` : `8.16, 8.17, 8.18` 
- `backport-active-9`: `9.0` 
- `backport-active-all`: `9.0,  8.16, 8.17, 8.18, 7.17`

v1v: That's a good question - I also had some thoughts about it when I raised this.

Let me ping you in the internal ticket, so we can agree with the naming all. and then I'll update this thread to circle back

v1v: Agreed to move forward with your proposal

.mergify.yml
```diff
@@ -176,6 +176,10 @@ pull_request_rules:
           * `backport-8./d` is the label to automatically backport to the `8./d` branch. `/d` is the digit.
           * `backport-9./d` is the label to automatically backport to the `9./d` branch. `/d` is the digit.
           * `backport-8.x` is the label to automatically backport to the `8.x` branch.
+          * `backport-active-all` is the label that automatically backports to all active branches.
+          * `backport-active-8` is the label that automatically backports to all active minor branches for the 8 major.
+          * `backport-active-9` is the label that automatically backports to all active minor branches for the 9 major.
+          * `backport-active-9` is the label that automatically backports to all active minor branches for the 9 major.
```
v1v, line 182:
```suggestion
          * `backport-active-all` is the label that automatically backports to all active branches.
          * `backport-active-8` is the label that automatically backports to all active minor branches for the 8 major.
          * `backport-active-9` is the label that automatically backports to all active minor branches for the 9 major.
```

endorama:
@v1v so this new labels would replace the usual `backport-X.Y`? i.e. backporting to 8.x line would be achieved with `backport-active-8` only?

In principle I agree, but I see this to create further confusion: when we act as release managers we look at previous PR to determine which release they targeted. With the current labels is fairly easy to see where we backported to.
With the new labels (where I expect there will be changing in the "active" definition going forward) we lose this information (unless I misunderstood and this PR would add the usual label, not act in place of them). As we have so many active branches this is already very time consuming and I fear this change will make it ever more so.

v1v:
> @v1v so this new labels would replace the usual backport-X.Y? i.e. backporting to 8.x line would be achieved with backport-active-8 only?

No, the same mechanism remains. This is adding new backport labels to help with targeting multiple branches

v1v:
> In principle I agree, but I see this to create further confusion: when we act as release managers we look at previous PR to determine which release they targeted. With the current labels is fairly easy to see where we backported to.
With the new labels (where I expect there will be changing in the "active" definition going forward) we lose this information (unless I misunderstood and this PR would add the usual label, not act in place of them). As we have so many active branches this is already very time consuming and I fear this change will make it ever more so.

I understand, we have some internal discussion about helping with generating a comprehensive list of PRs that were or not backported from main to the other branches.

Let me ping you to that thread. 

Feel free to dispose of this if you think it's not something beneficial.

endorama:
@v1v didn't want to be dismissive, this PR would be really helpful on one side, I'm just worried about adding further cognitive complexity to the release manager duty.

@inge4pres what do you think?

.mergify.yml
```diff
@@ -434,3 +436,59 @@ pull_request_rules:
...
+      - base=main
+      - label=backport-active-8
+    actions:
+      backport:
+        assignees:
+          - "{{ author }}"
+        # NOTE: this list needs to be changed when a new minor branch is created
+        #       or an existing minor branch reached EOL.
+        branches:
+          - "8.18"
```
v1v, line 452: Shall we consider `8.x` too?
```suggestion
          - "8.x"
          - "8.18"
```

1pkg (commented):
LGTM, but if the active branch list needs to be updated manually on every new release then we also need to update https://github.com/elastic/apm-server/blob/main/dev_docs/RELEASES.md to mention this

v1v:
> LGTM, but if the active branch list needs to be updated manually on every new release then we also need to update [main/dev_docs/RELEASES.md](https://github.com/elastic/apm-server/blob/main/dev_docs/RELEASES.md) to mention this

That's gonna be automated, there is an open GH issue in our internal repository to figure out how to do it.

1pkg (dismissed):

v1v:
> @[v1v](https://github.com/v1v) v1v dismissed stale reviews from [1pkg](https://github.com/1pkg) and [kruskall](https://github.com/kruskall) via [9b361e5](https://github.com/elastic/apm-server/commit/9b361e50a6adee65fe5a565a44a9b1c45c468f3d) [now](https://github.com/elastic/apm-server/pull/15550#event-16215542399)

Sorry folks, I enabled the support for `8.x` when using `backport-active-8`

1pkg (approved):

PR #15560, opened 2025-02-05
feat(fips): replace md5 with xxhash in agentcfg etag generation

kruskall:
## Motivation/summary

replace md5 with xxhash in agentcfg etag generation

## Checklist

## Related issues
closes https://github.com/elastic/apm-server/issues/15841

### DIFF
--- a/internal/beater/config/agentconfig.go
+++ b/internal/beater/config/agentconfig.go
@@ -18,12 +18,12 @@
 package config
 
 import (
-	"crypto/md5"
 	"encoding/json"
 	"fmt"
 	"strings"
 	"time"
 
+	"github.com/cespare/xxhash/v2"
 	"github.com/pkg/errors"
 
 	"github.com/elastic/apm-server/internal/elasticsearch"
@@ -125,7 +125,7 @@ func (s *FleetAgentConfig) setup() error {
 		if err != nil {
 			return fmt.Errorf("error generating etag for %s: %v", s.Service, err)
 		}
-		s.Etag = fmt.Sprintf("%x", md5.Sum(m))
+		s.Etag = fmt.Sprintf("%x", xxhash.New().Sum(m))
 	}
 	return nil
 }


internal/beater/config/agentconfig.go
```diff
@@ -121,11 +118,11 @@ func (s *FleetAgentConfig) setup() error {
...
 	}
 	if s.Etag == "" {
-		m, err := json.Marshal(s)
+		etag, err := generateEtag(s)
 		if err != nil {
-			return fmt.Errorf("error generating etag for %s: %v", s.Service, err)
+			return err
 		}
-		s.Etag = fmt.Sprintf("%x", md5.Sum(m))
+		s.Etag = etag
```
endorama, line 125: What do you think of not set Etag at all in FIPS scenario?

As the function already takes a `*FleetAgentConfig`, we could set `s.Etag` only in the non-FIPS implementation.

Something like
```
if s.Etag == "" {
err := generateEtag(s)
}
```

and then setting `s.Etag = etag` in the implementation.

1pkg (approved):

carsonip (changes requested):
I'm looking at this PR with fresh eyes.

- As shown in this PR, md5 is only used in agent config via fleet. That doesn't even concern ES. We can definitely switch to a compliant hashing algorithm for agent config via fleet.
- The ETag stored in ES is out of scope of apm-server. The code might be in kibana. Not sure if we need to do FIPS over there as well, and whether we need to perform any migrations for stored agent configs in ES.

internal/beater/config/agentconfig_fips.go
```diff
@@ -0,0 +1,24 @@
...
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+//go:build requirefips
+
+package config
+
+func generateEtag(s *FleetAgentConfig) (string, error) {
+	return "", nil
```
simitt, line 23: Looking at this functionality - why not generating an `etag` with a FIPS approved algorithm rather than returning an empty string? As @carsonip pointed out recently, the `etag` is used for caching on the apm agent side.

simitt (approved):

carsonip (approved):

isaacaflores2:
I validated the etag generation for the local `agent_config` for the 8.19 test plan


1. Create FIPS binary: `make apm-server-fips`
2. Create `local-apm-docker-agent.yml` which contains
```yaml
apm-server:
  host: "0.0.0.0:8200"
  agent_config:
    - service:
      name: "local-service"
      environment: "local"
      config:
        transaction_sample_rate: 0.1
        transaction_max_spans: 30
output.elasticsearch:
  hosts: ["http://host.docker.internal:9200"]
  enabled: true
  username: "elastic"
  password: "<redacted>"

```
3. Start the apm server
```bash
./apm-server-fips -c local-apm-docker-agent.yml -e -d "*"
```
4. Request the config
```bash
curl -i  "http://localhost:8200/config/v1/agents?service.name=local-service"
```
The response will look like
```bash
HTTP/1.1 200 OK
Access-Control-Expose-Headers: Etag
Cache-Control: max-age=30, must-revalidate
Content-Type: application/json
Etag: "7b2253657276696365223a7b224e616d65223a22222c22456e7669726f6e6d656e74223a22227d2c224167656e744e616d65223a22222c2245746167223a22222c22436f6e666967223a7b227472616e73616374696f6e5f6d61785f7370616e73223a223330222c227472616e73616374696f6e5f73616d706c655f72617465223a22302e31227d7def46db3751d8e999"
X-Content-Type-Options: nosniff
Date: Fri, 18 Jul 2025 00:29:34 GMT
Content-Length: 72

{
  "transaction_max_spans": "30",
  "transaction_sample_rate": "0.1"
}
```
6. Request the config with the etag to ensure we get the not modified response
Verify this returns a `304 Not Modified`
```bash
 curl -i  "http://localhost:8200/config/v1/agents?service.name=local-service" -H "If-None-Match: 7b2253657276696365223a7b224e616d65223a22222c22456e7669726f6e6d656e74223a22227d2c224167656e744e616d65223a22222c2245746167223a22222c22436f6e666967223a7b227472616e73616374696f6e5f6d61785f7370616e73223a223330222c227472616e73616374696f6e5f73616d706c655f72617465223a22302e31227d7def46db3751d8e999"
```
7. Modify the agent config and restart the Server
Request the config again with the same 
Verify this returns a `200 OK` with the new config
```bash
 curl -i  "http://localhost:8200/config/v1/agents?service.name=local-service" -H "If-None-Match: 7b2253657276696365223a7b224e616d65223a22222c22456e7669726f6e6d656e74223a22227d2c224167656e744e616d65223a22222c2245746167223a22222c22436f6e666967223a7b227472616e73616374696f6e5f6d61785f7370616e73223a223330222c227472616e73616374696f6e5f73616d706c655f72617465223a22302e31227d7def46db3751d8e999"
```

PR #15580, opened 2025-02-06
fix: overwrite output es config if malformed es config is passed

kruskall:
## Motivation/summary

if a password is provided without a username we should still overwrite the config for consistency

## Checklist

## How to test these changes

go test ./...


1pkg (approved):

simitt (approved):
LGTM, we should also backport to `8.19`

Issue #15581, opened 2025-02-06
ci go mod cache is not being used


kruskall:
**APM Server version** (`apm-server version`): main

**Description of the problem including expected versus actual behavior**:

For some reason github action jobs are not reusing the code cache. This is leading to increased CI time and costs

**Steps to reproduce**:

Please include a *minimal* but *complete* recreation of the problem,
including server configuration, agent(s) used, etc.  The easier you make it
for us to reproduce it, the more likely that somebody will take the time to
look at it.

 1. open a PR or watch the commit CI in main
 2.
 3.

**Provide logs (if relevant)**:

PR #15583, opened 2025-02-06
fix: use a single source of truth for the go version

kruskall:
## Motivation/summary

go version is defined in two places: go.mod and .go-version they often get out of sync and it's not clear which is used where. CI is using the version defined in go.mod which might be outdated compared to .go-version.

bump go version in go.mod and ensure scripts are using it consistently

remove old jenkins scripts

`.go-version` is kept around because it's still used in one place: the version bump automation. I plan to raise an issue to just create a workflow that runs `make update-go-version` since we already handle everything in a make task.

## Checklist

## Related issues

Related to https://github.com/elastic/apm-server/issues/15581

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,6 +6,7 @@ on:
       - main
       - 7.1*
       - 8.*
+      - 9.*
     paths-ignore: # When updating the list of expressions also update ci-docs.yml
       - '**.md'
       - '**.asciidoc'
--- a/.github/workflows/microbenchmark.yml
+++ b/.github/workflows/microbenchmark.yml
@@ -7,8 +7,8 @@ on:
   push:
     branches:
       - main
-      - "8.[0-9]+"
-      - "8.x"
+      - "8.*"
+      - "9.*"
     paths-ignore:
       - '**.md'
       - '**.asciidoc'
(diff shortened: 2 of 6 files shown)


1pkg (approved):

script/update_go_version.sh
```diff
@@ -11,6 +11,3 @@ cd $SDIR/..
 MAJOR_MINOR_VERSION=$(grep '^go' go.mod | cut -d' ' -f2 | cut -d. -f1-2)
 
 find ./ -type f -name "go.mod" -execdir go get go@$MAJOR_MINOR_VERSION \; -execdir go get toolchain@none \;
-
-GO_VERSION=$(grep '^go' go.mod | cut -d' ' -f2)
-echo $GO_VERSION > .go-version
```
simitt, line 16: >.go-version is kept around because it's still used in one place: the version bump automation. I plan to raise an issue to just create a workflow that runs make update-go-version since we already handle everything in a make task.

Shouldn't aligning the `.go-version` with the version from `go.mod` still be kept until `.go-version` is fully removed?

v1v:
After this is merged, I'll work on https://github.com/elastic/apm-server/blob/main/.ci/updatecli/bump-golang.yml, that's the [updatecli](https://www.updatecli.io/) pipeline we use for bumping the golang version.

PR #15623, opened 2025-02-07
ci: switch to custom setup-go action in windows

kruskall:
## Motivation/summary

windows runner has significantly slower disks and restoring the cache is wasting a huge amount of time

## Checklist

## Related issues

See https://github.com/actions/setup-go/issues/495
Slightly related to https://github.com/elastic/apm-server/issues/15581


.github/workflows/ci.yml
```diff
@@ -45,7 +45,13 @@ jobs:
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v4
+      - uses: antontroshin/setup-go@bda02de8887c9946189f81e7e59512914aeb9ea4
```
v1v, line 48: To play safe, this is a fork commit, and somehow I feel we could raise the fix to the upstream maybe, instead of using the fork from someone else.

I understand the benefits, but I'm a bit concerned about opening this

v1v: OTOH, we are using `permissions: contents: read` by default. But according to our guideline, we need to be cautious to use third-party actions.

v1v: We can give a go; the only thing could be the one whether the commit vanished

v1v (approved):

raultorrecilla (approved):

PR #15733, opened 2025-02-14
benchmark: increase worker instance size

marclop:
Doubles the worker instance size to see if it stabilizes the benchmark. The worker is currently running out of memory after 3h of running the benchmark.


simitt (approved):

carsonip:
I believe the root cause for OOM is https://github.com/elastic/apm-perf/issues/221

PR #15820, opened 2025-02-20
CI(smoketest): fetch previous version on first minor

inge4pres:
### Reason for this PR

Fixing smoke tests `basic-upgrade` that is failing fetching a negative minor version from the ECH API.

```
-> unspecified version, using 9.0
-> Running basic upgrade smoke test for version 9.0.0
-> Adding tfvar stack_version="9.-1.0"
```

### Details

Introduce a new function in smoke tests `lib.sh` to fetch the full latest version of a given prefix.
Use that function to populate the previous major latest minor.patch.

### DIFF
--- a/testing/smoke/basic-upgrade/test.sh
+++ b/testing/smoke/basic-upgrade/test.sh
@@ -29,9 +29,15 @@ elif [[ ${MAJOR_VERSION} -eq 8 ]] || [[ ${MAJOR_VERSION} -eq 9 ]]; then
     get_latest_patch "${MAJOR_VERSION}.${MINOR_VERSION}"
     LATEST_VERSION=${MAJOR_VERSION}.${MINOR_VERSION}.${LATEST_PATCH}
 
-    PREV_MINOR=$(( ${MINOR_VERSION} -1 ))
-    get_latest_patch "${MAJOR_VERSION}.${PREV_MINOR}"
-    PREV_LATEST_VERSION=${MAJOR_VERSION}.${PREV_MINOR}.${LATEST_PATCH}
+    # when the minor is 0, we want to fetch the latest patch of the previous major
+    if [[ ${MINOR_VERSION} -eq 0 ]]; then
+        PREV_MAJOR=$(( ${MAJOR_VERSION} -1 ))
+        PREV_LATEST_VERSION=$(get_latest_version "${PREV_MAJOR}")
+    else
+      PREV_MINOR=$(( ${MINOR_VERSION} -1 ))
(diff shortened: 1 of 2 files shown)


carsonip (commented):
I'm now a little bit confused. Do you mind listing out (in this PR or as a comment in the script) the upgrade scenarios that we want to test? Apparently in 7, we're testing patch upgrades but in 8 we're testing minor upgrade, and in 9 we're testing either major and minor upgrade.

inge4pres:
> Do you mind listing out (in this PR or as a comment in the script) the upgrade scenarios that we want to test?

I am not planning on changing the existing test strategy, simply fixing the bug that is currently happening when the minor version is 0.
I must be lacking understanding of the purpose of each smoke test, but from what i can see the `basic-upgrade` scenario tries to perform an upgrade fro the previous version to the next.
Currently this is broken and this PR tries to fix it.

> Apparently in 7, we're testing patch upgrades but in 8 we're testing minor upgrade, and in 9 we're testing either major and minor upgrade.

Are you referring to why there's a series of nested `if`s for the 3 major versions 7/8/9?
I have no idea to be honest, but the differences seem to be in how the tests are conducted, not much in the difference of version.
It looks to me like the intent is to always test the upgrade to the current latest version, starting from the version immediately prior to it. If that jump involves a mejor version upgrade, that becomes just a detail?

carsonip:
TBH i don't have all the answers here, but I believe we need to clarify what scenarios need to be tested, and only after that we can fix the scripts to do so.

e.g. smoke-tests-ess is only running against 7.17 and latest (which is 9.0 now), which doesn't look right: https://github.com/elastic/apm-server/blob/8e48d588bf0dd5520c039c99bbf5cc29ffde7eea/.github/workflows/smoke-tests-ess.yml#L54-L55

And also I still don't understand why basic-upgrade is testing patch upgrade for major 7, and minor upgrade for major 8, and possibly major upgrade for 9.0 and minor upgrade for 9.1+. Is a major version upgrade in scope for basic-upgrade?

Once we find out the answers, we should document them in a readme in `testing/smoke`. It is currently unclear what scenarios we are trying to cover.

cc @kruskall

inge4pres:
> Once we find out the answers, we should document them in a readme in `testing/smoke`. It is currently unclear what scenarios we are trying to cover.

LGTM, seems like we could still remove the bug and then iterate further to clarify the testin scenarios?

carsonip (approved):
I think it is probably fine to add 8.18->9.0 upgrade to basic-upgrade in this PR. We will need a follow-up to re-enable 8.x upgrade smoketest now that latest is 9, not 8.

PR #15837, opened 2025-02-22
Observability intake APIs

lcawl:
This PR makes the following changes to the OpenAPI files:

- Adds overlays to remove `patternProperties`, which is not yet supported in Bump.sh
- Adds overlays to set `additionalProperties` to `true` (if desired, we can put an explanation of the naming pattern in the descriptions and/or add more details about the supported objects there too, but this will be more effort to maintain if the original source files change).
- Adds an overlay to remove situations where `anyOf` and `allOf` were being used to indicate when/if properties are required, since this also seems to cause problems in Bump.sh
- Updates the `make api-docs` command to apply the overlays

The goal is to be able to publish this content on https://www.elastic.co/docs/api/ and not migrate the [APM Server API](https://www.elastic.co/guide/en/observability/current/apm-api.html) to the new docs system.

Here's an example of a preview before and after the fixes:

## Before

image

## After

image

image

## Next steps

I think that what's being asserted in places like this:

```
              allOf:
                - if:
                    properties:
                      counts:
                        type: array
                    required:
                      - counts
                  then:
                    properties:
                      values:
                        type: array
                    required:
                      - values
```
Is the same as what's being asserted in the description:

```
     properties:
        counts:
           description: >
              Counts holds the bucket counts for histogram metrics. 
              These numbers must be positive or zero.
              If Counts is specified, then Values is expected to be specified with the same number of elements, and with the same order.
```

So in my opinion given that this OpenAPI document is used only for documentation, it's sufficient to explain these dependencies in the text and we don't need to wait for additional functionality to publish these docs.


lcawl:
I guess in that case if you still want some version of these API docs published, rather than changing to yaml and commenting out the problematic sections, we could just use overlays to remove those sections as a kind of post-processing. I will try drafting that up in this PR.

lcawl:
> @lcawl maybe we could only publish the API paths and then link to the `docs/spec` in the apm-server repo for more details of the request body spec?

I've reset the PR to do the fixes via overlays and added details in the introduction about where to find the source files.

> The API docs are mainly relevant for anyone developing a custom client against it, which is not something we officially support.

IMO that sounds like something that should be explicitly added to the `info.description` too. The only thing close to that in the existing docs was in https://www.elastic.co/guide/en/observability/current/apm-api-events.html "Most users do not need to interact directly with the events intake API"... but it sounds like this ought to be a stronger warning.

docs/spec/openapi/apm-openapi.yaml
```diff
@@ -1,7 +1,13 @@
 openapi: 3.1.0
 info:
   title: Observability Intake APIs
-  description: OpenAPI schema for Observability Intake APIs
+  description: |
+    ## Documentation source and versions
+
+    This documentation is derived from the `main` branch of the [apm-server](https://github.com/elastic/apm-server) repository.
+    For the unabbreviated OpenAPI document, refer to [docs/spec/openapi](https://github.com/elastic/apm-server/tree/main/docs/spec/openapi).
```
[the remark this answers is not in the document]
lcawl: I presumed that it needed to be applied as an overlay but if you're fine with that info existing in the source files, that's great!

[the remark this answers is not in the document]
lcawl: I've regenerated the output and it seems to have fixed that link.

simitt:
@lcawl I am not sure how the overlay logic works, how can I take a look at the final output here?
I'm a bit confused about the docs now being in json and yaml and how to keep this up to date if something changes.

lcawl:
> @lcawl I am not sure how the overlay logic works, how can I take a look at the final output here? I'm a bit confused about the docs now being in json and yaml and how to keep this up to date if something changes.

To take a look at the final output you'd run the same `make api-docs` command but look at the new `public-docs.yaml` output file. Happy to discuss further if this is not satisfactory output or process.

docs/spec/openapi/public-docs.yaml
```diff
@@ -0,0 +1,5703 @@
...
+          name: service.environment
+          schema:
+            type: string
+      responses:
+        '200':
+          description: A successful response.
+          content:
+            application/json:
+              schema:
+                type: object
```
simitt, line 96: The response body is missing here; should be the same as for making a `post` request to this endpoint.

docs/spec/openapi/public-docs.yaml
```diff
@@ -0,0 +1,5703 @@
...
+                    example: 0.3
+                additionalProperties: true
+        '403':
+          description: >-
+            APM Server is configured to fetch agent configuration from
+            Elasticsearch but the configuration is invalid.
+        '503':
+          description: APM Server is starting up or Elasticsearch is unreachable.
+  /config/v1/rum/agents:
+    get:
```
simitt, line 176: this is also supported as `post` operation

docs/spec/openapi/public-docs.yaml
```diff
@@ -0,0 +1,5703 @@
...
+      tags:
+        - agent config
+      responses:
+        '200':
+          description: Successful response.
+  /intake/v2/events:
+    post:
+      summary: Send APM agent events
+      description: >
+        NOTE: Most users do not need to interact directly with the events intake
```
simitt, line 195: >Most users do not need to interact directly with the events intake API.

This NOTE applies to all of these APIs.

docs/spec/openapi/public-docs.yaml
```diff
@@ -0,0 +1,5703 @@
...
+      requestBody:
+        content:
+          application/ndjson:
+            schema:
+              anyOf:
+                - $ref: '#/components/schemas/ErrorEvent'
+                - $ref: '#/components/schemas/MetadataEvent'
+                - $ref: '#/components/schemas/MetricSetEvent'
+                - $ref: '#/components/schemas/SpanEvent'
+                - $ref: '#/components/schemas/TransactionEvent'
```
simitt, line 220: Every requests needs to have the `MetadataEvent` as the first line in the `ndjson` request, see https://www.elastic.co/docs/solutions/observability/apm/elastic-apm-events-intake-api#apm-api-metadata-overview.

docs/spec/openapi/public-docs.yaml
```diff
@@ -0,0 +1,5703 @@
...
+      responses:
+        '202':
+          description: Successful response; all events succeeded.
+  /intake/v3/rum/events:
```
simitt, line 251: The request body for v3 rum intake is actually different, but I don't think we had documented this correctly in the past. The spec is available in https://github.com/elastic/apm-server/tree/main/docs/spec/rumv3. 
I am not convinced it is worth documenting this API though, afaik it is barely used.

simitt (commented):
Apologies for stalling on this PR. I left a bunch of comments. 
I did not review all >5500k lines of definition for the concrete event specification, assuming that these have been auto-generated from some spec?

docs/spec/openapi/public-docs.yaml
```diff
@@ -0,0 +1,5703 @@
...
+        content:
+          application/json:
+            schema:
+              type: object
+              required:
+                - service
+              properties:
+                CAPTURE_BODY:
+                  type: string
+                  example: 'off'
```
carsonip, line 119: capture_body exists in the response, but I don't find anything about it in the request handling code. The fact that it is all caps makes it look a bit out of place as well.

docs/spec/openapi/public-docs.yaml
```diff
@@ -0,0 +1,5703 @@
...
+      summary: Send OTLP metrics
+      description: >
+        APM Server supports receiving metrics over the OpenTelemetry Protocol
+        (OTLP). This API uses the OTLP/HTTP (protobuf) communication protocol.
+      operationId: postOtlpHttpMetrics
+      tags:
+        - opentelemetry intake
+      requestBody:
+        content:
+          application/json:
```
carsonip, line 334: otlp/http should use content-type `application/x-protobuf` https://opentelemetry.io/docs/specs/otlp/#binary-protobuf-encoding. But not sure if it is supported in openapi

docs/spec/openapi/public-docs.yaml
```diff
@@ -0,0 +1,5703 @@
...
+      summary: Send OTLP traces
+      description: >
+        APM Server supports receiving traces over the OpenTelemetry Protocol
+        (OTLP). This API uses the OTLP/HTTP (protobuf) communication protocol.
+      operationId: postOtlpHttpTraces
+      tags:
+        - opentelemetry intake
+      requestBody:
+        content:
+          application/json:
```
carsonip, line 351: ditto

docs/spec/openapi/public-docs.yaml
```diff
@@ -0,0 +1,5703 @@
...
+      summary: Send OTLP logs
+      description: >
+        APM Server supports receiving logs over the OpenTelemetry Protocol
+        (OTLP). This API uses the OTLP/HTTP (protobuf) communication protocol.
+      operationId: postOtlpHttpLogs
+      tags:
+        - opentelemetry intake
+      requestBody:
+        content:
+          application/json:
```
carsonip, line 368: ditto

lcawl:
> The files in /docs/spec/rumv3 and /docs/spec/v2 are auto-generated from code in https://github.com/elastic/apm-data/blob/main/input/elasticapm/internal/modeldecoder/generator/cmd/main.go and then copied to the apm-server repo. The spec files are not mainly used for docs but for the development workflow. 

I think this means that all of the feedback provided on `docs/spec/openapi/public-docs.yaml` in this PR must actually be accomplished via updates to your code, since that ultimately affects what's generated by the `make-api-docs` command. 

> @lcawl maybe we could only publish the API paths and then link to the docs/spec in the apm-server repo for more details of the request body spec?

I think removing entire request and response body from each operation in the OpenAPI document is feasible via an overlay.
From my point of view as long as we end up with something that's useful to our readers, automated, and can replace the manually-maintained information in https://www.elastic.co/docs/solutions/observability/apm/apm-server-api I'd be happy.

cc'ing @colleenmcginnis in case there are alternative visions for the future of this content.

colleenmcginnis:
>@lcawl I am not convinced that switching these specs to yaml is the way forward. I'd much rather not have the API documented to this level, than changing the existing code and behaviour due to some documentation restrictions. The API docs are mainly relevant for anyone developing a custom client against it, which is not something we officially support. If needed, the spec files would still be available in the repositories, just not be part of official docs. There is also no active development planned for the APIs.

>>The files in /docs/spec/rumv3 and /docs/spec/v2 are auto-generated from code in https://github.com/elastic/apm-data/blob/main/input/elasticapm/internal/modeldecoder/generator/cmd/main.go and then copied to this apm-server repo. The spec files are not mainly used for docs but for the development workflow.
>
>I think this means that all of the feedback provided on docs/spec/openapi/public-docs.yaml in this PR must actually be accomplished via updates to your code, since that ultimately affects what's generated by the make-api-docs command.

@lcawl @simitt based on these comments... if it's complicated to make updates upstream in order for the changes to make it into the API docs _and_ these API docs are mainly for developing custom clients, we could keep them in their current format. We could even move those Markdown files to this repo for easier maintenance.

>>The API docs are mainly relevant for anyone developing a custom client against it, which is not something we officially support.
>
>IMO that sounds like something that should be explicitly added to the info.description too. The only thing close to that in the existing docs was in https://www.elastic.co/guide/en/observability/current/apm-api-events.html "Most users do not need to interact directly with the events intake API"... but it sounds like this ought to be a stronger warning.

If we really wanted to emphasize that these APIs are not helpful to the average APM user, we could consider moving them to https://www.elastic.co/docs/extend/ where we document workflows for creating integrations and plugins (and making other contributions to Elastic).

lcawl:
I agree that this PR is no longer the best path forward so I'm closing.

PR #15856, opened 2025-02-24
changelog: add 8.16.5 entry

1pkg:
## Motivation/summary

Add entry for 8.16.5 changelog.


1pkg:
`.asciidoc` is blocking, will try to recreate a new PR instead

1pkg:
finally works after recreating branch and using different commit hash in https://github.com/elastic/apm-server/pull/15983

PR #15884, opened 2025-02-26
fix: label terraform ECH deployments

kuisathaverat:
### What

Enable the tags module for the smoke tests
Set `project: apm-server` and `subproject` to the type of smoke tests

### Why

This deployment does not use any label, but they use the robots org for deployments. The CI that creates the deployments often fails to clean the resources because they are not tagged, we cannot identify the leftovers to clean them. This PR adds the proper tags for all manual terraform deployments that do not use the oblt tools and use the robots organization for deployment.


v1v (changes requested):
Let me contribute to your PR

v1v (dismissed):

testing/smoke/main.tf
```diff
@@ -10,6 +10,21 @@ terraform {
 
 provider "ec" {}
 
+module "tags" {
```
v1v, line 13: That's what https://github.com/elastic/apm-server/pull/10637 actually did

testing/smoke/main.tf
```diff
@@ -10,6 +10,21 @@ terraform {
...
+  source  = "../infra/terraform/modules/tags"
+  project = "smoke-test"
```
kuisathaverat, line 15: `smoke-test` does not give me any information valid to identify where the deployment come from, àpm-server` does

v1v: see `repo`:
- https://github.com/elastic/apm-server/pull/15884/files#diff-1b0205a341d2e4aec28dde78c2d93be8fc938d7ac1feb9228e4b4bb1154a478fR21

kuisathaverat: I said it, `repo` has the repository, I do not talk about that. I am saying that `project: smoke-test` means nothing. `smoke-test` is a kind of test is not a project, here the only project I see is `apm-server`

v1v: The `repo: apm-server` and `project: smoke-tests` tags/labels are meaningful, IMO. `project` is quite generic and is in fact, what the APM Server team has done in other places:

- https://github.com/search?q=repo%3Aelastic%2Fapm-server+project+language%3AHCL&type=code&l=HCL

If it is changed here, I'd suggest changing it to other places - to help with the semantic

I do understand caching is a bit easier than naming things 👼

v1v (approved):

endorama (approved):

PR #15901, opened 2025-02-27
smoke: Fix smoke test Terraform issue

ericywl:
## Motivation/summary

Fix smoke test issue:
```
-> Adding tfvar integrations_server="false"
-> Applying terraform configuration...

Error: Invalid index

  on ../../infra/terraform/modules/ec_deployment/outputs.tf line 53, in output "apm_component_id":
  53:   value       = ec_deployment.deployment.integrations_server[0].resource_id
    ├────────────────
    │ ec_deployment.deployment.integrations_server is empty list of object

The given key does not identify an element in this collection value: the
collection has no elements.

Error: Invalid index

  on ../../infra/terraform/modules/ec_deployment/outputs.tf line 58, in output "fleet_component_id":
  58:   value       = ec_deployment.deployment.integrations_server[0].resource_id
    ├────────────────
    │ ec_deployment.deployment.integrations_server is empty list of object

The given key does not identify an element in this collection value: the
collection has no elements.
```

## How to test these changes

Run smoke test and functional test again.

Smoke: https://github.com/elastic/apm-server/actions/runs/13650931063 (There are errors, but unrelated to the fix)
Functional: https://github.com/elastic/apm-server/actions/runs/13650934968


testing/infra/terraform/modules/ec_deployment/outputs.tf
```diff
@@ -50,11 +50,11 @@ output "admin_console_url" {
...
 
 output "apm_component_id" {
-  value       = ec_deployment.deployment.integrations_server[0].resource_id
+  value       = var.integrations_server ? ec_deployment.deployment.integrations_server[0].resource_id : ec_deployment.deployment.apm[0].resource_id
   description = "APM Cloud component ID"
 }
 
 output "fleet_component_id" {
-  value       = ec_deployment.deployment.integrations_server[0].resource_id
+  value       = var.integrations_server ? ec_deployment.deployment.integrations_server[0].resource_id : ""
```
ericywl, line 58: [For reviewers] Not sure what this should fallback to

ericywl:
Seems like there is still this issue (which is from before?):

```
Error: failed tracking create progress: 2 errors occurred:
	* found deployment plan errors: deployment [2ae8abcda33623d520648761a5f7b15a] - [apm][15bf782c863f4d8886efe95cc106e386]: caught error: "Plan change failed: Some instances were not running"
	* set "request_id" to "h8e5sbnwngz1qquqj7wm0a8emq4pqj9n4sgcwryug2ga41ljtkobsc49l4d7ruuy" to recreate the deployment resources

  with module.ec_deployment.ec_deployment.deployment,
  on ../../infra/terraform/modules/ec_deployment/deployment.tf line 17, in resource "ec_deployment" "deployment":
  17: resource "ec_deployment" "deployment" {
```

testing/infra/terraform/modules/ec_deployment/outputs.tf
simitt: Why is the distinction between `integrations_server` and `apm` resource even bleeding into our code? Having one variable defined in terraform that abstracts this sound the right path forward. @ericywl you are on a good path here already, adding the fallback for `apm` resource. @endorama is there any reason why `apm_component_id` and `fleet_component_id` would be needed? (Given that they currently point to the same resource I would not think so). 

Quick recap: in cloud the old resource was `apm`, and the newer one (introduced in `8.0`) is `integrations_server`. Currently both can exist, for deployments created in `7.x` the resource will be `apm`, for newer deployments resource will be `integrations_server` (if not explicitly migrated, which we don't do here).

endorama: @simitt I split them because that's what's done in the UI and what I expect someone to look for in this module output. I considered the fact they are referencing the same value as an implementation detail that should not leak from this module. We could have `apm_standalone_id` and `integrations_server_id` in place, but in 8.x both APM Server and Fleet Server have the same ID which to me just leaks an unnecessary information (if I'm searching for the APM Component ID I have to know that that's the Integrations Server ID). That's not the case with the current implementation.

I missed that `integrations_server` would have 0 items when running in 7.x, sorry about that!

@ericywl what you did is perfectly fine to me. In 7.x there is no fleet so it's reasonable and expected to me to be empty.

testing/infra/terraform/modules/ec_deployment/outputs.tf
```diff
@@ -50,11 +50,11 @@ output "admin_console_url" {
...
   description = "Fleet Server Cloud component ID"
```
endorama, line 59: I'd suggest to mention that this field is expected to be empty in some cases:
```suggestion
  description = "Fleet Server Cloud component ID; empty for 7.x clusters"
```

testing/infra/terraform/modules/ec_deployment/outputs.tf
simitt: >I considered the fact they are referencing the same value as an implementation detail that should not leak from this module.

They are not referencing the same value in ECH. If you really need a distinction between `apm` vs. `integrations_server` component, then the `apm` component cannot query for `integrations_server`, but rather can only assign `ec_deployment.deployment.apm[0].resource_id`. It is either or - a deployment can not have both components. 

>We could have `apm_standalone_id` and `integrations_server_id` in place, but in 8.x both APM Server and Fleet Server have the same ID which to me just leaks an unnecessary information (if I'm searching for the APM Component ID I have to know that that's the Integrations Server ID).

The `apm` resource can either carry a standalone or a fleet managed component. 
The `integrations_server` resource only holds fleet managed components. 
Therefore having the distinction between the two resources does not provide a lot of value, `apm` is not always `standalone`, it was simply the old resource name. 

I am still not convinced that this separation is needed. @endorama what is missing or bleeding through if we simply have this one component instead of two? 
```
output "ech_apm_component_id" {
  value       = var.integrations_server ? ec_deployment.deployment.integrations_server[0].resource_id : ec_deployment.deployment.apm[0].resource_id
  description = "APM or Integrations Server Cloud component ID"
}
```

endorama: > The apm resource can either carry a standalone or a fleet managed component.

The [documentation](https://registry.terraform.io/providers/elastic/ec/latest/docs/resources/deployment) for `apm` states:
> DEPRECATED APM cluster definition. This should only be used for deployments running a version lower than 8.0
whereas for `integrations_server`
> Integrations Server cluster definition. Integrations Server replaces apm in Stack versions > 8.0

Is there an edge case hidden here for clusters that migrate from 7 to 8? Otherwise I would not expect it to carry the 2 different components (from top of my mind I remember that in my tests `apm` in 8.x was not providing the correct values, but I', not 100% sure here).

I would bring back the focus here on unblocking this PR which is needed to fix smoke tests.
The `ech_apm_component_id` solution proposed here is fine (and quite elegant 👍). `fleet_id` could be skipped for the moment, even if I'd keep it for clarity.
@ericywl please also amend the functional tests where this variables are used 🙏

ericywl: Updated in [e47c471](https://github.com/elastic/apm-server/pull/15901/commits/e47c47161ae8658c284232d98df0e39270058a08). Hopefully we can merge this soon 🙏

endorama (dismissed):
Thanks!

ericywl:
Seems like smoke tests are broken by another PR: https://github.com/elastic/apm-server/pull/15884. I think considering that it was backported, it may be good to create a separate PR to fix that 🤦

testing/infra/terraform/modules/ec_deployment/outputs.tf
simitt: Moved the discussion to a different forum; should not be blocking this PR.

simitt (approved):

endorama:
@ericywl I see 2 failures in smoke tests `legacy-standalone-major-managed` and `basic-upgrade`. Was this expected?

ericywl:
> @ericywl I see 2 failures in smoke tests legacy-standalone-major-managed and basic-upgrade. Was this expected?

Yeah, those are unrelated to the fix. One of them is due to 7.x -> 9.0 migration being disallowed, the other one is an actual failure that I'm not sure what's the cause of.

PR #15903, opened 2025-02-27
smoke: Add note in smoke-tests about checkout ref

ericywl:
## Motivation

I kept trying to run `smoke-tests` with my changes by selecting the branch from the dropdown, only to have the smoke tests run without changes (because [`smoke-tests-schedule` checks out from `main`](https://github.com/elastic/apm-server/blob/main/.github/workflows/smoke-tests-schedule.yml#L41-L49)). Ultimately it is PEBKAC, but I thought it would be great to have a note there so others in the future will know.

## Changes

- Add note in `smoke-tests` about ref checkout
- Remove workflow dispatch input in `smoke-tests-ess` and `smoke-tests-es` since you can use the branch dropdown instead. 

## How to test these changes

Run smoke test workflows with this branch.


endorama (commented):
I understand the comment in `.github/workflows/smoke-tests-schedule.yml`, but why removing the `workflow_dispatch` configs in the other workflows?

ericywl:
> I understand the comment in .github/workflows/smoke-tests-schedule.yml, but why removing the workflow_dispatch configs in the other workflows?

I closed this PR previously because I was unsure about that. I'm not sure what's the use case of the `workflow_dispatch`. From my experience, when we want to trigger `smoke-test-ess` or `smoke-test-os`, we press the "Run workflow" button and get this:

Screenshot 2025-03-13 at 5 11 08 PM

In my case, I usually just put the same branch for both. So I don't see the need for "Branch to checkout". Putting `workflow_dispatch: ~` just means that there are no inputs for dispatching.

endorama:
@ericywl that make sense thanks! I think we should move forward with this, as I see no benefits in the current situation.

1pkg (approved):

Issue #13606, opened 2024-07-05
Flaky test TestServerElasticsearchOutput


lahsivjar:
```
=== RUN   TestServerElasticsearchOutput
    server_test.go:523: 
        	Error Trace:	D:/a/apm-server/apm-server/internal/beater/server_test.go:523
        	Error:      	Not equal: 
        	            	expected: map[string]interface {}{"elasticsearch":map[string]interface {}{"bulk_requests":map[string]interface {}{"available":9, "completed":0}, "indexers":map[string]interface {}{"active":1, "created":0, "destroyed":0}}}
        	            	actual  : map[string]interface {}{"elasticsearch":map[string]interface {}{"bulk_requests":map[string]interface {}{"available":8, "completed":0}, "indexers":map[string]interface {}{"active":1, "created":0, "destroyed":0}}}
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -3,3 +3,3 @@
        	            	   (string) (len=13) "bulk_requests": (map[string]interface {}) (len=2) {
        	            	-   (string) (len=9) "available": (int64) 9,
        	            	+   (string) (len=9) "available": (int64) 8,
        	            	    (string) (len=9) "completed": (int64) 0
        	Test:       	TestServerElasticsearchOutput
```

carsonip:
To reproduce, run `go test ./... -run=^TestServerElasticsearchOutput$ -count 1000 -failfast`

PR #13827, opened 2024-08-02
test: enable race detector

kruskall:
## Motivation/summary

Unfortunately we can only enable the race detector for ci and unit tests.
system-tests are building the apm-server binary (CGO disabled by default) so we can't enable the race detector as it requires cgo.

## Checklist

- [ ] Update [CHANGELOG.asciidoc](https://github.com/elastic/apm-server/blob/main/CHANGELOG.asciidoc)
- [ ] Documentation has been updated

For functional changes, consider:
- Is it observable through the addition of either **logging** or **metrics**?
- Is its use being published in **telemetry** to enable product improvement?
- Have system tests been added to avoid regression?

PR #15932, opened 2025-02-28
test: Enable race detector

ericywl:
## Motivation/summary

Enable race detector for tests, continuation of https://github.com/elastic/apm-server/pull/13827.

Blocked by https://github.com/elastic/apm-server/pull/16061 and https://github.com/elastic/apm-server/pull/15992.

## How to test these changes

Run unit test, system test and CI.


internal/beater/server_test.go
```diff
@@ -489,10 +490,13 @@ func TestServerElasticsearchOutput(t *testing.T) {
...
-		"elasticsearch.events.count":            5,
-		"elasticsearch.events.queued":           5,
-		"elasticsearch.bulk_requests.available": 9,
+	monitoringtest.ExpectContainOtelMetricsFunc(t, reader, map[string]monitoringtest.AssertionFunc{
+		"elasticsearch.events.count":  func(t *testing.T, actual any) bool { return actual.(int64) == 5 },
+		"elasticsearch.events.queued": func(t *testing.T, actual any) bool { return actual.(int64) == 5 },
+		"elasticsearch.bulk_requests.available": func(t *testing.T, actual any) bool {
+			// Available should be less than maxRequests since some will be used
+			return 0 < actual.(int64) && actual.(int64) < maxRequests
+		},
```
ericywl, line 499: [For reviewer] See commit https://github.com/elastic/apm-server/pull/15932/commits/a6f9618851c66727f59e858370a9ea00e040ba1d for reasoning behind change.

carsonip (commented):
is the part that fixes flaky TestServerElasticsearchOutput related to the race detector? If no, I'd prefer it to be a separate PR to fix https://github.com/elastic/apm-server/issues/13606 , as we might want to backport it to make our lives easier.

ericywl:
> is the part that fixes flaky TestServerElasticsearchOutput related to the race detector? If no, I'd prefer it to be a separate PR to fix https://github.com/elastic/apm-server/issues/13606 , as we might want to backport it to make our lives easier.

Not exactly, I can extract that PR out.

1pkg (dismissed):

internal/beater/server_test.go
```diff
@@ -490,9 +490,10 @@ func TestServerElasticsearchOutput(t *testing.T) {
 	}
 
 	monitoringtest.ExpectContainOtelMetrics(t, reader, map[string]any{
-		"elasticsearch.events.count":            5,
-		"elasticsearch.events.queued":           5,
-		"elasticsearch.bulk_requests.available": 9,
+		"elasticsearch.events.count":  5,
+		"elasticsearch.events.queued": 5,
+		// Disable this flaky test to be fixed later
+		// "elasticsearch.bulk_requests.available": 9,
```
simitt, line 496: Can you please create a follow up task to fix this issue and link from here? 
cc @raultorrecilla for scheduling work to look into why this is flaky.

raultorrecilla: @ericywl paste it here once created please!

ericywl: Yup apparently its a long standing issue: https://github.com/elastic/apm-server/issues/13606

.github/workflows/ci.yml
```diff
@@ -55,16 +55,13 @@ jobs:
         with:
           go-version-file: go.mod
           cache: true
+      - run: go test -v -race ./...
```
endorama, line 58: Why are we removing `CGO_ENABLED: "0"` here?

ericywl: Race detector unfortunately requires CGO to be enabled: https://go.dev/doc/articles/race_detector.

> The race detector requires cgo to be enabled, and on non-Darwin systems requires an installed C compiler. The race detector supports linux/amd64, linux/ppc64le, linux/arm64, linux/s390x, freebsd/amd64, netbsd/amd64, darwin/amd64, darwin/arm64, and windows/amd64.

internal/beater/server_test.go
carsonip: @ericywl can you revert this change in this PR? The flaky test shouldn't block this PR. (Even if the bugfix isn't merged and it unfortunately fails, rerunning the CI should be good enough)

docker-compose.yml
```diff
@@ -20,6 +20,7 @@ services:
       interval: 1s
     environment:
       - "ES_JAVA_OPTS=-Xms1g -Xmx1g"
+      - _JAVA_OPTIONS # Workaround on OSX for https://bugs.openjdk.org/browse/JDK-8345296
```
carsonip, line 23: Q: Is the key `_JAVA_OPTIONS` without value good enough to workaround your issue? I don't see any mention of `_JAVA_OPTIONS` in the bug report.

internal/beater/otlp/http.go
```diff
@@ -57,8 +61,9 @@ func NewHTTPHandlers(logger *zap.Logger, processor modelpb.BatchProcessor, semap
 
 	// TODO we should add an otel counter metric directly in the
 	// apm-data consumer, then we could get rid of the callback.
+	metricRegistrationMu.Lock() // Temporary locking to fix data race issue
```
carsonip, line 64: Any reason to not use `defer` for unlock?

carsonip, line 64: Also, is this a bugfix? If so, I prefer a separate PR, and if it is user-facing, it needs a changelog as well.

.github/workflows/ci.yml
endorama: I think we explicitly set it to `0` to make sure we do not include any C based library. Has this change been discussed previously?

docker-compose.yml
ericywl: It's mainly so that I can run the system tests using `_JAVA_OPTIONS="-XX:UseSVE=0" make system-test` without having to update the docker compose.

internal/beater/otlp/http.go
ericywl: This has data race because the NewHTTPHandlers can be called by different goroutines concurrently.

.github/workflows/ci.yml
ericywl: Hmmm, @kruskall I based this off your previous PR to enable the race detector. Not sure if that's something that has been discussed.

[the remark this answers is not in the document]
ericywl: Sure, let me update it.

internal/beater/server_test.go
ericywl: Sure.

ericywl:
System test failed previously due to `Test_warmup` but is suddenly passing again when I tried to look into it 🤡

internal/beater/otlp/http.go
ericywl: Used defer instead.

.github/workflows/ci.yml
endorama: I confirmed with @kruskall that this is ok. `systemtests` run without CGO, so we are covered on that side.
I would add a comment to document this for future reference

internal/beater/otlp/http.go
carsonip: - This is not changing test code (`*_test.go`). This is a code path that is actually used in apm-server. I think there is a possibility (very slim) that it runs concurrently when there's a hot reload triggered by EA.
- This may not need a changelog, because it is an unreleased change in 9.0 only. This is potentially a 9.0 regression that we should fix before the actual 9.0 release.
- If you search for `RegisterCallback`, there seems to be similar usages that look vulnerable to data race.
- Can we have a separate PR that fixes this race, backported to 9.0?

1pkg (approved):

PR #15948, opened 2025-02-28
test: Fix flaky esoutput

ericywl:
## Summary

Fixes https://github.com/elastic/apm-server/issues/13606.

## How to test these changes

Run `go test ./... -run=^TestServerElasticsearchOutput$ -count 1000 -failfast`

### DIFF
--- a/internal/beater/monitoringtest/opentelemetry.go
+++ b/internal/beater/monitoringtest/opentelemetry.go
@@ -83,7 +83,7 @@ func ExpectContainOtelMetricsKeys(t *testing.T, reader sdkmetric.Reader, expecte
 // assertOtelMetrics gathers all the metrics from `reader` and runs the AssertionFunc for all metrics specified
 // in `expectedMetrics`.
 //
-// If `fullMatch` is true, all metrics from `reader` must be found in `expectedMetrics`.
+// If `fullMatch` is true, all metrics from `reader` must be found in `expectedMetrics` vice versa.
 // Otherwise, only the metrics in `expectedMetrics` will be checked.
 //
 // If `skipValAssert` is true, the AssertionFunc(s) will be skipped entirely.
--- a/internal/beater/monitoringtest/opentelemetry.go
+++ b/internal/beater/monitoringtest/opentelemetry.go
@@ -26,23 +26,73 @@ import (
 	"go.opentelemetry.io/otel/sdk/metric/metricdata"
 )
 
-func ExpectOtelMetrics(t *testing.T, reader sdkmetric.Reader, expectedMetrics map[string]any) {
-	assertOtelMetrics(t, reader, expectedMetrics, true, true)
+type AssertionFunc func(t *testing.T, actual any) bool
+
+func ExpectOtelMetricsFunc(
+	t *testing.T,
+	reader sdkmetric.Reader,
+	expectedMetrics map[string]AssertionFunc,
+) {
+	assertOtelMetrics(t, reader, expectedMetrics, true, false)
 }
 
-func ExpectContainOtelMetrics(t *testing.T, reader sdkmetric.Reader, expectedMetrics map[string]any) {
-	assertOtelMetrics(t, reader, expectedMetrics, false, true)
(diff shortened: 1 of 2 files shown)


internal/beater/server_test.go
```diff
@@ -489,10 +490,13 @@ func TestServerElasticsearchOutput(t *testing.T) {
...
-		"elasticsearch.events.count":            5,
-		"elasticsearch.events.queued":           5,
-		"elasticsearch.bulk_requests.available": 9,
+	monitoringtest.ExpectContainOtelMetricsFunc(t, reader, map[string]monitoringtest.AssertionFunc{
+		"elasticsearch.events.count":  func(t *testing.T, actual any) bool { return actual.(int64) == 5 },
+		"elasticsearch.events.queued": func(t *testing.T, actual any) bool { return actual.(int64) == 5 },
+		"elasticsearch.bulk_requests.available": func(t *testing.T, actual any) bool {
+			// Available should be less than maxRequests since some will be used
+			return 0 < actual.(int64) && actual.(int64) < maxRequests
+		},
```
ericywl, line 499: [For reviewer] See commit https://github.com/elastic/apm-server/commit/a6f9618851c66727f59e858370a9ea00e040ba1d for reasoning behind change.

internal/beater/monitoringtest/opentelemetry.go
```diff
@@ -56,49 +106,39 @@ func assertOtelMetrics(t *testing.T, reader sdkmetric.Reader, expectedMetrics ma
...
 
-				if v, ok := expectedMetrics[m.Name]; ok {
-					if dp, ok := v.(int); ok {
-						assert.Equal(t, int64(dp), d.DataPoints[0].Value, m.Name)
-					} else {
-						assert.Fail(t, "expected an int value", m.Name)
-					}
-				} else if match {
+				if fn, ok := expectedMetrics[m.Name]; ok {
+					assert.True(t, fn(t, d.DataPoints[0].Value), m.Name)
```
carsonip, line 114: wouldn't it assert one time within the AssertionFunc and another time in `assert.True`?

internal/beater/server_test.go
```diff
@@ -463,13 +463,14 @@ func TestServerElasticsearchOutput(t *testing.T) {
...
 
+	const maxRequests = 10
 	srv := beatertest.NewServer(t, beatertest.WithMeterProvider(mp), beatertest.WithConfig(agentconfig.MustNewConfigFrom(map[string]interface{}{
 		"output.elasticsearch": map[string]interface{}{
 			"hosts":          []string{elasticsearchServer.URL},
 			"flush_interval": "1ms",
 			"backoff":        map[string]interface{}{"init": "1ms", "max": "1ms"},
 			"max_retries":    0,
-			"max_requests":   10,
+			"max_requests":   maxRequests,
```
carsonip, line 473: Will it be easier to just to set it to 1, and assert for 0?
```suggestion
			"max_requests":   1,
```

carsonip (commented):
Thanks for attempting to fix this annoying flaky test! It's been on my TODO for months. I understand the need to refactor to check if available bulk request is in a range. However it adds a bit of complexity to the tests only for this single use case. Do you think we can fix the flakiness just by setting max_requests to 1?

carsonip:
Adding backport-active-8 and backport-active-9 so that we won't run into the same flaky test during development of active versions

internal/beater/server_test.go
ericywl: Oh man that is indeed easier... Let me simplify it.

carsonip (approved):
lgtm, the flakiness is fixed. Thanks for the detailed comment. I don't mind the refactoring in `opentelemetry.go` but expect some conflict backporting to 8.x as the matchVal option doesn't exist over there.

ericywl:
> I don't mind the refactoring in opentelemetry.go but expect some conflict backporting to 8.x as the matchVal option doesn't exist over there.

No problem, I will resolve the conflict where necessary 😄

PR #15982, opened 2025-03-03
changelog: add 8.16.5 entry

1pkg:
## Motivation/summary

Add entry for 8.16.5 changelog.


1pkg:
branch protection rule still picking the wrong pipeline, will try a different branch name

1pkg:
finally works after recreating branch and using different commit hash

Issue #15991, opened 2025-03-04
Flaky system tests


ericywl:
There are 2 system tests that are flaky currently.
Note: I could not reproduce it on local environment, but it happens a lot in CI.

```
=== RUN   Test_warmup/4_agent_2s
    main_test.go:95: 
        	Error Trace:	/home/runner/work/apm-server/apm-server/systemtest/benchtest/main_test.go:95
        	Error:      	"0" is not greater than "4"
        	Test:       	Test_warmup/4_agent_2s
```

```
=== RUN   TestOTLPRateLimit
    otlp_test.go:479: 
        	Error Trace:	/home/runner/work/apm-server/apm-server/systemtest/otlp_test.go:479
        	Error:      	Not equal: 
        	            	expected: 0x8
        	            	actual  : 0x4
        	Test:       	TestOTLPRateLimit
    otlp_test.go:480: 
        	Error Trace:	/home/runner/work/apm-server/apm-server/systemtest/otlp_test.go:480
        	Error:      	Not equal: 
        	            	expected: "traces export: rpc error: code = ResourceExhausted desc = rate limit exceeded"
        	            	actual  : "traces export: rpc error: code = DeadlineExceeded desc = received context error while waiting for new LB policy update: context deadline exceeded"
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1 +1 @@
        	            	-traces export: rpc error: code = ResourceExhausted desc = rate limit exceeded
        	            	+traces export: rpc error: code = DeadlineExceeded desc = received context error while waiting for new LB policy update: context deadline exceeded
        	Test:       	TestOTLPRateLimit
```

PR #15992, opened 2025-03-04
systemtest: Fix flaky tests

ericywl:
## Summary

Fixes https://github.com/elastic/apm-server/issues/15991.

## How to test these changes

Trigger CI system tests multiple times and confirm that the listed tests don't fail anymore.


systemtest/benchtest/main.go
```diff
@@ -268,17 +268,24 @@ func warmup(agents int, duration time.Duration, url, token string) error {
...
-			defer wg.Done()
-			h.SendBatchesInLoop(ctx)
-		}()
+		eg.Go(func() error {
+			sendErr := h.SendBatchesInLoop(ctx)
+			if sendErr != nil && !errors.Is(sendErr, context.DeadlineExceeded) {
+				return sendErr
+			}
+			return nil
+		})
```
ericywl, line 282: [For reviewers] The behavior is changed slightly here, which might affect `apmbench` warmup i.e. [this benchmark](https://github.com/elastic/apm-server/blob/main/systemtest/cmd/apmbench/main.go#L184-L200) will fail if warmup agents fail to send due to other errors aside from context deadline.

systemtest/benchtest/main_test.go
```diff
@@ -48,17 +48,26 @@ func Test_warmup(t *testing.T) {
...
 			srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 				if r.URL.Path == "/debug/vars" {
+					// Wait until there are completed requests before reporting idle.
+					for completedRequests.Load() == 0 {
+						// NOTE: There were cases in CI environment where this test handler reports idle
+						// before any previous requests were even completed.
+					}
 					// Report idle APM Server.
-					w.Write([]byte(`{"libbeat.output.events.active":0}`))
+					_, _ = w.Write([]byte(`{"libbeat.output.events.active":0}`))
```
rubvs, line 60: I like this explicit style actually.

systemtest/otlp_test.go
```diff
@@ -444,34 +444,36 @@ func TestOTLPRateLimit(t *testing.T) {
...
 			otlptracegrpc.WithRetry(otlptracegrpc.RetryConfig{Enabled: false}),
 		)
-		resource := sdkresource.NewSchemaless(
+		res := sdkresource.NewSchemaless(
 			attribute.String("service.name", "service2"),
 			attribute.String("telemetry.sdk.name", "iOS"),
 			attribute.String("telemetry.sdk.language", "swift"),
 		)
-		return sendOTLPTrace(ctx, newOTLPTracerProvider(exporter, sdktrace.WithResource(resource)))
+		return sendOTLPTrace(ctx, newOTLPTracerProvider(exporter, sdktrace.WithResource(res)))
```
rubvs, line 452: Changing `resource` to `res` is religious, since `res` can also indicate `result`. So I'd keep it explicit with the full name.

systemtest/benchtest/main_test.go
```diff
@@ -48,17 +48,26 @@ func Test_warmup(t *testing.T) {
 	for _, c := range cases {
 		t.Run(fmt.Sprintf("%d_agent_%s", c.agents, c.duration.String()), func(t *testing.T) {
 			var received atomic.Uint64
+			var completedRequests atomic.Uint64
 			srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 				if r.URL.Path == "/debug/vars" {
+					// Wait until there are completed requests before reporting idle.
+					for completedRequests.Load() == 0 {
```
1pkg, line 55: nit: better use a channel to signal or at least add a call to [Gosched](https://pkg.go.dev/runtime#Gosched) here, since spinning in a tight loop could be problematic

1pkg (dismissed):

systemtest/benchtest/main_test.go
ericywl: Good idea, I used channel but switched to this for some reason 🤔, don't think there was an issue.

systemtest/benchtest/main_test.go
```diff
@@ -48,17 +49,43 @@ func Test_warmup(t *testing.T) {
...
-					w.Write([]byte(`{"libbeat.output.events.active":0}`))
+					_, _ = w.Write([]byte(`{"libbeat.output.events.active":0}`))
 					return
 				}
 
 				if !strings.HasPrefix(r.URL.Path, "/intake") {
 					return
 				}
 
+				// Signal that there is at least one request received.
```
1pkg, line 82: If I understand the flow correctly you probably could simplify the synchronization with just:

```
select {
 case requestActive <- struct{}:
 default:
}
```

ericywl: That would not wait until all ongoing requests are done though.

ericywl: Nevermind, I tried but this doesn't work. Because `/debug/vars` is called multiple times, if there are no more goroutines running the `case requestActive <- struct{}{}:` it will block the whole test from completing.

1pkg (approved):

PR #16031, opened 2025-03-05
feat(fips): enforce requirements for fips mode

kruskall:
## Motivation/summary

enforce fips140=only in fips mode
enforce -tags=requirefips in fips mode

## Checklist

### DIFF
--- a/packaging/docker/Dockerfile.fips
+++ b/packaging/docker/Dockerfile.fips
@@ -1,0 +1,86 @@
+ARG GOLANG_VERSION
+
+################################################################################
+# Build stage 0
+# Build the apm-server binary. The golang image version is kept
+# up to date with go.mod by Makefile.
+################################################################################
+FROM docker.elastic.co/wolfi/go-msft-fips:${GOLANG_VERSION} as builder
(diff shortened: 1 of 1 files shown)


internal/fips140/fips.go
```diff
@@ -0,0 +1,24 @@
...
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+//go:build requirefips
+
+package fips140
+
+func CheckFips() {
+	// all good
```
1pkg, line 23: should we actually panic if `!cryptofips.Enabled()` here?

internal/fips140/nofips.go
```diff
@@ -0,0 +1,28 @@
...
+// under the License.
+
+//go:build !requirefips
+
+package fips140
+
+import cryptofips "crypto/fips140"
+
+func CheckFips() {
+	if cryptofips.Enabled() {
```
1pkg, line 25: Why do we need to panic if fips is not required but was enabled anyway?

1pkg (approved):

PR #16060, opened 2025-03-06
 feat(fips): add docker image 

kruskall:
## Motivation/summary

add fips Dockerfile

~blocked until we have go-msft-fips 1.24 image~ image is available

## Checklist


1pkg (approved):

simitt (approved):
There's a lot of copy&paste code between the three Dockerfile files that we now have. WDYT about opening a follow up task to work with templates for code reusage rather than copying?

PR #16061, opened 2025-03-07
fix: Add locks for unsupported metric registration

ericywl:
## Motivation/summary

Related to https://github.com/elastic/apm-server/pull/15932#discussion_r1983473261. This change needs to be backported to 9.0, and so it is in a separate PR.


carsonip (approved):

carsonip:
Added backport-8.x as @kruskall confirmed that the code is in 8.x as well.

1pkg (approved):

PR #16062, opened 2025-03-07
functionaltests: Copy terraform files at start of each test

ericywl:
## Motivation/summary

This PR aims to solve the issue of duplicated Terraform files in functional tests with the simplest possible solution: copy the files at the start of each test instead of having them in Git.

## How to test these changes

Run functional tests: https://github.com/elastic/apm-server/actions/runs/13718916247.


functionaltests/utils_test.go
```diff
@@ -50,6 +51,23 @@ func createAPMAPIKey(t *testing.T, ctx context.Context, ecc *esclient.Client) st
...
+}
+
+// copyTerraforms copies the static Terraform files to the Terraform directory for this test.
+// It will remove all existing files from the test Terraform directory if it exists, before copying into it.
+func copyTerraforms(t *testing.T) {
+	t.Helper()
+	dirName := terraformDir(t)
+	err := os.RemoveAll(dirName)
+	require.NoError(t, err)
+	err = os.CopyFS(terraformDir(t), os.DirFS("terraforms"))
```
endorama, line 67: To keep consistency let's use `_infra/deployment` as folder for the terraform code.

I would also add a README in this folder to briefly explain that:
- the code can be use standalone
- tests will copy it to a separate folder to avoid Terraform state conflicts
- the code must support as many tests cases as possible exposing variables for configuration

functionaltests/README.md
```diff
@@ -22,5 +22,5 @@ This means that the test will start from `from_version`, and be upgraded to `to_
 The file that each test is in is named after the last minor version in the upgrade chain. For example, if the test name
 is `TestUpgrade_7_17_0_to_8_18_0_to_9_0_0_Something`, it should be written in `9_0_test.go`.
 
-The Terraforms for each test is located at the directory with the same name as the test.
-At this moment the terraform code must be copied in a separate folder for each test. We plan to address this duplication later on.
\ No newline at end of file
+The Terraform files for each test are copied from `terraforms/` at the start of each test, into `tf-<test_name>/`
```
endorama, line 25:
```suggestion
The Terraform files for each test are copied from `_infra/deployment/` at the start of each test, into `tf-<test_name>/` 
```

endorama (changes requested):
A minor suggest but looks good!

functionaltests/utils_test.go
ericywl: Sounds good, but what does the `_` in front of `infra` mean in this case?

endorama: `_` makes the file ignored by the Go compiler (see [code](https://github.com/golang/go/blob/b143c981693a0f405f16eade1cccf4933fde8e21/src/go/build/build.go#L571)). 
You'll find it around Go code where we want to signal a folder that we know won't host any Go code (or that must be excluded for other purposes).

endorama (approved):

endorama (approved):

Issue #15844, opened 2025-02-24
smoke: Clarify smoke tests scenarios and support 2 minors


carsonip:
- [x] It is unclear what scenarios are the smoke tests covering. Clarify and document them in the repo.
- [x] Update smoke tests to test 2 minor versions, as we have 2 active minors now. https://github.com/elastic/apm-server/pull/15885
- [ ] Ensure that SNAPSHOT versions are tested

simitt:
Please let me know your thoughts about the following suggestion for what to test in the current state.

**Testing Goals:**

- test next patch version of released and supported minor versions
- test next minor version of released and supported major version
- test next major version if the development for the next major version has started
- test version in FF and in developer branch
- do not test released versions but versions under development

**Currently released versions:**
- 7.17.28
- 8.16.5
- 8.17.3

**Upcoming releases:**
- 8.16.6
- 8.17.4
- 8.18.0
- 8.19.0
- 9.0.0
- 9.1.0

**OS Support Tests**:
- 7.17.29-SNAPSHOT
- 8.16.0-SNAPSHOT
- 8.17.0-SNAPSHOT
- 8.18.0-SNAPSHOT
- 8.19.0-SNAPSHOT
- 9.0.0-SNAPSHOT
- 9.1.0-SNAPSHOT

**Upgrade Tests:**
- 7.17.29-SNAPSHOT -> 8.18.0-SNAPSHOT (somewhat arbitrary to which of the 8.x versions to upgrade to)
- 8.16.6-SNAPSHOT -> 8.17.4-SNAPSHOT
- 8.17.4-SNAPSHOT -> 8.18.0-SNAPSHOT
- 8.18.0-SNAPSHOT -> 8.19.0-SNAPSHOT
- 8.18.0-SNAPSHOT -> 9.0.0-SNAPSHOT
- 8.19.0-SNAPSHOT -> 9.0.0-SNAPSHOT
- 9.0.0-SNAPSHOT -> 9.1.0-SNAPSHOT

ericywl:
For reference, currently available tests:

**Legacy Managed:**
- 7.17.28-SNAPSHOT
- 7.17.29-SNAPSHOT

** Legacy Standalone to Managed**:
- 7.17.28 standalone -> 8.18.0 -> 8.18.0 managed
- 7.17.29-SNAPSHOT standalone -> 9.1.0-SNAPSHOT -> 9.1.0-SNAPSHOT managed

[...]

PR #16147, opened 2025-03-12
smoke: Use snapshots for all test scenarios

ericywl:
## Motivation/summary

Related to https://github.com/elastic/apm-server/issues/15844.
 - Update smoke tests to use not-yet released SNAPSHOT versions
 - Update 7.17 basic upgrade test to test `7.17.29-SNAPSHOT` -> `8.19.0-SNAPSHOT` instead of `7.17.28` -> `7.17.29` (have trouble deploying `7.17.28-SNAPSHOT` on ECH)

## How to test these changes

Run `smoke-test-ess` with this branch: https://github.com/elastic/apm-server/actions/runs/13828755582 (all passing except 7.x -> 9.x test which is not allowed).


marclop:
I don't have anything against this but we should be mindful that we won't be testing against the exact artefact that we release to customers, since there may have been additional unreleased commits post the "public" release. If we're fine with that we should merge this since I still think there are tangible benefits to always testing with SNAPSHOT.

ericywl:
> I don't have anything against this but we should be mindful that we won't be testing against the exact artefact that we release to customers, since there may have been additional unreleased commits post the "public" release. If we're fine with that we should merge this since I still think there are tangible benefits to always testing with SNAPSHOT.

Noted, maybe I should put this in draft for now. Still pending @simitt to clarify some stuff about this issue: https://github.com/elastic/apm-server/issues/15844#issuecomment-2716337877.

ericywl:
Closed in favor of functionaltests replacing this.

Issue #16178, opened 2025-03-13
TestOTLPRateLimit occasionally failing in systemtest


ericywl:
TestOTLPRateLimit occasionally failing in `systemtest`.

```
=== RUN   TestOTLPRateLimit
    otlp_test.go:480: 
        	Error Trace:	/home/runner/work/apm-server/apm-server/systemtest/otlp_test.go:480
        	Error:      	Not equal: 
        	            	expected: 0x8
        	            	actual  : 0x4
        	Test:       	TestOTLPRateLimit
    otlp_test.go:481: 
        	Error Trace:	/home/runner/work/apm-server/apm-server/systemtest/otlp_test.go:481
        	Error:      	Not equal: 
        	            	expected: "traces export: rpc error: code = ResourceExhausted desc = rate limit exceeded"
        	            	actual  : "traces export: rpc error: code = DeadlineExceeded desc = context deadline exceeded"
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1 +1 @@
        	            	-traces export: rpc error: code = ResourceExhausted desc = rate limit exceeded
        	            	+traces export: rpc error: code = DeadlineExceeded desc = context deadline exceeded
        	Test:       	TestOTLPRateLimit
    server.go:133: log file: /home/runner/work/apm-server/apm-server/systemtest/logs/TestOTLPRateLimit/apm-server
```

ericywl:
I can't seem to reproduce it locally using `go test -run=TestOTLPRateLimit -race -v -count=100 -failfast ./...`. It might be something specific to the CI environment.

ericywl:
APM Server logs show:
```
fatal error: concurrent map writes
```

ericywl:
The issue comes from https://github.com/elastic/apm-server/blob/main/internal/beater/interceptors/metrics.go#L100-L119, concurrent map access on `counters` and `histograms`. 😖

PR #16182, opened 2025-03-13
fix: metrics interceptor concurrent map read write

ericywl:
## Motivation/summary

Closes https://github.com/elastic/apm-server/issues/16178.

## How to test these changes

Run `go test -run=TestMetrics_ConcurrentMapSafe -race -v -count=10 -failfast ./...` and see that there are no fatal errors.

### DIFF
--- a/internal/beater/interceptors/metrics.go
+++ b/internal/beater/interceptors/metrics.go
@@ -19,6 +19,7 @@ package interceptors
 
 import (
 	"context"
+	"sync"
 	"time"
 
 	"go.opentelemetry.io/otel/metric"
@@ -38,8 +39,11 @@ type metricsInterceptor struct {
 	logger *logp.Logger
 	meter  metric.Meter
 
+	countersMu sync.Mutex
 	counters   map[string]metric.Int64Counter
-	histograms map[string]metric.Int64Histogram
+
+	histogramsMu sync.Mutex
+	histograms   map[string]metric.Int64Histogram
 }
 
 func (m *metricsInterceptor) Interceptor() grpc.UnaryServerInterceptor {
@@ -98,6 +102,9 @@ func (m *metricsInterceptor) inc(legacyMetricsPrefix string, id request.ResultID
 }
 
 func (m *metricsInterceptor) getCounter(prefix, n string) metric.Int64Counter {
+	m.countersMu.Lock()
+	defer m.countersMu.Unlock()
+
 	name := prefix + n
 	if met, ok := m.counters[name]; ok {
 		return met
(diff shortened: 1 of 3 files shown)


internal/beater/interceptors/metrics.go
```diff
@@ -98,6 +102,9 @@ func (m *metricsInterceptor) inc(legacyMetricsPrefix string, id request.ResultID
 }
 
 func (m *metricsInterceptor) getCounter(prefix, n string) metric.Int64Counter {
+	m.countersMu.Lock()
+	defer m.countersMu.Unlock()
```
ericywl, line 106: Might need to check the performance impact after this change.

carsonip (commented):
what about a sync.map?

carsonip:
> Considered it, will take a look tmr. This was just a quick fix to see if the issue still happens.

If you look at http metrics, there's a sync map usage where I fixed a similar error. That PR had a similar discussion around locking vs sync map.

Thanks for taking care of this.

rubvs (dismissed):

internal/beater/interceptors/metrics.go
```diff
@@ -99,23 +100,23 @@ func (m *metricsInterceptor) inc(legacyMetricsPrefix string, id request.ResultID
...
-		return met
+	if met, ok := m.counters.Load(name); ok {
+		return met.(metric.Int64Counter)
 	}
+
 	nm, _ := m.meter.Int64Counter(name)
-	m.counters[name] = nm
-	return nm
+	met, _ := m.counters.LoadOrStore(name, nm)
+	return met.(metric.Int64Counter)
```
carsonip, line 109: Great attention to detail. This is correct, as opposed to my bad attempt to fix it in https://github.com/elastic/apm-server/blob/b62fd62bd5f6249f7c06eadbf7fc89f2edc902d0/internal/beater/middleware/monitoring_middleware.go#L76 .

carsonip (commented):
Looks good.
Do you mind expanding the scope of this PR (update subject and description as well) to fix an undetectable race i introduced in https://github.com/elastic/apm-server/blob/b62fd62bd5f6249f7c06eadbf7fc89f2edc902d0/internal/beater/middleware/monitoring_middleware.go#L76 ?

internal/beater/interceptors/metrics_test.go
```diff
@@ -176,3 +179,48 @@ func TestMetrics(t *testing.T) {
...
+	logger := logp.NewLogger("interceptor.metrics.test")
+	interceptor := Metrics(logger, mp)
+
+	ctx := context.Background()
+	info := &grpc.UnaryServerInfo{
+		FullMethod: "/opentelemetry.proto.collector.trace.v1.TraceService/Export",
+	}
+
+	waitAndDoNothing := func(ctx context.Context, req interface{}) (interface{}, error) {
+		time.Sleep(10 * time.Millisecond)
```
carsonip, line 199: is this sleep necessary?

carsonip (commented):
Also, do you mind figuring out the correct backport? imo this can be in 9.0.1 but it also means that we'll have to remind ourselves to backport this only after 9.0.0 release

internal/beater/interceptors/metrics_test.go
ericywl: Now that I think about, not necessary

carsonip (approved):
code change lgtm. Do you mind figuring out the backports?

Issue #16180, opened 2025-03-13
smoketest: legacy upgrade to 9.x fails


ericywl:
Running the 7.x -> 8.x -> 9.x legacy upgrade test results in failure to upgrade to `9.1.0-SNAPSHOT`.
```
make smoketest/run-version TEST_DIR=./testing/smoke/legacy-standalone-major-managed SMOKETEST_VERSION=latest
```

The error logs in ECH:
```
[...]

Issue #16199, opened 2025-03-14
functionaltests: 7.x standalone -> 8.x standalone -> 9.x standalone -> 9.x managed


ericywl:
As per https://github.com/elastic/apm-server/issues/16180, we need to reindex the data streams and indices before upgrading to `9.0+` version. The new steps would then be:

1. Create 7.x standalone, ingest and check
2. Upgrade to 8.x standalone, ingest and check
3. Query deprecation warnings from ES, and perform reindexing / block writing to indices
4. Upgrade to 9.x standalone, ingest and check
5. Upgrade to 9.x managed, ingest and check

endorama:
Q: is not clear to me why step 3 (Query deprecation warnings from ES, and perform reindexing / block writing to indices) only applies to standalone or it only applies to 9.1.0.

endorama:
Additional question: as we are reacting to ES migration deprecation logs, should we consider including this in our tests? For example if there are unaddressed deprecations, we may need to act on them either to make the test work or because they affect APM.

PR #16202, opened 2025-03-14
smoke: Temporarily disable failing 9.0 smoke test upgrade in legacy major

ericywl:
## Motivation/summary

Closes https://github.com/elastic/apm-server/issues/16180.

Unfortunately, tests for legacy 9.x are failing and it's unlikely we will be able to fix them using the current smoke tests. Disabling the upgrade for now. Instead of 7.x -> 8.x -> 9.x, we will simply do 7.x -> 8.x for now.

We will follow up with https://github.com/elastic/apm-server/issues/16199, by using functional tests to perform the test instead.

### DIFF
--- a/testing/smoke/legacy-standalone-major-managed/test.sh
+++ b/testing/smoke/legacy-standalone-major-managed/test.sh
@@ -60,20 +60,22 @@ healthcheck 1
 send_events
 data_stream_assertions "${ASSERTION_VERSION_8}"
 
-# Version 9 (if exists)
 MANAGED_VERSION="${LATEST_VERSION_8}"
 ASSERTION_MANAGED_VERSION="${ASSERTION_VERSION_8}"
-if [[ -n ${LATEST_VERSION_9} ]]; then
-    cleanup_tfvar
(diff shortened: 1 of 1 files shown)


simitt (changes requested):
Please always ensure to point to a github card with follow up actions and more details when disabling test logic. 
Right now, it would be very hard to track any follow up work and understand whether this is a test failure or actual logic is broken.

rubvs:
I don't like code commented. It leaves things in the grey-zone. I'd rather remove it, and create an issue stating we should revisit this.

ericywl:
> Please always ensure to point to a github card with follow up actions and more details when disabling test logic.
Right now, it would be very hard to track any follow up work and understand whether this is a test failure or actual logic is broken.

Added the follow up, which is to test using functional tests instead, since we will need to query ES and perform reindexing on some data streams / indices, prior to upgrade.

> I don't like code commented. It leaves things in the grey-zone. I'd rather remove it, and create an issue stating we should revisit this.

Makes sense, especially given that we aren't likely to even re-enable this exact same test.

simitt (approved):

PR #16203, opened 2025-03-14
functionaltests: Add stack versions fetching

### LINKED ISSUE #16200: functionaltests: Add ability to fetch stack versions
Current functional tests use hard coded `fromVersion` and `toVersion` for running the tests. This is great for tests that require some additional setup, but for basic upgrade tests, it would be better to automate fetching of latest stack versions.

The first step is to add the ability to fetch stack versions. We can do so by emulating the way that smoke test does it currently: https://github.com/elastic/apm-server/blob/main/testing/smoke/lib.sh.

ericywl:
## Motivation/summary

Resolves #16200.

- Add `StackVersion` and `StackVersions` to more easily manage stack versions.
- Adds methods in Elastic Cloud `ecclient.Client` to fetch stack versions and snapshots.
- Add unit tests for the stack versions type and EC client.

## How to test these changes

There are unit tests for the stack versions logic and EC client. I have also tested the EC client and it works locally on `qa` and `pro` environments.


endorama (commented):
I minor comment but looks good!

To overcome having to rely on Elastic Cloud for testing you could leverage [`go-vcr`](https://github.com/dnaeon/go-vcr) which we are already using in `kbclient`: https://github.com/elastic/apm-server/blob/main/functionaltests/internal/kbclient/client_test.go#L42

ericywl:
Hmmm, QA API does not seem to have the stack versions. I get:
```
response status code does not match any response statuses defined for this endpoint in the swagger spec (status 401): {}
```

ericywl:
Seems like the functional tests in QA are failing for various reasons...

endorama (approved):
Looks good thank you. Left a comment but please do not address it as it's a minor now so we can merge and move on.

PR #16273, opened 2025-03-18
functionaltests: Rename ecc to esc to make way for ecclient

ericywl:
## Motivation/summary

Currently, the Elasticsearch client `esclient.Client` is named `ecc` in its usage. This conflicts with Elastic Cloud `ecclient.Client`, that we are going to use soon.


endorama (approved):
FIPS test are failing, there is a panic timeout on `TestNewInstrumentation` which is not related to this PR. Approving to unblock.

Issue #16274, opened 2025-03-18
functionaltests: Use stack version fetching for tests


ericywl:
Follow up from #16200. 

Use the `GetVersions` and `GetSnapshotVersions` in our tests to test using the latest patch.

PR #16295, opened 2025-03-19
functionaltests: Use stack version fetching for tests

ericywl:
## Motivation/summary

Resolves https://github.com/elastic/apm-server/issues/16274.

This PR uses the version fetching added in https://github.com/elastic/apm-server/pull/16203 to automatically run the functional tests using latest patch version.

The rough idea is that we will [always use latest SNAPSHOT patch for testing `X.Y` versions](https://github.com/elastic/apm-server/issues/15844), and that we have a list of active `X.Y` versions so that we can skip testing non-active versions in the CI.

## How to test these changes

Run functionaltests with this branch: https://github.com/elastic/apm-server/actions/runs/13965737490.


functionaltests/utils_test.go
```diff
@@ -32,18 +33,15 @@ import (
...
-// Functional tests are expected to run Terraform code to operate
-// on infrastructure required for each test and to query Elastic
-// Cloud APIs. In both cases a valid API key is required.
-func ecAPICheck(t *testing.T) {
+// getLatestSnapshotFor retrieves the latest snapshot version for the version prefix.
+func getLatestSnapshotFor(t *testing.T, prefix string) string {
 	t.Helper()
-	require.NotEmpty(t, os.Getenv("EC_API_KEY"), "EC_API_KEY env var not set")
+	version, ok := fetchedSnapshots.LatestFor(prefix)
+	require.True(t, ok, "no version with prefix '%s' found", prefix)
```
endorama, line 40: suggestion: could you clarify in the message where this version is missing from? (My understanding is that versions are region specific). Es :

```suggestion
	require.True(t, ok, "no version with prefix '%s' found in Elastic Cloud region %s", prefix, regionFrom(*environment))
```

functionaltests/README.md
```diff
@@ -6,21 +6,58 @@ The functional tests test that APM Server works as expected after version upgrad
...
+```
+
+Note: Before running tests, make sure to delete the Terraforms by running `rm -r tf-*`.
+
+### Debugging the Tests
+
+If you get some errors after running the test, you can try heading to the [Elastic Cloud console](https://cloud.elastic.co/home)
+in order to access the Kibana instance. From there, you can use Dev Tools to check the data streams etc.
+
+Note: If the tests failed due to deployment, you may need to access the Elastic Cloud admin console instead to check the
```
endorama, line 29: Do you have a specific case in mind that requires the Admin Console?

endorama (changes requested):
Two minor comments but this is the direction I was expecting we would go!

functionaltests/README.md
ericywl: For Terraform deployment failures, I usually go to the admin console to check the deployment error logs.

functionaltests/utils_test.go
ericywl: Done in [9c3bab4](https://github.com/elastic/apm-server/pull/16295/commits/9c3bab43bb6ad82388293f8fae8a41a54054d590).

endorama (approved):

PR #16304, opened 2025-03-19
[docs] Copy recent release note edits

colleenmcginnis:
Copies over recent release note edits from https://github.com/elastic/docs-content/pull/827 (which snuck in before I could merge https://github.com/elastic/docs-content/pull/820).

cc @KOTungseth


inge4pres (approved):
Thanks Colleen 🙏🏼

PR #16341, opened 2025-03-21
functionaltests: Add BC tests

ericywl:
## Motivation/summary

This PR introduces functional tests for build candidate (BC) versions. The previous idea of using a single `TestUpgradeBC` test to run tests for all BC versions does not work, as each version may have their own caveats, e.g. DSL vs ILM, lazy rollover, different ignored APM error logs etc.

Note: `TestUpgrade_8_14_to_8_16_Reroute` is not changed since subsequent PR to introduce reroute tests for all versions will replace it.

## How to test these changes

Run functionaltests workflow with this branch: https://github.com/elastic/apm-server/actions/runs/13989686769.

### DIFF
--- a/functionaltests/main_test.go
+++ b/functionaltests/main_test.go
@@ -65,13 +65,13 @@ const (
 	managedByILM = "Index Lifecycle Management"
 )
 
-const (
-	defaultNamespace = "default"
-)
-
 var (
+	// fetchedCandidates are the build-candidate stack versions prefetched from Elastic Cloud API.
+	fetchedCandidates ecclient.StackVersions
 	// fetchedSnapshots are the snapshot stack versions prefetched from Elastic Cloud API.
 	fetchedSnapshots ecclient.StackVersions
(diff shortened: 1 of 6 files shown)


functionaltests/main_test.go
```diff
@@ -108,22 +108,34 @@ func TestMain(m *testing.M) {
...
+	}
+	fetchedCandidates = candidates
+
 	snapshots, err := ecc.GetSnapshotVersions(ctx, ecRegion)
 	if err != nil {
 		log.Fatal(err)
 		return
 	}
 	fetchedSnapshots = snapshots
-	fetchedSnapshots.Sort()
```
endorama, line 117: question(non-blocking): why are we not sorting anymore here?

endorama (approved):

functionaltests/main_test.go
ericywl: It's already sorted when we call `GetSnapshotVersions`

PR #16359, opened 2025-03-22
functionaltests: Move assertDocCount and assertDatastreams to asserts package

ericywl:
## Motivation/summary

- Move the assertion functions to `asserts` package
- ~~Fix post-upgrade document count assertion, previously it was checking before-ingestion count vs after-upgrade count, it should be before-upgrade vs after-upgrade~~ https://github.com/elastic/apm-server/pull/16402
- Rename some stuff to be more consistent

## How to test these changes

Run functionaltests with this branch: https://github.com/elastic/apm-server/actions/runs/14006903494.


ericywl:
@endorama This is an experimental PR, I'm actually not 100% convinced whether the change is better. Maybe moving the assertions to the `asserts` pkg is ok (considering that the assertions themselves are technically data stream agnostic, not APM specific), but the change to asserting via maps idk.

Curious to hear your thoughts about this.

functionaltests/single_upgrade_test.go
```diff
@@ -133,29 +134,33 @@ func (tt singleUpgradeTestCase) Run(t *testing.T) {
 	// and further assertions.
 	// We don't expect any change here unless something broke during the upgrade.
 	t.Log("check number of documents across upgrade")
-	assertDocCount(t, beforeUpgradeCount, esclient.APMDataStreamsDocCount{},
-		atStartCount, aggregationDataStreams(tt.dataStreamNamespace))
+	afterUpgradeCount, err := getDocCountPerDS(t, ctx, esc)
+	require.NoError(t, err)
+	asserts.CheckDocCount(t, afterUpgradeCount, beforeUpgradeCount,
+		emptyIngestForASingleRun(tt.dataStreamNamespace),
+		aggregationDataStreams(tt.dataStreamNamespace))
```
ericywl, line 141: > Fix post-upgrade document count assertion, previously it was not checking the right thing

Note: There is a fix inside this PR. So even if we close this, it will still be worth extracting this change out.

functionaltests/main_test.go
```diff
@@ -183,77 +182,12 @@ func allDataStreams(namespace string) []string {
 	return res
 }
 
-// getDocsCountPerDS retrieves document count.
-func getDocsCountPerDS(t *testing.T, ctx context.Context, esc *esclient.Client) (esclient.DataStreamsDocCount, error) {
+// getDocCountPerDS retrieves document count per data stream.
+func getDocCountPerDS(t *testing.T, ctx context.Context, esc *esclient.Client) (esclient.DataStreamsDocCount, error) {
```
endorama, line 186: We should probably remove this helper function, what do you think? Does it add value?

endorama:
100% on board with moving the asserts function to the package. They predate the creation of that package, that's the only reason they're in the main one.

On the map assertion I have 2 concerns:
- the output is more terse but not more clear than before: differences are clearer but provide less guidance and is harder to identify exactly what needs to be fixed
- the code for having the new output is way more complex.
Overall does not seem a good trade off in my opinion, we should investigate other direction (es rethink the `CheckDataStreamWant` struct to, for example, use a map of data stream names).

I would extract the `CheckDataStream` refactor from this PR, other changes are good!

ericywl:
> I would extract the CheckDataStream refactor from this PR, other changes are good!

Reverted in [b01faff](https://github.com/elastic/apm-server/pull/16359/commits/b01fafff2ddbfb3fe98f66af5db56533671ea3dc)!

functionaltests/main_test.go
endorama: Great idea moving the `NoError` check in this function 🚀

endorama (approved):
looks good!

PR #16174, opened 2025-03-12
test: add basic job for fips system-tests

kruskall:
## Motivation/summary

a lot of tests are failing but it's a start

add CI job for running system-tests in fips mode with fips binary

## Checklist

- [ ] Update [CHANGELOG.asciidoc](https://github.com/elastic/apm-server/blob/main/CHANGELOG.asciidoc)
- [ ] Documentation has been updated

For functional changes, consider:
- Is it observable through the addition of either **logging** or **metrics**?
- Is its use being published in **telemetry** to enable product improvement?
- Have system tests been added to avoid regression?

## Related issues

Related to https://github.com/elastic/apm-server/issues/15858
Related to https://github.com/elastic/apm-server/issues/15865

Issue #15863, opened 2025-02-25
(fips): Create test ensuring that that connections from and to non-fips compliant TLS setups are rejected


simitt:
Related to https://github.com/elastic/apm-server/issues/15862, set up system tests that ensure the APM Server rejects incoming connections for TLS handshakes with non-fips-supported configurations, as well as the server not establishing outgoing connections to components configured with non compliant TLS settings.

PR #16365, opened 2025-03-24
test: refactor tls systemtest for fips compliance

kruskall:
## Motivation/summary

add tls test for default configuration
add tls test for custom configuration
make sure tests pass in fips mode

## Checklist

## Related issues

Split from https://github.com/elastic/apm-server/pull/16174
Related to https://github.com/elastic/apm-server/issues/15863


inge4pres (approved):

systemtest/tls_test.go
```diff
@@ -79,6 +75,51 @@ func TestTLSConfig(t *testing.T) {
 	})
 }
 
+func TestCustomTLSConfig(t *testing.T) {
```
simitt, line 78: Would it make sense to also add tests that ensure non-fips configuration leads to failures? This is running the same test cases for fips and non-fips, but what about the cases that are allowed in non-fips but are expected to fail in fips?

PR #16385, opened 2025-03-24
TBS: Fix edge case where storage limit is exceeded but missed by drop loop check

### LINKED ISSUE #16386: apm-server does not drop and recreate badger DB after TTL when storage threshold is met
**APM Server version** (`apm-server version`): `8.17.3` confirmed but the problematic PR was merged in `8.17.1` and `8.16.3`

Once the storage threshold is hit, the cleanup procedure added with https://github.com/elastic/apm-server/pull/15106 does not kick in and APM Server is stuck in this state until manual action is taken.

A workaround is to decrease the storage threshold to below the current amount, wait for cleanup and then increase the threshold again.

**Steps to reproduce**:

unclear

carsonip:
## Motivation/summary

https://github.com/elastic/apm-server/pull/15106 contains a bug where drop loop condition does not match rw writeEntry storage limit check. The storageLimitThreshold was missing from the drop loop condition check. This creates an edge case where rw thinks storage limit is exceeded while drop loop mitigation does not kick in.

e.g. for 1GB storage limit and 90% storage limit threshold, if the db size is 950MB, it will be seen as storage limit exceeded by read writer but not by drop loop.

Also fix an edge case where pendingSize may not be correct at the end of writeEntry call if rw.txn.SetEntry fails. It shouldn't have any major impact either way.

## Checklist

## How to test these changes

The unit test change should cover it.

To manually test: It is a delicate one. DB size growth needs to be slow, such that DB size lands within `storage limit * 0.9` and `storage limit`. Before this PR, it will log `storage limit exceeded` but no `db will be dropped and recreated`. After this PR, it will log both `storage limit exceeded` and `db will be dropped and recreated`, and the db should be dropped and recreated.

## Related issues

Fixes #16386


simitt (approved):

x-pack/apm-server/sampling/eventstorage/storage_manager.go
```diff
@@ -164,15 +164,18 @@ func (s *StorageManager) runDropLoop(stopping <-chan struct{}, ttl time.Duration
...
 	checkAndFix := func() error {
 		lsm, vlog := s.Size()
-		// add buffer to avoid edge case storageLimitInBytes-lsm-vlog < buffer, when writes are still always rejected
-		buffer := int64(baseTransactionSize * len(s.rw.readWriters))
-		if uint64(lsm+vlog+buffer) >= storageLimitInBytes {
+		s.mu.RLock() // s.storage requires mutex RLock
+		pending := s.storage.pendingSize.Load()
+		s.mu.RUnlock()
+		total := lsm + vlog + pending
+		actualLimit := int64(float64(storageLimitInBytes) * storageLimitThreshold)
```
carsonip, line 171: [to reviewer] This is the actual fix. The `*0.9` was missing. That is it.

x-pack/apm-server/sampling/eventstorage/storage_manager.go
```diff
@@ -164,15 +164,18 @@ func (s *StorageManager) runDropLoop(stopping <-chan struct{}, ttl time.Duration
 	var firstExceeded time.Time
 	checkAndFix := func() error {
 		lsm, vlog := s.Size()
-		// add buffer to avoid edge case storageLimitInBytes-lsm-vlog < buffer, when writes are still always rejected
-		buffer := int64(baseTransactionSize * len(s.rw.readWriters))
-		if uint64(lsm+vlog+buffer) >= storageLimitInBytes {
+		s.mu.RLock() // s.storage requires mutex RLock
+		pending := s.storage.pendingSize.Load()
```
carsonip, line 168: [to reviewer] This is just a nice to have to rule out any possibility that the condition is inconsistent with what's used in read writer.

x-pack/apm-server/sampling/processor_test.go
```diff
@@ -873,7 +873,10 @@ func TestDropLoop(t *testing.T) {
 
 				config.Elasticsearch = pubsubtest.Client(nil, nil) // disable pubsub
 
-				config.StorageLimit = 100 * 1024 // lower limit to trigger storage limit error
+				// lower limit to trigger storage limit error
+				// make it 5% higher than the existing size, and dropAndRecreate should kick in
+				// because there's a hardcoded 90% threshold for storage limit
+				config.StorageLimit = uint64(float64(lsm+vlog) * 1.05)
```
carsonip, line 879: [to reviewer] with this change in test, it fails before the fix and passes after.

lahsivjar (approved):

Issue #16343, opened 2025-03-21
functionaltests: Add remaining active versions


ericywl:
Add the remaining active versions basic upgrade tests. Currently, that includes adding the basic upgrade tests (SNAPSHOTs and BCs) for 
- 8.16 -> 8.17
- 8.17 -> 8.18
- 8.18 -> 8.19
- 8.19 -> 9.1
- 9.0 -> 9.1

The 7.17 -> 8.x (latest) test is excluded because we have not implemented 7.x tests yet.

PR #16394, opened 2025-03-25
functionaltests: Add 8.17, 8.18, 8.19 and 9.1 active version tests

ericywl:
## Motivation/summary

Part of https://github.com/elastic/apm-server/issues/16343. Branched from https://github.com/elastic/apm-server/pull/16440.

Adds the following basic upgrades:
 - 8.16 -> 8.17 (DSL -> ILM, lazy rollover)
 - 8.17 -> 8.18 (ILM -> ILM)
 - 8.18 -> 8.19 (ILM -> ILM, lazy rollover)
 - 8.19 -> 9.1 (ILM -> ILM)
 - 9.0 -> 9.1 (ILM -> ILM, lazy rollover)
 
 See [comment](https://github.com/elastic/apm-server/pull/16394/commits/6e7b66907b18ad02780349de399778be8c6741a4) for why there is lazy rollover in 8.19 and 9.1.

## How to test these changes

Run functionaltests with this branch: https://github.com/elastic/apm-server/actions/runs/14267592272.


functionaltests/9_1_test.go
```diff
@@ -0,0 +1,121 @@
...
+import (
+	"testing"
+
+	"github.com/elastic/go-elasticsearch/v8/typedapi/types"
+)
+
+func TestUpgrade_9_0_to_9_1_Snapshot(t *testing.T) {
+	t.Parallel()
+
+	// TODO: Check why is there lazy rollover?
```
ericywl, line 29: @endorama I'm not quite sure why there is lazy rollover here for 9.0 -> 9.1

functionaltests/8_19_test.go
```diff
@@ -0,0 +1,73 @@
...
+import (
+	"testing"
+
+	"github.com/elastic/go-elasticsearch/v8/typedapi/types"
+)
+
+func TestUpgrade_8_18_to_8_19_Snapshot(t *testing.T) {
+	t.Parallel()
+
+	// TODO: Check why is there lazy rollover?
```
ericywl, line 29: @endorama I'm not quite sure why there is lazy rollover here for 8.18 -> 8.19

functionaltests/9_1_test.go
```diff
@@ -0,0 +1,125 @@
...
+			version:         getLatestBCOrSkip(t, "9.1"),
+			preferILM:       true,
+			indexManagement: managedByILM,
+		},
+		[]types.Query{
+			tlsHandshakeError,
+			esReturnedUnknown503,
+			refreshCache503,
+			refreshCache403,
+			refreshCacheInvalidESConfig,
```
ericywl, line 73: @endorama Have you seen these errors before?

functionaltests/apm_logs_filters.go
```diff
@@ -28,24 +28,26 @@ var (
...
 		MatchPhrase: map[string]types.MatchPhraseQuery{
 			"message": {Query: "ES returned unknown status code: 503 Service Unavailable"},
 		},
 	}
-	refreshCache503 = types.Query{
+
+	refreshCache403 = types.Query{
 		MatchPhrase: map[string]types.MatchPhraseQuery{
-			"message": {Query: "refresh cache elasticsearch returned status 503"},
+			"message": {Query: "refresh cache elasticsearch returned status 403"},
```
endorama, line 40: I don't think we should ignore the `403` as it indicates a permission issue, which in unexpected and should be investigated.

functionaltests/basic_upgrade_test.go
```diff
@@ -28,89 +28,85 @@ import (
 	"github.com/elastic/apm-server/functionaltests/internal/ecclient"
 )
 
-// runBasicUpgradeILMTest performs a basic upgrade test from `fromVersion` to
-// `toVersion`. The test assumes that all data streams are using Index
-// Lifecycle Management (ILM) instead of Data Stream Lifecycle Management
-// (DSL), which should be the case for most recent APM data streams.
-func runBasicUpgradeILMTest(
+type basicUpgradeVersionConfig struct {
```
endorama, line 31: I'm not fully convinced by this approach. What are the benefits (other that removing some duplication) of this approach compared to using `singleUpgradeTestCase` directly?

ericywl: Yeah... After this change it doesn't make as much sense anymore. I'm thinking to revert to just have a `runBasicUpgradeILMTest` for the most common case of all ILM no rollover, and the rest would just be using `singleUpgradeTestCase`.

ericywl: Updated in https://github.com/elastic/apm-server/pull/16394/commits/2051dcf6cfe450d8f354b1aa2e619c098fd922c6.

ericywl:
@endorama I removed 8.9 and 9.1 test for now to unblock the PR, since I'm not sure why there is lazy rollover. I will create a follow up PR to add those tests once investigation is done instead.

functionaltests/9_1_test.go
ericywl: https://github.com/elastic/apm-server/pull/16394/commits/6e7b66907b18ad02780349de399778be8c6741a4

functionaltests/8_19_test.go
ericywl: Updated in https://github.com/elastic/apm-server/pull/16394/commits/6e7b66907b18ad02780349de399778be8c6741a4

functionaltests/basic_upgrade_test.go
endorama: I have an idea on how to refactor this. So far it didn't seem that useful but it starting to look more reasonable as we progress.

Instead of having the "test" as a single unit, introduce the concept of "steps". A test is composed of multiple steps, run in sequence and interleaved with checks. Any step can reference the previous ones, but in theory not the next ones (how can this be done, is this necessary?). The test runner would run each step with a pre/run/post phases + our current checks.
The additional complexity is what made me avoid this route so far, but for tests with 3 or more steps this design would be quite effective IMO. It would also open the door for "custom steps" where custom logic could be introduced.

What do you think? This has the risk of being a rabbit hole and should not be addressed in this PR.

ericywl: Sounds interesting 🤔 I would have to experiment with it a bit to see how it would look like.

endorama (approved):

PR #16395, opened 2025-03-25
functionaltests: Use matrix to split test runs

ericywl:
## Motivation/summary

As we add more and more tests, I believe it's helpful to split the test runs, so that there aren't too many tests in one job (tests taking longer to run, makes checking for error logs harder etc.). 

In this PR, the active versions is moved to the workflow as a matrix instead. The test runs will then be split by version.

## How to test these changes

Run functionaltests with this branch: https://github.com/elastic/apm-server/actions/runs/14052852585.

Also tested using this PR: https://github.com/elastic/apm-server/pull/16394, and managed to cut down run time from 1h+ to 20min


endorama (approved):
thanks! Good speedup!

Issue #16339, opened 2025-03-20
functionaltests: support standalone mode


endorama:
APM Standalone mode needs to be supported for 7.x and 8.x clusters, with upgrade paths in standalone mode and before and after migrating to managed mode.

This will also require implementing the "standalone to managed" migration functionality.

ericywl:
I should be able to take this for this iteration. I assume we are just implementing it similar to the current smoke test?

PR #16418, opened 2025-03-26
functionaltests: Add integrations server option and EnableIntegrationsServer API

ericywl:
## Motivation/summary

Closes https://github.com/elastic/apm-server/issues/16339.

- Add `integrations_server` TF var to enable standalone vs managed deployment
- Add `EnableIntegrationsServer` API in Kibana client to migrate from standalone to managed
- Refactor ES config and Kibana client, to pass ES superuser username and password to Kibana (new API requires superuser privilege)

## How to test these changes

Run this simple test locally with this branch:

```go
func TestStandalone_7_17(t *testing.T) {
	ctx := context.Background()
	copyTerraforms(t)
	tf, err := terraform.New(t, terraformDir(t))
	require.NoError(t, err)

	version := getLatestSnapshot(t, "7.17")
	deployOut := createCluster(t, ctx, tf, *target, version.String(), false)

	esc := createESClient(t, deployOut)
	kbc := createKibanaClient(t, ctx, esc, deployOut)

	err = kbc.EnableIntegrationsServer(ctx)
	require.NoError(t, err)
}
```

Also run functionaltests workflow with this branch, to see if the refactoring broke anything: https://github.com/elastic/apm-server/actions/runs/14077966600.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -10,7 +10,7 @@ x-logging: &default-logging
     max-size: "1g"
 services:
   elasticsearch:
-    image: docker.elastic.co/elasticsearch/elasticsearch:9.1.0-15b6399f-SNAPSHOT
+    image: docker.elastic.co/elasticsearch/elasticsearch:9.1.0-bbb74424-SNAPSHOT
     ports:
       - 9200:9200
     healthcheck:
@@ -41,7 +41,7 @@ services:
     logging: *default-logging
 
   kibana:
-    image: docker.elastic.co/kibana/kibana:9.1.0-15b6399f-SNAPSHOT
+    image: docker.elastic.co/kibana/kibana:9.1.0-bbb74424-SNAPSHOT
     ports:
       - 5601:5601
     healthcheck:
@@ -60,7 +60,7 @@ services:
     logging: *default-logging
 
   metricbeat:
-    image: docker.elastic.co/beats/metricbeat:9.1.0-15b6399f-SNAPSHOT
+    image: docker.elastic.co/beats/metricbeat:9.1.0-bbb74424-SNAPSHOT
     environment:
       ELASTICSEARCH_HOSTS: '["http://elasticsearch:9200"]'
       ELASTICSEARCH_USERNAME: "${KIBANA_ES_USER:-admin}"
--- a/testing/infra/k8s/base/stack/apm-server.yaml
+++ b/testing/infra/k8s/base/stack/apm-server.yaml
@@ -4,7 +4,7 @@ metadata:
   name: apm-server
 spec:
   # TODO: replace with a pinned version such as 9.0.0-aaaaaaaa-SNAPSHOT
-  version: 9.1.0-15b6399f-SNAPSHOT
+  version: 9.1.0-bbb74424-SNAPSHOT
   count: 1
   http:
     tls:
--- a/testing/infra/k8s/base/stack/elasticsearch.yaml
+++ b/testing/infra/k8s/base/stack/elasticsearch.yaml
@@ -4,7 +4,7 @@ metadata:
   name: elasticsearch
 spec:
   # TODO: replace with a pinned version such as 9.0.0-aaaaaaaa-SNAPSHOT
-  version: 9.1.0-15b6399f-SNAPSHOT
+  version: 9.1.0-bbb74424-SNAPSHOT
   auth:
     fileRealm:
       - secretName: elasticsearch-admin
--- a/testing/infra/k8s/base/stack/kibana.yaml
+++ b/testing/infra/k8s/base/stack/kibana.yaml
@@ -4,7 +4,7 @@ metadata:
   name: kibana
 spec:
   # TODO: replace with a pinned version such as 9.0.0-aaaaaaaa-SNAPSHOT
-  version: 9.1.0-15b6399f-SNAPSHOT
+  version: 9.1.0-bbb74424-SNAPSHOT
   count: 1
   elasticsearchRef:
     name: elasticsearch


functionaltests/internal/kbclient/client.go
```diff
@@ -67,14 +76,63 @@ func prepareRequest(c *Client, method, path string, body any) (*http.Request, er
...
+	req = req.WithContext(ctx)
+	resp, err := c.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("cannot perform http request: %w", err)
+	}
+	defer resp.Body.Close()
+
+	b, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return nil, fmt.Errorf("cannot read response body: %w", err)
```
endorama, line 118:
```suggestion
		return nil, fmt.Errorf("cannot read response body for  %s %s: %w", method, path, err)
```

A minor suggestion to help locate this error if it happens. We may be reading the body in different places and the body may not contain enough information to pinpoint where the issue is happening.

functionaltests/utils_test.go
```diff
@@ -59,12 +60,36 @@ func copyTerraforms(t *testing.T) {
 	require.NoError(t, err)
 }
 
+type deploymentOutput struct {
```
endorama, line 63: maybe 
```suggestion
type deploymentInfo struct {
```
would be a better name? Output is coupled with Terraform but this does not represent a generic "output"

functionaltests/utils_test.go
```diff
@@ -59,12 +60,36 @@ func copyTerraforms(t *testing.T) {
...
+	// ElasticsearchURL holds the Elasticsearch URL.
+	ElasticsearchURL string
+
+	// Username holds the Elasticsearch username for basic auth.
+	Username string
+
+	// Password holds the Elasticsearch password for basic auth.
+	Password string
```
endorama, line 71: This is true but not complete. `Username` and `Password` hold the superuser credentials generated by the cluster creation. As you correctly mentioned these are necessary for some API calls. I think it's worth to mention this.

```suggestion
	// Username holds the Elasticsearch super user username for basic auth.
	Username string

	// Password holds the Elasticsearch super user password for basic auth.
	Password string
```

functionaltests/utils_test.go
```diff
@@ -90,15 +116,19 @@ func createCluster(t *testing.T, ctx context.Context, tf *terraform.Runner, targ
...
+	require.NoError(t, tf.Output("es_url", &output.ElasticsearchURL))
+	require.NoError(t, tf.Output("username", &output.Username))
+	require.NoError(t, tf.Output("password", &output.Password))
+	require.NoError(t, tf.Output("kb_url", &output.KibanaURL))
+
+	standaloneOrManaged := "standalone"
+	if enableIntegrations {
+		standaloneOrManaged = "managed"
+	}
+	t.Logf("created %s deployment %s with APM (%s)", standaloneOrManaged, deploymentID, apmID)
```
endorama, line 130: This is not a "standalone deployment" but a "deployment with APM standalone"

```suggestion
	t.Logf("created deployment %s with %s APM (%s)", deploymentID, standaloneOrManaged, apmID)
```

functionaltests/single_upgrade_test.go
```diff
@@ -73,19 +75,12 @@ func (tt singleUpgradeTestCase) Run(t *testing.T) {
 	require.NoError(t, err)
 
 	t.Log("------ cluster setup ------")
-	_, esCfg := createCluster(t, ctx, tf, *target, tt.fromVersion.String())
+	deployOut := createCluster(t, ctx, tf, *target, tt.fromVersion.String(), !tt.deployStandalone)
```
endorama, line 78: I'm confused as of why `!tt.deployStandalone` is needed here. Is this because we want the default to be "deploy managed"? If yes I'd suggest to use that and avoid the negation.

functionaltests/single_upgrade_test.go
```diff
@@ -50,6 +47,9 @@ type additionalFunc func(t *testing.T, ctx context.Context, esc *esclient.Client
 type singleUpgradeTestCase struct {
 	fromVersion ecclient.StackVersion
 	toVersion   ecclient.StackVersion
+	// deployStandalone determines whether to deploy the cluster in
+	// standalone mode as opposed to managed mode (default)
+	deployStandalone bool
```
endorama, line 52: Small naming suggestion to clarify that we are referring to the APM component and not the entire deployment:
```suggestion
	APMStandalone bool
```

See also the comment on the negation, maybe this should be `APMManaged` instead.

functionaltests/internal/esclient/client.go
```diff
@@ -39,16 +39,15 @@ type Client struct {
 	es *elasticsearch.TypedClient
 }
 
-// New returns a new Client for querying APM data.
-func New(cfg Config) (*Client, error) {
+// New returns a new Client for accessing Elasticsearch.
+func New(esURL, username, password string) (*Client, error) {
 	transport := http.DefaultTransport.(*http.Transport).Clone()
-	transport.TLSClientConfig = &tls.Config{InsecureSkipVerify: cfg.TLSSkipVerify}
+	transport.TLSClientConfig = &tls.Config{InsecureSkipVerify: false}
```
endorama, line 45: Do we need to keep the cloned transport if we don't allow modifying this value? I'm not sure if there is a use case for configuring `InsecureSkipVerify`. If there isn't I would remove this.

endorama (commented):
Left some comments, mostly cosmetic. What I'm not sure about is if the ingestion would work, but I don't think so. Can you confirm you were able to ingest data in a APM standalone scenario?

ericywl:
> What I'm not sure about is if the ingestion would work, but I don't think so. Can you confirm you were able to ingest data in a APM standalone scenario?

Let me try if I can deploy 8.x standalone and try ingesting.

functionaltests/utils_test.go
ericywl: Renamed in [11e29e3](https://github.com/elastic/apm-server/pull/16418/commits/11e29e3c9cfea6811a8491c2ef6bcd1d968a9e7b).

functionaltests/single_upgrade_test.go
ericywl: Introduced `apmDeploymentMode` in [11e29e3](https://github.com/elastic/apm-server/pull/16418/commits/11e29e3c9cfea6811a8491c2ef6bcd1d968a9e7b).

ericywl:
> > What I'm not sure about is if the ingestion would work, but I don't think so. Can you confirm you were able to ingest data in a APM standalone scenario?
> 
> Let me try if I can deploy 8.x standalone and try ingesting.

Here's what worked:
1. Deploy 7.17 with standalone
2. Upgrade to 8.x
3. Migrate to managed (enable integrations server)
4. Ingest and check doc count

With our current code, we cannot ingest to 7.17 (generator doesn't support) and we also can't properly query ES (we are using v8 ES client).

endorama:
@ericywl let's split these 2 problems.

> we cannot ingest to 7.17 (generator doesn't support)

For the generator supporting ingestion in 7.x we need https://github.com/elastic/apm-perf/pull/241. I'll send another PR to update the generator to support this use case.

> we also can't properly query ES v7 (we are using ES v8 client).

The cause is not the client version but the fact that `APMDocCount` is using EQ|QL, which didn't exist in v7. I have an implementation ready to tackle this, I'll create the PR.

ericywl:
> For the generator supporting ingestion in 7.x we need https://github.com/elastic/apm-perf/pull/241. I'll send another PR to update the generator to support this use case.

> The cause is not the client version but the fact that `APMDocCount` is using EQ|QL, which didn't exist in v7. I have an implementation ready to tackle this, I'll create the PR.

In that case, should we get this PR merged first so that you can rebase yours?

endorama (dismissed):

endorama:
@ericywl let's move on with this, I'll add my changes on top of it

ericywl:
> @ericywl let's move on with this, I'll add my changes on top of it

@endorama Sure! I had to resolve some conflicts from the merge from main. Please approve again.

endorama (approved):

PR #16433, opened 2025-03-26
feat(fips): package fips artifacts

kruskall:
## Motivation/summary

add linux amd64 and arm64
add apm-server-fips docker image

## Checklist

## How to test these changes

- `GENERATE_FIPS_ARTIFACTS=true` make package
or
- `make build/apm-server-fips-linux-amd64`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -143,7 +143,7 @@ jobs:
       - uses: actions/checkout@v4
       - name: Get changed files
         id: changed-files
-        uses: tj-actions/changed-files@dcc7a0cba800f454d79fff4b993e8c3555bcc0a8 # v45.0.7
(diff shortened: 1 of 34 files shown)


.github/workflows/ci.yml
```diff
@@ -105,6 +105,13 @@ jobs:
           cache: false
           cache-dependency-path: |
             go.sum
+      # necessary for the fips base image
+      - name: Log in to the Elastic Container registry
+        uses: docker/login-action@74a5d142397b4f367a81961eba4e8cd7edddf772 # v3.4.0
+        with:
```
v1v, line 111: Unfortunately, this won't work on forked PRs, GH secrets are inaccessible in that case.

The `test-package-and-push:` job is the one that can access the GH secrets. Is it possible to skip the FIPS when running `make package-snapshot`?

endorama (approved):
approving to unblock but is not clear to me how to verify that this works as expected.

PR #16440, opened 2025-03-27
functionaltests: Experiment with steps

ericywl:
## Motivation/summary

TODO

## How to test these changes

TODO

### DIFF
--- a/functionaltests/steps_test.go
+++ b/functionaltests/steps_test.go
@@ -1,0 +1,145 @@
+package functionaltests
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/elastic/go-elasticsearch/v8/typedapi/types"
+	"github.com/stretchr/testify/require"
+
+	"github.com/elastic/apm-server/functionaltests/internal/asserts"
+	"github.com/elastic/apm-server/functionaltests/internal/ecclient"
+	"github.com/elastic/apm-server/functionaltests/internal/esclient"
+	"github.com/elastic/apm-server/functionaltests/internal/gen"
+	"github.com/elastic/apm-server/functionaltests/internal/kbclient"
+	"github.com/elastic/apm-server/functionaltests/internal/terraform"
+)
+
+func runTestSteps(t *testing.T, namespace string, startVersion ecclient.StackVersion, steps []testStep) {
+	if namespace == "" {
+		namespace = "default"
+	}
+
+	start := time.Now()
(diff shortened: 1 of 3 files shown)


functionaltests/steps_test.go
```diff
@@ -0,0 +1,172 @@
...
+	"github.com/elastic/apm-server/functionaltests/internal/asserts"
+	"github.com/elastic/apm-server/functionaltests/internal/ecclient"
+	"github.com/elastic/apm-server/functionaltests/internal/esclient"
+	"github.com/elastic/apm-server/functionaltests/internal/gen"
+	"github.com/elastic/apm-server/functionaltests/internal/kbclient"
+	"github.com/elastic/apm-server/functionaltests/internal/terraform"
+)
+
+type testStepsRunner struct {
+	namespace    string
```
endorama, line 38:
```suggestion
	dataStreamNamespace    string
```

functionaltests/steps_test.go
```diff
@@ -0,0 +1,172 @@
...
+	startVersion ecclient.StackVersion
```
endorama, line 39: What do you think of having a `CreateStep`? I don't think we have an use today but maybe there may be the need to perform some operation before create, and it'd fit nicely with the overall pattern instead of having a special case.

ericywl: sounds good

functionaltests/steps_test.go
```diff
@@ -35,65 +35,63 @@ import (
...
+	e.kbc = createKibanaClient(t, ctx, e.esc, esCfg)
+	e.gen = createAPMGenerator(t, ctx, e.esc, esCfg)
+
+	docCount := getDocsCountPerDS(t, ctx, e.esc)
+	return testStepResult{DSDocCount: docCount}
+}
+
+var _ testStep = ingestionStep{}
+
 type ingestionStep struct {
```
endorama, line 93: nit: this step should probably be named `ingestStep` for coherence with other steps.

endorama (commented):
Overall this looks very promising to me. I'd suggest to add some comments (es to all `steps`) to clarify behavior and usage and then ask the team for feedback.

functionaltests/steps_test.go
endorama: that's right, let's keep it like it is for the moment.

endorama (approved):
Looks good! 🚀

PR #16505, opened 2025-04-02
functionaltests: execute cleanup on failure

endorama:
## Motivation/summary

Always cleanup test infra.  This prevents any leftover from CI run. We expect test logs to
be informative enough to provide enough troubleshooting aid.

We can consider to set this again in case of race conditions
that we can't reproduce in other ways.

## Checklist


functionaltests/internal/esclient/v7.go
```diff
@@ -0,0 +1,85 @@
+// Licensed to Elasticsearch B.V. under one or more contributor
```
ericywl, line 1: Seems like this got added by accident

marclop (approved):

PR #16519, opened 2025-04-03
functionaltests: Add support for interacting with v7 fleet

ericywl:
## Motivation/summary

We are planning to add a test for 7.x standalone -> 7.x managed -> 8.x managed. However, the 7.x Fleet API, that we use to trigger flushing of APM metrics, is only accessible to superuser.

- Add new fields to `PackagePolicy` that are necessary for update in 7.x
- Remove API key from `kbclient.Client`, since we are only using superuser basic auth
- Increase waiting time after ingestion (aggregation metrics occasionally take longer than expected to arrive in ES)

## How to test these changes

Run functionaltests with this branch.

I have also tested locally that versions from 8.15 to 9.1 are not affected negatively by this change, i.e. the `UpdatePackagePolicyByID` function works as per normal (no changes to unexpected policy fields).

### DIFF
--- a/functionaltests/internal/kbclient/client.go
+++ b/functionaltests/internal/kbclient/client.go
@@ -30,17 +30,13 @@ import (
 	"time"
 )
 
-func New(kibanaURL, apiKey string, username, password string) (*Client, error) {
+func New(kibanaURL, username, password string) (*Client, error) {
 	if kibanaURL == "" {
 		return nil, fmt.Errorf("kbclient.New kibanaURL must not be empty")
-	}
-	if apiKey == "" {
-		return nil, fmt.Errorf("kbclient.New apiKey must not be empty")
 	}
 
 	return &Client{
 		url:                 kibanaURL,
-		apiKey:              apiKey,
 		superUsername:       username,
 		superPassword:       password,
 		SupportedAPIVersion: "2023-10-31",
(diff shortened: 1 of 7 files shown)


functionaltests/utils_test.go
```diff
@@ -170,12 +170,10 @@ func createESClient(t *testing.T, deployInfo deploymentInfo) *esclient.Client {
 
 // createKibanaClient instantiate an HTTP API client with dedicated methods to query the Kibana API.
 // This function will also create an Elasticsearch API key with full permissions to be used by the HTTP client.
```
endorama, line 172:
```suggestion
```

functionaltests/internal/kbclient/client.go
```diff
@@ -187,19 +193,14 @@ func (c *Client) GetPackagePolicyByID(ctx context.Context, policyID string) (Pac
 	return policyResp.Item, nil
 }
 
-type UpdatePackagePolicyRequest struct {
-	PackagePolicy
-	Force bool `json:"force"`
```
endorama, line 192: May you clarify why are we removing this?

functionaltests/internal/kbclient/client.go
```diff
@@ -30,17 +30,13 @@ import (
 	"time"
 )
 
-func New(kibanaURL, apiKey string, username, password string) (*Client, error) {
+func New(kibanaURL, username, password string) (*Client, error) {
```
endorama, line 33: I think it's ok to keep using only username and password also when we decommission 7.x line, so I'm 100% onboard with this change and not making this 7.x specific.

endorama (commented):
A couple of questions but overall 👍

functionaltests/internal/kbclient/client.go
ericywl: Because we don't need to use `force=false` since that is the default, and that v7 doesn't support `force` to begin with. So we are only left with `PackagePolicy`. I'm ok to keep it as `UpdatePackagePolicyRequest`, but thought it was simpler to just remove it.

endorama: make sense, agree on removing it.

endorama (approved):

PR #16611, opened 2025-04-10
feat: use yaml.v2 in export command

kruskall:
## Motivation/summary

beats and related libs are using yaml.v2 to parse the config

yaml.v3 usage is causing the whole lib to be linked in the binary

use yaml.v2 so we standardize on one yaml library

## Checklist


ericywl:
Seems like there were slight differences in yaml.v3 and yaml.v2 😅

1pkg (approved):

Issue #16636, opened 2025-04-15
EC deployment terraform local execution error


ericywl:
When running functional tests, we occasionally get this error:

```
Error: local-exec provisioner error 
    with module.ec_deployment.null_resource.enable_features,
    on ../../testing/infra/terraform/modules/ec_deployment/deployment.tf line 168, in resource "null_resource" "enable_features":
    168:   provisioner "local-exec" {
        Error running command 'scripts/enable_features.sh': exit status 127. Output:
        /bin/bash: line 1: scripts/enable_features.sh: No such file or directory
```

Relevant sections of the [deployment terraform](https://github.com/elastic/apm-server/blob/main/testing/infra/terraform/modules/ec_deployment/deployment.tf#L116-L126):
```
resource "local_file" "enable_features" {
  content = templatefile("${path.module}/scripts/enable_features.tftpl", {
    kibana_url                  = ec_deployment.deployment.kibana.0.https_endpoint,
    elastic_password            = ec_deployment.deployment.elasticsearch_password,
    enable_expvar               = var.apm_server_expvar
    enable_pprof                = var.apm_server_pprof
    enable_tail_sampling        = var.apm_server_tail_sampling
    tail_sampling_storage_limit = var.apm_server_tail_sampling_storage_limit
  })
  filename = "${path.module}/scripts/enable_features.sh"
}

resource "null_resource" "enable_features" {
  triggers = {
    shell_hash          = local_file.enable_features.id
    integrations_server = var.integrations_server
  }
  provisioner "local-exec" {
    command     = "scripts/enable_features.sh"
    interpreter = ["/bin/bash", "-c"]
    working_dir = path.module
  }
}
```

The issue seems to be due to the local file being created in the `ec_deployment/scripts` directory instead of each respective deployment's working directory. Therefore, if 2 or more deployments were created (with different working directories) at the same time, there may be conflicts resulting in the scripts being missing / wrong etc.

endorama:
😢 looking at Terraform docs (https://developer.hashicorp.com/terraform/language/expressions/references#filesystem-and-workspace-info) this is a documented behaviour :

> [path.module](https://developer.hashicorp.com/terraform/language/expressions/references#path-module) is the filesystem path of the module where the expression is placed. We do not recommend using path.module in write operations because it can produce different behavior depending on whether you use remote or local module sources. Multiple invocations of local modules use the same source directory, overwriting the data in path.module during each call. This can lead to race conditions and unexpected results.

[...]

PR #16637, opened 2025-04-15
Fix ec_deployment local exec provisioner error

ericywl:
## Motivation/summary

The existing implementation creates local scripts in the Terraform module directory i.e. `ec_deployment/scripts`. This means that if we were to have multiple deployments at once (for example in [functionaltests](https://github.com/elastic/apm-server/blob/main/functionaltests/utils_test.go#L54-L70) which [references `ec_deployment`](https://github.com/elastic/apm-server/blob/main/functionaltests/infra/terraform/main.tf#L2)), the scripts from each deployment would overwrite one another.

Fixed by using `path.cwd` instead of `path.module` for `ec_deployment` scripts. Closes https://github.com/elastic/apm-server/issues/16636.

## How to test these changes

Run functionaltests and benchmark workflow with this branch, to ensure they are working properly.

### DIFF
--- a/testing/infra/terraform/modules/ec_deployment/deployment.tf
+++ b/testing/infra/terraform/modules/ec_deployment/deployment.tf
@@ -122,7 +122,7 @@ resource "local_file" "enable_features" {
     enable_tail_sampling        = var.apm_server_tail_sampling
     tail_sampling_storage_limit = var.apm_server_tail_sampling_storage_limit
   })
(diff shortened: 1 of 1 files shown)


ericywl:
Not sure if we need to backport, since this technically only affects functionaltests due to its parallel deployment.

1pkg (approved):

PR #16695, opened 2025-04-23
functionaltests: Move and rename tests

ericywl:
## Motivation/summary

- To avoid confusion, all files not containing Go tests are renamed to remove the `_test` suffix. Only the files containing Go tests e.g. `8_17_test.go` will have the suffix.
- `single_upgrade_test.go` has also been removed, as that is the old way that we used to write tests, which has been replaced by `steps.go` and `basic_upgrade.go`.
- Merged `8_x_test.go` and `9_x_test.go`, renamed to `standalone_test.go` since they only contained standalone tests.

## How to test these changes

Run functionaltests with this branch: https://github.com/elastic/apm-server/actions/runs/14726436275. The failures seems to be unrelated to the change, and is caused by some regression in ES.


.github/workflows/functional-tests.yml
```diff
@@ -28,13 +28,12 @@ jobs:
         environment:
           - 'pro'
         active-version:
-          - '8.16'
```
ericywl, line 31: 8.16 is no longer active

1pkg (approved):

simitt (approved):

PR #16812, opened 2025-05-05
changelog: add entry for 8.17.6

1pkg:
Add changelog entry for 8.17.6 release

### DIFF
--- a/changelogs/8.17.asciidoc
+++ b/changelogs/8.17.asciidoc
@@ -1,12 +1,36 @@
 [[apm-release-notes-8.17]]
 == APM version 8.17
 
+* <<apm-release-notes-8.17.6>>
 * <<apm-release-notes-8.17.5>>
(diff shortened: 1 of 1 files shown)


changelogs/8.17.asciidoc
```diff
@@ -1,13 +1,37 @@
...
+[float]
+[[apm-release-notes-8.17.6]]
+=== APM version 8.17.6
+
+https://github.com/elastic/apm-server/compare/v8.17.5\...v8.17.6[View commits]
+
+[float]
+==== Bug fixes
+
+- Tail-based sampling: Fix an issue with corrupted subscriber position files by handling file read error {pull}16736[16736]
```
carsonip, line 21:
```suggestion
- Tail-based sampling: ignore subscriber position read error and proceed as if file does not exist to avoid crash looping {pull}16736[16736]
```

PR #16849, opened 2025-05-07
fix: golang-crossbuild requires host specific for arm

v1v:
remove crossbuild for arm4 on amd64

## Motivation/summary

See https://github.com/elastic/golang-crossbuild/pull/560

## Checklist

### DIFF
--- a/.buildkite/package.yml
+++ b/.buildkite/package.yml
@@ -5,7 +5,6 @@ notify:
     if: "build.state != 'passed'"
 
 env:
-  PLATFORMS: "darwin/amd64 linux/386 linux/amd64 windows/386 windows/amd64"
   IMAGE_UBUNTU_X86_64: "family/core-ubuntu-2004"
   IMAGE_UBUNTU_ARM_64: "core-ubuntu-2004-aarch64"
   BEAT_DOC_BRANCH: "7.x"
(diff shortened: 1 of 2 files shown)


.buildkite/package.yml
```diff
@@ -5,6 +5,7 @@ notify:
     if: "build.state != 'passed'"
 
 env:
+  PLATFORMS: "+all linux/amd64 windows/amd64 darwin/amd64"
```
v1v, line 8: I'm not sure if we support `darwin/amd64` or any other platforms

v1v: See https://github.com/elastic/beats/blob/a156701b0302ca0e4e8676d2de96ebf2c1b7879d/dev-tools/mage/platforms.go/#L28

.buildkite/scripts/package.sh
```diff
@@ -8,14 +8,7 @@
...
 
 TYPE="$1"
-PLATFORM_TYPE=$(uname -m)
 
-PLATFORMS=''
-PACKAGES=''
-if [[ ${PLATFORM_TYPE} == "arm" || ${PLATFORM_TYPE} == "aarch64" ]]; then
-  PLATFORMS='linux/arm64'
-  PACKAGES='docker'
-fi
```
v1v, line 12: `PLATFORMS` was defaulted, henve the BK env variables was not working as expected. Avoid using the script but honour the env variables defined in the Buildkite pipeline itself.

v1v:
It works:

image

image


probably something in the scripts when no DRA is needed... but for now I think this PR is good to go

rubvs (approved):

PR #16914, opened 2025-05-14
[docs] Add `products` to `docset.yml`

colleenmcginnis:
Related to https://github.com/elastic/docs-builder/issues/1200

Add `products` to `docset.yml` to be used in the search experience.


kruskall:
@colleenmcginnis should this be backported ?

Issue #16697, opened 2025-04-23
functionaltests: Generalized upgrade tests


ericywl:
## Problem

Current upgrade tests are not automatic. Each version upgrade test tests (most of the time) only upgrades between consecutive minors e.g. `8.17 -> 8.18`. This is due to the fact that version upgrades have some intricacies in the way that we assert them currently. 

Our assertions include lazy rollover (no. indices per data stream) and per-index lifecycle management. This is good for catching issues that happened e.g. in `8.15`, but also means that we have to manually add the tests every time a new minor comes out. 

## Generalized tests

What we can do is to add generalized tests with moving targets. 

#### Generalization

These generalized tests will not go into the specifics (lazy rollover / per-index lifecycle management), and only focus on two things: document counts and data stream lifecycle. We know that in the foreseeable future, the data streams in APM should be managed solely by ILM.

#### Moving Targets

They will also be towards moving targets i.e. `8.x` or `9.x` instead of any specific minor versions. For example, if we start from `8.15`, we will implement `TestUpgrade_8_15_to_8_x` or `TestUpgrade_8_15_to_8_x_to_9_x`. We will need to handpick a few source versions to do this, as implementing for all of them would defeat the purpose of having the generalized tests in the first place (not manually updating tests)

ericywl:
I will experiment on this next week (May 12 - 16)

ericywl:
After discussion with @simitt and @endorama, there are a few things to be done in subsequent PRs.

1. Clean up the internal package
    - Move versions.go into `internal/ecclient`
    - Merging files with similar functionalities together e.g. `internal/kbclient/client.go` and `internal/kbclient/upgrade_assist.go`
    - Rename `ecclient` -> `ech`, `kbclient` -> `kibana`, `esclient` -> `elasticsearch`
    - `internal/asserts` should probably be in the main test folder instead (?)
2. Rename functionaltests to integrationservertest
3. Double check old upgrade smoke test and remove them

One remaining question:
1. What to do with the existing `smoke-test` after we remove the upgrade tests?

cc @raultorrecilla

simitt:
>What to do with the existing smoke-test after we remove the upgrade tests?

These will be trimmed down to the `OS smoke-tests`, I would rename them to `os-smoke-tests` and then keep them running as such for now.

raultorrecilla:
@ericywl did we update our docs to reflect the current testing strategy?

PR #16917, opened 2025-05-15
functionaltests: Add generalized test

ericywl:
## Motivation/summary

Closes https://github.com/elastic/apm-server/issues/16697.

This PR introduces generalized upgrade test. The new generalized upgrade test runs the exact same checks as the previous upgrade tests, and is configurable via YAML (see `upgrade-config.yaml`). The upgrade paths for the tests are provided via the `upgrade-path` argument. We use the GitHub workflow matrix to provide the `upgrade-path` argument to the test and split each run into its own instance.

All other upgrade tests have now been removed.

## How to test these changes

Run functionaltests workflow: https://github.com/elastic/apm-server/actions/runs/15318189642


functionaltests/8_15_test.go
```diff
@@ -21,27 +21,34 @@ import (
...
+// In 8.15, the data stream management was migrated from ILM to DSL.
+// However, a bug was introduced, causing data streams to be unmanaged.
+// See https://github.com/elastic/apm-server/issues/13898.
+//
+// It was fixed by defaulting data stream management to DSL, and eventually
+// reverted back to ILM in 8.17. Therefore, data streams created in 8.15 and
+// 8.16 are managed by DSL instead of ILM.
 
-func TestUpgrade_8_18_to_8_19_Snapshot(t *testing.T) {
+func TestUpgrade_8_15_to_8_16_Snapshot(t *testing.T) {
```
simitt, line 32: This should also be "rolling" to the latest snapshots (`8.19`). 
The version `8.16` won't receive any more patch releases.

functionaltests/8_15_test.go
```diff
@@ -21,27 +21,34 @@ import (
...
 		to.Version,
 		apmErrorLogs{
 			tlsHandshakeError,
 			esReturnedUnknown503,
-			refreshCache503,
+			preconditionFailed,
+			populateSourcemapServerShuttingDown,
+			refreshCacheCtxDeadline,
+			refreshCacheCtxCanceled,
+			// TODO: remove once fixed
```
simitt, line 51: good start; would be even better to link to a github issue

functionaltests/steps.go
```diff
@@ -314,8 +337,97 @@ type customStep struct {
...
+		fmt.Sprintf("traces-apm-%s", namespace):                     15013,
+		fmt.Sprintf("metrics-apm.app.opbeans_python-%s", namespace): 1437,
+		fmt.Sprintf("metrics-apm.internal-%s", namespace):           1351,
+		fmt.Sprintf("logs-apm.error-%s", namespace):                 364,
+		// Ignore aggregation data streams.
+		fmt.Sprintf("metrics-apm.service_destination.1m-%s", namespace): -1,
+		fmt.Sprintf("metrics-apm.service_transaction.1m-%s", namespace): -1,
+		fmt.Sprintf("metrics-apm.service_summary.1m-%s", namespace):     -1,
+		fmt.Sprintf("metrics-apm.transaction.1m-%s", namespace):         -1,
+	}
```
simitt, line 360: Would it not be enough to test for number of docs in datastreams > 0 or higher than at the start of a version test?

functionaltests/steps.go
```diff
@@ -314,8 +337,97 @@ type customStep struct {
...
+		fmt.Sprintf("metrics-apm.app.opbeans_python-%s", namespace): 0,
+		fmt.Sprintf("metrics-apm.internal-%s", namespace):           0,
+		fmt.Sprintf("logs-apm.error-%s", namespace):                 0,
+		// Ignore aggregation data streams.
+		fmt.Sprintf("metrics-apm.service_destination.1m-%s", namespace): -1,
+		fmt.Sprintf("metrics-apm.service_transaction.1m-%s", namespace): -1,
+		fmt.Sprintf("metrics-apm.service_summary.1m-%s", namespace):     -1,
+		fmt.Sprintf("metrics-apm.transaction.1m-%s", namespace):         -1,
+	}
+}
```
simitt, line 380: Why are aggregated metrics data streams not expected to be 0 if no docs were indexed?

simitt (commented):
Are the `8_17_test.go`, `8_18_test.go`, `9_0_test.go`, `9_1_test.go` files eventually going to be replaced by the more generic `upgrade_test.go` ?

ericywl:
> Are the 8_17_test.go, 8_18_test.go, 9_0_test.go, 9_1_test.go files eventually going to be replaced by the more generic upgrade_test.go ?

Yes

functionaltests/8_15_test.go
ericywl: Makes sense, I will update it in a separate PR instead.

functionaltests/steps_v7.go
```diff
@@ -87,8 +87,6 @@ func emptyDataStreamsIngestV7(namespace string) esclient.DataStreamsDocCount {
 // NOTE: Only works for versions 7.x.
 type ingestV7Step struct{}
 
-var _ testStep = ingestV7Step{}
```
simitt, line 90: Can you please merge `steps.go` and `steps_v7.go` and instead of having two files, organize things properly with `structs. This `steps_v7.go` has several things defined unrelated to `v7`, eg. [emptyIndicesIngest](https://github.com/elastic/apm-server/pull/16917/files#diff-b16cdc37470b93531b9fa9e7d9af76e3c6a757ff58c641b15ed9205c45a51785L33-R54), or the migration step, which can also happen on version 8.

.github/workflows/functional-tests.yml
```diff
@@ -20,22 +20,17 @@ env:
...
-          - '8.17'
-          - '8.18'
-          - '8.19'
-          - '9.0'
-          - '9.1'
-          - '8.x' # Latest 8 version
-          - '9.x' # Latest 9 version
+        upgrade-path:
+          - '8.15, 8.16, 8.17, 8.18, 9.0'
+          - '8.17, 8.19, 9.1'
```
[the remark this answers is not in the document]
ericywl: I agree with that. Would make the entire test finish faster as well.

functionaltests/steps.go
endorama: This ties to the comment [above](https://github.com/elastic/apm-server/pull/16917/files#r2101712325). The reason why the are ignored is because the aggregation doc count can change independently from ingestion. This leads to tests failing based on timing.

The way this variable is used, similarly to `expectedDataStreamsIngest`, is to add a certain amount to the current known amount and use this value in assertions. As the doc count for aggregation data streams is not known, they are ignored in this case too.

But if we move in a different direction, where we get the document count immediately before the action that (may) affect it, we can also get rid of the logic that uses these functions. I think it's a simplification worth pursuing.

functionaltests/README.md
```diff
@@ -4,64 +4,67 @@
...
-e.g. `TestUpgrade_7_17_to_8_x_Standalone_to_Managed`. Otherwise, if there is no upgrade simply omit the `Upgrade`
-prefix, e.g. `Test_7_17_Standalone_to_Managed`.
+For standalone tests:
+```sh
+go test -run=TestStandaloneManaged -v -timeout=30m -cleanup-on-failure=false -target="pro" ./
+```
+
+## Debugging the Tests
+
+If you get some errors after running the test, you can try heading to the [Elastic Cloud console](https://cloud.elastic.co/home) in order to access the Kibana instance.
```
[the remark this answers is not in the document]
ericywl: 👍 I added the following line previously, but did not mention CI. I will change it to:
```
Note: If the tests failed due to deployment or in CI, you may need to access the Elastic Cloud admin console instead to check the errors.
```

functionaltests/upgrade-config.yaml
```diff
@@ -0,0 +1,18 @@
...
+# Exceptions are specified as a list, i.e. no lazy rollover if upgrade is from
+# that minor version.
+lazy-rollover-with-exceptions:
+  8.16:
+  8.17:
+  # 8.19 and 9.1 have the same template due to
+  # https://github.com/elastic/elasticsearch/pull/123166.
+  8.19:
+  9.1:
+    - "8.19"
```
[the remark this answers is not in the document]
ericywl: By flipping the default, do you mean adding versions that will have lazy rollover?
```
lazy-rollover:
  8.16:
    - "8.15"
  8.17:
   - "8.15"
   - "8.16"
  8.19:
   - "8.17"
   - "8.18"
  9.1:
    - "9.0"
```

functionaltests/README.md
```diff
@@ -4,64 +4,67 @@
...
-Then, from the current directory, simply run:
-```sh
-go test -v -timeout=30m -cleanup-on-failure=false -target="pro" ./
+The following is a simplified directory structure of functional tests.
 ```
-
-You can also specify a specific test you want to run, for example:
-```sh
-go test -run=TestUpgrade_8_18_to_9_0 -v -timeout=30m -cleanup-on-failure=false -target="pro" ./
+- functionaltests/
```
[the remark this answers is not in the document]
simitt: Agreed; let's keep this as comment for next steps, but outside this PR

functionaltests/steps.go
```diff
@@ -145,7 +157,7 @@ func (c createStep) Step(t *testing.T, ctx context.Context, e *testStepEnv, _ te
 
 	t.Logf("------ cluster setup %s ------", c.DeployVersion)
 	e.tf = initTerraformRunner(t)
-	deployInfo := createCluster(t, ctx, e.tf, *target, c.DeployVersion, integrations)
+	deployInfo := createCluster(t, ctx, e.tf, e.target, c.DeployVersion, integrations, c.CleanupOnFailure)
 	e.esc = createESClient(t, deployInfo)
 	e.kbc = createKibanaClient(t, deployInfo)
 	e.gen = createAPMGenerator(t, ctx, e.esc, e.kbc, deployInfo)
```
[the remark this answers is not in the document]
simitt: You're totally right, my bad.

functionaltests/steps_v7.go
simitt: Even when support for `v7` stops, there could still be deployments that have not switched from standalone to managed. I'm not convinced we will be able to completely remove the test logic for this soon-ish. 
The logic can and should still be grouped together, however the way it is grouped now has some implicit knowledge in it that I am not sure is necessary. Let's discuss in our sync.

functionaltests/setup.go
```diff
@@ -277,7 +164,7 @@ func createCluster(
 	require.NoError(t, tf.Apply(ctx, ecTarget, ecRegion, ecDeploymentTpl, version, integrations, name))
 
 	t.Cleanup(func() {
-		if !t.Failed() || (t.Failed() && *cleanupOnFailure) {
+		if !t.Failed() || (t.Failed() && cleanupOnFailure) {
```
1pkg, line 167: nit: `!t.Failed() || (t.Failed() && cleanupOnFailure)` could be simplified to `!t.Failed() || cleanupOnFailure`

functionaltests/upgrade-config.yaml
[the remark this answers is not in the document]
1pkg: What is the point of having this configurable in `yaml` vs using static map in the code to cover these exceptions? Is this for future proofing? I hope that there won't be a need to add many more of them in the future.

1pkg (approved):

functionaltests/upgrade_test.go
```diff
@@ -0,0 +1,234 @@
...
+					DSManagedBy:      lifecycle,
+					IndicesManagedBy: indicesManagedBy,
+				}),
+			})
+			continue
+		}
+
+		// Upgrade deployment to new version and ingest.
+		prev := versionInfos[i-1].Version
+		oldIndicesManagedBy := copySlice(indicesManagedBy)
```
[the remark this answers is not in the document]
ericywl: TIL

functionaltests/upgrade-config.yaml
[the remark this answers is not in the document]
ericywl: I think both way works. We will need to make changes whenever there are new "template breaking" minor releases regardless.

1pkg (approved):

ericywl:
After discussion with @simitt and @endorama, there are a few things to be done in subsequent PRs.
1. Clean up the `internal` package
    - Move `versions.go` into `internal/ecclient`
    - Merging files with similar functionalities together e.g. `internal/kbclient/client.go` and `internal/kbclient/upgrade_assist.go`
    - Rename `ecclient` -> `ech`, `kbclient` -> `kibana`, `esclient` -> `elasticsearch`
    - `internal/asserts` should probably be in the main test folder instead (?)
2. Rename `functionaltests` to `integrationservertest`
3. Double check old upgrade smoke test and remove them

endorama (approved):

1pkg (approved):

functionaltests/internal/asserts/doc_count.go
```diff
@@ -18,52 +18,61 @@
...
-	// Check document counts for all data streams.
-	for ds, v := range currDocCount {
-		e, ok := expectedDiff[ds]
-		if !ok {
-			t.Errorf("unexpected documents (%d) for data stream %s", v, ds)
-			continue
-		}
+// DocCountStayedTheSame checks if current document counts for all data streams / indices
+// stayed the same from the previous.
+func DocCountStayedTheSame(t *testing.T, currDocCount, prevDocCount map[string]int) {
```
simitt, line 62: nit:
```
func DocCountUnchanged(t *testing.T, currDocCount, prevDocCount map[string]int) {
```

simitt (approved):
Changes look good; appreciate to see these tests generalized. 
Feel free to ignore the nit and get this merged.

functionaltests/internal/asserts/doc_count.go
ericywl: I will incorporate this in the next PR 😄

Issue #16838, opened 2025-05-07
Data race in TestServerTracingEnabled


carsonip:
**APM Server version** (`apm-server version`): main

**Description of the problem including expected versus actual behavior**:

data race in tests

**Steps to reproduce**:

```
go test -count=10 -race ./internal/beater/...
```

**Provide logs (if relevant)**:

```
==================
WARNING: DATA RACE
Read at 0x00c002016e43 by goroutine 1319:
  testing.(*common).logDepth()
      /home/carson/sdk/go1.24.2/src/testing/testing.go:1059 +0x4e4
  testing.(*common).log()
      /home/carson/sdk/go1.24.2/src/testing/testing.go:1046 +0x9e
  testing.(*common).Logf()
      /home/carson/sdk/go1.24.2/src/testing/testing.go:1097 +0x66
  go.uber.org/zap/zaptest.TestingWriter.Write()
      /home/carson/go/pkg/mod/go.uber.org/zap@v1.27.0/zaptest/logger.go:146 +0x11c
  go.uber.org/zap/zaptest.(*TestingWriter).Write()
      <autogenerated>:1 +0x74
  go.uber.org/zap/zapcore.(*ioCore).Write()
      /home/carson/go/pkg/mod/go.uber.org/zap@v1.27.0/zapcore/core.go:99 +0x18d
  go.uber.org/zap/zapcore.(*CheckedEntry).Write()
      /home/carson/go/pkg/mod/go.uber.org/zap@v1.27.0/zapcore/entry.go:253 +0x1ec
  go.uber.org/zap.(*SugaredLogger).log()
      /home/carson/go/pkg/mod/go.uber.org/zap@v1.27.0/sugar.go:355 +0x12c
  go.uber.org/zap.(*SugaredLogger).Info()
      /home/carson/go/pkg/mod/go.uber.org/zap@v1.27.0/sugar.go:155 +0x6c
  github.com/elastic/elastic-agent-libs/logp.(*Logger).Info()
      /home/carson/go/pkg/mod/github.com/elastic/elastic-agent-libs@v0.19.3/logp/logger.go:163 +0x2c
  github.com/elastic/apm-server/internal/beater/api.apmMiddleware.LogMiddleware.func1.1()
      /home/carson/projects/apm-server/internal/beater/middleware/log_middleware.go:63 +0x5ba
[...]

PR #17029, opened 2025-05-28
fix: close tracer after main server and final processor

kruskall:
## Motivation/summary

wait for the main server and final batch processor to finish and only then close the tracer processor

ensure the tracer server and processor are closed last to avoid missing instrumentation data.
Call Shutdown on the main thread to avoid goroutine leak

## Checklist

## Related issues

Related to https://github.com/elastic/apm-server/issues/16838


internal/beater/beater.go
```diff
@@ -512,23 +512,24 @@ func (s *Runner) Run(ctx context.Context) error {
...
-
 		tracerProcessor := append(preBatchProcessors, tracerBatchProcessor)
 		tracerServer, err := newTracerServer(s.config, tracerServerListener, s.logger, tracerProcessor, serverParams.Semaphore, serverParams.MeterProvider)
 		if err != nil {
 			return fmt.Errorf("failed to create self-instrumentation server: %w", err)
 		}
-		g.Go(func() error {
+		go func() error {
 			if err := tracerServer.Serve(tracerServerListener); err != http.ErrServerClosed {
 				return err
```
1pkg, line 522: This error is now getting silently ignored, we should probably log it at least.

PR #17042, opened 2025-05-30
fix: decrease dynamic group metric count

kruskall:
## Motivation/summary

dynamic group metric count is always growing

correctly decrease the counter when finalizing sample trace

## Checklist


lahsivjar (approved):

carsonip (commented):
Can we encapsulate `numDynamicServiceGroups` and `numDynamicServiceGroupsCounter` into 1 struct expose `Increment` and `Decrement` to prevent future hiccups?

Issue #17055, opened 2025-06-02
Replace smoke-test-ess with integrationservertest


ericywl:
- Rename `functionaltests` to `integrationservertest`
- Double check if we missed anything from old smoke test, and remove it
- Rename `smoke-test` to `os-smoke-test`

PR #17085, opened 2025-06-04
Remove smoke-tests-ess and enable integration-server-test schedule

ericywl:
## Motivation/summary

Closes https://github.com/elastic/apm-server/issues/17055

- Remove `smoke-tests-ess` and its references
- Rename `functional-tests` to `integration-server-test`
- Enable `integration-server-test` schedule

## How to test these changes

Run both `smoke-tests-os-sched`: https://github.com/elastic/apm-server/actions/runs/15438258662 (failures not relevant to this PR) and `integration-server-test`: https://github.com/elastic/apm-server/actions/runs/15457628569

NOTE: How these runs worked is that I renamed the workflows back to its original name and run them, so you might not be able to run them directly. Workflows need to be merged to main before appearing in Actions it seems.


Makefile
```diff
@@ -350,62 +350,6 @@ testing/rally/corpora:
 	@rm -fr testing/rally/corpora && mkdir testing/rally/corpora
 	@cd systemtest/cmd/gencorpora && go run . -write-dir $(CURRENT_DIR)/testing/rally/corpora/ -replay-count $(RALLY_GENCORPORA_REPLAY_COUNT)
 
-##############################################################################
```
endorama, line 353: As we are promoting the integration tests to our only tests, what do you think about adding make targets to run them?

endorama (commented):
Left a minor comment, overall looks good!

Makefile
ericywl: Yea that makes sense, let me add them.

endorama: What do you think of also using these make targets in the CI? That would help having a single entry point for running these tests.

ericywl: Done in latest commit

1pkg (approved):

Issue #16970, opened 2025-05-21
Data race in TestInvalidation


ericywl:
Data race in the macOS CI test. To be investigated.

```
=== RUN   TestInvalidation/hash_changed
==================
WARNING: DATA RACE
Read at 0x00c000103e03 by goroutine 168:
  testing.(*common).logDepth()
      /Users/runner/hostedtoolcache/go/1.24.3/arm64/src/testing/testing.go:1059 +0x380
  testing.(*common).log()
      /Users/runner/hostedtoolcache/go/1.24.3/arm64/src/testing/testing.go:1046 +0x80
  testing.(*common).Logf()
      /Users/runner/hostedtoolcache/go/1.24.3/arm64/src/testing/testing.go:1097 +0x58
  go.uber.org/zap/zaptest.TestingWriter.Write()
      /Users/runner/go/pkg/mod/go.uber.org/zap@v1.27.0/zaptest/logger.go:146 +0xf0
  go.uber.org/zap/zaptest.(*TestingWriter).Write()
      <autogenerated>:1 +0x64
  go.uber.org/zap/zapcore.(*ioCore).Write()
      /Users/runner/go/pkg/mod/go.uber.org/zap@v1.27.0/zapcore/core.go:99 +0x11c
  go.uber.org/zap/zapcore.(*CheckedEntry).Write()
      /Users/runner/go/pkg/mod/go.uber.org/zap@v1.27.0/zapcore/entry.go:253 +0x158
  go.uber.org/zap.(*SugaredLogger).log()
      /Users/runner/go/pkg/mod/go.uber.org/zap@v1.27.0/sugar.go:355 +0x104
  go.uber.org/zap.(*SugaredLogger).Info()
      /Users/runner/go/pkg/mod/go.uber.org/zap@v1.27.0/sugar.go:155 +0x450
  github.com/elastic/elastic-agent-libs/logp.(*Logger).Info()
      /Users/runner/go/pkg/mod/github.com/elastic/elastic-agent-libs@v0.19.5/logp/logger.go:163 +0x408
  github.com/elastic/apm-server/internal/sourcemap.(*MetadataESFetcher).startBackgroundSync.func1()
      /Users/runner/work/apm-server/apm-server/internal/sourcemap/metadata_fetcher.go:129 +0x3d0

Previous write at 0x00c000103e03 by goroutine 126:
  testing.tRunner.func1()
[...]

PR #17098, opened 2025-06-04
test: wait for fetcher background goroutine to exit

kruskall:
## Motivation/summary

fix TestInvalidation race condition due to goroutine leak

## Checklist

## How to test these changes

`go test -race -count=1000 -failfast -run=TestInvalidation ./...`

## Related issues

Closes https://github.com/elastic/apm-server/issues/16970


internal/sourcemap/metadata_fetcher_test.go
```diff
@@ -297,6 +297,10 @@ func TestInvalidation(t *testing.T) {
 			}
 
 			cancel()
+			// wait for invalidationChan to be closed so
+			// the background goroutine is done
+			for range invalidationChan {
+			}
 			<-invCh
```
carsonip:
```suggestion
			<-invCh
			// wait for invalidationChan to be closed so
			// the background goroutine is done
			for range invalidationChan {
			}
```
nit: is it slightly better to do it after `<-invCh`, because this `for range` may compete with the `invalidationChan` read from line 264.

kruskall: good catch!

kruskall:
> invalidationChan is closed _very_ close to the `return` in the goroutine in startBackgroundSync but not exactly after all threads are cleaned up. Does it matter, or would it still race?

nah because at that point there's nothing else (logger calls) that could race.

> any backports to 8.18 and 9.0?

I don't think 8.18 has the logger changes but I guess we coudl backport this

internal/sourcemap/metadata_fetcher_test.go
```diff
@@ -298,6 +298,10 @@ func TestInvalidation(t *testing.T) {
 
 			cancel()
 			<-invCh
+			// wait for invalidationChan to be closed so
+			// the background goroutine is done
```
carsonip:
```suggestion
			// the background goroutine created by NewMetadataFetcher is done
```

Issue #17104, opened 2025-06-05
monitoring: TBS lsm_size and value_log_size are reported less frequently than other TBS metrics


carsonip:
**APM Server version** (`apm-server version`): 9.0

**Description of the problem including expected versus actual behavior**:

`apm-server.sampling.tail.storage.lsm_size` and `apm-server.sampling.tail.storage.value_log_size` are reported less frequently (~15mins) than other TBS metrics e.g. `apm-server.sampling.tail.dynamic_service_groups` (~1min).

Also, `apm-server.sampling.tail.storage.lsm_size` and `apm-server.sampling.tail.dynamic_service_groups` are never reported together.

**Steps to reproduce**:

Please include a *minimal* but *complete* recreation of the problem,
including server configuration, agent(s) used, etc.  The easier you make it
for us to reproduce it, the more likely that somebody will take the time to
look at it.

 1.
 2.
 3.

**Provide logs (if relevant)**:

isaacaflores2:
`apm-server.sampling.tail.storage.lsm_size` and `apm-server.sampling.tail.storage.value_log_size` appear to be the only two asynchronous (observable) TBS metrics which explains the different reporting interval.  

My understanding is the `meterProvider` can be configured with a different interval using something like a `PeriodicReader` but I do not see such configuration. The only meterProvider I find is [here](https://github.com/elastic/apm-server/blob/main/internal/beatcmd/beat.go#L128-L132) which is using `ManualReaders`. I cannot find any collection interval in the apm-server codebase. 

Since we want some more control on when these two metrics are reported I will work on converting to synchronous measurements.

carsonip:
> which explains the different reporting interval.

My experience is that lsm_size and value_log_size just won't show up in `localhost:5066/stats`.

> Since we want some more control on when these two metrics are reported I will work on converting to synchronous measurements.

++ on converting to synchronous, as long as you confirm that the synchronous measurement is regularly updated by a background goroutine, which should be the case.

>  the only two asynchronous (observable) TBS metrics 

A quick search seems to show there are some other observable non-TBS metrics. If they exhibit the same issue, which I suppose they would, we can tackle them together now or in a follow up PR, or create a separate issue to tackle them with a lower urgency.

isaacaflores2:
Sounds good. I think we can review the other observable metrics and create another issue. I still need understand how current metric interval is determined for apm server. 


PR is open which takes advantage of an existing storage manager background goroutiner.

[...]

PR #17154, opened 2025-06-10
tbs: Update storage metrics to be reported synchronously in the existing `runDiskUsageLoop` method

isaacaflores2:
## Motivation/summary

Updated metrics for `apm-server.sampling.tail.storage.lsm_size` and `apm-server.sampling.tail.storage.value_log_size` to be synchronous so we can control the reporting interval.  The existing `runDiskUsageLoop` is run every 1s. 

## Checklist

## How to test these changes
1. Unit test
2. Calling the stats endpoint ` localhost:5066/stats`
   ```json
   "sampling": {
		"tail": {
			"storage": {
				"lsm_size": 6544,
				"value_log_size": 0
			}
		}
	}		
   ```

## Related issues

Closes https://github.com/elastic/apm-server/issues/17104


isaacaflores2:
@carsonip or @ericywl not sure if v9.0.3 requires a backport so please adjust the label if needed.

x-pack/apm-server/main_test.go
```diff
@@ -52,6 +53,12 @@ func TestMonitoring(t *testing.T) {
 
 	// Wrap & run the server twice, to ensure metric registration does not panic.
 	runServerError := errors.New("runServer")
+	runServerFunc := func(ctx context.Context, args beater.ServerParams) error {
+		// run server for some time to allow storage metrics to be reported
+		// should be wait at least x-pack/apm-server/sampling/eventstorage/storage_manager.go:diskUsageFetchInterval
```
ericywl: Might be better to provide a direct link to the line, e.g. https://github.com/elastic/apm-server/blob/main/x-pack/apm-server/sampling/eventstorage/storage_manager.go#L49

carsonip: i'm not too concerned about this, as lines may change

x-pack/apm-server/sampling/eventstorage/storage_manager.go
```diff
@@ -285,7 +294,7 @@ func (sm *StorageManager) diskUsed() uint64 {
 	return sm.cachedDiskStat.used.Load()
 }
 
-// runDiskUsageLoop runs a loop that updates cached disk usage regularly.
+// runDiskUsageLoop runs a loop that updates cached disk usage regularly and reports usage.
 func (sm *StorageManager) runDiskUsageLoop(stopping <-chan struct{}) error {
```
carsonip, line 300: Seems that running `updateDiskUsage` before the first ticker can save us up to 1 second in tests, and I also don't see why the user has to wait for up to 1s before seeing it in `/stats`. I think we should do it.

x-pack/apm-server/main_test.go
```diff
@@ -61,9 +68,7 @@ func TestMonitoring(t *testing.T) {
 			BatchProcessor:         modelpb.ProcessBatchFunc(func(ctx context.Context, b *modelpb.Batch) error { return nil }),
 			Namespace:              "default",
 			NewElasticsearchClient: elasticsearch.NewClient,
-		}, func(ctx context.Context, args beater.ServerParams) error {
-			return runServerError
-		})
+		}, runServerFunc)
```
carsonip, line 76: I'm thinking about a runServerFunc that uses `assert.Eventually` and returns runServerError. Then we can potentially save some test time combining with the runDiskUsageLoop change

x-pack/apm-server/sampling/processor_test.go
```diff
@@ -590,21 +590,28 @@ func TestGroupsMonitoring(t *testing.T) {
 	})
 }
 
-func getGauge(t testing.TB, reader sdkmetric.Reader, name string) int64 {
+// getGaugeValues collects metrics and searches for gauge values that match the provided names.
+// Values will be returned in the same order as the names.
+//
+// It is helpful to provide multiple names for synchronous metrics to avoid losing data when collecting.
+// Observable metrics report everytime Collect is called, so there will be no data loss.
+func getGaugeValues(t testing.TB, reader sdkmetric.Reader, names []string) []int64 {
```
carsonip: nit, non-blocker: seems to me that it is a good candidate for a variadic function, as name is the last param accepting a slice.

x-pack/apm-server/sampling/eventstorage/storage_manager.go
```diff
@@ -259,7 +262,13 @@ func (sm *StorageManager) dbSize() uint64 {
...
 
 func (sm *StorageManager) updateDiskUsage() {
-	sm.cachedDBSize.Store(sm.getDBSize())
+	lsmSize := sm.getDBSize()
+	sm.cachedDBSize.Store(lsmSize)
+
+	if sm.meterProvider != nil {
+		sm.storageMetrics.lsmSizeGauge.Record(context.Background(), int64(lsmSize))
+		sm.storageMetrics.valueLogSizeGauge.Record(context.Background(), int64(defaultValueLogSize))
+	}
```
carsonip:
```suggestion
	if sm.storageMetrics.lsmSizeGauge != nil {
		sm.storageMetrics.lsmSizeGauge.Record(context.Background(), int64(lsmSize))
	}
	if sm.storageMetrics.valueLogSizeGauge != nil {
		sm.storageMetrics.valueLogSizeGauge.Record(context.Background(), int64(defaultValueLogSize))
	}
```
nit: a nil check on the gauge itself eliminates the need to depend on another variable

x-pack/apm-server/sampling/eventstorage/storage_manager.go
isaacaflores2: Good idea, I updated to call `updateDiskUsage` right away

x-pack/apm-server/main_test.go
isaacaflores2: The changes to `runDiskUsageLoop` allow us to reduce the sleep to at least 10ms, so I was able to add assert.Eventually with a 10ms tick

x-pack/apm-server/sampling/processor_test.go
isaacaflores2: Updated thanks

x-pack/apm-server/sampling/eventstorage/storage_manager.go
isaacaflores2: Updated as well thanks

x-pack/apm-server/main_test.go
isaacaflores2: The reference is no longer needed due to the updates to `runDiskUsageLoop`, thanks

carsonip (dismissed):
perfect, thanks! Just some lint errors in CI.

carsonip (approved):

PR #17209, opened 2025-06-12
integration-server-test: Don't notify if not run from main

ericywl:
## Motivation/summary

~~Due to changes to `apm-data` introduced in https://github.com/elastic/elasticsearch/pull/128913, 8.19 and 9.1 no longer share the same templates. As a result, when upgrading from 8.19 to 9.1, there will be index lazy-rollover on ingest.~~

~~This PR fixes the failure caused by the above, and also update notify step to fire on main branch only.~~

It seems the failures is simply due to the `apm-data` change not being backported to 8.19 yet. I have reverted the relevant changes here, but left the PR up since there are some good changes.

### DIFF
--- a/integrationservertest/logs_filters.go
+++ b/integrationservertest/logs_filters.go
@@ -102,9 +102,15 @@ var (
 		},
 	})
 
-	syncSourcemapContextCanceled = apmErrorLog(types.Query{
+	initialSearchQueryContextCanceled = apmErrorLog(types.Query{
 		MatchPhrase: map[string]types.MatchPhraseQuery{
 			"message": {Query: "failed to sync sourcemaps metadata: failed to run initial search query: fetcher unavailable: context canceled"},
+		},
+	})
+
+	scrollSearchQueryContextCanceled = apmErrorLog(types.Query{
+		MatchPhrase: map[string]types.MatchPhraseQuery{
+			"message": {Query: "failed to sync sourcemaps metadata: failed scroll search: failed to run scroll search query: context canceled"},
 		},
 	})
 
--- a/.github/workflows/benchmarks-matrix.yml
+++ b/.github/workflows/benchmarks-matrix.yml
@@ -1,0 +1,65 @@
+name: benchmarks-matrix
+
+on:
+  workflow_dispatch:
+    inputs:
+      profilesJSON:
+        description: 'JSON list of system profiles for benchmark, e.g. ["system-profiles/8GBx1zone.tfvars"]'
+        required: true
+        type: string
+      essRegionsJSON:
+        description: 'JSON list of ESS regions, e.g. ["gcp-us-west2", "azure-westus2", "us-west-2"]'
+        required: true
+        type: string
+      runStandalone:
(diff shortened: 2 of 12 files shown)


endorama:
Just noted your test run failed. Tests that failed where:
- [Standalone-to-managed tests (Managed7)](https://github.com/elastic/apm-server/actions/runs/15603488719/job/43950047732#logs): due to a network error when contacting Elasticsearch
- [Upgrade tests (8.15, 8.16, 8.17, Reroute)](https://github.com/elastic/apm-server/actions/runs/15603488719/job/43950052952): not clear, the GitHub UI just says it failed

Both errors don't seem related to this change. Worth retrying them before merging.

ericywl:
Made some last minute changes, will rerun the test again.

carsonip:
fyi the change in https://github.com/elastic/elasticsearch/pull/128913 is going to be backported to 8.19. This means you should expect a change in template between "8.18 and 8.19" and "9.0 and 9.1"

ericywl:
Thanks for the heads up. I will revert the changes then.

integrationservertest/logs_filters.go
```diff
@@ -102,12 +102,18 @@ var (
...
 
-	syncSourcemapContextCanceled = apmErrorLog(types.Query{
+	initialSearchQueryContextCanceled = apmErrorLog(types.Query{
 		MatchPhrase: map[string]types.MatchPhraseQuery{
 			"message": {Query: "failed to sync sourcemaps metadata: failed to run initial search query: fetcher unavailable: context canceled"},
 		},
 	})
 
+	scrollSearchQueryContextCanceled = apmErrorLog(types.Query{
+		MatchPhrase: map[string]types.MatchPhraseQuery{
```
endorama, line 112: Should we think of excluding all "context canceled" logs that we observe? I expect there are multiple occasions where this happens on shutdown and filtering them one by one manually seems to require a lot of manual intervention. This is for a follow-up, worth a separate issue.

ericywl: Yup, "context canceled" (and maybe even "context deadline exceeded") could be excluded in a follow up.

endorama (approved):
@ericywl I would have preferred if you closed this PR instead of renaming it with a complete different scope, this may create some confusion (think already sent notifications/emails).


Said that let's merge this. I'd need you to clarify if with the last comment from Carson you reverted the changes because not relevant or because you plan to introduce them in another PR. My understanding is that with the template change there would be a rollover in 8.19, but maybe this already happens so it's already covered?

ericywl:
> I would have preferred if you closed this PR instead of renaming it with a complete different scope, this may create some confusion (think already sent notifications/emails).

Got it, will take not for future PRs.

> I'd need you to clarify if with the last comment from Carson you reverted the changes because not relevant or because you plan to introduce them in another PR. My understanding is that with the template change there would be a rollover in 8.19, but maybe this already happens so it's already covered?

Once the `apm-data` changes get backported to 8.19, both 8.19 and 9.1 will have the same templates (which is our existing assumption), so no change needed.

PR #17220, opened 2025-06-12
ci: Fix benchmarks-matrix permissions, Remove ess region but add per-CSP profiles

carsonip:
## Motivation/summary

Fix error
```
 Invalid workflow file: .github/workflows/benchmarks-matrix.yml#L50
The workflow is not valid. .github/workflows/benchmarks-matrix.yml (Line: 50, Col: 3): Error calling workflow 'elastic/apm-server/.github/workflows/benchmarks.yml@e95bb60f69efe154829c59455875ba5d63776810'. The nested job 'benchmarks' is requesting 'contents: write, id-token: write', but is only allowed 'contents: read, id-token: none'.
```

Add system profiles to test different CSPs instead of using ess region because instance sizes are not the same across CSPs.

## Checklist

## Related issues

Follow up from #17211


testing/benchmark/system-profiles/16GBx2zone-azure.tfvars
```diff
@@ -0,0 +1,24 @@
...
+ess_region = "azure-westus2"
+deployment_template = "azure-cpu-optimized"
+# The number of AZs the APM Server should span.
+apm_server_zone_count = 1
+# The Elasticsearch cluster node size.
+elasticsearch_size = "120g"
+# The number of AZs the Elasticsearch cluster should have.
+elasticsearch_zone_count = 2
+# APM server instance size
+apm_server_size = "15g"
```
isaacaflores2, line 17: is this 15g intentional?

testing/benchmark/system-profiles/32GBx2zone-azure.tfvars
```diff
@@ -0,0 +1,26 @@
...
+# The number of AZs the APM Server should span.
+apm_server_zone_count = 1
+# The Elasticsearch cluster node size.
+elasticsearch_size = "240g"
+# The number of AZs the Elasticsearch cluster should have.
+elasticsearch_zone_count = 2
+# Run the cluster with a dedicated master
+elasticsearch_dedicated_masters = true
+# APM server instance size
+apm_server_size = "30g"
```
isaacaflores2, line 19: Is 30g also intentional? Just double checking since the file name and other regions specify 32g?

carsonip: Azure instance sizes are strange (which is also exactly the reason for this PR), they are 1,2,4,8,15,30,... while GCP and AWS are 1,2,4,8,16,32,...

testing/benchmark/system-profiles/16GBx2zone-azure.tfvars
rubvs: Yes, Azure works with `15g` intervals, while GCP and AWS with `16g`

rubvs: Yes, Azure works with `15g` intervals, while GCP and AWS with `16g`

rubvs (approved):

simitt:
SNAPSHOT testing would still be reasonable for the regions where we can test non-stable versions in case we want to re-run to test larger changes before a release. But overall the split into dedicated per-CSP profiles is a good improvement.

kruskall:
@mergifyio backport 8.19

PR #17293, opened 2025-06-19
fix: close default apm tracer

kruskall:
## Motivation/summary

the default tracer is leaking and its background goroutine spamming requests to this apm server (default endpoint) If TLS is enabled it causes 'http request sent to https endpoint'.

Close the default tracer since it's not used.

## Checklist

## How to test these changes

- run apm-server with tls enabled
- enable self instrumentation
- check the message is not logged on startup

### DIFF
--- a/internal/beater/beater.go
+++ b/internal/beater/beater.go
@@ -124,6 +124,12 @@ type RunnerParams struct {
 func NewRunner(args RunnerParams) (*Runner, error) {
 	fips140.CheckFips()
 
+	// the default tracer is leaking and its background
+	// goroutine is spamming requests to this apm server (default endpoint)
+	// If TLS is enabled it causes "http request sent to https endpoint".
+	// Close the default tracer since it's not used.
+	apm.DefaultTracer().Close()
+
 	var unpackedConfig struct {
 		APMServer  *agentconfig.C        `config:"apm-server"`
 		Output     agentconfig.Namespace `config:"output"`


endorama:
If I understand correctly this has no impact because the default tracer is never actually used?

> 'http request sent to https endpoint'.

This error was bugging me a lot, thank you for finding out the cause and fixing it!

carsonip (commented):
Thanks for digging into this. If go agent is leaking, should it be considered a go agent bug, and could you open an issue over there and attach a link to the issue in code?

carsonip:

## Testing notes

### 9.0.3

Checked out repo with tag v9.0.3, in apm-server.yml, set `apm-server.ssl.enabled: true` and set `apm-server.ssl.certificate` and `apm-server.ssl.key`, and set `instrumentation.enabled: true`.

Error log appears:
```
{"log.level":"error","@timestamp":"2025-07-11T16:32:02.478+0100","log.logger":"beater.http","log.origin":{"function":"net/http.(*Server).logf","file.name":"http/server.go","file.line":3638},"message":"http: TLS handshake error from 127.0.0.1:58532: client sent an HTTP request to an HTTPS server","service.name":"apm-server","ecs.version":"1.6.0"}
```

### 9.0.4 BC

Check out commit [40666db476271a74600b77917a6a8275551cbc64](https://github.com/elastic/apm-server/commits/40666db476271a74600b77917a6a8275551cbc64), start with the same config.

No error log `client sent an HTTP request to an HTTPS server`.

PR #17312, opened 2025-06-19
add integration server tests for resources version checks

endorama:
## Motivation/summary

Add additional integrationservertest for a recent bug affecting some upgrade paths in APM Server.

## Checklist

### DIFF
--- a/integrationservertest/apm-version-bug_test.go
+++ b/integrationservertest/apm-version-bug_test.go
@@ -11,6 +11,12 @@ import (
 )
 
 // See https://github.com/elastic/ingest-dev/issues/5701
+// Available tests:
+// TestAPMResourcesVersionBug/8.16.2_to_8.17.8
+// TestAPMResourcesVersionBug/8.17.7_to_8.17.8
+// TestAPMResourcesVersionBug/8.17.8_to_8.18.3
+// TestAPMResourcesVersionBug/8.18.2_to_8.18.3
+// TestAPMResourcesVersionBug/8.18.3_to_8.19.0
 func TestAPMResourcesVersionBug(t *testing.T) {
 	config, err := parseConfig("upgrade-config.yaml")
 	if err != nil {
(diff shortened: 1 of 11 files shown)


integrationservertest/apm-version-bug_test.go
```diff
@@ -0,0 +1,180 @@
...
+				steps[2], // upgrade
+				steps[3], // ingest
+				zeroEventIngestedDocs,
+				noEventIngestedInMappings,
+				steps[4], // upgrade
+				steps[5], // upgrade
+				someEventIngestedDocs,
+				eventIngestedHasMappings,
+			}
+		}
```
ericywl: This is definitely quite flimsy, I think having a way to tell when 1 "cycle" (create / upgrade -> ingest -> others) is done would be best. Perhaps having the cycles manually defined e.g. 
```
[]testCycle{
    {
        createStep,
        ingestStep,
    },
    {
        upgradeStep,
        ingestStep,
     },
     ...
}
```

Or some other way to detect it.

integrationservertest/apm-version-bug_test.go
```diff
@@ -0,0 +1,180 @@
...
+	}
+
+	// the event.ingested change was introduced in 8.16.3, as
+	// it only shows if we upgrade a cluster from a version with
+	// the bug we need to start from 8.16.2.
+	start := ech.NewVersion(8, 16, 2, "SNAPSHOT")
+
+	// this wraps the usual steps build to add additional checks in between.
+	// Is a bit brittle, maybe we should consider allowing this as a first class citizen.
+	buildteststepsForBug := func(t *testing.T, versions ech.Versions, config upgradeTestConfig) []testStep {
```
ericywl:
```suggestion
	buildTestStepsForBug := func(t *testing.T, versions ech.Versions, config upgradeTestConfig) []testStep {
```

integrationservertest/apm-version-bug_test.go
```diff
@@ -0,0 +1,180 @@
...
+		to := vsCache.GetLatestSnapshot(t, "8.18")
+		versions := []ech.Version{from, to}
+		runner := testStepsRunner{
+			Target: *target,
+			Steps:  buildTestSteps(t, versions, config, false),
+		}
+		runner.Run(t)
+	})
+
+	t.Run("8.18.2 to 8.18.3", func(t *testing.T) {
```
ericywl, line 169: If this includes `start`, shouldn't it be "8.16.2 to 8.17.x to 8.18.x"? Also `vsCache.GetLatestSnapshot` gets the latest patch for the minor if you specify `x.y`, so the description will be inaccurate when new patches come out.

integrationservertest/buglib.go
```diff
@@ -0,0 +1,152 @@
...
+		}
+		// fmt.Println("fieldMap", fieldMap)
+		subProps := fieldMap.Properties
+		current = subProps
+	}
+	return false
+}
+
+type checkMappingStep struct {
+	datastreamname string
```
ericywl, line 46:
```suggestion
	dataStreamName string
```

integrationservertest/setup.go
```diff
@@ -175,15 +175,28 @@ func createCluster(
...
 
-	t.Cleanup(func() {
-		if !t.Failed() || cleanupOnFailure {
-			t.Log("cleanup terraform resources")
-			require.NoError(t, tf.Destroy(ctx, ecTarget, ecRegion, ecDeploymentTpl, name, ver))
-		} else {
-			t.Log("test failed and cleanup-on-failure is false, skipping cleanup")
-		}
-	})
+	if *doCleanup {
```
ericywl, line 178: I don't think this works. The `doCleanup` is defined in `main_test.go` and so is only accessible in test files.

endorama: It does if you run the tests 😅 It's a quick hack for my use case.

Do you think there is value in bringing this to `main`? If yes I'll do a proper implementation.

integrationservertest/apm-version-bug_test.go
endorama: I named this after our test cases for this bug. I don't think we want to merge this PR as it is, with all these test cases for all these versions. This is now useful to ensure our fixes are working but I would make it more general if we want to include it as our regular testing.

I think a necessary test case to add is an upgrade test: from the "previous previous" minor (ie in this case 8.16.2) to the previous minor and than to the current minor. Such a test could find issues that only appears with an upgrade, a scenario we don't yet cover.

integrationservertest/apm-version-bug_test.go
endorama: You are suggesting to not use the `buildTestStep` function but create them manually so it's easier to follow the various cycles?

ericywl: My bad I wasn't clear enough. Creating them manually works, but we can also change how steps are defined in the runner. In the example above, `testCycle` is essentially just: `type testCycle []testStep`, and demarcates where one version ends and the other starts.

In your case, since you are inserting the steps in between each "cycle", it would be much less flimsy since you can inject the steps at the end of each cycle.

Just an idea, not sure if it would work well.

integrationservertest/setup.go
ericywl: TIL, my IDE complains that `doCleanup` doesn't exist 🤔

integrationservertest/apm-version-bug_test.go
endorama: Ah I see, I refactored it in a way that I think removes some of the ifs, but did not address this issue directly in [`9874d2d` (#17312)](https://github.com/elastic/apm-server/pull/17312/commits/9874d2d831637979d269d06379abe55530a41bb8)

I think there is value in doing something you propose if we decide is something we want to introduce but we should take some time to think the feature through, here I'm only interested in a quick way to have additional steps (we could also extract `createStep` and `upgradeStep` "builder" functions from the current one, so they are easier to compose, but I considered it out of scope here)

integrationservertest/setup.go
endorama: It's correct I'd say, mine too. I'd expect it's because it doesn't exist if you're running/compiling the go module 😅 it exists if you run the tests because `_tests.go` files are included in the compilation.

integrationservertest/apm-version-bug_test.go
```diff
@@ -128,12 +128,12 @@ func TestAPMResourcesVersionBug(t *testing.T) {
...
 
-	version8162 := ech.NewVersion(8, 16, 2, "SNAPSHOT")
-	version8177 := ech.NewVersion(8, 17, 7, "SNAPSHOT")
-	version8178 := ech.NewVersion(8, 17, 8, "SNAPSHOT") // latest 8.17
-	version8182 := ech.NewVersion(8, 18, 2, "SNAPSHOT")
-	version8183 := ech.NewVersion(8, 18, 3, "SNAPSHOT") // latest 8.18
-	version8190 := ech.NewVersion(8, 19, 0, "SNAPSHOT") // latest 8.19
+	version8162 := vsCache.GetLatestVersionOrSkip(t, "8.16.2")
+	version8177 := vsCache.GetLatestVersionOrSkip(t, "8.17.7")
+	version8178 := vsCache.GetLatestBCOrSkip(t, "8.17.8") // latest 8.17 as of today
```
ericywl, line 133: This will skip the whole test entirely if the BC does not exist. Since both `GetLatestVersionOrSkip` and `GetLatestBCOrSkip` are not used currently, you can update them to achieve what you need instead.

endorama:
Closing as this dedicated to a specific debug session.

PR #17376, opened 2025-06-25
feat: bump beats and drop global beat registries

kruskall:
## Motivation/summary

replace global registries with instance-scoped registries
update for latest beats changes

## Checklist

### DIFF
--- a/integrationservertest/upgrade-config.yaml
+++ b/integrationservertest/upgrade-config.yaml
@@ -13,7 +13,6 @@ lazy-rollover-with-exceptions:
   8.17:
   8.19:
   9.0:
-  # 8.19 and 9.1 have the same template due to
-  # https://github.com/elastic/elasticsearch/pull/128913.
   9.1:
-    - "8.19"
+  9.2:
+    - "9.1"


internal/beatcmd/beat.go
```diff
@@ -317,7 +318,12 @@ func (b *Beat) Run(ctx context.Context) error {
 	var apiServer *api.Server
 	if b.Config.HTTP.Enabled() {
 		var err error
-		apiServer, err = api.NewWithDefaultRoutes(b.Info.Logger, b.Config.HTTP, api.NamespaceLookupFunc())
+		apiServer, err = api.NewWithDefaultRoutes(b.Info.Logger, b.Config.HTTP,
+			b.Monitoring.InfoRegistry(),
+			b.Monitoring.StateRegistry(),
+			b.Monitoring.StatsRegistry(),
+			b.Monitoring.InputsRegistry(),
```
simitt, line 325: Where is the `InputsRegistry()` used?

simitt (commented):
Generally looks good, just a minor comment. Have you manually tested this change and confirmed that behaviour is as expected? Since this brings in some good amount of libbeat refactoring a manual test through would be good.

internal/beatcmd/beat.go
kruskall: it's used by beats, we don't use it on our side

kruskall:
we've a lot of test coverage for beat monitoring since the switch to otel metrics, I think that's already good enough imo

kruskall:
@mergifyio backport 9.1 9.0 8.19

Issue #17342, opened 2025-06-23
monitoring: TBS metrics are not reported together


isaacaflores2:
**APM Server version** (`apm-server version`): 9.0, 8.19

**Description of the problem including expected versus actual behavior**:

In a prior issue https://github.com/elastic/apm-server/issues/17104 the reporting interval for the following metrics was fixed`apm-server.sampling.tail.storage.lsm_size` and `apm-server.sampling.tail.storage.value_log_size`.  

But the second problem with all sampling configs never being reported was not addressed. 

> Also, `apm-server.sampling.tail.storage.lsm_size` and `apm-server.sampling.tail.dynamic_service_groups` are never reported together in 9.0.

Testing locally this is caused because both sets of metrics use the same namespace `apm-server.sampling` but they are created using different  instances of a Meter (`meterProvider.Meter()`). Changing `apm-server.sampling.tail.storage.lsm_size` to `apm-server.storage.lsm_size` allow both to be reported. 
<details>
<summary>Example stats </summary>

```json
{
	"apm-server": {
		"agentcfg": {
			"elasticsearch": {
				"cache": {
					"entries": {
						"count": 0
					},
					"refresh": {
						"successes": 1
					}
				}
			}
		},
		"otlp": {
			"grpc": {
				"logs": {
					"request": {
						"count": 1
					},
					"response": {
						"count": 1,
						"valid": {
							"count": 1
						}
					}
				},
				"metrics": {
					"request": {
						"count": 1
					},
					"response": {
						"count": 1,
						"valid": {
							"count": 1
						}
					}
				},
				"traces": {
					"request": {
						"count": 3
					},
					"response": {
						"count": 3,
						"valid": {
							"count": 3
						}
					}
				}
			}
		},
		"processor": {
			"error": {
				"transformations": 1
			},
			"log": {
				"transformations": 2
			},
			"metric": {
				"transformations": 2
			}
		},
		"sampling": {
			"tail": {
				"dynamic_service_groups": 1,
				"events": {
					"processed": 3,
					"stored": 3
				}
			}
		},
		"storage": {
			"lsm_size": 831932,
			"value_log_size": 0
		}
	},
	"libbeat": {
		"config": {
			"module": {
				"running": 0,
				"starts": 0,
				"stops": 0
			},
			"reloads": 0,
			"scans": 0
		},
		"output": {
			"events": {
				"acked": 5,
				"active": 0,
				"batches": 1,
				"failed": 0,
				"toomany": 0,
[...]

PR #17427, opened 2025-06-30
tbs: update metric creation to avoid conflicts with name prefix 'apm-server.sampling'

isaacaflores2:
## Motivation/summary

Currently `apm-server.sampling.tail.storage.lsm_size` and `apm-server.sampling.tail.dynamic_service_groups` are never reported together. 

Testing locally this is caused because both sets of metrics use the same namespace `apm-server.sampling` but they are created using different instances of a Meter.  

I opted to create a common instance of the meter to avoid renaming the metrics. This seems to not align with the  [docs](https://pkg.go.dev/go.opentelemetry.io/otel/metric@v1.37.0#MeterProvider.Meter) around a Meter but I do not think we want to change the metric names. Let me know if you feel otherwise

## Checklist

- [ ] View individual metrics documents to confirm reported metrics are correct

## How to test these changes

1. Run apm-server with TBS enabled
2. Send data to the server `./sendotlp -insecure -endpoint=http://localhost:8200 -secret-token=<token>`
3. View metrics via the stats endpoint (http://localhost:5066/stats?pretty) . `storage` should be visible along with other metrics under `apm-server.sampling.tail`

```json
"sampling": {
	"tail": {
		"dynamic_service_groups": 1,
		"events": {
			"processed": 3,
			"stored": 3
		},
		"storage": {
			"lsm_size": 7572,
			"value_log_size": 0
		}
	}
}
```

## Related issues

Closes https://github.com/elastic/apm-server/issues/17342

### DIFF
--- a/x-pack/apm-server/main.go
+++ b/x-pack/apm-server/main.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/gofrs/uuid/v5"
 	"go.opentelemetry.io/otel/metric"
+	"go.opentelemetry.io/otel/metric/noop"
 	"golang.org/x/sync/errgroup"
 
 	"github.com/elastic/beats/v7/libbeat/common/reload"
@@ -107,8 +108,17 @@ func newTailSamplingProcessor(args beater.ServerParams) (*sampling.Processor, er
 		return nil, fmt.Errorf("failed to create Elasticsearch client for tail-sampling: %w", err)
 	}
 
+	// create a common meter for the tail-sampling processor
+	// so multiple metrics with the prefix 'apm-server.sampling'
+	// in the name can be created without conflicts
+	// and without renaming them.
+	if args.MeterProvider == nil {
+		args.MeterProvider = noop.NewMeterProvider()
+	}
+	samplingMeter := args.MeterProvider.Meter("github.com/elastic/apm-server/x-pack/apm-server")
+
 	storageDir := paths.Resolve(paths.Data, tailSamplingStorageDir)
-	db, err := getDB(storageDir, tailSamplingConfig.DatabaseCacheSize, args.MeterProvider, args.Logger)
+	db, err := getDB(storageDir, tailSamplingConfig.DatabaseCacheSize, samplingMeter, args.Logger)
 	if err != nil {
 		return nil, fmt.Errorf("failed to get tail-sampling database: %w", err)
 	}
@@ -128,7 +138,7 @@ func newTailSamplingProcessor(args beater.ServerParams) (*sampling.Processor, er
 
 	return sampling.NewProcessor(sampling.Config{
 		BatchProcessor: args.BatchProcessor,
-		MeterProvider:  args.MeterProvider,
+		Meter:          samplingMeter,
 		LocalSamplingConfig: sampling.LocalSamplingConfig{
 			FlushInterval:         tailSamplingConfig.Interval,
 			MaxDynamicServices:    1000,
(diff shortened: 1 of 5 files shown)


kruskall:
> I opted to create a common instance of the meter to avoid renaming the metrics. This seems to not align with the [docs](https://pkg.go.dev/go.opentelemetry.io/otel/metric@v1.37.0#MeterProvider.Meter) around a Meter but I do not think we want to change the metric names

I think we shuld investigate this more, the two meters were scoped to different packages. I don't see why the metrics would conflict. Am I missing something ?

isaacaflores2:
> I think we shuld investigate this more, the two meters were scoped to different packages. I don't see why the metrics would conflict. Am I missing something ?

I am not sure why the conflict exists. I will investigate further to find the exact cause and to see if we can keep the meters scoped to two different packages

isaacaflores2:
@kruskall Update:  I do not think it this is OTEL related. It seems to be `beat` related.  

As you probably know the metrics are registered with beat here: https://github.com/elastic/apm-server/blob/main/internal/beatcmd/beat.go#L543.  It appears that `addAPMServerMetrics` is somehow overwriting an existing key.  I am not exactly sure why it is doing that yet but I setup the below test to reproduce the issue. I'll spend some time understanding the snapshot logic in elastic-agent-lib


```go
func TestMonitoringApmServerMetrics(t *testing.T) {
	r := monitoring.NewRegistry()

	exporter, err := apmotel.NewGatherer()
	if err != nil {
		t.Logf("Unable to create Gatherer: %v", err)
	}

	metricReader := sdkmetric.NewManualReader()
	meterProvider := sdkmetric.NewMeterProvider(
		sdkmetric.WithReader(exporter),
		sdkmetric.WithReader(metricReader),
	)
	otel.SetMeterProvider(meterProvider)

	// Add metrics similar to lsm_size in storage_manager.go and events.processed in processor.go
	meter := meterProvider.Meter("github.com/elastic/apm-server/x-pack/apm-server/sampling/eventstorage")
	lsmSizeGauge, _ := meter.Int64Gauge("apm-server.sampling.tail.storage.lsm_size")
	lsmSizeGauge.Record(context.Background(), 123)

	meter2 := meterProvider.Meter("github.com/elastic/apm-server/x-pack/apm-server/sampling")
	processedCounter, _ := meter2.Int64Counter("apm-server.sampling.tail.events.processed")
	processedCounter.Add(context.Background(), 456)

	meter3 := meterProvider.Meter("github.com/elastic/apm-server/x-pack/apm-server/foo")
	otherCounter, _ := meter3.Int64Counter("apm-server.foo.request")
	otherCounter.Add(context.Background(), 1)

	monitoring.NewFunc(r, "apm-server", func(_ monitoring.Mode, v monitoring.Visitor) {
		var rm metricdata.ResourceMetrics
		if err := metricReader.Collect(context.Background(), &rm); err != nil {
			return
		}
		v.OnRegistryStart()
		defer v.OnRegistryFinished()
		for _, sm := range rm.ScopeMetrics {
			switch {
			case strings.HasPrefix(sm.Scope.Name, "github.com/elastic/apm-server"):
				// All simple scalar metrics that begin with the name "apm-server."
				// in github.com/elastic/apm-server/... scopes are mapped directly.
				addAPMServerMetrics(v, sm)
			}
		}
	})

	snapshot := monitoring.CollectStructSnapshot(r, monitoring.Full, false)
	// Assert that the snapshot contains the expected metrics and values
	assert.Equal(t, map[string]any{
		"foo": map[string]any{
			"request": int64(1),
		},
		"sampling": map[string]any{
			"tail": map[string]any{
				"storage": map[string]any{
					"lsm_size": int64(123),
				},
				"events": map[string]any{
					"processed": int64(456),
				},
			},
		},
	}, snapshot)
}

```

carsonip:
@isaacaflores2 is it intentional to not backport to 9.0?

isaacaflores2:
> @isaacaflores2 is it intentional to not backport to 9.0?

Oh it was not intentional. I just missed it. I added the 9.0 backport

carsonip (commented):
any possible way that we can verify this bugfix with a test?

isaacaflores2:
Closing this PR since we would like to keep meters scoped to different packages and alternate solution has been found.

PR #17512, opened 2025-07-07
monitoring: update apm-server metrics collection to avoid conflicts

isaacaflores2:
## Motivation/summary

Currently `apm-server.sampling.tail.storage.lsm_size` and `apm-server.sampling.tail.dynamic_service_groups` are never reported together. 

Testing locally this is caused because both sets of metrics use the same namespace (metric name prefix) `apm-server.sampling` but they are created using different instances of a Meter.  

The related monitoring func calls `addAPMServerMetrics` multiple times for each scoped metric. Metric names with the same prefix in different scopes are somehow overwriting each other. This approach opts to collect all "apm-server" metrics and add them to the snapshot once.  Another approach would be to update the `elastic-agent-lib` to prevent metrics from overwriting each other.

## Checklist

- [ ] View individual metrics documents to confirm reported metrics are correct

## How to test these changes
1. Unit test cover the expected behavior

### manual test
1. Run apm-server with TBS enabled
2. Send data to the server `./sendotlp -insecure -endpoint=http://localhost:8200 -secret-token=<token>`
3. View metrics via the stats endpoint (http://localhost:5066/stats?pretty) . `storage` should be visible along with other metrics under `apm-server.sampling.tail`

```json
"sampling": {
		"tail": {
			"dynamic_service_groups": 0,
			"events": {
				"processed": 3,
				"sampled": 6,
				"stored": 3
			},
			"storage": {
				"lsm_size": 7431,
				"value_log_size": 0
			}
		}
	}
```

## Related issues

Closes https://github.com/elastic/apm-server/issues/17342
Alternate approach to https://github.com/elastic/apm-server/pull/17427


rubvs:
Quickly tested this manually. 

- Stats before the PR change
```json
"sampling": {
  "tail": {
    "dynamic_service_groups": 1,
    "events": {
      "processed": 3,
      "stored": 3
    }
  }
}
```

- Stats after the PR change
```json
"sampling": {
  "tail": {
    "dynamic_service_groups": 1,
    "events": {
      "processed": 3,
      "sampled": 3,
      "stored": 3
    },
    "storage": {
      "lsm_size": 5407,
      "value_log_size": 0
    }
  }
}
```

- Cat of APM Server config
```yaml
apm-server:
  host: "127.0.0.1:8200"

  sampling.tail:
    enabled: true
    policies:
      - sample_rate: 1.0

output.elasticsearch:
  hosts: ["https://216c4d53f99e4e48a556e4cfe72cb680.us-central1.gcp.qa.cld.elstc.co:443"]
  username: "elastic"
  password: "<REDACTED>"

http:
  enabled: true
  host: localhost
  port: 5066
```

rubvs (approved):

internal/beatcmd/beat.go
```diff
@@ -528,22 +528,41 @@ func (b *Beat) registerStatsMetrics() {
 			}
 		}
 	})
-	monitoring.NewFunc(monitoring.Default, "apm-server", func(_ monitoring.Mode, v monitoring.Visitor) {
+	monitoring.NewFunc(monitoring.Default, "apm-server", apmServerMonitoringFunc(b.metricReader))
+}
+
+// apmServerMonitoringFunc returns a monitoring function that tracks apm-server metrics.
+//
+// Note: Unit test with the global otel meter provider are difficult to test due to interference from other test cases or
```
kruskall: Q: does this still applies ? lastest beats version moved away from global registries and we should never use global otel providers ?

internal/beatcmd/beat.go
```diff
@@ -528,22 +528,41 @@ func (b *Beat) registerStatsMetrics() {
...
 		var rm metricdata.ResourceMetrics
-		if err := b.metricReader.Collect(context.Background(), &rm); err != nil {
+		if err := metricReader.Collect(context.Background(), &rm); err != nil {
 			return
 		}
 		v.OnRegistryStart()
 		defer v.OnRegistryFinished()
+
+		// first collect all apm-server metrics
+		apmServerMetrics := make([]metricdata.Metrics, 0)
```
kruskall: Q: do we need to collect them ? AFAICS the slice is used to temporarily stored the metrics, we want to ensure the map (beatsMetrics) contains all the metrics before passing them to the registry

internal/beatcmd/beat.go
isaacaflores2: I think it will still apply, but I will pul the latest changes to check again. When I say global otel provider I mean this line here: https://github.com/elastic/apm-server/blob/main/internal/beatcmd/beat.go#L133. 

The issue I had with the test is that the monitoring func for say "output.elasticsearch" would call [collect](https://github.com/elastic/apm-server/blob/main/internal/beatcmd/beat.go#L531) and read the apm-server metrics before the "apm-server" monitoring func was able to collect them.

internal/beatcmd/beat.go
isaacaflores2: Yup this update fixes the conflict by first collecting the apm-server metrics then adding them to the registry once. Otherwise for every ScopeMetric iteration when we call  `addAPMServerMetrics`  we would have conflicts. So if you had the following metrics in two different scopes: `apm-server.sampling.tail.storage.lsm_size` and `apm-server.sampling.tail.events.processed`. The last one would overwrite since they have the same name prefix `apm-server.sampling`

internal/beatcmd/beat.go
isaacaflores2: You were right, it no longer applies. I refactored to removed that helper func and updated the unit test

internal/beatcmd/beat.go
kruskall: > Yup this update fixes the conflict by first collecting the apm-server metrics then adding them to the registry once. 

That's a good idea but right now we're doing:

collect -> slice -> map -> registry

Can't we avoid the extra slice ? We could add them to the map and only publish to the registry after the loop

isaacaflores2: Oh sorry I misunderstood your initial question. Good point the slice is redundant. I should be able to update to collect the metrics in a map instead

carsonip (approved):
lgtm, thanks for adding the test

internal/beatcmd/beat.go
isaacaflores2: Updated to remove the temp slice

internal/beatcmd/beat.go
```diff
@@ -547,14 +547,22 @@ func (b *Beat) registerStatsMetrics() {
...
 				// in github.com/elastic/apm-server/... scopes are mapped directly.
-				addAPMServerMetrics(v, sm)
+				addAPMServerMetricsToMap(beatsMetrics, sm.Metrics)
 			}
 		}
+
+		// register all metrics once
+		// this prevents metrics with the same prefix in the name
+		// from different scoped meters from overwriting each other
+		reportOnKey(v, beatsMetrics)
```
simitt, line 565: Great finding!

carsonip:

## Testing notes

Tested with 9.0.4 BC on ECH. lsm_size and dynamic_service_groups appear in the same monitoring metric document. It does not happen on 9.0.3.

image

PR #17571, opened 2025-07-11
iserver-test: Fix BC 9.0 test

ericywl:
## Motivation/summary

8.19 cannot upgrade to 9.0, so our BC test for 9.0 will always fail. This PR adds an edge case handling for 9.0, and also increase the APM server ready wait time.

## How to test these changes

Run workflow: https://github.com/elastic/apm-server/actions/runs/16211738588


.github/workflows/generate-bc-upgrade-paths/bc.sh
```diff
@@ -15,6 +15,12 @@ for bc in ${bcs}; do
     if [[ ${major} -lt 8 ]]; then
         continue
     fi
+    # Specifically for 9.0, we set the upgrade path to be from 8.18,
+    # since 8.19 cannot upgrade to 9.0.
```
marclop: Why's that? Is there a legitimate reason for that?

ericywl: Its due to ES upgrade policy, upgrades can only be done in chronological order. Since latest 9.0 was released before latest 8.19, it cannot be upgraded to.

ES issue: https://elasticco.atlassian.net/browse/CP-10254.

marclop: But 8.19 can be upgraded to 9.1?

ericywl: Yes.

marclop: Can you add this information in the comment, please? I'm sure others may have the same questions.

PR #17575, opened 2025-07-11
ironbank: remove tmpl extension

v1v:
## Motivation/summary

Use the same names as in main since backports won't work for the UBI releases automatically after https://github.com/elastic/apm-server/pull/8732

## Checklist


v1v:
The BK pipeline failed since it used the feature branch and we only support release branches

image

Issue #17602, opened 2025-07-14
integrationservertest: Automate upgrade paths


ericywl:
Current upgrade paths are defined manually in the YAML. We can instead automate it by querying active versions API, and selecting previous versions (with some strategy?) for upgrade paths. This will work similarly to the BC testing: https://github.com/elastic/apm-server/blob/main/.github/workflows/integration-server-test.yml#L77.

PR #17709, opened 2025-07-17
iserver-test: Add 9.2 test and some improvements

ericywl:
## Motivation/summary

- Add 9.2 upgrade paths manually, since we don't have https://github.com/elastic/apm-server/issues/17602 yet
- Add input option in workflow dispatch to control which tests to run
- Add retry for `telemetrygen` in case of connection issues during ingestion
- Remove flush trigger since it seems to interfere with metrics aggregation sometimes

## How to test these changes

Run workflow: https://github.com/elastic/apm-server/actions/runs/16346566007


.github/workflows/integration-server-test.yml
```diff
@@ -31,18 +47,17 @@ jobs:
         uses: ./.github/workflows/generate-bc-upgrade-paths
 
   run-upgrade:
-    name: Upgrade tests (Snapshot)
+    if: ${{ inputs.run-upgrade-tests == true }}
```
v1v: I recall some edge cases when using `workflow_dispatch` default inputs and the `schedule`. Can you double-check if the default values are honoured when the schedule event runs?

I did run a quick search and found this [StackOverflow thread](https://stackoverflow.com/a/73495922) or [this](https://github.com/orgs/community/discussions/137750)

ericywl: I see, will try another way to do it instead. Thanks for the heads up!

simitt (approved):

PR #17753, opened 2025-07-21
iserver-test: Auto generate upgrade path

ericywl:
## Motivation/summary

Introduce new `generate-upgrade-paths` composite action to generate upgrade paths for both SNAPSHOTs and BCs testing. The upgrade paths are randomly generated by picking some amount of previous versions for each active version. For example, given versions `8.18` and `8.19`, the script may pick `8.16 -> 8.18` and `8.18 -> 8.19` (random). After picking the upgrade pairs, the script will then try to combine them to form as long a chain as possible. In the example above, it will form `8.16 -> 8.18 -> 8.19`, and that will be the upgrade path to be tested.

To avoid RNG not fully testing the upgrade paths that we want, the latest patches of each major version (in current case that would be `8.19` and `9.2`) will be forced to pick 1 major upgrade and 1 minor upgrade. That would mean forcing `8.19 -> 9.2` and `9.1 -> 9.2` for `9.2` and `8.18 -> 8.19` for `8.19` (notice duplicate from above, so we ignore it) to be in one of the upgrade paths generated.

If we combine our examples from above, the final generated upgrade paths would be:
- `8.16 -> 8.18 -> 8.19 -> 9.2`
- `9.1 -> 9.2`

To try out the script, export `EC_API_KEY` and run:
```
cd integrationservertest && go run ./scripts/genpath --target='pro' --snapshots=true
```

Closes https://github.com/elastic/apm-server/issues/17602.

## How to test these changes

Run workflow: https://github.com/elastic/apm-server/actions/runs/16495315885
Another run: https://github.com/elastic/apm-server/actions/runs/16589177207


.github/workflows/setup-integration-server-test/action.yml
```diff
@@ -0,0 +1,21 @@
+---
+name: setup-integration-server-test
+description: Setup Integration Server Test
+
+runs:
+  using: composite
+  steps:
+    - uses: actions/checkout@v4
```
v1v: IIUC, you don't need the `checkout` again. That should always be done in the consumers of the composite action called .`github/workflows/setup-integration-server-test`

integrationservertest/scripts/genpath/versions_bc.go
```diff
@@ -0,0 +1,115 @@
...
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+
+	"github.com/elastic/apm-server/integrationservertest/internal/ech"
+)
+
+const (
+	futureReleasesAPI = "https://artifacts.elastic.co/releases/TfEVhiaBGqR64ie0g0r0uUwNAbEQMu1Z/future-releases/stack.json"
```
v1v, line 32: Can you add a note regarding this endpoint? It might not be accessible eventually.

v1v (commented):
I've only reviewed the CI bits and pieces, and so far It looks good. I left a couple of comments.

.github/workflows/setup-integration-server-test/action.yml
ericywl: Got it, thanks.

integrationservertest/scripts/genpath/main.go
```diff
@@ -0,0 +1,313 @@
...
+	upgradeFromVersions := vsCache.GetUpgradeFromVersions(version).
+		Filter(func(v ech.Version) bool {
+			// Filter out non-SNAPSHOTs if we are testing SNAPSHOTs and vice-versa.
+			snapshotFilter := !v.IsSnapshot()
+			if *useSnapshots {
+				snapshotFilter = v.IsSnapshot()
+			}
+			// Filter out versions that don't meet our defined version cutoff.
+			// Also filter out versions that has same major-minor as current version,
+			// since we don't care about patch upgrades in this test.
```
simitt, line 157: Why do we not want to test patch upgrades? It would make sense for the actively maintained minor versions.

simitt (commented):
Need to do a more exhaustive review as there are a few changes brought in. 
Some preliminary comment for introducing this randomness:
- Specify a _why_ for choosing the random upgrade paths and list which guarantees for included versions are provided (e.g. always testing up to the latest version). Please add this to the README
- It's not guaranteed that the same path will be chosen in the next run, so please also add a section to the README for how to manually re-run tests. If something was broken and we need to ensure it's fixed, there needs to be a documented way for testing this.

integrationservertest/scripts/genpath/main.go
ericywl: Maybe "not caring" about patch upgrades was the wrong expression. We simply did not test patch upgrades both in the old smoke tests and new integration server test ever. We only did minor and major upgrade tests, so this test simply continues that trend.

ericywl:
> Specify a why for choosing the random upgrade paths and list which guarantees for included versions are provided (e.g. always testing up to the latest version). Please add this to the README

Added a section in the README about this 👍 

> It's not guaranteed that the same path will be chosen in the next run, so please also add a section to the README for how to manually re-run tests. If something was broken and we need to ensure it's fixed, there needs to be a documented way for testing this.

That should already be documented in README, essentially copy the upgrade path from the failed workflow run and use it in the documented command. Unless you mean to have the test re-run on workflow, in which case I might have to add an input to the workflow to allow running manually provided upgrade path.

integrationservertest/internal/ech/versions_cache.go
```diff
@@ -40,28 +33,32 @@ func NewVersionsCache(ctx context.Context, client *Client, ecRegion string) (*Ve
 		return nil, err
 	}
 
-	upgradeInfo := map[Version]Versions{}
-	candidates := unpackStackVersions(cs, upgradeInfo)
-	snapshots := unpackStackVersions(ss, upgradeInfo)
-	versions := unpackStackVersions(vs, upgradeInfo)
+	upgradeTo := map[Version]Versions{}
+	upgradeFrom := map[Version]Versions{}
```
simitt, line 37: Maybe add a comment here what the map keys and values are, took me a bit to understand what's the difference here.

integrationservertest/README.md
```diff
@@ -30,11 +32,41 @@ The upgrade tests reside in `upgrade_test.go`.
 These tests take an `upgrade-path` argument that represents a list of versions for the upgrade test.
 The test will create a deployment with the first version, perform ingestion and check that everything is expected.
 Then, it will consecutively upgrade to the next version, perform ingestion and check again.
-For example, if we provide an `upgrade-path` of `8.15, 8.16, 8.17`, a deployment will be created in `8.15`, upgraded to `8.16` and then to `8.17`.
+For example, if we provide an `upgrade-path` of `8.17.4 -> 8.18.0 -> 8.19.0`, a deployment will be created in `8.17.4`, upgraded to `8.18.0` and then to `8.19.0`.
 
 We provide the `upgrade-path` argument through GitHub workflow matrix, see `integration-server-test.yml`.
 Configuration for the upgrade test can be found in `upgrade-config.yaml`.
 
+#### Upgrade Path Generation
```
isaacaflores2, line 40: praise: Nice description of the upgrade path flow

isaacaflores2 (approved):

PR #17774, opened 2025-07-23
iserver-test: Add bulk indexing failure log entry to ignores

ericywl:
## Motivation/summary

Bulk indexing failure log message that shows up occasionally only happens when APM server is restarted on upgrade while metrics are still flushing. Adding it to ignore list, since it is failing the tests occasionally.

## How to test these changes

Run workflow: https://github.com/elastic/apm-server/actions/runs/16465185470


1pkg (approved):

ericywl:
Not merging for now since I'm hoping to find another way to remove the chance of getting bulk indexing failures instead of ignoring it.

Issue #17263, opened 2025-06-17
9.1 Test Plan


rubvs:
# Manual Test Plan

When picking up a test case, please add your name to this overview beforehand and tick the checkbox when finished.
Testing can be started when the first build candidate (BC) is available in the CFT region.

## Smoke Testing ESS setup

Thanks to https://github.com/elastic/apm-server/issues/8303 further smoke tests are run automatically on ESS now.
**Consider extending the smoke tests to include more test cases which we'd like to cover**

## ES apm-data plugin

- [x] https://github.com/elastic/elasticsearch/pull/119995 @isaacaflores2 
- [x] https://github.com/elastic/elasticsearch/pull/128913 @1pkg @rubvs 
- [x] https://github.com/elastic/elasticsearch/pull/129074 @rubvs 
- [x] https://github.com/elastic/elasticsearch/pull/129418 @1pkg 

## apm-aggregation

List of changes: https://github.com/elastic/apm-aggregation/compare/v1.2.0...v1.3.0

- [x] https://github.com/elastic/apm-aggregation/pull/258 @rubvs  

## go-docappender

List of changes: https://github.com/elastic/go-docappender/compare/v2.4.0...v2.10.0

- [x] https://github.com/elastic/go-docappender/pull/247 @rubvs 
- [x] https://github.com/elastic/go-docappender/pull/228 @rubvs 

## apm-data

List of changes: https://github.com/elastic/apm-data/compare/v1.16.0...v1.19.2

- [x] https://github.com/elastic/apm-data/pull/500 @rubvs 
- [x] https://github.com/elastic/apm-data/pull/492 @rubvs 
- [x] https://github.com/elastic/apm-data/pull/487 @rubvs 
- [x] https://github.com/elastic/apm-data/pull/434 @rubvs 

## apm-server

List of changes: https://github.com/elastic/apm-server/compare/v9.0.0...main

- [x] https://github.com/elastic/apm-server/pull/16549 @1pkg 
- [x] https://github.com/elastic/apm-server/pull/15976 @rubvs 
- [x] https://github.com/elastic/integrations/pull/13950 @rubvs 
- [x] https://github.com/elastic/kibana/pull/224479 @rubvs 
- [x] https://github.com/elastic/apm-server/pull/16579 @isaacaflores2 
- [x] https://github.com/elastic/apm-server/pull/16928 @1pkg

rubvs:
[...]

PR #17787, opened 2025-07-23
doc: add 9.1 changelog

rubvs:
Add 9.1.0 release notes derived from https://github.com/elastic/apm-server/issues/17263


docs/release-notes/index.md
```diff
@@ -22,6 +22,19 @@ To check for security updates, go to [Security announcements for the Elastic sta
 % ### Fixes [elastic-apm-next-fixes]
 % * 1 sentence describing the change. ([#PR number](https://github.com/elastic/apm-server/pull/PR number))
 
+## 9.1.0 [elastic-apm-9.1.0-release-notes]
+
+### Features and enhancements [elastic-apm-9.1.0-features-enhancements]
+
+* Enable `date_detection` for all apm data streams ([#128913](https://github.com/elastic/elasticsearch/pull/128913))
```
isaacaflores2: I think we ended up disabling `date_detection` and just updating the mapping for a known field in this https://github.com/elastic/elasticsearch/pull/130466 right? ( I am assuming this made it into 9.1 too)

isaacaflores2 (approved):
I think you covered everything. I did a quick scan of this [compare](https://github.com/elastic/apm-server/compare/v9.0.0...9.1) (not sure if it is the correct comparison) to double check.

docs/release-notes/index.md
rubvs: aaa you're right. We are only enabling it for `system.process.cpu.start_time`

rubvs: Also, this is a fix and not a feat. Got thrown of by the `feat:` in the PR desc. https://github.com/elastic/elasticsearch/pull/128913

docs/release-notes/index.md
```diff
@@ -22,6 +22,19 @@ To check for security updates, go to [Security announcements for the Elastic sta
...
+
+### Features and enhancements [elastic-apm-9.1.0-features-enhancements]
+
+* Add config for tail-based sampling discard on write ([#13950](https://github.com/elastic/integrations/pull/13950))
+* Add config for tail-based sampling TTL ([#16579](https://github.com/elastic/apm-server/pull/16579))
+
+### Fixes [elastic-apm-9.1.0-fixes]
+
+* Fix broken UI by explicitly enabling date detection for the `system.process.cpu.start_time` field ([#130466](https://github.com/elastic/elasticsearch/pull/130466))
+* Fix using representative count as `event.success_count` if available ([#119995](https://github.com/elastic/elasticsearch/pull/119995))
```
1pkg: `Fix using representative ...` fix what here? `ingest pipeline`?

rubvs: The fix is to use representative count for this metric, instead of 0 or 1, which is more aligned with what the metric tends to represent.

PR #17828, opened 2025-07-29
docs: qualifier

v1v:
## Motivation/summary
update docs after the recent changes from https://github.com/elastic/apm-server/pull/15350

## Checklist


rubvs:
If this is only something to consider for MAJOR, maybe we should put it under the **Major Release** header?

v1v:
> If this is only something to consider for MAJOR, maybe we should put it under the Major Release header?

I don't know, the qualifier is release agnostic but we used it only for `MAJOR` releases.

rubvs:
In that case, I'd still prefer it under Major Release, since that is the only case when we _should_ consider taking it into account. cc @carsonip for an extra take on this. We can always move it out, when we end up using it for non-majors.

v1v:
That's still part of the `Between feature freeze and release` section, regardless of the type of release.

image

PR #17975, opened 2025-08-06
Bump golang.org/x/sync to 0.16.0 to surface panic

carsonip:
## Motivation/summary

Action item from #17967

golang.org/x/sync errgroup 0.15.0 does not propagate panics immediately which makes it very hard to detect issues. This is fixed in 0.16.0. Bump dep to 0.16.0.

No changelog as there should be no change in behavior unless there is a panic.

## Checklist

## How to test these changes

Tested in #17967

Not feasible to add unit test as panic in goroutine cannot be caught unless there is implementation code change.

## Related issues

#17967


carsonip:
^ i triggered backport but accidentally deleted the comment

Issue #18152, opened 2025-08-14
autobump dependencies on 8.19 branch


kruskall:
dependabot is configured to only target the main branch but 8.19 is a "LTS" branch so it will be around for a long time

we should look into configuring auto-bump on the "old" branch to ensure we're not missing out on bug fixes

PR #18153, opened 2025-08-14
ci: add dependency bumps to 8.19 branch

kruskall:
## Motivation/summary

add a new entry in `.github/dependabot.yml` for Go module and docker images ecosystem updates targeting the "8.19" branch.

dependabot only targets main but 8.19 will keep being supported in the future so we want to bump dependencies to get bugfixes

use a weekly interval and group dependencies together to avoid too much PR noise

## Checklist

## Related issues

Closes https://github.com/elastic/apm-server/issues/18152

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -41,7 +41,8 @@ updates:
         patterns:
           - "*"
   # go dependencies 8.x branch
-  - package-ecosystem: "gomod-8.19"
+  - package-ecosystem: "gomod"
+    target-branch: "8.19"
     directory: "/"
     ignore:
       - dependency-name: "github.com/elastic/beats/v7"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -40,6 +40,28 @@ updates:
...
           - "*"
+  # go dependencies 8.x branch
+  - package-ecosystem: "gomod"
+    target-branch: "8.19"
+    directory: "/"
+    ignore:
+      - dependency-name: "github.com/elastic/beats/v7"
+    schedule:
+      interval: "weekly"
+    groups:
```
simitt, line 51: Why is this different than for `main`?

kruskall: > use a weekly interval and group dependencies together to avoid too much PR noise

PR #17872, opened 2025-07-31
fix: do not try to create apm-serve registry if it exists

kruskall:
## Motivation/summary

`apm-server` registry could already exist, causing a failure on agent reload when trying to be recreated

## Checklist

- [ ] Update [CHANGELOG.asciidoc](https://github.com/elastic/apm-server/blob/main/CHANGELOG.asciidoc)
- [ ] Documentation has been updated

For functional changes, consider:
- Is it observable through the addition of either **logging** or **metrics**?
- Is its use being published in **telemetry** to enable product improvement?
- Have system tests been added to avoid regression?

PR #18154, opened 2025-08-14
Remove last remaining NewRegistry method call

faec:
The `NewRegistry` method in `elastic-agent-libs` is inherently unsafe, especially in concurrent code: calling `NewRegistry` on a name that already exists causes a panic, but checking whether the name exists first is not atomic. The recently added `GetOrCreateRegistry` method is a safe alternative that checks the name's existence and initializes it atomically, and never panics. This PR removes the last remaining `NewRegistry` call in the apm-server repo.

## Related issues

This is a precautionary followup to https://github.com/elastic/fleet-server/issues/5170 and https://github.com/elastic/apm-server/pull/17872, where a different `NewRegistry` call site caused a panic on server reload.

### DIFF
--- a/internal/beater/beater.go
+++ b/internal/beater/beater.go
@@ -965,9 +965,12 @@ func queryClusterUUID(ctx context.Context, esClient *elasticsearch.Client, state
 	// Running under elastic-agent, the callback linked above is not
 	// registered until later, meaning we need to check and instantiate the
 	// registries if they don't exist.
-	elasticsearchRegistry := stateRegistry.GetRegistry(outputES)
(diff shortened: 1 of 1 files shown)


kruskall:
Thanks! :bow: 

older 9.x version don't have local registries so it's a bit more complicated to switch. Change LGTM!

Can you sign your commits ?

carsonip:
~~@kruskall Why remove backport-9.1? I thought it should go into both 8.19 and 9.1+~~

I see you have added it now

carsonip (approved):

Issue #17603, opened 2025-07-14
Add test for the absence of `tags` in apm documents


simitt:
## Problem Statement

We are missing system tests that execute the geo-ip pipeline and then test for it to have succeeded. Any problems with installing the geoip pipeline therefore go unnoticed. 

## Proposal
Add a check to the integration tests for the event documents to not contain any `tags`. 
Eg. with a current bug, we see following tags in the indexed documents:
```
  "tags": [
    "_geoip_database_unavailable_GeoLite2-City.mmdb"
  ],
```
If all pipeline execution succeeds, there shouldn't be any tags, so to avoid overfitting tests to this current example, we should check for the absence of any such tags.

simitt:
@raultorrecilla for prioritization - the impact for not catching such an issue is for the Geo-location functionalities in the UI to not work properly, which mainly impacts RUM usecases.

PR #8065, opened 2022-05-06
mount static geoip db into es container

stuartnelson3:
fixes flaky test that depended on the geoip db
being downloaded.

closes #7260

I suppose the way to test this would be to run it several times?

Issue #7260, opened 2022-02-11
systemtest: ensure geoIP database is available before running tests


axw:
The system tests intermittently fail, due to the geoIP database being unavailable. e.g. https://apm-ci.elastic.co/job/apm-server/job/apm-server-mbp/job/PR-7257/2/testReport/

```
=== RUN   TestErrorIngest
    errors_test.go:52:   map[string]interface{}{
          	"events": []interface{}{
          		map[string]interface{}{
          			... // 16 identical entries
          			"processor": map[string]interface{}{"event": string("error"), "name": string("error")},
          			"service":   map[string]interface{}{"environment": string("staging"), "framework": map[string]interface{}{"name": string("Express"), "version": string("1.2.3")}, "language": map[string]interface{}{"name": string("ecmascript"), "version": string("8")}, "name": string("1234_service-12a3"), ...},
        + 			"tags":      []interface{}{string("_geoip_database_unavailable_GeoLite2-City.mmdb")},
          			"timestamp": map[string]interface{}{"us": float64(1.533827045999e+15)},
          			"trace":     map[string]interface{}{"id": string("0123456789abcdeffedcba0123456789")},
          			... // 2 identical entries
          		},
          	},
          }
        
        Run `make update check-approvals` to verify the diff
        
    server.go:191: log file: /var/lib/jenkins/workspace/pm-server_apm-server-mbp_PR-7257/src/github.com/elastic/apm-server/systemtest/logs/TestErrorIngest/apm-server
--- FAIL: TestErrorIngest (1.39s)
```

We should ensure the geoIP database is available before running the tests. I'm not sure how yet. Maybe we can use the [simulate pipeline API](https://www.elastic.co/guide/en/elasticsearch/reference/current/simulate-pipeline-api.html) to force a download, and keep trying until it succeeds?

axw:
Apparently the implementation doesn't quite work: https://apm-ci.elastic.co/job/apm-server/job/apm-server-mbp/job/PR-7322/2/testReport/

axw:
I just learned of a trick from `elastic-package`, which downloads the geoIP database and mounts it into the container. We might be able to do something similar.

PR #18217, opened 2025-08-18
add rum test for geoip tags

rubvs:
Closes https://github.com/elastic/apm-server/issues/17603

This system test should fail if there are any `tags` in the indexed doc. This test ensure that if, for some reason, we cannot connect or download the `geoIP` database, the system tests will fail.

The systems test leverage docker volumes to link the geoip database, see [docker-compose](https://github.com/elastic/apm-server/blob/13e35d76bcf2c01a1491a933937f1e3ee4d16d3a/docker-compose.yml#L40). Therefore, properly testing this PR requires some manual steps.

- Either delete or comment the docker volume line linked above.
- Ensure you are pulling a clean ES image. I typically like to keep my local clean, so I'll remove all containers and images:
```shell
# Remove all running container instances
> docker rm -f $(docker ps -a -q)

# Remove all built images
> docker rmi -f $(docker images -a -q)
```
- Running `make system-test` should fail with the following error:
```shell
> make system-test

rum_test.go:94: unexpected tags field in document: [_geoip_database_unavailable_GeoLite2-City.mmdb]
```
- Repeat, but with the `ingest-geoip` volume enabled in `docker-compose.yml`, then the system test will pass.

### DIFF
--- a/systemtest/approvals/TestRUMXForwardedFor.approved.json
+++ b/systemtest/approvals/TestRUMXForwardedFor.approved.json
@@ -12,27 +12,6 @@
         "agent.version": [
             "5.5.0"
         ],
-        "client.geo.city_name": [
-            "dynamic"
-        ],
-        "client.geo.continent_name": [
-            "Oceania"
-        ],
-        "client.geo.country_iso_code": [
-            "AU"
-        ],
-        "client.geo.country_name": [
-            "Australia"
-        ],
-        "client.geo.location": [
-            "dynamic"
-        ],
-        "client.geo.region_iso_code": [
-            "dynamic"
-        ],
-        "client.geo.region_name": [
-            "dynamic"
-        ],
         "client.ip": [
             "220.244.41.16"
         ],
@@ -130,27 +109,6 @@
         ],
         "agent.version": [
             "5.5.0"
-        ],
-        "client.geo.city_name": [
-            "dynamic"
-        ],
-        "client.geo.continent_name": [
-            "Oceania"
-        ],
-        "client.geo.country_iso_code": [
-            "AU"
-        ],
-        "client.geo.country_name": [
-            "Australia"
-        ],
-        "client.geo.location": [
-            "dynamic"
-        ],
-        "client.geo.region_iso_code": [
-            "dynamic"
-        ],
-        "client.geo.region_name": [
-            "dynamic"
         ],
         "client.ip": [
             "220.244.41.16"
--- a/systemtest/rum_test.go
+++ b/systemtest/rum_test.go
@@ -89,7 +89,6 @@ func TestRUMGeoIpTags(t *testing.T) {
 
 	// Error if geoIp enrichment failed.
 	for _, hit := range result.Hits.Hits {
-		fmt.Println(hit)
 		if v, ok := hit.Source["tags"]; ok {
 			t.Errorf("unexpected tags field in document: %v", v)
 		}


systemtest/approvals/TestRUMXForwardedFor.approved.json
```diff
@@ -131,27 +110,6 @@
         "agent.version": [
             "5.5.0"
         ],
```
kruskall, line 112: Why was this changed ?

systemtest/rum_test.go
```diff
@@ -36,8 +38,63 @@ import (
 	"github.com/elastic/apm-server/systemtest/estest"
 	"github.com/elastic/apm-tools/pkg/approvaltest"
 	"github.com/elastic/apm-tools/pkg/espoll"
+	"github.com/elastic/go-elasticsearch/v8/esapi"
 )
 
+func TestRUMGeoIpTags(t *testing.T) {
+	systemtest.CleanupElasticsearch(t)
+
+	// Ensure the geo-database does not get downloaded.
```
kruskall, line 47: I might be missing something but I think the issue was about testing the geoip is correctly downloaded and applied. This PR seems to do the opposite :thinking:

systemtest/approvals/TestRUMXForwardedFor.approved.json
simitt: +1, the `client.geo` expansion should not be removed.

systemtest/rum_test.go
```diff
@@ -36,8 +38,63 @@ import (
...
+	b, err := json.Marshal(map[string]interface{}{
+		"persistent": map[string]interface{}{
+			"ingest.geoip.downloader.enabled": false,
+		},
+	})
+	require.NoError(t, err)
+
+	r := esapi.ClusterPutSettingsRequest{
+		Body: bytes.NewReader(b),
+	}
```
simitt, line 57: Why would this be set to `false`? The goal is to trigger the geoip downloader and have the relevant fields in the ingested events geoip enriched.
The `tags` field would only be set if an error happened during this download or enrichment, so we want to check that `tags` is empty.

rubvs:
@simitt, there is something flaky here. 

The bind volume (`ingest-geoip`) in the `docker-compose` has been bugging me, since this means our tests are never downloading the geo db.

I've also noticed, the db does net get installed properly the first time I run the systems tests. Even with the cluster settings enabled for downloading. The second time I run `make system-test`, things work as expected.

After looking into this a bit, I see the original [Issue](https://github.com/elastic/apm-server/issues/7260), solved in [PR](https://github.com/elastic/apm-server/pull/8065), introduced the docker volume for exactly this flaky reason.

I'll have a deeper look at this tomorrow.

simitt:
>I've also noticed, the db does net get installed properly the first time I run the systems tests. Even with the cluster settings enabled for downloading. The second time I run make system-test, things work as expected.

Please dig a bit more into why this fails on the first try and succeeds on second. Is it a timing issue?

rubvs:
It doesn't seem like a timing issue from my debugging yesterday. There was a previous PR that would loop twice for this same reason: https://github.com/elastic/apm-server/commit/b5d52d7856c941db6dfa3043cc0a14390e541934. I'll look at this today.

rubvs:
Quick update:

### Yesterday

I looked into why downloading the GeoIp database is so inconsistent. I added the following download funtions:

```go
func waitGeoIPDatabase(ctx context.Context) error {
	b, err := json.Marshal(map[string]any{
		"persistent": map[string]any{
			"ingest.geoip.downloader.enabled":        true,
			"ingest.geoip.downloader.eager.download": true,  // <- Crucial
		},
	})
	r := esapi.ClusterPutSettingsRequest{
		Body: bytes.NewReader(b),
	}
	res, err := r.Do(ctx, Elasticsearch)
	if err != nil {
		return err
	}
	defer res.Body.Close()
	if res.IsError() {
		return fmt.Errorf("response err: %d", res.StatusCode)
	}

	var geoIpResponse struct {
		Databases []struct {
			Id       string `json:"id"`
			Database struct {
				Name string `json:"name"`
			} `json:"database"`
		} `json:"databases"`
	}

	first := true
	for {
		gr, err := Elasticsearch.Ingest.GetGeoipDatabase()
		if err != nil {
			return err
		}
		if err := json.NewDecoder(gr.Body).Decode(&geoIpResponse); err != nil {
			return err
		}
		if len(geoIpResponse.Databases) > 0 {
			return nil
		}
		if first {
			log.Printf("Waiting for geoIP database to be downloaded")
			first = false
		}
		time.Sleep(waitHealthyInterval)
	}
}
```

The critical part, is the addition if `"ingest.geoip.downloader.eager.download": true`. Basically, the cluster settings as defined in the code above states:

`ingest.geoip.downloader.enabled` translates to "it is permissible to download the files when you feel like it", and `ingest.geoip.downloader.eager.download` to "and that time is now". 

With this setup, and a wait loop, we are able to consistently download the db. However, on my local machine this takes some time - around `20s`.

### Today

With the downloading taking some time, it's safe to assume this will lead to some networking issues in CI, hence the solution outlined in [PR](https://github.com/elastic/apm-server/pull/8065). 

I have been looking into programmatically manipulating the volume to toggle the presence of the database. Updates will follow.

axw:
> I have been looking into programmatically manipulating the volume to toggle the presence of the database. Updates will follow.

@rubvs another option to consider: maybe we can intentionally break the ingest processor configuration by updating the `database_file` configuration to something that doesn't exist. It's not exactly the same root cause, but similar enough IMO.

rubvs:
Created a PR with (un)mount the local database file instead of downloading: https://github.com/elastic/apm-server/pull/18277

@axw, do you mean renaming the mounted database file?

axw:
> @axw, do you mean renaming the mounted database file?

No, I mean modify the ingest pipeline. Your approach sounds fine.

rubvs:
@axw oh I see the `database_file` defined [Here](https://github.com/elastic/elasticsearch/blob/4bcc35309c2b6d7577f42202899a69b2fd11280a/x-pack/plugin/apm-data/src/main/resources/ingest-pipelines/apm%40pipeline.yaml#L14C5-L14C18). Okay, let me think about this tomorrow.

rubvs:
Superseded by https://github.com/elastic/apm-server/pull/18325

PR #18277, opened 2025-08-22
update system test for geoip database

rubvs:
Closes https://github.com/elastic/apm-server/issues/17603

Update system tests to ensure the [GeoIp Pipeline](https://github.com/elastic/elasticsearch/blob/4bcc35309c2b6d7577f42202899a69b2fd11280a/x-pack/plugin/apm-data/src/main/resources/ingest-pipelines/apm%40pipeline.yaml#L13) defined in `apm-data` plugin returns the proper result if the GeoIp database is found (not found).

GeoIp is mostly used with RUM, therefore the tests are added in `systemtest/rum_test.go`. We have a few different options to manipulate the database availability:

## 1. Download and remove GeoIp database

Initial look in https://github.com/elastic/apm-server/pull/18217. There are serval issues with this:

- On my local, this takes about `20` seconds to download, which will be a problem in CI.
- Since the test order is non-deterministic, we will have to download/remove it multiple times - depending on the test.
- The original reason for introducing the mount volume was to get rid of networking issues in CI, see [PR](https://github.com/elastic/apm-server/pull/8065).

## 2. Programmatically change the mounted GeoIp database file name:

- Due to this being a `bind` mount, changing the db filename within the container will also mutate the host filename in the `testdata` dir.
- Changing the file name dynamically on the host won't work either, since docker does not have hot-reloading.

## 3. Solution: Programmatically toggle the mount volume

Therefore, the most robust solution would be to toggle the mount volume by dynamically stopping and starting the container with the required mounts.

- Restarting a container is much faster than redownloading the db.
- The system tests becomes more expressive.

### DIFF
--- a/systemtest/containers.go
+++ b/systemtest/containers.go
@@ -87,6 +87,7 @@ func StartStackContainers() error {
 }
 
 func ToggleGeoIpMount(ctx context.Context, enable bool) error {
+
 	src := filepath.Join(
 		systemtestDir,
 		"..",
@@ -102,6 +103,7 @@ func ToggleGeoIpMount(ctx context.Context, enable bool) error {
 		return err
 	}
 	defer cli.Close()
+	cli.NegotiateAPIVersion(ctx)
 
 	c, err := stackContainerInfo(ctx, cli, "elasticsearch")
 	if err != nil {
--- a/systemtest/containers.go
+++ b/systemtest/containers.go
@@ -104,7 +104,6 @@ func ToggleGeoIpMount(ctx context.Context, enable bool) error {
 	defer cli.Close()
 
 	c, err := stackContainerInfo(ctx, cli, "elasticsearch")
-	fmt.Println("Container ID:", c.ID)
 	if err != nil {
 		return err
 	}
@@ -192,7 +191,6 @@ func ToggleGeoIpMount(ctx context.Context, enable bool) error {
 	}
 
 	kb, err := stackContainerInfo(ctx, cli, "kibana")
-	fmt.Println("Kibana Container ID:", kb.ID)
 	if err != nil {
 		return err
 	}
--- a/systemtest/containers.go
+++ b/systemtest/containers.go
@@ -37,6 +37,8 @@ import (
 	"github.com/docker/docker/api/types"
 	"github.com/docker/docker/api/types/container"
 	"github.com/docker/docker/api/types/filters"
+	"github.com/docker/docker/api/types/mount"
+	"github.com/docker/docker/api/types/network"
 	"github.com/docker/docker/client"
 	"github.com/docker/docker/pkg/stdcopy"
 	"github.com/docker/go-connections/nat"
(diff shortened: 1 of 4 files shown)


systemtest/containers.go
```diff
@@ -84,6 +86,124 @@ func StartStackContainers() error {
 	return waitContainerHealthy(ctx, "kibana")
 }
 
+func ToggleGeoIpMount(ctx context.Context, enable bool) error {
```
axw, line 89: Should this accept a `testing.TB` parameter, and not return an error? Then you can simplify the error checking - just use assertions everywhere. Also, that way you could use a cleanup function to revert the change after the test completes.

systemtest/containers.go
```diff
@@ -84,6 +86,124 @@ func StartStackContainers() error {
...
+		return err
+	}
+
+	if err := cli.ContainerRestart(ctx, kb.ID, container.StopOptions{Timeout: nil}); err != nil {
+		return fmt.Errorf("starting kibana container: %w", err)
+	}
+
+	fmt.Printf("Container recreated (Id: %s) with GeoIp mount enabled. Kibana restarted: %v\n", createResp.ID[:5], enable)
+
+	return waitContainerHealthy(ctx, "kibana")
```
axw, line 206: Should these be elasticsearch rather than kibana?

systemtest/containers.go
```diff
@@ -84,6 +86,124 @@ func StartStackContainers() error {
...
+	kb, err := stackContainerInfo(ctx, cli, "kibana")
+	if err != nil {
+		return err
+	}
+
+	if err := cli.ContainerRestart(ctx, kb.ID, container.StopOptions{Timeout: nil}); err != nil {
+		return fmt.Errorf("starting kibana container: %w", err)
+	}
+
+	fmt.Printf("Container recreated (Id: %s) with GeoIp mount enabled. Kibana restarted: %v\n", createResp.ID[:5], enable)
```
axw, line 204: If you pass in a testing.TB, you can use Logf. We shouldn't be printing directly to stdout in test helpers.

rubvs:
Tomorrow: Will have a look why the system tests are failing in CI only.

rubvs:
When I test for `TestRUMXForwardedFor` only, it seems to be failing on `main` with a clean pull and `make system-test`. Something is weird here. Why is the CI passing on `main`.

Screenshot 2025-08-22 at 7 18 09 PM

systemtest/containers.go
rubvs: It's meant to be `kibana`, I need to restart the Kibana container after restarting ES, otherwise I get the following error:

```
Saved object index alias .kibana_9.2.0 not found: index_not_found_exception
```

systemtest/containers.go
rubvs: Good point, I wanted to keep things consistent with the rest of the file.

rubvs:
Looking into why running `TestRUMXForwardedFor` alone causes db enrichment issues:

This happen when any of the following tests inside `systemtest/otlp_test.go` if *not* executed:
* `TestOTLPGRPCTraces`
* `TestOTLPGRPCTraceSpanLinks`
* `TestOTLPClientIP`
* `TestOTLPHTTP`

Indicates there is probably improper shutdown/cleanup happening. ES datastreams seems to be cleanup properly. My next step will be to isolate the RUM tests using a `fixture` or `suite` pattern, instead of having the entire state initialize with `TestMain`.

systemtest/containers.go
axw: Some of these functions are exported and don't take testing.T because they are used by "runapm" and "buildapm". I don't know if they're still used - but they don't use this function, so no need to follow suit.

systemtest/containers.go
axw: OK, makes sense - a comment explaining that would be good

rubvs:
Superseded by https://github.com/elastic/apm-server/pull/18325

Issue #18313, opened 2025-08-26
failing geoip system test when run in isolation


rubvs:
Running `TestRUMXForwardedFor` standalone is failing

```shell
> cd systemtest
> go test -v -race -timeout=20m ./... -run TestRUMXForwardedFor)
```

<img width="1205" height="1483" alt="Screenshot 2025-08-22 at 7 18 09 PM" src="https://github.com/user-attachments/assets/915fe724-f51b-4c94-acc6-944f6c881890" />

Looking into why running `TestRUMXForwardedFor` alone causes db enrichment issues:

This happen when any of the following tests inside `systemtest/otlp_test.go` if *not* executed:
- `TestOTLPGRPCTraces`
- `TestOTLPGRPCTraceSpanLinks`
- `TestOTLPClientIP`
- `TestOTLPHTTP`

isaacaflores2:
I ran into a similar issue in MIS. My solution was to add a [setup step](https://github.com/elastic/apm-managed-service/pull/1882/files#diff-581b976b7e20c79553b3b5586815f6072e74558c2603787538d82e6cf4555dafR132-R169) before running the assertions/test cases and run the setup until the `tags` field is [no longer present](https://github.com/elastic/apm-managed-service/pull/1882/files#diff-8fa992bda07692e66154e14f859237673a030e9fb9034ab3fe79d1c029bb1c94R137-R165). 

Without looking into the system test in detail my guess would be that one of the `TestOTLP` is being run before `TestRUMXForwardedFor`. So one of the `TestOTLP`  test is triggering the geoip db download. This would explain why running `TestRUMXForwardedFor` by standalone is failing

PR #18325, opened 2025-08-26
geoip system tests

rubvs:
Closes https://github.com/elastic/apm-server/issues/17603 & fixes https://github.com/elastic/apm-server/issues/18313

This PR contains some confusion. Mostly, originating from how ES handles GeoIp pipelines. The GeoIP processor operates in two different phases:

1. When the pipelines are created (think compile time)
2. When data are ingest into the pipelines (think runtime)

If the GeoIp database is unavailable on creation, all following ingested data will be tagged, see [Code](https://github.com/elastic/elasticsearch/blob/0814e2f6f6de6c0c5713f60f8843faf7322962f2/modules/ingest-geoip/src/main/java/org/elasticsearch/ingest/geoip/GeoIpProcessor.java#L347)

If the database was available on creation - but is currently unavailable on data ingestion (runtime) - the incoming data will **only** be tagged if `ignore_missing = false`, see [Code](https://github.com/elastic/elasticsearch/blob/8f424781dfe8dc42a6aa9c536677a7745e736ee7/modules/ingest-geoip/src/main/java/org/elasticsearch/ingest/geoip/GeoIpProcessor.java#L114). We have `ignore_missing: true` in our `geoip` pipeline for APM Server, see [Code](https://github.com/elastic/elasticsearch/blob/8f424781dfe8dc42a6aa9c536677a7745e736ee7/x-pack/plugin/apm-data/src/main/resources/ingest-pipelines/apm%40pipeline.yaml#L18).

Therefore, the system tests are testing for 3 things:

1. **Compile Time**: The database is unavailable on start (compile time). We have to restart the container without downloading - in which case, the `tags` field will be present the response.
2. **Runtime**: Enrichment succeeded due to the GeoIp database being available (downloaded) when the container starts.
3. **Runtime**: Enrichment fails due to missing ip address *or* the database was removed after initial creation (runtime). In this case, there should be **no** `tags`, but the `client.geo` fields in the response should be missing.

### DIFF
--- a/systemtest/main_test.go
+++ b/systemtest/main_test.go
@@ -48,14 +48,6 @@ func TestMain(m *testing.M) {
 	os.Exit(m.Run())
 }
 
-var stats struct {
-	GeoIPStats `json:"stats"`
-}
-
-type GeoIPStats struct {
-	DatabasesCount int `json:"databases_count"`
-}
-
 const requestBody = `{"metadata": { "service": {"name": "alice", "agent": {"version": "3.14.0", "name": "elastic-node"}}}}
 {"metricset": {"samples":{"a":{"value":3.2}}, "timestamp": 1496170422281000}}`
 
@@ -113,10 +105,15 @@ func waitGeoIPDownload() error {
 			if err != nil {
 				return err
 			}
+			var stats = struct {
+				Stats struct {
+					DatabasesCount int `json:"databases_count"`
+				} `json:"stats"`
+			}{}
 			if err := json.Unmarshal(body, &stats); err != nil {
 				return err
 			}
-			if stats.DatabasesCount == 3 {
+			if stats.Stats.DatabasesCount == 3 {
 				log.Println("GeoIP database downloaded")
 				return nil
 			}


docker-compose.yml
```diff
@@ -33,11 +33,12 @@ services:
...
       - "logger.org.elasticsearch=${ES_LOG_LEVEL:-error}"
       - "action.destructive_requires_name=false"
+      - "ingest.geoip.downloader.enabled=true"
+      - "ingest.geoip.downloader.eager.download=true"
     volumes:
       - "./testing/docker/elasticsearch/roles.yml:/usr/share/elasticsearch/config/roles.yml"
       - "./testing/docker/elasticsearch/users:/usr/share/elasticsearch/config/users"
       - "./testing/docker/elasticsearch/users_roles:/usr/share/elasticsearch/config/users_roles"
-      - "./testing/docker/elasticsearch/ingest-geoip:/usr/share/elasticsearch/config/ingest-geoip"
     logging: *default-logging
```
rubvs, line 40: The [GeoLite2-City.mmdb](https://github.com/elastic/apm-server/blob/main/testing/docker/elasticsearch/ingest-geoip/GeoLite2-City.mmdb) file is either corrupted or outdated, and is the root cause of https://github.com/elastic/apm-server/issues/18313.

With the `eager.download` option enabled in the container `env` vars, the db will download on pipeline creation (immediately).

systemtest/containers.go
```diff
@@ -435,18 +435,6 @@ func (c *ElasticAgentContainer) Exec(ctx context.Context, cmd ...string) (stdout
...
-	var status struct {
-		Name    string `json:"name"`
-		Version string `json:"version"`
-		Status  string `json:"status"`
-	}
-	if err := json.NewDecoder(r).Decode(&status); err != nil {
-		return false
-	}
-	return status.Status == "HEALTHY"
-}
```
rubvs, line 448: unused code. cleanup.

docker-compose.yml
```diff
@@ -33,11 +33,12 @@ services:
       - "xpack.security.authc.api_key.enabled=true"
       - "logger.org.elasticsearch=${ES_LOG_LEVEL:-error}"
       - "action.destructive_requires_name=false"
+      - "ingest.geoip.downloader.enabled=true"
+      - "ingest.geoip.downloader.eager.download=true"
     volumes:
       - "./testing/docker/elasticsearch/roles.yml:/usr/share/elasticsearch/config/roles.yml"
       - "./testing/docker/elasticsearch/users:/usr/share/elasticsearch/config/users"
       - "./testing/docker/elasticsearch/users_roles:/usr/share/elasticsearch/config/users_roles"
-      - "./testing/docker/elasticsearch/ingest-geoip:/usr/share/elasticsearch/config/ingest-geoip"
```
kruskall, line 40: **issue**: the `ingest-geoip` folder is unused now that this has been removed, can we remove it ?

systemtest/rum_test.go
```diff
@@ -38,6 +38,187 @@ import (
...
+		},
+	)
+
+	for _, hit := range result.Hits.Hits {
+		_, hasTags := hit.Source["tags"]
+		require.NotEmpty(t, hasTags)
+	}
+}
+
+func TestGeoIp_EnrichmentSuccess(t *testing.T) {
```
kruskall: **issue**: we already have system tests validating geoip enrichment, why are we duplicating the same "coverage" here ?

systemtest/rum_test.go
```diff
@@ -38,6 +38,187 @@ import (
 	"github.com/elastic/apm-tools/pkg/espoll"
 )
 
+const requestBody = `{"metadata":{"service":{"name":"rum-js-test","agent":{"name":"rum-js","version":"5.5.0"}}}}
+{"transaction":{"trace_id":"611f4fa950f04631aaaaaaaaaaaaaaaa","id":"611f4fa950f04631","type":"page-load","duration":643,"span_count":{"started":0}}}
+{"metricset":{"samples":{"transaction.breakdown.count":{"value":12},"transaction.duration.sum.us":{"value":12},"transaction.duration.count":{"value":2},"transaction.self_time.sum.us":{"value":10},"transaction.self_time.count":{"value":2},"span.self_time.count":{"value":1},"span.self_time.sum.us":{"value":633.288}},"transaction":{"type":"request","name":"GET /"},"span":{"type":"external","subtype":"http"},"timestamp": 1496170422281000}}
+`
+
+func TestGeoIp_DatabaseUnavailable(t *testing.T) {
```
kruskall: **issue**: this shouldn't be something we test in apm-server IMO, it seems we're crossing into "testing what ES is doing" and that's an implementation detail that we shouldn't be concerned about

systemtest/rum_test.go
```diff
@@ -38,6 +38,187 @@ import (
...
+
+		// The "tags" field should only be populated if database
+		// was unavailable when pipelines was created.
+		if _, hasTags := hit.Source["tags"]; hasTags {
+			t.Errorf("did not expect tags field to be present")
+		}
+	}
+}
+
+func TestGeoIp_EnrichmentFail(t *testing.T) {
```
kruskall: same as above, we shouldn't test this IMO

docker-compose.yml
rubvs: Yes thanks, forgot about that.

systemtest/rum_test.go
rubvs: We can remove `TestRUMXForwardedFor`. I'll move the approvals to this test also. The sucess test here is mostly a renaming for consistency with the failure of enrichment

simitt:
Here's how I understand this to work: The [`apm-data` plugin](https://github.com/elastic/elasticsearch/blob/0814e2f6f6de6c0c5713f60f8843faf7322962f2/x-pack/plugin/apm-data/src/main/resources/ingest-pipelines/apm%40pipeline.yaml#L13-L21) defines `download_database_on_pipeline_creation: false`. This means that the geoip database is not downloaded when the apm assets are installed, but only on first actual need, so when the first document with the `client.ip` field set, arrives.
The question is why the behaviour is flaky when the database is not pre-downloaded. In previous conversations @rubvs  indicated that this is not a timing issue, but I am not sure there is yet an alternative explanation for it.

masseyke:
Is the problem just that it takes a little bit of time for elasticsearch to download and install the database? It does not block requests while this is happening, so some will be tagged as if no geoip database exists.

rubvs:
Thanks @masseyke, the async parts cleared up some confusion. I have updated the system tests accordingly. @simitt the updated test will trigger a lazy download on the first request, which will then poll for the absence of `tags` for up to 3 minutes, while the database is being downloaded. On success, the resulting hits will be asserted with the approvals.

rubvs:
Mmh, don't know why the fips unit tests are failing though.

docker-compose.yml
```diff
@@ -33,11 +33,12 @@ services:
       - "xpack.security.authc.api_key.enabled=true"
       - "logger.org.elasticsearch=${ES_LOG_LEVEL:-error}"
       - "action.destructive_requires_name=false"
+      - "ingest.geoip.downloader.enabled=true"
+      - "ingest.geoip.downloader.eager.download=true"
```
simitt: Why is this enabled? Wasn't the conclusion that the GeoIP database would get lazy downloaded on demand?

systemtest/main_test.go
```diff
@@ -18,21 +18,77 @@
...
+}
+
 func TestMain(m *testing.M) {
 	log.Println("INFO: starting stack containers...")
 	initContainers()
 	if err := StartStackContainers(); err != nil {
 		log.Fatalf("failed to start stack containers: %v", err)
 	}
 	initElasticSearch()
+	if err := waitGeoIPDownload(); err != nil {
```
simitt: What would trigger a download here?

systemtest/rum_test.go
```diff
@@ -38,8 +39,59 @@ import (
 	"github.com/elastic/apm-tools/pkg/espoll"
 )
 
+const requestBody = `{"metadata":{"service":{"name":"rum-js-test","agent":{"name":"rum-js","version":"5.5.0"}}}}
+{"transaction":{"trace_id":"611f4fa950f04631aaaaaaaaaaaaaaaa","id":"611f4fa950f04631","type":"page-load","duration":643,"span_count":{"started":0}}}
+{"metricset":{"samples":{"transaction.breakdown.count":{"value":12},"transaction.duration.sum.us":{"value":12},"transaction.duration.count":{"value":2},"transaction.self_time.sum.us":{"value":10},"transaction.self_time.count":{"value":2},"span.self_time.count":{"value":1},"span.self_time.sum.us":{"value":633.288}},"transaction":{"type":"request","name":"GET /"},"span":{"type":"external","subtype":"http"},"timestamp": 1496170422281000}}
+`
```
simitt: I don't see how this request body should be triggering the GeoIP download. As mentioned in https://github.com/elastic/apm-server/pull/18325#issuecomment-3232166887, the `client.ip` is specified as `field`, so I'd expect this field needs to be present in the request.

systemtest/main_test.go
```diff
@@ -18,21 +18,77 @@
...
+		log.Fatalf("failed to download GeoIp database: %v", err)
+	}
 	initKibana()
 	initSettings()
 	initOTEL()
 	log.Println("INFO: running system tests...")
 	os.Exit(m.Run())
 }
+
+func waitGeoIPDownload() error {
```
simitt: This should be checked as part of the lazy download.

systemtest/main_test.go
rubvs: The `eager` enabled the download starts creation.

systemtest/rum_test.go
rubvs: The address is in the the request header

rubvs:
- Removed downloading the db on startup.
- When required, download the geoip db lazily.
- The approvals already check for `tags`. I tested this manually to make sure. 

Moreover, from an abstraction level this PR reintroduces locality: The test that requires the geoip db, is responsible for downloading the db.

The `GeoIpLazyDownload` sends an event containing `client.ip` to trigger the download, then waits until finished. The datastreams have to cleared, since they will contain noisy tagged items.

docker-compose.yml
isaacaflores2: In case anyone else reviews this later, we discovered that the file was not corrupt right? Even though the geoip db files were mounted, they are not available immediately. So we were just running into the race condition where the geoip database was not completely downloaded before the test started.

isaacaflores2 (dismissed):
Thanks for sticking with this issue and sharing the geoIP database details you learned about ElasticSearch. 

It looks like test requested in https://github.com/elastic/apm-server/issues/17603 were already present. Like you mentioned the approvals.json files do not contain a `tags` field  which will assert the absence of `tags` given that ElasticSearch has been properly setup with the geoip databases.

systemtest/rum_test.go
```diff
@@ -49,27 +51,43 @@ func TestRUMXForwardedFor(t *testing.T) {
...
+
 	resp, err := http.DefaultClient.Do(req)
 	require.NoError(t, err)
+
+	_, err = io.ReadAll(resp.Body)
+	require.NoError(t, err)
 	require.Equal(t, http.StatusAccepted, resp.StatusCode)
+
 	io.Copy(io.Discard, resp.Body)
 	resp.Body.Close()
```
kruskall: **question**: why was this changed ? The new version doesn't seem to be correct (readall + io.copy?)

systemtest/rum_test.go
```diff
@@ -49,27 +51,43 @@ func TestRUMXForwardedFor(t *testing.T) {
 	require.NoError(t, err)
 	serverURL.Path = "/intake/v2/rum/events"
 
-	// Send one transaction and one set of breakdown metrics.
-	// They should both have geoIP enrichment applied.
```
kruskall: **question**: why was this removed ?

systemtest/rum_test.go
```diff
@@ -49,27 +51,43 @@ func TestRUMXForwardedFor(t *testing.T) {
...
 	require.NoError(t, err)
+
+	_, err = io.ReadAll(resp.Body)
+	require.NoError(t, err)
 	require.Equal(t, http.StatusAccepted, resp.StatusCode)
+
 	io.Copy(io.Discard, resp.Body)
 	resp.Body.Close()
 
-	result := estest.ExpectMinDocs(t, systemtest.Elasticsearch, 2, "traces-apm*,metrics-apm*", espoll.TermsQuery{
```
kruskall: **question**: same as above, this looks identical

systemtest/agentconfig_test.go
```diff
@@ -39,6 +39,7 @@ import (
 
 func TestAgentConfig(t *testing.T) {
 	systemtest.CleanupElasticsearch(t)
+	systemtest.GeoIpLazyDownload(t)
```
simitt: Why is this needed here? Afaics this test has nothing to do with GeoIP (similar note for the other tests below, except for the RUM one).

rubvs: These tests would fail if the database is not installed and they are ran in isolation. After another look, it seems like the test failing is due to them pulling "almost" the full doc via `estest.ExpectDocs`, while other tests would filter specific field. Essentially, filtering out the `tags` in the response. So some tests approvals will pass, while others not.

My conclusion, is that due to the db not being available on pipeline creation, almost all tests will end up with  a `tags` field, since the db will tag them all, see [Code](https://github.com/elastic/elasticsearch/blob/0814e2f6f6de6c0c5713f60f8843faf7322962f2/modules/ingest-geoip/src/main/java/org/elasticsearch/ingest/geoip/GeoIpProcessor.java#L347). I outlined this in the PR desc.

That brings us back to one of my previous PR implementation, see [Diff](https://github.com/elastic/apm-server/pull/18277/files#diff-c390be9f1fecd83135b63177c345d13b20d15a91d433525763a655f4b6bad655R89), 

- where we have to download the db when the ES container spins up,
- then remove the db, and restart the container for RUM tests where we want to test for lazy downloads only.

I am hesitant to do this container manipulation, which is why this PR initially included the download relating your https://github.com/elastic/apm-server/pull/18325#discussion_r2309183424. 

In my opinion, the core issue is the `geoip` pipeline being applied to all `apm` datastreams.

- Which leads to the dichotomy of having to account for 2 different types of downloads in out system test. Essentially encoding the two-phase process ES using that I outlined in the PR desc.
- To my knowledge, this also caused some confusion with the data-accuracy tests in MIS, see [Code](https://github.com/elastic/apm-managed-service/blob/5707b0c7e1929ae2789f3149a0596ec2af5d85ec/dataaccuracytest/main_test.go#L132), where we have to account for the geoip pipeline even though we do not use RUM in MIS.

Maybe we can move the `geoip` pipeline to [traces-apm.rum](https://github.com/elastic/elasticsearch/blob/main/x-pack/plugin/apm-data/src/main/resources/ingest-pipelines/traces-apm.rum%40default-pipeline.yaml). But I'm not sure how to handle metrics.

simitt: It's surprising to me that the geoip database gets downloaded lazily whenever the pipeline is called, rather than only when the pipeline condition evaluates to true (in this case when the `client.ip` field exists). But at least it explains why seamingly unrelated tests are also dependent on the geoIP database. 

In this case, what do you think about moving the `systemtest.GeoIpLazyDownload` o the Testsuite setup step in `TestMain` so it is not a constant source of pain if new tests are added? 
If there's a reasonable argument against this, I am also fine with the current state of this PR.

masseyke: > It's surprising to me that the geoip database gets downloaded lazily whenever the pipeline is called, rather than only when the pipeline condition evaluates to true

Our goal was to avoid having customers unnecessarily download the geoip databases during their own CI processes that don't actually have any geoip pipelines, since that can get expensive. Our thinking was that if they _did_ have a geoip processor in some pipeline, they'd probably want it available. We didn't get into checking conditions.

simitt: thanks for confirming, I think this was the last missing piece of the puzzle.

simitt (approved):
🎉

simitt:
@kruskall please re-review as you had requested changes, so this can be brought over the finish line

systemtest/main_test.go
```diff
@@ -30,9 +38,90 @@ func TestMain(m *testing.M) {
...
 	initOTEL()
 	log.Println("INFO: running system tests...")
 	os.Exit(m.Run())
 }
+
+var stats struct {
+	GeoIPStats `json:"stats"`
+}
+
+type GeoIPStats struct {
```
1pkg: nit: any point why not move this support struct definitions inside `waitGeoIPDownload ` since they are only used by it there is no need to expose them inside the `sysemtest` package

1pkg (approved):

PR #18367, opened 2025-08-27
feat: integrate STIG check in buildkite pipeline

kruskall:
## Motivation/summary

add STIG check pipeline

depend on DRA to ensure docker images are pushed before running the check

## Checklist

## Related issues

Related to https://github.com/elastic/apm-server/issues/18155

### DIFF
--- a/.buildkite/package.yml
+++ b/.buildkite/package.yml
@@ -61,15 +61,6 @@ steps:
       - step: "package"
         allow_failure: false
 
-  - label: "Trigger stig-status pipeline docker.elastic.co/apm/apm-server-fips:SNAPSHOT"
-    trigger: stig-status
-    if: build.branch == "main"
-    build:
-      env:
-        CONTAINER_IMAGE: "docker.elastic.co/apm/apm-server-fips:SNAPSHOT"
-    depends_on:
-      - step: "dra"
-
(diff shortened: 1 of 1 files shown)


.buildkite/package.yml
```diff
@@ -60,3 +60,12 @@ steps:
     depends_on:
       - step: "package"
         allow_failure: false
+
+  - label: "Trigger stig-status pipeline docker.elastic.co/apm/apm-server-fips:${BUILDKITE_BRANCH}-SNAPSHOT"
+    trigger: stig-status
+    if: build.branch != "main"
```
carsonip, line 66: why not main? What about DRA master snapshot?

kruskall: given the apm-server (fips) image is basically empty (static binary + cert) the reasoning was to run this on stack branches instead of main to reduce noice and resource usage

simitt: If there is a problem in `main` and we don't discover it through this pipeline, it would create problems for the current `SNAPSHOT` version.

kruskall: Looking a bit more into this I'm afraid there's no easy way to pull the latest snapshot image and dynamically set an env variable

I'll raise this with the relevant team but I don't think we're losing much coverage if we only run this on stack branches

simitt: ok, let's move foward as is then for now.

PR #18379, opened 2025-08-28
feat: return an error if tbs config fails

kruskall:
## Motivation/summary

failure to process the tbs config will log a message and silently fallback to the default config with tbs disabled.
This is not ideal and doesn't match the behaviour of other configs which return an error to the caller.

Update the tbs config logic to return an error if processing fails

Update test to assert error on malformed configs

We might need to be careful with backporting (maybe only 9.x ?)

## Checklist

## How to test these changes

- start apm-server with malformed tbs config
- observe a failure

### DIFF
--- a/internal/beater/config/sampling_test.go
+++ b/internal/beater/config/sampling_test.go
@@ -39,7 +39,7 @@ func TestSamplingPoliciesValidation(t *testing.T) {
 		c, err := NewConfig(config.MustNewConfigFrom(map[string]interface{}{
 			"sampling.tail.enabled": true,
 		}), nil, logptest.NewTestingLogger(t, ""))
-		assert.Error(t, err)
+		assert.EqualError(t, err, "error processing configuration: invalid sampling.tail config: no policies specified accessing 'sampling.tail'")
 		assert.Nil(t, c)
 	})
 	t.Run("NoDefaultPolicies", func(t *testing.T) {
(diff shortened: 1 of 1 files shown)


internal/beater/config/sampling_test.go
```diff
@@ -39,8 +39,8 @@ func TestSamplingPoliciesValidation(t *testing.T) {
 		c, err := NewConfig(config.MustNewConfigFrom(map[string]interface{}{
 			"sampling.tail.enabled": true,
 		}), nil, logptest.NewTestingLogger(t, ""))
-		assert.NoError(t, err)
-		assert.False(t, c.Sampling.Tail.Enabled)
+		assert.Error(t, err)
```
carsonip: assert the error message to ensure it is returning the right error

internal/beater/config/sampling_test.go
```diff
@@ -49,7 +49,7 @@ func TestSamplingPoliciesValidation(t *testing.T) {
 				"sample_rate":  0.5,
 			}},
 		}), nil, logptest.NewTestingLogger(t, ""))
-		assert.NoError(t, err)
-		assert.False(t, c.Sampling.Tail.Enabled)
+		assert.Error(t, err)
```
carsonip: ditto

internal/beater/config/sampling_test.go
kruskall: Good catch! Updated!

internal/beater/config/sampling_test.go
kruskall: Updated!

PR #18550, opened 2025-09-11
smoke: Fix deprecated aws_region name

ericywl:
## Motivation/summary

`aws_region.name` is deprecated in the latest version, use `aws_region.region` instead.

## How to test these changes

Run `smoke-test-os`: https://github.com/elastic/apm-server/actions/runs/17650010191.


testing/infra/terraform/modules/standalone_apm_server/main.tf
```diff
@@ -136,7 +136,7 @@ data "aws_subnets" "public_subnets" {
   }
   filter {
     name   = "availability-zone"
-    values = ["${data.aws_region.current.name}a"]
```
ericywl, line 139: @1pkg I saw that you were the one that committed this,  but is the `a` at the back intentional?

carsonip: i have to assume it is intentional to specify `a` in region

1pkg: I do not recall the context here, but looks like the single 'a' zone is picked on purpose. Since `aws_instance` requires single `subnet_id` input.

marclop (approved):

ericywl:
Unrelated system test seems to be failing 😭

PR #18581, opened 2025-09-15
bk: support packaging for PRs too

v1v:
## Motivation/summary

Use the `BK PR Bot` to run the `apm-server-package` on PRs too.

This should help with detecting breaking changes in the Unified Release before merging when using forked PRs. The only way to test that currently is by using a feature branch in the upstream.

## Checklist


catalog-info.yaml
```diff
@@ -34,21 +34,27 @@ spec:
...
         filter_enabled: true
-        filter_condition: build.branch == "main" || build.branch == "8.x" || build.branch =~ /^[0-9]+\.[0-9]+$$/ || build.branch =~ /^feature\//
-      cancel_intermediate_builds: false
-      skip_intermediate_builds: false
+        filter_condition: >-
+          build.pull_request.id == null || (build.creator.name == 'elasticmachine' && build.pull_request.id != null)
+      cancel_intermediate_builds: true
+      cancel_intermediate_builds_branch_filter: "!main !7.* !8.* !9.*"
+      skip_intermediate_builds: true
+      skip_intermediate_builds_branch_filter: "!main !7.* !8.* !9.*"
```
fr4nc1sc0-r4m0n, line 50: I probably don't have enough context here, so my apologies if I say something wrong in advance. Is this filter with the `skip` and `cancel` flags allowing only executions when a PR is created by the `elasticmachine` service user but not when a release or main branch as it does before?

v1v: I don't follow your question.

If you mean about `filter_condition`, that's the one that was copied from the Elastic Agent, unfortunately when I asked about the specifics I dind't get a clear answer. I think `elasticmachine` was used as part of the BK PR Bot service, hence that was needed in the filter_condition. If you search for similar [configurations](https://github.com/search?q=org%3Aelastic+++build.pull_request.id+%21%3D+null+language%3AYAML+catalog-info.yaml&type=code) you can see we follow the same pattern.

If you ask about `skip_intermediate_builds` or `cancel_intermediate_builds_branch_filter`, they are independent of the `filter`condition.

If you go to this [private link](https://docs.elastic.dev/ci/getting-started-with-buildkite-at-elastic) [only for Elastic employees], you can see what those settings do:

> * filter_condition: - (Optional, Default: "") The condition to evaluate when deciding if a build should run. More details available in the [Buildkite documentation](https://buildkite.com/docs/pipelines/conditionals#conditionals-in-pipelines).
> * [cancel_intermediate_builds](https://buildkite.com/docs/pipelines/skipping#cancel-running-intermediate-builds): (Optional, Default: False) A boolean to enable automatically cancelling any running builds on the same branch when a new build is created.
> * cancel_intermediate_builds_branch_filter: (Optional, Default: "") Limit which branches build cancelling applies to, for example !master will ensure that the master branch won't have it's builds automatically cancelled.
> * [skip_intermediate_builds](https://buildkite.com/docs/pipelines/skipping): (Optional, Default: True) A boolean to enable automatically skipping any unstarted builds on the same branch when a new build is created.
> * skip_intermediate_builds_branch_filter: (Optional, Default: "") Limit which branches build skipping applies to, for example !master will ensure that the master branch won't have it's builds automatically skipped.


Let me know If I clarify your question

fr4nc1sc0-r4m0n: Thank you for the explanation. 
Regarding the `filter_condition` I was trying to understand the filter expression and what has changed. I suppose that before the changes the filter works when the branch is `main` or `8.x `or any `X.X` or any branch starting with `feature`. Now with the change, I'm just wondering if first condition  is when the pull_request.id is `null` means when there's no PR and basically contains all the previous conditions before the change.

Issue #18558, opened 2025-09-11
ci: unit tests are failing (randomly) more often


kruskall:
**APM Server version** (`apm-server version`):

**Description of the problem including expected versus actual behavior**:

CI tests have been failing a lot more often lately. unit test failure seems to be ~38% according to github action perf metrics.

We should look into any flaky test

**Steps to reproduce**:

 1.
 2.
 3.

**Provide logs (if relevant)**:

kruskall:
PRs have been merged and tests should be more stable now, there's one flaky test due to beats managerv2 goroutine leaking but that should be fixed in beats

simitt:
>PRs have been merged and tests should be more stable now, there's one flaky test due to beats managerv2 goroutine leaking but that should be fixed in beats

Could you link to the tracking issue in beats please?

PR #18692, opened 2025-09-22
test(TestMonitoring): pass the correct collect struct to helper method

kruskall:
## Motivation/summary

Update TestMonitoring to pass the collectT struct rather than the primary test T. Since an assert.Eventually call will cause the test to fail when the assertion uses the main T, we now forward the closure argument (collectT) to ensure proper handling.

## Checklist

## How to test these changes

make test

## Related issues

### DIFF
--- a/internal/beater/monitoringtest/opentelemetry.go
+++ b/internal/beater/monitoringtest/opentelemetry.go
@@ -42,7 +42,7 @@ func ExpectContainOtelMetrics(
 	assertOtelMetrics(t, reader, expectedMetrics, false, false)
 }
 
-func ExpectContainOtelMetricsKeys(t *testing.T, reader sdkmetric.Reader, expectedMetricsKeys []string) {
+func ExpectContainOtelMetricsKeys(t assert.TestingT, reader sdkmetric.Reader, expectedMetricsKeys []string) {
 	expectedMetrics := make(map[string]any)
 	for _, metricKey := range expectedMetricsKeys {
 		expectedMetrics[metricKey] = nil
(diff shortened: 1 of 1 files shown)


internal/beater/monitoringtest/opentelemetry.go
```diff
@@ -62,55 +63,72 @@ func assertOtelMetrics(
...
-
-				if v, ok := expectedMetrics[m.Name]; ok {
-					assert.EqualValues(t, v, d.DataPoints[0].Count, m.Name)
-				} else if fullMatch {
-					assert.Fail(t, "unexpected metric", m.Name)
-				}
+		metrics = append(metrics, sm.Metrics...)
+	}
+
+	// multiple metrics might
```
carsonip: incomplete comment?

kruskall: Fixed!

x-pack/apm-server/main_test.go
```diff
@@ -54,12 +54,10 @@ func TestMonitoring(t *testing.T) {
 	runServerError := errors.New("runServer")
 	runServerFunc := func(ctx context.Context, args beater.ServerParams) error {
 		// run server for some time until storage metrics are reported by the storage manager
-		assert.EventuallyWithT(t, func(c *assert.CollectT) {
```
carsonip, line 57: I believe the actual bug was that `c` was unused. Rename it to `t` then `EventuallyWithT` should function as expected.

kruskall: This is true, thanks for spotting this!

kruskall:

PR #18730, opened 2025-09-23
test(TestLibbeatMetrics): wait until batches count is expected value

kruskall:
## Motivation/summary

When there are no active indexers but only a subset of bulk requests (e.g., two) have been counted, the bulk‑request metrics become inconsistent and cause test failures.

This change waits until the number of active indexers drops to 0 and the bulk‑request counter reaches 4, guaranteeing that all batches and their statuses are fully accounted for.

## Checklist

## How to test these changes

- `go test -race -v -count=1000 -failfast -run=TestLibbeatMetrics ./...`

## Related issues

### DIFF
--- a/internal/beatcmd/beat_test.go
+++ b/internal/beatcmd/beat_test.go
@@ -155,10 +155,11 @@ func TestLibbeatMetrics(t *testing.T) {
 		assert.NoError(t, appender.Close(context.Background()))
 	}()
 
-	require.NoError(t, appender.Add(context.Background(), "index", strings.NewReader("{}")))
-	require.NoError(t, appender.Add(context.Background(), "index", strings.NewReader("{}")))
-	require.NoError(t, appender.Add(context.Background(), "index", strings.NewReader("{}")))
-	require.NoError(t, appender.Add(context.Background(), "index", strings.NewReader("{}")))
+	const totalRequests = 4
+
+	for range totalRequests {
+		require.NoError(t, appender.Add(context.Background(), "index", strings.NewReader("{}")))
+	}
 
 	statsRegistry := beat.Monitoring.StatsRegistry()
 	libbeatRegistry := statsRegistry.GetRegistry("libbeat")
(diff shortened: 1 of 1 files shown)


internal/beatcmd/beat_test.go
```diff
@@ -192,7 +192,7 @@ func TestLibbeatMetrics(t *testing.T) {
 		snapshot = monitoring.CollectStructSnapshot(libbeatRegistry, monitoring.Full, false)
 		output := snapshot["output"].(map[string]any)
 		events := output["events"].(map[string]any)
-		return events["active"] == int64(0)
+		return events["active"] == int64(0) && events["batches"] == int64(4)
```
rubvs: nit: Maybe move the `4` into a `const`, since it's used a lot in the test, and it's not obvious why `4` unless I'm reading the entire test.

kruskall: Good idea! Updated!

internal/beatcmd/beat_test.go
```diff
@@ -192,7 +193,7 @@ func TestLibbeatMetrics(t *testing.T) {
 		snapshot = monitoring.CollectStructSnapshot(libbeatRegistry, monitoring.Full, false)
 		output := snapshot["output"].(map[string]any)
 		events := output["events"].(map[string]any)
-		return events["active"] == int64(0)
+		return events["active"] == int64(0) && events["batches"] == int64(totalRequests)
```
carsonip, line 196: nit: moving to [EventuallyWithT](https://github.com/stretchr/testify/blob/v1.11.1/assert/assertions.go#L2083) allows this to be split into 2 assert statements. If EventuallyWithT fails, it'll print more useful info.

carsonip (commented):
since you're already on it, there's a nit

internal/beatcmd/beat_test.go
kruskall: I don't really like `EventuallyWithT`, we already had an issue with it recently :cry: 

It's changing the meaning of assertions and making it more complex to reason about failure imo

carsonip (approved):
lgtm thanks

kruskall:

PR #18926, opened 2025-10-03
ci: build apm-server before running systemtests

### LINKED ISSUE #17487: TestNewServerTB is too slow
systemtest is currently our slowest job in CI

looking at random systemtest run there seems to be some really odd tests:

--- PASS: TestNewServerTB (97.45s)

97s is definitely not ok, we should investigate what's wrong

kruskall:
## Motivation/summary

Go tests build a separate binary for each package. Because we have both systemtest and apmservertest packages the systemtest job creates two binaries which causes the make build task to run twice. Tests in multiple packages run in parallel so the build cache is not shared effectively.

Since the target OS (GOOS) and architecture (GOARCH) stay constant throughout a run, we only need a single binary. Storing cached paths in a map is unnecessary.

Changes:

Update the CI job to compile the apm‑server binary before executing the systemtests, allowing the build cache to be reused across packages.

Run the binary build in parallel with Docker image pulls to shave off additional time.

Replace the apmserverBinary mutex with a sync.Once, reducing contention within the same package.

This also appears to reduce systemtest job time by ~1m.

## Checklist

## How to test these changes

- watch CI systemtest job

## Related issues

Closes https://github.com/elastic/apm-server/issues/17487

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -118,10 +118,6 @@ jobs:
         with:
           go-version-file: systemtest/go.mod
           cache: false
-      - run: |
-            docker compose up -d &
-            make apm-server &
-            wait
       - uses: actions/cache@v4
         with:
           path: |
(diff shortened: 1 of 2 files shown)


axw (approved):

v1v (approved):

carsonip:
@Mergifyio backport 8.19

PR #18936, opened 2025-10-06
iserver-test: Skip 9.3 for now

ericywl:
The 9.3 snapshot is not available yet, skip 9.3 snapshot for now.

### DIFF
--- a/integrationservertest/scripts/genpath/versions_snapshot.go
+++ b/integrationservertest/scripts/genpath/versions_snapshot.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"slices"
 	"strconv"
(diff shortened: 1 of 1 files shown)


ericywl:
> nit: should we use errors.Is and return a static error

Yeah that sounds like a better idea.

integrationservertest/internal/ech/versions_cache.go
```diff
@@ -19,9 +19,14 @@ package ech
 
 import (
 	"context"
+	"errors"
 	"fmt"
 )
 
+var (
+	ErrNotFoundInEC = errors.New("not found in EC")
```
kruskall: suggestion: the error message should be more descriptive, e.g. version not found in EC

PR #18945, opened 2025-10-06
fix: disable tracing for the self‑instrumentation Elasticsearch client and Pub/Sub ES client.

kruskall:
## Motivation/summary

Self‑instrumentation must not emit extra traces that would be returned to the trace server and cause further tracing loops. We also want to exclude the Pub/Sub Elasticsearch client from tracing. Therefore, tracing is disabled for the Elasticsearch client used by the self‑instrumentation processor and for the Pub/Sub client.

## Checklist

## Related issues

### DIFF
--- a/internal/elasticsearch/client.go
+++ b/internal/elasticsearch/client.go
@@ -117,7 +117,7 @@ func NewClientParams(args ClientParams) (*Client, error) {
 	if _, ok := args.TracerProvider.(noop.TracerProvider); !ok {
 		// only enable tracing with apm agent if a non-noop tracerprovider
 		// has been passed.
-		// TODO replace apmelasticsearch with otel
+		// TODO replace apmelasticsearch with otel (https://github.com/elastic/apm-server/issues/18949)
 		transport = apmelasticsearch.WrapRoundTripper(transport)
 	}
 


internal/elasticsearch/client.go
```diff
@@ -109,13 +114,20 @@ func NewClientParams(args ClientParams) (*Client, error) {
 		apikey = base64.StdEncoding.EncodeToString([]byte(args.Config.APIKey))
 	}
 
+	if _, ok := args.TracerProvider.(noop.TracerProvider); !ok {
+		// only enable tracing with apm agent if a non-noop tracerprovider
+		// has been passed.
+		// TODO replace apmelasticsearch with otel
```
lahsivjar: Can we have an assigned TODO or create an issue for this and link it here?

kruskall: Done!

internal/beater/beater.go
```diff
@@ -466,7 +470,7 @@ func (s *Runner) Run(ctx context.Context) error {
 		SourcemapFetcher:       sourcemapFetcher,
 		PublishReady:           publishReady,
 		KibanaClient:           kibanaClient,
-		NewElasticsearchClient: newElasticsearchClient,
+		NewElasticsearchClient: newESClient(tracenoop.NewTracerProvider()),
```
simitt, line 473: Why is tracing also disabled here and not just for the self instrumentation with the `tracerServerListener` below?

PR #19115, opened 2025-10-14
Revert "smoke: Fix deprecated aws_region name"

ericywl:
Reverts elastic/apm-server#18550 since its causing benchmarks to fail.

### DIFF
--- a/testing/benchmark/main.tf
+++ b/testing/benchmark/main.tf
@@ -7,7 +7,7 @@ terraform {
     }
     aws = {
       source  = "hashicorp/aws"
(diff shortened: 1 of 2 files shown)


testing/benchmark/main.tf
```diff
@@ -7,7 +7,7 @@ terraform {
     }
     aws = {
       source  = "hashicorp/aws"
-      version = ">=6.0"
+      version = "~>4.17"
```
ericywl, line 10: This change caused some dependencies in benchmark to be incompatible.

PR #19151, opened 2025-10-16
dependabot: support elastic docker registries

v1v:
## Motivation/summary

apm-server dependabot needs access to elastic private registry to bump the fips base image

## Checklist

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,4 +1,11 @@
 version: 2
+registries:
+  docker-elastic:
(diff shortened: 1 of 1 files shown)


v1v:
> change lgtm. Do we need to backport to 9.1 and 9.2 as well? Or just 8.19 because it will live for longer?

I see dependabot was configured to target two branches:
- `main`
- `8.19`

The configuration for `main` uses the `backport-active-9` labels.

I think we are good here

Issue #18949, opened 2025-10-07
replace apmelasticsearch with otel


kruskall:
the go agent is deprecated, we should replace apmelasticsearch with otel like we did for otel modules

TODO from https://github.com/elastic/apm-server/pull/18945#discussion_r2410710892

PR #19386, opened 2025-10-29
feat: replace apmelasticsearch with otel tracing

kruskall:
## Motivation/summary

replace go agent apmelasticsearch module with otel

replace go agent apmtest with otel manualreader in tests

set span attributes to ensure span type is set in kibana

## Checklist

## Related issues

Closes https://github.com/elastic/apm-server/issues/18949

### DIFF
--- a/internal/elasticsearch/tracing.go
+++ b/internal/elasticsearch/tracing.go
@@ -22,18 +22,9 @@ import (
 	"strings"
 
 	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/propagation"
 	"go.opentelemetry.io/otel/trace"
 	"go.opentelemetry.io/otel/trace/noop"
-)
-
-const (
-	// W3CTraceparentHeader is the standard W3C Trace-Context HTTP
-	// header for trace propagation.
-	W3CTraceparentHeader = "Traceparent"
-
-	// TracestateHeader is the standard W3C Trace-Context HTTP header
-	// for vendor-specific trace propagation.
-	TracestateHeader = "Tracestate"
 )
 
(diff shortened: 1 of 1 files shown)


internal/elasticsearch/tracing.go
```diff
@@ -0,0 +1,120 @@
...
+
+	return resp, err
+}
+
+func SetHeaders(req *http.Request, traceContext trace.SpanContext) {
+	headerValue := FormatTraceparentHeader(traceContext)
+	req.Header[W3CTraceparentHeader] = []string{headerValue}
+	if tracestate := traceContext.TraceState().String(); tracestate != "" {
+		req.Header[TracestateHeader] = []string{tracestate}
+	}
```
lahsivjar: Can we use the OTel's [propogator API](https://opentelemetry.io/docs/specs/otel/context/api-propagators/) to do this?

kruskall: Sorry, can you expand on this ? I'm not sure how this would work (keep in mind we're not using upstream otel tracer (yet) but the apmotel abstraction in the go agent)

lahsivjar: I was suggesting something like:

```
	var propagator propagation.TraceContext
	propagator.Inject(ctx, propagation.HeaderCarrier(req.Header))
```

Implementation for inject: https://github.com/open-telemetry/opentelemetry-go/blob/d1ebd7b9ff97c9f400098e7d89978814f47afab8/propagation/trace_context.go#L39-L65

lahsivjar: We can also set a global propagator alongside tracer provider in our code using `otel.SetTextMapPropagator` and then just use the global one directly to inject.

kruskall: Thanks, updated!

Issue #19608, opened 2025-11-20
ci: system tests failing due to disk usage limit reached


kruskall:
**APM Server version** (`apm-server version`):

**Description of the problem including expected versus actual behavior**:

Recently apm-server CI has been failing more often, the cause seems to be a flaky test: https://github.com/elastic/apm-server/issues/19198

but the frequency of failure has spiked to the point that it's almost impossible to merge a PR.

Upon further investigation it seems we're reaching the disk usage limit causing apm-serve to bypass tbs and index events by default hence failing the tbs test.

The ubuntu-latest image of the running is already significantly bloated and has ~75% disk usage (note: we set tbs disk usage limit to ~80% of available disk)

**Steps to reproduce**:

Please include a *minimal* but *complete* recreation of the problem,
including server configuration, agent(s) used, etc.  The easier you make it
for us to reproduce it, the more likely that somebody will take the time to
look at it.

 1. open PR
 2. watch system test fail
 3.

**Provide logs (if relevant)**:

```
{"log.level":"warn","@timestamp":"2025-11-20T14:43:33.872Z","log.logger":"beater.sampling","log.origin":{"function":"github.com/elastic/apm-server/x-pack/apm-server/sampling.(*Processor).ProcessBatch","file.name":"sampling/processor.go","file.line":127},"message":"processing trace failed, indexing by default","service.name":"apm-server","error":{"message":"disk usage threshold 0.80: configured limit reached (current: 63403094016, limit: 61509723750)"},"ecs.version":"1.6.0"}
```

kruskall:
Closing this since we did what we could on our side. The issue should be fixed on github action side.

PR #19614, opened 2025-11-21
test(TestTailSampling): set disk usage threshold to 90% to accommodate Ubuntu runner size

kruskall:
## Motivation/summary

Increase disk usage warning threshold to 90%.
The default Ubuntu GitHub Actions runner already consumes ~70% of space, so the previous 80% limit caused flaky tests

## Checklist

## How to test these changes

watch CI system tests

## Related issues

try to put a patch on https://github.com/elastic/apm-server/issues/19608 to unblock CI until we improve the disk usage

### DIFF
--- a/systemtest/apmservertest/config.go
+++ b/systemtest/apmservertest/config.go
@@ -146,21 +146,23 @@ type TailSamplingConfig struct {
 	Enabled            bool
 	Interval           time.Duration
 	Policies           []TailSamplingPolicy
-	DiskUsageThreshold float64 `json:"disk_usage_threshold"`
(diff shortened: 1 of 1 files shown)


systemtest/sampling_test.go
```diff
@@ -101,6 +101,7 @@ func TestTailSampling(t *testing.T) {
 					SampleRate: 0.5,
 				},
 			},
+			DiskUsageThreshold: 0.9,
```
carsonip, line 104: nit: do you mind adding a comment that it is to avoid failing on small disk CI?

Issue #17503, opened 2025-07-07
docs: Update TBS perf numbers after improvements


carsonip:
Update https://www.elastic.co/docs/solutions/observability/apm/transaction-sampling#_tail_based_sampling_performance_and_requirements 

TODO
- Change section from 9.0 to 9.0.3 to be specific and because it contains significant perf improvements for larger instances (PR: https://github.com/elastic/apm-server/pull/17254)
- Update deployment template to use ARM CPU Optimized for AWS because the x86 CPU optimized option is not available on ECH UI (Use c6gd instead of c6id)

carsonip:
Closes via https://github.com/elastic/docs-content/pull/4092

PR #19619, opened 2025-11-24
benchmark: Support ARM, add local scripts for parametrized benchmarks

carsonip:
## Motivation/summary

- Add ARM system profiles for benchmark
- Reduce benchmark flakiness
- Fix benchmark error handling and debug output
- Add scripts for producing TBS benchmark numbers in documentation https://www.elastic.co/docs/solutions/observability/apm/transaction-sampling#_tail_based_sampling_performance_and_requirements

## Checklist

## Related issues

Part of #17503

This PR generates doc change https://github.com/elastic/docs-content/pull/4092

### DIFF
--- a/testing/benchmark/Makefile
+++ b/testing/benchmark/Makefile
@@ -1,7 +1,7 @@
 TFVARS_SOURCE ?= terraform.tfvars.example
 
 # Automatically detect architecture based on instance type in TFVARS_SOURCE
-# ARM instance families: a1, t4g, c6g, c7g, m6g, m7g, r6g, r7g, x2gd
+# If we need to change this, remember to grep for other instance type checks in the codebase.
 DETECTED_ARCH = $(shell grep -E "standalone_apm_server_instance_size|standalone_moxy_instance_size|worker_instance_type" $(TFVARS_SOURCE) | head -1 | cut -d'=' -f2 | tr -d ' "' | grep -qE '^(a1|t4g|c6g|c7g|m6g|m7g|r6g|r7g|x2gd)' && echo "arm64" || echo "amd64")
 
 APMBENCH_PATH ?= ../../systemtest/cmd/apmbench


.github/workflows/benchmarks/analyze-artifacts.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
```
v1v, line 1: could we move these scripts to another location? This will avoid us, robots team, to review them, as part of the CODEOWNERS file

https://github.com/elastic/apm-server/blob/main/.github/CODEOWNERS#L5

.github/workflows/benchmarks.yml
```diff
@@ -171,14 +171,35 @@ jobs:
         id: init
         run: make init
 
+      - name: Detect instance architecture
+        id: detect-arch
+        run: |
+          # Extract instance type from the profile file
```
v1v: `question`: can we move this logic to a makefile instead? This will help with running it locally, and the CI will just run `make detect-artch` or whatever name we agree on.

You will need to use something like the above snippet to avoid issues with `>> "$GITHUB_OUTPUT"`
```bash
[ -n "${CI}" ] && echo "goarch=arm64" >> "$GITHUB_OUTPUT" || true
```

v1v (commented):
A few comments regarding the ownership/review model and simplicity in the GH workflows by using a Makefile or something similar

.github/workflows/benchmarks/analyze-artifacts.sh
carsonip: I was a bit conflicted, whether they should be placed closer to .github or closer to the benchmarks. My thought process was:
- Placing it in `.github` is slightly strange because it isn't run on github
- However, if anything in `.github/workflows/benchmarks.yml` change, these scripts likely need to be updated.
- Also, the scripts call `gh workflow`, so `.github` files are technically the intermediate step between these scripts and the actual benchmark implementation. So it makes less sense to place these scripts close to benchmark implementation, only for it to go through `benchmarks.yml` then back into the actual benchmark implementation.

In short, IMO it makes sense to be placed closer to `.github` files, but I agree that robots team doesn't need to be pinged on any changes here. Any suggestions?

.github/workflows/benchmarks.yml
carsonip: the goarch needs to be propagated to APMBENCH_GOARCH, APM_SERVER_GOARCH and MOXY_GOARCH. Getting goarch locally is not really useful. Either:
- we go all the way to set these env vars upon `make set-arch`; or
- we go half way to so that the benchmarks.yml looks like `echo "goarch=$(make detect-arch)" >> "$GITHUB_OUTPUT"` and keep the following usages of `${{ steps.detect-arch.outputs.goarch }}`.

carsonip: let me think more about this. The makefile can actually try to read from TFVARS_SOURCE and set `*_GOARCH`. That's the best case scenario and will be good for local use

carsonip: trying with b3ac004fced989df2951c3ea8d99fa4b36c5986c

carsonip: the run looks good: https://github.com/elastic/apm-server/actions/runs/19699722243

testing/benchmark/system-profiles/8GB_NVMe_ARM-x1zone.tfvars
```diff
@@ -0,0 +1,21 @@
...
+elasticsearch_size = "64g"
+# The number of AZs the Elasticsearch cluster should have.
+elasticsearch_zone_count = 2
+# APM server instance size
+apm_server_size = "8g"
+
+# Standalone
+
+standalone_apm_server_instance_size = "c6gd.xlarge"
+standalone_moxy_instance_size       = "c6g.2xlarge"
```
simitt, line 21: Not blocking, but potential future improvement: not a huge fan of repeating the majority of the settings between the profiles. It makes it hard to spot the actual differences and potential bugs

.github/workflows/benchmarks/analyze-artifacts.sh
simitt: I don't think there is an ideal solution here, but alse tend to keep this in the `.github/workflows/benchmarks` location as it becomes even harder to follow the flow otherwise. 
@v1v can your team be excepted as CODEOWNER for this subfolder?

carsonip: I'll merge this PR first before I log off for PTO. We can move it out of `.github` later if we later decide that's better. But I don't expect changes to these scripts to be frequent TBH

v1v: I'll create the releavnat entry in the CODEOWNERS to change the ownership for `.github/workflows/benchmarks`

v1v: althouth those scripts could be moved to `testing/benchmark/scripts`, that's the location for the Makefile and other bits and pieces

PR #19717, opened 2025-11-26
Add changelog for 8.19.8

ericywl:
Add changelog for 8.19.8


endorama (commented):
Should this release changelog be empty? If yes we usually add a note about it, "No significant changes.". Look on line 61 for an example.

Issue #17878, opened 2025-07-31
[APM Server/Integration] Report health on major errors


lucabelluccini:
It would be nice to expose some major health events to the Kibana UI without the need for the users to enable the whole monitoring and logging collection.

A key example is when TBS reaches the storage limit.

At the moment, those errors can be seen only if one actively looks at logs.

Once we have the status reported, it would be good to somehow expose it to the Kibana UI and/or Cloud Admin UI.

### Possible implementation via Fleet Protocol

Fleet exposes an API to report a status and health of an integration.

<img width="2093" height="556" alt="Image" src="https://github.com/user-attachments/assets/76850bec-e0bb-4c62-a894-24698357e60e" />

It would be nice to report `Degraded` or `Unhealthy` to Fleet based on errors encountered in APM Server/Integration.

It is feasible using https://github.com/elastic/beats/blob/main/libbeat/management/status/status.go 

For Filebeat, they use [`context.UpdateStatus()`](https://github.com/elastic/beats/blob/9190f776f5fb612a880cb013a9eec8a49fe17afd/filebeat/input/v2/input.go#L112).

Status changes should only be reported when the status actually changes (don't repeatedly send HEALTHY if the status is already HEALTHY).

The state can also contain hints on what to do (e.g. point to a documentation page for remediation).

As a safety measure, we could make this health reporting configurable at APM Integration level (e.g. `report health to fleet` or something like that). If this setting is enabled, APM Integration would not only log, but report the state to Fleet.

Advantages:
- Covers self-hosted and ECH APM Integrations
- Should be light to implement
- Removes the need for the customer to enable logging and metrics collection, at least for major errors

Disadvantage:
- Strongly tied to Fleet, which might not be reusable for the ECH Health API for Cloud Admin

raultorrecilla:
this is in review, @kruskall is there any development pending? @ericywl  as you were assigned to https://github.com/elastic/apm-server/issues/18084 I'm going to assign this one to you as well.

kruskall:
@carsonip reviewed the PR yesterday: https://github.com/elastic/apm-server/pull/19913

I'll address the review comments this week

ericywl:
I will try to follow up on this.

ericywl:
Update: Currently trying to test the changes on ECH using CFT but the changes aren't applied for some reason. Trying to debug it...

PR #19913, opened 2025-12-16
feat: report status on major health issues

kruskall:
## Motivation/summary

pass status reporter and report degraded state on health issues

## Checklist

## Related issues

Closes https://github.com/elastic/apm-server/issues/17878

### DIFF
--- a/x-pack/apm-server/sampling/processor_test.go
+++ b/x-pack/apm-server/sampling/processor_test.go
@@ -799,19 +799,19 @@ func TestReadSubscriberPositionFile(t *testing.T) {
 		{
 			name: "valid json",
 			setupFile: func(path string) error {
-				return os.WriteFile(path, []byte(`{}`), 0o644)
+				return os.WriteFile(path, []byte(`{}`), 0644)
 			},
 		},
 		{
 			name: "invalid json",
 			setupFile: func(path string) error {
-				return os.WriteFile(path, []byte(`not_json`), 0o644)
+				return os.WriteFile(path, []byte(`not_json`), 0644)
 			},
 		},
 		{
 			name: "bad perm",
 			setupFile: func(path string) error {
-				return os.WriteFile(path, []byte{}, 0o000)
+				return os.WriteFile(path, []byte{}, 0000)
 			},
 		},
 	} {


internal/beatcmd/beat.go
```diff
@@ -286,12 +286,14 @@ func openRegular(filename string) (*os.File, error) {
...
 func (b *Beat) Run(ctx context.Context) error {
 	defer b.Info.Logger.Sync()
+	var panicErr error
 	defer func() {
 		if r := recover(); r != nil {
 			b.Info.Logger.Fatalw("exiting due to panic",
 				"panic", r,
 				zap.Stack("stack"),
 			)
+			panicErr = fmt.Errorf("exiting due to panic: %v", r)
```
carsonip, line 296: realistically, does anything run after Fatalw though?

x-pack/apm-server/sampling/processor.go
```diff
@@ -203,11 +222,13 @@ func (p *Processor) processTransaction(event *modelpb.APMEvent) (report, stored
...
 		// Too many trace groups, drop the transaction.
-		p.rateLimitedLogger.Warn(`
+		msg := `
 Tail-sampling service group limit reached, discarding trace events.
 This is caused by having many unique service names while relying on
 sampling policies without service name specified.
-`[1:])
+`[1:]
+		p.rateLimitedLogger.Warn(msg)
+		p.statusReporter.UpdateStatus(status.Degraded, msg)
```
carsonip, line 231: I wonder if this "too many trace groups" degraded status would be almost immediately undone by any following spans, because spans walk a different code path vs transactions.

Is there any cost in flip-flopping between statuses?

carsonip (commented):
did a first pass. Sorry I'm not very familiar with the status report feature. High level q:
- Does it make sense for TBS report 1 status at a single point periodically instead of doing so per request? I'm asking because it is possible request A passes without issues while request B encounters issues
- Are there any costs if we flip-flop between statuses frequently, or if it generates an entry in fleet and ES every time it changes?
- Does the status reporter report back via push or pull? If we push, do we need to implement any kind of "rate limiting" or "debounce" within apm server code?
- Does it affect standalone apm-server?
- How are we testing this?

ericywl:
Closing, will create a new one later.

Issue #17772, opened 2025-07-22
TBS: potential data loss in race condition between event arrival and receiving decision


carsonip:
As TBS pebble implementation does not use db transaction and there are no locks, theoretically it is possible for the following sequence of events to happen.

1. goroutine A: a trace event T belonging to trace id 123 to arrive
2. goroutine A: confirm that no decision is made for trace id 123
3. goroutine B: make a sampling decision for trace id 123
4. goroutine B: publish all events for trace id 123
5. goroutine A: write trace event T to pebble
6. trace event T will sit in pebble until being deleted at TTL

Confirm if this is possible by adding a unit test, and if so, add a fix and measure the perf hit.

ericywl:
Update: Currently looking into reproducing this

ericywl:
@carsonip I'm not sure if my logic here is correct, but I wrote a test that sort-of confirms this: https://github.com/elastic/apm-server/pull/19948. It also requires the trace event T to not be sampled ofc. Hope you can help review it.

PR #19948, opened 2025-12-19
tbs: Fix potential data race

ericywl:
## Summary

Fix potential data race between `WriteTraceEvent` in `ProcessBatch` and `ReadTraceEvent` in the sampling goroutine. Closes https://github.com/elastic/apm-server/issues/17772.

### Baseline

```
goos: darwin
goarch: arm64
pkg: github.com/elastic/apm-server/x-pack/apm-server/sampling
cpu: Apple M4 Pro
BenchmarkProcess-14              3096828               357.5 ns/op
BenchmarkProcess-14              3284749               359.1 ns/op
BenchmarkProcess-14              3191538               353.3 ns/op
BenchmarkProcess-14              3333675               345.3 ns/op
BenchmarkProcess-14              3331615               345.4 ns/op
BenchmarkProcess-100             3439828               344.9 ns/op
BenchmarkProcess-100             3526257               325.7 ns/op
BenchmarkProcess-100             3461000               322.2 ns/op
BenchmarkProcess-100             3480249               394.0 ns/op
BenchmarkProcess-100             2877158               608.7 ns/op
```

### Single Mutex

```
goos: darwin
goarch: arm64
pkg: github.com/elastic/apm-server/x-pack/apm-server/sampling
cpu: Apple M4 Pro
BenchmarkProcess-14              3323389               337.5 ns/op
BenchmarkProcess-14              3482792               347.8 ns/op
BenchmarkProcess-14              3349486               333.5 ns/op
BenchmarkProcess-14              3437163               334.9 ns/op
BenchmarkProcess-14              3362293               333.8 ns/op
BenchmarkProcess-100             3516130               331.9 ns/op
BenchmarkProcess-100             3251674               339.8 ns/op
BenchmarkProcess-100             3432068               333.6 ns/op
BenchmarkProcess-100             3561802               368.2 ns/op
BenchmarkProcess-100             3290845               416.4 ns/op
```

### ShardLockReadWriter

```
goos: darwin
goarch: arm64
pkg: github.com/elastic/apm-server/x-pack/apm-server/sampling
cpu: Apple M4 Pro
BenchmarkProcess-14              3272188               346.3 ns/op
BenchmarkProcess-14              3415772               330.7 ns/op
BenchmarkProcess-14              3487447               333.2 ns/op
BenchmarkProcess-14              3470158               337.4 ns/op
BenchmarkProcess-14              3467367               338.6 ns/op
BenchmarkProcess-100             3626730               319.7 ns/op
BenchmarkProcess-100             3722044               369.1 ns/op
BenchmarkProcess-100             3123934               349.0 ns/op
BenchmarkProcess-100             3771914               381.9 ns/op
BenchmarkProcess-100             3462182               340.4 ns/op
```

### ShardLockReadWriter with RWMutex

```
goos: darwin
goarch: arm64
pkg: github.com/elastic/apm-server/x-pack/apm-server/sampling
cpu: Apple M4 Pro
BenchmarkProcess-14              3384834               343.1 ns/op
BenchmarkProcess-14              3284800               338.6 ns/op
BenchmarkProcess-14              3483038               348.2 ns/op
BenchmarkProcess-14              3217903               343.9 ns/op
BenchmarkProcess-14              3444128               334.4 ns/op
BenchmarkProcess-100             3115171               328.8 ns/op
BenchmarkProcess-100             3383545               329.2 ns/op
BenchmarkProcess-100             3069316               328.6 ns/op
BenchmarkProcess-100             3483396               334.0 ns/op
BenchmarkProcess-100             3359440               345.3 ns/op
```

### DIFF
--- a/x-pack/apm-server/sampling/processor.go
+++ b/x-pack/apm-server/sampling/processor.go
@@ -454,16 +454,17 @@ func (p *Processor) Run() error {
 				}
 			}
 
+			// We lock before WriteTraceSampled here to prevent race condition with IsTraceSampled from incoming events.
 			p.shardLock.Lock(traceID)
 			if err := p.eventStore.WriteTraceSampled(traceID, true); err != nil {
 				p.rateLimitedLogger.Warnf(
 					"received error writing sampled trace: %s", err,
 				)
 			}
+			p.shardLock.Unlock(traceID)
 
 			events = events[:0]
 			err = p.eventStore.ReadTraceEvents(traceID, &events)
-			p.shardLock.Unlock(traceID)
 			if err != nil {
 				p.rateLimitedLogger.Warnf(
 					"received error reading trace events: %s", err,
@@ -560,9 +561,6 @@ func newShardLock(numShards int) *shardLock {
 		panic("shardLock numShards must be greater than zero")
 	}
 	locks := make([]sync.RWMutex, numShards)
-	for i := 0; i < numShards; i++ {
-		locks[i] = sync.RWMutex{}
-	}
 	return &shardLock{locks: locks}
 }
 
--- a/x-pack/apm-server/sampling/processor_test.go
+++ b/x-pack/apm-server/sampling/processor_test.go
@@ -834,156 +834,8 @@ func TestGracefulShutdown(t *testing.T) {
 	assert.Equal(t, int(sampleRate*float64(totalTraces)), count)
 }
 
-type blockingRW struct {
-	ctx                 context.Context
-	unblockWriteEvent   chan struct{}
-	unblockWriteSampled chan struct{}
-	next                eventstorage.RW
-}
-
-func (m blockingRW) ReadTraceEvents(traceID string, out *modelpb.Batch) error {
-	return m.next.ReadTraceEvents(traceID, out)
-}
-
-func (m blockingRW) WriteTraceEvent(traceID, id string, event *modelpb.APMEvent) error {
-	select {
-	case <-m.ctx.Done():
-		return m.ctx.Err()
-	case <-m.unblockWriteEvent:
-		return m.next.WriteTraceEvent(traceID, id, event)
-	}
-}
-
-func (m blockingRW) WriteTraceSampled(traceID string, sampled bool) error {
-	select {
-	case <-m.ctx.Done():
-		return m.ctx.Err()
-	case <-m.unblockWriteSampled:
-		return m.next.WriteTraceSampled(traceID, sampled)
-	}
-}
-
-func (m blockingRW) IsTraceSampled(traceID string) (bool, error) {
-	return m.next.IsTraceSampled(traceID)
-}
-
-func (m blockingRW) DeleteTraceEvent(traceID, id string) error {
-	return m.next.DeleteTraceEvent(traceID, id)
-}
-
-func TestPotentialRaceCondition(t *testing.T) {
-	unblockWriteEvent := make(chan struct{}, 1)
-	unblockWriteSampled := make(chan struct{}, 1)
-
-	flushInterval := time.Second
-	tempdirConfig := newTempdirConfig(t)
-	timeoutCtx, cancel := context.WithTimeout(context.Background(), 10*flushInterval)
-	defer cancel()
-
-	tempdirConfig.Config.FlushInterval = flushInterval
-	// Wrap existing storage with blockingRW so we can control the flow.
-	tempdirConfig.Config.Storage = &blockingRW{
-		ctx:                 timeoutCtx,
-		unblockWriteEvent:   unblockWriteEvent,
-		unblockWriteSampled: unblockWriteSampled,
-		next:                tempdirConfig.Config.Storage,
-	}
-	// Collect reported transactions.
-	reported := make(chan modelpb.Batch)
-	tempdirConfig.Config.BatchProcessor = modelpb.ProcessBatchFunc(func(ctx context.Context, batch *modelpb.Batch) error {
-		select {
-		case <-ctx.Done():
-			return ctx.Err()
-		case reported <- batch.Clone():
-			return nil
-		}
-	})
-
-	processor, err := sampling.NewProcessor(tempdirConfig.Config, logptest.NewTestingLogger(t, ""))
-	require.NoError(t, err)
-	go processor.Run()
-	defer processor.Stop(context.Background())
-
-	// Send root transaction, which will be written to storage since write event is unblocked.
-	unblockWriteEvent <- struct{}{}
-	batch1 := modelpb.Batch{{
-		Trace: &modelpb.Trace{Id: "trace1"},
-		Transaction: &modelpb.Transaction{
-			Type:    "type",
-			Id:      "transaction1",
-			Sampled: true,
-		},
-	}}
-	err = processor.ProcessBatch(context.Background(), &batch1)
-	require.NoError(t, err)
-	assert.Len(t, batch1, 0)
-	monitoringtest.ExpectContainAndNotContainOtelMetrics(t, tempdirConfig.metricReader,
-		map[string]any{
-			"apm-server.sampling.tail.events.processed": 1,
-			"apm-server.sampling.tail.events.stored":    1,
-		},
-		[]string{
-			"apm-server.sampling.tail.events.sampled",
-		},
-	)
-
-	// Send child transaction, which will be blocked from writing to storage for now.
-	var wg sync.WaitGroup
-	wg.Go(func() {
-		batch2 := modelpb.Batch{{
-			Trace: &modelpb.Trace{Id: "trace1"},
-			Transaction: &modelpb.Transaction{
-				Type:    "type",
-				Id:      "transaction2",
-				Sampled: true,
-			},
-			ParentId: "transaction1",
-		}}
-		err = processor.ProcessBatch(context.Background(), &batch2)
-		require.NoError(t, err)
-		assert.Len(t, batch2, 0)
-	})
-
-	// Unblock write sampled so that the first transaction is published.
-	unblockWriteSampled <- struct{}{}
-	time.Sleep(2 * flushInterval)
-	// Unblock write event so that the second transaction is written to storage.
-	unblockWriteEvent <- struct{}{}
-	wg.Wait()
-	monitoringtest.ExpectContainAndNotContainOtelMetrics(t, tempdirConfig.metricReader,
-		map[string]any{
-			// This comes from second transaction being stored.
-			"apm-server.sampling.tail.events.stored":    1,
-			"apm-server.sampling.tail.events.processed": 1,
-		},
-		[]string{
-			"apm-server.sampling.tail.events.sampled",
-		},
-	)
-
-	select {
-	case batch := <-reported:
-		if len(batch) != 2 {
-			t.Fatal("expected two events in publish")
-		}
-		assert.Equal(t, batch[0].Transaction.Id, "transaction1")
-		assert.Equal(t, batch[1].Transaction.Id, "transaction2")
-	case <-timeoutCtx.Done():
-		t.Fatal("test timed out waiting for publish")
-	}
-	monitoringtest.ExpectContainAndNotContainOtelMetrics(t, tempdirConfig.metricReader,
-		map[string]any{
-			"apm-server.sampling.tail.events.sampled": 2,
-		},
-		[]string{
-			"apm-server.sampling.tail.events.processed",
-			"apm-server.sampling.tail.events.stored",
-		},
-	)
-}
-
 func TestPotentialRaceConditionConcurrent(t *testing.T) {
-	flushInterval := 5 * time.Second
+	flushInterval := 1 * time.Second
 	tempdirConfig := newTempdirConfig(t)
 	tempdirConfig.Config.FlushInterval = flushInterval
 	tempdirConfig.Config.Policies = []sampling.Policy{
@@ -1014,7 +866,7 @@ func TestPotentialRaceConditionConcurrent(t *testing.T) {
 			first := true
 			index := i * 100000000
 
-			timer := time.NewTimer(flushInterval + 2*time.Second)
+			timer := time.NewTimer(flushInterval * 2)
 			defer timer.Stop()
 
 			for {
@@ -1027,7 +879,7 @@ func TestPotentialRaceConditionConcurrent(t *testing.T) {
 				}
 
 				batch := modelpb.Batch{{
-					Trace: &modelpb.Trace{Id: fmt.Sprintf("trace%d", i)},
+					Trace: &modelpb.Trace{Id: "trace1"},
 					Transaction: &modelpb.Transaction{
 						Type:    "type",
 						Id:      fmt.Sprintf("transaction%08d", index),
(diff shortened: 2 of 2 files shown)


x-pack/apm-server/sampling/processor_test.go
```diff
@@ -830,6 +832,209 @@ func TestGracefulShutdown(t *testing.T) {
...
+	// to kick second transaction out from getting sampled.
+	var wg sync.WaitGroup
+	tx2Done := make(chan struct{})
+	wg.Go(func() {
+		batch2 := modelpb.Batch{{
+			Trace: &modelpb.Trace{Id: "trace1"},
+			Transaction: &modelpb.Transaction{
+				Type:    "type",
+				Id:      "transaction2",
+				Sampled: true,
```
carsonip: both txn1 and txn2 share the same trace id but none of them have a parent transaction. This is probably not a case that we can / should correctly handle. Did you mean to set txn2 parent id to txn1?

x-pack/apm-server/sampling/processor_test.go
```diff
@@ -830,6 +832,209 @@ func TestGracefulShutdown(t *testing.T) {
...
+			"apm-server.sampling.tail.events.stored":    1,
+		},
+		[]string{
+			"apm-server.sampling.tail.events.sampled",
+		},
+	)
+
+	// Send second transaction, which will be blocked from writing to storage for now.
+	// At the same time, send some other transaction with high event duration
+	// to kick second transaction out from getting sampled.
```
carsonip: as explained below, both txn1 and txn2 are root transactions which will enter the pool, and that's why you needed to "kick 2nd txn out of the pool". But if txn2 has a parent, it wouldn't go into the pool in the first place.

carsonip (commented):
Thanks. The setup looks about right, but the specific case that needs to be tested is not what I'm thinking about.

In this PR, there are txn1 and txn2 both part of trace1, which don't make sense to me. 1 trace should only have 1 root transaction. In your current setup, what should happen to txn2 is undefined.

To be clearer, what I'd like to test is similar: txn1 is root txn of trace1, and txn2 is a child of txn1.

At time t1: apm server receives txn1
t2: background sampling goroutine: apm server makes sampling decision for txn1
t2': apm server receives txn2
t3: background sampling goroutine: marks trace1 as sampled

^ the above is a race, because apm server receives txn2 between t2 and t3, and the result is txn2 is lost forever. If it happens either before t2 or after t3, txn2 is exported correctly.

It gets a bit theoretical but I believe it is possible. Lmk if you have any questions.

x-pack/apm-server/sampling/processor_test.go
ericywl: Updated to fix this 👍

x-pack/apm-server/sampling/processor_test.go
```diff
@@ -830,6 +832,177 @@ func TestGracefulShutdown(t *testing.T) {
...
+			"apm-server.sampling.tail.events.stored": 1,
+		},
+		[]string{
+			"apm-server.sampling.tail.events.sampled",
+			"apm-server.sampling.tail.events.processed",
+		},
+	)
+
+	// Unblock write sampled to check if second transaction gets sampled.
+	// It should not be sampled since the trace is already be sampled previously.
```
carsonip: comment is inaccurate. The expectation is that 2nd transaction should also be sampled and sent to next processor as the trace is sampled, no matter it arrives before or after the root transaction (+ sampling decision delay).

x-pack/apm-server/sampling/processor_test.go
```diff
@@ -830,6 +832,177 @@ func TestGracefulShutdown(t *testing.T) {
...
+	assert.NoError(t, tempdirConfig.Config.DB.Flush())
+	close(reported)
+	db := tempdirConfig.Config.DB
+	reader := newUnlimitedReadWriter(db)
+
+	var batch modelpb.Batch
+	assert.NoError(t, reader.ReadTraceEvents("trace1", &batch))
+	assert.Len(t, batch, 2)
+	assert.Equal(t, batch[0].Transaction.Id, "transaction1")
+	assert.Equal(t, batch[1].Transaction.Id, "transaction2")
```
carsonip: I'm afraid the end state of the local db is always txn1 + txn2, no matter the race happened or not, because trace events are never deleted by local sampling decisions. This assertion is no-op other than revalidating assumptions.

x-pack/apm-server/sampling/processor_test.go
ericywl: Yup, this is just explaining what's happening. I will convert this to a failure test when implementing the fix.

x-pack/apm-server/sampling/processor.go
```diff
@@ -449,47 +491,41 @@ func (p *Processor) Run() error {
...
+				if !ok {
+					return nil
+				}
+				// TODO(eric): Update this to be configurable.
+				if err := p.waitForOngoingTrace(gracefulContext, traceID, 10, 500*time.Millisecond); err != nil {
+					return err
 				}
+				// This only happens to local sampled trace IDs due to potential race condition
+				// between receiving the transaction / span and making the sampling decision.
+				p.readAndReportEvents(gracefulContext, traceID, false)
```
carsonip: q: is this critical section protected though? Is is still possible to race inside this readAndReportEvents func?

ericywl: Hmm the `ReadTraceEvents` is not protected. I was thinking of using per-trace mutex but thought it might be overkill 🤔

x-pack/apm-server/sampling/eventstorage/rw.go
```diff
@@ -139,3 +141,93 @@ func (s StorageLimitReadWriter) DeleteTraceEvent(traceID, id string) error {
...
+	rw.mu.Lock()
+	defer rw.mu.Unlock()
+	return rw.rw.WriteTraceSampled(traceID, sampled)
+}
+
+func (rw *lockedReadWriter) IsTraceSampled(traceID string) (bool, error) {
+	rw.mu.Lock()
+	defer rw.mu.Unlock()
+	return rw.rw.IsTraceSampled(traceID)
+}
```
carsonip: I acknowledge 8.x doesn't have the optimization to use a RWMutex, but given the ingest hot path is IsTraceSampled->WriteTraceEvent, I wonder if swapping it to RWMutex would further reduce lock contention. However, it may or may not show up in the benchmark, depending on GOMAXPROCS, as well as trace sampled hit rate (which will be 0 because the benchmark always generates new trace ID)

carsonip (commented):
thanks, the approach looks good now

x-pack/apm-server/sampling/eventstorage/rw.go
ericywl: I updated the code and benchmark results. Not that much of a difference with our current benchmark.

x-pack/apm-server/sampling/processor_test.go
```diff
@@ -830,6 +832,154 @@ func TestGracefulShutdown(t *testing.T) {
...
+
+func (m blockingRW) IsTraceSampled(traceID string) (bool, error) {
+	return m.next.IsTraceSampled(traceID)
+}
+
+func (m blockingRW) DeleteTraceEvent(traceID, id string) error {
+	return m.next.DeleteTraceEvent(traceID, id)
+}
+
+func TestPotentialRaceCondition(t *testing.T) {
```
carsonip: nit: a more descriptive name, or perhaps a comment explaining the race?

carsonip (commented):
I'm terribly sorry - Yes the PR in its existing state will address new race conditions introduced in 9.0 due to the lack of db transactions, but CMIIW theoretically there's another type of race that is inherent in the RW design, which is also in 8.x, where:
- goroutine A ProcessBatch: IsTraceSampled(traceID) returns ErrNotFound
- background goroutine B responsible for publishing: WriteTraceSampled(traceID, true)
- background goroutine B responsible for publishing: ReadTraceEvents(traceID)
- goroutine A ProcessBatch: WriteTraceEvent(traceID, event1)

In this case, `event1` will be written to DB and dropped silently.

Maybe we'll have to zoom out and rethink this. Either:
- we'll give up addressing flush time races; or
- introduce processor level locking; or
- implement some less expensive handling that will help us get back events fallen victim to this race; or
- merge this PR as is to resolve this kind of race as an improvement, but create a follow up issue to address this newly identified kind of race. I don't want us to merge this PR with the impression that we've already fixed publish time race.

thoughts?

ericywl:
In that case, my previous solution (on top of ShardLockRW) should be able to catch this. The publishing will be deferred to another goroutine that waits until `event1` is written.

carsonip (commented):
> In that case, my previous solution (on top of ShardLockRW) should be able to catch this. The publishing will be deferred to another goroutine that waits until event1 is written.

I studied f187e90e15d6dd697d26c740bed8b7daf58039ac. My understanding is that it increments and decrements a per-trace-id counter before and after WriteTraceEvent in the ingest goroutine. And in the background publishing goroutine, between WriteTraceSampled and ReadTraceEvents it checks if the counter is 0. If it isn't 0, retry later.

The issue with this is it doesn't prevent the following sequence of events:
- t1: ingest goroutine A IsTraceSampled
- t2: background goroutine B WriteTraceSampled
- t3: background goroutine B performs counter==0 check
- t4: background goroutine B ReadTraceEvents
- t4': ingest goroutine A +1 counter, WriteTraceEvent, -1 counter

Race happens and there is data loss when t4 < t4'. Therefore, in terms of correctness, f187e90e15d6dd697d26c740bed8b7daf58039ac isn't race proof.

(On a side note, if the +1 counter happens before IsTraceSampled instead of WriteTraceEvent, I _think_ it might be correct. But even in that case I'd prefer a simpler design with performance and memory implications that are easier to reason about.)

I have some ideas on how to fix it and let's take it offline. In any case we might want to have a more generic test (in addition to / replacing the existing one) that sends a lot of events at around sampling decision time. It may not be deterministic but will give us confidence on whether this class of race conditions is eliminated, without specifying the exact sequence.

x-pack/apm-server/sampling/processor_test.go
```diff
@@ -830,6 +834,233 @@ func TestGracefulShutdown(t *testing.T) {
...
```
carsonip: q: what are your thoughts on this? should we keep this test?

x-pack/apm-server/sampling/processor_test.go
```diff
@@ -830,6 +834,233 @@ func TestGracefulShutdown(t *testing.T) {
...
+		},
+		[]string{
+			"apm-server.sampling.tail.events.processed",
+			"apm-server.sampling.tail.events.stored",
+		},
+	)
+}
+
+func TestPotentialRaceConditionConcurrent(t *testing.T) {
+	flushInterval := 5 * time.Second
```
carsonip: it can be way shorter, in ms or max 1s, because this affects total test time

x-pack/apm-server/sampling/processor_test.go
```diff
@@ -830,6 +834,233 @@ func TestGracefulShutdown(t *testing.T) {
...
+		})
+	}
+
+	require.NoError(t, eg.Wait())
+	require.NoError(t, processor.Stop(context.Background()))
+
+	reportedMu.Lock()
+	defer reportedMu.Unlock()
+	reportedPlusLateArrivals := int64(len(reported)) + lateArrivals.Load()
+	assert.Equal(t, reportedPlusLateArrivals, processed.Load())
```
carsonip:
```suggestion
	assert.Equal(t, processed.Load(), reportedPlusLateArrivals)
```
nit

x-pack/apm-server/sampling/processor_test.go
```diff
@@ -830,6 +834,233 @@ func TestGracefulShutdown(t *testing.T) {
...
+func TestPotentialRaceConditionConcurrent(t *testing.T) {
+	flushInterval := 5 * time.Second
+	tempdirConfig := newTempdirConfig(t)
+	tempdirConfig.Config.FlushInterval = flushInterval
+	tempdirConfig.Config.Policies = []sampling.Policy{
+		{SampleRate: 1.0},
+	}
+
+	var reportedMu sync.Mutex
+	reported := map[string]struct{}{}
```
carsonip, line 846: nit: simplify to atomic int?

x-pack/apm-server/sampling/processor_test.go
```diff
@@ -830,6 +834,233 @@ func TestGracefulShutdown(t *testing.T) {
...
+
+	var processed atomic.Int64
+	var lateArrivals atomic.Int64
+	eg, ctx := errgroup.WithContext(context.Background())
+	for i := 0; i < 1000; i++ {
+		eg.Go(func() error {
+			first := true
+			index := i * 100000000
+
+			timer := time.NewTimer(flushInterval + 2*time.Second)
```
carsonip:
```suggestion
			timer := time.NewTimer(flushInterval * 2)
```

x-pack/apm-server/sampling/processor.go
```diff
@@ -536,3 +550,40 @@ func sendTraceIDs(ctx context.Context, out chan<- string, traceIDs []string) err
...
+}
+
+func newShardLock(numShards int) *shardLock {
+	if numShards <= 0 {
+		panic("shardLock numShards must be greater than zero")
+	}
+	locks := make([]sync.RWMutex, numShards)
+	for i := 0; i < numShards; i++ {
+		locks[i] = sync.RWMutex{}
+	}
```
carsonip:
```suggestion
```
nit: zero value is ready to use

x-pack/apm-server/sampling/processor.go
```diff
@@ -443,14 +454,17 @@ func (p *Processor) Run() error {
...
 			if err := p.eventStore.WriteTraceSampled(traceID, true); err != nil {
 				p.rateLimitedLogger.Warnf(
 					"received error writing sampled trace: %s", err,
 				)
 			}
 
 			events = events[:0]
-			if err := p.eventStore.ReadTraceEvents(traceID, &events); err != nil {
+			err = p.eventStore.ReadTraceEvents(traceID, &events)
+			p.shardLock.Unlock(traceID)
```
carsonip: CMIIW but I believe it is sufficient to just lock `WriteTraceSampled` without locking `ReadTraceEvents`, given we know if `IsTraceSampled` returns true, we shortcircuit and write to batch processor immediately. Holding the write lock for as short as possible would be good for perf.

x-pack/apm-server/sampling/processor.go
```diff
@@ -443,14 +454,17 @@ func (p *Processor) Run() error {
 				}
 			}
 
+			p.shardLock.Lock(traceID)
```
carsonip, line 458: do you mind adding a comment about why we lock, and what scenario we are trying to prevent?

x-pack/apm-server/sampling/processor_test.go
```diff
@@ -830,6 +834,233 @@ func TestGracefulShutdown(t *testing.T) {
...
+				select {
+				case <-timer.C:
+					return nil
+				case <-ctx.Done():
+					return ctx.Err()
+				default:
+				}
+
+				batch := modelpb.Batch{{
+					Trace: &modelpb.Trace{Id: fmt.Sprintf("trace%d", i)},
```
carsonip: it'll be way easier to hit the race if different goroutines all share the same trace id, likely before the loop (remember to increment the processed count).

x-pack/apm-server/sampling/processor_test.go
ericywl: I will remove it since the concurrent one catches it too.

x-pack/apm-server/sampling/processor_test.go
ericywl: It's to deduplicate the transaction IDs.

PR #20095, opened 2026-01-12
ci: fix version validation on prepare-release

endorama:
Trying to run patch release workflow for `9.1.10` [resulted](https://github.com/elastic/apm-server/actions/runs/20914244789/job/60083823300) in a failure: "version is not a patch one but a minor (only support for <major>.<minor>.[1-9]+[0-9]*)"

The failure is not correct, this is a legit patch version.

Update the check to use `.0` to validate patch version. Update the check to use `.0` also to validate minors. Update wording to remove reference to other version, as they may mislead.

### DIFF
--- a/.github/workflows/prepare-release/action.yml
+++ b/.github/workflows/prepare-release/action.yml
@@ -52,23 +52,23 @@ runs:
     - name: validate version format (major only)
       if: ${{ inputs.type == 'major' && ! endsWith(inputs.version, '.0.0') }}
       run: |-
-        FAILURE_MESSAGE='version is not a major one but a patch (only support for <major>.0.0)'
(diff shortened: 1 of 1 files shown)


endorama:
I don't think this needs to be backported as per [docs](https://github.com/elastic/apm-server/blob/main/dev_docs/RELEASES.md#patch-release) we always run this workflow from `main`.

endorama:
I was wrong on the backport. The same issue exists in branch `8.19` and would thus affect all `8.19` patch releases.

For an unfortunate coincidence we are also releasing `8.19.10`. Backporting to 8.19.

endorama:
@mergify backport to 8.19

Issue #20119, opened 2026-01-14
iserver-test: Change in EC API behavior


ericywl:
## Problem

The EC API `https://api.elastic-cloud.com/api/v1/stack/versions` no longer returns already released version SNAPSHOTs. This causes our current `genpath` script to fail since we expect there to be SNAPSHOTs, and will have to be updated to work with the current EC API behavior.

## Proposed Solution

Instead of generating upgrade paths that only consist of SNAPSHOTs, we should use released versions and upgrade to SNAPSHOTs. For example, previous upgrade path like `8.19.10-SNAPSHOT -> 9.1.0-SNAPSHOT` would become `8.19.10 -> 9.1.0-SNAPSHOT` instead. If there are no SNAPSHOTs for an active version, we should simply just skip testing for that version.

ericywl:
The API change has been partially reverted, and the test mostly passes now: https://github.com/elastic/apm-server/actions/runs/21132218317. The failures are due to missing 7.17 SNAPSHOT, which is expected because it is now EOL. I will update the test to remove the 7.17 test instead.

PR #20126, opened 2026-01-15
iserver-test: Fix genpath script due to recent EC API change

ericywl:
Closes https://github.com/elastic/apm-server/issues/20119.

The EC API that we use to get available stack versions was recently changed to remove previously release SNAPSHOTs. This caused our integration server test to fail due to enforcing the upgrade path to be full SNAPSHOTs (for legacy reason, there's no real explanation for it) e.g. `9.2.0-SNAPSHOT -> 9.3.0-SNAPSHOT`.

Since SNAPSHOTs are now only available for future releases, we will only use SNAPSHOTs as the tested version, and the previous versions in the upgrade path will use released ones instead. In our previous example, we want to test `9.3.0-SNAPSHOT`, so the upgrade path will instead be `9.2.0 -> 9.3.0-SNAPSHOT` since `9.2.0` is a released version.

### DIFF
--- a/integrationservertest/main_test.go
+++ b/integrationservertest/main_test.go
@@ -24,8 +24,6 @@ import (
 	"os"
 	"testing"
 
-	"github.com/stretchr/testify/require"
-
 	"github.com/elastic/apm-server/integrationservertest/internal/ech"
 )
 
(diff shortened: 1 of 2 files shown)


ericywl:
> q: is this a valid upgrade path though? I am trying to upgrade from 9.2.x to 9.3.0-SNAPSHOT manually and no luck so far

That's weird, it should be since we get the valid upgrade paths from the EC API itself. I will test it out myself and check with the ECH(?) team.

ericywl:
I tried the iserver-test and there seems to be an error with Kibana?
```
found deployment plan errors: deployment [5b46d9a42c48739e3bc5e4d0524720fe] - [integrations_server][ec88aeb196764af7bba49ba3085d1f97]: caught error: "Plan change failed: Cluster not reachable: [kibana]... Please validate the cluster is in a healthy state and retry."
```

I will try to follow up on this next working day. On the other hand, the smoke test is also broken due to the API change: https://github.com/elastic/apm-server/pull/20137.

PR #20137, opened 2026-01-16
smoke: Fix smoke test issue due to EC API change

ericywl:
The EC API that we use to get available stack versions was recently changed to remove previously release SNAPSHOTs. This causes some OS smoke tests to fail since we use SNAPSHOTs for all of the tests and some of them are missing. This PR fixes the issue by falling back to released versions instead of SNAPSHOTs if they are not available.

### DIFF
--- a/testing/smoke/lib.sh
+++ b/testing/smoke/lib.sh
@@ -9,7 +9,7 @@ get_versions() {
         return 1
     fi
     # initialize terraform, so we can obtain the configured region.
-    # terraform_init
+    terraform_init
 
     local REGION=$(echo var.region | terraform console | tr -d '"')
     local EC_VERSION_ENDPOINT="https://cloud.elastic.co/api/v1/regions/${REGION}/stack/versions?show_deleted=false&show_unusable=false"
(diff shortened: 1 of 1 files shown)


testing/smoke/lib.sh
```diff
@@ -21,33 +21,24 @@ get_versions() {
...
     if [[ -z ${EC_API_KEY} ]]; then
         echo "-> ESS API Key not set, please set the EC_API_KEY environment variable."
         return 1
     fi
     # initialize terraform, so we can obtain the configured region.
-    terraform_init
+    # terraform_init
 
-    local REGION=$(echo var.region | terraform console | tr -d '"')
+    # local REGION=$(echo var.region | terraform console | tr -d '"')
```
endorama: Do we plan to use this terraform code again? If not, shouldn't we remove it?

```suggestion
```

The terraform code we are initializing here could be removed (at least partially)?

ericywl: Oh no I made a mistake here

ericywl:
Smoke test should work as per normal now that the API change has been partially reverted.

PR #20151, opened 2026-01-19
iserver-test: Remove 7.17 standalone tests

ericywl:
Closes https://github.com/elastic/apm-server/issues/20184.

7.17 is now EOL, so we no longer need the 7.17 tests. From testing, it was not possible to disable integrations server when creating latest 8.x versions. 

Based on dashboard https://overview.elastic-cloud.com/app/r/s/VwuAL, there's only 12 7.x deployments left.

### DIFF
--- a/integrationservertest/README.md
+++ b/integrationservertest/README.md
@@ -13,7 +13,6 @@ The following is a simplified directory structure of integration server test.
    |- internal/
    |- scripts/
    |- main_test.go
-   |- standalone_test.go
    |- upgrade_test.go
 ```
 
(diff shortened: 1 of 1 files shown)


v1v (dismissed):

endorama:
It's still possible to migrate to managed in 8.x, which is still supported, so we may need to update these tests instead of removing them.

Have you assess how many customers may still follow that flow?

inge4pres:
> Have you assess how many customers may still follow that flow?

@endorama there are 12 deployments in ECH, I guess it is safe for us to remove the test

PR #20166, opened 2026-01-20
Report status on major health issues

ericywl:
## Motivation/summary

Pass status reporter and report degraded state on health issues

## Checklist

## How to test these changes

1. Deploy the custom APM server binary to ECH CFT:
```
cd testing/cloud
export EC_API_KEY=<REDACTED>
make
make apply
```

Note: In order to test this, I additionally modified the `ProcessBatch` function to always have error e.g. `err = errors.New("this is forced error").

2. Enable TBS in the Elastic APM integration via Kibana:

Screenshot 2026-02-03 at 5 12 44 PM

3. Send some traces to the APM endpoint:

```go
func TestTrace(t *testing.T) {
	serverURL, err := url.Parse(<APM_ENDPOINT>)
	if err != nil {
		t.Fatal(err)
	}
	secretToken := <REDACTED>

	httpTransport, err := transport.NewHTTPTransport(transport.HTTPTransportOptions{
		ServerURLs:  []*url.URL{serverURL},
		SecretToken: secretToken,
	})
	if err != nil {
		t.Fatal(err)
	}
	tracer, err := apm.NewTracerOptions(apm.TracerOptions{
		Transport: httpTransport,
	})
	if err != nil {
		t.Fatal(err)
	}
	t.Cleanup(tracer.Close)

	for i := 0; i < 10; i++ {
		tracer.StartTransaction("name", "type").End()
	}
	tracer.Flush(nil)
}
```

4. Check the APM health on Fleet:

Screenshot 2026-02-03 at 5 09 48 PM

## Related issues

Closes https://github.com/elastic/apm-server/issues/17878.

### DIFF
--- a/x-pack/apm-server/sampling/processor.go
+++ b/x-pack/apm-server/sampling/processor.go
@@ -18,12 +18,11 @@ import (
 	"go.opentelemetry.io/otel/metric"
 	"golang.org/x/sync/errgroup"
 
+	"github.com/elastic/beats/v7/libbeat/management/status"
 	"github.com/elastic/beats/v7/x-pack/libbeat/statusreporterhelper"
+	"github.com/elastic/elastic-agent-libs/logp"
 
 	"github.com/elastic/apm-data/model/modelpb"
-
-	"github.com/elastic/beats/v7/libbeat/management/status"
-	"github.com/elastic/elastic-agent-libs/logp"
 
 	"github.com/elastic/apm-server/internal/logs"
 	"github.com/elastic/apm-server/x-pack/apm-server/sampling/eventstorage"


ericywl:
Some comments in the previous PR by @carsonip:

```
Did a first pass. Sorry I'm not very familiar with the status report feature. High level q:

- Does it make sense for TBS report 1 status at a single point periodically instead of doing so per request? I'm asking because it is possible request A passes without issues while request B encounters issues
- Are there any costs if we flip-flop between statuses frequently, or if it generates an entry in fleet and ES every time it changes?
- Does the status reporter report back via push or pull? If we push, do we need to implement any kind of "rate limiting" or "debounce" within apm server code?
- Does it affect standalone apm-server?
- How are we testing this?
```

ericywl:
> Does it make sense for TBS report 1 status at a single point periodically instead of doing so per request? I'm asking because it is possible request A passes without issues while request B encounters issues

Judging by some of the other usages of `StatusReporter` in libbeat, it seems that reporting status per request is not out of the usual.

> Are there any costs if we flip-flop between statuses frequently, or if it generates an entry in fleet and ES every time it changes?
> Does the status reporter report back via push or pull? If we push, do we need to implement any kind of "rate limiting" or "debounce" within apm server code?

The status is stored locally by the `StatusReporter` and pushed out in intervals, so I think we don't have do anything of the sort.

> Does it affect standalone apm-server?
> How are we testing this?

Not quite sure about these questions yet.

x-pack/apm-server/sampling/processor.go
```diff
@@ -141,6 +156,10 @@ func (p *Processor) ProcessBatch(ctx context.Context, batch *modelpb.Batch) erro
 			i--
 		}
 
+		if report || stored {
```
carsonip:
```suggestion
		if stored {
```
if we're trying to capture a recovery from storage issues, `report` is not part of the concern. Also, adding `report` to the condition actually immediately undoes the Degraded status from the `DiscardOnWriteFailure==false`, which is the default.

x-pack/apm-server/sampling/processor.go
```diff
@@ -211,11 +230,13 @@ func (p *Processor) processTransaction(event *modelpb.APMEvent) (report, stored
...
 		// Too many trace groups, drop the transaction.
-		p.rateLimitedLogger.Warn(`
+		msg := `
 Tail-sampling service group limit reached, discarding trace events.
 This is caused by having many unique service names while relying on
 sampling policies without service name specified.
-`[1:])
+`[1:]
+		p.rateLimitedLogger.Warn(msg)
+		p.statusReporter.UpdateStatus(status.Degraded, msg)
```
carsonip: not super sure about this, because if a subsequent request comes in that doesn't hit errTooManyTraceGroups because the request belongs to an existing trace group, it'll return stored == true and immediately undo the degraded state.

ericywl: Gotcha, I will think about this for a bit to come up with something.

x-pack/apm-server/sampling/processor.go
```diff
@@ -124,13 +136,16 @@ func (p *Processor) ProcessBatch(ctx context.Context, batch *modelpb.Batch) erro
 		if err != nil {
 			failed = true
 			stored = false
+			msg := ""
```
carsonip:
```suggestion
			var msg string
```
nit: style

x-pack/apm-server/sampling/processor_test.go
```diff
@@ -1023,3 +1079,8 @@ func waitFileModified(tb testing.TB, filename string, after time.Time) ([]byte,
 func newUnlimitedReadWriter(sm *eventstorage.StorageManager) eventstorage.RW {
 	return sm.NewReadWriter(0, 0)
 }
+
+type noopStatusReport struct{}
+
+func (noopStatusReport) UpdateStatus(status.Status, string) {
+}
```
carsonip, line 1143: since we're already mocking this, can we check that the function is called with what we're expecting when storage limit is exceeded?

testing/docker/elastic-agent/Dockerfile
```diff
@@ -4,4 +4,4 @@ FROM --platform=$TARGETPLATFORM ${ELASTIC_AGENT_IMAGE}
 ARG STACK_VERSION # e.g. 8.5.0-SNAPSHOT
 ARG VCS_REF_SHORT # e.g. abc123
 ARG TARGETARCH
-ONBUILD COPY --chmod=0755 --chown=elastic-agent apm-server-linux-${TARGETARCH} ./data/elastic-agent-${VCS_REF_SHORT}/components/apm-server
+COPY --chmod=0755 --chown=elastic-agent apm-server-linux-${TARGETARCH} ./data/elastic-agent-${VCS_REF_SHORT}/components/apm-server
```
carsonip, line 7: thanks for digging into the bottom of the docker image override issue. It appears to me as a regression from https://github.com/elastic/apm-server/pull/18466

Do you mind creating a separate PR to fix this issue to keep this PR scoped to EA report status?

ericywl: Got it! https://github.com/elastic/apm-server/pull/20256

carsonip (approved):
`x-pack/apm-server/sampling/processor.go` imports look unsorted

carsonip (approved):

Issue #20189, opened 2026-01-22
test: Flaky Test_warmup


carsonip:
Example run: https://github.com/elastic/apm-server/actions/runs/21253562786/job/61161576505?pr=20186

```
 === RUN   Test_warmup
=== RUN   Test_warmup/1_agent_2s
=== RUN   Test_warmup/4_agent_2s
    main_test.go:115: 
        	Error Trace:	/home/runner/work/apm-server/apm-server/systemtest/benchtest/main_test.go:115
        	Error:      	Received unexpected error:
        	            	received error waiting for server inactive: context deadline exceeded
        	Test:       	Test_warmup/4_agent_2s
    main_test.go:116: 
        	Error Trace:	/home/runner/work/apm-server/apm-server/systemtest/benchtest/main_test.go:116
        	Error:      	"0" is not greater than "4"
        	Test:       	Test_warmup/4_agent_2s
--- FAIL: Test_warmup (95.76s)
    --- PASS: Test_warmup/1_agent_2s (2.48s)
    --- FAIL: Test_warmup/4_agent_2s (93.28s)
```

isaacaflores2:
Francesco has passed the baton over to me. I will look into this test failure next

raultorrecilla:
following up on the discussion in the weekly meeting this morning, @inge4pres mentioned that this might be useless?? cc @lahsivjar  @isaacaflores2 your thoughts?

Issue #20098, opened 2026-01-12
test: Flaky TestSubscribeSampledTraceIDsError


carsonip:
TestSubscribeSampledTraceIDsError failing on CI

```
=== RUN   TestSubscribeSampledTraceIDsError
=== RUN   TestSubscribeSampledTraceIDsError/stats=500
    logger.go:146: 2026-01-06T15:55:34.014Z	ERROR	sampling	error searching for trace IDs	{"error": "index stats request failed with status code 500: ", "position": {}}
    logger.go:146: 2026-01-06T15:55:34.015Z	ERROR	sampling	error searching for trace IDs	{"error": "index stats request failed with status code 500: ", "position": {}}
    logger.go:146: 2026-01-06T15:55:34.016Z	ERROR	sampling	error searching for trace IDs	{"error": "index stats request failed with status code 500: ", "position": {}}
    logger.go:146: 2026-01-06T15:55:34.017Z	ERROR	sampling	error searching for trace IDs	{"error": "index stats request failed with status code 500: ", "position": {}}
    logger.go:146: 2026-01-06T15:55:34.018Z	ERROR	sampling	error searching for trace IDs	{"error": "index stats request failed with status code 500: ", "position": {}}
    logger.go:146: 2026-01-06T15:55:34.019Z	ERROR	sampling	error searching for trace IDs	{"error": "index stats request failed with status code 500: ", "position": {}}
    logger.go:146: 2026-01-06T15:55:34.020Z	ERROR	sampling	error searching for trace IDs	{"error": "index stats request failed with status code 500: ", "position": {}}
    logger.go:146: 2026-01-06T15:55:34.021Z	ERROR	sampling	error searching for trace IDs	{"error": "index stats request failed with status code 500: ", "position": {}}
    logger.go:146: 2026-01-06T15:55:34.022Z	ERROR	sampling	error searching for trace IDs	{"error": "index stats request failed with status code 500: ", "position": {}}
=== RUN   TestSubscribeSampledTraceIDsError/stats=429
    logger.go:146: 2026-01-06T15:55:34.025Z	WARN	sampling	error searching for trace IDs	{"error": "index stats request failed with status code 429: ", "position": {}}
    logger.go:146: 2026-01-06T15:55:34.026Z	WARN	sampling	error searching for trace IDs	{"error": "index stats request failed with status code 429: ", "position": {}}
[...]

Issue #19715, opened 2025-11-26
test: flaky TestRunManager_Reloader_newRunnerError data race


carsonip:
TestRunManager_Reloader_newRunnerError is flaky e.g. https://github.com/elastic/apm-server/actions/runs/19664829682/job/56319090910?pr=19687

```
WARNING: DATA RACE
Write at 0x00c0001bc20b by goroutine 151:
  testing.tRunner.func1()
      /opt/hostedtoolcache/go/1.25.4/x64/src/testing/testing.go:1921 +0x904
  runtime.deferreturn()
      /opt/hostedtoolcache/go/1.25.4/x64/src/runtime/panic.go:589 +0x5d
  testing.(*T).Run.gowrap1()
      /opt/hostedtoolcache/go/1.25.4/x64/src/testing/testing.go:1997 +0x44

Previous read at 0x00c0001bc20b by goroutine 159:
  testing.(*outputWriter).writeLine()
      /opt/hostedtoolcache/go/1.25.4/x64/src/testing/testing.go:1159 +0x70
  testing.(*outputWriter).Write()
      /opt/hostedtoolcache/go/1.25.4/x64/src/testing/testing.go:1149 +0x2ad
  testing.(*common).log()
      /opt/hostedtoolcache/go/1.25.4/x64/src/testing/testing.go:1040 +0x164
  testing.(*common).Logf()
      /opt/hostedtoolcache/go/1.25.4/x64/src/testing/testing.go:1191 +0x8e
  go.uber.org/zap/zaptest.TestingWriter.Write()
      /home/runner/go/pkg/mod/go.uber.org/zap@v1.27.0/zaptest/logger.go:146 +0x119
  go.uber.org/zap/zaptest.(*TestingWriter).Write()
      <autogenerated>:1 +0x74
  go.uber.org/zap/zapcore.(*ioCore).Write()
      /home/runner/go/pkg/mod/go.uber.org/zap@v1.27.0/zapcore/core.go:99 +0x18d
  go.uber.org/zap/zapcore.(*CheckedEntry).Write()
      /home/runner/go/pkg/mod/go.uber.org/zap@v1.27.0/zapcore/entry.go:253 +0x1ec
  go.uber.org/zap.(*SugaredLogger).log()
      /home/runner/go/pkg/mod/go.uber.org/zap@v1.27.0/sugar.go:355 +0x12c
  go.uber.org/zap.(*SugaredLogger).Errorf()
[...]

Issue #17309, opened 2025-06-19
test: TestAPMServerInstrumentation failing


carsonip:
**APM Server version** (`apm-server version`): main

**Description of the problem including expected versus actual behavior**:

```
 === RUN   TestAPMServerInstrumentation
    instrumentation_test.go:92: 
        	Error Trace:	/home/runner/work/apm-server/apm-server/systemtest/instrumentation_test.go:92
        	Error:      	Not equal: 
        	            	expected: "102633cf232dd5dfc20f962b726aeb21"
        	            	actual  : "11ae2dc65a581af9577f527e27ea6c87"
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1 +1 @@
        	            	-102633cf232dd5dfc20f962b726aeb21
        	            	+11ae2dc65a581af9577f527e27ea6c87
        	Test:       	TestAPMServerInstrumentation
        	Messages:   	expecting log with trace id 102633cf232dd5dfc20f962b726aeb21; got trace id 11ae2dc65a581af9577f527e27ea6c87 and message "request ok" instead
    instrumentation_test.go:94: 
        	Error Trace:	/home/runner/work/apm-server/apm-server/systemtest/instrumentation_test.go:94
        	Error:      	Not equal: 
        	            	expected: "102633cf232dd5df"
        	            	actual  : "11ae2dc65a581af9"
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1 +1 @@
        	            	-102633cf232dd5df
        	            	+11ae2dc65a581af9
        	Test:       	TestAPMServerInstrumentation
        	Messages:   	expecting log with transaction id 102633cf232dd5df; got transaction id 11ae2dc65a581af9 and message "request ok" instead
    server.go:145: log file: /home/runner/work/apm-server/apm-server/systemtest/logs/TestAPMServerInstrumentation/apm-server
--- FAIL: TestAPMServerInstrumentation (1.54s)
```

**Steps to reproduce**:

Please include a *minimal* but *complete* recreation of the problem,
including server configuration, agent(s) used, etc.  The easier you make it
for us to reproduce it, the more likely that somebody will take the time to
look at it.

 1.
 2.
 3.

**Provide logs (if relevant)**:

inge4pres:
Closed with https://github.com/elastic/apm-server/pull/20290.
Re-open if flakyness persists

PR #20290, opened 2026-02-03
Fix flaky systemtests

inge4pres:
## Motivation/summary

This Pr attempts to fixseveral flaky tests in `make system-test`:

* #20189
* #20098
* #19715 
* #17309 

Details of the changes applied:

### #20189

The `WaitUntilServerInactive` function in `systemtest/benchtest/expvar/expvar.go` has a busy loop with no delay between polling attempts. The function continuously polled the server's `/debug/vars` endpoint to check if `ActiveEvents` reached 0.
Added a `ticker` to reduce the frequency.

### #20098 

Modified the channel, make it buffered with the same number of fake requests added.
This prevents blocking on send, allowing the httptest Server to close.

### #19715

Data race addressed by modifying the position of test to call `manager.Stop()` before test ends, adding a grace period.

### #17309 

Removed one assertion that could fail due to non-guaranteed ordering of log messages, changed to search and wrapped the relevant assertion in the if condition.
The behavior of the test doesn't change because if the trace ID is not found we still have a `Fatal` call at the end.

## Checklist

## How to test these changes

I was able to reproduce the errors for all issues but #17309, even with a high number of consecutive test calls.
After applying the (supposed) fixes to the flakyness, I tested individual tests with

```
go test -v -race -timeout=10m -count=30 -run "Test_xxx"
```

using a count of 30 to try and reproduce.

### DIFF
--- a/systemtest/benchtest/expvar/expvar.go
+++ b/systemtest/benchtest/expvar/expvar.go
@@ -148,6 +148,8 @@ func doExpvar(req *http.Request, out *expvar) (string, error) {
 // * Context is done.
 func WaitUntilServerInactive(ctx context.Context, server string) error {
 	result := expvar{LibbeatStats: LibbeatStats{ActiveEvents: 1}}
+	// Avoid a busy loop querying the expvar endpoint by using a ticker.
+	// 100ms is completely arbitrary.
 	ticker := time.NewTicker(100 * time.Millisecond)
 	defer ticker.Stop()
 
--- a/systemtest/benchtest/main.go
+++ b/systemtest/benchtest/main.go
@@ -43,7 +43,7 @@ import (
 	"github.com/elastic/apm-server/systemtest/benchtest/expvar"
 )
 
-const waitInactiveTimeout = 90 * time.Second
+const waitInactiveTimeout = 30 * time.Second
 
 // BenchmarkFunc is the benchmark function type accepted by Run.
 type BenchmarkFunc func(*testing.B, *rate.Limiter)


internal/beatcmd/beat_test.go
```diff
@@ -627,14 +627,23 @@ func TestRunManager_Reloader_newRunnerError(t *testing.T) {
...
+
+	// Stop manager before test ends to prevent data race.
+	// The manager starts background goroutines that log via t.Log().
+	// If we use defer, the test may finish before Stop() completes,
+	// causing the logger to access test state after the test ends.
+	manager.Stop()
+
+	// Give goroutines time to fully exit after Stop().
+	// Stop() cancels contexts but goroutines may still be logging.
+	time.Sleep(100 * time.Millisecond)
```
kruskall, line 646: > Stop() cancels contexts but goroutines may still be logging.

this is a bug, time.sleep is not a proper solution :(

inge4pres: The PR is aimed at removing tests flakyness.
If there's a bug in the logic we can tackle it separately, addressing it with a proper card.

inge4pres:
mhmhm the flakyness persisted in the first test fixed

https://github.com/elastic/apm-server/actions/runs/21644858759/job/62394131276?pr=20290#step:6:338

back to the IDE

inge4pres:
> https://github.com/elastic/apm-server/actions/runs/21644858759/job/62394131276?pr=20290#step:6:338

@kruskall do you remember if there's a a difference in execution of the test `Test_warmup/4_agent_2s` in FIPS mode?
I can't find any but maybe I'm looking in the wrong place

internal/beatcmd/beat_test.go
inge4pres: btw now that I look at this again, I don't think we can control the behavior of the logger, we use zap which AFAIR has its own "eventing" loop for dispatching logs

inge4pres:
I re-triggered the run in CI for systemtests, to see if a another run is successful

inge4pres:
Pushed a commit to improve concurrency and error handling in `warmup`.
Will trigger CI 3/4 times on system-tests to try and replicate the error.
Note that I am unable to reproduce locally the error reported in CI.

systemtest/benchtest/main.go
```diff
@@ -291,19 +291,24 @@ func warmup(agents int, duration time.Duration, url, token, apiKey string) error
...
+	g, ctx := errgroup.WithContext(ctx)
+
 	for i := 0; i < agents; i++ {
-		go func() {
-			defer wg.Done()
-			sendErr := h.SendBatchesInLoop(ctx)
-			if sendErr != nil && !errors.Is(sendErr, context.DeadlineExceeded) {
-				log.Printf("failed to send batches: %v", sendErr)
+		g.Go(func() error {
+			if sendErr := h.SendBatchesInLoop(ctx); sendErr != nil {
```
aelnahas, line 298: Is it okay to swallow errors that are not deadline exceeded errors? wondering if it would be helpful to log them just incase they happen

inge4pres: do you mean that we don't propagate sendErr if it's something else than `context.DeadlineExceed`?
what we want here is to return the first error that is not deadline exceeded, cause that means we failed to ingest.

aelnahas: gah I missed the negation sign 😓

systemtest/benchtest/expvar/expvar.go
```diff
@@ -147,11 +148,14 @@ func doExpvar(req *http.Request, out *expvar) (string, error) {
 // * Context is done.
 func WaitUntilServerInactive(ctx context.Context, server string) error {
 	result := expvar{LibbeatStats: LibbeatStats{ActiveEvents: 1}}
+	ticker := time.NewTicker(100 * time.Millisecond)
```
isaacaflores2, line 153: suggestion: could we add the reasoning behind this ticker here? Similar to the other helpful comments you added for the other changes

inge4pres: sure I'll add it 👍🏼

inge4pres: Just updated

inge4pres:
Would be a tragedy if the tests on last commit would block merging 😅

PR #20258, opened 2026-02-03
package: use ubi10

v1v:
## Motivation/summary

Use ubi10 for Ironbank images.

### Questions

@elastic/obs-ds-intake-services, I need your help here in terms of what set of actions you can run to validate changes are in a good shape. OTOH, can you help with the changelog?

## Checklist

- [ ] Update [CHANGELOG.asciidoc](https://github.com/elastic/apm-server/blob/main/CHANGELOG.asciidoc)
- [ ] Documentation has been updated

For functional changes, consider:
- Is it observable through the addition of either **logging** or **metrics**?
- Is its use being published in **telemetry** to enable product improvement?
- Have system tests been added to avoid regression?

PR #20297, opened 2026-02-05
packaging(8.19): ubi10-micro does not have sed

v1v:
## Motivation/summary

Fixes errors in `8.19` when bumping the version to ubi10.

Unfortunately, it failed in the DRA rather than in the CI in the original PR (#20295), and those checks are not mandatory! :/

https://buildkite.com/elastic/apm-server-package/builds/8944

Further details, see https://buildkite.com/elastic/apm-server-package/builds/8946#_

## Checklist

### DIFF
--- a/packaging/docker/Dockerfile
+++ b/packaging/docker/Dockerfile
@@ -94,12 +94,14 @@ RUN echo 'apm-server:*:1000:1000::/usr/share/apm-server:/bin/false' >> /etc/pass
 RUN chown -R 1000:1000 /usr/share/apm-server
 
 # remove unnecessary packages in ubuntu-based images
-# use rpm to ignore ubi images
-RUN rpm -v || (dpkg -P --force-depends libgssapi-krb5-2 libtirpc3 libnsl2 libkrb5-3 libk5crypto3 libkrb5support0 && \
-    rm /var/log/dpkg.log)
+# only if rpm is not available and dpkg exists
+RUN if ! rpm -v && dpkg --version; then \
+      dpkg -P --force-depends libgssapi-krb5-2 libtirpc3 libnsl2 libkrb5-3 libk5crypto3 libkrb5support0 && \
+      rm /var/log/dpkg.log; \
+    fi
 # remove unnecessary packages in ubi-based images
-# use dpkg to ignore ubuntu images
-RUN dpkg --version || (rpm -v -e --nodeps libxml2)
+# only if dpkg is not available and rpm exists
+RUN if ! dpkg --version && rpm -v; then rpm -v -e --nodeps libxml2; fi
 
 USER apm-server
 EXPOSE 8200


packaging/docker/Dockerfile
```diff
@@ -82,21 +87,21 @@ COPY --from=builder-certs /etc/ssl/certs/ca-certificates.crt /etc/pki/tls/certs/
...
 RUN mkdir --mode=0777 data
 RUN echo 'apm-server:*:1000:1000::/usr/share/apm-server:/bin/false' >> /etc/passwd
 RUN chown -R 1000:1000 /usr/share/apm-server
 
 # remove unnecessary packages in ubuntu-based images
-# use rpm to ignore ubi images
-RUN rpm -v || (dpkg -P --force-depends libgssapi-krb5-2 libtirpc3 libnsl2 libkrb5-3 libk5crypto3 libkrb5support0 && \
-    rm /var/log/dpkg.log)
+# only if rpm is not available and dpkg exists
+RUN if ! rpm -v && dpkg --version; then \
```
fr4nc1sc0-r4m0n, line 98: missing brackets in the if statement?

v1v: There is no concept of brackets in if conditions for bash, or what do you mean?

v1v:
@carsonip , as you helped with the review for https://github.com/elastic/apm-server/pull/20258, please let me know if these changes make sense

packaging/docker/Dockerfile
carsonip, line 98: q: so there's no rpm and dpkg in ubi10?

carsonip (approved):
thanks, are we going to fix in main as well?

kruskall:
> thanks, are we going to fix in main as well?

main should be fine, we ended up fixing this when we switched to ubi-micro: https://github.com/elastic/apm-server/commit/ab2a066101228cfd055fadd29c654d69a8bff061

It was never an issue in 8.19 because we were using ubi-minimal (https://github.com/elastic/apm-server/commit/f140913c42eec16db2fe26c26534a7a0ec1350e7) . I wonder if we should keep using ubi-minimal in 8.19 :thinking:

v1v:
> I wonder if we should keep using ubi-minimal in 8.19 🤔

good point, happy to change it to minimal instead of `micro` if that's the best approach here

packaging/docker/Dockerfile
v1v: It's a micro without any packaging system.

fr4nc1sc0-r4m0n: I meant square brackets. 
Maybe here is not required because we are not testing the output of a command but the command itself.

v1v: There is no need here:

```bash
$ if rpm -v && brew --version; then echo found; fi
command not found: rpm

$ if ! rpm -v && brew --version; then echo found; fi
command not found: rpm
Homebrew 5.0.13
found

$ if [ ! rpm -v && brew --version ] ; then echo found; fi
[: ']' expected
```

carsonip: sigh I mixed up minimal and micro. Thanks for pointing that out

v1v:
I've just raised https://github.com/elastic/apm-server/pull/20299 to use `minimal` rather than `micro` - let's see how it goes and if it works, then let's decide what to do here, either this PR or the other one.

PR #20299, opened 2026-02-05
packaging(ironbank): use minimal

v1v:
## Motivation/summary

Use `minimal` rather than micro as `8.19` was always `ubi-minimal`, but https://github.com/elastic/apm-server/pull/20295 was backported and hence changed to `micro` as it's the flavour in `main`

## Checklist


v1v:
Unrelated failures... it used my forked branch and it failed in the DRA but we are interested in the packaging only for now

mallendem (approved):

carsonip:
let's go ahead with https://github.com/elastic/apm-server/pull/20297 instead

PR #20321, opened 2026-02-09
bk: skip slack alerts if retried tasks succeeded

v1v:
## Motivation/summary

Currently for pipelines that have steps with auto retries, if there is a failure of a step, even if the subsequent retry succeeds, we receive Slack alerts.
This commits uses new functionality[^1] available in the Buildkite PR Bot to avoid Slack notifications in those cases and changes the settings for all pipelines where we have, or expect to have, retries.

### Why is it important/What is the impact to the user?

Reduce alert noise for non-actionable events in the Slack channel.

## Checklist

## Related issues

See https://github.com/elastic/logstash/pull/15869

### DIFF
--- a/catalog-info.yaml
+++ b/catalog-info.yaml
@@ -55,3 +55,4 @@ spec:
           access_level: READ_ONLY
       env:
         ELASTIC_PR_COMMENTS_ENABLED: 'true'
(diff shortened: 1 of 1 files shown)


carsonip (approved):
rubberstamp

PR #20484, opened 2026-02-24
Fix request context and increase duration for 4 agents test

lahsivjar:
- Use the updated request with the context to do expvar query
- Increase duration for 4 agents - this attempts to fix the flakiness in https://github.com/elastic/apm-server/issues/20189 as the logs suggest that the server didn't go to "inactive" state. The hypothesis here is that FIPS takes more CPU and can cause CPU starvation.

Fixes: https://github.com/elastic/apm-server/issues/20189

### DIFF
--- a/systemtest/benchtest/expvar/expvar.go
+++ b/systemtest/benchtest/expvar/expvar.go
@@ -20,7 +20,6 @@ package expvar
 import (
 	"context"
 	"encoding/json"
-	"log"
 	"net/http"
 	"runtime"
 	"time"


systemtest/benchtest/expvar/expvar.go
```diff
@@ -87,7 +87,7 @@ func queryExpvar(ctx context.Context, out *expvar, srv string) error {
 	if err != nil {
 		return err
 	}
-	req.WithContext(ctx)
+	req = req.WithContext(ctx)
```
carsonip, line 90: 🤦

systemtest/benchtest/expvar/expvar.go
```diff
@@ -154,6 +155,7 @@ func WaitUntilServerInactive(ctx context.Context, server string) error {
 	defer ticker.Stop()
 
 	for result.ActiveEvents > 0 {
+		log.Println("ActiveEvents", result.ActiveEvents)
```
lahsivjar:
```suggestion
```

systemtest/benchtest/expvar/expvar.go
```diff
@@ -20,6 +20,7 @@ package expvar
 import (
 	"context"
 	"encoding/json"
+	"log"
```
lahsivjar:
```suggestion
```

carsonip (approved):

PR #20529, opened 2026-03-02
Fix error handling in runScrollSearchQuery

newtmltch:
## Checklist

when a function returns `(T, error)`, and `T` can be nil on error, all uses of `T` must occur only after confirming that `err == nil`. Here, `http.NewRequestWithContext` returns `(*http.Request, error)` and may return a nil request on error. The fix is to check `err` immediately after the call, before any dereference of `req`, and return appropriately if `err` is non-nil.

Concretely, in `runScrollSearchQuery` in `internal/sourcemap/metadata_fetcher.go`, we should move the error check to immediately follow the call to `http.NewRequestWithContext` and before the usage of `req`.  right after the call, and removing the later `if err != nil` block (or leaving the later block empty is not acceptable). This keeps existing functionality intact: on failure to construct the request, the function returns the error as before; on success, it sets the query parameters and performs the request.

### DIFF
--- a/internal/sourcemap/metadata_fetcher.go
+++ b/internal/sourcemap/metadata_fetcher.go
@@ -385,12 +385,12 @@ func (s *MetadataESFetcher) scrollsearch(ctx context.Context, scrollID string, u
 
 func (s *MetadataESFetcher) runScrollSearchQuery(ctx context.Context, id string) (*http.Response, error) {
 	req, err := http.NewRequestWithContext(ctx, http.MethodPost, "/_search/scroll", nil)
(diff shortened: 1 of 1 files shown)


endorama:
Thanks for the patch @newtmltch, the change looks correct to me. I'll run the workflows, could you please address the Contributor Agreement requirement?

carsonip:
@newtmltch did you have the chance to sign the CLA? We may have to close the PR if that remains unsigned

carsonip:
Closing as this PR is inactive

PR #20607, opened 2026-03-10
Fix notice generation CRLF

ericywl:
`make notice` sometimes generates notices with CRLF e.g. https://github.com/elastic/apm-server/actions/runs/22902103895/job/66450978532?pr=20577. This PR adds `tr` command to remove the added CRLF.


carsonip:
> Why ? It feels like this is a workaround trying to address an underlying issue. What's the root cause ?

CI is failing due to minlz license. Do you have a better workaround in mind?

Issue #20533, opened 2026-03-02
Add integrations tests for timestamp.us upgrade


endorama:
Once the fix for https://github.com/elastic/apm-server/issues/20496 lands, we should ensure to cover it with integration tests to avoid regressions.

endorama:
Partially addressed by https://github.com/elastic/apm-server/pull/20560 (ensure we expect rollover in 9.4.0)

Issue #20496, opened 2026-02-26
Broken waterfall view due to wrong mapping of timestamp.us field


R0ky:
**APM Server version** (`apm-server version`): 8.15.2
We confirmed this bug affected all outputs except Elasticsearch, from 8.15.0 up to and including 8.19.12, 9.2.6 and 9.3.1.

**Description of the problem including expected versus actual behavior**:
After bundling APM related templates in Elasticsearch (confirmed Elasticsearch version 8.19.8, but others are likely also impacted), new templates do not contain an explicit mapping for `timestamp.us`, while previous versions did. If APM-server ships data directly to Elasticsearch, dynamic mapping works as expected and `timestamp.us` is mapped as `long`: 

```
"timestamp": {
    "properties": {
        "us": {
            "type": "long"
        }
    }
}
```

If we introduce Logstash in between (APM Server -> Logstash -> Elasticsearch), then the field can get mapped as float which breaks the waterfall view: 
```
   "timestamp": {
          "properties": {
            "us": {
              "type": "float"
            }
          }
        }
```

This is due to how the transformation between `APMEvent` to `libbeat.Event` works. This code path is used only for some outputs that travel through `libbeat`. The result is sometimes an output like this that is mapped by Elasticsearch `dymanic` mapping to `float`:

```
 "timestamp": {
        "us": 1.772028649517038E15
```

**Steps to reproduce**:

 1. Deploy APM server 8.15.2
 2. Deploy Logstash 8.19.8
 3. Deploy Elasticsearch and Kibana 8.19.8
 4. Configure APM server to ship data to Logstash
 5. Configure Logstash to forward data to Elasticsearch without any filters
 6. Instrument an application (java in this example) 
 7. Once APM data is indexed in Elasticsearch verify the mapping `traces-apm*` indices and notice wrong mapping of `timestamp.us` field
 8. Review broken waterfall in traces UI

lucabelluccini:
Workaround until this gets fixed is to create a `traces-apm@custom` component template with the following mapping:
```
  "mappings": {
      "properties": {
        "timestamp": {
          "properties": {
            "us": {
              "type": "long"
            }
          }
        }
      }
    }
```

endorama:
I updated the description to update our understanding of the culprit for the issue. It does not lie on Logstash but on APM Server.

All APM Server outputs that are not Elasticsearch travel through `libbeat`, using a JSON marshalling implementation provided by the `beater` framework.

This is the problematic function:
[...]

PR #20696, opened 2026-03-12
add timestamp.us system and integration tests

endorama:
## Motivation/summary

Provides automated regression tests for https://github.com/elastic/apm-server/issues/20496

Closes https://github.com/elastic/apm-server/issues/20533

## How to test these changes

Run system and integration tests.

## Related issues

Related https://github.com/elastic/apm-server/issues/20496

### DIFF
--- a/integrationservertest/timestamp_us_test.go
+++ b/integrationservertest/timestamp_us_test.go
@@ -1,3 +1,20 @@
+// Licensed to Elasticsearch B.V. under one or more contributor
+// license agreements. See the NOTICE file distributed with
+// this work for additional information regarding copyright
+// ownership. Elasticsearch B.V. licenses this file to you under
+// the Apache License, Version 2.0 (the "License"); you may
+// not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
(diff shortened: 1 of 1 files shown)


endorama:
Submitted as draft because trying the upgrade path (in prod) fails with a 503 after upgrade. I'm investigating.

endorama:
**For reviewers**: This PR adds 2 tests:
1. a system test that verifies mapping on current version
2. an integration test that verifies the bug does not appear when upgrading from known affected version to a version with the fix.

This second test is not run a part of the usual Upgrade tests we run in CI and I don't think it should be added for 2 reasons:
- some tested paths would not exhibit the bug
- testing this occurrence across the upgrade paths we have would need additional logic to select which tests to run and which version introduces the fix in the upgrade chain

I don't believe it's worth the effort and the additional complexity. Introducing a regression is already guarded by the system test.

integrationservertest/upgrade_test.go
```diff
@@ -188,3 +191,141 @@ func dataStreamsExpectations(expect asserts.DataStreamExpectation) map[string]as
...
 	}
 }
+
+// TestTimestampUSMappingUpgrade is a regression test for
+// https://github.com/elastic/apm-server/issues/20496.
+// It verifies that upgrading from a pre-fix to a post-fix version
+// produces backing indices with timestamp.us mapped as long.
+// NOTE: This test is not run as part of usual integration tests upgrade
+// runs.
+func TestTimestampUSMappingUpgrade(t *testing.T) {
```
ericywl: Might be worth moving to a separate file so that it's clear that it is separate from the usual upgrade tests.

endorama: Updated. That also dismissed your review unfortunately.

PR #20773, opened 2026-03-25
feat: move docker labels to opencontainers

kruskall:
## Motivation/summary

drop deprecated org.label-schema labels and use
opencontainers labels everywhere.

Overwrite ubi base image non-standard labels to avoid confusion.

label-schema is deprecated and images should move to https://specs.opencontainers.org/image-spec/annotations/

This should help with retrieving base image and maintainers from our images.

## Checklist

### DIFF
--- a/packaging/docker/Dockerfile
+++ b/packaging/docker/Dockerfile
@@ -33,33 +33,36 @@ ARG VCS_REF
 
 # Statically defined labels.
 LABEL \
-  org.label-schema.schema-version="1.0" \
-  org.label-schema.vendor="Elastic" \
-  org.label-schema.license="Elastic License" \
-  org.label-schema.name="apm-server" \
-  org.label-schema.url="https://www.elastic.co/apm" \
-  org.label-schema.vcs-url="github.com/elastic/apm-server" \
-  io.k8s.description="Elastic APM Server" \
-  io.k8s.display-name="Apm-Server image" \
-  org.opencontainers.image.licenses="Elastic License" \
+  org.opencontainers.image.authors="obs-ds-intake-services" \
+  org.opencontainers.image.base.name="registry.access.redhat.com/ubi10-micro:latest" \
+  org.opencontainers.image.description="Elastic APM Server" \
+  org.opencontainers.image.documentation="https://www.elastic.co/docs/solutions/observability/apm/apm-server" \
+  org.opencontainers.image.licenses="Elastic-2.0" \
+  org.opencontainers.image.source="https://github.com/elastic/apm-server" \
   org.opencontainers.image.title="Apm-Server" \
+  org.opencontainers.image.url="https://github.com/elastic/apm-server" \
   org.opencontainers.image.vendor="Elastic" \
-  name="apm-server" \
-  maintainer="infra@elastic.co" \
-  vendor="Elastic" \
-  release="1" \
-  url="https://www.elastic.co/apm" \
-  summary="apm-server" \
-  license="Elastic License" \
-  description="Elastic APM Server"
+  io.k8s.description="Elastic APM Server" \
+  io.k8s.display-name="Apm-Server image"
 
 # Dynamic labels, only set in published images.
 LABEL \
-  org.label-schema.build-date=${BUILD_DATE} \
-  org.label-schema.version=${VERSION} \
-  org.label-schema.vcs-ref=${VCS_REF} \
   org.opencontainers.image.created=${BUILD_DATE} \
-  version=${VERSION}
+  org.opencontainers.image.revision=${VCS_REF} \
+  org.opencontainers.image.version=${VERSION}
+
+# Overwrite ubi base image non-standard labels
+LABEL \
+  build-date="" \
+  description="" \
+  maintainer="" \
+  name="" \
+  release="" \
+  summary="" \
+  url="" \
+  vcs-ref="" \
+  vendor="" \
+  version=""
 
 ENV ELASTIC_CONTAINER="true"
 
--- a/packaging/docker/Dockerfile.fips
+++ b/packaging/docker/Dockerfile.fips
@@ -30,33 +30,23 @@ ARG VCS_REF
 
 # Statically defined labels.
 LABEL \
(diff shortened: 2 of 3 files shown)


packaging/docker/Dockerfile
```diff
@@ -33,33 +33,36 @@ ARG VCS_REF
...
-  org.label-schema.schema-version="1.0" \
-  org.label-schema.vendor="Elastic" \
-  org.label-schema.license="Elastic License" \
-  org.label-schema.name="apm-server" \
-  org.label-schema.url="https://www.elastic.co/apm" \
-  org.label-schema.vcs-url="github.com/elastic/apm-server" \
-  io.k8s.description="Elastic APM Server" \
-  io.k8s.display-name="Apm-Server image" \
-  org.opencontainers.image.licenses="Elastic License" \
+  org.opencontainers.image.authors="obs-ds-intake-services" \
```
carsonip, line 36: q: Does it make sense to align this with vendor field to use `Elastic` instead of exposing org internals? What's the practice within Elastic, or in other well known repos?

packaging/docker/Dockerfile
```diff
@@ -33,33 +33,36 @@ ARG VCS_REF
...
+  org.opencontainers.image.base.name="registry.access.redhat.com/ubi10-micro:latest" \
```
carsonip, line 37: q: is `latest` an accurate description, or should we specify the actual version and have tooling to bump it together?

kruskall: `latest` is correct because that's the tag we're using. We're also pinning to a digest which should be added to `org.opencontainers.image.base.digest` but it was omitted it because it would be a pain to keep in sync

packaging/docker/Dockerfile
kruskall: `Elastic` would be the vendor, this is based on the fact that the previous maintainer label was "infra" and the team name is already mentioned in the repo/codeowner

EDIT: s/infra-mki/infra

carsonip: what about aligning with beats to use `infra@elastic.co`? https://github.com/elastic/beats/blob/dcda43e551107fb14fb35900d26557fe874710e3/dev-tools/packaging/templates/docker/Dockerfile.tmpl#L76 Seems like a more helpful one than obs-ds-intake-services

kruskall: that's the old address we were using, it's a leftover from a time in which the images were published by the infra team. That's not the case anymore, intake services is the maintainer of the image.

IMO this is no different than the catalog info file we have in the repo with owner:obs-ds-intake-services

carsonip: I think it is slightly different from catalog info file because this will show up in the published artifact, but it isn't a blocker. We can update it later when we see fit.

packaging/docker/Dockerfile
```diff
@@ -33,33 +33,36 @@ ARG VCS_REF
...
+  org.opencontainers.image.description="Elastic APM Server" \
+  org.opencontainers.image.documentation="https://www.elastic.co/docs/solutions/observability/apm/apm-server" \
+  org.opencontainers.image.licenses="Elastic-2.0" \
```
carsonip, line 40:
```suggestion
  org.opencontainers.image.licenses="Elastic-License-2.0" \
```

nit: for consistency with other elastic images e.g.

https://github.com/elastic/elasticsearch/blob/7e16024fe6ef1ababcfbda237b7c49ccc303ad6b/distribution/docker/build.gradle#L167

packaging/docker/Dockerfile.fips
```diff
@@ -22,41 +22,31 @@ RUN sed -i 's/localhost:9200/elasticsearch:9200/' apm-server.yml
...
-  org.label-schema.url="https://www.elastic.co/apm" \
-  org.label-schema.vcs-url="github.com/elastic/apm-server" \
-  io.k8s.description="Elastic APM Server" \
-  io.k8s.display-name="Apm-Server image" \
-  org.opencontainers.image.licenses="Elastic License" \
+  org.opencontainers.image.authors="obs-ds-intake-services" \
+  org.opencontainers.image.base.name="docker.elastic.co/wolfi/chainguard-base-fips:latest" \
+  org.opencontainers.image.description="Elastic APM Server" \
+  org.opencontainers.image.documentation="https://www.elastic.co/docs/solutions/observability/apm/apm-server" \
+  org.opencontainers.image.licenses="Elastic-2.0" \
```
carsonip, line 37:
```suggestion
  org.opencontainers.image.licenses="Elastic-License-2.0" \
```

packaging/docker/Dockerfile.wolfi
```diff
@@ -23,41 +23,31 @@ RUN sed -i 's/localhost:9200/elasticsearch:9200/' apm-server.yml
...
-  org.label-schema.url="https://www.elastic.co/apm" \
-  org.label-schema.vcs-url="github.com/elastic/apm-server" \
-  io.k8s.description="Elastic APM Server" \
-  io.k8s.display-name="Apm-Server image" \
-  org.opencontainers.image.licenses="Elastic License" \
+  org.opencontainers.image.authors="obs-ds-intake-services" \
+  org.opencontainers.image.base.name="cgr.dev/chainguard/static:latest" \
+  org.opencontainers.image.description="Elastic APM Server" \
+  org.opencontainers.image.documentation="https://www.elastic.co/docs/solutions/observability/apm/apm-server" \
+  org.opencontainers.image.licenses="Elastic-2.0" \
```
carsonip, line 38:
```suggestion
  org.opencontainers.image.licenses="Elastic-License-2.0" \
```

packaging/docker/Dockerfile
kruskall: I'm afraid the other elastic images are wrong :sweat_smile: 

This is a SPDX identifier, the value must be picked from the list: https://spdx.org/licenses/

carsonip: oof, good one. Probably on you to fix all the other repos then :+1:

PR #20257, opened 2026-02-03
doc: improve dev_docs release instructions

inge4pres:
## Motivation/summary

The dev docs on the 9.x release process are wrong.

PR #20843, opened 2026-04-07
Fix patch release workflow version

### LINKED ISSUE #20731: Fix patch release action
## Summary

Current patch release action (https://github.com/elastic/apm-server/actions/workflows/run-patch-release.yml) creates wrong changelog PR and sends misleading Slack message to the `apm-server` channel.

Image

## Proposal

1. Fix the wrong changelog PR or remove it entirely. Currently, we are manually adding the changelog e.g. https://github.com/elastic/apm-server/pull/20723.
2. Fix Slack message patch release version, it is off-by-one, e.g. it should be `9.2.7` instead of `9.2.8` due to the way we provide the version.
3. Fix Slack message feature freeze date, it always says `Today` currently which is wrong.

ericywl:
## Motivation/summary

- Fix patch release workflow to take in to-be-released version, but use bumped version for update PRs.
- Remove changelog PRs from workflow, since we create them manually.
- Remove `update-version-makefile` since it is no longer used.

## How to test these changes

Run the workflow.

Example: https://github.com/elastic/apm-server/actions/runs/24068944954.

## Related issues

Closes https://github.com/elastic/apm-server/issues/20731.

### DIFF
--- a/.github/workflows/run-patch-release.yml
+++ b/.github/workflows/run-patch-release.yml
@@ -5,7 +5,7 @@ on:
   workflow_dispatch:
     inputs:
       version:
-        description: 'The version to be bumped to (semver format: major.minor.patch). e.g. when told "bump the project and documentation versions on the 9.2 branch to 9.2.6", fill in `9.2.6`.'
+        description: 'The upcoming feature freeze / release version (semver format: major.minor.patch)'
         required: true
         type: string
 
@@ -38,13 +38,27 @@ jobs:
           version: ${{ inputs.version }}
           type: 'patch'
 
-  run-patch:
+  bump-patch:
     runs-on: ubuntu-latest
     needs: [ prepare ]
+    steps:
+      - id: bump
+        shell: bash
+        run: |
+          VERSION="${{ needs.prepare.outputs.release-version }}"
+          MAJOR=$(echo "$VERSION" | cut -d. -f1)
+          MINOR=$(echo "$VERSION" | cut -d. -f2)
+          PATCH=$(echo "$VERSION" | cut -d. -f3)
+          NEW_VERSION="${MAJOR}.${MINOR}.$((PATCH + 1))"
+          echo "bump-version=${NEW_VERSION}" >> "$GITHUB_OUTPUT"
+
+  run-patch:
+    runs-on: ubuntu-latest
+    needs: [ prepare, bump-patch ]
     env:
       RELEASE_BRANCH: ${{ needs.prepare.outputs.release-branch }}
       RELEASE_TYPE: ${{ needs.prepare.outputs.release-type }}
-      RELEASE_VERSION: ${{ needs.prepare.outputs.release-version }}
+      BUMP_VERSION: ${{ needs.bump-patch.outputs.bump-version }}
     steps:
       - name: Get token
         id: get_token
@@ -89,7 +103,7 @@ jobs:
           bot-token: ${{ secrets.SLACK_BOT_TOKEN }}
           channel-id: ${{ env.SLACK_CHANNEL }}
           message: |-
-            PRs to bump versions in `${{ github.repository }}@${{ env.RELEASE_BRANCH }}` branch to `${{ env.RELEASE_VERSION }}` have been created.
+            PRs to bump versions in `${{ github.repository }}@${{ env.RELEASE_BRANCH }}` branch to `${{ env.BUMP_VERSION }}` have been created.
           thread-timestamp: ${{ needs.prepare.outputs.slack-thread || '' }}
 
       - if: failure()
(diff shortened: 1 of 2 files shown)


.github/workflows/run-patch-release.yml
```diff
@@ -5,7 +5,7 @@ on:
   workflow_dispatch:
     inputs:
       version:
-        description: 'The version to be bumped to (semver format: major.minor.patch). e.g. when told "bump the project and documentation versions on the 9.2 branch to 9.2.6", fill in `9.2.6`.'
+        description: 'The upcoming feature freeze / release version (semver format: major.minor.patch)'
```
carsonip: nit: "Upcoming FF / release version" may be ambiguous, when the team receives a slack message `We are releasing 9.2.6 today. If the following teams would please bump the project and documentation versions on the 9.2 branch to 9.2.7`. In that case the release version for today is 9.2.6, but which one is "upcoming"? Could we make the workflow version description more foolproof?

release.mk
```diff
@@ -80,7 +78,6 @@ minor-release:
 	@echo "INFO: Create release branch and update new version $(RELEASE_VERSION)"
 	$(MAKE) create-branch NAME=$(RELEASE_BRANCH) BASE=$(BASE_BRANCH)
 	$(MAKE) update-version VERSION=$(RELEASE_VERSION)
-	$(MAKE) update-version-makefile VERSION=$(PROJECT_MAJOR_VERSION)\.$(PROJECT_MINOR_VERSION)
```
carsonip, line 83: q: why is this removed for minor release?

.github/workflows/run-patch-release.yml
```diff
@@ -60,7 +76,7 @@ jobs:
           # 0 indicates all history for all branches and tags.
           fetch-depth: 0
           # Use the makefile in the given release branch.
-          ref: ${{ env.RELEASE_BRANCH }}
+          # ref: ${{ env.RELEASE_BRANCH }}
```
carsonip: TIL we're using makefile in release branch. I agree that it'll be easier to reason about when the workflow definition is entirely on main. Could you remove the comment or update the stale comment?

.github/workflows/run-patch-release.yml
```diff
@@ -38,13 +38,29 @@ jobs:
           version: ${{ inputs.version }}
           type: 'patch'
 
-  run-patch:
+  bump-patch:
```
carsonip: q: how slow is an extra job, compared to a step?

carsonip (commented):
Thanks for looking into it.
- Do you mind incorporating any required changes in doc including anything relevant in https://github.com/elastic/apm-server/pull/20257 ?
- If we're only running this workflow from main there's no need to backport. Also for 8.19 patch should we run the workflow from 8.19 branch or main?

.github/workflows/run-patch-release.yml
ericywl: Got it, it should be `9.2.6` in your example.

.github/workflows/run-patch-release.yml
ericywl: I should just convert it into a step, I originally thought it would be more involved.

.github/workflows/run-patch-release.yml
ericywl: I commented to test, since otherwise it wouldn't work at all. But I believe all of our active branches have same / similar Makefile so I agree with removing this ref.

release.mk
ericywl: My mistake, restored it here: 91328cc543b1aa32fe7a7d234187f3742c1d06dc.

ericywl:
> Do you mind incorporating any required changes in doc including anything relevant in https://github.com/elastic/apm-server/pull/20257?

I will update the docs as well 👍

release.mk
carsonip: nit: stale comment

carsonip (dismissed):

carsonip (approved):

ericywl:
Ran workflow again: https://github.com/elastic/apm-server/actions/runs/24130770168. It works as expected.

v1v:
I think this PR introduced a regression, https://buildkite.com/elastic/apm-server-version-bump/builds/9 triggered https://github.com/elastic/apm-server/actions/runs/25365191087 with the correct version

```
executing: gh workflow run run-patch-release.yml --ref main --field version=9.4.1
```


but it used `9.4.2`

```
Run make patch-release
  make patch-release
  shell: /usr/bin/bash -e {0}
  env:
    JOB_URL: https://github.com/elastic/apm-server/actions/runs/25365191087
    SLACK_CHANNEL: #apm-server
    RELEASE_BRANCH: 9.4
    RELEASE_TYPE: patch
    GIT_USER: obltmachine
    GIT_EMAIL: obltmachine@users.noreply.github.com
    GH_TOKEN: ***
    BUMP_VERSION: 9.4.2
```

therefore https://github.com/elastic/apm-server/pull/21015 is not correct

PR #20881, opened 2026-04-13
Implement paging for TBS

### LINKED ISSUE #20903: TBS: Bound memory usage when reading out sampled trace events
Trace events are read from local db (pebble in 9.x, badger in 8.x) when a sampling decision is received, local or remote. Currently, the number of trace events that is read into memory is unbounded because of the lack of batching. In some edge cases where a trace contains many events, reading them all at once may cause a spike in memory or OOM.

lahsivjar:
## Motivation/summary
TBS doesn't implement paging for reading trace events causing memory spike proportional to the trace size. This PR attempts a paging based read to avoid those spikes and load large traces without OOMs.

## Checklist

## Related issues

Fixes #20903


x-pack/apm-server/sampling/processor.go
```diff
@@ -516,14 +509,17 @@ func (p *Processor) Run() error {
...
 				}
-				p.eventMetrics.sampled.Add(gracefulContext, int64(len(events)))
-				if err := p.config.BatchProcessor.ProcessBatch(gracefulContext, &events); err != nil {
-					p.logger.With(logp.Error(err)).Warn("failed to report events")
-				}
-
-				for i := range events {
-					events[i] = nil // not required but ensure that there is no ref to the freed event
-				}
+				totalEvents += int64(len(events))
```
carsonip: nit: i have a slight preference to increment the metric immediately here, in case ProcessBatch stalls due to backpressure. But debug log it at once at the end is good.

x-pack/apm-server/sampling/eventstorage/rw.go
```diff
@@ -20,6 +20,12 @@ var (
 // RW is a read writer interface that has methods to read and write trace event and sampling decisions.
 type RW interface {
 	ReadTraceEvents(traceID string, out *modelpb.Batch) error
```
carsonip: q: can we get rid of this entirely?

x-pack/apm-server/sampling/eventstorage/prefix.go
```diff
@@ -81,6 +81,57 @@ func (rw PrefixReadWriter) ReadTraceEvents(traceID string, out *modelpb.Batch) e
...
+	iter, err := rw.db.NewIter(&pebble.IterOptions{})
+	if err != nil {
+		return err
+	}
+	defer iter.Close()
+
+	if valid := iter.SeekPrefixGE(b.Bytes()); !valid {
+		return nil
+	}
+	var batch modelpb.Batch
```
carsonip: nit: this is the only perf regression I can see from this PR. In the past, the sampling decision handling goroutine would reuse 1 batch forever, but here it is only reused within the function call within batches created by soft memory limit. Not sure if reusing batch falls into the premature optimization category.

carsonip (commented):
thanks, first pass looks good. This should get us far enough

x-pack/apm-server/sampling/eventstorage/prefix.go
```diff
@@ -59,14 +68,11 @@ func (rw PrefixReadWriter) ReadTraceEvents(traceID string, out *modelpb.Batch) e
 	}
 	defer iter.Close()
 
-	// SeekPrefixGE uses prefix bloom filter for on disk tables.
```
carsonip, line 62: why did the LLM not like my comment 😢

x-pack/apm-server/sampling/eventstorage/rw.go
```diff
@@ -19,7 +19,17 @@ var (
 
 // RW is a read writer interface that has methods to read and write trace event and sampling decisions.
 type RW interface {
-	ReadTraceEvents(traceID string, out *modelpb.Batch) error
+	// ReadTraceEventsCallback reads trace events in batches, calling fn
```
carsonip, line 22: nit: dup comment in implementation

carsonip (dismissed):
code change lgtm, 2 nit on comment

x-pack/apm-server/sampling/eventstorage/prefix.go
lahsivjar: Ah, this was me 😓 . I removed it thinking SeekPrefixGE has better information on what it did but now that I think about it, having this is good to get a summary of what is happening. Adding it back

carsonip (approved):

Issue #20454, opened 2026-02-20
ci: support minor versions in create-test-plan-patch workflow


carsonip:
create-test-plan-patch only handles patch versions, i.e. versions without suffix .0, hence the workflow's name. Expand this workflow to support minor version.

This may involve
- figuring out the common ancestor between x.y and x.(y-1) for x.y.0 release.

When the feature is ready
- update workflow name or use a separate workflow / scripts 
- update docs

carsonip:
@aelnahas has the patch ready, it just needs a bit more testing.

aelnahas:
Made changes to the script to generate these test plans. I been testing it locally and it seems like it is producing the right diffs, but would love for other folks to confirm my changes.

Issue #19947, opened 2025-12-19
9.3.0 Test Plan


marclop:
# Manual Test Plan

- When picking up a test case, please add your name to this overview beforehand and tick the checkbox when finished.
- Testing can be started when the first build candidate (BC) is available in the CFT region.
- For each repository, update the compare version range to get the list of commits to review.

## apm-aggregation

List of changes: https://github.com/elastic/apm-aggregation/compare/v1.2.0...v1.3.0

## go-docappender

List of changes: https://github.com/elastic/go-docappender/compare/v2.4.0...v2.10.0

## apm-data

List of changes: https://github.com/elastic/apm-data/compare/v1.16.0...v1.19.2

## apm-server

List of changes: https://github.com/elastic/apm-server/compare/v9.2.2...9.3

## Test cases from the GitHub board

Add yourself as _assignee_ on the PR before you start testing.

marclop:
I'll add the script I used to pull the commits below, but essentially these are the "potential changes" we may want to test:

═══════════════════════════════════════════════════════════════
  Commits by Author
═══════════════════════════════════════════════════════════════

┌─ @kruskall (5 commits)
│
│  f025f61a1 fix: add missing span kind to otel traces (#19732)
~│  febabd530 feat: use pebblegozstd tags everywhere (#19814)~ -> Test only
~│  44b4e028d lint: remove unused go tools entries (#19672)~ -> Test only
~│  350e400d9 feat: remove references to es output and unused callbacks (#19554)~ -> Cleans up old callbacks.
│  dece0323e fix: pass noop otel providers to pubsub docappender (#18891)
└────────────────────────────────────────────


So for the changes above, we seem to have fixed the telemetry when auto-instrumentation is enabled. Is that right @kruskall?

marclop:
Joseph confirmed. Nothing to test.

raultorrecilla:
@marclop can we close this one right?

Issue #18894, opened 2025-10-02
9.2.0 Test Plan


inge4pres:
# Manual Test Plan

- When picking up a test case, please add your name to this overview beforehand and tick the checkbox when finished.
- Testing can be started when the first build candidate (BC) is available in the CFT region.
- For each repository, update the compare version range to get the list of commits to review.

## ES apm-data plugin

- [ ] https://github.com/elastic/elasticsearch/pull/131296
- [ ] https://github.com/elastic/elasticsearch/pull/131395

## apm-aggregation

List of changes: https://github.com/elastic/apm-aggregation/compare/v1.3.0...v1.4.0

## go-docappender

List of changes: https://github.com/elastic/go-docappender/compare/v2.11.0...v2.11.2

## apm-data

List of changes: https://github.com/elastic/apm-data/compare/v1.19.2...v1.19.5

## apm-server

List of changes: https://github.com/elastic/apm-server/compare/9.1...9.2

## Test cases from the GitHub board

Add yourself as _assignee_ on the PR before you start testing.

inge4pres:
Only change tested by @carsonip

PR #20969, opened 2026-04-22
support minor release create test plan workflow

aelnahas:
## Motivation/summary

This PR will allow us to automate the test plan creation for minor releases

Changes that were done:
- update the `generate-test-plan.sh` to now handle minor releases
- update the documentation
- renamed `create-test-plan-patch` to `create-test-plan` to avoid confusion now that it handles both minor and patch releases
- update documentation

## Checklist

## How to test these changes

```sh
# since the workflow name has been releases I invoked the script from my terminal e.g.

➜  via 🐹 v1.26.2 bash .ci/scripts/generate-test-plan.sh 9.4.0
Upcoming release: v9.4.0
Release branch: 9.4
Previous tag: v9.3.3
Commits analyzed: 165
Generated build/test-plan-9.4.0.md

# then checked the output by 
cat build/test-plan-9.4.0.md


# did this for the following versions:
# 9.4.0 <- minor 
# 9.3.4 <- patch
# 9.0.0 <- when minor version is zero it should error out
# 9.6.0 <- when no tags exist for 9.5.* then it will error out as well
```

for convenience here are the test plans that were created by invoking this script:
- 9.3.4: https://github.com/aelnahas/apm-server/issues/3
- 9.4.0: https://github.com/aelnahas/apm-server/issues/2
they werent created by the workflow rather I ran the script and pasted the output test plan as an issue to my fork.

## Related issues

https://github.com/elastic/apm-server/issues/20454

### DIFF
--- a/.ci/scripts/generate-test-plan.sh
+++ b/.ci/scripts/generate-test-plan.sh
@@ -17,13 +17,22 @@ MAJOR="${BASH_REMATCH[1]}"
 MINOR="${BASH_REMATCH[2]}"
 PATCH="${BASH_REMATCH[3]}"
 
-if (( PATCH == 0 )); then
-  echo "Error: Cannot infer previous tag for x.y.0 releases. Patch version must be >= 1"
-  exit 1
-fi
-
 BRANCH="${MAJOR}.${MINOR}"
-PREVIOUS_TAG="v${MAJOR}.${MINOR}.$((PATCH - 1))"
+
+if (( PATCH == 0 )); then
+  if (( MINOR == 0 )); then
+    echo "Error: Cannot infer previous tag for x.0.0 releases"
+    exit 1
+  fi
+  PREVIOUS_MINOR=$((MINOR - 1))
+  PREVIOUS_TAG=$(git tag -l "v${MAJOR}.${PREVIOUS_MINOR}.*" | sort -V | tail -1)
+  if [[ -z "${PREVIOUS_TAG}" ]]; then
+    echo "Error: Could not find any tags for v${MAJOR}.${PREVIOUS_MINOR}.*"
+    exit 1
+  fi
+else
+  PREVIOUS_TAG="v${MAJOR}.${MINOR}.$((PATCH - 1))"
+fi
 OUTPUT_FILE="build/test-plan-${VERSION}.md"
 REPO_CACHE_DIR="build/test-plan-repos"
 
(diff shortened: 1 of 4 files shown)


mallendem (approved):

carsonip (commented):
a way to test this would be to reuse the same workflow name, push it to a branch under elastic/apm-server, i.e. not your fork, then you can run it on GH actions. Only after testing you may rename it as you like. This is because GH requires a workflow to be on main for it to be shown on the UI.

In this case, since the workflow doesn't do much other than invoking the script and creating the issue, you may decide if the local validation is enough.

I'll do a review separately

aelnahas:
https://github.com/elastic/apm-server/pull/20969#pullrequestreview-4161791415 

Worth doing just to make sure the renaming doesnt cause issues

.ci/scripts/generate-test-plan.sh
```diff
@@ -17,13 +17,22 @@ MAJOR="${BASH_REMATCH[1]}"
...
+
 if (( PATCH == 0 )); then
-  echo "Error: Cannot infer previous tag for x.y.0 releases. Patch version must be >= 1"
-  exit 1
+  if (( MINOR == 0 )); then
+    echo "Error: Cannot infer previous tag for x.0.0 releases"
+    exit 1
+  fi
+  PREVIOUS_MINOR=$((MINOR - 1))
+  PREVIOUS_TAG=$(git tag -l "v${MAJOR}.${PREVIOUS_MINOR}.*" | sort -V | tail -1)
```
carsonip, line 28: I'm not sure if the base tag detection logic is 100% correct. And at the very least it should not depend on when the generate test plan script is run. It is not a simple question though, under the existing 2 active minors model - how do you decide if a change should show up in minor changelog, e.g. 9.4.0?
- Obviously, if a change is merged to 9.4 (either before or after FF) but not 9.3, it should be in 9.4.0 changelog
- But what if a bugfix is merged and due to be released in both 9.3.x and 9.4.0?

carsonip (commented):
q on base tag decision logic. Since you already spent some time on it you might have a convincing answer ready

.github/workflows/create-test-plan.yml
fr4nc1sc0-r4m0n, line 21:
```suggestion
        uses: actions/checkout@v7
```

fr4nc1sc0-r4m0n, line 28: To prevent script injection use an env var for the inputs

.ci/scripts/generate-test-plan.sh
aelnahas: Personally I thought the desired behaviour is to include all the changes since the latest applicable tag (e.g. v9.3.x for a 9.4 release), but there is a bit of inconsistency with past test plans.

Examples
9.4: https://github.com/elastic/apm-server/issues/20863 (all changes from 9.3.3)
9.3: https://github.com/elastic/apm-server/issues/19947 (all changes from 9.2.2)
9.2: https://github.com/elastic/apm-server/issues/18894 (all changes from 9.1)

I based this logic on two assumptions ( which might be incorrect):
- Do tags always represent a release?  ( I assumed yes but perhaps this isn't always true. I can tighten the the PREVIOUS_TAG search to be more strict? )
- Additionally, could we have a situation where a past commit/fix that was released in a patch fail in a future minor?  ( I assumed perhaps maybe and I thought it is safer to include than to exclude).

aelnahas:
Closing this PR. I will open a new one, once I get a good understanding of what a minor release should encompass.

Issue #20991, opened 2026-04-27
EA integration regression dropping telemetry from fleet managed / EA standalone apm-server


carsonip:
https://github.com/elastic/integrations/pull/12793 introduced a regression in `packages/elastic_agent/data_stream/apm_server_metrics/fields/beat-fields.yml` such that many apm-server metrics become unmapped. https://github.com/elastic/integrations/pull/16560 then fixed some, but a lot are still missing. Add them all back.

Suspected impact is that unmapped metrics from apm-server are dropped because TSDB is used and dynamic mapping is false.

Issue #15533, opened 2025-02-03
monitoring: expose storage limit and disk metrics for TBS monitoring


carsonip:
## Summary

From comment https://github.com/elastic/apm-server/issues/14247#issuecomment-2576116925

8.x and 9.x: As apm-server exposes lsm_size and value_log_size as monitoring metrics, expose configured storage limit as well. Then it will be possible to plot db size vs storage limit, and removes the need to dig into the logs for the configured storage limit.

9.x: Add monitoring metrics to monitor the disk utilization check.

It is up to the implementer to decide what metric to emit, whether it is a combined metric or a few separate metrics. The actual work involves adding this metric in apm-server code, and update relevant mappings in ES, integrations and metricbeat repo. See https://github.com/elastic/apm-server/issues/13475 for an example.

## Dev Notes

PRs in https://github.com/elastic/apm-server/issues/15533#issuecomment-3754389427.

raultorrecilla:
this is important as we would like to add some UI to tell the customers how much disk they are using https://github.com/elastic/kibana/issues/226600

carsonip:
This is less straightforward than before. In 8.x there is one configured storage limit, and we compare db size against that. However, from 9.0, by default the storage limit is 0, and instead of comparing db size against that, we compare the disk_used vs disk_total so that TBS doesn't write to the last 20% of the disk. It does not involve db size and storage limit. I'm a bit hesitant to plumb these implementation details to monitoring metrics and all the way to the UI. We should rethink what should be exposed.

carsonip:
`apm-server.sampling.tail.events.failed_writes` is a counter metric that records the number of failed writes, no matter it is then discarded or directly indexed to ES. If this per-apm-server counter increases, it means apm-server is running into storage issues. I believe it is good enough for https://github.com/elastic/kibana/issues/226600

carsonip:
That said, for https://github.com/elastic/kibana/issues/226600 , I wonder how kibana would have access to monitoring metrics (think about stack monitoring). If we cannot surface this by default on ECH, I wonder how useful this will be.

carsonip:
In case we move ahead with implementing this, the actual work involves adding this metric in apm-server code, and update relevant mappings in ES, integrations and metricbeat repo. See https://github.com/elastic/apm-server/issues/13475 for an example.

carsonip:
[...]

PR #20993, opened 2026-04-27
monitoring: eagerly register apm-server metric instruments

### LINKED ISSUE #20996: monitoring: Tail-based sampling `disk_usage_threshold_pct` stat silently dropped from /stats
**APM Server version** (`apm-server version`): 9.2.0+, 9.3.0+.

The tail-based sampling storage metric `apm-server.sampling.tail.storage.disk_usage_threshold_pct` is recorded by the storage manager as a `Float64Gauge`, but it never appears in the `/stats` HTTP endpoint output even when TBS is enabled and the gauge is being actively recorded.

The metric was introduced in #20464 (backports: #20477 to 9.2, #20478 to 9.3). It has been silently dropped since it was added — it has never been observable via `/stats`.

Root cause: in `internal/beatcmd/beat.go`, the OTel-to-libbeat translator that walks `apm-server.*` instruments (`addAPMServerMetricsToMap`) only handles `Sum[int64]` / `Gauge[int64]` aggregations. Any float64 instrument falls through and is silently dropped:

```go
// before
if v, ok := getScalarInt64(m.Data); ok {
    beatsMetrics[m.Name] = v
}
// no float64 branch — float-typed instruments under apm-server.* vanish
```

The int64-only translator was introduced as part of the OTel monitoring migration in #15094. `disk_usage_threshold_pct` is the only `apm-server.*` Float64 instrument today, so it's the only one missing from `/stats`. But the gap is general: any future float-typed `apm-server.*` metric (e.g. ratios, percentages, durations recorded as seconds) would hit the same trap.

Expected: when TBS is enabled, `apm-server.sampling.tail.storage.disk_usage_threshold_pct` should appear in `/stats` with the configured threshold value (e.g. `0.8`).

Actual: the field is absent regardless of configuration.

**Steps to reproduce**:

1. Run apm-server 9.2.0+ with tail-based sampling enabled and a configured `apm-server.sampling.tail.storage_limit` / disk threshold.
2. Curl the stats endpoint: `curl localhost:5066/stats | jq '.apm-server.sampling.tail.storage'`
3. Note that `disk_usage_threshold_pct` is missing, while sibling int64 fields (e.g. `lsm_size`, `value_log_size`) are present.

N/A — silent drop, no log output.

---

This was discovered while fixing the lazy-counter regression in #20993; the fix for this specific issue is included there (extend the translator to try `getScalarFloat64` when the int64 path fails). Filing this as a separate issue so the regression is tracked independently.

carsonip:
## Motivation/summary

apm-server's monitoring tree (the `/stats` HTTP endpoint and the libbeat monitoring snapshot) is the source we use to generate three sets of upstream field mappings:

- the metricbeat `beat` module's [`_meta/fields.yml`](https://github.com/elastic/beats/blob/main/metricbeat/module/beat/_meta/fields.yml) and [`stats/_meta/fields.yml`](https://github.com/elastic/beats/blob/main/metricbeat/module/beat/stats/_meta/fields.yml)
- the Elasticsearch stack-monitoring templates [`monitoring-beats.json`](https://github.com/elastic/elasticsearch/blob/main/x-pack/plugin/core/template-resources/src/main/resources/monitoring-beats.json) and [`monitoring-beats-mb.json`](https://github.com/elastic/elasticsearch/blob/main/x-pack/plugin/core/template-resources/src/main/resources/monitoring-beats-mb.json)
- the elastic_agent integration package's [`beat-stats-fields.yml`](https://github.com/elastic/integrations/blob/main/packages/elastic_agent/data_stream/apm_server_metrics/fields/beat-stats-fields.yml)

Mappings can include fields that aren't always emitted at runtime — the problem is the tooling we use to derive those mappings. The stats-to-mapping helper from the closed [#13638](https://github.com/elastic/apm-server/pull/13638) (and any equivalent script) reads a single `/stats` snapshot and infers field types from what's there. If a counter only appears in `/stats` after some specific code path has fired, it's not in the snapshot and silently gets dropped from the regenerated mappings.

After [#15094](https://github.com/elastic/apm-server/pull/15094) "Translate otel metrics to libbeat monitoring", that's exactly what happens: the OpenTelemetry instruments backing the new monitoring path are lazy. An `Int64Counter` only contributes a data point — and therefore only appears in `/stats` — once `.Add()` has been called on it. Counters whose code path hasn't fired yet (most error responses, OTLP variants we haven't seen traffic for, agent-config cache hits/misses, etc.) are absent from the snapshot. Pre-#15094, libbeat monitoring counters were registered up front and emitted as `0` by default, which gave `/stats` an exhaustive enumeration of every counter the server defines.

This PR restores that property under the OpenTelemetry path so the mapping-generation tooling can regenerate field mappings from a single `/stats` capture without needing to drive every error path.

### Eager zero-init helper

Adds a small `internal/otelmetric` package whose constructors (`NewInt64Counter`, `NewInt64UpDownCounter`) call `.Add(0)` at instrument creation, ensuring each instrument has at least one data point so the SDK exports it. Gauges and histograms are intentionally not in the helper: a synthetic `Record(0)` would falsely declare a gauge's current value as 0 and would inflate a histogram's observation count.

### Per-package eager registration

Each registration site that defines `apm-server.*` metrics is converted to use the helpers:

- `internal/agentcfg/elasticsearch.go` (6 counters under `apm-server.agentcfg.elasticsearch.*` plus the cache-entries gauge converted to an `Int64UpDownCounter` with delta tracking, since the cumulative sum of its deltas equals the current cache size)
- `internal/beater/api/intake/handler.go` (3 counters under `apm-server.processor.stream.*`)
- `internal/beater/beater.go` (`apm-server.sampling.transactions_dropped`)
- `internal/model/modelprocessor/eventcounter.go` (`apm-server.processor.<event-type>.transformations`)
- `x-pack/apm-server/sampling/processor.go` (6 counters under `apm-server.sampling.tail.events.*`)
- `x-pack/apm-server/sampling/groups.go` (`apm-server.sampling.tail.dynamic_service_groups`)

### HTTP middleware and gRPC interceptor

`MonitoringMiddleware` and `metricsInterceptor` produce counter names that depend on a `request.ResultID` chosen per-request. A new `request.AllResultIDs` slice is iterated at construction so every `(prefix, ResultID)` pair has its counter created up front. This is the largest single contributor — it covers `apm-server.acm.*`, `apm-server.root.*`, `apm-server.server.*`, `apm-server.otlp.{grpc,http}.*` error counters that previously only appeared after a matching error response.

The interceptor's and middleware's per-request counter storage was previously a `sync.Map` (added in #16182 to fix a concurrent-map race). Since every `(prefix, ResultID)` pair is now created in advance, the maps are never written after construction and `sync.Map`'s concurrent-mutation guarantees aren't needed. Both refactor to a plain map populated at startup and read-only at request time. The map keys avoid the per-call `prefix + string(id)` string concat (interceptor uses a `{prefix, ResultID}` struct key; middleware uses two `map[ResultID]Counter` since the prefix is fixed per-instance), so the hot path becomes allocation-free. A missing entry now logs `"BUG: monitoring counter missing from eager registration"` instead of silently creating a counter on first use — drift becomes runtime-loud.

The OTLP gRPC dispatch (three method names → three legacy prefixes) is consolidated into a single `otlpGRPCLegacyMetricsPrefixes` map keyed by gRPC method. The same map drives both per-call dispatch in `Interceptor()` and the eager-registration loop in `Metrics()`; adding a new OTLP signal is one map entry.

### Single source of truth for the canonical ResultID set

`internal/beater/request/result.go` previously declared the same set of ResultIDs in three places: a const block, the `MapResultIDToStatus` map, and a `ResultID`-slice literal in `AllResultIDs`. A new ID could land in the const block and any inc() call site without ever being added to `AllResultIDs`, silently breaking the eager-registration property.

The file now has a single `resultIDStatus` map keyed by ResultID; `init()` derives both `AllResultIDs` and `MapResultIDToStatus` from it. Adding a ResultID is one new entry; forgetting the entry surfaces immediately because the new ID isn't in `AllResultIDs`, the eager loop skips it, and the first inc() logs the BUG line which the strict tests fail on.

Two dead constants (`IDEventReceivedCount`, `IDEventDroppedCount` — leftovers from the Jaeger removal in #14791) are removed in the same change since they had no remaining callers.

### Float64 translator gap (fixes #20996)

`apm-server.sampling.tail.storage.disk_usage_threshold_pct` is a `Float64Gauge` introduced in #20464 (present in 9.2.0+) and has **never** appeared in `/stats`, even with TBS enabled and the storage manager actively recording values. `addAPMServerMetricsToMap` in `internal/beatcmd/beat.go` only handled `Sum[int64]` / `Gauge[int64]`; float64 instruments fell through the `else`. The translator is extended with a `getScalarFloat64` path so float-typed `apm-server.*` instruments are reported via `monitoring.ReportFloat`. The validation rule for "single dimensionless data point" is hoisted into a generic helper `scalarValue[T int64 | float64]` shared by both getters. `TestMonitoringApmServer` is extended with a `Float64Gauge` to lock the float branch in.

### `unsupported_dropped` guard removal

`apm-server.otlp.{grpc,http}.metrics.consumer.unsupported_dropped` was guarded by `if stats.UnsupportedMetricsDropped > 0` inside the observable-counter callback, so it never appeared until at least one unsupported metric was dropped. The guard is removed; the callback now always observes the (possibly zero) counter.

### libbeat.output / libbeat.pipeline lazy emission

The docappender → libbeat translation in `addDocappenderLibbeatOutputMetrics` and `addDocappenderLibbeatPipelineMetrics` (`internal/beatcmd/beat.go`) used to report each `libbeat.output.events.*` / `libbeat.output.write.bytes` / `libbeat.pipeline.events.total` field only inside the matching metric-name switch arm. Before docappender's first publish — i.e., at process start with ES output configured — none of the source metric names had data points, so eight fields were missing from `/stats`:

```
libbeat.output.events.{acked,active,batches,failed,toomany,total}
libbeat.output.write.bytes
libbeat.pipeline.events.total
```

Two earlier attempts to address this were rejected: an always-emit-zero adapter (10a8d1a4, reverted in 702bc903) fabricated zeros even when docappender stopped reporting an instrument, masking renames; an upstream `Add(0)` PR in [`elastic/go-docappender#317`](https://github.com/elastic/go-docappender/pull/317) (now closed) put the eager init at the right architectural layer but couldn't seed the three status-attributed fields without changing docappender's test contract for `events.processed`.

This PR converts the two `NewFunc` callbacks to use per-instance persistent state structs (`libbeatOutputState`, `libbeatPipelineState`) backed by `atomic.Bool` / `atomic.Int64` fields. The structs zero-initialise at process start, so every field surfaces in `/stats` from the first scrape, before any indexing has occurred. Each scrape updates only the fields docappender currently exports and reports the full set; a rename or drop of an upstream instrument leaves the field at its last-known value rather than snapping back to a fabricated zero. The `if writeBytes > 0` guard is removed in the same change, since `writeBytes` is now a struct field rather than a per-call local. `"type": "elasticsearch"` and the field emission are gated on having ever observed the docappender scope, so console-output runs keep an empty `libbeat.output` sub-tree as before.

Drift in docappender's instrument names is caught by `TestLibbeatMetrics`, which exercises real traffic and asserts specific cumulative values for every field — a rename fails that test loudly. `TestLibbeatMetricsEagerZero` locks the eager-zero contract directly: build a Beat with `output.elasticsearch`, create a docappender with no traffic, assert every field appears at zero in the snapshot.

### Walker performance

`addAPMServerMetricsToMap` walks each metric's dotted path. The original `strings.Split` allocated a `[]string` per call; replaced with `strings.SplitAfterSeq` which yields each segment with its trailing `.` preserved, and `strings.CutSuffix(seg, ".")` distinguishes intermediate (descend) from leaf (assign) without lookahead state. Benchstat over a 3-metric input:

```
                            │ split        │ splitafterseq                  │
                            │ sec/op       │ sec/op     vs base             │
AddAPMServerMetricsToMap-16   1.175µ ± 1%   1.078µ ± 2%  -8.22% (p=0.000)
                              2.492Ki ± 0%  2.305Ki ± 0% -7.52% (p=0.000) [B/op]
                              18.00 ± 0%    15.00 ± 0% -16.67% (p=0.000) [allocs]
```

### Interceptor and middleware performance

The sync.Map → plain map swap, combined with the struct/ID-keyed lookup, takes both hot paths to zero allocations:

```
                            │ syncmap       │ plain map                          │
                            │ sec/op        │ sec/op       vs base               │
Interceptor-16                738.1n ± 0%    405.2n ± 2%   -45.11% (p=0.000)
                              304.0 ± 0%     0.0 ± 0%     -100.00% (p=0.000) [B/op]
                              9.000 ± 0%     0.000 ± 0%   -100.00% (p=0.000) [allocs]

MonitoringMiddleware-16       906.6n ± 2%    432.2n ± 1%   -52.32% (p=0.000)
                              320.0 ± 0%     0.0 ± 0%     -100.00% (p=0.000) [B/op]
                              11.00 ± 0%     0.00 ± 0%    -100.00% (p=0.000) [allocs]
```

### Tests

Strict-match assertions stay strict. `monitoringtest/opentelemetry.go` is byte-identical to `main`. `TestMonitoringHandler` (middleware) and `TestMetrics` (interceptor) build the full expected map in-place — `prefix × request.AllResultIDs → int64(0)` plus the per-test non-zero overrides — before calling the existing `ExpectOtelMetrics`. Tests with an unbounded eager-init footprint (mux api tests, otlp http/grpc tests, eventcounter test) stay on `ExpectContainOtelMetrics` (subset semantics), which is what they used before #15094.

`TestMonitoringApmServer` extended with a `Float64Gauge` so the `getScalarFloat64` path is verified.

Two benchmarks added so future regressions are caught:

- `BenchmarkAddAPMServerMetricsToMap` in `internal/beatcmd/beat_test.go`.
- `BenchmarkInterceptor` in `internal/beater/interceptors/metrics_test.go`.
- `BenchmarkMonitoringMiddleware` in `internal/beater/middleware/monitoring_middleware_test.go`.

## How to test these changes

```shell
# Build and run with no traffic.
go build -o /tmp/apm-server ./x-pack/apm-server
/tmp/apm-server -e -c apm-server.yml &
sleep 4
curl -s http://127.0.0.1:5066/stats | jq '."apm-server"'
```

Before this change, `apm-server.*` shows only metrics that have actually been incremented (typically just the few `sampling.tail.storage.*` int64 gauges if TBS is enabled, plus whatever traffic has triggered). After this change, the full enumerated set (`acm`, `agentcfg`, `otlp`, `processor`, `root`, `sampling`, `server`) is present at zero — and `disk_usage_threshold_pct` shows up under `sampling.tail.storage` for the first time.

Unit tests:

```shell
go test -race ./internal/... ./x-pack/...
```

## Related issues

Unblocks the mapping-regeneration tooling in [#13638](https://github.com/elastic/apm-server/pull/13638) and the periodic upstream-mapping-sync work in [#15533](https://github.com/elastic/apm-server/issues/15533) and [#13475](https://github.com/elastic/apm-server/issues/13475).

Part of #20991.
Fixes #20996 as a side effect (float64 translator gap).

### DIFF
--- a/internal/beatcmd/beat.go
+++ b/internal/beatcmd/beat.go
@@ -577,39 +577,67 @@ func getScalarInt64(data metricdata.Aggregation) (int64, bool) {
 	return 0, false
 }
 
+// getScalarFloat64 returns a single-value, dimensionless gauge or counter
+// float value, or (0, false) if the data does not match these constraints.
+func getScalarFloat64(data metricdata.Aggregation) (float64, bool) {
+	switch data := data.(type) {
+	case metricdata.Sum[float64]:
+		if len(data.DataPoints) != 1 || data.DataPoints[0].Attributes.Len() != 0 {
+			break
+		}
+		return data.DataPoints[0].Value, true
+	case metricdata.Gauge[float64]:
+		if len(data.DataPoints) != 1 || data.DataPoints[0].Attributes.Len() != 0 {
+			break
+		}
+		return data.DataPoints[0].Value, true
+	}
+	return 0, false
+}
+
 // addAPMServerMetricsToMap adds simple scalar metrics with the "apm-server." prefix
-// to the map.
+// to the map. Both int64 and float64 instruments are supported.
 func addAPMServerMetricsToMap(beatsMetrics map[string]any, metrics []metricdata.Metrics) {
 	for _, m := range metrics {
-		if suffix, ok := strings.CutPrefix(m.Name, "apm-server."); ok {
-			if value, ok := getScalarInt64(m.Data); ok {
-				current := beatsMetrics
-				suffixSlice := strings.Split(suffix, ".")
-				for i := 0; i < len(suffixSlice)-1; i++ {
-					k := suffixSlice[i]
-					if _, ok := current[k]; !ok {
-						current[k] = make(map[string]any)
-					}
-					if currentmap, ok := current[k].(map[string]any); ok {
-						current = currentmap
-					}
-				}
-				current[suffixSlice[len(suffixSlice)-1]] = value
+		suffix, ok := strings.CutPrefix(m.Name, "apm-server.")
+		if !ok {
+			continue
+		}
+		var value any
+		if v, ok := getScalarInt64(m.Data); ok {
+			value = v
+		} else if v, ok := getScalarFloat64(m.Data); ok {
+			value = v
+		} else {
+			continue
+		}
+		current := beatsMetrics
+		suffixSlice := strings.Split(suffix, ".")
+		for i := 0; i < len(suffixSlice)-1; i++ {
+			k := suffixSlice[i]
+			if _, ok := current[k]; !ok {
+				current[k] = make(map[string]any)
+			}
+			if currentmap, ok := current[k].(map[string]any); ok {
+				current = currentmap
 			}
 		}
+		current[suffixSlice[len(suffixSlice)-1]] = value
 	}
 }
 
 func reportOnKey(v monitoring.Visitor, m map[string]any) {
 	for key, value := range m {
-		if valueMap, ok := value.(map[string]any); ok {
+		switch value := value.(type) {
+		case map[string]any:
 			v.OnRegistryStart()
 			v.OnKey(key)
-			reportOnKey(v, valueMap)
+			reportOnKey(v, value)
 			v.OnRegistryFinished()
-		}
-		if valueMetric, ok := value.(int64); ok {
-			monitoring.ReportInt(v, key, valueMetric)
+		case int64:
+			monitoring.ReportInt(v, key, value)
+		case float64:
+			monitoring.ReportFloat(v, key, value)
 		}
 	}
 }
--- a/internal/beatcmd/beat.go
+++ b/internal/beatcmd/beat.go
@@ -651,17 +651,17 @@ func reportOnKey(v monitoring.Visitor, m map[string]any) {
 	}
 }
 
-// addDocappenderLibbeatOutputMetrics adapts go-docappender's OTel metrics to
-// the libbeat.output.* monitoring tree. Every reported field is accumulated
-// into a local variable and reported unconditionally, so the full field set
-// appears with zero values even before docappender has published anything.
-// This keeps /stats's libbeat.output.* shape stable from process start, which
-// downstream mapping-regeneration tooling relies on.
+// Adapt go-docappender's OTel metrics to beats stack monitoring metrics,
+// with a mixture of libbeat-specific and apm-server specific metric names.
 func addDocappenderLibbeatOutputMetrics(ctx context.Context, v monitoring.Visitor, sm metricdata.ScopeMetrics) {
-	var acked, active, batches, failed, toomany, total, writeBytes int64
+	var writeBytes int64
+
+	v.OnRegistryStart()
+	v.OnKey("events")
 	for _, m := range sm.Metrics {
 		switch m.Name {
 		case "elasticsearch.events.processed":
+			var acked, toomany, failed int64
 			data, _ := m.Data.(metricdata.Sum[int64])
 			for _, dp := range data.DataPoints {
 				status, ok := dp.Attributes.Value(attribute.Key("status"))
@@ -678,13 +678,16 @@ func addDocappenderLibbeatOutputMetrics(ctx context.Context, v monitoring.Visito
 					failed += dp.Value
 				}
 			}
+			monitoring.ReportInt(v, "acked", acked)
+			monitoring.ReportInt(v, "failed", failed)
+			monitoring.ReportInt(v, "toomany", toomany)
 		case "elasticsearch.events.count":
 			if value, ok := getScalarInt64(m.Data); ok {
-				total = value
+				monitoring.ReportInt(v, "total", value)
 			}
 		case "elasticsearch.events.queued":
 			if value, ok := getScalarInt64(m.Data); ok {
-				active = value
+				monitoring.ReportInt(v, "active", value)
 			}
 		case "elasticsearch.flushed.bytes":
 			if value, ok := getScalarInt64(m.Data); ok {
@@ -692,45 +695,33 @@ func addDocappenderLibbeatOutputMetrics(ctx context.Context, v monitoring.Visito
 			}
 		case "elasticsearch.bulk_requests.count":
 			if value, ok := getScalarInt64(m.Data); ok {
-				batches = value
+				monitoring.ReportInt(v, "batches", value)
 			}
 		}
 	}
-
-	v.OnRegistryStart()
-	v.OnKey("events")
-	monitoring.ReportInt(v, "acked", acked)
-	monitoring.ReportInt(v, "active", active)
-	monitoring.ReportInt(v, "batches", batches)
-	monitoring.ReportInt(v, "failed", failed)
-	monitoring.ReportInt(v, "toomany", toomany)
-	monitoring.ReportInt(v, "total", total)
 	v.OnRegistryFinished()
 
-	v.OnRegistryStart()
-	v.OnKey("write")
-	monitoring.ReportInt(v, "bytes", writeBytes)
-	v.OnRegistryFinished()
+	if writeBytes > 0 {
+		v.OnRegistryStart()
+		v.OnKey("write")
+		monitoring.ReportInt(v, "bytes", writeBytes)
+		v.OnRegistryFinished()
+	}
 }
 
-// addDocappenderLibbeatPipelineMetrics adapts go-docappender's OTel metrics
-// to the libbeat.pipeline.* monitoring tree. Reports the field unconditionally
-// so it surfaces at zero before any publish, matching the eager-emission
-// contract used by addDocappenderLibbeatOutputMetrics.
 func addDocappenderLibbeatPipelineMetrics(ctx context.Context, v monitoring.Visitor, sm metricdata.ScopeMetrics) {
-	var total int64
-	for _, m := range sm.Metrics {
-		if m.Name == "elasticsearch.events.count" {
-			if value, ok := getScalarInt64(m.Data); ok {
-				total = value
-			}
-		}
-	}
-
 	v.OnRegistryStart()
 	defer v.OnRegistryFinished()
 	v.OnKey("events")
-	monitoring.ReportInt(v, "total", total)
+
+	for _, m := range sm.Metrics {
+		switch m.Name {
+		case "elasticsearch.events.count":
+			if value, ok := getScalarInt64(m.Data); ok {
+				monitoring.ReportInt(v, "total", value)
+			}
+		}
+	}
 }
 
 // Add non-libbeat Elasticsearch output metrics under "output.elasticsearch".
--- a/internal/beatcmd/beat.go
+++ b/internal/beatcmd/beat.go
@@ -558,21 +558,27 @@ func (b *Beat) registerStatsMetrics() {
 	})
 }
 
+// scalarValue extracts the single dimensionless data point from a Sum or
+// Gauge aggregation's DataPoints slice, or returns (zero, false) if the
+// data has the wrong shape: zero or many points, or attribute-tagged
+// points.
+func scalarValue[T int64 | float64](dp []metricdata.DataPoint[T]) (T, bool) {
+	if len(dp) != 1 || dp[0].Attributes.Len() != 0 {
+		var zero T
+		return zero, false
+	}
+	return dp[0].Value, true
+}
+
 // getScalarInt64 returns a single-value, dimensionless
 // gauge or counter integer value, or (0, false) if the
 // data does not match these constraints.
 func getScalarInt64(data metricdata.Aggregation) (int64, bool) {
 	switch data := data.(type) {
 	case metricdata.Sum[int64]:
-		if len(data.DataPoints) != 1 || data.DataPoints[0].Attributes.Len() != 0 {
-			break
-		}
-		return data.DataPoints[0].Value, true
+		return scalarValue(data.DataPoints)
 	case metricdata.Gauge[int64]:
-		if len(data.DataPoints) != 1 || data.DataPoints[0].Attributes.Len() != 0 {
-			break
-		}
-		return data.DataPoints[0].Value, true
+		return scalarValue(data.DataPoints)
 	}
 	return 0, false
 }
@@ -582,15 +588,9 @@ func getScalarInt64(data metricdata.Aggregation) (int64, bool) {
 func getScalarFloat64(data metricdata.Aggregation) (float64, bool) {
 	switch data := data.(type) {
 	case metricdata.Sum[float64]:
-		if len(data.DataPoints) != 1 || data.DataPoints[0].Attributes.Len() != 0 {
-			break
-		}
-		return data.DataPoints[0].Value, true
+		return scalarValue(data.DataPoints)
 	case metricdata.Gauge[float64]:
-		if len(data.DataPoints) != 1 || data.DataPoints[0].Attributes.Len() != 0 {
-			break
-		}
-		return data.DataPoints[0].Value, true
+		return scalarValue(data.DataPoints)
 	}
 	return 0, false
 }
--- a/internal/beatcmd/beat.go
+++ b/internal/beatcmd/beat.go
@@ -612,17 +612,22 @@ func addAPMServerMetricsToMap(beatsMetrics map[string]any, metrics []metricdata.
 			continue
 		}
 		current := beatsMetrics
-		suffixSlice := strings.Split(suffix, ".")
-		for i := 0; i < len(suffixSlice)-1; i++ {
-			k := suffixSlice[i]
-			if _, ok := current[k]; !ok {
-				current[k] = make(map[string]any)
+		// SplitAfterSeq yields each segment with its trailing "." preserved,
+		// except the last. CutSuffix tells us which we're looking at: hasDot
+		// true => intermediate (descend), false => leaf (assign).
+		for seg := range strings.SplitAfterSeq(suffix, ".") {
+			name, hasDot := strings.CutSuffix(seg, ".")
+			if !hasDot {
+				current[name] = value
+				continue
 			}
-			if currentmap, ok := current[k].(map[string]any); ok {
-				current = currentmap
+			if _, ok := current[name]; !ok {
+				current[name] = make(map[string]any)
+			}
+			if next, ok := current[name].(map[string]any); ok {
+				current = next
 			}
 		}
-		current[suffixSlice[len(suffixSlice)-1]] = value
 	}
 }
 
--- a/internal/beatcmd/beat.go
+++ b/internal/beatcmd/beat.go
@@ -701,12 +701,10 @@ func addDocappenderLibbeatOutputMetrics(ctx context.Context, v monitoring.Visito
 	}
 	v.OnRegistryFinished()
 
-	if writeBytes > 0 {
-		v.OnRegistryStart()
-		v.OnKey("write")
-		monitoring.ReportInt(v, "bytes", writeBytes)
-		v.OnRegistryFinished()
-	}
+	v.OnRegistryStart()
+	v.OnKey("write")
+	monitoring.ReportInt(v, "bytes", writeBytes)
+	v.OnRegistryFinished()
 }
 
 func addDocappenderLibbeatPipelineMetrics(ctx context.Context, v monitoring.Visitor, sm metricdata.ScopeMetrics) {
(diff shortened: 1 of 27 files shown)


internal/beatcmd/beat.go
```diff
@@ -577,39 +577,67 @@ func getScalarInt64(data metricdata.Aggregation) (int64, bool) {
...
+			break
+		}
+		return data.DataPoints[0].Value, true
+	}
+	return 0, false
+}
+
 // addAPMServerMetricsToMap adds simple scalar metrics with the "apm-server." prefix
-// to the map.
+// to the map. Both int64 and float64 instruments are supported.
```
carsonip, line 625: [to reviewer] this fixes https://github.com/elastic/apm-server/issues/20996 as a side effect

carsonip (commented):
i'll review the code again tomorrow, but it looks right to me at a high level. At the very least it is fixing a bug.

internal/beatcmd/beat.go
```diff
@@ -577,39 +577,67 @@ func getScalarInt64(data metricdata.Aggregation) (int64, bool) {
...
+		if v, ok := getScalarInt64(m.Data); ok {
+			value = v
+		} else if v, ok := getScalarFloat64(m.Data); ok {
+			value = v
+		} else {
+			continue
+		}
+		current := beatsMetrics
+		suffixSlice := strings.Split(suffix, ".")
+		for i := 0; i < len(suffixSlice)-1; i++ {
```
ericywl: q: why not `for _, k := range suffixSlice` since `i` is not used anywhere else?

internal/beater/interceptors/metrics.go
```diff
@@ -119,6 +119,15 @@ func (m *metricsInterceptor) getHistogram(n string, opts ...metric.Int64Histogra
 	return met.(metric.Int64Histogram)
 }
 
+// otlpLegacyMetricsPrefixes lists every legacy metric prefix the metrics
+// interceptor records under. Used to eagerly register counters at startup
+// so /stats enumerates each metric from process start.
+var otlpLegacyMetricsPrefixes = []string{
+	"apm-server.otlp.grpc.metrics.",
+	"apm-server.otlp.grpc.traces.",
+	"apm-server.otlp.grpc.logs.",
```
ericywl: Might be a good idea to have these prefixes as consts since it is also used in `Interceptor()` above.

ericywl (commented):
Looks good overall, just minor nits.

internal/beatcmd/beat.go
carsonip: now added a test to check

internal/beatcmd/beat.go
carsonip: thanks for calling it out. I figured it would be better to fix the loop altogether and use SplitAfterSeq instead, to reduce allocs. Added a benchmark for it

```
                              │   split.txt    │       splitafterseq.txt            │
                              │     sec/op     │   sec/op     vs base               │
  AddAPMServerMetricsToMap-16    1.175µ ± 1%    1.078µ ± 2%  -8.22%  (p=0.000 n=10)

                              │   split.txt    │       splitafterseq.txt            │
                              │      B/op      │     B/op      vs base              │
  AddAPMServerMetricsToMap-16    2.492Ki ± 0%   2.305Ki ± 0%  -7.52%  (p=0.000 n=10)

                              │   split.txt    │       splitafterseq.txt            │
                              │   allocs/op    │ allocs/op   vs base                │
  AddAPMServerMetricsToMap-16     18.00 ± 0%    15.00 ± 0%  -16.67%  (p=0.000 n=10)
```

internal/beatcmd/beat_test.go
```diff
@@ -285,17 +285,23 @@ func TestMonitoringApmServer(t *testing.T) {
...
+	// Three independently-scoped meters whose metrics share the
+	// "apm-server.sampling" prefix; the assertion below verifies they
+	// merge into one snapshot subtree.
+	storageMeter := b.meterProvider.Meter("github.com/elastic/apm-server/x-pack/apm-server/sampling/eventstorage")
+	lsmSizeGauge, _ := storageMeter.Int64Gauge("apm-server.sampling.tail.storage.lsm_size")
 	lsmSizeGauge.Record(context.Background(), 123)
+	// Float-typed instruments under apm-server.* must surface in the
+	// snapshot too, not only int64 ones.
+	diskThresholdGauge, _ := storageMeter.Float64Gauge("apm-server.sampling.tail.storage.disk_usage_threshold_pct")
+	diskThresholdGauge.Record(context.Background(), 0.8)
```
carsonip, line 360: [to reviewer] new test that catches missing float reporting

internal/beater/interceptors/metrics.go
carsonip: good call!

internal/beater/interceptors/metrics.go
```diff
@@ -28,38 +27,53 @@ import (
...
 )
 
-type metricsInterceptor struct {
-	logger *logp.Logger
-	meter  metric.Meter
+// otlpGRPCLegacyMetricsPrefixes maps each OTLP gRPC service method to its
+// legacy "apm-server.otlp.grpc.<signal>." metric prefix. Single source of
+// truth for both the per-call dispatch in Interceptor() and the eager
+// registration loop in Metrics(); adding a new signal means one entry.
+var otlpGRPCLegacyMetricsPrefixes = map[string]string{
```
carsonip, line 44: [to reviewer] Single source of truth for method name to metric prefix. A _slight_ overhead in map lookup compared to switch case string comparison but negligible in the grand scheme of things

internal/beater/interceptors/metrics.go
```diff
@@ -28,38 +27,53 @@ import (
...
+	prefix string
+	id     request.ResultID
+}
 
-	counters   sync.Map
-	histograms sync.Map
+type metricsInterceptor struct {
+	logger              *logp.Logger
+	counters            map[counterKey]metric.Int64Counter
+	requestDurationHist metric.Int64Histogram
```
carsonip, line 61: [to reviewer] sorry for the major refactor here. sync.Map was added in https://github.com/elastic/apm-server/pull/16182 to workaround a concurrent map access. Now that we change the paradigm to always create up front, we can cut the sync.Map overhead and benefit from metric appearing in `/stats` from registration time

with this change, see benchstat
```
                 │ /tmp/syncmap.txt │          /tmp/plainmap.txt          │                                                                                                                                           
                 │      sec/op      │   sec/op     vs base                │
  Interceptor-16        738.1n ± 0%   405.2n ± 2%  -45.11% (p=0.000 n=10)                                                                                                                                             
                                                                     
                 │ /tmp/syncmap.txt │         /tmp/plainmap.txt         │                                                                                                                                             
                 │       B/op       │   B/op    vs base                 │
  Interceptor-16         304.0 ± 0%   0.0 ± 0%  -100.00% (p=0.000 n=10)                                                                                                                                               
                                                                                                                                                                                                                      
                 │ /tmp/syncmap.txt │          /tmp/plainmap.txt          │
                 │    allocs/op     │ allocs/op   vs base                 │                                                                                                                                           
  Interceptor-16         9.000 ± 0%   0.000 ± 0%  -100.00% (p=0.000 n=10)    
```

internal/beater/interceptors/metrics.go
```diff
@@ -28,38 +27,53 @@ import (
...
+var otlpGRPCLegacyMetricsPrefixes = map[string]string{
+	"/opentelemetry.proto.collector.metrics.v1.MetricsService/Export": "apm-server.otlp.grpc.metrics.",
+	"/opentelemetry.proto.collector.trace.v1.TraceService/Export":     "apm-server.otlp.grpc.traces.",
+	"/opentelemetry.proto.collector.logs.v1.LogsService/Export":       "apm-server.otlp.grpc.logs.",
+}
+
+// counterKey is the map key for metricsInterceptor.counters. A struct key
+// is used instead of the concatenated "prefix+id" string so the lookup at
+// each m.inc() call avoids allocating a fresh string.
+type counterKey struct {
```
carsonip, line 53: [to reviewer] this cuts the string prefix and name concat alloc per `inc` call

internal/beatcmd/beat.go
```diff
@@ -558,58 +558,91 @@ func (b *Beat) registerStatsMetrics() {
...
+		} else if v, ok := getScalarFloat64(m.Data); ok {
+			value = v
+		} else {
+			continue
+		}
+		current := beatsMetrics
+		// SplitAfterSeq yields each segment with its trailing "." preserved,
+		// except the last. CutSuffix tells us which we're looking at: hasDot
+		// true => intermediate (descend), false => leaf (assign).
+		for seg := range strings.SplitAfterSeq(suffix, ".") {
```
carsonip, line 648: [to reviewer] strings.Split -> strings.SplitAfterSeq perf gain
```
                              │   split.txt    │       splitafterseq.txt            │
                              │     sec/op     │   sec/op     vs base               │
  AddAPMServerMetricsToMap-16    1.175µ ± 1%    1.078µ ± 2%  -8.22%  (p=0.000 n=10)

                              │   split.txt    │       splitafterseq.txt            │
                              │      B/op      │     B/op      vs base              │
  AddAPMServerMetricsToMap-16    2.492Ki ± 0%   2.305Ki ± 0%  -7.52%  (p=0.000 n=10)

                              │   split.txt    │       splitafterseq.txt            │
                              │   allocs/op    │ allocs/op   vs base                │
  AddAPMServerMetricsToMap-16     18.00 ± 0%    15.00 ± 0%  -16.67%  (p=0.000 n=10)
```

internal/beater/middleware/monitoring_middleware.go
```diff
@@ -62,41 +68,52 @@ func (m *monitoringMiddleware) Middleware() Middleware {
...
-func MonitoringMiddleware(legacyMetricsPrefix string, mp metric.MeterProvider) Middleware {
 	mid := &monitoringMiddleware{
-		meter:               mp.Meter("github.com/elastic/apm-server/internal/beater/middleware"),
+		logger:              logger,
 		legacyMetricsPrefix: legacyMetricsPrefix,
-		counters:            sync.Map{},
-		histograms:          sync.Map{},
+		serverCounters:      serverCounters,
+		legacyCounters:      legacyCounters,
+		requestDurationHist: requestDurationHist,
```
carsonip, line 121: [to reviewer] this gives us

```
                            │ syncmap         │ plain map                          │                                                                                                                                  
                            │ sec/op          │ sec/op     vs base                 │                                                                                                                                  
  MonitoringMiddleware-16     906.6n ± 2%      432.2n ± 1%  -52.32% (p=0.000 n=10)                                                                                                                                    
                              320.0 ± 0%       0.0 ± 0%    -100.00% (p=0.000 n=10) [B/op]
                              11.00 ± 0%       0.00 ± 0%   -100.00% (p=0.000 n=10) [allocs]          
```

internal/beater/request/result.go
```diff
@@ -36,10 +35,6 @@ const (
 	IDResponseErrorsCount ResultID = "response.errors.count"
 	// IDResponseValidCount identifies all successful responses
 	IDResponseValidCount ResultID = "response.valid.count"
-	// IDEventReceivedCount identifies amount of received events
-	IDEventReceivedCount ResultID = "event.received.count"
-	// IDEventDroppedCount identifies amount of dropped events
-	IDEventDroppedCount ResultID = "event.dropped.count"
```
carsonip, line 42: [to reviewer] according to Claude: IDEventReceivedCount / IDEventDroppedCount removed — dead consts since the Jaeger removal in #14791, no callers

internal/beater/request/result.go
```diff
@@ -78,28 +73,56 @@ const (
...
+// populated by init() from resultIDStatus above; tests and the
+// monitoring middleware/interceptor iterate it to eagerly register
+// every counter at startup. Treat as read-only after package init.
+var AllResultIDs []ResultID
+
+// MapResultIDToStatus is the subset of resultIDStatus with non-zero
+// Status (every ID c.Result.ID can be set to). Populated by init().
+var MapResultIDToStatus = map[ResultID]Status{}
+
+func init() {
```
carsonip, line 118: [to reviewer] sorry for major refactor here again. This is to allow eager registration to be aware of all the result IDs and make the code drift-proof.

internal/beatcmd/beat.go
```diff
@@ -558,58 +558,91 @@ func (b *Beat) registerStatsMetrics() {
...
+		if !ok {
+			continue
+		}
+		var value any
+		if v, ok := getScalarInt64(m.Data); ok {
+			value = v
+		} else if v, ok := getScalarFloat64(m.Data); ok {
+			value = v
+		} else {
+			continue
```
ericywl, line 642: Would it be a good idea to have some indication that these metrics are being dropped?

carsonip (commented):
TODO
- double check if `addDocappenderLibbeatOutputMetrics` and `addDocappenderLibbeatPipelineMetrics` are working

Issue #21000, opened 2026-04-28
monitoring: `/stats` field mappings drift across 3 upstream repos with no automated sync


carsonip:
## The drift problem

When apm-server adds, renames, or removes a metric exposed by `apm-server:5066/stats`, five files in three other repositories need a matching update so the metric flows correctly through the stack-monitoring pipeline:

| File | Repo | Purpose |
| --- | --- | --- |
| [`monitoring-beats.json`](https://github.com/elastic/elasticsearch/blob/main/x-pack/plugin/core/template-resources/src/main/resources/monitoring-beats.json) | `elastic/elasticsearch` | Stack-monitoring index template (legacy) |
| [`monitoring-beats-mb.json`](https://github.com/elastic/elasticsearch/blob/main/x-pack/plugin/core/template-resources/src/main/resources/monitoring-beats-mb.json) | `elastic/elasticsearch` | Stack-monitoring index template (metricbeat variant) |
| [`metricbeat/module/beat/_meta/fields.yml`](https://github.com/elastic/beats/blob/main/metricbeat/module/beat/_meta/fields.yml) | `elastic/beats` | Alias entries used by metricbeat |
| [`metricbeat/module/beat/stats/_meta/fields.yml`](https://github.com/elastic/beats/blob/main/metricbeat/module/beat/stats/_meta/fields.yml) | `elastic/beats` | Concrete-typed entries under `beat.stats.*` |
| [`elastic_agent/data_stream/apm_server_metrics/fields/beat-stats-fields.yml`](https://github.com/elastic/integrations/blob/main/packages/elastic_agent/data_stream/apm_server_metrics/fields/beat-stats-fields.yml) | `elastic/integrations` | EA integration package fields |

Today this is a manual three-PR dance. There is no CI that fails when the trees diverge, so drift accumulates silently and is only discovered when a stack-monitoring chart breaks for a customer or when someone happens to look. Concrete examples of drift that has already accumulated:

- `apm-server.server.response.errors.concurrency` was removed from the server in #1749 (2019). It is still present in `monitoring-beats*.json`.
- `apm-server.jaeger.*` was removed in #14791. Still present in `monitoring-beats*.json`.
- `apm-server.sampling.tail.storage.disk_usage_threshold_pct` was added in #20464 (9.2.0+). The five mapping files have it; `/stats` did not actually emit it until #20993 / #20996, because of an unrelated translator gap.
[...]

PR #21001, opened 2026-04-28
ci: weekly drift check between /stats and upstream mapping files

carsonip:
## Motivation/summary

Implements the scheduled drift-detection job proposed in #21000 — the part of that issue that catches drift between apm-server's `/stats` endpoint and the five upstream mapping files when nobody has manually run the regen tool. The active-developer side of the same issue is implemented in [elastic/apm-tools#245](https://github.com/elastic/apm-tools/pull/245); this PR is the CI side.

## How and when it runs

A single new workflow at `.github/workflows/stats-to-mapping-drift.yml`. Triggers:

- `schedule: cron: '0 6 * * MON'` — every Monday at 06:00 UTC.
- `workflow_dispatch` — on-demand from the Actions tab or `gh workflow run`.

Each run, in order:

1. Builds apm-server from the workflow's checked-out commit (`go build ./x-pack/apm-server`).
2. Installs the regen tool: `go install github.com/elastic/apm-tools/cmd/stats-to-mapping@latest`.
3. Starts apm-server with TBS enabled and captures `/stats` from the `http.enabled` endpoint, then stops apm-server. The captured `stats.json` is the source of truth for the comparison.
4. Sparse-clones the three upstream repos (`elastic/elasticsearch`, `elastic/beats`, `elastic/integrations`) with `--depth 1 --filter=blob:none` and `sparse-checkout`, so only the relevant subtrees materialise.
5. Runs the regen tool over the five tracked files, then `git diff`s each one to detect drift.

`UPSTREAM_FILES` (job-level env) is the single source of truth for the file list: the sparse-checkout, the regen invocation, the drift loop, and the recipe in the issue body all derive from it. Adding or removing a tracked file means editing one list.

The apm-server config used to capture `/stats` is written to a file in the CI run and then read back (with `path.data` rewritten) when generating the recipe, so the issue's repro recipe and the CI run cannot disagree about what config produced the captured `/stats`.

## What the run produces

If no file drifted, the workflow exits green with no other output.

If at least one file drifted:

- **Workflow artifact** `drift-diff` (30-day retention) — full `git diff` of every drifted file, plus the captured `stats.json`.
- **Issue** with labels `bug` + `stats-to-mapping-drift`, body containing:
  - One-line summary stating how many files drifted and linking the workflow run.
  - Per-file change summary (`<repo>/<path> — +N / -M`).
  - A copy-pasteable local reproduction recipe (apm-server build, /stats capture, three upstream clones, regen invocation).

The issue is opened only if no open issue with the `stats-to-mapping-drift` label already exists. A sustained drift event therefore produces one issue, not a weekly stream of duplicates.

## Permissions

- Workflow level: `contents: read`.
- `drift-check` job: `contents: read` plus `issues: write` (required only to file the drift issue). No write access to the codebase, no PR creation, no other scopes.

## Checklist

- [ ] Update [CHANGELOG.asciidoc](https://github.com/elastic/apm-server/blob/main/CHANGELOG.asciidoc) — n/a, internal CI.
- [ ] Documentation has been updated — workflow has a header comment that points at #21000.

## How to test these changes

After merge, trigger the workflow manually from the Actions tab (`workflow_dispatch`). With current upstream state and a /stats capture from `main`, the workflow is expected to detect drift and file one issue.

Locally, run the recipe written into a generated drift issue's body to obtain the same diffs.

## Related issues

- Implements the CI piece of #21000.
- Pairs with [elastic/apm-tools#245](https://github.com/elastic/apm-tools/pull/245). The cron's first run will fail at `go install ...stats-to-mapping@latest` until that PR is merged; the cron is not scheduled to fire until the next Monday after merge.
- Adjacent to #20993, which makes `/stats` an exhaustive enumeration of every counter apm-server defines. Without it, the workflow would compare an incomplete `/stats` against the upstream files and produce false positives.

### DIFF
--- a/.github/workflows/stats-to-mapping-drift.yml
+++ b/.github/workflows/stats-to-mapping-drift.yml
@@ -3,10 +3,11 @@ name: stats-to-mapping-drift
 
 # Weekly cron that detects drift between apm-server's /stats endpoint and the
 # upstream mapping files in elastic/elasticsearch, elastic/beats, and
-# elastic/integrations. When drift is detected and no open drift issue exists,
-# a new issue is opened with the per-file change summary, a reproduction
-# recipe, and a link to the workflow run (where the full diff is attached as
-# an artifact).
+# elastic/integrations. The orchestration lives in
+# script/stats-to-mapping-drift.sh (also runnable as `make
+# stats-to-mapping-drift`); this workflow wraps the script with
+# GitHub-Actions-specific concerns: artifact upload and drift-issue
+# creation.
 #
 # Background: elastic/apm-server#21000.
 
@@ -29,18 +30,6 @@ jobs:
     env:
       DRIFT_LABEL: stats-to-mapping-drift
       WORK: ${{ github.workspace }}/.drift
-      STATS_PORT: 15066
-      # Single source of truth for the files this job tracks. Each line is
-      # "<repo>/<path-within-repo>". The first slash-separated segment is
-      # treated as the github.com/elastic/<repo> name; the remainder is the
-      # file's path inside that repo. Sparse-checkout, the regen invocation,
-      # the drift loop, and the issue-body recipe all derive from this.
-      UPSTREAM_FILES: |
-        elasticsearch/x-pack/plugin/core/template-resources/src/main/resources/monitoring-beats.json
-        elasticsearch/x-pack/plugin/core/template-resources/src/main/resources/monitoring-beats-mb.json
-        beats/metricbeat/module/beat/_meta/fields.yml
-        beats/metricbeat/module/beat/stats/_meta/fields.yml
-        integrations/packages/elastic_agent/data_stream/apm_server_metrics/fields/beat-stats-fields.yml
     steps:
       - name: Checkout apm-server
         uses: actions/checkout@v6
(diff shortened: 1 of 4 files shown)


.github/workflows/stats-to-mapping-drift.yml
```diff
@@ -0,0 +1,259 @@
...
+      # the drift loop, and the issue-body recipe all derive from this.
+      UPSTREAM_FILES: |
+        elasticsearch/x-pack/plugin/core/template-resources/src/main/resources/monitoring-beats.json
+        elasticsearch/x-pack/plugin/core/template-resources/src/main/resources/monitoring-beats-mb.json
+        beats/metricbeat/module/beat/_meta/fields.yml
+        beats/metricbeat/module/beat/stats/_meta/fields.yml
+        integrations/packages/elastic_agent/data_stream/apm_server_metrics/fields/beat-stats-fields.yml
+    steps:
+      - name: Checkout apm-server
+        uses: actions/checkout@v4
```
v1v:
```suggestion
        uses: actions/checkout@v6
```

v1v (commented):
Wonder if those steps should be added to a specific Makefile so it can be easy tested locally?

mallendem (dismissed):

marclop (dismissed):
Nice work!

PR #21457, opened 2026-07-12
refactor: run go fix during fmt step

kruskall:
## Motivation/summary

similar approach used in elastic agent but use the stdlib tool instead of golangci lint

## Checklist


carsonip (dismissed):
thanks! Let's backport this to active branches so we don't run into merge conflicts in the future on backport.

For backporting this PR, I'd suggest we merge all right away except 9.5 until 9.5.0 release.

x-pack/apm-server/sampling/groups.go
```diff
@@ -238,11 +238,8 @@ func (g *traceGroup) finalizeSampledTraces(traceIDs []string, ingestRateDecayFac
...
 	// Resize the reservoir, so that it can hold the desired fraction of
 	// the observed ingest rate.
-	newReservoirSize := int(math.Ceil(g.samplingFraction * g.ingestRate))
-	if newReservoirSize < minReservoirSize {
-		newReservoirSize = minReservoirSize
-	}
+	newReservoirSize := max(math.Ceil(g.samplingFraction*g.ingestRate), minReservoirSize)
 	g.reservoir.Reset()
-	g.reservoir.Resize(newReservoirSize)
+	g.reservoir.Resize(int(newReservoirSize))
```
carsonip:
```suggestion
	newReservoirSize := max(int(math.Ceil(g.samplingFraction*g.ingestRate)), minReservoirSize)
	g.reservoir.Reset()
	g.reservoir.Resize(newReservoirSize)
```

nit: prefer having `int()` called sooner because it doesn't make sense for a variable named "newReservoirSize" to not be a whole number.

x-pack/apm-server/sampling/pubsub/position.go
```diff
@@ -27,9 +30,7 @@ func (p *SubscriberPosition) UnmarshalJSON(data []byte) error {
 
 func copyPosition(pos SubscriberPosition) SubscriberPosition {
 	observedSeqnos := make(map[string]int64, len(pos.observedSeqnos))
-	for index, seqno := range pos.observedSeqnos {
-		observedSeqnos[index] = seqno
-	}
+	maps.Copy(observedSeqnos, pos.observedSeqnos)
```
carsonip, line 33: nit: Maybe `maps.Clone` is even better here, but I assume there are some caveats depending on whether observedSeqnos is sparse

carsonip: well let's keep it as `maps.Copy` as we call `delete` on it and it can be sparse.

carsonip (dismissed):
thanks!

carsonip (dismissed):

carsonip (approved):

PR #21520, opened 2026-07-15
Add max sourcemap size configurable limit

isaacaflores2:
## Motivation/summary

There is no current safe guard against large source maps that are retrieved from elasticsearch. The goal of this PR is to add a configurable limit which prevents APM-Server from exhausting memory resources when it reads/decompresses a large source map.

### Kibana
The Kibana API has a max upload size of 1MB: https://www.elastic.co/docs/api/doc/kibana/v9/operation/operation-uploadsourcemap. 

The default limit of `500MiB` is meant to be generous to prevent any existing customers from experiencing issues (in case they have increased the default Kibana max payload) but also low enough to protect APM-Server. 

### Negative Caching

I added the negative caching to avoid repeated attempts to requests to Elasticsearch to read/decompress the large source map. 

The only downside I see to this approach is that the first time the large source map is read, we will emit a debug log with error: `decompressed source map exceeds the size limit error`. Subsequent attempts to fetch the source map will emit a debug log with a more opaque [error](https://github.com/elastic/apm-server/blob/042491dba8e055370fff8af8a8f0b94cda77c2b0/internal/sourcemap/sourcemap_fetcher.go#L103). 

## Checklist

- [x] Add steps to test/validate

   * Yes logging exists
   *  N/A
   * A test case to `TestRUMSpanSourcemapping` has been added

## How to test these changes

<details><summary>Testing instructions</summary>
<p>

Disclaimer the majority of the instructions were generated with CladeCode. 

# RUM Source Map — Manual Testing Guide

Use these steps to manually exercise source map uploading and RUM span
source-mapping against a locally running stack. The curl commands reuse the
same test data as `TestRUMSpanSourcemapping` in `systemtest/sourcemap_test.go`
so the results are predictable and match the automated system test.

## Prerequisites
Start a local apm-server that is configured to connect to a local Elastic stack. Minium config:
```yaml
apm-server:
  host: "127.0.0.1:8200"

  rum:
    enabled: true
    source_mapping:
      enabled: true
      #max_sourcemap_size: 1b

output.elasticsearch:
  hosts: [ "http://localhost:9200" ]
  enabled: true
  username: "elastic"
  password: "changeme"
```

Credentials — export before running any command so they aren't baked into
shell history:

```bash
export ES_USER=elastic
export ES_PASSWORD=<your-elastic-password>
```

> All `@testdata/...` paths below are relative — **run every command from the
> repo root**.

---

## Step 1 — Upload a source map via the Kibana API

```bash
curl -sS -u "$ES_USER:$ES_PASSWORD" \
  -X POST "http://localhost:5601/api/apm/sourcemaps" \
  -H "kbn-xsrf: 1" \
  -F "service_name=apm-agent-js" \
  -F "service_version=1.0.0" \
  -F "bundle_filepath=http://localhost:8000/test/e2e/general-usecase/bundle.js.map" \
  -F "sourcemap=@testdata/sourcemap/bundle.js.map;type=application/octet-stream"
```

**Expected response** (`200 OK`):

```json
{"id":"...","created":"...","body":{...}}
```

> The `service_name`, `service_version`, and `bundle_filepath` values above
> must match the metadata inside the RUM intake payload (step 2). If they
> don't match, source mapping will not be applied.

---

## Step 2 — Send a RUM span payload to trigger source mapping

```bash
curl -sS \
  -X POST "http://localhost:8200/intake/v2/rum/events?verbose=true" \
  -H "Content-Type: application/x-ndjson" \
  --data-binary "@testdata/intake-v2/transactions_spans_rum_2.ndjson"
```

**Expected response** (`202 Accepted`):

```json
{"accepted":2}
```

The intake payload (`transactions_spans_rum_2.ndjson`) contains:
- A metadata line declaring `service.name: apm-agent-js`, `service.version: 1.0.0`
- A transaction and a span with stacktrace frames whose `abs_path` points to
  `http://localhost:8000/test/e2e/general-usecase/bundle.js.map`

APM Server normalizes the path and looks up the source map uploaded in step 1.

---

## Step 3 — Verify source mapping was applied

Query Elasticsearch for the indexed span and confirm source mapping fields are
present. Note: APM Server syncs source map metadata in the background — wait a
few seconds after step 1 before sending the intake payload if mapping doesn't
appear immediately.

```bash
curl -sS -u "$ES_USER:$ES_PASSWORD" \
  "http://localhost:9200/traces-apm*/_search?q=processor.event:span&pretty" \
  | grep -A5 '"sourcemap"'
```

Look for `"updated": true` in each stacktrace frame:

```json
"stacktrace": [
  {
    "sourcemap": { "updated": true },
    "original": { "function": "<anonymous>", "lineno": 1, "colno": 18, "abs_path": "..." }
  }
]
```

---

## (Optional) Testing the `max_sourcemap_size` limit

The `max_sourcemap_size` setting (added on this branch) lets you cap how large
a decompressed source map APM Server will accept from Elasticsearch. To
exercise it:

1. Add the following to your `local-apm.yml` under `apm-server.rum.source_mapping`:

```yaml
apm-server:
  host: "127.0.0.1:8200"

  rum:
    enabled: true
    source_mapping:
      enabled: true
      max_sourcemap_size: 1b 
```

2. Restart APM Server, then re-run steps 1–3.

3. With the limit below the map size (~4 787 bytes for the test map), APM
   Server should log a size-exceeded error and leave stacktrace frames
   **without** `sourcemap.updated: true`. The span is still indexed but
   source mapping is skipped. Example log:
```json
{
   "log.level": "debug",
   "@timestamp": "2026-07-16T15:37:40.246-0700",
   "log.logger": "beater.handler.stacktrace",
   "log.origin": {
      "function": "github.com/elastic/apm-server/internal/sourcemap.BatchProcessor.processStacktraceFrame",
      "file.name": "sourcemap/processor.go",
      "file.line": 124
   },
   "message": "failed to fetch sourcemap with name (apm-agent-js), version (1.0.0), path (http://subdomain1.localhost:8000/test/e2e/general-usecase/bundle.js.map): unable to find sourcemap for service.name=apm-agent-js service.version=1.0.0 bundle.path=http://localhost:8000/test/e2e/general-usecase/bundle.js.map",
   "service.name": "apm-server",
   "ecs.version": "1.6.0"
}

```

4. Restore limit (e.g. `10mb`) or remove the setting to return to normal
   behavior.

</p>
</details>


internal/sourcemap/elasticsearch.go
```diff
@@ -38,10 +38,16 @@ import (
 	"github.com/elastic/apm-server/internal/logs"
 )
 
+const (
+	// defaultMaxSourceMapSizeBytes default max is 100MB.
+	defaultMaxSourceMapSizeBytes = 100 * 1024 * 1024
```
carsonip: nit: duplicate source of truth. If this is only needed for tests, move into test files

internal/beater/config/rum.go
```diff
@@ -104,23 +119,53 @@ func (c *RumConfig) setup(log *logp.Logger, outputESCfg *config.C) error {
...
 	}
 	return nil
 }
 
+// parseUserDefinedBytesConfig parses the provided size configuration to bytes.
+// The default is used for empty values.
+func parseUserDefinedBytesConfig(configValue string, defaultValue uint64) (uint64, error) {
+	if configValue == "" {
+		return defaultValue, nil
+	}
```
carsonip, line 150: nit: I'm inclined to fail fast for empty value handling to be consistent with TailSamplingConfig

internal/beater/config/rum.go
```diff
@@ -104,23 +119,53 @@ func (c *RumConfig) setup(log *logp.Logger, outputESCfg *config.C) error {
...
 func defaultSourcemapping() SourceMapping {
 	return SourceMapping{
-		Enabled:  true,
-		ESConfig: elasticsearch.DefaultConfig(),
-		Timeout:  defaultSourcemapTimeout,
+		Enabled:                true,
+		ESConfig:               elasticsearch.DefaultConfig(),
+		Timeout:                defaultSourcemapTimeout,
+		MaxSourceMapSizeParsed: defaultMaxSourceMapSizeBytes,
+		// MaxSourcemapSize left empty; the const-derived parsed default above is authoritative.
```
carsonip: nit: having an already parsed MaxSourcemapSize would maintain the assumption that defaultSourcemapping is readily usable, like defaultTailSamplingConfig: https://github.com/elastic/apm-server/blob/862167ba26b30a526b8ac9fd58d35beff2512310/internal/beater/config/sampling.go#L169

internal/sourcemap/elasticsearch.go
```diff
@@ -113,11 +127,19 @@ func (s *esFetcher) Fetch(ctx context.Context, name, version, path string) (*sou
 	}
 	defer r.Close()
 
-	uncompressedBody, err := io.ReadAll(r)
+	limitReader := io.LimitReader(r, s.maxSourceMapSizeBytes+1)
```
carsonip: quite clever to read N+1 bytes and check if it exceeds N bytes.

nit: Do you mind adding a short comment just in case anyone thought the +1 was meaningless?

carsonip (commented):
thanks for the thorough work! The default you introduced here is fair IMO, and thanks for adding an escape hatch in yaml. Most nits I left won't make any difference in practice, but being consistent with the other config handling gives me a peace of mind because all the Unpack magic makes me nervous. IIRC from my digging a few years ago, `a.Unpack(b)` sometimes means parsing b into a and sometimes the opposite 🤯

carsonip (commented):
added backport-active-all. lmk if you disagree

internal/sourcemap/elasticsearch.go
isaacaflores2: I moved this const to a test file. It was used as a default in case the construction received a zero limit value. I updated the elasticsearch `Fetch` method to only use a LimitReader when the max size is valid. This should not happen in practice since the config layer provides a default

internal/beater/config/rum.go
isaacaflores2: Updated to match the tail sampling config

internal/sourcemap/elasticsearch.go
```diff
@@ -142,3 +142,44 @@ func parse(body io.ReadCloser, name, version, path string, logger *logp.Logger)
...
+
+	decompress := func(r io.Reader) ([]byte, error) {
+		uncompressedBody, err := io.ReadAll(r)
+		if err != nil {
+			return nil, fmt.Errorf("failed to read sourcemap content: %w", err)
+		}
+		return uncompressedBody, nil
+	}
+
+	if s.maxSourceMapSizeBytes <= 0 {
```
isaacaflores2, line 161: note: this should never happen since the config applies a default. Added this a safeguard to prevent using a limit reader which reads 1B.

carsonip (dismissed):
lgtm thanks! On backports, it will likely catch a 9.5.0 BC but miss other active branch BCs. I won't worry too much about it.

carsonip (approved):

carsonip:

Change works as intended.

There's a slight risk that the user is unaware of this change, since any sourcemap error is only logged in debug, but we can tackle that in a follow up.

isaacaflores2:
> ✔️ test-plan-ok
> 
> Change works as intended.
> 
> There's a slight risk that the user is unaware of this change, since any sourcemap error is only logged in debug, but we can tackle that in a follow up.

@carsonip let me know if you think we should update the sourcemap processor to use an error log instead.

carsonip:
> let me know if you think we should update the sourcemap processor to use an error log instead.

The current log location is on Fetch, so it is possible an error log there will spam the logs when there are a lot of fetches for the same sourcemap. I _think_ it makes sense to log a warn during sourcemap download if that happens (as opposed to the later fetch), but it also depends on whether that will churn through the cache and spam logs.

isaacaflores2:
Apologies for the late reply @carsonip. Here is the follow up PR to add a warn log: https://github.com/elastic/apm-server/pull/21582

PR #21597, opened 2026-08-01
fix: Implement path validation and sanitization functions

odaysec:
## Checklist

### DIFF
--- a/cmd/intake-receiver/main.go
+++ b/cmd/intake-receiver/main.go
@@ -35,6 +35,7 @@ import (
 	"os"
 	"os/signal"
 	"path/filepath"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


ekuboo100 (approved):

odaysec (commented):
Summary description (detailed)
- Problem: agent file paths were constructed directly from agent name and version metadata. Malicious or malformed metadata (for containing "../" or absolute paths) could cause the server to read or write files outside the intended directory.
- Fix: add sanitizePathComponent to normalize each path segment (allow only alphanumerics, dot, underscore, hyphen; replace other chars with '_', return "unknown" for empty results). Add safeAgentPath which uses filepath.Abs + filepath.Join and filepath.Rel to resolve the full path and confirm the resolved path does not escape the configured base directory (rejecting absolute/parent-traversal attempts). Replace direct concatenation in agentFileName with sanitized components.
- Impact: prevents directory traversal and unsafe file access from crafted agent metadata. No behavioral changes for normal metadata; malformed metadata now yields a validated error instead of potentially accessing unexpected files.

kruskall:
The way to do this is to use go dirfs instead of rolling your own sanitizer.

> Accessing files using paths constructed from user-controlled data can allow an attacker to access unexpected resources. This can result in sensitive information being revealed or deleted, or an attacker being able to influence behavior by modifying unexpected files.

This doesn’t really make sense. intake-receiver is a developer tool, it runs locally, and the folder argument is provided directly to the application. There isn’t an external attacker here, you’d effectively be performing path traversal against yourself.

Issue #21375, opened 2026-07-07
test: panic after TestPublisherStopShutdownInactive


carsonip:
```
=== RUN   TestPublisherStop
    logger.go:146: 2026-07-07T10:12:48.194Z	DEBUG	beat.publish	start pipeline event consumer
    logger.go:146: 2026-07-07T10:12:48.194Z	DEBUG	beat.publish	pipeline event consumer queue reader: start
    logger.go:146: 2026-07-07T10:12:49.297Z	DEBUG	beat.publish.memqueue	ackloop: return ack to broker loop:3
    logger.go:146: 2026-07-07T10:12:49.297Z	DEBUG	beat.publish.memqueue	ackloop:  done send ack
    logger.go:146: 2026-07-07T10:12:49.297Z	DEBUG	beat.publish.memqueue	ackloop: return ack to broker loop:16
    logger.go:146: 2026-07-07T10:12:49.297Z	DEBUG	beat.publish.memqueue	ackloop:  done send ack
    logger.go:146: 2026-07-07T10:12:49.297Z	DEBUG	beat.publish.memqueue	ackloop: return ack to broker loop:16
    logger.go:146: 2026-07-07T10:12:49.297Z	DEBUG	beat.publish.memqueue	ackloop:  done send ack
    logger.go:146: 2026-07-07T10:12:49.297Z	DEBUG	beat.publish.memqueue	ackloop: return ack to broker loop:1
    logger.go:146: 2026-07-07T10:12:49.297Z	DEBUG	beat.publish.memqueue	ackloop:  done send ack
    logger.go:146: 2026-07-07T10:12:49.298Z	DEBUG	beat.publish.memqueue	ackloop: return ack to broker loop:2
    logger.go:146: 2026-07-07T10:12:49.298Z	DEBUG	beat.publish.memqueue	ackloop:  done send ack
    logger.go:146: 2026-07-07T10:12:49.298Z	DEBUG	beat.publish.memqueue	ackloop: return ack to broker loop:1
    logger.go:146: 2026-07-07T10:12:49.298Z	DEBUG	beat.publish.memqueue	ackloop:  done send ack
    logger.go:146: 2026-07-07T10:12:49.298Z	DEBUG	beat.publish	client: close queue producer
    logger.go:146: 2026-07-07T10:12:49.298Z	DEBUG	beat.publish	client: done producer close
    logger.go:146: 2026-07-07T10:12:49.298Z	DEBUG	beat.publish	close pipeline
[...]

PR #21640, opened 2026-08-07
avoid panics in TestPublisherStopShutdownInactive by using in memory logs

aelnahas:
## Motivation/summary

TestPublisherStopShutdownInactive is flaky due to a potential for a routine to write to the logs after the test function has returned. The shutdown sequence used in the test cleanup:
https://github.com/elastic/beats/blob/50ef57e75fbd5b2923a308da39a27d1ce24185c3/libbeat/publisher/pipeline/consumer.go#L237 might happen before the queueReader had a chance to log its stop message. Therefore, the cleanup might exit too early. This commit bypasses this problem by using logp.NewTestingLoggerWithObserver which still returns a logging output using *testing.T but an in memory slice of the logs to bypass the cleanup panics.

Reference for ogp.NewTestingLoggerWithObserver: https://github.com/elastic/elastic-agent-libs/blob/main/logp/logptest/logptest.go#L46

## Checklist

## How to test these changes
I used the following steps to reproduce the problem and also check if the fix works

1. Clone beats and add the replace directive + go mod tidy
2. Add the delay to queueReader after the "stop" log: https://github.com/elastic/beats/blob/50ef57e75fbd5b2923a308da39a27d1ce24185c3/libbeat/publisher/pipeline/queue_reader.go#L76
3. In newTestLogger, swap the body to return logptest.NewTestingLogger(t, selector)
4. Run go test -count=2 ./internal/publish/... and observe the panic
5. Revert step 3 and re-run to confirm it's fixed

## Related issues
https://github.com/elastic/apm-server/issues/21375

### DIFF
--- a/internal/publish/pub_test.go
+++ b/internal/publish/pub_test.go
@@ -19,6 +19,7 @@ package publish_test
 
 import (
 	"context"
+	"fmt"
 	"testing"
 	"time"
 
@@ -30,6 +31,7 @@ import (
 	"github.com/elastic/beats/v7/libbeat/publisher"
 	"github.com/elastic/beats/v7/libbeat/publisher/pipeline"
 	"github.com/elastic/elastic-agent-libs/config"
+	"github.com/elastic/elastic-agent-libs/logp"
 	"github.com/elastic/elastic-agent-libs/logp/logptest"
 	"github.com/elastic/elastic-agent-libs/mapstr"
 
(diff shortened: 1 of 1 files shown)


carsonip (commented):
> given that this is likely caused by a regression from beats, could you gauge how much more effort it is to address the root cause (or are we using the pipeline correctly in our usage), before we jump to a workaround?

carsonip:
Superseded by https://github.com/elastic/apm-server/pull/21806 with simpler workaround

PR #21781, opened 2026-09-04
[8.19] Fix DRA annotate link to point at the published summary, not the temp path

ninalee12:
## What this changes

Same fix as #21780 (main), applied to `8.19` since the original DRA-prep migration backport (already merged) carried the same bug.

`dra-annotate.sh` linked the DRA build summary at `dra-prep`'s temp GCS upload path (`dra-builds/<pipeline>/<build_number>/<build_id>/summary-<version>.html`). That path never has a `summary-*.html` — `dractl prep` only writes the manifest there. The rendered summary only exists once `unified-release-dra-processing` publishes it to the final location (`<prefix>/<build_id>/summary-<version>.html`), so the annotation link 404s permanently.

Fix: read the `prefix` field already present in the manifest and link the final published path instead.

## Testing

Verified against a real production annotation from cloudbeat (same reference implementation, same bug): the temp-path URL 404s, the final-path URL (same shape this PR produces) returns 200 and renders the real summary page.

### DIFF
--- a/.buildkite/scripts/dra-annotate.sh
+++ b/.buildkite/scripts/dra-annotate.sh
@@ -4,11 +4,9 @@
 ##  extracts build_id and version, and annotates the build with a link to
 ##  the workflow's published summary.
 ##
-##  dractl prep never writes a summary-*.html at its temp upload path (only
-##  the manifest) — the rendered summary only exists once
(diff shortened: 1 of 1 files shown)


ninalee12:
closing as duplicate to https://github.com/elastic/apm-server/pull/21780

